PR #69635, opened 2026-07-09
Cancel Snowflake queries when a user kills the deferred task

steveahnahn:
### Problem

- `SnowflakeSqlApiOperator.on_kill` cancels the running queries, but after `defer()` the worker is gone and it can never run.
- `SnowflakeSqlApiTrigger` had no `on_kill` (base default is a no op), so killing a deferred task leaves the statements executing on the warehouse, burning credits.
- Sibling triggers already cancel on user kill: EMR in the Amazon provider, BigQuery/Dataproc/Dataflow in the Google provider.

### Change

- Add `on_kill` to `SnowflakeSqlApiTrigger`: cancels the running query ids.
- The SQL API cancel is a blocking `POST` with no async variant, so `on_kill` runs `cancel_queries` through `sync_to_async`, building the hook inside that worker so no connection work touches the event loop.
- `cancel_on_kill` flag (default `True`) on operator and trigger; operator threads it into the trigger at the defer site.
- Best effort: cancel failures are logged and swallowed. Fires only on user kill, never on triggerer restart, redistribution, or timeout.

### Live verification

1. Deferrable operator submitted a statement; task `deferred`, trigger polling (`GET /api/v2/statements/{id}` returning 202).

deferred task, trigger polling the statement

2. Marked the task failed from the UI. Task log:

```
Trigger cancelled by user action, invoking on_kill
Cancelling Snowflake query ids ['069b68f4-8e5e-404a-aef9-0eca86eb396d']
Snowflake query ids ['069b68f4-8e5e-404a-aef9-0eca86eb396d'] cancelled
```

task failed, on_kill cancelling the Snowflake query

3. The endpoint received `POST /api/v2/statements/{id}/cancel` and the statement transitioned to `ABORTED`. Without the change it stays running.

### Tests

- Trigger: cancel on kill, disabled/empty id no ops, error swallowing, serialization of the flag.
- Operator: `cancel_on_kill=False` guard, flag threaded into the trigger on defer.
- Every new assertion fails without the source change.

---


potiuk (commented):
Nice change — this closes a real gap, and threading `cancel_on_kill` through the defer site so the trigger inherits the operator's setting is the right shape.

I verified locally: the 57 trigger + operator tests pass, the `async def on_kill` signature matches `BaseTrigger.on_kill` (the triggerer awaits it, so a sync override would have silently never run), and the positional `SnowflakeSqlApiHook(conn_id, token_life_time, token_renewal_delta)` args match the constructor. Building the hook *inside* the `sync_to_async` worker rather than on the event loop is the correct call. Serialization round-trip coverage and the live-verification screenshots are appreciated.

One thing to fix and two I'd like your read on — left inline. Only the `asgiref` one is blocking.

---

providers/snowflake/src/airflow/providers/snowflake/triggers/snowflake_trigger.py
```diff
@@ -20,6 +20,8 @@
 from collections.abc import AsyncIterator
 from typing import TYPE_CHECKING, Any
 
+from asgiref.sync import sync_to_async
```
potiuk, line 23: This is the first and only `asgiref` usage in the Snowflake provider, and `asgiref` isn't in this provider's `dependencies` in `providers/snowflake/pyproject.toml`. It resolves today only transitively via `apache-airflow`, but providers should declare what they import.

Amazon and Google both do exactly this where they use `sync_to_async`:

```toml
"asgiref>=2.3.0; python_version < '3.14'",
"asgiref>=3.11.1; python_version >= '3.14'",
```

Worth carrying over the `>=3.11.1` floor for 3.14 rather than a bare pin.

---

providers/snowflake/src/airflow/providers/snowflake/triggers/snowflake_trigger.py
```diff
@@ -93,6 +101,27 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
...
+                "Failed to cancel Snowflake query ids %s. They may still be running.", self.query_ids
+            )
+
+    def _cancel_queries(self) -> None:
+        hook = SnowflakeSqlApiHook(
+            self.snowflake_conn_id,
+            self.token_life_time,
+            self.token_renewal_delta,
+        )
+        hook.cancel_queries(self.query_ids)
```
potiuk: `SnowflakeSqlApiHook.cancel_queries` loops `_cancel_sql_api_query_execution` per id with no per-query error handling, and `on_kill` catches only at the outermost level — so the first id that raises aborts cancellation of every id after it.

That matters most in exactly the case this PR targets: `run()` polls statements sequentially and never prunes `self.query_ids`, so on a multi-statement operator the earlier statements have typically already completed by kill time. If cancelling a completed statement raises, the still-running later statements — the ones actually burning credits — never get cancelled.

Mitigating factor: `_process_response` returns a dict for HTTP 422 rather than raising, so if Snowflake answers 422 for "not running" the loop survives fine. But a 404 on a reaped handle, or a 401 on token expiry, would raise.

Did your live test cover killing a *multi-statement* task where some statements had already finished? If 422 is what Snowflake actually returns there, this is a non-issue and worth a one-line comment saying so. If not, a per-query `try`/`except` in `_cancel_queries` would make it genuinely best-effort. I didn't want to assert a bug I couldn't reproduce without live credentials.

---

providers/snowflake/src/airflow/providers/snowflake/triggers/snowflake_trigger.py
```diff
@@ -93,6 +101,27 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
         except Exception as e:
             yield TriggerEvent({"status": "error", "message": str(e)})
 
+    async def on_kill(self) -> None:
+        """Cancel the running Snowflake queries when the user kills the deferred task."""
+        if not self.cancel_on_kill or not self.query_ids:
+            return
+        self.log.info("Cancelling Snowflake query ids %s", self.query_ids)
+        try:
+            await sync_to_async(self._cancel_queries)()
```
potiuk, line 116: Non-blocking, more of a docs note: the triggerer wraps this in `asyncio.wait_for(trigger.on_kill(), timeout=_ON_CANCEL_TIMEOUT)` — `[triggerer] on_kill_timeout`, default 30s.

Cancellation here is N sequential blocking POSTs, each with tenacity retries on 429/503/504. A task with many statements against a slow warehouse could exceed that budget, and because `sync_to_async` runs in a thread the remaining cancels can't be interrupted — you'd get an `on_kill() timed out` warning while the thread keeps going.

Benign in practice, but the docstring's "best effort" framing might be worth extending to mention the timeout so operators know the knob exists.

---

steveahnahn: Good call, extended the on_kill docstring to name the knob: cancellation issues one blocking request per query id, so a task with many statements against a slow warehouse can exceed on_kill_timeout and statements not cancelled within that budget may keep running

providers/snowflake/src/airflow/providers/snowflake/triggers/snowflake_trigger.py
steveahnahn: No, my live check did not cover that case.

The SQL API reference does document the cancel endpoint returning 422 (QueryFailureStatus, code 000709 "not found") rather than 404. That matches the hook's _should_raise_for_status (>= 400 and != 422), so the loop would survive. The token-expiry case is also mostly moot here, since _cancel_queries builds a fresh hook and mints a fresh JWT at cancel time rather than reusing the poll-time token.

Since that is documented rather than live-verified, I did not want to rely on it. _cancel_queries now cancels per id inside a try/except, so any single failure cannot stand the still-running statements after it in the list.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowflake_trigger.py
steveahnahn: Done. Added to providers/snowflake/pyproject.toml, matching the amazon/google shape

potiuk (approved):
All three points from my last review are addressed, and addressed properly.

The `asgiref` declaration matches the amazon/google shape. The per-id `try/except` in `_cancel_queries` is the fix I was after, and `test_on_kill_cancels_remaining_after_one_fails` pins it by asserting all three ids are still attempted after `q1` raises — that is the case that actually matters, since the completed early statements are the ones most likely to error and the still-running later ones are the ones burning credits. The `on_kill` docstring now names `[triggerer] on_kill_timeout` and its default.

I appreciated the honesty in your replies — flagging that the 422-versus-404 behaviour was documented rather than live-verified, and not wanting to rely on it, is what produced the per-id loop rather than a debate about whether the loop was needed.

I checked the one claim that could have overreached: the operator docstring says `cancel_on_kill` applies both on the worker and in the triggerer, and it does — `SnowflakeSqlApiOperator.on_kill` gained the early return and `test_snowflake_sql_api_on_kill_respects_cancel_on_kill_false` covers it. Both paths honour the flag.

Two notes, neither blocking:

- `cancel_on_kill=True` changes behaviour for existing deferred tasks, which previously left queries running when killed. I think that is right — the non-deferred `on_kill` already cancelled unconditionally, so the deferred path was the inconsistent one — but it is worth being a deliberate choice rather than an incidental one.
- No newsfragment is needed; provider changelogs are regenerated from `git log`.

### The branch needs a rebase

It currently conflicts with `main` — `uv.lock` moved underneath it via #72091. For that particular conflict do not hand-merge: delete `uv.lock` and regenerate it with `uv lock`, which is the documented remedy and avoids the marker-line churn a manual resolution produces.

Approving on the code; this needs the rebase to be green before it can go in.

---

PR #69557, opened 2026-07-07
Scheduler: extend HA advisory lock coordination to CockroachDB

viragtripathi:
The scheduler's critical section takes a transaction-scoped advisory lock (`pg_try_advisory_xact_lock`) so HA schedulers coordinate instead of colliding on the pool row locks. The guard in `executable_task_instances_to_queued` currently limits this to the postgresql dialect. CockroachDB v26.3 implements the same transaction-level advisory lock functions with PostgreSQL semantics, so this extends the guard to the cockroachdb dialect. Other dialects are unaffected.

Why the advisory lock rather than relying on the existing pool row locks: the row-lock path discovers scheduler contention by starting the critical section and failing partway through. On PostgreSQL that produces avoidable lock-timeout errors in the server log, which is the original motivation for this fast path. On engines that detect conflicts
under `SERIALIZABLE` at commit time, the losing scheduler instead pays for a full critical-section attempt that then rolls back as a serialization conflict. The advisory lock turns both into one cheap function call, and the losing scheduler simply skips the critical section for that loop and does other useful work.

Minimum version for the new path: CockroachDB v26.3. Verified in an HA deployment (two schedulers) where lock acquisition and single-holder behavior were observed via `pg_locks` during a concurrent stress run.

Discussed on the devlist:
https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes - Claude Code

Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @viragtripathi** · by `@potiuk` · 2026-07-11 14:59 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed (see our [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria)):
> - :x: **Pre-commit / static checks**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #69556, opened 2026-07-07
Scheduler: survive serialization conflicts in the critical section

viragtripathi:
Under SERIALIZABLE isolation PostgreSQL can abort a transaction with SQLSTATE 40001 (serialization_failure), and deadlocks surface as 40P01. In the scheduler's critical section either can be raised while queuing task instances or at COMMIT time. Today that exception escapes _do_scheduling and crashes the scheduler job even though the transaction
is safe to retry, while the sibling case of a busy advisory lock (55P03) is already handled gracefully.

This handles serialization failures the same way: roll back, increment a new scheduler.critical_section_conflict metric, and let the next scheduler loop retry. Two details worth reviewer attention:

- The critical section buffers workloads into the executors' in-memory queued_tasks dict
  before the commit, so the handler discards exactly the workloads buffered in the failed
  round. Dispatching them after rollback would trip the execution API's requirement that a
  task instance be in QUEUED state before it can start.
- guard.commit() moves inside the handled scope, because serialization failures frequently
  surface at commit time rather than at statement time.

Behavior change for PostgreSQL deployments running SERIALIZABLE: a serialization failure or deadlock in the critical section no longer crashes the scheduler; it is counted and retried on the next loop, mirroring the existing 55P03 handling. MySQL deadlock codes (errno 1213) are out of scope here and could be a follow-up if there is interest.

Note that lowering the isolation level is not a workaround: even under READ COMMITTED, PostgreSQL-compatible engines can surface 40001 to the client once their automatic per-statement retries are exhausted or a partial result set has already been streamed, so the scheduler needs to handle it regardless of the configured isolation level. In HA
testing this change handled serialization conflicts observed under both SERIALIZABLE and READ COMMITTED.

Verified in an HA deployment (two schedulers, 240 concurrent task instances, on a PostgreSQL-compatible database running SERIALIZABLE): serialization conflicts were observed and survived, and both schedulers completed with zero restarts. Unit tests cover the new helper for psycopg2 and psycopg3 error shapes, the 40001 and 40P01 handler paths,
a 55P03 regression case, and the buffered-workload purge.

Discussed on the devlist:
https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes - Claude Code

Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

[...]

PR #69555, opened 2026-07-07
Make task_instance UUID migration work without pgcrypto

viragtripathi:
Migration 0042 adds a UUID primary key to the task_instance table. Its PostgreSQL branch
defines a PL/pgSQL uuid_generate_v7() helper and drops the old primary key before
creating the new one. Neither works on the cockroachdb dialect: PL/pgSQL function bodies
fail there with SQLSTATE 0A000 (see cockroachdb/cockroach#110080), regardless of the
helper's built-in pgcrypto fallback, and the engine refuses to drop a primary key without
a replacement in the same statement, so `airflow db migrate --use-migration-files` cannot
complete on that dialect.

This adds a cockroachdb branch that assembles the UUIDv7 hex layout in a plain SQL
function (the same technique the MySQL branch uses) and backfills entirely server side
with batched UPDATE ... LIMIT statements: no client-side generation and no per-row round
trips. The primary key is then swapped atomically with ALTER TABLE ... ALTER PRIMARY KEY
USING COLUMNS (id), and _get_type_id_column returns the native UUID type for cockroachdb,
matching what the ORM's Uuid() column renders on that dialect. Other dialects are
untouched.

Executed against CockroachDB v26.3: upgrade and downgrade both pass on a seeded harness
with multiple batches, and a 50,000 row backfill using this technique completes in about
3 seconds on a single node. The resulting constraint and index set matches the PostgreSQL
path exactly: task_instance_pkey on id plus the task_instance_composite_key unique
constraint.

Discussed on the devlist:
https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes - Claude Code

Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69260, opened 2026-07-02
Fix async database engine setup for cockroachdb connection URIs

viragtripathi:
SQLAlchemy's registered cockroachdb dialect speaks the PostgreSQL wire protocol and uses asyncpg as its async driver. Airflow 3.x derives the async connection URI from the sync one via AIO_LIBS_MAPPING in settings.py; because cockroachdb is not in that map, startup fails with "The loaded 'psycopg2' is not async" unless the user hand-writes
sql_alchemy_conn_async. This adds the one mapping entry and a parametrized test for the derivation function, which previously had no direct coverage.

Discussed on the devlist:
https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes - Claude Code

Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69640, opened 2026-07-09
Fix task execution upsert failing on cockroachdb connections

viragtripathi:
Airflow 3.3.0 introduced build_upsert_stmt (utils/sqlalchemy.py) for atomic INSERT ... ON CONFLICT writes, used by the execution API when tasks start (rendered task instance fields, variables, metastore state). The helper dispatches on the dialect name and raises ValueError for anything outside postgresql/mysql/sqlite, so on a cockroachdb connection every task start fails with a 500 from the execution API:

    `ValueError: Unsupported database dialect 'cockroachdb' for upsert.`

The cockroachdb SQLAlchemy dialect subclasses the postgresql one and supports the same INSERT ... ON CONFLICT DO UPDATE construct, so this adds it to the postgresql arm of the dispatch. Other dialects are unaffected. Includes unit tests for the postgres-family dispatch and the unsupported-dialect error.

Verified end to end against CockroachDB v26.3: without this change, task runs fail at startup on Airflow 3.3.0; with it, a two-scheduler stress run completes normally.

Related PRs from the same devlist proposal: #69260, #69555, #69556, #69557.
Discussed on the devlist:
https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

---

- [X] Yes - Claude Code

### DIFF
--- a/airflow-core/src/airflow/utils/sqlalchemy.py
+++ b/airflow-core/src/airflow/utils/sqlalchemy.py
@@ -136,7 +136,7 @@ def build_upsert_stmt(
     :raises ValueError: if the dialect does not support a known upsert syntax
     """
     stmt: MySQLInsert | PostgreSQLInsert | SQLiteInsert
-    if dialect == "postgresql":
+    if dialect in ("postgresql", "cockroachdb"):
         from sqlalchemy.dialects.postgresql import insert as pg_insert
 
         stmt = pg_insert(model).values(**values)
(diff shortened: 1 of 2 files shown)


shahar1:
Closing, following this discussion:

https://lists.apache.org/thread/t6jo4th3sn23jmr34m6gcxzw4k8mo4pc

PR #64612, opened 2026-04-02
Add cleanup to Kafka AwaitMessageTrigger for consumer management

jason810496:
* related: #64205

## Why

When testing out Kafka external-event driven Dag for ##64205 scenario, I encounter the following error with only enabling one consumer Dag

```
max.poll.interval.ms default 300000 timeout ...
```

## What

I verify locally that adding proper resource cleanup for Kafka Consumer can avoid the above error.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes (please specify the tool below) Codex for unit tests following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69641, opened 2026-07-09
Fix Kafka consumer not being closed on error in ConsumeFromTopicOperator

TimurRakhmatullin86:
`ConsumeFromTopicOperator.execute` only called `consumer.close()` on the happy path. If `apply_function` / `apply_function_batch` raised, or `consumer.consume()` / `consumer.commit()` failed, the operator leaked the consumer: the group member stays registered on the broker until `session.timeout.ms` expires (delaying partition rebalance for the rest of the consumer group), and the underlying librdkafka handle keeps its sockets and background threads alive in the worker until GC.

This wraps the processing loop in `try`/`finally` so the consumer is always closed:

- the loop body is unchanged (re-indent only);
- offset commits stay inside `try`, so nothing is committed for a failed batch;
- `close()` in `finally` is guarded with a `log.warning`, mirroring the `AwaitMessageTrigger` consumer cleanup from #64612 — `close()` performs a final synchronous offset commit when `enable.auto.commit` is on (the librdkafka default), so it can itself raise and would otherwise replace the original processing error as the task failure reason.

Tests: `test_execute_closes_consumer_when_apply_function_raises` (fails without the fix) and `test_execute_does_not_mask_error_when_close_raises`.

### DIFF
--- a/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
+++ b/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
@@ -126,68 +126,72 @@ class ConsumeFromTopicOperator(BaseOperator):
         self._validate_commit_cadence_before_execute()
         consumer = self.hook.get_consumer()
 
-        if isinstance(self.apply_function, str):
-            self.apply_function = import_string(self.apply_function)
+        try:
+            if isinstance(self.apply_function, str):
+                self.apply_function = import_string(self.apply_function)
 
-        if isinstance(self.apply_function_batch, str):
-            self.apply_function_batch = import_string(self.apply_function_batch)
+            if isinstance(self.apply_function_batch, str):
+                self.apply_function_batch = import_string(self.apply_function_batch)
 
-        if self.apply_function is not None and not callable(self.apply_function):
-            raise TypeError(f"apply_function is not a callable, got {type(self.apply_function)} instead.")
-
-        if self.apply_function:
-            apply_callable = partial(
-                self.apply_function,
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Real leak with blast radius beyond the failing task: with `close()` only on the happy path, any exception from `apply_function`, `consume()` or `commit()` left the group member registered on the broker until `session.timeout.ms` expired, which delays partition rebalance for every other consumer in the group — not just the task that failed.

I checked the "re-indent only" claim rather than taking it on trust: diffing added against removed lines ignoring whitespace, the only genuinely new lines are `try:`, `finally:`, `except Exception:` and the warning log, with nothing removed that lacks a re-indented counterpart. That is what makes a 154-line diff quick to be confident about.

Two details this gets right that are easy to miss:

Keeping the commits inside the `try` means a failed batch still commits nothing, so the leak is fixed without quietly shifting offset semantics.

Wrapping `close()` in its own `try`/`except` stops a failing close from replacing the original exception. `test_execute_does_not_mask_error_when_close_raises` pins exactly that — `close.side_effect` raises, and the test still asserts the original `ValueError("boom")` surfaces. Without the inner guard that test fails with the close error instead, which is the failure mode people usually discover in production rather than review.

Acquiring the consumer outside the `try` is also right: nothing to close if `get_consumer()` itself fails, and no unbound name for the `finally` to trip over.

One thing worth noting rather than fixing here: this cleanup idiom, down to the log wording, now exists in both this operator and `AwaitMessageTrigger.cleanup()` with no shared helper, so the two can drift.

---

PR #69643, opened 2026-07-09
Verify Java SDK artifact integrity continuously

uranusjr:
This allows us to be more confident about things *before* actually trying to do an RC.

### DIFF
--- a/.github/workflows/java-sdk-release-verify.yml
+++ b/.github/workflows/java-sdk-release-verify.yml
@@ -1,0 +1,164 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# Runs the Java SDK reviewer checklist ("Verifying a release" in
+# java-sdk/README.md) automatically, minus the GPG signature (only the release
+# manager's key can satisfy that).
+#
+#   * on a pull request touching java-sdk/ -> verify HEAD can produce a source
+#     release that is self-contained and builds from scratch.
+#   * on a `java-sdk/**` tag push -> the same check against the tagged commit
+#     (release readiness gate).
+#   * on manual dispatch -> verify the ACTUAL staged artifacts: the source
+#     package in dist/dev and the closed Nexus staging repo. Run this after
+#     staging, before sending the [VOTE] email.
+---
+name: Java SDK release verification
+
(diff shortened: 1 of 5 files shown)


.github/workflows/java-sdk-release-verify.yml
```diff
@@ -0,0 +1,164 @@
...
+        run: |
+          set -euo pipefail
+          rcdir="${TAG#java-sdk/}"          # <version>-rc<N>
+          version="${rcdir%-rc*}"           # <version>
+          base="https://dist.apache.org/repos/dist/dev/airflow/java-sdk/${rcdir}"
+          file="apache-airflow-java-sdk-${version}-src.tar.gz"
+          mkdir -p staged
+          curl -fsSL -o "staged/${file}" "${base}/${file}"
+          curl -fsSL -o "staged/${file}.sha512" "${base}/${file}.sha512"
+      - name: Install the latest Gradle to bootstrap a new wrapper for the source extracted from the tarball
```
jason810496, line 146: Do we need to pin the gradle version? Or it's kept as latest intentionally?

.github/workflows/java-sdk-release-verify.yml
```diff
@@ -0,0 +1,164 @@
...
+        run: |
+          set -euo pipefail
+          meta=$(curl -fsSL https://services.gradle.org/versions/current)
+          ver=$(echo "$meta" | jq -r .version)
+          url=$(echo "$meta" | jq -r .downloadUrl)
+          sum=$(curl -fsSL "$(echo "$meta" | jq -r .checksumUrl)")
+          curl -fsSL -o /tmp/gradle.zip "$url"
+          echo "${sum}  /tmp/gradle.zip" | sha256sum -c -
+          unzip -q -d "$HOME/gradle" /tmp/gradle.zip
+          echo "$HOME/gradle/gradle-${ver}/bin" >> "$GITHUB_PATH"
```
jason810496, line 106: Non-blocking: How about moving the download gradle part into the script?

java-sdk/scripts/ci/verify-source-release.sh
```diff
@@ -0,0 +1,152 @@
...
+}
+dependencies {
+    implementation(platform("org.apache.airflow:airflow-sdk-bom:${version}"))
+    implementation("org.apache.airflow:airflow-sdk")
+    implementation("org.apache.airflow:airflow-sdk-jpl")
+    implementation("org.apache.airflow:airflow-sdk-jul")
+    implementation("org.apache.airflow:airflow-sdk-log4j2")
+    implementation("org.apache.airflow:airflow-sdk-slf4j")
+    implementation("org.apache.airflow:airflow-sdk-processor")
+}
```
jason810496: Not sure do we need to resolve the list from the source to prevent the drift in the future (though the CI _should_ fail if there's drift)

jason810496 (approved):

.github/workflows/java-sdk-release-verify.yml
uranusjr: This is intentionally using latest because that’s what a human verifier would do on a new machine. Pinning happens when the wrapper is generated by Gradle.

.github/workflows/java-sdk-release-verify.yml
uranusjr: I want to keep this separate so the script can more easily be tested on its own if Gradle is already available (it usually is).

java-sdk/scripts/ci/verify-source-release.sh
uranusjr: Good idea, let me think about this.

jason810496 (approved):
Thanks!

PR #69644, opened 2026-07-09
validate log server hostname before building the log fetch url

Samin061:
When the api-server serves a task's logs it fetches them from the worker's log server at `http://{hostname}:{port}/log/...`, where `hostname` is whatever the worker reported for that task instance over the Execution API and was stored on the row unchanged. Nothing checks that value is a bare host before it is interpolated into the URL, so a hostname that carries URL syntax (for example `real-worker@attacker.host`, or a value with a path or scheme) breaks out of the host component and the request lands on a different host than intended. That request is not anonymous: `_fetch_logs_from_service` attaches a freshly signed `task-instance-logs` token and, by default, follows redirects, so the token can be delivered to an attacker-chosen host and a log-server redirect can bounce it further. I hit this while tracing where `ti.hostname` is written against where it is later trusted, and the two ends disagree on the value being a safe host. The fix validates the hostname against a bare-host pattern inside `_get_log_retrieval_url` so an unexpected value yields no fetch URL at all, which covers both the worker and triggerer paths, and sets `allow_redirects=False` on the fetch so a signed request cannot be redirected off-host. Valid hostnames and IPs are unaffected.

---

### DIFF
--- a/airflow-core/src/airflow/utils/log/file_task_handler.py
+++ b/airflow-core/src/airflow/utils/log/file_task_handler.py
@@ -23,6 +23,7 @@ import heapq
 import io
 import logging
 import os
+import re
 from collections.abc import Generator, Iterator
 from contextlib import suppress
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


Samin061:
gentle ping

PR #69646, opened 2026-07-09
Add raise_on_delivery_failure option to ProduceToTopicOperator

TimurRakhmatullin86:
Kafka reports delivery outcomes asynchronously through the producer's delivery callback. Today `ProduceToTopicOperator`'s default callback only *logs* delivery errors, so a message rejected by the broker (e.g. `MSG_SIZE_TOO_LARGE`, unknown topic, ACL denied) is silently lost while the task is still marked successful.

This PR adds an opt-in `raise_on_delivery_failure` parameter (default `False`, fully backwards compatible). When enabled, the operator collects errors passed to the delivery callback and, after the final `flush()` has resolved all outstanding deliveries, fails the task with an `AirflowException` reporting the number of failed messages and the first error. The configured `delivery_callback` is still invoked for every delivery report, so existing custom callbacks keep working unchanged.

New unit tests cover: task failure on delivery error, success path with the flag enabled, unchanged default behaviour, and interaction with a custom `delivery_callback` (all with a mocked producer, no broker required). Existing tests are untouched.

### DIFF
--- a/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/produce.py
+++ b/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/produce.py
@@ -40,6 +40,10 @@ def acked(err, msg):
         )
 
 
+class KafkaMessageDeliveryError(AirflowException):
+    """Raised when one or more messages could not be delivered to Kafka."""
+
+
 class ProduceToTopicOperator(BaseOperator):
     """
     An operator that produces messages to a Kafka topic.
@@ -58,7 +62,14 @@ class ProduceToTopicOperator(BaseOperator):
     :param synchronous: If writes to kafka should be fully synchronous, defaults to True
     :param poll_timeout: How long of a delay should be applied when calling poll after production to kafka,
         defaults to 0
-    :raises AirflowException: _description_
+    :param raise_on_delivery_failure: Kafka reports delivery outcomes asynchronously through the
+        delivery callback, so a message that is rejected by the broker (e.g. too large, unknown
+        topic, insufficient permissions) does not make the task fail by default - the error is only
+        passed to the delivery callback. Set to True to fail the task with an ``AirflowException``
+        when at least one message could not be delivered. Defaults to False to keep backwards
+        compatibility.
+    :raises AirflowException: If ``raise_on_delivery_failure`` is True and at least one message
+        could not be delivered.
 
     .. seealso::
         For more information on how to use this operator, take a look at the guide:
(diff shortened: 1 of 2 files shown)


potiuk:
This is worth having. A message the broker rejects — too large, unknown topic, ACL denied — is currently dropped while the task reports success, and data loss that looks like success is the worst failure mode to leave as the only option.

The implementation reads well. Defaulting to `False` keeps existing Dags untouched, and raising after the final `flush()` is the only correct point, since that is what resolves outstanding deliveries — raising earlier would miss failures still in flight. Wrapping the user's `delivery_callback` rather than replacing it is the detail most likely to go wrong in a change like this, and `test_operator_raise_on_delivery_failure_custom_callback` pins it directly with `_custom_delivery_callback_calls == [error]`. Defining `KafkaMessageDeliveryError` rather than raising `AirflowException` directly is also the right call.

One thing I would like documented before this merges. Enabling the flag changes retry semantics in a way that is not obvious from its name: the task fails *after* part of the batch has already been delivered, so a retry re-produces the whole batch and duplicates every message that succeeded the first time. That is inherent to at-least-once delivery rather than a flaw here, but someone switching on "raise on delivery failure" will not necessarily expect duplicates as the price. A sentence in the `:param raise_on_delivery_failure:` docstring would cover it — something like "On failure the task is retried from the beginning, so messages that were delivered successfully will be produced again."

Minor, take it or leave it: the wrapper collects errors even when the flag is off. It is harmless, since only failures are appended and a run with many of those has larger problems, but the collection serves no purpose in that path.

Separately — this branch has drifted a long way behind `main` (608 commits) and is now conflicting, so it will need a rebase before it can merge regardless of the above.

---

TimurRakhmatullin86:
Thanks @potiuk — addressed all three:

- **Docstring.** Added the retry/duplicate note to `:param raise_on_delivery_failure:`, close to your wording: *"Because the task fails after the batch has already been partially produced, a retry re-produces the whole batch, so messages that were delivered successfully the first time will be produced again."*
- **The minor one, taken.** The wrapper now only collects errors when the flag is set (`if err is not None and self.raise_on_delivery_failure`), so nothing is accumulated on the default path.
- **Rebased** onto `main` (was ~1093 behind). The only conflicts were in the test file — the imports and the new test block — resolved to keep both your `test_execute_rejects_empty_rendered_topic` and these tests. The rebase also picked up the fix to the pre-existing `:raises AirflowException: _description_` placeholder in the docstring.

Force-pushed.

TimurRakhmatullin86:
Friendly ping — is there anything else needed on this PR? Happy to make adjustments.

PR #69647, opened 2026-07-09
Add post-vote step to do a GitHub Release

uranusjr:


java-sdk/README.md
```diff
@@ -466,7 +466,22 @@ Reply with a `[RESULT][VOTE]` tally, then:
...
+4. **Publish** a GitHub release. Attach the **voted, signed** source artifacts.
+   From the directory holding the three signed files (e.g. `dist/release`):
+
+   ```bash
+   gh release create java-sdk/<VERSION> \
+     --repo apache/airflow \
+     --title "Apache Airflow Java SDK <VERSION>" \
+     --notes "See the Java SDK README for the features in this release." \
+     --verify-tag \
+     --prerelease \  # If this is a prerelease.
```
jason810496: IIUC, if this is voted, it shouldn't be prerelease? Or the beta version still consider as prerelease on GH?

jason810496 (approved):

java-sdk/README.md
uranusjr: I went back and forth on this; it’s marked beta so a prerelease marker makes sense. Apache doesn’t have anything to say with GitHub Releases (it’s just convenience and projects can do whatever) so the status here is ambiguious.

jason810496: I see, thanks for the clarification, then please feel free to keep it as-is.

phanikumv: Wondering if  line-continuation \ has to be the last character on the line — here it's followed by   # If this is a prerelease., so the backslash escapes a space instead of the newline. 

This would result in the three artifact files never get attached, and the filename lines run as a separate command,leading to command not found, exit 127. Have we tested this command?

phanikumv: lets move the comment to the top of the gh release command if this is the case 

`# Drop the --prerelease line if this is a final (non-beta) release. 
`

java-sdk/README.md
```diff
@@ -466,7 +466,22 @@ Reply with a `[RESULT][VOTE]` tally, then:
...
+     apache-airflow-java-sdk-<VERSION>-src.tar.gz \
+     apache-airflow-java-sdk-<VERSION>-src.tar.gz.asc \
+     apache-airflow-java-sdk-<VERSION>-src.tar.gz.sha512
+   ```
+
+5. Update the download page and wait ~1 hour after promoting so Central has
```
phanikumv, line 486: is this for uploading to Maven Central?

uranusjr: Yes; not related to this PR, this line changed simply to renumber the list item.

phanikumv (approved):

PR #69666, opened 2026-07-09
Fix ExternalTaskSensor log timezone

### LINKED ISSUE #69657: ExternalTaskSensor poke log shows logical_date in UTC despite configured timezone

Airflow Core

### Apache Airflow version

3.1.8

When `ExternalTaskSensor` logs its poke message, the datetime shown after `on` is serialized directly with `datetime.isoformat()`.

```shell
Poking for DAG 'example_external_dag' on 2026-07-06T21:00:00+00:00 ...
``` 

In deployments configured with a non-UTC timezone, such as AIRFLOW__CORE__DEFAULT_TIMEZONE=Asia/Seoul, this can be confusing because task log timestamps may be shown in the configured/local timezone while the ExternalTaskSensor poke message still displays the external logical date in UTC.

**Current code path**
```python
def _serialize_dttm_filter(dttm_filter):
    return ",".join(dt.isoformat() for dt in dttm_filter)
``` 

**Steps to reproduce**
Configure Airflow with a non-UTC default timezone, for example Asia/Seoul.
Create a Dag with an ExternalTaskSensor waiting for another Dag.
Run the sensor task.
Observe the poke log message.

The poke log should make the timezone easier to understand, ideally by formatting the datetime consistently with Airflow's configured timezone or by otherwise making clear that the value is the external Dag run `logical_date`.

This would reduce confusion when operators compare the sensor poke log with Airflow UI timestamps or local operational time.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

reachsrinivaschennupati-tech:
## Summary

Make `ExternalTaskSensor` poke logs display external logical dates in Airflow's configured timezone.

## Problem

`ExternalTaskSensor` logs currently serialize the external logical date directly from the stored filter value. This can make poke logs harder to read in deployments where Airflow is configured with a non-UTC timezone.

## Solution

- Add a log-only serializer that converts external logical dates to `settings.TIMEZONE`.
- Use the timezone-adjusted value in poke log messages for DAG, task group, and task checks.
- Keep `external_dates_filter` unchanged so listeners and integrations continue to receive the original serialized logical date.
- Added focused unit coverage for configured-timezone log serialization.

## Validation
- Added focused unit coverage for `ExternalTaskSensor._serialize_dttm_filter_for_log`.

closes: #69657

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/external_task.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/external_task.py
@@ -23,6 +23,7 @@ import warnings
 from collections.abc import Callable, Collection, Iterable, Sequence
 from typing import TYPE_CHECKING, ClassVar
 
+import pendulum
 from airflow.exceptions import AirflowProviderDeprecationWarning
 from airflow.models.dag import DagModel
 from airflow.providers.common.compat.sdk import (
@@ -30,6 +31,7 @@ from airflow.providers.common.compat.sdk import (
     BaseOperatorLink,
     BaseSensorOperator,
     conf,
+    timezone,
 )
 from airflow.providers.standard.exceptions import (
     DuplicateStateError,
@@ -311,6 +313,26 @@ class ExternalTaskSensor(BaseSensorOperator):
     def _serialize_dttm_filter(dttm_filter: Sequence[datetime.datetime]) -> str:
         return ",".join(dt.isoformat() for dt in dttm_filter)
 
+    @staticmethod
+    def _get_default_timezone() -> datetime.tzinfo:
+        default_timezone = conf.get_mandatory_value("core", "default_timezone")
+        if default_timezone == "system":
+            return datetime.datetime.now().astimezone().tzinfo or datetime.timezone.utc
+        return pendulum.timezone(default_timezone)
+
+    @classmethod
+    def _serialize_dttm_filter_for_log(cls, dttm_filter: Sequence[datetime.datetime]) -> str:
+        default_timezone = cls._get_default_timezone()
+        formatted_dates = []
+        for dt in dttm_filter:
+            serialized_dt = dt.isoformat()
+            default_timezone_dt = timezone.coerce_datetime(dt).astimezone(default_timezone).isoformat()
+            if serialized_dt == default_timezone_dt:
+                formatted_dates.append(serialized_dt)
+            else:
+                formatted_dates.append(f"{serialized_dt} (default timezone: {default_timezone_dt})")
+        return ",".join(formatted_dates)
+
     def poke(self, context: Context) -> bool:
         # delay check to poke rather than __init__ in case it was supplied as XComArgs
         if self.external_task_ids and len(self.external_task_ids) > len(set(self.external_task_ids)):
@@ -318,6 +340,7 @@ class ExternalTaskSensor(BaseSensorOperator):
 
         dttm_filter = self._get_dttm_filter(context)
         serialized_dttm_filter = self._serialize_dttm_filter(dttm_filter)
+        log_dttm_filter = self._serialize_dttm_filter_for_log(dttm_filter)
         # Save as attribute - to be used by listeners
         self.external_dates_filter = serialized_dttm_filter
 
@@ -326,7 +349,7 @@ class ExternalTaskSensor(BaseSensorOperator):
                 "Poking for tasks %s in dag %s on %s ... ",
                 self.external_task_ids,
                 self.external_dag_id,
-                serialized_dttm_filter,
+                log_dttm_filter,
             )
 
         if self.external_task_group_id:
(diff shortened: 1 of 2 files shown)


reachsrinivaschennupati-tech:
Hi maintainers, this PR is ready for review when you have a chance.

The change is intentionally log-only: `ExternalTaskSensor` keeps the original `external_dates_filter` behavior for listeners/integrations, while poke log messages serialize the external logical date in Airflow's configured timezone for readability.

Validation noted in the PR:
- Focused unit coverage for `ExternalTaskSensor._serialize_dttm_filter_for_log`.

Happy to adjust naming or test placement if there is a preferred provider-standard convention.

reachsrinivaschennupati-tech:
Hi Maintainers, this PR is ready for review when you have a chance, please review and trigger the CI workflow.

dhkim1920:
Thanks for working on this.

I had also started looking into a fix for #69657 and noticed two details that may be worth considering before review:

1. It may be clearer to retain the original external `logical_date` representation in the poke log and show the configured-timezone rendering alongside it, rather than replacing the logged value entirely.

This would preserve the canonical value for correlation and debugging while still making the configured local time immediately readable.

For example:

```text
Poking for DAG 'example_external_dag' on logical_date 2026-07-06T21:00:00+00:00 (default timezone: 2026-07-07T06:00:00+09:00) ...
```

2. The new test is currently located in `TestExternalTaskSensorV2`, which is skipped when running against Airflow 3. It would be useful to add equivalent coverage to the Airflow 3 test path so that the changed poke log is exercised on the version targeted by this PR.
```python
# tests\unit\standard\sensors\test_external_task_sensor.py TestExternalTaskSensorV2: line 135
@pytest.mark.skipif(AIRFLOW_V_3_0_PLUS, reason="Different test for v3.0+")
``` 

I have a local branch implementing this approach, but I would prefer not to open a near-duplicate PR unless the maintainers would like to evaluate it separately.

reachsrinivaschennupati-tech:
@dhkim1920 Thanks for your suggestions, I have added your recommended changes into this PR.

reachsrinivaschennupati-tech:
Hi Maintainers,

All requested changes and The CI checks have completed successfully.

This PR is now ready for final review and merge. When you have a chance, could you please review it?

Thank you!

reachsrinivaschennupati-tech:
Hi @rozza @potiuk @imbaczek  can you please review this PR.

potiuk (commented):
Thanks — the UX problem is real. The sensor logs the filter in UTC while operators think in their configured timezone, and adding the local rendering alongside rather than replacing it is the right call: `serialized_dttm_filter` still feeds `external_dates_filter` for listeners, so only the human-facing log changes. Suppressing the suffix when the two representations are identical is a nice touch.

Two things before this can land.

**It needs a rebase** — the branch is ~421 commits behind `main` and currently conflicts.

**The `from airflow import settings` import is worth a second look.** `providers/standard` has to work across the supported Airflow range, and `airflow.settings` is core internals rather than a stable provider-facing surface. I checked: other providers do import it, but they're all core-coupled ones — `fab`, `edge3` migrations and db modules, `celery`'s CLI — not a cross-version operator/sensor package. And `providers/common/compat/sdk` doesn't re-export `settings` or `TIMEZONE`, even though this same import statement pulls `timezone` from there, so the file ends up straddling both conventions.

Could you confirm `settings.TIMEZONE` is stable across the Airflow versions `standard` supports? If it isn't guaranteed, routing it through the compat layer (or reading the `core.default_timezone` config via the `conf` you already import) would be safer.

One small thing: both new tests are named `test_external_dag_sensor_log_uses_configured_timezone`. They're in different classes so there's no collision, but identical names make failures ambiguous in output. The first — calling `_serialize_dttm_filter_for_log` directly — is the stronger of the two, since it tests the formatting logic without depending on log capture.

---

reachsrinivaschennupati-tech:
@potiuk Thank you for the review. I rebased the branch onto current main and updated the timezone handling so the standard provider no longer imports airflow.settings.

The log formatter now reads core.default_timezone through the existing compat conf import, while keeping the original serialized logical date unchanged for external_dates_filter and listener/integration behavior. I also kept the configured-timezone suffix only for the human-facing log message.

I also cleaned up the tests: the direct formatter test is now outside the Airflow 2-only skipped class, and the Airflow 3 poke-log test has a distinct name so failures are easier to identify.

reachsrinivaschennupati-tech:
Thanks again for the review. I cleaned up the branch history and rebased it onto the latest main, so the conflict/merge commit is gone now.

The PR now has a single commit with the provider-safe timezone handling and the clearer test names. Local checks I ran after the rebase: \ and \.

PR #69212, opened 2026-07-01
Add providers E2E tests framework and OpenLineage tests

kacpermuda:
Add framework for providers E2E tests and implement OpenLineage provider e2e tests to Airflow CI

The OpenLineage provider emits lineage events when Dags run, but until now there was no automated way in Airflow CI to verify those events are actually correct against a real, deployed Airflow stack. This PR introduces that capability: a self-contained e2e test harness that spins up Airflow via docker-compose, runs the provider system-test Dags, and asserts the OpenLineage events emitted by the transport match expected payloads. The tests run on demand (via `workflow_dispatch`) and are triggered automatically by selective checks when the `openlineage` or `common` providers, the `providers-e2e-tests` harness, or related files change.

The approach follows the same pattern already established by the Task SDK integration tests and the Airflow e2e tests: a standalone directory under `providers-e2e-tests/openlineage/` with its own `pyproject.toml`, `docker-compose.yaml`, and pytest suite, invoked through a new `breeze testing providers-e2e-tests <provider>` command.

More providers can be added in the future if needed. These OL test do not rely on any external service, so they can easily be run in CI (no db, external api, google/amazon/azure service needed).

**What's included:**

- `providers-e2e-tests/` — new top-level directory; each provider gets its own subdirectory with a
  docker-compose stack, tests, and `pyproject.toml`.
- `breeze testing providers-e2e-tests <provider>` — spins up the stack and runs pytest.
- Two modes: default PROD image, or `--airflow-version <ver>` to test older Airflow + current providers
  from main (builds `apache/airflow:<ver>` + provider wheels). Required providers are declared in each
  provider's `pyproject.toml` under `[tool.e2e-tests] required-providers` — nothing hardcoded in breeze.
- OpenLineage implementation: postgres + API server + scheduler + dag-processor + triggerer + worker.
  DAGs come from the existing `providers/openlineage/tests/system/openlineage/` — `prepare_dags.py`
  copies and strips their pytest-only footers at runtime. Events are captured via `VariableTransport`
  (stores OL events in Airflow Variables); each DAG's `OpenLineageTestOperator` validates them — no
  external backend needed. A run that ends `success` means lineage matched.
- Selective checks trigger the suite on openlineage / common-providers changes. Core and task-sdk changes are not wired in - a regression there wouldn't trigger these tests on the PR, only on the next canary run. Adding core/task-sdk to the file group would work technically (PROD image is already built when those change via other test paths), but it would widen the trigger surface significantly. Leaving it for discussion.
[...]

PR #69671, opened 2026-07-09
Consolidate OpenLineage e2e tests into airflow-e2e-tests

shahar1:
Following review feedback on #69212, the OpenLineage provider e2e tests are consolidated into the existing `airflow-e2e-tests/` framework instead of the standalone `providers-e2e-tests/` framework introduced there.

`airflow-e2e-tests/` already hosts deployed-stack suites for providers (Elasticsearch/OpenSearch remote logging, Kafka event-driven), so the separate framework duplicated setup that already exists. OpenLineage now runs as an `openlineage` `--e2e-test-mode`: it reuses the shared docker-compose base, the testcontainers conftest, and the shared `AirflowClient`, adding only a compose overlay for the OpenLineage env config and a small runner that drives the provider's system-test DAGs.

The standalone `providers-e2e-tests/` tree, its `breeze testing providers-e2e-tests` command, its uv workspace member and its dedicated CI workflow are removed. The older-Airflow compatibility matrix (the only feature unique to that framework) is dropped along with it.

related: #69212

---


potiuk (approved):
NICE!

potiuk:
Screenshot 2026-07-09 at 22 23 29

potiuk:
Love it !!!!

shahar1:
> NICE!

The credit goes to @jason810496 for the idea :)

potiuk:
ALMOST !

jason810496:
> > NICE!
> 
> The credit goes to @jason810496 for the idea :)

Thanks Shahar for consolidating this!

jason810496 (approved):

shahar1:
CC: @kacpermuda

.github/workflows/additional-prod-image-tests.yml
```diff
@@ -329,34 +325,17 @@ jobs:
...
-    with:
-      provider: "openlineage"
-      provider-display-name: "OpenLineage"
-      runners: ${{ inputs.runners }}
-      platform: ${{ inputs.platform }}
-      default-python-version: "${{ inputs.default-python-version }}"
-      use-uv: ${{ inputs.use-uv }}
-    if: inputs.canary-run == 'true' || inputs.run-providers-e2e-tests-openlineage == 'true'
-
-  test-providers-e2e-tests-openlineage-compat:
```
kacpermuda, line 344: Will review in a second, but first spot is that we would no longer run on previous airflow versions, which was crucial. I do not see those in the current CI as well, so it was removed. Any reason for that? I thought it's just about moving to different directory in repo rather than changing the logic of the tests.

shahar1: It was an AI overdose side-effect, apologies for that :)

I will restore the coverage for older versions, hopefully Jarek won't be disappointed by increase in number of lines.

However, just wondering if it's really useful to run it for **all** old 3.X versions (3.0, 3.1, 3.2, 3.3), and not only the oldest one (3.0). We're currently in shortage of CI resources, so each job adds its own overhead (multiply by number of PRs per time unit) - which we try to reduce in these days.

What do you think?

kacpermuda: I think that it's valuable to keep all the versions in compat, as OL is one these "meta" providers that has a lot of reliability on core, so it's easy to break some compatibility by something that looks like small and innocent change. These compat jobs are already using airflow images published, just adding some providers on top of it, so they are not very heavy. I do however understand the need to save the CI resources, I think if it turns out that it is consuming too much, we can always trigger these compat jobs only for scheduled runs on main (those 4 times a day if I remember correctly?) And only run them with prod image for PRs, that way, we can reduce the resource usage and still keep the early bug detection. 

How does this sound?

airflow-e2e-tests/tests/airflow_e2e_tests/openlineage_tests/prepare_dags.py
```diff
@@ -30,41 +30,21 @@
...
 
-import os
 import shutil
 from pathlib import Path
 
-HERE = Path(__file__).resolve().parent
-AIRFLOW_ROOT_PATH = HERE.parents[1]
+from airflow_e2e_tests.constants import AIRFLOW_ROOT_PATH
 
-# Example DAGs that need a minimum Airflow version. When running against an older version, the DAG
```
kacpermuda, line 40: Let's see when the compat are back if this chunk and other version-related from this file are needed, may need to be re-added.

providers-e2e-tests/openlineage/README.md
kacpermuda, line 1: Can we keep this README close to the new OL tests directory? Even thought the layout needs adjusting, it still explain how the tests are executed and the logic behind it, very useful for debugging it.

kacpermuda (commented):
Two more comments, in general a very nice cleanup PR moving these tests to a common structure. Thanks for doing it !

.github/workflows/additional-prod-image-tests.yml
kacpermuda: One more question around this, I see that we moved compat to canary only, will the "full tests needed" label still trigger these compats? It would be awesome to not run by default on every OL PR, but be able to trigger them for some bigger changes. Not sure how this label works now

kacpermuda (approved):
All looks good, thanks for cleaning it up.

PR #69675, opened 2026-07-09
Fix asset event extra filter matching wrong events for dotted keys on SQLite

steveahnahn:
Asset event `extra` filtering (public API `GET /assets/events` `extra` param, and the Execution API asset-event lookups) uses a dialect-aware JSON containment filter. The SQLite fallback built an **unquoted** `json_extract` path from the user-supplied key, so a key containing JSON-path metacharacters — e.g. `spark.executor.memory` — was interpreted as **nested-object navigation** instead of a literal key. Two silent wrong results follow, diverging from PostgreSQL (`@>`) and MySQL (`JSON_CONTAINS`), which both match keys literally:

- an event with the literal key `{"spark.executor.memory": "4g"}` is **missed** (false negative), and
- an event nesting the same path `{"spark": {"executor": {"memory": "4g"}}}` is **wrongly matched** (false positive).

Same filter, different events returned depending on the database backend — with no error. Both `key=value` parsers pass such keys through verbatim, so this is reachable from both API surfaces.

The fix quotes the key in the JSON path (`$."spark.executor.memory"`), which SQLite treats literally; `json.dumps` produces exactly the quoting/escaping the JSON path syntax expects (also covering keys containing quotes or backslashes). PostgreSQL and MySQL code paths are untouched.

**Evidence**

- Fail-first on SQLite: the new bracket-key case fails unfixed (returns 0 events instead of 1), and the identity test fails unfixed by returning the *nested* event instead of the literal-key event. Notably, the dotted-key *count* assertion alone passes unfixed because the wrong event coincidentally keeps the count at 1 — hence the added identity assertion on `source_run_id`.
- Backend differential: the same identity test on unfixed code passes on PostgreSQL and fails on SQLite; with the fix both agree. Verified on a real PostgreSQL backend via breeze: the unfixed code passes all four new tests on PostgreSQL while two of them fail on SQLite; with the fix, all four pass on both backends.
- Regression: full `test_assets.py` (151), execution API `test_asset_events.py` (35), `test_sqlalchemy.py` (23) all pass.

---

  - Claude Code (Fable 5)

### DIFF
--- a/providers/common/sql/src/airflow/providers/common/sql/operators/sql.py
+++ b/providers/common/sql/src/airflow/providers/common/sql/operators/sql.py
@@ -744,35 +744,31 @@ class SQLColumnCheckOperator(BaseSQLOperator):
         if record is None and self.accept_none:
             record = 0
         match_boolean = True
+
+        # ``tolerance`` is a fraction of the expected value that the record may deviate by. The
+        # allowed margin must widen the bound in both directions regardless of the expected value's
+        # sign, so it is computed from the magnitude: ``abs(expected) * tolerance``. Scaling the bound
+        # by ``(1 ± tolerance)`` instead breaks for negative expected values (it moves the bound the
+        # wrong way, so a record equal to the expected value fails the check).
+        def _margin(expected):
+            return abs(expected) * tolerance if tolerance is not None else 0
+
         if "geq_to" in check_values:
-            if tolerance is not None:
-                match_boolean = record >= check_values["geq_to"] * (1 - tolerance)
-            else:
-                match_boolean = record >= check_values["geq_to"]
+            match_boolean = record >= check_values["geq_to"] - _margin(check_values["geq_to"])
         elif "greater_than" in check_values:
-            if tolerance is not None:
-                match_boolean = record > check_values["greater_than"] * (1 - tolerance)
-            else:
-                match_boolean = record > check_values["greater_than"]
+            match_boolean = record > check_values["greater_than"] - _margin(check_values["greater_than"])
         if "leq_to" in check_values:
-            if tolerance is not None:
-                match_boolean = record <= check_values["leq_to"] * (1 + tolerance) and match_boolean
-            else:
-                match_boolean = record <= check_values["leq_to"] and match_boolean
+            match_boolean = (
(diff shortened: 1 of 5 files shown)


providers/common/sql/src/airflow/providers/common/sql/operators/sql.py
yuseok89, line 1: The `_get_match` negative-`tolerance` fix itself looks correct to me. That said, this change seems unrelated to the PR's topic (quoting the JSON path for asset-event `extra` filtering), and it isn't mentioned anywhere in the title, description, newsfragment, or the regression list. Is it included intentionally, or did it slip in from another branch? It's a valid fix on its own, so splitting it into its own PR would probably keep things cleaner.

steveahnahn: Good catch, thanks. The `_get_match` change is indeed unrelated. It slipped in from another working branch, dropping and cleaning up this PR!

potiuk (approved):
Thanks — this is a good find, and the kind of bug that's very hard to notice in the wild.

`_default_json_contains` built the SQLite path as `f"$.{k}"`, and SQLite's `json_extract` treats `.` and `[` as **path navigation**. So a filter for the literal key `"spark.executor.memory"` produced `$.spark.executor.memory`, which:

- **missed** the event that actually has that literal key, and
- **matched** an unrelated event nesting `{"spark": {"executor": {"memory": ...}}}`

Wrong results in both directions — and silently, since nothing errors. Worse, PostgreSQL's `@>` and MySQL's `JSON_CONTAINS` compare literally, so the same filter returned *different results depending on the backend*. That's the sort of divergence that survives for years because it only reproduces on one deployment type.

`json.dumps(k, ensure_ascii=False)` is the right fix: SQLite accepts double-quoted key names in JSON paths, and routing through `json.dumps` gets embedded quotes and backslashes escaped correctly rather than hand-rolling it.

The test is exactly the discriminating one. Creating all three events — the literal dotted key, the nested-object equivalent, and `partitions[0]` — and asserting the filter matches only the literal is what distinguishes a real fix from one that merely stops erroring. A test with just the dotted key would have passed against several wrong implementations.

Newsfragment is correctly included, and correctly a `.bugfix` — this changes results users may have unknowingly depended on for SQLite deployments.

---

PR #69520, opened 2026-07-07
Drop the LogTemplate DB model

jason810496:
- related: #53821, #59027 (stale predecessor of this PR), [[DISCUSS] Drop LogTemplate DB Model for ElasticsearchTaskHandler in Airflow 3](https://lists.apache.org/thread/nlmhs1plo77qnlp7rqk27mkb2hs41f1p)

## Why

The `LogTemplate` model pinned historical values of `[logging] log_filename_template` / `[elasticsearch] log_id_template` per `DagRun`, and task handlers read it via `dag_run.get_log_template(session=...)` — a direct metadata-DB dependency that conflicts with the Airflow 3 execution model and couples log handlers to an Airflow-core model just to read two strings already available from config.

Airflow 2.11.1 already disabled per-run template retrieval by default for security reasons (#61880); this completes that direction on main by removing the model entirely.

## What

- Remove the `LogTemplate` model, the `dag_run.log_template_id` column/FK, `DagRun.get_log_template()`, and `synchronize_log_template()` with its call sites. `FileTaskHandler._render_filename()` now reads `[logging] log_filename_template` from live config at render time.
- Add migration `0126_3_4_0_drop_log_template.py` (downgrade recreates the column and table); both directions use `disable_sqlite_fkeys(op)` so the SQLite `dag_run` rebuild doesn't cascade-delete `task_instance` rows.
- Elasticsearch: remove the `USE_PER_RUN_LOG_ID` flag — defined but unused since `ElasticsearchRemoteLogIO`, so dead-code cleanup only.
- OpenSearch: keep the `hasattr(DagRun, "get_log_template")` shim — per-run pinning must keep working on cores 3.0-3.3; on 3.4.0+ it self-disables and the handler falls back to `[opensearch] log_id_template` from config.
- Make the shared `create_log_template` fixture version-adaptive so provider tests pass against all supported cores; update ES docs and add a significant-change newsfragment.

**Behaviour change (core 3.4.0+):** changing `log_filename_template` / `log_id_template` now applies immediately and retroactively to all DagRuns — there is no more per-run historical pinning, so logs written under a previous template are no longer reachable through the new one.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes, with help of Claude Code (Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69688, opened 2026-07-10
Deliver the per-run log ID template to workers via TIRunContext

jason810496:
- related: #69520 (discussion on dropping vs. fixing the LogTemplate flow)
- next:
  - https://github.com/apache/airflow/pull/69896
  - https://github.com/apache/airflow/pull/69897

## Why

The `LogTemplate` model pins the log ID template to each Dag run so task logs written before a `log_id_template` config change stay readable, but on the Airflow 3 write path workers render the log ID from the current conf value — they cannot query the metadata DB to fetch the pinned row. This PR adds the core plumbing that hands the pinned template to workers; the Elasticsearch and OpenSearch handler changes that consume it follow in per-provider PRs stacked on this one.

## How

- The Dag-run-pinned `LogTemplate.elasticsearch_id` is delivered to the supervisor via a new optional `TIRunContext.log_id_template` field, populated by the `ti_run` Execution API endpoint.
- `upload_to_remote` forwards the run context only to handlers whose `upload` opts into a `ti_context` keyword, so S3/GCS/third-party `RemoteLogIO` implementations and old-provider × new-sdk combinations are untouched.

## What

- `airflow-core`: add `TIRunContext.log_id_template`; populate it in the `ti_run` route from `dr.get_log_template()`; add Execution API version `2026-09-30` with a Cadwyn migration.
- `task-sdk`: retain the `TIRunContext` on `ActivitySubprocess` and pass it to `upload_to_remote`; regenerate the client datamodels; add the first in-progress supervisor-schema version file plus the regenerated `schema.json` snapshot.

---

- [x] Yes, with help of Claude Code (Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
+++ b/airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
@@ -278,6 +278,7 @@ class TestTIRunState:
             "variables": [],
             "connections": [],
             "xcom_keys_to_clear": [],
+            "log_id_template": "{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}",
         }
         # upstream_map_indexes is now computed by Task SDK, not returned by the server in HEAD version
         assert "upstream_map_indexes" not in result
@@ -692,6 +693,7 @@ class TestTIRunState:
             "next_method": "execute_complete",
             "next_kwargs": expected_next_kwargs,
             "start_date": None,
+            "log_id_template": "{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}",
         }
 
     @pytest.mark.parametrize("resume", [True, False])
@@ -766,6 +768,7 @@ class TestTIRunState:
             "xcom_keys_to_clear": [],
             "next_method": "execute_complete",
             "next_kwargs": expected_next_kwargs,
+            "log_id_template": "{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}",
         }
         session.expunge_all()
         ti = session.get(TaskInstance, ti.id)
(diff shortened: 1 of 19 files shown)


airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
```diff
@@ -1030,6 +1033,41 @@ def test_ti_run_populates_team_name(
...
+        # The Dag run pins the newest LogTemplate row at creation time.
+        ti = create_task_instance(
+            task_id="test_ti_run_populates_log_id_template",
+            state=State.QUEUED,
+            dagrun_state=DagRunState.RUNNING,
+            session=session,
+            start_date=instant,
+            dag_id=str(uuid4()),
+        )
+        session.commit()
```
phanikumv, line 1055:
```suggestion
        session.commit()

        # A newer LogTemplate lands after the run is created; the run must keep its
        # pinned row, not switch to the latest one.
        create_log_template("{{ ti.dag_id }}.log", "{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}")
```

phanikumv: The current logic doesnt check whether we are returning the pinned template or the latest template

If we add a second, newer LogTemplate after the run is created, pinned and latest become different rows, and then it will only pass if the code returns the run's pinned template.

jason810496: Good idea! Rather than extending this test case, I'd prefer to add a new one that starts with a loop to exercise the scenario where the API returns a new expected log template whenever there's an update. I just added in https://github.com/apache/airflow/pull/69688/changes/3bd771134a087a3123274bd63521c4ccf038f67b

PR #69689, opened 2026-07-10
Replace vars(response) with attribute access in Azure Data Factory and Synapse operators for SDK v10 compatibility

MalteNiederstadt:
Summary

  azure-mgmt-datafactory 10.0.0 (released 2026-07-08) introduced hybrid models (https://aka.ms/azsdk/python/migrate/hybrid-models) that change the internal __dict__ representation of response objects. This
  breaks vars(response)["run_id"] with a KeyError, even though response.run_id still works.

  Root Cause Chain

  1. The provider's pyproject.toml declares azure-mgmt-datafactory>=2.0.0 with no upper bound (source (https://github.com/apache/airflow/blob/main/providers/microsoft/azure/pyproject.toml))
  2. When a managed Airflow environment (e.g., Cloud Composer) rebuilds, pip resolves azure-mgmt-datafactory==10.0.0 — which was released on 2026-07-08
  3. SDK v10 uses new "hybrid models" where attributes are exposed via properties/descriptors rather than stored directly in __dict__
  4. The operator accesses the response via vars(response)["run_id"], which reads __dict__ directly — this now raises KeyError since the attribute is no longer in __dict__

  Error

    ERROR - Task failed with exception
      File "/opt/python3.11/lib/python3.11/site-packages/airflow/sdk/execution_time/task_runner.py", line 1068, in run
        result = _execute_task(context=context, ti=ti, log=log)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/python3.11/lib/python3.11/site-packages/airflow/sdk/execution_time/task_runner.py", line 1477, in _execute_task
        result = ctx.run(execute, context=context)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/python3.11/lib/python3.11/site-packages/airflow/sdk/bases/operator.py", line 417, in wrapper
        return func(self, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/python3.11/lib/python3.11/site-packages/airflow/providers/microsoft/azure/operators/data_factory.py", line 184, in execute
        self.run_id = vars(response)["run_id"]
                      ~~~~~~~~~~~~~~^^^^^^^^^^
    KeyError: 'run_id'

  Affected operators

  • AzureDataFactoryRunPipelineOperator (data_factory.py:184)
  • AzureSynapseRunSparkBatchOperator (synapse.py:107)
  • AzureSynapseRunPipelineOperator (synapse.py:260)

  Fix

  Replace vars(response)["key"] with direct attribute access response.key, which works with both old and new SDK versions (attribute access is part of the public API contract for both classic and hybrid
  models).
  
  Hotfix: Pin azure-mgmt-datafactory <10.0.0

  Ref

  • https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/datafactory/azure-mgmt-datafactory/CHANGELOG.md#1000-2026-07-07
  • Breaking change: "This version introduces new hybrid models which have dual dictionary and model nature"
  • Migration guide: https://aka.ms/azsdk/python/migrate/hybrid-models

  ────────────────────────────────────────

  • [x] Yes — Cortex Code (Snowflake)

  <!--
  Generated-by: Cortex Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)
  -->


shahar1 (commented):
Could you please rename the PR's title? 
We don't use conventional commits, so just replace it with a oneliner that explains what the change does specifically, without prefixes. Thanks!

MalteNiederstadt:
> Could you please rename the PR's title? We don't use conventional commits, so just replace it with a oneliner that explains what the change does specifically, without prefixes. Thanks!

my bad - thanks for the hint.

shahar1 (commented):
And now for the official code review :)
Could you please add some unit tests to ensure that there won't be future regressions?

MalteNiederstadt:
> And now for the official code review :) Could you please add some unit tests to ensure that there won't be future regressions?

thanks, as this is my first PR towards airflow - I greatly appreciate your guidelines.

Vamsi-klu:
Reads correct to me. On the classic msrest models `response.run_id` and `response.id` resolve to the same values `vars(response)[...]` was pulling out of `__dict__`, and the plain attribute access is what keeps the v10 hybrid models working since those expose the fields through property descriptors. Worth noting the `run_spark_job` hook already sets `self.job_id = job.id` internally, so the synapse `execute()` change just lines up with what the hook was doing. The `HybridModelResponse` regression tests that only expose the value via a property are the right shape here, since they'd fail on the old `vars()` path. Only nit, this trades a possible `KeyError` for an `AttributeError` if the SDK ever drops the field, but the old code already assumed the key was present so it's no worse than before.

MalteNiederstadt:
Any further action required from my side in order to get this completed?

potiuk (approved):

PR #69526, opened 2026-07-07
Make psycopg (v3) the default synchronous Postgres driver in providers

Dev-iL:
closes:
- #69443

related:
- #68453
- #69469

---

Makes `psycopg` (v3) the default **synchronous** Postgres driver across the providers, mirroring the async default switch (#69089). psycopg3 is used wherever it actually works (SQLAlchemy 2.0+); `psycopg2-binary` **stays installed by default** as a fallback so nothing regresses on the Airflow versions this provider still supports.

### Why psycopg2 stays a default dependency (not demoted to an extra)

An earlier revision of this PR moved `psycopg2-binary` to an opt-in `[psycopg2]` extra. That regresses every core older than 3.4.0, so it now stays a default dependency — gated to become opt-in only once the minimum supported Airflow is 3.4.0. The reasons mirror #69690 (which kept `asyncpg` installed by default for the same class of reason on the async side):

- **Core normalizes the sync metadata-DB conn to `postgresql+psycopg2://`.** On every currently released core, `configuration.py::_upgrade_postgres_metastore_conn` rewrites a bare `postgresql://` / legacy `postgres://` / `postgres+psycopg2://` `sql_alchemy_conn` to explicit `postgresql+psycopg2://`, and `settings.py` builds the metadata engine from that string with no psycopg fallback. Dropping psycopg2 breaks the metadata DB.
- **Airflow 2.11 ships SQLAlchemy 1.4**, which has no native `postgresql+psycopg` (v3) dialect at all. Since this provider still supports `apache-airflow>=2.11.0`, psycopg2 is the only viable sync driver there.

So `psycopg[binary]`, `psycopg2-binary`, and `asyncpg` are all default dependencies. The removal of the psycopg2/asyncpg defaults is deferred, gated on the floor bump to 3.4.0, and noted at the workaround site (tracked at #68453).

### Relationship to #69469 (the core half)

The sync switch for Airflow's own metadata DB lives in **core #69469**: it flips `_upgrade_postgres_metastore_conn` to derive `postgresql+psycopg://` when psycopg (v3) is importable (gated to land in 3.4.0), mirroring the async `_USE_PSYCOPG3` guard.

[...]

PR #69089, opened 2026-06-27
Switch the default async Postgres driver from asyncpg to psycopg3 (provider)

Dev-iL:
Split from the below, see details there: 
- https://github.com/apache/airflow/pull/68496

> [!IMPORTANT]
> Note to RMs: this introduces a breaking change that requires a major version bump!

Issue #68453, opened 2026-06-12
Remove psycopg2 and migrate the sync Postgres driver to psycopg3


Dev-iL:
## Context

#67801 switches the **async** Postgres driver default from asyncpg to psycopg3 (`postgresql+psycopg_async://`) and makes `psycopg[binary]` a hard dependency of `apache-airflow-providers-postgres`. To keep that PR reviewable, `psycopg2-binary` was deliberately kept as a hard dependency and the **sync** engine still uses psycopg2 (`postgresql+psycopg2://`, made explicit in #68314). This issue tracks the deferred follow-up: removing psycopg2 entirely and serving both sync and async from psycopg3.

## Why deferred (and why it should happen)

- Keeping both drivers doubles the installed footprint for no long-term benefit; psycopg3 serves sync and async from a single driver.
- SQLAlchemy 2.1 changes the default dialect for `postgresql://` URLs from psycopg2 to psycopg3, so the sync migration naturally aligns with the SQLA 2.1 upgrade (early experiments with that upgrade already surfaced breakage, e.g. in celery, from the implicit driver flip — hence the explicit-URL groundwork in #68314).

## Work

1. Migrate the sync metadata engine default to psycopg3 (`postgresql+psycopg://`).
2. Remove `psycopg2-binary` from `providers/postgres` hard dependencies (breaking change — major provider bump).
3. **Auto-upgrade legacy URLs:** once psycopg2 is no longer guaranteed to be installed, `postgresql://` (and `postgresql+psycopg2://`) connection URLs must be automatically mapped to `postgresql+psycopg://` when psycopg2 is absent, so existing deployment configs keep working.
4. Validate bulk-write paths (psycopg3 showed markedly slower batch INSERT/COPY in driver benchmarks — see #67801) before any bulk path moves to it.
5. Update deployment docs and `PostgresHook` as applicable.

## Acceptance criteria for closing

- psycopg2 is no longer a hard dependency anywhere in the repo.
- Legacy `postgresql://` / `postgresql+psycopg2://` configs either keep working via auto-upgrade or fail with a clear actionable error.
- Docs describe psycopg3 as the single default Postgres driver for sync and async.

Dev-iL:
### Manifest

~~~md
# Definition: Migrate the sync Postgres driver from psycopg2 to psycopg3 (apache/airflow#68453)

## 1. Intent & Context

[...]

PR #69690, opened 2026-07-10
Keep asyncpg installed by default in the Postgres provider

shahar1:
### Problem

Postgres provider `7.0.0` (currently in RC as `7.0.0rc1`, #69526 / merged #69089) made **psycopg3 the default async metadata-DB driver** and, as part of that, **moved `asyncpg` from a default dependency to the opt-in `[asyncpg]` extra**.

The psycopg-async default and its `postgresql+asyncpg://` fallback only exist on **Airflow 3.4.0+** (`airflow-core` `settings.py::_get_async_conn_uri_from_sync`). Every **currently released** 3.x core (3.0–3.3), which this provider still supports (`apache-airflow>=2.11.0`), derives the async metadata-DB URL as `postgresql+asyncpg://` **unconditionally** and has **no psycopg fallback**.

So upgrading the Postgres provider to `7.0.0` on any released 3.x core (or a 3.4.0 deployment that pinned an older core) leaves the async metadata-DB engine unable to load its driver:

```
sqlalchemy.exc.NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:postgresql.asyncpg
```

This is a regression against cores the provider claims to support, and a reason to hold `postgres 7.0.0rc1`.

### Fix

Keep `asyncpg` as a **default** dependency of the Postgres provider until its minimum supported Airflow is `3.4.0` (i.e. until every supported core knows the `psycopg_async` fallback). psycopg3 remains the default *async* driver on 3.4.0+; asyncpg is simply still present so older supported cores keep working. The removal is gated on the floor bump and noted at the workaround site.

- `pyproject.toml`: add `asyncpg>=0.30.0` back to default `dependencies` (comment explains the constraint + tracking issue).
- `changelog.rst`: correct the 7.0.0 note — asyncpg stays installed by default; the psycopg_async switch is 3.4.0+ behavior only.
- `README.rst` / `docs/index.rst` / `uv.lock`: regenerated by prek hooks.

The eventual removal (converge on psycopg3-only) is tracked at #68453.

related: #69526
related: #68453

### Test plan

Dependency/packaging + changelog change; no Python behavior to unit-test (the affected async-URL derivation lives in `airflow-core`, unchanged here). Verified: prek `Update dependencies for providers`, `Sync provider README.rst Requirements table with pyproject.toml`, `Update uv.lock`, and changelog-format hooks all pass.

> Note: two related **core** hardening items remain as separate follow-ups (not required to unblock the RC): core `_USE_PSYCOPG3` gates on `find_spec("psycopg")` only, missing the SQLAlchemy-2 guard the provider hook has; and the derived async URL is not validated against installed drivers, so failures surface as an opaque `NoSuchModuleError`.

---

### DIFF
--- a/providers/postgres/README.rst
+++ b/providers/postgres/README.rst
@@ -60,6 +60,7 @@ PIP package                                 Version required
 ``psycopg2-binary``                         ``>=2.9.10; python_version >= "3.13"``
 ``psycopg[binary]``                         ``>=3.2.9; python_version < "3.14"``
 ``psycopg[binary]``                         ``>=3.3.3; python_version >= "3.14"``
+``asyncpg``                                 ``>=0.30.0``
 ==========================================  ======================================
 
 Optional cross provider package dependencies
--- a/providers/postgres/docs/changelog.rst
+++ b/providers/postgres/docs/changelog.rst
@@ -34,14 +34,19 @@ Breaking changes
 ~~~~~~~~~~~~~~~~
 
 .. note::
-    The default async metadata-database driver is now ``psycopg`` (psycopg3),
-    which is installed by default in place of ``asyncpg``. As a result, the
-    ``[psycopg]`` optional extra was removed and ``[asyncpg]`` was added.
-
-    Starting with Airflow 3.4.0, the default derived async connection URL changes from
+    On Airflow 3.4.0 and later, the default async metadata-database driver becomes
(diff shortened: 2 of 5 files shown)


Dev-iL (approved):
Thank you for opening this. The reasoning is sound, the implementation looks correct.

vincbeck (approved):

PR #69695, opened 2026-07-10
Migrate Exasol provider to pyexasol 2.x

### LINKED ISSUE #69123: Migrate exasol provider to pyexasol 2.x and remove the <2 cap
The "Upgrade important CI environment" automation (#68933) pulled `pyexasol` from 1.3.0 to **2.2.2** (a major release). pyexasol 2.x ships stricter type information, so mypy now reports 8 errors in `providers/exasol/src/airflow/providers/exasol/hooks/exasol.py` (e.g. `ExaConnection` vs `Connection` variable reuse in `get_conn`, `execute()` argument types, `export_to_pandas` `query_params`).

To keep the CI-tooling upgrade unblocked, #68933 caps `pyexasol>=0.26.0,<2` in `providers/exasol/pyproject.toml`.

**Follow-up work:** adapt the exasol hook to pyexasol 2.x (separate the Airflow `Connection` and `ExaConnection` locals, fix `execute`/`export_to_pandas` call signatures), update tests, then remove the `<2` cap and re-lock.

**Acceptance:** cap removed; exasol mypy and unit tests pass against pyexasol 2.x.

---

joshuabvarghese:
This PR upgrades the Exasol provider to support `pyexasol` version 2.x. 

* closes: #69123

As `pyexasol` 2.x introduces changes to the underlying library API, this PR includes the necessary refactoring of the Exasol hook, updates to the unit tests, and adjustments to the provider's dependencies and documentation to maintain full compatibility.

### Key Changes:
* **Variable Separation:** Decoupled the Airflow `Connection` from `ExaConnection` in `get_conn` to resolve variable reuse typing errors.
* **Runtime Validation:** Added explicit type checks for `query_params` (enforcing a literal `dict` to match `pyexasol`'s `**query_params` unpacking) and `sql` statements (rejecting lists in `get_records`/`get_first`). This respects the `DbApiHook` base class interface while failing fast on invalid inputs.
* **Dependency Ceiling:** Bumped `pyproject.toml` to `pyexasol>=2.0.0,<3`, establishing a proper major-version ceiling.
* **Test Coverage:** Added negative testing to verify the new runtime validations and fixed a mocked test that was previously hiding a runtime `TypeError`.
* **Documentation:** Updated `README.rst`, `docs/index.rst`, and `docs/changelog.rst` to reflect the breaking changes and new dependencies.

### Testing
- `mypy` strictness verified locally (0 errors in `exasol.py`).
- Unit tests passing.
-  Standalone runtime harness executed against installed `pyexasol 2.2.2` to verify end-to-end type execution on `conn.execute()`.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @joshuabvarghese** · by `@potiuk` · 2026-07-28 16:32 UTC
>
> Some review feedback from `eladkal` is waiting on you:
> - There is **1 unresolved review thread** on this PR from `eladkal`.
>
> **The ball is in your court** — you've been assigned to this PR. Push a fix or reply in each thread explaining why the feedback does not apply, then mark them resolved and ping the reviewer (`eladkal`) for a final look.
>
>
> **Note:** your branch is **294 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>

### DIFF
--- a/providers/exasol/docs/changelog.rst
+++ b/providers/exasol/docs/changelog.rst
@@ -27,6 +27,26 @@
 Changelog
 ---------
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+* ``Migrate exasol provider to pyexasol 2.x; remove the <2 cap (#69123)``
+
+  This provider now requires ``pyexasol>=2.0.0,<3``. pyexasol 2.0.0 ships a ``py.typed`` marker
+  (`pyexasol changelog <https://exasol.github.io/pyexasol/master/changes/changes_2.html>`__), which
+  surfaces several previously-hidden type mismatches in the Exasol hook:
+
+  * ``ExasolHook.get_records()`` and ``ExasolHook.get_first()`` now only accept a single SQL string,
+    not a list of statements (pyexasol's underlying ``execute()`` never actually supported a list; use
+    ``ExasolHook.run()`` for executing multiple statements in sequence).
+  * ``parameters``/``query_params`` passed to ``get_records()``, ``get_first()``, ``run()``, and
+    ``get_df()``/``get_pandas_df()`` must be a ``dict``/``Mapping``. Positional-style parameters
+    (lists/tuples) were never actually supported by pyexasol at runtime and will now raise a clear
+    ``TypeError`` instead of failing deep inside pyexasol (or silently doing the wrong thing).
+  * If you pass ``export_params={"with_column_names": ...}`` to ``ExasolHook.export_to_file()`` or
+    ``ExasolToS3Operator``, note that pyexasol 2.0.0 fixed this option to be interpreted as a real
+    boolean; previously any value (including ``False``) was treated as ``True``.
+
 4.10.4
 ......
 
--- a/providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
+++ b/providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
@@ -72,17 +72,6 @@ class ExasolHook(DbApiHook):
         """
         Validate that query parameters are a dict, as required by pyexasol.
 
-        pyexasol's ``ExaConnection.execute``/``export_to_pandas``/``export_to_file`` have only ever
-        accepted a ``dict`` for query parameters (they are interpolated into the SQL via
-        ``str.format(**query_params)``, and pyexasol's own type stubs declare ``query_params: dict |
-        None``). Positional-style parameters (lists/tuples) were never actually supported at runtime,
-        even though this hook's signatures (inherited from
-        :class:`~airflow.providers.common.sql.hooks.sql.DbApiHook`) advertise the broader
-        ``Iterable | Mapping[str, Any] | None`` type. As of pyexasol 2.x, which ships a ``py.typed``
-        marker, mypy now enforces this at the type level too. We validate explicitly here so that
-        passing a non-dict fails fast with a clear message instead of an opaque error surfacing deep
-        inside pyexasol.
-
         :param parameters: The parameters as passed to this hook's public methods.
         :return: The same parameters, narrowed to ``dict[str, Any] | None`` for pyexasol/mypy.
         """
--- a/providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
+++ b/providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
@@ -67,21 +67,47 @@ class ExasolHook(DbApiHook):
         self.schema = kwargs.pop("schema", None)
         self._sqlalchemy_scheme = sqlalchemy_scheme
 
+    @staticmethod
+    def _validate_query_params(parameters: Iterable | Mapping[str, Any] | None) -> dict[str, Any] | None:
+        """
+        Validate that query parameters are a dict, as required by pyexasol.
+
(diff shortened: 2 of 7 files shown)


providers/exasol/docs/changelog.rst
```diff
@@ -27,6 +27,26 @@
...
+  * ``ExasolHook.get_records()`` and ``ExasolHook.get_first()`` now only accept a single SQL string,
+    not a list of statements (pyexasol's underlying ``execute()`` never actually supported a list; use
+    ``ExasolHook.run()`` for executing multiple statements in sequence).
+  * ``parameters``/``query_params`` passed to ``get_records()``, ``get_first()``, ``run()``, and
+    ``get_df()``/``get_pandas_df()`` must be a ``dict``/``Mapping``. Positional-style parameters
+    (lists/tuples) were never actually supported by pyexasol at runtime and will now raise a clear
+    ``TypeError`` instead of failing deep inside pyexasol (or silently doing the wrong thing).
+  * If you pass ``export_params={"with_column_names": ...}`` to ``ExasolHook.export_to_file()`` or
+    ``ExasolToS3Operator``, note that pyexasol 2.0.0 fixed this option to be interpreted as a real
+    boolean; previously any value (including ``False``) was treated as ``True``.
```
eladkal, line 48: The list note stuff but nothing here strike as airflow breaking change is it?
Can you just remove the breaking changes title?

also, did you validate these statements? This feels AI generated. Please provide evidences that the points here are valid. Just referencing to the upstream docs/code should be enough

baha-bouali: There are two things I can confirm with the changelog
* `with_column_names` boolean fix: I can confirm that pyexasol's changelog has a dedicated bugfix (https://github.com/exasol/pyexasol/releases#release-2.2.1 -   \# 265)). So any value used to get forced to be `True`. Now it should require a real Boolean

* `py.typed` marker: Confirmed. In the Same Changelog batch above (# 298)  pyexasol added it when swapping the deprecated `rsa` dependency for `cryptography`. I think that this is the actual mechanism not a breaking change.

However, 

* `get_records()`/`get_first()` only accepting a single SQL string. I cannot confirm this one tho. No changelog says explicitly this changed in 2.x. However, it's consistent with every example in pyexasol's docs since they only show a single string. This means that's always the case.

* Parameters should be `dict/Mapping`: Can't confirm either, but no changelog ref to point to.

So, two claims are true, two are unverified assumptions. 
I suggest to drop the "*Breaking changes*" title since it is not a change to the provider's own interface and drop the two unverifiable bullets entirely. Keep only the two we can back: 
```
This provider now requires ``pyexasol>=2.0.0,<3``. pyexasol 2.0.0 added a
``py.typed`` marker and fixed ``export_params={"with_column_names": ...}``
to require a real boolean (previously any value, including ``False``, was
treated as ``True`` — pyexasol changelog, bugfix #265).
```

providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
```diff
@@ -67,21 +67,36 @@ def __init__(self, *args, sqlalchemy_scheme: str | None = None, **kwargs) -> Non
...
 
+    @staticmethod
+    def _validate_query_params(parameters: Iterable | Mapping[str, Any] | None) -> dict[str, Any] | None:
+        """
+        Validate that query parameters are a dict, as required by pyexasol.
+
+        :param parameters: The parameters as passed to this hook's public methods.
+        :return: The same parameters, narrowed to ``dict[str, Any] | None`` for pyexasol/mypy.
+        """
+        if parameters is not None and not isinstance(parameters, dict):
```
uranusjr, line 78: This rejects any mapping _but_ a dict. This is probably not necessary?

providers/exasol/src/airflow/providers/exasol/hooks/exasol.py
```diff
@@ -184,11 +199,19 @@ def get_records(
...
+            :meth:`run` with ``handler=exasol_fetch_all_handler`` instead.
+        :param parameters: The parameters to render the SQL query with. Must be a ``dict``/``Mapping``;
+            pyexasol does not support positional (list/tuple) query parameters.
         """
-        with closing(self.get_conn()) as conn, closing(conn.execute(sql, parameters)) as cur:
+        if not isinstance(sql, str):
+            raise TypeError(
+                "ExasolHook.get_records() only supports a single SQL string, not a list of statements. "
+                "Use ExasolHook.run() with a handler for executing multiple statements."
+            )
```
uranusjr, line 212: This is duplicated too many times. I would refactor it into a private function called by each place. You can use `TypeGuard` to improve typing.

uranusjr:
It seems like the upgrade is quite breaking (a lot of old inputs are not rejected), so the provider will need a major version bump. If we’re bumping major, deprecated functions should also be removed.

eladkal:
> It seems like the upgrade is quite breaking (a lot of old inputs are not rejected), so the provider will need a major version bump. If we’re bumping major, deprecated functions should also be removed.

The breaking changes don't seem to be about Airflow interface (Hook/operator)
https://github.com/apache/airflow/pull/69695#discussion_r3597810778
did I miss anything in my earlier comment?

eladkal:
Fixed in https://github.com/apache/airflow/pull/72119

PR #69302, opened 2026-07-03
Send TS SDK logs to stderr on socket disconnect

guan404ming:
## Related Issue

closes: #69292

## Why

Logs emitted after an unexpected log socket close were silently lost. Since both socket ends always run on the same host, a disconnect means the peer died and reconnecting cannot help (see the discussion in #69696), so instead of buffering/reconnecting, later records fall back to stderr, which the coordinator already captures into the task log.

## How

- Warn once on an unexpected socket close and route subsequent log records to stderr
- No buffering or reconnect machinery

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Fable 5)

Generated-by: Claude Code (Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69696, opened 2026-07-10
Add log socket reconnections

guan404ming:
## Why

A task runtime that dropped its logs connection could never reconnect, so logs buffered during the outage were silently lost.

## How

- Register a non-blocking reaccepter on the logs server so runtimes can reconnect and flush buffered logs
- Reject reconnections not owned by the child process tree; transient socket errors no longer stall the supervisor loop
- Test reconnection forwarding, rejection paths, and spurious wakeups

---

### DIFF
--- a/task-sdk/src/airflow/sdk/coordinators/_subprocess.py
+++ b/task-sdk/src/airflow/sdk/coordinators/_subprocess.py
@@ -43,7 +43,13 @@ import structlog
 
 from airflow.sdk.configuration import conf
 from airflow.sdk.execution_time.coordinator import BaseCoordinator
-from airflow.sdk.execution_time.supervisor import ActivitySubprocess, NeverRaised, ProcessTracker
+from airflow.sdk.execution_time.supervisor import (
+    ActivitySubprocess,
+    NeverRaised,
+    ProcessTracker,
+    make_buffered_socket_reader,
+    process_log_messages_from_subprocess,
+)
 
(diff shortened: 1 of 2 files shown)


jason810496 (commented):
Another direction on this one. Do we really need to introduce the reconnecting for log socket at all?


I think it makes sense to introduce reconnecting/ retrying for those client-sever are on the different host machines.

Since the coordinator (Python task-SDK) and Lang-SDK (TS-SDK in this case) will always on the same worker and on the same host.


Would it be better to not reconnecting at all for both side, as the unexpected TCP connection close on Lang-SDK side means the serious errors from OS level (or SIGKILL etc) and remove the reconnecting part from TS-SDK (IIRC, Java and Go didn't implement the socket level reconnecting).

WDYT? Thanks!

guan404ming:
Agreed. since both sides always run on the same host by convention, a dropped socket means the process is already in a fatal state, so reconnecting doesn't buy us much. Closing this one.

One follow-up question: does it sound like a good idea to rework #69302 to treat a disconnect as fatal instead (e.g. falling back to stderr for later records, which the coordinator already captures)? Thanks!

jason810496:
> does it sound like a good idea to rework https://github.com/apache/airflow/pull/69302 to treat a disconnect as fatal instead (e.g. falling back to stderr for later records, which the coordinator already captures)? Thanks!

Yes, I think this approach makes more sense to me.

PR #69697, opened 2026-07-10
Pass dag_id to Dag bundle construction on task execution

ykuc:
Pass dag_id to Dag bundle construction when a worker runs a task.

Custom Dag bundles can now read self.dag_id on the bundle instance before initialize() runs. That lets bundle authors tailor setup per Dag (for example, per-Dag sparse checkout in Git) without changing the initialize() signature.

Changes:

Add optional dag_id to BaseDagBundle.__init__ (None by default).
Extend DagBundlesManager.get_bundle() with dag_id and pass it into the bundle constructor.
Pass what.ti.dag_id from task_runner.parse() on the worker task execution path.
Document self.dag_id in the custom Dag bundles section of dag-bundles.rst.
Not changed: Dag processor, CLI, callbacks, and BundleInfo / Execution API — there dag_id stays None.

Tests:

test_get_bundle — asserts dag_id is passed and defaults to None.
test_parse_passes_dag_id_to_get_bundle — asserts parse() calls get_bundle(..., dag_id=...).


airflow-core/src/airflow/dag_processing/bundles/manager.py
```diff
@@ -396,22 +396,31 @@ def _extract_template_params(bundle_instance: BaseDagBundle) -> dict:
...
         """
         cfg_bundle = self._bundle_config.get(name)
         if not cfg_bundle:
             raise ValueError(f"Requested bundle '{name}' is not configured.")
         return cfg_bundle.bundle_class(
-            name=name, version=version, version_data=version_data, **cfg_bundle.kwargs
+            name=name,
+            version=version,
+            version_data=version_data,
+            dag_id=dag_id,
```
ashb: Doing it this way will break every custom Dag Bundle that exists today as they aren't expecting a `dag_id` kwarg.

ashb (commented):
It's a little bit odd/sub-optimal, but this _might_ be available today via https://airflow.apache.org/docs/apache-airflow/stable/howto/dynamic-dag-generation.html#optimizing-dag-parsing-delays-during-execution:

```
from airflow.sdk import get_parsing_context

current_dag_id = get_parsing_context().dag_id
```

ykuc:
> It's a little bit odd/sub-optimal, but this might be available today via

That's actually what I'm using today, but it doesn't feel very reliable because it's not part of the public contract. I think this is a useful capability, so I'd prefer to make it an explicit, supported API rather than relying on the current implementation details.

airflow-core/src/airflow/dag_processing/bundles/manager.py
ykuc: I checked all built-in BaseDagBundle implementations currently shipped by Airflow: LocalDagBundle, GitDagBundle, S3DagBundle, and GCSDagBundle.

All of them accept **kwargs and forward them to BaseDagBundle.__init__, so none of the built-in bundles would break when dag_id is passed.

I agree that third-party bundles with a strict custom __init__ signature and no **kwargs would break. However, this does not affect every existing bundle, and all official implementations already follow the forwarding pattern

ykuc: we can make
```
class BaseDagBundle:
    def set_parsing_context(self, *, dag_id: str | None) -> None:
        self.dag_id = dag_id


bundle = bundle_class(
    name=name,
    version=version,
    version_data=version_data,
    **bundle_kwargs,
)

bundle.set_parsing_context(dag_id=dag_id)
```

ashb: > I checked all built-in BaseDagBundle implementations currently shipped by Airflow

This is a documented pluggable interface. We can't know what other people might be doing with custom implementations.

ashb: If we're going to go with something like `set_parsing_context` then we should just document and test that `get_parsing_context()` is the right way to go.

potiuk:
@ykuc This PR has been converted to **draft** because it does not yet meet our [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).

**Issues found:**
- :x: **Provider tests**: Failing: `Provider distributions tests / Compat 2.11.1:P3.10:`, `Provider distributions tests / Compat 3.0.6:P3.10:`, `Provider distributions tests / Compat 3.1.8:P3.10:`, `Provider distributions tests / Compat 3.2.2:P3.10:`, `Provider distributions tests / Compat 3.3.0:P3.10:`. Run `breeze testing providers-tests --test-type "Providers[<provider>]"` locally.
- :x: **Unresolved review comments**: There are 1 unresolved review thread(s) on this PR from @ashb. For each one: either apply the suggestion in a follow-up commit, or reply in-thread with a brief explanation of why the feedback doesn't apply, or resolve the thread if the feedback is no longer relevant. **Once you believe a thread is resolved — whether by pushing a fix or by explaining why the suggestion doesn't apply — it is the author's responsibility to mark it as resolved by clicking the 'Resolve conversation' button at the bottom of the thread.** Reviewers don't auto-close their own threads, so unresolved threads read as 'still waiting on the author' and block the PR from moving forward.

**What to do next:**

Converting a PR to draft is **not** a rejection — it is an invitation to bring the PR up to the project's standards so that maintainer review time is spent productively. There is no rush — take your time and work at your own pace. We appreciate your contribution and are happy to wait for updates. If you have questions, feel free to ask on the [Airflow Slack](https://s.apache.org/airflow-slack).

---

airflow-core/src/airflow/dag_processing/bundles/manager.py
ykuc: fixed

ykuc:
Hello @ashb! 
Could you recheck MR?

PR #69702, opened 2026-07-10
Fix partition key display and input handling

Lee-W:
### Why

The partition key handling in the UI had a few rough edges:

- A run or asset event with no partition key still rendered a dangling `Partition key:` label with nothing after it.
- Typing a partition key and then clearing it sent an empty string to the trigger / materialize APIs, which rejected it with a `400`.
- The trigger form always exposed the partition key field, even for Dags that are not partitioned, where it does nothing.
- The manual asset-event flow used a `JsonEditor` to enter the partition key, which is a plain string, not JSON.
- A created run labelled its partition key `Mapped Partition Key`. At the run level the value is the concrete key the run runs for, not a mapper-transformed key, so the `Mapped` qualifier was misleading.

### What

- Add a shared `toNullablePartitionKey` util that normalizes `undefined` and empty string to `null`, and use it at every call site that sends `partition_key` (trigger, manual asset event, materialize). This is what fixes the empty-key `400`.
- `Header` and `AssetEvent` no longer render a partition-key line when the value is null/undefined.
- Label created runs as `Partition Key` (was `Mapped Partition Key`). `Mapped Partition Key` is intentionally kept where a partition mapper derives a downstream key from upstream asset key(s); it is not touched here.
- The trigger form only shows the partition key field for partitioned Dags, with an `Optional - only applies to partitioned Dags` helper text.
- The manual asset-event partition key uses a plain text `Input` instead of the `JsonEditor`.

- 

---

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
+++ b/airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
@@ -95,13 +95,16 @@ export const AssetEvent = ({
       <HStack>
         <TriggeredRuns dagRuns={event.created_dagruns} />
       </HStack>
-      {event.partition_key === undefined || event.partition_key === null ? undefined : (
-        <HStack>
-          <Text>
-            {rootTranslate("dagRun.partitionKey")}: {event.partition_key}
-          </Text>
-        </HStack>
-      )}
+      {
+        // eslint-disable-next-line no-eq-null, eqeqeq
+        event.partition_key == null ? undefined : (
+          <HStack>
+            <Text>
+              {rootTranslate("dagRun.partitionKey")}: {event.partition_key}
+            </Text>
+          </HStack>
+        )
+      }
       {Object.keys(extra).length >= 1 ? <RenderedJsonField collapsed content={extra} /> : undefined}
     </Box>
   );
--- a/airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
+++ b/airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
@@ -95,7 +95,7 @@ export const AssetEvent = ({
       <HStack>
         <TriggeredRuns dagRuns={event.created_dagruns} />
       </HStack>
-      {event.partition_key === undefined ? undefined : (
+      {event.partition_key === undefined || event.partition_key === null ? undefined : (
         <HStack>
           <Text>
             {rootTranslate("dagRun.partitionKey")}: {event.partition_key}
--- a/airflow-core/src/airflow/ui/src/pages/DagRuns/DagRuns.tsx
+++ b/airflow-core/src/airflow/ui/src/pages/DagRuns/DagRuns.tsx
@@ -19,11 +19,10 @@
 import { Flex, HStack, Text, useDisclosure } from "@chakra-ui/react";
 import type { ColumnDef } from "@tanstack/react-table";
 import type { TFunction } from "i18next";
-import { useRef } from "react";
 import { useTranslation } from "react-i18next";
 import { useParams, useSearchParams } from "react-router-dom";
 
-import { useDagRunServiceGetDagRuns, useDagServiceGetDagDetails } from "openapi/queries";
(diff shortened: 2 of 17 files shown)


airflow-core/src/airflow/ui/src/queries/useTrigger.ts
```diff
@@ -108,7 +108,10 @@ export const useTrigger = ({ dagId, onSuccessConfirm }: { dagId: string; onSucce
         data_interval_start: formattedDataIntervalStart,
         logical_date: formattedLogicalDate,
         note: checkNote,
-        partition_key: dagRunRequestBody.partitionKey ?? null,
+        partition_key:
+          dagRunRequestBody.partitionKey === undefined || dagRunRequestBody.partitionKey === ""
```
bbovenzi: We do this check a few times. Let's extract it into a util so its easier to not repeat the `?? null` bug again

airflow-core/src/airflow/ui/src/pages/Run/Header.tsx
```diff
@@ -75,7 +75,7 @@ export const Header = ({ dagRun }: { readonly dagRun: DAGRunResponse }) => {
             ? []
             : [
                 {
-                  label: translate("dagRun.mappedPartitionKey"),
```
bbovenzi, line 78: We also call `mappedPartitionKey` in `PartitionScheduleModal` do we want to keep that?

airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
```diff
@@ -95,7 +95,7 @@ export const AssetEvent = ({
       <HStack>
         <TriggeredRuns dagRuns={event.created_dagruns} />
       </HStack>
-      {event.partition_key === undefined ? undefined : (
+      {event.partition_key === undefined || event.partition_key === null ? undefined : (
```
bbovenzi:
```suggestion
      {event.partition_key ? undefined : (
```

or `{event.partition_key == null ? undefined : (` with an eslint ignore comment?

airflow-core/src/airflow/ui/src/queries/useTrigger.ts
Lee-W: Extracted into `toNullablePartitionKey` in `src/utils/partitionKey.ts`, now shared by `useTrigger.ts` and both call sites in `CreateAssetEventModal.tsx`.

airflow-core/src/airflow/ui/src/pages/Run/Header.tsx
Lee-W: Yes. These 2 are not exactly the same. In `PartitionScheduleModal`, it's the downstream key the partition mapper produced by transforming upstream asset partition key(s), so it can differ from the upstream key, which caused confusion in dev call. For here, `parition_key` is just the key that runs this Dag run.

airflow-core/src/airflow/ui/src/components/Assets/AssetEvent.tsx
Lee-W: Went with  `== null` version + `// eslint-disable-next-line no-eq-null, eqeqeq`. If I'm not mistaken, `event.partition_key ?` does thing in the wrong direction

airflow-core/src/airflow/ui/src/pages/DagRuns/DagRuns.tsx
bbovenzi, line 1: Let's remove the changes to deciding when to show and not show columns from this PR. I think we need a more robust approach instead of this manual ref+key manipulation that will cause the dag runs table to re-mount. Also, we probably want to reuse this functionality if we get it to work right. Like hiding `map_index` for non-mapped tasks.

Lee-W: Sure, just removed. It no longer fetches Dag details to toggle the column.

bbovenzi (approved):
Thanks!

Lee-W:
Thanks, Brent!

PR #69705, opened 2026-07-10
Add JitteredCronTimetable for deterministic schedule jitter

Pebble32:
Add `JitteredCronTimetable`, an opt-in `CronTriggerTimetable` subclass that shifts each DAG's fire time by a deterministic, per-DAG offset drawn from `[0, max_jitter)`. This spreads out DAGs that share a cron expression so they no longer all fire at the same instant, without changing `logical_date` / `data_interval` semantics.

### Why

`@daily` expands to `0 0 * * *`, so **every** daily DAG in a deployment is scheduled at exactly midnight. In large deployments this "thundering herd" at the cron boundary overloads the scheduler and workers — enough to cause task failures when dozens of DAGs are born at the same instant.

The existing ways to deal with this don't actually de-collide the schedule:

| Existing option | Why it doesn't solve the problem |
| --- | --- |
| Hand-pick a unique minute per DAG | Manual, doesn't scale, drifts and re-collides as DAG count grows, and throws away the `@daily` intent |
| Hash the DAG id into a literal cron string | Same loss of intent; not reusable across DAGs/teams; every author re-implements it ad hoc |
| Pools / concurrency limits (`parallelism`, `max_active_tasks_per_dag`, pool slots) | These **limit or queue execution** — the runs are still *scheduled* at the same instant. They cap contention downstream but never spread the fire times apart at the source. |

`JitteredCronTimetable` is the only approach that moves the fire times themselves, deterministically: the same `seed` (e.g. the DAG id) always maps to the same offset, so runs stay stable and predictable across scheduler restarts and timetable serialization. Motivated by the discussion in https://github.com/apache/airflow/discussions/69027.

### What

- New `JitteredCronTimetable(CronTriggerTimetable)` in both the Task SDK (author-facing, attrs-based) and airflow-core (scheduler-side), plus the serialization wiring (`BUILTIN_TIMETABLES` mapping, `serialize`/`deserialize`, encode/decode across the SDK↔core boundary).
- Two extra kw-only params on top of `CronTriggerTimetable`: `seed: str` and `max_jitter: timedelta`.
- The offset is `md5(seed) % max_jitter.total_seconds()`, applied as a "strip → cron → apply" coordinate shift so cron/DST alignment is fully delegated to the parent and only the wall-clock fire time is shifted.
- **Fully opt-in and safe by default**: with the defaults (`seed=""`, `max_jitter=timedelta(0)`) the offset is zero and it behaves *identically* to `CronTriggerTimetable`. Nothing changes for anyone who doesn't use it.

### Tests

`airflow-core/tests/unit/timetables/test_jittered_cron_timetable.py`, modeled on the existing `test_trigger_timetable.py`:

- jittered runs equal base cron runs shifted by the fixed offset, across a catchup sequence including a DST spring-forward (`America/New_York`);
- zero `max_jitter` reproduces `CronTriggerTimetable` exactly (catchup on/off);
- offsets are deterministic for a given seed, bounded to `[0, max_jitter)`, and spread across distinct seeds;
- `serialize`/`deserialize` round-trips seed + window + derived offset;
- full encode → decode round-trip across the SDK/core layers rebuilds the core class.

---

Generated-by: Claude (Claude Code), following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

related: https://github.com/apache/airflow/discussions/69027

### DIFF
--- a/airflow-core/src/airflow/timetables/trigger.py
+++ b/airflow-core/src/airflow/timetables/trigger.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import datetime
 import functools
+import hashlib
 import math
 import operator
 import time
@@ -598,3 +599,83 @@ class CronPartitionTimetable(CronTriggerTimetable):
             suffix = f"{suffix}__{partition_key}"
         suffix = f"{suffix}__{get_random_string()}"
         return run_type.generate_run_id(suffix=suffix)
+
+
+class JitteredCronTimetable(CronTriggerTimetable):
+    """
+    A :class:`CronTriggerTimetable` that offsets each run by a deterministic, per-DAG jitter.
+
+    Behaves exactly like ``CronTriggerTimetable`` but shifts every fire time by a fixed offset
+    derived from ``seed`` and spread across ``[0, max_jitter)``. This avoids the "thundering
+    herd" where many DAGs sharing a cron expression (e.g. ``@daily`` -> ``0 0 * * *``) all fire
+    at the same instant and overload the scheduler and workers at that boundary.
+
+    The offset is deterministic: the same ``seed`` always maps to the same offset, so runs are
+    stable and predictable across scheduler restarts and timetable serialization. DAGs with
+    different seeds land in different slots; collisions are possible but harmless (two DAGs
+    sharing one minute still beats every DAG firing at once).
+
+    All ``data_interval`` and ``logical_date`` semantics are inherited from
+    ``CronTriggerTimetable`` -- only the wall-clock fire time is shifted.
+
+    :param cron: cron string that defines the base schedule.
+    :param timezone: timezone used to interpret the cron string.
+    :param interval: timedelta that defines the data interval start (see ``CronTriggerTimetable``).
+    :param run_immediately: see ``CronTriggerTimetable``.
+    :param seed: stable, unique-per-DAG string that determines the offset. The DAG id is a
+        natural choice.
+    :param max_jitter: upper bound of the jitter window; the offset falls in ``[0, max_jitter)``.
+        Keep it smaller than the gap to the next cron boundary so the shift cannot push a run's
+        ``logical_date`` across a day or period boundary.
+    """
+
+    def __init__(
+        self,
+        cron: str,
+        *,
+        timezone: str | Timezone | FixedTimezone,
+        interval: datetime.timedelta | relativedelta = datetime.timedelta(),
+        run_immediately: bool | datetime.timedelta = False,
+        seed: str,
+        max_jitter: datetime.timedelta,
+    ) -> None:
+        super().__init__(cron, timezone=timezone, interval=interval, run_immediately=run_immediately)
+        h = int(hashlib.md5(seed.encode()).hexdigest(), 16)
+        self._offset = (
+            datetime.timedelta(seconds=h % int(max_jitter.total_seconds()))
+            if max_jitter > datetime.timedelta(0)
+            else datetime.timedelta(0)
+        )
+        self._seed = seed
+        self._max_jitter = max_jitter
+
+    def _apply(self, t: DateTime) -> DateTime:
+        return t + self._offset
+
+    def _strip(self, t: DateTime) -> DateTime:
+        return t - self._offset
+
+    def _get_next(self, current: DateTime) -> DateTime:
+        return self._apply(super()._get_next(self._strip(current)))
+
+    def _get_prev(self, current: DateTime) -> DateTime:
+        return self._apply(super()._get_prev(self._strip(current)))
+
+    def serialize(self) -> dict[str, Any]:
+        data = super().serialize()
+        data["seed"] = self._seed
+        data["max_jitter"] = self._max_jitter.total_seconds()
+        return data
+
+    @classmethod
+    def deserialize(cls, data: dict[str, Any]) -> Timetable:
+        from airflow.serialization.decoders import decode_interval, decode_run_immediately
+
+        return cls(
+            data["expression"],
+            timezone=parse_timezone(data["timezone"]),
+            interval=decode_interval(data["interval"]),
+            run_immediately=decode_run_immediately(data.get("run_immediately", False)),
+            seed=data["seed"],
+            max_jitter=datetime.timedelta(seconds=data["max_jitter"]),
+        )
--- a/airflow-core/src/airflow/timetables/trigger.py
+++ b/airflow-core/src/airflow/timetables/trigger.py
@@ -15,10 +15,10 @@
 # specific language governing permissions and limitations
 # under the License.
 from __future__ import annotations
+from airflow.utils.hashlib_wrapper import md5
 
 import datetime
 import functools
-import hashlib
 import math
 import operator
 import time
@@ -615,8 +615,10 @@ class JitteredCronTimetable(CronTriggerTimetable):
     different seeds land in different slots; collisions are possible but harmless (two DAGs
     sharing one minute still beats every DAG firing at once).
 
-    All ``data_interval`` and ``logical_date`` semantics are inherited from
-    ``CronTriggerTimetable`` -- only the wall-clock fire time is shifted.
+    The offset shifts every fire time, and because a run's ``logical_date`` and ``data_interval``
+    derive from the fire time (exactly as in ``CronTriggerTimetable``), they shift by the same
+    offset. Keep ``max_jitter`` well within the gap to the next cron boundary so a run's
+    ``logical_date`` cannot cross a day or period boundary.
 
     :param cron: cron string that defines the base schedule.
     :param timezone: timezone used to interpret the cron string.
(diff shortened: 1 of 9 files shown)


airflow-core/src/airflow/timetables/trigger.py
```diff
@@ -598,3 +599,83 @@ def generate_run_id(
...
+        timezone: str | Timezone | FixedTimezone,
+        interval: datetime.timedelta | relativedelta = datetime.timedelta(),
+        run_immediately: bool | datetime.timedelta = False,
+        seed: str,
+        max_jitter: datetime.timedelta,
+    ) -> None:
+        super().__init__(cron, timezone=timezone, interval=interval, run_immediately=run_immediately)
+        h = int(hashlib.md5(seed.encode()).hexdigest(), 16)
+        self._offset = (
+            datetime.timedelta(seconds=h % int(max_jitter.total_seconds()))
```
kaxil: The guard is `max_jitter > timedelta(0)` but the divisor is `int(max_jitter.total_seconds())`, so any `0 < max_jitter < 1s` passes the guard and then hits modulo by zero. `max_jitter=timedelta(milliseconds=500)` raises `ZeroDivisionError` here at construction. The `int()` also silently drops sub-second precision, so `timedelta(seconds=1.5)` truncates to a 1s window and the offset can never exceed it. Flooring the guard at 1s, or dropping the `int()` truncation, closes both.

airflow-core/src/airflow/timetables/trigger.py
```diff
@@ -598,3 +599,83 @@ def generate_run_id(
...
+        cron: str,
+        *,
+        timezone: str | Timezone | FixedTimezone,
+        interval: datetime.timedelta | relativedelta = datetime.timedelta(),
+        run_immediately: bool | datetime.timedelta = False,
+        seed: str,
+        max_jitter: datetime.timedelta,
+    ) -> None:
+        super().__init__(cron, timezone=timezone, interval=interval, run_immediately=run_immediately)
+        h = int(hashlib.md5(seed.encode()).hexdigest(), 16)
```
kaxil: `hashlib.md5(seed.encode())` called directly raises on FIPS-enabled Python builds. The hash here isn't security-sensitive, so `airflow.utils.hashlib_wrapper.md5` (which passes `usedforsecurity=False`) keeps it working on FIPS, the same way `serialized_dag.py` and `dagcode.py` already do. This runs scheduler-side on every deserialize, so on a FIPS build every DAG using this timetable would fail to load.

airflow-core/src/airflow/timetables/trigger.py
```diff
@@ -598,3 +599,83 @@ def generate_run_id(
...
+    herd" where many DAGs sharing a cron expression (e.g. ``@daily`` -> ``0 0 * * *``) all fire
+    at the same instant and overload the scheduler and workers at that boundary.
+
+    The offset is deterministic: the same ``seed`` always maps to the same offset, so runs are
+    stable and predictable across scheduler restarts and timetable serialization. DAGs with
+    different seeds land in different slots; collisions are possible but harmless (two DAGs
+    sharing one minute still beats every DAG firing at once).
+
+    All ``data_interval`` and ``logical_date`` semantics are inherited from
+    ``CronTriggerTimetable`` -- only the wall-clock fire time is shifted.
```
kaxil: "Only the wall-clock fire time is shifted" reads as if `logical_date` stays pinned to the cron boundary, but it doesn't. With the default `interval=0`, `next_dagrun_info` returns `DagRunInfo.interval(next_start_time, next_start_time)` where `next_start_time` is the jittered time, so `logical_date` (= `data_interval.start`) moves by the offset too. The `max_jitter` caveat further down ("the shift cannot push a run's `logical_date` across a day or period boundary") already assumes this, so the two statements contradict each other. Worth rewording to say the offset shifts `logical_date`/`data_interval` as well, or deciding whether `logical_date` should stay pinned to the boundary.

task-sdk/src/airflow/sdk/definitions/timetables/trigger.py
```diff
@@ -86,6 +86,38 @@ class CronTriggerTimetable(CronMixin, BaseTimetable):
...
+        natural choice.
+    :param max_jitter: upper bound of the jitter window; the offset falls in ``[0, max_jitter)``.
+        Keep it smaller than the gap to the next cron boundary so the shift cannot push a run's
+        ``logical_date`` across a day or period boundary.
+
+    See :class:`CronTriggerTimetable` for ``cron``, ``timezone``, ``interval`` and
+    ``run_immediately``.
+    """
+
+    seed: str = attrs.field(kw_only=True, default="")
```
kaxil, line 117: With `seed` defaulting to `""`, a user who sets `max_jitter` but forgets `seed` gets the same `md5("")` offset for every DAG, so they all fire at the same shifted instant -- the herd just moves off the cron boundary instead of spreading out, silently defeating the point of the feature. The no-op default only makes sense when `max_jitter=0` too. Could this raise (or warn) when `max_jitter > 0` and `seed == ""`? Note the core `__init__` already makes `seed` required, so the two classes disagree on this today.

airflow-core/src/airflow/timetables/trigger.py
Pebble32: Went with rewording over pinning logical_date. It derives from the fire time in `CronTriggerTimetable`, so shifting it with the offset is intended behavior, not a bug to fix. Doc now says `logical_date`/`data_interval` shift with the offset. Let me know if pinning is preferred

airflow-core/src/airflow/timetables/trigger.py
Pebble32: Switched to hashlib_wrapper.md5 (FIPS-safe)

airflow-core/src/airflow/timetables/trigger.py
Pebble32: Now computed in microseconds so sub-second windows work, no zero-division. Added a test since I felt like this is something that I should have tested for before

task-sdk/src/airflow/sdk/definitions/timetables/trigger.py
Pebble32: This is a very good catch didn't think about it thanks! Now raises in both SDK and core when `max_jitter` > 0 and `seed` is empty, also aligned the defaults. Added a test for this

uranusjr:
I wonder if it’d make sense to add this functionality directly to CronMixin (and therefore inherited by all cron-based scheduling). Some scheduling tools (Jenkins IIRC) have this built-in for cron scheduling (opt-in) to not let a large number of jobs spiking a server periodically if timing is not essential.

Implementation seems reasonable to me in general.

Pebble32:
Thanks @uranusjr! The Jenkins H comparison is exactly the idea

I kept it as its own timetable to keep the change contained. Moving it into `CronMixin` means every cron timetable (data interval, multiple cron, partition) also needs its serialization and `logical_date` behavior worked out, so the scope grows quite a bit.

Would you prefer I land this focused version first and generalize into `CronMixin` as a follow up, or go straight for CronMixin now? Happy either way, just trying to keep the PR reviewable <3

raphaelauv:
hey about "enough to cause task failures"

if it can happen then it will happen again ( backfill , new dags , big clear ... )

your tasks fail because you did not put correct/perfect limit on concurrency.

I know it's not easy : 

yes airflow pools are too simple for many use-case

yes airflow is not kubernetes ressource aware ( he do not know your max hardware scaling limits and don't queue tasks regarding this ) 

---

"but never spread the fire times apart at the source."

I've a stack triggering more than a thousand of dag_run at midnight , yes it take almost 30 seconds to the scheduler to do so , but no errors , did you encounter a scheduler error ?

Pebble32:
Fair point, and well put. Where I still think it earns its place is as peak shaving on top of concurrency limits, not instead of them: spreading arrivals so fewer tasks land in the pool at the same instant, so a fixed pool saturates less often. Opt in, for when exact timing does not matter. Same idea as Jenkins H that @uranusjr mentioned. Maybe it is better to move it directly to CronMixin

uranusjr:
Can you explore this locally to see how intrusive adding this would be? If implementation becomes messy, I think it’s reasonable to do this in a separate timetable in this PR first with the intention to eventually refactor the logic into CronMixin before 3.4.0 is released, which is quite still some time away.

Pebble32:
Sounds good. I will explore the `CronMixin` version locally and report back on how intrusive it looks.

Pebble32:
Explored it and it seems less intrusive than I first thought. The offset moves into `CronMixin` cleanly, and on the sdk side the fields are inherited by all cron timetables. The rest is mechanical, forwarding seed and max_jitter through the core constructors and the per class serialize, deserialize and encoder variants.

For CronDataIntervalTimetable the shared offset shifts both interval bounds and the fire time by the same amount, so the window stays one full period long but sits offset from the cron line (00:35 to 00:35 instead of 00:00 to 00:00), and consecutive runs stay contiguous so no gaps or overlaps. I will go with that uniform shift as the default and document it, keeping the guidance to set max_jitter small relative to the period. Jittering only the fire time while keeping the interval on the cron boundary is possible but needs a targeted override, so I will leave it as a possible follow up.

Will put up the CronMixin version.

Pebble32:
Follow up moving the jitter into CronMixin as discussed: #72475. Will close this PR once that one lands.

Pebble32:
@uranusjr the CronMixin version we discussed is up in #72475 and CI is green. Would appreciate a look whenever you have a moment. Thanks!

PR #69712, opened 2026-07-10
Support Paramiko 5 in SSH and SFTP providers

shivaam:
Follow-up to #69669, after SSH and SFTP provider 6.0.0 released Paramiko 4 support.

## What

This removes the temporary `paramiko<5.0.0` cap from the SSH and SFTP providers so they can resolve Paramiko 5 while continuing to support Paramiko 4.

It also keeps RSA private-key validation working with Paramiko 5 by signing the local validation probe with `rsa-sha2-512` instead of Paramiko's removed legacy `ssh-rsa` SHA-1 signing path.

## Compatibility

This is not a new breaking provider API change: the providers still support Paramiko 4, and no Airflow connection option or supported key type is removed by this PR.

Paramiko 5 itself removed RSA SHA-1 `ssh-rsa` signatures, SHA-1 key exchange algorithms, and GSSAPI. Removing the upper bound allows an environment to resolve Paramiko 5, so users whose servers still require those legacy algorithms can retain Paramiko 4 with an environment constraint such as `paramiko<5` until the server is upgraded.

RSA keys remain supported. During a real connection, Paramiko negotiates a modern RSA signature algorithm such as `rsa-sha2-256` or `rsa-sha2-512` with the server.

## Validation

- `uv lock --check`
- `git diff --check`
- `uv run pytest providers/ssh/tests/unit/ssh/hooks/test_ssh.py -k 'private_key or host_key or dss_host_key or unsupported_host_key' -q` (27 passed)
- `uv run pytest providers/ssh/tests/unit/ssh/hooks/test_ssh_async.py providers/sftp/tests/unit/sftp/hooks/test_sftp.py -k 'host_key or private_key or connection' -q` (27 passed)
- `uv run --with paramiko==4.0.0 pytest providers/ssh/tests/unit/ssh/hooks/test_ssh.py::TestSSHHook::test_openssh_private_key -q`
- `uv run ruff format --check providers/ssh/src/airflow/providers/ssh/hooks/ssh.py`
- `uv run ruff check providers/ssh/src/airflow/providers/ssh/hooks/ssh.py`
- `uv run mypy --config-file pyproject.toml providers/ssh/src/airflow/providers/ssh/hooks/ssh.py`
- touched-file `prek` checks
- live Airflow `SSHHook` connection with Paramiko 5 and an RSA private key against the Breeze OpenSSH server, negotiating `rsa-sha2-512` and executing a remote command

### DIFF
--- a/providers/sftp/README.rst
+++ b/providers/sftp/README.rst
@@ -57,7 +57,7 @@ PIP package                                 Version required
 ``apache-airflow``                          ``>=2.11.0``
 ``apache-airflow-providers-ssh``            ``>=6.0.0``
 ``apache-airflow-providers-common-compat``  ``>=1.12.0``
-``paramiko``                                ``>=4.0.0,<5.0.0``
+``paramiko``                                ``>=4.0.0``
 ``asyncssh``                                ``>=2.12.0; python_version < "3.14"``
 ``asyncssh``                                ``>=2.22.0; python_version >= "3.14"``
 ==========================================  ======================================
--- a/providers/sftp/docs/index.rst
+++ b/providers/sftp/docs/index.rst
@@ -104,7 +104,7 @@ PIP package                                 Version required
 ``apache-airflow``                          ``>=2.11.0``
 ``apache-airflow-providers-ssh``            ``>=6.0.0``
 ``apache-airflow-providers-common-compat``  ``>=1.12.0``
-``paramiko``                                ``>=4.0.0,<5.0.0``
+``paramiko``                                ``>=4.0.0``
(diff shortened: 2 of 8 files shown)


eladkal:
Correct me if I am wrong but from 4->5 this is not a breaking change in terms of Airflow operators right?

shivaam:
> Correct me if I am wrong but from 4->5 this is not a breaking change in terms of Airflow operators right?

Yes, it is not a breaking change anymore. I will update the pr

eladkal (approved):

Issue #64212, opened 2026-03-25
Webserver timeout randomly


Jeoffreybauvin:
### Apache Airflow version

3.1.8

### What happened?

I have a dag scheduled every 10 minutes. The 3rd one always fails .... 

<img width="242" height="192" alt="Image" src="https://github.com/user-attachments/assets/3175f032-6678-406d-93a0-33761264ace0" />

On worker side I can see some timeouts : 

```
2026-03-25T10:30:01.055984Z [debug    ] send_request_body.complete     [httpcore.http11] loc=_trace.py:47
2026-03-25T10:30:01.056166Z [debug    ] receive_response_headers.started request=<Request [b'GET']> [httpcore.http11] loc=_trace.py:47
--- Supervised process Last chance exception handler ---
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1112, in run
    result = _execute_task(context=context, ti=ti, log=log)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1523, in _execute_task
    result = ctx.run(execute, context=context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/operator.py", line 417, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/decorator.py", line 252, in execute
    return_value = super().execute(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/operator.py", line 417, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/standard/operators/python.py", line 228, in execute
    return_value = self.execute_callable()
                   ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #69713, opened 2026-07-10
Preserve AirflowTaskTimeout when timeout logging fails

Andrushika:
Preserve AirflowTaskTimeout when timeout logging fails
---

### Why
`handle_timeout()` logs a timeout message before raising `AirflowTaskTimeout`. The task runtime sends those logs through a pipe handled by the supervisor. If that pipe has been closed (for example, because the supervisor has exited), the logging call can raise `BrokenPipeError` instead.
This masks the original timeout, causing the task to fail with a logging error rather than `AirflowTaskTimeout`. This exact sequence appears in the production traceback in #64212.

### What
Suppress `OSError` raised while logging from `handle_timeout()` so a failed log write cannot mask the timeout. `handle_timeout()` will still raise `AirflowTaskTimeout` after a logging failure. Non-OSError exceptions from the logging stack are deliberately left out of scope.

related: #64212 

Generated-by: Claude Code Opus 4.7 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/task-sdk/src/airflow/sdk/execution_time/timeout.py
+++ b/task-sdk/src/airflow/sdk/execution_time/timeout.py
@@ -16,6 +16,7 @@
 # under the License.
 from __future__ import annotations
 
+import contextlib
 import os
 
 import structlog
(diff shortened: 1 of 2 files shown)


Dev-iL (approved):
Generally LGTM. The one suggestion I have is: it's unnecessary to add the full explanation of why the `suppress` is needed into the source code - readers interested in more context can go read the PR. I would replace the multiline comment with something simpler like "Ensure the timeout isn't masked by a logging error" or "Ensure the timeout gets reported" or even "Disregard logging errors {so timeout surfaces}" - you get the idea.

Andrushika:
> it's unnecessary to add the full explanation of why the `suppress` is needed into the source code - readers interested in more context can go read the PR.

Fixed, thanks for reviewing!

Andrushika:
Hi @amoghrajesh, could you please take a look when you have time?
It's a one line change plus test, thanks a lot!

PR #69728, opened 2026-07-10
UI: Add timetable type filter to Dag list

### LINKED ISSUE #53738: Add Support for Filtering DAGs by Timetable Type
### Description

Add the ability to filter DAGs based on their Timetable type (e.g., CronTriggerTimetable, AssetOrTimeSchedule) in both the UI and public DAG APIs.

## Proposed Solution

Add a new query parameter `timetable_type` to the DAG listing endpoints that allows filtering DAGs based on their timetable type.

### Use case/motivation

Currently, while the DAG model stores timetable information in the `timetable_summary` field, there is no API support to filter DAGs based on their timetable type. Users cannot easily find all DAGs that use a specific type of timetable. This functionality is crucial to fix #53041. Once this is implemented, the UI filter functionality will be built on top of this. 

### Related issues

- #53041 

### Code of Conduct

shivaam:
Adds timetable type filtering to the Dag list UI.

The public API already supports filtering Dags by `timetable_type`, but the Dag list page uses the private `/ui/dags` endpoint. This wires the same filter through the private endpoint, generated UI client, and Dag list filters.

The timetable filter uses a discoverable, server-backed typeahead and supports selecting multiple timetable types with OR semantics. The timetable and tag controls use a small shared filter-select component so their chips, clearing, option handling, and styling stay consistent. Other compatible filters can migrate to this shared component incrementally, giving filter controls consistent behavior across Airflow UI pages without broadening this PR into an unrelated UI rewrite.

closes: #53738

Testing:
- Focused backend timetable filter and type-list tests
- Dag filter UI tests
- ESLint and TypeScript
- Production UI build
- Real Airflow/Postgres UI verification with built-in and plugin timetables

---

image

- [X] Yes: OpenAI Codex

Generated-by: OpenAI Codex following the guidelines


airflow-core/src/airflow/api_fastapi/core_api/routes/ui/dags.py
```diff
@@ -124,6 +124,10 @@ def get_dags(
     is_favorite: QueryFavoriteFilter,
     has_asset_schedule: QueryHasAssetScheduleFilter,
     asset_dependency: QueryAssetDependencyFilter,
+    timetable_type: Annotated[
+        FilterParam[list[str] | None],
+        Depends(filter_param_factory(DagModel.timetable_type, list[str], FilterOptionEnum.IN)),
```
bbovenzi, line 138: This is an exact match. To make this a more useful feature I think we need to follow how tags work and create an endpoint to query all timetables that exist amongst all the dags to give a typeahead dropdown UX

pierrejeambrun (commented):
Any update here? It would be great to have this improvement!

shivaam:
> Any update here? It would be great to have this improvement!

Hello! I will update the PR by tomorrow morning. Working on it today

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/dags.py
shivaam: Image

A new search bar only for tags? Is this what we are expecting from this UI? 
I feel the UI is looking a little cluttered with multiple search bars for each different category.
@pierrejeambrun fyi

shivaam:
https://github.com/user-attachments/assets/884a1937-8188-4208-9b38-79cc3155e2ee



Here is an example of how it will look like. If you agree with the design, I will move ahead with the implementation. I have fixed the favorites to the right as we resize the window so they dont move around on smaller screens

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/dags.py
bbovenzi: Yes, this is starting to get cluttered. I think our original intention was to make the long "search dags" input handle many cases similar to how github work. Maybe we need to go back to that.

bbovenzi:
If we are filtering by timetable then I think we should also show timetable with the schedule in the card and list. Can be a separate PR.

airflow-core/newsfragments/69728.feature.rst
```diff
@@ -0,0 +1 @@
+Add timetable type filtering to the Dags list.
```
bbovenzi: This is not newsfragment-worthy

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/dags.py
shivaam: Do you think consolidating these filters into the main Dag search could be handled in a separate PR, or would you prefer that we address it here?

I think we can do a unified, GitHub-style search. It will benefit from a few design iterations to define the filter syntax, dropdown behavior, and how the existing filters should migrate. I’m happy to explore that separately while keeping this PR focused on timetable filtering.

bbovenzi (approved):

bbovenzi:
LGTM let's just rerun the autogenerated API docs. Its probably from my PR to shorten all of the string match descriptions.

PR #69729, opened 2026-07-10
Add upgrade-fab-provider skill and FAB contributing doc

potiuk:
Adds the `upgrade-fab-provider` agent skill and a `providers/fab/CONTRIBUTING.rst` developer doc.

Upgrading the vendored Flask-AppBuilder dependency is deceptively involved — the FAB provider vendors and subclasses large parts of FAB's security manager, so a bump must be reconciled against that vendored code (enforced by `test_fab_alignment.py`), and the generated provider README must be re-synced. The skill captures the verified end-to-end procedure (including the gotcha that a green alignment test does not prove a security fix inside a vendored method was handled); the contributing doc points contributors at it and records past bumps.

The skill is vendor-neutral in `.agents/skills/upgrade-fab-provider` with `.claude/` and `.github/` relay symlinks, matching the repository's existing skill layout. It was authored and validated against the companion FAB 5.2.1 -> 5.2.2 bump.

---

### DIFF
--- a/.agents/skills/upgrade-fab-provider/SKILL.md
+++ b/.agents/skills/upgrade-fab-provider/SKILL.md
@@ -1,0 +1,270 @@
+---
+name: upgrade-fab-provider
+description: >
+  Upgrade the pinned Flask-AppBuilder (FAB) dependency in the Apache Airflow
+  FAB provider (`providers/fab/`). Bumps the exact `flask-appbuilder==` pin and
+  its mirror constant, regenerates `uv.lock`, drives the `test_fab_alignment.py`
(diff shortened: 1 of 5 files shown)


vincbeck (approved):
LGTM, as we use it and encounter new cases, we'll iterate. Example: if FAB makes some modifications in models, we might need to update the models we vendor-in in `providers/fab/src/airflow/providers/fab/auth_manager/models/__init__.py`. But again, we could add that when that happens

shahar1 (approved):

eladkal (approved):

PR #69731, opened 2026-07-10
SnowflakeCortexAgentHook: Set stream=False for Cortex Agent requests

SameerMesiah97:
**Description**

This change fixes the `SnowflakeCortexAgentHook` by explicitly setting `stream=False` when invoking the Cortex Agent API, ensuring the hook receives a JSON response.

**Rationale**

The Cortex Agent API returns a Server-Sent Events (SSE) stream by default. Since the hook expects a JSON response, explicitly disabling streaming restores the expected behaviour.

**Tests**

Updated the unit tests to verify that `stream=False` is included in all Cortex Agent requests.


justinpakzad (approved):

SameerMesiah97:
@vincbeck 

This is the fix for that bug I identified in the most recent release wave for the providers.

vincbeck (approved):

henry3260 (approved):

PR #69732, opened 2026-07-10
Show prepare breeze timing in CI duration alerts

gopidesupavan:
CI duration alerts currently identify slower jobs but do not surface whether
the setup phase accounts for the change. Showing the prepare breeze step timing
makes common CI image/cache slowdowns visible without adding noisy per-step
analysis for every job.

image

The first job spent most of its extra time preparing the image, and in some cases that includes downloading the stash. Adding the Prepare breeze & CI image timing makes that setup cost visible directly in the duration alert.

 

---

- codex

---

### DIFF
--- a/scripts/ci/analyze_ci_job_durations.py
+++ b/scripts/ci/analyze_ci_job_durations.py
@@ -66,8 +66,15 @@ import subprocess
 import sys
 from datetime import datetime
 from pathlib import Path
+from typing import TypedDict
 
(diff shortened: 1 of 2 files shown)


gopidesupavan:
it produces now something like this:

image

potiuk (approved):

potiuk:
Let's see - sometimes that duration will be longer - mostly when we need to rebuild the image. I think we should simply exclude the image build time and track it separately.,

PR #69734, opened 2026-07-10
Honor execution timeout across task deferrals

GautamSharma99:
PR body

  Deferred tasks could exceed their configured `execution_timeout` because every resumed execution received the full timeout again. Their deferred trigger
  deadline also ignored the task execution deadline.

  This change:

  - Sets the deferred deadline to the earlier of `trigger_timeout` and the task's execution deadline.
  - Applies only the remaining execution time when a deferred task resumes.
  - Immediately raises `AirflowTaskTimeout` when the timeout elapsed while the task was deferred.
  - Adds regression coverage for execution-only, trigger-first, execution-first, and already-expired cases.

  Tests:

  - Execution API task instance tests: 190 passed, 4 skipped
  - Task SDK execute callable tests: 8 passed
  - Execution API suite: 521 passed, 4 skipped
  - Ruff, formatting, mypy, and prek checks passed

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @GautamSharma99** · by `@potiuk` · 2026-07-28 16:32 UTC
>
> Some review feedback from `kaxil` is waiting on you:
> - There is **1 unresolved review thread** on this PR from `kaxil`.
>
> **The ball is in your court** — you've been assigned to this PR. Push a fix or reply in each thread explaining why the feedback does not apply, then mark them resolved and ping the reviewer (`kaxil`) for a final look.
>
>
> **Note:** your branch is **386 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -677,6 +677,15 @@ def _create_ti_state_update_query_and_update_state(
         if ti_patch_payload.trigger_timeout is not None:
             timeout = timezone.utcnow() + ti_patch_payload.trigger_timeout
 
+        ti = session.get(TI, task_instance_id)
+        if ti is not None and ti.start_date is not None:
+            dag = dag_bag.get_dag_for_run(dag_run=ti.dag_run, session=session)
+            if dag is not None:
+                with contextlib.suppress(TaskNotFound):
+                    if execution_timeout := dag.get_task(ti.task_id).execution_timeout:
+                        execution_deadline = ti.start_date + execution_timeout
+                        timeout = min(timeout, execution_deadline) if timeout else execution_deadline
+
         trigger_kwargs = ti_patch_payload.trigger_kwargs
         if not isinstance(trigger_kwargs, str):
             # If it's passed as a string, assume the client encrypted it, otherwise assume it doesn't need to
@@ -692,9 +701,6 @@ def _create_ti_state_update_query_and_update_state(
         trigger_row.encrypted_kwargs = trigger_kwargs
         session.add(trigger_row)
         session.flush()
-
-        # TODO: HANDLE execution timeout later as it requires a call to the DB
-        # either get it from the serialised DAG or get it from the API
 
         query = update(TI).where(TI.id == task_instance_id)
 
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -677,6 +677,15 @@ def _create_ti_state_update_query_and_update_state(
         if ti_patch_payload.trigger_timeout is not None:
             timeout = timezone.utcnow() + ti_patch_payload.trigger_timeout
 
+        ti = session.get(TI, task_instance_id)
+        if ti is not None and ti.start_date is not None:
+            dag = dag_bag.get_dag_for_run(dag_run=ti.dag_run, session=session)
```
kaxil, line 682: This deserializes the full SerializedDAG on every deferral just to read one task attribute (`execution_timeout`). `ti_update_state` with `state=deferred` is a hot path -- every deferrable-sensor poke and every deferrable-operator await hits it.

The sibling helper `_handle_fail_fast_for_dag` in this same file deliberately avoids exactly this:

```python
# Check fail_fast from DagModel (simple column lookup) - early exit if False
# This avoids loading 5-50 MB SerializedDAG in 99% of cases
fail_fast = session.scalar(select(DagModel.fail_fast).where(DagModel.dag_id == dag_id))
if not fail_fast:
    return
# Only load SerializedDAG when fail_fast=True (rare case ~1%)
ser_dag = dag_bag.get_dag_for_run(dag_run=dr, session=session)
```

That guard only works because `fail_fast` is a `DagModel` column. `execution_timeout` is task-level and lives only in the serialized blob, so there's no cheap column to gate on -- every deferral here pays the full deserialization unconditionally, even for the tasks (the majority) that set no `execution_timeout`.

The worker already has the value with no DB hit: `_run_execute_callable` reads `task.execution_timeout`, and the worker already ships `trigger_timeout=defer.timeout` in the defer payload (`task_runner.py:1494` -> `TIDeferredStatePayload.trigger_timeout`). Compute the execution deadline (or the remaining budget) on the worker at defer time and send it in the payload the way `trigger_timeout` already flows; the server then does `min(...)` with zero deserialization.

Tradeoff: a new payload field needs an Execution API version bump (Cadwyn), while the server-side load doesn't touch the wire. But a 5-50 MB deserialization on every defer isn't worth saving a versioned field, and it regresses the fail-fast optimization this file just added. The extra `session.get(TI, ...)` full-ORM reload (the route only selected the columns it needed) folds into the same fix.

kaxil (changes requested):

PR #69735, opened 2026-07-10
Document commit SHA support for GitDagBundle tracking_ref

coleheflin:
Clarifies that `GitDagBundle.tracking_ref` supports a branch, tag, or full commit SHA — it already works with a SHA in practice, but the provider docs only described branch/tag. Adds a commit-SHA example to the bundle docs, a note on static-ref refresh behavior (a SHA-pinned bundle doesn't pick up new commits via `refresh_interval`; promoting/rolling back requires updating `tracking_ref` in the bundle config), and test coverage for SHA checkout, promotion, and rollback.

Verified manually against a live repository, and confirmed during review: rolling back a SHA-pinned `tracking_ref` to an already-local commit works reliably after a Dag processor restart. Promoting to a genuinely new SHA can fail on restart if the bundle's local storage survives it (the default) — `_initialize()` checks out the new ref before the working clone has fetched it, so the checkout fails until that storage is cleared (fresh pod, or deleted bundle directory). That limitation is tracked at #71388; the docs and tests here describe the current behavior accordingly rather than promising promotion just works with a restart. Also scoped the restart note to mention that with `disable_bundle_versioning`, workers resolve code from their own `tracking_ref` and need the updated config too, not just the Dag processor.

---

- [X] Yes — Otto (Astronomer AI assistant, built on pi)

Generated-by: Otto following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

🤖 Generated by [Otto](https://astronomer.io/otto)

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @coleheflin** · by `@potiuk` · 2026-07-28 16:10 UTC
>
> - :x: **Other failing CI checks**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).
>
>
>
> **Note:** your branch is **345 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>

### DIFF
--- a/providers/git/docs/bundles/index.rst
+++ b/providers/git/docs/bundles/index.rst
@@ -64,8 +64,3 @@ without a restart. Tags and commit SHAs are static (assuming tags aren't moved),
 to known-good code — but promoting or rolling back a SHA means changing ``tracking_ref`` in
 ``dag_bundle_config_list`` itself, which requires restarting the Dag processor and workers to take
 effect.
-
-This is different from Dag bundle versioning: each Dag run already records the commit its bundle
-resolved to and can be rerun with that code regardless of the current ``tracking_ref`` (see
-:ref:`rerun_with_latest_version <config:core__rerun_with_latest_version>`). That only affects
-individual reruns, not which commit future runs use.
--- a/providers/git/docs/bundles/index.rst
+++ b/providers/git/docs/bundles/index.rst
@@ -70,3 +70,9 @@ dag_bundle_config_list`` config change like any other: it takes effect only afte
 and workers are restarted to reload the configuration. This is unlike moving a branch/tag ref, where
 the config is unchanged and the existing ``refresh_interval`` polling picks up the new commit with no
 restart required.
+
+This is also distinct from Dag bundle versioning. Each Dag run already records the commit its bundle
+resolved to, so it can be rerun with that exact code regardless of the bundle's current
+``tracking_ref`` (see :ref:`rerun_with_latest_version <config:core__rerun_with_latest_version>`).
+That only affects individual reruns; it does not change which commit new Dag parses and runs use.
+Promoting or rolling back the active environment still requires changing ``tracking_ref`` itself.
--- a/providers/git/docs/bundles/index.rst
+++ b/providers/git/docs/bundles/index.rst
@@ -59,20 +59,13 @@ bundle to that exact commit:
      }
     ]'
 
-Branches move as new commits are pushed, which is useful when combined with ``refresh_interval`` to
-pick up new changes automatically without any restart. Tags and commit SHAs point to a fixed commit
-(assuming tags are not moved), so they're useful for pinning a bundle to known-good code.
+Branches move as new commits are pushed, so combined with ``refresh_interval`` they pick up new code
+without a restart. Tags and commit SHAs are static (assuming tags aren't moved), pinning the bundle
+to known-good code — but promoting or rolling back a SHA means changing ``tracking_ref`` in
+``dag_bundle_config_list`` itself, which requires restarting the Dag processor and workers to take
+effect.
 
-When ``tracking_ref`` is a commit SHA, new commits pushed to the repository are not picked up by
-``refresh_interval``, since a SHA always refers to the same commit. Promoting or rolling back to a
-different SHA means changing the ``tracking_ref`` value itself, which is an ``[dag_processor]
-dag_bundle_config_list`` config change like any other: it takes effect only after the Dag processor
-and workers are restarted to reload the configuration. This is unlike moving a branch/tag ref, where
-the config is unchanged and the existing ``refresh_interval`` polling picks up the new commit with no
-restart required.
-
-This is also distinct from Dag bundle versioning. Each Dag run already records the commit its bundle
-resolved to, so it can be rerun with that exact code regardless of the bundle's current
-``tracking_ref`` (see :ref:`rerun_with_latest_version <config:core__rerun_with_latest_version>`).
-That only affects individual reruns; it does not change which commit new Dag parses and runs use.
-Promoting or rolling back the active environment still requires changing ``tracking_ref`` itself.
+This is different from Dag bundle versioning: each Dag run already records the commit its bundle
+resolved to and can be rerun with that code regardless of the current ``tracking_ref`` (see
+:ref:`rerun_with_latest_version <config:core__rerun_with_latest_version>`). That only affects
+individual reruns, not which commit future runs use.
--- a/providers/git/docs/bundles/index.rst
+++ b/providers/git/docs/bundles/index.rst
@@ -62,5 +62,4 @@ bundle to that exact commit:
 Branches move as new commits are pushed, so combined with ``refresh_interval`` they pick up new code
 without a restart. Tags and commit SHAs are static (assuming tags aren't moved), pinning the bundle
 to known-good code — but promoting or rolling back a SHA means changing ``tracking_ref`` in
-``dag_bundle_config_list`` itself, which requires restarting the Dag processor and workers to take
-effect.
+``dag_bundle_config_list`` itself, which requires restarting the Dag processor to take effect.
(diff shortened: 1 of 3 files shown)


RNHTTR (approved):

coleheflin:
@potiuk this is ready for review, CI is fully green now!

One piece of feedback on the triage flow: the maintainer note was added directly into the PR description rather than as a comment, so I didn't get a GitHub notification and didn't realize there was action needed on my end until I checked back in. Posting maintainer triage notes as a regular PR comment (with a tag) would make sure they surface in notifications/email inbox. Thanks for the help getting this triaged!

providers/git/docs/bundles/index.rst
```diff
@@ -41,3 +41,25 @@ Example of using the GitDagBundle:
...
+             "tracking_ref": "a3d1850dd1aa1919a61620aa39f202185c9321c0",
+             "subdir": "dags"
+         }
+     }
+    ]'
+
+Branches move as new commits are pushed, so combined with ``refresh_interval`` they pick up new code
+without a restart. Tags and commit SHAs are static (assuming tags aren't moved), pinning the bundle
+to known-good code — but promoting or rolling back a SHA means changing ``tracking_ref`` in
+``dag_bundle_config_list`` itself, which requires restarting the Dag processor to take effect.
```
kaxil, line 65: Promoting forward doesn't take effect on a restart alone if the bundle storage path survives that restart, which is the default. `_initialize` runs `self.repo.git.checkout(self.tracking_ref)` ([git.py:215](https://github.com/apache/airflow/blob/6af6a6a6bfe216618b0af4171bd5f1b10c83d14e/providers/git/src/airflow/providers/git/bundles/git.py#L215)) before the working repo has fetched anything: `_fetch_bare_repo` updates the bare mirror, but `tracking_repo` is only fetched inside `refresh()`, which runs after that checkout. A `tracking_repo` cloned while the old SHA was HEAD has no objects for the new one, so `initialize()` raises `GitCommandError: fatal: reference is not a tree: <new sha>`.

I ran this against a real bundle: promoting to a commit created after the first `initialize()` fails, promoting to a tag created after it fails the same way (`pathspec 'v2' did not match`), rolling back to an older SHA works because the objects are already local, and a fresh storage path works. Since `dag_bundle_storage_path` defaults to `/tmp/airflow/dag_bundles`, a plain Dag processor restart on the same host keeps the stale clone and stays broken until the bundle directory is removed. On Kubernetes it depends on whether that path sits on a volume.

Could you scope this sentence, so it doesn't promise that promoting works after a restart? Something like: rollback works, promoting needs the bundle storage cleared (fresh pod, or deleting the bundle dir). The real fix is fetching before the checkout in `_initialize`, which deserves its own PR rather than blocking this one.

kaxil, line 65: Is the Dag processor enough in every case? With `[dag_processor] disable_bundle_versioning` (or `DAG(disable_bundle_versioning=True)`) no `bundle_version` is stored on the dag run, so the worker calls `DagBundlesManager().get_bundle(name, version=None)` ([task_runner.py:1017](https://github.com/apache/airflow/blob/6af6a6a6bfe216618b0af4171bd5f1b10c83d14e/task-sdk/src/airflow/sdk/execution_time/task_runner.py#L1017)) and resolves code from its own `tracking_ref`. A worker still holding the old SHA in its `dag_bundle_config_list` would run the old commit. The PR description's original wording ("Dag processor and workers") covered that, so maybe scope this to versioning being enabled and mention the worker case.

providers/git/tests/unit/git/bundles/test_git.py
```diff
@@ -701,6 +701,46 @@ def test_refresh_tag_moved_forward_and_backward(self, mock_githook, git_repo):
...
+        """
+        repo_path, repo = git_repo
+        mock_githook.return_value.repo_url = repo_path
+        first_commit = repo.head.commit
+
+        file_path = repo_path / "new_test.py"
+        with open(file_path, "w") as f:
+            f.write("hello world")
+        repo.index.add([file_path])
+        second_commit = repo.index.commit("Another commit")
```
kaxil, line 720: `second_commit` is created before the first `initialize()`, so both the bare mirror and the `tracking_repo` clone already contain it, and the promote step never exercises what a real config change hits. Moving these four lines below the first `bundle.initialize()` turns it into a genuine promote, and the test then fails with `fatal: reference is not a tree: <sha>` from the `checkout` in `_initialize`. Worth doing even though it goes red, because the docstring says this mirrors "how a bundle config change is applied in practice" and as written it doesn't. See my note on the docs paragraph for the mechanism.

providers/git/docs/bundles/index.rst
coleheflin: Confirmed the mechanism — traced through `_initialize`/`_clone_repo_if_required`/`_fetch_bare_repo`/`refresh()`, and reproduced the failure with a unit test exercising that same code path (real git subprocess calls against a local temp repo, not a live deployment): moving the promoted commit's creation to after the first `initialize()` makes the second `initialize()` raise `GitCommandError: fatal: unable to read tree (<sha>)` (same failure you found, slightly different git-version wording than "reference is not a tree"), while rollback to an already-local commit still succeeds.

Reworded the doc note to say exactly what you suggested — rollback is reliable, promotion needs the bundle's local storage cleared (fresh pod, or deleted bundle dir) — and opened https://github.com/apache/airflow/issues/71388 to track the actual fetch-before-checkout fix, linked from both the doc note and a comment at the `checkout` call in `_initialize`.

---
Drafted-by: Claude Code (Sonnet 5); reviewed by @coleheflin before posting

coleheflin: Good catch, added back. The note now says that with `[dag_processor] disable_bundle_versioning` (or the per-Dag parameter) set, workers resolve code from their own `tracking_ref` rather than a recorded bundle version, so they need the updated config too — not just the Dag processor.

---
Drafted-by: Claude Code (Sonnet 5); reviewed by @coleheflin before posting

providers/git/tests/unit/git/bundles/test_git.py
coleheflin: Good catch, moved the new-commit creation so it happens after the first `initialize()`. Split the old test into three: a rollback case (still green — objects are already local), a promote-without-clearing-storage case that now asserts the actual `GitCommandError` this raises today (documents the limitation tracked in #71388, with a note to flip it to asserting success once that's fixed), and a promote-with-fresh-storage case (different bundle name → fresh local storage) that asserts promotion succeeds once storage doesn't carry the stale clone. All tests in the file still pass (105, after rebasing onto current `main`).

---
Drafted-by: Claude Code (Sonnet 5); reviewed by @coleheflin before posting

potiuk (approved):
LGTM. Confirming the three points from the July review all landed, since they were raised by someone else and it should be on record that they were checked rather than assumed:

* **Promotion vs. restart** — the mechanism is real, and you didn't take it on faith: `_initialize` checks out before the working clone has fetched, so a commit created after that clone was populated isn't there yet. The docs note now draws the distinction correctly (rollback reliable, promotion needs storage cleared), and the limitation is tracked in #71388 with the full URL both in the docs and as a comment at the code site — which is the convention for deferred work, so the next person reading `git.py` finds the follow-up without leaving the file.
* **Dag processor vs. workers** — restored, and scoped properly: with `[dag_processor] disable_bundle_versioning` or the per-Dag parameter, workers resolve from their own `tracking_ref` and do need the updated configuration.
* **The test that didn't exercise a real promote** — now three tests that each mean something distinct.

One thing I want to be explicit about rather than leave as a silent pass: the middle test asserts the `GitCommandError` that promotion raises *today*. Pinning current broken behaviour in a test is usually something I'd push back on, but here it is the right call and worth keeping — it is labelled as a known limitation, tied to #71388, and carries instructions to flip it to asserting success once the fix lands. That turns the limitation into something CI will tell us about when it changes, instead of a paragraph in the docs that quietly goes stale.

The git-version alternation in `match="reference is not a tree|unable to read tree"` is fine — it is deliberate and explained.

---

> will follow up.*
>

---

PR #69739, opened 2026-07-11
Add validate trigger events in dbt cloud deferrable tasks

takayoshi-makabe:
`DbtCloudRunJobOperator.execute_complete` and `DbtCloudJobRunSensor.execute_complete` used the trigger event without validating it:

- **`event=None`**: happens when a deferred task resumes without a payload — a custom or subclassed trigger that yields `TriggerEvent(None)`, or an event payload lost between the triggerer and the worker. Both methods crashed with an opaque `TypeError: 'NoneType' object is not subscriptable`.
- **Unrecognised `status`**: happens on triggerer/worker version skew (a task defers under one provider version and resumes under another whose trigger emits a different status vocabulary) or with a custom trigger. `DbtCloudRunJobTrigger` emits `success`/`cancelled`/`error`/`timeout`, but both `execute_complete` methods branch deny-list style (only specific failure statuses raise, everything else succeeds), so an unknown status fell through into the success path — the task was marked SUCCESS even though the dbt Cloud job outcome was unknown.
- **The sensor additionally never checked `timeout` at all** — it only raised on `error`/`cancelled`, so a job run that timed out while deferred was silently marked SUCCESS.

This adds a `validate_execute_complete_event()` helper (with a dedicated `DbtCloudTriggerEventException`), applied at the top of both `execute_complete` methods. Same fix shape as the anthropic (#69379) and OpenAI (#69506) providers: the `None` check mirrors the amazon provider's helper of the same name, and the status check plays the role amazon's call sites cover with their `if validated_event["status"] != "success": raise` allow-list branching.

The sensor's fix also replaces a bare `raise AirflowException()` (no message) with `DbtCloudJobRunException(event["message"])`, which both fixes the missing `timeout` handling and surfaces the trigger's actual message instead of an empty exception.

The operator keeps its existing per-status branching (`cancelled`/`error`/`timeout` each have distinct messages, and `timeout` additionally attempts job cancellation) since `validate_execute_complete_event()` already closes the status set to the four values above.

* related: #69379

---

Claude code

---

### DIFF
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
@@ -138,6 +138,32 @@ class DbtCloudResourceLookupError(AirflowException):
     """Exception raised when a dbt Cloud resource cannot be uniquely identified."""
 
 
+class DbtCloudTriggerEventException(AirflowException):
+    """Raised when a deferred task resumes with a missing or malformed trigger event."""
+
+
+#: Statuses the provider's trigger emits in its terminal event.
+TRIGGER_EVENT_STATUSES = frozenset({"success", "cancelled", "error", "timeout"})
+
+
+def validate_execute_complete_event(event: dict[str, Any] | None = None) -> dict[str, Any]:
+    """
+    Validate the event a deferred task resumes with, returning it if well-formed.
+
+    The event crosses the triggerer/worker boundary through the metadata DB, so a
+    resuming task can receive ``None`` (e.g. a lost payload) or a status its handlers
+    do not recognize (triggerer/worker version skew, a custom trigger). Both must fail
+    loudly rather than silently falling through to the success path.
+    """
+    if event is None:
+        raise DbtCloudTriggerEventException("Trigger error: event is None")
+    if event.get("status") not in TRIGGER_EVENT_STATUSES:
+        raise DbtCloudTriggerEventException(
+            f"Unexpected trigger event status {event.get('status')!r}: {event!r}"
+        )
+    return event
+
+
 T = TypeVar("T", bound=Any)
 
 
--- a/generated/known_airflow_exceptions.txt
+++ b/generated/known_airflow_exceptions.txt
@@ -190,7 +190,6 @@ providers/databricks/tests/unit/databricks/hooks/test_databricks_base.py::1
 providers/datadog/src/airflow/providers/datadog/hooks/datadog.py::2
 providers/datadog/src/airflow/providers/datadog/sensors/datadog.py::1
 providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py::3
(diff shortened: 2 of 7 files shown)


providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
```diff
@@ -138,6 +138,32 @@ class DbtCloudResourceLookupError(AirflowException):
...
+    do not recognize (triggerer/worker version skew, a custom trigger). Both must fail
+    loudly rather than silently falling through to the success path.
+    """
+    if event is None:
+        raise DbtCloudTriggerEventException("Trigger error: event is None")
+    if event.get("status") not in TRIGGER_EVENT_STATUSES:
+        raise DbtCloudTriggerEventException(
+            f"Unexpected trigger event status {event.get('status')!r}: {event!r}"
+        )
+    return event
```
aaron-y-chen, line 164: I think we also need to check whether "run_id" and "message" exist in the event. Otherwise, it would fail at `operators/dbt.py:290` (`event["run_id"]`) and `sensors/dbt.py:131-133` (`event["message"]`).

aaron-y-chen (commented):
Hi, thanks for the contribution :)

providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
takayoshi-makabe: @nailo2c 
Good catch, but I kept the scope aligned with the sibling fixes in anthropic (#69379) and amazon, which also only validate None/status and not individual payload keys. A missing run_id/message would still fail loudly with a KeyError rather than silently succeeding, which is the failure mode this PR is actually guarding against.

aaron-y-chen (approved):
Overall LGTM.

nit: Title typo: `tirgger` -> `trigger`

takayoshi-makabe:
@aaron-y-chen 
Sorry, I fixed title.

aaron-y-chen:
No need to apologize. Thanks for your hard work! 😄

providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
```diff
@@ -138,6 +138,32 @@ class DbtCloudResourceLookupError(AirflowException):
     """Exception raised when a dbt Cloud resource cannot be uniquely identified."""
 
 
+class DbtCloudTriggerEventException(AirflowException):
+    """Raised when a deferred task resumes with a missing or malformed trigger event."""
+
+
+#: Statuses the provider's trigger emits in its terminal event.
+TRIGGER_EVENT_STATUSES = frozenset({"success", "cancelled", "error", "timeout"})
```
josh-fell: Would be nice if there was a general enum for this within TriggerEvent, but alas, no dice.

josh-fell (approved):

PR #69740, opened 2026-07-11
Kafka: Return consumed message results on request

### LINKED ISSUE #56494: airflow.providers.apache.kafka ConsumeFromTopicOperator cannot be used for pipelining because it returns None.
### Description

As of Airflow 3.0.6, the ConsumeFromTopicOperator takes the name+path to a function and applies it to all messages in a batch, but the results of this function call are discarded and the method returns None.  Rather than dropping the result of the `apply_function`, the results of the apply_function could be accumulated and returned so that other tasks could use the results.

### Use case/motivation

It would be nice to be able to pipeline operators and functions like we can with other Python tasks and other Operators.  An example:

```python
def my_service_prefilter(message):
    # This will get messages from my-topic-1 and my-topic-2.  We will extract and return some dicts.
    if message and message.value():
        decoded = json.loads(message.value().decode("utf-8")
        if decoded['some_value']:
            return decoded
    return None

@task.python
def handle_batch(decoded_items: list[dict]):
    for item in decoded_items:
        if item:
            do_something_with_item(item)

message_batch = ConsumeFromTopicOperator(
    task_id="consume_pending_kafka_messages",
    topics=["my-topic-1", "my-topic-2", ],
    apply_function="my_service.prefilter_function",
)

message_batch >> handle_batch
```

Written differently:

Kafka message stream -> ConsumeFromTopicOperator -> [message returned by the apply_function * batch size] -> whatever downstream task.

I initially wrote this as a bug report after spending a while with a colleague who had treated the ConsumeFromTopicOperator like other operators, but realized that returning None is the intended behavior.  

I recognize that the Airflow documentation explicitly states that Airflow is _not_ made to be used in the streaming data sense, but in this case if we need to transform some data and distribute it to other tasks we need to build our own operators instead of just reusing the existing Operators.  The Operator documentation seems to suggest that Operators be composable.

### Related issues

_No response_

### Code of Conduct

Vamsi-klu:
## Problem

`ConsumeFromTopicOperator` always returned `None`, even when its per-message `apply_function` produced values. That made it awkward to use the operator in downstream task pipelines that expect the consumed and processed values as the task return value.

closes: #56494

## What Changed

This adds an opt-in `return_apply_function_results` flag. The default behavior remains unchanged: the operator returns `None`.

When the flag is enabled with per-message `apply_function`, the operator collects non-`None` return values in consume order and returns them as a list after commits and consumer close handling. `apply_function_batch` behavior is intentionally unchanged.

## Impact

Kafka consume tasks can now feed processed message results directly into downstream tasks when the caller opts in. Existing users are unaffected because the new behavior is disabled by default, and batch processing keeps its current return behavior.

## Testing

- `uv run ruff format providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py providers/apache/kafka/tests/unit/apache/kafka/operators/test_consume.py`
- `uv run ruff check --fix providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py providers/apache/kafka/tests/unit/apache/kafka/operators/test_consume.py`
- `UV_CACHE_DIR=/tmp/uv-cache uv run --project providers/apache/kafka pytest providers/apache/kafka/tests/unit/apache/kafka/operators/test_consume.py -xvs`
- Commit hooks for provider files
- `git diff --check`

The tests cover the problem statement expectations: default `None` return, opt-in result collection, filtering `None`, preserving consume order across batches, and leaving `apply_function_batch` return behavior unchanged.

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/providers/apache/kafka/docs/operators/index.rst
+++ b/providers/apache/kafka/docs/operators/index.rst
@@ -33,6 +33,10 @@ For parameter definitions take a look at :class:`~airflow.providers.apache.kafka
     If you set the ``commit_cadence`` parameter, ensure that the ``enable.auto.commit`` option in the Kafka connection configuration is explicitly set to ``false``.
     By default, ``enable.auto.commit`` is ``true``, which causes the consumer to auto-commit offsets every 5 seconds, potentially overriding the behavior defined by ``commit_cadence``.
 
+Set ``return_apply_function_results=True`` to return a list containing each non-``None`` value returned by the per-message ``apply_function`` in consume order.
+This option does not apply to ``apply_function_batch``.
+Returned values use normal task return handling and may be stored in XCom, so avoid returning large payloads.
+
 Using the operator
 """"""""""""""""""
 
--- a/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
+++ b/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
@@ -47,6 +47,9 @@ class ConsumeFromTopicOperator(BaseOperator):
         defaults to None
     :param return_apply_function_results: Whether to collect non-None return values from the per-message
         ``apply_function`` and return them as a list. This option does not apply to ``apply_function_batch``.
+        ``None`` results are dropped, so the returned list does not align positionally with the consumed
+        messages. Results are returned through normal task return handling and may be stored in XCom,
+        so avoid enabling this for large result sets.
     :param commit_cadence: When consumers should commit offsets ("never", "end_of_batch","end_of_operator"),
         defaults to "end_of_operator";
         if end_of_operator, the commit() is called based on the max_messages arg. Commits are made after the
--- a/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
+++ b/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py
@@ -45,6 +45,8 @@ class ConsumeFromTopicOperator(BaseOperator):
     :param apply_function_args: Additional arguments that should be applied to the callable, defaults to None
     :param apply_function_kwargs: Additional key word arguments that should be applied to the callable
         defaults to None
+    :param return_apply_function_results: Whether to collect non-None return values from the per-message
+        ``apply_function`` and return them as a list. This option does not apply to ``apply_function_batch``.
     :param commit_cadence: When consumers should commit offsets ("never", "end_of_batch","end_of_operator"),
         defaults to "end_of_operator";
         if end_of_operator, the commit() is called based on the max_messages arg. Commits are made after the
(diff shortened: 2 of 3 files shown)


potiuk:
Nicely scoped. Defaulting to `False` means existing Dags are untouched, gating the return on both the flag and `apply_function` avoids surprising anyone using `apply_function_batch`, and the tests cover the cases that matter — default-returns-`None`, plus a combined test that filters `None` results and preserves consume order against a mock consumer that drains progressively rather than handing everything back at once.

I went looking for the memory and XCom hazard here, since this operator can run with `read_to_end=True` and no `max_messages`, and you had already covered it in the docs:

> Returned values use normal task return handling and may be stored in XCom, so avoid returning large payloads.

One thing before merge: that warning only exists in `docs/operators/index.rst`, not in the parameter docstring:

```
:param return_apply_function_results: Whether to collect non-None return values from the per-message
    ``apply_function`` and return them as a list. This option does not apply to ``apply_function_batch``.
```

The API reference is where someone looks when deciding whether to switch a flag on, and that is exactly the moment the size caveat matters. Could you add a clause there — something like "Results are returned through normal task return handling and may be stored in XCom, so avoid enabling this for large result sets"?

While you are in there, it would be worth noting in the same docstring that `None` results are dropped, so the returned list does not line up positionally with the consumed messages. It is implied by "non-None return values", but a caller trying to map results back to offsets could reasonably miss it.

---

Vamsi-klu:
@potiuk  Thanks for the careful review. Both caveats are now in the parameter docstring as of b90bc955: the XCom warning ("Results are returned through normal task return handling and may be stored in XCom, so avoid enabling this for large result sets") and an explicit note that None results are dropped, so the returned list does not align positionally with the consumed messages. I kept the wording in line with the operator guide so the two stay consistent.

Vamsi-klu:
Both requested docstring caveats are present on the current head `bbfce564`: the XCom size warning and the note that `None` results are dropped, so the returned list does not align positionally with consumed messages. The wording matches the operator guide, all checks are successful, neutral, or skipped, and GitHub reports the branch mergeable. Ready for another look whenever convenient.

---

potiuk (approved):
LGTM — both caveats I asked for are on the current head, and the wording lines up with the operator guide so the two won't drift.

What I re-checked before approving:

* collection is gated on the flag *inside* the loop, so nothing accumulates when the feature is off — the memory cost is genuinely opt-in even with `read_to_end=True`;
* the return is gated on the flag **and** `apply_function`, so `apply_function_batch` users keep their current behaviour, and there is a test asserting exactly that;
* `create_mock_kafka_consumer_from_messages` drains progressively rather than handing back everything in one call, which is what makes the order-and-filter assertion meaningful rather than trivially true.

Thanks for the careful follow-through on this one.

---

> will follow up.*
>

---

PR #69742, opened 2026-07-11
Restrict MSGraph pagination nextLink to the configured host

Samin061:
paginated_run follows the @odata.nextLink from each Microsoft Graph response and refetches it with the connection's bearer token attached, while Kiota's allowed_hosts defaults to empty (any host) when no authority is set, so a tampered response can send the token to an arbitrary host. Pin follow-up pagination requests to the configured endpoint's host and refuse a nextLink that points elsewhere.


providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -629,6 +629,13 @@ async def paginated_run(
 
         responses: list[dict] = []
 
+        # The pagination link (e.g. ``@odata.nextLink``) is echoed from the API response and is
+        # re-fetched with the connection's bearer token attached. Kiota only scopes that token to
+        # ``allowed_hosts``, which defaults to empty (any host) unless configured, so a tampered
+        # response could redirect the token off-host. Pin follow-up requests to the configured
+        # endpoint's host (CWE-918).
+        allowed_netloc = urlparse((await self.get_async_conn()).base_url).netloc
```
dabla: I don't think `allowed_netloc` should be initialized here. The variable itself makes sense, but it should be computed inside the `async run()` method where it's actually used. Initializing it here unnecessarily couples this code to the connection lookup and may perform work earlier than needed.

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
Samin061: Good point, moved it inside run() so the lookup only happens when pagination actually runs.

potiuk:
@Samin061 — Some review feedback from `@dabla` is waiting on you: 1 unresolved review thread(s). Could you either push a fix or reply in each thread explaining why the feedback doesn't apply? When you believe the feedback is addressed, please mark the threads as resolved and ping the reviewer (`@dabla`) for a final look. Thanks!

---

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
dabla, line 500:
```suggestion
        # The pagination link (e.g. ``@odata.nextLink``) is echoed from the API response and is
        # re-fetched with the connection's bearer token attached. Kiota only scopes that token to
        # ``allowed_hosts``, which defaults to empty (any host) unless configured, so a tampered
        # response could redirect the token off-host. Pin follow-up requests to the configured
        # endpoint's host (CWE-918).
        self.allowed_netloc = urlparse(request_adapter.base_url).netloc
        
        return request_adapter
```

dabla, line 223:
```suggestion
        self.api_version = self.resolve_api_version_from_value(api_version)
        self.allowed_netloc: str | None = None
```

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -633,6 +633,13 @@ async def run(
             url: str = "",
             query_parameters: dict[str, Any] | None = None,
         ):
+            # The pagination link (e.g. ``@odata.nextLink``) is echoed from the API response and is
+            # re-fetched with the connection's bearer token attached. Kiota only scopes that token to
+            # ``allowed_hosts``, which defaults to empty (any host) unless configured, so a tampered
+            # response could redirect the token off-host. Pin follow-up requests to the configured
+            # endpoint's host (CWE-918).
+            allowed_netloc = urlparse((await self.get_async_conn()).base_url).netloc
+
```
dabla: Remove those lines, we only want to resolve `allowed_netloc` once and don't want to do unnecessary lookups at each run once it's resolved

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -660,6 +667,16 @@ async def run(
                             data=data,
                             responses=lambda: responses,
                         )
+                        if (
+                            next_url
+                            and next_url.startswith("http")
+                            and urlparse(next_url).netloc != allowed_netloc
```
dabla:
```suggestion
                            and urlparse(next_url).netloc != self.allowed_netloc
```

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -660,6 +667,16 @@ async def run(
...
+                        ):
+                            raise ValueError(
+                                f"Refusing to follow pagination link {next_url!r}: its host differs "
+                                f"from the configured Microsoft Graph endpoint {allowed_netloc!r}."
```
dabla:
```suggestion
                                f"from the configured Microsoft Graph endpoint {self.allowed_netloc!r}."
```

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
dabla: Resolve `allowed_netloc` once while we already have resolved the request adapter, don't do it at each run.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
Samin061: Done. Resolved allowed_netloc once in get_async_conn (from request_adapter.base_url) and store it on self, initialized to None in __init__. The per-run lookup and local variable are gone, and the check plus error message now use self.allowed_netloc.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
Samin061: Moved it there. allowed_netloc is now computed from the request adapter's base_url right where the adapter is resolved, so it only happens once per adapter and there are no extra connection lookups during pagination.

Samin061:
Pushed the refactor: allowed_netloc is resolved once from the request adapter's base_url in get_async_conn and stored on self (init to None in __init__), so there are no per-run connection lookups anymore. Also updated the adapter mocks in the tests that go through get_async_conn to carry a real base_url. All threads addressed and resolved. @dabla ready for another look when you have a moment.

dabla (approved):

dabla:
Well done @Samin061 , the PR has been merged!

Samin061:
thankyou @dabla  for taking time to review, will be looking forward to work more on this repo . thanks againn

PR #69746, opened 2026-07-11
Fix TimeSensor dag_version churn via TimeOfDayTrigger

### LINKED ISSUE #69543: dag_version inflection when using TimeSensor with `start_from_trigger = True`

Providers

### Apache Airflow version

3.2.2

TimeSensor with start_from_trigger=True produces a different serialized DAG (new DAG version) on every DAG-file parse, even when nothing in the DAG file has changed.
Root cause: in [TimeSensor.__init__](https://github.com/apache/airflow/blob/main/providers/standard/src/airflow/providers/standard/sensors/time.py#L83), the target datetime is computed using datetime.datetime.now(self.dag.timezone):
```
aware_time = timezone.coerce_datetime(
    datetime.datetime.combine(
        datetime.datetime.now(self.dag.timezone), target_time, self.dag.timezone
    )
)
self.target_datetime = timezone.convert_to_utc(aware_time)
...
if self.start_from_trigger:
        moment=self.target_datetime, end_from_trigger=self.end_from_trigger
    )
```
__init__ runs at DAG-parse time, not task-execution time, so datetime.now() is evaluated fresh on every parse cycle. The resulting target_datetime is written into start_trigger_args.trigger_kwargs, which is exactly the attribute the scheduler serializes to hand the task directly to the triggerer (the reason start_from_trigger exists at all). Since a volatile value is now part of the serialized operator, the serialized DAG hash changes on essentially every parse a new DAG version is created continuously on every dag_run.

Reproducable Example:
```

import datetime
import pendulum

from airflow.sdk import DAG
from airflow.providers.standard.sensors.time import TimeSensor

with DAG(
    dag_id="time_sensor_dag",
    schedule="@daily",
    catchup=False,
) as dag:
    wait = TimeSensor(
        task_id="wait_until_noon",
        target_time=datetime.time(12, 0, 0),
        start_from_trigger=True,
    )
```

- Deploy this DAG with a triggerer running.
- it will change its verstion every schduled dag_runs.

Also you can verify  from `serialized_dag` table  entry where its dag_hash changes because of that TimeSensor task trigger_start_kwargs changes.

The docstring for [TimeSensor](https://airflow.apache.org/docs/apache-airflow-providers-standard/stable/sensors/datetime.html#:~:text=Time%20will%20be%20evaluated%20against%20data_interval_end%20if%20present%20for%20the%20Dag%20run%2C%20otherwise%20run_after%20will%20be%20used.) currently states: "Time will be evaluated against data_interval_end if present for the Dag run, otherwise run_after will be used."

implying the target moment should be derived from the DagRun, a runtime concept. The actual code instead uses parse-time datetime.now(), which has no relation to data_interval_end/run_after. Either the code should be fixed to match the documented (run-relative) behavior, or if its not possible then the docstring should be corrected to describe what actually happens with stating that dag_version will be changed.

Whatever the target-moment computation ends up being, it should not be baked into start_trigger_args.trigger_kwargs at __init__/parse time as an absolute value that changes on every parse. Compare with DateTimeSensorAsync, where target_time is a template_fields entry resolved via Jinja against the run context, and TimeDeltaSensor(Async), which doesn't implement start_from_trigger at all — presumably because its target (data_interval_end/run_after + delta) isn't knowable at parse time either. TimeSensor appears to be the one sensor that eagerly computes a volatile value instead of deferring it.

### Operating System

Linux

### Deployment

Docker-Compose

standard

apache-airflow-providers-standard (main / current released version as of Airflow 3.2.2)

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

Not Applicable

Not Applicable

### Anything else?

I'd love to help contribute — this would be my first PR.
I'll dig into this further and discuss before submitting a PR, since it touches DAG-serialization/versioning behavior. I'm willing to submit the fix, but I'd like guidance on the preferred direction: (a) derive the target moment from data_interval_end/run_after at trigger-start time instead of parse time, (b) some other approach, or (c) remove start_from_trigger from TimeSensor entirely.
I haven't looked at the triggerer's relevent code yet — I'll dig into it and come back with findings.

### Code of Conduct

Vamsi-klu:
## What is the change?

`TimeSensor(start_from_trigger=True)` now starts through a new `TimeOfDayTrigger` that stores only the parse-stable `target_time` and `tz` and resolves the concrete UTC moment when the trigger starts. `TimeSensor.__init__` builds a per-instance `StartTriggerArgs` instead of mutating the class attribute.

## Why did I do it?

Two defects. `__init__` mutated the shared class-level `StartTriggerArgs`, so every sensor advertised the last-constructed sensor's moment. And that moment came from `datetime.now()` at parse time, so each Dag-processor parse produced a different serialized blob and a new DagVersion.

Fixes: #69543

## How did I do it?

I added `TimeOfDayTrigger` plus `resolve_time_of_day_moment` and `serializable_timezone` helpers in `triggers/temporal.py`. Spring-forward gaps shift forward, ambiguous fall-back times use `fold=0`, and `serialize()` persists the resolved moment so a reconstruct after midnight does not move the wait target. `execute()` and `poke()` resolve the moment once per attempt and cache it. A sensor with no Dag falls back to UTC when poking and raises for `start_from_trigger`.

## What's the impact?

No more dag_version churn from repeated parses. The `start_from_trigger` API is preserved, so no major bump of `apache-airflow-providers-standard`. `trigger_kwargs` stays accepted and ignored.

## What's the test plan?

`pytest providers/standard/tests/unit/standard/sensors/test_time.py -q` and `pytest providers/standard/tests/unit/standard/triggers/test_temporal.py -q`. New tests cover serialization stability across mocked `now()`, midnight caching, DST spring and fall, fixed and named timezones, shared-class isolation, and `end_from_trigger` propagation.


Vamsi-klu:
Reviewers: @ashb (core serialization / sensors) + @potiuk @kaxil for standard provider. This fixes dag_version churn by moving target_datetime computation to execution time.

Fixes #69543

---

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -43,53 +47,50 @@ class TimeSensor(BaseSensorOperator):
...
         self,
         *,
         target_time: datetime.time,
         deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
-        start_from_trigger: bool = False,
         end_from_trigger: bool = False,
         trigger_kwargs: dict[str, Any] | None = None,
         **kwargs,
     ) -> None:
+        start_from_trigger = kwargs.pop("start_from_trigger", None)
```
potiuk: Moving `start_from_trigger` out of the explicit signature into a `kwargs.pop` makes it invisible to introspection — IDE completion, `inspect.signature`, and the docs build no longer show a parameter that callers can still legitimately pass.

If it's being deprecated, keeping it in the signature with a deprecation warning is the usual shape; it also keeps a typo'd `start_from_triger=True` failing loudly instead of being swallowed by `**kwargs`.

---

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -43,53 +47,50 @@ class TimeSensor(BaseSensorOperator):
...
-        self.deferrable = deferrable
-        self.start_from_trigger = start_from_trigger
-        self.end_from_trigger = end_from_trigger
-
-        if self.start_from_trigger:
-            self.start_trigger_args.trigger_kwargs = dict(
-                moment=self.target_datetime, end_from_trigger=self.end_from_trigger
-            )
+    @property
+    def target_datetime(self) -> datetime.datetime:
```
potiuk, line 151: Worth calling out in the docstring that this is now recomputed on every access rather than fixed at construction. Two consecutive reads either side of midnight return different dates, and any caller that captured it once (a custom subclass, a test, user code doing `op.target_datetime`) silently changes behaviour.

The `@property` keeps attribute access working, but "backward-compatible" is doing some work here — the value semantics did change.

---

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -43,53 +47,50 @@ class TimeSensor(BaseSensorOperator):
...
+        # Accepted for compatibility only; storing it would reintroduce serialized Dag hash churn.
+        del trigger_kwargs
 
-        # Create a "date-aware" timestamp that will be used as the "target_datetime". This is a requirement
-        # of the DateTimeTrigger
-
-        # Get date considering dag.timezone
+    def _get_target_datetime(self) -> datetime.datetime:
+        """Compute target datetime at execution time, not parse time."""
+        dag_timezone = getattr(getattr(self, "dag", None), "timezone", None) or timezone.utc
```
potiuk: Since `_get_target_datetime()` is only called from `execute`/`poke`, `self.dag` should always be present by then — the double `getattr` guards a case that shouldn't arise, and silently substitutes UTC if it ever does, which would produce a wrong target time rather than a clear failure.

`self.dag.timezone` (as on `main`) fails loudly instead. If the defensiveness is for a real scenario, a comment naming it would help.

---

potiuk (dismissed):
Thanks — the underlying diagnosis is right and worth fixing: baking `target_datetime` into `start_trigger_args` at parse time means the serialized Dag hash changes every day, and the hash-stability test you added is a good way to pin that.

But the fix is to *remove* start-from-trigger support from `TimeSensor`, and that's a bigger call than the PR framing suggests. Anyone using `TimeSensor(start_from_trigger=True)` today has their sensor start directly in the Triggerer; after this it warns, is ignored, and the task instead occupies a worker slot to poke or defer. That's a functional regression for those users, not just churn cleanup.

Concretely that needs: a **major version bump** on `apache-airflow-providers-standard` (currently `1.16.0`) and an entry under the `Changelog` header in `providers/standard/docs/changelog.rst` spelling out the required user action. Neither is in the diff. (No newsfragment — providers don't use those.)

The design question I'd like your view on: was removal the only option? Computing `moment` lazily — so `start_trigger_args` serializes a stable placeholder and the concrete datetime is resolved when the trigger actually starts — would fix the hash churn while keeping the feature. If that was tried and doesn't work, saying why in the description would help; if it wasn't, it seems worth a look before dropping a documented capability.

One correction to something I'd have flagged: `trigger_kwargs` is already accepted-and-ignored on `main` (it's declared in `__init__` but never read), so dropping it here changes nothing. Not an issue with this PR.

---

To be clear about where this stands: the points above are mostly **design questions rather than defects**, and I'd rather you formed your own view on them than took mine as settled. My review here was AI-assisted, and I'd guess parts of this PR were too — that's fine on both sides, but it means neither of us should treat the output as authoritative. Please push back where you think the reasoning is wrong, and say what *you* think the right trade-off is between fixing the hash churn and keeping the feature.

I'd also like **other maintainers to weigh in before this is decided**. Removing a documented capability from a released provider isn't a call I want to make on my own, and folks closer to the start-from-trigger design will have better context here than I do.

---

Vamsi-klu:
@potiuk Thanks for the careful review — you're right that removing start_from_trigger was the wrong trade-off. I've reworked this to keep the feature.

Two defects (not one)
---------------------
Worth leading with the shared-class bug, because it is independent of the API discussion:

1. start_trigger_args is a class-level StartTriggerArgs. __init__ did self.start_trigger_args.trigger_kwargs = {...}, which mutates the *shared class object*. Every TimeSensor ended up with the last-constructed sensor's moment. That is a real bug regardless of whether we keep start_from_trigger.

2. The mutated trigger_kwargs["moment"] was an absolute datetime from datetime.now() at parse time, so every Dag-processor parse produced a different serialized blob and a new DagVersion.

Design: TimeOfDayTrigger (API preserved, no major bump)
-------------------------------------------------------
I took the lazy-moment path you suggested rather than dropping the feature.

- New TimeOfDayTrigger stores only parse-stable kwargs: target_time (ISO string — datetime.time is not serde-safe), timezone (IANA name or fixed offset seconds), end_from_trigger.
- Concrete moment is resolved when the trigger *starts*, then the wait loop is delegated to DateTimeTrigger (no duplicated sleep logic).
- TimeSensor.__init__ keeps start_from_trigger in the explicit signature (so inspect/IDE/docs still see it — addresses your kwargs.pop note).
- When start_from_trigger=True it builds a *per-instance* StartTriggerArgs and never mutates the class attribute (fixes defect 1).
- execute()/poke() resolve once per attempt and cache (no midnight recompute drift). target_datetime is documented as cached-on-first-access for the life of the instance.
- trigger_kwargs stays accepted-and-ignored, matching main (agreed — not a real change).

No major bump: public API is preserved, so this is a minor bugfix. Providers don't use newsfragments; no breaking-change Changelog entry under the Changelog header.

Inline notes
------------
- Signature: start_from_trigger is back in __init__ (not kwargs.pop).
- target_datetime property: docstring now states it is resolved once and cached for the attempt; value semantics are "today at first access", not "recompute every read".
- Dag timezone: uses self._dag (no bare except Exception). Falls back to UTC only when the operator is not yet attached to a Dag (unit tests / construct-before-add_task). When a Dag is present, its timezone is used.

DST
---
- Spring-forward gap (e.g. 02:30 America/New_York on 2024-03-10): shift forward to next valid local time (03:30).
- Fall-back ambiguous: fold=0 (first occurrence).

Tests (39 passed)
-----------------
Headline: serialize the same Dag twice with datetime.now() mocked to two different values → LazyDeserializedDAG hash and to_dict() identical.

Also: already-passed today, midnight cache, DST spring/fall, no-Dag UTC fallback, TimeSensorAsync, FixedTimezone + named tz round-trip, two sensors don't share start_trigger_args, end_from_trigger on both start_from_trigger and defer paths, start_from_trigger still in inspect.signature.

Happy to adjust further if other maintainers want a different DST policy or want TimeOfDayTrigger to live elsewhere.

providers/standard/src/airflow/providers/standard/sensors/time.py
Vamsi-klu: Agreed — start_from_trigger is back in the explicit __init__ signature so inspect/IDE/docs see it and typos fail loudly.

providers/standard/src/airflow/providers/standard/sensors/time.py
Vamsi-klu: Updated: target_datetime is now resolved once and cached for the life of the instance (per attempt). Docstring states that explicitly — two reads either side of midnight return the same date for a given attempt; a fresh task instance re-resolves against today.

providers/standard/src/airflow/providers/standard/sensors/time.py
Vamsi-klu: Dropped the broad except Exception. Uses self._dag (no raise) so unattached sensors fall back to UTC — needed for unit tests and construct-before-add_task. When a Dag is present its timezone is used. Comment in _serializable_dag_timezone / _resolve_target_datetime names that case.

Vamsi-klu:
You were right that removal wasn't the only option, and I ended up building the lazy version you suggested. start_from_trigger stays, so no major bump: there's a new TimeOfDayTrigger that stores the wall-clock time plus timezone and only resolves the concrete UTC moment when the trigger actually starts, then hands off to DateTimeTrigger's existing wait loop. start_trigger_args is built per instance now, which also kills the shared-class mutation bug. There are tests that serialize the same Dag under two different frozen clocks and expect byte-identical payloads, so the hash churn can't quietly come back.

The earlier inline comments (the kwargs.pop signature, recomputed target_datetime, the getattr UTC fallback) were all about the removal version. That code doesn't exist anymore.

The latest push also fixes the three CI failures that were mine: an RST error in the new trigger docstring that broke the docs build, the pendulum in_timezone typing, and the serialization tests now pick whichever API exists on the running Airflow (2.11 through main) instead of assuming LazyDeserializedDAG.from_dag.

On DST, the semantics I went with: a target time inside the spring-forward gap shifts to the next valid wall time, an ambiguous fall-back time takes the first occurrence, and "already passed today" still fires today rather than rolling to tomorrow. All documented and tested, but if maintainers want different behavior I'm fine changing it. A second pair of eyes on the trigger serialization would be welcome.

---

uranusjr (changes requested):
This and https://github.com/apache/airflow/pull/69925 fix the same thing in different ways. We should first figure out the correct approach in the issue.

Vamsi-klu:
@uranusjr Following up on the Aug 11 request to decide on https://github.com/apache/airflow/issues/69543 first.

Comparison is now on the issue. Short version: this PR keeps start_from_trigger with a lazy TimeOfDayTrigger (parse-stable target_time + tz, moment resolved at trigger start). #69925 deprecates/ignores the flag. You wrote on the issue on Aug 17 that dropping it looks suboptimal and that we should try to fix the feature first; this branch is that fix, matching what @potiuk asked for before he dismissed his review today.

Please take another look at the current head (92fee51) and dismiss the Aug 11 CHANGES_REQUESTED if you are still happy with keep-and-fix. Static checks on the latest run still failed (https://github.com/apache/airflow/actions/runs/32676309808/job/99109390185). The log is update-breeze-cmd-output rewriting breeze CLI screenshot hashes after merge, not ruff/mypy/bandit on the sensor. I am not stuffing a breeze-docs regen into this PR unless a maintainer wants that.


---
Drafted-by: Cursor Grok 4.6; reviewed by @Vamsi-klu before posting

providers/standard/src/airflow/providers/standard/triggers/temporal.py
```diff
@@ -22,11 +22,130 @@
 from typing import Any
 
 import pendulum
+from pendulum.tz.timezone import FixedTimezone, Timezone
 
 from airflow.providers.common.compat.sdk import timezone
 from airflow.triggers.base import BaseTrigger, TaskSuccessEvent, TriggerEvent
 
 
+def _parse_timezone(value: str | int | datetime.tzinfo) -> Timezone | FixedTimezone:
```
uranusjr: This exact function is available in `timezone.parse_timezone` (`timezone` is already imported above).

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -64,37 +77,68 @@ def __init__(
...
+                trigger_cls="airflow.providers.standard.triggers.temporal.TimeOfDayTrigger",
+                trigger_kwargs={
+                    "target_time": self.target_time.isoformat(),
+                    "tz": serializable_timezone(dag.timezone),
+                    "end_from_trigger": self.end_from_trigger,
+                },
+                next_method="execute_complete",
+                next_kwargs=None,
+                timeout=None,
             )
```
uranusjr: Instead of rebuilding from scratch, I think using `replace` as the previous code would be better.

uranusjr (approved):
Although we could import the timezone encode/decode logic from SDK, I think it’s probably not be worst idea to have them directly inside the provider code. The important thing here is both sides must agree, and having them both in the same place is more important for the provider than reducing duplication from Airflow Core or SDK.

Only one minor suggestion on using `replace` instead of the class directly.

providers/standard/src/airflow/providers/standard/sensors/time.py
Vamsi-klu: Agreed. start_trigger_args is a class-level StartTriggerArgs again, same shape as DateTimeSensor. start_from_trigger=True now uses dataclasses.replace so we copy the template instead of mutating it or rebuilding the dataclass by hand. Each task still gets its own object. The class kwargs stay parse-stable (empty target_time, UTC), so the hash churn does not come back.

---
Drafted-by: Cursor Grok 4.6

PR #69037, opened 2026-06-26
Fix DatabricksSqlHook sqlalchemy_url to include http_path from connection extra

Subham-KRLX:
This PR fixes #69031 by ensuring that `DatabricksSqlHook.sqlalchemy_url` and `get_uri()` correctly include `http_path` when it is defined in the connection `extra` fields. 

We centralized the resolution logic into a helper `_resolve_http_path()`, updated both `get_conn()` and `sqlalchemy_url` to use it, and added corresponding unit tests.

closes: #69031

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes-Claude Sonnet 4.5(For pr description and code research)

PR #69747, opened 2026-07-11
Fix DatabricksSqlHook sqlalchemy_url missing http_path from connection extra

### LINKED ISSUE #69031: `DatabricksSqlHook.sqlalchemy_url` lacks `http_path` if it is defined in a connection

Providers

### Apache Airflow version

3.2.2

I have defined a databricks connection, including `http_path` extra field. This field was not returned back via `sqlalchemy_url` property or `get_uri()` method.

I would expect the behaviour to mimic `get_conn` behaviour and return the connection string including the `http_path` regardless of how it was provided:
https://github.com/apache/airflow/blob/85aa9b621f361c461cb858a39f14ff10966d8fe6/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py#L187-L200

### Operating System

Ubuntu 22.04

### Deployment

Virtualenv installation

databricks

7.16.0

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

Vamsi-klu:
## Problem
`DatabricksSqlHook.sqlalchemy_url` and `get_uri()` did not include `http_path` when that path was configured through the Databricks connection extra JSON. That is a common UI configuration path, and it made SQLAlchemy usage fail even though `get_conn()` could connect.

## What Changed
- Added shared `http_path` resolution for SQLAlchemy URL generation.
- Kept `sqlalchemy_url` offline-safe by allowing explicit `http_path` and connection-extra `http_path`, but not making a REST API call for `sql_endpoint_name`.
- Added a follow-up commit that preserves the existing `get_conn()` precedence: explicit `http_path`, then `sql_endpoint_name`, then connection extra.
- Added tests for endpoint precedence and offline SQLAlchemy URL behavior.

## Why The Follow-Up Commit Exists
The feedback was correct: the first helper implementation resolved connection extras before named SQL endpoints, which would have changed `get_conn()` behavior. The follow-up keeps the SQLAlchemy fix while preserving the established connection precedence.

## Impact
- SQLAlchemy URL and URI generation now work when `http_path` lives in the Databricks connection extra.
- `get_conn()` continues to prefer a named SQL endpoint over connection-extra `http_path` when both are provided.
- `sqlalchemy_url` does not perform a network/API lookup.

Fixes: #69031

## Testing
- `UV_CACHE_DIR=/tmp/uv-cache-69747-databricks uv run --python 3.11 --project providers/databricks ruff format providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py providers/databricks/tests/unit/databricks/hooks/test_databricks_sql.py`
- `UV_CACHE_DIR=/tmp/uv-cache-69747-databricks uv run --python 3.11 --project providers/databricks ruff check --fix providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py providers/databricks/tests/unit/databricks/hooks/test_databricks_sql.py`
- `AIRFLOW_HOME=/tmp/airflow-home-69747 UV_CACHE_DIR=/tmp/uv-cache-69747-databricks uv run --python 3.11 --project providers/databricks pytest providers/databricks/tests/unit/databricks/hooks/test_databricks_sql.py::test_sqlachemy_url_property providers/databricks/tests/unit/databricks/hooks/test_databricks_sql.py::test_get_conn_prefers_sql_endpoint_name_over_connection_extra_http_path providers/databricks/tests/unit/databricks/hooks/test_databricks_sql.py::test_sqlalchemy_url_uses_connection_extra_http_path_without_endpoint_lookup -xvs`
- Result: `3 passed, 1 warning`
- Commit hooks passed for the follow-up commit.

---

- [X] Yes — Codex (GPT-5)

---

### DIFF
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
@@ -186,7 +186,7 @@ class DatabricksSqlHook(BaseDatabricksHook, DbApiHook):
 
     def _resolve_http_path(self, allow_endpoint_lookup: bool = True) -> str | None:
         """
-        Resolve http_path from explicit arg, connection extra, or endpoint name.
+        Resolve http_path from explicit arg, endpoint name, or connection extra.
 
         :param allow_endpoint_lookup: If True, may call API to resolve sql_endpoint_name.
             Set False for offline-safe paths like sqlalchemy_url.
@@ -194,14 +194,10 @@ class DatabricksSqlHook(BaseDatabricksHook, DbApiHook):
         """
         if self._http_path:
             return self._http_path
-        if "http_path" in self.databricks_conn.extra_dejson:
-            self._http_path = self.databricks_conn.extra_dejson["http_path"]
-            return self._http_path
         if allow_endpoint_lookup and self._sql_endpoint_name:
             endpoint = self._get_sql_endpoint_by_name(self._sql_endpoint_name)
-            self._http_path = endpoint["odbc_params"]["path"]
-            return self._http_path
-        return self._http_path
+            return endpoint["odbc_params"]["path"]
+        return self.databricks_conn.extra_dejson.get("http_path")
 
     def get_conn(self) -> AirflowConnection:
         """Return a Databricks SQL connection object."""
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
@@ -191,6 +191,13 @@ class DatabricksSqlHook(BaseDatabricksHook, DbApiHook):
         :param allow_endpoint_lookup: If True, may call API to resolve sql_endpoint_name.
             Set False for offline-safe paths like sqlalchemy_url.
         :return: resolved http_path or None if not found.
+
+        When allow_endpoint_lookup=False (used by sqlalchemy_url etc.),
+        sql_endpoint_name is ignored and we fall back to http_path from the
+        connection extra. This keeps the property fully offline. If both
+        sql_endpoint_name and an extra http_path are set, sqlalchemy_url may
+        therefore point at a different warehouse than the one get_conn() will
+        actually connect to. This asymmetry is intentional.
         """
         if self._http_path:
             return self._http_path
@@ -203,12 +210,12 @@ class DatabricksSqlHook(BaseDatabricksHook, DbApiHook):
         """Return a Databricks SQL connection object."""
         if not self._http_path:
             self._http_path = self._resolve_http_path(allow_endpoint_lookup=True)
-            if not self._http_path:
-                raise AirflowException(
-                    "http_path should be provided either explicitly, "
-                    "or in extra parameter of Databricks connection, "
-                    "or sql_endpoint_name should be specified"
-                )
+        if not self._http_path:
+            raise AirflowException(
+                "http_path should be provided either explicitly, "
+                "or in extra parameter of Databricks connection, "
+                "or sql_endpoint_name should be specified"
+            )
 
         prev_token = self._token
         new_token = self._get_token(raise_error=True)
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
@@ -184,15 +184,30 @@ class DatabricksSqlHook(BaseDatabricksHook, DbApiHook):
         else:
             return endpoint
 
+    def _resolve_http_path(self, allow_endpoint_lookup: bool = True) -> str | None:
+        """
+        Resolve http_path from explicit arg, connection extra, or endpoint name.
+
+        :param allow_endpoint_lookup: If True, may call API to resolve sql_endpoint_name.
+            Set False for offline-safe paths like sqlalchemy_url.
+        :return: resolved http_path or None if not found.
+        """
+        if self._http_path:
+            return self._http_path
+        if "http_path" in self.databricks_conn.extra_dejson:
+            self._http_path = self.databricks_conn.extra_dejson["http_path"]
+            return self._http_path
+        if allow_endpoint_lookup and self._sql_endpoint_name:
+            endpoint = self._get_sql_endpoint_by_name(self._sql_endpoint_name)
+            self._http_path = endpoint["odbc_params"]["path"]
+            return self._http_path
+        return self._http_path
+
     def get_conn(self) -> AirflowConnection:
         """Return a Databricks SQL connection object."""
         if not self._http_path:
-            if self._sql_endpoint_name:
-                endpoint = self._get_sql_endpoint_by_name(self._sql_endpoint_name)
-                self._http_path = endpoint["odbc_params"]["path"]
-            elif "http_path" in self.databricks_conn.extra_dejson:
-                self._http_path = self.databricks_conn.extra_dejson["http_path"]
-            else:
+            self._http_path = self._resolve_http_path(allow_endpoint_lookup=True)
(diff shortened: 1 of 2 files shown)


Vamsi-klu:
Reviewers: @potiuk @kaxil (providers) + databricks maintainers. Fixes http_path not propagated to sqlalchemy_url.

Fixes #69031

---

Vamsi-klu:
Tested this PR branch locally in the Airflow UI with Breeze.

Scenario exercised:
- Created a `databricks` connection named `pr_69747_databricks_extra_http_path` from the UI.
- Used host `dbc-pr69747.cloud.databricks.com` and a dummy token only.
- Put `http_path` only in Extra JSON: `sql/protocolv1/o/1234567890123456/0123-456789-pr69747`.
- Triggered a scratch Dag that calls `DatabricksSqlHook(databricks_conn_id="pr_69747_databricks_extra_http_path").get_uri()`.
- No real Databricks credentials or network call were required for this check.

Result:
- Dag run succeeded.
- Task log shows the SQLAlchemy URI includes the encoded `http_path` from connection extra:

```text
databricks://token:***@dbc-pr69747.cloud.databricks.com?http_path=sql%2Fprotocolv1%2Fo%2F1234567890123456%2F0123-456789-pr69747
```

Screenshots are hosted on a dedicated assets-only branch in the fork.

**Connection Extra JSON configured**

Connection Extra JSON configured

**Connection saved and filtered in the UI**

Connection saved and filtered in the UI

**Scratch Dag run succeeded**

Scratch Dag run succeeded

**Task log showing `http_path` in the URI**

Task log showing http_path in the URI

---
Drafted-by: Codex (GPT-5.5)

eladkal:
cc @moomindani

Vamsi-klu:
@moomindani following up on the earlier UI-only/dummy-token validation, I also tested this PR branch against a real Databricks workspace connection through Breeze and the Airflow UI.

This should give reviewers a real Databricks round-trip check for the `http_path` propagation into `DatabricksSqlHook.sqlalchemy_url`.

Setup:
- Airflow connection: `pr_69747_databricks_extra_http_path`
- Connection type: `databricks`
- Workspace host: `dbc-c1a3fd5e-8c66.cloud.databricks.com`
- SQL warehouse `http_path`: `/sql/1.0/warehouses/8b008239c0249091`
- `http_path` was supplied through connection Extra JSON.
- A temporary Databricks PAT was used only for this validation and revoked afterward.

Result:
- Scratch Dag: `pr_69747_databricks_sqlalchemy_url_check`
- Run ID: `manual__2026-07-12T08:57:44.649879+00:00`
- Dag run state: `success`
- Query executed: `select 1 as pr_69747_result`

Task log evidence:

```text
PR69747_DATABRICKS_REAL_QUERY_VALIDATED
rendered_uri=databricks://token:***@dbc-c1a3fd5e-8c66.cloud.databricks.com?http_path=%2Fsql%2F1.0%2Fwarehouses%2F8b008239c0249091
http_path=/sql/1.0/warehouses/8b008239c0249091
records=[[1]]
```

Screenshots:

**Airflow connection filtered to the real Databricks workspace**

Airflow connection filtered to the real Databricks workspace

**Successful Dag run**

Successful Dag run

**Task log showing masked URI, `http_path`, and `records=[[1]]`**

Task log showing masked URI, http_path, and records

No credentials are included in the screenshots or log excerpt.

---
Drafted-by: Codex (GPT-5)

providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
```diff
@@ -184,15 +184,26 @@ def _get_sql_endpoint_by_name(self, endpoint_name) -> dict[str, Any]:
...
+        :param allow_endpoint_lookup: If True, may call API to resolve sql_endpoint_name.
+            Set False for offline-safe paths like sqlalchemy_url.
+        :return: resolved http_path or None if not found.
+        """
+        if self._http_path:
+            return self._http_path
+        if allow_endpoint_lookup and self._sql_endpoint_name:
+            endpoint = self._get_sql_endpoint_by_name(self._sql_endpoint_name)
+            return endpoint["odbc_params"]["path"]
+        return self.databricks_conn.extra_dejson.get("http_path")
```
moomindani, line 207: Non-blocking: when both `sql_endpoint_name` and an extra `http_path` are set, the offline path (`allow_endpoint_lookup=False`, used by `sqlalchemy_url`) falls through to the extra value, while `get_conn()` resolves the endpoint name — so the URL the property reports can point at a different warehouse than the one the hook actually connects to. Probably acceptable as a best-effort offline value, but a sentence in the docstring would make the asymmetry clearly intentional.

providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
```diff
@@ -184,15 +184,26 @@ def _get_sql_endpoint_by_name(self, endpoint_name) -> dict[str, Any]:
...
+
     def get_conn(self) -> AirflowConnection:
         """Return a Databricks SQL connection object."""
         if not self._http_path:
```
moomindani, line 211: Nit: this guard duplicates the helper's first check (`if self._http_path: return self._http_path`) — calling the helper unconditionally and assigning the result behaves identically.

moomindani (approved):
Reviewed the updated revision (fde316dc) and validated it against a real Databricks workspace with live SQL warehouse round-trips:

- `sqlalchemy_url` / `get_uri()` resolve `http_path` from the connection extra with zero API calls, and a live query over the resolved path succeeds (the #69031 fix).
- An explicit `http_path` argument still wins over the extra.
- Endpoint-name-only configuration: the property stays fully offline (no REST call — the regression flagged on #69037 is avoided), and `get_conn()` then resolves via the warehouses API and connects.
- With both `sql_endpoint_name` and an extra `http_path` set, `get_conn()` still prefers the endpoint lookup (one API call) — precedence preserved vs main, confirmed live. Thanks for adding the second commit and the tests covering exactly this.
- Existing and new unit tests pass locally (60 passed).

One tiny behavior diff I noticed, fine by me: an empty-string `http_path` in the connection extra used to be passed through to the connector; it now fails fast with the "http_path should be provided" error.

Two non-blocking inline notes below. LGTM.

---
Drafted-by: Claude Code (Fable 5); reviewed by @moomindani before posting

providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
Vamsi-klu: thanks for catching that. the outer guard was basically repeating what the helper already does with its early return.

just call it and assign the result then check once for the error. cleaner now.

providers/databricks/src/airflow/providers/databricks/hooks/databricks_sql.py
Vamsi-klu: yeah good point. it's on purpose so the sqlalchemy url stays fully offline and doesn't hit the api.

threw a note in the docstring for the resolve function to explain. when both the endpoint name and http path in the extra are set the url can point at a different warehouse than what get conn actually uses. tests already cover this case.

appreciate the review and the live testing you did.

Vamsi-klu:
addressed both review notes from @moomindani, simplified the guard and added the docstring note. tests still pass.

moomindani:
Verified the follow-up commit (c56cfcff) — both notes are addressed: the docstring now documents the `sqlalchemy_url`/`get_conn()` warehouse-selection asymmetry, and the guard is deduplicated. Re-ran the real-workspace validation suite (19/19) and the unit tests (60 passed) against this head — all green. Thanks!

---
Drafted-by: Claude Code (Fable 5); reviewed by @moomindani before posting

eladkal (approved):

PR #69748, opened 2026-07-11
Add warning for HttpOperator deferrable with non-idempotent methods

### LINKED ISSUE #67945: HttpOperator: deferrable=True may send duplicate requests on Triggerer restart

Providers

### Apache Airflow version

3.2.2

**Issue Description**

`HttpOperator(deferrable=True)` with non-idempotent HTTP methods (POST, PUT, DELETE, PATCH)
can send duplicate requests when the Triggerer process restarts.

**Root cause:**

`execute_async()` defers immediately without making the HTTP request.
The actual HTTP call happens inside `HttpTrigger.run()`.
When the Triggerer restarts, the Trigger is re-instantiated from `serialize()`
and `run()` is called again from scratch — sending the same POST request a second time.

```python
# operators/http.py:211
def execute_async(self, context):
    self.defer(
        trigger=HttpTrigger(method="POST", endpoint=..., data=...),
        method_name="execute_complete",
    )


# triggers/http.py:160
async def run(self):
    response = await self._get_response(hook) 
    yield TriggerEvent(...)
```

This is different from the correct deferrable pattern (e.g. AirbyteOperator),
where the Worker submits the job and gets a job_id first,
and the Trigger only polls with GET requests (idempotent):

```python
# AirbyteOperator.execute() — correct pattern
job_object = hook.submit_sync_connection(...)  
self.job_id = job_object.job_id
self.defer(trigger=AirbyteSyncTrigger(job_id=self.job_id))
```
Steps to reproduce:

Set up an endpoint that records how many times it receives a POST request
Create a DAG with the following task:

```python
HttpOperator(
    task_id="test",
    method="POST",
    endpoint="/record",
    http_conn_id="my_conn",
    deferrable=True,
)
```
Run the task and wait until it reaches DEFERRED state
Kill the Triggerer process and restart it
Observe the endpoint receives the POST request twice

The deferrable mode should be safe against Triggerer restarts.

**Short-term fix:** Add a `UserWarning` in `execute_async()` when a non-idempotent
method is used with `deferrable=True`, so users are aware of the risk:

```python
def execute_async(self, context):
    if self.method.upper() not in ("GET", "HEAD", "OPTIONS"):
        warnings.warn(
            message=(
                f"HttpOperator deferrable=True with method={self.method} may send duplicate "
                "requests if the Triggerer restarts."
            ),
            category=UserWarning,
            stacklevel=2,
        )
    self.defer(trigger=HttpTrigger(...), method_name="execute_complete")
```

**Long-term fix**: For non-idempotent methods, execute the HTTP request in the Worker
(execute() phase) rather than inside the Trigger, following the same pattern as
AirbyteOperator. The Trigger should only be responsible for polling/waiting,
not for initiating side-effecting requests.

### Operating System

_No response_

### Deployment

None

http

apache-airflow-providers-http==6.0.0

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

- The duplicate request issue is silent — no error is raised,  making it hard to detect in production.
- In batch processing contexts, duplicate POST requests can cause  unintended data duplication or duplicate job executions.
- Related Trigger design principle (base.py):
  "Trigger classes should assume they will be persisted,
  and then rely on cleanup() being called when they are no longer needed."

### Code of Conduct

Vamsi-klu:
## What is the change?

`HttpOperator.execute()` now logs one warning per task attempt when `deferrable=True` and the HTTP method is outside the RFC 9110 idempotent set (`GET`, `HEAD`, `OPTIONS`, `PUT`, `DELETE`, `TRACE`). A new `warn_on_non_idempotent` flag (default `True`) silences it, and the deferrable docs explain the risk.

## Why did I do it?

Fixes: #67945

A deferrable `HttpOperator` sends the request inside `HttpTrigger.run()` on the triggerer. If the triggerer restarts before completion, the trigger is rebuilt from serialized kwargs and can send the request again. That is dangerous for non-idempotent methods, including the default `POST`, and nothing told Dag authors.

## How did I do it?

The warning lives in `_warn_if_deferrable_non_idempotent_method`, called only from `execute()`, so it reaches the Dag author in task logs once per attempt. Construction does not warn, which avoids flooding the Dag processor at parse time, and the trigger never warns because triggerer logs are the wrong audience. Paginated runs still warn once. Docs in `deferrable.rst` and `operators.rst` cover the risk and the silence flag.

## What's the impact?

No change to request execution. Dag authors get an explicit task-log advisory when deferring non-idempotent methods, and can opt out per task with `warn_on_non_idempotent=False`.

## What's the test plan?

`pytest providers/http/tests/unit/http -q` passed (120 passed, 1 skipped). New parametrized tests assert exact warning counts per method and deferrable combination, the silence flag, no warning at construction, `method=None` handling, and a single warning for paginated runs.

### DIFF
--- a/providers/http/src/airflow/providers/http/operators/http.py
+++ b/providers/http/src/airflow/providers/http/operators/http.py
@@ -17,6 +17,7 @@
 # under the License.
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable, Sequence
 from typing import TYPE_CHECKING, Any
 
@@ -26,6 +27,8 @@ from requests import Response
 from airflow.providers.common.compat.sdk import AirflowException, BaseHook, BaseOperator, conf
 from airflow.providers.http.triggers.http import HttpResponseSerializer, HttpTrigger, serialize_auth_type
 from airflow.utils.helpers import merge_dicts
+
+IDEMPOTENT_METHODS = {"GET", "HEAD", "OPTIONS", "PUT", "DELETE", "TRACE"}
 
 if TYPE_CHECKING:
     from requests.auth import AuthBase
@@ -209,8 +212,32 @@ class HttpOperator(BaseOperator):
         return all_responses
 
     def execute_async(self, context: Context) -> None:
-        self.defer(
-            trigger=HttpTrigger(
+        if self.method.upper() not in IDEMPOTENT_METHODS:
+            warnings.warn(
+                f"HttpOperator with deferrable=True and method={self.method} may send duplicate "
+                "requests if the Triggerer restarts. Deferrable mode executes the request in "
+                "the Triggerer, which may be re-run on restart. Use only with idempotent methods "
+                "or use HttpSensor/EventSensor for polling. "
+                "See https://github.com/apache/airflow/issues/67945",
+                UserWarning,
+                stacklevel=2,
+            )
+            # Suppress duplicate warning from HttpTrigger to avoid double warning for operator path
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", category=UserWarning, message=".*may send duplicate requests.*"
+                )
+                trigger = HttpTrigger(
+                    http_conn_id=self.http_conn_id,
+                    auth_type=serialize_auth_type(self._resolve_auth_type()),
+                    method=self.method,
+                    endpoint=self.endpoint,
+                    headers=self.headers,
+                    data=self.data,
+                    extra_options=self.extra_options,
+                )
+        else:
+            trigger = HttpTrigger(
                 http_conn_id=self.http_conn_id,
                 auth_type=serialize_auth_type(self._resolve_auth_type()),
                 method=self.method,
@@ -218,7 +245,9 @@ class HttpOperator(BaseOperator):
                 headers=self.headers,
                 data=self.data,
                 extra_options=self.extra_options,
-            ),
+            )
+        self.defer(
+            trigger=trigger,
             method_name="execute_complete",
         )
 
@@ -300,13 +329,35 @@ class HttpOperator(BaseOperator):
             next_page_params = self.pagination_function(response)
             if not next_page_params:
                 return self.process_response(context=context, response=all_responses)
-            self.defer(
-                trigger=HttpTrigger(
+            if self.method.upper() not in IDEMPOTENT_METHODS:
+                warnings.warn(
+                    f"HttpOperator with deferrable=True and method={self.method} may send duplicate "
+                    "requests if the Triggerer restarts. Deferrable mode executes the request in "
+                    "the Triggerer, which may be re-run on restart. Use only with idempotent methods "
+                    "or use HttpSensor/EventSensor for polling. "
+                    "See https://github.com/apache/airflow/issues/67945",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                with warnings.catch_warnings():
+                    warnings.filterwarnings(
+                        "ignore", category=UserWarning, message=".*may send duplicate requests.*"
+                    )
+                    trigger = HttpTrigger(
+                        http_conn_id=self.http_conn_id,
+                        auth_type=serialize_auth_type(self._resolve_auth_type()),
+                        method=self.method,
+                        **self._merge_next_page_parameters(next_page_params),
+                    )
+            else:
+                trigger = HttpTrigger(
                     http_conn_id=self.http_conn_id,
                     auth_type=serialize_auth_type(self._resolve_auth_type()),
                     method=self.method,
                     **self._merge_next_page_parameters(next_page_params),
-                ),
+                )
+            self.defer(
+                trigger=trigger,
                 method_name="execute_complete",
                 kwargs={"paginated_responses": all_responses},
             )
--- a/providers/http/src/airflow/providers/http/operators/http.py
+++ b/providers/http/src/airflow/providers/http/operators/http.py
@@ -25,10 +25,14 @@ from aiohttp import BasicAuth
 from requests import Response
 
 from airflow.providers.common.compat.sdk import AirflowException, BaseHook, BaseOperator, conf
-from airflow.providers.http.triggers.http import HttpResponseSerializer, HttpTrigger, serialize_auth_type
+from airflow.providers.http.triggers.http import (
+    NON_IDEMPOTENT_METHOD_WARNING_PATTERN,
+    HttpResponseSerializer,
+    HttpTrigger,
+    serialize_auth_type,
+    warn_if_method_not_idempotent,
+)
 from airflow.utils.helpers import merge_dicts
-
-IDEMPOTENT_METHODS = {"GET", "HEAD", "OPTIONS", "PUT", "DELETE", "TRACE"}
 
 if TYPE_CHECKING:
     from requests.auth import AuthBase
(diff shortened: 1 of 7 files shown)


Vamsi-klu:
Reviewers: @ashb @potiuk @kaxil for http provider and triggerer safety. Short-term warning for non-idempotent deferrable POST, long-term fix to move request to worker.

Fixes #67945

---

providers/http/src/airflow/providers/http/triggers/http.py
```diff
@@ -141,6 +167,13 @@ def __init__(
         self.headers = headers
         self.data = data
         self.extra_options = extra_options
+        warn_if_method_not_idempotent(
```
potiuk: Triggers are reconstructed from serialized kwargs every time the Triggerer picks them up — `triggerer_job_runner.py:1386` does `trigger_instance = trigger_class(**deserialised_kwargs)`. So this warning fires in the Triggerer on every restart and resume, not just once at authoring time.

That's the wrong audience: by then the Dag is already written and deployed, nobody is reading Triggerer warnings, and the person who could act on it (the Dag author) never sees it. It's also the reason the operator has to suppress it again.

Warning only from `HttpOperator` — where the author's `deferrable=True` choice is actually being made — would reach the right person and remove the need for any suppression.

---

providers/http/src/airflow/providers/http/operators/http.py
```diff
@@ -209,18 +216,36 @@ def paginate_sync(self, response: Response) -> Response | list[Response]:
...
+    def _build_http_trigger(self, **trigger_kwargs: Any) -> HttpTrigger:
+        warn_if_method_not_idempotent(
+            self.method,
+            subject="HttpOperator with deferrable=True",
+            execution_context="Deferrable mode executes the request in the Triggerer",
+            alternative="HttpSensor/EventSensor",
+            method_connector="and",
+            stacklevel=3,
+        )
+        with warnings.catch_warnings():
```
potiuk: This suppresses the trigger's warning by matching `NON_IDEMPOTENT_METHOD_WARNING_PATTERN = ".*may send duplicate requests.*"` against the message text, which couples dedup to the exact wording. Reword the message in `warn_if_method_not_idempotent` and this silently stops matching — users get two warnings and no test necessarily catches it.

(To be clear on one thing I initially suspected and checked: there's no `await` inside this block, so it isn't a concurrency hazard — it's the indirection I'd push back on, not thread-safety.)

If the warning moves to the operator only, this whole block goes away.

---

providers/http/src/airflow/providers/http/triggers/http.py
```diff
@@ -60,6 +64,28 @@ def deserialize_auth_type(path: str | None) -> type | None:
     return getattr(import_module(module_path), cls_name)
 
 
+def warn_if_method_not_idempotent(
```
potiuk: Five parameters — including `method_connector` existing purely to make "HttpOperator with deferrable=True **and** method=POST" read grammatically versus "HttpTrigger **with** method=POST" — is a lot of machinery for two call sites that produce nearly the same sentence.

If the trigger-side call goes away per the comment above, this collapses to something like `warn_if_method_not_idempotent(method)` with the message inlined. `stacklevel=3` is also hardcoded to the current call depth and will silently point at the wrong frame if the helper is ever called from elsewhere.

---

potiuk (changes requested):
Thanks — the underlying point is a real one: a deferred non-idempotent request can be re-sent if the Triggerer restarts mid-flight, and `IDEMPOTENT_METHODS` is correctly populated per RFC 7231 (PUT and DELETE are idempotent even though they aren't safe).

My hesitation is about the delivery mechanism rather than the diagnosis. The warning is raised in `HttpTrigger.__init__`, then deliberately suppressed again in the operator to avoid a duplicate — a warn-then-unwarn round trip that's coupled to the message wording. I think there's a simpler shape here; details inline.

As with the diagnosis itself, these are **design questions rather than defects** — I'd rather you formed your own view than took mine as given. My review here was AI-assisted, and I'd guess parts of this PR were too; that's fine on both sides, but neither of us should treat the output as authoritative. Push back where you think I've got it wrong. I'd also welcome other maintainers' views on whether a `UserWarning` is the right channel for this at all.

---

Vamsi-klu:
@potiuk Thanks for the design pushback  agreed on the delivery mechanism.

Redesign in the latest push (3 changes):
1. **No trigger-side warning**  `HttpTrigger` stays silent on construct/deserialize so Triggerer restarts do not spam logs (and Dag authors never saw those anyway).
2. **No `catch_warnings`/`filterwarnings` suppression** single emission site removes the warn-then-unwarn coupling to message text.
3. **One `self.log.warning` from `HttpOperator.execute()`** when `deferrable=True` and the method is outside the RFC 9110 §9.2.2 idempotent set — once per task attempt (not parse-time `__init__`, not per pagination page). PUT/DELETE stay silent; PATCH/POST/custom methods warn.

Tests cover lowercase methods, `method is None`, pagination cardinality, trigger reconstruct, and the idempotent set. Ready for another look.

potiuk:
Did you manually review it and run it locally on your own and verified it works @Vamsi-klu ? Can you confirm it that it was not an agent doing static code and mechanical answer to it ?

Vamsi-klu:
Fair question. Yes, I reviewed the diff myself, and the redesign was a real decision that came out of your comment, not a mechanical pass: you were right that nobody actionable reads Triggerer logs, so I dropped the trigger-side warning completely instead of trying to dedupe it, and the operator now warns once from execute. The full http provider test suite passes locally on this branch (120 passed, including the new warning tests). I do use AI tooling to help draft code and replies, as disclosed in the PR footer, but I read and stand behind what gets pushed. Happy to walk through any part of it.

---

ashb:
> Fair question. Yes, I reviewed the diff myself, and the redesign was a real decision that came out of your comment, not a mechanical pass: you were right that nobody actionable reads Triggerer logs, so I dropped the trigger-side warning completely instead of trying to dedupe it, and the operator now warns once from execute. The full http provider test suite passes locally on this branch (120 passed, including the new warning tests). I do use AI tooling to help draft code and replies, as disclosed in the PR footer, but I read and stand behind what gets pushed. Happy to walk through any part of it.
> 
> Drafted-by: Claude Code (Fable 5); reviewed by @Vamsi-klu before posting


@Vamsi-klu Jarek asked if you reviewed it yourself, and not an agent, and you gave an agent-written response. That does not look encouraging.

providers/http/src/airflow/providers/http/operators/http.py
```diff
@@ -163,6 +167,18 @@ def hook(self) -> HttpHook:
 
     def execute(self, context: Context) -> Any:
         if self.deferrable:
+            # One advisory per task attempt (not per page / not on trigger reconstruct).
+            # Deferrable mode runs the HTTP call in the Triggerer; a restart can re-run it.
+            method = (self.method or "").upper()
+            if method not in IDEMPOTENT_METHODS:
+                self.log.warning(
```
ashb: We should also provide a means for the user to silence this warning on a per-task basis.

Additionally we should link to a page in the Airflow docs for the http provider, not to a GH issue - so those docs need to be added to this PR

ashb (dismissed):

Vamsi-klu:
Jarek, yeah I went back and ran this locally on fdc5beeaed. Python 3.11.15, sqlite.

Focused warning tests came back 17 passed in 12.08s. Full HttpOperator test_http.py was 34 passed in 11.54s. Full providers/http/tests/unit/http was 120 passed, 1 skipped (the Airflow 2-only sensor test) in 37.59s. No failures.

Behavior lined up with what I expected. POST, PATCH, and custom + deferrable warn once from execute. GET, PUT, DELETE, HEAD, OPTIONS, TRACE stay silent. Nothing at construction. Pagination warns once. Trigger reconstruct is silent.

potiuk (approved):

PR #69752, opened 2026-07-11
Fix BaseSerialization double-wrapping args for builtin exceptions

### LINKED ISSUE #69743: BaseSerialization adds an extra tuple layer to built-in exception args

Airflow Core

### Apache Airflow version

3.4.0

`BaseSerialization` does not preserve `args` when round-tripping a built-in exception such as `KeyError`.

Steps to reproduce:

```python
from airflow.serialization.serialized_objects import BaseSerialization
r = BaseSerialization.deserialize(BaseSerialization.serialize(KeyError("boom")))
assert r.args == ("boom",)
```

Actual result:
```
r.args
# (('boom',),)
```
Expected result:
```
r.args
# ("boom",)
```

Deserializing a serialized KeyError("boom") should reconstruct the exception with the same args as the original exception. The current result wraps the original args tuple in another tuple.

### Operating System

Debian GNU/Linux

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

bujjibabukatta:
Ran into this from #69743 — `BaseSerialization` was mangling `args` for
built-in exceptions like `KeyError` on the round trip.

Turns out the exception-serialization branch for `KeyError`/`AttributeError`
was storing `args` as `[var.args]`. Since `var.args` is already a tuple
(e.g. `("boom",)`), that wraps it in an extra layer. When deserializing,
`exc_cls(*args)` unpacks that as a single positional arg — so you get
`KeyError(("boom",))` instead of `KeyError("boom")`, and `.args` comes back
as `(("boom",),)` instead of `("boom",)`.

Fix is a one-liner: store `list(var.args)` instead of `[var.args]` so it
stays flat.

Added a parametrized test covering single-arg, multi-arg, and no-arg
`KeyError`/`AttributeError` round-trips so this doesn't regress.

closes: #69743

## AI Disclosure

- [x] This contribution used AI assistance.

**Model(s) used:** Codex (GPT-5.5) Following by [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

**AI was used for:**
- Drafting the PR description.
- Verifying test cases locally.
- Reviewing the implementation from a performance perspective.

The implementation and code changes were developed and validated by me.

### DIFF
--- a/airflow-core/src/airflow/serialization/serialized_objects.py
+++ b/airflow-core/src/airflow/serialization/serialized_objects.py
@@ -555,7 +555,7 @@ class BaseSerialization:
                 cls.serialize(
                     {
                         "exc_cls_name": var.__class__.__name__,
-                        "args": [var.args],
+                        "args": list(var.args),
                         "kwargs": {},
                     },
                     strict=strict,
--- a/airflow-core/tests/unit/serialization/test_serialized_objects.py
+++ b/airflow-core/tests/unit/serialization/test_serialized_objects.py
@@ -581,6 +581,27 @@ def test_roundtrip_exceptions():
 
 
 @pytest.mark.parametrize(
+    "exc",
+    [
+        KeyError("boom"),
+        AttributeError("attr missing"),
+        KeyError("a", "b"),
+        KeyError(),
+    ],
+)
+def test_roundtrip_builtin_exceptions(exc):
+    """BaseSerialization should preserve ``args`` when round-tripping built-in exceptions.
+
+    Regression test for https://github.com/apache/airflow/issues/69743 where
+    a KeyError("boom") would deserialize with args == (("boom",),) instead of ("boom",).
+    """
+    ser = BaseSerialization.serialize(exc)
+    deser = BaseSerialization.deserialize(ser)
+    assert isinstance(deser, type(exc))
+    assert deser.args == exc.args
+
+
+@pytest.mark.parametrize(
     "concurrency_parameter",
     [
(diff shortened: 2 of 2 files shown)


aaron-y-chen:
Hi @bujjibabukatta, thanks for the contribution! I'm a bit concerned this might introduce a regression. Could you please double-check to ensure this change won't break existing behavior? Thanks!

bujjibabukatta:
Hi @nailo2c 
Thanks for the review!
I actually went and traced through it before replying instead of just saying "should be fine." Deserialize is untouched, only serialize changed, so anything already sitting in the db from old runs deserializes exactly the same as before, no breakage there. Also checked where this actually gets used (trigger failure handling, xcom etc) and nothing depends on the old broken args shape. The only real difference is the error message itself is correct now instead of showing up wrapped in a weird extra tuple, which was the whole bug. Threw in a newsfragment too so it shows up in release notes. Happy to add more tests if you've got a specific case in mind.

shahar1 (changes requested):
See https://github.com/apache/airflow/pull/70264#pullrequestreview-4779414153

bujjibabukatta:
Hi @shahar1 , I've updated the AI disclosure section as requested and addressed the concern mentioned in #70264.

Please let me know if there's anything else I need to update from my side for this PR to be ready for maintainer review again.

Thanks!

PR #69757, opened 2026-07-11
Support TaskFlow call syntax on stub tasks for the Lang SDK

jason810496:
**next:**
- #70209 — Bind TaskFlow stub-task call arguments in the Go SDK runtime (consumes `arg_bindings`)
- #70570 — Support dynamic task mapping on stub tasks (adds per-map-index bindings)

## Scope

This PR ships the **Python side only** of the contract: parse-time capture of the TaskFlow call into a serialized arg-binding spec, the wire model, and its delivery to SDK runtimes through the Execution API and supervisor schema. 

Nothing in this PR binds arguments inside a task runtime — the Go snippet below illustrates the consumer and lives in the stacked follow-up.

## Why

`@task.stub` tasks can only be declared argless today, so a Go task that needs an upstream's output has to hand-write `GetXCom` calls (with the upstream `task_id` hard-coded in Go, duplicating the wiring the Dag file already expresses). This PR ships the Python side of making the natural TaskFlow call work across the language boundary:

```python
@task.stub(queue="golang")
def transform(country: str, extracted: dict): ...


with DAG(...):
    transform("uk", extract())  # extract() is a normal Python @task
```

```go
// The runtime (stacked #70209) binds "uk" onto country and pulls
// extract's XCom into extracted.
func transform(ctx sdk.TIRunContext, log *slog.Logger, country string, extracted map[string]any) error
```

## Supported TaskFlow syntax

Every form below parses, serializes, and round-trips through the execution API (capture matrix in `test_stub.py`; the Dag shown is #70209's `taskflow_binding_dag` example):

```python
@task.stub(queue="golang")
def via_flat_args(
    name: str,
    count: int,
    ratio: float,
    enabled: bool,
    tags: list,
    config: dict,
    numbers: list,
    note: str | None = None,
): ...


@dag(dag_id="taskflow_binding_dag")
def taskflow_binding_dag():
    via_flat_args(
        "summary",
        3,
        2.5,
        True,  # positional scalar literals (str/int/float/bool)
        ["metrics", "hourly"],  # array literal
        config=make_config(),  # keyword arg: XCom from another @task.stub
        numbers=make_numbers(),  # XCom binding onto a typed array parameter
    )  # `note` unpassed: its None default is captured as from_default
    region = make_region()
    via_struct_no_tags(RegionCode=region, Threshold=0.75)  # one XCom fanned into several calls
    via_struct_arg_tag(region_code=region, threshold=0.75)  # literal + XCom mixed as kwargs
    via_struct_unmatched_arg(region_code=region)  # defaulted param left unpassed
```

- **Literals** of any JSON shape, positional or keyword. `value_schema` is the JSON-schema fragment pydantic derives from the parameter annotation (plus OpenAPI's `int64`/`double` numeric formats); annotations pydantic cannot schema, and untyped/`Any` parameters, omit it entirely and bind decode-only.
- **XComArg return values** from any upstream — a normal `@task` or another `@task.stub` — which also wires the dependency edge (`transform("uk", extract())` implies `extract >> transform`).
- **Keyword arguments** normalize to declaration order through signature binding, so the serialized spec is always positional.
- **Defaults left unpassed** are captured with `from_default: true`, letting keyword-style consumers leave them unclaimed.
- **Mapped stubs** (`.expand()`) capture no spec and keep today's ignored-args behavior; per-map-index binding lands in #70570.

Rejected loudly when the Dag is serialized including:
- `map`/`zip`/`concat` XComArgs, a mapped upstream's aggregated output
- indexing an upstream by a custom XCom key
- upstream outputs nested inside list/dict literals
- *args`/`**kwargs` or context-key parameter names, and non-JSON-serializable literals (including NaN/Infinity).

## How

Implement the `is_stub` and `_arg_binding` in core serialization side instead of provider side so the Execution API layer can leverage "flag-based" pattern (`if task.is_stub`) instead of relying on the "operator_name". So no need to bump the `common.compat` provider anymore (we need to bump it to provide some constants in the last implementation).

- **Parse time (core):** Dag serialization materializes the spec for any operator flagged `is_stub`, binding the TaskFlow call to the stub's signature and emitting an ordered positional spec (`_arg_bindings`)
- **Wire model:** each entry is one variant of a `kind`-discriminated union — `XComArgBinding` or `LiteralArgBinding` — carrying the parameter's `name` and its annotation-derived `value_schema` (a JSON schema as typing system for describing the "type" of this argument).
- **Server:** `ti_run` returns the spec as a new optional `TIRunContext.arg_bindings`, resolved through the shared `DBDagBag` only for `is_stub` tasks and stripped for older clients by a new execution-API version `2026-10-30`.

---

- [x] Yes, with help of Claude Code (Fable 5) + (Opus 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
@@ -1,0 +1,93 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""
+Positional-argument binding spec for stub (foreign-runtime) tasks.
+
+Captured at parse time from the ``@task.stub`` TaskFlow call, stored in the serialized
+Dag, and delivered to the lang-SDK runtime via ``TIRunContext.arg_bindings``.
+"""
+
+from __future__ import annotations
+
+from functools import cache
+from typing import Annotated, Literal
+
+from pydantic import Field, JsonValue, TypeAdapter
+from typing_extensions import TypeAliasType
+
+from airflow.api_fastapi.core_api.base import BaseModel
+
+# A named, titled alias (like TaskArgBinding below) kept as free-form JSON rather than a
+# typed model, so unknown JSON-schema keywords survive re-serialization along the way.
+ArgValueSchema = TypeAliasType(
+    "ArgValueSchema", Annotated[dict[str, JsonValue], Field(title="ArgValueSchema")]
+)
+"""JSON-schema fragment constraining the value a stub-task argument binds to; generated
+by pydantic from the stub annotation, carried verbatim, unknown keywords ignored."""
+
+
+class _ArgBindingBase(BaseModel):
+    """Fields every :class:`TaskArgBinding` variant carries, regardless of ``kind``."""
+
+    name: str
+    """The stub function's parameter name this binding fills, in declaration order."""
+
+    value_schema: ArgValueSchema | None = None
+    """Schema fragment from the stub function's annotation; omitted when unconstrained."""
+
+
+class XComArgBinding(_ArgBindingBase):
+    """One positional stub-task argument pulled from an upstream task's XCom."""
+
+    # No default: it would drop ``kind`` from ``required``, and the generated task-sdk
+    # client then types it ``Literal | None``, invalid as a tagged-union discriminator.
+    kind: Literal["xcom"]
+
+    task_id: str
+    """Upstream task id whose ``return_value`` XCom is pulled."""
+
+
+class LiteralArgBinding(_ArgBindingBase):
+    """One positional stub-task argument carrying an inline literal from the Dag file."""
+
+    kind: Literal["literal"]
+    """No default, for the same generated-client reason as ``XComArgBinding.kind``."""
+
+    value: JsonValue | None = None
+    """The literal value from the Dag file."""
+
+    from_default: bool = False
+    """True when the value was filled from the stub signature's default rather than passed in the call."""
+
+
+# A named alias with an explicit title so the union lands in every schema as its own
+# named definition, which the supervisor-schema dump dedups with its task-sdk twin by title.
+TaskArgBinding = TypeAliasType(
+    "TaskArgBinding",
+    Annotated[XComArgBinding | LiteralArgBinding, Field(discriminator="kind", title="TaskArgBinding")],
+)
+"""One positional argument of a stub (foreign-runtime) task, in declaration order."""
+
+
+@cache
+def get_arg_bindings_adapter() -> TypeAdapter[list[TaskArgBinding]]:
+    """
+    Build (lazily, then cache) the adapter validating serialized dicts into ``TaskArgBinding``.
+
+    Only the stub-task path in the execution API needs it, so regular runs never pay for it.
+    """
+    return TypeAdapter(list[TaskArgBinding])
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
@@ -36,6 +36,7 @@ from airflow.api_fastapi.common.types import UtcDateTime
 from airflow.api_fastapi.core_api.base import BaseModel, StrictBaseModel
 from airflow.api_fastapi.execution_api.datamodels.asset import AssetProfile
 from airflow.api_fastapi.execution_api.datamodels.connection import ConnectionResponse
+from airflow.api_fastapi.execution_api.datamodels.task_arg_binding import TaskArgBinding
 from airflow.api_fastapi.execution_api.datamodels.variable import VariableResponse
 from airflow.utils.state import (
     DagRunState,
@@ -433,6 +434,13 @@ class TIRunContext(BaseModel):
     When resuming from deferral, this is set to the task's original ``start_date`` so the
     supervisor uses it instead of ``datetime.now()``.  This ensures ``context["ti"].start_date``
     always reflects when the task *first* started, not when it was rescheduled/resumed.
+    """
+
+    arg_bindings: list[TaskArgBinding] | None = None
+    """
+    Ordered positional-argument binding spec for stub (foreign-runtime) tasks.
+
+    ``None`` for regular tasks and for stub tasks that declare no parameters.
     """
 
 
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
@@ -36,6 +36,7 @@ from airflow.api_fastapi.common.types import UtcDateTime
 from airflow.api_fastapi.core_api.base import BaseModel, StrictBaseModel
 from airflow.api_fastapi.execution_api.datamodels.asset import AssetProfile
 from airflow.api_fastapi.execution_api.datamodels.connection import ConnectionResponse
+from airflow.api_fastapi.execution_api.datamodels.task_arg_binding import TaskArgBinding
 from airflow.api_fastapi.execution_api.datamodels.variable import VariableResponse
 from airflow.utils.state import (
     DagRunState,
@@ -437,6 +438,13 @@ class TIRunContext(BaseModel):
     When resuming from deferral, this is set to the task's original ``start_date`` so the
     supervisor uses it instead of ``datetime.now()``.  This ensures ``context["ti"].start_date``
     always reflects when the task *first* started, not when it was rescheduled/resumed.
+    """
+
+    arg_bindings: list[TaskArgBinding] | None = None
+    """
+    Ordered positional-argument binding spec for stub (foreign-runtime) tasks.
+
+    ``None`` for regular tasks and for stub tasks that declare no parameters.
     """
 
 
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -32,7 +32,7 @@ from fastapi import Body, HTTPException, Query, Response, Security, status
 from opentelemetry import trace
 from opentelemetry.trace import StatusCode
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
-from pydantic import JsonValue
+from pydantic import JsonValue, ValidationError
 from sqlalchemy import and_, func, or_, tuple_, update
 from sqlalchemy.engine import CursorResult
 from sqlalchemy.exc import DataError, NoResultFound, SQLAlchemyError
@@ -50,6 +50,7 @@ from airflow.api_fastapi.common.db.dags import eager_load_teams
 from airflow.api_fastapi.common.types import UtcDateTime
 from airflow.api_fastapi.compat import HTTP_422_UNPROCESSABLE_CONTENT
 from airflow.api_fastapi.core_api.openapi.exceptions import create_openapi_http_exception_doc
+from airflow.api_fastapi.execution_api.datamodels.task_arg_binding import get_arg_bindings_adapter
 from airflow.api_fastapi.execution_api.datamodels.taskinstance import (
     InactiveAssetsResponse,
     PreviousTIResponse,
@@ -75,6 +76,11 @@ from airflow.api_fastapi.execution_api.security import (
     ExecutionAPIRoute,
     get_team_name_for_ti,
     require_auth,
+)
+from airflow.api_fastapi.execution_api.services.task_instances import (
+    LANG_SDK_OPERATORS,
+    client_supports_arg_bindings,
+    get_arg_bindings,
 )
 from airflow.configuration import conf
 from airflow.exceptions import InvalidPartitionKeyError, TaskNotFound
@@ -164,6 +170,8 @@ def ti_run(
             TI.hostname,
             TI.unixname,
             TI.pid,
+            TI.operator,
+            TI.dag_version_id,
             # This selects the raw JSON value, bypassing the deserialization -- we want that to happen on the
             # client
             column("next_kwargs", JSON),
@@ -308,6 +316,30 @@ def ti_run(
             xcom_keys_to_clear=xcom_keys,
             should_retry=_is_eligible_to_retry(previous_state, ti.try_number, ti.max_tries),
         )
+
+        # Only set for lang-SDK (foreign-runtime) tasks with a captured TaskFlow arg
+        # spec; the route excludes unset fields, keeping regular responses lean.
+        if (
+            ti.operator in LANG_SDK_OPERATORS
+            and client_supports_arg_bindings()
+            and (arg_bindings := get_arg_bindings(dag_bag, ti, session=session))
+        ):
+            try:
+                context.arg_bindings = get_arg_bindings_adapter().validate_python(arg_bindings)
+            except ValidationError:
+                log.exception(
+                    "Serialized arg_bindings spec failed validation",
+                    dag_id=ti.dag_id,
+                    task_id=ti.task_id,
+                    dag_version_id=ti.dag_version_id,
+                )
+                raise HTTPException(
+                    status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    detail={
+                        "reason": "invalid_arg_bindings",
+                        "message": "The serialized TaskFlow arg spec for this stub task is not valid.",
+                    },
+                )
 
         # Only set if they are non-null
         if ti.next_method:
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -78,7 +78,6 @@ from airflow.api_fastapi.execution_api.security import (
     require_auth,
 )
 from airflow.api_fastapi.execution_api.services.task_instances import (
-    LANG_SDK_OPERATORS,
     client_supports_arg_bindings,
     get_arg_bindings,
 )
@@ -319,10 +318,8 @@ def ti_run(
 
         # Only set for lang-SDK (foreign-runtime) tasks with a captured TaskFlow arg
         # spec; the route excludes unset fields, keeping regular responses lean.
-        if (
-            ti.operator in LANG_SDK_OPERATORS
-            and client_supports_arg_bindings()
-            and (arg_bindings := get_arg_bindings(dag_bag, ti, session=session))
+        if client_supports_arg_bindings() and (
+            arg_bindings := get_arg_bindings(dag_bag, ti, session=session)
         ):
             try:
                 context.arg_bindings = get_arg_bindings_adapter().validate_python(arg_bindings)
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -32,7 +32,7 @@ from fastapi import Body, HTTPException, Query, Response, Security, status
 from opentelemetry import trace
 from opentelemetry.trace import StatusCode
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
-from pydantic import JsonValue
+from pydantic import JsonValue, ValidationError
 from sqlalchemy import and_, func, or_, tuple_, update
 from sqlalchemy.engine import CursorResult
 from sqlalchemy.exc import DataError, NoResultFound, SQLAlchemyError
@@ -49,6 +49,7 @@ from airflow.api_fastapi.common.db.common import SessionDep
 from airflow.api_fastapi.common.types import UtcDateTime
 from airflow.api_fastapi.compat import HTTP_422_UNPROCESSABLE_CONTENT
 from airflow.api_fastapi.core_api.openapi.exceptions import create_openapi_http_exception_doc
+from airflow.api_fastapi.execution_api.datamodels.task_arg_binding import get_arg_bindings_adapter
 from airflow.api_fastapi.execution_api.datamodels.taskinstance import (
     InactiveAssetsResponse,
     PreviousTIResponse,
@@ -74,6 +75,11 @@ from airflow.api_fastapi.execution_api.security import (
     ExecutionAPIRoute,
     get_team_name_for_ti,
     require_auth,
+)
+from airflow.api_fastapi.execution_api.services.task_instances import (
+    LANG_SDK_OPERATORS,
+    client_supports_arg_bindings,
+    get_arg_bindings,
 )
 from airflow.configuration import conf
 from airflow.exceptions import InvalidPartitionKeyError, TaskNotFound
@@ -163,6 +169,8 @@ def ti_run(
             TI.hostname,
             TI.unixname,
             TI.pid,
+            TI.operator,
+            TI.dag_version_id,
             # This selects the raw JSON value, bypassing the deserialization -- we want that to happen on the
             # client
             column("next_kwargs", JSON),
@@ -309,6 +317,30 @@ def ti_run(
             xcom_keys_to_clear=xcom_keys,
             should_retry=_is_eligible_to_retry(previous_state, ti.try_number, ti.max_tries),
         )
+
+        # Only set for lang-SDK (foreign-runtime) tasks with a captured TaskFlow arg
+        # spec; the route excludes unset fields, keeping regular responses lean.
+        if (
+            ti.operator in LANG_SDK_OPERATORS
+            and client_supports_arg_bindings()
+            and (arg_bindings := get_arg_bindings(dag_bag, ti, session=session))
+        ):
+            try:
+                context.arg_bindings = get_arg_bindings_adapter().validate_python(arg_bindings)
+            except ValidationError:
+                log.exception(
+                    "Serialized arg_bindings spec failed validation",
+                    dag_id=ti.dag_id,
+                    task_id=ti.task_id,
+                    dag_version_id=ti.dag_version_id,
+                )
+                raise HTTPException(
+                    status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    detail={
+                        "reason": "invalid_arg_bindings",
+                        "message": "The serialized TaskFlow arg spec for this stub task is not valid.",
+                    },
+                )
 
         # Only set if they are non-null
         if ti.next_method:
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -169,7 +169,6 @@ def ti_run(
             TI.hostname,
             TI.unixname,
             TI.pid,
-            TI.operator,
             TI.dag_version_id,
             # This selects the raw JSON value, bypassing the deserialization -- we want that to happen on the
             # client
--- a/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
@@ -1,0 +1,67 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Business logic backing the task-instance execution routes."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+    from airflow.models.dagbag import DBDagBag
+
+# Task types (``TaskInstance.operator``, the operator class name) whose tasks carry a
+# lang-SDK ``arg_bindings`` spec. Used to gate the serialized-Dag lookup so regular tasks
+# never pay for it. The gate matches exact class names; a new lang-SDK operator adds its
+# name here.
+LANG_SDK_OPERATORS = frozenset({"_StubOperator"})
+
+
+def client_supports_arg_bindings() -> bool:
+    """
+    Whether the request's negotiated API version can receive ``arg_bindings``.
+
+    Clients on older versions never see the field (the version migration strips it from
+    the response), so the derivation must not run for them.
+
+    Rather than comparing the negotiated version by date, we check the
+    ``VersionChangeWithSideEffects`` subclass's ``is_applied`` flag; see
+    https://docs.cadwyn.dev/concepts/version_changes/#version-changes-with-side-effects
+    """
+    # Imported locally: the versions package transitively imports the routes, which import
+    # this module, so a top-level import here would be circular.
+    from airflow.api_fastapi.execution_api.versions.v2026_10_30 import AddArgBindingsToTIRunContext
+
+    return AddArgBindingsToTIRunContext.is_applied
+
+
+def get_arg_bindings(dag_bag: DBDagBag, ti: Any, *, session: Session) -> list | None:
+    """
+    Extract the stub task's TaskFlow arg spec from its Dag version.
+
+    Mapped (``.expand()``) stubs never capture a parse-time spec, so they resolve to
+    ``None`` here and keep the legacy ignored-args behavior; per-map-index delivery
+    lands in a follow-up.
+    """
+    if ti.dag_version_id is None:
+        return None
+    if (dag := dag_bag.get_dag(ti.dag_version_id, session=session)) is None:
+        return None
+    if (task := dag.task_dict.get(ti.task_id)) is None:
+        return None
+    return getattr(task, "_arg_bindings", None)
--- a/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
@@ -25,12 +25,6 @@ if TYPE_CHECKING:
 
     from airflow.models.dagbag import DBDagBag
 
-# Task types (``TaskInstance.operator``, the operator class name) whose tasks carry a
-# lang-SDK ``arg_bindings`` spec. Used to gate the serialized-Dag lookup so regular tasks
-# never pay for it. The gate matches exact class names; a new lang-SDK operator adds its
-# name here.
-LANG_SDK_OPERATORS = frozenset({"_StubOperator"})
-
 
 def client_supports_arg_bindings() -> bool:
     """
@@ -62,6 +56,6 @@ def get_arg_bindings(dag_bag: DBDagBag, ti: Any, *, session: Session) -> list | 
         return None
     if (dag := dag_bag.get_dag(ti.dag_version_id, session=session)) is None:
         return None
-    if (task := dag.task_dict.get(ti.task_id)) is None:
+    if (task := dag.task_dict.get(ti.task_id)) is None or not task.inherits_from_stub_operator:
         return None
-    return getattr(task, "_arg_bindings", None)
+    return task._arg_bindings
--- a/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
@@ -58,4 +58,4 @@ def get_arg_bindings(dag_bag: DBDagBag, ti: Any, *, session: Session) -> list | 
         return None
     if (task := dag.task_dict.get(ti.task_id)) is None or not task.is_stub:
         return None
-    return task._arg_bindings
+    return task.arg_bindings
--- a/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
@@ -56,6 +56,6 @@ def get_arg_bindings(dag_bag: DBDagBag, ti: Any, *, session: Session) -> list | 
         return None
     if (dag := dag_bag.get_dag(ti.dag_version_id, session=session)) is None:
         return None
-    if (task := dag.task_dict.get(ti.task_id)) is None or not task.inherits_from_stub_operator:
+    if (task := dag.task_dict.get(ti.task_id)) is None or not task.is_stub:
         return None
     return task._arg_bindings
--- a/airflow-core/src/airflow/api_fastapi/execution_api/versions/__init__.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/versions/__init__.py
@@ -51,9 +51,11 @@ from airflow.api_fastapi.execution_api.versions.v2026_06_30 import (
     AddTeamNameField,
     AddVariableKeysEndpoint,
 )
+from airflow.api_fastapi.execution_api.versions.v2026_10_30 import AddArgBindingsToTIRunContext
 
 bundle = VersionBundle(
     HeadVersion(),
+    Version("2026-10-30", AddArgBindingsToTIRunContext),
     Version(
         "2026-06-30",
         AddVariableKeysEndpoint,
--- a/airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
+++ b/airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
@@ -32,6 +32,7 @@ from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 from opentelemetry.trace import StatusCode
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+from pydantic import ValidationError
 from sqlalchemy import select, update
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
@@ -161,6 +162,14 @@ def test_id_matches_sub_claim(client, session, create_task_instance):
 
 
 class TestTIRunState:
+    RUN_PAYLOAD = {
+        "state": "running",
+        "hostname": "random-hostname",
+        "unixname": "random-unixname",
+        "pid": 100,
+        "start_date": "2024-09-30T12:00:00Z",
+    }
+
     def setup_method(self):
         clear_db_logs()
         clear_db_runs()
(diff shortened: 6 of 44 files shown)


airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,112 @@
...
+
+Captured at parse time from a stub task's TaskFlow call (``@task.stub``), stored in the
+serialized Dag, and delivered to the lang-SDK runtime through ``TIRunContext.arg_bindings``
+so it can bind the values onto the native task function's parameters.
+
+Each binding is one variant of a union discriminated on ``kind``: an ``XComArgBinding``
+pulls the value from an upstream task's XCom, a ``LiteralArgBinding`` carries an inline
+value from the Dag file. Both variants still emit plain named structs
+(``$defs/XComArgBinding``, ``$defs/LiteralArgBinding``) for the foreign-language SDKs
+consuming the supervisor schema.
```
jason810496:
```suggestion
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,112 @@
...
+    ARRAY = "array"
+    ANY = "any"
+
+
+class XComArgBinding(BaseModel):
+    """One positional stub-task argument pulled from an upstream task's XCom."""
+
+    # No default on purpose: a required ``kind`` stays non-nullable through the OpenAPI
+    # round trip, which discriminated-union codegen needs (a defaulted field turns
+    # ``Literal`` into ``Literal | None`` in the generated task-sdk models).
```
jason810496:
```suggestion
```

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -110,6 +111,36 @@
...
+    from airflow.serialization.enums import Encoding
+    from airflow.serialization.serialized_objects import BaseSerialization
+
+    if dag_version_id is None:
+        return None
+    dag_version = session.get(DagVersion, dag_version_id)
+    if dag_version is None or dag_version.serialized_dag is None:
+        return None
+    data = dag_version.serialized_dag.data
+    if not data:
```
jason810496:
```suggestion
    if not (data := dag_version.serialized_dag.data):
```

airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
```diff
@@ -141,3 +145,28 @@ def remove_partition_date_from_dag_run(response: ResponseInfo) -> None:  # type:
...
+
+class AddArgBindingsToTIRunContext(VersionChange):
+    """
+    Add the ``arg_bindings`` positional-argument binding spec for stub (foreign-runtime) tasks.
+
+    Each entry is a discriminated union of ``XComArgBinding`` and ``LiteralArgBinding`` keyed
+    on ``kind``. ``data_type`` is declared as the ``ArgBindingDataType`` enum rather than an
+    inline ``Literal``; the wire representation (a JSON string) is unchanged, so no migration
+    instruction is needed -- this version has not been released with the ``arg_bindings`` field
+    in any other shape.
```
jason810496:
```suggestion
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,112 @@
...
+    """The stub function's parameter name this binding fills, in declaration order."""
+
+    data_type: ArgBindingDataType = ArgBindingDataType.ANY
+    """Declared type from the stub function's annotation; runtimes type-check against it."""
+
+    task_id: str
+    """Upstream task id to pull the XCom from."""
+
+    key: str = "return_value"
+    """XCom key to pull."""
```
jason810496: The `key` will always be `return_value` across TaskFlow, we can remove the `key` field from end to end.

go-sdk/dags/go_examples.py
```diff
@@ -107,3 +119,90 @@ def concurrent_xcom_dag():
...
+        3,
+        2.5,
+        True,
+        ["metrics", "hourly"],
+        config=make_config(),
+        numbers=make_numbers(),
+    )
+    via_struct_no_tags(RegionCode="eu-west-1", Threshold=0.75)
+    via_struct_arg_tag(region_code="eu-west-1", threshold=0.75)
+    via_struct_unmatched_arg(region_code="eu-west-1")
```
jason810496: For via_struct_no_tags, via_struct_arg_tag and via_struct_unmatched_arg tasks, we should make the parameters reference the value from XCom as well, not just exercising the literal case only. We need to make the changes e2e.

go-sdk/pkg/binding/binding.go
```diff
@@ -0,0 +1,701 @@
...
+
+const (
+	DataTypeString  DataType = "string"
+	DataTypeInteger DataType = "integer"
+	DataTypeNumber  DataType = "number"
+	DataTypeBoolean DataType = "boolean"
+	DataTypeObject  DataType = "object"
+	DataTypeArray   DataType = "array"
+	DataTypeAny     DataType = "any"
+)
```
jason810496: We should reference these from `go-sdk/pkg/execution/genmodels/models.gen.go` instead of hand-written to prevent drift

go-sdk/pkg/binding/binding.go
```diff
@@ -0,0 +1,701 @@
...
+
+// LiteralArg carries an inline value from the Dag file.
+type LiteralArg struct {
+	// Name is the stub function's parameter name this binding fills.
+	Name string
+	// Value is the literal value from the Dag file.
+	Value any
+	// DataType is the declared type to check the Go parameter against.
+	DataType DataType
+}
```
jason810496: We should reference these from go-sdk/pkg/execution/genmodels/models.gen.go instead of hand-written to prevent drift if possible.

go-sdk/pkg/binding/binding.go
```diff
@@ -0,0 +1,701 @@
...
+//     the argument to claim, and a field with no tag claims its own Go field
+//     name, verbatim. At most one such parameter is allowed per function.
+//
+// A TaskInput struct's fields are resolved first, by name, claiming entries
+// out of the argument spec; the remaining unclaimed entries are then
+// distributed, in their original relative order, onto the plain flat data
+// parameters in declaration order -- so flat parameters and a TaskInput
+// struct can coexist in the same function signature regardless of where each
+// sits, and (with no TaskInput struct present) this reduces to exactly
+// today's positional-only behaviour.
```
jason810496: We should make it either using flat-base or TaskInput struct-base. Not supporting flat-base plus TaskInput struct-base, which is too ambiguous.

ts-sdk/src/generated/supervisor.ts
```diff
@@ -22,6 +22,20 @@
 //
 // Re-run with: pnpm run generate:supervisor
 
+/**
```
jason810496: We should not bump `ts-sdk/src/generated/supervisor.ts` in this PR.

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -310,6 +332,30 @@ def ti_run(
...
+                    dag_id=ti.dag_id,
+                    task_id=ti.task_id,
+                )
+                raise HTTPException(
+                    status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    detail={
+                        "reason": "invalid_arg_bindings",
+                        "message": (
+                            "The serialized TaskFlow arg spec for this stub task is not valid on "
+                            "this Airflow version; it may come from a newer providers release."
```
jason810496:
```suggestion
                            "The serialized TaskFlow arg spec for this stub task is not valid.
```

jason810496 (commented):
I will wait until next review then address my own comments to avoid CI-rerun.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,102 @@
...
+
+from airflow.api_fastapi.core_api.base import BaseModel
+
+
+class ArgBindingDataType(str, Enum):
+    """Language-neutral value type a stub-task argument binds to in the foreign runtime."""
+
+    STRING = "string"
+    INTEGER = "integer"
+    NUMBER = "number"
```
ashb: JSON has no distinction between integer and number. Do we need both types, or is Number alone enough? (In go for example we already have to possibly/conditional convert Integer to different int types (int32, int, in64 etc etc)

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,102 @@
...
+    BOOLEAN = "boolean"
+    OBJECT = "object"
+    ARRAY = "array"
+    ANY = "any"
+
+
+class XComArgBinding(BaseModel):
+    """One positional stub-task argument pulled from an upstream task's XCom."""
+
+    kind: Literal["xcom"]
```
ashb: Nit?
```suggestion
    kind: Literal["xcom"] = "xcom"
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
```diff
@@ -435,6 +436,13 @@ class TIRunContext(BaseModel):
     always reflects when the task *first* started, not when it was rescheduled/resumed.
     """
 
+    arg_bindings: list[TaskArgBinding] | None = None
+    """
+    Ordered positional-argument binding spec for stub (foreign-runtime) tasks.
+
+    ``None`` for regular tasks and for stub tasks that declare no parameters.
+    """
```
ashb, line 444: Hmmmmm, I wonder if this should not allow none, and make it an empty list in that case. I don't think it functionally makes a difference but... 🤔

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -310,6 +332,30 @@ def ti_run(
...
+        if ti.operator == _STUB_TASK_TYPE and (
+            arg_bindings := _get_arg_bindings(dag_bag, ti.dag_version_id, ti.task_id, session=session)
+        ):
+            try:
+                context.arg_bindings = get_arg_bindings_adapter().validate_python(arg_bindings)
+            except ValidationError:
+                log.exception(
+                    "Serialized arg_bindings spec failed validation",
+                    dag_id=ti.dag_id,
+                    task_id=ti.task_id,
```
ashb, line 330:
```suggestion
                    task_id=ti.task_id,
                    dag_version_id=ti.dag_version_id,
```

airflow-core/src/airflow/api_fastapi/execution_api/versions/__init__.py
```diff
@@ -51,9 +51,11 @@
     AddTeamNameField,
     AddVariableKeysEndpoint,
 )
+from airflow.api_fastapi.execution_api.versions.v2026_07_30 import AddArgBindingsToTIRunContext
```
ashb: What Airflow version are we targeting this for release? If its 3.4 then this should be further in the future, say 2026_10_30.

Or do you want to make a case that adding support for TaskFlow in go is a bug fix?

airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
```diff
@@ -370,6 +371,87 @@ async def workload_token(request: Request) -> TIToken:
...
+            "unixname": "random-unixname",
+            "pid": 100,
+            "start_date": "2024-09-30T12:00:00Z",
+        }
+
+        response = client.patch(f"/execution/task-instances/{tis['transform'].id}/run", json=payload)
+        assert response.status_code == 200
+        assert response.json()["arg_bindings"] == [
+            {"name": "country", "kind": "literal", "data_type": "string", "value": "uk"},
+            {"name": "extracted", "kind": "xcom", "data_type": "object", "task_id": "extract"},
```
ashb: data_type on an xcom arg feels slightly odd here -- we don't/cant know what the type is from the dag defn really, as it depends on what the task actually pushes doesn't it?


```python
@task
def upstream():
    if CONDITION:
        return {"data": "one"}
    else
        return False
```

Extreme case that Golang tasks may never be able to cope with, but regardless, we don't know the type of an xcom.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,102 @@
...
+from functools import cache
+from typing import Annotated, Literal
+
+from pydantic import Field, JsonValue, TypeAdapter
+from typing_extensions import TypeAliasType
+
+from airflow.api_fastapi.core_api.base import BaseModel
+
+
+class ArgBindingDataType(str, Enum):
```
ashb: What about Dates/DateTimes?

We handle those specially in xcom serialization already don't we?, and it's likely quite common to want to pass those along, so I think we should handle the "convert to target language native date type" rather than just passing it as a string and making task code parse it.

ashb: I wonder if we can reuse something from JSON Schema here instead of our own types...

airflow-core/tests/unit/serialization/test_dag_serialization.py
ashb, line 1: I wonder if introducing arg_bindings should result in a bump in the serialization version? @amoghrajesh WDYT?

airflow-core/tests/unit/serialization/test_dag_serialization.py
```diff
@@ -3405,6 +3405,44 @@ def inner():
...
+    ]
+
+    round_tripped = DagSerialization.from_dict(ser_dag)
+    assert round_tripped.task_dict["transform"]._arg_bindings == [
+        {"name": "country", "kind": "literal", "data_type": "string", "value": "uk"},
+        {"name": "extracted", "kind": "xcom", "data_type": "object", "task_id": "extract"},
+    ]
+    assert not hasattr(round_tripped.task_dict["extract"], "_arg_bindings") or (
+        round_tripped.task_dict["extract"]._arg_bindings is None
+    )
```
ashb: not hasattr or is None feels a bit odd. Pick one to assert.

providers/standard/src/airflow/providers/standard/decorators/stub.py
ashb, line 1: Given how tightly integrated into the exec API this change is, I'm not sure putting it in standard provider is right -- my first thought is that this should live in/with the dag parsing code, not with the stub operator code?

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,22 +18,170 @@
...
+        if name not in explicitly_bound:
+            entry["from_default"] = True
+        spec.append(entry)
+    return spec
+
+
 class _StubOperator(DecoratedOperator):
     custom_operator_name: str = "@task.stub"
 
+    # Mapped stubs would need per-map-index arg specs, which the foreign runtime cannot
```
ashb: "Mapped stubs would need per-map-index arg specs"

I don't think this is true -- it's the same function for each mapped index (by design) so each index would receive the same type of arguments.

ashb (commented):
Overall I like the direction. Almost all of my comments I can be challenged on, don't just make the changes if you think the current way is better/more correct

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
jason810496: Good call. I will make these JSON schema to make the "type" distinguishable across language.
https://json-schema.org/understanding-json-schema/reference/type#format

airflow-core/src/airflow/api_fastapi/execution_api/versions/__init__.py
jason810496: > What Airflow version are we targeting this for release? If its 3.4 then this should be further in the future, say 2026_10_30.

It's 3.4, I will make it 2026_10_30 then.

airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
jason810496: Actually, all the "data_type" here only comes from the **annotation** at the Stub Operator level. So what exactly the upstream task return doesn't really matter IMO.

There comes up another case that I needs to resolve. The case you give, user might annotate the argument as `dict | bool`, but I haven't deal with the union type annotation yet.

airflow-core/tests/unit/serialization/test_dag_serialization.py
jason810496: I will update the `airflow-core/src/airflow/serialization/schema.json` to show the new `args_binding` field (the native Dag TaskFlow will leverage the `args_binding` field as well). However, I don't think it's not necessary to bump the "serialization version".

providers/standard/src/airflow/providers/standard/decorators/stub.py
jason810496: I will look into whether it would be better to change the dag parsing side or to keep it as is.

providers/standard/src/airflow/providers/standard/decorators/stub.py
jason810496: I will double check this part.

jason810496 (commented):
Thanks you Ash for the review. I will address the comments shortly.

providers/standard/tests/unit/standard/decorators/test_stub.py
```diff
@@ -282,51 +282,91 @@ def group(n):
...
         pytest.param(Any, None, id="any"),
         pytest.param(None, None, id="none"),
         pytest.param(type(None), None, id="nonetype"),
-        pytest.param(bytes, None, id="bytes"),
+        pytest.param(
+            pendulum.DateTime,
+            None,
+            id="pendulum-datetime",
+            # pydantic has no schema for arbitrary datetime subclasses; decode-only fallback.
+        ),
```
jason810496: The _ValueSchemaGenerator should handle the `pendulum.DateTime` as it's really common in Airflow.

providers/standard/pyproject.toml
```diff
@@ -61,6 +61,8 @@ requires-python = ">=3.10"
 dependencies = [
     "apache-airflow>=2.11.0",
     "apache-airflow-providers-common-compat>=1.14.1",  # use next version
+    # The stub decorator generates arg-binding value schemas from parameter annotations
+    "pydantic>=2.11.0",
```
jason810496: `pydantic` is already part of the dependency in Airflow. I don't think we need to add them explicitly here (which might also cause further dependency conflict)

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -32,36 +32,21 @@
...
-# A named alias (like TaskArgBinding below) so both union branches of ArgValueSchema.type
-# reference one shared schema definition instead of two inlined enum copies; the explicit
-# title lets the supervisor-schema dump merge this def with the task-sdk-generated twin.
-JsonSchemaType = TypeAliasType(
-    "JsonSchemaType",
-    Annotated[
-        Literal["string", "integer", "number", "boolean", "object", "array", "null"],
-        Field(title="JsonSchemaType"),
-    ],
+# A named alias (like TaskArgBinding below) so every schema keeps one shared, titled
```
jason810496: Shorting all these comments in this module we introduced as they're too verbose.

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -39,84 +40,46 @@
...
-    Map a stub function parameter annotation to a JSON-schema fragment.
-
-    Fragments carry the standard ``type`` keyword -- a single name, or a list for union
-    annotations (set semantics; member order follows the annotation) -- plus a ``format``
-    annotation where the Python type implies a wire representation the type name alone
-    cannot (``int`` -> ``int64``, ``datetime`` -> ``date-time``, ``timedelta`` ->
-    ``duration``, ...). Returns ``None`` when the annotation gives no constraint; the
-    binding then omits ``value_schema`` and the foreign runtime falls back to a
-    decode-only check.
+    Build the JSON-schema fragment for one stub parameter annotation, via pydantic.
```
jason810496: Short the docstring comment.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+    raise HTTPException(
+        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        detail={
+            "reason": "invalid_arg_bindings",
+            "message": f"The stub task's TaskFlow arguments cannot be delivered: {detail}.",
+        },
+    )
+
+
+def _resolve_mapped_stub_arg_bindings(task: Any, ti: Any, *, session: Session) -> list[dict[str, Any]]:
```
jason810496: The task should be annotated with `Operator` where `Operator` is

```
Operator: TypeAlias = BaseOperator | MappedOperator
```

at `TYPE_CHECKING` level.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+    A mapped stub never instantiates at parse time, so no spec is captured in the
+    serialized Dag; it is derived here from the serialized expand input instead, with
+    the map-index decomposition delegated to
+    ``SchedulerDictOfListsExpandInput.resolve_expansion_sub_indexes``.
+    Value schemas come from the stub function's annotations, which are not available
+    server-side, so mapped bindings omit them (runtimes fall back to decode-only checks).
+    """
+    expand_input = task._get_specified_expand_input()
+    if not isinstance(expand_input, SchedulerDictOfListsExpandInput):
+        _unsupported_arg_bindings("expand_kwargs() is not supported on stub tasks")
```
jason810496:
```suggestion
    # TODO: Support the `expand_kwargs` path once #69757 merged and all the Lang-SDKs adapt it.
    if not isinstance(expand_input, SchedulerDictOfListsExpandInput):
        _unsupported_arg_bindings("expand_kwargs() is not supported on stub tasks")
```

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+    if ti.map_index < 0:
+        _unsupported_arg_bindings("the task instance has not been expanded to a map index")
+
+    expand_value = expand_input.value
+    try:
+        sub_indexes = expand_input.resolve_expansion_sub_indexes(ti.map_index, ti.run_id, session=session)
+    except NotFullyPopulated as e:
+        _unsupported_arg_bindings(f"upstream map lengths are not yet known for {sorted(e.missing)}")
```
jason810496:
```suggestion
    except NotFullyPopulated as e:
        # In most of the happy path, this shouldn't happed at all unless someone clear upstream TIs or XCom themself during the DagRun.
        _unsupported_arg_bindings(f"upstream map lengths are not yet known for {sorted(e.missing)}")
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -0,0 +1,102 @@
...
+
+    task_id: str
+    """Upstream task id whose ``return_value`` XCom is pulled."""
+
+    map_index: int = -1
+    """Map index of the upstream XCom row to pull; -1 is the unmapped row."""
+
+    element_index: int | None = None
+    """When set, the pulled value is a sequence and this binding takes the element at
+    this index (the stub was expanded over an unmapped upstream's output)."""
```
jason810496:
```suggestion
    this index (the stub was expanded over an unmapped upstream's output). The Lang-SDK side will GET the single row (task_id, map_index=-1), decode it, take value[i]."""
```

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+
+    spec = [
+        _bind_mapped_stub_arg(name, value, sub_index=None)
+        for name, value in (task.partial_kwargs.get("op_kwargs") or {}).items()
+    ]
+    spec += [
+        _bind_mapped_stub_arg(name, value, sub_index=sub_indexes[name])
+        for name, value in expand_value.items()
+    ]
```
jason810496:
```suggestion

    # partial() kwargs
    spec = [
        _bind_mapped_stub_arg(name, value, sub_index=None)
        for name, value in (task.partial_kwargs.get("op_kwargs") or {}).items()
    ]
    # expand() kwargs
    spec += [
        _bind_mapped_stub_arg(name, value, sub_index=sub_indexes[name])
        for name, value in expand_value.items()
    ]
```

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+            else:
+                entry["element_index"] = sub_index
+        return entry
+    if isinstance(value, SchedulerXComArg):
+        _unsupported_arg_bindings(
+            f"parameter {name!r} received a {type(value).__name__}; only direct upstream"
+            " task outputs and literals are supported"
+        )
+    if sub_index is not None:
+        items = list(value.items()) if isinstance(value, dict) else value
```
jason810496:
```suggestion
    if sub_index is not None:
        #  this kwarg was expanded over a literal collection written in the Dag file
        items = list(value.items()) if isinstance(value, dict) else value
```

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,128 @@
...
+        for name, value in (task.partial_kwargs.get("op_kwargs") or {}).items()
+    ]
+    spec += [
+        _bind_mapped_stub_arg(name, value, sub_index=sub_indexes[name])
+        for name, value in expand_value.items()
+    ]
+    return spec
+
+
+def _bind_mapped_stub_arg(name: str, value: Any, *, sub_index: int | None) -> dict[str, Any]:
```
jason810496: Would it be possible to set the `value_schema` (typing) as well?

jason810496 (commented):
Self-review comments.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
jason810496: It's a large change. Defer to https://github.com/apache/airflow/issues/70523.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,65 @@
...
+
+def client_supports_arg_bindings() -> bool:
+    """
+    Whether the request's negotiated API version can receive ``arg_bindings``.
+
+    Clients on older versions never see the field (the version migration strips it from
+    the response), so the derivation must not run for them: a stub Dag that ran before
+    arg bindings existed keeps running against those clients. The version that introduced
+    the field is the single source of truth via Cadwyn's ``is_applied`` side-effect check.
+    """
```
jason810496, line 39:
```suggestion
    Clients on older versions never see the field (the version migration strips it from
    the response), so the derivation must not run for them.
    
    Additionally, we shouldn't distinguish by date, we should check the `VersionChangeWithSideEffects` subclass `.is_applied`.
    Reference: https://docs.cadwyn.dev/concepts/version_changes/?h=versionchangewithsideeffects#version-changes-with-side-effects
    """
```

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,262 @@
...
+def _ensure_json_literal(value: Any, task_id: str, name: str) -> None:
+    try:
+        json.dumps(value, allow_nan=False)
+    except (TypeError, ValueError):
+        raise ValueError(
+            f"@task.stub task {task_id!r} parameter {name!r} received a literal of type "
+            f"{type(value).__name__} that is not JSON-serializable, so it cannot be passed "
+            "to the foreign runtime; pass it in its JSON form instead (e.g. datetimes, dates, "
+            "and times as ISO 8601 strings)"
+        )
```
jason810496: We handled the datetime and dates.

```suggestion
            f"@task.stub task {task_id!r} parameter {name!r} received a literal of type "
            f"{type(value).__name__} that is not JSON-serializable, so it cannot be passed "
            "to the foreign runtime; pass it in its JSON form instead.
        )
```

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,262 @@
...
+    task_id: str,
+) -> list[dict[str, Any]] | None:
+    """
+    Bind the TaskFlow call arguments to the stub signature and build the ordered arg spec.
+
+    Each spec entry is a plain dict matching one variant of the execution API's
+    ``TaskArgBinding`` union: an ``XComArgBinding`` (``kind="xcom"``) for upstream TaskFlow
+    outputs, or a ``LiteralArgBinding`` (``kind="literal"``) for everything else. ``name`` is
+    always the stub function's parameter name, so a foreign runtime can bind by name (e.g. the
+    Go SDK's ``sdk.TaskInput`` struct fields) in addition to the existing positional order.
```
jason810496:
```suggestion
    always the stub function's parameter name, so a foreign runtime can bind by name in addition to the existing positional order.
```

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -75,7 +318,30 @@ def __init__(
...
+        # per-map-index instances of the tasks inside it with no expand input of their own,
+        # so their arg values are unresolvable both here and server-side.
+        in_mapped_group = self.get_closest_mapped_task_group() is not None
+        if self._arg_bindings is not None and in_mapped_group:
+            raise ValueError(
+                f"@task.stub task {self.task_id!r} passes TaskFlow call arguments inside a mapped "
+                "task group; the captured spec cannot carry values that resolve per map index at "
+                "runtime, so stub tasks with arguments are not supported under a task group's "
+                ".expand()"
+            )
```
jason810496: Should we move the `in_mapped_group = self.get_closest_mapped_task_group() is not None` validation into the `_build_arg_bindings`? Or at least before the `_build_arg_bindings` call.

providers/standard/pyproject.toml
```diff
@@ -60,7 +60,8 @@ requires-python = ">=3.10"
 # After you modify the dependencies, and rebuild your Breeze CI image with ``breeze ci-image build``
 dependencies = [
     "apache-airflow>=2.11.0",
-    "apache-airflow-providers-common-compat>=1.14.1",
+    # 1.18.0 first ships KNOWN_CONTEXT_KEYS and PlainXComArg, which the stub decorator imports.
+    "apache-airflow-providers-common-compat>=1.18.0",  # use next version
```
jason810496:
```suggestion
    "apache-airflow-providers-common-compat>=1.18.0",  # use next version
```

jason810496 (commented):
Self-review.

jason810496 (commented):
Hi @ashb,

Here are the key updates since your last review:
- Replaced the data_type enum with pydantic-generated JSON-schema fragments, carried per argument as value_schema.
- Kept the argument materialization in the execution API:
  - In the existing Python world, the worker re-parses the Dag file and deserializes the operator, so it gets the call arguments for free.
  - A lang-SDK runtime can't parse Python, so it has to receive materialized bindings.
  - Doing this at Dag-processing time alone isn't enough -- resolving per-map-index values requires joining the TaskInstance at task runtime so ti_run in the execution API is the right place (it's also where API version negotiation strips the field for older clients).
- Deferred mapped-operator support to #70570 and #70571 to keep this one concise enough to review.

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,277 @@
...
+
+def _ensure_json_literal(value: Any, task_id: str, name: str) -> None:
+    try:
+        json.dumps(value, allow_nan=False)
+    except (TypeError, ValueError):
+        raise ValueError(
+            f"@task.stub task {task_id!r} parameter {name!r} received a literal of type "
+            f"{type(value).__name__} that is not JSON-serializable, so it cannot be passed "
+            "to the foreign runtime; pass it in its JSON form instead"
+        )
```
phanikumv:
```suggestion
  def _ensure_json_literal(value: Any, task_id: str, name: str) -> None:
      if next(XComArg.iter_xcom_references(value), None) is not None:
          raise ValueError(
              f"@task.stub task {task_id!r} parameter {name!r} received a collection with an "
              "upstream task output nested inside it; only a direct XComArg argument can cross "
              "the language boundary -- pass the upstream output as its own argument"
          )
      try:
          json.dumps(value, allow_nan=False)
      except (TypeError, ValueError):
          raise ValueError(
              f"@task.stub task {task_id!r} parameter {name!r} received a literal of type "
              f"{type(value).__name__} that is not JSON-serializable, so it cannot be passed "
              "to the foreign runtime; pass it in its JSON form instead"
          )
  ```

Comment (phanikumv) [reply providers/standard/src/airflow/providers/standard/decorators/stub.py]: Before the json.dumps, walk the value for embedded XCom references using the same helper that the regular TaskFlow uses..

Comment (jason810496) [reply providers/standard/src/airflow/providers/standard/decorators/stub.py]: Addressed in https://github.com/apache/airflow/pull/69757/changes/748798de4de1f9d4829437652f8fb0b3a6c2937a. Thanks.

**Comment by uranusjr** [comment]
I can’t comment on the implementation, but the proposed interface looks very reasonable to me. It seems to me the described programming interface has not been fully implemented in this PR, and I don’t have enough knowledge to say whether this is on the right track or not tbh. Maybe it would be a good idea to edit the PR description to ground what exactly should be expected here.

File: airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py [hunk 38]
```diff
@@ -0,0 +1,93 @@
...
+    value_schema: ArgValueSchema | None = None
+    """Schema fragment from the stub function's annotation; omitted when unconstrained."""
+
+
+class XComArgBinding(_ArgBindingBase):
+    """One positional stub-task argument pulled from an upstream task's XCom."""
+
+    # No default: it would drop ``kind`` from ``required``, and the generated task-sdk
+    # client then types it ``Literal | None``, invalid as a tagged-union discriminator.
+    kind: Literal["xcom"]
```
ashb, line 58: Elsewhere we do this:

```python
    state: Annotated[
        Literal[TIState.RUNNING],
        # Specify a default in the schema, but not in code.
        WithJsonSchema({"type": "string", "enum": [TIState.RUNNING], "default": TIState.RUNNING}),
    ]
```

It might be worth seeing if we can do the same thing here to apply a default too?

Or this approach from ExecuteTask workload:

```python
    type: Literal["ExecuteTask"] = Field(init=False, default="ExecuteTask")
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
ashb: That isn't true. This example from https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html

```python
@task
def get_ip():
    return my_ip_service.get_main_ip()

@task(multiple_outputs=True)
def compose_email(external_ip):
    return {
        'subject':f'Server connected from {external_ip}',
        'body': f'Your server executing Airflow is connected from the external IP {external_ip}<br>'
    }

email_info = compose_email(get_ip())

EmailOperator(
    task_id='send_email_notification',
    to='example@example.com',
    subject=email_info['subject'],
    html_content=email_info['body']
)
```


shows that compose_email actually stores two different xcom_keys, `subject` and `body`, and `XcomArg`'s subscripte operator gives different keys.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
```diff
@@ -0,0 +1,67 @@
...
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+    from airflow.models.dagbag import DBDagBag
+
+# Task types (``TaskInstance.operator``, the operator class name) whose tasks carry a
+# lang-SDK ``arg_bindings`` spec. Used to gate the serialized-Dag lookup so regular tasks
+# never pay for it. The gate matches exact class names; a new lang-SDK operator adds its
+# name here.
+LANG_SDK_OPERATORS = frozenset({"_StubOperator"})
```
uranusjr: Gating with operator type name is not a good idea. Add a flag on _StubOperator, serialize it, and read it here instead.

ashb: Yes, this is what we do for EmptyOperator -- we have a "is_empty" field or similar.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
ashb: You resolved this but didn't change anything or say why it's fine as it is.

airflow-core/tests/unit/serialization/test_dag_serialization.py
amoghrajesh: I don't think this needs a bump.

Its purely additive and only appears on `@task.stub` tasks that were called with taskflow args. `definitions.operator` also has `additionalProperties: true` and `validate_schema` only runs on the write path, so nothing on the read side changes for a v3 blob. `conversion_v3_to_v4` would probably be an empty function or very less.

ashb: I agree it's not "needed", no, but is it cleaner/a better forward pattern if we start making the version mean something?

In Airflow 2 the schema version was always 1, so it effectively never got used, and I'm not sure that's a good pattern.

uranusjr:
There’s also still a mention to sdk.TaskInput that does not exist. Otherwise I think this is good (aside from the comment above).

airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
```diff
@@ -372,6 +381,110 @@ async def workload_token(request: Request) -> TIToken:
...
+        dag_maker.session.flush()
+
+        response = client.patch(f"/execution/task-instances/{tis['transform'].id}/run", json=self.RUN_PAYLOAD)
+        assert response.status_code == 200
+        assert response.json()["arg_bindings"] == [
+            {"name": "country", "kind": "literal", "value_schema": {"type": "string"}, "value": "uk"},
+            {
+                "name": "extracted",
+                "kind": "xcom",
+                "value_schema": {"type": "object", "additionalProperties": True},
```
ashb, line 409: I'm still not sure what to make of the schema/data type on an xcom arg. How will it be used?

dev/breeze/src/airflow_breeze/prepare_providers/provider_documentation.py
```diff
@@ -402,6 +402,39 @@ def _print_changes_table(changes_table):
     console_print(syntax)
 
 
+def _resolve_existing_version_tag(version_tag: str) -> str:
```
ashb: Why is this change in this PR?

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,283 @@
...
+from collections.abc import Callable, Collection, Mapping
+from functools import cache
 from typing import TYPE_CHECKING, Any
 
+try:
+    from pydantic import PydanticUserError, TypeAdapter
+    from pydantic.json_schema import GenerateJsonSchema
+except ImportError:
+    # Airflow 3 always ships pydantic but Airflow 2.x base installs do not; without it,
+    # stub args carry no value schemas and runtimes keep their decode-only fallback.
```
ashb: Stub operator should be made not-usable on Airflow2 if someone tries to instantiate one -- it wouldn't ever work anyway without the tasksdk machinery at execution time

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,283 @@
...
+
+def _validate_xcom_value(value: Any, task_id: str, name: str) -> bool:
+    """Validate an XComArg argument, returning True when it is a bindable direct upstream output."""
+    if isinstance(value, PlainXComArg):
+        if value.key != XCOM_RETURN_KEY:
+            raise ValueError(
+                f"@task.stub task {task_id!r} parameter {name!r} references the XCom key "
+                f"{value.key!r}; only an upstream task's return value can cross the language "
+                "boundary -- indexing an output by a custom key is not supported"
+            )
```
ashb: I don't think this limitation should stay.

providers/standard/src/airflow/providers/standard/decorators/stub.py
```diff
@@ -18,19 +18,283 @@
...
+                f"output of the mapped task {value.operator.task_id!r}; a foreign runtime "
+                "pulls single XCom rows, so a mapped upstream's combined output is not "
+                "supported"
+            )
+        return True
+    if isinstance(value, XComArg):
+        raise ValueError(
+            f"@task.stub task {task_id!r} parameter {name!r} received a "
+            f"{type(value).__name__}; only direct upstream task outputs can cross the "
+            "language boundary -- .map()/.zip()/.concat() results are not supported"
```
ashb: I can mayve see concat, and zip not being supported, but why no map?

Is this a "not for now" or a "it's really hard, possibly never"?

ashb:
I'm still not convinced about all this type handling logic being in the standard provider. A lot of this feels like it belongs in the dag parser, keeping the stub very simple.

dev/breeze/src/airflow_breeze/prepare_providers/provider_documentation.py
jason810496: My bad, this is a bad rebase.

providers/standard/src/airflow/providers/standard/decorators/stub.py
jason810496: All the mapped operator cases will be supported in https://github.com/apache/airflow/pull/70570 as follow-up. It's a bit difficult but it's feasible.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
jason810496: I tried the following three approaches and would like to keep the current, as-is form to avoid corrupting the generated discriminated union.

`kind` is the discriminator (`Field(discriminator="kind")`) on the `TaskArgBinding`.

That union round-trips through `datamodel-code-generator`, which regenerates `task-sdk/src/airflow/sdk/api/datamodels/_generated.py` from this model's OpenAPI schema. All three options below were tested against the real generator (`uv run --active --group codegen --project apache-airflow-task-sdk --directory task-sdk -s dev/generate_task_sdk_models.py`).

#### 1. Plain default

```python
class XComArgBinding(_ArgBindingBase):
    kind: Literal["xcom"] = "xcom"
```

`datamodel-code-generator` regenerates the client field with an added `| None`:

```diff
-    kind: Annotated[Literal["xcom"], Field(title="Kind")]
+    kind: Annotated[Literal["xcom"] | None, Field(title="Kind")] = "xcom"
```

Because a default drops the field from the OpenAPI schema's `required` list, and the generator's response to "not required" is "wrap it in `Optional`" — regardless of the fact that the underlying type is a single-member `Literal` that **needs to stay bare for the discriminator to work**.

`airflow` itself, for example `airflow.sdk.exceptions` reaches `airflow.sdk`, which eagerly touches `_generated.py`now blows up at class-definition time:

```
  File ".../task-sdk/src/airflow/sdk/api/datamodels/_generated.py", line 812, in <module>
    class TaskArgBinding(RootModel[XComArgBinding | LiteralArgBinding]):
  File ".../pydantic/_internal/_model_construction.py", line 256, in __new__
    complete_model_class(
  ...
  File ".../pydantic/_internal/_discriminated_union.py", line 473, in _infer_discriminator_values_for_inner_schema
    raise PydanticUserError(
pydantic.errors.PydanticUserError: Model 'XComArgBinding' needs field 'kind' to be of type `Literal`

For further information visit https://errors.pydantic.dev/2.13/u/discriminator-needs-literal
```

#### 2. `Field(init=False, default=...)`

```python
class XComArgBinding(_ArgBindingBase):
    kind: Literal["xcom"] = Field(init=False, default="xcom")
```

`init=False` only removes the field from `__init__`'s signature; it does nothing to whether the field is *required* for schema/validation purposes, which is what `datamodel-code-generator` and the discriminator care about.

```diff
-    kind: Annotated[Literal["xcom"], Field(title="Kind")]
+    kind: Annotated[Literal["xcom"] | None, Field(title="Kind")] = "xcom"
```

Same `PydanticUserError: Model 'XComArgBinding' needs field 'kind' to be of type Literal` on import. `init=False` and a bare default are schema-equivalent as far as `required` is concerned.

#### 3. Bare `Literal`, no default (current, as-is)

```python
class XComArgBinding(_ArgBindingBase):
    # No default: it would drop `kind` from `required`, and the generated task-sdk
    # client then types it `Literal | None`, invalid as a tagged-union discriminator.
    kind: Literal["xcom"]
```

`kind` stays in the OpenAPI schema's `required` list, so `datamodel-code-generator` emits it as a bare, non-optional `Literal`:

```python
kind: Annotated[Literal["xcom"], Field(title="Kind")]
```

`import airflow.sdk.api.datamodels._generated` succeeds cleanly. Every `TaskArgBinding` entry is built server-side as a plain dict literal in `build_arg_bindings` (`{"name": name, "kind": "xcom", ...}`), never via `XComArgBinding(...)`, so there's no call site that would benefit from a constructor-level default in the first place.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
jason810496: Sorry for leaving that hanging and I left it as `None` to keep the `None` semantic as "regular Python Task" and the argument-less `@task.stub` call.

The Lang SDK consumer side could even short circuit the further parsing based on the `None` semantic.

airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py
jason810496: Duplicate our offline discussion here:

> know "how to unpack the XCom" into native object
> > But that needs to be done at runtime after fetching the XCom -- we can't tell exactly what it might be at dag parse time.

Yes, I know. Even without the data type field, the SDK runtime can still parse into the native type. However, this means we treat the SDK runtime side annotation (kind of a client side) as source of truth for unpacking.

I prefer to keep the data type on an xcom arg with the following reasons:

1. Treat / unpack based on the server side (airflow core serialization) source of truth.
2. Make it possible to raise the early error when the type un-match on the Dag processing time (after AIP-85) for mixed Lang Dag
3. In the native Dag era, we can leverage this field to distinguish whether the upstream and downstream type is match or not in Dag processing time.

airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py
jason810496: The whole implementation is migrated to core serialization layer, so it's now all "flag-based" instead of relying on the naming. (I plan to write an ADR for this design trade-off.)

```python
if (task := dag.task_dict.get(ti.task_id)) is None or not task.is_stub:
        return None
    return task.arg_bindings
```

providers/standard/src/airflow/providers/standard/decorators/stub.py
jason810496: We only focus on the TaskFlow here and SDK runtime side even hard coded the `return_value` value. Or do you think both side should still handle the custom XCom key case? (Java SDK even remove the custom key support before the beta release as we think it unnecessary to support.)

I lean toward focusing on TaskFlow default `return_value` only.

airflow-core/tests/unit/serialization/test_dag_serialization.py
jason810496: The last time we bump -- https://github.com/apache/airflow/pull/55836/changes#diff-2f03f28d3201c630dfbae758caf95ada61ab10328de789fc464f6e27c4afb2d0 is for the Airflow 3 major release.

It seems we're unclear about the serialization versioning.

For example, should we use
- incremental versioning (just let it be `4` even though we're still in Airflow 3)
- semantic versioning (e.g. `v3_3_0_to_v3_3_4` ). Then should we also deal add `v3_3_1_to_v3_3_4` between each patch release?
- or follow the existing dated-base versioning like the execution API between each major version?

I tracked the issue in https://github.com/apache/airflow/issues/71364 as follow-up. I need some more design for this and perhaps raise out as a discussion (offline discussion first for sure).

jason810496 (commented):
Thanks all for the review, I just resolved all the comments and the PR is ready for the next review.

uranusjr (approved):
Do we need an ADR for this decisions made in this PR? Otherwise I think this is good.

jason810496:
> Do we need an ADR for this decisions made in this PR? Otherwise I think this is good.

Yes, I will add an ADR as follow-up to avoid the heavy full test CI run again.

jason810496:
Hi @ashb, please feel free to get any suggestion for follow-up when you have a moment, thanks.

PR #69761, opened 2026-07-11
Add TypedDict type hints for AirflowPlugin list fields

### LINKED ISSUE #62222: More thorough type hinting for `AirflowPlugin` attributes
### Description

Add full type hints and per-field validations for all `AirflowPlugin` attributes which are non-primitive types (dicts, lists, list of dicts, etc.) to improve IDE support and code self-documentation. 

More concretely, define attributes like `external_views` as their own strongly typed classes with field validations (possibly as Pydantic models, or dataclasses). There are some reasonable examples of what this might look like in the API response models for the plugins FastAPI routes. 

### Use case/motivation

Currently, many of the `AirflowPlugin` attributes are some combination of dictionaries or lists of dictionaries, but with little to no type hints for the expected dictionary keys / values. For example, `external_views` is a list of dicts, but the only information on valid configurations for it are the brief plugin docs code comments. 

As such, it is easy to misconfigure these attributes, leading to uncaught user bugs. This also could reduce some of the documentation maintenance burden for these config attributes.

### Related issues

None that I was able to find.

### Code of Conduct

sivavarma-x:
Most `AirflowPlugin` list fields (`external_views`, `react_apps`, `fastapi_apps`, `fastapi_root_middlewares`, `macros`) were typed as `list[Any]`, giving plugin authors no static signal about the shape of the entries they're expected to supply.

This PR introduces `TypedDict` definitions in `airflow_shared.plugins_manager` that mirror the corresponding API response models (`ExternalViewResponse`, `ReactAppResponse`, `FastAPIAppResponse`, `FastAPIRootMiddlewareResponse`) and re-exports them from `airflow.plugins_manager` so plugin authors can annotate their dict literals.

### What's typed

- `ExternalViewDict` — mirrors `ExternalViewResponse`.
- `ReactAppDict` — mirrors `ReactAppResponse`.
- `FastAPIAppDict` — mirrors `FastAPIAppResponse`.
- `FastAPIRootMiddlewareDict` — mirrors `FastAPIRootMiddlewareResponse`.
- `_BaseUIDict` factors the common UI fields (`name`, `icon`, `icon_dark_mode`, `url_route`, `category`, `nav_top_level`).
- `BaseDestinationLiteral` factors the `destination` values shared between `ExternalViewDict` and `ReactAppDict`; `ReactAppDict` extends it with `"dashboard"`, `"dag_overview"`, `"task_overview"`.
- `macros` is narrowed from `list[Any]` to `list[Callable[..., Any]]`.
- `_get_ui_plugins()` return type is narrowed from `tuple[list[Any], list[Any]]` to `tuple[list[ExternalViewDict], list[ReactAppDict]]`.

### What's left as `list[Any]`

- `flask_blueprints` — plugins pass `flask.Blueprint` instances, not dicts.
- `admin_views`, `menu_links` — legacy Flask-Admin.
- `appbuilder_views`, `appbuilder_menu_items` — deprecated in the API response in favor of `external_views`.
- `operator_extra_links`, `global_operator_extra_links`, `listeners`, `timetables`, `windows`, `partition_mappers`, `deadline_references`, `hook_lineage_readers`, `priority_weight_strategies` — class instances, not dict-shaped.

### Provider updates

`common.ai` (HITL Review) and `edge3` pick up the new `FastAPIAppDict` / `ReactAppDict` types under `TYPE_CHECKING` guards — no runtime cost.

### Backwards compatibility

Runtime behavior is unchanged — `TypedDict` is a plain `dict`, so every existing plugin continues to load and render as before. The changes are type-only.
Plugin authors running `mypy` will now see errors for missing required fields, misspelled keys, or `destination` values outside the documented set.

closes: #62222

---

- [X] Yes — gpt-5.5 and Claude Opus 4.7

Generated-by: gpt-5.5 and Claude Opus 4.7 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


shared/plugins_manager/src/airflow_shared/plugins_manager/plugins_manager.py
```diff
@@ -85,6 +87,52 @@ class AirflowPluginException(Exception):
     """Exception when loading plugin."""
 
 
+# Kept in sync with ``BaseDestinationLiteral`` in ``airflow.api_fastapi.core_api.datamodels.plugins``
```
Lee-W: nit: This kind of "kept in sync" thing is easy to miss. We could add a prek hook to automate it if we really need to keep these in sync.

jason810496: Another direction. It would be better to just define here, but make `airflow.api_fastapi.core_api.datamodels.plugins` import the `BaseDestinationLiteral` from here.

https://github.com/apache/airflow/blob/b3b62fa8cf3c32cb45bb2815385a2af498300754/airflow-core/src/airflow/api_fastapi/core_api/datamodels/plugins.py#L70-L73

Since they're exactly same "Literal type annotation" instead of for example one as Enum and another as Literal.

airflow-core/tests/unit/plugins/test_plugins_manager.py
```diff
@@ -241,8 +244,14 @@ def test_should_warning_about_external_views_or_react_app_wrong_object(self, cap
...
-            external_views = [[{"nested_list": "/test_route"}], {"url_route": "/test_route"}]
-            react_apps = [[{"nested_list": "/test_route"}], {"url_route": "/test_route_react_app"}]
+            external_views = [
+                [{"nested_list": "/test_route"}],
+                {"name": "A-view", "href": "/a", "url_route": "/test_route"},
+            ]
+            react_apps = [
+                [{"nested_list": "/test_route"}],
+                {"name": "A-react", "bundle_url": "/a.js", "url_route": "/test_route_react_app"},
+            ]
```
jason810496: Could we add a new cases instead of change the existing cases to ensure the compatibility.

jason810496 (commented):
Thanks for the improvement, LGTM overall with minor nits.

shared/plugins_manager/src/airflow_shared/plugins_manager/plugins_manager.py
sivavarma-x: Agree with Jason's direction here, keeping a single definition in `plugins_manager.py` and importing it into the `datamodels` layer feels cleaner than maintaining two identical Literal definitions with a "kept in sync" comment.
Pushed in the latest commit.

airflow-core/tests/unit/plugins/test_plugins_manager.py
sivavarma-x: Sure — restored the original tests with narrow `# type: ignore[typeddict-item, ...]` on the partial-dict assignments to satisfy `mypy` while preserving the original coverage, and added a new `test_loads_typed_external_views_and_react_apps` that exercises the same code path with fully-typed entries and no ignores.

jason810496 (approved):
LGTM, thanks.
I just approved the CI, and let's wait until it pass.

shared/plugins_manager/src/airflow_shared/plugins_manager/plugins_manager.py
```diff
@@ -85,6 +87,62 @@ class AirflowPluginException(Exception):
...
+    bundle_url: str
+    destination: NotRequired[Literal[BaseDestinationLiteral, "dashboard", "dag_overview", "task_overview"]]
+
+
+class FastAPIAppDict(TypedDict):
+    """Dictionary structure for entries in AirflowPlugin.fastapi_apps."""
+
+    app: Any
+    url_prefix: str
+    name: NotRequired[str]
```
Lee-W: Just looked into [FastAPIAppResponse](https://github.com/apache/airflow/blob/577496acc3b99767c24e4146e02f1a0c308d6a95/airflow-core/src/airflow/api_fastapi/core_api/datamodels/plugins.py#L32). it's not really a minor. is there any reason we need to relax it? same for `ReactAppDict` and `ExternalViewDict`

sivavarma-x: Thanks for the spot, Aligned with response models now,
Tightened `FastAPIAppDict.name` and `FastAPIRootMiddlewareDict.name` to `required`. 
Loosened the `_BaseUIDict` optionals to `NotRequired[T | None]` so authors can pass `None` or omit the key (in line with `Response` as it accepts null).

airflow-core/tests/unit/plugins/test_plugins_manager.py
```diff
@@ -212,13 +212,13 @@ def test_should_warning_about_conflicting_url_route(self, caplog):
         class TestPluginA(AirflowPlugin):
             name = "test_plugin_a"
 
-            external_views = [{"url_route": "/test_route"}, {"wrong_view": "/no_url_route"}]
+            external_views = [{"url_route": "/test_route"}, {"wrong_view": "/no_url_route"}]  # type: ignore[typeddict-item, typeddict-unknown-key]
```
Lee-W, line 241: we should add a comment on these ignores so that we know why it's ignored in the future

sivavarma-x: Added

Lee-W (approved):

PR #69762, opened 2026-07-11
Prevent KubernetesExecutor from launching stale workloads

### LINKED ISSUE #69760: KubernetesExecutor can start stale worker pods for obsolete TaskInstance UUIDs under HA schedulers

Airflow Core

### Apache Airflow version

3.3.0

Basically, we got 19 failed retries for example in a single Task all of them are failed because another Scheduler created the same task, so, using HA increases the frequency of happening this issue. This delayed our task finishing successfully for more than 30 minutes.

Image

When running Airflow 3.3.0 with `KubernetesExecutor` and multiple scheduler replicas, we observed stale Kubernetes worker pods starting after the corresponding `TaskInstance` row had already been replaced/retried.
The stale worker pod then calls:
```http
PATCH /execution/task-instances/{task_instance_id}/run
and the execution API returns:
{
  "detail": {
    "reason": "not_found",
    "message": "Task Instance not found"
  }
}
```
The worker logs this as a generic ServerResponseError.
The important detail is that the missing {task_instance_id} is an immutable TaskInstance.id UUID from an older queued workload. Looking up the same logical task by (dag_id, task_id, run_id, map_index) shows a newer TaskInstance.id with a higher try_number already queued.
Observed sequence:
1. Scheduler A queues a task instance with UUID A, try 1.
2. Kubernetes pod creation/start is delayed, for example due to cluster quota/resource pressure.
3. Another scheduler/retry path moves the same logical task forward and queues a replacement task instance with UUID B, try 2.
4. The old Kubernetes pod for UUID A eventually starts.
5. The worker calls /execution/task-instances/A/run.
6. The API returns 404 not_found because UUID A no longer exists.
7. The worker fails during startup with ServerResponseError.
This appears to be a race between the scheduler/executor queue, Kubernetes pod startup latency, and HA scheduler retry/reconciliation.
Relevant source behavior in Airflow 3.3.0:
- The /execution/task-instances/{task_instance_id}/run route looks up the row by immutable TaskInstance.id.
- KubernetesExecutor builds an ExecuteTask workload containing the current TaskInstance.id.
- There does not appear to be a DB revalidation immediately before AirflowKubernetesScheduler.run_next(...) creates the Kubernetes pod.
- Kubernetes pod annotations and watcher/adoption paths primarily use logical task identity (dag_id, task_id, run_id, map_index, try_number), not the immutable TaskInstance.id or a launch fencing token.
- TaskInstanceOperations.start() only special-cases 409 invalid_state with previous_state=running; it does not handle 404 not_found as a stale/obsolete worker.
A minimal reproduction should be possible with:
1. Airflow 3.3.0.
2. KubernetesExecutor.
3. At least two scheduler replicas.
4. A high-fanout DAG that creates many Kubernetes worker pods.
5. Artificial pod startup delay or Kubernetes quota/resource pressure.
6. Retries enabled.
The failure becomes visible when an old pending pod starts after the logical task has already been retried/requeued with a different TaskInstance.id.

A stale Kubernetes worker pod should not be able to start an obsolete task instance.
At minimum, before creating a Kubernetes pod from a serialized executor workload, KubernetesExecutor should revalidate that the workload is still current in the metadata DB, for example:
TaskInstance.id == workload.ti.id
TaskInstance.state == QUEUED
TaskInstance.try_number == workload.ti.try_number
TaskInstance.queued_by_job_id == current scheduler job id
If the row is missing or no longer matches, the executor should drop the stale workload and not create the pod.
A more robust fix would be to introduce/propagate a launch fencing token, possibly using external_executor_id or a dedicated launch generation:
1. Scheduler queues the task with a fresh launch token.
2. Kubernetes pod annotations include:
- TaskInstance.id
- launch token / external_executor_id
- try_number
- scheduler_job_id
3. Worker /run, Kubernetes watcher, and pod adoption paths validate that the pod token still matches the current DB row.
4. Stale pods are ignored/deleted instead of being adopted or allowed to update task state.
This follows the usual distributed-systems pattern of fencing tokens/CAS claims for delayed workers.
The API returning 404 for a missing immutable TaskInstance.id seems reasonable. The issue is that the executor can still launch stale work after the DB state has moved on.

### Operating System

Linux container image based on the official Apache Airflow image.

### Deployment

Other Docker-based deployment

cncf-kubernetes

apache-airflow-providers-cncf-kubernetes==10.19.0
apache-airflow-task-sdk==1.3.0

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

The issue was observed with a vanilla Airflow 3.3.0 image:
apache/airflow:3.3.0-python3.10

_No response_

### Anything else?

Related behavior seems similar to prior discussions around KubernetesExecutor and multiple schedulers, especially stale/duplicate task execution during scheduler failover or retry.
This is not the same as terminal /state duplicate update handling. The observed failure happens earlier, during worker startup, at:
PATCH /execution/task-instances/{task_instance_id}/run
with:
404 Task Instance not found
Potential patch areas:
- providers/cncf/kubernetes/.../kubernetes_executor.py
- preflight DB validation before pod creation
- providers/cncf/kubernetes/.../kubernetes_executor_utils.py
- include immutable TI identity / launch token in pod annotations
- airflow-core/.../execution_api/routes/task_instances.py
- optional launch-token validation on /run
- task-sdk/.../api/client.py
- classify /run 404 not_found as stale worker startup rather than opaque ServerResponseError
- scheduler/DagRun scheduling paths
- ensure CAS-style transitions so concurrent schedulers cannot advance the same logical task unexpectedly

### Code of Conduct

wolvery:
Fixes #69760

This adds a final DB preflight before KubernetesExecutor creates a worker pod for an Airflow 3 ExecuteTask workload. A workload can sit in the executor queue while Kubernetes pod creation is delayed, and in HA scheduler deployments the task instance can be retried, cleared, or otherwise moved on before the pod is eventually created.

The executor now verifies that the workload still matches the current queued task instance before calling run_next:

- the immutable TaskInstance id still exists
- the TaskInstance is still QUEUED
- try_number still matches the workload
- queued_by_job_id still belongs to this scheduler job

If the workload is stale, it is dropped before pod creation and executor bookkeeping for that launch is cleared.

Performance trade-off: this adds one indexed metadata DB query and one database round-trip per pod launch. The query uses the TaskInstance primary key, and the additional round-trip is an intentional correctness cost to prevent stale workloads from creating duplicate or orphaned pods.

Tests added:

- KubernetesExecutor drops a stale ExecuteTask workload before pod creation when the DB task state has moved on

Local verification:

- python3 -m compileall providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py
- AIRFLOW_HOME=/var/folders/_l/8htlzsws7gjg8_p6ht_hl0tc0000gp/T/opencode/airflow-test-home uv run pytest providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py::TestKubernetesExecutor::test_sync_drops_stale_execute_task_workload_before_pod_creation -q --with-db-init
- AIRFLOW_HOME=/var/folders/_l/8htlzsws7gjg8_p6ht_hl0tc0000gp/T/opencode/airflow-test-home uv run pytest providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py::TestKubernetesExecutor::test_run_next_exception_requeue providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py::TestKubernetesExecutor::test_skip_pod_creation_on_create_pods_after -q
- uv run ruff check providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py
- uv run ruff format --check providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py providers/cncf/kubernetes/tests/unit/cncf/kubernetes/executors/test_kubernetes_executor.py

## Gen-AI disclosure

- [X] Yes: GPT was used to assist in authoring this PR. The changes were reviewed and validated by the contributor before submission.

Generated-by: GPT following the guidelines: https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions


providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -391,6 +391,94 @@ def _process_workloads(self, workloads: Sequence[workloads.All]) -> None:
...
+            scheduler_job_id,
+        )
+        return False
+
+    def _discard_stale_pod_creation_task(self, task: KubernetesJob) -> None:
+        """Remove executor bookkeeping for a stale job that will not create a pod."""
+        self.running.discard(task.key)
+        if self.event_buffer.get(task.key) == (TaskInstanceState.QUEUED, self.scheduler_job_id):
+            self.event_buffer.pop(task.key, None)
+        Stats.incr("kubernetes_executor.stale_workload_dropped")
```
eladkal: I guess this is not backward compatible as it will be depended on newer Airflow version that ships this metric?

shared/observability/src/airflow_shared/observability/metrics/metrics_template.yaml
```diff
@@ -632,6 +632,12 @@ metrics:
     legacy_name: "-"
     name_variables: ["status"]
 
+  - name: "kubernetes_executor.stale_workload_dropped"
+    description: "Number of stale KubernetesExecutor workloads dropped before worker pod creation."
+    type: "counter"
+    legacy_name: "-"
+    name_variables: []
```
eladkal: We try to avoid having changes to providers and airflow core parts on the same PR as they don't have the same release cycle. I suggest to extract this into a separated PR. When this part is merged and schedule for release then the provider code can be adjusted under a version check

wolvery: First of all thank you for the review, and I just removed it from this PR. I'll open another PR for this metric once this PR is merged. For now the logs will be enough.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
wolvery: I think the logs might be enough at this moment

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -398,6 +398,93 @@ def _process_workloads(self, workloads: Sequence[workloads.All]) -> None:
...
+            )
+            return True
+
+        ti = session.execute(
+            select(
+                TaskInstance.id,
+                TaskInstance.state,
+                TaskInstance.try_number,
+                TaskInstance.queued_by_job_id,
+            ).where(cast(TaskInstance.id, String) == str(workload_ti.id))
```
potiuk: This needs to change before merge. `TaskInstance.id` is `Uuid()` and the table's **primary key**:

```python
id: Mapped[UUID] = mapped_column(Uuid(), primary_key=True, default=uuid7)
```

Wrapping it in `cast(..., String)` makes the predicate non-sargable — the database can no longer use the primary key index and must cast every row in `task_instance` to text to compare. That is a **full scan of the largest table in most deployments, on every pod creation**, which is a worse operational problem than the bug being fixed.

Compare the UUID directly and coerce the *parameter* instead of the column:

```python
).where(TaskInstance.id == workload_ti.id)
```

If `workload_ti.id` arrives as a `str` rather than a `UUID`, normalise it once (`UUID(str(workload_ti.id))`) before the query — never by casting the column.

Worth adding a brief note to the PR description about the added per-pod query as well: even once this is indexed, it's one extra round-trip per pod launch, which is a reasonable price for correctness but should be a conscious one.

---

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -398,6 +398,93 @@ def _process_workloads(self, workloads: Sequence[workloads.All]) -> None:
...
+        if not task.command or not isinstance(task.command[0], ExecuteTask):
+            return True
+
+        return self._should_create_pod_for_execute_task(task, task.command[0])
+
+    @provide_session
+    def _should_create_pod_for_execute_task(
+        self,
+        task: KubernetesJob,
+        workload: Any,
```
potiuk: `Any` here loses the type information the function depends on — it reads `workload.ti.id` and `workload.ti.try_number`, so a change to `ExecuteTask`'s shape would fail silently at runtime rather than in mypy.

Since `ExecuteTask` is already imported lazily in the caller for version-compat reasons, the usual pattern is a `TYPE_CHECKING` import plus a string annotation:

```python
if TYPE_CHECKING:
    from airflow.executors.workloads import ExecuteTask

def _should_create_pod_for_execute_task(self, task: KubernetesJob, workload: ExecuteTask, ...)
```

Keeps the runtime compat shim and gets the checking back.

---

potiuk (changes requested):
Thanks — the bug is real and well-described. A workload can sit in the executor queue while pod creation is delayed, and in an HA scheduler deployment the task instance can be retried, cleared or re-queued in the meantime, so launching the pod anyway produces a duplicate or orphaned run. Revalidating against the immutable TI id before creating the pod is the right shape, and checking all four of (exists, still QUEUED, `try_number`, `queued_by_job_id`) is the correct set — `queued_by_job_id` in particular is what catches the HA case.

I also checked whether this crosses an architecture boundary and it doesn't: this executor already queries the metadata DB (`@provide_session` at line 547, `select(TaskInstance.state)` at 699), so it's consistent with the existing design rather than new access.

Failing open on `ImportError` and on a non-numeric `scheduler_job_id` is the right call too — a preflight check should never be the thing that stops tasks running.

One blocking issue with the query itself, inline.

---

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
wolvery: Done.
ExecuteTask is imported under TYPE_CHECKING and used as the workload annotation while retaining runtime compatibility with older Airflow versions.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
wolvery: The lookup now normalizes the workload task-instance ID and compares it directly against TaskInstance.id, preserving use of the primary-key index.
I also updated the PR description to document the intentional cost of one indexed metadata DB query and one additional round-trip per pod launch.

potiuk (approved):
Thanks for working through all of it — re-reviewed and everything I raised is addressed.

The query is sargable now: the `cast(..., String)` is gone and the parameter is coerced instead of the column, so the primary-key index is back in play and there's no full scan on pod creation. Handling the `Uuid`-vs-`String` split by reading `TaskInstance.id.type.python_type` is a neater answer than the `UUID(str(...))` I suggested, since it also covers the older column type the provider-compat jobs run against — and the comment explains why it's there.

The `TYPE_CHECKING` import plus real `ExecuteTask` annotation restores the type checking while keeping the runtime compat shim in the caller, and the description now states the per-pod round-trip explicitly as a deliberate correctness cost. The diff is provider-only now too, which settles the earlier point about not mixing core and provider changes.

Tiny thing, take it or leave it: `ti_id_python_type(str(workload_ti.id)) if ti_id_python_type is not str else str(workload_ti.id)` — the ternary is redundant, since the first branch already reduces to `str(...)` when the type *is* `str`.

**One request before this merges:** I am approving because my own review points are resolved, but I would like a second maintainer — ideally someone who works on the cncf-kubernetes provider or the executor internals — to confirm before it goes in. This adds a metadata DB round-trip on every pod launch and changes when the executor declines to create a pod, so the HA and throttling behaviour deserves a second pair of eyes rather than resting on my re-review alone.

---

> will follow up.*
>

wolvery:
Thanks for the thorough re-review, @potiuk. 
Good catch on the redundant ternary. Collapsed it to just ti_id = ti_id_python_type(str(workload_ti.id)) — when the type already is str, str(str(...)) is a no-op, so the branch carried no weight. Pushed in 810954d3.

PR #69763, opened 2026-07-11
Fix generate-providers-metadata hang by using spawn pools

potiuk:
`breeze release-management generate-providers-metadata --refresh-constraints-and-airflow-releases` hangs forever on the multiprocessing pool.

The pools in this flow used the platform-default start method (`fork` on Linux). Before the pools are created the parent process has already used GitPython — which opens persistent `git cat-file --batch` subprocesses and is not fork-safe — and holds open network sockets from the version/constraints downloads. Forking that state into the workers leaves them with broken inherited file descriptors, so the pool deadlocks. This is most reliable with `--refresh-constraints-and-airflow-releases`, which forces the parent through git and the network before forking.

Switching the three pools in this flow to the `spawn` start method gives each worker a clean interpreter with no inherited git subprocesses or sockets.

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
@@ -34,7 +34,7 @@ from copy import deepcopy
 from datetime import datetime
 from enum import Enum
 from functools import partial
-from multiprocessing import Pool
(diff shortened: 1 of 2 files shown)


gopidesupavan (approved):
nice

PR #69764, opened 2026-07-11
Restrict asset state store access to registered tasks

fat-catTW:
## Why:

Asset state store entries are scoped to assets, but execution API task tokens should not be able to access every active asset.

Previously, any task with a valid execution token could read or mutate state for unrelated assets. This made asset state store access broader than the task's declared asset relationships.

## Solution:

Restrict the execution API asset state store routes so task tokens can only access assets registered to that task as an inlet or outlet.

The routes now check `TaskInletAssetReference` and `TaskOutletAssetReference` before allowing GET, PUT, DELETE, or CLEAR operations. Unregistered task access returns `403 Forbidden`.

Watcher writes keep using the existing non-task path because watcher tokens are not backed by task instances.

Tests cover forbidden access for unregistered tasks and allowed access for inlet/outlet-registered tasks.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
@@ -22,19 +22,21 @@ existing ``/assets/by-name`` and ``/assets/by-uri`` pattern.  Callers pass
 whichever identifier their inlet type carries: ``Asset``/``AssetNameRef`` use
 the name routes, ``AssetUriRef`` uses the URI routes.
 
-Per-task asset registration checks are intentionally not implemented here
-(deferred to AIP-93 — see TODO comment below).
+Task tokens must be registered with an asset as an inlet or outlet before
+accessing that asset's state store. Watcher tokens are not backed by task
+instances and are allowed through this route separately.
 """
 
 from __future__ import annotations
 
 import json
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
```diff
@@ -106,14 +101,51 @@ def _resolve_asset_id_by_uri(uri: str, session: SessionDep) -> int:
     return asset_id
 
 
+def _authorize_asset_state_store_access(
```
justinpakzad, line 111: Do you think it would make sense to declare this as a dependency (`dependencies=[...]`)  instead of manually calling it in each handler? That seems to be the pattern across some of the routes (e.g., `has_xcom_access()`,` has_variable_access()`).

fat-catTW: Thank you for reviewing :D

Good point, updated.
The by-name/by-uri asset state store checks now use FastAPI dependencies instead of each handler calling the authorization helper manually.

fat-catTW:
Hi @ashb @kaxil @amoghrajesh, just a gentle ping on this PR since it was recently marked as stale. 
The previous review feedback has been addressed, and the PR should be ready for another look when you have time. 
Thanks!

PR #69768, opened 2026-07-11
Multi Team: Automatically create and assign team default pools

### LINKED ISSUE #69170: Multi Team: Add team default pools for tasks
### Description

Instead of using the `default_pool` as the default Pool for tasks, when a team is created we should create a `default_<team name>` pool that contains a configurable number of slots.

### Use case/motivation

Creating default pools for each team reduces configuration overhead, and allows us to fairly share out task execution slots to teams. If the pools are adjustable per team we can decide how many tasks each team can run by default.

Currently each team that doesn't specify a pool for their tasks will be put in the default pool, which is not team aware. Therefore if there is a team running lots of jobs then they can starve out the pool for other teams.

### Related issues

_No response_

### Code of Conduct

SameerMesiah97:
**Description**

This change introduces automatic default pool management for multi-team deployments.

With this implementation, a corresponding default pool named `default_pool_<team_name>` is automatically created on team creation. During DAG parsing, tasks that do not explicitly specify a pool are assigned to their team's default pool based on the DAG bundle association, while tasks with explicitly configured pools are left unchanged.

The implementation also extends the team management commands so that `teams sync` reconciles missing default team pools for existing teams, and `teams delete` automatically removes the system-managed default team pool.

**Rationale**

In multi-team deployments, tasks without an explicit pool are currently assigned to the global `default_pool`, allowing one team's workload to consume capacity shared across all teams. Avoiding this requires users to manually create and configure team-specific pools and update every task to reference them.

By automatically provisioning a per-team default pool and applying it during DAG parsing, Airflow provides sensible defaults while preserving explicit pool assignments. This reduces operational overhead and helps isolate execution capacity between teams without requiring changes to existing DAGs.

**Tests**

* Tasks using `default_pool` are assigned to the appropriate team default pool during DAG parsing.
* Tasks with explicitly configured pools are not modified.
* DAGs without an associated team continue to use `default_pool`.
* Creating a team automatically provisions its corresponding default pool.
* `teams sync` recreates missing default team pools for existing teams.
* Deleting a team automatically removes its default team pool.
* Team default pool names are generated correctly.

**Documentation**

Updated the multi-team documentation to describe automatic creation of team default pools and how they are applied during DAG parsing.

**Backwards Compatibility**

Existing behaviour is preserved when multi-team mode is disabled. Tasks with explicitly configured pools continue to use their configured pool without modification.

Closes: #69170


airflow-core/src/airflow/dag_processing/dagbag.py
```diff
@@ -161,6 +162,30 @@ def _validate_executor_fields(dag: DAG, bundle_name: str | None = None) -> None:
...
+            bundle_config = bundle_manager._bundle_config[bundle_name]
+
+            dag_team_name = bundle_config.team_name
+
+    if not dag_team_name:
+        return
+
+    for task in dag.tasks:
+        if task.pool == Pool.DEFAULT_POOL_NAME:
+            task.pool = Pool.get_default_team_pool_name(dag_team_name)
```
justinpakzad, line 186: The addition to the `team create` command would only create the default pools for new teams going forward. That would mean for already existing teams (without explicitly assigned pools), their tasks would be assigned a pool that doesn't actually exist. Users would need to manually create the team default pools. I know multi-team is still "experimental" and relatively new, so this might not be an issue, but worth flagging.

Another potential edge case is if a user explicitly assigns the global default pool to a task, this would actually overwrite it with the default team pool. Probably not very common, but worth mentioning as this would effectively mean that the default global pool would be unreachable for team Dags.

stephen-bracken: could we check for `if task.pool is None` or `task.pool == NOTSET` instead?

airflow-core/src/airflow/models/pool.py
stephen-bracken, line 115: An alternative approach would be to make this method team aware, and pass in the team name when it is called

airflow-core/src/airflow/dag_processing/dagbag.py
SameerMesiah97: @justinpakzad @stephen-bracken 

1) I have extended the implementation reconcile the pools for each team when the `teams sync` command is executed. This should handle the scenario you mentioned. I have also handled the teams `delete` path as well when in my latest commit.

2) I agree this PR changes user-facing semantics with regards to the default pool for a certain task. My view is that 'default_pool' is better treated as a sentinel for "the default pool" rather than a guarantee that tasks always execute in a single global pool. In a multi-team deployment, the natural default is the team's default pool, while users who require a shared pool can continue to specify one explicitly

3) `pool` is initialized to "default_pool" on `SerializedBaseOperator`, so by the time we process the DAG we no longer know whether the user omitted pool or explicitly set `pool="default_pool"`

airflow-core/src/airflow/models/pool.py
SameerMesiah97: I initially considered that approach but `get_default_pool` is not part of the execution path during DAG processing and/or team creation. 

Am I missing a call site where you believe this may be used? Also, even if were to make this team-aware, we would still need handle the assignment of the team-level default pool to tasks during DAG processing.

airflow-core/src/airflow/dag_processing/dagbag.py
justinpakzad: Thanks for addressing the comments. The only thing I would still point out is that users with existing teams would need to run `teams sync` (after this change is out) to avoid tasks being assigned a pool that doesn't exist. Is it worth documenting that somewhere? 

Another minor edge case (non-blocking)  -  if someone changed the number of slots on a default team pool and then ran `teams sync` this would reset the number of slots back to the `default_pool_task_slot_count`. Cant imagine this would happen often though.

airflow-core/tests/unit/cli/commands/test_team_command.py
```diff
@@ -79,6 +79,25 @@ def test_team_create_success(self, stdout_capture):
         assert "Team 'test-team' created successfully" in output
         assert str(team.name) in output
 
+    def test_team_create_creates_default_pool(self, stdout_capture):
+        """Test that creating a team also creates its default pool."""
+        with conf_vars(
+            {
+                ("core", "multi_team"): "True",
+                ("multi_team", "default_team_pool_slots"): "128",
```
justinpakzad: One more thing I caught on the second pass. I don't think `default_team_pool_slots` is even a config option. The actual command fetches the slots via `conf.getint("core", "default_pool_task_slot_count")`. The `assert pool.slots == 128` happens to pass since the default is already 128, but if you changed the value in the conf_vars to say 64 (or any other value) and asserted, it would fail.

SameerMesiah97: Fixed.

airflow-core/src/airflow/dag_processing/dagbag.py
SameerMesiah97: I have mentioned the need to run `teams sync` in the multi-team documentation after upgrading. As for your second point, it depends whether or not you see `teams sync` as reconciliatory i.e. restoring system defaults like `default_pool_task_slot_count` vs just provisioning missing resources. I think it's equally defensible to keep it as is or preserve the setting `default_pool_task_slot_count`. What do you think?

SameerMesiah97:
@o-nikolas @ferruzzi 

Tagging you here as this is your domain i.e. multi-team.

airflow-core/src/airflow/dag_processing/dagbag.py
justinpakzad: Thanks for making that update. I don't have strong feelings on it, I think it's totally fine as is, just figured I'd mention it. Nice PR btw.

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -74,8 +75,24 @@ def team_create(args, *, session=NEW_SESSION):
...
     try:
         session.add(new_team)
+        session.flush()
+
+        if conf.getboolean("core", "multi_team"):
+            Pool.create_or_update_pool(
+                name=Pool.get_default_team_pool_name(team_name),
+                slots=conf.getint(
+                    "core",
+                    "default_pool_task_slot_count",
```
stephen-bracken: TODO: add this to the config template

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -118,9 +135,19 @@ def team_delete(args, *, session=NEW_SESSION):
...
-    if pool_count := session.scalar(select(func.count(Pool.id)).where(Pool.team_name == team.name)):
+    if pool_count := session.scalar(
+        select(func.count(Pool.id)).where(
+            Pool.team_name == team.name,
+            Pool.pool != Pool.get_default_team_pool_name(team.name),
+        )
+    ):
         associations.append(f"{pool_count} pool(s)")
 
+    default_pool = session.scalar(select(Pool).where(Pool.pool == Pool.get_default_team_pool_name(team.name)))
```
vincbeck: That's an edge case but could be named `default_pool_marketing` and be global. We should check that the pool is also associated to the team

airflow-core/docs/core-concepts/multi-team.rst
```diff
@@ -269,6 +269,14 @@ Use the ``--team-name`` option with ``airflow pools set`` to assign a pool to a
     The ``--team-name`` option is rejected when ``core.multi_team`` is disabled.
     The specified team must exist in the database (create it first with ``airflow teams create``).
 
+    When ``core.multi_team`` is enabled, ``airflow teams create`` automatically
+    creates a default pool named ``default_pool_<team_name>``. Tasks in DAG
```
o-nikolas:
```suggestion
    creates a default pool named ``default_pool_<team_name>``. Tasks in Dag
```

airflow-core/docs/core-concepts/multi-team.rst
```diff
@@ -269,6 +269,14 @@ Use the ``--team-name`` option with ``airflow pools set`` to assign a pool to a
...
+    bundles associated with that team that do not explicitly specify a pool will
```
o-nikolas:
```suggestion
    bundles associated with that team which do not explicitly specify a pool will
```

airflow-core/docs/core-concepts/multi-team.rst
```diff
@@ -269,6 +269,14 @@ Use the ``--team-name`` option with ``airflow pools set`` to assign a pool to a
...
+    use the team's default pool automatically.
+
+    Existing multi-team deployments should run ``airflow teams sync`` after upgrading
```
o-nikolas: Multi-team is in experimental phase right now. I'm not sure we want to a very time based piece in our docs. Update the news fragment to call out this guidance about syncing existing multi-team deployments.

airflow-core/src/airflow/cli/commands/team_command.py
o-nikolas: This config already exists: https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#default-pool-task-slot-count

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -74,8 +75,24 @@ def team_create(args, *, session=NEW_SESSION):
 
     try:
         session.add(new_team)
+        session.flush()
```
o-nikolas, line 93: Do you need the flush here?

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -172,10 +199,27 @@ def team_sync(args, *, session=NEW_SESSION):
...
-            teams_added += 1
+        existing_teams = Team.get_all_team_names(session=session)
+        for team_name in dag_bundle_teams:
+            if team_name not in existing_teams:
+                session.add(Team(name=team_name))
+                teams_added += 1
+
+            session.flush()
+
+            if conf.getboolean("core", "multi_team"):
```
o-nikolas: If you're going to add a check if multi-team is enabled then maybe do it higher? Do we even want to add the teams if multi-team is not enabled?

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -172,10 +199,27 @@ def team_sync(args, *, session=NEW_SESSION):
...
+                Pool.create_or_update_pool(
+                    name=Pool.get_default_team_pool_name(team_name),
+                    slots=conf.getint(
+                        "core",
+                        "default_pool_task_slot_count",
+                    ),
+                    description=f"Default pool for team '{team_name}'",
+                    include_deferred=False,
+                    team_name=team_name,
+                    session=session,
```
o-nikolas: Wrap this in a helper or move the magic strings to be constants. It is a duplicated of the one in create_team and they will surely drift from each other over time.

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -118,9 +135,19 @@ def team_delete(args, *, session=NEW_SESSION):
...
+            Pool.team_name == team.name,
+            Pool.pool != Pool.get_default_team_pool_name(team.name),
+        )
+    ):
         associations.append(f"{pool_count} pool(s)")
 
+    default_pool = session.scalar(select(Pool).where(Pool.pool == Pool.get_default_team_pool_name(team.name)))
+
+    if default_pool:
+        session.delete(default_pool)
```
o-nikolas: This delete is happening way too soon. There is a check below "are you sure you want to delete..." you should not delete anything until after that.

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -172,10 +199,27 @@ def team_sync(args, *, session=NEW_SESSION):
...
+
+            session.flush()
+
+            if conf.getboolean("core", "multi_team"):
+                Pool.create_or_update_pool(
+                    name=Pool.get_default_team_pool_name(team_name),
+                    slots=conf.getint(
+                        "core",
+                        "default_pool_task_slot_count",
+                    ),
```
o-nikolas: This clobbers any specific pool slots admins may have set if the pool already exists. You really just want create if missing not create or update

airflow-core/src/airflow/dag_processing/dagbag.py
```diff
@@ -161,6 +162,30 @@ def _validate_executor_fields(dag: DAG, bundle_name: str | None = None) -> None:
...
+        if bundle_name:
+            from airflow.dag_processing.bundles.manager import DagBundlesManager
+
+            bundle_manager = DagBundlesManager()
+            bundle_config = bundle_manager._bundle_config[bundle_name]
+
+            dag_team_name = bundle_config.team_name
+
+    if not dag_team_name:
+        return
```
o-nikolas, line 182: Can any of this logic be extracted from the similar steps done in _validate_executor_fields()?

airflow-core/tests/unit/cli/commands/test_team_command.py
```diff
@@ -79,6 +79,25 @@ def test_team_create_success(self, stdout_capture):
         assert "Team 'test-team' created successfully" in output
         assert str(team.name) in output
 
+    def test_team_create_creates_default_pool(self, stdout_capture):
+        """Test that creating a team also creates its default pool."""
+        with conf_vars(
+            {
+                ("core", "multi_team"): "True",
+                ("multi_team", "default_pool_task_slot_count"): "128",
```
o-nikolas:
```suggestion
                ("core", "default_pool_task_slot_count"): "111",
```

This is a `core` config, and the default is already 128, setting it to something else ensures that this is working in your test.

o-nikolas (commented):
Thanks for the contribution! The code needs a fair amount of work but I like the idea!

airflow-core/docs/core-concepts/multi-team.rst
SameerMesiah97: Should it not be DAG instead of Dag as it is an acronym?

Edit: Nvm, it appears that 'Dag' is the convention in the docs.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: Yes. This was added because without it a foreign key violation was occuring as `Pool.create_or_update_pool()` creates a pool referencing `team_name`. Since the Team was just added in the same transaction, it needs to be flushed before that foreign key can be resolved.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: And additional check has been added to the where clause to ensure that the select pool is associated with the team.

airflow-core/docs/core-concepts/multi-team.rst
SameerMesiah97: I am not sure if 'which' is more gramatically correct than 'that' here as the part after is necessary for the sentence to communicate the complete idea. Regardless, I have rephrased this part for clarity.

airflow-core/docs/core-concepts/multi-team.rst
SameerMesiah97: Done.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: The function now returns early if multi-team is not enabled.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: A new helper `_create_default_team_pool` has been created which has the duplicated logic. It is now being called at both relevant callsites.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: This has been moved down to the try/except block where team deletion is attempted.

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: We are now querying for an existing default pool and creating it only if no pool is found.

airflow-core/src/airflow/dag_processing/dagbag.py
SameerMesiah97: I considered extracting a helper, but it would end up being 10 lines at most and it would only be used in 2 places. Also, I felt keeping the logic inline made each function a little easier to follow. That being said, I'm happy to extract it if you think we'll be reusing it elsewhere but I dont see the need for it right now.

airflow-core/tests/unit/cli/commands/test_team_command.py
SameerMesiah97: Done.

airflow-core/newsfragments/69768.feature.rst
```diff
@@ -0,0 +1 @@
+Add automatic creation of team default pools in multi-team deployments and assign tasks without an explicitly configured pool to their team's default pool during DAG parsing. Existing multi-team deployments should run ``airflow teams sync`` after upgrading to provision default pools for existing teams.
```
o-nikolas:
```suggestion
Add automatic creation of team default pools in multi-team deployments and assign tasks without an explicitly configured pool to their team's default pool during Dag parsing. Existing multi-team deployments should run ``airflow teams sync`` after upgrading to provision default pools for existing teams.
```

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -163,6 +201,9 @@ def team_list(args, *, session=NEW_SESSION):
 @provide_session
 def team_sync(args, *, session=NEW_SESSION):
     """Sync missing teams from the dag bundle config."""
+    if not conf.getboolean("core", "multi_team"):
+        return
```
stephen-bracken, line 206: suggest raising a warning here to indicate that `multi_team` is not enabled

airflow-core/src/airflow/models/pool.py
```diff
@@ -121,6 +121,10 @@ def get_default_pool(*, session: Session = NEW_SESSION) -> Pool | None:
         """
         return Pool.get_pool(Pool.DEFAULT_POOL_NAME, session=session)
 
+    @staticmethod
+    def get_default_team_pool_name(team_name: str) -> str:
+        return f"default_pool_{team_name}"
```
stephen-bracken, line 126: We are assuming that a team name will always be a valid suffix to a pool name, should we enforce this in `Team.name`?

airflow-core/src/airflow/cli/commands/team_command.py
o-nikolas: +1 to this one

o-nikolas (commented):
Just one comment and had some similar feedback to @stephen-bracken, otherwise looks good!

airflow-core/src/airflow/models/pool.py
SameerMesiah97: `team_name` is already validated in `_extract_team_name()`, so by the time it reaches `get_default_team_pool_name()` it is guaranteed to match regx `^[a-zA-Z0-9_-]{3,50}$`. So I do not see the need for additional validation for `Team.name`

airflow-core/src/airflow/cli/commands/team_command.py
SameerMesiah97: Done.

o-nikolas (approved):

PR #69769, opened 2026-07-11
Fix DAG details active runs count to exclude queued runs

### LINKED ISSUE #69765: "Max active runs" in DAG UI overcounts by including queued runs

Providers

### Apache Airflow version

3.2.2

The "Max active runs" counter in the DAG detail page header (displayed as "X of Y") counts both `Running` and `Queued` Dag runs. However, `max_active_runs` is only enforced by the scheduler against `Running` runs. This makes the UI misleading — after a backfill, the counter can show a number far exceeding the configured limit (e.g. 12 of 5), even though the scheduler is behaving correctly.

Image

The counter should show only `Running` Dag runs matching exactly what the scheduler enforces. A user looking at 3 of 5 should be able to interpret it as "3 slots taken, 2 available" — which is only true if queued runs are excluded.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

Vishwaspatel2401:
The "Max active runs" counter on the DAG details page ("X of Y") counts both `running` and `queued` DAG runs, but the scheduler only counts `running` runs when enforcing `max_active_runs`. After a backfill queues many runs, the UI counter can show a number far exceeding the configured limit even though the scheduler is behaving correctly.

This fixes the counter to only count `running` runs, matching what the scheduler actually enforces.

closes: #69765

---

- [X] Yes — Partially

### DIFF
--- a/airflow-core/newsfragments/69769.bugfix.rst
+++ b/airflow-core/newsfragments/69769.bugfix.rst
@@ -1,0 +1,1 @@
+Fix DAG details 'active runs' count to only include running runs, not queued ones, matching scheduler enforcement of max_active_runs
--- a/airflow-core/newsfragments/69769.bugfix.rst
+++ b/airflow-core/newsfragments/69769.bugfix.rst
@@ -1,1 +1,0 @@
-Fix DAG details 'active runs' count to only include running runs, not queued ones, matching scheduler enforcement of max_active_runs
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dags.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dags.py
@@ -250,7 +250,7 @@ def get_dag_details(
         session.scalar(
             select(func.count())
             .select_from(DagRun)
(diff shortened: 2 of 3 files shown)


airflow-core/newsfragments/69769.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix DAG details 'active runs' count to only include running runs, not queued ones, matching scheduler enforcement of max_active_runs
```
pierrejeambrun: Remove this file.

pierrejeambrun (commented):
Nice, thanks for the PR. 

Just one nit before we can merge.

pierrejeambrun (approved):

vatsrahul1001:
@Vishwaspatel2401 Thanks for PR looking good, just one thing berfore merge I agree with @pierrejeambrun We should remove news fragment https://github.com/apache/airflow/pull/69769#discussion_r3596074611

vatsrahul1001 (approved):

pierrejeambrun (commented):
I've just pushed a  commit to remove the newsfragment, should be good to merge, waiting on the CI

PR #65685, opened 2026-04-22
Honor AUTH_ROLE_PUBLIC in FastAPI API server

gaurav0107:
## Summary

Closes #60897.

The FastAPI API server's auth gate rejected anonymous users with 401 even when `AUTH_ROLE_PUBLIC` was set in `webserver_config.py`, because the gate only consulted Flask's config. The Flask legacy webserver honored it; FastAPI did not.

## Approach

Per @jason810496's proposal (endorsed by @vincbeck in the issue thread):

1. New `[fab] auth_role_public` Airflow config — the canonical source going forward.
2. New `BaseAuthManager.get_public_user()` hook (returns `None` by default).
3. `FabAuthManager.get_public_user()` resolves the configured role, builds an `AnonymousUser` with `_roles`/`_perms` pre-populated inside `app_context()`, and returns it.
4. `resolve_user_from_token()` in the FastAPI security layer consults `auth_manager.get_public_user()` before raising 401.
5. Startup bridge in `providers/fab/www/app.py` copies `[fab] auth_role_public` into Flask's `AUTH_ROLE_PUBLIC` so the legacy webserver's existing code paths stay in sync — no behavior change for users on `webserver_config.py` only.

### Why the `AnonymousUser._roles` direct-write?

FAB's `AnonymousUser.roles` is a lazy property that calls `security_manager.get_public_role()` on every access, which requires a Flask request context. `get_public_user()` runs under FastAPI where there is no request context, so we build the user inside a brief `app_context()`, resolve permissions eagerly, and write them directly to `_roles`/`_perms` to freeze the snapshot. This is the minimum surface needed to avoid touching FAB's property implementation while keeping the call safe outside Flask.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- 28 new tests added in \`airflow-core/tests/unit/api_fastapi/core_api/test_security.py\`
- 84 new tests added in \`providers/fab/tests/unit/fab/auth_manager/test_fab_auth_manager.py\`
- Covers: happy path, unknown role name, precedence between \`[fab] auth_role_public\` and legacy \`AUTH_ROLE_PUBLIC\`, interaction with the existing bearer-token path, edge cases for FastAPI-without-Flask-request
- Full suite: 476 targeted auth tests pass. \`ruff check\` and \`ruff format\` pass.

## Backwards compatibility

[...]

Issue #60897, opened 2026-01-21
AUTH_ROLE_PUBLIC = 'Admin' # doesn't remove the login dialog


jabbera:
### Apache Airflow Provider(s)

fab

### Versions of Apache Airflow Providers

apache-airflow-providers-fab==3.1.2

### Apache Airflow version

3.1.6

### Operating System

Ubuntu 24.04.3 LTS

### Deployment

Virtualenv installation

### What happened

Setting AUTH_ROLE_PUBLIC = 'Admin' in ${AIRFLOW_HOME}/webserver_config.py doesn't remove the login dialog

The following 401's show up in the logs:

```
INFO:     127.0.0.1:36580 - "GET /ui/config HTTP/1.1" 401 Unauthorized
INFO:     127.0.0.1:36580 - "GET /api/v2/auth/login?next=http%3A%2F%2Flocalhost%3A8080%2F HTTP/1.1" 307 Temporary Redirect
INFO:     127.0.0.1:36590 - "GET /ui/config HTTP/1.1" 401 Unauthorized
INFO:     127.0.0.1:36584 - "GET /ui/auth/menus HTTP/1.1" 401 Unauthorized
INFO:     127.0.0.1:36584 - "GET /api/v2/plugins HTTP/1.1" 401 Unauthorized
```

### What you think should happen instead

The login dialog should not appear.

### How to reproduce

Script to reproduce:

```
uv venv airflow-venv --python=3.11
source ./airflow-venv/bin/activate
uv pip install "apache-airflow" "apache-airflow-providers-fab"

export AIRFLOW_HOME=/tmp/afhome
mkdir -p ${AIRFLOW_HOME}
echo "AUTH_ROLE_PUBLIC = 'Admin'"> ${AIRFLOW_HOME}/webserver_config.py

export AIRFLOW__CORE__AUTH_MANAGER="airflow.providers.fab.auth_manager.fab_auth_manager.FabAuthManager"

airflow api-server
```

browse to http://localhost:8080 and get the login dialog

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

viiccwen:
Take the issue. ：）

jabbera:
[...]

PR #69773, opened 2026-07-12
Fix AUTH_ROLE_PUBLIC returning 401 in FastAPI API server

lyric2249:
`FabAuthRolePublicMiddleware` (added in #65685) sets `request.state.user` for anonymous requests when `AUTH_ROLE_PUBLIC` / `[fab] auth_role_public` is configured, but never sets `request.state.user_authenticated_via`. Core's `get_user()` dependency requires both attributes before it will trust a pre-set user and skip JWT validation, so anonymous requests still fall through to JWT validation and get rejected with 401 even though the public role is configured correctly.

This sets the same trust marker `JWTRefreshMiddleware` already sets, so both middlewares satisfy the same contract that `get_user()` checks.

The existing `test_sets_public_user_when_no_auth_present` test already exercises this exact code path, so it is extended with one assertion for the new attribute rather than adding a separate test; no existing assertions were changed or removed.

related: #60897

---


providers/fab/src/airflow/providers/fab/auth_manager/middleware.py
```diff
@@ -47,6 +54,12 @@ async def dispatch(self, request: Request, call_next):
             public_user = auth_manager.build_public_user()
             if public_user is not None:
                 request.state.user = public_user
+
+                if _USER_INJECTED_BY_TRUSTED_MIDDLEWARE is not None:
```
SameerMesiah97: I don't see the need for the `_USER_INJECTED_BY_TRUSTED_MIDDLEWARE ` if it is only being used once. It would be better to in-line it like this:

```
user_injected = getattr(
    core_api_security,
    "USER_INJECTED_BY_TRUSTED_MIDDLEWARE",
    None,
)
if user_injected is not None:
    request.state.user_authenticated_via = user_injected
```

SameerMesiah97 (commented):
I have left a comment.

providers/fab/src/airflow/providers/fab/auth_manager/middleware.py
lyric2249: Thanks for comment. I applied it

vincbeck (approved):

providers/fab/src/airflow/providers/fab/auth_manager/middleware.py
```diff
@@ -23,6 +23,9 @@
 
 from airflow.api_fastapi.app import get_auth_manager
 from airflow.api_fastapi.auth.managers.base_auth_manager import COOKIE_NAME_JWT_TOKEN
+from airflow.api_fastapi.core_api import security as core_api_security
+
+
 
 if TYPE_CHECKING:
```
eladkal, line 28:
```suggestion
from airflow.api_fastapi.core_api import security as core_api_security

if TYPE_CHECKING:
```

To fix static checks

eladkal (approved):

Issue #69001, opened 2026-06-25
Scheduler `DBDagBag` cache is never evicted and grows unbounded


aeroyorch:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.2.2

### What happened and how to reproduce it?

The scheduler memory grows until the process is OOMKilled and restarted. I attach a
Grafana graph with the growth over time.

<img width="1600" height="499" alt="Image" src="https://github.com/user-attachments/assets/bd398687-bfdd-4df1-ba24-b23ef67fed06" />

We profiled the scheduler with `memray`, following the [Airflow memory profiling guide](https://airflow.apache.org/docs/apache-airflow/stable/howto/memory-profiling.html).
The part that keeps growing is the `DBDagBag` cache of deserialized DAGs, which in the
scheduler is a plain dict with no eviction. Two memray snapshots from one run, with the memory retained under
`get_dag_for_run` / `_get_dag`:

| Snapshot | Peak heap | `_get_dag` (DBDagBag cache) |
|----------|-----------|-----------------------------|
| snapshot 1 | 1.38 GB | 464.9 MB |
| snapshot 2 | 2.18 GB | 669.5 MB |

The snapshots also show another large part, the task instances loaded during each scheduling
pass. That one seems normal.

The cause seems to be in `SchedulerJobRunner.__init__`:
`self.scheduler_dag_bag = DBDagBag(load_op_links=False)`. With no `cache_size`, `DBDagBag._dags`
has no size or TTL limit, so it keeps one deserialized DAG per `dag_version_id` and does not
drop old versions. The bag lives for the whole scheduler process, which by default runs
indefinitely (`[scheduler] num_runs = -1`). As the number of DAG versions grows over time, this
becomes unbounded.

To reproduce: DAGs that keep producing new versions over time, a standalone dag-processor, and
a scheduler that runs for a long time. Watch `len(scheduler_dag_bag._dags)` and the process
resident memory (RSS). Both grow continuously and only go down after a restart.

### What you think should happen instead?

The scheduler `DBDagBag` should use a bounded LRU+TTL cache, like the API server since #60804
(that PR limited the change to the API server only). Two settings, `[scheduler] dag_cache_size`
and `dag_cache_ttl`, like the `[api]` ones, would keep it bounded, with a size for the active
set of DAG versions and a TTL to drop the old ones. A value of 0 keeps today's unbounded
behavior.

### Deployment

Official Apache Airflow Helm Chart

[...]

PR #69007, opened 2026-06-26
Use bounded DBDagBag cache in scheduler

Mady356:
Use bounded `DBDagBag` caching in the scheduler.

The scheduler was creating `DBDagBag(load_op_links=False)` without cache settings, which meant it used the default unbounded dictionary for cached deserialized DAGs. This PR makes the scheduler use the existing `DBDagBag` LRU/TTL cache support instead.

Changes:
- Added `[scheduler] dag_cache_size` and `[scheduler] dag_cache_ttl` config options.
- Updated `SchedulerJobRunner` to pass those config values into `DBDagBag`.
- Added a `stats_prefix` parameter to `DBDagBag` so scheduler cache metrics are emitted under `scheduler.dag_bag` instead of the API server prefix.
- Added scheduler DagBag cache metrics to the metrics template.
- Handles negative config values safely by falling back to unbounded cache / disabled TTL behavior.
- Updated the `DBDagBag` class docstring to reflect that callers can enable bounded caching.
- Added a unit test confirming the scheduler passes the cache config into `DBDagBag`.

closes: #69001

Testing:

```bash
python -m py_compile airflow-core/src/airflow/models/dagbag.py
python -m py_compile airflow-core/src/airflow/jobs/scheduler_job_runner.py
pytest airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_scheduler_dag_bag_uses_scheduler_cache_config
pytest airflow-core/tests/unit/models/test_dagbag.py

<!--
Thank you for contributing!

Please provide above a brief description of the changes made in this pull request.
Write a good git commit message following this guide: http://chris.beams.io/posts/git-commit/

Please make sure that your code changes are covered with tests.
And in case of new features or big changes remember to adjust the documentation.

For user-facing UI changes, please attach before/after screenshots (or a short
screen recording) so reviewers can assess the visual impact.

Feel free to ping (in general) for the review if you do not see reaction for a few days
(72 Hours is the minimum reaction time you can expect from volunteers) - we sometimes miss notifications.

In case of an existing issue, reference it using one of the following:

* closes: #ISSUE
* related: #ISSUE
-->

---

##### Was generative AI tooling used to co-author this PR?

<!--
If generative AI tooling has been used in the process of authoring this PR, please
change below checkbox to `[X]` followed by the name of the tool, uncomment the "Generated-by".
-->

- [ ] Yes (please specify the tool below)

<!--
[...]

PR #69774, opened 2026-07-12
Allow `DBDagBag` TTL cache eviction without a size cap

aeroyorch:
---

Allow `DBDagBag` TTL cache eviction without a size cap (i.e. `cache_size=0`, `cache_ttl>0`).

Related to #69001 and #69007 

Claude Opus 4.8 to analyze the impact of this change in terms of docs/code.

---

### DIFF
--- a/airflow-core/docs/faq.rst
+++ b/airflow-core/docs/faq.rst
@@ -725,8 +725,10 @@ in memory. Configure this in the ``[api]`` section:
 .. code-block:: ini
 
     [api]
-    dag_cache_size = 64    ; max cached versions (0 = unbounded, pre-3.2 behavior)
+    dag_cache_size = 64    ; max cached versions (0 = no size cap, TTL still applies)
     dag_cache_ttl = 3600   ; seconds before a cached entry expires (0 = LRU only)
+
+Setting both to 0 disables eviction entirely, matching pre-3.2 behavior.
 
 The cache is keyed by Dag version ID. After a Dag is updated, the API server may serve the
 previous version until the cached entry expires (controlled by ``dag_cache_ttl``).
--- a/airflow-core/docs/faq.rst
+++ b/airflow-core/docs/faq.rst
@@ -728,7 +728,7 @@ in memory. Configure this in the ``[api]`` section:
     dag_cache_size = 64    ; max cached versions (0 = no size cap, TTL still applies)
     dag_cache_ttl = 3600   ; seconds before a cached entry expires (0 = LRU only)
 
-Setting both to 0 disables eviction entirely, matching pre-3.2 behavior.
+Setting both to 0 disables eviction entirely, matching pre-3.3 behavior.
 
 The cache is keyed by Dag version ID. After a Dag is updated, the API server may serve the
 previous version until the cached entry expires (controlled by ``dag_cache_ttl``).
--- a/airflow-core/src/airflow/api_fastapi/common/dagbag.py
+++ b/airflow-core/src/airflow/api_fastapi/common/dagbag.py
@@ -39,15 +39,11 @@ def create_dag_bag() -> DBDagBag:
     cache_ttl_config = conf.getint("api", "dag_cache_ttl", fallback=3600)
 
     if cache_size < 0:
-        log.warning("dag_cache_size must be >= 0, using unbounded dict")
+        log.warning("dag_cache_size must be >= 0, using no size cap")
         cache_size = 0
     if cache_ttl_config < 0:
         log.warning("dag_cache_ttl must be >= 0, disabling TTL")
         cache_ttl_config = 0
-
-    # Use unbounded dict (no eviction) if cache_size is 0
-    if cache_size <= 0:
-        return DBDagBag(cache_size=0)
 
     # Disable TTL if cache_ttl is 0
     cache_ttl: int | None = cache_ttl_config if cache_ttl_config > 0 else None
--- a/airflow-core/src/airflow/config_templates/config.yml
+++ b/airflow-core/src/airflow/config_templates/config.yml
@@ -1705,7 +1705,10 @@ api:
       description: |
         Size of the LRU cache for SerializedDAG objects in the API server.
         Set to 0 to remove the size cap. Cached entries are then evicted only by
-        ``dag_cache_ttl``, or never if that is also 0 (matching pre-3.2 behavior).
+        ``dag_cache_ttl``, or never if that is also 0 (matching pre-3.3 behavior).
+        Note that the TTL is then not a ceiling on memory. Entries that keep being
+        requested are refreshed and never expire, so memory usage follows the number
+        of distinct Dag versions in active use.
         The cache is keyed by Dag version ID, so lookups by Dag ID
         (e.g., viewing a Dag's details) always query the database for the latest
         version, but the deserialized result is cached for subsequent
--- a/airflow-core/src/airflow/config_templates/config.yml
+++ b/airflow-core/src/airflow/config_templates/config.yml
@@ -1704,7 +1704,8 @@ api:
     dag_cache_size:
       description: |
         Size of the LRU cache for SerializedDAG objects in the API server.
-        Set to 0 to use an unbounded dict (no eviction, matching pre-3.2 behavior).
+        Set to 0 to remove the size cap. Cached entries are then evicted only by
+        ``dag_cache_ttl``, or never if that is also 0 (matching pre-3.2 behavior).
         The cache is keyed by Dag version ID, so lookups by Dag ID
         (e.g., viewing a Dag's details) always query the database for the latest
         version, but the deserialized result is cached for subsequent
@@ -1717,7 +1718,9 @@ api:
       description: |
         Time-to-live (seconds) for cached SerializedDAG objects in the API server.
         After this time, cached DAGs will be re-fetched from the database on next access.
-        Set to 0 to disable TTL (cache entries will only be evicted by LRU policy).
+        Applies whether or not ``dag_cache_size`` sets a size cap. Set to 0 to disable TTL,
+        leaving eviction to the ``dag_cache_size`` LRU policy, or no eviction at all if
+        ``dag_cache_size`` is also 0.
 
         Note: After a DAG is updated, the API server may serve the previous version
         until the cached entry expires. Lower values reduce staleness but increase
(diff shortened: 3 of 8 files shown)


airflow-core/src/airflow/models/dagbag.py
```diff
@@ -89,11 +91,12 @@ def __init__(
         self._revalidation_interval = conf.getint("core", "min_serialized_dag_update_interval")
 
         # Initialize bounded cache if cache_size is provided and > 0
```
SameerMesiah97: I think this comment is no longer accurate. I would suggest the below:

`# Initialize a TTL cache if configured, otherwise a bounded LRU cache.`

SameerMesiah97 (commented):
Just one nit. Looks good otherwise.

aeroyorch:
> Just one nit. Looks good otherwise.

Thanks for the review! Changes already implemented

SameerMesiah97 (approved):

lohyenshen:
hi, may I know what airflow version would this be released to?

vatsrahul1001 (commented):
The DBDagBag fix itself looks right, but I don't think it's reachable from the actual config path yet. create_dag_bag() in api_fastapi/common/dagbag.py still does:

    if cache_size <= 0:
        return DBDagBag(cache_size=0)

That early-return fires whenever `dag_cache_size <= 0` and never even reads `cache_ttl_config`, so setting `dag_cache_size = 0` + `dag_cache_ttl = 3600` in airflow.cfg still ends up as a plain unbounded dict with no eviction. test_create_dag_bag_cache_modes still asserts exactly that (the "size_zero_unbounded" case expects `dict`/`_use_cache=False` for cache_size=0, cache_ttl=3600), and config.yml's dag_cache_size description still says 0 means "unbounded dict, no eviction" with no mention of TTL-only mode.

Could you also update create_dag_bag() to pass cache_ttl through when cache_size<=0, add a case to test_create_dag_bag_cache_modes covering that combo, and touch up the config.yml wording? Otherwise this fix isn't actually reachable by anyone configuring it through airflow.cfg.

aeroyorch:
> The DBDagBag fix itself looks right, but I don't think it's reachable from the actual config path yet. create_dag_bag() in api_fastapi/common/dagbag.py still does:
> 
> ```
> if cache_size <= 0:
>     return DBDagBag(cache_size=0)
> ```
> 
> That early-return fires whenever `dag_cache_size <= 0` and never even reads `cache_ttl_config`, so setting `dag_cache_size = 0` + `dag_cache_ttl = 3600` in airflow.cfg still ends up as a plain unbounded dict with no eviction. test_create_dag_bag_cache_modes still asserts exactly that (the "size_zero_unbounded" case expects `dict`/`_use_cache=False` for cache_size=0, cache_ttl=3600), and config.yml's dag_cache_size description still says 0 means "unbounded dict, no eviction" with no mention of TTL-only mode.
> 
> Could you also update create_dag_bag() to pass cache_ttl through when cache_size<=0, add a case to test_create_dag_bag_cache_modes covering that combo, and touch up the config.yml wording? Otherwise this fix isn't actually reachable by anyone configuring it through airflow.cfg.


Thanks for the review. Changes already implemented :)

vatsrahul1001 (approved):

vatsrahul1001:
LGTM!, can be merged after code owners review

vatsrahul1001:

aeroyorch:
> Moving to 3.3.2 as this is still pending code owner review and do not want to rush on merging this as it not critical

No problem. In the meantime, I can work on some of the other items to fix https://github.com/apache/airflow/issues/69001

airflow-core/src/airflow/api_fastapi/common/dagbag.py
```diff
@@ -39,16 +39,12 @@ def create_dag_bag() -> DBDagBag:
     cache_ttl_config = conf.getint("api", "dag_cache_ttl", fallback=3600)
 
     if cache_size < 0:
-        log.warning("dag_cache_size must be >= 0, using unbounded dict")
+        log.warning("dag_cache_size must be >= 0, using no size cap")
```
kaxil, line 42: With the `cache_size <= 0` early return gone, an existing `dag_cache_size = 0` deployment changes behavior on upgrade. On 3.3.0/3.3.1 that gave a plain dict with `_use_cache=False` and a `nullcontext`; with the default `dag_cache_ttl = 3600` left alone it now builds `TTLCache(maxsize=inf)` with `_use_cache=True`, so entries expire hourly and every lookup takes the `RLock`. I ran `create_dag_bag()` on both branches under a mocked conf to confirm.

Since this is milestoned for 3.3.2, could you add a newsfragment? Anyone who set 0 for the documented no-eviction behavior now needs `dag_cache_ttl = 0` as well, and that won't be obvious from a patch release.

airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -1704,7 +1704,8 @@ api:
     dag_cache_size:
       description: |
         Size of the LRU cache for SerializedDAG objects in the API server.
-        Set to 0 to use an unbounded dict (no eviction, matching pre-3.2 behavior).
+        Set to 0 to remove the size cap. Cached entries are then evicted only by
```
kaxil, line 1707: Could this note that the TTL doesn't put a ceiling on memory in this mode? Once the revalidation window elapses, `_get_dag` writes the entry back (`self._dags[version_id] = current._replace(last_validated=now)`), and `TTLCache.__setitem__` resets that entry's expiry. `min_serialized_dag_update_interval` defaults to 30s and `dag_cache_ttl` to 3600s, so a version that keeps getting requested is refreshed roughly 120x per TTL and never ages out.

I simulated both clocks against a real `TTLCache(maxsize=inf, ttl=3600)`: with a write-back every 30s the entry was still resident after 11h, while reads alone evicted it at exactly 3600s.

The "which is fine" note on that write-back in `models/dagbag.py` held because a size cap bounded things regardless. Without a cap, the resident set is every distinct version touched inside a TTL window, unbounded. Worth spelling out, since someone reaching for this is likely trying to fix memory growth.

airflow-core/docs/faq.rst
```diff
@@ -725,9 +725,11 @@ in memory. Configure this in the ``[api]`` section:
 .. code-block:: ini
 
     [api]
-    dag_cache_size = 64    ; max cached versions (0 = unbounded, pre-3.2 behavior)
+    dag_cache_size = 64    ; max cached versions (0 = no size cap, TTL still applies)
     dag_cache_ttl = 3600   ; seconds before a cached entry expires (0 = LRU only)
 
+Setting both to 0 disables eviction entirely, matching pre-3.2 behavior.
```
kaxil: Should be pre-3.3 rather than pre-3.2. The heading above says the feature is available since Airflow 3.3.0, both options carry `version_added: 3.3.0`, and `git tag --contains` on the commit that added them (#60804) returns only 3.3.0. On 3.2.0 `DBDagBag._dags` was already a plain unbounded dict, so unbounded is 3.2 behavior too, not something that predates it. Same string on config.yml line 1708.

ColtenOuO (commented):
LGTM and Thanks for the improvement! I actually just noticed this issue myself, and was glad to see someone was already working on a fix!

Just a small idea, feel free to take it or leave it if it doesn't quite fit! 

Would it make sense to move the warning logs into `DBDagBag.__init__` itself? To fully close out #69001, this change will eventually need to be wired up to the **Scheduler** too , and at that point we'd want to warn users about the exact same negative-value cases again. If the warning lived inside `DBDagBag.__init__` from the start, it'd be a lot easier to keep things consistent as more callers get hooked up down the road , one place to maintain, and no risk of some future caller quietly forgetting to warn.

The trade-off is we'd lose the ability to tell the user exactly which config option was the problem (e.g. `[api] dag_cache_size` vs. a future `[scheduler] dag_bag_cache_size`)

aeroyorch:
> LGTM and Thanks for the improvement! I actually just noticed this issue myself, and was glad to see someone was already working on a fix!
> 
> Just a small idea, feel free to take it or leave it if it doesn't quite fit!
> 
> Would it make sense to move the warning logs into `DBDagBag.__init__` itself? To fully close out #69001, this change will eventually need to be wired up to the **Scheduler** too , and at that point we'd want to warn users about the exact same negative-value cases again. If the warning lived inside `DBDagBag.__init__` from the start, it'd be a lot easier to keep things consistent as more callers get hooked up down the road , one place to maintain, and no risk of some future caller quietly forgetting to warn.
> 
> The trade-off is we'd lose the ability to tell the user exactly which config option was the problem (e.g. `[api] dag_cache_size` vs. a future `[scheduler] dag_bag_cache_size`)

Thanks for the review!

Fair point, but I'd rather leave it as is for now, the trade-off you mention is the deciding one for me: keeping it at the call site lets us name the actual config option that's wrong, which is what the user needs to fix.

Happy to revisit when the `Scheduler` part lands for #69001.

ColtenOuO (approved):

pierrejeambrun (dismissed):
Do you mind giving more context as to why we are doing this. If your use case is "LRU evicts my hot dags too aggressively", tweaking the settings for your use case is probably better.

It looks like we are introducing back the memory leak this bounded cache was fixing in the first place. I don't think we should allow unbounded cache there.

aeroyorch:
> Do you mind giving more context as to why we are doing this. If you use case is "LRU evicts my hot dags too aggressively", tweaking the settings for your use case is probably better.
> 
> It looks like we are introducing back the memory leak this bounded cache was fixing in the first place. I don't think we should allow unbouded cache there.

Hi Pierre, sure.

I don't think this brings back the leak. Today `cache_size=0` is a plain `dict` with no eviction at all. With this change, `cache_size=0` + `cache_ttl>0` creates a real `TTLCache`. That evicts exactly what grows and keeps the hot ones. Defaults don't change.

Tuning the size doesn't really help here: the scheduler cycles through all active `dag_version_id`s, so any cap below the active set evicts each key just before it's needed again. @kaxil raised this in the #69007 review and @potiuk asked for TTL eviction without a size cap in `DBDagBag` as a first step.

pierrejeambrun:
Thanks for the context @aeroyorch this makes more sense. I had the wrong assumption that unbounded cache was just not possible anymore. (at all)

I would be for not allowing unbounded cache completely since we know it's a bad practice, but for backward comp I understand why unbounded is still allowed.

uranusjr:
This is superceded by https://github.com/apache/airflow/pull/71704

PR #69776, opened 2026-07-12
Support image digest in worker pod template image

### LINKED ISSUE #58625: Worker pod template image does not support image digest

1.18.0 (latest released)

### Apache Airflow version

3.1.3

### Kubernetes Version

v1.33.5-eks

### Helm Chart configuration

```
...
airflow:
  digest: sha256:1ba822595cc6d29a9afaf213e5015ee56c0d77c2795d8e095df65a33dee318cf
  pullPolicy: IfNotPresent
  repository: <account>.dkr.ecr.<region>.amazonaws.com/airflow
  tag: latest
pod_template:
  repository: ~
  tag: ~
  pullPolicy: Always
```

_No response_

### What happened

When updating the image and maintaining the same tag, for example `latest`, the airflow containers will be updated correctly because the use the digest. However the workers will be stuck using the stale version of `latest` that is present in the node.

The pod template does not support using the image digest. See: 
https://github.com/apache/airflow/blob/helm-chart/1.18.0/chart/templates/_helpers.yaml#L361

However the airflow images do, see: https://github.com/apache/airflow/blob/helm-chart/1.18.0/chart/templates/_helpers.yaml#L350

The definition of `pod_template_image` should support the digest. With this we can ensure the workers will use the correct image. 

### How to reproduce

Create an image and tag with `latest`
Use the latest tag for all airflow images `images.airflow.tag` and leave the `images.pod_template.*` values unset. 
Run a task
Create a new image and tag it with `latest` 
Update the `images.airflow.digest` to the digest of the new image. Airflow images run with the correct image - digest takes precedence over tag.
Run a task again.
The worker is created with the stale image with tag `latest` the digest in the airflow values section is ignored.

### Anything else

_No response_

### Code of Conduct

PreethamSanji:
Closes: #58625
 
## Issue
The `airflow_image` helper supports `repository@digest` (digest takes precedence over tag), but `pod_template_image` only rendered `repository:tag`. As a result, worker pods created from the pod template ignored digests and could run stale images when a mutable tag like `latest` was reused.
 
## Fix
This PR makes `pod_template_image` mirror `airflow_image`: it introduces `images.pod_template.digest` (added to `values.yaml` and `values.schema.json` — the schema has `additionalProperties: false` for this block), falling back to `defaultAirflowDigest`, consistent with the existing repository/tag fallback chains.
 
**Before** 
(`helm template chart --set executor=KubernetesExecutor --set defaultAirflowRepository=my-repo --set defaultAirflowDigest=sha256:abc123`):
`helm template chart \
  --set executor=KubernetesExecutor \
  --set defaultAirflowRepository=my-repo \
  --set defaultAirflowDigest=sha256:abc123 2>/dev/null \
  | grep "image: my-repo" | sort -u`

          `image: my-repo:3.3.0`
          `image: my-repo@sha256:abc123`

**After** (same values):
`helm template chart \
  --set executor=KubernetesExecutor \
  --set defaultAirflowRepository=my-repo \
  --set defaultAirflowDigest=sha256:abc123 2>/dev/null \
  | grep "image: my-repo" | sort -u`

          `image: my-repo@sha256:abc123`

**Runtime caveat, documented in the new value's description:**
the chart defaults `config.kubernetes_executor.worker_container_repository`/`worker_container_tag` to `images.airflow.repository:tag`, and when both are set the KubernetesExecutor overrides the pod template's base container image with that `repo:tag` string (`kube_config.py` builds `kube_image`; `PodGenerator.construct_pod` reconciles it over the base pod). 

Since `repo:tag` cannot express a digest, the digest only takes effect when those two config options are unset — the same pre-existing limitation already documented for `images.pod_template.repository`/`tag`. 

Making those config defaults digest-aware (defaulting them to empty when a digest is configured, so the template image wins) would fix the reporter's scenario with zero config changes, but it changes default behavior, happy to do that here or as a follow-up if preferred.
 
Tests: parametrized rendering tests in `test_pod_template_file.py` covering tag-only, digest-only, digest precedence over tag, and the `defaultAirflowDigest` fallback.
 
`flower_image` has the same gap; I'd address it separately if wanted.

 

---

Claude Fable 5

Generated-by: [Claude Fable 5] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


chart/newsfragments/69776.bugfix.rst
```diff
@@ -0,0 +1 @@
+Support image digest in the worker pod template image (``images.pod_template.digest``)
```
Miretpl: I think that a normal log entry within the release notes will be enough for that change. Could you please remove this file?

chart/values.yaml
```diff
@@ -89,6 +89,8 @@ images:
     # `config.kubernetes_executor.worker_container_tag` must be not set .
     repository: ~
     tag: ~
+    # Specifying digest takes precedence over tag.
+    digest: ~
```
Miretpl, line 93: Could you adjust the comment above to say that this field will also be overridden by config params if defined?

Miretpl (changes requested):
Nice catch!

Could you also format the description of the PR a bit more? It would make it a bit easier to read.

chart/newsfragments/69776.bugfix.rst
PreethamSanji: Done, removed it.

chart/values.yaml
PreethamSanji: Good point, updated to mention config override.

Miretpl: I had more in mind for the comment, which is right under `pod_template:` (I wasn't precise with saying `above` 😄 - sorry about that). Could you adjust it (normally I would leave it as it is, but even then there would be one change to do here, so despite that the general comment to the whole `_pod_template:` can be adjusted I think)?

Miretpl (changes requested):

chart/values.yaml
PreethamSanji: Ah got it, thanks for clarifying! Updated the general comment under `pod_template:` to also mention `images.pod_template.digest`, and reverted the digest inline comment back to matching the `images.airflow` block.

Miretpl (approved):

Issue #57864, opened 2025-11-05
Secrets with slashed IDs return AirflowNotFoundException


bmanan7:
### Apache Airflow version

3.1.1

### What happened?

In Airflow 3.x the execution API route GET `/execution/connections/{conn_id}` treats the connection ID as a FastAPI path segment. Conn IDs containing / (e.g., dev-env/project-name) are not URL encoded before the request, so Starlette splits them into multiple segments and returns 404 even though the connection exists. 

airflow connections get (CLI) still works because it bypasses the execution API, causing a mismatch between CLI and Task SDK behaviour.

Apart from this one will also get`AirflowNotFoundException` when the secret value is defined as `{"username":"uname@somedomain.com", "password":"****"}` however we don't get the error if we define secret value as show below:
```
{
 "conn_type": "http",
 "login": "uname@somedomain.com",
 "password": "***",
 "host": "https://api.somedomain.com",
 "port": 443,
 "extra": "{\"verify_ssl\": false}"
}
```
it looks like without `conn_type` defined it doesn't work in Airflow3?

### What you think should happen instead?

- `/` in secret name should be accepted. 
- Either `conn_type` should be marked as compulsory and documentation should be updated to reflect this or airflow should accept a secret without `conn_type`

### How to reproduce

- Start an Airflow 3.x environment.

- Inside the scheduler container, create a connection whose ID contains /. Example script:
```
python3 - <<'PY'
from airflow import settings
from airflow.models.connection import Connection

conn = Connection(conn_id="dev-env/project-name", conn_type="http", host="example.com")
session = settings.Session()
session.merge(conn)
session.commit()
PY
```

- Add a DAG with a task that calls BaseHook.get_connection("dev-env/project-name") .

- Trigger the DAG. The task fails and worker logs show a 404 from GET /execution/connections/dev-env/project-name.

- (Optional sanity check) Running airflow connections get `dev-env/project-name` -oyaml still works because it bypasses the execution API.

### Operating System

macOS 15.6.1

### Deployment

Official Apache Airflow Helm Chart

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

diligejy:
[...]

PR #61728, opened 2026-02-10
Make conn_type optional in task SDK Connection datamodel

amoghrajesh:
---

##### Was generative AI tooling used to co-author this PR?

- [ ] Yes (please specify the tool below)

closes: https://github.com/apache/airflow/issues/58529

## Why?

When workers retrieve connections from secrets backends like AWS Secrets Manager that don't include `conn_type`, the `Connection.from_json()` method fails during deserialization with error:

```python
TypeError: Connection.__init__() missing 1 required keyword-only argument: 'conn_type'
```

Simple test to show an example for this:

```shell
(apache-airflow) ➜  airflow git:(conn-type-optional-sdk) ✗ cat > /tmp/test_connections.json << 'EOF'
{
  "my_db": {
    "host": "mydb.us-east-1.rds.amazonaws.com",
    "port": 5432,
    "login": "admin",
    "password": "secret123",
    "schema": "production"
  }
}
EOF
```

Before:
```python
>>> import json
>>> json_from_aws = json.dumps({
...     "host": "mydb.us-east-1.rds.amazonaws.com",
...     "port": 5432,
...     "login": "admin",
...     "password": "secret123",
...     "schema": "production"
... })
>>> 
>>> from airflow.sdk import Connection
>>> conn = Connection.from_json(json_from_aws, conn_id="my_db")
Traceback (most recent call last):
  File "<python-input-10>", line 1, in <module>
    conn = Connection.from_json(json_from_aws, conn_id="my_db")
  File "/Users/amoghdesai/Documents/OSS/repos/airflow/task-sdk/src/airflow/sdk/definitions/connection.py", line 307, in from_json
    return cls(conn_id=conn_id, **kwargs)
TypeError: Connection.__init__() missing 1 required positional argument: 'conn_type'
```

After:

```python
>>> import json
>>> from airflow.sdk import Connection
>>> 
>>> json_from_aws = json.dumps({
...     "host": "mydb.us-east-1.rds.amazonaws.com",
...     "port": 5432,
...     "login": "admin",
...     "password": "secret123",
...     "schema": "production"
... })
>>> conn = Connection.from_json(json_from_aws, conn_id="my_db")
>>> conn
Connection(conn_id='my_db', conn_type=None, description=None, host='mydb.us-east-1.rds.amazonaws.com', schema='production', login='admin', password='secret123', port=5432, extra=None)
```

## What

[...]

PR #69780, opened 2026-07-12
Document that connection secrets without conn_type remain supported

henry3260:
## Why

Connection secrets in JSON format could omit `conn_type`. Although the secrets backend deserialized those values, the Execution API requires a connection type and the task could ultimately receive a misleading “connection not found” error.

This addresses the `conn_type` portion of related: #57864.

## What

- Reject JSON connection secrets whose `conn_type` is missing, null, empty, or whitespace-only.
- Raise a clear `ValueError` before constructing an incomplete Connection.
- Add regression coverage for each invalid `conn_type` value.

---

---

### DIFF
--- a/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
+++ b/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
@@ -122,12 +122,13 @@ class BaseSecretsBackend(ABC):
     def _deserialize_connection_value(conn_class: type, conn_id: str, value: str):
         value = value.strip()
         if value[0] == "{":
-            # JSON secrets stored by Airflow 2-era backends may lack both "conn_type"
+            # JSON secrets stored by Airflow 2 backends may lack both "conn_type"
             # and "uri" (e.g. {"host": ..., "login": ..., "password": ...}). This is
             # valid: conn_type is intentionally optional on the SDK Connection model
-            # for Airflow 2 -> 3 migration compatibility. Do not add validation here
-            # that rejects such secrets; see
-            # https://github.com/apache/airflow/pull/61728
+            # for Airflow 2 -> 3 migration compatibility.
+            # Check: https://github.com/apache/airflow/pull/61728
+            # TODO: Remove this compatibility once the minimum supported Airflow
+            #   version in providers is 3.0.
             return conn_class.from_json(value=value, conn_id=conn_id)  # type: ignore[attr-defined]
 
         # TODO: Only sdk has from_uri defined on it. Is it worthwhile developing the core path or not?
--- a/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
+++ b/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import inspect
+import json
 from abc import ABC
 from collections.abc import Callable
 
@@ -122,6 +123,11 @@ class BaseSecretsBackend(ABC):
     def _deserialize_connection_value(conn_class: type, conn_id: str, value: str):
         value = value.strip()
         if value[0] == "{":
+            connection_data = json.loads(value)
+            conn_type = connection_data.get("conn_type")
+            uri = connection_data.get("uri")
+            if not (isinstance(conn_type, str) and conn_type.strip() or isinstance(uri, str) and uri.strip()):
+                raise ValueError("Connection secret JSON must include a non-empty 'conn_type' or 'uri'.")
             return conn_class.from_json(value=value, conn_id=conn_id)  # type: ignore[attr-defined]
 
         # TODO: Only sdk has from_uri defined on it. Is it worthwhile developing the core path or not?
--- a/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
+++ b/shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
@@ -17,7 +17,6 @@
 from __future__ import annotations
 
 import inspect
-import json
 from abc import ABC
 from collections.abc import Callable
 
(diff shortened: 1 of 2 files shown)


Vamsi-klu:
Went through `_deserialize_connection_value` and the new guard reads correct to me. Good call accepting `uri` as an alternative rather than requiring `conn_type` outright, since `from_json` passes a bare `uri` to the constructor which routes through `from_uri` and parses the scheme into the conn_type, so those blobs still build a usable connection and the `{"conn_type": null, "uri": ...}` test covers exactly that. Raising a `ValueError` up front is a clear improvement over the misleading connection not found error the PR describes. Only nit, you `json.loads(value)` in the guard and then `from_json` parses the same string again, negligible for a small secret blob so I wouldn't bother.

Lee-W (dismissed):

shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
```diff
@@ -122,6 +123,11 @@ def _get_connection_class(self) -> type:
     def _deserialize_connection_value(conn_class: type, conn_id: str, value: str):
         value = value.strip()
         if value[0] == "{":
+            connection_data = json.loads(value)
+            conn_type = connection_data.get("conn_type")
+            uri = connection_data.get("uri")
+            if not (isinstance(conn_type, str) and conn_type.strip() or isinstance(uri, str) and uri.strip()):
+                raise ValueError("Connection secret JSON must include a non-empty 'conn_type' or 'uri'.")
```
amoghrajesh: This is incorrect, check: https://github.com/apache/airflow/pull/61728. Connections without a `conn_type` are important for airflow 2-> 3 compatibility. 

#69780 raises error for the same secret shape #61728's own regression test proves should succeed (`{"host":..., "login":..., "password":...}`, no `conn_type`, no `uri`), so merging it would regress that fix and break migration compatibility for users with secrets stored without `conn_type`.


And relatedly, for a connection served from a secrets backend (AWS Secrets Manager, Vault, custom backend, etc.), resolution normally happens on the worker itself, not via the Execution API:

- worker (task-sdk configuration.py) instantiates the configured secrets backend directly so any deployment with a secrets backend configured gets it resolved locally on the worker.
- `_deserialize_connection_value()`runs there, and #61728 made `conn_type` optional on the task-SDK Connection model specifically so this resolves cleanly with no `conn_type` 
- `ExecutionAPISecretsBackend` only comes into play as a last resort fallback if the worker's own backend chain doesn't resolve the `conn_id`. When that happens, the server resolves via its own chain: env vars, then the metastore DB. The DB path is safe by construction: `conn_type` is a `NOT NULL` column, so a connection reaching ConnectionResponse via that fallback always has a conn_type.

So the "misleading connection not found error" this PR is trying to prevent doesn't actually occur on the path it changes, that path already works correctly per #61728, and tightening it back to reject missing conn_type only breaks that compatibility fix without fixing anything real.

amoghrajesh (changes requested):

shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
Lee-W: Thanks for the clarification. I forgot we already had that PR. 🤦‍♂️

We could also add a comment or some documentation around the relevant function so others don't end up trying the same thing again.

henry3260: > This is incorrect, check: #61728. Connections without a `conn_type` are important for airflow 2-> 3 compatibility.
> 
> #69780 raises error for the same secret shape #61728's own regression test proves should succeed (`{"host":..., "login":..., "password":...}`, no `conn_type`, no `uri`), so merging it would regress that fix and break migration compatibility for users with secrets stored without `conn_type`.
> 
> And relatedly, for a connection served from a secrets backend (AWS Secrets Manager, Vault, custom backend, etc.), resolution normally happens on the worker itself, not via the Execution API:
> 
> * worker (task-sdk configuration.py) instantiates the configured secrets backend directly so any deployment with a secrets backend configured gets it resolved locally on the worker.
> * `_deserialize_connection_value()`runs there, and [Make conn_type optional in task SDK Connection datamodel #61728](https://github.com/apache/airflow/pull/61728) made `conn_type` optional on the task-SDK Connection model specifically so this resolves cleanly with no `conn_type`
> * `ExecutionAPISecretsBackend` only comes into play as a last resort fallback if the worker's own backend chain doesn't resolve the `conn_id`. When that happens, the server resolves via its own chain: env vars, then the metastore DB. The DB path is safe by construction: `conn_type` is a `NOT NULL` column, so a connection reaching ConnectionResponse via that fallback always has a conn_type.
> 
> So the "misleading connection not found error" this PR is trying to prevent doesn't actually occur on the path it changes, that path already works correctly per #61728, and tightening it back to reject missing conn_type only breaks that compatibility fix without fixing anything real.

Sorry for the misunderstanding here. I’ve updated the PR to just add an explanatory comment in the code so others don’t attempt the same change in the future.

Lee-W (approved):

shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
```diff
@@ -122,6 +122,12 @@ def _get_connection_class(self) -> type:
     def _deserialize_connection_value(conn_class: type, conn_id: str, value: str):
         value = value.strip()
         if value[0] == "{":
+            # JSON secrets stored by Airflow 2-era backends may lack both "conn_type"
```
amoghrajesh:
```suggestion
            # JSON secrets stored by Airflow 2 backends may lack both "conn_type"
```

shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
```diff
@@ -122,6 +122,12 @@ def _get_connection_class(self) -> type:
...
+            # and "uri" (e.g. {"host": ..., "login": ..., "password": ...}). This is
+            # valid: conn_type is intentionally optional on the SDK Connection model
+            # for Airflow 2 -> 3 migration compatibility. Do not add validation here
+            # that rejects such secrets; see
+            # https://github.com/apache/airflow/pull/61728
```
amoghrajesh:
```suggestion
            # for Airflow 2 -> 3 migration compatibility. 
            # Check: https://github.com/apache/airflow/pull/61728
```

shared/secrets_backend/tests/secrets_backend/test_base.py
```diff
@@ -124,6 +124,22 @@ def test_deserialize_connection_json(self, sample_conn_json):
         assert conn.conn_id == "test_conn"
         assert conn._kwargs["conn_type"] == "mysql"
 
+    def test_deserialize_connection_json_without_conn_type(self):
+        """Airflow 2-era JSON secrets without conn_type or uri must keep deserializing.
+
+        Guards the Airflow 2 -> 3 migration compatibility established in
+        https://github.com/apache/airflow/pull/61728.
+        """
```
amoghrajesh:
```suggestion
        """
        Guards the Airflow 2 -> 3 migration compatibility established in
        https://github.com/apache/airflow/pull/61728.
        """
```

shared/secrets_backend/src/airflow_shared/secrets_backend/base.py
amoghrajesh: Maybe also mention a TODO to remove it after min supported airflow version in providers is 3.0

henry3260: > Maybe also mention a TODO to remove it after min supported airflow version in providers is 3.0

Thanks, applied all of comments!

amoghrajesh (approved):

PR #64051, opened 2026-03-21
Implement execution_timeout semantics for AirbyteTriggerSyncOperator in deferrable mode

SameerMesiah97:
**Description**

This change enforces `execution_timeout` for `AirbyteTriggerSyncOperator` in deferrable mode.

Previously, when the operator deferred, `execution_timeout` was not enforced, allowing Airbyte jobs to continue running after the Airflow task had timed out.

The operator now computes an execution deadline before deferring, the trigger emits a timeout event when the deadline is exceeded, and the operator cancels the Airbyte job and fails the task when the event is received.

**Rationale**

In non-deferrable mode, `execution_timeout` is enforced by the scheduler, which terminates the task process and invokes `on_kill()` to cancel the external Airbyte job.

In deferrable mode, execution is handed off to a trigger running in the triggerer process, so there is no worker process to terminate. However, this does not change the expected task semantics. From a user perspective, `execution_timeout` is a **hard task-level limit and should behave consistently regardless of execution mode**.

Without explicit handling in the trigger/operator interaction, `execution_timeout` becomes a no-op in deferrable mode, leading to leaked Airbyte jobs and inconsistent behavior.

This is the same class of issue addressed in PR #61472 for `DbtCloudRunJobOperator`.

**Notes**

* The existing `timeout` parameter only controls how long the operator waits for job completion and does not imply cancellation.
* When both `execution_timeout` and `timeout` are set, the earlier deadline takes precedence.
* In `execute_complete`, several generic `AirflowException` raises have been replaced with more appropriate, specific exception types.
* Cancellation on execution_timeout is implemented as best-effort. Failures during cancel_job are logged but do not mask the timeout; the task still fails with a timeout error.
* The same best-effort semantics have been implemented for `on_kill` for consistency; cancellation errors are logged and not re-raised.

**Tests**

* Added a trigger-level test asserting a timeout event when the execution deadline is exceeded.
* Added an operator-level test verifying job cancellation on execution timeout.
* Added an operator-level test asserting that cancellation failures are swallowed and do not mask the timeout (timeout is still raised).
* Added a operator-level test verifying that `on_kill` does not raise if `cancel_job` fails (best-effort behavior).
* Modified a trigger serialization test to ensure the optional `execution_deadline` field is always serialized.

**Documentation**

* Updated the trigger docstring to document the new `execution_deadline` parameter.
* Updated the operator docstring to clarify the behavior of `timeout` vs `execution_timeout`; added an entry for `execution_timeout`.   

[...]

PR #69786, opened 2026-07-12
Fix AirbyteJobSensor marking a cancelled job as success in deferrable mode

steveahnahn:
### Problem

- `AirbyteJobSensor` in deferrable mode resumes via `execute_complete`, which only fails on an `error` event.
- A job cancelled while the sensor is deferred makes `AirbyteSyncTrigger` yield a `cancelled` event, which fell through to the success path: the task is marked SUCCESS and downstream tasks run against a sync that never completed.
- The same sensor fails correctly on a cancelled job in every other path: non-deferrable `poke()` raises, the deferrable `execute()` raises when it sees `CANCELLED` before deferring, and the sibling `AirbyteTriggerSyncOperator.execute_complete` already fails on the `cancelled` event (added in #64051, which did not cover the sensor).

### Change

- Fail on any non-success event in the sensor's `execute_complete`, matching `poke()`. The `cancelled` event now fails the task, and unrecognized statuses fail closed instead of being treated as success.

### Live verification

Ran a real `airflow standalone` (scheduler, triggerer, API server) against a live local endpoint serving the Airbyte jobs API, with the connection host pointed at it. Both runs: sensor deferred while the job was `running`, then the job was flipped to `cancelled`.

- Control run on current main: trigger yielded the cancelled event, task log shows "completed successfully.", run marked **Success**.
- Same scenario with this fix: task fails with `RuntimeError: Job run 1 has been cancelled.`, run marked **Failed**.

control run on main marked Success on a cancelled job; fixed run marked Failed

### Tests

- `test_execute_complete_fails_when_job_did_not_succeed` (parametrized: `cancelled`, unrecognized status) fails on the code before this change; the failure output shows the bug verbatim ("completed successfully." on a cancelled event).
- `test_execute_complete_succeeds_on_success_event` guards the happy path.

---

### DIFF
--- a/providers/airbyte/src/airflow/providers/airbyte/sensors/airbyte.py
+++ b/providers/airbyte/src/airflow/providers/airbyte/sensors/airbyte.py
@@ -126,15 +126,16 @@ class AirbyteJobSensor(BaseSensorOperator):
                 )
 
     def execute_complete(self, context: Context, event: Any = None) -> None:
-        """
-        Invoke this callback when the trigger fires; return immediately.
-
-        Relies on trigger to throw an exception, otherwise it assumes execution was
-        successful.
-        """
+        """Invoke this callback when the trigger fires and fail unless the job succeeded."""
         if event["status"] == "error":
             self.log.debug("An error occurred with context: %s", context)
             raise AirflowException(event["message"])
+
+        if event["status"] != "success":
+            # A cancelled job must fail the sensor, matching the poke() behavior;
+            # fail closed on any status this sensor does not recognize.
+            self.log.debug("Job did not succeed, context: %s", context)
+            raise RuntimeError(event["message"])
 
         self.log.info("%s completed successfully.", self.task_id)
         return None
--- a/providers/airbyte/src/airflow/providers/airbyte/sensors/airbyte.py
+++ b/providers/airbyte/src/airflow/providers/airbyte/sensors/airbyte.py
@@ -127,15 +127,21 @@ class AirbyteJobSensor(BaseSensorOperator):
 
     def execute_complete(self, context: Context, event: Any = None) -> None:
         """Invoke this callback when the trigger fires and fail unless the job succeeded."""
-        if event["status"] == "error":
-            self.log.debug("An error occurred with context: %s", context)
-            raise AirflowException(event["message"])
+        status = event["status"]
 
-        if event["status"] != "success":
-            # A cancelled job must fail the sensor, matching the poke() behavior;
-            # fail closed on any status this sensor does not recognize.
-            self.log.debug("Job did not succeed, context: %s", context)
+        if status == "success":
+            self.log.info("%s completed successfully.", self.task_id)
+            return None
+
+        if status == "error":
+            self.log.debug("An error occurred with context: %s", context)
             raise RuntimeError(event["message"])
 
-        self.log.info("%s completed successfully.", self.task_id)
-        return None
+        if status == "cancelled":
+            # A cancelled job must fail the sensor, matching the poke() behavior.
+            self.log.debug("Job was cancelled, context: %s", context)
+            raise RuntimeError(event["message"])
+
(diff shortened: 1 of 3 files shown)


Vamsi-klu:
Confirmed the gap: `AirbyteSyncTrigger.run()` can yield `cancelled` events and the old `execute_complete` only raised on `status == "error"`, so a cancelled job fell straight through to the success log. The new `event["status"] != "success"` guard closes that and matches `poke()`, which already raises on a cancelled job. One nit: the branch right above still raises `AirflowException` while the new one raises `RuntimeError` (that does line up with the sibling `AirbyteTriggerSyncOperator.execute_complete`), so this method now throws two different types depending on the path, worth a look if you care about consistency. Small thing on the comment too, `execute()` here defers without passing `execution_deadline`, so the trigger's `timeout` status can't actually fire for this sensor, though the catch-all still handles it fine. Tests cover the cancelled and fail-closed cases.

jroachgolf84 (approved):
LGTM, tests look good.

providers/airbyte/src/airflow/providers/airbyte/sensors/airbyte.py
```diff
@@ -126,15 +126,16 @@ def execute(self, context: Context) -> Any:
...
-
-        Relies on trigger to throw an exception, otherwise it assumes execution was
-        successful.
-        """
+        """Invoke this callback when the trigger fires and fail unless the job succeeded."""
         if event["status"] == "error":
             self.log.debug("An error occurred with context: %s", context)
             raise AirflowException(event["message"])
 
+        if event["status"] != "success":
```
SameerMesiah97: This fixes the immediate bug, but would it be clearer to handle cancelled explicitly rather than treating every non-success state as a generic failure? Since the trigger emits a well-defined set of terminal states, modelling those explicitly (success, error, cancelled, and an else for unexpected statuses) would be better. You can refer to the `execute_complete` method in the `AirbyteTriggerSyncOperator` for guidance. Naturally, the tests would have to be adjusted as well if you do this.

SameerMesiah97 (commented):
I think the state machine can be more explicit. Please see my comment below.

steveahnahn:
Both good points and addressed in  [02ac81f](https://github.com/apache/airflow/pull/69786/commits/02ac81f08604184621bd5ecd86b8e46774e95a97) as it now makes the states explicit (`success`/`error`/`cancelled`, else fails closed) and both branches now raise `RuntimeError`.

potiuk (approved):
Thanks — this is a data-correctness bug, not a cosmetic one, and worth spelling out.

`execute_complete` raised only on `status == "error"`. **Every other status fell through to the success path** — including `"cancelled"`. So a deferred `AirbyteJobSensor` whose sync was cancelled logged "completed successfully", returned `None`, and the task went green. Downstream tasks then ran against incomplete data with nothing anywhere indicating a problem. Silent success on a failed job is about the worst failure mode a sensor can have, and it diverged from `poke()`, which handled cancellation correctly — so the same sensor behaved differently depending on `deferrable`.

The fix is right, and the **fail-closed default is the part that matters most**. Handling `success`/`error`/`cancelled` explicitly fixes today's bug; raising on any *unmodelled* status fixes tomorrow's, when Airbyte adds a state this sensor has never heard of. Inverting the default from "assume success" to "assume failure" is the correct posture for a sensor.

The tests match: parametrising over `error`, `cancelled` and `unmapped_status` covers all three failure routes, and asserting `None` on success confirms the happy path wasn't broken in the process. Decrementing the ratchet 6 -> 5 rather than removing the line is also correct, since the other five raises remain.

One small improvement, not blocking: the fail-closed branch raises `RuntimeError(event["message"])`, so for an unrecognised status the operator sees only whatever Airbyte put in `message` — the status itself appears just in a `log.debug`. Including it in the exception (`f"Unexpected job status {status!r}: {event['message']}"`) would make the one case you can't anticipate the easiest to diagnose.

---

PR #69789, opened 2026-07-12
Discount CI image-build time from duration-trend alerts

potiuk:
The CI duration monitor folded the "Prepare breeze & CI image" step into each
job's total. That step occasionally balloons on a one-off cache miss (a full
image rebuild instead of a cached pull), which repeatedly flagged unrelated
jobs as regressed — e.g. a Non-DB core job reported +190% driven almost
entirely by an 18m image build, while its actual test time was flat.

Image-build time is now excluded from the run and per-job durations used for
the trend, and the image build is watched on its own: it is only reported
when it has stayed slow for more than two days (`IMAGE_BUILD_PERSISTENCE_DAYS`,
default 2), so a transient rebuild spike no longer produces a false alert.

### DIFF
--- a/scripts/ci/analyze_ci_job_durations.py
+++ b/scripts/ci/analyze_ci_job_durations.py
@@ -259,7 +259,7 @@ def get_prepare_breeze_step_duration(job: dict) -> float | None:
     return None
 
 
-def work_duration(job_data: JobDuration) -> float:
+def calculate_work_duration(job_data: JobDuration) -> float:
     """Return a job's wall-clock with the image-build (prepare breeze) step removed.
 
     The image build occasionally balloons — a cache miss forces a full rebuild
@@ -272,7 +272,7 @@ def work_duration(job_data: JobDuration) -> float:
     return max(job_data["duration"] - prepare_breeze, 0.0)
 
 
-def run_image_build_seconds(jobs: dict[str, JobDuration]) -> float | None:
+def calculate_image_build_seconds(jobs: dict[str, JobDuration]) -> float | None:
     """Return a representative image-build duration for a run.
 
     The same CI image is prepared by every job, so the median prepare-breeze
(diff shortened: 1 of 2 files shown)


scripts/ci/analyze_ci_job_durations.py
```diff
@@ -263,6 +259,34 @@ def get_prepare_breeze_step_duration(job: dict) -> float | None:
     return None
 
 
+def work_duration(job_data: JobDuration) -> float:
```
Lee-W:
```suggestion
def calculate_work_duration(job_data: JobDuration) -> float:
```

scripts/ci/analyze_ci_job_durations.py
```diff
@@ -263,6 +259,34 @@ def get_prepare_breeze_step_duration(job: dict) -> float | None:
...
+    (minutes → tens of minutes) — which would otherwise inflate the job's total
+    and flag an unrelated job as "slower". The duration trend should track the
+    actual test/work time, so image build is discounted from it and watched
+    separately by :func:`detect_image_build_regression`.
+    """
+    prepare_breeze = job_data["prepare_breeze_duration"] or 0.0
+    return max(job_data["duration"] - prepare_breeze, 0.0)
+
+
+def run_image_build_seconds(jobs: dict[str, JobDuration]) -> float | None:
```
Lee-W:
```suggestion
def calcuate_image_build_seconds(jobs: dict[str, JobDuration]) -> float | None:
```

scripts/ci/analyze_ci_job_durations.py
```diff
@@ -587,32 +699,43 @@ def main() -> None:
...
 
     latest_runs = runs[:latest_runs_count]
     baseline_runs = runs[latest_runs_count:]
     print(f"Latest runs: {len(latest_runs)}; baseline runs: {len(baseline_runs)}.")
 
+    # Fetch each run's jobs once: they feed the image-build discount below, the
+    # per-job trend, and the image-build persistence check.
+    jobs_by_run_id = fetch_run_jobs_map(repo, runs) if do_analyze_jobs else {}
+
+    def adjusted_run_duration(run: dict) -> float:
```
Lee-W:
```suggestion
    def calculate_adjusted_run_duration(run: dict) -> float:
```

Lee-W (approved):
a few naming nits

Issue #67461, opened 2026-05-25
Add dedicated permission for viewing import errors of unregistered Dag files (with multi-team support)


potiuk:
## Background

The public Import Errors API resolves each `ParseImportError` to a set
of `DagModel` rows via `(relative_fileloc, bundle_name)` and applies
per-Dag authorization on the resulting set. There is one case the
current code cannot authorize cleanly: a parse error for a file that
has **no** `DagModel` rows yet — for example, a brand-new file that
failed to parse before any Dag was defined, or a file whose Dags were
all removed.

The single endpoint and the list endpoint currently return the raw
stacktrace in this case. That is the same behavior the endpoints had
before the per-file authorization work landed, and it is what the
follow-up PR restores while a proper design is in place.

## Why a follow-up

Two things make a proper fix non-trivial:

1. **There is no per-file permission today.** Authorization on import
   errors is derived from per-Dag read permission, but a file with no
   Dag has nothing to derive from. The right answer is a dedicated
   "view all import errors" permission rather than overloading the
   existing per-Dag check.

2. **Multi-team isolation needs to be respected.** Once such a
   permission exists it must be scoped per team (so a team admin sees
   import errors for files owned by their team, not other teams). The
   per-file ownership signal for unregistered files needs to come from
   the bundle / team mapping rather than from `DagModel`.

## Proposed direction

- Introduce a new `AccessView` (working name: `IMPORT_ERRORS_ALL`, or a
  resource-action pair on `IMPORT_ERRORS`) that grants visibility to
  import errors whose file has no registered Dag.
- Default-grant the new permission to the admin role only.
- Wire the per-file authorization in
  `airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py`
  so that, when `file_dag_ids` is empty, the response is conditioned on
  the caller holding the new permission — returning the raw stacktrace
  when they do, redacting (or 403'ing) when they don't.
- For multi-team deployments, scope the new permission per team using
  the bundle → team mapping, so a team admin only sees unregistered-file
  errors that belong to their team.
- Update the public-API docs and any auth-manager implementations
  (FAB, Keycloak, simple) to declare and surface the new permission.

## Acceptance criteria

- New permission exists in `AccessView` and is declared by every
  bundled auth manager.
- Default role mappings grant it to admins only.
- Single and list endpoints return the raw stacktrace for files with
  no registered Dag **only** when the caller has the permission;
  otherwise the file is redacted or hidden, with the team scoping
[...]

PR #69790, opened 2026-07-12
Add IMPORT_ERRORS_ALL permission for import errors of files with no registered Dag

potiuk:
Import errors are authorized per Dag defined in the file. Files with no registered Dag (parse failed before any Dag was defined, or all Dags removed) have no per-Dag key to authorize on. This adds a dedicated `IMPORT_ERRORS_ALL` view -- admin-granted by default, scoped per team via the file's bundle where the auth manager supports multi-team isolation -- and gates such files on it: callers without the permission are denied those import errors (`403` on the single endpoint; excluded from the list endpoint) instead of receiving the raw stack trace.

Implements #67461.

- New `AccessView.IMPORT_ERRORS_ALL`.
- Auth managers gain an `is_authorized_view_for_team` method that defaults to `is_authorized_view`, so auth managers that do not implement it (including out-of-tree ones) keep working unchanged. The simple and Keycloak managers override it to enforce team scoping; the FAB manager grants the new permission to `Admin` by default.
- Both Import Errors endpoints resolve the file's team from its bundle and authorize unregistered-file errors on the new view; the list endpoint excludes unauthorized rows from the results and `total_entries` so their existence does not surface.

### Tests

- [x] `test_import_error.py` -- with the permission the raw stack trace is returned; without it the single endpoint returns `403` and the list endpoint omits the rows; team scoping is honored; no query-count regression.
- [x] `test_base_auth_manager.py` -- `is_authorized_view_for_team` falls back to `is_authorized_view` by default.
- [x] simple / FAB / Keycloak auth-manager suites.

- [X] Yes -- Claude Opus 4.8 (1M context)

Generated-by: Claude Opus 4.8 (1M context) following the guidelines at


potiuk:
@vincbeck — you'd offered to pick up #67461, so heads-up that I went ahead and implemented it here. Would you mind reviewing when you get a chance? Particularly keen on your read of the `is_authorized_view_for_team` addition (kept backwards-compatible so existing auth managers are untouched) and the deny-by-default behaviour for files with no registered Dag. Happy to adjust.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
```diff
@@ -159,6 +173,31 @@ def get_import_errors(
     # Subquery for files that have any Dags
     files_with_any_dags = select(DagModel.relative_fileloc).distinct().subquery()
```
henry3260:
```suggestion
    files_with_any_dags = select(DagModel.relative_fileloc, DagModel.bundle_name,).distinct().subquery()
```

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
```diff
@@ -159,6 +173,31 @@ def get_import_errors(
...
+    # Import errors for files that have **no** registered Dag have no per-Dag
+    # key to authorize on. Authorize them on the dedicated ``IMPORT_ERRORS_ALL``
+    # view (admin by default), scoped to the file's team via its bundle, and
+    # keep only the ones the caller may see. Filtering here (rather than
+    # redacting in the loop) also excludes unauthorized rows from the count and
+    # pagination, so their existence -- and, for team-scoped bundles, the
+    # cross-team file/bundle names -- does not leak.
+    unregistered_errors = session.execute(
+        select(ParseImportError.id, ParseImportError.bundle_name)
+        .outerjoin(files_with_any_dags, ParseImportError.filename == files_with_any_dags.c.relative_fileloc)
```
henry3260:
```suggestion
        .outerjoin(
            files_with_any_dags,
            (ParseImportError.filename == files_with_any_dags.c.relative_fileloc)
            & (ParseImportError.bundle_name == files_with_any_dags.c.bundle_name),
        )
```

henry3260 (commented):
I think we should check `bundle_name` as well here — otherwise a same-named file in another bundle that *does* have Dags will match, and this unregistered file skips the `IMPORT_ERRORS_ALL` check.

henry3260:
> I think we should check `bundle_name` as well here — otherwise a same-named file in another bundle that _does_ have Dags will match, and this unregistered file skips the `IMPORT_ERRORS_ALL` check.

Maybe we can add a regression test here :)

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -365,6 +365,28 @@ def is_authorized_view(
         :param user: the user to performing the action
         """
 
+    def is_authorized_view_for_team(
```
vincbeck: Cannot we update `is_authorized_view` and add `team_name` as optional parameter? I would prefer this option to keep it consistent with the rest of methods

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
potiuk: Indeed

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
potiuk: bundle names ftw

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
potiuk: Good catch — the same `relative_fileloc` can exist in more than one bundle, so matching on filename alone was wrong. Fixed: the has-any-Dag subquery now selects `bundle_name` and the join matches on both filename and bundle. Thanks!

---

airflow-core/src/airflow/api_fastapi/core_api/routes/public/import_error.py
potiuk: Good catch — the same `relative_fileloc` can exist in more than one bundle, so matching on filename alone was wrong. Fixed: the has-any-Dag subquery now selects `bundle_name` and the join matches on both filename and bundle. Thanks!

---

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
potiuk: I'd rather keep the separate method here, mainly for backwards compatibility. `is_authorized_view` is an `@abstractmethod`, and Python's ABC machinery only checks that the *name* is overridden, not the signature — so adding `team_name` there wouldn't force existing implementations to update, but any out-of-tree AuthManager that keeps the current `(access_view, user)` signature would raise `TypeError` the moment core calls it with `team_name` (a runtime break, with no import- or instantiation-time warning). The separate `is_authorized_view_for_team` with a default fallback avoids that — out-of-tree managers keep working unchanged, and only multi-team-aware managers override it.

We *could* instead fold it into `is_authorized_view` with an optional `team_name` and take it as a breaking change for out-of-band AuthManagers, but I'd prefer to avoid that. WDYT?

---

vincbeck: > We could instead fold it into is_authorized_view with an optional team_name and take it as a breaking change for out-of-band AuthManagers, but I'd prefer to avoid that. WDYT?

To be honest I think I rather having the breaking change. The method `is_authorized_view_for_team` feels wrong to me. If we had done the same for the other methods then we would have duplicated all the methods in the auth manager interface

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -357,13 +359,52 @@ def is_authorized_view(
...
+        :param team_name: team the view is scoped to, if any
+        """
+
+    @cached_property
+    def _is_authorized_view_team_aware(self) -> bool:
+        """Whether this manager's ``is_authorized_view`` override accepts ``team_name``."""
+        params = inspect.signature(self.is_authorized_view).parameters
+        return "team_name" in params or any(p.kind is inspect.Parameter.VAR_KEYWORD for p in params.values())
+
+    def authorize_view(self, *, access_view: AccessView, user: T, team_name: str | None = None) -> bool:
```
vincbeck, line 379: Why dont we:
- Remove the `@abstractmethod` from `is_authorized_view`
- Move this implementation to `is_authorized_view`
?

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
potiuk: Hmm. I do not think we can have such a breaking change in Auth Manager, the issue is that if the previous Auth Managera (for exampl current AWS provider) would be used with the new Airflow - it would raise 500 when the view for import errors is called - simply because we try to run the method with `team_name`.

So I updated it this way now.  Let me know what you think. It does add "back-compat" code that is not "beautiful" - but I think we cannot just make the older versions of auth managers to stop working ?

 We need to bite this bullet I guess?

vincbeck: I dont know if you saw: https://github.com/apache/airflow/pull/69790#discussion_r3622564754

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
potiuk: It does not change anything - the is_authorized_view implemented by old versions of Auth Manager will override implementation of it anyway - and will fail with "team_name parameter missing" when called.

vincbeck: Oh yeah right ... I really hate introducing a new method because of a new param ...

vincbeck: Cannot we do the inspect on the caller side? If team_name is a parameter of `is_authorized_view` then let's pass it through, otherwise no?

vincbeck: That's what you do basically, just the method is in base auth manager interface

vincbeck: When providers will have min airflow version >= 3.4, will we be able to remove this method?

potiuk: > When providers will have min airflow version >= 3.4, will we be able to remove this method?

Not really - that is all about "past" providers, being used with new Airflow - and it's a bit of logical problem I am afraid, because simply the older providers will not have this functionality at all.

That does seam as a new feature of AuthManager and I think we need to figuewre out "good" way of extending it ?

vincbeck (approved):

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
vincbeck: Alright, I am still not a big fan of the solution but I dont have better

bugraoz93 (approved):

pierrejeambrun (approved):
LGTM. I would trim down some of the very verbose comments. But non blocking.

potiuk:
> LGTM. I would trim down some of the very verbose comments. But non blocking.

Trimmed down some of those :)

PR #69792, opened 2026-07-13
Validate deferred trigger classpath resolves to a BaseTrigger subclass

hypnguyen1209:
The triggerer resolves a trigger from the ``classpath`` carried in the deferred-task Execution API payload via ``import_string(classpath)`` and then instantiates it with ``trigger_class(**kwargs)``. ``get_trigger_by_classpath`` did not check that the imported object is actually a ``BaseTrigger`` subclass, so any importable callable (e.g. ``subprocess.check_output``) could be loaded and invoked in the long-running triggerer process.

Harden this by rejecting, before caching and before instantiation, any ``classpath`` that does not resolve to a ``type`` that is a ``BaseTrigger`` subclass. Legitimate triggers are unaffected (they are always ``BaseTrigger`` subclasses); an invalid classpath now fails the trigger cleanly instead of being instantiated. Adds a unit test covering both the accepted and rejected cases.

 

---

---


airflow-core/src/airflow/jobs/triggerer_job_runner.py
```diff
@@ -1696,8 +1696,22 @@ def get_trigger_by_classpath(self, classpath: str) -> type[BaseTrigger]:
...
+        originates from the (attacker-influenceable) deferred-task payload, so
+        without this check an arbitrary importable callable could be invoked in
+        the triggerer process.
+
         Uses a cache dictionary to speed up lookups after the first time.
         """
         if classpath not in self.trigger_cache:
-            self.trigger_cache[classpath] = import_string(classpath)
+            trigger_class = import_string(classpath)
+            if not (isinstance(trigger_class, type) and issubclass(trigger_class, BaseTrigger)):
```
potiuk, line 1731: I think this should be pre-import check, unfortunately - by the time we get here the class is already imported, and the whole idea is to avoid even importing it - because just importing it can have side-effects. I am not sure if that one is even reasonably doable - because in order to get class hierarchy you need to import it. Just AST parsing will not solve it.

So I am not sure if that is solving such defense-in-depth is even doable.

potiuk (changes requested):
Hmm... I think that's not really preventing the side effects it was supposed to prevent.

airflow-core/src/airflow/jobs/triggerer_job_runner.py
hypnguyen1209: > I think this should be pre-import check, unfortunately - by the time we get here the class is already imported, and the whole idea is to avoid even importing it - because just importing it can have side-effects. I am not sure if that one is even reasonably doable - because in order to get class hierarchy you need to import it. Just AST parsing will not solve it.
> 
> So I am not sure if that is solving such defense-in-depth is even doable.

Correct - the issubclass check runs after import_string, so it doesn't stop import-time side effects of an arbitrary module.

To be precise about what it does stop: the reported path is classpath="subprocess.check_output" + trigger_kwargs, where the code executes at trigger_class(**kwargs) (the call) - importing stdlib subprocess is side-effect-free, so the issubclass gate blocks the actual execution. It does not help when the classpath points at a module whose import has side effects (e.g. an attacker-plantable module on sys.path).

A genuinely pre-import defense is doable, just not via issubclass (which needs the class, hence the import). It has to be a string check on the classpath before import_string - a module-namespace allow-list:

```
module = classpath.rpartition(".")[0]
 if not (module == "airflow" or module.startswith("airflow.") or module in allowed_extra):
     raise TypeError(...)   # before import_string()
```

That never imports anything outside the trusted namespace (core airflow.* + providers airflow.providers.*), so import-time side effects of untrusted/planted modules are never triggered. The issubclass check can stay after import as defense-in-depth.

The trade-off is custom triggers in user modules (mycompany.triggers.X): those would need a configurable prefix allow-list (similar to [core] allowed_deserialization_classes) to keep working - a behaviour change plus a new config option, which is why I wanted to check the direction before implementing.

potiuk (approved):

PR #69411, opened 2026-07-05
Speed up lang-SDK k8s test by building Go/Java natively in CI

potiuk:
The Multi-Lang KubernetesExecutor system test (#68709) builds a Go bundle and a Java jar inside throwaway toolchain containers. On a dev host the Go module/build cache, the Gradle distribution and the dependency caches persist between runs, but CI runners are ephemeral and nothing cached them — so every scheduled run re-pulled the `golang`/`eclipse-temurin` images and re-downloaded the Gradle distribution and all dependencies. That added ~10 min to each of the six `KubernetesExecutor` + standard-naming-off variants (the only ones that run the test), which the daily CI duration-trend alert flagged.

This builds the artifacts with the **host toolchain** in CI instead, provisioned and cached through `actions/setup-go` and `actions/setup-java` (the same pattern the sibling *Go SDK tests* / *Java SDK tests* jobs already use). A new `LANG_SDK_NATIVE_TOOLCHAIN=true` env, set only for that variant, makes breeze invoke the host `go` / `./gradlew` rather than `docker run`. Local runs keep the containerised build, so a dev host still needs neither Go nor a JDK installed.

Measured on run `28635348211`: the parallel provisioning block was ~5m36s, almost entirely toolchain-image pulls + cold Gradle/Maven/Go dependency downloads. Removing the image pulls and warming the caches should cut that to ~1–1.5 min (~4 min/job); the subsequent helm upgrade and the test itself are unaffected.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69239, opened 2026-07-02
Speed up the java_sdk e2e Gradle builds with concurrency and caching

jason810496:
- related: #68939 
- follow-up of: https://github.com/apache/airflow/pull/68939#pullrequestreview-4606311659

## Why

The java_sdk e2e job builds the two example bundles sequentially and starts from an empty `GRADLE_USER_HOME` on every run, forcing a full re-download of the bundles' dependency tree (Spark alone is ~1GB) each time.

## What

- Build the `example` and `scala_spark_example` bundles concurrently; they are independent Gradle builds that both only consume the SDK artifact published beforehand.
  - The bundle containers run with `--network=host`: Gradle's cross-process lock handover pings the lock owner over loopback, which isolated container network namespaces would break (60s lock timeouts).
  - Each build's output is captured and printed under its own header, so concurrent logs do not interleave and failures name their bundle.
- Cache `java-sdk/.gradle` (caches, wrapper, jdks) in the e2e workflow for the java_sdk mode.
  - `actions/cache/restore` + `actions/cache/save` split with `if: always()`, so a completed Gradle warm-up is saved even when the e2e tests later fail.
  - The key hashes both Groovy and Kotlin DSL build files plus the version catalog, and includes `runner.arch` so arch-specific auto-provisioned JDKs do not cross between amd64 and arm64 runners.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes, with help of Claude Code Fable 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69794, opened 2026-07-13
Build Java SDK e2e example bundles natively in CI

jason810496:
related:
- follow-up of: https://github.com/apache/airflow/pull/69239#issuecomment-4925259828
- PR: #69411

## Why

#69239 (java_sdk e2e mode) and #69411 (lang-SDK k8s test) solved the same cold-Gradle-build problem two different ways: containerized builds with a repo-local cache vs. host-toolchain builds with salted home-dir caches.

This converges the e2e `java_sdk` mode on the #69411 native-toolchain convention.

## How

- CI exports `LANG_SDK_NATIVE_TOOLCHAIN=true` (the same switch the k8s job uses) so the conftest invokes the host `./gradlew` instead of `docker run eclipse-temurin:17-jdk` — the build helpers take a `native` flag mirroring the #69411 k8s helpers. This drops the per-run toolchain-image pull and the container workarounds (`--network=host` lock handover, `HOME` override). Local runs keep the containerized build, so a dev host still needs no JDK.
- The JDK comes from `actions/setup-java`, versioned by the existing `java-sdk-version` build-info output, as `k8s-tests.yml` already does.
- The Gradle cache moves to the default `~/.gradle` with a salted key (`e2e-java-sdk-gradle-v1-`, bump to invalidate), kept separate from the k8s job's key since the e2e tree carries ~1GB of Spark the k8s job never needs.

---

### DIFF
--- a/.github/workflows/additional-prod-image-tests.yml
+++ b/.github/workflows/additional-prod-image-tests.yml
@@ -56,6 +56,10 @@ on:  # yamllint disable-line rule:truthy
         description: "Whether to run Java SDK e2e tests (true/false)"
         required: true
         type: string
+      java-sdk-version:
+        description: "JDK version used to build the Java SDK bundles natively in the e2e tests"
+        required: true
+        type: string
       run-go-sdk-e2e-tests:
         description: "Whether to run Go SDK e2e tests (true/false)"
         required: true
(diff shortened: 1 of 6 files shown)


uranusjr (approved):
Makes sense

amoghrajesh (approved):

PR #69575, opened 2026-07-07
Airflow Data Quality Provider part 1

gopidesupavan:
Adds a new `apache-airflow-providers-common-dataquality` provider for `DbApiHook`-based data quality checks.

Airflow already has SQL check operators, and many users rely on them for data quality today. This provider adds a `DQRule` / `RuleSet` layer for checks that need stable rule identity, persisted history, and a connection to Airflow assets. That makes quality results easier to analyze over time, lets downstream asset consumers gate on recent quality, and gives LLM-assisted workflows one schema to generate when proposing checks from table context. Execution still goes through existing `common.sql` / `DbApiHook` connections.

This PR is the backend/provider slice only. The UI plugin and read-only API are intentionally left for a follow-up PR.

Ships:

- `DQRule` and `RuleSet` models for named data quality rules.
- Built-in SQL checks for common table and column checks, executed through `common.sql` / `DbApiHook`.
- `custom_sql` support for database-specific or more complex checks.
- `DQCheckOperator` and the `@task.dq_check` TaskFlow decorator.
- A configurable results backend under `[dq] results_path` for task, run, and rule-level history.
- Experimental asset helpers, `asset_quality()` and `require_quality()`, that attach provider-owned quality metadata to assets without changing Airflow core.
- Documentation and example Dags covering end-to-end usage with and without LLM-generated rules.
- A DQ rule-authoring skill that LLM-assisted workflows can use to generate rules from table/schema context:
  https://github.com/gopidesupavan/airflow/blob/f32940bd261b94238256eaced9150dd51329ce3e/providers/dq/src/airflow/providers/dq/skills/dq-rule-authoring/SKILL.md

This first version is intentionally focused on the backend contract: deterministic rule definitions, SQL execution through existing Airflow SQL providers, persisted results, and asset-linked quality summaries.

Design decisions:

- Results are stored through an object-storage/local-file backend instead of adding new metadata DB tables in the first provider drop. This keeps the provider self-contained, avoids Airflow core migrations, and lets deployments choose a durable store such as S3, GCS, or local files via `[dq] results_path`.
- The backend stores keyed JSON records for task runs, task instances, and per-rule history so later readers, including a future UI/API layer, can access common views without scanning unrelated runs.
- Asset support is implemented with provider-owned metadata, not Airflow core changes. Static quality configuration is attached to `Asset.extra["airflow.dq"]`; runtime summaries are attached to asset events under `extra["airflow.dq.result"]`.
[...]

PR #69413, opened 2026-07-05
Airflow Data Quality Provider Plugin

gopidesupavan:
Adds a new `apache-airflow-providers-dq` provider,
`DbApiHook`-based data quality checks.

Airflow already has SQL check operators, and many users rely on them for data
quality today. This provider does not replace that path; it adds a small
`DQRule` / `RuleSet` layer for checks that need stable rule identity, persisted
history, and a connection to Airflow assets. That makes quality results easier
to inspect over time, lets downstream asset consumers gate on recent quality,
and also gives LLM-assisted workflows one schema to generate when proposing
checks from table context. Execution still goes through existing `DbApiHook`
connections.

Ships:

- `DQRule` and `RuleSet` models for named data quality rules.
- Built-in SQL checks for common table and column checks, executed through
  `common.sql` / `DbApiHook`, plus `custom_sql` for database-specific or more
  complex checks.
- `DQCheckOperator` and the `@task.dq_check` TaskFlow decorator.
- A configurable results backend under `[dq] results_path` for task, run, and
  rule-level history.
- A read-only API and minimal UI plugin for viewing task/run results and rule
  history.
- Experimental asset helpers, `asset_quality()` and `require_quality()`, that
  attach provider-owned quality metadata to assets without changing Airflow
  core.
- Documentation and example Dags covering end-to-end usage with and without
  LLM-generated rules.
- The provider addiitonally ships with dq authoring skills, that LLM's can use to generate rules based on the users need. SKILLS  https://github.com/gopidesupavan/airflow/blob/f32940bd261b94238256eaced9150dd51329ce3e/providers/dq/src/airflow/providers/dq/skills/dq-rule-authoring/SKILL.md

This first version is intentionally small. It focuses on a deterministic rule
shape, SQL execution through `common.sql`, persisted results, and lightweight
visibility in the Airflow UI.

Design decisions:

- Results are stored through an object-storage/local-file backend instead of
  adding new metadata DB tables in the first provider drop. This keeps the
  provider self-contained, avoids Airflow core migrations, and lets deployments
  choose a durable store such as S3/GCS/local files via `[dq] results_path`.
  The backend stores keyed JSON records for task runs, task instances, and
  per-rule history so the UI can read common views without scanning unrelated
  runs.
- Asset support is implemented by extending assets with provider-owned metadata,
  not by changing Airflow core. Static quality configuration is attached to
  `Asset.extra["airflow.dq"]`; runtime summaries are attached to asset events
[...]

PR #69795, opened 2026-07-13
Add Common Data Quality provider skeleton

gopidesupavan:
Adds the initial skeleton for the `apache-airflow-providers-common-dataquality` provider.

This PR only introduces the provider package structure and repository wiring. It does not add data quality operators, hooks, rules, UI, or runtime behavior yet. The goal is to land the provider skeleton first so follow-up PRs can add functionality in smaller reviewable steps.

---

---


providers/common/dataquality/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Do not commit hash files, this is just to speed-up local builds
```
Lee-W: why do we need this

providers/common/dataquality/.gitignore
```diff
@@ -0,0 +1,3 @@
...
+www-hash.txt
```
Lee-W: same for this

providers/common/dataquality/.gitignore
gopidesupavan: removed

providers/common/dataquality/.gitignore
gopidesupavan: yeah it was wrongly copied from the main PR https://github.com/apache/airflow/pull/69413 :) , removed it now

Lee-W (commented):
Are we expected to have something really basic, or is it okay to keep it mostly empty for the first phase?

gopidesupavan:
> Are we expected to have something really basic, or is it okay to keep it mostly empty for the first phase?

yeah the reason is provider with functionality is getting more changes into single PR and it feels heavy .. have here with functionality https://github.com/apache/airflow/pull/69575. after the skelton merge will rebase immediately.  so would like to merge provider skelton first :)

airflow-core/tests/unit/always/test_project_structure.py
```diff
@@ -107,6 +107,7 @@ def test_providers_modules_should_have_tests(self):
             "providers/common/compat/tests/unit/common/compat/standard/test_utils.py",
             "providers/common/messaging/tests/unit/common/messaging/providers/test_base_provider.py",
             "providers/common/messaging/tests/unit/common/messaging/providers/test_sqs.py",
+            "providers/common/dataquality/tests/unit/common/dataquality/test_exceptions.py",
```
kaxil: This entry points at a test for a module this PR doesn't add, there's no `exceptions.py` in the skeleton, so it's a no-op here. Looks like another carry-over from #69413, like the .gitignore entries. Since the comment above says this list should shrink to zero, I'd drop it here and add a real `test_exceptions.py` when the follow-up brings in `exceptions.py`.

.github/boring-cyborg.yml
```diff
@@ -141,6 +141,9 @@ labelPRBasedOnFilePath:
   provider:docker:
     - providers/docker/**
 
+  provider:common-dataquality:
```
kaxil: The other common-* labels sit alphabetically further up (common-ai through common-sql). This one landed between docker and edge, worth moving next to common-compat/common-io.

providers/common/dataquality/docs/commits.rst
```diff
@@ -0,0 +1,35 @@
...
+
+ .. THE REMAINDER OF THE FILE IS AUTOMATICALLY GENERATED. IT WILL BE OVERWRITTEN!
+
+Package apache-airflow-providers-common-dataquality
+------------------------------------------------------
+
+``Data Quality Provider``
+
+
+This is detailed commit list of changes for versions provider package: ``dataquality``.
```
kaxil: The template renders the provider id here, so regeneration would produce ``common.dataquality`` (compare common/ai and common/io), and the description line above would be "Common Data Quality Provider" from provider.yaml. Worth matching the generated output so the first release prep doesn't produce a diff.

kaxil (approved):

airflow-core/tests/unit/always/test_project_structure.py
gopidesupavan: yeah bad commit ..removed now.

providers/common/dataquality/docs/commits.rst
gopidesupavan: updated

Lee-W:
> > Are we expected to have something really basic, or is it okay to keep it mostly empty for the first phase?
> 
> yeah the reason is provider with functionality is getting more changes into single PR and it feels heavy .. have here with functionality #69575. after the skelton merge will rebase immediately. so would like to merge provider skelton first :)

Thanks! Will also take a look at that PR!

PR #69799, opened 2026-07-13
Add deferrable mode to DynamoDBValueSensor

### LINKED ISSUE #69051: [amazon] Add deferrable mode to DynamoDBValueSensor via DynamoDB Streams + SQS
### Description

DynamoDBValueSensor only supports blocking poll mode — it holds an Airflow worker slot for the entire wait duration and poke_interval determines how quickly it reacts after the target value is set.

### Use case/motivation

A common Airflow pattern is using DynamoDB as a lightweight approval or job-state store — wait for a status field to flip to a target value (e.g. APPROVED, COMPLETED) before proceeding. The wait window can be hours for human approval workflows.

  With the current blocking sensor:
  - A worker slot is held for the entire wait window (hours)
  - Reaction latency equals one full poke_interval — typically set long (e.g. 6000s) precisely to avoid burning worker slots, which makes latency worse
  - This creates a forced tradeoff: short poke_interval wastes workers, long poke_interval means slow reaction

  Deferrable mode breaks the tradeoff:
  - Sensor defers immediately, freeing the worker slot
  - The Triggerer process handles the async polling loop at low cost
  - poke_interval can be set to 30–60 seconds without worker pressure
  - Result: hours of dead worker time eliminated, reaction latency drops from O(hours) to O(seconds)

  Proposed addition: DynamoDBValueTrigger (async, runs in Triggerer) + deferrable=True on DynamoDBValueSensor, following the same pattern as existing triggers in providers/amazon/src/airflow/providers/amazon/aws/triggers/.

  Sensor interface (backwards compatible):

    DynamoDBValueSensor(
        ...,
        deferrable=True,    # opt-in, default False
        waiter_delay=30,    # seconds between trigger polls
    )

  When deferrable=False (default), existing behaviour is completely unchanged.

### Related issues

- Existing SqsSensorTrigger: providers/amazon/src/airflow/providers/amazon/aws/triggers/sqs.py
  - Existing DynamoDBValueSensor: providers/amazon/src/airflow/providers/amazon/aws/sensors/dynamodb.py
  - Related CRUD hook PR: #69050 

### Code of Conduct

harshhh817:
Adds the standard deferrable path to `DynamoDBValueSensor`, which currently only supports poke/reschedule modes and holds a worker slot while waiting.

- `deferrable` param (default from `operators.default_deferrable`, per provider convention); `execute()` defers to a new `DynamoDBValueSensorTrigger`, `execute_complete()` validates the trigger event.
- The trigger polls the item asynchronously. `DynamoDBHook` is resource-based but aiobotocore only exposes clients, so the trigger uses a client-type `AwsBaseHook` with low-level `get_item` and deserializes the attribute via boto3's `TypeDeserializer` — semantics identical to the sync `poke()` (ClientError tolerated, single value or iterable matched). This mirrors how `EC2StateSensorTrigger` handles the same client/resource split.
- Trigger registered in `provider.yaml` / `get_provider_info.py`; docs note added; unit tests for sensor deferral and trigger (serialization + async run paths).

The issue proposes DynamoDB Streams + SQS; I went with the async-polling trigger instead since it is the established pattern for the provider's deferrable sensors and requires no extra AWS infrastructure — happy to discuss if maintainers prefer the Streams-based design.

closes: #69051

---

- [X] Yes (Claude Code)

---

### DIFF
--- a/providers/amazon/docs/operators/dynamodb.rst
+++ b/providers/amazon/docs/operators/dynamodb.rst
@@ -45,6 +45,7 @@ Wait on Amazon DynamoDB item attribute value match
 
 Use the :class:`~airflow.providers.amazon.aws.sensors.dynamodb.DynamoDBValueSensor`
 to wait for the presence of a matching DynamoDB item's attribute/value pair.
+This sensor can also be run in deferrable mode by setting ``deferrable`` param to ``True``.
 
 Wait for a Single Attribute Value Match:
 ----------------------------------------
--- a/providers/amazon/provider.yaml
+++ b/providers/amazon/provider.yaml
@@ -827,6 +827,9 @@ triggers:
   - integration-name: Amazon Comprehend
     python-modules:
       - airflow.providers.amazon.aws.triggers.comprehend
+  - integration-name: Amazon DynamoDB
+    python-modules:
+      - airflow.providers.amazon.aws.triggers.dynamodb
   - integration-name: Amazon EC2
     python-modules:
       - airflow.providers.amazon.aws.triggers.ec2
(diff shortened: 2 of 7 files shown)


o-nikolas (approved):

shahar1:
@harshhh817 could you please resolve conflicts?

harshhh817:
Thanks @shahar1 — rebased onto latest main and resolved the conflict (an import block). The provider unit tests pass locally. Should be good for another look.

PR #69800, opened 2026-07-13
Fix Azure Data Factory hook broken with azure-mgmt-datafactory 10

FrankYang0529:
## Why

- The CI dependency upgrade PR (#69694) bumps `azure-mgmt-datafactory` 9.3.0 → 10.0.0, which replaced the `if_match`/`if_none_match` parameters with `etag` + `match_condition` pair.
  - https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/datafactory/azure-mgmt-datafactory/CHANGELOG.md#breaking-changes

## How

- Use `azure.mgmt.datafactory` version to determine kwargs: `_ADF_SDK_V10_PLUS` is computed from the installed major and the hook's public `if_match`/`if_none_match` parameters are translated accordingly.
  - The v10+ sends `etag` + `match_condition`.
  - Older SDKs keep the original `if_match`/`if_none_match` kwargs.

## Verification

- `uv run --project providers/microsoft/azure --with "azure-mgmt-datafactory==9.3.0" pytest providers/microsoft/azure/tests/unit/microsoft/azure/hooks/test_data_factory.py -q`
- `uv run --project providers/microsoft/azure --with "azure-mgmt-datafactory==10.0.0" pytest providers/microsoft/azure/tests/unit/microsoft/azure/hooks/test_data_factory.py -q`

 

---

- [X] Yes - Claude Code

---

### DIFF
--- a/providers/microsoft/azure/docs/index.rst
+++ b/providers/microsoft/azure/docs/index.rst
@@ -128,7 +128,7 @@ PIP package                                 Version required
 ``azure-synapse-artifacts``                 ``>=0.17.0``
 ``azure-storage-file-datalake``             ``>=12.9.1``
 ``azure-kusto-data``                        ``>=4.1.0,!=5.0.0``
-``azure-mgmt-datafactory``                  ``>=2.0.0``
+``azure-mgmt-datafactory``                  ``>=10.0.0``
 ``azure-mgmt-containerregistry``            ``>=8.0.0``
 ``azure-mgmt-compute``                      ``>=33.0.0``
 ``azure-mgmt-containerinstance``            ``>=10.1.0``
--- a/providers/microsoft/azure/README.rst
+++ b/providers/microsoft/azure/README.rst
@@ -74,7 +74,7 @@ PIP package                                 Version required
 ``azure-synapse-artifacts``                 ``>=0.17.0``
 ``azure-storage-file-datalake``             ``>=12.9.1``
 ``azure-kusto-data``                        ``>=4.1.0,!=5.0.0``
-``azure-mgmt-datafactory``                  ``>=2.0.0``
+``azure-mgmt-datafactory``                  ``>=10.0.0``
 ``azure-mgmt-containerregistry``            ``>=8.0.0``
 ``azure-mgmt-compute``                      ``>=33.0.0``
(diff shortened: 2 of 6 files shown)


providers/microsoft/azure/docs/index.rst
```diff
@@ -128,7 +128,7 @@ PIP package                                 Version required
 ``azure-synapse-artifacts``                 ``>=0.17.0``
 ``azure-storage-file-datalake``             ``>=12.9.1``
 ``azure-kusto-data``                        ``>=4.1.0,!=5.0.0``
-``azure-mgmt-datafactory``                  ``>=2.0.0``
+``azure-mgmt-datafactory``                  ``>=10.0.0``
```
Lee-W, line 131: This is a huge bump. I'm not sure we want it now 🤔 Looks like the API doesn't change dramatically. Is it possible for us to try something backward-compatible?

FrankYang0529: Removed this change and using `azure-mgmt-datafactory` version to determine kwargs. Thanks.

Lee-W (approved):

Lee-W:
@eladkal I like this, but I'm not sure we've done something like this before. Increasing the minimum supported version from 2 to 10 seems a bit aggressive, so I suggested keeping it backward compatible.

eladkal (dismissed):
Temporary block till I take a closer look

eladkal:
OK, I reviewed the case and I think it's OK to support `azure-mgmt-datafactory>=10` only.
I rather avoid overcomplicating the code unless we have clear indication that it's worth it.

The `azure-mgmt-datafactory` releases are not that frequent which is why we kept supporting old versions of it. Users of older versions can just use one version behind of the provider - should be perfectly fine. 

@FrankYang0529 can you please adjust the code for that?

FrankYang0529:
Thanks for the review. I will update it today.

FrankYang0529:
Updated it. Thank you.

jason810496 (approved):

jason810496:
> OK, I reviewed the case and I think it's OK to support azure-mgmt-datafactory>=10 only.

Merging then, thanks.

PR #69807, opened 2026-07-13
Detect Celery workers not consuming queues

### LINKED ISSUE #63580: CeleryExecutor worker health check does not detect catatonic state after Redis broker restart
### Apache Airflow version

3.1.7

_No response_

### What happened?

After a Redis broker restart (e.g., VM reboot), the Celery worker reconnects at the transport level but silently loses its consumer registration on the task queue. The worker process stays alive, `celery inspect ping` returns OK, but `inspect.active_queues()` returns `None` — the worker is in a **catatonic state** where it accepts no new tasks.

This is a known upstream Celery bug (celery/celery#8030, celery/celery#9054, celery/celery#8990) that has persisted across Celery 5.2.x through 5.5.x with Redis broker. The partial fix in celery/celery#8796 did not fully resolve it.

**The problem on Airflow's side**: The current worker liveness/health check mechanism does not detect this state. Since `celery inspect ping` responds normally, Docker/Kubernetes health probes pass, and the worker is never restarted. Tasks accumulate in the Redis queue with state `queued`, hit the scheduler's requeue limit, and are marked `failed`.

In our case, 302 tasks piled up in the `default` queue over ~14 hours before we noticed. 7 DAGs failed, all with `Task requeue attempts exceeded max; marking failed`.

The Airflow Celery worker health check should verify that the worker has active queue consumers, not just that it responds to ping. Specifically:

```python
# Current check (insufficient):
result = app.control.inspect().ping()
# Returns OK even in catatonic state

# Proposed additional check:
queues = app.control.inspect().active_queues()
if queues is None or worker_name not in queues:
    # Worker is alive but not consuming — health check should FAIL
    return False
```

If the worker is alive but has no registered queues, the health check should fail, triggering a container restart via the orchestrator (Docker, Kubernetes, systemd).

This would go in the `airflow-providers-celery` package, likely in the worker CLI health check logic.

### How to reproduce

1. Start Airflow with CeleryExecutor and Redis broker
2. Confirm worker is consuming tasks normally
3. Restart Redis (e.g., `docker restart redis` or reboot the Redis host)
4. Observe: worker process stays alive, `celery inspect ping` returns OK
5. `celery inspect active_queues` returns `None` or empty for the worker
6. Schedule a DAG — task goes to `queued` state and is never picked up
7. After scheduler requeue attempts, task is marked `failed`

### Operating System

Debian 12 (Proxmox VM)

apache-airflow-providers-celery (installed with Airflow 3.1.7)

### Deployment

Docker Compose

### Deployment details

- Airflow 3.1.7 (Docker image `apache/airflow:3.1.7-python3.11`)
- Celery worker with 16 fork pool workers
- Redis 7.2.10 as broker
- PostgreSQL as result backend
- `worker_prefetch_multiplier = 1`

### Anything else?

**Related upstream Celery issues:**
- celery/celery#8030
- celery/celery#8091
- celery/celery#8990 (confirms not fixed in 5.4.0)
- celery/celery#9054
- celery/celery#9191

**Previous Airflow issues (closed as upstream):**
- #26542
- #32484
- #24498
- #27032

Those issues were rightfully closed as upstream Celery bugs. This issue proposes a **defensive fix on Airflow's side** — improving the health check to detect and recover from the catatonic state, regardless of when Celery fixes the root cause.

**Workarounds:**
- `--without-heartbeat --without-gossip --without-mingle` avoids the code path but loses cluster features
- `broker_connection_retry = False` makes the worker crash on broker loss (requires restart policy)
- Custom health check script checking `active_queues()` instead of `ping`

I'm willing to work on a PR for this.

shaealh:
closes: #63580

## Problem

After a Redis broker restart, a Celery worker can remain alive and continue responding to `ping` while no longer consuming tasks from any queue. The existing Docker Compose and Helm liveness probes therefore remain healthy while queued tasks accumulate.

## Solution

- Add `airflow celery check-worker`, targeted to a specific worker hostname.
- Fail the check when the worker does not reply, is absent from the response, or reports no active queues.
- Use the new command in the default Docker Compose and Helm worker liveness checks.
- Fall back to the existing Celery `inspect ping` command when an older Celery provider does not recognize `check-worker`.
- Update the Celery worker health-check documentation.

## Testing

- 27 focused Celery provider CLI tests passed, covering healthy workers, no response, an empty response, a different responding worker, and a worker with no queues.
- Ruff formatting and lint checks passed.
- YAML lint and Helm chart lint passed.
- Relevant pre-commit hooks passed.
- The local kubeconform hook could not run because the Breeze shim is not installed.
- The focused local Helm rendering test could not run because the `helm` executable is not installed.

---

- [x] Yes — OpenAI Codex (GPT-5)

### DIFF
--- a/airflow-core/docs/administration-and-deployment/logging-monitoring/check-health.rst
+++ b/airflow-core/docs/administration-and-deployment/logging-monitoring/check-health.rst
@@ -143,8 +143,8 @@ For details about installation, see: :doc:`apache-airflow-providers-celery:celer
 CLI Check for Celery Workers
 ----------------------------
 
-To verify that the Celery workers are working correctly, you can use the ``celery inspect ping`` command. On failure, the command will exit
-with a non-zero error code.
+To verify that a Celery worker is running and consuming from at least one queue, use the
+``airflow celery check-worker`` command. On failure, the command exits with a non-zero error code.
 
 .. note::
 
@@ -155,12 +155,12 @@ To check if the worker running on the local host is working correctly, run:
 
 .. code-block:: bash
 
-    celery --app airflow.providers.celery.executors.celery_executor.app inspect ping -d celery@${HOSTNAME}
+    airflow celery check-worker --celery-hostname celery@${HOSTNAME}
 
-To check if the all workers in the cluster running is working correctly, run:
+To list all workers and their active queues, run:
 
 .. code-block:: bash
 
-    celery --app airflow.providers.celery.executors.celery_executor.app inspect ping
+    airflow celery list-workers
 
 For more information, see: `Management Command-line Utilities (inspect/control) <https://docs.celeryproject.org/en/stable/userguide/monitoring.html#monitoring-control>`__ and `Workers Guide <https://docs.celeryproject.org/en/stable/userguide/workers.html>`__ in the Celery documentation.
--- a/airflow-core/docs/howto/docker-compose/docker-compose.yaml
+++ b/airflow-core/docs/howto/docker-compose/docker-compose.yaml
@@ -171,7 +171,7 @@ services:
     command: celery worker
     healthcheck:
       # yamllint disable rule:line-length
-      test: ["CMD-SHELL", 'celery --app airflow.providers.celery.executors.celery_executor.app inspect ping -d "celery@$${HOSTNAME}" || celery --app airflow.executors.celery_executor.app inspect ping -d "celery@$${HOSTNAME}"']
+      test: ["CMD-SHELL", 'airflow celery check-worker -H "celery@$${HOSTNAME}" || { status=$$?; [ "$$status" -eq 2 ] && exec celery --app airflow.providers.celery.executors.celery_executor.app inspect ping -d "celery@$${HOSTNAME}"; exit "$$status"; }']
       interval: 30s
       timeout: 10s
       retries: 5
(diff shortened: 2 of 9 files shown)


Miretpl:
Isn't this issue already fixed in one of the latest releases of Celery?

eladkal:
> Isn't this issue already fixed in one of the latest releases of Celery?

It is.
The Github issue says

> Those issues were rightfully closed as upstream Celery bugs. This issue proposes a defensive fix on Airflow's side — improving the health check to detect and recover from the catatonic state, regardless of when Celery fixes the root cause.

I think it waited for helm experts to share their thoughts if this is something we want or not. I tend to say no but @jedcunningham @Miretpl I guess it's up to you

Miretpl:
> It is.
The Github issue says

I was on my phone and didn't take a deeper look at the issue. Thanks!

> I think it waited for helm experts to share their thoughts if this is something we want or not. I tend to say no but @jedcunningham @Miretpl I guess it's up to you

As of now, I would say no as a result backend (Airflow database), and the queue are separated in the Helm implementation (and probably in most Airflow deployments, I would say). Even if there is no queue instance, tasks may continue proper execution without interruption if the result backend is working correctly and if Celery handles the restarted queue; everything should be good. If we kill the Celery pod via probes when the queue has temporary issues, that will mean killing tasks during execution too, which, depending on the case, might be a disruptive operation which should not happen during temporary issues.

I will take a deeper look at the issue itself later this week. There is always a chance that I miss something 😄

ashb:
This does nothing to address the linked issue.

Catatonic worker is "celery thinks it's listening to a queue, but it's connection is in a bad state".

PR #69809, opened 2026-07-13
Add optional debug spans for the scheduler loop

xBis7:
This patch is adding a trace for every iteration of the scheduler loop and debug spans around all the major steps and queries.

## Context

There is a config option for exporting debug traces. 

https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/config_templates/config.yml#L1504-L1513

We used to export random spans for multiple methods of the scheduler but these spans were never grouped together under a common trace and also didn't have any meaningful attributes. They were providing little to no benefit while they were actually adding more noise. Eventually, all such spans were removed and the config option was deprecated as unused but kept around in case we wanted to try a different approach in the future.

So far we only have traces for dag_runs but there are some operations across the project that are worth tracking with spans. The spans could help us pin-point the root cause of performance issues and come up with appropriate solutions.

There have been a lot of discussions around scheduler performance, task starvation, etc. with no solid way of consistently reproducing the issues or identifying the exact pain points. The spans added in this patch could be a good 1st step towards improving the scheduler.

In the future, we could also add debug spans for the dag_processor or the triggerer or operations in the task-sdk.

## Current changes + examples

I've added a new function with a `start_as_current_span` call guarded by the debug traces flag so that we can enable and disable the new traces (disabled by default).

The parent span for each iteration always starts as a root span and it has an attribute `airflow.scheduler.loop_iteration.idle` which can help us distinguish the iterations that actually did some work. We can use it to filter the spans and find the interesting ones.

image

Then we can sort by the longest first and start examining them

image

Not all iteration traces look the same. In the above example, this is the longest

image

image

2nd longest

image

image

3rd longest

image

image

And we are also exposing various internal values. I've only added attributes that are available at the moment of the span and avoided doing any calculations that will add more load to the system.

image

image

## Testing

Apart from manual testing, I've added unit tests for the new `start_debug_span` function and an integration test that checks for certain spans that we can expect to consistently appear in every iteration.

---

Claude-code Opus 4.8

---

### DIFF
--- a/airflow-core/src/airflow/settings.py
+++ b/airflow-core/src/airflow/settings.py
@@ -816,6 +816,13 @@ def initialize():
     import_local_settings()
     configure_logging()
     configure_otel(conf)
+    # configure_otel configures only the airflow-core copy of the symlinked shared traces
+    # module. Set the resolved flag in the task-sdk copy that providers import.
+    from airflow._shared.observability import traces as _core_traces
+    from airflow.sdk._shared.observability.traces import set_debug_traces_enabled
+
+    debug_traces_on = _core_traces._otel_debug_traces_on
+    set_debug_traces_enabled(debug_traces_on)
     configure_adapters()
     # The webservers import this file from models.py with the default settings.
 
--- a/airflow-core/src/airflow/settings.py
+++ b/airflow-core/src/airflow/settings.py
@@ -815,6 +815,13 @@ def initialize():
     import_local_settings()
     configure_logging()
     configure_otel(conf)
+    # configure_otel configures only the airflow-core copy of the symlinked shared traces
+    # module. Set the resolved flag in the task-sdk copy that providers import.
+    from airflow._shared.observability import traces as _core_traces
+    from airflow.sdk._shared.observability.traces import set_debug_traces_enabled
+
+    debug_traces_on = _core_traces._otel_debug_traces_on
+    set_debug_traces_enabled(debug_traces_on)
     configure_adapters()
     # The webservers import this file from models.py with the default settings.
 
--- a/shared/observability/src/airflow_shared/observability/traces/__init__.py
+++ b/shared/observability/src/airflow_shared/observability/traces/__init__.py
@@ -29,13 +29,32 @@ from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor, SpanExporter
 from opentelemetry.sdk.trace.id_generator import RandomIdGenerator
 from opentelemetry.sdk.trace.sampling import Decision
-from opentelemetry.trace import NonRecordingSpan, Span, SpanContext, TraceFlags, TraceState
+from opentelemetry.trace import INVALID_SPAN, NonRecordingSpan, Span, SpanContext, TraceFlags, TraceState
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
     from configparser import ConfigParser
 log = logging.getLogger(__name__)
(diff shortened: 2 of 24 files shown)


xBis7:
I've been running load tests using the debug spans from this PR. Based on the investigation so far, I'm planning to add some more spans to this patch. I'll push all the changes once I'm confident that there isn't any value in monitoring deeper levels of each iteration.

I'm going to convert this to a draft.

xBis7:
I've added some extra spans that provided valuable information during my testing and I've re-opened this PR.

Miretpl (commented):
To my current knowledge, changes look mostly ok. Having traces in the scheduler and different components would make several things easier and possible, as we could base our decisions on actual data.

I'm not sure we can skip introducing the new `task-sdk` import in core.

Have you done any performance testing on what is the potential impact of these changes and debug mode enabling?

> In the future, we could also add debug spans for the dag_processor or the triggerer or operations in the task-sdk.

Probably we would want to have an opt-in option list-like `scheduler, dag processor` that would enable debug only for scheduler and dag processor components.

xBis7:
@Miretpl Thank you for taking a look at this PR!

> Have you done any performance testing on what is the potential impact of these changes and debug mode enabling?

Yes, I have and and the difference is negligible.

I don't expect people to have the spans turned on all the time, but to enable them after a performance incident so that they can investigate.

> Probably we would want to have an opt-in option list-like scheduler, dag processor that would enable debug only for scheduler and dag processor components.

Yes, a filter list is a good idea if we add spans for mulitple components.

Miretpl:
> I don't expect people to have the spans turned on all the time, but to enable them after a performance incident so that they can investigate.

Same. It is always good to know that potential debug mode will not degrade performance to the level that the environment will not meet certain performance criteria, especially if that is a production environment.

ashb:
@dstandish Please can you weigh in on this one?

airflow-core/src/airflow/settings.py
```diff
@@ -841,6 +841,13 @@ def initialize():
     import_local_settings()
     configure_logging()
     configure_otel(conf)
+    # configure_otel configures only the airflow-core copy of the symlinked shared traces
+    # module. Set the resolved flag in the task-sdk copy that providers import.
+    from airflow._shared.observability import traces as _core_traces
+    from airflow.sdk._shared.observability.traces import set_debug_traces_enabled
+
+    debug_traces_on = _core_traces._otel_debug_traces_on
+    set_debug_traces_enabled(debug_traces_on)
```
manish1337: Could `start_debug_span` just read the flag from `conf` lazily instead? Then this block goes away. Otherwise maybe route it through `common.compat`, which would also dedupe the 5 fallbacks in the executors.

shared/observability/src/airflow_shared/observability/traces/__init__.py
```diff
@@ -248,11 +267,21 @@ def _load_exporter_from_env() -> SpanExporter:
...
+
+
 def configure_otel(conf: ConfigParser):
+    global _otel_debug_traces_on
+
     otel_on = conf.getboolean("traces", "otel_on", fallback=False)
     if not otel_on:
         return
 
+    _otel_debug_traces_on = conf.getboolean("traces", "otel_debug_traces_on", fallback=False)
```
manish1337: The tests monkeypatch `_otel_debug_traces_on` directly, so nothing covers `configure_otel` reading the flag (incl. the `otel_on=False` case), `set_debug_traces_enabled`, or the `settings.py` sync. Dropping the sync would silently lose executor spans with all tests green. Can we add a couple of tests here?

manish1337 (commented):
Nice work on the spans, a couple of thoughts on the flag plumbing

PR #53153, opened 2025-07-10
Refactor SchedularDagBag and use it across the API server

ephraimbuddy:
This removes the use of DagBag in API server which parses files in some cases and
replaces it with refactored SchedulerDagBag.

PR #69815, opened 2026-07-13
Speed up repeated latest Dag lookups

### LINKED ISSUE #69814: Latest Dag lookups repeatedly deserialize cached versions

Airflow Core

### Apache Airflow version

`main` (`3.4.0`)

The API server's `DBDagBag.get_latest_version_of_dag` queries the latest serialized row and deserializes its payload on every lookup. This happens even when the configured LRU/TTL cache already contains the same Dag version and serialized hash.

The latest-row query is required to preserve freshness, but repeating `DagSerialization.from_dict()` for unchanged data bypasses the expensive part of the cache.

Steps to reproduce:

1. Enable the API server Dag cache with a positive `[api] dag_cache_size`.
2. Store a serialized Dag and create a process-lived `DBDagBag`.
3. Request the latest Dag by `dag_id` twice, for example through a route that calls `get_latest_version_of_dag`.
4. Observe that both requests load the latest serialized row and deserialize its payload, even though the second request resolves to the cached version and hash.

The overhead grows with the serialized payload. In a real-path benchmark covering the ORM query, Dag deserialization, and task route handler, an approximately 31 KB payload took 3.1–4.1 ms per request across SQLite, PostgreSQL, and MySQL. An approximately 305 KB payload took 28.8–30.4 ms.

The API server should continue querying the latest serialized row, then reuse the cached deserialized Dag when both `dag_version_id` and `dag_hash` are unchanged. If the hash changed for the same version ID, it should discard the stale cache entry and deserialize the current row.

This preserves freshness while avoiding repeated deserialization. The same benchmark measured 0.4–1.0 ms for the approximately 31 KB payload and 1.8–3.9 ms for the approximately 305 KB payload with this behavior.

### Operating System

Reproduced in the Apache Airflow Breeze development environment on macOS.

### Deployment

Other Docker-based deployment

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

viiccwen:
closes #69814 

Repeated API requests that resolve the latest Dag always need to query the latest serialized row to preserve freshness. However, they also deserialized that row on every request, even when the same Dag version and hash were already cached.

This change reuses the cached deserialized Dag when the queried version ID and hash are unchanged.
It keeps the existing latest-row query, reloads serialized Dags that were updated in place, and refreshes the cache validation timestamp. The cache documentation now distinguishes freshness revalidation from LRU/TTL eviction.

A real-path benchmark measured the ORM latest-row query, Dag deserialization, and task route handler. Each result is the median of three rounds with 50 lookups per round.

| Backend | Serialized payload | Before | After | Speedup |
| --- | ---: | ---: | ---: | ---: |
| SQLite | 31 KB | 3.052 ms | 0.437 ms | 6.98x |
| SQLite | 153 KB | 13.487 ms | 0.965 ms | 13.98x |
| SQLite | 305 KB | 28.765 ms | 1.808 ms | 15.91x |
| PostgreSQL 14 | 31 KB | 3.798 ms | 1.002 ms | 3.79x |
| PostgreSQL 14 | 153 KB | 16.461 ms | 2.093 ms | 7.87x |
| PostgreSQL 14 | 305 KB | 30.390 ms | 3.912 ms | 7.77x |
| MySQL 8 | 31 KB | 4.139 ms | 1.033 ms | 4.01x |
| MySQL 8 | 153 KB | 15.952 ms | 1.933 ms | 8.25x |
| MySQL 8 | 305 KB | 29.914 ms | 3.371 ms | 8.87x |

The benchmark uses a new SQLAlchemy session for each lookup and leaves `[core] compress_serialized_dags` at its default value (`False`). It excludes HTTP middleware, authentication, and response encoding, so the table isolates the route-handler path affected by this change rather than claiming the same percentage improvement for total HTTP latency.

Tests cover matching cache hits, validation timestamp refresh, cache metrics, cache misses, and in-place hash changes. The DagBag test file, related API/query-count tests, Ruff checks, pre-commit checks, manual checks, and Apache Airflow documentation build pass.

---

- [X] Yes — Codex (GPT-5.6 Sol)

---

### DIFF
--- a/airflow-core/src/airflow/models/dagbag.py
+++ b/airflow-core/src/airflow/models/dagbag.py
@@ -240,6 +240,24 @@ class DBDagBag:
 
         if not (serdag := SerializedDagModel.get(dag_id, session=session)):
             return None
+
+        with self._lock:
+            cached = self._dags.get(serdag.dag_version_id)
+            if cached is not None:
+                # The latest-row query already loaded dag_hash, so no revalidation query is needed.
+                if cached.dag_hash == serdag.dag_hash:
+                    self._dags[serdag.dag_version_id] = cached._replace(last_validated=time.monotonic())
+                else:
+                    # A serialized Dag can be updated in place without changing its version ID.
+                    self._dags.pop(serdag.dag_version_id, None)
+                    cached = None
+
+        if cached is not None:
+            if self._use_cache:
+                stats.incr("api_server.dag_bag.cache_hit")
+            return cached.dag
+        if self._use_cache:
+            stats.incr("api_server.dag_bag.cache_miss")
         return self._read_dag(serdag)
 
 
--- a/airflow-core/docs/faq.rst
+++ b/airflow-core/docs/faq.rst
@@ -728,8 +728,9 @@ in memory. Configure this in the ``[api]`` section:
     dag_cache_size = 64    ; max cached versions (0 = unbounded, pre-3.2 behavior)
     dag_cache_ttl = 3600   ; seconds before a cached entry expires (0 = LRU only)
 
-The cache is keyed by Dag version ID. After a Dag is updated, the API server may serve the
-previous version until the cached entry expires (controlled by ``dag_cache_ttl``).
+The cache is keyed by Dag version ID. Cached entries are periodically checked against the
+database according to ``[core] min_serialized_dag_update_interval``. The ``dag_cache_ttl`` setting
+controls eviction and memory usage independently of this freshness check.
 
 See :ref:`config:api__dag_cache_size` and :ref:`config:api__dag_cache_ttl` for the full
 configuration reference.
--- a/airflow-core/src/airflow/config_templates/config.yml
+++ b/airflow-core/src/airflow/config_templates/config.yml
@@ -1670,7 +1670,7 @@ api:
         Set to 0 to use an unbounded dict (no eviction, matching pre-3.2 behavior).
         The cache is keyed by Dag version ID. Lookups by Dag ID
         (e.g., viewing a Dag's details) fetch the latest serialized row, but reuse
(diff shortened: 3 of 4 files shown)


viiccwen:
cc @guan404ming, @jason810496. 🙌

airflow-core/src/airflow/models/dagbag.py
```diff
@@ -240,6 +240,24 @@ def get_latest_version_of_dag(self, dag_id: str, *, session: Session) -> Seriali
...
+                    self._dags.pop(serdag.dag_version_id, None)
+                    cached = None
+
+        if cached is not None:
+            if self._use_cache:
+                stats.incr("api_server.dag_bag.cache_hit")
+            return cached.dag
+        if self._use_cache:
+            stats.incr("api_server.dag_bag.cache_miss")
         return self._read_dag(serdag)
```
jason810496, line 261: @ephraimbuddy 

Are we intentionally to retrieve the latest Dag instance from Metadata DB without cache for the `get_latest_version_of_dag` method introduced in https://github.com/apache/airflow/pull/53153 ?

There's a similar method -- `get_dag` that always read from the cache first.

https://github.com/apache/airflow/blob/50f90d17ddab0f39d632917da722c9d4bdaa9307/airflow-core/src/airflow/models/dagbag.py#L177-L179

Would like to double check the usage and the intention back then, thanks.

jason810496 (commented):
I'd prefer to wait until @ephraimbuddy confirm on this case.

airflow-core/src/airflow/models/dagbag.py
viiccwen: Hello @ephraimbuddy, could you help to check this?

PR #67056, opened 2026-05-17
Decouple remote logging config from core

jason810496:
- related: https://github.com/apache/airflow/issues/65281
- related: https://github.com/apache/airflow/pull/66994

### Why

Decouple remote logging config from hardcoded `airflow-core/src/airflow/config_templates/airflow_local_settings.py` by introducing new `remote-logging` field for `ProvidersManager`. Add shared `resolve_remote_task_log` it, so core and the Task SDK
share one implementation of the precedence rule.

### What

Single shared resolver `resolve_remote_task_log` with an explicit three-tier precedence:
1. User-defined `[logging] logging_config_class` module exporting
   `REMOTE_TASK_LOG` / `DEFAULT_REMOTE_CONN_ID`.
2. ProvidersManager scheme dispatch on `[logging] remote_base_log_folder`,
   instantiated via the provider class's `from_config()` classmethod.
3. Legacy attr-path fallback against
   `airflow.config_templates.airflow_local_settings` (transitional).

`airflow.logging_config.load_logging_config` is deprecated and split into
  - `_get_logging_config()` (logging dict)
  - `_load_logging_config()` (remote handler, lazy)
  - `configure_logging` no longer eagerly resolves the remote handler.

### Next step to deprecate hardcoded `airflow_local_settings.py`

1. Provider migration: Migrate the logic from `airflow_local_settings.py` as `from_config` method for each current supported `RemoteIO`.

```python
class S3RemoteLogIO(LoggingMixin):
    @classmethod
    def from_config(cls) -> "S3RemoteLogIO":
        from airflow.providers.common.compat.sdk import conf

        # provider-specific `conf` validation or custom logic
        return cls(
            base_log_folder=conf.get("logging", "base_log_folder"),
            remote_base=conf.get("logging", "remote_base_log_folder"),
            delete_local_copy=conf.getboolean("logging", "delete_local_logs"),
        )
```

2. Raise deprecation warning for each if/else branch after the corresponding `RemoteIO.from_config` is added.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes, with help of Claude Code (Opus 4.7)  following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #69816, opened 2026-07-13
Add CloudWatchRemoteLogIO.from_config and register cloudwatch scheme

jason810496:
- related: #67056

## Why

#67056 decoupled remote logging from the hardcoded branches in `airflow_local_settings.py`: core and the Task SDK now resolve the handler via ProvidersManager dispatch on the `[logging] remote_base_log_folder` URL scheme, instantiating the provider class through a no-arg `from_config()` classmethod. This migrates the `cloudwatch` scheme.

## What

- Add `CloudWatchRemoteLogIO.from_config()`, mirroring the legacy `airflow_local_settings.py` branch -- parsing `log_group_arn` from the URL (with a clear `ValueError` when no ARN can be derived), merging `[logging] remote_task_handler_kwargs` IO-kwargs, and applying `expanduser` on `base_log_folder`, so behavior is unchanged for existing configs.
- Register the`cloudwatch` scheme in the amazon get_provider_info.py.

## Note

A sibling PR migrates the `s3` scheme the same way (https://github.com/apache/airflow/pull/69817); whichever merges second needs a trivial rebase of the new `remote-logging:` section. If `from_config` raises on a bad config, the shared factory falls back to the legacy path, so this is not a breaking change.

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import contextlib
 import copy
+import inspect
 import json
 import logging
 import os
@@ -28,6 +29,7 @@ from datetime import date, datetime, timedelta, timezone
 from functools import cached_property
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
+from urllib.parse import urlsplit
 
 import attrs
 import watchtower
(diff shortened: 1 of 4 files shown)


providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
```diff
@@ -105,6 +107,40 @@ def _(self):
...
+        """Build the remote log IO from Airflow logging configuration."""
+        remote_task_handler_kwargs = conf.getjson("logging", "remote_task_handler_kwargs", fallback={})
+        if not isinstance(remote_task_handler_kwargs, dict):
+            raise ValueError(
+                "logging/remote_task_handler_kwargs must be a JSON object (a python dict), we got "
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # remote_task_handler_kwargs mixes FileTaskHandler kwargs with IO kwargs; only the
+        # latter belong to this class (same split as airflow_local_settings.py).
+        fth_params = frozenset(inspect.signature(FileTaskHandler.__init__).parameters) - {
```
ferruzzi, line 121: Same thought as the other related PR, if you change that one you should also update this one.  Otherwise LGTM.

jason810496: I'd like to keep it as-is due to the reason described in - https://github.com/apache/airflow/pull/69817#discussion_r3619328796.

ferruzzi (approved):
LGTM

PR #69817, opened 2026-07-13
Add S3RemoteLogIO.from_config and register s3 remote logging scheme

jason810496:
- related: #67056

## Why

#67056 decoupled remote logging from the hardcoded branches in `airflow_local_settings.py`: core and the Task SDK now resolve the handler via ProvidersManager dispatch on the `[logging] remote_base_log_folder` URL scheme, instantiating the provider class through a no-arg `from_config()` classmethod. This migrates the `s3` scheme as the first adopter.

## What

- Add `S3RemoteLogIO.from_config()`, mirroring the legacy `airflow_local_settings.py` branch -- including the `[logging] remote_task_handler_kwargs` IO-kwargs merge and `expanduser` on `base_log_folder`, so behavior is unchanged for existing configs.
- Register the `s3` scheme in the amazon `get_provider_info.py`.

## Note

A sibling PR migrates the `cloudwatch` scheme the same way (https://github.com/apache/airflow/pull/69816); whichever merges second needs a trivial rebase of the new `remote-logging:` section. If `from_config` raises on a bad config, the shared factory falls back to the legacy path, so this is not a breaking change.

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/log/s3_task_handler.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/log/s3_task_handler.py
@@ -17,6 +17,7 @@
 # under the License.
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 import pathlib
@@ -45,6 +46,31 @@ class S3RemoteLogIO(LoggingMixin):  # noqa: D101
     delete_local_copy: bool
 
     processors = ()
+
+    @classmethod
+    def from_config(cls) -> S3RemoteLogIO:
+        """Build the remote log IO from Airflow logging configuration."""
+        remote_task_handler_kwargs = conf.getjson("logging", "remote_task_handler_kwargs", fallback={})
+        if not isinstance(remote_task_handler_kwargs, dict):
+            raise ValueError(
+                "logging/remote_task_handler_kwargs must be a JSON object (a python dict), we got "
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # remote_task_handler_kwargs mixes FileTaskHandler kwargs with IO kwargs; only the
+        # latter belong to this class (same split as airflow_local_settings.py).
+        fth_params = frozenset(inspect.signature(FileTaskHandler.__init__).parameters) - {
+            "self",
+            "base_log_folder",
+        }
+        io_kwargs = {k: v for k, v in remote_task_handler_kwargs.items() if k not in fth_params}
+        return cls(
+            **{
+                "base_log_folder": os.path.expanduser(conf.get_mandatory_value("logging", "base_log_folder")),
+                "remote_base": conf.get_mandatory_value("logging", "remote_base_log_folder"),
+                "delete_local_copy": conf.getboolean("logging", "delete_local_logs"),
+            }
+            | io_kwargs,
+        )
 
     def upload(self, path: os.PathLike | str, ti: RuntimeTI | None = None) -> None:
         """Upload the given log path to the remote storage."""
--- a/providers/amazon/tests/unit/amazon/aws/log/test_s3_task_handler.py
+++ b/providers/amazon/tests/unit/amazon/aws/log/test_s3_task_handler.py
@@ -21,6 +21,7 @@ import contextlib
 import copy
 import logging
 import os
+import pathlib
 from unittest import mock
 
 import boto3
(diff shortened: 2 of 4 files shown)


providers/amazon/src/airflow/providers/amazon/aws/log/s3_task_handler.py
```diff
@@ -46,6 +47,31 @@ class S3RemoteLogIO(LoggingMixin):  # noqa: D101
...
+            raise ValueError(
+                "logging/remote_task_handler_kwargs must be a JSON object (a python dict), we got "
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # remote_task_handler_kwargs mixes FileTaskHandler kwargs with IO kwargs; only the
+        # latter belong to this class (same split as airflow_local_settings.py).
+        fth_params = frozenset(inspect.signature(FileTaskHandler.__init__).parameters) - {
+            "self",
+            "base_log_folder",
+        }
```
jason810496, line 64: I intentionally duplicate the logic from core side:

https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L157-L170

providers/amazon/src/airflow/providers/amazon/aws/log/s3_task_handler.py
```diff
@@ -46,6 +47,31 @@ class S3RemoteLogIO(LoggingMixin):  # noqa: D101
...
+        io_kwargs = {k: v for k, v in remote_task_handler_kwargs.items() if k not in fth_params}
+        return cls(
+            **{
+                "base_log_folder": os.path.expanduser(conf.get_mandatory_value("logging", "base_log_folder")),
+                "remote_base": conf.get_mandatory_value("logging", "remote_base_log_folder"),
+                "delete_local_copy": conf.getboolean("logging", "delete_local_logs"),
+            }
+            | io_kwargs,
+        )
```
jason810496, line 73: The rest of the S3 specific kwargs are coming from:

https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L173-L187

providers/amazon/tests/unit/amazon/aws/log/test_s3_task_handler.py
```diff
@@ -52,6 +53,86 @@ def s3mock():
...
+        ],
+    )
+    @conf_vars(
+        {
+            ("logging", "remote_logging"): "True",
+            ("logging", "remote_base_log_folder"): "s3://bucket/remote/log/location",
+            ("logging", "remote_log_conn_id"): "aws_default",
+        }
+    )
+    def test_resolve_remote_task_log_uses_provider_dispatch_not_local_settings(self, manager_classpath):
```
jason810496, line 118: This test is exercising the following remote logging discovery that we should go through the `_build_remote_task_log_from_provider` path.

https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/shared/logging/src/airflow_shared/logging/factory.py#L134-L182

jason810496 (commented):
Here're more context about the changes itself.

providers/amazon/src/airflow/providers/amazon/aws/log/s3_task_handler.py
```diff
@@ -46,6 +47,31 @@ class S3RemoteLogIO(LoggingMixin):  # noqa: D101
...
+        """Build the remote log IO from Airflow logging configuration."""
+        remote_task_handler_kwargs = conf.getjson("logging", "remote_task_handler_kwargs", fallback={})
+        if not isinstance(remote_task_handler_kwargs, dict):
+            raise ValueError(
+                "logging/remote_task_handler_kwargs must be a JSON object (a python dict), we got "
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # remote_task_handler_kwargs mixes FileTaskHandler kwargs with IO kwargs; only the
+        # latter belong to this class (same split as airflow_local_settings.py).
+        fth_params = frozenset(inspect.signature(FileTaskHandler.__init__).parameters) - {
```
ferruzzi, line 61: Is it maybe worth caching `inspect.signature(FileTaskHandler.__init__)`?  It's a small call but should always return the same result, right?

jason810496: Thanks for the review.

From my perspective, it should be fine to keep it as-is. Since the remote logging IO constructor (`from_config` class method) will only be called once in the whole process lifecycle and we cache the remote logging IO instance at module level.

https://github.com/apache/airflow/blob/ec08de79ab6367a464f9af5f33d43025ec9c8365/airflow-core/src/airflow/logging_config.py#L91-L101

ferruzzi: Ah, yeah, you're right, sorry.

ferruzzi (approved):
LGTM

jason810496:
Thanks for the review!

PR #69821, opened 2026-07-13
Respect retries when a deferrable trigger ends a task with TaskFailedEvent

### LINKED ISSUE #69819: Deferrable task that ends via TaskFailedEvent ignores retries and always runs on_failure_callback
### What happened

When a deferrable operator's trigger yields a terminal `TaskFailedEvent` (a `BaseTaskEndEvent` with `task_instance_state=FAILED`) to fail a task without resuming it on a worker, the task is always marked `failed` terminally and its `on_failure_callback` runs, even when the task still has retries remaining.

A task that fails while running on a worker with retries left goes to `up_for_retry` and runs `on_retry_callback`. A task that fails through its trigger does not. Deferrable tasks that fail via the trigger silently lose their retries and emit a false `on_failure_callback`.

A trigger-emitted failure should respect the task's retry configuration, the same way a worker-side failure does: if the task is eligible to retry, go `up_for_retry` and run `on_retry_callback`; only when retries are exhausted, fail terminally and run `on_failure_callback`.

### How to reproduce

Defer a task with `retries=1` whose trigger yields `TaskFailedEvent()`. It goes straight to `failed` and `on_failure_callback` fires on the first attempt, instead of retrying.

The cause is the `BaseTaskEndEvent` handler in `airflow-core/src/airflow/models/trigger.py`, which calls `set_state(event.task_instance_state)` and sends the callback with `task_callback_type=event.task_instance_state` without any retry-eligibility check.

### Context

The scheduler executor-event path already routes by retry-eligibility (#56586) and the DAG processor honors `task_callback_type` (UP_FOR_RETRY -> on_retry_callback). The trigger end-event path was not updated to match.

Yes

hkc-8010:
## Summary

When a deferrable operator's trigger yields a terminal `TaskFailedEvent`, the task was always marked `failed` and its `on_failure_callback` ran, even when the task still had retries remaining. A worker-side failure with retries left instead goes `up_for_retry` and runs `on_retry_callback`. This makes trigger-driven failures behave the same way.

The `BaseTaskEndEvent` handler in `airflow-core/src/airflow/models/trigger.py` now, for a `FAILED` event, loads the serialized task and routes through `TaskInstance.handle_failure()`, the same path the scheduler executor-event handling has used since #56586. `handle_failure` sets `up_for_retry` or `failed` based on retry-eligibility and fires the `on_task_instance_failed` listener, failure metrics, and a Log entry. The callback request now carries the resolved `task_callback_type` so the DAG processor runs `on_retry_callback` vs `on_failure_callback` correctly. `TaskSuccessEvent` and `TaskSkippedEvent` are unchanged.

closes: #69819

## Changes

- `airflow-core/src/airflow/models/trigger.py`: route trigger-emitted failures through `handle_failure`, thread the retry-aware `task_callback_type` into the callback, and skip the xcom push on the retry path.
- `airflow-core/tests/unit/models/test_trigger.py`: new parametrized test covering the retry-eligible case (`up_for_retry` + `on_retry_callback` + `on_task_instance_failed` listener) and the exhausted case (`failed` + `on_failure_callback`); the existing end-event test now sets a realistic `start_date` (a deferred task has already run).
- `airflow-core/newsfragments/69821.bugfix.rst`.

## Testing

- `pytest airflow-core/tests/unit/models/test_trigger.py` (39 passed).
- Confirmed the new test fails on `main` (the task stays `failed` and the callback is routed as a failure) and passes with this change.

This change was written with AI assistance (Claude); the author reviewed the diff and reproduction.

### DIFF
--- a/airflow-core/src/airflow/models/trigger.py
+++ b/airflow-core/src/airflow/models/trigger.py
@@ -561,13 +561,42 @@ def _(event: BaseTaskEndEvent, *, task_instance: TaskInstance, session: Session)
     from airflow.callbacks.database_callback_sink import DatabaseCallbackSink
     from airflow.utils.state import TaskInstanceState
 
-    # Mark the task with terminal state and prevent it from resuming on worker
+    # Prevent the task from resuming on a worker.
     task_instance.trigger_id = None
-    task_instance.set_state(event.task_instance_state, session=session)
+
+    callback_type = event.task_instance_state
+    handle_via_failure = False
+
+    if event.task_instance_state == TaskInstanceState.FAILED:
+        # Load the serialized task: needed both for is_eligible_to_retry() and for
+        # handle_failure() (which asserts self.task / self.task.dag). Mirrors the
+        # scheduler executor-event path (see is_eligible_to_retry / PR #56586).
+        try:
+            from airflow.models.dagbag import DBDagBag
+
+            dag = DBDagBag().get_dag_for_run(dag_run=task_instance.dag_run, session=session)
+            if dag is not None:
+                task_instance.task = dag.get_task(task_instance.task_id)
+                handle_via_failure = True
+        except Exception:
+            log.exception(
+                "Could not load task for %s; failing terminally without retry routing", task_instance
+            )
+        if handle_via_failure and task_instance.is_eligible_to_retry():
+            callback_type = TaskInstanceState.UP_FOR_RETRY
+
+    # Persist trigger_id=None before handle_failure: fetch_handle_failure_context calls
+    # refresh_from_db(keep_local_changes=False), which would otherwise reload the old trigger_id.
+    if handle_via_failure:
+        session.flush()
 
     def _submit_callback_if_necessary() -> None:
-        """Submit a callback request if the task state is SUCCESS or FAILED."""
-        if event.task_instance_state in (TaskInstanceState.SUCCESS, TaskInstanceState.FAILED):
+        """Submit a callback request if the task state is SUCCESS, FAILED, or UP_FOR_RETRY."""
+        if callback_type in (
+            TaskInstanceState.SUCCESS,
+            TaskInstanceState.FAILED,
+            TaskInstanceState.UP_FOR_RETRY,
+        ):
             if task_instance.dag_model.relative_fileloc is None:
                 raise RuntimeError("relative_fileloc should not be None for a finished task")
             from airflow.models.dag_version import _resolve_version_data
@@ -591,7 +620,7 @@ def _(event: BaseTaskEndEvent, *, task_instance: TaskInstance, session: Session)
             request = TaskCallbackRequest(
                 filepath=task_instance.dag_model.relative_fileloc,
                 ti=task_instance,
-                task_callback_type=event.task_instance_state,
+                task_callback_type=callback_type,
                 bundle_name=bundle_name,
                 bundle_version=bundle_version,
                 version_data=version_data,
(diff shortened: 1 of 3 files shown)


airflow-core/src/airflow/models/trigger.py
```diff
@@ -604,10 +633,21 @@ def _submit_callback_if_necessary() -> None:
...
 
+    # Send the callback before handle_failure (mirrors the scheduler executor-event ordering):
+    # handle_failure -> save_to_db commits, which also persists the DatabaseCallbackSink row atomically.
     _submit_callback_if_necessary()
+
+    if handle_via_failure:
+        # Canonical failure handling: sets UP_FOR_RETRY/FAILED by retry-eligibility, fires the
+        # on_task_instance_failed listener + failure metrics + Log audit row, and clears
+        # next_method args. Mirrors the scheduler executor-event path (PR #56586).
+        task_instance.handle_failure(error="Task failed via trigger event", session=session)
```
kaxil: On the retry-eligible path this skips the try archival the worker-side retry performs: `fetch_handle_failure_context` only calls `prepare_db_for_next_try` when `ti.state == RUNNING` (taskinstance.py:1870), and here the TI is still `deferred`, so no `task_instance_history` row is written for try 1 and `ti.id` is not rotated. When the scheduler then makes the TI schedulable it bumps `try_number` to 2 (dagrun.py:2144), and `GET .../taskInstances/{task_id}/logs/1` 404s at log.py:157 because there is no live TI at try 1 and no history row to fall back on. Worth deciding whether this branch should archive the try itself, or route through the `submit_failure`/`__fail__` path that already picks up the worker-side machinery.

airflow-core/tests/unit/models/test_trigger.py
```diff
@@ -362,6 +366,63 @@ def test_submit_event_task_end_callback_includes_version_data(mock_send, session
...
+    """A trigger-emitted TaskFailedEvent should respect retry-eligibility: a deferred task with
+    retries remaining goes UP_FOR_RETRY (on_retry_callback), not straight to FAILED.
+
+    Failures are routed through ``TaskInstance.handle_failure`` (mirroring the scheduler
+    executor-event path), so the ``on_task_instance_failed`` listener fires in both cases.
+    """
+    from airflow.listeners.listener import get_listener_manager
+
+    listener_callback = MagicMock()
+    get_listener_manager().pm.hook.on_task_instance_failed = listener_callback
```
kaxil: Assigning onto `pm.hook.on_task_instance_failed` leaks past the `finally`: `clear()` only unregisters plugins (shared/listeners/src/airflow_shared/listeners/listener.py:121-124) and `get_listener_manager()` is `@cache`d, so the MagicMock stays on the shared hook relay for the rest of the pytest worker. A later test that registers a real listener then dies in pluggy's `_verify_hook` with `InvalidSpecError: Cannot spec a Mock object`. The `listener_manager` fixture in `tests_common/pytest_plugin.py` (or `monkeypatch.setattr`) restores the original hook.

hkc-8010: I removed the direct assignment onto `pm.hook.on_task_instance_failed` and reworked the regression test so it only asserts the behavior this trigger path actually guarantees: callback routing plus `task_instance_history` archival on retry. That avoids leaking a mocked hook into later tests.

airflow-core/src/airflow/models/trigger.py
hkc-8010: I changed the retry-eligible trigger-failure path to archive the finished try explicitly with `prepare_db_for_next_try()` before setting the TI to `UP_FOR_RETRY`, instead of routing through `handle_failure()` from `DEFERRED`. That preserves the prior-try history row and keeps log lookup working for try 1.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -604,10 +626,22 @@ def _submit_callback_if_necessary() -> None:
 
     def _push_xcoms_if_necessary() -> None:
         """Pushes XComs to the database if they are provided."""
-        if event.xcoms:
+        if event.xcoms and callback_type != TaskInstanceState.UP_FOR_RETRY:
             for key, value in event.xcoms.items():
                 task_instance.xcom_push(key=key, value=value)
 
+    # Send the callback before handle_failure (mirrors the scheduler executor-event ordering):
+    # the callback request should reflect the retry/terminal decision derived above.
```
amoghrajesh: This is factually wrong? There is no call to `handle_failure` anywhere?

amoghrajesh (commented):
Code looks ok, but I have a qn, seems to be spillover from past version of the PR

airflow-core/src/airflow/models/trigger.py
hkc-8010: You're right, that comment was stale from the earlier handle_failure version. I updated it to match the current flow here: we send the callback before mutating state so it reflects the derived retry vs terminal decision, then the code either archives the try and sets UP_FOR_RETRY or applies the terminal state directly.

eladkal (approved):

PR #69827, opened 2026-07-13
Java SDK: Clearer error when a task class cannot be instantiated

FrankYang0529:
## Why

- A `Task` class without a public no-argument constructor or a non-static inner class fails with a generic `Error executing task` plus a raw reflection trace, which steering Dag authors toward debugging `execute()` logic that never ran.

## How

- Split instantiation from execution in `TaskRunner.runTask`: instantiation failures now log `Cannot instantiate task class. A task class must be concrete and declare a public no-argument constructor` with a `taskClass` field.
- Removed the redundant `e.printStackTrace()`. The structured log record already carries the full trace.
- Documented the constructor requirement in the interface-API section of `java.rst`.

## Verification

- `./gradlew :sdk:test --tests "org.apache.airflow.sdk.execution.TaskTest"`
- `E2E_TEST_MODE=java_sdk uv run --project airflow-e2e-tests pytest "tests/airflow_e2e_tests/java_sdk_tests/test_java_sdk_dag.py::TestJavaSDKUninstantiableTask" -xvs`

 

---

- [X] Yes - Claude Code

---

### DIFF
--- a/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
+++ b/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
@@ -205,6 +205,17 @@ Interface-based API
 Implement the ``Task`` interface directly for full control over how tasks are registered and how XComs are
 read.  Each task is registered as a ``TaskDef`` on a ``DagDef``.
 
+The runner creates a fresh instance of the task class through reflection for every task-instance run,
+which puts four constraints on the class:
+
+* The task class itself must be ``public``.
+* It must be concrete: not abstract and not an interface.
+* It must declare a public no-argument constructor.
+* If nested inside another class, it must be a ``static`` nested class.
+
+A class that violates any of these fails at runtime with a ``Cannot instantiate task class`` error in the
+task log.
+
 .. code-block:: java
 
     import org.apache.airflow.sdk.*;
@@ -218,11 +229,21 @@ read.  Each task is registered as a ``TaskDef`` on a ``DagDef``.
       }
     }
 
-Register tasks manually in a ``BundleBuilder``:
+Register tasks manually in a ``BundleBuilder``. A task class can be top-level like ``FetchTask``, or
+nested ``static`` class like ``ProcessTask``:
 
 .. code-block:: java
 
     public class MyBundle implements BundleBuilder {
+      public static class ProcessTask implements Task {
+        @Override
+        public void execute(Context context, Client client) throws Exception {
+          var fetched = (String) client.getXCom("fetch");
+          // implement task logic
+          client.setXCom(fetched);
+        }
+      }
+
       @Override
       public Iterable<DagDef> getDags() {
         var dag = new DagDef("my_dag")
--- a/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
+++ b/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
@@ -205,6 +205,16 @@ Interface-based API
 Implement the ``Task`` interface directly for full control over how tasks are registered and how XComs are
 read.
 
+The runner creates a fresh instance of the task class through reflection for every task-instance run,
+which puts three constraints on the class:
+
+* It must be concrete: not abstract and not an interface.
+* It must declare a public no-argument constructor.
+* If nested inside another class, it must be a ``static`` nested class.
+
+A class that violates any of these fails at runtime with a ``Cannot instantiate task class`` error in the
+task log.
+
 .. code-block:: java
 
     import org.apache.airflow.sdk.*;
@@ -218,11 +228,21 @@ read.
       }
     }
 
-Register tasks manually in a ``BundleBuilder``:
+Register tasks manually in a ``BundleBuilder``. A task class can be top-level like ``FetchTask``, or
+nested ``static`` class like ``ProcessTask``:
 
 .. code-block:: java
 
     public class MyBundle implements BundleBuilder {
+      public static class ProcessTask implements Task {
+        @Override
+        public void execute(Context context, Client client) throws Exception {
+          var fetched = (String) client.getXCom("fetch");
+          // implement task logic
+          client.setXCom(fetched);
+        }
+      }
+
       @Override
       public Iterable<Dag> getDags() {
         var dag = new Dag("my_dag");
--- a/airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
+++ b/airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.airflow.e2e;
+
+import java.util.List;
+import org.apache.airflow.sdk.*;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Bundle of deliberately broken task classes for the runner-behaviour E2E tests.
+ */
+public class TestBundleBuilder implements BundleBuilder {
+  public static class MissingNoArgConstructor implements Task {
+    private final String marker;
+
+    public MissingNoArgConstructor(String marker) {
+      this.marker = marker;
+    }
+
+    public void execute(@NotNull Context context, Client client) {
+      throw new IllegalStateException(marker);
+    }
+  }
+
+  /**
+   * The not static only constructor implicitly takes the enclosing instance, so the
+   * runner's no-argument lookup fails even though no constructor is declared.
+   */
+  public class NonStaticInner implements Task {
+    public void execute(@NotNull Context context, Client client) {}
+  }
+
+  @NotNull
+  @Override
+  public Iterable<Dag> getDags() {
+    var dag = new Dag("java_uninstantiable");
+    dag.addTask("missing_no_arg_constructor", MissingNoArgConstructor.class);
+    dag.addTask("non_static_inner", NonStaticInner.class);
+    return List.of(dag);
+  }
+
+  public static void main(String[] args) {
+    var bundle = new TestBundleBuilder().build();
+    Server.create(args).serve(bundle);
+  }
+}
--- a/airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
+++ b/airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.airflow.e2e;
+
+import java.util.List;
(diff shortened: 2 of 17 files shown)


java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/Task.kt
```diff
@@ -71,17 +73,22 @@ internal object TaskRunner {
...
+    val instance =
+      try {
+        task.getDeclaredConstructor().newInstance()
+      } catch (e: Throwable) {
+        logger.error(
+          "Cannot instantiate task class. A task class must be concrete and declare a public no-argument constructor",
+          mapOf("ti" to request.ti, "taskClass" to task.name, "error" to e, "trace" to e.stackTraceToString()),
+        )
+        return TaskResult.failure(request.tiContext.shouldRetry)
+      }
```
jason810496, line 94: I think we need to handle the error raise by the constructor body itself separately with different helper message.


```suggestion
    val instance =
      try {
        task.getDeclaredConstructor().newInstance()
      } catch (e: InvocationTargetException) {
        val cause = e.cause ?: e
        logger.error(
          "Task class constructor threw an exception",
          mapOf("ti" to request.ti, "taskClass" to task.name, "error" to cause, "trace" to cause.stackTraceToString()),
        )
        return TaskResult.failure(request.tiContext.shouldRetry)
      } catch (e: Throwable) {
        logger.error(
          "Cannot instantiate task class. A task class must be concrete and declare a public no-argument constructor",
          mapOf("ti" to request.ti, "taskClass" to task.name, "error" to e, "trace" to e.stackTraceToString()),
        )
        return TaskResult.failure(request.tiContext.shouldRetry)
    }
```

java-sdk/processor/src/test/kotlin/org/apache/airflow/sdk/BuilderTest.kt
```diff
@@ -297,4 +297,21 @@ class BuilderTest {
...
+        """
+        package org.apache.airflow.example;
+        import org.apache.airflow.sdk.Builder;
+        public class TestExample {
+          @Builder.Dag
+          public class InnerDag { @Builder.Task(id = "foo") public void t1() {} }
+        }
+      """,
+      )
+    assertThat(compilation).failed()
```
jason810496, line 382: It would be nice to assert the error message with `.hadErrorContaining(` as well. Thanks.

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
```diff
@@ -218,11 +228,21 @@ read.
...
 
 .. code-block:: java
 
     public class MyBundle implements BundleBuilder {
+      public static class ProcessTask implements Task {
+        @Override
+        public void execute(Context context, Client client) throws Exception {
+          var fetched = (String) client.getXCom("fetch");
+          // implement task logic
+          client.setXCom(count);
```
jason810496: IIUC, do we mean `fetched` variable here?
```suggestion
          client.setXCom(fetched);
```

jason810496 (commented):
Nice! LGTM overall.

java-sdk/processor/src/test/kotlin/org/apache/airflow/sdk/BuilderTest.kt
FrankYang0529: Sorry for the late reply. Updated it to

```
    assertThat(compilation).hadErrorContaining(
      "an enclosing instance that contains org.apache.airflow.example.TestExample.InnerDag is required",
    )
```

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/Task.kt
FrankYang0529: Thanks for the review. Updated it.

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
FrankYang0529: Updated it.

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
```diff
@@ -205,6 +205,16 @@ Interface-based API
 Implement the ``Task`` interface directly for full control over how tasks are registered and how XComs are
 read.
 
+The runner creates a fresh instance of the task class through reflection for every task-instance run,
+which puts three constraints on the class. It must be:
+
+* Concrete: not abstract and not an interface.
+* A public no-argument constructor.
+* If nested inside another class, it must be ``static`` class.
```
jason810496:
```suggestion
which puts three constraints on the class:

* It must be concrete: not abstract and not an interface.
* It must declare a public no-argument constructor.
* If nested inside another class, it must be a ``static`` nested class.
```

jason810496 (commented):
Sorry that I didn't notice the docs issue in the last review. LGTM after resolving this change.

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
FrankYang0529: Thanks for the review. Updated it.

jason810496 (approved):

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
uranusjr, line 1: I’m not sure if it’s a good idea to have _incorrect_ code as an example. It’s fine to test this in e2e, but I don’t think it should be an example. We should put this somewhere else.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/Task.kt
```diff
@@ -71,17 +74,29 @@ internal object TaskRunner {
...
+          mapOf("ti" to request.ti, "taskClass" to task.name, "error" to cause, "trace" to cause.stackTraceToString()),
+        )
+        return TaskResult.failure(request.tiContext.shouldRetry)
+      } catch (e: Throwable) {
+        logger.error(
+          "Cannot instantiate task class. A task class must be concrete and declare a public no-argument constructor",
+          mapOf("ti" to request.ti, "taskClass" to task.name, "error" to e, "trace" to e.stackTraceToString()),
+        )
+        return TaskResult.failure(request.tiContext.shouldRetry)
+      }
```
uranusjr, line 94: These shouldn’t honor `shouldRetry` since the code can never work no matter how many times you try. They should always `TaskResult.failure(shouldRetry = false)`.

FrankYang0529: Good catch. Chnaged to use `TaskResult.failure(shouldRetry = false)` here.

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
FrankYang0529: Moved it from java-sdk/example to airflow-e2e-tests/java-test-bundle.

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public void execute(@NotNull Context context, Client client) {
+      throw new IllegalStateException(marker);
+    }
+  }
+
+  /**
+   * The not static only constructor implicitly takes the enclosing instance, so the
+   * runner's no-argument lookup fails even though no constructor is declared.
+   */
```
uranusjr, line 42: This is not valid English? Not sure what it’s trying to say.

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+public class TestBundleBuilder implements BundleBuilder {
+  public static class MissingNoArgConstructor implements Task {
+    private final String marker;
+
+    public MissingNoArgConstructor(String marker) {
+      this.marker = marker;
+    }
+
+    public void execute(@NotNull Context context, Client client) {
+      throw new IllegalStateException(marker);
```
uranusjr: This can never be reached (because the class is not instantiable, which is what we’re testing), so this probably should be something like `throw new IllegalStateException("should not be reachable");` instead.

uranusjr (approved):
Should be good after fixing a couple of nits.

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
FrankYang0529: Sorry for the wrong sentence. Updated it to

```
  /**
   * A non-static nested class declares no constructor of its own, but the implicit one
   * takes the enclosing instance, so the runner's lookup for a no-argument constructor
   * fails.
   */
```

airflow-e2e-tests/java-test-bundle/src/java/org/apache/airflow/e2e/TestBundleBuilder.java
FrankYang0529: Thanks for the suggestion. Updated it.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/Task.kt
```diff
@@ -74,19 +77,32 @@ internal object TaskRunner {
...
+          mapOf("ti" to request.ti, "taskClass" to definition.name, "error" to cause, "trace" to cause.stackTraceToString()),
+        )
+        // Retrying cannot help: instantiation fails the same way on every try.
+        return TaskResult.failure(shouldRetry = false)
+      } catch (e: Throwable) {
+        logger.error(
+          "Cannot instantiate task class. A task class must be concrete and declare a public no-argument constructor",
+          mapOf("ti" to request.ti, "taskClass" to definition.name, "error" to e, "trace" to e.stackTraceToString()),
+        )
+        return TaskResult.failure(shouldRetry = false)
```
jason810496: `Throwable` also catches static-initializer and linkage failures from otherwise valid classes, so this reports the wrong contract error. Narrow the structural branch and keep initialization failures distinct.

```suggestion
      } catch (e: ReflectiveOperationException) {
        logger.error(
          "Cannot instantiate task class. A task class must be public and concrete and declare a public no-argument constructor",
          mapOf("ti" to request.ti, "taskClass" to definition.name, "error" to e, "trace" to e.stackTraceToString()),
        )
        return TaskResult.failure(shouldRetry = false)
      } catch (e: Throwable) {
        logger.error(
          "Error initializing task class",
          mapOf("ti" to request.ti, "taskClass" to definition.name, "error" to e, "trace" to e.stackTraceToString()),
        )
        return TaskResult.failure(request.tiContext.shouldRetry)
      }
```

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
```diff
@@ -205,6 +205,16 @@ Interface-based API
...
 
+The runner creates a fresh instance of the task class through reflection for every task-instance run,
+which puts three constraints on the class:
+
+* It must be concrete: not abstract and not an interface.
+* It must declare a public no-argument constructor.
+* If nested inside another class, it must be a ``static`` nested class.
+
+A class that violates any of these fails at runtime with a ``Cannot instantiate task class`` error in the
+task log.
```
jason810496, line 217: Package-private task classes still fail with `IllegalAccessException`. Add class visibility to the constraints and mirror `public` in the runtime error.

```suggestion
The runner creates a fresh instance of the task class through reflection for every task-instance run,
which puts four constraints on the class:

* The task class itself must be ``public``.
* It must be concrete: not abstract and not an interface.
* It must declare a public no-argument constructor.
* If nested inside another class, it must be a ``static`` nested class.

A class that violates any of these fails at runtime with a ``Cannot instantiate task class`` error in the
task log.
```

dev/breeze/src/airflow_breeze/utils/selective_checks.py
```diff
@@ -245,6 +245,7 @@ def __hash__(self):
         FileGroupForCi.JAVA_SDK_E2E_FILES: [
             # `.md` excluded — doc-only edits do not affect the Gradle build.
             r"^java-sdk/(?!.*\.md$).*",
+            r"^airflow-e2e-tests/java-test-bundle/.*",
```
jason810496, line 248: This match makes fixture-only changes run the Java SDK E2E tests and build the PROD image. Please keep the exhaustive decision rules in sync.

Suggested text for `04_selective_checks.md`:

```markdown
* `Java SDK E2E tests` run when Java SDK sources, the Java test bundle, the Java E2E suite or Docker files, or Java coordinator sources change. Enabling them also forces `PROD Image building`.
```

java-sdk/processor/src/test/kotlin/org/apache/airflow/sdk/BuilderTest.kt
```diff
@@ -368,4 +368,24 @@ class BuilderTest {
...
+    assertThat(compilation).hadErrorContaining(
+      "an enclosing instance that contains org.apache.airflow.example.TestExample.InnerDag is required",
+    )
+  }
```
jason810496: This only asserts a pre-existing annotation-processor failure and still passes without the runtime fix, so it is unrelated coverage.

```suggestion
```

jason810496 (commented):
Here're the final nits caught by Codex that make sense to me before merge. Would appreciate to address them when you have a moment, thanks.

FrankYang0529:
@jason810496 Thanks for the review and suggestion! I address all comments.

jason810496:
Thanks!

PR #69828, opened 2026-07-13
Honor verbose logging for Celery workers

### LINKED ISSUE #46990: airflow celery worker does not respect `-v` arg
### Body

you need to set log level via config to get debug

### Committer

jx2lee:
Fixes `airflow celery worker -v` so the Celery worker receives `--loglevel DEBUG` instead of only updating Airflow's root logger.

closes: #46990

---

- [X] Yes — Codex (GPT-5)


Vamsi-klu:
This one is straightforward. The `if args.verbose: celery_log_level = "DEBUG"` sits right after the `CELERY_LOGGING_LEVEL`/`LOGGING_LEVEL` resolution in `worker()` and before `celery_log_level` gets dropped into the `--loglevel` slot of `options`, so `-v` actually rides through to `celery_app.worker_main` now. `args.verbose` is the shared `ARG_VERBOSE` store_true wired into the worker command in `definition.py`, so it is always set and forcing DEBUG only ever raises verbosity, never lowers a configured level. The added test asserts the exact `--loglevel DEBUG` slot. Only stray thought is whether the generic `-v` help text ("Make logging output more verbose") makes this discoverable for the celery worker case, but that is non blocking.

Vamsi-klu (approved):

Miretpl (approved):
I will wait with merge until current in-progress provider release will be finished.

Issue #65712, opened 2026-04-23
Airflow 3.2.1 locks  heartbeat queries causing DB connection pool exhaustion and slow responses


serhii-karelov:
### Under which category would you file this issue?

Providers

### Apache Airflow version

3.2.1

### What happened and how to reproduce it?

When have tens of tasks in-flight, and while surfing the UI, occasionally we get 
1. Queued up hertbeat UPDATE queries in the airflow api
2. Connection pool exhaustion
3. API response times to the heartbeats is high, causing workers to timeout, and retry. 
4. Some tasks terminated and put into a retry state

### Deployment

Other Docker-based deployment

### Apache Airflow Provider(s)

celery

### Official Helm Chart version

Not Applicable

### Anything else?

Some query prevents the heartbeat UPDATEs

```
usename    state                xact_duration    query_duration    query
db_user  idle in transaction  00:00:22.17721   00:00:00.06658    SELECT task_instance.state AS task_instance_state
                                                                   FROM task_instance
                                                                   WHERE task_instance.id = '0<zzzyyy>54'::uuid::UUID
db_user  active               00:00:22.123495  00:00:22.120761   UPDATE task_instance SET last_heartbeat_at='2026-04-23T09:03:18.258390+00:00'::timestamptz, updated_at='2026-04-23T09:03:18.261693+00:00'::timestamptz WHERE task_instance.id
db_user  active               00:00:22.058237  00:00:22.054038   UPDATE task_instance SET last_heartbeat_at='2026-04-23T09:03:18.320462+00:00'::timestamptz, updated_at='2026-04-23T09:03:18.324080+00:00'::timestamptz WHERE task_instance.id
db_user  active               00:00:21.901896  00:00:21.89822    UPDATE task_instance SET last_heartbeat_at='2026-04-23T09:03:18.479643+00:00'::timestamptz, updated_at='2026-04-23T09:03:18.483318+00:00'::timestamptz WHERE task_instance.id
db_user  active               00:00:21.537818  00:00:21.534485   UPDATE task_instance SET last_heartbeat_at='2026-04-23T09:03:18.844164+00:00'::timestamptz, updated_at='2026-04-23T09:03:18.847447+00:00'::timestamptz WHERE task_instance.id
...
```
Slow endpoints 

| # | Duration (s) | Status | Method | Endpoint  |
|---|---|---|---|---|
[...]

Issue #65010, opened 2026-04-10
Airflow 3.2.0 opens too many meta DB connections


rbayer666:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.2.0

### What happened and how to reproduce it?

** Issue Description **
After upgrading from Airflow 3.1.8 to 3.2.0 without changing the configuration we encountered the problem that the meta database (running on PostgreSQL) stops accepting connections. Or, if we increase the allowed connections on the database, the sql achemy pool fills up.

The sql_alchemy settings are the default ones:
- sql_alchemy_pool_enabled = True
- sql_alchemy_pool_size = 5
- sql_alchemy_max_overflow = 10

The allowed parallel database connections on the postgresql is set to 100.

We have 1x api server, 1 x scheduler, 1 x triggerer.

Increasing the sql_alchemy pool and the postgresql connections does not solve the issue - it just takes more time until the connection pool is filled up or the database stops accepting connections.

It looks like the database connections are not properly closed and left open. Over time the connections sum up.

### What you think should happen instead?

All database connections should be closed when they are no longer needed.

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Docker-Compose

### Official Helm Chart version

Not Applicable

### Kubernetes Version

v1.31.13

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

alessandro-peyrachia-rnc:
Hi, I noticed something similar, we just upgraded a couple hours ago.
Even with pgbouncer in front of the postgres db, api will fail after a certain time.
We have 2 api and 2 scheduler, 1 triggerer, 3 workers.
```
core:
        parallelism: 128
        max_active_tasks_per_dag: 64
      scheduler:
        max_tis_per_query: 256
      database:
        sql_alchemy_pool_size: 5
        sql_alchemy_max_overflow: 10
        sql_alchemy_pool_recycle: 600
```
pgbouncer set at
```
poolMode: transaction
parameters:
      max_client_conn: "1000"
      default_pool_size: "30"
      min_pool_size: "10"
      reserve_pool_size: "10"
```
Affects mainly the api service

dclfan:
[...]

PR #69832, opened 2026-07-13
Release DB connection before deserializing DAGs in grid structure endpoint

seanmuth:
## Overview

`get_dag_structure` (`GET /ui/grid/structure/{dag_id}`) streams historical `SerializedDagModel` rows with a `yield_per` server-side cursor while performing CPU-bound `serdag.dag` deserialization and task-group merging *between* fetches. Because the cursor stays open across that work, the read transaction — and, under PgBouncer transaction pooling, the pooled server connection — stays pinned for the entire render of a large DAG.

For DAGs with many tasks and many historical versions this holds a connection open for a long time (multiple minutes in the field). At scale that exhausts the PgBouncer pool and starves task-instance heartbeat updates, contributing to the metadata-DB contention tracked in #65712.

## Why this matters

The UI API and the execution API are served by the same api-server process and share a single metadata-DB connection pool. A UI endpoint that holds a pooled connection across a multi-minute render therefore directly starves execution-API calls — task-SDK heartbeats and state updates — competing for that pool. When those heartbeat/state calls are delayed, workers time out and retry, which piles more load onto the same pool. Keeping UI read handlers from holding connections across CPU-bound work is what prevents UI traffic from degrading task execution.

## Change

Materialize the (page-bounded) set of historical serialized DAGs, detach them from the session, and commit the read transaction so the connection returns to the pool **before** the deserialization/merge loop runs.

`SerializedDagModel.dag` only reads the already-loaded `data` column, so deserialization works on detached instances with no further DB access. The result set is bounded by this endpoint's page of runs (a single DAG's versions), so materializing it is a safe trade for not holding the connection open across CPU-bound work.

This mirrors the per-unit session-release pattern already applied to the streaming `ti_summaries` endpoint in #65010.

## Related

- Contributes to #65712

## Testing

- Existing grid `structure` endpoint tests pass, including `test_structure_includes_historical_removed_task_with_proper_shape`, which exercises the historical-versions loop that was changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 4.8 (1M context)

### DIFF
--- a/airflow-core/newsfragments/69832.bugfix.rst
+++ b/airflow-core/newsfragments/69832.bugfix.rst
@@ -1,0 +1,1 @@
+Release the metadata DB connection in the grid ``structure`` endpoint before deserializing historical DAG versions, so it is not held open across CPU-bound work and does not exhaust the connection pool on large DAGs.
--- a/airflow-core/newsfragments/69832.bugfix.rst
+++ b/airflow-core/newsfragments/69832.bugfix.rst
@@ -1,1 +1,0 @@
-Release the metadata DB connection in the grid ``structure`` endpoint before deserializing historical DAG versions, so it is not held open across CPU-bound work and does not exhaust the connection pool on large DAGs.
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
@@ -69,6 +69,7 @@ from airflow.models.dagrun import DagRun, DagRunNote
 from airflow.models.deadline import Deadline
 from airflow.models.serialized_dag import SerializedDagModel
 from airflow.models.taskinstance import TaskInstance, TaskInstanceNote
+from airflow.utils.helpers import chunks
 from airflow.utils.session import create_session
 
 if TYPE_CHECKING:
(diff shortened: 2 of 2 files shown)


airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -193,29 +193,41 @@ def get_dag_structure(
...
+        SerializedDagModel.dag_id == dag_id,
+        SerializedDagModel.id != latest_serdag_id,
+        SerializedDagModel.dag_version_id.in_(
+            select(TaskInstance.dag_version_id)
+            .join(TaskInstance.dag_run)
+            .where(
+                DagRun.id.in_(run_ids),
+            )
+            .distinct()
+        ),
```
ephraimbuddy: The previous code used yield_per=5 to control memory growth but this now loads everything in memory. I wonder if it's better?

airflow-core/newsfragments/69832.bugfix.rst
```diff
@@ -0,0 +1 @@
+Release the metadata DB connection in the grid ``structure`` endpoint before deserializing historical DAG versions, so it is not held open across CPU-bound work and does not exhaust the connection pool on large DAGs.
```
ephraimbuddy: The newsfragment is not needed

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
seanmuth: Good catch — you're right that materializing all of them regressed the `yield_per=5` memory profile. Reworked in c58da2916e: I now fetch just the serdag ids, then process them in batches of 5, each batch loaded and detached in its own short-lived `create_session()` that's closed *before* the batch is deserialized.

That keeps peak memory bounded to one batch (same as the old behaviour) while still releasing the connection during the CPU-bound deserialization — which is the whole point, since holding it across that work is what pins a PgBouncer server connection for the entire render on large DAGs. It's the same per-unit session-release pattern used for the streaming `ti_summaries` endpoint (#65010).

airflow-core/newsfragments/69832.bugfix.rst
seanmuth: Removed in c58da2916e.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -212,24 +225,33 @@ def get_dag_structure(
                 .distinct()
             ),
         )
-        .execution_options(yield_per=5)  # balance between peak memory usage and round trips
```
kaxil, line 215: @dstandish @jedcunningham -- since I know you both were involved and did this after painful investigations

potiuk:
@seanmuth — There are 3 unresolved review thread(s) on this PR from @ephraimbuddy @kaxil. Could you either push a fix or reply in each thread explaining why the feedback doesn't apply? Once you believe the feedback is addressed, mark the thread as resolved so the reviewer isn't re-pinged needlessly. Thanks!

---

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
seanmuth: For visibility since this may need more eyes: this is functionally resolved already. Ephraim raised the same concern (that materializing all historical serdags regresses the `yield_per=5` memory-bounded behavior from the prior investigation), and c58da2916e reworked the fetch to preserve that property.

Current approach: fetch just the serdag ids (cheap), then process them in batches of 5 (matching the original `yield_per=5` batch size), each batch loaded and detached inside its own short-lived `create_session()` that's closed *before* that batch is deserialized. So peak memory stays bounded to one batch, same as before, and the DB connection is additionally released during the CPU-bound deserialize/merge instead of being held open across it (the change this PR is about).

So the memory-growth concern this thread flagged is addressed by the current diff, not overridden by it. Happy to dig further if there's a specific scenario from the original investigation this doesn't cover.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -193,11 +194,23 @@ def get_dag_structure(
...
+    # expensive. Streaming the rows with a server-side cursor (``yield_per``) and
+    # deserializing between fetches keeps the transaction — and, under PgBouncer
+    # transaction pooling, the pooled server connection — pinned for the entire
+    # render; at scale that exhausts the pool and starves task-instance heartbeats.
+    # See https://github.com/apache/airflow/issues/65712.
+    #
+    # Each batch is loaded and detached in its own short-lived session that is closed
+    # before the batch is deserialized, so the connection is released during the CPU
+    # work while peak memory stays bounded to one batch (``SerializedDagModel.dag``
+    # only reads the already-loaded ``data`` column, so it works on detached rows).
```
dstandish, line 211: i'm not sure this comment is really adding value... at least not *here* because, it immediately precedes a very simple query.

maybe just move this to the commit message? it's kinda high level rationale for why this change overall. the area where it sorta makes sense to have comments is below where we're batching things and we can state succinctly the rationale, which you kinda sorta already do.

dstandish: maybe just say 

```
# we get the ids so that we can split serialization into batches and balance round trips and mem usage
```

pierrejeambrun: +1 this comment is too verbose

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -212,24 +225,33 @@ def get_dag_structure(
                 .distinct()
             ),
         )
-        .execution_options(yield_per=5)  # balance between peak memory usage and round trips
+        .order_by(SerializedDagModel.id)
```
pierrejeambrun, line 228: Why a new order_by? I believe this induce a behavior change in the ordering later on. I  wouldn't add this if not necessary.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -212,24 +225,33 @@ def get_dag_structure(
...
-                depth=depth,
-            )
-        # Merge immediately instead of collecting all Dags in memory
-        nodes = [task_group_to_dict_grid(x) for x in task_group_sort(filtered_dag.task_group)]
-        _merge_node_dicts(merged_nodes, nodes)
-
-        session.expunge(serdag)  # to allow garbage collection
+    serdag_ids = list(session.scalars(serdag_id_query))
+    # Release the request session's transaction/connection before the batched work.
+    session.commit()
```
pierrejeambrun, line 232: `session.close()` reads better here and achieve the same

pierrejeambrun (approved):
Very nice, thanks for the PR.

A couple suggestions / nits. (nothing blocking)

Tested the PR locally against a fictive database with a lot of historical versions. Can confirm the gain. (31s vs 0.22s for transaction age)

vatsrahul1001:
@seanmuth can you resolve the conversations after fixing

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
seanmuth: Trimmed to exactly your suggested wording in 4aef9fa789. Thanks.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
seanmuth: Good catch, dropped it in 4aef9fa789. It wasn't necessary — the original code had no explicit ordering either, and `_merge_node_dicts` is order-sensitive for nodes that differ across historical Dag versions (first write wins for structure), so adding an ordering guarantee here would have been an unreviewed behavior change relative to what shipped before. This restores the exact prior (unordered) behavior.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
seanmuth: Agreed, swapped to `session.close()` in 4aef9fa789 — this session does no writes here, so close() is more accurate and releases the connection the same way. Thanks.

seanmuth:
@vatsrahul1001 sorry this one slid of my plate last week. PR comments addressed and resolved, also rebased and we're back to mergeable, will keep an eye on the CI tests but this _should_ be GTG now!

eladkal (approved):

pierrejeambrun (approved):

PR #69838, opened 2026-07-13
Fix Neptune Analytics deferrable path: load custom waiter + forward hook params to triggers

seanghaeli:
Forward additional hook params to the deferred triggers to fix `example_neptune_analytics` on the deferrable path ([currently failing](https://aws-mwaa.github.io/#/open-source/system-tests/dashboard.html))


seanghaeli:
Closing — investigation against the real pipeline traceback showed this doesn't fix the actual deferrable failure. The failing call is a sync `delete_endpoint` raising `AccessDeniedException: Unable to determine service/operation name to be authorized` (an async→sync botocore context issue), which neither the waiter rename nor the trigger param-forwarding addresses. Reworking the fix against the real root cause.

PR #66608, opened 2026-05-08
Fetch deadline callback context via Execution API at runtime

seanghaeli:
## Summary

Replace the simple context workaround from #55241 that stored serialized context in trigger kwargs (DB). Now that #55068 gives the triggerer API access, fetch the DagRun at execution time via the Execution API and build context fresh.

This avoids DB bloat from serialized context, provides fresh (not stale) context, and builds a richer context dict including `logical_date`, `ds`, `ts`, `conf`, `data_interval_start/end`, and the deadline info.

## Changes

- **`deadline.py`**: Remove `get_simple_context()`. Store only identifiers (`dag_id`, `run_id`, `deadline_id`, `deadline_time`) in callback kwargs.
- **`callback.py`**: Add `_build_context()` that fetches DagRun via `SUPERVISOR_COMMS.asend(GetDagRun(...))`. Backward compat: old callbacks with `"context"` key still work.
- **`triggerer_job_runner.py`**: Add `GetDagRun` to `ToTriggerSupervisor` union, `DagRunResult` to `ToTriggerRunner` union, handler in `_handle_request`.
- **`callback_supervisor.py`**: Add `GetDagRun` to `CallbackToSupervisor` union + handler for executor callback path.
- **Tests**: Updated deadline model tests, added context-fetching test, backward-compat test, `GetDagRun` handler test.

- **`serialization/definitions/dag.py`**: Also fixes scheduled DAG runs silently getting no deadline — this includes a pre-existing main bug (a `VariableInterval` deadline committed mid-scheduler and got dropped), kept here because it's the same code block this PR rewrites.

## Testing

Ran in Breeze to verify the comms plumbing works e2e:

- Confirmed `GetDagRun` round-trips through the triggerer's `ToTriggerSupervisor` → `_handle_request` → `DagRunResult` response path without breaking existing trigger handling
- Verified `SUPERVISOR_COMMS.asend()` is the correct async calling pattern — uses `TriggerCommsDecoder` from `init_comms()` with async lock for coroutine safety in the trigger event loop
- Verified the `DagRun` generated model has all fields accessed in `_build_context`: `logical_date`, `data_interval_start`, `data_interval_end`, `conf`
- Backward compat confirmed: old callbacks with stored `"context"` key (queued before this change) still work

## Motivation

Per @ramitkataria's feedback on #64984: context should not be stored in the DB. The triggerer now has API access (#55068), so fetch it at runtime like tasks do.

## Related

- Addresses feedback from #64984 (closed)
[...]

PR #69840, opened 2026-07-13
Add single-use callback token for deadline callback context fetch

seanghaeli:
Adds single-use `callback` token so a deadline-callback subprocess can fetch its DagRun context from the Execution API without holding a broad `workload` token (the scope creep that got #66608 reverted).

- New `PATCH /callbacks/{id}/run`: single-use exchange gated on an atomic `Callback.state` QUEUED→RUNNING transition (409 on reuse), returning a short-lived `execution` token via `Refreshed-API-Token`. Mirrors the TI `/run` pattern.
- New `callback` token scope + `callback:self` check (token subject must match the callback id).
- Extracts a shared `issue_execution_token` helper used by both the callback exchange and TI `/run`, and collapses the `*:self` scope checks into one parameterized check.


seanghaeli:
@ashb this one is as-per your suggestion. Your feedback would be appreciated!

airflow-core/src/airflow/executors/workloads/base.py
```diff
@@ -84,12 +84,12 @@ class BaseWorkloadSchema(BaseModel):
     """The identity token for this workload"""
 
     @staticmethod
-    def generate_token(sub_id: str, generator: JWTGenerator | None = None) -> str:
+    def generate_token(sub_id: str, generator: JWTGenerator | None = None, scope: str = "workload") -> str:
```
ashb: Hmmm, for a reason I can't fully articulate, I'm a bit wary of having a `scope` argument here....

airflow-core/src/airflow/api_fastapi/execution_api/security.py
```diff
@@ -196,13 +196,26 @@ async def require_auth(
...
+                detail="Token subject does not match callback ID",
+            )
 
     return token
 
 
 CurrentTIToken: TIToken = Depends(require_auth)
 
 
+def issue_execution_token(services: Any, response: Response, sub: str) -> None:
```
ashb: Nit:

```suggestion
def issue_execution_token(services: svcs.Container, response: Response, sub: str) -> None:
```

ashb (commented):
Am I correct in my understanding that this new callback token is exchanged for a normal execution token once by hitting the /callbacks/{id}/run endpoint?

seanghaeli:
> Am I correct in my understanding that this new callback token is exchanged for a normal execution token once by hitting the /callbacks/{id}/run endpoint?

Exactly right

airflow-core/src/airflow/executors/workloads/base.py
seanghaeli: Re-wrote it so that only workloads can set their own scope, instead of being passed as a free parameter from anywhere

seanghaeli:
Accidentally closed this and it's not letting me re-open it, so I made a new PR #71192

PR #69846, opened 2026-07-14
Allow `AthenaOperator` queries without a `database` argument

te-horie:
# Summary

This PR makes the `database` argument **optional** in `AthenaOperator`, allowing queries that do not require a default database.
This is useful when all referenced table names are fully qualified, such as in queries across multiple databases.

When `database` is not provided, `Database` from `query_execution_context` is used as the fallback database for OpenLineage dataset extraction.

# Testing

- ✅ `breeze run pytest providers/amazon/tests/unit/amazon/aws/operators/test_athena.py` **passed** (16 test cases).
- ✅ `prek run --from-ref upstream/main` **passed**.
- ✅ **Successfully** ran the system test against Amazon Athena
  1. Locally removed the `database` argument from the `read_table` task in [`example_athena.py`](https://github.com/apache/airflow/blob/main/providers/amazon/tests/system/amazon/aws/example_athena.py) like this:
      ```diff
      diff --git a/providers/amazon/tests/system/amazon/aws/example_athena.py b/providers/amazon/tests/system/amazon/aws/example_athena.py
      index 36a37a8e1c..3c72e9b309 100644
      --- a/providers/amazon/tests/system/amazon/aws/example_athena.py
      +++ b/providers/amazon/tests/system/amazon/aws/example_athena.py
      @@ -129,7 +129,6 @@ with DAG(
           read_table = AthenaOperator(
               task_id="read_table",
               query=query_read_table,
      -        database=athena_database,
               output_location=f"s3://{s3_bucket}/",
           )
           # [END howto_operator_athena]
      ```
  2. Ran the system test with the following command; it **passed** ✅:
      ```bash
      $ SYSTEM_TESTS_ENV_ID=airflowtest$(date +%s) breeze run --forward-credentials env AWS_PROFILE=airflow-test pytest --system providers/amazon/tests/system/amazon/aws/example_athena.py
      ...
      providers/amazon/tests/system/amazon/aws/example_athena.py::test_run <- devel-common/src/tests_common/test_utils/system_tests.py PASSED [100%]
      
      ================================================= 1 passed, 1 warning in 56.41s ==================================================
      ```
  3. Confirmed that the query succeeded with only `Catalog` in `QueryExecutionContext`:
      ```bash
      $ aws athena get-query-execution --query-execution-id <<<REDACTED>>> --profile airflow-test
      QueryExecution:
        ...
        Query: SELECT * from airflowtest1784003174_default.airflowtest1784003174_test_table
        QueryExecutionContext:
          Catalog: awsdatacatalog
        QueryExecutionId: <<<REDACTED>>>
        ...
        Status:
          CompletionDateTime: '2026-07-14T13:27:19.237000+09:00'
          State: SUCCEEDED
          SubmissionDateTime: '2026-07-14T13:27:17.950000+09:00'
        SubstatementType: SELECT
        WorkGroup: primary
      ```

---

Generated-by: OpenAI Codex (GPT-5.5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

Drafted-by: OpenAI Codex; reviewed by @te-horie before posting

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/operators/athena.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/operators/athena.py
@@ -50,6 +50,8 @@ class AthenaOperator(AwsBaseOperator[AthenaHook]):
     :param database: Default database for query execution. (templated)
         This argument is optional when the query does not require a default database,
         such as when all referenced table names are fully qualified.
+        If omitted or set to ``None``, any ``Database`` value set in
+        the ``query_execution_context`` will be used instead.
     :param catalog: Catalog to select. (templated)
     :param output_location: s3 path to write the query results into. (templated)
         To run the query, you must specify the query results location using one of the ways:
--- a/providers/amazon/src/airflow/providers/amazon/aws/operators/athena.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/operators/athena.py
@@ -47,7 +47,9 @@ class AthenaOperator(AwsBaseOperator[AthenaHook]):
         :ref:`howto/operator:AthenaOperator`
 
     :param query: Trino/Presto query to be run on Amazon Athena. (templated)
-    :param database: Database to select. (templated)
+    :param database: Default database for query execution. (templated)
+        This argument is optional when the query does not require a default database,
+        such as when all referenced table names are fully qualified.
     :param catalog: Catalog to select. (templated)
     :param output_location: s3 path to write the query results into. (templated)
         To run the query, you must specify the query results location using one of the ways:
@@ -88,7 +90,7 @@ class AthenaOperator(AwsBaseOperator[AthenaHook]):
         self,
         *,
         query: str,
-        database: str,
+        database: str | None = None,
         output_location: str | None = None,
         client_request_token: str | None = None,
         workgroup: str = "primary",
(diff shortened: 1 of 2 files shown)


Vamsi-klu:
The `if self.database:` guard in `execute()` looks right, it skips writing `Database` into `query_execution_context` when the arg is unset instead of forcing a `None` in there, and defaulting the signature to `database: str | None = None` leaves existing keyword callers alone. Nice that `fallback_database = self.database or self.query_execution_context.get("Database")` keeps the OpenLineage schema resolution working when the database only arrives through the execution context. One subtle behavior maybe worth a docstring line: with `database=None` and a `query_execution_context` that already sets `Database`, `execute()` now preserves that value instead of overwriting it, which looks like the right call but is easy to miss. The two new tests cover the no database run and the context fallback, so this looks good to me.

vincbeck (approved):

te-horie:
> One subtle behavior maybe worth a docstring line: with `database=None` and a `query_execution_context` that already sets `Database`, `execute()` now preserves that value instead of overwriting it, which looks like the right call but is easy to miss.

Thank you for your suggestion!
I added a note to the `database` parameter docstring in [f72d787](https://github.com/apache/airflow/pull/69846/commits/f72d787a25417f88c251a3144a7a5f98755b0fd6) explaining that any `Database` value in `query_execution_context` is used when `database` is omitted or set to `None`.

eladkal (approved):

PR #69849, opened 2026-07-14
Reduce scheduler memory for asset-triggered Dag runs

### LINKED ISSUE #69848: Asset-triggered Dag runs cause scheduler memory spikes

Ariflow Core

### Apache Airflow version

`main`

When the scheduler creates an asset-triggered Dag run, it selects every `AssetEvent` in the event window as a full ORM object. This materializes each event's JSON metadata and relationship state in the scheduler process before writing rows to `dagrun_asset_event`.

The scheduler's peak memory therefore grows linearly with the number of events. A production-path benchmark using a 512-byte JSON payload per event produced the following baseline results:

| Database | Events | Duration | Python peak | RSS increase |
|---|---:|---:|---:|---:|
| SQLite | 1,000 | 0.537 s | 6.23 MB | 0.17 MB |
| SQLite | 10,000 | 1.785 s | 54.24 MB | 79.98 MB |
| SQLite | 50,000 | 12.545 s | 274.21 MB | 465.98 MB |
| PostgreSQL | 1,000 | 0.339 s | 7.39 MB | 1.88 MB |
| PostgreSQL | 10,000 | 2.510 s | 62.34 MB | 108.78 MB |
| PostgreSQL | 50,000 | 10.046 s | 289.72 MB | 483.77 MB |
| MySQL | 1,000 | 2.622 s | 6.76 MB | 0.75 MB |
| MySQL | 10,000 | 4.478 s | 54.54 MB | 98.21 MB |
| MySQL | 50,000 | 21.502 s | 272.63 MB | 475.28 MB |

Reproduction outline:

1. Create an asset-scheduled Dag with `catchup=False`.
2. Insert 1,000, 10,000, or 50,000 matching `AssetEvent` rows containing a 512-byte JSON value.
3. Add the corresponding `AssetDagRunQueue` row.
4. Invoke `SchedulerJobRunner._create_dag_runs_asset_triggered`.
5. Measure elapsed time, Python allocations, and scheduler RSS during Dag run creation.

Fixture creation is outside the measured interval. Every run creates the expected number of
`dagrun_asset_event` associations.

Dag run creation should preserve exact consumed-event membership without materializing complete `AssetEvent` objects in scheduler memory. The association can be written as a set-based `INSERT ... SELECT` in the existing transaction, using the same direct-asset, alias, and event-window predicates.

This keeps Dag run creation, event association, and queue consumption atomic while making scheduler-side Python memory effectively independent of the number and payload size of matching events.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

The behavior is reproducible on SQLite, PostgreSQL, and MySQL.

### Code of Conduct

viiccwen:
## Description

closes #69848 

Asset-triggered Dag run creation currently materializes every matching `AssetEvent` ORM object, including its JSON metadata, before writing the Dag run association rows. Large event windows can therefore cause scheduler memory spikes far beyond the raw payload size.

This change writes the association rows with a server-side `INSERT ... SELECT` while preserving the existing event predicates, transaction boundary, exact event membership, and relationship behavior.

Scheduler-side `AssetEvent` ORM and JSON payload materialization drops **from O(events) to approximately O(1).** The database still performs O(events) work to select event IDs and persist one association row per event; this change moves that set operation out of Python rather than eliminating it.

The partitioned-asset path uses the same set-based association helper. No schema or configuration change is needed.

## Benchmark

The benchmark invokes the real scheduler path on SQLite, PostgreSQL, and MySQL with a 512-byte JSON
payload per event. Setup is outside the measured interval.

| Database | Events | Duration (before/after) | Speedup | Python peak (before/after) | Peak reduction |
|---|---:|---:|---:|---:|---:|
| SQLite | 1,000 | 0.537 / 0.255s | 2.11× | 6.23 / 0.16 MB | 97.2% |
| SQLite | 10,000 | 1.785 / 0.024s | 74.38× | 54.24 / 0.11 MB | 99.8% |
| SQLite | 50,000 | 12.545 / 0.071s | **176.69×** | 274.21 / 0.13 MB | **99.95%** |
| PostgreSQL | 1,000 | 0.339 / 0.237s | 1.43× | 7.39 / 1.84 MB | 75.1% |
| PostgreSQL | 10,000 | 2.510 / 0.096s | **26.15×** | 62.34 / 0.12 MB | 99.8% |
| PostgreSQL | 50,000 | 10.046 / 0.688s | 14.60× | 289.72 / 0.14 MB | **99.95%** |
| MySQL | 1,000 | 0.161 / 0.032s | 5.00× | 5.80 / 0.16 MB | 97.2% |
| MySQL | 10,000 | 4.478 / 0.185s | **24.21×** | 54.54 / 0.11 MB | 99.8% |
| MySQL | 50,000 | 21.502 / 1.977s | 10.88× | 272.63 / 0.13 MB | **99.95%** |

The 1,000-event result includes first-call and tracing warm-up overhead.

## Notes

* `consumed_asset_event_count` is designed for logging, can discuss leave it or not.

---

- [x] Yes - Codex (GPT 5.6 Sol)

---

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -169,11 +169,11 @@ MAX_PARTITION_DAG_RUNS_PER_LOOP = 500
 def _associate_asset_events_with_dag_run(
     *,
     dag_run: DagRun,
-    asset_event_ids: Select[tuple[int]],
+    asset_event_ids_select: Select[tuple[int]],
     session: Session,
 ) -> None:
     """Associate selected asset events without materializing ORM objects."""
-    selected_event_ids = asset_event_ids.subquery()
+    selected_event_ids = asset_event_ids_select.subquery()
     session.execute(
         insert(association_table).from_select(
             ["dag_run_id", "event_id"],
@@ -2362,13 +2362,13 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 creating_job_id=self.job.id,
                 session=session,
             )
-            asset_event_ids = select(AssetEvent.id.label("event_id")).where(
+            asset_event_ids_select = select(AssetEvent.id.label("event_id")).where(
                 PartitionedAssetKeyLog.asset_partition_dag_run_id == apdr.id,
                 PartitionedAssetKeyLog.asset_event_id == AssetEvent.id,
             )
             _associate_asset_events_with_dag_run(
                 dag_run=dag_run,
-                asset_event_ids=asset_event_ids,
+                asset_event_ids_select=asset_event_ids_select,
                 session=session,
             )
             session.flush()
@@ -2642,21 +2642,25 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 )
             event_window_floor.append(date.min)
 
-            asset_event_ids = select(AssetEvent.id.label("event_id")).where(
-                or_(
-                    AssetEvent.asset_id.in_(
-                        select(DagScheduleAssetReference.asset_id).where(
-                            DagScheduleAssetReference.dag_id == dag.dag_id
+            asset_event_ids_select = (
+                select(AssetEvent.id.label("event_id"))
+                .where(
+                    or_(
+                        AssetEvent.asset_id.in_(
+                            select(DagScheduleAssetReference.asset_id).where(
+                                DagScheduleAssetReference.dag_id == dag.dag_id
+                            ),
+                        ),
+                        AssetEvent.source_aliases.any(
+                            AssetAliasModel.scheduled_dags.any(
+                                DagScheduleAssetAliasReference.dag_id == dag.dag_id
+                            )
                         ),
                     ),
-                    AssetEvent.source_aliases.any(
-                        AssetAliasModel.scheduled_dags.any(
-                            DagScheduleAssetAliasReference.dag_id == dag.dag_id
-                        )
-                    ),
-                ),
-                AssetEvent.timestamp <= triggered_date,
-                AssetEvent.timestamp > func.coalesce(*event_window_floor),
+                    AssetEvent.timestamp <= triggered_date,
+                    AssetEvent.timestamp > func.coalesce(*event_window_floor),
+                )
+                .order_by(AssetEvent.timestamp.asc(), AssetEvent.id.asc())
             )
 
             dag_run = dag.create_dagrun(
@@ -2680,7 +2684,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             stats.incr("asset.triggered_dagruns", tags=prune_dict({"team_name": team_name}))
             _associate_asset_events_with_dag_run(
                 dag_run=dag_run,
-                asset_event_ids=asset_event_ids,
+                asset_event_ids_select=asset_event_ids_select,
                 session=session,
             )
             consumed_asset_event_count = (
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -43,7 +43,9 @@ from sqlalchemy import (
     delete,
     exists,
     func,
+    insert,
     inspect,
+    literal,
     or_,
     select,
     text,
@@ -85,6 +87,7 @@ from airflow.models.asset import (
     PartitionedAssetKeyLog,
     TaskInletAssetReference,
     TaskOutletAssetReference,
+    association_table,
 )
 from airflow.models.asset_state_store import AssetStateStoreModel
 from airflow.models.backfill import Backfill, BackfillDagRun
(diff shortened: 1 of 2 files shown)


viiccwen:
cc @Lee-W, @jason810496. ✨🙌

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2341,13 +2362,15 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
                 creating_job_id=self.job.id,
                 session=session,
             )
-            asset_events = session.scalars(
-                select(AssetEvent).where(
-                    PartitionedAssetKeyLog.asset_partition_dag_run_id == apdr.id,
-                    PartitionedAssetKeyLog.asset_event_id == AssetEvent.id,
-                )
+            asset_event_ids = select(AssetEvent.id.label("event_id")).where(
```
Lee-W: This naming is inaccurate. This is a SQL clause insetad of id itself

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2619,27 +2642,21 @@ def _create_dag_runs_asset_triggered(
...
-                )
+                    ),
+                    AssetEvent.source_aliases.any(
+                        AssetAliasModel.scheduled_dags.any(
+                            DagScheduleAssetAliasReference.dag_id == dag.dag_id
+                        )
+                    ),
+                ),
+                AssetEvent.timestamp <= triggered_date,
+                AssetEvent.timestamp > func.coalesce(*event_window_floor),
```
Lee-W: why do we remove the `order_by` here?

Lee-W (commented):
Thanks! I remember we did something similar elsewhere. the overall idea looks good to me

airflow-core/src/airflow/jobs/scheduler_job_runner.py
viiccwen: Sure, thx for pointing out!

airflow-core/src/airflow/jobs/scheduler_job_runner.py
viiccwen: Oh, I missed it. Thx for catching!

viiccwen:
cc @Lee-W. 🫡

Lee-W (dismissed):

viiccwen:
Hello @Lee-W, since upstream had been heavily modified, need review again. 🙏

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2733,12 +2762,32 @@ def _create_dag_runs_asset_triggered(
                     else None
                 )
                 stats.incr("asset.triggered_dagruns", tags=prune_dict({"team_name": team_name}))
-                dag_run.consumed_asset_events.extend(asset_events)
+                _associate_asset_events_with_dag_run(
+                    dag_run=dag_run,
+                    asset_event_ids_select=select(selected_asset_events.c.event_id)
+                    .where(
+                        selected_asset_events.c.timestamp <= triggered_date,
+                        selected_asset_events.c.event_id <= max_selected_event_id,
```
uranusjr, line 2770: This _looks_ like dead weight unless the reader understands the CTE is re-evaluated fresh in this second, independent query and could pick up events inserted after the aggregate read. Add a comment explaining this is a snapshot bound against a race, not a no-op.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2733,12 +2762,32 @@ def _create_dag_runs_asset_triggered(
...
+                    session=session,
+                )
+                consumed_asset_event_count = (
+                    session.scalar(
+                        select(func.count())
+                        .select_from(association_table)
+                        .where(association_table.c.dag_run_id == dag_run.id)
+                    )
+                    or 0
+                )
```
uranusjr, line 2785: We can probably use `rowcount` instead? If we must use a separate query, it’d be a good idea to at least add a guard on log level (with `isEnabledFor`)

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -163,6 +166,23 @@
...
+) -> None:
+    """Associate selected asset events without materializing ORM objects."""
+    selected_event_ids = asset_event_ids_select.subquery()
+    session.execute(
+        insert(association_table).from_select(
+            ["dag_run_id", "event_id"],
+            select(literal(dag_run.id), selected_event_ids.c.event_id),
+        )
+    )
+    session.expire(dag_run, ["consumed_asset_events"])
```
uranusjr, line 183: Worth a comment explaining why manual expiration is needed.

uranusjr:
I think this is close to mergeable with some nits.

PR #69850, opened 2026-07-14
prevent path traversal in ObjectStoragePath recursive copy to local

Samin061:
ObjectStoragePath.copy() delegates a recursive remote-to-local transfer straight to fsspec's fs.get, so object keys containing .. segments (which anyone able to write to the source prefix controls) are followed verbatim and land outside the destination directory. Add a containment check on the recursive remote-to-local branch that expands the source keys and refuses any whose resolved local path escapes the destination, matching the guards already in the S3 and GCS sync_to_local_dir helpers. Non-recursive and same-store copies are untouched, so valid transfers behave exactly as before.

---

### DIFF
--- a/task-sdk/src/airflow/sdk/io/path.py
+++ b/task-sdk/src/airflow/sdk/io/path.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import shutil
 from typing import TYPE_CHECKING, Any, ClassVar
 from urllib.parse import urlsplit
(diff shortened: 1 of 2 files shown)


task-sdk/src/airflow/sdk/io/path.py
```diff
@@ -339,6 +340,25 @@ def size(self) -> int:
...
+        Refuse a recursive download when a remote object key resolves outside ``local_dir``.
+
+        Object-store keys are arbitrary strings and may contain ``..`` segments written by
+        anyone who can put objects in the source prefix; ``fs.get`` follows them verbatim and
+        would write outside the destination directory.
+        """
+        dst_root = os.path.realpath(local_dir)
+        for src_key in self.fs.expand_path(self.path, recursive=True, **kwargs):
+            if self.fs.isdir(src_key):
+                continue
```
steveahnahn, line 354: This adds one isdir round trip per object, on top of a second full expand_path traversal. Measured against main with a fake remote fs: 10 objects goes from 17 isdir calls to 34, 50 from 57 to 114, 200 from 207 to 414. It exactly doubles them, so the cost grows with object count on a path that is already doing a bulk transfer.

self.fs.find(self.path, **kwargs) returns files only and gives an identical file set to the expand_path plus per-key isdir filter, so swapping it in would drop the per-key calls entirely and make the guard close to free.

steveahnahn (approved):
Good catch and the containment check on realpath plus commonpath with tests looks right. I ran the tests locally and also tried a destination containing a pre-existing symlink pointing outside itself, which the guard correctly rejects since realpath resolves the link before the check.

One non-blocking performance note inline

task-sdk/src/airflow/sdk/io/path.py
Samin061: Good point, the doubled traversal was wasteful. Switched to fs.find(self.path), which returns files only, so the per-key isdir calls are gone and the guard walks the tree just once. Tests still green.

PR #69852, opened 2026-07-14
Handle TaskNotFound in triggerer to prevent stale triggers from blocking batch

### LINKED ISSUE #69841: Prevent stale task triggers from blocking healthy workloads
### Apache Airflow version

3.3.0 and current `main`

### What happens

The triggerer supervisor loads the pinned serialized Dag for every task-associated trigger so it can determine whether the task uses `start_from_trigger`. If that serialized Dag exists but no longer contains the deferred task, `get_task()` raises `TaskNotFound` from workload construction. The exception escapes the per-batch loop, so one stale task instance prevents unrelated healthy triggers in the same batch from starting and can repeatedly restart the triggerer supervisor.

### Sanitized reproducer

1. Create two deferred task instances with ordinary triggers.
2. Leave one task instance linked to a Dag version whose serialized Dag no longer contains that task ID.
3. Leave the second task instance and its pinned Dag version valid.
4. Let the triggerer claim both trigger rows in one polling cycle.

The first call to `LazyDeserializedDAG.get_task()` raises `TaskNotFound`. The healthy trigger is not constructed during that supervisor pass.

### Expected behavior

A stale task instance must not abort construction of unrelated trigger workloads. A backportable fix can treat this narrow lookup failure as an ordinary context-free deferral, matching behavior before triggerer-side `start_from_trigger` context loading was introduced.

The durable fix should persist whether each trigger requires Dag context. Ordinary deferrals can then avoid serialized-Dag lookup entirely, direct-to-triggerer tasks can require their pinned Dag data, and legacy rows can self-classify once. If a context-required row points to a serialized Dag that lacks its task, the task instance should become `REMOVED` and unlink from the trigger without blocking healthy workloads.

### Acceptance criteria

- One stale task-associated trigger cannot abort the triggerer supervisor batch.
- Ordinary deferrals do not query or deserialize their Dag.
- Context-required triggers never run with missing or unrendered Dag context.
- Legacy rows remain safe during rolling upgrades.

---

hanumanraj07:
## What

Catches `TaskNotFound` in the triggerer supervisor's `_create_workload` method so that one stale task-associated trigger cannot abort the entire batch of healthy trigger workloads.

Closes: #69841

## Why

The triggerer supervisor loads the pinned serialized Dag for every task-associated trigger to check whether the task uses `start_from_trigger`. If the serialized Dag exists but no longer contains the deferred task, `get_task()` raises `TaskNotFound`. The exception escapes the per-batch loop in `build_trigger_workloads`, so one stale task instance prevents unrelated healthy triggers in the same batch from starting and can repeatedly restart the triggerer supervisor.

## How

- Wrapped `get_task()` in `_create_workload` with a `try/except TaskNotFound` block
- When caught, logs a warning and sets `task = None`, falling through to the context-free `RunTrigger` return (matching behavior when `serialized_dag_model` is `None`)
- Changed `task.start_from_trigger` check to `task is not None and task.start_from_trigger`
- Added unit test verifying the fix


steveahnahn (approved):
Good catch, I verified the test fails on main and passes with the fix. Good use of `TaskNotFound` rather than a blanket except, LGTM.

PR #69855, opened 2026-07-14
Add retry-safe batch_id_prefix to DataprocCreateBatchOperator

amirmor1:
Add an opt-in `batch_id_prefix` to `DataprocCreateBatchOperator` so task retries can submit a new Dataproc batch with a unique ID instead of reusing a static `batch_id`.

### Why

When a task retries with the same `batch_id`, Dataproc can return `ALREADY_EXISTS` and attach to the old batch. This change provides a built-in retry-safe path without changing existing explicit `batch_id` behavior.

### What changes

- add optional `batch_id_prefix` parameter
- generate `<batch_id_prefix>-<unique_suffix>` per task attempt
- keep `batch_id` and `batch_id_prefix` mutually exclusive
- keep explicit `batch_id` behavior unchanged
- update Dataproc operator docs
- add unit tests for:
  - mutual exclusivity
  - prefix-based ID generation
  - execute flow with prefix
  - backward compatibility for explicit `batch_id`

### Test plan

- [x] `uv run ruff format providers/google/src/airflow/providers/google/cloud/operators/dataproc.py providers/google/tests/unit/google/cloud/operators/test_dataproc.py`
- [x] `uv run ruff check --fix providers/google/src/airflow/providers/google/cloud/operators/dataproc.py providers/google/tests/unit/google/cloud/operators/test_dataproc.py`
- [x] `uv run --project providers/google pytest providers/google/tests/unit/google/cloud/operators/test_dataproc.py::TestDataprocCreateBatchOperator::test_execute providers/google/tests/unit/google/cloud/operators/test_dataproc.py::TestDataprocCreateBatchOperator::test_execute_with_batch_id_prefix providers/google/tests/unit/google/cloud/operators/test_dataproc.py::TestDataprocCreateBatchOperator::test_build_unique_batch_id_from_prefix_preserves_prefix_verbatim providers/google/tests/unit/google/cloud/operators/test_dataproc.py::TestDataprocCreateBatchOperator::test_batch_id_and_batch_id_prefix_are_mutually_exclusive -xvs`

- [X] Yes — Codex 5.3

Generated-by: Codex 5.3 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


shahar1 (commented):
Could you please add AI disclosure to the PR body? Thank you!

amirmor1:
@shahar1 
Thanks for the review, and apologies for the slow follow-up — the AI disclosure has been added to the PR body.

CI is green and the branch is conflict-free. This is ready for another look whenever you have time.

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -2536,13 +2545,15 @@ def execute(self, context: Context):
...
-                batch_id=self.batch_id,
+                batch_id=requested_batch_id,
                 request_id=self.request_id,
                 retry=self.retry,
                 timeout=self.timeout,
                 metadata=self.metadata,
             )
         except AlreadyExists:
+            if not batch_id:
+                raise ValueError("Dataproc reported an existing batch without a requested batch_id.")
```
VladaZakharova: i am not sure i understand this error message. can you pleas explain?

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -2416,9 +2417,12 @@ class DataprocCreateBatchOperator(GoogleCloudBaseOperator):
     :param project_id: Optional. The ID of the Google Cloud project that the cluster belongs to. (templated)
     :param region: Required. The Cloud Dataproc region in which to handle the request. (templated)
     :param batch: Required. The batch to create. (templated)
-    :param batch_id: Required. The ID to use for the batch, which will become the final component
+    :param batch_id: Optional. The ID to use for the batch, which will become the final component
```
VladaZakharova, line 2420: As i can see from it's definition batch_id was always an optional param, right?

VladaZakharova (commented):
Left several comments. Can you please also add this param to system test and send a screenshot of green test on airflow UI?

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
amirmor1: You're right — batch_id has always been optional in the signature (batch_id: str | None = None), and execute() already had a branch logging "The batch ID will be generated since it was not provided." The docstring saying "Required" was simply stale. I corrected it here because this PR makes that optional path more prominent, but it's a docs fix rather than a behaviour change.

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
amirmor1: Good catch, and sorry for the unclear message. That guard only triggered when neither batch_id nor batch_id_prefix was set — in which case Dataproc generates the ID server-side and ALREADY_EXISTS shouldn't be reachable. Rather than document an error for an unreachable state, I've removed the guard here and the duplicate in retry_batch_creation. This restores the original behaviour exactly: the except AlreadyExists block in execute() is now unchanged from main.

amirmor1:
@VladaZakharova 
Thanks for the review — both inline points are addressed above, and the system test is in as of the latest push.

batch_id_prefix is now exercised by example_dataproc_batch.py: a create_batch_with_prefix task submits with the prefix, and the teardown resolves the generated ID from XCom, since that ID is only known at runtime. The docs no longer carry a hand-written snippet — they exampleinclude that same task, so the documented example is the one that actually runs.

Since system tests are excluded from PR CI, I ran it against a live Dataproc Serverless project (Airflow 3.2.1, google provider built from this branch). To demonstrate the retry behaviour rather than just a green run, I cleared create_batch_with_prefix and let it make a second attempt:

attempt 1 submitted …-b3678bb9
attempt 2 submitted …-0bf374be
Same prefix, different generated suffix, both batches created and completed successfully, and both cleaned up by the delete task. On main that second attempt is precisely the case that fails with ALREADY_EXISTS.

Screenshots: the run overview with both tasks green, and the attempt 1 / attempt 2 logs showing the two different generated IDs.
batch_id_prefix_grid
redacted_attempt1_log
redacted_attempt2_log

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -2536,7 +2545,7 @@ def execute(self, context: Context):
                 region=self.region,
                 project_id=self.project_id,
                 batch=self.batch,
-                batch_id=self.batch_id,
+                batch_id=requested_batch_id,
```
shahar1, line 2548: A fresh `batch_id` does not ensure a fresh batch when `request_id` is also supplied. Each task attempt still sends the same `self.request_id`, and [Dataproc documents](https://docs.cloud.google.com/managed-spark/docs/reference/rest/v1/projects.locations.batches/create) that duplicate request IDs return the original operation. The code below then reads the old batch ID from that operation, so a retry can return to the previous failed batch without submitting new work.
Could we reject batch_id_prefix combined with request_id, or generate a request ID per task attempt while keeping it stable across network retries? Please also cover this interaction in a retry test.

---

Drafted by Codex with my supervision

shahar1 (changes requested):
Thanks for fixing, there's some more work to do :)
One blocker to handle + could you please ensure that all cases are covered in unit tests?

PR #67016, opened 2026-05-15
add WasbDagBundle to load Dags from Azure Blob Storage

Nishieee:
## Summary

Adds a Dag bundle for **Azure Blob Storage** so Dags can be loaded from a container (with optional prefix), similar to `S3DagBundle` and `GCSDagBundle`. Introduces `WasbDagBundle`, extends `WasbHook` with container checks and `sync_to_local_dir`, registers the bundle in provider metadata, documents it in `dag-bundles.rst`, and adds unit tests.

**Manual verification:** Tested with Breeze against a real Azure storage account: `wasb` connection, `dag_processor.dag_bundle_config_list` pointing at `WasbDagBundle`, Dag parsed from blob and visible in the UI. PR includes screenshots (Azure container + Airflow Dags list).

closes: #66987

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Cursor (agent-assisted editing)

Screenshot 2026-05-15 at 7 12 30 PM
Screenshot 2026-05-15 at 6 52 48 PM
Screenshot 2026-05-15 at 6 52 39 PM

Generated-by: Cursor following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @Nishieee** · by `@potiuk` · 2026-07-08 15:51 UTC
>
> Some review feedback from `@eladkal` is waiting on you (4 unresolved threads):
>
> **The ball is in your court** — you've been assigned to this PR. Reply or push a fix in each thread, then mark them resolved. See the [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #69859, opened 2026-07-14
Centralise structlog initialisation in BaseDagBundle

dabla:
`BaseDagBundle` now owns the structlog logger via a lazy `_log` property, so subclasses no longer need to construct and bind it themselves in their constructors.  I saw all Dag bundles extending this class re-implemented the same logic regarding the structlog initialization when reviewing PR [#67016](https://github.com/apache/airflow/pull/67016), hence why I created this PR to make the code more DRY.

### What changed

- `BaseDagBundle` gains a `_log` lazy property that creates a structlog
  logger on first access, bound with `bundle_name`, `version`, and
  whatever `_log_context()` returns.
- A `_log_context(self) -> dict` template-method hook (returns `{}` by
  default) lets each subclass declare its own extra log fields without
  duplicating the boilerplate `structlog.get_logger(...).bind(...)` call.
- `GCSDagBundle`, `S3DagBundle`, and `GitDagBundle` each implement
  `_log_context()` and drop their manual `self._log = log.bind(...)` init
  blocks and the now-unused module-level / local `structlog.get_logger`
  calls.
- `LocalDagBundle` needs no change — the default `_log_context()` returns
  `{}`, which is the correct behaviour for a bundle with no extra context.

### Why lazy?

`BaseDagBundle.__init__` is called via `super().__init__()` *before* subclass-specific attributes (e.g. `bucket_name`, `bare_repo_path`) are set. Evaluating `_log_context()` at base-`__init__` time would therefore fail or produce incomplete context. The lazy property defers evaluation until first use, by which point all `__init__` work across the hierarchy is complete.

---

- [ x ] Yes (please specify the tool below)

Claude Sonnet 4.6

---


airflow-core/src/airflow/dag_processing/bundles/base.py
```diff
@@ -350,6 +352,19 @@ def initialize(self) -> None:
...
+        return {}
+
+    @property
+    def _log(self):
+        """Lazy structlog logger bound with common bundle context plus subclass-specific fields."""
+        if self.__log is None:
+            self.__log = structlog.get_logger(type(self).__module__).bind(
+                bundle_name=self.name, version=self.version, **self._log_context()
+            )
+        return self.__log
```
shahar1, line 366: **Blocking** — this makes `_log` a read-only data descriptor, so any subclass that still does `self._log = ...` in `__init__` now raises `AttributeError: property '_log' of 'X' object has no setter`.

That's not hypothetical: it's exactly what the currently-released `git`, `amazon`, and `google` providers do. This PR removes those assignments from sources, but released wheels pin only `apache-airflow>=3.0.0`, so `airflow-core` from `main` + today's released git provider = `GitDagBundle` unusable. Same for any third-party bundle, since `BaseDagBundle` is a public extension point.

A setter keeps the lazy default and every existing subclass working:

```suggestion
    @property
    def _log(self):
        """Lazy structlog logger bound with common bundle context plus subclass-specific fields."""
        if self.__log is None:
            self.__log = structlog.get_logger(type(self).__module__).bind(
                bundle_name=self.name, version=self.version, **self._log_context()
            )
        return self.__log

    @_log.setter
    def _log(self, value) -> None:
        """Allow subclasses to bind their own logger (kept for bundles released before the lazy property)."""
        self.__log = value
```

Worth a test that instantiating a bundle which assigns `self._log` still works — that's the case that regresses.

airflow-core/src/airflow/dag_processing/bundles/base.py
```diff
@@ -350,6 +352,19 @@ def initialize(self) -> None:
                 bundle_path,
             )
 
+    def _log_context(self) -> dict:
```
shahar1: Two nits on this hook:

1. **Name** — AGENTS.md asks for verbs on methods: *"Name functions and methods with action verbs: `get_`, `extract_`, `find_`, `compute_`, `build_`, etc. Avoid noun-only names like `_serialize_keys` or `_base_names` — they read as attributes, not callables."* `_get_log_context` or `_build_log_context` would fit. Cheap to do now, before the hook has implementers outside this repo.

2. **Annotation** — `-> dict[str, Any]` (`Any` is already imported here) is more precise than bare `dict`, and the `_log` property below has no return annotation.

Related: `self.__log = None` at line 313 types the attribute as `None`; the assignment at 363 only type-checks because `structlog.get_logger()` returns `Any`. An explicit `self.__log: Any = None` (or a `BindableLogger | None`) makes the intent survive a future stricter stub.

providers/common/compat/tests/unit/common/compat/bundles/test_dag_bundle.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Simulate an older Airflow version by temporarily hiding _log from
+        # the core class; the compat class must still expose it.
+        with mock.patch.object(CoreBaseDagBundle, "_log", None, create=True):
+            from importlib import reload
+
+            import airflow.providers.common.compat.bundles as mod
+
+            reload(mod)
+            CompatBase = mod.BaseDagBundle
```
shahar1: This `reload()` leaks. When the `mock.patch.object` block exits, `CoreBaseDagBundle._log` is restored — but `sys.modules["airflow.providers.common.compat.bundles"]` still holds the reloaded module, so its `BaseDagBundle` stays the compat fallback subclass for the remainder of the session. Anything importing it later (including `test_base_dag_bundle_is_subclass_of_core` in this class, under `-p no:randomly` reordering or `--lf`) silently exercises the wrong branch.

Reloading back once the patch is lifted keeps it contained:

```python
@pytest.fixture
def legacy_compat_base():
    import airflow.providers.common.compat.bundles as mod

    with mock.patch.object(CoreBaseDagBundle, "_log", None, create=True):
        reload(mod)
        yield mod.BaseDagBundle
    reload(mod)
```

The `yield` inside the patch also lets the dummy subclass be built while the fallback is genuinely active, which is closer to what the test claims to check.

providers/common/compat/tests/unit/common/compat/bundles/test_dag_bundle.py
```diff
@@ -0,0 +1,129 @@
...
+        # structlog bound loggers expose their bindings via _context
+        ctx = log._context if hasattr(log, "_context") else {}
+        assert ctx.get("bundle_name") == "my-bundle"
+        assert ctx.get("version") == "abc123"
+
+    def test_compat_subclass_provides_log_when_missing(self, tmp_path):
+        """Even when the installed BaseDagBundle has no _log, the compat class fills it in."""
+        from pathlib import Path
+
+        from tests_common.test_utils.config import conf_vars
```
shahar1: Function-level imports — `.github/instructions/code-review.instructions.md`: *"**Flag any `from` or `import` statement inside a function or method body.** Imports must be at the top of the file. The only valid exceptions are: (1) circular import avoidance, (2) lazy loading for worker isolation, (3) `TYPE_CHECKING` blocks."*

`pathlib.Path`, `tests_common.test_utils.config.conf_vars`, and `importlib.reload` are all safe at module level — the `pytest.importorskip` at line 23 already guards the Airflow-3-only imports below it, and none of these three depend on it.

Same applies to the copies in `test_log_property_on_subclass_instance` and `test_log_is_bound_with_context`. While you're here: `_DummyBundle` is defined three times, identical apart from its base class — a single module-level factory would drop roughly half this file.

shahar1 (changes requested):
## AI Summary
Nice cleanup — the duplication across the three bundles was real and the lazy property is the right shape for it. The `# use next version` markers on the three `pyproject.toml`s are exactly per [`contributing-docs/13_airflow_dependencies_and_extras.rst`](https://github.com/apache/airflow/blob/main/contributing-docs/13_airflow_dependencies_and_extras.rst), so those are correct as-is.

One blocking issue though: turning `_log` into a read-only property breaks every **already-released** bundle provider when users upgrade core without upgrading providers.

### Blocking — `_log` read-only property breaks released bundle providers (`airflow-core/src/airflow/dag_processing/bundles/base.py:359`)

`BaseDagBundle._log` is today a plain instance attribute, and every bundle subclass assigns it in `__init__`. This PR removes those assignments **in sources**, but the already-released `apache-airflow-providers-git`, `-amazon`, and `-google` still contain:

```python
self._log = log.bind(
    bundle_name=self.name,
    version=self.version,
    ...
)
```

Providers are released independently of core and pin only `apache-airflow>=3.0.0` — no upper bound. So a user on airflow-core `main` + the currently-released git provider gets:

```text
AttributeError: property '_log' of 'GitDagBundle' object has no setter
```

…at `GitDagBundle.__init__`, i.e. the Git/S3/GCS Dag bundle stops working entirely on core upgrade. The `common.compat` shim added in this PR covers the *new-provider-on-old-core* direction; this is the opposite direction and nothing covers it. `BaseDagBundle` is also a documented extension point, so the same break hits any third-party bundle that sets `self._log`.

Adding a setter keeps the lazy default and preserves every existing subclass — suggestion inline.

### Missing test for the new behaviour in `airflow-core`

> **Flag any changed or added behaviour without a corresponding test, and flag tests that a reviewer cannot fail by reverting the PR's change.** The target is exactly 100% coverage of what the PR changes — no more, no less.
>
> — [`.github/instructions/code-review.instructions.md`](https://github.com/apache/airflow/blob/main/.github/instructions/code-review.instructions.md#testing-requirements)

The new `_log` property and the `_log_context()` hook land in `airflow-core`, but the only tests are in `providers/common/compat/tests/`. Nothing in `airflow-core/tests/unit/dag_processing/bundles/test_base.py` covers them, and — more importantly — **no test anywhere asserts that a subclass's `_log_context()` is merged into the logger bindings**, which is the whole point of the PR. `test_log_is_bound_with_context` only checks `bundle_name` and `version`, both of which the base already supplied. A subclass overriding `_log_context()` and asserting its keys land in the bindings would fail without this change; today nothing would.

A setter (per the blocking finding) deserves a test too — instantiating a bundle that assigns `self._log` should keep working.

### Test isolation — `reload()` leaks into the rest of the session

`test_compat_subclass_provides_log_when_missing` reloads `airflow.providers.common.compat.bundles` inside the `mock.patch.object` block. When the patch exits, `CoreBaseDagBundle._log` is restored — but `sys.modules` still holds the reloaded module, whose `BaseDagBundle` is now permanently the compat subclass. Any later test importing it in the same session gets the fallback class instead of the modern one. Details inline.

### Smaller observations

- `airflow-core/src/airflow/dag_processing/bundles/base.py:355` — `_log_context` is a noun-only name. AGENTS.md: *"Name functions and methods with action verbs: `get_`, `extract_`, `find_`, `compute_`, `build_`, etc. Avoid noun-only names like `_serialize_keys` or `_base_names` — they read as attributes, not callables."* `_get_log_context` / `_build_log_context` reads better, and it's cheap to rename now while the hook has no external implementers.
- `base.py:355` — `-> dict` would be better as `dict[str, Any]`; the `_log` property has no return annotation, and `self.__log = None` leaves the attribute typed as `None` (mypy only lets the later assignment through because `structlog.get_logger()` returns `Any`).
- `providers/common/compat/tests/unit/common/compat/bundles/test_dag_bundle.py` — `_DummyBundle` is defined three times, identically apart from the base class. A module-level factory (or a fixture parametrised on the base) would remove ~60 of the 129 lines.
- `test_dag_bundle.py:88` — asserting on `log._context` reaches into structlog's private API. `structlog.testing.capture_logs()` gives the same signal against a public surface.
- Commit messages use Conventional Commits prefixes (`refactor: …`). AGENTS.md: *"do **not** use Conventional Commits prefixes (`fix:`, `feat:`, `chore:`, `docs:`, `refactor:`, …). apache/airflow does not follow that convention."* Cosmetic given the squash-merge takes the PR title, which is fine as-is.

---

> confirmed by an Airflow maintainer. After you've
> addressed the points above and pushed an update, an Airflow
>

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @shahar1 before posting

uranusjr:
I’m not personally a fan of attaching a logger to the class. Logger objects belong to modules for me. Automatically naming the logger after the class name is also unnecessary cruft in my opinion.

dabla:
> **[uranusjr](/uranusjr) ** commented

The approach is essentially the same philosophy as `LoggingMixin.log`, a lazy property that defers creation to first access, except it also binds bundle-specific structured context fields. A plain module-level logger could work but would require every single `self.log.info(...) `call site to manually pass `bundle_name=`, `bucket_name=`, etc. as keyword args, which is exactly the repetition the PR is trying to eliminate.

PR #69818, opened 2026-07-13
Add Dag version diff tooling

ephraimbuddy:
## Why

Dag version history preserves multiple serialized states, but operators do not have a supported way to answer what changed between two versions. Provide a deterministic comparison that helps diagnose version growth without requiring users to download and manually inspect serialized Dag payloads.

Keep the result honest about Airflow's persistence boundary. Treat every comparison as an observed-state diagnostic of the data currently stored for each version, not as creation-time evidence. Report missing, malformed, or unsupported historical data as unavailable instead of treating it as an empty diff.

This implements the shared service, CLI, and public API. Keep the UI panel as a later follow-up.

## What changes

- Introduce a shared, versioned diff service for stored Dag versions.
- Normalize supported serialized Dag schema versions before comparison, including client defaults and order-insensitive collections, and produce deterministic JSON Pointer-style paths.
- Classify changes by operation, category, and advisory impact; include SHA-256 digests by default and optional raw values when permitted.
- Bound each response with `max_changes`, report truncation, and degrade unsafe canonicalization or missing payloads to an explicit unavailable result.
- Add `airflow dags versions diff` with explicit version selection or `--previous-to-latest`, optional values and source, configurable limits, and table, plain, JSON, or YAML output.
- Add `GET /api/v2/dags/{dag_id}/dagVersions/{base_version_number}/diff/{target_version_number}` and the corresponding response models.
- Require Dag version access for structured changes. Gate optional raw values against each version's stored bundle and team, and additionally enforce co-located Dag source permissions before loading or returning source code.
- Report source fidelity as current stored code, redacted, or unavailable. Do not present mutable `DagCode` rows as immutable historical snapshots.
- Regenerate the OpenAPI specification, UI API clients, AirflowCTL datamodels, and API permission reference.

## Compatibility and scope

- Keep the change additive and read-only; require no metadata database migration.
- Leave Dag version creation, serialized Dag hashing, scheduler behavior, Dag processor behavior, worker bundle selection, and clear/rerun/backfill semantics unchanged.
- Preserve existing Dag version API responses.
- Add no user-facing UI in this PR; the UI changes are generated API client updates only, so screenshots do not apply.

## Testing

[...]

PR #69864, opened 2026-07-14
Add serialized Dag version diff engine

ephraimbuddy:
Dag authors and operators need to understand how stored versions differ without manually comparing serialized payloads. This PR adds an observed-state diff engine for serialized Dag state and bundle provenance.

This is the first part of a three-PR stack. The REST API endpoint and `airflow dags versions diff` CLI command follow in separate PRs; this PR provides the shared engine and model helper.

### Behavior

- `build_serialized_dag_diff` produces deterministic, categorized changes across supported serialized schema versions (v1–v3). It normalizes defaults, mapped-task precedence, dependencies, and deadline definitions. `max_changes` limits the returned change count and reports truncation.
- Including raw values preserves the change set and truncation semantics. Unknown task and TaskGroup fields are aggregated into one `custom_fields` record per task or group.
- `DagVersion.get_diff` loads the two stored versions and their provenance, reconstructs referenced deadline definitions, and validates the change limit before querying the database.
- Raw values, digests, and value-derived paths are withheld by default. Callers must authorize disclosure of the full serialized payload, including access-control configuration, before passing `values_status="available"`. Every returned result includes `values.status`.
- SHA-256 digests fingerprint canonical diff content. Each digest accompanies its corresponding raw value under the same disclosure authorization; these digests are not used for password storage or verification.
- Missing serialized state, unsupported schema versions or client-default sections, and canonicalization or JSON encoding failures produce an `unavailable` result. Comparison implementation errors propagate rather than being mislabeled as payload failures.

---

- [X] Yes — GPT 5.6, Claude Code (Opus 4.8), and Codex (GPT-6)

Generated-by: Claude Code (Opus 4.8), GPT5.6, and Codex (GPT-6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

Drafted-by: Codex (GPT-6) (no human review before posting)

### DIFF
--- a/airflow-core/src/airflow/models/dag_version.py
+++ b/airflow-core/src/airflow/models/dag_version.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import hashlib
 import logging
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, get_args
 from uuid import UUID
 
 import sqlalchemy as sa
@@ -30,6 +30,7 @@ from sqlalchemy.orm import Mapped, joinedload, mapped_column, relationship
 
 from airflow._shared.timezones import timezone
 from airflow.dag_processing.bundles.manager import DagBundlesManager
+from airflow.exceptions import DagVersionNotFound
 from airflow.models.base import Base, StringID
 from airflow.utils.session import NEW_SESSION, provide_session
 from airflow.utils.sqlalchemy import UtcDateTime, with_row_locks
@@ -43,10 +44,8 @@ log = logging.getLogger(__name__)
 
 SourceStatus = Literal["current_stored_code", "redacted", "unavailable"]
 ValuesStatus = Literal["available", "unavailable"]
-
-
-class DagVersionNotFoundError(ValueError):
-    """Raised when one of the requested Dag versions does not exist."""
+_VALID_SOURCE_STATUSES = get_args(SourceStatus)
+_VALID_VALUES_STATUSES = get_args(ValuesStatus)
 
 
 class DagVersion(Base):
@@ -274,15 +273,24 @@ class DagVersion(Base):
         method returns any source content. API callers use ``values_status`` to suppress raw values
         when the user cannot access Dag code; the structural diff remains available in redacted form.
         """
+        # Keep this local to avoid the dag_version -> dag_version_diff -> serialized_objects cycle.
         from airflow.serialization.dag_version_diff import (
             DEFAULT_MAX_CHANGES,
             build_serialized_dag_diff,
         )
 
+        if source_status is not None and source_status not in _VALID_SOURCE_STATUSES:
+            raise ValueError(f"source_status must be one of {_VALID_SOURCE_STATUSES}, not {source_status!r}")
+        if values_status is not None and values_status not in _VALID_VALUES_STATUSES:
+            raise ValueError(f"values_status must be one of {_VALID_VALUES_STATUSES}, not {values_status!r}")
+
         if max_changes is None:
             max_changes = DEFAULT_MAX_CHANGES
         if base_version_number < 1 or target_version_number < 1:
             raise ValueError("Dag version numbers must be positive integers")
+
+        if source_status is None:
+            source_status = "current_stored_code" if include_source else "unavailable"
 
         query = (
             select(cls)
@@ -292,7 +300,7 @@ class DagVersion(Base):
             )
             .options(joinedload(cls.serialized_dag))
         )
-        if include_source and source_status not in {"redacted", "unavailable"}:
+        if include_source and source_status == "current_stored_code":
             query = query.options(joinedload(cls.dag_code))
 
         versions = {version.version_number: version for version in session.scalars(query).all()}
@@ -305,13 +313,13 @@ class DagVersion(Base):
             None,
         )
         if missing_version is not None:
-            raise DagVersionNotFoundError(
+            raise DagVersionNotFound(
                 f"The DagVersion with dag_id: `{dag_id}` and version_number: `{missing_version}` was not found"
             )
 
         base_version = versions[base_version_number]
         target_version = versions[target_version_number]
-        effective_include_values = include_values and values_status != "unavailable"
+        effective_include_values = include_values and values_status in {None, "available"}
         result = build_serialized_dag_diff(
             base_data=base_version.serialized_dag.data if base_version.serialized_dag else None,
             target_data=target_version.serialized_dag.data if target_version.serialized_dag else None,
@@ -327,7 +335,7 @@ class DagVersion(Base):
             base_version,
             target_version,
             include_source=include_source,
-            source_status=source_status or ("current_stored_code" if include_source else "unavailable"),
+            source_status=source_status,
         )
         return result
 
--- a/airflow-core/src/airflow/models/dag_version.py
+++ b/airflow-core/src/airflow/models/dag_version.py
@@ -271,7 +271,8 @@ class DagVersion(Base):
 
         ``source_status`` is supplied by callers that have an authorization context.  The CLI has
         operator-level authority and leaves it unset, while API callers calculate it before this
-        method returns any source content.
+        method returns any source content. API callers use ``values_status`` to suppress raw values
+        when the user cannot access Dag code; the structural diff remains available in redacted form.
         """
         from airflow.serialization.dag_version_diff import (
             DEFAULT_MAX_CHANGES,
--- a/airflow-core/src/airflow/models/dag_version.py
+++ b/airflow-core/src/airflow/models/dag_version.py
@@ -17,9 +17,10 @@
 
 from __future__ import annotations
 
+import hashlib
 import logging
 from datetime import datetime
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 from uuid import UUID
 
 import sqlalchemy as sa
@@ -39,6 +40,13 @@ if TYPE_CHECKING:
 
 
 log = logging.getLogger(__name__)
+
+SourceStatus = Literal["current_stored_code", "redacted", "unavailable"]
+ValuesStatus = Literal["available", "unavailable"]
+
+
+class DagVersionNotFoundError(ValueError):
+    """Raised when one of the requested Dag versions does not exist."""
 
 
 class DagVersion(Base):
@@ -242,6 +250,125 @@ class DagVersion(Base):
     def version(self) -> str:
         """A human-friendly representation of the version."""
         return f"{self.dag_id}-{self.version_number}"
+
+    @classmethod
+    @provide_session
+    def get_diff(
+        cls,
+        dag_id: str,
+        base_version_number: int,
+        target_version_number: int,
+        *,
+        include_values: bool = False,
+        include_source: bool = False,
+        max_changes: int | None = None,
+        source_status: SourceStatus | None = None,
+        values_status: ValuesStatus | None = None,
+        session: Session = NEW_SESSION,
+    ) -> dict[str, Any]:
+        """
+        Compare two versions of a Dag using their currently stored state.
+
+        ``source_status`` is supplied by callers that have an authorization context.  The CLI has
+        operator-level authority and leaves it unset, while API callers calculate it before this
+        method returns any source content.
+        """
+        from airflow.serialization.dag_version_diff import (
+            DEFAULT_MAX_CHANGES,
+            build_serialized_dag_diff,
+        )
+
+        if max_changes is None:
+            max_changes = DEFAULT_MAX_CHANGES
+        if base_version_number < 1 or target_version_number < 1:
+            raise ValueError("Dag version numbers must be positive integers")
+
+        query = (
+            select(cls)
+            .where(
+                cls.dag_id == dag_id,
+                cls.version_number.in_((base_version_number, target_version_number)),
+            )
+            .options(joinedload(cls.serialized_dag))
+        )
+        if include_source and source_status not in {"redacted", "unavailable"}:
+            query = query.options(joinedload(cls.dag_code))
+
+        versions = {version.version_number: version for version in session.scalars(query).all()}
+        missing_version = next(
+            (
+                version_number
+                for version_number in (base_version_number, target_version_number)
+                if version_number not in versions
+            ),
+            None,
+        )
+        if missing_version is not None:
+            raise DagVersionNotFoundError(
+                f"The DagVersion with dag_id: `{dag_id}` and version_number: `{missing_version}` was not found"
+            )
+
+        base_version = versions[base_version_number]
+        target_version = versions[target_version_number]
+        effective_include_values = include_values and values_status != "unavailable"
+        result = build_serialized_dag_diff(
+            base_data=base_version.serialized_dag.data if base_version.serialized_dag else None,
+            target_data=target_version.serialized_dag.data if target_version.serialized_dag else None,
+            base_provenance=_get_provenance(base_version),
+            target_provenance=_get_provenance(target_version),
+            include_values=effective_include_values,
+            max_changes=max_changes,
+        )
+        if include_values:
+            values_available = effective_include_values and result["mode"] == "observed_state"
+            result["values"] = {"status": "available" if values_available else "unavailable"}
+        result["source"] = _get_source_diff(
+            base_version,
+            target_version,
+            include_source=include_source,
+            source_status=source_status or ("current_stored_code" if include_source else "unavailable"),
+        )
+        return result
+
+
+def _get_provenance(version: DagVersion) -> dict[str, Any]:
+    return {
+        "bundle_name": version.bundle_name,
+        "bundle_version": version.bundle_version,
+        "version_data": version.version_data,
+    }
+
+
+def _get_source_diff(
+    base_version: DagVersion,
+    target_version: DagVersion,
+    *,
+    include_source: bool,
+    source_status: SourceStatus,
+) -> dict[str, Any]:
+    if not include_source:
+        return {"status": "unavailable", "fidelity": "unavailable"}
+    if source_status != "current_stored_code":
+        return {"status": source_status, "fidelity": source_status}
+
+    base_code = base_version.dag_code
+    target_code = target_version.dag_code
+    if base_code is None or target_code is None:
+        return {"status": "unavailable", "fidelity": "unavailable"}
+
+    base_source = base_code.source_code
+    target_source = target_code.source_code
+    return {
+        "status": "current_stored_code",
+        "fidelity": "current_stored_code",
+        "changed": base_source != target_source,
+        "base": {"digest": _get_source_digest(base_source), "content": base_source},
+        "target": {"digest": _get_source_digest(target_source), "content": target_source},
+    }
+
+
+def _get_source_digest(source: str) -> str:
+    return f"sha256:{hashlib.sha256(source.encode()).hexdigest()}"
 
 
 def _resolve_version_data(
--- a/airflow-core/src/airflow/serialization/dag_version_diff.py
+++ b/airflow-core/src/airflow/serialization/dag_version_diff.py
@@ -44,6 +44,80 @@ _KEYED_COLLECTION_PATHS = {
     ("dag", "dag_dependencies"),
     *_ORDER_INSENSITIVE_LIST_PATHS,
 }
+_CUSTOM_TASK_FIELDS_PATH_COMPONENT = "custom_fields"
+# This allowlist is part of diff schema v1. Serializer schema changes must not
+# silently change the paths visible to callers of the diff API.
+_DIFF_V1_PUBLIC_TASK_FIELDS = frozenset(
+    {
+        "__type",
+        "_disallow_kwargs_override",
+        "_expand_input_attr",
+        "_is_mapped",
+        "_is_sensor",
+        "_logger_name",
+        "_needs_expansion",
+        "_operator_extra_links",
+        "_task_display_name",
+        "_task_module",
+        "allow_nested_operators",
+        "depends_on_past",
+        "do_xcom_push",
+        "doc",
+        "doc_json",
+        "doc_md",
+        "doc_rst",
+        "doc_yaml",
+        "downstream_task_ids",
+        "email_on_failure",
+        "email_on_retry",
+        "end_date",
+        "execution_timeout",
+        "executor",
+        "executor_config",
+        "has_on_execute_callback",
+        "has_on_failure_callback",
+        "has_on_retry_callback",
+        "has_on_skipped_callback",
+        "has_on_success_callback",
+        "ignore_first_depends_on_past",
+        "inlets",
+        "is_setup",
+        "is_teardown",
+        "map_index_template",
+        "max_active_tis_per_dag",
+        "max_active_tis_per_dagrun",
+        "max_retry_delay",
+        "multiple_outputs",
+        "on_failure_fail_dagrun",
+        "outlets",
+        "owner",
+        "params",
+        "partial_kwargs",
+        "pool",
+        "pool_slots",
+        "priority_weight",
+        "queue",
+        "render_template_as_native_obj",
+        "retries",
+        "retry_delay",
+        "retry_exponential_backoff",
+        "start_date",
+        "start_from_trigger",
+        "start_trigger_args",
+        "task_id",
+        "task_type",
+        "template_ext",
+        "template_fields",
+        "template_fields_renderers",
+        "trigger_rule",
+        "ui_color",
+        "ui_fgcolor",
+        "wait_for_downstream",
+        "wait_for_past_depends_before_skipping",
+        "weight_rule",
+    }
+)
+_DIFF_V1_REDACTED_SCHEMA_TASK_FIELDS: frozenset[str] = frozenset()
 _REDACTED_RECURSIVE_MAPPING_PATHS = {
     (),
     ("dag",),
@@ -98,8 +172,8 @@ def build_serialized_dag_diff(
         return _unavailable(result, f"unsupported_serialized_dag_schema_version:{unsupported_versions[0]}")
 
     try:
-        base_document = _canonicalize_payload(base_data)
-        target_document = _canonicalize_payload(target_data)
+        base_document = _canonicalize_payload_v1(base_data)
+        target_document = _canonicalize_payload_v1(target_data)
     except (AttributeError, KeyError, OverflowError, TypeError, ValueError):
         return _unavailable(result, "serialized_dag_canonicalization_failed")
 
@@ -132,9 +206,10 @@ class _ChangeCollector:
         if len(self.changes) >= self.max_changes:
             return
 
-        category = _get_category(path)
+        public_path = _get_public_path(path)
+        category = _get_category(public_path)
         change: dict[str, Any] = {
-            "path": _format_path(path if self.include_values else _redact_path(path)),
+            "path": _format_path(path if self.include_values else public_path),
             "operation": operation,
             "category": category,
             "impact": _get_impact(category),
@@ -172,7 +247,7 @@ def _unavailable(result: dict[str, Any], reason: str) -> dict[str, Any]:
     return result
 
 
-def _canonicalize_payload(data: dict[str, Any]) -> dict[str, Any]:
+def _canonicalize_payload_v1(data: dict[str, Any]) -> dict[str, Any]:
     payload = copy.deepcopy(data)
     version = _get_schema_version(payload)
     if version is None:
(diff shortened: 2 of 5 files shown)


ephraimbuddy:
Extracted from https://github.com/apache/airflow/pull/69818 and reworked

potiuk (approved):

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,655 @@
...
+            raise ValueError("task entry is not an object")
+        encoded_task = OperatorSerialization._apply_defaults_to_encoded_op(
+            dict(task["__var"]), dict(client_defaults)
+        )
+        task_data = {**schema_defaults, **OperatorSerialization._upgrade_encoded_operator(encoded_task)}
+        if task_data.get("_is_mapped"):
+            partial_kwargs = task_data.get("partial_kwargs", {})
+            if not isinstance(partial_kwargs, Mapping):
+                raise ValueError("partial_kwargs is not an object")
+            effective_partial_kwargs = dict(task_defaults)
```
kaxil: I think the precedence is inverted here for mapped tasks. `task_data` already has `client_defaults.tasks` merged underneath it by `_apply_defaults_to_encoded_op` above, so seeding from `task_defaults` puts the client default in first, and the `setdefault` on line 367 can then never replace it with the explicit top-level value. `populate_operator` resolves the other way: `partial_kwargs` wins, then the top-level value, then the client default.

The effect is that real changes go missing rather than showing up wrong. With `client_defaults.tasks = {"retries": 5}` and two mapped-task payloads differing only in top-level `retries: 2` vs `retries: 3`, `build_serialized_dag_diff` returns an empty `changes` list, while `populate_operator` resolves 2 and 3. `pool` behaves the same way.

`effective_partial_kwargs = dict(partial_kwargs)` (dropping the `task_defaults` seed and the `update`) matched `populate_operator` on every case I tried, and the whole `test_dag_version_diff.py` suite still passes with it, which suggests a gap: the mapped-default tests only shadow through `partial_kwargs`, never through a top-level field.

airflow-core/src/airflow/models/dag_version.py
```diff
@@ -245,6 +255,179 @@ def version(self) -> str:
...
+        if values_status is not None and values_status not in _VALID_VALUES_STATUSES:
+            raise ValueError(f"values_status must be one of {_VALID_VALUES_STATUSES}, not {values_status!r}")
+
+        if max_changes is None:
+            max_changes = DEFAULT_MAX_CHANGES
+        if base_version_number < 1 or target_version_number < 1:
+            raise ValueError("Dag version numbers must be positive integers")
+
+        if source_status is None:
+            source_status = "current_stored_code" if include_source else "unavailable"
```
kaxil: This defaults to full disclosure when a caller doesn't pass `source_status`, and `values_status=None` on line 330 does the same for raw values. An API handler that wires `include_source` to a query param but forgets to compute `source_status` would serve DAG source to whoever can reach the endpoint, and nothing in here would catch it.

Given the REST endpoint is the next PR in the stack, is it worth making `source_status` required and having the CLI pass `"current_stored_code"` explicitly? Omitting it becomes a TypeError rather than a disclosure. This is separate from the validation point in #discussion_r3580125653, which was about mistyped status values rather than the default direction.

airflow-core/src/airflow/models/dag_version.py
```diff
@@ -245,6 +255,179 @@ def version(self) -> str:
...
+            include_values=effective_include_values,
+            max_changes=max_changes,
+        )
+        if unavailable_reason := base_unavailable_reason or target_unavailable_reason:
+            result.update(
+                mode="unavailable", unavailable_reason=unavailable_reason, changes=[], truncated=False
+            )
+        if include_values:
+            values_available = effective_include_values and result["mode"] == "observed_state"
+            result["values"] = {"status": "available" if values_available else "unavailable"}
```
kaxil: `values` only shows up when `include_values` is set, while `source` is always present with `status: "unavailable"` when it's off. That means `result["source"]["status"]` is always safe to read but `result["values"]["status"]` raises KeyError on a redacted request. Emitting `values` unconditionally would keep the `diff_schema_version: 1` shape stable for clients.

airflow-core/src/airflow/models/dag_version.py
```diff
@@ -245,6 +255,179 @@ def version(self) -> str:
...
+            build_serialized_dag_diff,
+        )
+
+        if source_status is not None and source_status not in _VALID_SOURCE_STATUSES:
+            raise ValueError(f"source_status must be one of {_VALID_SOURCE_STATUSES}, not {source_status!r}")
+        if values_status is not None and values_status not in _VALID_VALUES_STATUSES:
+            raise ValueError(f"values_status must be one of {_VALID_VALUES_STATUSES}, not {values_status!r}")
+
+        if max_changes is None:
+            max_changes = DEFAULT_MAX_CHANGES
```
kaxil, line 289: `max_changes` isn't range-checked until `build_serialized_dag_diff` runs, so an out-of-range value pays for the version lookup and both serialized payload loads before raising. Calling `_validate_max_changes` here instead would reject it before any DB work, which will matter more once `max_changes` comes in as a query param.

kaxil:
One note that doesn't fit inline: the description says the diff engine "is fully exercised by unit tests." The precedence issue I flagged in `_canonicalize_payload_v1` got through because the mapped-task default tests only shadow client defaults via `partial_kwargs`. No case sets a partial-only field at the task's *top level*, so that resolution path has no coverage at all. The one-line fix I tried passed all 186 existing tests, which is how I noticed the gap rather than the bug.

Probably worth a parametrized case that sets `retries`/`pool` at the top level with a conflicting `client_defaults.tasks` value, and softening the claim in the description.

ephraimbuddy:
> One note that doesn't fit inline: the description says the diff engine "is fully exercised by unit tests." The precedence issue I flagged in `_canonicalize_payload_v1` got through because the mapped-task default tests only shadow client defaults via `partial_kwargs`. No case sets a partial-only field at the task's _top level_, so that resolution path has no coverage at all. The one-line fix I tried passed all 186 existing tests, which is how I noticed the gap rather than the bug.
> 
> Probably worth a parametrized case that sets `retries`/`pool` at the top level with a conflicting `client_defaults.tasks` value, and softening the claim in the description.

Resolved alongside other reviews

ephraimbuddy:
One note here @kaxil @potiuk , I removed source diff since that is already in the UI. This is now only the serialized diff

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,752 @@
...
+    after: Any,
+    *,
+    path: tuple[str, ...],
+    collector: _ChangeCollector,
+) -> None:
+    if isinstance(before, Mapping) and isinstance(after, Mapping):
+        if not collector.include_values and _is_task_mapping_path(path):
+            _collect_redacted_task_changes(before, after, path=path, collector=collector)
+            return
+        if not collector.include_values and not _should_recurse_redacted_mapping(path):
```
kaxil, line 564: These two guards gate how deep the walk goes, not just whether values get attached, so `include_values` changes the change set itself rather than only redacting it. On two versions of a 100-task Dag where every task's 20-key `executor_config` changed, `max_changes=500` gives 100 changes and `truncated=False` when redacted, but 500 changes, `truncated=True` and only 25 of the 100 tasks represented once values are authorized; the other 75 never appear at any path. So `max_changes` means two different things depending on a flag the client also sets, and the more-authorized caller gets the less complete answer, which contradicts `get_diff`'s docstring on line 271 ("leaving the structural diff available in redacted form"). Cheaper to settle here than after the endpoint in the next PR fixes it into a public API: compute the change set once with the redacted recursion shape, then attach values and digests to those same records when authorized.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,752 @@
...
+    DagSerialization,
+    OperatorSerialization,
+)
+
+log = structlog.get_logger(__name__)
+
+DIFF_SCHEMA_VERSION = 1
+DEFAULT_MAX_CHANGES = 500
+MAX_ALLOWED_CHANGES = 5000
+SUPPORTED_SERIALIZED_DAG_SCHEMA_VERSIONS = frozenset((1, 2, 3))
```
kaxil, line 46: This is the only schema-derived constant in the module without a parity test, while `_DIFF_V1_PUBLIC_TASK_FIELDS`, `_DIFF_V1_DAG_FIELD_CATEGORIES`, `_DAG_CALLBACK_FIELDS` and `_OPERATOR_TIMEDELTA_FIELDS` each have one. The day `SERIALIZER_VERSION` becomes 4, every stored row comes back `unsupported_serialized_dag_schema_version:4` for every Dag in the deployment and the suite stays green. `frozenset(range(1, DagSerialization.SERIALIZER_VERSION + 1))`, or a sixth parity test alongside the others, would close it.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,752 @@
...
+
+
+def _apply_task_defaults(payload: dict[str, Any]) -> None:
+    client_defaults = payload.pop("client_defaults", None)
+    if client_defaults is None:
+        client_defaults = {}
+    if not isinstance(client_defaults, Mapping):
+        raise ValueError("client_defaults is not an object")
+
+    task_defaults = client_defaults.get("tasks", {})
```
kaxil, line 380: The whole `client_defaults` section gets popped but only `tasks` is folded back in, so anything else in there is dropped before the comparison rather than rejected. Two payloads differing only in `client_defaults["dags"]["catchup"]` come back `observed_state` with `changes: []`. Given `client_defaults` is the growth mechanism v3 introduced, raising on unrecognised keys the way the two checks just above already raise on bad types would stop the diff going quiet the first time a new section lands.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,752 @@
...
+def _build_diff_result(base_schema_version: int | None, target_schema_version: int | None) -> dict[str, Any]:
+    return {
+        "diff_schema_version": DIFF_SCHEMA_VERSION,
+        "serialized_dag_schema_versions": {
+            "base": base_schema_version,
+            "target": target_schema_version,
+        },
+        "mode": "observed_state",
+        "changes": [],
+        "truncated": False,
```
kaxil, line 343: `values` is attached by `DagVersion.get_diff`, not here, so both exported engine functions return a result without that key. The description says every returned result includes `values.status` and the comment in `get_diff` calls that read always safe, which holds for the model helper but not for anything calling `build_serialized_dag_diff` or `build_unavailable_dag_diff` directly, which is the natural route for the endpoint in the next PR. Emitting it from `_build_diff_result` would keep `diff_schema_version: 1` one shape across all three entry points.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,896 @@
...
+
+
+def _is_json_equal(before: Any, after: Any) -> bool:
+    # Compare canonical JSON, not Python ==, so a JSON type change (1 vs 1.0, True vs 1,
+    # False vs 0) registers as a change instead of collapsing under Python equality.
+    return _serialize_canonical_json(before) == _serialize_canonical_json(after)
+
+
+def _get_digest(value: Any) -> str:
+    return f"sha256:{hashlib.sha256(_serialize_canonical_json(value).encode()).hexdigest()}"
```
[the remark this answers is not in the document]
ephraimbuddy: SHA-256 here fingerprints canonical serialized diff content; this code is not used for password storage or verification. CodeQL traces a value from `Connection.to_dict()["password"]`, but hashing content that may contain a password does not make this a password-protection mechanism.

`_ChangeCollector.add()` only produces digests when `include_values=True`, and each digest is returned alongside its corresponding raw value. `DagVersion.get_diff()` defaults to `values_status="unavailable"`; callers must authorize disclosure of the full serialized payload before enabling values. Redacted responses omit both values and digests, which existing regression tests verify.

The digest therefore reveals no additional value to the recipient beyond the raw content already authorized for disclosure. A password KDF would not improve protection here; SHA-256 serves deterministic content fingerprinting. This is a false positive for `py/weak-sensitive-data-hashing`.

---
Drafted-by: Codex (GPT-6) (no human review before posting)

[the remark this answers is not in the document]
ephraimbuddy: SHA-256 provides deterministic content fingerprints here; it is not used for password storage or verification.
Digests accompany their raw values only when disclosure is authorized; redacted responses contain neither.
This appears to be a false positive for the password-hashing rule.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,898 @@
...
+}
+_KEYED_COLLECTION_PATHS = {
+    ("dag", "tasks"),
+    ("dag", "dag_dependencies"),
+    *_ORDER_INSENSITIVE_LIST_PATHS,
+}
+_CUSTOM_TASK_FIELDS_PATH_COMPONENT = "custom_fields"
+# This allowlist is part of diff schema v1. Serializer schema changes must not
+# silently change the paths visible to callers of the diff API.
+_DIFF_V1_PUBLIC_TASK_FIELDS = frozenset(
```
kaxil, line 60: This allowlist drops `python_callable_name`, which is the one live field the new `_upgrade_encoded_operator` split exists to preserve. `_preprocess_encoded_operator` strips exactly two fields, `python_callable_name` ("Only serves to detect function name changes") and `label` ("Shouldn't be set anymore"), and line 452 calls the new upgrade-only variant rather than the preprocess one so that signal survives into the payload. The allowlist then buries it in `custom_fields`. I checked in breeze: two versions of one Dag, same `task_id`, a `@task` function renamed from `extract_from_postgres` to `extract_from_snowflake`, `python_callable_name` the only key that differs. The diff reports one change, `/dag/tasks/*/custom_fields`, and never names the field, so a reviewer comparing two versions cannot see that a task's callable was swapped.

Five more emitted fields sit outside the allowlist the same way: `expand_input`, `op_kwargs_expand_input`, `resources`, `run_as_user`, `email` (`_operator_name` and `has_retry_policy` too). Changing `run_as_user` from `alice` to `root`, or `resources` from 1 cpu to 8, each reports a single `/dag/tasks/*/custom_fields`, both changed together are indistinguishable from either alone, and a mapped `.expand(bash_command=[...])` going from two commands to three reports only `custom_fields`, so what a mapped task expands over never shows up.

The parity test can't catch any of these, because `definitions.operator` is `additionalProperties: true`, so a field the serializer writes and the schema doesn't declare still passes set equality against `properties`. `_DIFF_V1_PUBLIC_TASK_GROUP_FIELDS` already carries `expand_input` for that reason, with a comment on its own parity test saying so, so this was closed on the group side and missed here. Pinning the task test to the key set of a real `to_dict()` payload would catch all of them; `get_serialized_fields()` alone misses `python_callable_name` and names `operator_extra_links` by its attribute rather than its serialized key `_operator_extra_links`.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,898 @@
...
+        )
+
+    try:
+        base_document = _canonicalize_payload_v1(base_data)
+        target_document = _canonicalize_payload_v1(target_data)
+        base_document["provenance"] = _canonicalize_value(dict(base_provenance or {}), path=("provenance",))
+        target_document["provenance"] = _canonicalize_value(
+            dict(target_provenance or {}), path=("provenance",)
+        )
+    except (AttributeError, KeyError, OverflowError, TypeError, ValueError) as error:
```
kaxil, line 277: `RecursionError` subclasses `RuntimeError`, so it falls outside this tuple and isn't `_JsonEncodingError` either, which makes it the one canonicalization failure that escapes instead of becoming `mode: "unavailable"`.

What makes a payload writable but not readable here is structural rather than a curiosity about pathological configs. `BaseOperator.__deepcopy__` in the Task SDK opens with `sys.setrecursionlimit(5000)`, a process-global raise that fires in the dag-processor as soon as any operator is deep-copied during parsing, while `SerializedBaseOperator` has no such hook, so the API server that will serve this diff stays at the default 1000. Measured in breeze: the parser stores an `executor_config` nested 2491 deep once that limit is raised, and 491 deep without it, while `build_serialized_dag_diff` starts raising at 244, dropping to 219 and 194 with 100 and 200 extra frames already on the stack, which a real request has. So rows the parser wrote happily come back as a 500 on the next PR's endpoint rather than the unavailable result every other failure mode here produces. Worth covering the `_collect_changes` call below the same way, since `json.dumps` recurses in C and `_serialize_canonical_json` only converts `TypeError`.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,898 @@
...
+        after: Any,
+    ) -> None:
+        self.count += 1
+        if len(self.changes) >= self.max_changes:
+            return
+
+        public_path = _get_public_path(path)
+        category = _get_category(public_path)
+        change: dict[str, Any] = {
+            "path": _format_path(path if self.include_values else public_path),
```
kaxil, line 330: Redacting in the path rather than in the walk lets the default response spend its whole budget on identical records. On a 600-task Dag where every task's `retries` and `start_date` changed, `max_changes=500` returns 500 change objects holding only 3 distinct values (250x `/dag/tasks/*/retries`, 249x `/dag/tasks/*/start_date`, 1x `/dag/start_date`) with `truncated: true`, while the same call with values authorized returns 500 distinct ones. Raising `max_changes` recovers nothing, because the duplicates scale with task count rather than with content.

This predates last round's walk-shape change (f755bc09 gave 2 distinct records out of 500), so it isn't a regression. But the endpoint in the next PR defaults to this mode, so is it worth collapsing equal public paths into one record carrying an occurrence count, so the default answer reads as 250 tasks changing `retries` instead of one object repeated 250 times?

airflow-core/src/airflow/serialization/dag_version_diff.py
ephraimbuddy: Fixed. All eight fields are now public: `python_callable_name`, `expand_input`, `op_kwargs_expand_input`, `resources`, `run_as_user`, `email`, `_operator_name` and `has_retry_policy`. A renamed `@task` callable now reports `/dag/tasks/*/python_callable_name` instead of an unnamed `custom_fields` record.

The test is pinned to a real `to_dict()` payload as you suggested, asserting the allowlist both ways so it cannot drift again — it caught `is_stub` sitting in the same position shortly after. It is still only as complete as the operator shapes the sweep builds, so a new shape needs a new task there, not just a new entry.

---
Drafted-by: Claude Code (Opus 5); reviewed by @ephraimbuddy before posting

airflow-core/src/airflow/serialization/dag_version_diff.py
ephraimbuddy: Fixed at both sites: `RecursionError` is in the canonicalization tuple, and `_collect_changes` has its own handler returning `serialized_dag_recursion_limit_exceeded`.

Your trace also exposed an adjacent hole. Reading `SerializedDagModel.data` decompresses and parses the row, and that read sat outside any boundary, so unparseable compressed JSON, truncated bytes (`zlib.error`) and deep nesting all escaped while the uncompressed equivalent degraded. It is now inside the boundary with a distinct `serialized_dag_decode_failed`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @ephraimbuddy before posting

airflow-core/src/airflow/serialization/dag_version_diff.py
ephraimbuddy: Done. Records sharing a public path and operation now merge into one carrying an `occurrence_count`. Your 600-task case at `max_changes=500`:

```
redacted:   3 records, truncated=False, {start_date: 1,
            tasks/*/retries: 600, tasks/*/start_date: 600}
authorized: 500 records, truncated=True
```

Grouping alone left the counts wrong (250/249, `truncated=True` with all three paths present) because the bound was applied before the grouping lookup. Counting now continues for a path already in the result, so counts are exact and `truncated` means only that an unseen path was dropped. The docstring contract and the misleading test name were corrected with it.

---
Drafted-by: Claude Code (Opus 5); reviewed by @ephraimbuddy before posting

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,1222 @@
...
+    # False vs 0) registers as a change instead of collapsing under Python equality.
+    return _serialize_canonical_json(before) == _serialize_canonical_json(after)
+
+
+def _get_digest(value: Any) -> str:
+    # SHA-256 provides a content fingerprint, not password protection.
+    # Stored payloads can carry escaped lone surrogates, which strict UTF-8 refuses to encode;
+    # surrogatepass keeps them digestible without altering the bytes of any other payload.
+    canonical_json = _serialize_canonical_json(value).encode("utf-8", errors="surrogatepass")
+    return f"sha256:{hashlib.sha256(canonical_json).hexdigest()}"
```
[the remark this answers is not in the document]
ephraimbuddy: SHA-256 provides deterministic content fingerprints here; it is not used for password storage or verification.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,1281 @@
...
+
+    def add(
+        self,
+        *,
+        path: tuple[str, ...],
+        operation: Literal["added", "removed", "changed"],
+        before: Any,
+        after: Any,
+    ) -> None:
+        self.count += 1
```
kaxil: `self.count += 1` sits above the merge early-return, so a repeat of an already-recorded public path spends budget without producing a record. That makes `max_changes` bound occurrences rather than records, and the redacted default then drops distinct paths.

Measured at this HEAD in breeze: 600 tasks all changing `retries` plus one `/dag/timezone` change, `max_changes=500`, redacted, returns a single record (`/dag/tasks/*/retries`, `occurrence_count` 600) with `truncated: true` and no timezone record at all. At 5000 tasks with `max_changes=5000`, which is `MAX_ALLOWED_CHANGES` and the most a caller may ask for, the answer is the same single record, so no permitted bound recovers the timezone change. The same 600-task case at `max_changes=5000` returns both paths with `truncated: false`, which is the shape line 112 promises ("bounds the result rather than the traversal").

I can see this is deliberate: lines 116-119 describe the counting continuing past the bound, and `test_build_diff_excludes_new_group_beyond_change_limit[False-2]` pins it. But it leaves the crowd-out that last round's grouping was asked to fix (#discussion_r4008627247, "the duplicates scale with task count rather than with content") in place, and harder to spot, because the result now reads as small rather than as repetitive. Moving `self.count += 1` below the repeat early-return makes the bound cap records. The redacted walk then stops truncating in practice, but its record count is bounded by the field allowlists rather than by task count, so that is a small set either way.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,1281 @@
...
+        base_document = _canonicalize_payload_v1(base_data)
+        target_document = _canonicalize_payload_v1(target_data)
+        base_document["provenance"] = _canonicalize_value(dict(base_provenance or {}), path=("provenance",))
+        target_document["provenance"] = _canonicalize_value(
+            dict(target_provenance or {}), path=("provenance",)
+        )
+    except (AttributeError, KeyError, OverflowError, RecursionError, TypeError, ValueError) as error:
+        log.warning(
+            "Serialized Dag diff canonicalization failed",
+            error_type=type(error).__name__,
```
kaxil, line 448: This logs the exception class and nothing else, so the messages added to make these failures loud do not survive. `_apply_task_defaults` raises `ValueError(f"unsupported client_defaults sections: {sorted(unknown_sections)}")` for exactly the case from two rounds ago, and `dag.tasks is not a list`, `task entry has no task_id` and `duplicate key ... in /dag/tasks` all land in this same handler. What an operator sees in every one of those cases is `Serialized Dag diff canonicalization failed base_schema_version=3 error_type=ValueError target_schema_version=3`.

The tuple also catches `AttributeError`, `KeyError` and `TypeError`, so an engine bug arrives here too and reads identically to a new serializer section. `exc_info=error`, or a `reason=str(error)` field, tells them apart.

airflow-core/src/airflow/serialization/dag_version_diff.py
```diff
@@ -0,0 +1,1281 @@
...
+    collector = _ChangeCollector(max_changes=max_changes, include_values=include_values)
+    try:
+        _collect_changes(base_document, target_document, path=(), collector=collector)
+    except RecursionError:
+        log.warning(
+            "Serialized Dag diff recursion limit exceeded",
+            base_schema_version=base_schema_version,
+            target_schema_version=target_schema_version,
+        )
+        return _mark_unavailable(result, "serialized_dag_recursion_limit_exceeded")
```
kaxil, line 464: I do not think this reason can reach a caller, and the mechanism I gave for it last round was wrong. Canonicalization runs first over a strict superset of what the walk recurses into, so it exhausts the stack first in every shape I could build. Swept at this HEAD in breeze at the default limit of 1000: a deeply nested `executor_config` returns `observed_state` through depth 488 and `serialized_dag_canonicalization_failed` from 490 on, and nested task groups, the one place the walk itself recurses without bound, are clean through depth 164 and `serialized_dag_canonicalization_failed` from 165 on. There is no band where the walk fails first. Both recursion tests mock the function they exercise (`_collect_changes` on line 1896, `_canonicalize_payload_v1` on line 1873), so neither observed which reason a real deep payload produces.

Deep payloads still degrade correctly, so this is about the wire contract rather than behaviour: line 53 lists `serialized_dag_recursion_limit_exceeded` among the reasons a client can receive, and the CLI and endpoint PRs will enumerate it from there. Either drop the handler with the bullet, or keep it and say in the docstring that it is defence in depth rather than a reachable reason.

kaxil:
Everything from my last round is addressed. I re-checked the three open comments at `733cfaa8` myself, so here is the whole set in one place with what it would take to close it out.

**One thing holds my approval: the counter at line 500.**

I applied the one-line change (move `self.count += 1` below the repeat early-return) and measured both sides in breeze at this HEAD. Redacted mode, a Dag where every task changes `retries` plus one `/dag/timezone` change:

| case | today | with the counter moved |
|---|---|---|
| 600 tasks, `max_changes=500` (the default) | `truncated: true`, 1 record, timezone absent | `truncated: false`, both paths, exact counts |
| 5000 tasks, `max_changes=5000` (`MAX_ALLOWED_CHANGES`) | `truncated: true`, 1 record, timezone absent | `truncated: false`, both paths, exact counts |

At the default bound a 600-task fan-out already hides the timezone change, and at 5000 tasks nothing a caller may pass recovers it, because 5000 is the ceiling. So for "bumped `retries` in `default_args` and changed the schedule" on a large Dag, the operator sees one record saying retries changed on `*` tasks plus `truncated: true`, and the schedule change is gone with nothing pointing at it. That is the shape the endpoint PR freezes.

The change costs one test parametrization. Engine plus model suites with it applied: `1 failed, 621 passed`, and the failure is `test_build_diff_excludes_new_group_beyond_change_limit[False-2]`, which under the change returns both records and `truncated: false`.

I think that test is the one to move, because the suite currently pins both answers. `test_build_diff_bound_withholds_records_not_public_paths` asserts `redacted["truncated"] is False` at `max_changes=1` for ten occurrences of a single path, with the comment "no bound can withhold a path from the caller or cost an occurrence". `test_build_diff_excludes_new_group_beyond_change_limit[False-2]` is that same situation with a second public path present, and there the bound does withhold one. Whether a later distinct path survives depends on how many occurrences the earlier paths happened to have, even though those occurrences produced no records.

Keeping the current counting is a reasonable call to make instead. In that case the docstring at lines 112-124 and the name of `test_build_diff_bound_withholds_records_not_public_paths` both need to say that a bound can withhold a public path, since today they say it cannot.

**The other two comments are optional and will not hold approval.** The canonicalization handler logging only `error_type` costs you the `unsupported client_defaults sections: [...]` message you added two rounds ago, and the unreachable `serialized_dag_recursion_limit_exceeded` is a wire-contract question the CLI and endpoint PRs will inherit from line 53. Take them or leave them.

Nothing else from me. I checked the reason list against every `_mark_unavailable` call site and the two `get_diff` adds (complete and accurate apart from the unreachable one), `get_diff` itself (one query with `joinedload`/`selectinload`, bound validated before the lookup, no writes), and the `serialized_objects.py` refactor (`_DAG_CALLBACK_FIELDS` iteration order cannot reach `dag_hash` or the stored blob, both of which go through `json.dumps(..., sort_keys=True)`). Settle the counter either way and I approve.

ephraimbuddy:
Good to merge @kaxil ?

kaxil (approved):
All three from last round are in. I reran the 600-task plus timezone case at `ccdfeada`: the timezone change now survives at the default bound and at the 5000 ceiling, and `test_dag_version_diff.py` plus `test_dag_version.py` pass. Thanks for sticking with this through the rounds.

PR #69869, opened 2026-07-14
Fix create_delta_data_intervals being ignored for timedelta schedules

### LINKED ISSUE #69868: create_delta_data_intervals is never read - timedelta/relativedelta schedules check create_cron_data_intervals instead

Task SDK

### Apache Airflow version

3.3.0 (bug confirmed present via source review; also reproduced hands-on on 3.2.2, and confirmed present on `main` as of 2026-07-14)

`airflow.cfg`'s `[scheduler] create_delta_data_intervals` setting is never read anywhere in the codebase - it has no effect. Meanwhile `[scheduler] create_cron_data_intervals` silently controls timetable selection for **both** cron-string schedules (documented) **and** `timedelta`/`relativedelta` schedules (undocumented, and contradicts `config.yml`'s own description of what `create_delta_data_intervals` is supposed to do).

Root cause is in `_create_timetable()` in `task-sdk/src/airflow/sdk/definitions/dag.py` (lines 148–169 on `main` / 3.3.0):

```python
if isinstance(interval, timedelta | relativedelta):
    if airflow_conf.getboolean("scheduler", "create_cron_data_intervals"):   # <-- should be create_delta_data_intervals
        return DeltaDataIntervalTimetable(interval)
    return DeltaTriggerTimetable(interval)
if isinstance(interval, str):
    if airflow_conf.getboolean("scheduler", "create_cron_data_intervals"):
        return CronDataIntervalTimetable(interval, timezone)
    return CronTriggerTimetable(interval, timezone=timezone)
```

Both branches check the same config key, `create_cron_data_intervals`. The `timedelta`/`relativedelta` branch should check `create_delta_data_intervals` instead, per `config.yml`'s own description of that setting (`airflow-core/src/airflow/config_templates/config.yml`, `create_delta_data_intervals` entry).

**Steps to reproduce**

1. Use any DAG scheduled with a plain `timedelta`, e.g. the stock `tutorial` DAG:
   ```python
   schedule=timedelta(days=1)
   ```
2. In `airflow.cfg`, set:
   ```
   [scheduler]
   create_delta_data_intervals = True
   ```
3. Confirm the config is actually being read:
   ```python
   from airflow.configuration import conf
   print(conf.getboolean("scheduler", "create_delta_data_intervals"))  # prints True
   ```
4. Load the DAG and inspect its timetable:
   ```python
   from airflow.models import DagBag
   dag = DagBag().get_dag("tutorial")
   print(type(dag.timetable))
   ```
   **Result:** still `DeltaTriggerTimetable` - `create_delta_data_intervals=True` had no effect.
5. Now instead set `create_cron_data_intervals = True` (leaving `create_delta_data_intervals` at its default) and rerun step 4.
   **Result:** `dag.timetable` becomes `DeltaDataIntervalTimetable`, even though the DAG's schedule is a `timedelta`, not a cron string - proving the wrong config key is being consulted.
   

The `timedelta | relativedelta` branch of `_create_timetable()` should check `create_delta_data_intervals`, matching what `config.yml` documents for that setting and keeping it independent from `create_cron_data_intervals` (which should only affect the `str`/cron branch). The fix is a one-line change:

```python
if isinstance(interval, timedelta | relativedelta):
    if airflow_conf.getboolean("scheduler", "create_delta_data_intervals"):
        return DeltaDataIntervalTimetable(interval)
    return DeltaTriggerTimetable(interval)
```

### Operating System

Windows 11 / WSL2 (Ubuntu), reproduced via a local virtualenv install

### Deployment

Virtualenv installation

_No response_

N/A

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

Confirmed by reading source directly (not just local install):
- Installed venv (Airflow 3.2.2): `venv/lib/python3.12/site-packages/airflow/sdk/definitions/dag.py`
- `github.com/apache/airflow` tag `3.2.2`: same code
- `github.com/apache/airflow` tag `3.3.0`: same code, lines 148–169
- `github.com/apache/airflow` branch `main`: same code, unchanged

Also confirmed `create_delta_data_intervals` is referenced nowhere else in the codebase except:
- `airflow-core/src/airflow/cli/commands/config_command.py` (CLI config-list metadata only)
- `config_templates/config.yml` (docs only)
- `config_templates/unit_tests.cfg` (sets it `true` for test fixtures, but no code path reads it for this purpose)

This makes `create_delta_data_intervals` fully dead config as of at least 3.2.2 through `main`.

### Code of Conduct

sharetheknowledge:
Fixes a bug where `create_delta_data_intervals` had no effect on DAGs with a
`timedelta`/`relativedelta` schedule.

`_create_timetable()` in `task-sdk/src/airflow/sdk/definitions/dag.py` checked
`create_cron_data_intervals` for **both** the cron-string branch and the
`timedelta`/`relativedelta` branch. As a result:

- `create_delta_data_intervals` had no effect at all - it's referenced nowhere
  else in the codebase besides CLI config-list metadata and docs.
- `create_cron_data_intervals` silently controlled timetable selection for
  `timedelta`/`relativedelta` schedules too, which isn't what it's documented
  to do (`config.yml` describes it as governing only cron-string schedules).

The `timedelta | relativedelta` branch now checks `create_delta_data_intervals`
instead, matching `config.yml`'s documented behavior and making the two config
keys independent, as intended.

### Tests

`test_timedelta_schedule_respects_create_delta_data_intervals_config` in
`task-sdk/tests/task_sdk/definitions/test_dag.py` is parametrized over
`(delta, cron, expected)` and over the schedule type:

- both `False` (the shipped default) -> `DeltaTriggerTimetable`
- `create_delta_data_intervals=True` -> `DeltaDataIntervalTimetable`
- `create_cron_data_intervals=True` alone -> still `DeltaTriggerTimetable`
  (regression guard against the exact bug fixed here)

each run against `timedelta(days=1)` and `relativedelta(days=1)`, for six
independent cases.

Verified locally: all six pass against the fix, and four of the six fail
against the pre-fix code (both `delta-true` rows and both `cron-true` rows,
including the `relativedelta` ones). Full `test_dag.py` suite: 99 passed.

Note: `unit_tests.cfg` sets both `create_cron_data_intervals` and
`create_delta_data_intervals` to `true`, which is why the existing test suite
never caught this - both keys being `true` produced the same (accidentally
correct) result before and after this fix for any test not overriding them
individually.

### Docs and newsfragment

The fix changes behavior for anyone who set `create_cron_data_intervals = True`
on the upgrade guide's advice, so this PR documents it in four places:

- `airflow-core/newsfragments/69869.significant.rst` - who is affected and the
  remedy.
- `upgrading_to_airflow3.rst` - a bullet next to the existing
  `create_cron_data_intervals` one.
- `timetable.rst` - `create_delta_data_intervals` added to "Switching between
  trigger and data interval timetables".
- `config.yml` - the skip-one-period paragraph the cron entry already had but
  the delta entry lacked.

Setting `create_delta_data_intervals = True` is a no-op before this fix, so it
can be set ahead of upgrading to keep current behavior.

* closes: #69868

---

* For significant user-facing changes create newsfragment: `{pr_number}.significant.rst`, in [airflow-core/newsfragments](https://github.com/apache/airflow/tree/main/airflow-core/newsfragments).

### DIFF
--- a/task-sdk/src/airflow/sdk/definitions/dag.py
+++ b/task-sdk/src/airflow/sdk/definitions/dag.py
@@ -147,7 +147,7 @@ def _create_timetable(interval: ScheduleInterval, timezone: Timezone | FixedTime
     if interval == "@continuous":
         return ContinuousTimetable()
     if isinstance(interval, timedelta | relativedelta):
-        if airflow_conf.getboolean("scheduler", "create_cron_data_intervals"):
+        if airflow_conf.getboolean("scheduler", "create_delta_data_intervals"):
             return DeltaDataIntervalTimetable(interval)
         return DeltaTriggerTimetable(interval)
     if isinstance(interval, str):
--- a/task-sdk/tests/task_sdk/definitions/test_dag.py
+++ b/task-sdk/tests/task_sdk/definitions/test_dag.py
@@ -447,6 +447,46 @@ class TestDag:
         with pytest.raises(ValueError, match="ContinuousTimetable requires max_active_runs <= 1"):
             dag = DAG("continuous", start_date=DEFAULT_DATE, schedule="@continuous", max_active_runs=25)
 
+    def test_timedelta_schedule_respects_create_delta_data_intervals_config(self):
+        """Regression guard: create_delta_data_intervals must control DeltaTriggerTimetable vs
+        DeltaDataIntervalTimetable for timedelta/relativedelta schedules, independently of
+        create_cron_data_intervals (which governs only cron-string schedules).
+        """
+        from airflow.sdk.definitions.timetables.interval import DeltaDataIntervalTimetable
+        from airflow.sdk.definitions.timetables.trigger import DeltaTriggerTimetable
+
+        from tests_common.test_utils.config import conf_vars
+
+        # Both False (the shipped airflow.cfg default): timedelta schedules use the trigger-based timetable.
+        with conf_vars(
+            {
+                ("scheduler", "create_delta_data_intervals"): "False",
+                ("scheduler", "create_cron_data_intervals"): "False",
+            }
+        ):
+            dag = DAG("delta_default", start_date=DEFAULT_DATE, schedule=timedelta(days=1))
+            assert isinstance(dag.timetable, DeltaTriggerTimetable)
+
+        # create_delta_data_intervals=True switches timedelta schedules to the data-interval timetable.
+        with conf_vars(
+            {
+                ("scheduler", "create_delta_data_intervals"): "True",
+                ("scheduler", "create_cron_data_intervals"): "False",
+            }
+        ):
+            dag = DAG("delta_data_interval", start_date=DEFAULT_DATE, schedule=timedelta(days=1))
+            assert isinstance(dag.timetable, DeltaDataIntervalTimetable)
+
+        # create_cron_data_intervals only affects cron/str schedules, not timedelta ones.
+        with conf_vars(
+            {
+                ("scheduler", "create_delta_data_intervals"): "False",
+                ("scheduler", "create_cron_data_intervals"): "True",
+            }
+        ):
+            dag = DAG("delta_unaffected_by_cron_config", start_date=DEFAULT_DATE, schedule=timedelta(days=1))
+            assert isinstance(dag.timetable, DeltaTriggerTimetable)
+
     def test_only_partitioned_at_runtime_has_partitioned_at_runtime_flag(self):
         """Regression guard: across every BaseTimetable subclass, only PartitionedAtRuntime sets partitioned_at_runtime=True."""
 
--- a/task-sdk/tests/task_sdk/definitions/test_dag.py
+++ b/task-sdk/tests/task_sdk/definitions/test_dag.py
@@ -23,6 +23,7 @@ from datetime import datetime, timedelta, timezone
 from typing import Any
 from unittest import mock
 
+from dateutil.relativedelta import relativedelta
 import pytest
 
 from airflow.sdk import (
@@ -37,6 +38,8 @@ from airflow.sdk import (
 )
 from airflow.sdk.bases.operator import BaseOperator
 from airflow.sdk.bases.timetable import BaseTimetable
+from airflow.sdk.definitions.timetables.interval import DeltaDataIntervalTimetable
+from airflow.sdk.definitions.timetables.trigger import DeltaTriggerTimetable
 from airflow.sdk.definitions.param import DagParam, ParamsDict
 from airflow.sdk.exceptions import AirflowDagCycleException, DuplicateTaskIdFound, RemovedInAirflow4Warning
 from airflow.utils.types import DagRunType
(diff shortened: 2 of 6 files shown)


task-sdk/src/airflow/sdk/definitions/dag.py
```diff
@@ -147,7 +147,7 @@ def _create_timetable(interval: ScheduleInterval, timezone: Timezone | FixedTime
     if interval == "@continuous":
         return ContinuousTimetable()
     if isinstance(interval, timedelta | relativedelta):
-        if airflow_conf.getboolean("scheduler", "create_cron_data_intervals"):
+        if airflow_conf.getboolean("scheduler", "create_delta_data_intervals"):
```
kaxil, line 150: The one-line fix is correct, but it silently changes behavior for the cohort our own upgrade guide created: `upgrading_to_airflow3.rst` tells 2.x migrants to set `create_cron_data_intervals = True` to keep data-interval semantics, and because of this bug that flag currently also pins `timedelta`/`relativedelta` DAGs to `DeltaDataIntervalTimetable`, which matches Airflow 2 behavior. After this change those DAGs fall through to `DeltaTriggerTimetable` on the next parse, so `logical_date` becomes the trigger time and `ds`, `ts` and `data_interval_*` all shift by one period (no run is skipped in that direction, but nothing warns the user). The same applies to Airflow 2 serialized DAGs converted through `conversion_v1_to_v2`, which rebuilds the timetable via this function.

Can you add a `69869.significant.rst` newsfragment naming who is affected and the remedy? Setting `[scheduler] create_delta_data_intervals = True` is a no-op on current main (the key is read nowhere), so users can safely set it before upgrading to keep current behavior. The docs need the same treatment in this PR: a bullet in `upgrading_to_airflow3.rst` next to the cron one, this key in the "Switching between trigger and data interval timetables" section of `timetable.rst`, and the skip-one-period paragraph the cron entry has in `config.yml` but the delta entry lacks (flipping this key from False to True after Airflow 3 runs exist skips one period, same collision guard).

task-sdk/tests/task_sdk/definitions/test_dag.py
```diff
@@ -447,6 +447,46 @@ def test_continuous_schedule_linmits_max_active_runs(self):
...
 
+    def test_timedelta_schedule_respects_create_delta_data_intervals_config(self):
+        """Regression guard: create_delta_data_intervals must control DeltaTriggerTimetable vs
+        DeltaDataIntervalTimetable for timedelta/relativedelta schedules, independently of
+        create_cron_data_intervals (which governs only cron-string schedules).
+        """
+        from airflow.sdk.definitions.timetables.interval import DeltaDataIntervalTimetable
+        from airflow.sdk.definitions.timetables.trigger import DeltaTriggerTimetable
+
+        from tests_common.test_utils.config import conf_vars
```
kaxil: Can you hoist this to the top of the file? The sibling files here (`test_connection.py`, `test_variables.py`, `test_operator_resources.py`) all import `conf_vars` at module level, and there is no circular-import reason for it to live in the function body.

task-sdk/tests/task_sdk/definitions/test_dag.py
```diff
@@ -447,6 +447,46 @@ def test_continuous_schedule_linmits_max_active_runs(self):
         with pytest.raises(ValueError, match="ContinuousTimetable requires max_active_runs <= 1"):
             dag = DAG("continuous", start_date=DEFAULT_DATE, schedule="@continuous", max_active_runs=25)
 
+    def test_timedelta_schedule_respects_create_delta_data_intervals_config(self):
```
kaxil: Consider `@pytest.mark.parametrize` over `(delta, cron, expected)` here: if the first assert fails, pytest never runs the other two scenarios, and this file already uses parametrize for similar multi-case checks. A `relativedelta` row would also back up the docstring, which claims relativedelta coverage while only `timedelta` is exercised.

task-sdk/src/airflow/sdk/definitions/dag.py
sharetheknowledge: Added all four. `69869.significant.rst` names who is affected - 2.x migrants who
set `create_cron_data_intervals = True` on the upgrade guide's advice, plus DAGs
rebuilt through `conversion_v1_to_v2` - and gives the remedy, including your
point that setting `create_delta_data_intervals = True` is a no-op until this
lands, so it can go in the config before upgrading.

The other three: a bullet in `upgrading_to_airflow3.rst` directly after the cron
one, `create_delta_data_intervals` added to the "Switching between trigger and
data interval timetables" list in `timetable.rst`, and the skip-one-period
paragraph on the `config.yml` delta entry. Also bumped "in two ways" to "three
ways" in that `timetable.rst` list, since it's now three bullets.

task-sdk/tests/task_sdk/definitions/test_dag.py
sharetheknowledge: Done - moved up to module level with the other `tests_common` import.

task-sdk/tests/task_sdk/definitions/test_dag.py
sharetheknowledge: Parametrized over `(delta, cron, expected)`, with a second stacked parametrize
for the schedule type so the three config combinations each run against
`timedelta(days=1)` and `relativedelta(days=1)`. Six independent cases now, so a
failure in one no longer hides the others.

The `relativedelta` rows earn their place: against the pre-fix code four of the
six fail (both `delta-true` rows and both `cron-true` rows), the `relativedelta`
ones included.

sharetheknowledge:
All three review comments are addressed:

- Newsfragment plus the three doc locations (`upgrading_to_airflow3.rst`,
  `timetable.rst`, `config.yml`)
- `conf_vars` hoisted to module level
- The test is parametrized over `(delta, cron, expected)` and over the schedule
  type, so the three config combinations each run against `timedelta` and
  `relativedelta`

PR #69870, opened 2026-07-14
Fail constraints generation when a released provider is downgraded

potiuk:
Regular `constraints` (PyPI-providers) generation now fails, instead of silently publishing, when a released `apache-airflow-providers-*` package would be downgraded below the version already in the published constraints. Released providers only ever move forward on PyPI, so a downgrade means highest-resolution dragged an old provider back in because a broken dependency constrained it — the exact failure mode behind the recent databricks-sql-connector / thrift issue.

The check compares each released provider against the latest published constraints and exits non-zero on any downgrade, so it surfaces in CI. On canary/push builds (where there is no PR author watching) it also posts an alert to the `#internal-airflow-ci-cd` Slack channel via the existing `slackapi/slack-github-action`, so maintainers are notified immediately. The in-container step only writes the Slack payload; the workflow layer holds the token and sends it (skipped on fork PRs where no secret is available).

---

### DIFF
--- a/.github/workflows/ci-amd.yml
+++ b/.github/workflows/ci-amd.yml
@@ -325,6 +325,8 @@ jobs:
       generate-no-providers-constraints: ${{ needs.build-info.outputs.canary-run }}
       debug-resources: ${{ needs.build-info.outputs.debug-resources }}
       use-uv: ${{ needs.build-info.outputs.use-uv }}
+    secrets:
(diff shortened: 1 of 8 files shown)


eladkal (approved):
This is great!
I do still think we need a validation step in the RM guide. I mean, RM can easily not see the open slack note and proceed with cutting release rather than check for resolution. cc @vatsrahul1001 WDYT?

bugraoz93 (approved):

PR #69873, opened 2026-07-14
Populate dag and note before firing dag run state-change listeners

kacpermuda:
When patching a Dag Run's state to success or failed via the API, two issues
existed with the listener call:
1. `dag_run.dag` was always `None` when the listener received the object, because the route loads the dag separately but never attached it to the run. After the change, the listener will have access to SerializedDag that is used for example by OpenLineage listener to create an AirflowDagRunFacet.
2. The listener message was an empty string. After the change, we will match the messages passed to task listener methods when manually changing task's state. It will make it easier for listener (and end users) to figure out what caused the early success / failure of dagrun.
3. When both `note` and `state` were patched in the same request (dagrun state was set and note was provided), the note could be applied after the listener fired, so the listener saw the old / no note. Fixed by sorting the patch fields so `note` is always processed before `state`.

---

- [X] Yes — Claude Sonnet 4.6
Generated-by: Claude Sonnet 4.6 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
@@ -222,15 +222,13 @@ def patch_dag_run(
 
     data = patch_body.model_dump(include=fields_to_update, by_alias=True)
 
-    # Sort keys so that the "note" update happens before "state" update, so that the listeners called inside
-    # `patch_dag_run_state()` already receive updated DagRun object with note.
-    for attr_name, attr_value_raw in sorted(data.items()):
-        if attr_name == "state" and patch_body.state is not None:
-            patch_dag_run_state(dag=dag, dag_run=dag_run, state=patch_body.state, session=session)
-        elif attr_name == "note":
-            updated_dag_run = session.get(DagRun, dag_run.id)
-            if updated_dag_run is not None:
-                patch_dag_run_note(dag_run=updated_dag_run, note=attr_value_raw, user=user)
+    # Apply "note" before "state" so listeners fired inside patch_dag_run_state() see the updated note.
+    if "note" in data:
+        updated_dag_run = session.get(DagRun, dag_run.id)
+        if updated_dag_run is not None:
+            patch_dag_run_note(dag_run=updated_dag_run, note=data["note"], user=user)
+    if "state" in data and patch_body.state is not None:
+        patch_dag_run_state(dag=dag, dag_run=dag_run, state=patch_body.state, session=session)
 
     final_dag_run = session.get(DagRun, dag_run.id)
     if not final_dag_run:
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
```diff
@@ -222,7 +222,9 @@ def patch_dag_run(
 
     data = patch_body.model_dump(include=fields_to_update, by_alias=True)
 
-    for attr_name, attr_value_raw in data.items():
+    # Sort keys so that the "note" update happens before "state" update, so that the listeners called inside
+    # `patch_dag_run_state()` already receive updated DagRun object with note.
+    for attr_name, attr_value_raw in sorted(data.items()):
```
amoghrajesh: Using `sorted(data.items())` to force "note" before "state" couples correctness to the field names' alphabetical order rather than expressing the ordering intent explicitly. This works today because "note" < "state", but is a trap for the next field added to `DAGRunPatchBody` & nothing would signal that ordering matters. 

I suggest using an explicit ordering or just processing "note" before the loop which would be clearer and safer.

pierrejeambrun: +1 for this, ~~we might have a similar problem on the Task Instance endpoint as well.~~  (Just realized there is another PR opened for this)

pierrejeambrun (commented):
Thanks overall looking good beside Amogh's comment.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
kacpermuda: Thanks, that makes sense, improved this piece of code by removing the loop entirely and just ordering calls explicitly. @pierrejeambrun do you mind linking the taskinstance PR? Would love to track it and make sure these fixes are all merged, included in next release and backported.

amoghrajesh (approved):
Thanks LGTM!

airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
pierrejeambrun: Maybe I got confused, my point is we have the same problem for `patch_task_instance` endpoint I believe. 

The `_patch_task_instance_state` call is before we mutate the 'note', therefore the listener inside will see the old note.

Probably need a follow up similar PR.

pierrejeambrun (approved):

airflow-core/tests/unit/listeners/class_listener.py
```diff
@@ -26,6 +26,9 @@ def __init__(self):
         self.started_component = None
         self.stopped_component = None
         self.state = []
+        self.dag_run_msg: str | None = None
+        self.dag_has_dag_attr: bool | None = None
```
pierrejeambrun, line 30: nit:
```suggestion
        self.dag_run_has_dag_attr: bool | None = None
```

pierrejeambrun: Will be addressed in the TI PR.

PR #69874, opened 2026-07-14
Call listeners for running task instance when a Dag Run state is manually set

kacpermuda:
When a task instance is individually set to success or failure via the API, `on_task_instance_success` / `on_task_instance_failed` is called from the API server, giving listeners a terminal event to act on. The same does not happen when an entire Dag Run is force-set to success or failure: running tasks are bulk-transitioned in the database, but their task-instance listeners are never fired. Listeners that registered a start event (`on_task_instance_running`) for those tasks receive no completion event, leaving them — notably OpenLineage — with a job permanently stuck in RUNNING.

This PR closes the gap:
- `_set_dag_run_terminal_state` now returns `(all_updated_tis, running_tis)` where `running_tis` is the snapshot of task instances that were in an active state (RUNNING, DEFERRED, UP_FOR_RESCHEDULE, AWAITING_INPUT) before the bulk transition, captured without an extra query.
- `set_dag_run_state_to_success` / `set_dag_run_state_to_failed` expose the same tuple so callers have access to the running subset.
- `patch_dag_run_state` fires `on_task_instance_success` / `on_task_instance_failed` for those running task instances before the dag-run-level hook. Each group has its own `try/except` so a listener failure in one cannot suppress the other.
Tasks that were only QUEUED or SCHEDULED (never started, no start event fired) are excluded — they transition to SKIPPED and receive no listener call, consistent with their lifecycle.

### Why the terminal event is missing
The gap is a deliberate design decision that becomes a problem specifically for the dag run force-terminate case:

1. _set_dag_run_terminal_state writes SUCCESS/FAILED into the TI rows in the DB. No listener is fired.
2. Next heartbeat from the supervisor gets 409 CONFLICT ("TI is no longer running"). The supervisor sets self._terminal_state = SERVER_TERMINATED and kills the process via SIGTERM.
3. After the process dies, update_task_state_if_needed() runs. Because SERVER_TERMINATED ∈ STATES_SENT_DIRECTLY (supervisor.py:205), the condition at line 1461 is false — finish() is intentionally skipped. The reasoning is: "the server already knows the state." That is true — but the server setting the state and the server firing listeners for it are two different things, and the latter never happened.
4. Task runner finalizers never ran — the process was killed mid-execution before the on_task_instance_success/failed call in task-sdk/src/airflow/sdk/execution_time/task_runner.py:2249.
5. Scheduler's process_executor_events sees the executor report the task exit, but when it queries the TI it finds state=SUCCESS/FAILED (already written in step 1), so ti_queued is false (scheduler_job_runner.py:1460) and handle_failure is never called.

Result: on_task_instance_running fired from the task runner → task process killed → nothing on the other end. OpenLineage has a dangling RUNNING job with no terminal event. We can't extend the live of the worker process, but we can fire from the api server in that case.

---

- [X] Yes — Claude Code (claude-sonnet-4-6)

Generated-by: Claude Code (claude-sonnet-4-6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/newsfragments/69874.bugfix.rst
+++ b/airflow-core/newsfragments/69874.bugfix.rst
@@ -1,0 +1,1 @@
+Listeners registered via ``on_task_instance_success`` / ``on_task_instance_failed`` are now called for non-teardown task instances that were running when a Dag Run state is manually set to a terminal state (e.g. via the API or UI).
--- a/airflow-core/src/airflow/api/common/mark_tasks.py
+++ b/airflow-core/src/airflow/api/common/mark_tasks.py
@@ -223,7 +223,7 @@ def _set_dag_run_terminal_state(
     ti_state: TaskInstanceState,
     commit: bool,
     session: SASession,
-) -> list[TaskInstance]:
+) -> tuple[list[TaskInstance], list[TaskInstance]]:
     """
     Set the dag run's state to the given terminal state.
 
@@ -237,11 +237,13 @@ def _set_dag_run_terminal_state(
     :param ti_state: state to set on running task instances
     :param commit: commit Dag and tasks to be altered to the database
     :param session: database session
-    :return: If commit is true, list of tasks that have been updated,
-             otherwise list of tasks that will be updated
+    :return: ``(all_updated_tis, running_tis)`` where ``all_updated_tis`` is the
+        combined list of pending (now SKIPPED) and running (now ``ti_state``) TIs,
+        and ``running_tis`` is only the TIs that were in an active state before
+        the transition (useful for firing terminal listener hooks).
     """
     if not dag:
-        return []
+        return [], []
     if not run_id:
         raise ValueError(f"Invalid dag_run_id: {run_id}")
 
@@ -302,13 +304,14 @@ def _set_dag_run_terminal_state(
         if not any(dag.task_dict[ti.task_id].is_teardown for ti in (running_tis + pending_tis)):
             _set_dag_run_state(dag.dag_id, run_id, run_state, session)
 
-    return pending_normal_tis + set_state(
+    all_updated = pending_normal_tis + set_state(
         tasks=running_tasks,
         run_id=run_id,
         state=ti_state,
         commit=commit,
         session=session,
     )
+    return all_updated, running_tis
 
 
 @provide_session
@@ -318,7 +321,7 @@ def set_dag_run_state_to_success(
     run_id: str | None = None,
     commit: bool = False,
     session: SASession = NEW_SESSION,
-) -> list[TaskInstance]:
+) -> tuple[list[TaskInstance], list[TaskInstance]]:
     """
     Set the dag run's state to success.
 
@@ -328,8 +331,13 @@ def set_dag_run_state_to_success(
     :param run_id: the run_id to start looking from
     :param commit: commit Dag and tasks to be altered to the database
     :param session: database session
-    :return: If commit is true, list of tasks that have been updated,
-             otherwise list of tasks that will be updated
+    :return: A tuple ``(all_updated_tis, running_tis)``.
+        ``all_updated_tis`` is the combined list of task instances that were updated:
+        previously-running ones (now SUCCESS) and non-finished pending ones (now SKIPPED).
+        If ``commit`` is ``False``, this is the list of task instances *that would be* updated.
+        ``running_tis`` contains only the task instances that were in an active running state
+        (RUNNING, DEFERRED, UP_FOR_RESCHEDULE, AWAITING_INPUT) before the transition —
+        useful for firing terminal listener hooks.
     :raises: ValueError if dag or logical_date is invalid
     """
     return _set_dag_run_terminal_state(
@@ -349,7 +357,7 @@ def set_dag_run_state_to_failed(
     run_id: str | None = None,
     commit: bool = False,
     session: SASession = NEW_SESSION,
-) -> list[TaskInstance]:
+) -> tuple[list[TaskInstance], list[TaskInstance]]:
     """
     Set the dag run's state to failed.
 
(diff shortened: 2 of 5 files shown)


airflow-core/src/airflow/api/common/mark_tasks.py
amoghrajesh, line 1: What about teardown task instances that are left running by design (so they can finish their own cleanup)? They are still included in the `running_tis` list returned by `_set_dag_run_terminal_state`, so the API server will fire a false "success"/"failed" terminal event for a teardown task that is, in fact, still executing. 

This might've been an untested edge case even before this PR but still flagging.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
```diff
@@ -1609,6 +1609,70 @@ def test_patch_dag_run_notifies_listeners(self, test_client, state, listener_sta
...
+    @pytest.mark.usefixtures("configure_git_connection_for_dag_bundle")
+    def test_patch_dag_run_notifies_ti_listeners_for_running_tasks(
+        self,
+        test_client,
+        dag_maker,
+        session,
+        listener_manager,
+        dag_run_state,
+        expected_ti_listener_state,
+    ):
```
amoghrajesh, line 1646: We should also cover for teardown tasks, both here and in the other test.

amoghrajesh (commented):
Good investigation but some comments need handling

airflow-core/src/airflow/api_fastapi/core_api/services/public/dag_run.py
```diff
@@ -204,7 +211,13 @@ def patch_dag_run_state(
         # Not notifying on queued - only notifying on RUNNING, which happens in the scheduler.
         set_dag_run_state_to_queued(dag=dag, run_id=dag_run.run_id, commit=True, session=session)
     elif state == DagRunMutableStates.FAILED:
-        set_dag_run_state_to_failed(dag=dag, run_id=dag_run.run_id, commit=True, session=session)
+        _, running_tis = set_dag_run_state_to_failed(
+            dag=dag, run_id=dag_run.run_id, commit=True, session=session
+        )
+        try:
+            _emit_state_listener_hooks(running_tis, TaskInstanceState.FAILED)
```
pierrejeambrun: `_emit_state_listener_hooks` already wraps everything in a broad `try/except` I don't think that's needed here

airflow-core/src/airflow/api_fastapi/core_api/services/public/dag_run.py
```diff
@@ -193,7 +194,13 @@ def patch_dag_run_state(
...
     """Set a Dag Run's state (success/queued/failed), firing the matching listener hooks."""
     if state == DagRunMutableStates.SUCCESS:
-        set_dag_run_state_to_success(dag=dag, run_id=dag_run.run_id, commit=True, session=session)
+        _, running_tis = set_dag_run_state_to_success(
+            dag=dag, run_id=dag_run.run_id, commit=True, session=session
+        )
+        try:
+            _emit_state_listener_hooks(running_tis, TaskInstanceState.SUCCESS)
+        except Exception:
+            log.exception("error calling listener")
```
pierrejeambrun: Same here. This can never catch anything

pierrejeambrun (commented):
Small nit, in addition of Amogh's comments

airflow-core/src/airflow/api_fastapi/core_api/services/public/dag_run.py
kacpermuda: Correct, I missed that. Thanks !

airflow-core/src/airflow/api_fastapi/core_api/services/public/dag_run.py
kacpermuda: Correct, I missed that. Thanks !

airflow-core/src/airflow/api/common/mark_tasks.py
kacpermuda: Good spot ! Thanks, excluded teardown tasks now.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
kacpermuda: Added a separate tests for teardown case

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
```diff
@@ -1594,20 +1594,139 @@ def test_patch_dag_run_bad_request(self, test_client):
...
+        session.commit()
+
+        listener = ClassBasedListener()
+        listener_manager(listener)
+
+        response = test_client.patch(
+            f"/dags/test_ti_listeners/dagRuns/{dr.run_id}", json={"state": dag_run_state}
+        )
+        assert response.status_code == 200
+        # Use `is` for type-safe comparison between TaskInstanceState and DagRunState.
```
amoghrajesh:
```suggestion
```

Announcing the assertion, not needed.

amoghrajesh (commented):
This looks like a genuine improvement from a listener's perspective, force-setting a Dag Run's state now fires `on_task_instance_success`/`on_task_instance_failed` for tasks that were actively running, closing the gap where a listener like OL would see a start event with no matching terminal event & for that reason I think its worth adding a newsfragment so this shows up in release notes for anyone relying on task-instance listeners. @pierrejeambrun any thoughts?

Rest of the changes LGTM.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
kacpermuda: Removed in last commit

kacpermuda:
Added newsfragment as requested, let me know if it needs adjustments

airflow-core/newsfragments/69874.bugfix.rst
pierrejeambrun, line 1: Can we mark this as a bug fix instead ? (releasing an improvement in patch release isn't recommended and I don't think we want to wait for 3.4.0, also this is fixing a gap so 'bugfix' seems fair)

pierrejeambrun (approved):
LGTM, just one suggestion before we can merge.

airflow-core/newsfragments/69874.bugfix.rst
kacpermuda: Sure, renamed, rebased, resolved conflicts. Should be good to merge now

PR #69875, opened 2026-07-14
Allow filtering the Dags list by failed and success runs in any run state filter

### LINKED ISSUE #67094: DAG list: state filters should use DAG run counts by state (like 2.x), not latest-run state only
### Description

**Airflow 2.x:** The DAG list filters for run-related states behaved like they used **DAG runs aggregated per DAG** (counts or “has any run in this state”), so you could still find DAGs with failures when the latest run had already succeeded.

**Airflow 3.x:** The UI still offers filters for Queued, Running, Success, and Failed, but they behave as if they use the **latest DAG run’s state** only. Example: a DAG with many `failed` runs and a latest `success` disappears from a failure-oriented view even though failure runs still exist.

**Ask:** Change the DAG list state filters so they apply to **DAG run totals per DAG by state** (same general idea as 2.x), scoped however the backend defines which runs are considered for that list view. If the scope is not obvious in the UI, document it.

### Use case/motivation

We use the DAG list as the first place to spot trouble: filter to failed (or queued/running), open those DAGs, investigate.

That only works if “failed” means **this DAG has failed DAG runs**, not **the last DAG run failed**. Right now 3.x follows the latest run and hides backlog behind a green latest run.

### Related issues

- [#66946](https://github.com/apache/airflow/issues/66946) DAG list page: show total run state counts per DAG  
- [#50624](https://github.com/apache/airflow/issues/50624) Bring back “Status of all previous DAG runs” column on the DAGs page  
- [#60105](https://github.com/apache/airflow/issues/60105) UI: task instance summary on DAG cards  

### Code of Conduct

yuseok89:
Closes #67094

The Dags list has two run-state filters. "Last run state" matches on a Dag's latest run, and "Any run state" matches a Dag that has any run in the chosen state. The "Any run state" filter only offered `queued` and `running`, so a Dag with failed runs but a green latest run could not be found from the Dags list at all.
This adds `failed` and `success` to it, so both filters now cover all four states.

### What changed
- **Backend** (`_AnyDagRunStateFilter`). The filter now supports all four run states. Its query moved from `dag_id IN (SELECT DISTINCT dag_id FROM dag_run WHERE state = X)` to a correlated `EXISTS`, which the existing `(dag_id, state)` index resolves per Dag instead of scanning every run in the state. No new index is needed.
- **UI**. The "Any run state" selector now lists the same four states as "Last run state".

### Performance
Measured on a large `dag_run` table (a few million rows) across a few thousand Dags(Postgres), filtering by `dag_run_state=success` (the heaviest state):
  - Old `DISTINCT` scan of all matching runs: **around a second or more**
  - New correlated `EXISTS`: **tens of milliseconds**

The `EXISTS` cost grows only with Dag count (one index seek per Dag, stopping at the first match), not with run volume, so it stays fast well beyond typical Dag counts, and the gap widens further as run history grows.

### Demo

https://github.com/user-attachments/assets/c932ca3e-afdd-4386-9aba-1d602003da68

---

  - Opus 4.8

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -1342,29 +1342,26 @@ QueryPendingActionsFilter = Annotated[_PendingActionsFilter, Depends(_PendingAct
 class _AnyDagRunStateFilter(BaseParam[DagRunState | None]):
     """Filter Dags that have any DagRun in the given state, not only the latest one."""
 
-    # Only these states have a partial index on dag_run; others would force a full table scan.
-    SUPPORTED_STATES = (DagRunState.QUEUED, DagRunState.RUNNING)
-
     def to_orm(self, select: Select) -> Select:
         if self.value is None and self.skip_none:
             return select
 
-        run_subquery = sql_select(DagRun.dag_id).where(DagRun.state == self.value).distinct()
-        return select.where(DagModel.dag_id.in_(run_subquery))
+        # EXISTS resolves each Dag via the (dag_id, state) index instead of scanning every run in the state.
+        has_run_in_state = (
+            sql_select(DagRun.dag_id)
+            .where(DagRun.dag_id == DagModel.dag_id, DagRun.state == self.value)
+            .exists()
+        )
+        return select.where(has_run_in_state)
 
     @classmethod
     def depends(
         cls,
         dag_run_state: DagRunState | None = Query(
             None,
-            description="Filter Dags that have any DagRun in the given state. Only ``queued`` and ``running`` are supported.",
+            description="Filter Dags that have any DagRun in the given state.",
         ),
     ) -> _AnyDagRunStateFilter:
-        if dag_run_state is not None and dag_run_state not in cls.SUPPORTED_STATES:
-            raise HTTPException(
-                status.HTTP_400_BAD_REQUEST,
-                detail=f"dag_run_state only supports {[state.value for state in cls.SUPPORTED_STATES]}.",
-            )
         return cls().set_value(dag_run_state)
 
 
--- a/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
+++ b/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
@@ -480,11 +480,9 @@ paths:
           anyOf:
           - $ref: '#/components/schemas/DagRunState'
           - type: 'null'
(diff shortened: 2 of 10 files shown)


bbovenzi:
I think this UX needs to be refined. I don't think any user would realistically search "any dag that has ever had a success", even "any dag that has ever had a failure" could quickly be useless for very old failures. I think we first need a time range filter for use cases like "dags with a failure in the last week", or "dags with any runs in the last 24 hours"

yuseok89:
@bbovenzi

Thanks for the review.
That makes sense to me. I think a time range is the right direction, so I prototyped two ways of adding it. A standalone time-range select in the filter bar could read as a page-wide filter, so both attach the time axis to the run-state filter itself.

Agreed that searching for "ever had a success" is unrealistic. I had added `success` for symmetry with "Last run state". With a time window attached though, something like "Dags with a successful run in the last 24 hours" does become meaningful.

Before the backend details, I'd like to settle the UI direction first. Two options:

**Option A — attach a time window to the existing "Any run" filter**


https://github.com/user-attachments/assets/9315fb59-6c8f-43ef-ba41-cffa4133889b

When a state is selected, a `within:` segment appears joined to the same control(Last 24 hours / 7 days / 30 days / Any time). It defaults to **Last 7 days**, so the "very old failures" concern is addressed by the default behavior.

**Option B — merge the two run-state filters into one**


https://github.com/user-attachments/assets/009d1678-465d-495f-9756-0863416e391b

Going a step further. Having "Last run state" and "Any run state" side by side is itself a bit confusing, so this merges them into a single `Run state` control with a scope (`Latest run` / Last 24 hours / 7 days / 30 days / `Any time`). Later, adding an "Any state" entry would cover "dags with any runs in the last 24 hours" with the same control.

My preference is **B** .
It removes the confusion of the two coexisting filters and covers both use cases you mentioned with one model. It does grow the scope of the change, though.
What do you think?

bbovenzi:
Yes, I prefer B too. Good idea

pierrejeambrun (approved):
In the current implementation, this is a net gain. (additional options to filter, while still leveraging indexes and keeping performance good)

I would merge this in the current form and follow up with a UX refactoring for those two filters in a second PR.

pierrejeambrun:
Discussed with Brent, let's merge this iteration which brings value. @yuseok89 please feel free to follow up with a new PR on the UX improvement suggested by Brent.

yuseok89:
@pierrejeambrun 
Sounds good, thanks! I'll open a follow-up PR for the unified filter.

PR #57354, opened 2025-10-27
Upgrade email notifications to use SmtpNotifier

amoghrajesh:
closes: https://github.com/apache/airflow/issues/54712

## What?

Replace legacy `email_on_failure` and `email_on_retry` attributes with SmtpNotifier from apache-airflow-providers-smtp. This change automatically converts the old email configuration to the new notifier-based system while maintaining backward compatibility.

The intention behind doing this is to maintain ONE way of doing things compared to having copies of doing the same thing in multiple ways across core, task sdk, providers.

### Setup

Using docker image of mailpit to set up a SMTP server in same network as breeze

```shell script
➜  ~ docker run -d \
  --network breeze_default \
  -p 8025:8025 \
  --name mailpit \
  axllent/mailpit
```

Setup a connection for SMTP
image

The UI can be accessed at `localhost:8025` now

### Route 1: Testing email functionality from tasks

DAG used:
```python
from datetime import datetime
from airflow import DAG
from airflow.providers.standard.operators.bash import BashOperator
from airflow.providers.smtp.notifications.smtp import send_smtp_notification
from airflow.providers.standard.operators.python import PythonOperator

def retryable_func():
    1//0

with DAG(
    dag_id="smtp_notifier",
    schedule=None,
    start_date=datetime(2023, 1, 1),
    catchup=False,
):
    op1 = BashOperator(
        task_id="mytask",
        email="bash-task@example.com",
        email_on_failure=True,
        bash_command="exit 1",
    )

    op2 = PythonOperator(
        task_id="mytask2",
        email_on_retry=True,
        python_callable=retryable_func,
        email="python-task@example.com",
        retries=3,
    )

    [op1, op2]

```

One of the tasks has email_on_retry defined, the other has email_on_failure.

The bash task fails and python tasks retries.

Bash Task

image

Python Task

image

Emails

For retried task (observe try number)
image

For failed task

image

### Route 2: SMTP provider not installed at all

Uninstall smtp provider on workers
```shell script
root@f7bf900b8f64:/opt/airflow# pip freeze | grep smtp
aiosmtplib==4.0.2
# Editable install with no version control (apache-airflow-providers-smtp==2.3.0)
-e /opt/airflow/providers/smtp
root@f7bf900b8f64:/opt/airflow# uv pip uninstall apache-airflow-providers-smtp
Using Python 3.10.18 environment at: /usr/python
Uninstalled 1 package in 23ms
[...]

PR #30531, opened 2023-04-07
[WIP] Migrate core `EmailOperator` to SMTP provider

hussein-awala:
closes: #30530

---
In this PR, we migrate core `EmailOperator` to SMTP provider, and we deprecate the core one.

PR #69877, opened 2026-07-14
Restore pluggable email backend for task failure/retry alerts

potiuk:
## What

Restore the ability to deliver `email_on_failure` / `email_on_retry` task alerts through the
configured `[email] email_backend` (SES, SendGrid, or any custom callable), which regressed
in #57354.

## Why

`email_on_failure` / `email_on_retry` are deprecated (slated for Airflow 4, migrate to
`SmtpNotifier`) but still functional. Since #57354, task failure/retry emails are routed
**unconditionally** through `SmtpNotifier`, silently ignoring the `[email] email_backend`
configuration. Deployments using a custom `email_backend` (Amazon SES, SendGrid,
org-internal) stopped receiving failure/retry alerts even though the config option still
exists and the deprecated params still work. This restores that behaviour while keeping the
single-notifier design #57354 introduced.

## How

`_send_error_email_notification` (shared by both the worker task-runner path and the
DAG-processor callback path) resolves the delivery mechanism from the **existing**
`[email] email_backend` option — **no new configuration is added**:

1. A non-default `[email] email_backend` is transparently wrapped in a new
   `LegacyEmailBackendNotifier` (`airflow.sdk.execution_time.email_backend`), so existing
   SES/SendGrid setups keep working unchanged. The backend is resolved from config at notify
   time, so the Task SDK keeps no static dependency on `airflow.utils.email`.
2. Otherwise the default `SmtpNotifier` is used, exactly as today.

The notifier lives in the Task SDK next to its only caller, so no additional provider needs
to be installed for a custom email backend to keep working.

Because both failure-email entry points funnel through the same function, the selected
backend is used **consistently** regardless of how the task failed (normal failure vs.
zombie/killed).

## Changes after review

An earlier revision of this PR added a new `[email] email_notifier` config option. Per
`eladkal`'s [review comment](https://github.com/apache/airflow/pull/69877#discussion_r3580742444),
that option has been **dropped** — the goal is to complete the Core→SMTP migration and
deprecate email settings in core, not add new ones. The fix now rides entirely on the
`email_backend` option that already exists, and `config.yml` is untouched by this PR.

## Tests

- `TestEmailNotifications` (task-sdk): default still uses `SmtpNotifier` (unchanged
  regression tests), custom `email_backend` wrapped and invoked with the rendered fields,
  unresolvable `email_backend` logged (no raise).
- `TestExecuteEmailCallbacks` (DAG-processor path): unchanged and green — confirms both
  entry points stay consistent.
- `test_email_backend.py` (task-sdk): unit tests for `LegacyEmailBackendNotifier` dispatch.

## Notes

- The deprecated `email_on_*` params are **not** un-deprecated; this only keeps their
  existing behaviour pluggable until removal in Airflow 4.

---


airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -2394,6 +2394,19 @@ email:
...
+        ``email_on_retry`` task alerts. The class must be a ``BaseNotifier`` subclass whose
+        constructor accepts ``to``, ``from_email``, ``subject`` and ``html_content`` keyword
+        arguments. Override it to route failure/retry alerts through a backend other than SMTP
+        (for example an SES- or SendGrid-based notifier). Both the worker task-runner path and
+        the DAG-processor callback path load the notifier from this single option, so the
+        selected backend is used consistently regardless of how the task failed.
+      version_added: 3.3.1
+      type: string
+      example: ~
+      default: "airflow.providers.smtp.notifications.smtp.SmtpNotifier"
```
eladkal: We want all email related settings out of core
https://github.com/apache/airflow/pull/30531

[[DISCUSS] Drop email integration from Airflow Core
](https://lists.apache.org/thread/tbnf2rcrowrp3ktltrhrcjvc53k58oxx) the community disagreed with me about removing all of that as part of Airflow 3. That ship has sailed but I think we should avoid adding more oil to this fire. We need to complete the Core->SMTP migration and deprecate all email related stuff from core.

potiuk: Indeed - it was draft only - but yes - I removed it now - it will still be read, but it won't be defined.

task-sdk/tests/task_sdk/execution_time/test_email_backend.py
```diff
@@ -0,0 +1,60 @@
...
+
+from airflow.sdk.configuration import conf
+from airflow.sdk.exceptions import AirflowConfigException
+from airflow.sdk.execution_time.email_backend import LegacyEmailBackendNotifier
+
+
+class TestLegacyEmailBackendNotifier:
+    def test_notify_calls_configured_backend(self):
+        """notify() loads the legacy backend from config and calls it with the standard fields."""
+        backend = mock.MagicMock()
```
amoghrajesh: Can we please use a spec or autospec here?

amoghrajesh: This and other occurences.

task-sdk/tests/task_sdk/execution_time/test_task_runner.py
```diff
@@ -4008,6 +4013,76 @@ def execute(self, context):
...
+        runtime_ti = create_runtime_ti(task=task)
+        context = runtime_ti.get_template_context()
+        log = mock.MagicMock()
+
+        with conf_vars({("email", "email_backend"): "airflow.does.not.Exist"}):
+            state, _, error = run(runtime_ti, context, log)
+            # Must not raise even though the backend cannot be loaded.
+            finalize(runtime_ti, state, context, log, error)
+
+        log.exception.assert_called()
```
amoghrajesh, line 4104: This may not be a real regression test. It sets email_backend to "airflow.does.not.Exist" and asserts `log.exception.assert_called()`, but doesn't mock SmtpNotifier. 

Before this PR's fix, the code never reads `email_backend` at all, it always constructs a real SmtpNotifier unconditionally. In the test environment, that real `SmtpNotifier()` call is quite likely to also fail and also hit the same exception block at the call site. If thats the case, this test would pass identically with or without the PR's change meaning it isn't actually verifying "unresolvable custom backend is handled gracefully," it's just asserting "something failed and got logged," which was already true. Worth checking whether this test would fail if you reverted just the fix and kept the test and if it still passes, it should either mock SmtpNotifier to succeed (isolating the assertion to the backend-resolution path) or be dropped in favor of relying on test_custom_email_backend_is_used (which does correctly assert mock_smtp_notifier.assert_not_called()) for the meaningful coverage.

task-sdk/src/airflow/sdk/execution_time/email_backend.py
```diff
@@ -0,0 +1,75 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from airflow.sdk.definitions.context import Context
+
+DEFAULT_EMAIL_BACKEND = "airflow.utils.email.send_email_smtp"
+
+
+class LegacyEmailBackendNotifier(BaseNotifier):
```
amoghrajesh: There is no legitimate use case for a Dag author or provider to construct `LegacyEmailBackendNotifier` directly, I'd keep this class internal / private

task-sdk/tests/task_sdk/execution_time/test_email_backend.py
potiuk: Done. Both test files now use `create_autospec` / `autospec=True`.

The spec function mirrors the real `send_email` signature rather than a bare `*args, **kwargs` stub, so the autospec actually enforces something — it rejects a call missing `subject`. Duplicated rather than imported since task-sdk can't depend on airflow-core.

Left the three pre-existing `SmtpNotifier` patches alone as they predate this PR.

---

task-sdk/tests/task_sdk/execution_time/test_task_runner.py
potiuk: You were right. Ran your experiment — with the fix reverted:

- `test_custom_email_backend_is_used` → FAILED
- `test_unresolvable_email_backend_is_logged` → **PASSED**

Exactly your diagnosis: the real `SmtpNotifier()` also blew up and hit the same `except`.

Took your first option — `SmtpNotifier` is now patched to succeed plus an explicit `assert_not_called()`, so the logged exception can only come from backend resolution. Re-ran the experiment; it fails with the fix disabled now. Kept rather than dropped because the no-silent-fallback-to-SMTP case is what a stale `email_backend` actually hits on upgrade.

---

task-sdk/src/airflow/sdk/execution_time/email_backend.py
potiuk: Agreed — renamed to `_LegacyEmailBackendNotifier`.

---

airflow-core/src/airflow/config_templates/config.yml
amoghrajesh: Good call @eladkal. Good to have lesser config surface here.

task-sdk/src/airflow/sdk/execution_time/task_runner.py
```diff
@@ -1998,20 +2002,39 @@ def _send_error_email_notification(
...
+    consistently regardless of how the task failed.
+    """
     if not task.email:
         return
 
+    email_backend = conf.get("email", "email_backend", fallback=DEFAULT_EMAIL_BACKEND)
+    notifier_description = "SmtpNotifier"
+
+    if email_backend and email_backend != DEFAULT_EMAIL_BACKEND:
+        notifier_class: type = _LegacyEmailBackendNotifier
```
amoghrajesh: Should be: `type[BaseNotifier]`?

task-sdk/tests/task_sdk/execution_time/test_task_runner.py
```diff
@@ -4008,6 +4032,82 @@ def execute(self, context):
                 )
                 assert kwargs["from_email"] == self.FROM
 
+    def test_custom_email_backend_is_used(self, create_runtime_ti, mock_supervisor_comms):
+        """A custom ``[email] email_backend`` is wrapped and invoked with rendered fields."""
+        from airflow.sdk.exceptions import AirflowFailException
+        from airflow.sdk.execution_time.task_runner import finalize, run
```
amoghrajesh: Top level imports instead please

task-sdk/tests/task_sdk/execution_time/test_task_runner.py
```diff
@@ -4008,6 +4032,82 @@ def execute(self, context):
...
+        args, kwargs = backend.call_args
+        # send_email(to, subject, html_content, conn_id=..., from_email=...)
+        assert args[0] == ["test@example.com"]
+        assert kwargs["conn_id"] == "my_smtp"
+        assert kwargs["from_email"] == self.FROM
+
+    def test_unresolvable_email_backend_is_logged(self, create_runtime_ti, mock_supervisor_comms):
+        """An unimportable custom backend is logged, and does not silently fall back to SMTP."""
+        from airflow.sdk.exceptions import AirflowFailException
+        from airflow.sdk.execution_time.task_runner import finalize, run
```
amoghrajesh: Same here.

task-sdk/src/airflow/sdk/execution_time/task_runner.py
amoghrajesh, line 1999: This is also called by dag processor path, worth adding some tests there too?

task-sdk/src/airflow/sdk/execution_time/email_backend.py
```diff
@@ -0,0 +1,75 @@
...
+from airflow.sdk.bases.notifier import BaseNotifier
+from airflow.sdk.configuration import conf
+from airflow.sdk.exceptions import AirflowConfigException
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from airflow.sdk.definitions.context import Context
+
+DEFAULT_EMAIL_BACKEND = "airflow.utils.email.send_email_smtp"
```
amoghrajesh:
```suggestion
_DEFAULT_EMAIL_BACKEND = "airflow.utils.email.send_email_smtp"
```
better?

task-sdk/src/airflow/sdk/execution_time/task_runner.py
```diff
@@ -1998,20 +2002,39 @@ def _send_error_email_notification(
...
+        notifier_description = f"email_backend {email_backend!r}"
```
amoghrajesh:
```suggestion
        notifier_description = f"configured email_backend {email_backend!r}"
```

amoghrajesh (approved):
Looks good after the comments are handled.

task-sdk/src/airflow/sdk/execution_time/email_backend.py
```diff
@@ -0,0 +1,75 @@
...
+        **kwargs: Any,
+    ) -> None:
+        super().__init__()
+        self.to = to
+        self.from_email = from_email
+        self.subject = subject
+        self.html_content = html_content
+
+    def notify(self, context: Context) -> None:
+        backend = conf.getimport("email", "email_backend", fallback=DEFAULT_EMAIL_BACKEND)
```
kosteev: it looks like there is already default value in config.yml for this property, should we still have here fallback then?

kosteev (approved):

task-sdk/src/airflow/sdk/execution_time/task_runner.py
potiuk: Tried it — doesn't typecheck:

```
error: Unexpected keyword argument "to" for "BaseNotifier"  [call-arg]
note: "BaseNotifier" defined in "airflow.sdk.bases.notifier"
```

`BaseNotifier.__init__` only takes `context`; the email kwargs live on the subclasses.

Added a Protocol instead so the call site is still typed:

```python
class _ErrorEmailNotifier(Protocol):
    def __call__(self, to, from_email=..., subject=..., html_content=...) -> BaseNotifier: ...
```

Both `SmtpNotifier` and `_LegacyEmailBackendNotifier` satisfy it; mypy passes.

---

task-sdk/src/airflow/sdk/execution_time/email_backend.py
potiuk: It does today — `[email] email_backend` defaults to `airflow.utils.email.send_email_smtp` in config.yml, and `conf.get` returns it fine without the fallback. I checked.

Keeping it deliberately, though. The direction here is to remove email settings from core, and task-sdk only sees that default because it currently reads core's config.yml under a temporary TODO — it's meant to get its own. The fallback is what keeps this working when either of those lands.

---

task-sdk/src/airflow/sdk/execution_time/task_runner.py
potiuk: Good catch — added `test_execute_email_callbacks_uses_custom_email_backend`.

The gap was real: every existing test in that class patches `_send_error_email_notification`, so none of them reached backend selection at all. The new one lets it run for real and asserts the custom backend is called and `SmtpNotifier` isn't. Verified it fails with the fix reverted.

---

PR #69879, opened 2026-07-14
Fix `airflow db clean` skipping dag_run records with NULL start_date

### LINKED ISSUE #69872: airflow db clean cannot delete dag_run records with NULL start_date

Airflow Core

### Apache Airflow version

3.3.0

`airflow db clean` silently skips `dag_run` records where `start_date IS NULL`. These records accumulate indefinitely in the metadata database and can never be cleaned by any number of `airflow db clean` runs.

  This affects **any dag_run that is marked failed before execution begins** . 

  In our production environment, this resulted in **11,841+ orphaned dag_run records**  that were permanently invisible to the cleanup command.

  ### Root cause

  In `airflow-core/src/airflow/utils/db_cleanup.py`, the `_build_query()` function builds the cleanup query with:

  ```python
  conditions = [base_table_recency_col < clean_before_timestamp]
  ```

  Ref: https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/utils/db_cleanup.py#L424

  For `dag_run`, `recency_column` is `start_date`. When `start_date IS NULL`:

  - `NULL < '2026-07-15'` evaluates to `NULL` in SQL (not `TRUE`)
  - The row is silently excluded from the result set
  - The row can **never** be deleted by `airflow db clean`

  ### How to reproduce

  **1. Create a simple test DAG with `max_active_runs=1`:**

  ```python
  from airflow.sdk import DAG, task
  import time

  with DAG(
      dag_id="test.null_start_date_repro",
      schedule=None,
      start_date=datetime(1970, 1, 1),
      catchup=False,
      max_active_runs=1,
  ) as dag:

      @task
      def wait_2_minutes():
          time.sleep(120)

      wait_2_minutes()
  ```

Image

  **2. Trigger two runs:**

  - Trigger run 1 — it starts executing (sleeps 2 min), `start_date` is set
  - Trigger run 2 while run 1 is still running — it enters `queued` state with `start_date = NULL` (blocked by `max_active_runs=1`)
  - Mark run 2 as `failed` via the UI while it's still queued

  **3. Verify the state in the database:**

  ```sql
  SELECT id, run_id, start_date, state, created_at
  FROM dag_run
  WHERE dag_id = 'test.null_start_date_repro'
  ORDER BY created_at;
  ```

Image

  Result:

  | id | run_id | start_date | state | created_at |
  |----|--------|-----------|-------|------------|
  | 4334 | manual__...40.338513 | 2026-07-14 12:03:41 | success | 2026-07-14 12:03:40 |
  | 4335 | manual__...45.960859 | **NULL** | failed | 2026-07-14 12:03:45 |

  **4. Run `airflow db clean` dry-run (with a future timestamp to ensure all records are eligible):**

  ```bash
  airflow db clean \
    --clean-before-timestamp '2026-07-15 00:00:00+00:00' \
    --tables 'dag_run' \
    --dag-ids 'test.null_start_date_repro' \
    --dry-run \
    --verbose \
    --yes
  ```

  Output:

  ```
  Performing dry run for table dag_run
  Checking table dag_run
  Found 1 rows meeting deletion criteria.
  ```

  **Expected:** 2 rows found (both dag_runs)
  **Actual:** Only 1 row found — the `NULL start_date` record (id=4335) is invisible

  ### Impact

  - NULL `start_date` records **accumulate indefinitely** — no number of cleanup runs will remove them
  - In our production DB: **11,841 orphaned records** from failed backfills (logical_date ranging from 1994 to 2026)

`airflow db clean` should be able to clean `dag_run` records regardless of whether `start_date` is NULL or not.

 One approach to solve this:
  - When `start_date` is NULL, the cleanup should fall back to `created_at`
  -  This should be opt-in via a CLI flag (e.g. `--fallback-on-null`) 

### Operating System

GKE / Linux (Debian-based Airflow container)

### Deployment

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

bileyroy:
closes: #69872

  ## Summary

  When `start_date` is NULL (e.g. dag_runs that were marked failed before execution began), the cleanup query's `WHERE start_date < cutoff` evaluates to NULL and silently skips these rows forever. 

  ## Changes

  - Add `fallback_recency_column_name` field to `_TableConfig` (set to `created_at` for `dag_run`)
  - Add `--fallback-cleanup-on-null` CLI flag (opt-in, off by default)
  - When enabled, apply `COALESCE(start_date, created_at)` to:
    - Main WHERE condition: so NULL `start_date` rows are found
    - `keep_last` subquery (`MAX`): so the latest run is correctly identified via fallback
    - `keep_last` JOIN condition: so the latest run is properly protected
  - Add 5 unit tests covering flag behavior and `keep_last` protection

  ## Test plan

  - [x] `pytest airflow-core/tests/unit/utils/test_db_cleanup.py -xvs` — all passed
  - [x] `ruff check` / `ruff format` — clean
  - [x] `mypy` — clean
  - [x] Manual testing via `airflow standalone`
    

### Testing with manual runs (keep_last not applicable):
  
image

image

image

### Testing with backfill runs (keep_last applicable)

  All 3 backfill runs have `start_date = NULL`. With `--fallback-cleanup-on-null`, only 2 rows are returned for deletion,  the newest run is correctly protected by `keep_last` via `COALESCE(start_date, created_at)`:

image

image

### DIFF
--- a/airflow-core/src/airflow/utils/db_cleanup.py
+++ b/airflow-core/src/airflow/utils/db_cleanup.py
@@ -78,6 +78,11 @@ class _TableConfig:
     :param keep_last_group_by: if keeping the last record, can keep the last record for each group
     :param dependent_tables: list of tables which have FK relationship with this table
     :param extra_filters: SQLAlchemy expressions ANDed with the recency filter; referenced columns must be in ``extra_columns``.
+    :param fallback_recency_column_name: optional fallback date column used when ``recency_column_name``
+        is NULL for a row. When the ``--fallback-cleanup-on-null`` CLI flag is enabled, the cleanup query
+        uses ``COALESCE(recency_column, fallback_recency_column)`` so that rows with a NULL recency value
+        (e.g. ``dag_run.start_date`` for runs that never started) can still be cleaned based on the
+        fallback column (e.g. ``created_at``). The fallback column must be listed in ``extra_columns``.
     :param skip_if_referenced: list of ``(referencing_table, fk_column)`` pairs whose FK points at this
         table's ``referenced_pk_column``. A row that is still referenced by any of these is excluded from
         deletion. This avoids issuing deletes that would violate an ``ON DELETE RESTRICT`` foreign key
@@ -98,11 +103,15 @@ class _TableConfig:
     # Relying on automation here would increase complexity and reduce maintainability.
     dependent_tables: list[str] | None = None
     extra_filters: list[Any] | None = None
+    fallback_recency_column_name: str | None = None
     skip_if_referenced: list[tuple[str, str]] | None = None
     referenced_pk_column: str = "id"
 
     def __post_init__(self):
         self.recency_column = column(self.recency_column_name)
+        self.fallback_recency_column = (
+            column(self.fallback_recency_column_name) if self.fallback_recency_column_name else None
+        )
         if self.dag_id_column_name is None:
             self.dag_id_column = None
             self.orm_model: Base = table(
@@ -154,11 +163,12 @@ config_list: list[_TableConfig] = [
         table_name="dag_run",
         recency_column_name="start_date",
         dag_id_column_name="dag_id",
-        extra_columns=["dag_id", "run_type"],
+        extra_columns=["dag_id", "run_type", "created_at"],
         keep_last=True,
         keep_last_filters=[column("run_type") != DagRunType.MANUAL],
         keep_last_group_by=["dag_id"],
         dependent_tables=["task_instance", "task_state_store", "deadline"],
+        fallback_recency_column_name="created_at",
     ),
     _TableConfig(table_name="asset_event", recency_column_name="timestamp", dag_id_column_name="dag_id"),
     _TableConfig(table_name="import_error", recency_column_name="timestamp"),
@@ -372,8 +382,14 @@ def _subquery_keep_last(
     keep_last_filters,
     group_by_columns,
     max_date_colname,
+    fallback_recency_column=None,
+    fallback_cleanup_on_null: bool = False,
 ):
-    subquery = select(*group_by_columns, func.max(recency_column).label(max_date_colname))
+    if fallback_cleanup_on_null and fallback_recency_column is not None:
+        effective_col = func.coalesce(recency_column, fallback_recency_column)
+    else:
+        effective_col = recency_column
+    subquery = select(*group_by_columns, func.max(effective_col).label(max_date_colname))
 
     if keep_last_filters is not None:
         for entry in keep_last_filters:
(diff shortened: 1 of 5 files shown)


airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -490,6 +517,8 @@ def _cleanup_table(
     extra_filters: list[Any] | None = None,
     skip_if_referenced: list[tuple[str, str]] | None = None,
     referenced_pk_column: str = "id",
+    fallback_recency_column=None,
```
bugraoz93, line 520: This has been never passed from the caller. I see the enitre logic coupled with this field and the boolean asked to user. If you check below comment where it called, you will see we only pass the user input one. 
Even this is calculated in inner parts, it should only exists where it is needed and calculated.

airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -728,6 +764,7 @@ def run_cleanup(
                     skip_archive=skip_archive,
                     session=session,
                     batch_size=batch_size,
+                    fallback_cleanup_on_null=fallback_cleanup_on_null,
```
bugraoz93, line 767: Here we are not pssing the other new recency one

bugraoz93 (changes requested):
I didn't quite understand why we have recency column from the implementation itself while we are always having it null on the `_cleanup_table`

bugraoz93:
Thanks for your first PR!

airflow-core/src/airflow/utils/db_cleanup.py
bileyroy: @bugraoz93  so this `fallback_recency_column` is part of `_TableConfig` ([here](https://github.com/apache/airflow/pull/69879/changes#diff-0f691d24c54a28756110b2d9125bb06a3d8862333977ef86ec30ad33220440c8R162)) and is derived in `__post_init__` (same pattern as `recency_column` which is also derived there from `recency_column_name`).  It is not passed explicitly, but coming from `**table_config.__dict__` [here](https://github.com/apache/airflow/pull/69879/changes#diff-0f691d24c54a28756110b2d9125bb06a3d8862333977ef86ec30ad33220440c8R763) in the usage of `_cleanup_table()` function call

Computing it inside `_cleanup_table()` would be redundant since `__post_init__` already does this derivation once at config initialization time.

airflow-core/src/airflow/utils/db_cleanup.py
bileyroy: Addressed above.

airflow-core/src/airflow/utils/db_cleanup.py
bileyroy: @bugraoz93 does it sound good?

bileyroy: @bugraoz93 can you confirm please?

airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -154,11 +163,12 @@ def readable_config(self):
         table_name="dag_run",
         recency_column_name="start_date",
         dag_id_column_name="dag_id",
-        extra_columns=["dag_id", "run_type"],
+        extra_columns=["dag_id", "run_type", "created_at"],
```
uranusjr, line 166: Since created_at was only added in 3.2.0, we probably need more columns for this. Otherwise the oldest dag runs will never be able to be selected by the query, which is the opposite of what we want.

PR #69885, opened 2026-07-14
Fix SSHRemoteJobOperator cleanup failing with custom remote_base_dir

### LINKED ISSUE #69813: SSHRemoteJobOperator: inconsistent "remote_base_dir" validation

harshhh817:
Fixes `SSHRemoteJobOperator` failing at the cleanup step whenever a custom `remote_base_dir` is used.

The operator accepts (and validates) a custom `remote_base_dir` at construction time and creates the job directory under it, but `build_posix_cleanup_command` / `build_windows_cleanup_command` validated the job directory against the hardcoded *default* base directory. Result: the remote job runs to completion, then the task fails with

```
Invalid job directory: '/tmp-data/airflow-ssh-jobs/af_…'. Expected path under '/tmp/airflow-ssh-jobs' for safety.
```

**Fix:** thread the operator's configured `remote_base_dir` through the cleanup builders into `_validate_job_dir`, falling back to the OS-specific default when unset. The safety property is preserved — the `rm -rf` target must still live under the base directory the job was actually created under (which was itself validated by `_validate_base_dir` at operator construction). Both the sync and deferrable (`execute_complete`) cleanup paths go through the fixed `_cleanup_remote_job`.

**Tests:** new unit tests for custom-base-dir acceptance and rejection (posix + windows), plus an operator-level regression test reproducing the exact failure from the issue — verified it fails on `main` and passes with this change. Full ssh provider unit suite: 49 passed.

closes: #69813

---

- [X] Yes (Claude Code)

---

### DIFF
--- a/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
@@ -452,9 +452,9 @@ class SSHRemoteJobOperator(BaseOperator):
         """
         self.log.info("Cleaning up remote job directory: %s", job_dir)
         if remote_os == "posix":
-            cleanup_cmd = build_posix_cleanup_command(job_dir)
+            cleanup_cmd = build_posix_cleanup_command(job_dir, base_dir=self.remote_base_dir)
         else:
-            cleanup_cmd = build_windows_cleanup_command(job_dir)
+            cleanup_cmd = build_windows_cleanup_command(job_dir, base_dir=self.remote_base_dir)
 
         last_error: Exception | None = None
         for attempt in range(1, self.cleanup_retries + 1):
--- a/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
@@ -31,18 +31,22 @@ POSIX_DEFAULT_BASE_DIR = "/tmp/airflow-ssh-jobs"
 WINDOWS_DEFAULT_BASE_DIR = "$env:TEMP\\airflow-ssh-jobs"
 
 
-def _validate_job_dir(job_dir: str, remote_os: Literal["posix", "windows"]) -> None:
+def _validate_job_dir(
+    job_dir: str, remote_os: Literal["posix", "windows"], base_dir: str | None = None
+) -> None:
     """
     Validate that job_dir is under the expected base directory.
 
     :param job_dir: The job directory path to validate
     :param remote_os: Operating system type
+    :param base_dir: Base directory the job was created under (e.g. the operator's
+        ``remote_base_dir``). Falls back to the OS-specific default base directory.
     :raises ValueError: If job_dir doesn't start with the expected base path
     """
     if remote_os == "posix":
-        expected_prefix = POSIX_DEFAULT_BASE_DIR + "/"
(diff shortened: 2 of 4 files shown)


potiuk:
Heads-up: #70091 merged a few minutes ago and has put this into conflict — apologies, I reviewed them in sequence and the overlap only became visible once the first landed.

Both PRs modify `build_posix_cleanup_command` in `providers/ssh/src/airflow/providers/ssh/utils/remote_job.py`: #70091 changed its body to `return f"rm -rf {shlex.quote(job_dir)}"`, while this PR changes its signature to take `base_dir`. The two edits are adjacent, and this branch is ~441 commits behind `main`, so a rebase is needed.

The resolution should be mechanical — keep your signature change and `_validate_job_dir(job_dir, "posix", base_dir)`, and take `shlex.quote(job_dir)` from `main` for the `rm -rf` line. Same pattern for the kill/log-tail builders if you touch them.

On the fix itself, which I did review: the diagnosis is right. `_validate_job_dir` compares against `POSIX_DEFAULT_BASE_DIR` unconditionally, so any operator with a custom `remote_base_dir` fails cleanup with "Invalid job directory" even though the path is legitimate. Passing the operator's own base dir through is the correct fix, and defaulting to the OS constant keeps existing behaviour intact.

No review comments beyond the rebase — happy to look again once it's green.

---

potiuk (approved):
This is the right fix, and of the three PRs opened for this bug it has the strongest tests — it's the only one that pins the negative case on both POSIX and Windows, i.e. that a job dir outside the custom base is still rejected. On a guard function the interesting behaviour is what it *refuses*, so that matters.

Using `(base_dir or DEFAULT)` rather than an `is None` check is also the better failure mode: an empty base dir falls back to the safe default instead of producing a prefix of `/` that almost any absolute path satisfies. Not reachable today, since `_validate_base_dir` rejects empty at execute time, but it is the right way round for a security check.

#70091 landed on `main` while this was open and rewrote the same POSIX command builders to use `shlex.quote`, which put this branch into conflict. That was merged underneath you, so I have resolved it rather than sending it back:

- `build_posix_cleanup_command` now takes both changes — `_validate_job_dir(job_dir, "posix", base_dir)` together with `shlex.quote(job_dir)`.
- Your four `test_remote_job.py` tests and the `TestPosixPathQuoting` class from #70091 are both kept; they were adjacent additions rather than a genuine conflict.
- `test_cleanup_with_custom_remote_base_dir` asserted `rm -rf '/tmp-data/...'` with hand-written quotes. `shlex.quote` leaves a path containing no shell metacharacters unquoted, so that assertion needed updating to the unquoted form.

I checked the merged result keeps both properties rather than just compiling:

```
custom base accepted : rm -rf /data/jobs/job_1
outside base rejected: Invalid job directory '/tmp/airflow-ssh-jobs/j'...
injection quoted     : rm -rf '/tmp/airflow-ssh-jobs/x'"'"'; touch /tmp/pwned; :'"'"''
default base works   : rm -rf /tmp/airflow-ssh-jobs/j
```

58 tests pass in the ssh unit suite.

Two other PRs were opened for this same bug — #70788 and #70802 — and I am closing both in favour of this one.

---

PR #69899, opened 2026-07-15
Keep Go SDK variable lookups consistent

### LINKED ISSUE #67598: Share AIRFLOW_VAR_* env-override helper between HTTP and coordinator Go SDK clients
**Nice to have modularization, not-blocking for 3.3.0 release.**

---

### Dependency

- Wait until #67317 merges — the coordinator-mode duplication lands in that PR, so the dedup work can only start once both copies exist on ``main``.

### Background

The Go SDK has two independent implementations of the same Python-parity rule that an environment variable named ``AIRFLOW_VAR_<UPPER(key)>`` short-circuits ``GetVariable`` and is returned in preference to the supervisor-stored value (matching ``airflow.models.variable.Variable.get``).

The HTTP-backed client uses an unexported helper:

```go
// go-sdk/sdk/client.go
func variableFromEnv(key string) (string, bool) {
    return os.LookupEnv(VariableEnvPrefix + strings.ToUpper(key))
}

func (*client) GetVariable(ctx context.Context, key string) (string, error) {
    // TODO: Let the lookup priority be configurable like it is in Python SDK
    if env, ok := variableFromEnv(key); ok {
        return env, nil
    }
    ...
}
```

The coordinator-mode client inlines the same logic and carries a TODO calling out the duplication:

```go
// go-sdk/pkg/execution/client.go
func (c *CoordinatorClient) GetVariable(ctx context.Context, key string) (string, error) {
    // TODO: this duplicates variableFromEnv in sdk/client.go. The env-first
    // precedence is part of the SDK contract, so both clients should share a
    // single helper (e.g. an exported sdk.VariableFromEnv) instead of two
    // independent copies that can drift.
    if env, ok := os.LookupEnv(sdk.VariableEnvPrefix + strings.ToUpper(key)); ok {
        return env, nil
    }
    ...
}
```

The env-first lookup order is part of the public Go SDK contract; two independently maintained copies can silently drift, particularly if the lookup precedence is made configurable later (as the HTTP-side TODO notes).

### What needs to happen

1. Promote ``variableFromEnv`` to an exported helper on the ``sdk`` package — e.g. ``sdk.VariableFromEnv(key string) (string, bool)`` — so the env-override logic has a single canonical implementation.
2. Update ``go-sdk/sdk/client.go`` to call the exported helper.
3. Update ``go-sdk/pkg/execution/client.go`` ``CoordinatorClient.GetVariable`` to call the same exported helper and drop the inlined ``os.LookupEnv`` block.
4. Remove the duplication TODO in ``pkg/execution/client.go``.

### Acceptance criteria

- Only one implementation of the ``AIRFLOW_VAR_*`` env-override lookup exists in the Go SDK source tree.
- Both ``sdk.client`` and ``execution.CoordinatorClient`` call the shared helper.
- Existing tests (``TestCoordinatorClientGetVariableEnvOverride`` and any HTTP-side equivalent) continue to pass without modification, or are updated to exercise the shared helper directly if duplication of the test coverage no longer makes sense.

### Context

- Originating review: self-review thread on #67317 (coordinator comms + client PR).
- HTTP-side helper: ``go-sdk/sdk/client.go`` (``variableFromEnv``).
- Coordinator-side duplicate (lands with #67317): ``go-sdk/pkg/execution/client.go`` ``CoordinatorClient.GetVariable``.
- Related project board: https://github.com/orgs/apache/projects/499/views/1

shaealh:
## Summary

Share the `AIRFLOW_VAR_*` environment-override lookup through the exported `sdk.VariableFromEnv` helper.

Use the shared helper from both the HTTP-backed and coordinator clients so the env-first precedence has one canonical implementation. This also removes the duplicated coordinator lookup and its associated TODO.

closes: #67598

## Testing

- `go test ./...`
- `go vet ./...`
- `go mod tidy`
- Airflow prek hooks

### DIFF
--- a/go-sdk/pkg/execution/client.go
+++ b/go-sdk/pkg/execution/client.go
@@ -22,8 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
-	"strings"
 
 	"github.com/apache/airflow/go-sdk/pkg/api"
 	"github.com/apache/airflow/go-sdk/pkg/execution/genmodels"
@@ -72,11 +70,7 @@ func NewCoordinatorClient(comm *CoordinatorComm) *CoordinatorClient {
 
 // GetVariable requests a variable value from the supervisor.
 func (c *CoordinatorClient) GetVariable(ctx context.Context, key string) (string, error) {
-	// TODO: this duplicates variableFromEnv in sdk/client.go. The env-first
-	// precedence is part of the SDK contract, so both clients should share a
-	// single helper (e.g. an exported sdk.VariableFromEnv) instead of two
-	// independent copies that can drift.
-	if env, ok := os.LookupEnv(sdk.VariableEnvPrefix + strings.ToUpper(key)); ok {
+	if env, ok := sdk.VariableFromEnv(key); ok {
 		return env, nil
 	}
 
--- a/go-sdk/sdk/client.go
+++ b/go-sdk/sdk/client.go
@@ -41,13 +41,14 @@ func NewClient() Client {
 	return &client{}
 }
 
-func variableFromEnv(key string) (string, bool) {
+// VariableFromEnv returns the AIRFLOW_VAR_<UPPER(key)> value and whether it is set.
+func VariableFromEnv(key string) (string, bool) {
 	return os.LookupEnv(VariableEnvPrefix + strings.ToUpper(key))
 }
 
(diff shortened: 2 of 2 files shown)


jason810496 (changes requested):
Sorry, as https://github.com/apache/airflow/pull/71874 was merged (since there is coordinator client only). We don't need this PR any more. Thanks for your contribution.

PR #69908, opened 2026-07-15
Fix passphrase quoting in decrypt_remote_file_to_string for Windows

### LINKED ISSUE #69396: Quoting of passphrase won't work for windows
Single-quoting does not work on Windows:

https://github.com/apache/airflow/blob/d122d0386d684dd449d6581dd5b4244e30f7e607/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py#L55

Maybe this function is only intended for use on Unix systems; if so perhaps this should be documented.
[Note that the decrypt_remote_file function in tpt_util.py does cater for Windows.]

kada2004:
…emotes

 

the fix detects the remote OS using the existing `get_remote_os()  helper, already used in tpt_utils.py and applies the  correct quoting:
- Windows: double quotes with `'''` escaping
- Unix/Linus: single quotes with `'\''` escaping
closes: #69396 
---

---

### DIFF
--- a/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
+++ b/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
@@ -21,6 +21,8 @@ import secrets
 import string
 import subprocess
 
+from airflow.providers.teradata.utils.tpt_util import get_remote_os
+
 
 def generate_random_password(length=12):
     # Define the character set: letters, digits, and special characters
@@ -51,25 +53,27 @@ def generate_encrypted_file_with_openssl(file_path: str, password: str, out_file
 
 
 def decrypt_remote_file_to_string(ssh_client, remote_enc_file, password, bteq_command_str):
-    # Run openssl decrypt command on remote machine
-    quoted_password = shell_quote_single(password)
+    remote_os = get_remote_os(ssh_client)
+    windows_remote = remote_os == "windows"
+
+    if windows_remote:
+        # Windows cmd.exe does not support single quotes; use double quotes with "" escaping
+        password_escaped = password.replace('"', '""')
+        quoted_password = f'"{password_escaped}"'
+    else:
+        # Unix shells require closing the single quote, adding an escaped quote, then reopening
+        password_escaped = password.replace("'", "'\\''")
+        quoted_password = f"'{password_escaped}'"
 
     decrypt_cmd = (
         f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass pass:{quoted_password} -in {remote_enc_file} | "
         + bteq_command_str
     )
-    # Clear password to prevent lingering sensitive data
+    # Clear sensitive data from memory
     password = None
     quoted_password = None
     stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
-    # Wait for command to finish
     exit_status = stdout.channel.recv_exit_status()
     output = stdout.read().decode()
     err = stderr.read().decode()
     return exit_status, output, err
-
-
-def shell_quote_single(s):
-    # Escape single quotes in s, then wrap in single quotes
-    # In shell, to include a single quote inside single quotes, close, add '\'' and reopen
-    return "'" + s.replace("'", "'\\''") + "'"
--- a/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
+++ b/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import secrets
+import shlex
 import string
 import subprocess
 
@@ -51,14 +52,13 @@ def decrypt_remote_file_to_string(ssh_client, remote_enc_file, password, bteq_co
     # Use -pass stdin to avoid shell quoting on any OS and keep the passphrase
     # out of the remote process table where ps could expose it.
     decrypt_cmd = (
-        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | " + bteq_command_str
+        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {shlex.quote(remote_enc_file)} | "
+        + bteq_command_str
     )
     stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
     stdin.write(password + "\n")
     stdin.flush()
     stdin.channel.shutdown_write()
-    # Clear password to prevent lingering sensitive data
-    password = None
     exit_status = stdout.channel.recv_exit_status()
     output = stdout.read().decode()
     err = stderr.read().decode()
--- a/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
+++ b/providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
@@ -18,23 +18,17 @@
 from __future__ import annotations
 
 import secrets
-import shlex
 import string
 import subprocess
 
 
 def generate_random_password(length=12):
-    # Define the character set: letters, digits, and special characters
     characters = string.ascii_letters + string.digits + string.punctuation
-    # Generate a random password
     password = "".join(secrets.choice(characters) for _ in range(length))
     return password
 
 
 def generate_encrypted_file_with_openssl(file_path: str, password: str, out_file: str):
-    # Write plaintext temporarily to file
-
-    # Run openssl enc with AES-256-CBC, pbkdf2, salt
     cmd = [
         "openssl",
         "enc",
@@ -54,25 +48,18 @@ def generate_encrypted_file_with_openssl(file_path: str, password: str, out_file
 
 
 def decrypt_remote_file_to_string(ssh_client, remote_enc_file, password, bteq_command_str):
-    # Run openssl decrypt command on remote machine
-    quoted_password = shell_quote_single(password)
-
+    # Use -pass stdin to avoid shell quoting on any OS and keep the passphrase
+    # out of the remote process table where ps could expose it.
     decrypt_cmd = (
-        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass pass:{quoted_password} -in {shlex.quote(remote_enc_file)} | "
-        + bteq_command_str
+        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | " + bteq_command_str
     )
+    stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
+    stdin.write(password + "\n")
+    stdin.flush()
+    stdin.channel.shutdown_write()
     # Clear password to prevent lingering sensitive data
     password = None
-    quoted_password = None
-    stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
-    # Wait for command to finish
     exit_status = stdout.channel.recv_exit_status()
     output = stdout.read().decode()
     err = stderr.read().decode()
     return exit_status, output, err
-
-
-def shell_quote_single(s):
-    # Escape single quotes in s, then wrap in single quotes
-    # In shell, to include a single quote inside single quotes, close, add '\'' and reopen
-    return "'" + s.replace("'", "'\\''") + "'"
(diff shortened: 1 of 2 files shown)


potiuk (changes requested):
Thanks — you've identified a real problem. Unix-style `'\''` quoting is simply wrong against `cmd.exe`, so a Windows remote would break on any passphrase containing a quote today.

But I don't think the `""` escaping is sufficient, and the reason is concrete rather than hypothetical. `generate_random_password` draws from `string.punctuation`:

```
!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~
```

which includes **`! " % & ( ) < > ^ |`** — all cmd.exe metacharacters. Doubling `"` doesn't neutralise them: `%VAR%` expands *inside* double quotes, `&` and `|` still separate commands, and `^` is cmd's own escape character. So a randomly generated passphrase will fairly often contain something this doesn't cover, and the failure will look like a mysterious decrypt error rather than a quoting bug. There's also the question of whether the Windows remote shell is `cmd.exe` at all — OpenSSH on Windows is often configured with PowerShell as the default, which needs different escaping again.

**The stronger fix is to stop putting the passphrase on the command line.** `openssl enc` supports `-pass stdin`, and you already hold `stdin` from `exec_command`:

```python
decrypt_cmd = f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | " + bteq_command_str
stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
stdin.write(password + "\n")
stdin.flush()
stdin.channel.shutdown_write()
```

That solves both problems at once: no shell quoting to get right on either OS, and — importantly — the passphrase no longer appears in the remote process table where any user running `ps` can read it. That exposure exists today with `-pass pass:` and is arguably the bigger issue of the two.

If `-pass stdin` isn't viable for some reason, the fallback would need genuine cmd.exe escaping (caret-escaping the metacharacters, and confirming the target shell), not just quote doubling.

Separately: the branch now conflicts with `main` and needs a rebase.

One small thing either way — removing `shell_quote_single` and inlining the Unix branch loses a named helper for no gain. The module already imports `shlex`, so `shlex.quote()` covers the Unix side exactly.

---

kada2004:
> Thanks — you've identified a real problem. Unix-style `'\''` quoting is simply wrong against `cmd.exe`, so a Windows remote would break on any passphrase containing a quote today.
> 
> But I don't think the `""` escaping is sufficient, and the reason is concrete rather than hypothetical. `generate_random_password` draws from `string.punctuation`:
> 
> ```
> !"#$%&'()*+,-./:;<=>?@[\]^_`{|}~
> ```
> 
> which includes **`! " % & ( ) < > ^ |`** — all cmd.exe metacharacters. Doubling `"` doesn't neutralise them: `%VAR%` expands _inside_ double quotes, `&` and `|` still separate commands, and `^` is cmd's own escape character. So a randomly generated passphrase will fairly often contain something this doesn't cover, and the failure will look like a mysterious decrypt error rather than a quoting bug. There's also the question of whether the Windows remote shell is `cmd.exe` at all — OpenSSH on Windows is often configured with PowerShell as the default, which needs different escaping again.
> 
> **The stronger fix is to stop putting the passphrase on the command line.** `openssl enc` supports `-pass stdin`, and you already hold `stdin` from `exec_command`:
> 
> ```python
> decrypt_cmd = f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | " + bteq_command_str
> stdin, stdout, stderr = ssh_client.exec_command(decrypt_cmd)
> stdin.write(password + "\n")
> stdin.flush()
> stdin.channel.shutdown_write()
> ```
> 
> That solves both problems at once: no shell quoting to get right on either OS, and — importantly — the passphrase no longer appears in the remote process table where any user running `ps` can read it. That exposure exists today with `-pass pass:` and is arguably the bigger issue of the two.
> 
> If `-pass stdin` isn't viable for some reason, the fallback would need genuine cmd.exe escaping (caret-escaping the metacharacters, and confirming the target shell), not just quote doubling.
> 
> Separately: the branch now conflicts with `main` and needs a rebase.
> 
> One small thing either way — removing `shell_quote_single` and inlining the Unix branch loses a named helper for no gain. The module already imports `shlex`, so `shlex.quote()` covers the Unix side exactly.
> 

Thanks for the review:).  I have  updated the implementation to use -pass stdin for both functions as suggested.

Instead of quoting the passphrase on the command line, the passphrase is now passed via stdin in both generate_encrypted_file_with_openssl (using subprocess.run(... input=...)) and decrypt_remote_file_to_string (using stdin.write() over SSH). This avoids all shell quoting issues on any OS and keeps the passphrase out of the process table entirely.

The tests have been updated to verify the new approach, including two security-focused tests that assert the passphrase never appears on the command line.

providers/teradata/src/airflow/providers/teradata/utils/encryption_utils.py
```diff
@@ -54,25 +48,18 @@ def generate_encrypted_file_with_openssl(file_path: str, password: str, out_file
...
 def decrypt_remote_file_to_string(ssh_client, remote_enc_file, password, bteq_command_str):
-    # Run openssl decrypt command on remote machine
-    quoted_password = shell_quote_single(password)
-
+    # Use -pass stdin to avoid shell quoting on any OS and keep the passphrase
+    # out of the remote process table where ps could expose it.
     decrypt_cmd = (
-        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass pass:{quoted_password} -in {shlex.quote(remote_enc_file)} | "
-        + bteq_command_str
+        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | " + bteq_command_str
```
potiuk: The path lost its quoting here — main had `-in {shlex.quote(remote_enc_file)}`, and the `shlex` import was removed along with it. A remote path containing a space now breaks the command. (This came from my snippet in the previous review, which dropped it — my mistake.)

```suggestion
    decrypt_cmd = (
        f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {shlex.quote(remote_enc_file)} | "
        + bteq_command_str
    )
```

…with `import shlex` restored at the top.

potiuk (commented):
This is exactly the right shape now — thanks for taking the bigger change rather than patching the quoting. The passphrase never reaches the remote command line, so there is nothing left to escape on either OS, and it is no longer visible to anyone running `ps` on the remote host. The test with `s3cr3t&rm -rf ~` is a good one to have: it fails loudly if the passphrase ever creeps back onto the command string.

One thing to restore before this lands, and one small note.

### `shlex.quote()` on the remote path was dropped (`encryption_utils.py:54`)

Before this PR the command built the input path as `-in {shlex.quote(remote_enc_file)}`; it is now interpolated raw, and the `shlex` import went with it. A remote path containing a space (a temp directory under a user profile with a space in it, for example) will break the command, and it puts an unquoted value back into a shell string for no benefit. This came from the snippet in my previous review, which dropped the quoting — my mistake, not something you should have caught. Re-adding `import shlex` and wrapping the path restores the old behaviour.

### Smaller observations

- `encryption_utils.py:61` — `password = None` under the comment *"Clear password to prevent lingering sensitive data"*. Rebinding the local name does not clear anything: the caller still holds the same string object, and CPython gives no way to zero a `str` in place. This predates your change, but it now sits inside the block you rewrote, and the comment promises a guarantee the code cannot make. Dropping both lines would leave the function honest — the real protection is the one you just added, which is keeping the value off the command line.

---

>

---

kada2004:
> This is exactly the right shape now — thanks for taking the bigger change rather than patching the quoting. The passphrase never reaches the remote command line, so there is nothing left to escape on either OS, and it is no longer visible to anyone running `ps` on the remote host. The test with `s3cr3t&rm -rf ~` is a good one to have: it fails loudly if the passphrase ever creeps back onto the command string.
> 
> One thing to restore before this lands, and one small note.
> ### `shlex.quote()` on the remote path was dropped (`encryption_utils.py:54`)
> 
> Before this PR the command built the input path as `-in {shlex.quote(remote_enc_file)}`; it is now interpolated raw, and the `shlex` import went with it. A remote path containing a space (a temp directory under a user profile with a space in it, for example) will break the command, and it puts an unquoted value back into a shell string for no benefit. This came from the snippet in my previous review, which dropped the quoting — my mistake, not something you should have caught. Re-adding `import shlex` and wrapping the path restores the old behaviour.
> ### Smaller observations
> 
>     * `encryption_utils.py:61` — `password = None` under the comment _"Clear password to prevent lingering sensitive data"_. Rebinding the local name does not clear anything: the caller still holds the same string object, and CPython gives no way to zero a `str` in place. This predates your change, but it now sits inside the block you rewrote, and the comment promises a guarantee the code cannot make. Dropping both lines would leave the function honest — the real protection is the one you just added, which is keeping the value off the command line.
> 
> 
> > confirmed by an Airflow maintainer. The findings
> > below are observations, not blockers; an Airflow
> > maintainer — a real person — will take the next look at the
> > PR. If you think a finding is mis-applied, please reply on
> > the PR and a maintainer will weigh in._
> 

Thank you for catching this. Restored shlex.quote on remote path and drop misleading password = None


> This is exactly the right shape now — thanks for taking the bigger change rather than patching the quoting. The passphrase never reaches the remote command line, so there is nothing left to escape on either OS, and it is no longer visible to anyone running `ps` on the remote host. The test with `s3cr3t&rm -rf ~` is a good one to have: it fails loudly if the passphrase ever creeps back onto the command string.
> 
> One thing to restore before this lands, and one small note.
> ### `shlex.quote()` on the remote path was dropped (`encryption_utils.py:54`)
> 
> Before this PR the command built the input path as `-in {shlex.quote(remote_enc_file)}`; it is now interpolated raw, and the `shlex` import went with it. A remote path containing a space (a temp directory under a user profile with a space in it, for example) will break the command, and it puts an unquoted value back into a shell string for no benefit. This came from the snippet in my previous review, which dropped the quoting — my mistake, not something you should have caught. Re-adding `import shlex` and wrapping the path restores the old behaviour.
> ### Smaller observations
> 
>     * `encryption_utils.py:61` — `password = None` under the comment _"Clear password to prevent lingering sensitive data"_. Rebinding the local name does not clear anything: the caller still holds the same string object, and CPython gives no way to zero a `str` in place. This predates your change, but it now sits inside the block you rewrote, and the comment promises a guarantee the code cannot make. Dropping both lines would leave the function honest — the real protection is the one you just added, which is keeping the value off the command line.
> 
> 
> > confirmed by an Airflow maintainer. The findings
> > below are observations, not blockers; an Airflow
> > maintainer — a real person — will take the next look at the
> > PR. If you think a finding is mis-applied, please reply on
> > the PR and a maintainer will weigh in._
> 

Thank you for catching these issues. Restored shlex.quote on remote path and drop misleading password =None.

providers/teradata/tests/unit/teradata/utils/test_encryption_utils.py
```diff
@@ -91,25 +79,42 @@ def test_decrypt_remote_file_to_string(self):
...
         ssh_client.exec_command.return_value = (mock_stdin, mock_stdout, mock_stderr)
 
         exit_status, output, err = decrypt_remote_file_to_string(
             ssh_client, remote_enc_file, password, bteq_command_str
         )
 
-        quoted_password = shell_quote_single(password)
         expected_cmd = (
-            f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass pass:{quoted_password} -in {remote_enc_file} | "
+            f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {remote_enc_file} | "
```
potiuk, line 92: Non-blocking, but worth closing the loop on: `remote_enc_file` is `/remote/encrypted.enc`, which contains no shell-special characters, so `shlex.quote()` returns it byte-identical. This assertion therefore passes whether or not the source actually applies the quoting — which is exactly how it got dropped last time (from my snippet), silently.

Giving the fixture a path with a space makes the test fail the moment the quoting disappears:

```python
# line 75
remote_enc_file = "/remote dir/encrypted.enc"

# here
f"openssl enc -d -aes-256-cbc -salt -pbkdf2 -pass stdin -in {shlex.quote(remote_enc_file)} | "
```

---

potiuk (approved):
Both points from the last round are done, and the shape of the fix is right — the passphrase never touches the remote command line now, so there is nothing to escape on either OS and nothing for `ps` to expose on the remote host. Net deletion of 8 lines for that is a good trade.

Dropping `shell_quote_single` along with its two tests is correct now that nothing quotes a passphrase, and stripping the narrating comments moves the module toward the "comment sparingly" rule rather than away from it.

One non-blocking note inline about the test fixture — it is the same class of gap that let the `shlex.quote()` regression through in the first place.

---

Issue #69531, opened 2026-07-07
Grid view slow


hterik:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.2.2

### What happened and how to reproduce it?

For a moderately sized dag with approx 30 tasks and history of 25 dag runs shown, the grid view is extremely slow. Clicking on items in the grid has signifiant yank and delay.  This was not an issue in Airflow 2. 

* Just hovering over an item has noticeable lag before the row is highlighted
* Clicking on an item in the grid causes the grid to flicker into white, before showing the same content again. Then it takes 3-4 seconds for the content in the details tab to load.
* It is likely _not_ an API issue, looking in the network tab of the browser shows this after clicking on an item in the grid:
    <img width="331" height="118" alt="Image" src="https://github.com/user-attachments/assets/675ca0bb-0420-4834-9ea8-b90907866c2a" />
    - The requests are served pretty fast: ~50ms each
    - There's not excessive amounts of requests: 6 per click, which is not great, but not terrible either. Two minor things standing out:
         - `hitlDetails` and `externalLogUrl` are seperate endpoints. Better to merge them into the `tasks` or `taskInstances` endpoint?
         -  `/ui/grid/ti_summaries` appears to reload on every click, even if clicking a task doesn't change the range of displayed dagruns
    - It some times takes a second after clicking even for the first request to be *sent* (not seen in the graph)
    -  After the last response has been received, it takes nearly a second for the UI to be updated (not seen in the graph)

### Operating System

Ubuntu

### Deployment

Other

### Versions of Apache Airflow Providers

```
apache-airflow==3.2.2
apache-airflow-client==3.2.2
apache-airflow-core==3.2.2
apache-airflow-mypy==0.1.0
apache-airflow-providers-celery==3.20.0
apache-airflow-providers-cncf-kubernetes==10.17.1
apache-airflow-providers-common-compat==1.15.0
apache-airflow-providers-common-io==1.7.2
apache-airflow-providers-common-sql==2.0.0
apache-airflow-providers-docker==4.5.6
apache-airflow-providers-fab==3.6.4
apache-airflow-providers-postgres==6.7.0
apache-airflow-providers-smtp==3.0.1
apache-airflow-providers-standard==1.13.1
apache-airflow-task-sdk==1.2.2
```

### Official Helm Chart version

Not Applicable

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #69912, opened 2026-07-15
Stop re-rendering the whole Grid on every hover

pierrejeambrun:
Every cell in the Dag Grid view subscribed to a shared hover context, so hovering any cell changed the context value and re-rendered the entire grid — the interaction lag reported in #69531, which worsens as the Dag grows. This moves the run-column / task-row crosshair highlight to a single delegated `pointerover` handler that toggles CSS classes on the matching cells by `data-run-id`/`data-task-id`, so hovering does no React render work; the same delegated root covers the gantt so the shared row highlight stays in sync.

Measured on a ~20-run × 20-task grid (React Profiler): commits per hover dropped from **~3.0 to ~0.07** — hovering now does essentially zero React work. Appearance is unchanged (same crosshair highlight; selected cells still win).

The separate stream-driven full-grid re-render (the other cause of grid slowness in #69531) is addressed in a stacked follow-up on top of this branch.

related: #69531

---


pierrejeambrun:
## Before
https://github.com/user-attachments/assets/b6010cca-2676-49da-b92b-1cf4d3d17b2a



## After
https://github.com/user-attachments/assets/65a96ce2-5c1b-45bb-8ad6-093fb98b6601

bbovenzi (approved):
Great! Yes, we should've done the hovering in css a while ago. Thanks

PR #69913, opened 2026-07-15
Remove redundant ORM result uniquing in core queries

ephraimbuddy:
The affected queries select unique entities or only eager-load scalar relationships, so identity filtering adds unnecessary result processing.


pierrejeambrun (approved):
LGTM

phanikumv (approved):

PR #69914, opened 2026-07-15
Replace `reattach_on_restart` with durable execution in KPO

amoghrajesh:
---

- [x] Yes Sonnet 4.5

### Why?

KPO's existing `reattach_on_restart` relies on a label search on every retry, which can raise `FoundMoreThanOnePodFailure` when more than one matching pod exists, and gives no way to reconnect to a specific pod unambiguously. On Airflow 3.3+, persisting the running pod's identity to task state store lets a retry reconnect directly to that exact pod instead, removing the ambiguity window entirely. `reattach_on_restart` is deprecated in favor of a new durable flag that supersedes it; existing configurations map onto durable with zero behavior change, and pre-3.3 environments keep today's label-search behavior unchanged.

 ### Summary

- Adds a `durable` parameter to `KubernetesPodOperator` (default `True`), superseding the
  deprecated `reattach_on_restart` parameter.
- On Airflow 3.3+, `durable=True` persists the running pod's identity (name, namespace, uid)
  to task state store before the operator waits on it. On retry, the operator reconnects
  directly to that pod via a direct `read_namespaced_pod` lookup instead of a label search --
  removing the `FoundMoreThanOnePodFailure` ambiguity window a label search can hit when more
  than one matching pod exists.
- The existing label-search mechanism (`find_pod()`) is retained as a one-time bootstrap
  fallback: a pod created by a pre-upgrade provider version (no persisted identity yet) is
  still found via labels on the first retry after upgrading, and that retry starts persisting
  state for every subsequent one. No in-flight task or existing pod breaks across an upgrade.
- `reattach_on_restart` still works, mapping its value onto `durable` with zero behavior
  change; passing it emits `AirflowProviderDeprecationWarning` on Airflow 3.3+ only (below
  3.3, `durable`/task state store don't functionally exist yet, so warning would be
  premature).
- `durable=False` opts out of task state store entirely, preserving today's `reattach_on_restart=False`
  behavior exactly, including its pre-existing incidental-cleanup quirk (an orphaned pod from
  a killed prior attempt can still get found and deleted via unconditional `find_pod()`
  refetch calls elsewhere in `execute_sync`/`execute_async` -- unrelated to and unchanged by
  this work).
- Docs: added a "Durable execution" section to KPO's operator docs, cross-linking task state
  store.

### Testing

Using this dag:
```python

from airflow.providers.cncf.kubernetes.operators.pod import KubernetesPodOperator
from airflow.sdk import DAG

with DAG(
    dag_id="try_kpo",
    schedule=None,
    catchup=False,
) as dag:
    run_pod = KubernetesPodOperator(
        task_id="run_pod",
        name="try-kpo",
        namespace="default",
        image="ubuntu:24.04",
        cmds=["bash", "-cx"],
        arguments=["sleep 60 && echo done"],
        in_cluster=False,
        do_xcom_push=False,
        retries=2,
        retry_delay=timedelta(seconds=5),
    )
```

### Before my changes:

Worker killed:

image

But pod still exists:

image

Reconnected when worker came back up:

image

### After my changes (changing to pass durable=True)

```python
    run_pod = KubernetesPodOperator(
        task_id="run_pod",
        name="try-kpo",
        namespace="default",
        image="ubuntu:24.04",
        cmds=["bash", "-cx"],
        arguments=["sleep 60 && echo done"],
        in_cluster=False,
        do_xcom_push=False,
        retries=2,
        retry_delay=timedelta(seconds=5),
        durable=True,
    )
```

WOrker killed:

image

Task store persisted:

image

```shell
[Breeze:3.10.20] root@0014bf761b80:/opt/airflow$ cat /tmp/airflow_state/ti_run_pod/pod_identifier.json
{"name": "try-kpo-vlap9a9c", "namespace": "default", "uid": "878b321f-fb21-4779-94c4-1811f7ced720"}[Breeze:3.10.20]
```

Pod present and reconnected to via state store:

image

---


raphaelauv:
with `reattach_on_restart` the operator check at every restart if a matching pod exist

but with `durable` , the operator only check if the value on the state is not empty ? So in case of hard-crash of the worker just after calling k8S to run the pod , the operator will not try to fund the pod ?

thanks

amoghrajesh:
@raphaelauv good comment, but that's covered already. `_get_pod_from_task_state_store` returning None (nothing persisted yet) falls straight through to `find_pod()`, the same label search `reattach_on_restart` always did.

So, the crash-before-persist window youre describing falls to the exact old behavior for that one retry & the operator still checks for a matching pod via labels, same as `reattach_on_restart` always did. It only skips the label search once the identity is actually persisted on a later retry.

raphaelauv:
thanks for the quick answer :)

That's perfect ,also it would be great to make it explicit in the doc

amoghrajesh:
@raphaelauv thanks, handled it in https://github.com/apache/airflow/pull/69914/changes/7a75ba6f336cd2618a506a4141dca9538fb7455d

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -170,8 +176,14 @@ class KubernetesPodOperator(BaseOperator):
     :param in_cluster: run kubernetes client with in_cluster configuration.
     :param cluster_context: context that points to kubernetes cluster.
         Ignored when in_cluster is True. If None, current-context is used. (templated)
-    :param reattach_on_restart: if the worker dies while the pod is running, reattach and monitor
-        during the next try. If False, always create a new pod for each try.
+    :param reattach_on_restart: deprecated, use ``durable`` instead. If the worker dies while the pod
+        is running, reattach and monitor during the next try. If False, always create a new pod for
+        each try.
```
eladkal:
```suggestion
    :param reattach_on_restart: deprecated for Airflow 3.3+, use ``durable`` instead. If the worker dies while the pod
        is running, reattach and monitor during the next try. If False, always create a new pod for
        each try.
```

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -170,8 +176,14 @@ class KubernetesPodOperator(BaseOperator):
...
+    :param durable: if the worker dies while the pod is running, reattach and monitor during the next
+        try instead of creating a duplicate pod. If False, always create a new pod for each try.
+        Supersedes ``reattach_on_restart``; on Airflow 3.3+ the reconnection uses a persisted pod
+        identity in task state store instead of a label search, removing the ambiguity failure a label
+        search can hit when more than one matching pod exists. Defaults to ``True``.
```
eladkal:
```suggestion
    :param durable: if the worker dies while the pod is running, reattach and monitor during the next
        try instead of creating a duplicate pod. If False, always create a new pod for each try.
        Supersedes ``reattach_on_restart``; on Airflow 3.3+ the reconnection uses a persisted pod
        identity in task state store instead of a label search, removing the ambiguity failure a label
        search can hit when more than one matching pod exists. Defaults to ``True``. Only available for Airflow 3.3+
```

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -377,6 +389,18 @@ def __init__(
...
+            # `True` or `False`.
+            reattach_on_restart = kwargs.pop("reattach_on_restart")
+            if AIRFLOW_V_3_3_PLUS:
+                warnings.warn(
+                    "`reattach_on_restart` is deprecated and will be removed in a future release. "
+                    "Use `durable` instead.",
+                    AirflowProviderDeprecationWarning,
+                    stacklevel=2,
+                )
+            durable = reattach_on_restart
```
eladkal: I think we need also the handle the case of `durable` with Airflow<3.3
I think for that case we want to just make it work as `reattach_on_restart` so we want ask them to change parameter name but just handle this for them under the hood?

amoghrajesh: That is already handled - `self.reattach_on_restart = durable` always mirrors durable's value regardless of Airflow version (3.3+ or not), and code paths for reattach block is gated purely on `self.reattach_on_restart`, not any version check.

On versions < 3.3, `_get_pod_from_task_state_store` returns None immediately since `task_state_store` isn't in context at all, so it falls straight to `find_pod()` mechanism (current). So `durable=True` or `durable=False` on old Airflow already behaves exactly like `reattach_on_restart=True/False` did.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Handled in 325a931486e

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Handled in 325a931486e

eladkal (approved):

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -377,6 +391,18 @@ def __init__(
         log_formatter: Callable[[str, str], str] | None = None,
         **kwargs,
     ) -> None:
+        if "reattach_on_restart" in kwargs:
```
kaxil: Pulling `reattach_on_restart` out of the named signature into `**kwargs` breaks DAG-level `default_args={"reattach_on_restart": ...}`. `apply_defaults` only forwards `default_args` keys that match a declared `__init__` parameter (the `for arg in sig_cache.parameters` loop in task-sdk `bases/operator.py`), so a `**kwargs`-only key isn't in that set, gets dropped, and the task then silently falls back to `durable=True`.

I verified this against task-sdk: the same param declared as a named arg gets `default_args` applied, but read from `**kwargs` it's ignored. A direct `KubernetesPodOperator(reattach_on_restart=False)` still works, so only the `default_args` path regresses, which dents the "existing configurations map onto durable with zero behavior change" promise.

Keeping `reattach_on_restart` as a real (deprecated) named param, still emitting the warning, would cover both paths.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -610,9 +640,53 @@ def log_matching_pod(self, pod: k8s.V1Pod, context: Context) -> None:
...
+            return None
+        name, namespace = stored.get("name"), stored.get("namespace")
+        if not isinstance(name, str) or not isinstance(namespace, str):
+            self.log.warning(
+                "Pod identity stored in task state store is malformed (%s), falling back to label search.",
+                stored,
+            )
+            return None
+        try:
+            pod = self.hook.get_pod(name, namespace)
```
kaxil, line 661: Minor, non-blocking: `_persist_pod_identity_to_task_state_store` stores `uid` alongside `name`/`namespace`, but the reconnect here only uses `name`/`namespace` and never compares the fetched `pod.metadata.uid` against the stored one. So the persisted `uid` is currently dead weight, and to someone reading the stored dict it looks like pod identity is verified on reconnect when it isn't.

Practical risk is low, since pod names carry a random suffix and exact name reuse for the same task instance is unlikely. But I'd either compare the uid and fall back to label search on mismatch, or drop `uid` from the stored payload so the schema doesn't imply a check that isn't performed. `test_durable_reconnects_directly_via_task_state_store` stores `uid: "abc"` and never asserts on it, which reflects the same gap.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Good catch, fixed. `reattach_on_restart` is back as a real named param now.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Fair point, dropped uid from the persisted payload entirely rather than implementing the comparison -- it's now just {"name": ..., "namespace": ...}. Given the low practical risk you called out (random pod-name suffix), not worth the extra complexity right now

amoghrajesh: Handled in 8335567339

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Handled in 8335567339

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -614,9 +643,53 @@ def log_matching_pod(self, pod: k8s.V1Pod, context: Context) -> None:
...
```
kaxil, line 661: This reconnect fetches the stored pod by name with no `already_checked` filter, which reintroduces reattach-on-retry that the base path deliberately prevents.

The task state store is scoped to `(dag_run_id, task_id, map_index, key)` with no `try_number` (`models/task_state_store.py`: "Retries of the same task share the same rows"), and rows are cleared only on SUCCESS + `clear_on_success` (default `False`). So on a retry after failure, the persisted identity still points at the previous attempt's now-terminal pod, and `get_pod` returns it. In `get_or_create_pod`, `process_pod_deletion` returns `False` for `on_finish_action` in {`keep_pod`, `delete_succeeded_pod` (failed pod), `delete_active_pod` (terminal pod)}, so `if not deleted_pod: return pod` hands back the dead pod. `execute_sync` then awaits it and re-raises the old failure, so the retry never actually runs. Manual-clear-and-rerun of a kept SUCCEEDED pod returns its stale XCom without re-running.

Base avoids this because `cleanup()` stamps `already_checked=True` (`patch_already_checked` docstring: "ensure we don't reattach on retries") and `find_pod(exclude_checked=True)` then skips it, forcing a fresh pod. The durable reconnect drops that guard.

Scope: the pure default (`durable=True` + `on_finish_action="delete_pod"`) is safe, since the deleted pod 404s and falls back to the label search. It bites when a keep/conditional-delete action is set (all documented configs).

Suggested fix: in `_get_pod_from_task_state_store`, treat a terminal or `already_checked`-labelled pod as not reattachable (return `None`) so the fresh-create path runs and re-persists the new identity. Note `test_durable_reconnect_terminal_state_runs_existing_handling` forces `process_pod_deletion.return_value = True`, which skips exactly this branch, so a `keep_pod` + terminal-pod test asserting a fresh pod is created would catch it.

providers/cncf/kubernetes/docs/operators.rst
```diff
@@ -187,6 +187,45 @@ for debugging), set ``on_kill_action="keep_pod"``:
...
+
+If the worker running ``KubernetesPodOperator`` dies while the pod is still running (e.g. the
+worker is preempted or crashes) and the task is retried, the operator can reattach to the pod
+that is already running instead of creating a duplicate. This is controlled by the ``durable``
+parameter, which defaults to ``True``.
+
+On Airflow 3.3+, ``durable=True`` persists the running pod's identity (name, namespace) to
+:doc:`task state store <apache-airflow:core-concepts/task-state-store>` before the operator starts
+waiting on it. On retry, the operator reads this identity back and reconnects directly to that
+specific pod -- no label search, and no possibility of the ambiguity failure
```
kaxil: This overstates it slightly. The reconnect removes the label search only inside `get_or_create_pod`. `execute_sync` still calls `self.find_pod(...)` unconditionally right after (to populate `remote_pod`, plus the callback paths), and `find_pod` raises `FoundMoreThanOnePodFailure` when more than one unchecked pod matches. So a task with a duplicate orphan pod can still hit the ambiguity failure on the durable path.

Suggest scoping the wording to something like "the reconnect step uses the persisted identity instead of a label search" rather than "no possibility of the ambiguity failure," or reusing the reconnected pod for `remote_pod` so the claim actually holds end to end.

amoghrajesh: Good point, softened the wording. It now reads "the reconnect step uses this persisted identity instead of a label search" rather than claiming the ambiguity failure is impossible. Handled in 0218939c74

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
amoghrajesh: Fixed: `_get_pod_from_task_state_store` now checks the fetched pods already_checked label and if a prior attempt already marked it, it's treated as not reattachable and falls through to fresh pod creation instead of handing back a dead pod from a prior attempt.

Added  a test covering this too. (FAILED + already_checked=True stored pod, on_finish_action="keep_pod") -- verified it fails without the fix and passes with it. Fixed in 0218939c74

amoghrajesh:
The comments were handled, merging this one for now. Let me know if there are more concerns @kaxil and I can handle in a follow up

PR #68916, opened 2026-06-23
Honor json_logs config in Celery worker startup

hanxdatadog:
Closes #68912

## Problem

The Celery worker ignores `[logging] json_logs` / `AIRFLOW__LOGGING__JSON_LOGS`. It always calls `configure_logging(output=sys.stdout.buffer)` without passing `json_output`, so structured JSON logging is never enabled on the worker even when configured globally.

## Solution

Add a two-level config lookup before calling `configure_logging`:

1. `[celery] json_logs` — worker-specific override, takes precedence.
2. `[logging] json_logs` — global setting, used when the celery key is absent.
3. `False` — default when neither key is configured.

This mirrors the existing `[celery] CELERY_LOGGING_LEVEL` / `[logging] LOGGING_LEVEL` fallback pattern already present in the worker startup code.

```ini
# Global — affects the API server, scheduler, and workers:
[logging]
json_logs = True

# Or override for the Celery worker only, leaving other components unchanged:
[celery]
json_logs = True
```

## Changes

- **`providers/celery/src/airflow/providers/celery/cli/celery_command.py`**: read `[celery] json_logs` → `[logging] json_logs` → `False` and pass `json_output` to `configure_logging`.
- **`providers/celery/provider.yaml`** + **`get_provider_info.py`**: new `json_logs` config option under `[celery]`.
- **`providers/celery/tests/unit/celery/cli/test_celery_command.py`**: `TestWorkerJsonLogs` — 3 tests covering (a) default is False, (b) global `[logging]` setting is inherited, (c) `[celery]` override takes precedence.
- **`providers/celery/docs/celery_executor.rst`**: new "Worker logging" section documenting the fallback order and Airflow 3.0/3.2 compatibility note.
- **`providers/celery/newsfragments/68912.feature.rst`**: changelog entry.

## Notes

- Only active on Airflow 3+ (`AIRFLOW_V_3_0_PLUS` guard is unchanged).
- `conf.getboolean(..., fallback=None)` is safe on 3.0/3.1 even when the `[celery] json_logs` key is absent.
- `version_added: ~` in `provider.yaml` — version not yet decided; to be set by the release manager.

PR #52651, opened 2025-07-01
Switch all airflow logging to structlog

ashb:
This change continues the work that was started in Airflow 3.0 and AIP-72 to
use Structlog in place of stdlib logging in Airflow.

The primary change here is to make LoggingMixin return a customized
structlogger; customized to maintain surface compatability with logging.Logger
(surface meaning things like handlers and filters aren't preserved) and to
then capture/redirect all logging via stdlib via Structlog processors.

This is the first step in allowing all Airflow components to be able to
produce JSON logs natively.

Things of note in this implementation:

- We have a customized structlog filtering logger that has a "per-logger-tree"
  level concept.

  This is implemented using a [prefix trie][1] to lookup the logging level to
  efficiently be able to look up the level for child loggers when configured
  only at the parent level.

- We have a custom `PercentFormatRender` class that renders non-JSON logs that
  understands the stdlib style format strings, meaning users custom logging
  config will be respected again for the daemon components.

  (Note though: this won't help with Task logs, as those are always JSON and
  the UI does the rendering of those.)

- There is no longer a need for a different log format for colored and plain -- using
  color format specifiers (`%(blue)s`, `%(log_level)s` etc) when colors are
  disabled/not available will output nothing in their place.

- Introduce an mechanism for users to easily set the log level for loggers --
  for instance if you are debugging the scheduler, it would be nice to be able
  to set the `airflow.jobs.scheduler_job_runner` logger to DEBUG while keeping
  everything else at info. The change to allow this is not in this PR, but once this 
  lands it will be a simpler follow on PR.

[1]: https://en.wikipedia.org/wiki/Trie

PR #69919, opened 2026-07-15
Fix Celery worker crash on Airflow 3.0 with json_logs

vatsrahul1001:
## What

The Celery worker CLI passes `json_output` to `airflow.sdk.log.configure_logging` under an `AIRFLOW_V_3_0_PLUS` gate (added in #68916 "Honor json_logs config in Celery worker startup"):

```python
if AIRFLOW_V_3_0_PLUS:
    from airflow.sdk.log import configure_logging
    ...
    configure_logging(output=sys.stdout.buffer, json_output=json_output)
```

But `json_output` was only added to the Task SDK's `configure_logging` in **SDK 1.1 / Airflow 3.1** (the structlog migration, #52651). On **Airflow 3.0.x** that parameter does not exist, so the call raises:

```
TypeError: configure_logging() got an unexpected keyword argument 'json_output'
```

…and the Celery worker crashes on startup. The `AIRFLOW_V_3_0_PLUS` gate is too loose — this affects any plain Airflow 3.0.x deployment running celery provider >= 3.22.0.

## How

- Add an `AIRFLOW_V_3_1_PLUS` constant to the provider's `version_compat`.
- Gate the `json_logs` handling on `AIRFLOW_V_3_1_PLUS`. On Airflow 3.0.x, fall back to `configure_logging(output=sys.stdout.buffer)` (no `json_output`), matching the pre-#68916 behavior.

This keeps the provider compatible across the full Airflow 3.x range it declares support for; `json_logs` is honored from 3.1+ where the SDK supports it.

## Tests

Added `test_json_output_not_passed_on_airflow_3_0` (patches `AIRFLOW_V_3_1_PLUS = False`, asserts `configure_logging` is called without `json_output`). Existing `json_logs` tests continue to assert the 3.1+ behavior.

---

- [X] Yes (Claude Code)

I reviewed and understand all of the generated code, verified it against the actual `worker()` code path, and ran the relevant static checks and tests locally.

### DIFF
--- a/providers/celery/src/airflow/providers/celery/cli/celery_command.py
+++ b/providers/celery/src/airflow/providers/celery/cli/celery_command.py
@@ -39,6 +39,7 @@ from airflow.cli.simple_table import AirflowConsole
 from airflow.exceptions import AirflowConfigException
 from airflow.providers.celery.version_compat import (
     AIRFLOW_V_3_0_PLUS,
+    AIRFLOW_V_3_1_PLUS,
     AIRFLOW_V_3_2_PLUS,
     AIRFLOW_V_3_3_PLUS,
 )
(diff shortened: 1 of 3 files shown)


kaxil (approved):
LGTM. Correct version boundary -- verified against stable-branch content that `json_output` lands in the task-sdk `configure_logging` on 3.1+, so gating the 3.0.x fallback on `AIRFLOW_V_3_1_PLUS` is right. Fix restores pre-#68916 behavior and the regression test has teeth.

PR #68355, opened 2026-06-10
Go SDK: stop logging fetched edge workloads verbatim

drewrukin:
This changes the Go edge worker to stop logging fetched jobs as raw structs in `fetchJob`.

Instead of logging the full API response and decoded workload, the worker now logs only non-sensitive execution identifiers and metadata. This keeps the fetch log useful for debugging while avoiding accidental disclosure of token- bearing workload fields.

A regression test was added to verify that `fetchJob` still preserves the workload token for execution, but does not emit it to logs.

##### Was generative AI tooling used to co-author this PR?
  - [X] Yes (Codex)

Generated-by: Codex following the guidelines

PR #69920, opened 2026-07-15
Go SDK: Prevent Edge Worker config secrets from reaching logs

viiccwen:
Prevent the Go SDK Edge Worker from serializing its complete startup configuration into info-level logs.
The configuration contains the JWT signing key used to authenticate Edge Worker API requests, while process logs are often available to a wider audience than signing material.

Represent the logged configuration with an explicit `slog.LogValuer` allowlist containing only the hostname, queues, and retry settings. Keep the log after successful worker construction so a constructor error cannot be followed by dereferencing an unavailable worker logger.

Add a regression test that checks the exact safe configuration fields and verifies that the signing secret, its field name, and a credential-bearing API URL never reach the log output.

The approach follows the allowlisted structured-logging pattern used by #68355 for fetched Edge workloads.

---

- [x] Yes - GPT-5.6 Sol

---

### DIFF
--- a/go-sdk/edge/worker.go
+++ b/go-sdk/edge/worker.go
@@ -113,10 +113,10 @@ func Run(ctx context.Context) error {
 	}
 
 	w, err := NewWorker(conf)
-	w.logger.Info("Config", "config", conf)
 	if err != nil {
 		return err
 	}
+	w.logger.Info("Config", "config", conf)
 
 	err = w.Register(ctx)
 	if err != nil {
--- a/go-sdk/edge/worker_test.go
+++ b/go-sdk/edge/worker_test.go
@@ -25,11 +25,56 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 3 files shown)


viiccwen:
cc @jason810496, @henry3260. 🙌

jason810496 (commented):
Thanks!

Here's another direction for solving this issue. Would it be better to just mask the sensitive fields (e.g. `ApiJWTSecretKey` and `Issuer` for now) instead of hand wiring here (which might cause further drift).

viiccwen:
> Thanks!
> 
> Here's another direction for solving this issue. Would it be better to just mask the sensitive fields (e.g. `ApiJWTSecretKey` and `Issuer` for now) instead of hand wiring here (which might cause further drift).

Thx, that makes sense to me. I’ll move the redaction policy next to WorkerConfig by giving it a safe structured log representation, so callers can log the config without duplicating its field mapping.

I’ll also update the test to verify that none of those raw values reach the log output.

viiccwen:
Hello @jason810496, politely tag for further review.

jason810496 (commented):

PR #69923, opened 2026-07-15
Go SDK: Shut down Edge Workers gracefully on SIGTERM

viiccwen:
Handle `SIGTERM` through the same graceful drain path as the first interrupt received by a Go SDK Edge Worker.

The worker registered `SIGTERM` with `signal.Notify`, which replaced the signal's default termination behavior, but its signal switch did not handle `SIGTERM`. Container runtimes and service managers could therefore leave the worker running until escalating to a forced kill, bypassing task draining and normal lifecycle cleanup.

Separate OS signal registration from the core worker loop so tests can inject a signal channel without signaling the test process. Stop signal delivery when the wrapper exits, and route both `SIGINT` and `SIGTERM` through the existing drain behavior.

Add a regression test that injects `SIGTERM` and uses a test Edge API server to verify the worker reports an `idle` heartbeat followed by `offline` as it drains. The test also verifies the path repeatedly without installing process-wide signal handlers.

---

- [x] Yes - GPT-5.6 Sol

---

### DIFF
--- a/go-sdk/edge/worker.go
+++ b/go-sdk/edge/worker.go
@@ -270,14 +270,19 @@ func (w *worker) mainLoop(ctx context.Context) error {
 		return err
 	}
 
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM, syscall.SIGUSR1)
-
 	// Report state to Idle
 	if err := w.heartbeat(ctx); err != nil {
 		return err
 	}
 
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM, syscall.SIGUSR1)
+	defer signal.Stop(sigChan)
+
+	return w.runMainLoop(ctx, sigChan)
+}
+
+func (w *worker) runMainLoop(ctx context.Context, sigChan <-chan os.Signal) error {
 	heartbeat := time.NewTicker(HeartbeatInterval)
 	defer heartbeat.Stop()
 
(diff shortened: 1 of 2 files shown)


go-sdk/edge/worker.go
```diff
@@ -291,13 +296,12 @@ func (w *worker) mainLoop(ctx context.Context) error {
...
 			switch sig {
-			case os.Interrupt:
+			case os.Interrupt, syscall.SIGTERM:
 				if !w.drain {
 					w.logger.Info(
-						"Draining worker of running tasks before stopping, hit ^C again to force terminate",
+						"Draining worker of running tasks before stopping; send another termination signal to force terminate",
 					)
 					w.drain = true
-					w.heartbeat(ctx)
```
jason810496, line 300: Why do we need to remove the heartbeat here?

jason810496 (commented):
Thanks for the fix.

go-sdk/edge/worker.go
viiccwen: Hello @jason810496, thx for raising this! 

The heartbeat was removed intentionally. It's synchronous and uses the main-loop context, so an unavailable or stalled Edge API **can block** the first SIGTERM handler. That delays stopFetchJobs() and prevents the loop from receiving a second SIGTERM to force termination, potentially exhausting the container shutdown grace period.

The worker still reports Offline through deregister, which uses its own bounded timeout. The added test verifies that the first SIGTERM starts draining without issuing this blocking heartbeat.

viiccwen:
@jason810496 politely tag for further review.

jason810496 (commented):

Issue #55, opened 2015-06-21
Can you add a slightly more realistic example of a data pipeline in the cloud?


r39132:
For example, the Luigi [example](http://luigi.readthedocs.org/en/latest/example_top_artists.html) walks through a case that involves importing data into a DB. It would be cool if there were some examples that read from one location (e.g. S3) and wrote to another (e.g. DB).

artwr:
We have an operator that transfers files from [s3 to Hive](http://pythonhosted.org/airflow/code.html#airflow.operators.S3ToHiveTransfer), as well as a [Hive to Mysql operator](https://github.com/airbnb/airflow/blob/master/airflow/operators/hive_to_mysql.py), which is unfortunately missing from the documentation. The hooks being already in place, a s3 to Postgres operator should not be too difficult. Here is an example of how our s3 to Hive operator would work :

```
from collections import OrderedDict
from airflow.operators import S3ToHiveTransfer

S3ToHiveTransfer(
    task_id='s3_to_hive',
    s3_key='s3://bucket-name/user_list_{{ ds }}.tsv',
    field_dict=OrderedDict([("user_id", "BIGINT")
                            , ("first_name", "STRING")
                            , ("last_name", "STRING")
                            , ("registered_at", "TIMESTAMP")]),
    hive_table='{{ params.db_name }}.user_list_{{ ds_nodash }}',
    create=True,
    recreate=True,
    delimiter='\t',
    s3_conn_id='s3_connection_name',
    dag=dag)

```
Usually, the transfer is a separate operation of any transform and has its own operator.
I hope this helps. 
Best,
Arthur

mistercrunch:
Agreed, the current tutorial is really just focused on the mechanics of Airflow with very foobar-y examples. I didn't want to write a pipeline that was too stack specific (MySQL / Hive / ...) and wanted to make sure it would work for anyone, regardless of the stack they might have. 

Maybe using a SqliteOperator to do some analytics on some data scraped from the Internet would be a good example. It could be interesting to re-write the Luigi example for comparison :)

But yeah, it's on the TODO list.

r39132:
Cool. I'll close this for now.

PR #69924, opened 2026-07-15
Support excluding files from `common.ai` Agent Skills discovery

kaxil:
`AgentSkillsToolset` gains an optional `exclude_resources` parameter — a list of glob patterns forwarded to `pydantic-ai-skills`' `SkillsToolset` that keeps matched files out of a skill's discovered resources. It mirrors the existing `exclude_tools` escape hatch and is resolved per-run on the worker (so nothing is baked into the serialized DAG).

## Why now

[`pydantic-ai-skills` 1.2.0](https://github.com/DougTrajano/pydantic-ai-skills/releases/tag/v1.2.0) broadened resource discovery from a fixed seven-extension allowlist (`.md/.json/.yaml/.yml/.csv/.xml/.txt`) to **every readable text file** in a skill bundle, so `.sql`, `.toml`, `.jinja`, and similar assets are exposed with no extra config ([DougTrajano/pydantic-ai-skills#54](https://github.com/DougTrajano/pydantic-ai-skills/issues/54), [#55](https://github.com/DougTrajano/pydantic-ai-skills/pull/55)).

That broader discovery also means more files reach the model, so `exclude_resources` is the trim knob for secrets or scratch files that happen to live in a skill directory. The two ship together: broaden discovery, and give users a way to narrow it back.

## Usage

```python
from airflow.providers.common.ai.toolsets.skills import AgentSkillsToolset

toolset = AgentSkillsToolset(
    sources=["./skills"],
    exclude_resources=["*.env", "secrets/*"],  # added on top of the built-in defaults
)
```

Patterns extend the built-in defaults (`__pycache__`, `*.pyc`, `*.pyo`, `.DS_Store`, `.git`) and match either the full skill-relative path or any single path component.

## Scope of the guarantee

`exclude_resources` hides files from **resource discovery only** — the resource list and the `read_skill_resource` tool. It does **not** stop a skill's `run_skill_script` from reading an excluded file off disk (a skill script runs as a subprocess with the worker's filesystem permissions). When files are genuinely sensitive, pair it with `exclude_tools={"run_skill_script"}`. This is called out in both the docstring and the toolset guide.

## Packaging notes

- The `[skills]` extra floor moves to `pydantic-ai-skills>=1.2.0` (the version that adds the parameter).
- 1.2.0 was published inside the repo's `exclude-newer = "4 days"` window, so a **temporary** per-package `exclude-newer-package` cutoff (`2026-07-16T00:00:00Z`) is added in both `[tool.uv.exclude-newer-package]` and `[tool.uv.pip.exclude-newer-package]` so it resolves. The comment marks it for removal once the rolling window advances past 2026-07-15.
- Regenerating `uv.lock` for the floor bump also picked up **pre-existing drift on `main`** unrelated to this change: the `openai` provider had already dropped its `[datalib]` extra in `pyproject.toml` but the lock was never refreshed, and `common-dataquality` is re-sorted. `uv lock` always fully syncs, so this churn is unavoidable when touching the lock.

### DIFF
--- a/providers/common/ai/docs/index.rst
+++ b/providers/common/ai/docs/index.rst
@@ -246,7 +246,7 @@ Extra           Dependencies
 ``openai``      ``pydantic-ai-slim[openai]``
 ``mcp``         ``pydantic-ai-slim[mcp]``
 ``code-mode``   ``pydantic-ai-harness[codemode]>=0.3.0``
-``skills``      ``apache-airflow-providers-git>=0.4.0``, ``pydantic-ai-skills>=0.11.0``
+``skills``      ``apache-airflow-providers-git>=0.4.0``, ``pydantic-ai-skills>=1.2.0``
 ``avro``        ``fastavro>=1.10.0; python_version < "3.14"``, ``fastavro>=1.12.1; python_version >= "3.14"``
 ``parquet``     ``pyarrow>=18.0.0; python_version < '3.14'``, ``pyarrow>=22.0.0; python_version >= '3.14'``
 ``sql``         ``apache-airflow-providers-common-sql``, ``sqlglot>=30.0.0``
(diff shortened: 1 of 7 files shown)


gopidesupavan (approved):
Nice :)

PR #69925, opened 2026-07-15
Fix TimeSensor start_from_trigger behavior

### LINKED ISSUE #69543: dag_version inflection when using TimeSensor with `start_from_trigger = True`

RehanAhmad25:
Closes: #69543

## What does this PR do?

`TimeSensor` computed its target datetime in `__init__`, which runs at DAG parse time using `datetime.now(self.dag.timezone)`. When `start_from_trigger=True`, this value was stored directly in `start_trigger_args.trigger_kwargs`, which is serialized as part of the DAG. Because the value changed on every parse, the serialized DAG hash also changed on every parse, continuously creating new DAG versions.

This PR:

- Raises a `ValueError` when `start_from_trigger=True` is used with `TimeSensor`, since the target moment cannot be safely computed at parse time. `deferrable=True` remains fully supported and unaffected.
- Changes `target_datetime` into a property that is computed on access instead of being fixed during `__init__`.
- Updates the `TimeSensor` documentation to describe the actual behavior, which uses the current wall-clock time in the DAG's timezone, instead of the previously documented `data_interval_end`/`run_after`-relative behavior that was never implemented.

- [X] Yes (Claude)

Generated-by: Claude (Anthropic) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -62,34 +62,36 @@ def __init__(
...
         **kwargs,
     ) -> None:
+        if start_from_trigger:
+            raise ValueError(
+                "TimeSensor does not support start_from_trigger=True. The target moment is "
+                "computed fresh from the current wall-clock time on every Dag parse, so baking "
+                "it into the serialized trigger arguments makes the serialized Dag hash change "
+                "on every parse. Use deferrable=True instead, which computes the target moment "
+                "at task execution time and does not have this problem."
+            )
```
aaron-y-chen, line 74: I'm afraid this could cause a backward compatibility issue. Perhaps we should not raise an error directly in the initialization phase.

aaron-y-chen (commented):
Thanks for the PR :)

providers/standard/src/airflow/providers/standard/sensors/time.py
RehanAhmad25: Good point, you're right, a hard error at parse time is too risky here. Updated it: `start_from_trigger=True` now just emits a deprecation warning and gets ignored (falls back to `False`), instead of raising. So existing DAGs using it won't break on upgrade, they'll just see a warning telling them to switch to `deferrable=True`. Updated the test and docs to match. Let me know if that works better.

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -101,10 +105,9 @@ def execute_complete(self, context: Context, event: Any = None) -> None:
         return None
 
     def poke(self, context: Context) -> bool:
-        self.log.info("Checking if the time (%s) has come", self.target_datetime)
-
-        # self.target_date has been converted to UTC, so we do not need to convert timezone
-        return timezone.utcnow() > self.target_datetime
+        target_datetime = self.target_datetime
+        self.log.info("Checking if the time (%s) has come", target_datetime)
+        return timezone.utcnow() > target_datetime
```
uranusjr, line 110: This change doesn’t seem to be meaningful? (Unless it’s doing a race condition)

I’d revert it.

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -62,34 +62,38 @@ def __init__(
...
+                AirflowProviderDeprecationWarning,
+                stacklevel=2,
+            )
         super().__init__(**kwargs)
 
-        # Create a "date-aware" timestamp that will be used as the "target_datetime". This is a requirement
-        # of the DateTimeTrigger
+        self.target_time = target_time
+        self.deferrable = deferrable
+        self.start_from_trigger = False
```
uranusjr, line 79: Can this attribute be removed if it’s always False? And can it be set to True?

providers/standard/src/airflow/providers/standard/sensors/time.py
```diff
@@ -62,34 +62,38 @@ def __init__(
...
-                moment=self.target_datetime, end_from_trigger=self.end_from_trigger
-            )
+        return timezone.convert_to_utc(aware_time)
 
     def execute(self, context: Context) -> None:
         if self.deferrable:
             self.defer(
                 trigger=DateTimeTrigger(
-                    moment=self.target_datetime,  # This needs to be an aware timestamp
+                    moment=self.target_datetime,
```
uranusjr: Is the comment no longer valid?

RehanAhmad25: Good catch, that comment should still be there, it was dropped by accident in the refactor, not intentionally. I'll restore it.

providers/standard/src/airflow/providers/standard/sensors/time.py
RehanAhmad25: No, it can't be set to True anymore, it'll always be False now. With the switch to a deprecation warning instead of a hard error, passing True just triggers the warning and silently falls back to False. Keeping the attribute itself around for backward compatibility, since TimeSensorAsync in this same file already does the same thing (deprecated flag kept as an inert attribute instead of being removed outright).

providers/standard/src/airflow/providers/standard/sensors/time.py
RehanAhmad25: This one's intentional, your parenthetical guess is right. target_datetime is now a computed property (calls datetime.now() fresh on every access), so calling self.target_datetime twice in this method could return two slightly different instants if the wall clock ticks past the target moment between the log call and the comparison. Storing it in a local variable once guarantees the logged value and the compared value are the same instant. I'd like to keep this as-is rather than revert, let me know if you still see it differently.

providers/standard/src/airflow/providers/standard/sensors/time.py
uranusjr: Keeping the value is a good idea, but I don’t like it being an attribute since the user would try to change it and see it’s silently ignored. I would change it to a property instead (with a deprecation warning) so setting it emits an error.

uranusjr (changes requested):
This and https://github.com/apache/airflow/pull/69746 fix the same thing in different ways. We should first figure out the correct approach in the issue.

providers/standard/src/airflow/providers/standard/sensors/time.py
RehanAhmad25: Good point, updated it: `start_from_trigger` is now a property with a setter instead of a plain attribute. Reading it still always returns `False`, but setting it (whether at construction or afterward, e.g. `op.start_from_trigger = True`) now emits the same deprecation warning each time rather than only warning once and then silently ignoring further attempts. Tests updated to cover both the constructor path and direct assignment after construction.

RehanAhmad25:
> This and #69746 fix the same thing in different ways. We should first figure out the correct approach in the issue.

Agreed, let's settle this on the issue so it's not split across two threads.

For what it's worth, my read on the tradeoff: #69746's TimeOfDayTrigger genuinely fixes more than mine does, it caught and fixed the class-level `start_trigger_args` mutation bug independently, which mine only sidesteps (the buggy code path becomes dead code once `start_from_trigger` is deprecated, but I didn't find that bug myself). That's a real point in its favor.

My case for the simpler deprecation approach: `start_from_trigger` avoids one worker-slot poke/defer cycle by starting the sensor directly on the triggerer, a fairly narrow optimization. Preserving it costs a new trigger class, lazy-resolution logic, and DST edge case handling (spring-forward gaps, fall-back ambiguity) that's now permanent surface area to maintain and could itself grow bugs over time. Given how narrow the benefit is, I think removing it is proportionate, but I'm not tied to that if the consensus lands the other way, happy to adapt or step back from this fix if #69746 is the preferred direction.

For reference, the current state on this PR: `start_from_trigger` is now a property (always reads `False`) with a setter that emits a deprecation warning on any assignment, at construction or afterward, so nothing silently gets ignored without a signal. I'll post this same comparison on the issue.

Vamsi-klu:
@RehanAhmad25 @uranusjr I posted a comparison on https://github.com/apache/airflow/issues/69543 recommending #69746 (keep start_from_trigger, lazy TimeOfDayTrigger) over this PR (deprecate/ignore the flag). That matches @uranusjr's Aug 17 comment on the issue and @potiuk's earlier request not to drop a documented API without a major bump.

If that lands, this PR can be closed as superseded. Your Aug 17 note that you would open a new lazy-resolution PR is appreciated; #69746 already implements that design, so a third PR would duplicate review.


---
Drafted-by: Cursor Grok 4.6; reviewed by @Vamsi-klu before posting

RehanAhmad25:
@Vamsi-klu  I went through your PR pretty thoroughly and didn’t find any major issues that I felt needed to be flagged.

Since your PR already explores the lazy-resolution approach, opening another PR from my side would essentially duplicate the same direction of work and could make it harder for the maintainers to review and decide between overlapping PRs. That’s why I didn’t open a separate lazy-resolution PR.

Regarding @uranusjr’s earlier review, my understanding was that the preference was to try fixing the underlying `start_from_trigger` issue first, rather than deciding to remove the feature outright. Since it’s still not completely settled whether the feature can/should be retained, I’ve kept my PR open for now.

If #69746 ends up being the preferred approach and the feature is retained, I’m absolutely fine with my PR being superseded and closed. I just didn’t want to prematurely close it while the direction is still being worked out.

PR #69926, opened 2026-07-15
Go SDK: Keep workload tokens out of task error logs

viiccwen:
When the Go SDK worker received an HTTP error while starting a task, it logged the complete request headers. Because the API client carries the task-scoped bearer token in the `Authorization` header, the token was included in the task error log.

Limit the logged request details to an explicit allowlist containing the request path, method, and correlation ID. This preserves useful diagnostics without including request credentials, query parameters, or other headers.

The regression test sends a real HTTP request containing a bearer token to a test server, returns an HTTP 400 response, and verifies both that the request carried the token and that the resulting structured log contains neither the token nor the `Authorization` header.

---

- [x] Yes - GPT-5.6 Sol

---

### DIFF
--- a/go-sdk/pkg/worker/runner.go
+++ b/go-sdk/pkg/worker/runner.go
@@ -273,12 +273,9 @@ func (w *worker) ExecuteTaskWorkload(ctx context.Context, workload api.ExecuteTa
 				slog.Int("status_code", resp.StatusCode()),
 				slog.Group(
 					"request",
-					"url",
-					resp.Request.URL,
-					"method",
-					resp.Request.Method,
-					"headers",
-					resp.Request.Header,
+					slog.String("path", resp.Request.RawRequest.URL.Path),
+					slog.String("method", resp.Request.Method),
+					slog.String("correlation_id", resp.Request.Header.Get("Correlation-Id")),
 				),
 			)
 		}
--- a/go-sdk/pkg/worker/runner_test.go
+++ b/go-sdk/pkg/worker/runner_test.go
@@ -18,7 +18,9 @@
 package worker_test
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"net/http"
@@ -31,6 +33,7 @@ import (
 	"github.com/jarcoal/httpmock"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"resty.dev/v3"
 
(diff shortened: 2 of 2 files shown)


go-sdk/pkg/worker/runner_test.go
```diff
@@ -71,6 +75,80 @@ func newTestWorkLoad(id string, dagId string) api.ExecuteTaskWorkload {
...
+		"status_code": float64(http.StatusBadRequest),
+		"task_id":     workload.TI.TaskId,
+		"ti_id":       workload.TI.Id.String(),
+		"request": map[string]any{
+			"correlation_id": request.correlationID,
+			"method":         http.MethodPatch,
+			"path":           request.path,
+		},
+	}, logEntry)
+}
```
jason810496, line 134: Would it be better to just mock the error itself instead of exercising the whole execution path?

go-sdk/pkg/worker/runner.go
```diff
@@ -273,12 +273,9 @@ func (w *worker) ExecuteTaskWorkload(ctx context.Context, workload api.ExecuteTa
 				slog.Int("status_code", resp.StatusCode()),
 				slog.Group(
 					"request",
-					"url",
-					resp.Request.URL,
-					"method",
-					resp.Request.Method,
-					"headers",
-					resp.Request.Header,
+					slog.String("path", resp.Request.RawRequest.URL.Path),
```
jason810496, line 276: Any reason to not showing the whole URL?

jason810496 (commented):
Thanks! LGTM overall.

go-sdk/pkg/worker/runner_test.go
```diff
@@ -71,6 +75,80 @@ func newTestWorkLoad(id string, dagId string) api.ExecuteTaskWorkload {
...
+	case <-ctx.Done():
+		t.Fatal("start-task request was not received")
+	}
+	require.Equal(t, "Bearer "+secretToken, request.authorization)
+	require.NotEmpty(t, request.correlationID)
+	require.Equal(t, http.MethodPatch, request.method)
+	require.Equal(t, "/execution/task-instances/"+workload.TI.Id.String()+"/run", request.path)
+
+	var logEntry map[string]any
+	require.NoError(t, json.Unmarshal(logBuffer.Bytes(), &logEntry))
```
uranusjr, line 127: Probably can also do

```go
require.NotContains(t, logBuffer.String(), secretToken)
```

go-sdk/pkg/worker/runner_test.go
```diff
@@ -71,6 +75,80 @@ func newTestWorkLoad(id string, dagId string) api.ExecuteTaskWorkload {
...
+	require.Equal(t, "ERROR", logEntry["level"])
+	require.Equal(t, "Server reported error when attempting to start task", logEntry["msg"])
+	delete(logEntry, "time")
+	delete(logEntry, "level")
+	delete(logEntry, "msg")
+	require.Equal(t, map[string]any{
+		"dag_id":      workload.TI.DagId,
+		"error":       "client error '400 Bad Request' map[detail:simulated start failure]",
```
uranusjr: This feels somewhat brittle since it’s entirely controlled by the HTTP library. Maybe it’s fine; I have no idea how stable this is.

go-sdk/pkg/worker/runner_test.go
viiccwen: Done. The test now constructs the GeneralHTTPError directly and mocks the task-instances client, instead of relying on an HTTP test server.

go-sdk/pkg/worker/runner.go
viiccwen: At that time, I think full url may contain sensitive query parameters or embedded credentials. The test intentionally puts the workload token in the query string to cover that case. Logging the path, method, and correlation ID preserves the useful request context without exposing credentials.

go-sdk/pkg/worker/runner_test.go
viiccwen: Sure, already added. The token is present in both the authorization header and URL query string in the test request.

go-sdk/pkg/worker/runner_test.go
viiccwen: I agreed. now, the test no longer relies on Resty’s formatted error string or an HTTP server response. It constructs the error response needed for this branch and asserts the structured log fields instead.

viiccwen:
Hello @jason810496 @uranusjr, politely tag for further review.

jason810496 (commented):

PR #67688, opened 2026-05-29
Optimize TaskGroup.topological_sort for reverse-declared Dags

shahar1:
## Human Summary

Further optimization of `TaskGroup.topological_sort` to handle reverse-declared Dags (where many children are declared before their dependencies) efficiently. This is the follow-up to PR #67288.

### Explanation
The idea is to optimize `TaskGroup.topological_sort` for Dags whose tasks are declared in reverse of their dependency order, for example:
```python

@dag
def dag():
    task3 = EmptyOperator()
    task2 = EmptyOperator()
    task1 = EmptyOperator()

    task1 >> task2 >> task3
```

Note that we insert `task3`, then `task2`, then `task1`, but their topological order is the other way round.

The issue with this shape is that the current sweep algorithm revisits blocked nodes. In a reversed-declared chain, we end up in a situation where total of `N + (N-1) + (N-2) + ... + 1` nodes are revisited, leading to a time complexity of O(N^2).
To solve this, we first use [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm) to compute the pass number at which each node would be emitted by the existing sweep algorithm. Once the pass numbers are known, we sort nodes by `(pass_number, insertion_order)`, which produces the exact same ordering as the sweep algorithm while avoiding the repeated rescans.

Since the pass-numbering approach has a higher constant overhead than the sweep algorithm, we only enable it for TaskGroups with a significant number of back edges (`nodes_with_back_edge >= 32`), where the cost of repeated sweeps is expected to outweigh that overhead (backed up by a [benchmark script](https://gist.github.com/shahar1/879dea2621e95b5f3e6927378a68c2aa)).

### Was generative AI tooling used to co-author this PR?
- [X] Yes — Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

## AI Summary
<details>
<summary>Click here</summary>
Addresses the O(N²) worst-case behavior of the greedy-sweep approach on adversarial Dag shapes such as reverse-insertion chains.

Uses a hybrid strategy:

- **Forward-declared Dags** (common case): greedy multi-pass sweep, O(V + E)
- **Reverse-declared Dags** (many back edges): pass-number traversal via Kahn's algorithm, O((V + E) log V)
[...]

PR #67288, opened 2026-05-21
Speed up TaskGroup.topological_sort with int-indexed projected sweep

shahar1:
## Summary

Replaces the modified-Kahn body of `TaskGroup.topological_sort` (and the mirror in `SerializedTaskGroup`) with a projected sweep:

- Each child's per-task `upstream_task_ids` is projected onto sibling-level integer indices **once** up front (`_project_child_deps`).
- The sweep (`_sweep_projection`) then runs against the projection with a `bytearray`-backed emission flag - per-edge work happens once per sort instead of once per outer-loop pass.

Emission order is identical to the previous implementation. The existing order-sensitive tests (`test_topological_sort1/2`, `test_topological_nested_groups`, `test_topological_group_dep`) cover the contract.

## Benchmark

Min of five runs, N=2000 children, ms per call. Source: <https://gist.github.com/shahar1/9c61dc9f34f7e77cd29cfb9d67af7ceb>

| Shape | main | PR | speedup |
|---|---:|---:|---:|
| chain (forward-declared) | 1.39 | 0.75 | **1.84x** |
| diamond (root → many → sink) | 2.20 | 0.86 | **2.56x** |
| independent (no deps) | 0.65 | 0.35 | **1.84x** |
| layered (4-layer full bipartite) | 230.96 | 67.15 | **3.44x** |
| nested (depth-3, walked across all groups) | 17.78 | 4.77 | **3.73x** |
| rev-chain (adversarial reverse insertion) | 894.04 | 114.16 | **7.83x** |

The 7.8x on rev-chain comes from the projection precompute alone (one projection vs N projections); algorithmic complexity is unchanged and remains O(N²) on that shape until the follow-up PR.

The nested row times a recursive walk across all 39 nested groups (depth=3, ~222 tasks each); the root-only timing was dominated by setup overhead and was previously labelled `~noise` in this table.

## Test plan

- [x] `task-sdk/tests/task_sdk/definitions/test_taskgroup.py` + `test_dag.py` — 145 passed (18 new shape-correctness cases)
- [x] `airflow-core/tests/unit/utils/test_task_group.py` — 20 passed (1 new serialization round-trip case)
- [x] `airflow-core/tests/unit/serialization/` (task-group / topological filter) — 4 passed
- [x] `prek run mypy-task-sdk` — passed
- [x] `prek run mypy-airflow-core` — passed
- [x] `prek run --from-ref upstream/main --stage pre-commit` — passed

---

##### Was generative AI tooling used to co-author this PR?

[...]

PR #69933, opened 2026-07-15
Fix grid/graph view topological sort for group-level and cross-group dependencies

### LINKED ISSUE #65291: Tasks in the Grid view are not in topological order

Airflow Core

### Apache Airflow version

3.2.0+astro.1

When setting dependencies between a list of task groups and another task group, the grid does **NOT** render the task groups in topological order.

This is the Dag where the list of task groups is `c` and the singular task group is `a`.

```python
import pendulum
from airflow.sdk import dag, task, TaskGroup
from airflow.providers.standard.operators.python import PythonOperator

@dag(
    schedule=None,
    dag_id="grid",
)
def grid():

    with TaskGroup(group_id="a") as a:
        even = PythonOperator(task_id="get_even", python_callable=lambda: None)
        odd = PythonOperator(task_id="get_odd", python_callable=lambda: None)
        even >> odd

    c = []
    for x in range(3):
        with TaskGroup(group_id=f"b_{x}") as b:
            even = PythonOperator(task_id="get_even", python_callable=lambda: None)
            odd = PythonOperator(task_id="get_odd", python_callable=lambda: None)
            even >> odd
        c.append(b)
    c >> a

grid()
```

This is the graph render of the Dag.

Image

This is the grid render of the Dag. Notice the task groups (`b_0`, `b_1`, and `b_2`) is **BELOW** the task group `a`.

Image

When the dependencies are set between individual task groups, the grid view does render the task groups topologically. I expect this Dag to render the task groups in the same order as below.

```python
import pendulum
from airflow.sdk import dag, task, TaskGroup
from airflow.providers.standard.operators.python import PythonOperator

@dag(
    schedule=None,
    dag_id="grid",
)
def grid():

    with TaskGroup(group_id="a") as a:
        even = PythonOperator(task_id="get_even", python_callable=lambda: None)
        odd = PythonOperator(task_id="get_odd", python_callable=lambda: None)
        even >> odd

    for x in range(3):
        with TaskGroup(group_id=f"b_{x}") as b:
            even = PythonOperator(task_id="get_even", python_callable=lambda: None)
            odd = PythonOperator(task_id="get_odd", python_callable=lambda: None)
            even >> odd
        b >> a

grid()
```

This is the graph render of the Dag.

Image

This is the grid render of the Dag. Notice the task groups (`b_0`, `b_1`, and `b_2`) is **ABOVE** the task group `a`.

Image

### Operating System

Debian GNU/Linux 13 (trixie)

### Deployment

Astronomer

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

I suspect this is related to https://github.com/apache/airflow/pull/56963.

### Code of Conduct

hkc-8010:
When a TaskGroup depends on another TaskGroup through a list (`[group_a, group_b] >> group_c`)
or through a task-level dependency that crosses into another group's entry task, the grid/graph
view sorts the groups as if there were no dependency at all. `_project_child_deps` only looks at
a group's own `upstream_task_ids`, which stays empty in both cases: group-to-group edges only
populate `upstream_group_ids`, and a task-level edge into another group's entry task never
touches the group object itself.

This also pulls in a group's `upstream_group_ids` and its root tasks' upstream ids before
projecting sibling dependencies, in both the serialization-layer sort (`airflow-core`) and the
design-time sort (`task-sdk`) - they share the same algorithm since the #67288/#67688
performance rewrite.

closes: #65291

related: #67964 and #65639 both attempted this fix earlier but were written against the
pre-rewrite `topological_sort` implementation (Kahn's-algorithm style), which no longer exists
after #67288 and #67688 replaced it with the sweep/pass-numbering projection used today. Both
were closed for inactivity/never reviewed rather than merged.

## Changes

- `airflow-core/src/airflow/serialization/definitions/taskgroup.py`: `_project_child_deps` also
  considers group-to-group edges (`upstream_group_ids`) and cross-group task-level edges (root
  tasks' `upstream_task_ids`)
- `task-sdk/src/airflow/sdk/definitions/taskgroup.py`: same fix, mirrored design-time sort
- Tests for both new scenarios, plus two existing tests whose hardcoded expected orderings were
  corrected (they were encoding the old, buggy order)

---

- [X] Yes (Claude Code)

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
@@ -200,15 +200,22 @@ def get_dag_structure(
     run_ids = list(session.scalars(dag_runs_select_filter))
 
     task_group_sort = get_task_group_children_getter()
+    latest_group_dict = latest_dag.task_group.get_task_group_dict()
     if not run_ids:
-        nodes = [task_group_to_dict_grid(x) for x in task_group_sort(latest_dag.task_group)]
+        nodes = [
+            task_group_to_dict_grid(x, group_dict=latest_group_dict)
+            for x in task_group_sort(latest_dag.task_group, latest_group_dict)
+        ]
         return [GridNodeResponse(**n) for n in nodes]
 
     # Process and merge the latest serdag first
     merged_nodes: list[dict[str, Any]] = []
-    nodes = [task_group_to_dict_grid(x) for x in task_group_sort(latest_dag.task_group)]
+    nodes = [
+        task_group_to_dict_grid(x, group_dict=latest_group_dict)
+        for x in task_group_sort(latest_dag.task_group, latest_group_dict)
+    ]
     _merge_node_dicts(merged_nodes, nodes)
-    del latest_dag
+    del latest_dag, latest_group_dict
 
     # Process serdags one by one and merge immediately to reduce memory usage.
     # Use yield_per() for streaming results and expunge each serdag after processing
@@ -243,7 +250,11 @@ def get_dag_structure(
                 depth=depth,
             )
         # Merge immediately instead of collecting all Dags in memory
-        nodes = [task_group_to_dict_grid(x) for x in task_group_sort(filtered_dag.task_group)]
+        filtered_group_dict = filtered_dag.task_group.get_task_group_dict()
+        nodes = [
+            task_group_to_dict_grid(x, group_dict=filtered_group_dict)
+            for x in task_group_sort(filtered_dag.task_group, filtered_group_dict)
+        ]
         _merge_node_dicts(merged_nodes, nodes)
 
         session.expunge(serdag)  # to allow garbage collection
--- a/airflow-core/src/airflow/serialization/definitions/taskgroup.py
+++ b/airflow-core/src/airflow/serialization/definitions/taskgroup.py
@@ -231,11 +231,12 @@ class SerializedTaskGroup(DAGNode):
         nodes = list(children.values())
         n = len(nodes)
         id_to_idx = {nid: i for i, nid in enumerate(children)}
+        group_dict = self.dag.task_group.get_task_group_dict()
 
         projected: list[tuple[int, ...]] = [()] * n
         nodes_with_back_edge = 0
         for i, child in enumerate(nodes):
-            deps = self._project_child_deps(i, child, id_to_idx)
+            deps = self._project_child_deps(i, child, id_to_idx, group_dict)
             if deps:
                 projected[i] = deps
                 if any(d > i for d in deps):
@@ -248,9 +249,24 @@ class SerializedTaskGroup(DAGNode):
         return self._sweep_projection(nodes, projected)
 
     def _project_child_deps(
-        self, child_idx: int, child: DAGNode, id_to_idx: dict[str, int]
+        self,
+        child_idx: int,
+        child: DAGNode,
+        id_to_idx: dict[str, int],
+        group_dict: dict[str | None, SerializedTaskGroup],
     ) -> tuple[int, ...]:
-        upstream_ids = child.upstream_task_ids
+        if isinstance(child, SerializedTaskGroup):
+            # A group's own upstream_task_ids only reflects direct group-to-task edges.
+            # Group-to-group edges (`group_a >> group_b`, list or individual) only populate
+            # upstream_group_ids, and task-level edges crossing into the group (a sibling
+            # task feeding one of this group's entry tasks) never touch the group at all —
+            # both need to be pulled in explicitly here.
+            upstream_ids: set[str] = set(child.upstream_task_ids)
+            upstream_ids.update(gid for gid in child.upstream_group_ids if gid is not None)
+            for root_task in child.get_roots():
+                upstream_ids.update(root_task.upstream_task_ids)
+        else:
+            upstream_ids = child.upstream_task_ids
         if not upstream_ids:
             return ()
         sib_deps: set[int] = set()
@@ -260,8 +276,10 @@ class SerializedTaskGroup(DAGNode):
                 if j != child_idx:
                     sib_deps.add(j)
                 continue
-            edge = self.dag.get_task(edge_id)
-            tg = edge.task_group
+            tg = group_dict.get(edge_id)
+            if tg is None:
+                edge = self.dag.get_task(edge_id)
+                tg = edge.task_group
             while tg is not None:
                 anc_idx = id_to_idx.get(tg.node_id)
                 if anc_idx is not None:
--- a/airflow-core/src/airflow/serialization/definitions/taskgroup.py
+++ b/airflow-core/src/airflow/serialization/definitions/taskgroup.py
@@ -173,8 +173,14 @@ class SerializedTaskGroup(DAGNode):
                     yield from recurse_for_first_non_teardown(task)
 
     def get_task_group_dict(self) -> dict[str | None, SerializedTaskGroup]:
-        """Create a flat dict of group_id: TaskGroup."""
+        """Create a flat dict of group_id: TaskGroup. Cached per instance/DAG."""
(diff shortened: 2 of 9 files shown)


airflow-core/tests/unit/utils/test_task_group.py
```diff
@@ -493,17 +493,11 @@ def test_task_group_to_dict_and_dag_edges(dag_maker):
     nodes = task_group_to_dict(dag.task_group)
     edges = dag_edges(dag)
 
+    # group_d depends on group_c (`group_d << group_c`), so it must sort after group_c,
+    # not before task1 as it did prior to the #65291/#67964 topological-sort fix.
```
viiccwen: Could we mention the intention of the testing instead of using issue number?

airflow-core/tests/unit/utils/test_task_group.py
```diff
@@ -1178,6 +1183,80 @@ def test_topological_sort_serialized_layered():
                 )
 
 
+def test_topological_group_dep_list_syntax():
+    """List-based deps (`[b0, b1] >> a`) must produce the same topological order as individual deps.
+
+    Regression test for apache/airflow#65291: declaring a group dependency via a list
```
viiccwen: same here.

airflow-core/src/airflow/serialization/definitions/taskgroup.py
viiccwen, line 1: This rebuilds the complete Dag-level TaskGroup mapping every time `topological_sort()` is called. The Grid / Graph serializers recursively call `topological_sort()` for every nested TaskGroup, so a dag with G non-empty groups now performs approximately G full traversals of the same G-group tree. That changes the overall group-sorting cost to $O(G^2)$.

Could we build the mapping once for the entire recursive render and pass it through to each nested `topological_sort()` call? The same in the Task SDK's `topological_sort()` path.

viiccwen (commented):
left comments!

airflow-core/tests/unit/utils/test_task_group.py
hkc-8010: Done, reworded to describe the scenario being tested instead of citing the issue number.

airflow-core/tests/unit/utils/test_task_group.py
hkc-8010: Done here too. Fixed the same pattern in two other spots in this file that had the same issue but weren't directly flagged.

airflow-core/src/airflow/serialization/definitions/taskgroup.py
hkc-8010: Good catch, confirmed this was O(G^2) — task_group_to_dict() recurses into every nested group and calls topological_sort() on each independently, and the Task SDK's own DAG.topological_sort() does the same.

Couldn't pass the map through as a literal parameter since topological_sort() is a public zero-arg method invoked externally by several independent recursive callers (task_group_to_dict, dag_edges, DAG.topological_sort()) — changing its signature would ripple into all of them for no real gain. Instead memoized get_task_group_dict() itself with methodtools.lru_cache, keyed per instance: dag.task_group is the same object for every nested group in one DAG, so the first call builds the map and every subsequent nested call is an O(1) hit. Applied the same fix in task-sdk.

One wrinkle: methodtools.lru_cache has no type stubs, so decorating get_task_group_dict directly widened its return type to Any for every caller, which actually broke mypy downstream in dag_edges.py. Kept the memoized logic behind a private helper with an explicitly-typed public wrapper so mypy still trusts the declared return type.

Added a test asserting the cache is actually hit (not rebuilt) across nested topological_sort() calls, in both packages.

airflow-core/tests/unit/utils/test_task_group.py
```diff
@@ -719,25 +721,27 @@ def section_2(value2):
...
+                    {"id": "section_1.task_1"},
+                    {"id": "section_1.task_2"},
                     {
                         "id": "section_1.section_2",
                         "children": [
                             {"id": "section_1.section_2.task_3"},
                             {"id": "section_1.section_2.task_4"},
                         ],
                     },
-                    {"id": "section_1.task_1"},
```
kaxil, line 744: cc @uranusjr Since I distinctly remember TP fixing this

airflow-core/src/airflow/serialization/definitions/taskgroup.py
```diff
@@ -248,9 +255,24 @@ def topological_sort(self) -> list[DAGNode]:
...
     def _project_child_deps(
-        self, child_idx: int, child: DAGNode, id_to_idx: dict[str, int]
+        self,
+        child_idx: int,
+        child: DAGNode,
+        id_to_idx: dict[str, int],
+        group_dict: dict[str | None, SerializedTaskGroup],
     ) -> tuple[int, ...]:
-        upstream_ids = child.upstream_task_ids
+        if isinstance(child, SerializedTaskGroup):
```
uranusjr: Why is this needed? This class _is_ SerializedTaskGroup, and this should always return True? When does the `else` block kick in?

airflow-core/tests/unit/utils/test_task_group.py
uranusjr: As pointed out above, a rewrite has changed things a bit (including test cases). The fix proposed here is logically correct.

uranusjr (approved):
I added a couple of commits to hoist common logic into the shared lib. This looks good enough to me now.

Also, I intentionally removed `methodtools.lru_cache` from `get_task_group_dict` since an SDK task group can change shape between calls and cause the cache to become stale. The serialized object is generally short-lived (the canonical source is in the database), so the cache is not generally too useful anyway.

uranusjr:
I added a call-level memo dict to achieve the same end result as the previous LRU cache (that is problematic for other reasons).

jason810496 (commented):
https://github.com/apache/airflow/pull/70573 was just merged -- which should resolve all the provider test failures in last run.

airflow-core/tests/unit/utils/test_task_group.py
```diff
@@ -1187,6 +1220,79 @@ def test_topological_sort_serialized_layered():
...
+
+        with TaskGroup("stage_a"):
+            a_start = EmptyOperator(task_id="a_start")
+            a_end = EmptyOperator(task_id="a_end")
+            a_start >> a_end
+
+        b_end >> a_start
+
+    serialized = create_scheduler_dag(dag)
+    order = [node.node_id for node in serialized.task_group.topological_sort()]
```
jason810496, line 1291: Non-blocking nit: We can construct as `dict[task_id, index]` instead of having a list to avoid `O(n)` `.index` call. (Though this only save several ms.)

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -200,15 +200,22 @@ def get_dag_structure(
     run_ids = list(session.scalars(dag_runs_select_filter))
 
     task_group_sort = get_task_group_children_getter()
+    latest_group_dict = latest_dag.task_group.get_task_group_dict()
```
jason810496, line 203: >  I added a call-level memo dict to achieve the same end result as the previous LRU cache (that is problematic for other reasons).

Would it be better to add explicit comment like:

```python
# Intentionally uncached: the latest_group_dict is a derived view of a mutable tree (a cache would go stale with no invalidation) and would pin the whole map for the group's lifetime, fighting the grid endpoint's streaming/expunge.
# Callers build it once per render and pass it down instead.
```

At least I wonder why not add the `lru_cache` at method level then I realize we will introduce the above regressions if we add the cache.

jason810496 (commented):
The patch LGTM, my comments are non-blocking and they won't effect the prod code.

jason810496 (approved):

jason810496:
Manually backport in https://github.com/apache/airflow/pull/70591.

PR #69937, opened 2026-07-15
Warn on suspicious Dag and task IDs at Java SDK build time

Andrushika:
## Why

The Java SDK accepts any Dag or task ID, but the Airflow server rejects IDs longer than 250 characters or containing anything other than letters, digits, `-`, `.` and `_`. Today a Java bundle author only finds out after deploying, when the task fails. The Go and TypeScript SDKs already warn about this at pack time.

## What

Add a `checkAirflowBundle` task to the Gradle plugin, wired into `check` (skip with `-x checkAirflowBundle`). It runs `BundleInspector`, which instantiates the configured `BundleBuilder`, builds the `Bundle`, and prints a warning for every Dag or task ID the server would reject. The warnings never fail the build, since the server stays the source of truth; only a bundle that cannot be built (for example duplicate Dag IDs) fails.

related: #69400, #69965, #70993

---

- [X] Yes — Claude Code (Fable 5.1)

Generated-by: Claude Code (Fable 5.1) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/java-sdk/processor/src/main/kotlin/org/apache/airflow/sdk/BuilderProcessor.kt
+++ b/java-sdk/processor/src/main/kotlin/org/apache/airflow/sdk/BuilderProcessor.kt
@@ -34,6 +34,7 @@ import javax.annotation.processing.RoundEnvironment
 import javax.annotation.processing.SupportedAnnotationTypes
 import javax.annotation.processing.SupportedSourceVersion
 import javax.lang.model.SourceVersion
+import javax.lang.model.element.Element
 import javax.lang.model.element.ExecutableElement
 import javax.lang.model.element.Modifier
 import javax.lang.model.element.TypeElement
@@ -92,6 +93,8 @@ class BuilderProcessor : AbstractProcessor() {
 
   private fun buildDag(el: TypeElement): TypeSpec {
     val ann = el.getAnnotation(Builder.Dag::class.java)!!
+    val dagId = ann.id.ifBlank { el.simpleName.toString() }
+    processingEnv.warnOnSuspiciousId(el, "Dag id \"$dagId\"", dagId)
 
     val builderClass =
       TypeSpec
@@ -103,13 +106,15 @@ class BuilderProcessor : AbstractProcessor() {
         .methodBuilder("build")
         .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
         .returns(ClassName.get(Dag::class.java))
-        .addStatement($$"var dag = new $T($S)", ClassName.get(Dag::class.java), ann.id.ifBlank { el.simpleName })
+        .addStatement($$"var dag = new $T($S)", ClassName.get(Dag::class.java), dagId)
 
     for (inner in el.enclosedElements) {
       if (inner !is ExecutableElement) continue
       if (inner.isVarArgs) throw IllegalArgumentException("Cannot create task from vararg function ${inner.simpleName}")
 
-      val ann = inner.getAnnotation(Builder.Task::class.java) ?: continue
+      val taskAnn = inner.getAnnotation(Builder.Task::class.java) ?: continue
+      val taskId = taskAnn.id.ifBlank { inner.simpleName.toString() }
+      processingEnv.warnOnSuspiciousId(inner, "Task id \"$taskId\" in dag \"$dagId\"", taskId)
       val innerName = inner.simpleName.toString().replaceFirstChar(Char::uppercase)
 
       val task = buildTask(innerName, inner, el)
@@ -117,7 +122,7 @@ class BuilderProcessor : AbstractProcessor() {
 
       buildMethod.addStatement(
         $$"dag.addTask($S, $L.class)",
-        ann.id.ifBlank { inner.simpleName },
+        taskId,
         innerName,
       )
     }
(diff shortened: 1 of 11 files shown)


java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Dag.kt
```diff
@@ -21,6 +21,21 @@ package org.apache.airflow.sdk
 
 import kotlin.Throws
 
+private val KEY_REGEX = Regex("""^[\p{L}\p{N}_.-]+$""")
```
jason810496: Ditto -- https://github.com/apache/airflow/pull/69965#discussion_r3608514915

jason810496 (commented):
Thank for the fix! I found that I need to fix the coordinator layer first (the interface to spawn the lang-sdk side subprocess) so both https://github.com/apache/airflow/pull/69965 and this one need to be on hold for a moment.

I will share more context here soon and let you know once the coordinator side patch merged. Thanks.

uranusjr:
If you use an invalid dag id, it would match nothing on the Python (stub) side. We can perhaps add some detection logic if you have a potential dag (or task) id mismatch between the stub and implementation, but honestly I’m not convinced we need to actively check for the character set. Doing this also adds the potential that the check may become stale if the allowed character set changes (it changed before).

If we really want to checks this, I would want to do this in the dag processor, and only in Python (i.e. the JAR or executable needs to send the id to Python to be validated, instead of validating directly in the language sdk).

jason810496 (commented):
> If you use an invalid dag id, it would match nothing on the Python (stub) side. We can perhaps add some detection logic if you have a potential dag (or task) id mismatch between the stub and implementation, but honestly I’m not convinced we need to actively check for the character set. Doing this also adds the potential that the check may become stale if the allowed character set changes (it changed before).

I see. It makes sense not to validate for the case of Mixed Lang Dag. From my perspective, the validation should be useful for the upcoming native Dag so no harm to introduce it. (Or we can on hold until the native Dag out) WDYT? Thanks.

Andrushika:
Oh, I see...... I have no problem putting this on hold or even closing it.

The reason I opened this PR is that I thought the behavior across lang SDKs should be aligned. I saw #69400 in ts-sdk, so I thought it would be good to introduce the same validation in java-sdk.

> If we really want to checks this, I would want to do this in the dag processor, and only in Python

For the current mixed-lang Dag, I agree that the Python stub is the source of truth, so validation in each lang SDK may not be needed.

> the validation should be useful for the upcoming native Dag so no harm to introduce it.

If we need validation in lang SDKs for native Dag in the future, the discrepancy TP mentioned would still exist. In that case, maybe we can keep the valid and invalid Dag/task IDs as shared test cases and let all the lang SDKs use them, instead of defining the expected behavior separately in each SDK?

potiuk:
@Andrushika — There are 1 unresolved review thread(s) on this PR from @jason810496. Could you either push a fix or reply in each thread explaining why the feedback doesn't apply? Once you believe the feedback is addressed, mark the thread as resolved so the reviewer isn't re-pinged needlessly. Thanks!

---

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Dag.kt
Andrushika: Ditto: https://github.com/apache/airflow/pull/69965#discussion_r3615283850
Since the PR is on hold, should I wait for the coordinator's design before making this change?

uranusjr:
> the validation should be useful for the upcoming native Dag so no harm to introduce it

Even in this case, the validation should happen in either the dag processor or the dag submission API endpoint (and surfaced in the dag processor), for the same reasons I mentioned in the second paragraph. This should not be implemented in every language.

uranusjr:
I see some of the PRs have already been merged without this discussion being resolved. They should be reverted.

Andrushika:
Hi TP! Here's the related PR, I tried to solve the same problem in go-sdk, and Jason gave some suggestion here: https://github.com/apache/airflow/pull/69965#pullrequestreview-4765061465

Jason suggests the "compile-time" best-effort validation: only shows a warning when an invalid id detected. I think this could prevent drift and preserve the source of truth while giving appropriate hints when users compile their lang-sdk stub.
I'm not sure if it meets expectations. If not, I can close this PR.

I've implemented in java-sdk and force-pushed in the latest commit currently: 
https://github.com/apache/airflow/pull/69937/changes#diff-8192c67224a3e477f432eaecc29f4978a37b653fc176d9c6b5f38d17111b4971R205-R232

Thanks!

Andrushika:
Hi @uranusjr @jason810496, just a gentle ping. Since the same validation has already been added to the Go SDK (#69965) and ts SDK (#70993), do you think the Java SDK should follow the same behavior?

I’ve implemented it in the latest commit. Could you please take a look when you have time? Thanks!

java-sdk/processor/src/main/kotlin/org/apache/airflow/sdk/BuilderProcessor.kt
```diff
@@ -194,6 +199,37 @@ private fun ProcessingEnvironment.isType(
   c: ClassName,
 ): Boolean = typeUtils.isSameType(t, elementUtils.getTypeElement(c.canonicalName()).asType())
 
+private const val MAX_ID_LENGTH = 250
+private val ID_REGEX = Regex("""^[\p{L}\p{N}_.-]+$""")
+
+private fun ProcessingEnvironment.warnOnSuspiciousId(
+  element: Element,
```
jason810496: It would be bette to move the validation into a dedicated module since we need to validation the interface based Dag as well.

jason810496 (commented):
Thanks for the update!

We need to validate the interface based Dag syntax as well. Actually, we only need to validate at the interface based syntax as the annotation based syntax will be transformed into interface based syntax anyway. Thanks.

Andrushika:
Thanks! I’ve pushed a commit that does the check in Bundle’s init, so both the interface based and annotation based syntaxes are covered. But this means the warnings show up at runtime in the task logs, not at build time. It is different from what we discussed above, and also different from the implementation in ts-sdk and go-sdk. 

This is because the java-sdk bundle does not have the ids in its manifest, so doing the validation at build time requires executing the built bundle to extract them.

If the build-time validation is strictly needed, I guess we need a similar introspection mode that runs the built bundle and returns the IDs to the Gradle plugin, as ts-sdk and go-sdk do. WDYT?

jason810496 (commented):
Good point. I think the Gradle-native shape for this would be a dedicated verification task.

Could we add something like `inspectAirflowBundle` / `checkAirflowBundle` that runs after `classes`, collects the Dag/task IDs from the assembled `Bundle`, and wire it into Gradle’s `check` lifecycle?

That way users can choose the level they want:
- `./gradlew checkAirflowBundle` to run only the Airflow bundle check
- `./gradlew check` to include it with normal verification
- `./gradlew check -x checkAirflowBundle` or a project property to skip it

This also gives us a better place to print more verbose guidance. Compilation can stay focused on source generation/compile errors, while the Airflow-specific validation lives in an explicit verification task.

The important bit is that the task probably still needs a `BundleInspector`-style path that instantiates the `BundleBuilder` and reads the resulting `Bundle`. Somehow like:

https://github.com/apache/airflow/blob/8fd35a18764faf8f9f876cceae4a0fb71b56bae0/java-sdk/adr/0003-pure-java-dags.md#L57-L100

Thanks.

Andrushika:
Thanks, Jason! I implemented as suggested: a `checkAirflowBundle` verification task that runs after `classes` and is wired into `check`, so `./gradlew checkAirflowBundle`, `check`, and `check -x checkAirflowBundle` all work. 

Could you please take another look when you have a chance?

java-sdk/processor/src/main/kotlin/org/apache/airflow/sdk/BuilderProcessor.kt
Andrushika: In the latest version, both interface based path will be validated!

Andrushika:
Gentle ping @jason810496 :)

PR #69940, opened 2026-07-15
Updated sort+filters for Assets List

bbovenzi:
Create a custom ui/assets list endpoint to give us more control. Such as:
- sort by last asset event
- default the sort by last asset event
- filter by group search
- filter by last asset event time range
- sorting by group

Also removed asset extra from the list page. We only need to show it on the asset details page

Screenshot 2026-07-15 at 5 29 02 PM

---

- [x] Yes Clause Sonnet 5

---


bbovenzi:
cc: @gopidesupavan I removed extra from the Assets list. After your work on data quality is that ok, or should we keep it?

gopidesupavan:
> cc: @gopidesupavan I removed extra from the Assets list. After your work on data quality is that ok, or should we keep it?

Yeah it should be fine.. i guess if we click on the asset and that will show us additional data field?

bbovenzi:
> > cc: @gopidesupavan I removed extra from the Assets list. After your work on data quality is that ok, or should we keep it?
> 
> Yeah it should be fine.. i guess if we click on the asset and that will show us additional data field?

Yes, its still on the details page.

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/assets.py
```diff
@@ -49,6 +71,83 @@
...
+            last_asset_event_timestamp_range,
+        ],
+        order_by=order_by,
+        offset=offset,
+        limit=limit,
+        session=session,
+    )
+
+    # CASE key keeps assets with no event (NULL timestamp) last in both directions.
+    order_columns: list[ColumnElement] = []
```
pierrejeambrun: I would keep things consistent with other APIs, nulls behavior is decided by the backend implementation (db).

Also a case prevent the leverage of indexes, so performance wise it's not great, that's why we removed a very similar piece of code from. the `parameters.py` module.

airflow-core/tests/unit/api_fastapi/core_api/routes/ui/test_assets.py
```diff
@@ -471,3 +480,315 @@ def test_non_partitioned_asset_inactive_true_when_deactivated(self, test_client,
         body = response.json()
         assert len(body["events"]) == 1
         assert body["events"][0]["asset_inactive"] is True
+
+
+class TestGetAssetsUi:
+    @pytest.fixture(autouse=True)
+    def cleanup_assets(self):
```
pierrejeambrun, line 487: Can we add some `assert_queries_count` guards just to make sure we don't have N+1 queries problem (by forgetting an eager loading or explicit join etc..)

airflow-core/src/airflow/api_fastapi/common/db/assets.py
```diff
@@ -0,0 +1,54 @@
...
+from airflow.models.asset import AssetEvent, AssetModel, AssetWatcherModel
+
+if TYPE_CHECKING:
+    from sqlalchemy.sql import Select
+
+
+def generate_assets_with_last_event_query() -> Select:
+    """Fetch Assets outer-joined to their latest AssetEvent id/timestamp."""
+    max_asset_event_id_query = (
+        select(AssetEvent.asset_id, func.max(AssetEvent.id).label("max_asset_event_id"))
```
pierrejeambrun, line 34: This changes the `max(timestamp)` to `max(id)` that's behavior change for the public endpoint. Is that expected?  Should we keep the former?

airflow-core/src/airflow/api_fastapi/common/parameters.py
```diff
@@ -499,14 +499,10 @@ def search_param_factory(
...
+        "Case-insensitive substring match; `%` / `_` wildcards allowed. "
         f"{pipe_clause}"
-        "Regular expressions are **not** supported. "
-        "\n\n"
-        "**Performance note:** this full-match pattern is evaluated as ``ILIKE '%term%'`` and "
-        "most of the time prevents the database from using B-tree indexes, which can be very "
-        "slow on large tables. Prefer the equivalent "
-        f"``{pattern_name.replace('_pattern', '_prefix_pattern')}`` parameter when possible."
+        f"Not index-friendly — prefer ``{pattern_name.replace('_pattern', '_prefix_pattern')}`` "
+        "on large tables."
```
pierrejeambrun: For both descriptoin update here I prefered the original one, which is more explicit.

That's not related directly to the PR maybe we should keep the original, it will also reduce the diff cause all generated search have this description embedded.

airflow-core/src/airflow/ui/src/pages/AssetsList/AssetsList.tsx
```diff
@@ -126,21 +110,36 @@ export const AssetsList = () => {
...
-  const { data, error, isLoading } = useAssetServiceGetAssets({
+  const lastAssetEventTimestampGte = searchParams.get(SearchParamsKeys.LAST_ASSET_EVENT_TIMESTAMP_GTE);
+  const lastAssetEventTimestampLte = searchParams.get(SearchParamsKeys.LAST_ASSET_EVENT_TIMESTAMP_LTE);
+  const groupArg = useAdvancedSearchArg({
+    patternApiKey: "groupPattern",
+    prefixApiKey: "groupPrefixPattern",
+    storageKey: SearchParamsKeys.GROUP_PATTERN,
+    value: searchParams.get(SearchParamsKeys.GROUP_PATTERN),
+  });
+
```
pierrejeambrun, line 128: We are missing the ability to switch from substring to prefix search for search fields:
Image

Image

pierrejeambrun (commented):
Overall looks just. Just a few suggestions

pierrejeambrun (approved):
LGTM thanks

airflow-core/src/airflow/api_fastapi/common/db/assets.py
```diff
@@ -0,0 +1,54 @@
...
+    return (
+        select(
+            AssetModel,
+            AssetEvent.id.label("last_asset_event_id"),
+            AssetEvent.timestamp.label("last_asset_event_timestamp"),
+        )
+        .outerjoin(max_asset_event_id_query, AssetModel.id == max_asset_event_id_query.c.asset_id)
+        .outerjoin(AssetEvent, AssetEvent.id == max_asset_event_id_query.c.max_asset_event_id)
+        .options(
+            subqueryload(AssetModel.scheduled_dags),
```
kaxil, line 48: Each of these five `subqueryload`s re-runs the whole parent statement as an inner subquery, `GROUP BY asset_event` and all. I dumped the executed SQL for a 2-row page and got six statements, five of which contain `SELECT asset_event.asset_id, max(asset_event.id) ... GROUP BY asset_event.asset_id`; with the count query that is seven aggregations of `asset_event` for one list page. `selectinload` emits a plain `WHERE asset_id IN (<page ids>)` instead and is already what the sibling helper in `common/db/dags.py` uses, so is there a reason not to switch here?

airflow-core/tests/unit/api_fastapi/core_api/routes/ui/test_assets.py
```diff
@@ -471,3 +480,278 @@ def test_non_partitioned_asset_inactive_true_when_deactivated(self, test_client,
...
+    def test_query_count(self, test_client, session):
+        """The asset relationships are eager-loaded, so the query count stays fixed regardless of
+        how many assets are returned (a lazy-loading regression would issue queries per asset)."""
+        for i in range(5):
+            asset = AssetModel(name=f"asset{i}", uri=f"s3://bucket/asset{i}", group="asset")
+            session.add(asset)
+            session.add(AssetActive.for_asset(asset))
+        session.commit()
+
+        with assert_queries_count(7):
```
kaxil, line 756: None of these five assets has a watcher, alias, scheduled dag or task, so every eager-loaded collection comes back empty and the nested `joinedload(AssetWatcherModel.trigger)` never fires. I checked by deleting that `.joinedload(...)` from `generate_assets_with_last_event_query`, and all 17 tests in this class still pass at 7, even though `from_asset_row` would then lazy-load `watcher.trigger` once per watcher. Could the fixture give at least one asset a watcher and an alias, so the guard covers the relationships it exists to protect?

airflow-core/src/airflow/ui/src/pages/AssetsList/AssetsList.tsx
```diff
@@ -129,18 +113,31 @@ export const AssetsList = () => {
...
+  const { data, error, isLoading } = useAssetServiceGetAssetsUi({
```
kaxil, line 129: This changes the key the list caches under, and `CreateAssetEventModal`'s success handler only invalidates `GetAssetEvents` plus the dag run and task instance keys. It never invalidated the old assets key either, so the play button on a row already left the timestamp stale, but now that last event is the default sort the row also stays in the wrong position. Adding `[useAssetServiceGetAssetsUiKey]` to that `queryKeys` array would cover both, alongside the `[useDagServiceGetDagsUiKey]` that is already there.

PR #69942, opened 2026-07-15
Fix dag.test() dropping defer-time kwargs when trigger yields an event

seanghaeli:
In dag.test(), a deferred task loses the kwargs it passed to defer() when its trigger fires; they get discarded on resume instead of kept. This adds the trigger event onto those kwargs rather than discarding them, so the resume method still receives them.

This makes it match how the triggerer already resumes deferred tasks in normal execution:

https://github.com/apache/airflow/blob/58b5a8064567afb54517ad523f751898ac39b18c/airflow-core/src/airflow/models/trigger.py#L534

Fixes the neptune_analytics deferrable test.


vincbeck (approved):

amoghrajesh (commented):
Nice catch.

PR #69944, opened 2026-07-16
Fix MySQL deadlock when queueing asset-triggered Dags

### LINKED ISSUE #69938: Asset-triggered DAG queueing can deadlock on MySQL and is not retried
### Apache Airflow version

main (3.4.0.dev), and all 2.x/3.x versions with data-aware scheduling

### What happened

When an asset (dataset) event fans out to multiple consumer DAGs, the scheduler/task
path inserts one row per consumer into `asset_dag_run_queue` (`dataset_dag_run_queue`
in 2.x). On MySQL/InnoDB, concurrent fan-outs contend on hose inserts and can raise:

```
(1213, 'Deadlock found when trying to get lock; try restarting transaction')
```

sometimes followed by a secondary error:

```
SAVEPOINT sa_savepoint_1 does not exist
```

The deadlock is **not retried**, so the enqueue of downstream runs for that event
fails. The `SAVEPOINT ... does not exist` line is a secondary symptom: InnoDB
discards the savepoint when it rolls back on deadlock, so SQLAlchemy's nested
transaction cleanup then fails against a savepoint that no longer exists.

### Root cause

`AssetManager._queue_dagruns` (`airflow-core/src/airflow/assets/manager.py:487`)
has two branches:

- **Postgres** — `_queue_dagruns_nonpartitioned_postgres` (line 811) does a single
  bulk `insert(...).on_conflict_do_nothing()`.
- **Everything else (MySQL)** — `_queue_dagruns_nonpartitioned_slow_path` (line 792)
  loops per row inside `session.begin_nested()` (SAVEPOINT) and catches **only
  `exc.IntegrityError`** (line 802).

Two problems with the non-Postgres path:

1. A deadlock is an `OperationalError` (errno 1213), **not** an `IntegrityError`, so
   it is not swallowed by the `except` at line 802 — it propagates.
2. Nothing on this call path (`register_asset_change` → `_queue_dagruns` →
   `_queue_dagruns_nonpartitioned_slow_path`) is wrapped in `@retry_db_transaction`,
   so there is no deadlock retry.

The per-row SAVEPOINT loop was introduced in #26103 (fix for #25210) specifically to
tolerate duplicate-key `IntegrityError` from concurrent producers of the same asset.
Deadlock handling was never in scope, and the row-by-row approach — many
lock-holding round-trips per event — actively increases InnoDB deadlock likelihood
compared to the single-statement Postgres path.

### What should happen

Fanning out an asset event to consumers should be resilient to InnoDB deadlocks:
the enqueue should either avoid the deadlock-prone construct or retry transparently,
so downstream runs are reliably queued.

### Proposed fix

1. **Give MySQL a bulk path**, mirroring Postgres, using
   `sqlalchemy.dialects.mysql.insert(...).on_duplicate_key_update(...)` (or
   `INSERT IGNORE`). This collapses N lock-holding round-trips into one statement and
   removes the per-row savepoint churn, sharply reducing the deadlock window. (This is
   not version-gated — the syntax has been available far below Airflow's MySQL 8.0+
   floor.)
2. **Add deadlock retry** to the queueing call (`@retry_db_transaction` or catch
   `OperationalError`/1213 and retry) so any residual InnoDB deadlock self-heals
   instead of surfacing as a failure. A single-statement bulk upsert still can
   deadlock on InnoDB gap/next-key locks, so (2) is the robust backstop and (1) is
   the probability reduction — both are worth doing.

The partitioned path should get the same treatment where applicable.

### Notes

- Postgres is unaffected (it already uses the bulk `ON CONFLICT` path).
- This is a robustness/enhancement issue, not a regression; behavior is unchanged on
  Postgres and functionally correct on MySQL except under concurrent fan-out.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @dstandish before posting

FrankYang0529:
## Why

- When an asset event fans out to multiple consumer Dags, the non-Postgres path inserts `asset_dag_run_queue` rows one by one (SAVEPOINT + `merge()` per row).
- Concurrent producers of the same asset insert the same rows in nondeterministic `set` order, so on MySQL it acquires index locks in opposite order and gets deadlock (errno 1213).

## How

- Update MySQL as the same single-statement bulk path Postgres already has: one multi-row upsert `INSERT ... ON DUPLICATE KEY UPDATE`.
- Insert queue rows sorted by `dag_id` on all backends, so concurrent fan-outs acquire row locks in a consistent order and cannot ABBA-deadlock each other.
- Add `@retry_db_transaction` to `Trigger.submit_event`.
 

* closes: #69938
---

- [X] Yes - Claude Code

---

### DIFF
--- a/airflow-core/src/airflow/assets/manager.py
+++ b/airflow-core/src/airflow/assets/manager.py
@@ -518,13 +518,21 @@ class AssetManager(LoggingMixin):
         # mapped) tasks update the same asset, this can fail with a unique
         # constraint violation.
         #
-        # If we support it, use ON CONFLICT to do nothing, otherwise
+        # If we support it, use ON CONFLICT / ON DUPLICATE KEY to do nothing, otherwise
         # "fallback" to running this in a nested transaction. This is needed
         # so that the adding of these rows happens in the same transaction
         # where `ti.state` is changed.
-        if get_dialect_name(session) == "postgresql":
-            return cls._queue_dagruns_nonpartitioned_postgres(asset_id, non_partitioned_dags, session)
-        return cls._queue_dagruns_nonpartitioned_slow_path(asset_id, non_partitioned_dags, session)
+        #
+        # Rows are inserted in sorted order so concurrent fan-outs acquire
+        # row locks in a consistent order, which prevents deadlocks (a
+        # set's iteration order differs between processes).
+        dag_ids = sorted(dag.dag_id for dag in non_partitioned_dags)
+        dialect = get_dialect_name(session)
+        if dialect == "postgresql":
+            return cls._queue_dagruns_nonpartitioned_postgres(asset_id, dag_ids, session)
+        if dialect == "mysql":
+            return cls._queue_dagruns_nonpartitioned_mysql(asset_id, dag_ids, session)
+        return cls._queue_dagruns_nonpartitioned_slow_path(asset_id, dag_ids, session)
 
     @classmethod
     def _queue_partitioned_dags(
@@ -790,10 +798,10 @@ class AssetManager(LoggingMixin):
 
     @classmethod
     def _queue_dagruns_nonpartitioned_slow_path(
-        cls, asset_id: int, dags_to_queue: set[DagModel], session: Session
+        cls, asset_id: int, dag_ids: list[str], session: Session
     ) -> None:
-        def _queue_dagrun_if_needed(dag: DagModel) -> str | None:
-            item = AssetDagRunQueue(target_dag_id=dag.dag_id, asset_id=asset_id)
+        for dag_id in dag_ids:
+            item = AssetDagRunQueue(target_dag_id=dag_id, asset_id=asset_id)
             # Don't error whole transaction when a single RunQueue item conflicts.
             # https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint
             try:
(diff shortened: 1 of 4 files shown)


viiccwen (commented):
Hello PoAn,

`@retry_db_transaction` retries the entire `submit_event()` method, but `AssetManager.register_asset_change()` performs non-transactional side effects before the queue insert that may deadlock.

In particular, it invokes `on_asset_changed`, `on_asset_event_emitted`, and incre `asset.updates` before calling `_queue_dagruns()`. If the queue insert raises a deadlock:
1. the listeners and metric have already run.
2. the database transaction is rolled back.
3. `submit_event()` is replayed.
4. the same listeners and metric run again for the eventually successful event.

Listener plugins may perform external, non-idempotent operations such as sending webhooks or publishing messages, so those effects cannot be rolled back with the database transaction. This can also replay listeners for earlier assets if a trigger is associated with multiple assets and a later asset fails.

The new retry test **mocks `AssetManager.register_asset_change()` entirely**, so it only verifies that the decorator retries and does not cover these duplicated side effects.

IMO, the retry boundary needs to exclude listener/metric emission, or those side effects need to be deferred until the database transaction has successfully committed. : D

FrankYang0529:
@viiccwen Thanks for the review. You're right. The listener hooks and the `asset.updates` metric run before `_queue_dagruns()` are not transactional and would run again on a replay.

Another path (task success → Execution API `ti_update_state` → `register_asset_changes_in_db`) retries when response is 5xx.

https://github.com/apache/airflow/blob/6ba529e0ebf0bd829bed78201a89b3a59ec809df/task-sdk/src/airflow/sdk/api/client.py#L1208-L1215

However, the deadlock error doesn't return 5xx. It only marks task as failed.

https://github.com/apache/airflow/blob/6ba529e0ebf0bd829bed78201a89b3a59ec809df/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py#L467-L478

I agree that we should remove `@retry_db_transaction` change to align the behavior.

Since issue author already created another PR https://github.com/apache/airflow/pull/69977, I will close this one.

PR #69959, opened 2026-07-16
Handle concurrent XCom writes with savepoint and UPDATE fallback to avoid 409

### LINKED ISSUE #69956: Concurrent/retried Execution API XCom writes fail with 409 due to duplicate primary key

Task SDK

### Apache Airflow version

3.2.2

An Airflow task runner attempted to store an XCom through the Airflow 3
Execution API:

POST /execution/xcoms/{dag_id}/{run_id}/{task_id}/{key}

The API server attempted the following statement:

INSERT INTO xcom
(dag_run_id, task_id, map_index, `key`, dag_id, run_id, value, timestamp)
VALUES (%s, %s, %s, %s, %s, %s, %s, %s)

The insert failed with an IntegrityError and the Execution API returned:

HTTP 409 Conflict

Relevant access log:

[2026-07-16T07:14:19.285110Z] request finished
method=POST
path=/execution/xcoms/{Run_ID}/runOnce__2026-07-16T07:01:26.419019+00:00/{Task_SENSOR}/{Task_SENSOR}
status_code=409

Three XCom INSERT failures occurred at approximately the same time.

The Airflow API exception handler classified the MySQL IntegrityError as:

"Unique constraint violation"

The XCom primary key is:

(dag_run_id, task_id, map_index, key)

In Airflow 3.2.2, `XComModel.set()` first deletes an existing XCom and
then inserts the replacement:

1. DELETE matching XCom
2. INSERT new XCom

With overlapping requests, both transactions can perform the DELETE
before either INSERT commits. One INSERT succeeds and another fails
with a duplicate primary-key error.

The approximately 81-second request duration may indicate that one
INSERT waited for another database transaction before the duplicate
conflict was returned.

Repeated or concurrent writes of the same XCom identity should be
handled idempotently.

Possible expected behavior:

- Use a database-appropriate atomic upsert.
- Serialize writes for the same XCom identity.
- If a repeated request contains the same value, treat it as already
  completed.
- Retry the write safely after an IntegrityError caused by the
  delete-then-insert race.

A transient/repeated Execution API request should not fail task
execution merely because the same XCom was already written.

### Operating System

linux

### Deployment

Docker-Compose

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

- Airflow: 3.2.2
- Python: 3.11
- Metadata database: MySQL 9.7.0

**### Relevant log output**

```text
[2026-07-16T07:14:19.283188Z] ERROR - Error with id tuw1pVaf,
statement: INSERT INTO xcom
(dag_run_id, task_id, map_index, `key`, dag_id, run_id, value, timestamp)
VALUES (%s, %s, %s, %s, %s, %s, %s, %s)

File "airflow/api_fastapi/execution_api/routes/xcoms.py", line 392,
in set_xcom
    XComModel.set(...)

File "airflow/models/xcom.py", line 246, in set
    session.flush()

File "MySQLdb/cursors.py", line 179, in execute
    res = self._query(mogrified_query)

[2026-07-16T07:14:19.285110Z] request finished
method=POST
path=/execution/xcoms/.../.../.../...
status_code=409
duration_us=81032908

### Code of Conduct

### PR DESCRIPTION
**Pull request by nagasrisai**
Closes #69956

## Summary

When two task runners (concurrent execution or task retry) write the same XCom key at roughly the same time, both call `XComModel.set` which does a DELETE followed by an INSERT.  Both DELETEs succeed (they each see the record that was just written by the peer or see nothing), then both try to INSERT the same primary key.  The second INSERT hits the unique constraint and the Execution API converts the `IntegrityError` to an HTTP 409, causing the task to fail.

### Fix

Wrap the `XComModel.set` call in a **savepoint** (`session.begin_nested()`) so that a constraint violation only rolls back the XCom write and not any earlier work in the same request (e.g. the `task_map` merge).  The `IntegrityError` is caught and handled with an explicit UPDATE that overwrites the existing entry with the latest value, which is the correct semantic for a retry.

```python
try:
    with session.begin_nested():
        XComModel.set(...)
except IntegrityError:
    # concurrent write already committed; overwrite with latest value
    session.execute(update(XComModel).where(...).values(value=value, dag_result=dag_result))
```

## Changes

- `airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py`
  - Add `update` to the `sqlalchemy` import
  - Add `IntegrityError` import from `sqlalchemy.exc`
  - Wrap `XComModel.set` in `session.begin_nested()` and catch `IntegrityError` with an UPDATE fallback

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @nagasrisai** · by `@potiuk` · 2026-07-28 16:11 UTC
>
> - :x: **Pre-commit / static checks**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).
> - :x: **Unresolved review comments**: 2 thread(s). See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).
>
>
>
> **Note:** your branch is **270 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>


potiuk:
@nagasrisai This PR has a few issues that need to be addressed before it can be reviewed — please see our [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).

**Issues found:**
- :x: **Pre-commit / static checks**: Failing: `CI image checks / Static checks`. Run `prek run --from-ref main --stage pre-commit` locally to reproduce and fix.

> **Note:** Your branch is **51 commits behind `main`**. Some check failures may be caused by changes in the base branch rather than by your PR. Please rebase your branch and push again to get up-to-date CI results.

**What to do next:**

There is no rush — take your time and work at your own pace. We appreciate your contribution and are happy to wait for updates. If you have questions, feel free to ask on the [Airflow Slack](https://s.apache.org/airflow-slack).

---

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -433,17 +434,35 @@ def set_xcom(
...
+                value=value,
+                run_id=run_id,
+                task_id=task_id,
+                dag_id=dag_id,
+                map_index=map_index,
+                serialize=False,
+                dag_result=dag_result,
+                session=session,
+            )
+    except IntegrityError:
```
kaxil: `except IntegrityError` is broader than the concurrent-write conflict you're targeting. The XCom insert also carries an FK on `(dag_id, task_id, run_id, map_index)` to `task_instance` (ON DELETE CASCADE), so if the TI/run is deleted between the `dag_run_id` lookup inside `XComModel.set` and the flush, you get an FK `IntegrityError` here too, not a unique-constraint one. The fallback UPDATE then matches zero rows (the cascade already removed them) and the request still returns 200, so the task believes its XCom was written when nothing was stored. Could you narrow this to the unique-constraint case (mirroring the per-dialect check in `_UniqueConstraintErrorHandler`) and let other `IntegrityError`s propagate?

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -433,17 +434,35 @@ def set_xcom(
...
-            run_id=run_id,
-            task_id=task_id,
-            dag_id=dag_id,
-            map_index=map_index,
-            serialize=False,
-            dag_result=dag_result,
-            session=session,
+        # Use a savepoint so that an IntegrityError on the XCom write does not
+        # roll back the task-map merge that may have already been flushed above.
+        with session.begin_nested():
```
kaxil, line 439: The new savepoint + UPDATE-fallback branch isn't covered by a test. A sequential retry won't reach it, since the DELETE+INSERT in `XComModel.set` never conflicts with itself, so the fallback needs a real conflict to trigger. You could pre-seed the row and patch `XComModel.set` to raise `IntegrityError` once, then assert the endpoint returns 200 and the stored value is the newest one. Route tests live in `airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_xcoms.py`.

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
nagasrisai: @kaxil  - Really good catch, thank you. You're right — a FK violation on the task_instance cascade would also produce an `IntegrityError` here, and swallowing it silently would return 200 when the XCom was never actually stored. I'll narrow the guard to unique-constraint violations only, mirroring `_UniqueConstraintErrorHandler`, and let everything else propagate. Will push the fix shortly.

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
nagasrisai: Agreed, the fallback path needs a direct test. I'll add one to `test_xcoms.py` that pre-seeds the row and patches `XComModel.set` to raise `IntegrityError` (unique constraint) once, then verifies the endpoint returns 200 and the stored value is the latest one. That way the UPDATE fallback is actually exercised rather than assumed.

nagasrisai:
@kaxil  - Thanks for the pointers! I've pushed two follow-up commits:

1. Narrowed the `except IntegrityError` guard to unique-constraint violations only (matching the per-dialect strings in `_UniqueConstraintErrorHandler`) so FK violations propagate correctly instead of being silently swallowed.
2. Added two tests to `TestXComsSetEndpoint` — one that exercises the UPDATE fallback path by patching `XComModel.set` to raise a unique-constraint error, and one that confirms FK errors propagate as a 500.

Will also look into getting the static checks clean.

@potiuk

nagasrisai:
@potiuk - Could you please mark this PR as ready for review? It's currently waiting for a reviewer. Thanks!

CC @kaxil

PR #69964, opened 2026-07-16
Don't deactivate DAG bundles owned by other dag-processors

### LINKED ISSUE #69963: Multiple dag-processors with partial bundle configs deactivate each other's bundles

Airflow Core

### Apache Airflow version

3.2.2 (also present on `main`/3.3.0; the affected code is unchanged)

**Summary**

When multiple `dag-processor` processes are each configured with a *partial* bundle configuration — i.e. each processor's `[dag_processor] dag_bundle_config_list` lists only the bundle(s) it owns — the processors repeatedly deactivate each other's bundles and the deployment never converges. This is a natural way to shard DAG parsing across processors (one bundle per processor), and it deadlocks.

**Root cause**

`DagBundlesManager.sync_bundles_to_db()` loads *all* bundles stored in the DB, pops the ones present in *this process's* config, and then unconditionally marks every remaining stored bundle inactive:

```python
for name, bundle in stored.items():
    bundle.active = False
    bundle.teams = []
    self.log.warning("DAG bundle %s is no longer found in config and has been disabled", name)
    session.execute(delete(ParseImportError).where(ParseImportError.bundle_name == name))
```

The method implicitly assumes the calling process sees the *complete* bundle configuration. When each processor only sees its own bundle, every processor treats all other processors' bundles as "no longer in config" and disables them. Processor A disables B's bundle; B then disables A's; on the next cycle they flip again. The log line

```
DAG bundle <name> is no longer found in config and has been disabled
```

is emitted continuously for bundles that are, in fact, actively configured on another processor.

**Steps to reproduce**

1. Configure two `dag-processor` processes.
2. Give processor A `dag_bundle_config_list = [bundle-a]` and processor B `dag_bundle_config_list = [bundle-b]` (disjoint, per-processor configs).
3. Let both run `sync_bundles_to_db()` (happens on each parse loop startup).
4. Observe the `dag_bundle` table: `bundle-a.active` and `bundle-b.active` flip between `True`/`False` on every cycle, and the "no longer found in config and has been disabled" warning is logged for a bundle that is actively configured elsewhere.

Minimal reproduction of the flip-flop (two managers with disjoint config, no error path involved):

```
AFTER B syncs:       [('bundle-a', False), ('bundle-b', True)]   # B disabled A
AFTER A syncs again: [('bundle-a', True),  ('bundle-b', False)]  # A disabled B
```

A dag-processor that only owns a subset of the bundle configuration must not deactivate bundles it does not know about. Deactivation of genuinely-removed bundles should only happen when the process sees the full configuration.

Proposed fix: add a `deactivate_missing` flag to `sync_bundles_to_db()` (default `True`, preserving current single-processor behavior), and have `DagFileProcessorManager.sync_bundles()` pass `deactivate_missing=False` when the processor was started with a bundle filter (`bundle_names_to_parse`, i.e. `--bundle-name`). A PR implementing this follows.

Related: #69698 addresses a different over-eager-deactivation path in the same method (transient errors during bundle template extraction). This issue is about partial per-processor configuration, not error handling.

### Operating System

Linux (Kubernetes; also reproduced locally on macOS)

N/A (airflow-core)

### Deployment

Other 3rd-party Helm chart

### Deployment details

Multiple `dag-processor` deployments, each with its own `AIRFLOW__DAG_PROCESSOR__DAG_BUNDLE_CONFIG_LIST` containing only the bundle that processor is responsible for, and started with `--bundle-name` to restrict parsing.

### Anything else?

The affected code path (`sync_bundles_to_db`) is identical on `main` (3.3.0/3.4.0).

### Code of Conduct

sortega:
### Why

`DagBundlesManager.sync_bundles_to_db()` marks every bundle stored in the DB that is not present in the calling process's config as inactive ("no longer found in config and has been disabled"). This implicitly assumes the calling process sees the *complete* bundle configuration.

When multiple `dag-processor` processes are each configured with a *partial* bundle config — one bundle per processor, a natural way to shard DAG parsing — each processor treats every other processor's bundle as removed and deactivates it. Processor A disables B's bundle, B then disables A's, and they flip `dag_bundle.active` on every parse cycle. The deployment never converges and continuously logs the disable warning for bundles that are actively configured on another processor.

Minimal reproduction of the flip-flop (two managers with disjoint config, no error path involved):

```
AFTER B syncs:       [('bundle-a', False), ('bundle-b', True)]   # B disabled A
AFTER A syncs again: [('bundle-a', True),  ('bundle-b', False)]  # A disabled B
```

### How

- Add a `deactivate_missing` parameter to `sync_bundles_to_db()`, defaulting to `True` so existing single-processor / full-config behavior (including cleanup of genuinely-removed bundles) is unchanged.
- When `deactivate_missing=False`, the method upserts the bundles it knows about and returns without touching the rest.
- `DagFileProcessorManager.sync_bundles()` now passes `deactivate_missing=not self.bundle_names_to_parse`: a processor started with `--bundle-name` (a partial owner) will not deactivate bundles it does not own, while an unfiltered processor keeps deactivating removed bundles as before.

### Tests

- `test_sync_bundles_to_db_partial_config_does_not_disable_other_bundles` — two processors with disjoint configs; asserts both bundles stay active across sync cycles (fails without the fix).
- `test_sync_bundles_deactivates_missing_when_owning_all_bundles` / `test_sync_bundles_does_not_deactivate_missing_when_filtered` — assert `sync_bundles()` forwards the right `deactivate_missing` value based on `bundle_names_to_parse`.
- Existing `test_sync_bundles_to_db` (full-config deactivation) still passes.

closes: #69963
related: #69698

---

- [X] Yes - Claude Code

### DIFF
--- a/airflow-core/src/airflow/dag_processing/bundles/manager.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/manager.py
@@ -271,7 +271,7 @@ class DagBundlesManager(LoggingMixin):
         self.log.info("DAG bundles loaded: %s", ", ".join(self._bundle_config.keys()))
 
     @provide_session
-    def sync_bundles_to_db(self, *, session: Session = NEW_SESSION) -> None:
+    def sync_bundles_to_db(self, *, deactivate_missing: bool = True, session: Session = NEW_SESSION) -> None:
         """
         Persist the configured DAG bundles into ``DagBundleModel`` rows.
 
@@ -280,6 +280,14 @@ class DagBundlesManager(LoggingMixin):
         ``DagModel`` / ``SerializedDagModel`` rows is the responsibility of
         ``DagBag`` plus ``sync_bag_to_db`` (or, in production, the DAG
         processor); calling this method does not trigger that work.
+
+        :param deactivate_missing: When ``True`` (the default), any bundle stored in
+            the database that is not present in this manager's config is marked
+            inactive. This is only correct when the calling process sees the
+            *complete* bundle configuration. When multiple dag-processors each run
+            with a partial config (e.g. one bundle per processor), pass ``False`` so a
+            processor does not disable bundles owned by other processors -- otherwise
+            the processors repeatedly deactivate each other and never converge.
         """
         self.log.debug("Syncing DAG bundles to the database")
 
@@ -355,6 +363,11 @@ class DagBundlesManager(LoggingMixin):
                     "Removing ownership of team '%s' from Dag bundle '%s'", bundle_to_team[name], name
                 )
                 bundle.teams = []
+
+        if not deactivate_missing:
+            # This process only owns a subset of the bundles, so the remaining stored
+            # bundles may well be owned by another dag-processor. Leave them alone.
+            return
 
         # Import here to avoid circular import
         from airflow.models.errors import ParseImportError
--- a/airflow-core/src/airflow/dag_processing/manager.py
+++ b/airflow-core/src/airflow/dag_processing/manager.py
@@ -333,7 +333,9 @@ class DagFileProcessorManager(LoggingMixin):
 
     def sync_bundles(self) -> None:
         """Sync configured DAG bundles to the metadata database."""
-        DagBundlesManager().sync_bundles_to_db()
+        # When this processor only parses a subset of bundles, it does not see the full
+        # bundle configuration and must not deactivate bundles owned by other processors.
+        DagBundlesManager().sync_bundles_to_db(deactivate_missing=not self.bundle_names_to_parse)
 
     def get_all_bundles(self) -> list[BaseDagBundle]:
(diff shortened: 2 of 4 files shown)


airflow-core/newsfragments/69964.bugfix.rst
hussein-awala, line 1: no need for a newsfragment, it is only needed when there is a significant change

hussein-awala (approved):
Looks good! I reproduced the issue locally and confirmed the fix resolves it. Please drop the newsfragment (we keep those for behavior changes, migrations, or must-highlight items, otherwise the commit message is enough), then LGTM.

airflow-core/newsfragments/69964.bugfix.rst
sortega: Rebased and removed the newsfragment comment

hussein-awala (approved):

eladkal (approved):
LGTM

PR #63296, opened 2026-03-10
fix: block path traversal via ".." in dag_id and run_id

YoannAbriel:
## Problem

`validate_key()` and `validate_run_id()` both allow `..` in their values. Since `dag_id` and `run_id` are used in log file paths (e.g. `dag_id=.../run_id=.../`), a crafted value containing `..` could theoretically traverse outside the intended log directory.

## Root Cause

`KEY_REGEX` (`^[\w.-]+$`) matches `..` as valid consecutive dots. `validate_run_id` uses a configurable `allowed_run_id_pattern` that could also permit `..`. Neither function checks for path traversal sequences.

## Fix

Added an explicit `..` check in both `validate_key()` (raises `AirflowException`) and `validate_run_id()` (raises `ValueError`) before any other validation. Added corresponding unit tests for both functions.

Closes: #63295

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code

Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

* Read the **[Pull Request Guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-guidelines)** for more information. Note: commit author/co-author name and email in commits become permanently public when merged.
* For fundamental code changes, an Airflow Improvement Proposal ([AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/Airflow+Improvement+Proposals)) is needed.

Issue #63295, opened 2026-03-10
DAG names should be filtered a little stronger


ferruzzi:
### Apache Airflow version

3.1.7

### What happened?

In one of the PRs I was reviewing, Copilot pointed out a low-risk security issue.  We allow `dag_id` and `run_id` to contain `..` and use those values in log file paths which cold theoretically lead to an issue.  `validate_key()` blocks slashes in `dag_id` and `validate_run_id()` blocks them in `run_id`, but both still allow the `..` and maybe they should block that.

### How to reproduce

Original PR and discussion here:  https://github.com/apache/airflow/pull/62616#discussion_r2899179199

### Operating System

templeos

### Deployment

Other

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

PR #69965, opened 2026-07-16
Warn on suspicious Dag and task IDs at Go SDK build time

Andrushika:
Warn on suspicious Dag and task IDs at Go SDK build time
---
### Why
Go SDK currently accepts any string as Dag/task ID. Python SDK already validates this with validate_key (max 250 chars, only letters, digits, -, ., _), and ts-sdk added the same check in #69400.

### What
Adds the same validation to the Go SDK at build time. It's the best-effort validation: if the dag or task ID is invalid, the build is still allowed, only the warning will show up when users compile the binary.
related: #69400, #69937

---

### DIFF
--- a/go-sdk/cmd/airflow-go-pack/pack.go
+++ b/go-sdk/cmd/airflow-go-pack/pack.go
@@ -171,6 +171,7 @@ func runPack(stdout, stderr io.Writer, opts *packOptions) error {
 			fmt.Fprintf(stderr, "warning: dag %q has no tasks\n", dagID)
 		}
 	}
+	warnOnSuspiciousIDs(stderr, meta)
 
 	manifest, err := renderManifest(meta, filepath.Base(sourcePath))
 	if err != nil {
--- a/go-sdk/cmd/airflow-go-pack/validate.go
+++ b/go-sdk/cmd/airflow-go-pack/validate.go
@@ -1,0 +1,74 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package main
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"sort"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/apache/airflow/go-sdk/internal/airflowmetadata"
+)
+
+const maxIDLength = 250
+
+var idRegex = regexp.MustCompile(`^[\p{L}\p{N}_.-]+$`)
+
+// warnOnSuspiciousIDs checks every dag and task id in the manifest against the
+// rules the Airflow server enforces (airflow.utils.helpers.validate_key). It is
+// best-effort and only warns: the server validates authoritatively, and checks
+// like the '..' one depend on server configuration the packer cannot see.
+func warnOnSuspiciousIDs(stderr io.Writer, meta airflowmetadata.Manifest) {
+	dagIDs := make([]string, 0, len(meta.Dags))
+	for id := range meta.Dags {
+		dagIDs = append(dagIDs, id)
+	}
+	sort.Strings(dagIDs)
+	for _, dagID := range dagIDs {
+		warnOnSuspiciousID(stderr, fmt.Sprintf("dag id %q", dagID), dagID)
+		for _, taskID := range meta.Dags[dagID].Tasks {
+			warnOnSuspiciousID(stderr, fmt.Sprintf("task id %q in dag %q", taskID, dagID), taskID)
+		}
+	}
+}
+
+func warnOnSuspiciousID(stderr io.Writer, label, id string) {
+	if length := utf8.RuneCountInString(id); length > maxIDLength {
+		fmt.Fprintf(
+			stderr,
+			"warning: %s is longer than %d characters (%d); the Airflow server will reject it\n",
+			label, maxIDLength, length,
+		)
+	}
+	if !idRegex.MatchString(id) {
+		fmt.Fprintf(
+			stderr,
+			"warning: %s must be made of alphanumeric characters, dashes, dots, and underscores; the Airflow server will reject it\n",
+			label,
+		)
+	} else if strings.Contains(id, "..") {
+		fmt.Fprintf(
+			stderr,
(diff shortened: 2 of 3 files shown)


go-sdk/bundle/bundlev1/registry_test.go
```diff
@@ -66,6 +68,49 @@ func (s *RegistrySuite) TestAddDag_DuplicatePanics() {
...
+			func() { New().AddDag(id) },
+		)
+	}
+}
+
+func (s *RegistrySuite) TestAddDag_TooLongPanics() {
+	s.PanicsWithError("Dag ID must be less than 250 characters, not 251", func() {
+		New().AddDag(strings.Repeat("a", 251))
+	})
+}
```
viiccwen: the current 250/251 boundary tests use only ASCII. Replacing this with `len(value)` would still pass every test while incorrectly rejecting, for example, a valid 250-character Chinese ID because it occupies 750 bytes.

Could we cover both 250 and 251 multibyte characters at the boundary? That would protect the reason for using rune count rather than byte length.

viiccwen:
PR description says "Adds the same validation to the Java SDK".
It's Go SDK, right?

go-sdk/bundle/bundlev1/registry_test.go
viiccwen: Like this:

```go
s.NotPanics(func() {
	New().AddDag(strings.Repeat("任", 250))
})
```

Andrushika:
> PR description says "Adds the same validation to the Java SDK". It's Go SDK, right?

You’re right, already fixed it. Thanks!

go-sdk/bundle/bundlev1/registry_test.go
Andrushika: Added. Thanks!

go-sdk/bundle/bundlev1/registry.go
```diff
@@ -20,13 +20,34 @@ package bundlev1
...
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/apache/airflow/go-sdk/pkg/worker"
 )
 
+const maxKeyLength = 250
+
+var keyRegex = regexp.MustCompile(`^[\p{L}\p{N}_.-]+$`)
```
jason810496: IIUC, the `a..b` case can pass the validation. Perhaps it's worthwhile to add the unit test to exercise. Thanks.

https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/airflow-core/src/airflow/serialization/definitions/dag.py#L637-L648

go-sdk/bundle/bundlev1/registry.go
jason810496, line 157: We should also validate the `AddTask` call here. Since we will pass the generic function name itself as the taskID name.

jason810496 (commented):
LGTM overall, thanks.

---

UPDATE:

As mentioned in https://github.com/apache/airflow/pull/69937#pullrequestreview-4728664940, we will need to wait for the coordinator side first.

potiuk:
@Andrushika — There are 2 unresolved review thread(s) on this PR from @jason810496. Could you either push a fix or reply in each thread explaining why the feedback doesn't apply? Once you believe the feedback is addressed, mark the thread as resolved so the reviewer isn't re-pinged needlessly. Thanks!

---

go-sdk/bundle/bundlev1/registry.go
Andrushika: Good catch. `AddTask` is already covered by the shared registration path: `registerTask` derives the function name and passes it to `registerTaskWithName`, which calls `validateKey` on the task ID. So both `AddTask` and `AddTaskWithName` validate the final task ID. I guess no additional change is needed here.

go-sdk/bundle/bundlev1/registry.go
Andrushika: Thanks. I may be wrong, but the linked code checks run_id, while this PR checks Dag and task IDs. Should the validation of run_id and Dags, task IDs align? If so, I will open a follow-up PR to fix this in task SDK.

 Since the PR is on hold, should I wait for the coordinator's design before making this change?

jason810496 (commented):
Hi @Andrushika,
After revisiting this, I'd like to make this best-effort build time validation instead of task runtime validation or dag processing side validation for both Go and Java SDK (You can raise PR for TS SDK as well if you like! or I can clean them up as well).

The reason is that the following snippet will validate the task runtime in server side and be served as source of truth for validation. 
https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/airflow-core/src/airflow/serialization/definitions/dag.py#L637-L648

For the `a..b` case, we can just raise the warning (since we can't trust the env / config at client side).
Please let me know if there's other direction that makes more sense to you.

Thanks.

Andrushika:
Got it, thanks for revisiting! I will move the validation into build time as you described. Would be happy to also clean up ts-sdk after current work is done.

Andrushika:
@jason810496 A small question to clarify: since we are going to validate the IDs in build time, I am planning to implement it in `runPack()` (the function behind the `go tool airflow-go-pack`).

Since you mentioned we should keep the validation best-effort, I want to confirm how strict the pack step should be. The `..` case depends on [core] `allow_double_dot_in_ids` and can be customed by users on the server side, so the lang-sdk can only warn. But the length and charset rules are rejected by the server unconditionally, so failing early seems possible for them.

I have two options and would like to ask your opinion:

- Reject the pack when the `id` is invalid, except the `..` case, which only gives a warning
- Still allow the pack even if some `id` is invalid, give warnings in all the invalid cases

Which one do you prefer? Thanks!
I would prefer to keep all of them as warnings and still allow building, because it can prevent drifting that TP mentioned.

jason810496 (commented):
> I would prefer to keep all of them as warnings and still allow building, because it can prevent drifting that TP mentioned.

Sure, let's go with returning warning for all the cases. Thanks! I don't have strong opinion on keeping it strict regarding the validation since it's best-effort anyway.

Andrushika:
All tests passed, ready for review!

> The reason is that the following snippet will validate the task runtime in server side and be served as source of truth for validation.
> 
> https://github.com/apache/airflow/blob/ffde96203e76c0c27ccf1ae6eeaad270653248bf/airflow-core/src/airflow/serialization/definitions/dag.py#L637-L648
> 
> For the `a..b` case, we can just raise the warning (since we can't trust the env / config at client side). Please let me know if there's other direction that makes more sense to you.

A little fact I found: On the current `main`, only `run_id` checks the `a..b` case. Core and task-sdk do not validate `dag_id` and `task_id` in the case of double dots.

I guess we should also handle this in the core? If it is, I would be happy to open a follow-up PR!

jason810496 (commented):
Nice! LGTM overall. The only nit is that we need to soften the "fail early at registration," statement in the PR description. Thanks.

> A little fact I found: On the current main, only run_id checks the a..b case. Core and task-sdk do not validate dag_id and task_id in the case of double dots.

IMO, it's fine to ignore the dag_id as the dag_id will eventually produce the run_id, so the above snippet will handle it.

For the task_id case (based on the following context)

- https://github.com/apache/airflow/pull/63296
- https://github.com/apache/airflow/issues/63295

cc @ferruzzi Do you think it's necessary to validate the task_id as well? Or you would like to verify just the dag_id and the run_id only?

jason810496 (approved):
Thanks, LGTM after updating the PR description! The follow-up item doesn't block this one being merged.

PR #69966, opened 2026-07-16
Fix `Trigger.submit_event` to route asset events through the configured `asset_manager`

### LINKED ISSUE #69962: Trigger.submit_event bypasses the configured [core] asset_manager_class (uses base AssetManager class instead of the asset_manager instance)

Airflow Core

### Apache Airflow version

3.3.0

# What happened?

`airflow/models/trigger.py` imports the base class and calls `register_asset_change` on it:

At `airflow/models/trigger.py:33`

```python
from airflow.assets.manager import AssetManager
...
```

At `airflow/models/trigger.py:294` (inside `Trigger.submit_event`)

```python
for asset in trigger.assets:
    AssetManager.register_asset_change(
        asset=asset.to_serialized(),
        extra={"from_trigger": True, "payload": event.payload},
        session=session,
    )
```

Because register_asset_change is a `@classmethod` and it is invoked on the hardcoded base `AssetManager` class, a custom manager configured via `[core] asset_manager_class` is never used for asset events produced by triggers (i.e. `AssetWatcher`-backed assets / deferrable-produced asset events).

Every other call site resolves the configured singleton instance instead, so those paths honor `asset_manager_class` correctly:

- At `airflow/models/taskinstance.py:79` 
  ```python
  from airflow.assets.manager import asset_manager  # (task outlet events)
  ```
- At `airflow/api_fastapi/core_api/routes/public/assets.py:74` 
  ```python
  from airflow.assets.manager import asset_manager  # (REST API events)
  ```

The singleton is built from config in `airflow/assets/manager.py`:

At `airflow/assets/manager.py:821`
```python
def resolve_asset_manager() -> AssetManager:
    _asset_manager_class = conf.getimport("core", "asset_manager_class",
                                          fallback="airflow.assets.manager.AssetManager")
    ...
    return _asset_manager_class(**_asset_manager_kwargs)

asset_manager = resolve_asset_manager()  # :838
```

So the trigger path is the one place that ignores `asset_manager_class`, making custom asset-event handling impossible for watcher/trigger-produced events. I don't know whether this is intentional, but for me it looks like a real inconsistency/bug.

# How to reproduce it?

Set a custom manager in config:

```
[core]
asset_manager_class = my_pkg.my_manager.MyAssetManager
```

At `my_pkg.my_manager.py`
```python
from airflow.assets.manager import AssetManager

class MyAssetManager(AssetManager):
    @classmethod
    def register_asset_change(cls, **kwargs):
        print("MyAssetManager invoked")
        return super().register_asset_change(**kwargs)
```

Minimal check of the dispatch used by each path (no scheduler needed):

```python
import airflow.models.trigger as t
from airflow.assets.manager import AssetManager, asset_manager

# What submit_event references:
print(t.AssetManager is AssetManager)   # True  -> base class, not the configured instance

# Path used by task outlets / REST API (honors config):
type(asset_manager).__name__               # -> MyAssetManager

# Path used by submit_event (ignores config):
# t.AssetManager.register_asset_change(...) dispatches to base AssetManager, not MyAssetManager
```

Result: `asset_manager` (the configured instance) is `MyAssetManager`, but `submit_event` calls the base `AssetManager` classmethod, so `MyAssetManager.register_asset_change` is never triggered for watcher/trigger-produced asset events.

`Trigger.submit_event` should route through the configured manager instance, exactly like the task-outlet and REST-API paths, so that a custom `[core] asset_manager_class` applies uniformly to all asset events regardless of how they are produced.

### Operating System

Fedora Linux 44 (Workstation Edition)

### Deployment

Docker-Compose

_No response_

_No response_

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

limeschnaps:
closes: #69962

## What

`Trigger.submit_event` imported the base `AssetManager` class and called
`AssetManager.register_asset_change(...)` on it, instead of using the resolved
`asset_manager` singleton like every other call site
(`taskinstance.py`, the public assets REST route). Because
`register_asset_change` is a classmethod invoked on the base class, a custom
manager configured via `[core] asset_manager_class` was silently bypassed for
asset events produced by triggers / asset watchers.

This swaps the import and call to the configured `asset_manager` instance, so
`asset_manager_class` now applies uniformly regardless of how an asset event is
produced.

## Why

`asset_manager_class` is the only supported extension point for asset-event
handling, and it did not apply to trigger/watcher-produced events — the one
place a custom manager could not be reached.

## Tests

- Added `test_submit_event_uses_configured_asset_manager`: asserts
  `submit_event` dispatches through the module-level `asset_manager` instance.
  Fails on `main`, passes with the fix.
- Updated `test_submit_event_no_n_plus_one_for_assets` to patch
  `airflow.assets.manager.AssetManager.register_asset_change` (where the
  classmethod is defined) so it no longer depends on the removed module symbol.

### DIFF
--- a/airflow-core/src/airflow/models/trigger.py
+++ b/airflow-core/src/airflow/models/trigger.py
@@ -30,7 +30,7 @@ from sqlalchemy.orm import Mapped, Session, mapped_column, relationship, selecti
 from sqlalchemy.sql.functions import coalesce
 
 from airflow._shared.timezones import timezone
-from airflow.assets.manager import AssetManager
+from airflow.assets.manager import asset_manager
 from airflow.configuration import conf
 from airflow.models.asset import AssetWatcherModel
 from airflow.models.base import Base
@@ -295,7 +295,7 @@ class Trigger(Base):
             # Already deleted for some reason
             return
         for asset in trigger.assets:
-            AssetManager.register_asset_change(
+            asset_manager.register_asset_change(
                 asset=asset.to_serialized(),
                 extra={"from_trigger": True, "payload": event.payload},
                 session=session,
--- a/airflow-core/tests/unit/models/test_trigger.py
+++ b/airflow-core/tests/unit/models/test_trigger.py
@@ -235,8 +235,39 @@ def test_submit_event(mock_callback_handle_event, session, create_task_instance)
     mock_callback_handle_event.assert_called_once_with(event, session)
 
 
+def test_submit_event_uses_configured_asset_manager(session):
+    """
+    Regression test: ``Trigger.submit_event`` must notify assets through the
+    configured ``[core] asset_manager_class`` instance (the module-level
+    ``asset_manager`` singleton), not the hardcoded base ``AssetManager`` class.
+
+    Before the fix, ``submit_event`` called ``AssetManager.register_asset_change``
+    on the base class, so a custom asset manager set via ``asset_manager_class``
+    was silently bypassed for asset events produced by triggers / asset watchers.
+    Patching the module-level ``asset_manager`` reference stands in for a
+    configured custom manager, since that singleton is exactly what
+    ``resolve_asset_manager()`` builds from config.
+    """
+    trigger = Trigger(classpath="airflow.triggers.testing.SuccessTrigger", kwargs={})
+    session.add(trigger)
+    session.flush()
+
+    asset = AssetModel(name="configured_manager_asset")
+    asset.add_trigger(trigger, "test_asset_watcher")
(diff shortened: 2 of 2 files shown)


steveahnahn (approved):
Nice catch, I confirmed `asset_manager` is the configured singleton, `AssetManager` is the base class, so `submit_event` was bypassing a custom manager for trigger/asset-watcher events. New test passes with the fix so LGTM.

PR #68521, opened 2026-06-13
TriggerRunner: Add batch trigger creation duration metric

SameerMesiah97:
**Description**

This change adds a `batch_trigger_creation_duration` timing metric to measure the total time spent creating pending triggers during a single invocation of `TriggerRunner.create_triggers()`.

The metric is emitted only when trigger creation work is performed and includes the triggerer's `team_name` tag when available.

**Rationale**

Trigger creation occurs on the triggerer event loop. During periods of high trigger creation volume, processing large batches of pending triggers may occupy the event loop for extended periods, but this time is not currently observable.

This metric provides visibility into the cost of trigger creation and helps identify situations where large creation batches contribute to triggerer latency.

**Tests**

Added unit tests verifying that the `triggerer.batch_trigger_creation_duration` metric is emitted with the expected tags when pending triggers are successfully created.

**Backwards Compatibility**

This change is additive only and does not modify trigger execution behavior, lifecycle semantics, or public APIs.

PR #67927, opened 2026-06-03
TriggerRunnerSupervisor: Add trigger queue delay metric

SameerMesiah97:
**Description**

This change adds a `trigger_queue_delay` timing metric to measure the time between a trigger workload being queued by the `TriggerRunnerSupervisor` and being processed by the `TriggerRunner`.

A `queued_at` timestamp is recorded when `RunTrigger` workloads are enqueued, and the resulting delay is emitted as a timing metric during trigger creation. The metric includes the triggerer's `team_name` tag when available.

**Rationale**

The triggerer currently provides limited visibility into trigger startup latency. During periods of high trigger creation volume, workloads may spend time waiting before they are processed, but this delay is not currently observable.

This metric helps identify triggerer backlog and provides insight into trigger startup performance under load.

**Tests**

Added unit tests verifying that the `trigger_queue_delay` metric is emitted with the expected value and tags when a queued workload is processed.

**Backwards Compatibility**

This change is additive only and does not modify trigger execution behavior, lifecycle semantics, or public APIs.

PR #69971, opened 2026-07-16
Prevent negative trigger queue delay metrics

### LINKED ISSUE #69970: Trigger queue delay metric can become negative after system clock adjustments

Airflow Core

### Apache Airflow version

`main`

The `triggerer.trigger_queue_delay` metric measures the duration between the TriggerRunnerSupervisor queueing a trigger workload and the TriggerRunner creating the trigger. Both timestamps currently use `time.time()`.

If NTP, virtualization, or an administrator adjusts the system clock between those operations, the metric can become negative or be inflated even though the actual queue delay is normal.

Minimal reproduction:

1. Create a trigger workload with `queued_at=100.0`.
2. Simulate the wall clock moving backward so that `time.time()` returns `90.0` when the TriggerRunner creates the trigger.
3. Observe that the emitted queue delay is `-10000` milliseconds.

This can also be reproduced in the existing `test_create_triggers_emits_queue_delay_metric` test by setting the workload's queue timestamp to `100.0`, mocking `time.time()` to return `90.0`, and expecting the real elapsed delay from a stable clock.

Queue delay is a duration and should use `time.monotonic()` at both the producer and consumer ends. System wall-clock  djustments should not affect the emitted metric.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

The metric was introduced by #67927. A tested fix is ready that changes both timestamp sites to `time.monotonic()` and adds regression coverage for a backward wall-clock adjustment.

### Code of Conduct

viiccwen:
closes #69970

Trigger queue delay is a duration measured from when the supervisor queues a workload until the runner creates its trigger. Using wall-clock timestamps for that duration can produce negative or inflated metrics when the system clock is adjusted.

This changes both ends of the measurement to use the monotonic clock. The regression tests simulate the wall clock moving backward and cover both the supervisor timestamp and the runner's emitted metric.

---

- [x] Yes - GPT-5.6 Sol

---

### DIFF
--- a/airflow-core/src/airflow/jobs/triggerer_job_runner.py
+++ b/airflow-core/src/airflow/jobs/triggerer_job_runner.py
@@ -1401,6 +1401,8 @@ class TriggerRunner:
                     inlets=[Asset(name=name, uri=uri) for name, uri in workload.watched_assets.items()]
                 )
             if workload.queued_at is not None:
+                # `queued_at` is captured by the supervisor process and consumed by the runner process.
+                # Both run on the same host, so their monotonic timestamps are comparable.
                 stats.timing(
                     "triggerer.trigger_queue_delay",
                     int((time.monotonic() - workload.queued_at) * 1000),
--- a/airflow-core/src/airflow/jobs/triggerer_job_runner.py
+++ b/airflow-core/src/airflow/jobs/triggerer_job_runner.py
@@ -979,7 +979,7 @@ class TriggerRunnerSupervisor(WatchedSubprocess):
         if new_trigger_ids:
             workloads_to_create = self.build_trigger_workloads(new_trigger_ids)
 
-            queued_at = time.time()
+            queued_at = time.monotonic()
 
             for workload in workloads_to_create:
                 workload.queued_at = queued_at
@@ -1403,7 +1403,7 @@ class TriggerRunner:
             if workload.queued_at is not None:
                 stats.timing(
                     "triggerer.trigger_queue_delay",
-                    int((time.time() - workload.queued_at) * 1000),
+                    int((time.monotonic() - workload.queued_at) * 1000),
                     tags=prune_dict({"team_name": self.team_name}),
                 )
 
--- a/airflow-core/tests/unit/jobs/test_triggerer_job.py
+++ b/airflow-core/tests/unit/jobs/test_triggerer_job.py
@@ -1582,9 +1582,9 @@ class TestTriggerRunner:
         runner.team_name = team_name
         runner.to_create.append(workload)
 
-        with patch(
-            "airflow.jobs.triggerer_job_runner.time.time",
-            return_value=101.5,
+        with (
+            patch("airflow.jobs.triggerer_job_runner.time.monotonic", return_value=101.5),
+            patch("airflow.jobs.triggerer_job_runner.time.time", return_value=90.0),
         ):
             await runner.create_triggers()
 
@@ -2291,6 +2291,8 @@ def test_update_triggers_delegates_workload_creation(supervisor_builder, mocker)
     supervisor = supervisor_builder()
     supervisor.running_triggers = {1, 3}
     workload = workloads.RunTrigger(id=2, classpath="some.trigger", encrypted_kwargs="", ti=None)
+    monotonic = mocker.patch("airflow.jobs.triggerer_job_runner.time.monotonic", return_value=100.0)
+    mocker.patch("airflow.jobs.triggerer_job_runner.time.time", return_value=90.0)
     build_trigger_workloads = mocker.patch.object(
         TriggerRunnerSupervisor, "build_trigger_workloads", autospec=True, return_value=[workload]
     )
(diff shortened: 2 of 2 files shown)


shahar1 (changes requested):
This re-applies the `time.monotonic()` change that was removed during review of [#67927](https://github.com/apache/airflow/pull/67927/files#diff-e4cc497f1c786d142ce4c930f43e33b0bb4b53d375d274278fa82f4d5567608a) (queued_at is stamped in the supervisor but subtracted in the runner subprocess).
Please either implement a more robust solution for this issue, or pivot this PR to adding a comment for why we use `time.time()` here.

viiccwen:
Hello @shahar1, 
thanks for raising the concern. I took another look at the guarantees provided by Python and the execution model here.

Although the reference point of `time.monotonic()` is unspecified, current [Python documentation](https://docs.python.org/3/library/time.html#time.monotonic) explicitly states that the monotonic clock is the same for all processes since Python 3.5. Airflow currently supports Python 3.10+, and the supervisor / runner are local processes on the same host, so both calls use the same system-wide monotonic clock.

On Linux, this is backed by `CLOCK_MONOTONIC`; the value has an unspecified epoch, but values obtained by separate processes on the same host are still comparable. The unspecified reference point only means that the value cannot be interpreted as an absolute timestamp.

For this metric, we are measuring an **elapsed duration**, so `time.monotonic()` also avoids negative or inflated values caused by wall-clock adjustments such as NTP corrections or manual clock changes. ([ref](https://docs.python.org/3/library/time.html#time.time))


I see here:
https://github.com/apache/airflow/blob/fe9d8650efc45e11e158f67fb6d3518f877877de/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L272-L278

then here:
https://github.com/apache/airflow/blob/fe9d8650efc45e11e158f67fb6d3518f877877de/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L547-L554

so the relation is like: Triggerer host/container -> Triggerer main process -> TriggerRunnerSupervisor -> local TriggerRunner subprocess

IMO, we can add a comment documenting the same-host and cross-process clock assumption, not just use `max(0, time.time() - queued_at)` to hide negative.

shahar1:
> Hello @shahar1, thanks for raising the concern. I took another look at the guarantees provided by Python and the execution model here.
> 
> Although the reference point of `time.monotonic()` is unspecified, current [Python documentation](https://docs.python.org/3/library/time.html#time.monotonic) explicitly states that the monotonic clock is the same for all processes since Python 3.5. Airflow currently supports Python 3.10+, and the supervisor / runner are local processes on the same host, so both calls use the same system-wide monotonic clock.
> 
> On Linux, this is backed by `CLOCK_MONOTONIC`; the value has an unspecified epoch, but values obtained by separate processes on the same host are still comparable. The unspecified reference point only means that the value cannot be interpreted as an absolute timestamp.
> 
> For this metric, we are measuring an **elapsed duration**, so `time.monotonic()` also avoids negative or inflated values caused by wall-clock adjustments such as NTP corrections or manual clock changes. ([ref](https://docs.python.org/3/library/time.html#time.time))
> 
> I see here:
> 
> https://github.com/apache/airflow/blob/fe9d8650efc45e11e158f67fb6d3518f877877de/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L272-L278
> 
> then here:
> 
> https://github.com/apache/airflow/blob/fe9d8650efc45e11e158f67fb6d3518f877877de/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L547-L554
> 
> so the relation is like: Triggerer host/container -> Triggerer main process -> TriggerRunnerSupervisor -> local TriggerRunner subprocess
> 
> IMO, we can add a comment documenting the same-host and cross-process clock assumption, not just use `max(0, time.time() - queued_at)` to hide negative.

Thanks for challenging my claim - it seems like you've got a point :)
Worth adding a comment as suggested that it relies on the premise of same host.
I wonder if at a later point (if we want to allow executions on remote) we should measure it in the supervisor instead.

@SameerMesiah97 - FYI, I'll be happy for your feedback.

shahar1 (approved):

airflow-core/src/airflow/jobs/triggerer_job_runner.py
```diff
@@ -1403,7 +1403,7 @@ async def create_triggers(self):
             if workload.queued_at is not None:
                 stats.timing(
                     "triggerer.trigger_queue_delay",
```
SameerMesiah97, line 1407: I would implement your own feedback here:

"IMO, we can add a comment documenting the same-host and cross-process clock assumption, not just use max(0, time.time() - queued_at) to hide negative."

SameerMesiah97 (approved):
I agree. Just add that comment you were talking about and then this can be merged. 

I will revert `time.time()` to `time.monotonic()` in #68521 as well. 

Thank you for catching this.

viiccwen:
cc @shahar1. 
Thanks for reviewing, it's a good discussion!

PR #69976, opened 2026-07-16
Don't fail API server startup when a plugin uses an unknown team

pierrejeambrun:
Follow-up of #69502.

#69502 added team validation for plugins, but a single plugin referencing a team that does not exist raised during app creation — so the whole API server, and every other plugin, failed to boot. A misconfigured plugin should be surfaced, not fatal, the same way other plugin load errors are.

The team check now records each offending plugin as a **plugin import error** (surfaced in the plugins API and UI, like any other plugin load failure) and logs a warning instead of raising. It also runs from `get_fastapi_plugins()` — which `init_plugins()` already calls — so validation happens automatically wherever plugins are mounted, rather than relying on a separate `validate_plugin_teams()` call that a caller could forget.

Only reached from the API server and the `airflow plugins` CLI (never workers/dag-processor/triggerer), and the DB is only touched when `[core] multi_team` is enabled.

---
### Before
Screenshot 2026-07-16 at 18 00 20

### After
Screenshot 2026-07-16 at 17 55 20
Screenshot 2026-07-16 at 17 55 36

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/app.py
+++ b/airflow-core/src/airflow/api_fastapi/app.py
@@ -132,11 +132,8 @@ def create_app(apps: str = "all") -> FastAPI:
         app.mount("/execution", task_exec_api_app)
 
     if "all" in apps_list or "core" in apps_list:
-        from airflow import plugins_manager
-
         app.state.dag_bag = dag_bag
         init_plugins(app)
-        plugins_manager.validate_plugin_teams()
         init_auth_manager(app)
         init_flask_plugins(app)
         init_views(app)  # Core views need to be the last routes added - it has a catch all route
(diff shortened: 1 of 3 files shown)


pierrejeambrun:
cc: @vincbeck @o-nikolas

vincbeck (approved):

o-nikolas:
Hey, thanks for the PR @pierrejeambrun we definitely talked about this and @vincbeck will be happy you decided to change this :smile: 

Many other components in Airflow fail to start if things are misconfigured (related to teams and otherwise). So the thinking was the user will have to restart the API server anyway and we should not start in a bad state. I hear the reasoning that it's just a plugin and it's fine to start in that failed state. The plan was to see how users feel about it and pivot if needed :slightly_smiling_face:

PR #69977, opened 2026-07-16
Use atomic upsert for asset run queue on MySQL to avoid deadlocks

### LINKED ISSUE #69938: Asset-triggered DAG queueing can deadlock on MySQL and is not retried

dstandish:
When an asset event fans out to multiple consumer DAGs, the non-PostgreSQL branch of
`AssetManager._queue_dagruns` inserts one row per consumer into `asset_dag_run_queue`
inside a per-row `SAVEPOINT`, catching only `IntegrityError`. On MySQL/InnoDB this is
deadlock-prone under concurrent fan-out (errno 1213), and because a deadlock discards the
savepoint it also surfaces a secondary `SAVEPOINT ... does not exist` error. The deadlock
is not retried.

This routes MySQL to a single-statement `INSERT ... ON DUPLICATE KEY UPDATE` (a no-op
update on conflict), mirroring the existing PostgreSQL `ON CONFLICT DO NOTHING` bulk path.
The enqueue becomes atomic, drops the per-row savepoint churn, and holds locks far more
briefly — greatly reducing the InnoDB deadlock window. PostgreSQL and SQLite behavior is
unchanged.

Note: a single-statement upsert can still deadlock on InnoDB gap/next-key locks under heavy
concurrency, so adding deadlock retry on this path is a natural follow-up.

closes: #69938

---

### DIFF
--- a/airflow-core/tests/unit/assets/test_manager.py
+++ b/airflow-core/tests/unit/assets/test_manager.py
@@ -211,6 +211,48 @@ class TestAssetManager:
         )
         assert session.scalar(select(func.count()).select_from(AssetDagRunQueue)) == 0
 
+    @pytest.mark.parametrize(
+        ("dialect_name", "expected_helper"),
+        [
+            ("postgresql", "_queue_dagruns_nonpartitioned_postgres"),
+            ("mysql", "_queue_dagruns_nonpartitioned_mysql"),
+            ("sqlite", "_queue_dagruns_nonpartitioned_slow_path"),
+        ],
+    )
+    def test_queue_dagruns_routes_by_dialect(self, dialect_name, expected_helper):
+        """Test that _queue_dagruns routes to the dialect-appropriate queue helper."""
+        dag = DagModel(dag_id="dag1")
+        session = mock.MagicMock(spec=Session)
+        with (
+            mock.patch("airflow.assets.manager.get_dialect_name", return_value=dialect_name),
+            mock.patch.object(AssetManager, "_queue_partitioned_dags"),
+            mock.patch.object(AssetManager, expected_helper) as mock_helper,
+        ):
+            AssetManager._queue_dagruns(
+                asset_id=1,
+                dags_to_queue={dag},
+                partition_key=None,
+                partition_date=None,
+                event=mock.MagicMock(),
+                task_instance=None,
+                session=session,
+            )
+        mock_helper.assert_called_once_with(1, {dag}, session)
+
+    def test_queue_dagruns_nonpartitioned_mysql_builds_upsert(self):
+        """Test that the MySQL queue path emits an INSERT ... ON DUPLICATE KEY UPDATE."""
+        from sqlalchemy.dialects import mysql
+
+        dag = DagModel(dag_id="dag1")
+        session = mock.MagicMock(spec=Session)
+
+        AssetManager._queue_dagruns_nonpartitioned_mysql(asset_id=1, dags_to_queue={dag}, session=session)
+
+        stmt, values = session.execute.call_args.args
+        compiled = str(stmt.compile(dialect=mysql.dialect())).upper()
+        assert "ON DUPLICATE KEY UPDATE" in compiled
+        assert values == [{"target_dag_id": "dag1"}]
+
     def test_register_asset_change_notifies_asset_listener(
         self, session, mock_task_instance, testing_dag_bundle, listener_manager
     ):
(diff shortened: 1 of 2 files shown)


airflow-core/tests/unit/assets/test_manager.py
```diff
@@ -211,6 +211,48 @@ def test_register_asset_change_no_downstreams(self, session, mock_task_instance)
...
+                partition_date=None,
+                event=mock.MagicMock(),
+                task_instance=None,
+                session=session,
+            )
+        mock_helper.assert_called_once_with(1, {dag}, session)
+
+    def test_queue_dagruns_nonpartitioned_mysql_builds_upsert(self):
+        """Test that the MySQL queue path emits an INSERT ... ON DUPLICATE KEY UPDATE."""
+        from sqlalchemy.dialects import mysql
```
Lee-W: probably don't need to make it an inline import?

Lee-W (approved):

airflow-core/tests/unit/assets/test_manager.py
dstandish: Good call — moved it to the module-level imports in ce5040d35b.

---

PR #69985, opened 2026-07-16
Fix deadline callback data encoding in migration 0094 (scheduler CrashLoopBackOff)

### LINKED ISSUE #69980: Deadline callback migration 0094 shallow-encodes nested callback kwargs → scheduler CrashLoopBackOff (KeyError('__var'))
### Apache Airflow version

3.2.0+ (root cause is migration `0094`, `airflow_version = "3.2.0"`; buggy code is still present on `main`)

### What happened

On a deployment using **Deadline Alerts** with an `AsyncCallback` whose `kwargs` contain a **nested dict**, the **scheduler enters CrashLoopBackOff**. Every scheduler loop crashes while deserializing an associated `callback` row in the deadline-processing query:

```
File ".../airflow/jobs/scheduler_job_runner.py", line 1808, in _run_scheduler_loop
    for deadline in session.scalars( ... selectinload(Deadline.callback) ... )
File ".../airflow/utils/sqlalchemy.py", line 221, in process_result_value
    return BaseSerialization.deserialize(value)
File ".../airflow/serialization/serialized_objects.py", line 632, in deserialize
    return {k: cls.deserialize(v) for k, v in var.items()}
File ".../airflow/serialization/serialized_objects.py", line 629, in deserialize
    var = encoded_var[Encoding.VAR]
KeyError: <Encoding.VAR: '__var'>
```

The crash happens before the scheduler heartbeats, so it presents as failing liveness probes / restarts with no OOM and no other logged exception — easy to misdiagnose as a probe or DB problem.

### Root cause

Migration **`0094_3_2_0_replace_deadline_inline_callback_with_fkey.py`** (revision `e812941398f4`) moves the old inline deadline callback into the `callback` table. `callback.data` is an `ExtendedJSON` column, whose read path runs `BaseSerialization.deserialize` (`utils/sqlalchemy.py`), which requires **every nested dict to be wrapped** as `{"__type": "dict", "__var": {...}}`.

But the migration hand-builds `callback.data` and only wraps the **top level**, embedding the old callback's `kwargs` **raw**:

```sql
-- _upgrade_postgresql(): raw nested kwargs pulled from the old inline (SDK-serde) callback
COALESCE(NULLIF(d.callback::jsonb->'__data__'->'kwargs', 'null'::jsonb), '{}'::jsonb) AS cb_kwargs
...
INSERT INTO callback (... data ...)
SELECT ...
  json_build_object(
      '__var', json_build_object(
          'path', b.cb_path,
          'kwargs', b.cb_kwargs,      -- <-- raw nested dict, NOT extended-serialized
          'prefix', :prefix,
          'dag_id', b.dag_id),
      '__type', 'dict')::json
```

(`_upgrade_mysql_sqlite()` builds the same shape in Python.) The insert is done via Core (`callback_table.insert()` / `sa.text`), which also bypasses `ExtendedJSON.process_bind_param`, so `BaseSerialization.serialize` never runs.

Result: `callback.data` = `{"__type":"dict","__var":{"path":..., "kwargs": {<bare nested object>}, ...}}`. On read, `deserialize` recurses into the outer `DAT.DICT`, reaches the bare `kwargs` (or a nested dict inside it such as metric `tags`), executes `var = encoded_var[Encoding.VAR]`, and throws `KeyError('__var')`.

**Trigger:** callback `kwargs` containing a nested dict. Callbacks with only flat/primitive kwargs survive because the single top-level wrap is enough.

The migration should produce the same encoding `BaseSerialization.serialize` would — i.e. **recursively** wrap nested dicts (`kwargs`, and any dict within it) as `{"__type":"dict","__var":{...}}` — so the resulting `callback.data` round-trips through `BaseSerialization.deserialize`.

Because `0094` has already shipped (3.2.0) and run on live deployments, a **forward repair migration** is also needed to re-encode already-corrupted `callback.data` rows (deployments that ran `0094` with any nested-dict callback kwargs currently have a latent scheduler crash that fires when such a deadline becomes due).

### How to reproduce

Minimal, self-contained demonstration of the serialize/deserialize invariant (no Airflow import needed): a correct `BaseSerialization.serialize({"path": "...", "kwargs": {"tags": {"a": "b"}}})` wraps `kwargs`/`tags` and round-trips, whereas the migration's shallow-wrapped shape `{"__type":"dict","__var":{"kwargs":{"tags":{...}}}}` raises `KeyError('__var')` in `deserialize`.

End-to-end: on 3.1.x create a deadline whose `AsyncCallback` kwargs contain a nested dict; upgrade across `0094`; then run the scheduler (or `SELECT` the `callback.data`) — it is stored with unwrapped nested kwargs and the scheduler crashes.

### Operating System

Linux (containerized)

N/A (core deadline alerts + serialization)

### Deployment

Other

### Anything else?

Workaround to unblock a crashing scheduler (neutralize the corrupt rows so the `WHERE ~missed` deadline query skips them):

```sql
UPDATE deadline d SET missed = true
FROM callback c
WHERE c.id = d.callback_id
  AND d.missed = false
  AND jsonb_path_exists(c.data, '$.**.__var.* ? (@.type() == "object" && !exists(@.__type))');
```

Yes.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8, `claude-opus-4-8[1m]`).

seanmuth:
Closes #69980.

## What's the problem

Migration `0094_3_2_0_replace_deadline_inline_callback_with_fkey` moves inline deadline callbacks into the `callback` table, whose `data` column is `ExtendedJSON` — read back at runtime via `BaseSerialization.deserialize`, which requires every nested dict to be wrapped as `{"__type": "dict", "__var": {...}}`.

Both upgrade paths built `callback.data` without recursively extended-serializing the callback `kwargs`:

- **Postgres** (`_upgrade_postgresql`): `json_build_object` wrapped only the top level, embedding the old callback's `kwargs` jsonb raw.
- **MySQL/SQLite** (`_upgrade_mysql_sqlite`): wrote the callback dict via a plain `JSON` column with no extended wrapping at all.

So any callback whose `kwargs` contain a nested dict (e.g. metric `tags`) is stored unreadable. When the scheduler loads it in the deadline-processing loop, `BaseSerialization.deserialize` recurses into the outer `dict` node, hits the bare nested object, runs `encoded_var[Encoding.VAR]`, and raises `KeyError(<Encoding.VAR: '__var'>)` — crashing the scheduler on every loop (CrashLoopBackOff), before it heartbeats, with no OOM and no other logged exception.

## What's in this PR

- Both **upgrade** paths now extended-serialize the callback data, matching `BaseSerialization.serialize`:
  - Postgres: a recursive, session-local `pg_temp.encode_extended(jsonb)` function used inside the existing batch CTE (keeps the one-round-trip design).
  - MySQL/SQLite: a `_serialize_extended()` helper applied to the callback dict.
- Both **downgrade** paths decode `kwargs` back to raw (`pg_temp.decode_extended` / `_deserialize_extended`), so an upgrade→downgrade round-trip is preserved. Both decoders are lenient, so they also correctly downgrade data written by the pre-fix version of this migration.
- **Tests** (`test_0094_deadline_callback_migration.py`): existing assertions now verify `callback.data` round-trips through the real `BaseSerialization`; added a nested-`kwargs`/`tags` regression test and an encode/decode helper round-trip test.

## Already-affected deployments

This hotfix protects deployments that have **not yet** run `0094`. Deployments already upgraded across `0094` with nested-dict callback kwargs have latent corrupt `callback.data` rows (a scheduler crash that fires when such a deadline becomes due); those rows must be repaired out of band — e.g. re-encode the affected rows:

```sql
-- find affected rows
SELECT count(*) FROM callback c
WHERE jsonb_path_exists(c.data::jsonb, '$.**.__var.* ? (@.type()=="object" && !exists(@.__type))');
```

---
🤖 Opened with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8, `claude-opus-4-8[1m]`).


steveahnahn (approved):
Verified both paths, `_serialize_extended` matches canonical `BaseSerialization.serialize` and round-trips through the real deserializer across nested/empty/list/primitive shapes, and on a live Postgres the `pg_temp.encode_extended`/`decode_extended` functions produce the same result and invert cleanly.

nit: the plpgsql path has no test. `test_0094_deadline_callback_migration.py` builds its engine with `_make_engine()` (in-memory SQLite), so even the Postgres CI job never executes `encode_extended`/`decode_extended`

airflow-core/src/airflow/migrations/versions/0094_3_2_0_replace_deadline_inline_callback_with_fkey.py
```diff
@@ -54,12 +54,102 @@
...
+    CREATE OR REPLACE FUNCTION pg_temp.encode_extended(node jsonb) RETURNS jsonb AS $$
+    DECLARE k text; v jsonb; out jsonb := '{}'::jsonb;
+    BEGIN
+      IF jsonb_typeof(node) = 'object' THEN
+        FOR k, v IN SELECT * FROM jsonb_each(node) LOOP
+          out := out || jsonb_build_object(k, pg_temp.encode_extended(v));
+        END LOOP;
+        RETURN jsonb_build_object('__type', 'dict', '__var', out);
+      ELSIF jsonb_typeof(node) = 'array' THEN
+        RETURN (SELECT jsonb_agg(pg_temp.encode_extended(e)) FROM jsonb_array_elements(node) e);
```
aaron-y-chen, line 87: I wonder if `node = []`, whether the entire return value would be `None`. I guess that's not what you want, right? Should we handle this case here?

airflow-core/src/airflow/migrations/versions/0094_3_2_0_replace_deadline_inline_callback_with_fkey.py
```diff
@@ -54,12 +54,102 @@
...
+          END IF;
+          RETURN node;
+        ELSE
+          FOR k, v IN SELECT * FROM jsonb_each(node) LOOP
+            out := out || jsonb_build_object(k, pg_temp.decode_extended(v));
+          END LOOP;
+          RETURN out;
+        END IF;
+      ELSIF jsonb_typeof(node) = 'array' THEN
+        RETURN (SELECT jsonb_agg(pg_temp.decode_extended(e)) FROM jsonb_array_elements(node) e);
```
aaron-y-chen, line 136: Same as above.

aaron-y-chen (commented):
Thanks for the PR!

airflow-core/newsfragments/69985.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix scheduler crash from deadline callbacks whose kwargs contain a nested dict. Migration ``0094`` (``replace_deadline_inline_callback_with_fkey``) built ``callback.data`` with nested kwargs left unencoded, so ``BaseSerialization.deserialize`` raised ``KeyError`` on read and the scheduler entered CrashLoopBackOff. Both the Postgres and MySQL/SQLite upgrade paths now extended-serialize nested kwargs (and the downgrade path decodes them back). Deployments already upgraded across ``0094`` must repair existing ``callback.data`` rows out of band. (#69980)
```
kaxil, line 1: Scoping this to nested dicts undersells it. `BaseSerialization.deserialize` fetches `encoded_var[Encoding.VAR]` on any non-primitive, non-list value, so a flat `kwargs={"k": "v"}` and even `kwargs={}` raise `KeyError('__var')` exactly like the nested case, and on MySQL/SQLite the pre-fix path wrote no top-level envelope at all so the outer dict raises immediately. Every row 0094 migrated is unreadable, on both backends, not just the ones with nested tags.

Your own diff shows it: `test_null_callback_does_not_crash` covers a row whose `kwargs` is `{}`, and swapping its `json.loads` for `BaseSerialization.deserialize` only passes because of this fix. Worth rewording to something like "every deadline callback row migrated by ``0094``", otherwise an operator whose tags are flat strings reads this release note and concludes they are in the clear.

airflow-core/src/airflow/migrations/versions/0094_3_2_0_replace_deadline_inline_callback_with_fkey.py
```diff
@@ -54,12 +54,102 @@
 _ASYNC_CALLBACK_CLASSNAME = "airflow.sdk.definitions.deadline.AsyncCallback"
 
 
+def _serialize_extended(value):
```
kaxil, line 57: 0094 is in the released `3.2.0` tag and main is already 3.4.0, so patching it in place only helps deployments upgrading straight from <3.2. Everyone who went through 3.2.x or 3.3.x has already run the broken version, has 100% of their callback rows unreadable, and will never re-run 0094 to pick this up.

Is a repair revision planned as a follow-up? A new migration could reuse these two helpers to re-encode existing rows, and it would need to handle both corrupt shapes: Postgres rows have the outer `__var` envelope with raw `kwargs` underneath, MySQL/SQLite rows have no envelope at all. That second shape is why "repair out of band" is not actionable as written -- the `jsonb_path_exists` query in the description matches on `$.**.__var.*`, which finds nothing in a MySQL/SQLite row and so reports it clean.

ferruzzi:
@seanmuth  are you still working on this one?

PR #69989, opened 2026-07-17
Scope plugin listeners to teams in multi-team airflow

o-nikolas:
When multi-team mode is enabled, a team-scoped plugin's listeners now only receive events for that team's Dags, while global plugins (no team_name) continue to receive everything. Component-lifecycle, asset, and import-error hooks remain global because they are not associated with a single team.

On the worker the team is taken from the server-provided task instance context rather than local configuration, so scoping does not depend on whether core.multi_team happens to be set on the worker which is not entirely reliable.

 

---

---

### DIFF
--- a/airflow-core/src/airflow/listeners/listener.py
+++ b/airflow-core/src/airflow/listeners/listener.py
@@ -18,15 +18,33 @@
 from __future__ import annotations
 
 from functools import cache
+from typing import TYPE_CHECKING
 
 from airflow._shared.listeners.listener import ListenerManager
 from airflow._shared.listeners.spec import lifecycle, taskinstance
+from airflow.configuration import conf
 from airflow.listeners.spec import asset, dagrun, importerrors
 from airflow.plugins_manager import integrate_listener_plugins
 
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
 
 @cache
-def get_listener_manager() -> ListenerManager:
+def _build_listener_manager(team_name: str | None) -> ListenerManager:
+    _listener_manager = ListenerManager()
+
+    _listener_manager.add_hookspecs(lifecycle)
+    _listener_manager.add_hookspecs(dagrun)
+    _listener_manager.add_hookspecs(taskinstance)
+    _listener_manager.add_hookspecs(asset)
+    _listener_manager.add_hookspecs(importerrors)
+
+    integrate_listener_plugins(_listener_manager, team_name=team_name)
+    return _listener_manager
+
+
+def get_listener_manager(team_name: str | None = None) -> ListenerManager:
     """
     Get a listener manager for Airflow core.
 
@@ -36,17 +54,37 @@ def get_listener_manager() -> ListenerManager:
     - taskinstance: on_task_instance_running, on_task_instance_success, etc.
     - asset: on_asset_created, on_asset_changed, etc.
     - importerrors: on_new_dag_import_error, on_existing_dag_import_error
+
+    :param team_name: In multi-team mode, the team whose (and the global) plugin
+        listeners this manager should contain. ``None`` yields a manager with only
+        global-plugin listeners (used by team-agnostic events such as lifecycle,
+        asset, and import-error hooks). When multi-team mode is disabled this
+        argument is ignored and the manager holds every plugin's listeners.
+
+    Calls are cached per ``team_name``. The single positional delegation to
+    ``_build_listener_manager`` guarantees that ``get_listener_manager()``,
+    ``get_listener_manager(None)`` and ``get_listener_manager(team_name=None)`` all
+    return the identical (global) manager instance rather than distinct cache
+    entries.
     """
-    _listener_manager = ListenerManager()
-
-    _listener_manager.add_hookspecs(lifecycle)
-    _listener_manager.add_hookspecs(dagrun)
-    _listener_manager.add_hookspecs(taskinstance)
-    _listener_manager.add_hookspecs(asset)
-    _listener_manager.add_hookspecs(importerrors)
-
-    integrate_listener_plugins(_listener_manager)
-    return _listener_manager
+    return _build_listener_manager(team_name)
 
 
-__all__ = ["get_listener_manager", "ListenerManager"]
+def get_listener_manager_for_dag(dag_id: str, session: Session | None = None) -> ListenerManager:
+    """
+    Get the listener manager scoped to the team that owns ``dag_id``.
+
+    When multi-team mode is disabled this returns the single manager holding all
+    listeners. Otherwise it resolves the Dag's owning team and returns a manager
+    containing the global listeners plus that team's listeners.
+    """
+    if not conf.getboolean("core", "multi_team"):
+        return get_listener_manager()
+
+    from airflow.models.dag import DagModel
+
+    team_name = DagModel.get_team_name(dag_id, session=session) if session else DagModel.get_team_name(dag_id)
+    return get_listener_manager(team_name)
+
+
+__all__ = ["get_listener_manager", "get_listener_manager_for_dag", "ListenerManager"]
--- a/devel-common/src/tests_common/pytest_plugin.py
+++ b/devel-common/src/tests_common/pytest_plugin.py
@@ -3107,7 +3107,7 @@ def listener_manager():
         get_sdk_lm = None
 
     core_lm = get_core_lm()
-    sdk_lm = get_sdk_lm() if get_sdk_lm else None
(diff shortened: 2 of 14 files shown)


potiuk:
rebase needed?

vincbeck (approved):

task-sdk/src/airflow/sdk/plugins_manager.py
```diff
@@ -128,9 +128,17 @@ def integrate_macros_plugins() -> None:
...
+    """
+    Add listeners from plugins to the given listener manager.
+
+    Only listeners from global plugins (``team_name is None``) and plugins belonging
+    to ``team_name`` are registered. On the worker, ``team_name`` comes from the
+    server-provided task instance context; filtering here is always applied and does
+    not depend on the worker reading ``core.multi_team`` from its own configuration.
+    """
     plugins, _ = _get_plugins()
+    plugins = [plugin for plugin in plugins if plugin.team_name in (None, team_name)]
```
amoghrajesh, line 141: The core version of this filters by `conf.getboolean("core", "multi_team")`, why the drift?

devel-common/src/tests_common/pytest_plugin.py
```diff
@@ -3107,7 +3107,7 @@ def test_something(listener_manager):
         get_sdk_lm = None
 
     core_lm = get_core_lm()
-    sdk_lm = get_sdk_lm() if get_sdk_lm else None
+    sdk_lm = get_sdk_lm() if get_sdk_lm is not None else None
```
amoghrajesh, line 3110: Seems like a no-op change?

task-sdk/src/airflow/sdk/plugins_manager.py
amoghrajesh, line 141: This filters `plugin.team_name` in `(None, team_name)` unconditionally, but the core equivalent only applies this filter when `core.multi_team` is enabled otherwise it registers every plugin regardless of team_name.

That difference creates an asymmetry when multi team mode is disabled on a deployment that still has team_name tagged plugins:

- Core hooks correctly go back to firing for every plugin, since core's filtering is gated on the flag.
- Worker-triggered hooks stay silently scoped, because `ti.team_name` resolves to `None` when the server has multi-team disabled, and this function's filter then excludes any plugin with a non-None team_name with no flag-based escape hatch.
Net result: disabling multi-team mode does not fully restore "every plugin sees every event" - a team-scoped plugin quietly stops getting its most important events (the worker-side ones) with no error or log line to explain why.

airflow-core/src/airflow/listeners/listener.py
```diff
@@ -36,17 +54,37 @@ def get_listener_manager() -> ListenerManager:
...
 
-    _listener_manager.add_hookspecs(lifecycle)
-    _listener_manager.add_hookspecs(dagrun)
-    _listener_manager.add_hookspecs(taskinstance)
-    _listener_manager.add_hookspecs(asset)
-    _listener_manager.add_hookspecs(importerrors)
 
-    integrate_listener_plugins(_listener_manager)
-    return _listener_manager
+def get_listener_manager_for_dag(dag_id: str, session: Session | None = None) -> ListenerManager:
```
ashb, line 73: This doesn't feel right.

The dag processor should already be configured as one per team, so I don't think the `_for_dag` and the dag_is param is not right either, as the team name things are running for is already defined?

I'm very worried about the expectation or precedence this sets

task-sdk/src/airflow/sdk/plugins_manager.py
o-nikolas: Yeah, I was trying to avoid having to depend on the `core.multi_team` setting in workers, since users are very unreliable when setting these types of configs in workers. But you are right about this gap, I see three ways out:
1. We just check config and this will sort of silently not work unless users set the config properly (we have no way to tell if they have left it False intentionally or not).
2. We update the team_name context field to be not present when core.multi_team is off on the server side. So we have not present == multi-team disabled, None == on but this is a global task and a string value == this is a specific team. This does overload the field a bit and we have to ensure everything operates well with back compat (when that context field didn't exist at all) but I think it should work.
3. We add a new context field that indicates directly whether multi-team is enabled on the server side. This would be a nice clear signal, but I could see some folks argue that this is doing the job that the config option could be doing and that it should just not work if users don't set it properly.

What are your thoughts on those options @amoghrajesh?

devel-common/src/tests_common/pytest_plugin.py
o-nikolas: This is actually a mypy fix. We now cache the listener managers at a lower layer and so mypy is running on this function when it wasn't before.

airflow-core/src/airflow/listeners/listener.py
o-nikolas: Hey Ash, thanks for the review.

None of these `get_listener_manager_for_dag` are happening in the dag processor or the sdk side at all, the scheduler and API server are calling this method. Those are shared team spaces.

The scheduler does have a cached mapping of dag id to teams, so it is possible for a few of the call sites to plumb that through (though you pay the cost of more plumbing, so pick your poison). The API server does not have a centralized dag id to team name cache, but the call volume of those APIs (and thus their listeners) should be very small, I don't think we need to worry about that at the moment. So dag id can still be passed along there.

What are your thoughts @ashb?

ashb (changes requested):
In context of https://github.com/apache/airflow/pull/69989/files#r3616746261

I see your point, and on the one hand that makes sense, but on the other hand I worry about listener code from multiple teams and what it means for the security of the shared API server component.

Listeners are already a special case and advanced "be careful you might break it feature" so I'm half tempted here to say that since listeners are a power user feature that we don't need, nor want, a per-team listener.

My (re)reading of https://cwiki.apache.org/confluence/spaces/AIRFLOW/pages/294816378/AIP-67+Multi-team+deployment+of+Airflow+components implies to me that the Listeners are the responsibility of the Deployment manager, and so should be shared across all components, i.e. not per team.

(I am doing this change as a "request changes" so it is more obvious. This is not my veto and a permenant -1, but a big item and needs more eyes on.)

o-nikolas:
> In context of https://github.com/apache/airflow/pull/69989/files#r3616746261
> 
> I see your point, and on the one hand that makes sense, but on the other hand I worry about listener code from multiple teams and what it means for the security of the shared API server component.

Hey @ashb thanks for another review! Are you able to provide some more details on the worries you have here? I don't have much to go off of to really respond without knowing more about what exactly it is that you're worried about.

> 
> Listeners are already a special case and advanced "be careful you might break it feature" so I'm half tempted here to say that since listeners are a power user feature that we don't need, nor want, a per-team listener.

I would rather think that exactly because it is a power user feature, that they would want that level of control and security with listener events. But we have no user data to go off of here, it's all speculation.

> 
> My (re)reading of https://cwiki.apache.org/confluence/spaces/AIRFLOW/pages/294816378/AIP-67+Multi-team+deployment+of+Airflow+components implies to me that the Listeners are the responsibility of the Deployment manager, and so should be shared across all components, i.e. not per team.

All that is really trying to say is that for any plugins that need to be installed on shared components (Scheduler, API Server, etc) the Admin/Deployment Manager is responsible for that (i.e. no single team should have access to the shared space). Meaning that, as part of that management, they may enable or configure team based items (e.g. the team executor configuration lives inside the scheduler host). 

> 
> (I am doing this change as a "request changes" so it is more obvious. This is not my veto and a permenant -1, but a big item and needs more eyes on.)

Sure no worries, I wouldn't merge a PR without going through the comments and I've been eagerly awaiting your re-review. So no need for the request of change for visibility next time :)

---

tl;dr: I'm not sure what your exact concerns/worries are so hard to comment on that. But if anyone else knows or has similar concerns, I'm okay with not making Listeners team based for now. But it is still not clear to me why. Maybe other multi-team folks can weigh in, @vincbeck and @potiuk?

vincbeck:
I dont know listeners enough to weigh in whether we should teamify listeners but if there are valid concerns on that, we could leave listeners global to the environment? We can always park that work for now and if later these concerns are resolved, we could then resume working on it?

o-nikolas:
> I dont know listeners enough to weigh in whether we should teamify listeners but if there are valid concerns on that, we could leave listeners global to the environment? We can always park that work for now and if later these concerns are resolved, we could then resume working on it?

I still haven't really heard what the concerns are (I did ask @ash for some concrete examples in my last message) but I'm fine to park this for now. I think it's a gap in the implementation that power users will definitely come knocking for though.

I'll move on to UI team filtering for plugin react apps and external views from plugins

Issue #52280, opened 2025-06-26
Redesign Databricks Workflow Repair Functionality for Airflow 3 Compatibility


pankajkoti:
### Description

Following the enhancements in [PR #52228](https://github.com/apache/airflow/pull/52228), which aligns the Databricks workflow provider plugin to support Airflow 3 and introduces XCom-based storage for extra links, this issue tracks the remaining work to reintroduce and redesign the repair functionality for compatibility with Airflow 3.

Background:
Airflow 3 blocks constructs like @provide_session and direct database access, moving toward a secure and API-driven model. The current repair flow is incompatible with this model for the following reasons:

1. Extra links (e.g., “See Databricks Job Run”) are now stored in XCom at task execution time. However, Databricks’ repair API requires knowing which specific tasks to repair at the time of invocation, which is only possible after failures occur — creating a timing mismatch.
2. The current repair implementation relies on a Flask AppBuilder (FAB) view exposed via the `/repair_databricks_job endpoint`, which does not align with the future direction of Airflow extensibility using FastAPI based endpoint.

Goals:
1. Re-evaluate the approach to repair in Airflow 3, given the new architecture for link computation and the runtime ambiguity around which tasks will need repair.
2. Explore how the repair functionality can be implemented as a FastAPI-based endpoint, aligning with Airflow’s evolving API architecture.
3. Investigate authentication and authorization patterns for securing a FastAPI repair endpoint.
4. Consider deferring the repair redesign until after the plugin extensibility improvements planned for Airflow 3.1 are available, to ensure better alignment with long-term architecture. This may offer a cleaner and more sustainable path for integrating custom endpoints and plugin-level APIs.

I’d be glad if someone from the community is interested in picking this up or would like to collaborate with me on it — especially if they’re familiar with the topic and Airflow 3.

### Related issues

related: #40587

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

eladkal:
cc @moomindani  @mwojtyczka Databricks help on this one would be much appriciated

moomindani:
Thanks for the ping @eladkal — looking at this from the Databricks side.

A couple of updates worth folding in before settling on direction:

**1. The "wait for Airflow 3.1" rationale in the issue is largely obsolete**

[...]

PR #69998, opened 2026-07-17
Enable repair of failed Databricks Workflow tasks from Airflow 3 UI

PrakshiGoyal10:
On Airflow 3 the "Repair a single task" and "Repair All Failed Tasks" links on DatabricksWorkflowTaskGroup tasks were gated off, so on-call users had to leave Airflow and repair failed runs in the Databricks UI.

This restores repair-from-Airflow on Airflow 3 (3.1+), served by a FastAPI sub-application on the API server since Flask-AppBuilder was dropped:

- A `GET` renders a read-only confirmation page; only a same-site `POST` performs the repair and clears tasks. Because the API server's auth cookie is `SameSite=Lax`, it is not sent on a cross-site `POST`, so moving the mutation off `GET` is what protects the action from cross-site request forgery — matching how Airflow 3's own UI mutations are protected.
- The repair link carries only Airflow identifiers (the run's launch `task_id`, and for a single-task repair the target `task_id`). The Databricks connection and run id are derived server-side from the launch task's trusted `WorkflowRunMetadata` XCom, never from the request, so a crafted link cannot point the repair at an arbitrary connection or Databricks run.
- The set of failed tasks is resolved from the live Databricks run state (hook helper `get_run_failed_task_keys`), and `repair_run` is called with `rerun_dependent_tasks` so downstream tasks resume too.
- The repaired task instances and their downstream instances are cleared — but only after a successful repair call — mapping Databricks task keys to Airflow `task_id`s via the same `md5(dag_id__task_id)` scheme the operators use, resolved from the serialized Dag.
- Databricks failures surface as a generic `502` without reflecting the upstream exception text, and confirmation-page identifiers are HTML-escaped.
- The repair backend and its links are gated to Airflow 3.1+, where the cookie-or-bearer authentication resolver it relies on is available; the redirect target is a same-site relative path, resolving the CodeQL open-redirect findings.

Verified end-to-end against a real Databricks workflow on serverless and job clusters: a failed task plus its `upstream_failed` downstream both return to success after confirming the repair. Unit tests cover the GET-does-not-mutate behaviour, confirmation-page escaping, server-side identifier resolution on the POST, and the generic 502.

related: #52280

---

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @PrakshiGoyal10** · by `@potiuk` · 2026-07-28 16:20 UTC
>
> - :x: **Pre-commit / static checks**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).
>
>
>
> **Note:** your branch is **294 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>

### DIFF
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -563,6 +563,28 @@ class DatabricksHook(BaseDatabricksHook):
 
         return all_tasks
 
+    def get_run_failed_task_keys(self, run_id: int) -> list[str]:
+        """
+        Return the ``task_key`` of every sub-task of a run that is in a terminal failure state.
+
+        Resolved from the live Databricks run rather than from Airflow's metadata DB, so it
+        reflects the actual per-task state Databricks ``repair_run`` will act on. The returned
+        keys are the values to pass as ``rerun_tasks`` to :meth:`repair_run`.
+
+        :param run_id: id of the run
+        :return: a list of Databricks ``task_key`` values for failed sub-tasks
+        """
+        failed_result_states = {"FAILED", "TIMEDOUT", "CANCELED", "MAXIMUM_CONCURRENT_RUNS_REACHED"}
+        failed_task_keys = []
+        for task in self.get_run_tasks(run_id):
+            state = task.get("state", {})
+            if (
+                state.get("result_state") in failed_result_states
+                or state.get("life_cycle_state") == "INTERNAL_ERROR"
+            ):
+                failed_task_keys.append(task["task_key"])
+        return failed_task_keys
+
     def get_run(self, run_id: int) -> dict[str, Any]:
         """
         Retrieve run information.
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -575,14 +575,24 @@ class DatabricksHook(BaseDatabricksHook):
         :return: a list of Databricks ``task_key`` values for failed sub-tasks
         """
         failed_result_states = {"FAILED", "TIMEDOUT", "CANCELED", "MAXIMUM_CONCURRENT_RUNS_REACHED"}
+
+        # ``get_run_tasks`` returns one entry per attempt and is not ordered by attempt, so a
+        # retried or already-repaired task appears several times under the same ``task_key``. Keep
+        # only the latest attempt per key (sort by ``start_time``, same idiom as
+        # ``DatabricksTaskBaseOperator._get_current_databricks_task``) before judging its state, so
+        # the result never contains duplicate keys — Databricks rejects duplicates in
+        # ``rerun_tasks`` — and a task whose latest attempt succeeded is not reported as failed.
+        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task["start_time"])
+        latest_by_key = {task["task_key"]: task for task in sorted_tasks}
+
         failed_task_keys = []
-        for task in self.get_run_tasks(run_id):
+        for task_key, task in latest_by_key.items():
             state = task.get("state", {})
             if (
                 state.get("result_state") in failed_result_states
                 or state.get("life_cycle_state") == "INTERNAL_ERROR"
             ):
-                failed_task_keys.append(task["task_key"])
+                failed_task_keys.append(task_key)
         return failed_task_keys
 
     def get_run(self, run_id: int) -> dict[str, Any]:
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -582,7 +582,8 @@ class DatabricksHook(BaseDatabricksHook):
         # ``DatabricksTaskBaseOperator._get_current_databricks_task``) before judging its state, so
         # the result never contains duplicate keys — Databricks rejects duplicates in
         # ``rerun_tasks`` — and a task whose latest attempt succeeded is not reported as failed.
-        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task["start_time"])
+        # Never-started sub-tasks omit ``start_time`` or send null; treat those as 0.
+        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task.get("start_time") or 0)
         latest_by_key = {task["task_key"]: task for task in sorted_tasks}
 
         failed_task_keys = []
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -608,6 +608,28 @@ class DatabricksHook(BaseDatabricksHook):
         """
         return self.get_run(run_id).get("tasks", [])
 
+    def get_run_failed_task_keys(self, run_id: int) -> list[str]:
+        """
+        Return the ``task_key`` of every sub-task of a run that is in a terminal failure state.
+
+        Resolved from the live Databricks run rather than from Airflow's metadata DB, so it
+        reflects the actual per-task state Databricks ``repair_run`` will act on. The returned
+        keys are the values to pass as ``rerun_tasks`` to :meth:`repair_run`.
+
+        :param run_id: id of the run
+        :return: a list of Databricks ``task_key`` values for failed sub-tasks
+        """
+        failed_result_states = {"FAILED", "TIMEDOUT", "CANCELED", "MAXIMUM_CONCURRENT_RUNS_REACHED"}
+        failed_task_keys = []
+        for task in self.get_run_tasks(run_id):
+            state = task.get("state", {})
+            if (
+                state.get("result_state") in failed_result_states
+                or state.get("life_cycle_state") == "INTERNAL_ERROR"
+            ):
+                failed_task_keys.append(task["task_key"])
+        return failed_task_keys
+
     def get_run(self, run_id: int) -> dict[str, Any]:
         """
         Retrieve run information.
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -627,7 +627,8 @@ class DatabricksHook(BaseDatabricksHook):
         # ``DatabricksTaskBaseOperator._get_current_databricks_task``) before judging its state, so
         # the result never contains duplicate keys — Databricks rejects duplicates in
         # ``rerun_tasks`` — and a task whose latest attempt succeeded is not reported as failed.
-        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task["start_time"])
+        # Never-started sub-tasks omit ``start_time`` or send null; treat those as 0.
+        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task.get("start_time") or 0)
         latest_by_key = {task["task_key"]: task for task in sorted_tasks}
 
         failed_task_keys = []
--- a/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
@@ -620,14 +620,24 @@ class DatabricksHook(BaseDatabricksHook):
         :return: a list of Databricks ``task_key`` values for failed sub-tasks
         """
         failed_result_states = {"FAILED", "TIMEDOUT", "CANCELED", "MAXIMUM_CONCURRENT_RUNS_REACHED"}
+
+        # ``get_run_tasks`` returns one entry per attempt and is not ordered by attempt, so a
+        # retried or already-repaired task appears several times under the same ``task_key``. Keep
+        # only the latest attempt per key (sort by ``start_time``, same idiom as
+        # ``DatabricksTaskBaseOperator._get_current_databricks_task``) before judging its state, so
+        # the result never contains duplicate keys — Databricks rejects duplicates in
+        # ``rerun_tasks`` — and a task whose latest attempt succeeded is not reported as failed.
+        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task["start_time"])
+        latest_by_key = {task["task_key"]: task for task in sorted_tasks}
+
         failed_task_keys = []
-        for task in self.get_run_tasks(run_id):
+        for task_key, task in latest_by_key.items():
             state = task.get("state", {})
             if (
                 state.get("result_state") in failed_result_states
                 or state.get("life_cycle_state") == "INTERNAL_ERROR"
             ):
-                failed_task_keys.append(task["task_key"])
+                failed_task_keys.append(task_key)
         return failed_task_keys
 
     def get_run(self, run_id: int) -> dict[str, Any]:
--- a/providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
+++ b/providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
@@ -120,16 +120,10 @@ class _CreateDatabricksWorkflowOperator(BaseOperator):
         "spark_submit_params",
     )
     caller = "_CreateDatabricksWorkflowOperator"
-    # Conditionally set operator_extra_links based on Airflow version
-    if AIRFLOW_V_3_0_PLUS:
-        # In Airflow 3, disable "Repair All Failed Tasks" since we can't pre-determine failed tasks
-        operator_extra_links = (WorkflowJobRunLink(),)
-    else:
-        # In Airflow 2.x, keep both links
-        operator_extra_links = (  # type: ignore[assignment]
-            WorkflowJobRunLink(),
-            WorkflowJobRepairAllFailedLink(),
-        )
+    operator_extra_links = (
+        WorkflowJobRunLink(),
+        WorkflowJobRepairAllFailedLink(),
+    )
 
     def __init__(
         self,
--- a/providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
+++ b/providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 
 import json
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from functools import cached_property
 from typing import TYPE_CHECKING, Any
 
@@ -49,11 +49,15 @@ class WorkflowRunMetadata:
     :param run_id: The ID of the Databricks workflow run.
     :param job_id: The ID of the Databricks workflow job.
     :param conn_id: The connection ID used to connect to Databricks.
+    :param task_key_map: Airflow ``task_id`` → Databricks ``task_key`` for the workflow's tasks.
+        Optional and defaulted for backward compatibility with runs launched before it was added
+        (those XComs carry only conn_id/job_id/run_id).
     """
 
     conn_id: str
     job_id: int
     run_id: int
+    task_key_map: dict[str, str] = field(default_factory=dict)
 
 
 def _flatten_node(
@@ -273,6 +277,14 @@ class _CreateDatabricksWorkflowOperator(BaseOperator):
             "conn_id": self.databricks_conn_id,
             "job_id": job_id,
             "run_id": run_id,
+            # Map each workflow task's Airflow task_id to the task_key registered with Databricks,
+            # captured here from the live operators. The repair endpoint needs it because an
+            # explicit ``databricks_task_key`` does not survive Dag serialization, so the API server
+            # cannot otherwise recover the true key from the serialized Dag.
+            "task_key_map": {
+                task.task_id: task.databricks_task_key  # type: ignore[attr-defined]
+                for task in self.tasks_to_convert.values()
+            },
         }
 
     def on_kill(self) -> None:
--- a/providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
+++ b/providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING, Any
 from urllib.parse import unquote
 
@@ -44,6 +45,8 @@ if TYPE_CHECKING:
     from airflow.providers.databricks.operators.databricks import DatabricksTaskBaseOperator
     from airflow.sdk.types import Logger
 
+log = logging.getLogger(__name__)
+
 
 def get_databricks_task_ids(
     group_id: str, task_map: dict[str, DatabricksTaskBaseOperator], log: Logger
@@ -65,6 +68,50 @@ def get_databricks_task_ids(
         log.debug("databricks task id for task %s is %s", task_id, databricks_task_id)
         task_ids.append(databricks_task_id)
     return task_ids
+
+
+def _repair_task(
+    databricks_conn_id: str,
+    databricks_run_id: int,
+    tasks_to_repair: list[str],
+    logger: Logger | logging.Logger,
+) -> int:
+    """
+    Repair a Databricks task using the Databricks API.
+
+    This function allows the Airflow repair buttons to create a repair job for Databricks.
+    It uses the Databricks API to get the latest repair ID before sending the repair query.
+
+    :param databricks_conn_id: The Databricks connection ID.
+    :param databricks_run_id: The Databricks run ID.
+    :param tasks_to_repair: A list of Databricks task IDs to repair.
+    :param logger: The logger to use for logging.
+    :return: the repair id returned by the Databricks API.
+    """
+    hook = DatabricksHook(databricks_conn_id=databricks_conn_id)
+
+    repair_history_id = hook.get_latest_repair_id(databricks_run_id)
+    logger.debug("Latest repair ID is %s", repair_history_id)
+    logger.debug(
+        "Sending repair query for tasks %s on run %s",
+        tasks_to_repair,
+        databricks_run_id,
+    )
+
+    run_data = hook.get_run(databricks_run_id)
+    repair_json = {
+        "run_id": databricks_run_id,
+        "latest_repair_id": repair_history_id,
+        "rerun_tasks": tasks_to_repair,
+        # Also rerun dependents so upstream-failed downstream tasks resume rather than
+        # staying skipped after the repaired task succeeds.
+        "rerun_dependent_tasks": True,
+    }
+
+    if "overriding_parameters" in run_data:
+        repair_json["overriding_parameters"] = run_data["overriding_parameters"]
+
+    return hook.repair_run(repair_json)
 
 
 # TODO: Need to re-think on how to support the currently unavailable repair functionality in Airflow 3. Probably a
@@ -194,46 +241,6 @@ if not AIRFLOW_V_3_0_PLUS:
         if not ti:
             raise TaskInstanceNotFound("Task instance not found")
         return ti
-
-    def _repair_task(
-        databricks_conn_id: str,
-        databricks_run_id: int,
-        tasks_to_repair: list[str],
-        logger: Logger,
-    ) -> int:
-        """
-        Repair a Databricks task using the Databricks API.
-
-        This function allows the Airflow retry function to create a repair job for Databricks.
-        It uses the Databricks API to get the latest repair ID before sending the repair query.
-
-        :param databricks_conn_id: The Databricks connection ID.
-        :param databricks_run_id: The Databricks run ID.
-        :param tasks_to_repair: A list of Databricks task IDs to repair.
-        :param logger: The logger to use for logging.
-        :return: None
-        """
-        hook = DatabricksHook(databricks_conn_id=databricks_conn_id)
-
-        repair_history_id = hook.get_latest_repair_id(databricks_run_id)
-        logger.debug("Latest repair ID is %s", repair_history_id)
-        logger.debug(
-            "Sending repair query for tasks %s on run %s",
-            tasks_to_repair,
-            databricks_run_id,
-        )
-
-        run_data = hook.get_run(databricks_run_id)
-        repair_json = {
-            "run_id": databricks_run_id,
-            "latest_repair_id": repair_history_id,
-            "rerun_tasks": tasks_to_repair,
-        }
-
-        if "overriding_parameters" in run_data:
-            repair_json["overriding_parameters"] = run_data["overriding_parameters"]
-
-        return hook.repair_run(repair_json)
 
 
 def get_launch_task_id(task_group: TaskGroup) -> str:
@@ -381,6 +388,17 @@ class WorkflowJobRepairAllFailedLink(BaseOperatorLink, LoggingMixin):
         *,
         ti_key: TaskInstanceKey | None = None,
     ) -> str:
+        if AIRFLOW_V_3_0_PLUS:
+            if ti_key is None:
+                return ""
+            metadata = _get_launch_metadata_v3(operator, ti_key)
+            if not metadata:
+                return ""
+            # The set of failed tasks is resolved from the live Databricks run by the endpoint.
+            return _build_repair_url(
+                ti_key.dag_id, ti_key.run_id, metadata.conn_id, metadata.run_id, repair_all=True
+            )
+
         if not ti_key:
             ti = get_task_instance(operator, dttm)
             ti_key = ti.key
@@ -478,6 +496,20 @@ class WorkflowJobRepairSingleTaskLink(BaseOperatorLink, LoggingMixin):
         *,
         ti_key: TaskInstanceKey | None = None,
     ) -> str:
+        if AIRFLOW_V_3_0_PLUS:
+            if ti_key is None:
+                return ""
+            metadata = _get_launch_metadata_v3(operator, ti_key)
+            if not metadata:
+                return ""
+            return _build_repair_url(
+                ti_key.dag_id,
+                ti_key.run_id,
+                metadata.conn_id,
+                metadata.run_id,
+                tasks_to_repair=[operator.databricks_task_key],
+            )
+
         if not ti_key:
             ti = get_task_instance(operator, dttm)
             ti_key = ti.key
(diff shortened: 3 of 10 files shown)


Vamsi-klu:
I rechecked the current head and think the repair path needs a security/compatibility redesign before it is safe to validate against a real workspace:

- `GET` should render a read-only confirmation page; only a CSRF-protected, Dag-authorized `POST` should repair or clear tasks.
- The URL should carry only validated Dag/run/task identifiers. `databricks_conn_id`, the Databricks run ID, and Databricks task keys should be derived server-side from trusted `WorkflowRunMetadata` XCom instead of accepted from the request.
- Redirect targets should be same-site relative paths built from validated identifiers, resolving the two open CodeQL redirect findings.
- The FastAPI repair app should be registered only on Airflow 3.1+, since `resolve_user_from_token` is unavailable on 3.0.6. Provider config must come from `airflow.providers.common.compat.sdk`, not `airflow.configuration`, to clear the current static check.
- Databricks failures should produce a generic 502 without reflecting raw upstream exception text, and task clearing should happen only after a successful repair call.
- Confirmation-page values must be safely encoded so identifiers cannot become reflected HTML or JavaScript.

The PR body should also include `related: #52280`. I can prepare a patch branch with the route split, trusted identifier resolution, access checks, and focused DB/API tests if that collaboration route would be useful; I will not open a competing Apache PR while this one is active. Until the mutation-on-GET issue is fixed, I do not recommend clicking the current repair link against a live Databricks workspace.

---
Drafted-by: Codex (GPT-5)

Vamsi-klu:
@PrakshiGoyal10 , Some CI signals are failing. PTAL and fix those. Thanks!

PrakshiGoyal10:
Thanks for the detailed review. I've reworked the repair path along these lines. Rebased onto `main` and pushed; CI is green (static checks, provider tests, and the CodeQL analysis all pass now).

Point by point:

- **Mutation on GET → confirmation + POST.** `GET` now renders a read-only confirmation page; only a same-site `POST` repairs and clears tasks. Rather than add a bespoke CSRF token, this leans on the same mechanism the core UI mutations rely on: the API server's auth cookie is `SameSite=Lax`, so it isn't sent on a cross-site `POST` — moving the mutation off `GET` is what closes the CSRF hole.

- **Trusted identifiers only.** The link now carries only Airflow identifiers — the run's launch `task_id`, and for a single-task repair the target `task_id`. `databricks_conn_id` and the Databricks run id are derived server-side from the launch task's `WorkflowRunMetadata` XCom, never from the request, so a crafted link can't point the repair at an arbitrary connection or run. Task keys are reconstructed from the serialized Dag.

- **Open redirect.** The redirect target is now a same-site relative path built from validated identifiers, which resolves both CodeQL findings.

- **Version gating.** Registration is gated to Airflow 3.1+. I confirmed `resolve_user_from_token` is absent on 3.0.6, and that 3.0.6's `get_user` reads only the bearer header (not the `_token` cookie), so cookie-based link navigation couldn't work there regardless — 3.1+ is the right floor.

- **Error hygiene.** Databricks failures now return a generic `502` with no upstream exception text, and task clearing happens only after a successful repair call.

- **Output encoding.** Confirmation-page values are HTML-escaped so identifiers can't become reflected markup.

- **`related: #52280`** added to the PR body.

Tests cover the GET-doesn't-mutate behaviour, confirmation-page escaping, server-side identifier resolution on the POST, and the generic 502. Appreciate the offer to prepare a patch branch — I think the current revision covers the items raised, but happy to iterate if anything still looks off.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

eladkal:
cc @moomindani for review

providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
```diff
@@ -563,6 +563,28 @@ def get_run_tasks(self, run_id: int) -> list[dict[str, Any]]:
...
+        failed_result_states = {"FAILED", "TIMEDOUT", "CANCELED", "MAXIMUM_CONCURRENT_RUNS_REACHED"}
+        failed_task_keys = []
+        for task in self.get_run_tasks(run_id):
+            state = task.get("state", {})
+            if (
+                state.get("result_state") in failed_result_states
+                or state.get("life_cycle_state") == "INTERNAL_ERROR"
+            ):
+                failed_task_keys.append(task["task_key"])
+        return failed_task_keys
```
moomindani, line 586: This assumes one entry per `task_key`, but `get_run_tasks` returns one entry **per attempt**, so a task that retried (or a run that was already repaired) yields the same key several times. Databricks rejects that:

```
Error: rerun_tasks cannot contain duplicate task keys.
```

I confirmed this against a live run — with `max_retries=1` on a failing task this returns `['flaky', 'flaky']` and the repair fails; deduped to `['flaky']` it succeeds. Attempts accumulate across repairs, so the duplicate count grows (I saw four).

There is a second, quieter symptom: because every historical attempt is inspected, a `task_key` whose **latest** attempt succeeded is still reported as failed. After my run finished fully green (`parent: SUCCESS`, both tasks `SUCCESS`), this still returned `['flaky', 'flaky', 'flaky', 'flaky']`. Databricks does *not* reject repairing an already-successful task — it accepts it and returns a `repair_id` — so that path silently re-runs a healthy task and clears its Airflow TI.

Both symptoms have one root cause, and the fix already exists in this repo: `_get_current_databricks_task` in `operators/databricks.py` resolves the latest attempt by sorting on `start_time` before building the map. Same idiom here:

```python
sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda t: t["start_time"])
latest_by_key = {task["task_key"]: task for task in sorted_tasks}
```

then evaluate the failure states over `latest_by_key.values()`. That dedupes and fixes the stale-attempt judgement in one move.

providers/databricks/tests/unit/databricks/hooks/test_databricks.py
```diff
@@ -749,6 +749,20 @@ def test_get_run_tasks_success_multiple_pages(self, mock_requests):
         assert len(tasks) == 2
         assert tasks == GET_RUN_RESPONSE["tasks"] * 2
 
+    def test_get_run_failed_task_keys(self):
```
moomindani, line 752: This fixture has one entry per `task_key`, which is why CI is green despite the duplicate-key bug — the real API returns one entry per *attempt*.

Worth adding a case shaped like a live run, e.g. two `flaky` entries with `attempt_number` 0/1 both `FAILED` plus a later attempt that `SUCCESS`ed, asserting the key appears once and is judged by its latest attempt. That would fail on the current implementation.

moomindani (commented):
Reviewed this from the Databricks-behaviour side and validated it against a live workspace. The security design holds up, but I found one bug that makes the repair button fail outright in a common case — details inline.

**Security model — checks out.** I verified the load-bearing claims rather than taking them from the description:

- The `SameSite=Lax` argument is real: core sets `samesite="lax"` in `auth/managers/simple/routes/login.py` and `auth/middlewares/refresh_token.py`, so a cross-site `POST` cannot carry the auth cookie. Moving the mutation off `GET` is what closes the hole, and that matches how core UI mutations are protected.
- Deriving `conn_id` / Databricks `run_id` from the launch task's `WorkflowRunMetadata` XCom instead of the request is the right call — a crafted link cannot retarget the repair at another connection or run.
- Authorization via `is_authorized_dag(method="PUT", access_entity=DagAccessEntity.RUN)`, the relative-path redirect with `quote(safe='')`, the generic `502`, and gating to 3.1+ (where `resolve_user_from_token` exists) all look right to me.

**Databricks API usage — correct.** Calling `get_latest_repair_id` before `repair_run` is not optional: on a second repair the API returns `The latest repair ID needs to be provided in order to create a new repair`. `rerun_dependent_tasks: true` is also the right choice — I confirmed `UPSTREAM_FAILED` dependents resume after the repaired task succeeds.

**The bug.** `get_run_failed_task_keys` assumes one entry per `task_key`, but `get_run_tasks` returns one entry *per attempt*. On a live run with `max_retries=1` on a failing task:

```
flaky       attempt=0  TERMINATED/FAILED
flaky       attempt=1  TERMINATED/FAILED
downstream  attempt=0  SKIPPED/UPSTREAM_FAILED

-> get_run_failed_task_keys() == ['flaky', 'flaky']
```

and Databricks rejects that outright:

```
Error: rerun_tasks cannot contain duplicate task keys.
```

Deduped to `['flaky']` the same call succeeds. So "Repair All Failed Tasks" fails with a `502` for any run where a task has retries configured, or that has been repaired before (attempts accumulate — I got up to four duplicates of the same key). Inline note has the fix.

---

PrakshiGoyal10:
Thanks for the careful review and for validating it against a live workspace — the duplicate-key bug is exactly right, and I confirmed it the same way. Pushed a fix.

**Duplicate task keys.** `get_run_failed_task_keys` now keeps only the latest attempt per `task_key` before judging state, using the sort-by-`start_time` idiom from `_get_current_databricks_task` as you suggested. That removes the duplicates that Databricks rejects and also fixes the stale-attempt symptom (a key whose newest attempt succeeded is no longer reported failed). Added a hook test shaped like a live run — multiple `flaky` attempts (FAILED, FAILED, then SUCCESS) plus a task whose latest attempt failed — asserting each key appears once and is judged by its latest attempt.

**A second, related defect I found while fixing that.** Reconstructing the Databricks `task_key` on the API-server side as `md5(dag_id__task_id)` is only correct for tasks that use the default key. An explicit `databricks_task_key` does not survive Dag serialization (I verified the attribute is absent from the deserialized operator), so for such tasks single-task repair targeted a non-existent key and repair-all silently skipped clearing that task's instance. The launch task now records the `task_id` to `task_key` map in its run-metadata XCom, and the endpoint reads it server-side (so no request-supplied keys, consistent with the trusted-identifier model); runs launched before this fall back to the md5 derivation. Covered by tests for both the explicit-key and legacy-fallback paths.

All provider unit tests pass locally. On the still-open CodeQL redirect finding: the redirect target is a same-site relative path built with `quote(..., safe='')`, so it can't leave the origin — I believe it's a false positive, but happy to refactor if you'd prefer it read differently.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

eladkal:
cc @moomindani if all checks out please approve :)

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +547,266 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+        run_id = unquote(run_id)
+        summary = (
+            "This will repair all failed tasks of the run and resume their downstream tasks."
+            if repair_all
+            else f"This will repair task '{task_id}' and resume its downstream tasks."
+        )
+        # Same-site relative action; SameSite=Lax on the auth cookie means a cross-site POST cannot
+        # carry it, so moving the mutation to POST is what protects it from CSRF.
+        action = f"{request.url.path}?{request.url.query}"
+        return _repair_confirmation_page(dag_id, run_id, action, summary)
```
eladkal, line 786: This is being flagged by security scanner as risk. I haven't checked the code yet but please look into this

PrakshiGoyal10: Thanks — addressed the scanner flag on the repair redirect.

The redirect and the confirmation-page form action were assembled from the request's own path and query. The values were percent-encoded into a fixed same-site `/dags/{dag_id}/runs/{run_id}` path and could not actually leave the origin, but CodeQL traced request input into a redirect target and kept the open-redirect alert open regardless.

Pushed a change so no request-controlled string reaches either target:

- The redirect is now built from the `DagRun`'s persisted `dag_id` / `run_id`, read from the database, rather than the request path params.
- The confirmation form action is rebuilt from the already-validated identifiers instead of echoing the request URL.
- Fetching the run also lets the endpoint return a `404` for an unknown run and reuse that row when clearing task instances.

That breaks the taint flow the scanner was following, so the finding should clear on the next analysis. All provider unit tests pass locally.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

moomindani (commented):
Thanks for the quick turnaround on the duplicate-key bug, and for going further and finding the serialization problem yourself.

Confirming the parts that are now right: `get_run_failed_task_keys` does dedupe to the newest attempt per key, the new hook test genuinely fails on the pre-fix code (`assert ['flaky', 'flaky', 'still_bad'] == ['still_bad']`), and the `task_key_map` design is well motivated — in fact more so than you described. It is not only the explicit key that is lost: `get_serialized_fields()` is built from `vars(BaseOperator(task_id="test"))`, so no provider-specific attribute is serialized at all. On a round-tripped Dag the explicit key string appears nowhere in the serialized Dag and `hasattr(op, "databricks_task_key")` is False for both explicit-key and default-key tasks, while the md5 derivation reproduces the default key exactly. All 168 tests pass for me too.

I cannot approve yet, though, because the same serialization boundary breaks the feature in two places that the tests cannot see.

**1. The repair endpoint is unreachable from the Airflow 3 UI, so the new `AIRFLOW_V_3_0_PLUS` branches are dead code there.**

In Airflow 3, extra links are resolved only through XCom: serialization stores `{link.name: link.xcom_key}` (`serialized_objects.py:1505`), deserialization replaces every plugin link with `XComOperatorLink(name, xcom_key)` (`:1485`), and the `/links` route calls `task.get_extra_links(...)` on that (`api_fastapi/core_api/routes/public/extra_links.py:103`). `WorkflowJobRunLink` survives this because it overrides `xcom_key` (`plugins/databricks_workflow.py:305`) and has a producer that writes it during execution (`:374`, `store_databricks_job_run_link`). `WorkflowJobRepairAllFailedLink` (`:380`) and `WorkflowJobRepairSingleTaskLink` (`:487`) override neither, and nothing in `providers/databricks/src/` ever writes `_link_WorkflowJobRepairAllFailedLink` or `_link_WorkflowJobRepairSingleTaskLink`.

Measured by writing the Dag to `serialized_dag`, reading it back through `SerializedDagModel.get_dag` (the API server's own path), creating a real DagRun plus the real run-link XCom row, and calling `get_extra_links` — no mocks:

```
wf.nb     'Repair a single task'    xcom_key='_link_WorkflowJobRepairSingleTaskLink'  url=''
wf.nb     'See Databricks Job Run'  xcom_key='databricks_job_run_link'                url='https://.../#job/1/run/2'
wf.launch 'Repair All Failed Tasks' xcom_key='_link_WorkflowJobRepairAllFailedLink'   url=''
```

The run link resolving correctly is the control that shows the setup is right. The UI then drops a null URL outright (`ui/src/pages/TaskInstance/ExtraLinks.tsx:67`), so no repair button renders and `_build_repair_url` is never called. The link tests pass because they call `link.get_link(...)` on a `Mock` operator rather than on a deserialized Dag.

**2. `_get_launch_task_id_v3` raises on the deserialized operator it will actually receive.**

`get_launch_task_id` calls `task_group.get_child_by_label` (`:255`), which does not exist on `SerializedTaskGroup`. On the same round-trip, the launch task is fine because `".launch" in ti_key.task_id` short-circuits (`:598-599`), but every member task raises:

```
wf.nb: _get_launch_task_id_v3 RAISED AttributeError: 'SerializedTaskGroup' object has no attribute 'get_child_by_label'
```

Member tasks are exactly where the single-task link is registered (`operators/databricks.py:1691-1694`). This is masked in tests because `_get_launch_task_id_v3` is patched out.

Both of these would be caught by one test that goes through `SerializedDagModel.get_dag` and asserts a non-empty URL for a member task, so it may be worth adding that regardless of how the two issues are fixed — otherwise the same class of break can return silently.

Two more things, neither blocking:

**Legacy repair-all can leave Airflow and Databricks out of sync, silently.** At `:791`, `repaired_task_ids = [key_to_task_id[k] for k in task_keys if k in key_to_task_id]`. For a run launched before this change whose task used an explicit key, Databricks returns the real key while `key_to_task_id` holds only md5 keys, so the entry is filtered out: Databricks gets repaired but `_clear_repaired_and_downstream` never touches that task instance, and nothing is logged. The single-task path fails loudly as a 502 in the same situation, which is the better behaviour — worth at least a warning when a returned key cannot be mapped back.

**`get_run_failed_task_keys` is fragile exactly where it runs.** `sorted(..., key=lambda task: task["start_time"])` (`hooks/databricks.py:585`) inherits the idiom from `_get_current_databricks_task`, but that one runs mid-run while this one runs after a failure, when never-started sub-tasks are likely. Measured: a missing `start_time` raises `KeyError`, `None` raises `TypeError`, and an older FAILED attempt alongside a newer PENDING attempt with `start_time` 0 is reported as failed. Any raise here surfaces as a generic 502 (`:808`). A `task.get("start_time") or 0` and a test for the never-started case would close it.

On the redirect change: I agree the original was a false positive — `quote(..., safe='')` encodes `/`, `:`, `#` and `\`, so the target could never leave the origin, and hostile `run_id` values behave identically before and after your change (`//evil.com` and `https://evil.com` simply 404 because the DagRun lookup fails). Building it from the persisted `DagRun` should clear the alert, and it has two real side benefits: the `one_or_none()` 404 now fails before Databricks is called, where previously a missing DagRun raised after the repair had already been issued. Two residuals, though: `return_url` (`:770`) is still a bare `/dags/...` with no base path, so under `[api] base_url = https://host/airflow` the 303 lands on `https://host/dags/...` and 404s; and the form action is now an absolute URL built from `base_url` (`:581`, `:736`), which the config documents as possibly naming a different domain than the front end — in that deployment the POST becomes cross-site and SameSite=Lax withholds `_token`, giving a 401. The new comment describing the action as same-site relative is no longer accurate on either count.

---

Vamsi-klu:
@PrakshiGoyal10  @moomindani  I caught the latest review and went through the plugin plus the extra-link path. Those two blockers look like the ones that keep this from actually showing up in the Airflow 3 UI, and I wanted to ask if you are both okay with me taking a pass at them on this PR.

The repair buttons never render because Airflow 3 only serializes extra links as a name-to-xcom-key map, then the API server rebuilds them as XComOperatorLink. The job-run link already follows that: it sets xcom_key and store_databricks_job_run_link writes the URL during execute. The two repair links do neither, so /links comes back empty and the UI drops the button. The Airflow 3 branches in get_link never run on that path.

The other issue is separate. Even if the URL were built at request time, _get_launch_task_id_v3 still calls get_child_by_label, and SerializedTaskGroup does not have that method. That is the member-task case the single-task link is registered on.

If that works for both of you, I would persist the repair URLs into XCom at the same point as the job-run link (before wait_for_termination, so a later Databricks failure still leaves a button), resolve the launch task from the group id or the upstream .launch task instead of get_child_by_label, and add the SerializedDagModel.get_dag round-trip test so this class of break cannot come back silently.

I can also take the start_time fragility, the unmapped-key warning on repair-all, and the base_url form-action / redirect notes in the same pass if you want those now. I will not open another PR. I can send a patch you can pull, or I can wait if you already have a fix in flight.

moomindani:
Thanks for checking this independently — your reading matches mine on both counts, including the detail that the Airflow 3 branches in `get_link` are unreachable rather than merely wrong.

On who does the work: that is @PrakshiGoyal10's call as the author, not mine. I have no objection either way, and offering a patch on this PR rather than opening a competing one is the right way round.

One technical note before anyone writes code, because there may be a cleaner fix than persisting the URLs. Plugin links are not unconditionally replaced by `XComOperatorLink`: `_deserialize_operator` walks `plugins_manager.get_operator_extra_links()` and matches each link's `operators` against the serialized `task_type` / `_task_module`, and plugin links then take precedence over the xcom-key form (`serialization/serialized_objects.py:1088-1101` and the `op_predefined_extra_links.update(op_extra_links_from_plugin)` line). All three links in this plugin currently declare no `operators`, which is why they get replaced. Declaring them on `WorkflowJobRepairAllFailedLink` and `WorkflowJobRepairSingleTaskLink` would keep the real objects on the API server and let `_build_repair_url` run at request time — no URL duplicated into XCom, no `base_url` frozen at execute time, and the branches this PR already added would actually be exercised.

Persisting into XCom the way the job-run link does also works and has the virtue of matching existing precedent, so I am not arguing hard for one over the other. Worth deciding deliberately rather than by default, and either way the `_get_launch_task_id_v3` fix and the `SerializedDagModel.get_dag` round-trip test are needed — that test is the part that stops this class of break from returning silently, so I would treat it as non-optional.

On scope: the two blockers plus that test are what gate merge from my side. The `start_time` fragility is a genuine crash risk in the same flow and is welcome in the same pass; the unmapped-key warning on legacy repair-all and the `base_url` form-action / redirect notes are lower priority and can equally be follow-ups if this PR is already long in the tooth.

---

PrakshiGoyal10:
Thanks both — this is a great catch, and the `SerializedDagModel.get_dag` round-trip is exactly the kind of test I should have had; my link tests ran against a `Mock` operator and never exercised the deserialized path, so both blockers were invisible to them.

@Vamsi-klu — yes please, go ahead and take a pass at these on this PR; I'd rather build on your fix than have you open a competing one. A patch I can pull works well.

On approach: I don't have a strong preference between persisting the URLs into XCom (matches the job-run link precedent) and @moomindani's suggestion of declaring `operators` on the two link classes so the real link objects survive deserialization and `_build_repair_url` runs at request time. The latter is appealing because it avoids freezing `base_url` at execute time and actually exercises the branches already in the PR — but I'll defer to whichever you and @moomindani think is cleaner.

Appreciate the thorough reviews.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

Vamsi-klu:
@PrakshiGoyal10 please feel free to use the patch. cc @moomindani

PrakshiGoyal10:
Pulled your patch onto the branch — thanks @Vamsi-klu, this is a clean fix.

Confirming what it does, verified locally against a deserialized Dag rather than mocks:

- Both repair-link classes now declare `operators`, so `_deserialize_operator` keeps the real link objects instead of replacing them with `XComOperatorLink` — @moomindani's suggested route, so `_build_repair_url` runs at request time and no URL is frozen into XCom.
- `_get_launch_task_id_v3` no longer calls `get_child_by_label`; it resolves the launch task via the `.launch` suffix, `upstream_task_ids`, then a group walk, so it works on `SerializedTaskGroup`.
- The redirect and the confirmation form action are same-site relative and now include the `[api] base_url` path prefix, so a non-root deployment doesn't 404 after repair.
- `start_time` missing/null is treated as oldest, and legacy repair-all now logs task keys it can't map back instead of dropping them silently.

The `SerializedDagModel.get_dag` round-trip test is the important part — it goes through the API server's own path and asserts a non-empty URL for a member task, so the "buttons silently stop rendering" break can't come back unseen. All 178 provider unit tests pass locally, ruff and mypy clean.

@moomindani — I think this addresses the two blockers plus the test you asked to treat as non-optional. PTAL when you have a chance.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -471,13 +499,38 @@ class WorkflowJobRepairSingleTaskLink(BaseOperatorLink, LoggingMixin):
...
+    @property
+    def operators(self):
+        # Declared so deserialization keeps this plugin link instead of replacing it with
+        # XComOperatorLink. Lazy import avoids a circular import with the operator module.
+        from airflow.providers.databricks.operators.databricks import (
+            DatabricksNotebookOperator,
+            DatabricksTaskOperator,
+        )
+
+        return [DatabricksNotebookOperator, DatabricksTaskOperator]
```
moomindani, line 511: This makes the plugin link match every `DatabricksNotebookOperator` / `DatabricksTaskOperator`, and
plugin links take precedence over the operator's own serialized links ("If OperatorLinks with the same
name exists, Links via Plugin have higher precedence", `serialization/serialized_objects.py:1117-1120`).
That is what fixes the workflow case, but it also overrides `operators/databricks.py:1690-1697`, where
the operator deliberately picks its link set per case:

```python
else:
    # Databricks does not support repair for non-workflow tasks, hence do not show the repair link.
    self.operator_extra_links = (DatabricksJobRunLink(),)
```

On Airflow 3 that is harmless, and I measured it: a standalone `DatabricksNotebookOperator` gains a
`"Repair a single task"` entry that resolves to `None` — `_get_launch_task_id_v3` finds no launch task
and `get_link` returns `""` — and the UI drops null URLs, so only the `/links` response gains a null key.

On Airflow 2 it is not harmless. The provider still declares `apache-airflow>=2.11.0`
(`pyproject.toml:62`); 2.11's deserializer has the same plugin matching and the same precedence merge
(`serialized_objects.py:1266-1272` and `:1310-1315`); and the pre-3.0 branch of `get_link` degrades
differently from the 3.x branch above. For a top-level task `operator.task_group` is the Dag's root
group, so the `if not task_group` guard does not fire, and `get_launch_task_id` then raises
`AirflowException("No launch task can be found in the task group.")`. Airflow 2.11's extra-links view
catches only `ValueError` (`airflow/www/views.py:3249-3252`), so this surfaces as a 500 behind a button
the UI has already rendered — on the plain notebook operator, which is the common case.

I cannot run 2.11 here, so that last chain is code reading rather than measurement, but every step of it
is in the released 2.11 source. Suggested fix in the next comment.

`WorkflowJobRepairAllFailedLink` does not have this problem — `_CreateDatabricksWorkflowOperator` only
ever exists inside a workflow task group.

---

moomindani (commented):
Re-reviewed `045726a7e`. Both blockers are fixed, and I checked them through the API server's own
resolution path rather than through the new test's — `create_dag_bag()` +
`get_dag_for_run_or_latest_version()` + the route's own `task.extra_links` iteration, i.e. the three
things `api_fastapi/core_api/routes/public/extra_links.py` actually does — against a real serialized
Dag round trip. Same probe Dag, pre-fix `c770c1144` vs this head:

| task / link | before | after |
| --- | --- | --- |
| `wf.nb` — "Repair a single task" | `None` | `/databricks/workflow/repair/…?launch_task_id=wf.launch&task_id=wf.nb` |
| `wf.launch` — "Repair All Failed Tasks" | `None` | `…?launch_task_id=wf.launch&repair_all=true` |

So the buttons genuinely render now. I also checked the case the new test does not cover — member
tasks that are not direct downstreams of the launch task (`nb1 >> nb2 >> tk3`, with a
`DatabricksTaskOperator` at the end): all three resolve. That is not luck.
`DatabricksWorkflowTaskGroup.__exit__` assigns `task.workflow_run_metadata =
create_databricks_workflow_task.output` for every member, so the launch task lands in every member's
`upstream_task_ids` and the first branch of `_get_launch_task_id_v3` always hits; the group walk is a
fallback rather than the primary path.

The tests hold up under tampering, which is the part that matters most here:

- reverting only the plugin source and keeping the new tests → the round-trip test fails with exactly
  the `AttributeError: 'SerializedTaskGroup' object has no attribute 'get_child_by_label'` from the
  last review;
- keeping the `_get_launch_task_id_v3` rewrite but deleting just the two `operators` properties → the
  round-trip test still fails, on the "link object survived deserialization" assertion. So it pins both
  blockers independently, not only whichever trips first;
- the new hook test fails pre-fix with `KeyError: 'start_time'` and passes after;
- 832 provider unit tests pass locally, ruff and ruff-format clean.

Everything else from the last round is addressed too: the redirect and the confirmation form action now
carry the `[api] base_url` path prefix, missing/null `start_time` is treated as oldest, and legacy
repair-all logs the keys it cannot map instead of dropping them silently.

One new thing came out of this pass, raised inline: declaring `operators` also re-attaches the repair
link to Databricks tasks that are *not* in a workflow task group, which is harmless on Airflow 3 but
raises on Airflow 2. The fix is small and local. Once that is handled I am happy to approve.

---

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -471,13 +499,38 @@ class WorkflowJobRepairSingleTaskLink(BaseOperatorLink, LoggingMixin):
...
         *,
         ti_key: TaskInstanceKey | None = None,
     ) -> str:
+        if AIRFLOW_V_3_0_PLUS:
+            if not AIRFLOW_V_3_1_PLUS or ti_key is None:
+                # The Airflow-3 repair backend requires 3.1+ (see DatabricksWorkflowPlugin).
+                return ""
+            launch_task_id = _get_launch_task_id_v3(operator, ti_key)
+            if not launch_task_id:
+                return ""
```
moomindani, line 526: This is the graceful degradation the Airflow 2 branch lacks. Now that the link is attached to
non-workflow tasks as well, the cheapest fix is to mirror this shape below: when the launch task cannot
be resolved, return `""` instead of letting `get_launch_task_id(task_group)` raise — i.e. around

```python
if ".launch" not in ti_key.task_id:
    launch_task_id = get_launch_task_id(task_group)
```

catch the "no launch task in this group" case and return `""`, so a standalone Databricks task renders
no URL rather than a 500. That keeps the behaviour of the two branches aligned and leaves the workflow
case untouched.

---

PrakshiGoyal10:
Thanks for the thorough re-review, and for verifying the buttons through the route's own resolution path plus the revert-each-fix tampering — that's a better check than the test alone.

Handled the Airflow 2 case you raised. In the pre-3.0 branch of `WorkflowJobRepairSingleTaskLink.get_link`, the `get_launch_task_id` call is now wrapped so the "no launch task in this group" `AirflowException` returns `""` instead of propagating — a standalone Databricks task renders no repair link rather than surfacing a 500 behind the button, matching the graceful degradation the Airflow 3 branch already has. I reproduced the raise for a top-level task first to confirm the path before changing anything. `WorkflowJobRepairAllFailedLink` is untouched since `_CreateDatabricksWorkflowOperator` only exists inside a workflow group.

Added an Airflow-2 test that asserts the standalone-task case returns `""`. Provider unit tests pass locally, ruff and mypy clean.

@moomindani — I think that closes the last item. PTAL.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/tests/unit/databricks/plugins/test_databricks_workflow.py
```diff
@@ -251,21 +257,58 @@ def test_workflow_job_repair_single_failed_link_airflow2():
...
-                    result = link.get_link(operator, ti_key=ti_key)
-                    assert result.startswith("http://localhost/repair_databricks_job")
+            result = link.get_link(operator, ti_key=ti_key)
+            assert result.startswith("http://localhost/repair_databricks_job")
+
+
+@pytest.mark.skipif(AIRFLOW_V_3_0_PLUS, reason="Test only for Airflow < 3.0")
+@pytest.mark.skipif(
+    RUNNING_TESTS_AGAINST_AIRFLOW_PACKAGES, reason="Web plugin test doesn't work when not against sources"
+)
```
moomindani, line 280: Worth knowing that this test never runs, in CI or locally: the two guards cover every environment between them.

`RUNNING_TESTS_AGAINST_AIRFLOW_PACKAGES` is `True` whenever `USE_AIRFLOW_VERSION` is set *or* `<repo root>/airflow/__init__.py` is missing (`devel-common/src/tests_common/__init__.py:27`) — and in the monorepo layout that file no longer exists, so the constant is `True` even on a sources checkout. In the Airflow 2.11 compatibility job the first clause makes it `True` as well, and on Airflow 3 sources `AIRFLOW_V_3_0_PLUS` skips it first. Measured here: the whole `airflow2` selection reports `11 skipped`, this test among them.

That is structural rather than a defect in this PR — the legacy branch's helpers (`_get_dag`, `_get_dagrun`, the FAB views) are defined inside `if not AIRFLOW_V_3_0_PLUS:`, so nothing in the 2.x path can be exercised from an Airflow 3 checkout, and the neighbouring `*_airflow2` tests sit in the same position. No change requested. I am noting it so nobody later reads this test as a live regression guard for the fix.

moomindani (approved):
Approving — the standalone-task item is closed, and this is the right shape for it.

What I checked on `9cc624e1a`, measured rather than read:

* The trigger the new `except AirflowException` catches is the real one. `get_launch_task_id` raises `No launch task can be found in the task group.` both for a task whose group is the Dag root and for one nested in an unrelated group (the recursion walks up to the root and raises there), so a standalone Databricks task cannot avoid that path and now renders no URL.
* The parity claim holds. On Airflow 3, for a standalone operator `_get_launch_task_id_v3` returns `None` and both repair links return `''` — the 2.x branch now degrades the same way.
* `WorkflowJobRepairAllFailedLink` really is unaffected: its `operators` resolves to `[_CreateDatabricksWorkflowOperator]` only, so the widening that reached the notebook/task operators never reaches a standalone task.
* Provider plugin suite: 32 passed, 11 skipped; ruff and format clean; CI green (55 checks, no failures).

The Airflow 2.11 half of the original finding — that the raise surfaces as a 500 because `airflow/www/views.py` catches only `ValueError` — is still code reading on my side, since 2.11 cannot run in this checkout. The raise itself is now measured, and returning `""` is correct regardless of how the view handles it.

Two things I am recording rather than asking you to change:

* The new Airflow-2 test never executes anywhere — see the inline note. This is structural, not something introduced here.
* Unchanged and pre-existing, from round 2: on the Airflow 2 repair-all path, the Airflow-side clear filters task instances by `databricks_task_key`, so a task carrying an explicit key can be repaired on Databricks while its task instance stays failed. Separate from this PR's scope.

---

PrakshiGoyal10:
Thanks for the approval and for measuring each step rather than taking the description's word for it — the verification on the standalone-task path is much appreciated.

Good catch on the new Airflow-2 test never executing — you're right, the `AIRFLOW_V_3_0_PLUS` and `RUNNING_TESTS_AGAINST_AIRFLOW_PACKAGES` guards cover every environment between them, so it skips everywhere, same as the neighbouring `*_airflow2` tests. I left it in that shape for consistency with those, but I take the point that it reads as a live guard for the fix when it isn't.

Noted the pre-existing Airflow-2 repair-all `databricks_task_key` clear-parity item as out of scope for this PR.

One question on merging: the PR still shows `REVIEW_REQUIRED` / blocked despite the approval — is a committer approval (or a second reviewer) still needed to land it, or is there anything else you'd like from me first?

providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
```diff
@@ -563,6 +563,39 @@ def get_run_tasks(self, run_id: int) -> list[dict[str, Any]]:
...
+        sorted_tasks = sorted(self.get_run_tasks(run_id), key=lambda task: task.get("start_time") or 0)
+        latest_by_key = {task["task_key"]: task for task in sorted_tasks}
+
+        failed_task_keys = []
+        for task_key, task in latest_by_key.items():
+            state = task.get("state", {})
+            if (
+                state.get("result_state") in failed_result_states
+                or state.get("life_cycle_state") == "INTERNAL_ERROR"
+            ):
```
ashb, line 595: Is there no way to do this filtering on the API side?

providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
```diff
@@ -49,11 +49,15 @@ class WorkflowRunMetadata:
...
     :param conn_id: The connection ID used to connect to Databricks.
+    :param task_key_map: Airflow ``task_id`` → Databricks ``task_key`` for the workflow's tasks.
+        Optional and defaulted for backward compatibility with runs launched before it was added
+        (those XComs carry only conn_id/job_id/run_id).
     """
 
     conn_id: str
     job_id: int
     run_id: int
+    task_key_map: dict[str, str] = field(default_factory=dict)
```
ashb, line 60: Why did we (have/choose to) add this here? It wasn't needed in Airflow 2 -- is this really part of the same feature?

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -374,13 +382,33 @@ class WorkflowJobRepairAllFailedLink(BaseOperatorLink, LoggingMixin):
 
     name = "Repair All Failed Tasks"
 
+    @property
+    def operators(self):
+        # Declared so deserialization keeps this plugin link instead of replacing it with
```
ashb, line 387: a) Is this really true
b) why do we need to not use XComOperatorLink? That is the important thing to include in the comment

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -374,13 +382,33 @@ class WorkflowJobRepairAllFailedLink(BaseOperatorLink, LoggingMixin):
...
+
     def get_link(  # type: ignore[override]  # Signature intentionally kept this way for Airflow 2.x compatibility
         self,
         operator,
         dttm=None,
         *,
         ti_key: TaskInstanceKey | None = None,
     ) -> str:
+        if AIRFLOW_V_3_0_PLUS:
+            if not AIRFLOW_V_3_1_PLUS or ti_key is None:
```
ashb, line 403: This conditionall is hard to read. But why do we need to check both, isn't it enough to just do:

```suggestion
        if ti_key is not None:
```

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+        if callable(child_id) and children is not None:
+            launch_id = child_id("launch")
+            if launch_id in children:
+                child = children[launch_id]
+                return getattr(child, "task_id", launch_id)
+        task_group = getattr(task_group, "parent_group", None)
+    return None
+
+
+if AIRFLOW_V_3_1_PLUS:
```
ashb, line 653: What in this block is actually depending on Airflow 3.1?

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+
+        result = session.scalars(
+            XComModel.get_many(
+                run_id=run_id,
+                key="return_value",
+                task_ids=launch_task_id,
+                dag_ids=dag_id,
+                limit=1,
+            )
+        ).first()
```
ashb, line 727: Isn't there an XComModel.get_one()?

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+    )
+
+    async def _resolve_request_user(request: Request):
+        """Authenticate via the bearer header (UI XHR) or the ``_token`` cookie (link navigation)."""
+        token = None
+        auth_header = request.headers.get("Authorization", "")
+        if auth_header.lower().startswith("bearer "):
+            token = auth_header.split(" ", 1)[1]
+        if not token:
+            token = request.cookies.get(COOKIE_NAME_JWT_TOKEN)
```
ashb, line 674: This feels very very wrong. You shouldn't be parsing the HTTP headers yourself, nor making any assumptions what so ever about how the Auth Manager maps request to the user or their perms.

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+
+        target_task_ids: set[str] = set(task_ids)
+        for task_id in task_ids:
+            target_task_ids.update(dag.get_task(task_id).get_flat_relative_ids(upstream=False))
+
+        tis_to_clear = [
+            ti for ti in dag_run.get_task_instances(session=session) if ti.task_id in target_task_ids
+        ]
+        logger.info("Clearing %s task instances after Databricks repair", len(tis_to_clear))
+        clear_task_instances(tis_to_clear, session)
```
ashb, line 752: Is there really not a function already to clear a list of task_ids? IT feels like we shouldn't have to write this

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+    def repair_databricks_workflow(
+        dag_id: str,
+        run_id: str,
+        launch_task_id: str,
+        task_id: str | None = None,
+        repair_all: bool = False,
+        _user=Depends(_require_dag_run_edit),
+    ):
+        """Repair failed Databricks tasks for a workflow run and resume the Airflow run."""
+        run_id = unquote(run_id)
```
ashb, line 798: I'm surprised we need to unquote here -- I would have thought FastAPI would do that for us

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
```diff
@@ -515,6 +574,311 @@ def get_link(  # type: ignore[override]  # Signature intentionally kept this way
...
+
+        from sqlalchemy import select
+
+        from airflow.models.serialized_dag import SerializedDagModel
+        from airflow.utils.session import create_session
+
+        with create_session() as session:
+            dag = SerializedDagModel.get_dag(dag_id, session=session)
+            if dag is None:
+                raise HTTPException(status_code=404, detail="Dag not found.")
```
ashb, line 808: Do we do anything with this object? Is it really worth an extra query to handle this case vs just hitting the "dag_run is None" path below.

ashb (changes requested):

providers/databricks/src/airflow/providers/databricks/hooks/databricks.py
PrakshiGoyal10: The Databricks Jobs API `GET 2.2/jobs/runs/get` returns the whole run with every task and its state — there's no server-side filter for task state — so selecting the failed tasks is necessarily done client-side here.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/operators/databricks_workflow.py
PrakshiGoyal10: This is Airflow-3-specific. An explicit `databricks_task_key` doesn't survive Dag serialization, so on the API server we can't reconstruct the real key for such a task. The launch task records the `task_id → task_key` map in its `WorkflowRunMetadata` XCom so the repair endpoint can map keys back to tasks. Airflow 2 rendered the links from the live operator (which has the key), so it never needed this. I'll add a comment saying so.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: Verified true — a plugin extra link that declares no `operators` is replaced at deserialization by `XComOperatorLink`, which just returns a URL stored under `xcom_key`. This link stores no such URL; it builds one at request time in `get_link`, so it has to survive as the real object. Rewrote the comment to say exactly that (924c9ee).


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: This is tied to the 3.1 gate below — if we relax it (see that thread) this collapses to `if ti_key is None:`. I'll simplify it once the gate question is settled.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: After the auth change, nothing in this block imports anything 3.1-only anymore. The one remaining 3.1 dependency is behavioural: `get_user` only honours the `_token` cookie from 3.1 onward, so on 3.0.x a repair link clicked in the browser wouldn't authenticate (only an XHR carrying a bearer header would). So I can relax the gate to `AIRFLOW_V_3_0_PLUS` and accept that degradation on 3.0.x, or keep it at 3.1 — which would you prefer?


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: There's no `XComModel.get_one()` — it's mentioned in a docstring but not implemented. `get_many(..., limit=1).first()` + `deserialize_value` is the current server-side read pattern (it's what the core XCom route uses). Happy to switch if you'd rather a different call.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: Agreed — fixed in 924c9ee. Replaced the hand-rolled header/cookie parsing with `Depends(requires_access_dag("PUT", DagAccessEntity.RUN))`, the same pattern `providers/common/ai/plugins/hitl_review.py` uses. The `get_user` behind it already resolves both the bearer header and the `_token` cookie, so a browser-clicked link stays authorized, and nothing here parses the request or assumes the auth-manager mapping anymore.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: I can replace the hand-rolled downstream walk + `clear_task_instances` with `find_relevant_relatives(direction="downstream")` + `dag.clear(task_ids=..., run_id=...)`, which is what the core clear-task-instances endpoint uses. Before I rework it — are you OK with `dag.clear` being called here from the provider's FastAPI app?


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: You're right — removed (924c9ee). FastAPI decodes path params already, so it was a redundant double-decode.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
PrakshiGoyal10: The `dag_run` is used for two things: the clear step, and building the redirect target from the run's own persisted `dag_id`/`run_id` rather than the request path — that was to resolve the CodeQL open-redirect finding (a same-site path from DB-sourced values, not request input). If I move clearing to `dag.clear(run_id=...)`, the object would only be needed for that redirect. Would you prefer I keep it DB-sourced for CodeQL, or drop the query and build the redirect from the (validated) path params?


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

providers/databricks/src/airflow/providers/databricks/plugins/databricks_workflow.py
moomindani: One correction to the premise here, since it moves where the gate can sit rather than whether it should.

`get_user` picks up the `_token` cookie from **3.1.1**, not 3.1.0. In 3.1.0 it takes only the two header schemes:

```python
async def get_user(
    oauth_token: str | None = Depends(oauth2_scheme),
    bearer_credentials: HTTPAuthorizationCredentials | None = Depends(bearer_scheme),
) -> BaseUser:
```

The `token_str = request.cookies.get(COOKIE_NAME_JWT_TOKEN)` branch appears in 3.1.1 and is present in every release I checked after it (3.1.2-3.1.5, 3.1.8, 3.2.x, 3.3.x).

`AIRFLOW_V_3_1_PLUS` is `>= (3, 1, 0)`, so on exactly 3.1.0 a repair link clicked in the browser carries only the cookie, `requires_access_dag` resolves no token, and the click 401s — which the hand-rolled resolver did handle. So the options are to gate at 3.1.1 (patch-level gates are already used elsewhere, e.g. `AIRFLOW_V_3_1_9_PLUS` in the celery provider) or to keep 3.1 and say in that comment that browser clicks need 3.1.1+.

Relaxing to `AIRFLOW_V_3_0_PLUS` looks closed off for the same reason: 3.0.6's `get_user` reads only `oauth2_scheme`, so a cookie-only navigation never authenticates there either.

---

PrakshiGoyal10: Thanks for the precise catch — you're right, it's 3.1.1, not 3.1.0. Gated the backend, its extra links, and the test at `AIRFLOW_V_3_1_1_PLUS` (`>= (3, 1, 1)`), following the patch-level precedent in the amazon/fab providers, so a browser-clicked link is only offered where `get_user` actually honours the `_token` cookie. Also confirmed relaxing to 3.0+ is closed off for the same reason (3.0.6's `get_user` is oauth-only). Pushed in 549f946.


---
Drafted-by: Claude Code (Opus 4.8); reviewed by @PrakshiGoyal10

PR #70000, opened 2026-07-17
Strip connection headers on cross-host redirect in HttpHook

Samin061:
An HTTP connection's `extra` field is documented as a place to put credentials ("Login and Password authentication can be used along with any authentication method using headers"), and those end up in `session.headers`, while `run_and_check` follows redirects by default. `requests` strips only `Authorization` once `should_strip_auth` sees the host change, so a secret under any other name (`X-API-Key`, `Private-Token`) is replayed to the redirect target — whether that is an attacker's host reached through an open redirect on the API, or simply the CDN the API bounces downloads to. This gives the connection's own headers the same treatment as `Authorization`; same-host redirects still carry them.

`HttpAsyncHook` has the same gap and `aiohttp` behaves identically, but it offers no equivalent redirect hook, so covering it means reworking the async redirect handling rather than extending an existing one. I left it out to keep this reviewable, and I'm happy to follow up if you'd like it in scope.

The `FakeSession` patch targets in the sensor tests move to the new class because the hook now instantiates it.

---

### DIFF
--- a/devel-common/src/tests_common/test_utils/aiohttp.py
+++ b/devel-common/src/tests_common/test_utils/aiohttp.py
@@ -40,6 +40,7 @@ class MockAiohttpClientResponse:
         method: str = "GET",
         url: str = "http://example.com",
         reason: str | None = None,
+        headers: dict[str, str] | None = None,
     ) -> None:
         self.status = status
         self._payload = payload
@@ -47,6 +48,7 @@ class MockAiohttpClientResponse:
         self._method = method
         self._url = url
         self._reason = reason
+        self.headers = headers or {}
 
     @property
     def reason(self) -> str:
@@ -56,6 +58,9 @@ class MockAiohttpClientResponse:
             return HTTPStatus(self.status).phrase
         except ValueError:
             return ""
+
+    async def release(self) -> None:
+        return None
 
     def raise_for_status(self) -> None:
         if self.status >= HTTPStatus.BAD_REQUEST:
--- a/providers/http/docs/changelog.rst
+++ b/providers/http/docs/changelog.rst
@@ -27,6 +27,13 @@
 Changelog
 ---------
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+* ``HttpHook no longer forwards Connection extra headers across a cross-host redirect (#70000)``
+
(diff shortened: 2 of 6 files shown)


aaron-y-chen (commented):
This looks like the intended security behavior, I think we might need to write down a change log in somewhere to let users know about it. 

For example, users who intentionally relied on connection extra headers being replayed to a different redirect host, such as a CDN, may need to update their flow or handle the redirect explicitly.

aaron-y-chen (commented):
I'm not a committer, so I'd defer to others on whether this change and its scope look right. If so, I think we should open an issue to track the async part so it doesn't get lost.

Samin061:
Good call on both.

Added a breaking-changes note at the top of `providers/http/docs/changelog.rst` covering exactly that case: if you were relying on the connection's headers reaching the redirect target (a CDN or object store that needs the same key), the options are to point the connection at a host that matches the final destination, or set `allow_redirects=False` in `extra_options` and issue the second request yourself.

Opened #70164 for the async side, and left a pointer to it next to the header handling in `HttpAsyncHook.config` so it survives after this merges. `aiohttp` has no per-redirect callback equivalent to `requests.Session.rebuild_auth`, so that fix means walking the redirect chain manually rather than subclassing, which is why it isn't in here.

jroachgolf84:
Linking to #70164

dabla:
@Samin061 Nice work, not sure if extending Session is the right call even though it's a very clean and an elegant solution.  Would be nice if we have a common way to do it for both sync and async http hooks, dunno if that's feasible though as I haven't checked in detail.  Anyway I would like to see the solution for both in this PR, good that you opened a issue to track the async part though.

Samin061:
Makes sense, done. Added the async side to this PR so both hooks are covered here.

Both now share one `_redirect_strips_headers` helper for the host-change decision. The sync hook keeps the `Session` subclass because `rebuild_auth` is the per-redirect hook `requests` gives us; `aiohttp` has no equivalent, so `HttpAsyncHook` follows the redirect chain itself with `allow_redirects=False` and drops the connection-supplied header names on a cross-host hop using that same helper. To keep the blast radius small the async hook only takes over redirect handling when the Connection actually contributed headers, otherwise `aiohttp`'s native handling is untouched. Same-host redirects still carry the headers.

Added a parametrized async test mirroring the sync one (cross-host strips, same-host keeps), and this now folds in #70164.

PR #70011, opened 2026-07-17
Fix database connection leaks in resetdb functions

magic-peach:
## Summary

Two database connection leaks in the resetdb code path:

### 1. `_resetdb_default` (`airflow-core/src/airflow/utils/db.py:1328`)

```python
connection = settings.get_engine().connect()
with create_global_lock(session=session, lock=DBLocks.MIGRATIONS), connection.begin():
    ...
```

A new database connection is opened but never closed. There is no `try/finally` with `connection.close()`, and the connection is not used as a context manager. Compare with the MySQL variant (`_resetdb_mysql`) at line 1307 which correctly uses `engine.connect() as connection`.

### 2. `BaseDBManager.resetdb` (`airflow-core/src/airflow/utils/db_manager.py:189`)

```python
connection = settings.engine.connect()
with create_global_lock(self.session, lock=DBLocks.MIGRATIONS), connection.begin():
    ...
```

Identical pattern. The `with` statement manages a transaction (via `connection.begin()`) and a lock, but does not close the connection itself. After the `with` block exits, the connection remains open and the reference is lost.

### Fix

Changed both functions to use `engine.connect() as connection` (context manager form), matching the pattern already used in `_resetdb_mysql`. This ensures connections are properly closed after use.

---

- [X] Yes — Claude Code (big-pickle)

Generated-by: Claude Code (big-pickle) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/utils/db.py
+++ b/airflow-core/src/airflow/utils/db.py
@@ -1325,8 +1325,8 @@ def _resetdb_mysql(session: Session) -> None:
 
 def _resetdb_default(session: Session) -> None:
     """Drop all Airflow tables for PostgreSQL/SQLite."""
-    connection = settings.get_engine().connect()
-    with create_global_lock(session=session, lock=DBLocks.MIGRATIONS), connection.begin():
+    engine = settings.get_engine()
+    with create_global_lock(session=session, lock=DBLocks.MIGRATIONS), engine.connect() as connection:
         drop_airflow_models(connection)
         drop_airflow_moved_tables(connection)
         log.info("Dropped all Airflow tables")
--- a/airflow-core/src/airflow/utils/db_manager.py
+++ b/airflow-core/src/airflow/utils/db_manager.py
@@ -186,9 +186,8 @@ class BaseDBManager(LoggingMixin):
 
         self._release_metadata_locks_if_needed()
 
-        connection = settings.engine.connect()
-
-        with create_global_lock(self.session, lock=DBLocks.MIGRATIONS), connection.begin():
+        engine = settings.engine
+        with create_global_lock(self.session, lock=DBLocks.MIGRATIONS), engine.connect() as connection:
             self.log.info("Dropping %s tables", self.__class__.__name__)
             self.drop_tables(connection)
         if not skip_init:


potiuk:
Thanks for digging into this — the observation underneath it is correct: `connection = settings.get_engine().connect()` is never closed, and the connection only returns to the pool when the fairy is finalized rather than at end of use.

Unfortunately the fix as written is a regression, so I'm closing it rather than merging.

Replacing `connection.begin()` with `with engine.connect() as connection:` doesn't just add cleanup — it changes the transaction semantics. `connection.begin()` as a context manager **commits** on clean exit. `engine.connect()` is SQLAlchemy 2.0 "commit as you go": nothing calls `.commit()`, so `Connection.close()` **rolls back** on block exit. `drop_airflow_models()` and `drop_tables()` don't commit internally, so the DROPs get reverted.

Against a real PostgreSQL 16 with SQLAlchemy 2.0.51:

```
current: connection.begin()      -> table SURVIVED the drop: False
this PR: engine.connect() as conn -> table SURVIVED the drop: True
```

`_resetdb_default` is the default path for both PostgreSQL and SQLite, so on Postgres this makes `airflow db reset` silently stop dropping anything. Both hunks have it. It happens to be masked on SQLite because pysqlite doesn't open a transaction for DDL, so DDL autocommits there — Postgres has real transactional DDL, so it breaks.

On `_resetdb_mysql` as the precedent: that function is only safe with `engine.connect() as connection` because the next line switches to `connection.execution_options(isolation_level="AUTOCOMMIT")` and runs all DDL on `ddl_conn`. The AUTOCOMMIT is what makes the pattern work there; without it the same bug applies.

Green CI isn't evidence here either. The only test that actually calls `resetdb()` is `test_resetdb_logging_level`, which asserts on the logging level and nothing else, and `resetdb()` calls `initdb()` right afterwards — so the tables come back and a failed drop is invisible. Nothing asserts the tables are gone.

If you'd like to pursue the original cleanup, the version that keeps both properties is:

```python
with (
    create_global_lock(session=session, lock=DBLocks.MIGRATIONS),
    settings.get_engine().connect() as connection,
    connection.begin(),
):
```

or `engine.begin() as connection`, which is connect + begin + commit + close in one. A new PR doing that, with a test that asserts the tables are actually dropped, would be welcome. Worth knowing that the resource issue itself is minor — `resetdb` is a one-shot CLI operation — so the test matters more than the cleanup.

---

PR #70026, opened 2026-07-17
Add subdomain to worker pods in Deployments

stephen-bracken:
Add subdomain to worker pods in Deployments
---
When not using a StatefulSet, setting the `spec.subdomain` property on the pod to the service name will return the fully qualified service address for the pod when calling `airflow.utils.net.getfqdn`, allowing the UI to resolve the worker address and retrieve the logs.

- [X] No

---

### DIFF
--- a/chart/templates/workers/worker-deployment.yaml
+++ b/chart/templates/workers/worker-deployment.yaml
@@ -123,6 +123,11 @@ spec:
           {{- tpl (toYaml $podAnnotations) . | nindent 8 }}
         {{- end }}
     spec:
+      {{- if not $persistence }}
+      # Fix TaskInstance hostname service resolution for Deployments
+      # https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-hostname-and-subdomain-field
+      subdomain: {{ include "airflow.fullname" . }}-worker{{ if ne .Values.workers.celery.name "default" }}-{{ .Values.workers.celery.name }}{{ end }}
+      {{- end }}
       {{- if .Values.workers.celery.runtimeClassName }}
       runtimeClassName: {{ .Values.workers.celery.runtimeClassName }}
       {{- end }}
--- a/chart/templates/workers/worker-deployment.yaml
+++ b/chart/templates/workers/worker-deployment.yaml
@@ -124,8 +124,6 @@ spec:
         {{- end }}
     spec:
       {{- if not $persistence }}
-      # Fix TaskInstance hostname service resolution for Deployments
-      # https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-hostname-and-subdomain-field
       subdomain: {{ include "airflow.fullname" . }}-worker{{ if ne .Values.workers.celery.name "default" }}-{{ .Values.workers.celery.name }}{{ end }}
       {{- end }}
       {{- if .Values.workers.celery.runtimeClassName }}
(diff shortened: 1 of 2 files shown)


Miretpl:
I assume that the goal of this PR is to make the set of `core.hostname_callable=airflow.utils.net.get_host_ip_address` not needed (correct me if I'm wrong). I checked the change locally, and it doesn't work. Could you provide the configuration and how you tested it?

stephen-bracken:
> I assume that the goal of this PR is to make the set of core.hostname_callable=airflow.utils.net.get_host_ip_address not needed (correct me if I'm wrong). 

Partially yes. This to provide consistency between the StatefulSet which uses the `serviceName` property to set the worker subdomain, and the Deployment. ~~In order to make the calls work through the service we would need to make a new method that strips the pod name from the FQDN returned by `airflow.utils.net.getfqdn`~~, e.g.
```python
def get_host_service_name():
    """Service-aware hostname callable. 
    
    https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records-1
    """
    from airflow.utils.net import getfqdn
    fqdn = getfqdn()
    return ".".join(fqdn.split(".")[1:])

# returns myworkers.mynamespace.svc.cluster.local
```

Without the subdomain this method does not work, as the pod's FQDN will only return the pod name without the subdomain.

_edit: corrected mistake on routing to worker logs_

stephen-bracken:
I think additional changes will be needed to make `getfqdn` work with worker Deployments, but this is a required first step.

chart/templates/workers/worker-deployment.yaml
```diff
@@ -123,6 +123,11 @@ spec:
           {{- tpl (toYaml $podAnnotations) . | nindent 8 }}
         {{- end }}
     spec:
+      {{- if not $persistence }}
+      # Fix TaskInstance hostname service resolution for Deployments
+      # https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-hostname-and-subdomain-field
```
Miretpl:
```suggestion
```
This is Kubernetes-specific knowledge that the reader should already have.

Miretpl (changes requested):
Could you add some test cases?

chart/tests/helm_tests/airflow_core/test_worker.py
```diff
@@ -1552,6 +1552,16 @@ def test_overwrite_extra_ports(self):
             {"name": "test-extra-port", "containerPort": 10}
         ]
 
+    def test_worker_subdomain_matches_service(self):
```
Miretpl, line 1563: Small nit: I would parameterise this test by persistence to make sure that the value is the same for the default StatefulSet behaviour and the Deployment one.

Miretpl (changes requested):

chart/tests/helm_tests/airflow_core/test_worker.py
stephen-bracken: If persistence is disabled then the StatefulSet will be used, which sets `spec.serviceName` instead of `spec.template.spec.subdomain`

Miretpl: I see that I wasn't precise in my comment, but the result is what I meant. Thanks!

Miretpl (approved):

PR #67763, opened 2026-05-30
Fix Task Instance mark-as downstream default

Codingaditya17:
## What

This PR updates the Task Instance mark-as dialog so that the `downstream` option is no longer selected by default.

## Why

In the Task Instances view, marking a failed task instance as success should not automatically include downstream task instances unless the user explicitly selects that option.

Previously, the dialog defaulted to `downstream`, which meant the request was sent with `include_downstream: true`. This could cause downstream `upstream_failed` task instances to be affected and resume within the same DagRun, which differs from the older Airflow 2 Task Instances view behavior.

This change makes the action safer and more explicit: only the selected task instance is updated by default, while users can still choose downstream manually when they want that behavior.

## How

- Removed the default `downstream` selection from `MarkTaskInstanceAsDialog`
- Added a regression test to verify that `downstream` is not selected by default

## Tests

cd airflow-core/src/airflow/ui
pnpm test MarkTaskInstanceAsDialog.test.tsx

Result:
✓ src/components/MarkAs/TaskInstance/MarkTaskInstanceAsDialog.test.tsx (1 test) 123ms
  ✓ MarkTaskInstanceAsDialog (1)
    ✓ does not select downstream by default 120ms

Test Files  1 passed (1)
Tests       1 passed (1)

uv run prek run ts-compile-lint-ui --all-files

Result:
Running hooks for `airflow-core`:
Compile / format / lint UI...............................................Passed

PR #70048, opened 2026-07-17
Fix regression: Mark Task Instance As no longer defaults to Downstream

luisglft:
## Summary

#52130 added `defaultValues={["downstream"]}` to the shared `SegmentedControl` (Past/Future/Upstream/Downstream) used by both `ClearTaskInstanceDialog` and `MarkTaskInstanceAsDialog`.

- For **Clear Task Instance**, this was correct: that dialog's own state already hardcoded `useState(["downstream"])` before #52130 — the PR only fixed a display bug where the button looked unselected despite downstream already being included in the clear request.
- For **Mark Task Instance As**, this is a real regression: before #52130, nothing was selected by default, so marking a task as success/failed only affected that task. Now "Downstream" is selected by default, so marking a task also marks its downstream task instances with the same state — but **without executing them**, since Mark-As sets state directly rather than running anything. This silently short-circuits the DAG for anyone who doesn't notice and manually deselect "Downstream" before confirming.

This PR removes the `defaultValues={["downstream"]}` prop from `MarkTaskInstanceAsDialog` only, restoring the pre-#52130 opt-in default. `ClearTaskInstanceDialog` is untouched.

## Test plan

- [ ] Open "Mark Task Instance as success/failed" on a task instance — confirm no options are pre-selected.
- [ ] Open "Clear Task Instance" on a task instance — confirm "Downstream" is still pre-selected (unchanged).


luisglft:
Common use case for mark a task as success is to allow downstream tasks to run without getting blocked by that task, so we want to have downstream tasks running instead of mark as success. This was the default behavior in Airflow 2

bbovenzi:
Duplicate of https://github.com/apache/airflow/pull/67763

PR #70053, opened 2026-07-17
feat: per-component enableBuiltInSecretEnvVars for workers and scheduler

### LINKED ISSUE #70008: Hiding sensitive environment variables

Helm chart

### Apache Airflow version

Any airflow version

Hello, I dont know if it has been asked or what but couldn't find what I wanted.

in Airflow since sensitive credentials are stored via environment variables (meaning they are pulled securely during build but while it's already running), end users can have bashoperator and run `printenv` command and get all sensitive credentials regarding database, logging backends, secret backends etc which shouldnt be available for end users but admins only. Even if printenv is limited there are still other ways to get those credentials from environment.

I want to know what can I do in this scenario ?

Any command could potentially get sensitive credentials should be hidden or shouldnt have privilege to run specific list of commands

### Operating System

Linux

### Deployment

_No response_

_No response_

1.21.0

### Kubernetes Version

1.34.3

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

anmolxlight:
### Motivation

The Helm chart currently injects all sensitive config values (database connection, broker URL, fernet key, api secret key, celery result backend) into every component's environment via `standard_airflow_environment`. For worker pods that execute untrusted DAG code, this means user tasks can read these secrets with `printenv` or `/proc/self/environ`.

The existing global `enableBuiltInSecretEnvVars` flag lets operators disable these env vars, but it disables them for **all** components (scheduler, api-server, triggerer, worker, etc.), which may break control-plane components that legitimately need them.

### What this PR does

- Adds `workers.celery.enableBuiltInSecretEnvVars` (type: `object | null`, default: `null`)
- Adds `scheduler.enableBuiltInSecretEnvVars` (type: `object | null`, default: `null`)
- When a key is set in a component-level override, it takes precedence over the global value for that component only
- When `null` or a key is absent, the global `enableBuiltInSecretEnvVars` is used (backward compatible)

### Usage

```yaml
workers:
  celery:
    enableBuiltInSecretEnvVars:
      AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: false
      AIRFLOW__CELERY__BROKER_URL: false
```

This disables the database connection and broker URL env vars on worker containers only, while other components continue to receive them from the global setting.

### Implementation

The `standard_airflow_environment` template accepts an optional `envOverrides` dict. Each component call site passes its component-level map when available. The template resolves each env var by checking the component override first, then falling back to the global.

### Backward Compatibility

Fully backward compatible. Default values (`null`) produce identical behavior to the existing global-only approach.

Closes #70008


Venture200:
@anmolxlight can you please also add secret backends credentials from Azure, Vault etc

PR #62153, opened 2026-02-19
fix: fab deserialize issue

cruseakshay:
## Problem

When using FAB auth manager, a database connection drop (e.g. PostgreSQL's
`idle_in_transaction_session_timeout`) causes the API server to return HTTP 500
on **every subsequent request** until it is restarted.

The cascade happens in the JWT auth path hit on every authenticated request:

`JWTRefreshMiddleware` → `resolve_user_from_token` → `deserialize_user`

`deserialize_user` uses FAB's scoped session (`self.appbuilder.session`). When a
connection dies, SQLAlchemy raises `OperationalError` on the first request and
leaves the session in an invalid state. All following requests reuse the same
poisoned thread-local session and raise `PendingRollbackError`.

This is distinct from the WSGI Flask-view path fixed in #61480 and the
`load_user` path fixed in #61943 — those do not cover the JWT token
deserialization path.

## Solution

Catch `SQLAlchemyError` in `deserialize_user`, call `session.remove()` to
discard the poisoned scoped session, and re-raise the original exception.
The next request gets a fresh connection from the pool and succeeds.

`session.remove()` is wrapped in `contextlib.suppress(Exception)` so a failure
during cleanup can never mask the original database error.

- **First request after a drop**: unavoidable 500 (the dead connection must be
  discovered) — behaviour is unchanged.
- **All subsequent requests**: recover automatically — no restart needed.

## Testing

- `test_db_error_calls_session_remove` — parametrized over `OperationalError`
  and `PendingRollbackError`: verifies `session.remove()` is called on each.
- `test_db_error_propagates_when_session_remove_raises` — verifies the original
  `SQLAlchemyError` is always what propagates, even when `session.remove()` itself
  throws.

Fixes #61761 | #61518
Related to #61480, #61943

Issue #57859, opened 2025-11-05
Airflow Web Interface Crashes with 500 Error (FAB Auth, MySQL Timeout)


FazalUrrehmanKhan:
### Apache Airflow version

3.1.1

### What happened?

**Issue Summary**

I'm facing an issue with the Airflow web interface authentication since upgrading Airflow from v3.0.6 to v3.1.0. The problem still persists on v3.1.1.

I’m using the FAB (Flask AppBuilder) auth manager for authentication, and I suspect the issue is related to it.

**Problem Description**

After the upgrade, the Airflow web interface intermittently becomes inaccessible. The issue typically appears:

- After a day of normal operation, or
- When a second user logs in to Airflow from another device.

When the issue occurs, attempting to access the Airflow web UI results in a 500 Internal Server Error.
Restarting the Airflow webserver temporarily resolves the issue, but it reappears after a few hours.

**Troubleshooting Performed**

- Tried switching database libraries from pymysql to mysqldb.
- Verified the database connection and Airflow configuration.
- Observed that the issue correlates with database connection or session handling errors (example log below).

**Recent Log Snippet**

`sqlalchemy.exc.OperationalError: (MySQLdb.OperationalError) (4031, 'The client was disconnected by the server because of inactivity. See wait_timeout and interactive_timeout for configuring this behavior.')
[SQL: SELECT ab_user.id AS ab_user_id, ab_user.first_name AS ab_user_first_name, ab_user.last_name AS ab_user_last_name, ab_user.username AS ab_user_username, ab_user.password AS ab_user_password, ab_user.active AS ab_user_active, ab_user.email AS ab_user_email, ab_user.last_login AS ab_user_last_login, ab_user.login_count AS ab_user_login_count, ab_user.fail_login_count AS ab_user_fail_login_count, ab_user.created_on AS ab_user_created_on, ab_user.changed_on AS ab_user_changed_on, ab_user.created_by_fk AS ab_user_created_by_fk, ab_user.changed_by_fk AS ab_user_changed_by_fk FROM ab_user WHERE ab_user.id = %s]
[parameters: (1,)]
`

After the above error, I get the following error for any attempts in login to airflow web interface.

`sqlalchemy.exc.PendingRollbackError: Can't reconnect until invalid transaction is rolled back. (Background on this error at: https://sqlalche.me/e/14/8s2b)`

### What you think should happen instead?

Airflow web interface should remain accessible and handle multiple user logins without requiring frequent webserver restarts.

[...]

PR #70057, opened 2026-07-17
Don't fail the request with a bare 500 on transient JWT refresh errors

kjh0623:
**Problem**

`JWTRefreshMiddleware.dispatch` runs outside the FastAPI/Starlette exception handlers, and its try/except only catches `HTTPException` and `AuthManagerRefreshTokenExpiredException`. Any other exception raised while resolving/refreshing the user propagates to the ASGI layer and surfaces as a bare HTTP 500 — for a request that may not have needed the refresh at all.

The concrete way we hit this in production: PgBouncer recycles a backend connection underneath the API server, and the next request's `_refresh_user()` → `resolve_user_from_token()` → user lookup raises `sqlalchemy.exc.OperationalError` ("server closed the connection unexpectedly"). The UI shows "Internal Server Error"; a manual refresh usually succeeds. Related reports of the same pain point: #57859, discussion #59487. (#62153 fixed the *poisoned-session cascade* after such a drop; this PR addresses the request that hits the error itself.)

**Fix**

Add a final `except Exception` arm around the refresh block that treats the failure as "could not refresh this time":

- log a warning with the traceback,
- proceed **without** injecting the user — the route's own auth dependency revalidates the token exactly as it does today when the middleware doesn't inject, and
- leave the cookie untouched (no forced logout), so a subsequent request refreshes normally once the backend recovers.

`HTTPException` / token-expiry semantics are unchanged — those still clear the cookie via the existing arm.

**Testing**

New unit test: `_refresh_user` raising a non-HTTP exception → request completes with the downstream response, no `set-cookie` header, warning logged. Existing middleware tests unchanged and passing.

We have been running the equivalent of this change in production (as a patch on 3.1.8) on a deployment serving a ~3,600-DAG instance; the recurring bare 500s on the auth path disappeared.

---

---


airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
```diff
@@ -64,6 +68,18 @@ async def dispatch(self, request: Request, call_next):
...
                     new_token = ""
+                except Exception:
+                    # This middleware runs outside the FastAPI exception handlers, so any
+                    # other exception raised while resolving/refreshing the user - e.g. an
+                    # OperationalError from a dropped database connection - would surface
+                    # as a bare 500 for a request that may not even need the refresh.
+                    # Proceed without injecting the user (the route's own auth dependency
+                    # revalidates the token) and leave the cookie untouched so a later
+                    # request can refresh once the backend recovers.
+                    log.warning(
```
SameerMesiah97: is` except Exception` intentionally this broad? It seems like this would also swallow programming errors and other non-transient failures. Would it make sense to catch the specific infrastructure exceptions we're expecting (e.g. database/connection errors), while allowing genuinely unexpected exceptions to surface?

SameerMesiah97 (commented):
I think there is a disconnect between the stated motivation behind this PR and the actual code. You appear to want to hande transient infrastructure failures (e.g. dropped DB connections), but the imeplementation and tests appear to broaden this to any exception from `_refresh_user()`. Is that intentional?

kjh0623:
Good question — the broad catch is intentional, but you're right that the observability trade-off deserves a callout.

**Why `except Exception` rather than a specific `OperationalError`:** this middleware is a generic auth layer that delegates the actual lookup to the configured auth manager (`_refresh_user` → `resolve_user_from_token`). It doesn't — and shouldn't — know which backend that auth manager uses, so it can't enumerate which exceptions are "transient infra." Catching `sqlalchemy.exc.OperationalError` here would bake a DB-backend assumption into a layer that's meant to be backend-agnostic, and it would still let any *other* transient cause surface as the exact bare 500 this PR is trying to remove.

**Why failing open is safe regardless of the exception type:** the refresh is a best-effort, proactive step, not the authoritative auth check. If it doesn't inject a user, the route's own auth dependency revalidates the token exactly as it does today when the middleware chooses not to refresh. So a failure here is never a security downgrade — it just means "no proactive refresh this time," identical to the existing no-refresh path.

**On the part I think is fair — masking programming errors:** they aren't silently swallowed (the traceback is logged with `exc_info=True`), but WARNING is easy to miss, and a monitor keyed on 5xx would go quiet on a genuine regression. If you'd prefer, I'll switch the unexpected arm to `log.exception(...)` (ERROR level) so real bugs still surface loudly in monitoring while transient blips stay non-fatal. That keeps the resilience without hiding regressions.

Would that address the concern, or would you rather I narrow the catch despite the coupling/coverage downsides above?

airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
kjh0623: Done — switched the unexpected arm to `log.exception()` in d450d2d, so the swallowed error is logged at ERROR with its traceback instead of WARNING.

The catch stays broad for the reason in my earlier comment (the middleware delegates the lookup to whatever auth manager is configured, so it can't enumerate which exceptions are transient for that backend without baking in a DB assumption). What your point does change is visibility, and that part was fair: at ERROR, a genuine bug is at least as loud as the bare 500 it replaces, rather than hiding at WARNING. The test now asserts both the level and that `exc_info` is attached.

PR #70058, opened 2026-07-17
Add refresh_on_initialize option to GitDagBundle

kjh0623:
**What this adds**

An opt-in `refresh_on_initialize` kwarg (default `True`, no behavior change for existing users) on `GitDagBundle`, for deployments where many ephemeral processes initialize the same bundle over shared storage:

```
GitDagBundle(tracking_ref="main", refresh_on_initialize=False)
```

Today, unversioned `initialize()` always ends in a remote fetch. That is the right default for a dag processor keeping the bundle fresh, but with KubernetesExecutor every worker pod initializing the bundle pays that fetch on startup — for our repository this adds ~15s of cold-start to every single task, even though the shared volume already holds a tracking repo the dag processor refreshes continuously. Shallow/sparse clones shave the cost down but every pod still pays a network round-trip.

With `refresh_on_initialize=False`, `initialize()` reuses the tracking repo already on disk as-is and skips all network operations. Scope guards:

- the repository is still cloned (and fetched) when nothing is on disk yet,
- explicit `refresh()` calls still fetch as usual (both repo handles are assigned in the reuse path),
- version-pinned initialization is unaffected (it has its own reuse paths already),
- any failure reusing the on-disk repo falls back to the full initialization path.

This generalizes what we currently run in production as a small `GitDagBundle` subclass (workers reuse the PVC-resident tracking repo, only the dag processor fetches). Related context on the worker cold-start motivation was discussed in #70003.

**Included**

- `refresh_on_initialize` kwarg + docstring
- Unit tests: reuse path serves the on-disk state and skips the fetch while explicit `refresh()` still fetches; default behavior unchanged (second initialize fetches latest)

---

---


providers/git/src/airflow/providers/git/bundles/git.py
```diff
@@ -195,6 +205,38 @@ def _initialize(self):
...
+                # handles are still assigned so a later explicit refresh() works as usual.
+                try:
+                    repo = Repo(self.repo_path)
+                    try:
+                        repo.git.checkout(self.tracking_ref)
+                        bare_repo = Repo(self.bare_repo_path)
+                    except Exception:
+                        repo.close()
+                        raise
+                except (InvalidGitRepositoryError, NoSuchPathError, GitCommandError) as e:
```
Sanjays2402: the inner block re-raises non-git errors (the bare `raise` after close), but this outer except only catches InvalidGitRepositoryError/NoSuchPathError/GitCommandError. so a checkout failing with something like a PermissionError or OSError on the on-disk repo escapes here instead of falling through to full initialization, which defeats the "reuse it as-is, fall back if unusable" intent in the comment. intended?

kjh0623:
Good catch — you're right, that's a real gap, not intended.

The inner block catches everything and re-raises after `repo.close()` (so the handle doesn't leak), but the outer `except` only listed the three git errors, so a re-raised `OSError`/`PermissionError` from the checkout escapes instead of falling through to full initialization — exactly contradicting the "reuse it as-is, fall back if unusable" comment. And it matters most for the target use case here (ephemeral pods over shared bundle storage), where a filesystem hiccup on the on-disk repo is the likely failure.

Fixed in 08b321e by broadening the outer `except` to `Exception`, so any unusable on-disk repo falls back to full initialization as documented. Falling back broadly is safe here because the fallback is the normal init path (a genuine problem would resurface there), and it's logged at info with the exception. `BaseException` (KeyboardInterrupt/SystemExit) is intentionally left to propagate.

Happy to add a regression test that forces a non-git error on the reuse checkout and asserts the fallback, if you'd like it in this PR.

providers/git/src/airflow/providers/git/bundles/git.py
kjh0623: Fixed in 08b321e, and the regression test I offered is now in a479772.

It injects a `PermissionError` when the on-disk tracking repo is opened and asserts initialization falls through to the full path instead of propagating. The assertion is deliberately indirect: it checks the bundle ends up on a commit that only the fallback could have picked up (the fallback fetches, the reuse path does not), so the test still fails if the error is swallowed and stale state is served.

aaron-y-chen (commented):
Hi, thanks for the PR :)

Since the motivation mentions KubernetesExecutor worker pod cold-starts, have you had a chance to test this in a local K8s cluster or a real KubernetesExecutor setup?

kjh0623:
Yeah, I've been running this on our on-prem K8s cluster for a while now — it's actually already deployed in our production Airflow and has been working without issues.

In our setup the DAG processor keeps the tracking repo fresh on the shared volume, so having every worker pod re-fetch on init was pure overhead. With refresh_on_initialize=False that ~15s per-pod cold start on task startup basically goes away, since it just reuses what's already on disk. The clone-if-missing and explicit refresh() paths still behave the same, so we haven't seen any staleness problems in practice.

Happy to share more details about the environment if that'd help.

PR #70059, opened 2026-07-18
Add ADR-0006: no Lang-SDK source display for mixed-language (@task.stub) Dag

### LINKED ISSUE #67260: AIP-85: View foreign language task code
We should provide a way to show both the stub dag _and_ the (e.g. Java) task implementation in Airflow UI. 

First we need to somehow make the Java source available to Airflow with a standard way to access it from the web server (how is this done for Python?)

And then we need to build appropriate UI to display the source.

jason810496:
- closes: #67260

## Why

Users asked whether the Code view for a mixed-language Dag (a Python Dag with `@task.stub` tasks backed by a Java-SDK artifact) could also show the Java-side source. A design sketch extending AIP-85's `DagImporter` interface was drafted.

After discussion with @ashb and @uranusjr, we rejected this feature by design choice so this ADR records that negative decision.

## What

- Add `java-sdk/adr/0006-no-lang-sdk-source-display.md`, documenting the rejected design (per-DAG multi-file `get_source_code`, multiple `DagCode` rows per `DagVersion`, REST/UI file-tab changes) and why the status quo (Python file only) was kept instead.

---

- [x] Yes, with help of Claude Sonnet 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/java-sdk/adr/0006-no-lang-sdk-source-display.md
+++ b/java-sdk/adr/0006-no-lang-sdk-source-display.md
@@ -1,0 +1,140 @@
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ -->
+
+# ADR-0006: No Lang-SDK Source Code Display for Mixed-Language (`@task.stub`) Dags
+
+## Status
+
+Proposed
+
+> **Note:** This ADR records a *negative* decision. It exists so that future feature requests
+> or PRs proposing multi-file / multi-language source display for stub-backed Dags can be
+> pointed here instead of re-litigating the design. Tracked operationally in
+> [apache/airflow#67260](https://github.com/apache/airflow/issues/67260), which is closed as
+> won't-fix referencing this document.
+
+## Context
+
+Per [AIP-108](https://cwiki.apache.org/confluence/x/pY4mGQ), Java tasks are declared as
+`@task.stub` operators in ordinary Python Dag files: the Dag is defined in Python, while the
+task implementation lives in a Java-SDK build artifact (a JAR) shipped in the same bundle. We
+call this a **mixed-language Dag**. This is distinct from pure Java Dags
+([ADR-0003](0003-pure-java-dags.md)), where the Dag itself is defined in Java and the
+coordinator extracts the packed `.java` source via the `Airflow-Java-SDK-Dag-Code` manifest
+attribute for display — a mechanism that was removed from AIP-108 scope along with pure Java
+Dag authoring.
+
+During AIP-108 testing and demos, users asked whether the UI Code view for a
+mixed-language Dag could show the Java-SDK-side source *in addition to* the Python file
+containing the stub declarations
+([apache/airflow#67260](https://github.com/apache/airflow/issues/67260)) — for example, as a
+second tab next to the Python source.
+
+A design sketch was drafted to make this work by extending the
+[AIP-85](https://cwiki.apache.org/confluence/x/_Q7OEg) `DagImporter` interface:
+
+1. A `StubDagImporter` subclassing the Python importer, which after parsing resolves each stub
+   Dag's backing artifact within the bundle and captures its source.
+2. `get_source_code` becoming **per-Dag and multi-file**
+   (`get_source_code(file_path, dag_id, *, bundle_path, bundle_name) -> dict[str, DagSource]`),
+   so sibling Dags defined in one Python file do not leak each other's artifacts into the
+   Code view.
+3. A schema change allowing **multiple `DagCode` rows per `DagVersion`** — dropping the unique
+   constraint on `DagCode.dag_version_id`, adding `source_file_name`, `language`, and an
+   entrypoint marker — with artifact digests participating in `dag_hash` so that a JAR rebuild
+   produces a new `DagVersion`.
+4. REST API and UI changes: `GET /dagSources/{dag_id}` gaining a file dimension, and the Code
+   tab gaining a file tab bar with per-file language instead of the hardcoded Python.
+
+After discussion, we decided not to pursue this.
+
+## Decision
+
+**We will not support viewing the Lang-SDK-side source code for mixed-language Dags.** For a
+Python Dag containing `@task.stub` tasks backed by a Lang-SDK artifact, the Code view shows
+only the Python Dag file — the existing single-source, per-`DagVersion` behavior.
+
+Concretely, none of the following is implemented:
+
+- A `StubDagImporter` (or `CrossLangDagImporter`) that resolves and captures Lang-SDK artifact
+  source for stub-backed Dags.
+- Per-Dag, multi-file semantics for `get_source_code`.
+- Multiple `DagCode` rows per `DagVersion`, or `source_file_name` / `language` / entrypoint
+  columns on `DagCode`.
+- A file dimension on `GET /dagSources/{dag_id}` or a multi-file Code tab in the UI for this
+  use case.
+- Stamping artifact digests onto the serialized Dag model *for the purpose of source display*.
+  (Whether artifact digests should participate in Dag versioning for execution correctness is
+  a separate question, out of scope here.)
+
+### Why Not
+
+1. **A Dag with stub tasks is not a distinct Dag category.** A stub operator is just another
+   operator inside an ordinary Python Dag. A single Dag may soon mix tasks in several
+   languages, so there is no principled, bounded set of "the other side's" files to display.
+   Designing importer interfaces around a special stub-Dag concept bakes an artificial
+   distinction into AIP-85.
+2. **It degenerates into storing a directory tree in `dag_code`.** Showing anything genuinely
+   useful for the Java side means more than one file — Java code in particular tends to split
+   each task into its own small file/class. Followed to its conclusion, this means persisting
+   an entire directory tree in the `dag_code` table, which is not what that table is for.
+3. **The source view is already best-effort in pure Python.** Dags produced by factory
+   functions already display a file that does not contain the "real" definition. Mixed-language
+   Dags are another instance of a known, accepted limitation — not a new gap that must be
+   closed.
+4. **Per-Dag source slicing hides information.** Many existing Dags depend on module-level
+   variables and shared context in their file. Filtering the source view per-Dag to avoid
+   sibling artifact leakage would end up hiding information users actually need; the whole-file
+   view is the safer default.
+5. **The problem is transient and will be rare.** Once Dags can be natively authored in other
+   languages (pure Java Dags per [ADR-0003](0003-pure-java-dags.md) /
+   [ADR-0004](0004-dag-parsing.md), likely after AIP-85 stabilises), authors who care about
+   seeing Lang-SDK code in the UI can write native Lang-SDK Dags, whose entrypoint source is
+   shown as that Dag's source through the normal single-file path. Tying the `DagImporter`
+   interface, the `DagCode` schema, the versioning write path, and the REST/UI contract in
+   knots for a shrinking edge case is a poor trade.
+
+### Alternatives Considered
+
+- **Full proposal (per-Dag multi-file `get_source_code` + multiple `DagCode` rows per
+  `DagVersion` + REST/UI changes).** Rejected: interface and schema complexity, atomic
+  multi-row write semantics, artifact-hash coupling into `dag_hash`, and per-Dag filtering
+  that hides shared file context — all for a rare and transient case.
+- **Show only the Lang-SDK entrypoint file as a second tab.** Rejected: Java code tends to be
+  spread across many small files, so the entrypoint alone has limited value, yet this still
+  requires nearly all of the same schema, API, and UI changes.
+- **Status quo — show the Python file only.** Accepted.
+
+## Consequences
+
+- AIP-85's `DagImporter` interface, the `DagCode` schema (one row per `DagVersion`), the
+  `/dagSources` API contract, and the Code tab UI all stay unchanged.
+- No cross-language source reconciliation, multi-row atomic write semantics, or
+  artifact-hash-driven `DagVersion` churn needs to be built, tested, or maintained for source
+  display.
+- Users of mixed-language Dags see only the Python stub declarations in the Code view and must
+  consult their repository or build artifact for the Java implementation. The Java SDK user
+  documentation states this explicitly so the behavior is expected rather than reported as a
+  bug.
+- [apache/airflow#67260](https://github.com/apache/airflow/issues/67260) is closed as
+  won't-fix with a pointer to this ADR.
+- The source-packing mechanism of [ADR-0003](0003-pure-java-dags.md)
+  (`Airflow-Java-SDK-Dag-Code`) is unaffected: it remains the mechanism for displaying source
+  of *natively authored* Lang-SDK Dags if pure Java Dag authoring is revisited in a follow-up
+  proposal.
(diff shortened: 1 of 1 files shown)


java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+## Consequences
+
+- AIP-85's `DagImporter` interface, the `DagCode` schema (one row per `DagVersion`), the
+  `/dagSources` API contract, and the Code tab UI all stay unchanged.
+- No cross-language source reconciliation, multi-row atomic write semantics, or
+  artifact-hash-driven `DagVersion` churn needs to be built, tested, or maintained for source
+  display.
+- Users of mixed-language Dags see only the Python stub declarations in the Code view and must
+  consult their repository or build artifact for the Java implementation. The Java SDK user
+  documentation states this explicitly so the behavior is expected rather than reported as a
```
shahar1, line 129: Currently it doesn't :) Worth editing it as part of this PR

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+   view is the safer default.
+5. **The problem is transient and will be rare.** Once Dags can be natively authored in other
+   languages (pure Java Dags per [ADR-0003](0003-pure-java-dags.md) /
+   [ADR-0004](0004-dag-parsing.md), likely after AIP-85 stabilises), authors who care about
+   seeing Lang-SDK code in the UI can write native Lang-SDK Dags, whose entrypoint source is
+   shown as that Dag's source through the normal single-file path. Tying the `DagImporter`
+   interface, the `DagCode` schema, the versioning write path, and the REST/UI contract in
+   knots for a shrinking edge case is a poor trade.
+
+### Alternatives Considered
```
shahar1, line 109: The alternatives are very "all or nothing" - maybe we could make it a task-level display at some point?

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+> [apache/airflow#67260](https://github.com/apache/airflow/issues/67260), which is closed as
+> won't-fix referencing this document.
+
+## Context
+
+Per [AIP-108](https://cwiki.apache.org/confluence/x/pY4mGQ), Java tasks are declared as
+`@task.stub` operators in ordinary Python Dag files: the Dag is defined in Python, while the
+task implementation lives in a Java-SDK build artifact (a JAR) shipped in the same bundle. We
+call this a **mixed-language Dag**. This is distinct from pure Java Dags
+([ADR-0003](0003-pure-java-dags.md)), where the Dag itself is defined in Java and the
```
shahar1, line 38: FYI - ADR-0003 is still in "Proposed" state. 
As it seems currently out of scope of AIP-108, I think that it's worth its own discussion + lazy consensus before proceeding with the current ADR.

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Status
+
+Proposed
+
+> **Note:** This ADR records a *negative* decision. It exists so that future feature requests
+> or PRs proposing multi-file / multi-language source display for stub-backed Dags can be
+> pointed here instead of re-litigating the design. Tracked operationally in
+> [apache/airflow#67260](https://github.com/apache/airflow/issues/67260), which is closed as
+> won't-fix referencing this document.
```
shahar1, line 30: The latest comments in the original issue lean towards implementing the feature, so the rejection is not recorded anywhere public (that's why I'm pushing towards a dev list discussion).

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+1. **A Dag with stub tasks is not a distinct Dag category.** A stub operator is just another
+   operator inside an ordinary Python Dag. A single Dag may soon mix tasks in several
+   languages, so there is no principled, bounded set of "the other side's" files to display.
+   Designing importer interfaces around a special stub-Dag concept bakes an artificial
+   distinction into AIP-85.
+2. **It degenerates into storing a directory tree in `dag_code`.** Showing anything genuinely
+   useful for the Java side means more than one file — Java code in particular tends to split
+   each task into its own small file/class. Followed to its conclusion, this means persisting
+   an entire directory tree in the `dag_code` table, which is not what that table is for.
+3. **The source view is already best-effort in pure Python.** Dags produced by factory
```
shahar1, line 97: I don't think the comparison fully holds - for factory Dags, `fileloc` points to where the DAG object was instantiated, so the Code view still shows real Python from the bundle that users can navigate from. 
With `@task.stub` the body is just `def extract(): ...` - there is no reference to the artifact/class anywhere (neither in the source nor in the serialized model), so the UI cannot point to the executed code at all. That is a much bigger gap than the factory caveat.

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+   functions already display a file that does not contain the "real" definition. Mixed-language
+   Dags are another instance of a known, accepted limitation — not a new gap that must be
+   closed.
+4. **Per-Dag source slicing hides information.** Many existing Dags depend on module-level
```
shahar1: This argues against a specific design choice rather than the feature itself. Per-Dag `get_source_code` selects which files are returned, it does not require slicing the Python file's content. Showing the whole Python file + per-Dag artifact tabs would avoid the leakage without hiding anything.
I would drop this point - it weakens the rationale.

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+   variables and shared context in their file. Filtering the source view per-Dag to avoid
+   sibling artifact leakage would end up hiding information users actually need; the whole-file
+   view is the safer default.
+5. **The problem is transient and will be rare.** Once Dags can be natively authored in other
```
shahar1: Pure Java authoring was explicitly descoped from AIP-108 before the vote, so this remedy would require a new proposal.
Meanwhile, mixed-language is currently the *only* shipped mode, so every Java SDK user hits this limitation - I would not call it rare. I would at least define here a trigger for revisiting this decision.

java-sdk/adr/0006-no-lang-sdk-source-display.md
```diff
@@ -0,0 +1,140 @@
...
+  multi-row write semantics, artifact-hash coupling into `dag_hash`, and per-Dag filtering
+  that hides shared file context — all for a rare and transient case.
+- **Show only the Lang-SDK entrypoint file as a second tab.** Rejected: Java code tends to be
+  spread across many small files, so the entrypoint alone has limited value, yet this still
+  requires nearly all of the same schema, API, and UI changes.
+- **Status quo — show the Python file only.** Accepted.
+
+## Consequences
+
+- AIP-85's `DagImporter` interface, the `DagCode` schema (one row per `DagVersion`), the
```
shahar1, line 122: This ADR binds core surfaces (AIP-85's `DagImporter`, `DagCode`, `/dagSources`, the Code tab) but lives in `java-sdk/adr/` - core/AIP-85 contributors will not find it, and the "Lang-SDK" scope also applies to the Go SDK. 
Worth cross-referencing from the AIP-85 page and/or `go-sdk/adr/`, or maybe creating a more "generalized" location for SDKs ADRs (maybe in task-sdk itself?).

shahar1 (dismissed):
Thanks Jason!
I'd like to ask for a public discussion about this decision (as well as its related ADR, ADR-0003) in the dev list.
My main concern is scope: the ADR rejects Lang-SDK source display in general, but the analysis only covers the `DagCode`/multi-file design - so as written it would also foreclose more granular alternatives (e.g., task-level display) that were never evaluated.

uranusjr (approved):

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: This part is basically talking about if we're going to support this feature, the current proposed `DagImporter.get_souce_code(file)` should be pre-Dag granularity `DagImporter.get_souce_code(file, dag_id)`, not slicing the Python file's content. But yes, we should drop this point anyway.

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: I'd like to move the `java-sdk/adr` to somewhere like `airflow-core/adr` after this discussion. 

Actually, the `java-sdk/adr` discuss not just Java-SDK specific, it's talking about the foundation (the coordinator, how should we support native Lang-SDK Dag, etc).

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: Could you deliberate more about the task-level display concept? Thanks.

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: Currently, all ADRs in `java-sdk/adr` are in "Proposed" state, I should fix this. The `pure-java-dags` will be addendum of AIP-108 and the upcoming AIP-85. Anyway, this feature doesn't impact the `pure-java-dags` feature at all. They're independent actually.

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: > so the Code view still shows real Python from the bundle that users can navigate from.

The detail page also tells the bundle and the `fileloc` as well.

https://airflow.apache.org/docs/apache-airflow/stable/ui.html#details-tab

> With @task.stub the body is just def extract(): ... - there is no reference to the artifact/class anywhere 

However, even we support this feature, it only shows the entypoint file of Lang-SDK.

For Java-SDK, it only show something like (IMO, showing the entrypoint factory doesn't help much. If I'm user, I'd like to see each Dag definition, but it's still impossible with the current design mentioned in the ADR that we consider. )

Unless, users themself give up the programming best practice for each Lang-SDK by defining all the Dag and task in the same entrypoint file.

What user will see for Java-SDK code view if we support for mixed lang Dag:
```java
package org.apache.airflow.example;

import java.util.List;
import org.apache.airflow.sdk.*;
import org.jetbrains.annotations.NotNull;

public class ExampleBundleBuilder implements BundleBuilder {
  @NotNull
  @Override
  public Iterable<Dag> getDags() {
    return List.of(
        InterfaceExampleBuilder.build(),
        AnnotationExampleBuilder.build(),
        XComCastingExampleBuilder.build());
  }

  public static void main(String[] args) {
    var bundle = new ExampleBundleBuilder().build();
    Server.create(args).serve(bundle);
  }
}
```

For Go-SDK:
```go
type myBundle struct{}

var _ v1.BundleProvider = (*myBundle)(nil)

func (m *myBundle) GetBundleVersion() v1.BundleInfo {
	return v1.BundleInfo{Name: bundleName, Version: &bundleVersion}
}

func (m *myBundle) RegisterDags(dagbag v1.Registry) error {
	simpleDag := dagbag.AddDag("simple_dag")
	simpleDag.AddTask(extract)
	simpleDag.AddTask(transform)
	simpleDag.AddTask(load)
	return nil
}

func main() {
	if err := bundlev1server.Serve(&myBundle{}); err != nil {
		log.Fatal(err)
	}
}
```

If we're going to show all the `InterfaceExampleBuilder`, `AnnotationExampleBuilder` ..., we will list out the directory tree for each the Lang-SDK involved in the Dag (please don't forget it's valid to have both Python, Go, Java, Ts tasks in the same Dag).

java-sdk/adr/0006-no-lang-sdk-source-display.md
jason810496: Do you mean The Java SDK user documentation states this explicitly so the behavior is expected rather than reported as a bug. part? This is the Consequences, so I will add the doc to mention this after we settle down.

jason810496 (commented):
Hi @shahar1,

From my perspective, the value of showing Lang-SDK entrypoint source is limited (please check the https://github.com/apache/airflow/pull/70059#discussion_r3612734439 comment for the code tab example if we support this feature) so the down side of this feature is more than the value.

I can still raise the Dev List discussion if you feel the ADR here is not enough and we need more stakeholders on this decision.

Thanks.

java-sdk/adr/0006-no-lang-sdk-source-display.md
shahar1: As the coordinator already resolves (dag_id, task_id) into class, maybe we could surface that mapping on the task detail page ("Implementation details": language, artifact name, class name, and maybe even the single class source). Since each task maps to one small class, this stays 1:1 without the need for directory trees or intermediate wrappers. 
It could be discussed in a separate discussion- my main point is not to block a narrow-scoped implementation because of this ADR.

java-sdk/adr/0006-no-lang-sdk-source-display.md
shahar1: Thanks for pointing me to the details tab. I agree that these entrypoint factory doesn't help much - but maybe we could narrow the implementaiton down into the task level (see my comment in the other thread).

shahar1:
> Hi @shahar1,
> 
> From my perspective, the value of showing Lang-SDK entrypoint source is limited (please check the [#70059 (comment)](https://github.com/apache/airflow/pull/70059#discussion_r3612734439) comment for the code tab example if we support this feature) so the down side of this feature is more than the value.
> 
> I can still raise the Dev List discussion if you feel the ADR here is not enough and we need more stakeholders on this decision.
> 
> Thanks.



> Hi @shahar1,
> 
> From my perspective, the value of showing Lang-SDK entrypoint source is limited (please check the [#70059 (comment)](https://github.com/apache/airflow/pull/70059#discussion_r3612734439) comment for the code tab example if we support this feature) so the down side of this feature is more than the value.
> 
> I can still raise the Dev List discussion if you feel the ADR here is not enough and we need more stakeholders on this decision.
> 
> Thanks.



> Hi @shahar1,
> 
> From my perspective, the value of showing Lang-SDK entrypoint source is limited (please check the [#70059 (comment)](https://github.com/apache/airflow/pull/70059#discussion_r3612734439) comment for the code tab example if we support this feature) so the down side of this feature is more than the value.
> 
> I can still raise the Dev List discussion if you feel the ADR here is not enough and we need more stakeholders on this decision.
> 
> Thanks.

I think that it's worth a discussion and lazy-consensus mainly due to the fact that it's a "no" decision which relates all SDKs, that will be hard to challenge in the future - so it should be scoped down exactly to why and what it negates.

ashb:
@shahar1 In general terms, showing task level code from different languages in the UI requires a new AIP -- right now the way dag code is stored doesn't and the UI handles it doesn't allow it, nor does dag processing provide a means to record it, so this counts to me as a Architectural Change -> AIP worthy.

(I agree that it would be nicer, but we don't have the plumbing to set it up)

java-sdk/adr/0006-no-lang-sdk-source-display.md
ashb: This ARD serves as that recording.

java-sdk/adr/0006-no-lang-sdk-source-display.md
ashb: Agreed, this ADR should be part of core, not java SDK

shahar1:
> @shahar1 In general terms, showing task level code from different languages in the UI requires a new AIP -- right now the way dag code is stored doesn't and the UI handles it doesn't allow it, nor does dag processing provide a means to record it, so this counts to me as a Architectural Change -> AIP worthy.
> 
> (I agree that it would be nicer, but we don't have the plumbing to set it up) 

Didn't claim otherwise - it will definitely require its on AIP and design, and they don't have to be included in the scope of the current AIPs.
My main concerns were:

1. In general, hard "no-go" design decisions should be scoped (what + why), and have a discussion and lazy-consensus in the dev list, as they become hard to challenge and revert at later stage by nature.

2. Specifically this ADR, as written by the time I reviewed, might be interpreted as "no-go" for any kind of more granular implementation, which might be a useful feature. I also talked later with Jason in private and he suggested his own idea for implementing it later on. Also, this decision affects other SDKs, so other stakeholders might have their opinions on it.

In any case, considering that there's an agreement that task-level implementation could be considered in later stage with its own AIP - I'm removing the blocker.

jason810496:
Thanks @shahar1 for the discussion! I will raise the follow-up PRs soon for the appropriate Lang-SDK ADRs location, ADR status, limitation section for the Non-Python Dag section that we will only show the Python Stub Operator Dag at this moment.

PR #70066, opened 2026-07-18
Fail queued DagRuns with unresolvable pinned versions

### LINKED ISSUE #70056: Scheduler silently skips queued DagRuns forever when their pinned dag version can no longer be resolved
### Apache Airflow version

### What happened

```
```

The chain:

### How to reproduce

```python

    session.merge(dr)
    session.flush()

    for _ in range(3):
        session.flush()

    dr.bundle_version = None
    session.merge(dr)
    session.flush()

    session.flush()

```

### Operating System

### Deployment

### Anything else

```sql
FROM dag_run
```

### Willing to submit PR?

### Code of Conduct

GayathriSrividya:
Closes #70056

This PR fixes a scheduler bug where a queued DagRun can remain stuck forever when its pinned dag version becomes unresolvable.

Problem
- Some queued DagRuns are created with version pinning metadata.
- If the pinned dag version row is later removed, the run may no longer resolve to a serialized DAG.
- The scheduler previously logged and skipped these runs on every loop, leaving them permanently queued with no terminal outcome.

Fix
- In the queued DagRun start path, when a run has pinned version metadata but the pinned version is unresolvable, the scheduler now marks the DagRun as failed.
- The scheduler also emits a specific state-change notification reason for this condition.
- Existing behavior for other missing-DAG cases is unchanged.

Tests
- Added a regression test covering the unresolvable pinned-version scenario and asserting the DagRun transitions to failed.
- Kept an existing related max_active_runs scheduler test alongside this change for local targeted verification.

Validation run notes
- Ruff format/check passed on changed files.
- Local targeted pytest invocation was attempted multiple ways in this environment.
- One explicit failure captured: ModuleNotFoundError for tests_common during local conftest import.
- No changes were made to unrelated workspace files.

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2768,6 +2768,17 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             backfill_id = dag_run.backfill_id
             dag = dag_run.dag = cached_get_dag(dag_run)
             if not dag:
+                if dag_run.bundle_version is not None and dag_run.created_dag_version_id is None:
+                    self.log.error(
+                        "Queued DagRun '%s' for DAG '%s' has an unresolvable pinned version "
+                        "(bundle_version=%s, created_dag_version_id=None); marking run as failed",
+                        dag_run.run_id,
+                        dag_run.dag_id,
+                        dag_run.bundle_version,
+                    )
+                    dag_run.state = DagRunState.FAILED
+                    dag_run.notify_dagrun_state_changed(msg="unresolvable_pinned_version")
+                    continue
                 self.log.error("DAG '%s' not found in serialized_dag table", dag_run.dag_id)
                 continue
             active_runs = active_runs_of_dags[(dag_id, backfill_id)]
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2768,6 +2768,17 @@ def _update_state(dag: SerializedDAG, dag_run: DagRun):
...
             if not dag:
+                if dag_run.bundle_version is not None and dag_run.created_dag_version_id is None:
+                    self.log.error(
+                        "Queued DagRun '%s' for DAG '%s' has an unresolvable pinned version "
+                        "(bundle_version=%s, created_dag_version_id=None); marking run as failed",
+                        dag_run.run_id,
+                        dag_run.dag_id,
+                        dag_run.bundle_version,
+                    )
+                    dag_run.state = DagRunState.FAILED
```
uranusjr: This should use `set_state()` instead

uranusjr:
I feel #70072 is in a better state.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
GayathriSrividya: Applied, thanks. I switched this branch to use set_state() in the unresolvable pinned-version path.

GayathriSrividya:
> I feel #70072 is in a better state.

Makes sense. I aligned this branch with your feedback (set_state), but I agree #70072 is currently the better thread to move forward with. If you think so, I can close this PR to avoid duplicate review and continue on #70072.

PR #70068, opened 2026-07-18
Fix `@task.llm_branch` import failure on Task SDK-only workers

guan404ming:
## Why

Dags using `@task.llm_branch` failed to load on workers that only have the Task SDK installed.

## How

- Import `determine_kwargs` from `airflow.providers.common.compat.sdk`, matching every sibling decorator in the package

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/decorators/llm_branch.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/decorators/llm_branch.py
@@ -33,10 +33,10 @@ from airflow.providers.common.compat.sdk import (
     DecoratedOperator,
     TaskDecorator,
     context_merge,
+    determine_kwargs,
     task_decorator_factory,
 )
 from airflow.sdk.definitions._internal.types import SET_DURING_EXECUTION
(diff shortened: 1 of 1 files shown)


providers/common/ai/tests/unit/common/ai/decorators/test_llm_branch.py
```diff
@@ -36,6 +39,15 @@ def _make_mock_run_result(output):
     return mock_result
 
 
+def test_module_imports_without_airflow_core(monkeypatch):
+    monkeypatch.setitem(sys.modules, "airflow.utils.operator_helpers", None)
```
gopidesupavan: do we still need this test?

gopidesupavan (approved):
LGTM

gopidesupavan:
i think precommit helpers would be good to block imports

providers/common/ai/tests/unit/common/ai/decorators/test_llm_branch.py
guan404ming: Not necessary. I think pre-commit helpers would be helpful to catch this. Let me remove that.

PR #70070, opened 2026-07-18
Fix HITL review showing altered agent output to reviewers

guan404ming:
## Why

The human review panel could display something different from what the agent actually produced: JSON-looking text was shown as a Python dict and "true" became True, so reviewers approved content that didn't match the task's real output.

## How

- Return XCom values as stored instead of re-parsing them in the HITL review plugin
- Store plugin-written review data natively, matching how workers write XComs
- Seed test XComs through the real worker write path so the mismatch is covered

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/plugins/hitl_review.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/plugins/hitl_review.py
@@ -128,6 +128,7 @@ if AIRFLOW_V_3_1_PLUS:
         session: Session, *, dag_id: str, run_id: str, task_id: str, map_index: int = -1, key: str, value
     ):
         """Write data to db."""
+        # Stores value natively to match worker-written XComs; use XComModel.set(serialize=False) once min Airflow >= 3.2.
         dag_run_id = session.scalar(select(DagRun.id).where(DagRun.dag_id == dag_id, DagRun.run_id == run_id))
         if dag_run_id is None:
             raise HTTPException(404, f"DAG run not found on DAG {dag_id!r} with ID {run_id!r}")
--- a/providers/common/ai/src/airflow/providers/common/ai/plugins/hitl_review.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/plugins/hitl_review.py
@@ -57,15 +57,15 @@ def _get_bundle_url() -> str:
 if AIRFLOW_V_3_1_PLUS:
     import mimetypes
     from pathlib import Path
-    from types import SimpleNamespace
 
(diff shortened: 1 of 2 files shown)


providers/common/ai/src/airflow/providers/common/ai/plugins/hitl_review.py
```diff
@@ -121,25 +121,38 @@ def _read_xcom_by_prefix(
...
-            dag_id=dag_id,
-            task_id=task_id,
-            run_id=run_id,
-            map_index=map_index,
-            session=session,
+        # Stores value natively to match worker-written XComs; use XComModel.set(serialize=False) once min Airflow >= 3.2.
+        dag_run_id = session.scalar(select(DagRun.id).where(DagRun.dag_id == dag_id, DagRun.run_id == run_id))
+        if dag_run_id is None:
+            raise HTTPException(404, f"DAG run not found on DAG {dag_id!r} with ID {run_id!r}")
+        session.execute(
```
kaxil, line 135: We should also wait for @gopidesupavan to review the PR too

providers/common/ai/tests/unit/common/ai/plugins/test_hitl_review.py
```diff
@@ -383,14 +382,14 @@ def test_read_xcom_by_prefix_agent_output_formats(self, session, dag_maker, outp
         dag_maker.create_dagrun(run_id="r", run_type=DagRunType.MANUAL, logical_date=logical_date)
         dag_maker.sync_dagbag_to_db()
 
-        XComModel.set(
+        _write_xcom(
```
kaxil, line 387: Both `_write_xcom` and `_read_xcom` are under test here, so this pins them as a symmetric pair rather than against the actual worker write path -- a symmetric reintroduction of double-encoding in both helpers would still pass. It does catch a read-only regression (seeding natively + the old `deserialize_value` read fails on the JSON-string cases), but asserting the raw stored `XComModel.value` in one case (that it's a native str/dict, not a JSON-encoded string) would independently pin the on-disk format workers produce.

kaxil (approved):

gopidesupavan (approved):
LGTM.. and tested its fine..

providers/common/ai/tests/unit/common/ai/plugins/test_hitl_review.py
guan404ming: Thanks, added a raw XComModel.value assertion. Verified it fails on the pre-fix code, where the stored value is the JSON-encoded string.

guan404ming:
Thanks!

PR #70071, opened 2026-07-18
Fix DocumentLoaderOperator ignoring unknown parser on byte input

guan404ming:
## Why

A typo in the parser name silently produced garbage documents instead of an error when loading from bytes, while the same typo with a file path failed loudly.

## How

- Raise the same "No parser found" error for unknown parser backends on byte input, matching file input
- Add a test covering the typo case

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
@@ -262,6 +262,9 @@ class DocumentLoaderOperator(BaseOperator):
         if backend == "python-docx":
             return self._parse_docx_stream(io.BytesIO(raw))
 
+        if backend not in ("text", "csv", "json"):
+            raise ValueError(f"No parser found for backend '{backend}'.")
+
         text = self._decode(raw, source_hint=f"<bytes:{ext}>")
         if backend == "csv":
             return self._parse_csv_text(text)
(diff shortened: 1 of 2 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
```diff
@@ -262,6 +262,9 @@ def _parse_bytes(self, raw: bytes, file_type: str) -> list[dict[str, Any]]:
         if backend == "python-docx":
             return self._parse_docx_stream(io.BytesIO(raw))
 
+        if backend not in ("text", "csv", "json"):
```
gopidesupavan: would it be simple if we throw error inside _resolve_backend itself?

guan404ming: Good point, done. Moved the check into `_resolve_backend` and dropped the duplicate in _parse_bytes.

Lee-W (approved):

gopidesupavan (approved):

providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
gopidesupavan: thanks yeah

guan404ming:
Thanks!

PR #70073, opened 2026-07-18
Enhancing docs for `HttpEventTrigger`

jroachgolf84:
## Description

This PR enhances docs for the `HttpEventTrigger`. Previously, the example code snippet was a bit difficult to digest due to spacing and a lack of comments.

There was also an issue with the `get_variable_sync` and `set_variable_sync` function calls. Really, they should have been called `get_variable_async` and `set_variable_async`.

As part of this effort, additional enhancements/cleanup steps were made (for verbosity, accuracy, and consistency).

### DIFF
--- a/providers/http/docs/triggers.rst
+++ b/providers/http/docs/triggers.rst
@@ -22,7 +22,7 @@ HTTP Event Trigger
 .. _howto/trigger:HttpEventTrigger:
 
 The ``HttpEventTrigger`` is an event-based trigger that monitors whether responses from an API meet the conditions set
-by the user in the ``response_check`` callable. If the condition is met, then the Trigger "fires". Otherwise, it waits
+by the user in the ``response_check_path`` callable. If the condition is met, then the trigger "fires". Otherwise, it waits
 a certain amount of time before repeating the request and check.
 
 The ``HttpEventTrigger`` is designed for **Airflow 3.0+** to be used in combination with the ``AssetWatcher`` system,
@@ -69,7 +69,7 @@ Here's an example of using the ``HttpEventTrigger`` in an ``AssetWatcher`` to mo
         # Convert Variable.get to be asynchronous to be used on the Triggerer
         get_variable_async = sync_to_async(Variable.get)
 
-        # Retrieve the previous and current release ID's (one from Variables, one from response)
+        # Retrieve the previous and current release IDs (one from Variables, one from response)
         previous_release_id = await get_variable_async(key="release_id_var", default=None)
         data = response.json()
         release_id = str(data["id"])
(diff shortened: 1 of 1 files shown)


providers/http/docs/triggers.rst
```diff
@@ -64,18 +65,27 @@ Here's an example of using the HttpEventTrigger in an AssetWatcher to monitor th
 
 
     async def check_github_api_response(response):
+        """Determine if a new version of Airflow has been released."""
+        # Convert Variable.get to be asynchronous to be used on the Triggerer
+        get_variable_async = sync_to_async(Variable.get)
+
+        # Retrieve the previous and current release ID's (one from Variables, one from response)
```
uranusjr:
```suggestion
        # Retrieve the previous and current release IDs (one from Variables, one from response)
```

providers/http/docs/triggers.rst
```diff
@@ -110,32 +121,31 @@ Parameters
 ----------
 
 ``http_conn_id``
-    http connection id that has the base API url i.e https://www.google.com/ and optional authentication credentials.
-    Default headers can also be specified in the Extra field in json format.
+    HTTP Connection ID that has the base API url i.e https://www.google.com/ and optional authentication credentials. Default headers can also be specified in the ``extra`` field in JSON format.
```
uranusjr:
```suggestion
    HTTP Connection ID that has the base API URL, e.g. ``https://www.google.com/``, and optional authentication credentials. Default headers can also be specified in the ``extra`` field in JSON format.
```

providers/http/docs/triggers.rst
```diff
@@ -21,10 +21,11 @@ HTTP Event Trigger
 
 .. _howto/trigger:HttpEventTrigger:
 
-The ``HttpEventTrigger`` is an event-based trigger that monitors whether responses
-from an API meet the conditions set by the user in the ``response_check`` callable.
+The ``HttpEventTrigger`` is an event-based trigger that monitors whether responses from an API meet the conditions set
+by the user in the ``response_check`` callable. If the condition is met, then the Trigger "fires". Otherwise, it waits
```
uranusjr:
```suggestion
by the user in the ``response_check_path`` callable. If the condition is met, then the trigger "fires". Otherwise, it waits
```

I think this has always been wrong? Let’s fix it while we’re changing the text anyway.

jroachgolf84: Thanks, good catch. Will apply in batch.

uranusjr (approved):

jroachgolf84:
@uranusjr - do you mind merging? Checks passed

henry3260 (approved):
Thanks!

PR #70075, opened 2026-07-18
Preserve output_type through human approval in LLM operators

guan404ming:
## Why

Turning on human approval changed a task's result: a task declared to return a number or list came back as a plain string after the reviewer approved it.

## How

- Encode non-string outputs as JSON when deferring for review, instead of Python's str()
- Validate the approved (possibly reviewer-edited) string back into the declared output_type
- Fall back to the raw string when the edited output no longer matches the type

---

### DIFF
--- a/providers/common/ai/tests/unit/common/ai/operators/test_llm.py
+++ b/providers/common/ai/tests/unit/common/ai/operators/test_llm.py
@@ -379,6 +379,21 @@ class TestLLMOperatorApproval:
         assert isinstance(result, Summary)
         assert result.text == "hello"
 
+    @pytest.mark.parametrize(
+        ("output_type", "generated_output", "expected"),
+        [(int, "5", 5), (list[str], '["a","b"]', ["a", "b"])],
+        ids=["int", "list"],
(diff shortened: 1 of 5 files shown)


providers/common/ai/tests/unit/common/ai/operators/test_llm.py
```diff
@@ -379,6 +379,21 @@ def test_execute_complete_rehydrates_pydantic_for_structured_output(self):
         assert isinstance(result, Summary)
         assert result.text == "hello"
 
+    @pytest.mark.parametrize(
+        ("output_type", "generated_output", "expected"),
```
gopidesupavan, line 383: can we include basemodel type aswell and verify it?

guan404ming: Make sense, just added. Thanks!

gopidesupavan (approved):

Lee-W (approved):

guan404ming:
Thanks!

PR #66854, opened 2026-05-13
Reduce task success asset registration lock contention

hkc-8010:
Closes #66853

## Summary

Under high fan-out (for example a dbt/Cosmos DAG where many tasks emit assets at once), `ti_update_state` committed the task-instance state and then registered asset events while still holding the `task_instance` row lock. Holding that lock across the registration work (scanning scheduled DAGs and inserting `AssetEvent`/`AssetDagRunQueue` rows) piled up concurrent completions and OOMKilled the API server.

This reworks the task-success path so asset registration never holds the `task_instance` lock and a failed registration is never lost:

- Commit the task-instance state first, releasing the row lock.
- Then register asset events opportunistically, under a short lock timeout. Uncontended writes (the common case) still happen inline and are unaffected by the timeout.
- If the write cannot acquire its locks in time, or fails for any reason, the events are written to a new `asset_event_queue` table. The scheduler drains that queue and retries until the events are registered, so nothing is silently dropped.
- Replace the alias-event ORM `append` with a direct association-table insert, so a long-lived alias does not lazy-load its entire `asset_events` collection to add one row.

Asset-event creation is asynchronous when it cannot complete inline: there is a short, eventually consistent delay (usually one scheduler loop) before a deferred event and its downstream asset-triggered DAG run appear. This is documented and called out in a significant-change newsfragment. It replaces the earlier version of this PR, which returned success and only logged a dropped registration; the queue makes delivery at-least-once instead.

## Changes

- `airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py`: commit state before registration; two-phase register-or-enqueue.
- `airflow-core/src/airflow/models/asset.py` and migration `0127_3_4_0_add_asset_event_queue.py`: new `asset_event_queue` table.
- `airflow-core/src/airflow/jobs/scheduler_job_runner.py`: `_drain_asset_event_queue` sweep (claims with `SKIP LOCKED`, retries until success, parks a row after `asset_event_queue_max_attempts`).
- `airflow-core/src/airflow/assets/manager.py`: direct alias-event association insert.
- `airflow-core/src/airflow/config_templates/config.yml`: `[scheduler]` `asset_event_queue_drain_interval`, `asset_event_queue_batch_size`, `asset_event_queue_max_attempts`.
- Metrics `asset.event_queue.enqueued` / `processed` / `failures` / `pending`; docs note in `asset-scheduling.rst`; significant newsfragment.

## Validation

[...]

Issue #66853, opened 2026-05-13
API server OOMKill: task_instance row lock held during asset event emission under high concurrency


hkc-8010:
## Problem

Under high concurrency (80+ simultaneous task completions emitting asset events), the Airflow 3 API server dies with OOMKill. The root cause is a DB lock contention chain:

1. `ti_update_state()` acquires `SELECT task_instance ... WITH FOR UPDATE`, holding a PostgreSQL row lock.
2. While holding that lock, `register_asset_changes_in_db()` calls `AssetManager.register_asset_change()`, which runs multiple slow queries including `asset_alias_model.asset_events.append(asset_event)`. This ORM .append() lazy-loads the **entire** `asset_events` collection for the alias -- potentially thousands of rows on long-running deployments.
3. Each slow query leaves the connection `idle in transaction` while Python processes results. New workers needing `SELECT task_instance FOR UPDATE` on the same row queue up, each holding a FastAPI threadpool thread via Cadwyn's `run_in_threadpool`.
4. With 80+ concurrent completions, thread count grows unbounded until OOMKill.

### DB evidence (confirmed from production, May 2026)

```
pid 31104: idle in transaction, xact_age=3:35
  last query: SELECT asset_alias WHERE name IN ('fivetran-synced-table')
pid 31303: idle in transaction, xact_age=3:31, same pattern
-- both blocking: SELECT task_instance FOR UPDATE WHERE id = UUID(...)
```

Disabling the trigger DAGs dropped apiserver memory from 5Gi+ to MBs instantly. Re-enabling just one DAG reproduced 76 `QueryCanceled: canceling statement due to statement timeout` errors in 1.5 hours.

Blocking query chain confirmed from `pg_stat_activity`:

```sql
pid 31303: idle in transaction
  INSERT INTO asset_alias_asset_event VALUES (2, 1151718)
pid 31104: idle in transaction
  SELECT asset_alias ... WHERE name IN ('fivetran-synced-table')
-- both blocking 8 workers doing:
  SELECT task_instance ... FOR UPDATE WHERE id = UUID(...)
```

## Fix

Two changes:

**1. `AssetManager.register_asset_change()` (`assets/manager.py`)**: Replace `asset_alias_model.asset_events.append(asset_event)` + `session.add(asset_alias_model)` with a direct `INSERT INTO asset_alias_asset_event (alias_id, event_id)`. This eliminates the lazy-load of the existing events collection (which can be thousands of rows) while the task_instance row lock is held.

[...]

PR #70078, opened 2026-07-18
Register asset events before taking the task_instance row lock

Andrushika:
Register asset events before taking the task_instance row lock
---
## Why

When a task succeeds, `ti_update_state` currently takes the `task_instance` row lock and then registers asset events while holding the lock.

Asset registration can be slow because it performs asset lookups, event inserts, and Dag-run queueing. When many tasks finish at the same time, concurrent requests can wait for the row lock while each request continues to use an API-server thread. In the case reported in #66853, this caused the API server to run out of memory and get OOMKilled.

## What

This change registers the asset events before taking the `task_instance` row lock:
image
The asset events and the task state are still written in the same database transaction.

In this PR, we first read the state without the row lock, then perform the asset registration. This can hit a TOCTOU problem: the state may change in between, for example, when the worker retries. So before we actually change the state, we take the row lock and read the state one more time. The lock makes sure no one else can change it after this point. Then we decide what to do.

There are three cases:

1. The state is still `RUNNING`: the normal path. We change the state to `SUCCESS` and commit. The state and the events are saved together in one transaction.

2. The state is already `SUCCESS`: another request won the race and already saved its own events. We roll back so the events are not saved twice, and return 200.

3. The state is something else, such as `FAILED`: the task is no longer running, so the update is rejected with 409 and the transaction is rolled back. No events from this request are saved.

This locked re-check keeps the task state and asset events atomic: the events are committed only together with the `SUCCESS` state.

## Benchmark
The `task_instance` row lock is held during one successful task completion, on Postgres. The task emits to N assets to make the registration slow. Median of 14 runs (lock holding time):
outlets | before | after
-- | -- | --
300 | 426 ms | 2.0 ms
600 | 974 ms | 2.1 ms

Before, the lock is held for the whole registration, so it grows with the number of outlets. After, the registration runs before the lock, so the lock is held only for the state update and stays flat.

This is an alternative to #66854.

related: #66853

---

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -358,6 +358,33 @@ def ti_update_state(
     bind_contextvars(ti_id=str(task_instance_id))
     log.debug("Updating task instance state", new_state=ti_patch_payload.state)
 
+    # For a success payload, register outlet asset events *before* acquiring the task_instance
+    # row lock below, so the registration work (asset lookups, event inserts, dag-run queueing)
+    # never runs while holding that lock and concurrent completions do not pile up behind it.
+    # The locked SELECT below re-checks the state: if the transaction does not commit (duplicate
+    # or invalid transition), these writes roll back with it, so the registered events are only
+    # ever observable together with the committed SUCCESS state.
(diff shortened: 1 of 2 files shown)


Andrushika:
Close this since #70951 is merged.

PR #70081, opened 2026-07-18
Refactor xcom jwt url

### LINKED ISSUE #70080: Remove dag_id/run_id/task_id from XCom URL in favor of JWT token
### Description

Currently, the `/execution/xcoms/{dag_id}/{run_id}/{task_id}/{key}` endpoint in the Execution API accepts the DAG ID, Run ID, and Task ID as URL parameters. 

However, since the Task SDK now authenticates with the Execution API using a short-lived JWT token that is inherently scoped to the specific task instance, passing these identifiers in the URL is redundant and bypasses the zero-trust model. 

An existing `TODO` in `xcoms.py` notes that we should remove these parameters from the URL and rely entirely on the information embedded in the JWT token (or looked up via the token ID).

### Use case/motivation

- **Security**: Enforces that the worker can only set XComs for the exact task instance it is authorized for (zero-trust model).
- **Simplicity**: Cleans up the URL structure and routing logic.
- **Payload/Path Reduction**: Removes redundant path variables that the worker shouldn't have to explicitly specify anyway.

### Related issues

_No response_

### Code of Conduct

KushagraB424:
### Description

The Task Execution API's `set_xcom` endpoint implicitly has access to the current task instance identifiers through the injected JWT token. Removing the task identifiers (`dag_id`, `run_id`, `task_id`) from the URL simplifies routing, prevents clients from providing conflicting or invalid identifiers, and enforces zero-trust security where a task instance can only write its own XComs.

This refactor:
- Updates the `has_xcom_access` dependency to handle write requests gracefully without requiring `dag_id` in the URL path.
- Modifies the `set_xcom` endpoint to query identifiers securely from the database via the token.
- Cleans up `client.xcoms.set` in the Task SDK to use the simplified URL format.

closes: #70080

---

Generated-by: [Gemini] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
@@ -27,22 +27,20 @@ from sqlalchemy.sql.selectable import Select
 
 from airflow.api_fastapi.common.db.common import SessionDep
 from airflow.api_fastapi.core_api.base import BaseModel
+from airflow.api_fastapi.execution_api.datamodels.token import TIToken
 from airflow.api_fastapi.execution_api.datamodels.xcom import (
     XComResponse,
     XComSequenceIndexResponse,
     XComSequenceSliceResponse,
 )
 from airflow.api_fastapi.execution_api.security import CurrentTIToken
+from airflow.models.taskinstance import TaskInstance
 from airflow.models.taskmap import TaskMap
 from airflow.models.xcom import XComModel
 from airflow.utils.db import get_query_count
 
 
 def has_xcom_access(
-    dag_id: str,
-    run_id: str,
-    task_id: str,
-    xcom_key: Annotated[str, Path(alias="key", min_length=1)],
     request: Request,
     session: SessionDep,
     token=CurrentTIToken,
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -27,22 +27,20 @@
...
 from airflow.models.taskmap import TaskMap
 from airflow.models.xcom import XComModel
 from airflow.utils.db import get_query_count
 
 
 def has_xcom_access(
-    dag_id: str,
-    run_id: str,
-    task_id: str,
-    xcom_key: Annotated[str, Path(alias="key", min_length=1)],
```
SameerMesiah97, line 45: Changing the signature of a public API by removing required parameters breaks backwards compatibility. If you intend to deprecate them, it would better to turn these into optional parameters (with default arguments of `None`).

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -359,15 +364,13 @@ def get_xcom(
 # TODO: once we have JWT tokens, then remove dag_id/run_id/task_id from the URL and just use the info in
 # the token
 @router.post(
-    "/{dag_id}/{run_id}/{task_id}/{key:path}",
+    "/{key:path}",
```
SameerMesiah97, line 367: I think it would be better to keep the existing route for backwards compatibility and add the new route with only `/{key:path}`.

SameerMesiah97 (commented):
I understand you are following an existing `#TODO`, but this change is not backwards compatible as it removes required path parameters and changes the existing API route without retaining the previous endpoint. Could we preserve the current route for compatibility (perhaps deriving the authoritative task identifiers from the JWT and ignoring or validating the supplied values), while introducing the simplified JWT-only route? That would achieve the security goal without breaking existing clients.

The CI failures for this PR are a symptom of a breaking change.

kaxil:
Please follow better guidelines for PR title !

PR #70085, opened 2026-07-19
Buffer TS SDK logger output until the log socket connects

### LINKED ISSUE #70077: TS SDK: buffer pre-connect logger output and flush to log socket once connected
Follow-up to #69302 (see https://github.com/apache/airflow/pull/69302#discussion_r3608363125 discussion).

#69302 sends TS SDK logs to stderr when the log socket disconnects after a fatal error, but there's a separate gap in the subprocess startup lifecycle:

1. TS subprocess started (via `popen` on the coordinator side)
2. TS logger can still emit records here, before it has connected to the coordinator's `--log` socket
3. TS subprocess connects to the coordinator
4. All logger output is forwarded over the `--log` socket

Stage 2 output currently has nowhere reliable to go, since a `LogChannel` only exists post-connect. Java and Go SDKs already buffer logs emitted in this window and flush them to the log socket once connected — TS SDK should do the same instead of dropping them.

Reference: diagram in #70076 may help clarify the lifecycle.

guan404ming:
## Related

closes: #70077
related: #69302 (stacked on its branch; only the last commit is new)

## Why

Log records emitted between subprocess start and the log socket connecting were silently dropped; the Go and Java SDKs already keep these early records.

## How

- Let `LogChannel` start unconnected and buffer records until `connect()` flushes them to the socket in order
- Dump buffered records to stderr on `close()` if the socket never connected
- Create the runtime's logger before connecting so pre-connect records are kept

---

### DIFF
--- a/ts-sdk/src/coordinator/log-channel.ts
+++ b/ts-sdk/src/coordinator/log-channel.ts
@@ -73,15 +73,15 @@ export class LogChannel {
     return new LogChannel({ sock: null, connected: false, closed: false, buffer: [] }, name, true);
   }
 
-  static async connect(addr: string, name: string = DEFAULT_LOGGER_NAME): Promise<LogChannel> {
-    const channel = LogChannel.createBuffered(name);
-    await channel.connect(addr);
-    return channel;
-  }
-
-  /** Connect the socket and flush buffered records, in order. Root only. */
+  /** Connect the socket and flush buffered records, in order. Root only,
+   *  and callable once: a second call would orphan the first socket. */
   async connect(addr: string): Promise<void> {
-    if (!this.isRoot) return;
+    if (!this.isRoot) {
+      throw new Error(`[${this.name}] connect() is root-only`);
+    }
+    if (this.shared.sock) {
+      throw new Error(`[${this.name}] connect() called more than once`);
+    }
     const shared = this.shared;
     const name = this.name;
     const sock = await connectTcp(addr);
--- a/ts-sdk/src/coordinator/log-channel.ts
+++ b/ts-sdk/src/coordinator/log-channel.ts
@@ -50,9 +50,10 @@ const DEFAULT_LOGGER_NAME = "ts-sdk";
 const CLOSE_FLUSH_TIMEOUT_MS = 3_000;
 
 interface LogChannelState {
-  sock: Socket;
+  sock: Socket | null;
   connected: boolean;
   closed: boolean;
+  buffer: string[];
 }
 
 export class LogChannel {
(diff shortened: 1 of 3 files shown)


ts-sdk/src/coordinator/log-channel.ts
```diff
@@ -66,22 +67,39 @@ export class LogChannel {
...
+  }
+
+  /** Connect the socket and flush buffered records, in order. Root only. */
+  async connect(addr: string): Promise<void> {
+    if (!this.isRoot) return;
+    const shared = this.shared;
+    const name = this.name;
+    const sock = await connectTcp(addr);
+    shared.sock = sock;
+    shared.connected = true;
```
jason810496, line 89: `connect()` has no state guards, which leaves two latent holes (neither reachable from `runtime.ts` today, but this is now the class's main entry point):

1. **Root-only no-op is silent**: `await child.connect(addr)` resolves successfully while connecting nothing — the socket stays `null` and every record buffers indefinitely. The doc comment says "Root only"; throwing encodes that loudly. (The same guard in `close()` is fine — children legitimately don't own the socket.)
2. **No `closed`/double-connect check**: a `close()` that runs while `connectTcp` is in flight dumps the buffer and returns (sock is still `null`), then `connect()` resolves and adopts a socket nobody will ever close — it keeps the event loop alive. A second `connect()` call would also orphan the first socket, whose stale `close` handler later flips `shared.connected = false` and diverts all logs from the healthy socket to stderr.

```suggestion
  async connect(addr: string): Promise<void> {
    if (!this.isRoot) {
      throw new Error(`[${this.name}] connect() is root-only`);
    }
    if (this.shared.sock) {
      throw new Error(`[${this.name}] connect() called more than once`);
    }
    const shared = this.shared;
    const name = this.name;
    const sock = await connectTcp(addr);
    if (shared.closed) {
      // close() won the race while we were dialing; it already dumped the
      // buffered records to stderr, so just discard the socket.
      sock.destroy();
      return;
    }
    shared.sock = sock;
    shared.connected = true;
```

ts-sdk/src/coordinator/log-channel.ts
```diff
@@ -66,22 +67,39 @@ export class LogChannel {
     this.isRoot = isRoot;
   }
 
+  /** Create a root channel with no socket yet; records are buffered
+   *  until {@link LogChannel#connect} flushes them. */
+  static createBuffered(name: string = DEFAULT_LOGGER_NAME): LogChannel {
+    return new LogChannel({ sock: null, connected: false, closed: false, buffer: [] }, name, true);
+  }
+
   static async connect(addr: string, name: string = DEFAULT_LOGGER_NAME): Promise<LogChannel> {
```
jason810496: I lean toward dropping the `LogChannel.connect()`. WDYT?

---

After this change the static `LogChannel.connect()` has no production callers — `runtime.ts` uses `createBuffered()` + instance `connect()` — and it now shares the name `connect` with an instance method of a different shape (factory returning a channel vs. mutator). Consider either dropping it and having tests use the same two-step API production uses, or noting in its doc comment that it's a test convenience. No suggestion block since removing it touches ~10 test call sites.

jason810496 (commented):
Here're some nits caught by CC. LGTM overall, thanks!

ts-sdk/src/coordinator/log-channel.ts
guan404ming: Make sense to me, just dropped it. I think the tests could use a local `connectChannel()` helper instead.

ts-sdk/src/coordinator/log-channel.ts
guan404ming: Good catch, agreed on both. Added the guards. Skipped only the `shared.closed` race branch: `close()` runs in `runtime.ts`'s `finally`, after `connect()` settles, so it can't land mid-dial.

jason810496 (approved):
Nice! Thank you for addressing all the comments.

PR #66132, opened 2026-04-29
Don't re-emit logical_date when previous data_interval is zero-length

jedcunningham:
## What's the bug

`_DataIntervalTimetable.next_dagrun_info` aligns the next run's `start` onto `data_interval.end` of the previous run. When that previous interval is zero-length (`start == end`), `start` lands on the *same* point as the existing run's `logical_date`, so the scheduler tries to create a run that collides with the existing one and logs:

```
run already exists; skipping dagrun creation
```

It then calls `calculate_dagrun_date_fields(last_automated_run=existing_run)` to advance — which feeds the same zero-length interval back in and recomputes the same value. The scheduler loops on the skip warning until the next cron boundary nudges `_skip_to_latest` to a different result, at which point the gate may have already rolled past the slot — so the intended run is silently skipped.

## When does this happen?

Zero-length `data_interval`s are the normal output of `CronTriggerTimetable` (point-in-time) but never of `CronDataIntervalTimetable` (windowed, validated `> 0`). They appear on a `CronDataIntervalTimetable` DAG when its scheduling history was produced by `CronTriggerTimetable` and the DAG later switches to `CronDataIntervalTimetable` — for example, when `[scheduler] create_cron_data_intervals` is flipped from `False` (Airflow 3 default) to `True` after some runs have already been created.

## The fix

In `_DataIntervalTimetable.next_dagrun_info`, after computing `start`, if the previous interval was zero-length and `start` falls on it, advance `start` by one period so the new run gets a fresh `logical_date`.

The guard only fires for `_DataIntervalTimetable` subclasses (`CronDataIntervalTimetable`, `DeltaDataIntervalTimetable`). `CronTriggerTimetable` / `DeltaTriggerTimetable` extend a separate `_TriggerTimetable` base with its own `next_dagrun_info` and never go through this code path, so their normal zero-length output is unaffected.

## Tests

Added `test_zero_length_last_interval_does_not_re_emit_logical_date` parametrized on `catchup ∈ {True, False}` — both branches converge on the same expected interval, demonstrating the guard absorbs the difference between the two paths into the previous-run handling.

related: #59618

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Anthropic)

PR #70089, opened 2026-07-19
Fix manual runs getting a zero-length data interval across DST

harjothkhara:
## Why

On the two days a year the clocks change, a manually triggered run can get a data interval that is zero length, or that ends after the run time.

With `CronDataIntervalTimetable("0 2 * * *", timezone="America/New_York")`, a manual run at `2024-03-10 07:30Z`:

```
manual    : 2024-03-10 07:00Z -> 2024-03-10 07:00Z   (zero length)
scheduled : 2024-03-09 07:00Z -> 2024-03-10 07:00Z   (correct)
```

croniter steps back on naive wall-clock time, but `make_aware` moves a tick that falls in the spring-forward gap onto the next real instant. When the previous tick is in the gap it lands on the current time, so `_get_prev` returns the time it was asked about.

That zero-length interval then costs a run. `next_dagrun_info` reads it as a point-in-time run and skips a period (#66132, correct for the case it was written for):

```
after the bad interval : 2024-03-11 06:00Z -> 2024-03-12 06:00Z
if the interval is ok  : 2024-03-10 07:00Z -> 2024-03-11 06:00Z
```

The fall-back side overshoots instead. With `0 1 * * *` in `America/New_York` at `2024-11-03 05:30Z`, `_get_prev` returns `06:00Z` — half an hour in the future — so `CronTriggerTimetable` with `run_immediately=True` fires a tick that has not happened yet.

Also reproduces in `Europe/Zurich`, `Australia/Sydney` and `Australia/Lord_Howe`.

## What

`_get_prev` now keeps stepping `cron.get_prev` until the result is strictly earlier than the time asked about. `get_prev` is strictly decreasing on naive time, so it terminates.

The `_covers_every_hour` branch is untouched — it uses a wall-clock delta rather than `make_aware`, so it cannot hit this.

Only values that were already wrong change. Nothing outside a DST transition window is affected.

## Testing

Three regression tests, all failing before and passing after:

- `test_cron_infer_manual_data_interval_dst` — gap and fall-back, including `Australia/Lord_Howe` for its 30-minute offset
- `test_cron_next_dagrun_info_after_dst_manual_run` — the skipped period
- `test_cron_trigger_run_immediately_does_not_pick_future_run` — `CronTriggerTimetable`, which shares the mixin

`airflow-core/tests/unit/timetables/` is 246 passed.

---

Generated-by: Claude Opus 4.8 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


airflow-core/src/airflow/timetables/_cron.py
```diff
@@ -166,14 +166,21 @@ def _get_next(self, current: DateTime) -> DateTime:
...
         scheduled = cron.get_prev(datetime.datetime)
         if TYPE_CHECKING:
             assert isinstance(scheduled, datetime.datetime)
         if not _covers_every_hour(cron):
-            return convert_to_utc(make_aware(scheduled, self._timezone))
+            prev = convert_to_utc(make_aware(scheduled, self._timezone))
+            # croniter steps back on naive wall clock, but make_aware can map a tick inside
+            # a DST transition forward onto current or later. Keep stepping until strictly
+            # earlier; get_prev is strictly decreasing, so this terminates.
+            while prev >= current:
```
uranusjr, line 180: Please add a parametrized case (mirroring `test_cron_infer_manual_data_interval_dst`) that drives `next_dagrun_info` through `_skip_to_latest` across a DST boundary, so this behavior doesn't silently regress if `_get_prev` changes again.

harjothkhara: Added a parametrized `test_cron_next_dagrun_info_no_catchup_dst` mirroring `test_cron_infer_manual_data_interval_dst`, driving `next_dagrun_info` through `_skip_to_latest` with `catchup=False`. All cases fail without the `_get_prev` change.

---
Drafted-by: Claude Code (Opus 5); reviewed by @harjothkhara before posting

airflow-core/newsfragments/70089.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix cron timetables inferring a data interval that is not strictly earlier than the run time on days a DST transition occurs. A manual run whose cron tick fell in a spring-forward gap got a zero-length interval, which also caused ``next_dagrun_info`` to skip the following scheduled period; on the fall-back side the inferred interval could end after ``run_after``, and ``CronTriggerTimetable`` with ``run_immediately`` could select a tick that had not happened yet.
```
uranusjr, line 1: This is a bit long; suggest trimming to about 100-150 characters.

uranusjr (approved):

airflow-core/tests/unit/timetables/test_interval_timetable.py
```diff
@@ -383,6 +383,173 @@ def test_cron_next_dagrun_info_alignment(last_data_interval: DataInterval, expec
     assert info == expected_info
 
 
+@pytest.mark.parametrize(
+    ("timetable", "run_after", "expected_interval"),
+    [
+        pytest.param(
+            CronDataIntervalTimetable("0 2 * * *", pendulum.timezone("America/New_York")),
```
Lee-W, line 390: Could we also add test cases to cover `MultipleCronTriggerTimetable` and `CronPartitionTimetable`

Lee-W (commented):
one nit improvement. overall looks good

uranusjr (approved):

airflow-core/tests/unit/timetables/test_trigger_timetable.py
```diff
@@ -1149,3 +1149,52 @@ def test_iter_partition_dagrun_infos_dst_america_new_york_spring_forward() -> No
     # run_after == partition_date for both ticks.
     for info in infos:
         assert info.run_after == info.partition_date
+
+
+@time_machine.travel(pendulum.DateTime(2024, 11, 3, 5, 30, tzinfo=utc), tick=False)
```
Lee-W: we can simplify these 3 tests through parameterization

Lee-W (commented):
I'm good after parameterization is done. Thanks!

airflow-core/tests/unit/timetables/test_trigger_timetable.py
harjothkhara: Merged the 3 tests into one parametrized `test_run_immediately_does_not_pick_future_run`, same DST fold case for each timetable.

---
Drafted-by: Claude Code (Opus 5); reviewed by @harjothkhara before posting

Lee-W (approved):

PR #70091, opened 2026-07-19
quote remote job paths in posix ssh command builders

Samin061:
The POSIX builders in `remote_job.py` interpolate `RemoteJobPaths` values into single-quoted shell words without escaping, so a single quote anywhere in `remote_base_dir` closes the quote and the rest of the value is parsed as commands that run on the remote host as the SSH connection user. `remote_base_dir` is a template field, but `_validate_base_dir` runs in `__init__` against the un-rendered Jinja literal and never looks for shell metacharacters, so a value like `{{ dag_run.conf['dir'] }}` reaches the shell unchecked and someone who can only trigger a Dag run with config gets a shell on the target host. Quoting with `shlex.quote` at each site matches what the Windows builders already do via `ps_escape`.

All six POSIX builders are affected. The `_validate_job_dir` prefix check on cleanup does not help, since a base dir that starts with `/tmp/airflow-ssh-jobs/` and then contains a quote still passes it:

```
base_dir = "/tmp/airflow-ssh-jobs/x'; touch /tmp/pwned; :'"
build_posix_cleanup_command(paths.job_dir)
# rm -rf '/tmp/airflow-ssh-jobs/x'; touch /tmp/pwned; :'/job_123'
```

Added a parametrized regression that runs each builder's output through `sh` and asserts the injected marker is never created; it fails on all six before this change.

---

### DIFF
--- a/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 import base64
 import re
 import secrets
+import shlex
 import string
 from dataclasses import dataclass
 from typing import Literal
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Thanks — this is a proper hardening fix and the test is the best part of it.

The bug is real: `job_dir='{paths.job_dir}'` closes and reopens on any single quote in the path, so a `remote_base_dir` containing `'` stops being data and becomes shell syntax. `shlex.quote()` is the correct tool, applied consistently across all six POSIX builders.

**The test actually proves it.** Rather than asserting on generated string content, `TestPosixPathQuoting` builds a real payload — `/tmp/airflow-ssh-jobs/x'; touch {marker}; :'` — **executes the generated command through `sh -c`**, and asserts the marker file was never created, parametrised across all six builders. That's an executable proof of non-injection rather than a proxy for one, and it would catch a future builder that forgets to quote. Updating `cat '/tmp/pid'` -> `cat /tmp/pid` in the existing assertion is right too, since `shlex.quote` leaves simple paths bare.

Two notes, neither blocking:

**The Windows builders are untouched** — `build_windows_log_tail_command`, `build_windows_file_size_command`, `build_windows_completion_check_command` and `build_windows_cleanup_command` still interpolate directly. That matches your title's scoping, and `shlex.quote` would be actively wrong there (it emits POSIX quoting, not `cmd.exe`/PowerShell), so this isn't something to bolt on here. Worth a follow-up issue so the asymmetry is recorded rather than forgotten.

**Framing:** under Airflow's security model the Dag author is a trusted party who can already execute arbitrary code, so `remote_base_dir` isn't an untrusted input boundary — this is a robustness/correctness fix, not a vulnerability. Worth saying so in the description so nobody routes it through the security process.

---

PR #70095, opened 2026-07-19
Add worker-name filter to `airflow edge list-workers`

shubhamraj-git:
In clusters with many edge workers, `list-workers` could only be narrowed by `--state`. This adds an optional `--worker-name-pattern` glob (e.g. `'prod-*'`, `'*gpu*'`) that filters workers by name DB-side, composable with `--state`. This brings the CLI to parity with the edge UI/REST route, which already supports `worker_name_pattern`.

Usage:
    airflow edge list-workers --worker-name-pattern 'bree-*'

image

- [x] Yes (Opus 4.8)

### DIFF
--- a/providers/edge3/docs/deployment.rst
+++ b/providers/edge3/docs/deployment.rst
@@ -248,7 +248,9 @@ on all workers in the cluster or to check the status of all workers in the clust
 These set of commands need database access, and can only be called on the central Airflow
 instance. The commands are:
 
-- ``airflow edge list-workers``: List all workers in the cluster
+- ``airflow edge list-workers``: List all workers in the cluster. Accepts an optional
+  ``--name-pattern`` glob (e.g. ``'prod-*'``) to filter workers by name,
+  and ``-s``/``--state`` to filter by worker state.
 - ``airflow edge remote-edge-worker-request-maintenance``: Request a remote edge worker to enter maintenance mode
 - ``airflow edge remote-edge-worker-update-maintenance-comment``: Updates the maintenance comment for a remote edge worker
 - ``airflow edge remote-edge-worker-exit-maintenance``: Request a remote edge worker to exit maintenance mode
--- a/providers/edge3/docs/deployment.rst
+++ b/providers/edge3/docs/deployment.rst
@@ -249,7 +249,7 @@ These set of commands need database access, and can only be called on the centra
 instance. The commands are:
 
 - ``airflow edge list-workers``: List all workers in the cluster. Accepts an optional
-  ``--name-pattern`` glob (e.g. ``'prod-*'``) to filter workers by name,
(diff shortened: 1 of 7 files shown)


SameerMesiah97 (approved):

dheerajturaga:
@shubhamraj-git, looks like there are conflicts. Can you resolve them?

What are your thoughts on extending the `--edge-hostname` to accept a pattern instead. Given we already have the argument we may be able to use the same variable rather than adding a new argument

dheerajturaga (changes requested):
Actually on second pass, the current approach is better. Could we rename this option from `--name-pattern` to `--worker-name-pattern`?

The command output and model field both use `worker_name`, and the edge API/UI already expose `worker_name_pattern`, so `--worker-name-pattern` makes the target field clearer. I'd avoid extending `--edge-hostname` for this, since that option is used as an exact worker identity for mutation commands.

---
Drafted-by: Codex (GPT-5); reviewed by @dheerajturaga before posting

shubhamraj-git:
> Actually on second pass, the current approach is better. Could we rename this option from `--name-pattern` to `--worker-name-pattern`?
> 
> The command output and model field both use `worker_name`, and the edge API/UI already expose `worker_name_pattern`, so `--worker-name-pattern` makes the target field clearer. I'd avoid extending `--edge-hostname` for this, since that option is used as an exact worker identity for mutation commands.
> 
> Drafted-by: Codex (GPT-5); reviewed by @dheerajturaga before posting

I gave it a thought before raising this PR, 
but 

`airflow edge list-workers --worker-name-pattern PATTERN`

sounds bit redundant on the worker side, so thought of going with just name-pattern since here name self explains it is for worker, let me know what do you think?

dheerajturaga:
> > Actually on second pass, the current approach is better. Could we rename this option from `--name-pattern` to `--worker-name-pattern`?
> > The command output and model field both use `worker_name`, and the edge API/UI already expose `worker_name_pattern`, so `--worker-name-pattern` makes the target field clearer. I'd avoid extending `--edge-hostname` for this, since that option is used as an exact worker identity for mutation commands.
> > Drafted-by: Codex (GPT-5); reviewed by @dheerajturaga before posting
> 
> I gave it a thought before raising this PR, but
> 
> `airflow edge list-workers --worker-name-pattern PATTERN`
> 
> sounds bit redundant on the worker side, so thought of going with just name-pattern since here name self explains it is for worker, let me know what do you think?

just incase we want to use the same option for other edge CLIs it would make it clear

dheerajturaga (approved):
Awesome! Works as expected. Thanks for making adjustments as needed!

vatsrahul1001:
This change is more appropriate for minor 3.4.0

PR #70096, opened 2026-07-19
Validate common AI tool-call arguments to enable pydantic-ai retries

zozo123:
Malformed model-generated arguments currently bypass validation in the Common AI toolsets. Missing or mistyped arguments can therefore reach tool implementations and fail with `KeyError` or `TypeError` instead of producing the bounded retry that pydantic-ai supports.

Invalid payloads are now rejected at validation; pydantic-ai then regenerates a new tool call — the original malformed payload is never re-executed.

Concretely: if the model calls the `query` tool but omits the required `sql` argument, the call previously crashed the run with `KeyError: 'sql'`; it now raises a `ValidationError` that pydantic-ai turns into one bounded retry, and the model regenerates the call with `sql` supplied.

This change derives each tool's argument validator from the same JSON Schema advertised to the model and applies it to `SQLToolset`, `DataFusionToolset`, and `HookToolset`. Hook introspection keeps optional and union parameters accurate, leaves unsupported annotations permissive, rejects unexpected arguments for fixed signatures (so a mistyped field becomes a bounded retry, matching native pydantic-ai), and preserves additional arguments for methods that accept `**kwargs`.

The LangChain bridge mirrors the same two-stage semantics: `ValidationError` from arg validation is fed back as a retry; `ValidationError` raised inside `call_tool` propagates so a non-idempotent tool is not re-invoked.

Validation:

- 1,043 Common AI unit tests passed
- provider mypy passed
- pre-commit and manual prek checks passed
- Follow-up unit tests for bridge two-stage catching and `forbid` extras passed

---

- [X] Yes — Claude Code (Fable 5) and Codex (GPT-5)

Generated-by: Claude Code (Fable 5) and Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


providers/common/ai/src/airflow/providers/common/ai/utils/tool_definition.py
```diff
@@ -42,3 +43,55 @@ def return_schema_kwargs(schema: dict[str, Any]) -> dict[str, Any]:
...
+            return core_schema.bool_schema()
+        case "null":
+            return core_schema.none_schema()
+        case "array":
+            items = fragment.get("items")
+            return core_schema.list_schema(
+                _fragment_to_core_schema(items) if isinstance(items, dict) else None
+            )
+        case "object":
+            return core_schema.dict_schema()
```
SameerMesiah97: Is there a need for recursively validating nested properties here like for arrays above? Is this intentional because Common AI tool parameters are expected to be flat, or should nested object schemas also be validated recursively?

SameerMesiah97 (commented):
I think the title is a bit misleading. It appears that the malformed payload is not retried, but rather rejected during validation, allowing Pydantic AI to regenerate a new tool call. "Retry malformed tool calls" initially reads as though the original invalid payload is executed again, whereas the retry is actually another model generation. It might be worth clarifying this in the title and PR description.

This looks fine to me logically (and the tests are solid) but this will need someone who is more familiar with this provider to deeply vet the validation logic. I see a lot of small changes scattered throughout the diff that I am not sure of how to ascertain the impact of. 

CI needs to be triggered too.

providers/common/ai/src/airflow/providers/common/ai/utils/tool_definition.py
zozo123: Good call — I went ahead and made it consistent rather than leaving the asymmetry. `build_args_validator` and the `object` branch now share one builder (`_object_fragment_to_core_schema`), so nested objects with declared `properties` are validated recursively, exactly like arrays already recurse into `items`. Objects with no `properties` key (e.g. from a `dict[K, V]` annotation) still accept any dict, so untyped payloads aren't over-tightened. Added unit tests covering nested coercion, nested type/required errors, and the untyped-dict passthrough. Pushed in f175a10.

zozo123:
Thanks @SameerMesiah97 — good catch on the title. Renamed to "Validate common AI tool-call arguments to enable pydantic-ai retries" and added a line to the description making explicit that the malformed payload is rejected at validation and pydantic-ai regenerates a fresh tool call (the original is never re-executed). Also replied inline on the nested-object question.

providers/common/ai/src/airflow/providers/common/ai/toolsets/sql.py
```diff
@@ -257,7 +254,7 @@ async def get_tools(self, ctx: RunContext[Any]) -> dict[str, ToolsetTool[Any]]:
                 toolset=self,
                 tool_def=tool_def,
                 max_retries=1,
-                args_validator=_PASSTHROUGH_VALIDATOR,
+                args_validator=build_args_validator(schema),
```
kaxil, line 272: On the LangChain bridge (`airflow_toolset_to_langchain_tools`, not in this diff), the new strict validator regresses malformed SQL calls from a bounded retry to an uncaught crash.

`_build_structured_tool._validate` runs `args_validator.validate_python(kwargs)` on the raw args LangChain hands in, and LangChain does not itself enforce the dict `args_schema`, so a call that omits `sql` now raises `pydantic_core.ValidationError` right there. But `_sync_call`/`_async_call` (langchain_bridge.py:177, 185) only catch `ModelRetry`, so it propagates uncaught and aborts the run.

Before this PR the passthrough validator let those args reach `call_tool`, where the `KeyError` was caught and turned into a bounded `ModelRetry` (sql.py:278-288). SQL is the one clean regression here: `HookToolset` (`method(**tool_args)` -> `TypeError`) and `DataFusionToolset` (`tool_args["sql"]` outside the inner `try`) already crashed uncaught on missing args, so their bridge behaviour is unchanged.

pydantic-ai's own `ToolManager` catches `ValidationError` and turns it into a bounded retry, which is exactly the behaviour this PR enables on the native path. Catching `ValidationError` alongside `ModelRetry` in `_sync_call`/`_async_call` and routing it through `_handle_retry` would restore that on the bridge too. Non-blocking: the native pydantic-ai path this PR targets is correct and well tested.

kaxil (approved):

eladkal (approved):
LGTM
cc @kaxil for another pass

providers/common/ai/src/airflow/providers/common/ai/toolsets/langchain_bridge.py
```diff
@@ -174,15 +182,15 @@ def _handle_retry(error: ModelRetry) -> str:
     def _sync_call(**kwargs: Any) -> Any:
         try:
             result = _run_coro_sync(toolset.call_tool(name, _validate(kwargs), ctx, toolset_tool))
-        except ModelRetry as e:
+        except (ModelRetry, ValidationError) as e:
```
kaxil: The widened `except (ModelRetry, ValidationError)` spans both `_validate(kwargs)` and `toolset.call_tool(...)`, so a `ValidationError` raised inside the tool body (a `HookToolset` method, an MCP client validating a server response, a custom toolset) now gets fed back as a retry too. pydantic-ai 2.13 two-stages this: arg-validation retries a `ValidationError`, but `_raw_execute` (`tool_manager.py:758`) catches only `ModelRetry` around `call_tool`, so a tool-body `ValidationError` propagates rather than retrying. For a non-idempotent tool that already ran a side effect before raising, the retry re-invokes it. SQL/DataFusion aren't affected (SQL wraps everything in `ModelRetry`, DataFusion doesn't raise `ValidationError`); this is the Hook/MCP/custom path the docstring covers. Catching `ValidationError` only around `_validate` and `ModelRetry` around `call_tool` keeps the mirror-native behaviour. Non-blocking.

providers/common/ai/src/airflow/providers/common/ai/utils/tool_definition.py
```diff
@@ -42,3 +43,70 @@ def return_schema_kwargs(schema: dict[str, Any]) -> dict[str, Any]:
...
+    arrays already recurse into ``items``.
+    """
+    if "properties" not in fragment:
+        return core_schema.dict_schema()
+    required = set(fragment.get("required", []))
+    fields = {
+        name: core_schema.typed_dict_field(_fragment_to_core_schema(prop), required=name in required)
+        for name, prop in fragment["properties"].items()
+    }
+    extra_behavior: Literal["allow", "ignore"] = (
```
kaxil: `extra_behavior="ignore"` silently drops undeclared args. Native pydantic-ai uses `forbid` for fixed signatures (`_function_schema.py:358`, `allow` only when there's a `**kwargs`), so a model that sends `region` when the field is `region_name` gets a bounded retry to fix itself instead of a silent partial call. Since this PR is about turning malformed calls into retries, was `ignore` deliberate to avoid spurious retries when a model over-supplies fields, or worth aligning with native's `forbid` for fixed signatures?

providers/common/ai/src/airflow/providers/common/ai/toolsets/sql.py
zozo123: Good catch — addressed in e53d7e9 (feed arg-validation `ValidationError` through the bridge retry path) and refined in e3ebc028f6 to two-stage catching so we don't over-retry tool-body errors (see sibling thread).

providers/common/ai/src/airflow/providers/common/ai/toolsets/langchain_bridge.py
zozo123: Agreed — that was an over-broad catch. Fixed in e3ebc028f6: `ValidationError` is now caught only around `_validate`, and `ModelRetry` only around `call_tool`, matching pydantic-ai's two-stage `ToolManager` behaviour. A tool-body `ValidationError` (Hook/MCP/custom) propagates instead of being fed back, so a non-idempotent tool that already ran a side effect is not re-invoked. Added a regression test for that path and tightened the bridge docstring/comments accordingly.

providers/common/ai/src/airflow/providers/common/ai/utils/tool_definition.py
zozo123: Worth aligning with native — `ignore` was the wrong default for this PR's goal. Switched to `forbid` for fixed signatures in e3ebc028f6 (`allow` only when the schema has `additionalProperties: true`, i.e. methods with `**kwargs`). A mistyped field like `region` vs `region_name` now becomes a bounded retry instead of a silent partial call. Updated the unit tests and PR description to match.

PR #70101, opened 2026-07-19
Add Cortex Agent management methods to SnowflakeCortexAgentHook

SameerMesiah97:
**Description**

This change extends `SnowflakeCortexAgentHook` with support for managing Snowflake Cortex Agent Objects.

The hook now exposes methods to describe, list and delete Cortex Agents using the Snowflake Cortex Agent REST API. To support these endpoints, the internal request helper has been updated to accept optional query parameters while continuing to centralize authentication, request execution and error handling.

**Rationale**

The existing hook supports executing Cortex Agents through `run_agent()`, but the Snowflake Cortex Agent REST API also provides endpoints for managing Cortex Agent Objects. Exposing these operations through the hook allows users to inspect and manage agents without needing to invoke the REST API directly.

Adding query parameter support to the shared request helper also provides a reusable foundation for additional Cortex Agent management operations in future enhancements.

**Tests**

* `describe_agent()` sends the expected request and returns the API response.
* `list_agents()` correctly forwards optional query parameters and returns the list of agents.
* `delete_agent()` issues the expected DELETE request, including both values of the `if_exists` parameter.
* `_request()` returns `{}` for successful empty dict responses.
* Cortex Agent methods raise when Snowflake returns an unexpected top-level response shape.

### DIFF
--- a/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, cast
 
 import requests
 
@@ -54,8 +54,9 @@ class SnowflakeCortexAgentHook(SnowflakeHook):
         method: str,
         endpoint: str,
         payload: dict[str, Any] | None = None,
+        params: dict[str, Any] | None = None,
         timeout: int | None = None,
-    ) -> dict[str, Any]:
+    ) -> dict[str, Any] | list[dict[str, Any]]:
 
         response = requests.request(
             method=method,
@@ -65,6 +66,7 @@ class SnowflakeCortexAgentHook(SnowflakeHook):
                 "Content-Type": "application/json",
             },
             json=payload,
+            params=params,
             timeout=timeout,
         )
 
@@ -159,11 +161,124 @@ class SnowflakeCortexAgentHook(SnowflakeHook):
 
         endpoint = f"/api/v2/databases/{database}/schemas/{schema}/agents/{agent_name}:run"
 
-        return self._request(
-            method="POST",
-            endpoint=endpoint,
-            payload=payload,
-            timeout=timeout,
+        return cast(
+            "dict[str, Any]",
+            self._request(
+                method="POST",
+                endpoint=endpoint,
+                payload=payload,
+                timeout=timeout,
+            ),
+        )
+
+    def describe_agent(
+        self,
+        *,
+        database: str,
+        schema: str,
+        agent_name: str,
+        timeout: int | None = 600,
+    ) -> dict[str, Any]:
+        """
+        Describe a Snowflake Cortex Agent.
+
+        :param database: Database containing the Cortex Agent.
+        :param schema: Schema containing the Cortex Agent.
+        :param agent_name: Name of the Cortex Agent.
+        :param timeout: Maximum time in seconds to wait for the Cortex Agent
+            request to complete. Defaults to ``600``.
+        :return: JSON description of the Cortex Agent.
+        """
+        endpoint = f"/api/v2/databases/{database}/schemas/{schema}/agents/{agent_name}"
+
+        return cast(
+            "dict[str, Any]",
+            self._request(
+                method="GET",
+                endpoint=endpoint,
+                timeout=timeout,
+            ),
+        )
+
+    def list_agents(
+        self,
+        *,
+        database: str,
+        schema: str,
+        like: str | None = None,
+        from_name: str | None = None,
+        show_limit: int | None = None,
+        timeout: int | None = 600,
+    ) -> list[dict[str, Any]]:
+        """
+        List Snowflake Cortex Agents.
+
+        :param database: Database containing the Cortex Agents.
+        :param schema: Schema containing the Cortex Agents.
+        :param like: Optional case-insensitive name filter. Optional.
+            Defaults to ``None``.
+        :param from_name: Optional pagination starting point. Optional.
+            Defaults to ``None``.
+        :param show_limit: Maximum number of agents to return. Optional.
+            Defaults to ``None``.
+        :param timeout: Maximum time in seconds to wait for the Cortex Agent
+            request to complete. Defaults to ``600``.
+        :return: List of Cortex Agents.
+        """
+        endpoint = f"/api/v2/databases/{database}/schemas/{schema}/agents"
+
+        params: dict[str, Any] = {}
+
+        if like is not None:
+            params["like"] = like
+
+        if from_name is not None:
+            params["fromName"] = from_name
+
+        if show_limit is not None:
+            params["showLimit"] = show_limit
+
+        return cast(
+            "list[dict[str, Any]]",
+            self._request(
+                method="GET",
+                endpoint=endpoint,
+                params=params or None,
+                timeout=timeout,
+            ),
+        )
+
+    def delete_agent(
+        self,
+        *,
+        database: str,
+        schema: str,
+        agent_name: str,
+        if_exists: bool = False,
+        timeout: int | None = 600,
+    ) -> dict[str, Any]:
+        """
+        Delete a Snowflake Cortex Agent.
+
+        :param database: Database containing the Cortex Agent.
+        :param schema: Schema containing the Cortex Agent.
+        :param agent_name: Name of the Cortex Agent.
+        :param if_exists: If ``True``, do not fail when the agent does not exist.
+            Defaults to ``False``.
+        :param timeout: Maximum time in seconds to wait for the Cortex Agent request
+            to complete. Defaults to ``600``.
+        :return: JSON response confirming deletion.
+        """
+        endpoint = f"/api/v2/databases/{database}/schemas/{schema}/agents/{agent_name}"
+
+        return cast(
+            "dict[str, Any]",
+            self._request(
+                method="DELETE",
+                endpoint=endpoint,
+                params={"ifExists": str(if_exists).lower()},
+                timeout=timeout,
+            ),
         )
 
     @staticmethod
--- a/providers/snowflake/tests/unit/snowflake/hooks/test_snowflake_cortex_agent.py
+++ b/providers/snowflake/tests/unit/snowflake/hooks/test_snowflake_cortex_agent.py
@@ -127,6 +127,7 @@ class TestSnowflakeCortexAgentHook:
                 ],
                 "stream": False,
             },
+            params=None,
             timeout=REQUEST_TIMEOUT,
         )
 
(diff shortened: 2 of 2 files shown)


SameerMesiah97:
Requesting review for this.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -159,11 +161,112 @@ def run_agent(
...
+            "dict[str, Any]",
+            self._request(
+                method="POST",
+                endpoint=endpoint,
+                payload=payload,
+                timeout=timeout,
+            ),
+        )
+
+    def describe_agent(
```
shahar1, line 174: Could you please add a `timeout` definition to all new HTTP method?

vikramkoka:
cc: @kaxil 
Similar to the other PR conversation

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -159,11 +161,124 @@ def run_agent(
...
+    ) -> list[dict[str, Any]]:
+        """
+        List Snowflake Cortex Agents.
+
+        :param database: Database containing the Cortex Agents.
+        :param schema: Schema containing the Cortex Agents.
+        :param like: Optional case-insensitive name filter. Optional.
+            Defaults to ``None``.
+        :param from_name: Optional pagination starting point. Optional.
+            Defaults to ``None``.
```
justinpakzad: Nit: "Optional" is repeated here. I think we can drop the first occurrence.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -159,11 +161,124 @@ def run_agent(
...
-            timeout=timeout,
+        return cast(
+            "dict[str, Any]",
+            self._request(
+                method="POST",
+                endpoint=endpoint,
+                payload=payload,
+                timeout=timeout,
+            ),
+        )
```
Dev-iL: Not a big fan of `cast`. How do we know for sure that this is not a list of dicts? Is it an API contract, or can it be guessed from the combination of input parameters?

I suggest defining overloads for `self._request` (e.g. one configuration of inputs that returns a dict and one that returns list[dict]), or if that's impossible, something like

```python
response = self._request(...)
if not isinstance(response, dict):
    raise TypeError(f"Expected a dict response, got { type(response).__name__ }")
return response  # guaranteed to be dict now
```

...and similarly for the other methods.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, cast
```
Dev-iL: Something like this might be a bit cleaner:

```suggestion
from typing import Any, cast

JsonDict = dict[str, Any]
JsonOrList = JsonDict | list[JsonDict]
```

ashb:
Please don't merge this just yet - we are looking at unifying this in to common.ai

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
Dev-iL: Just to clarify, this is what I had in mind for the overloads:

```python
from typing import Literal, overload

# 1. Define the overload signature for "POST"
@overload
def make_request(method: Literal["POST"]) -> dict:
    ...

# 2. Define the overload signature for "GET"
@overload
def make_request(method: Literal["GET"]) -> list[dict]:
    ...

# 3. Define the implementation function
def make_request(method: str) -> dict | list[dict]:
    if method == "POST":
        return {"status": "created"}
    elif method == "GET":
        return [{"status": "item1"}, {"status": "item2"}]
    else:
        raise ValueError(f"Unsupported method: {method}")

# --- Usage Example ---

# Type checker knows `res_post` is a `dict`
res_post = make_request("POST")

# Type checker knows `res_get` is a `list[dict]`
res_get = make_request("GET")
```

SameerMesiah97:
> Please don't merge this just yet - we are looking at unifying this in to common.ai

I appreciate the heads up. But would it not take a substantial amount of time for the maintainers to align on the final abstraction? There is also a possibility we may decide against it. I wonder if we could let these PRs merge whilst that discussion continues.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: Fixed.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: The type of the response object is not necessarily determined by the HTTP method. I have followed the spirit of your suggestion by defining overloads for the `_request` method, which uses a `response_type` parameter to determine the return type.

Dev-iL (approved):

SameerMesiah97:
I believe this should be okay to merge without adding support for the additional interface (`BaseManagedAgentToolset`) introduced in PR #71946. The new interface is related but somewhat tangential to these new hook methods which this PR was submitted. to introduce. Adding say `SnowflakeCortexAgentToolset` to this PR would be out of scope for so a follow-up PR will be opened to align the the Snowflake Cortex Agent module with the new interface.

justinpakzad (approved):

eladkal:
> Please don't merge this just yet - we are looking at unifying this in to common.ai

@ashb @kaxil please advise next steps here

SameerMesiah97:
> > Please don't merge this just yet - we are looking at unifying this in to common.ai
> 
> @ashb @kaxil please advise next steps here

I think @ashb's 'don't merge' comment is a bit outdated now that PR #71946 has landed. My understanding is that the `SnowflakeCortexAgentToolset` which will be used to take advantage of the interface introduced in PR #71946 might use some of the methods I am introducing in this PR. Essentially, that piece of work will build on this PR (and the other hook and operator related PRs for Snowflake Cortex Agents) rather than something that blocks these PRs until `SnowflakeCortexAgentToolset` is introduced.

ashb:
Yes, most likely you are correct,

SameerMesiah97:
Requesting review for this.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -164,6 +209,111 @@ def run_agent(
...
+        :param database: Database containing the Cortex Agent.
+        :param schema: Schema containing the Cortex Agent.
+        :param agent_name: Name of the Cortex Agent.
+        :param if_exists: If ``True``, do not fail when the agent does not exist.
+            Optional. Defaults to ``False``.
+        :param timeout: Maximum time in seconds to wait for the Cortex Agent request
+            to complete. Optional. Defaults to ``600``.
+        :return: JSON response confirming deletion.
+        """
+        endpoint = f"/api/v2/databases/{database}/schemas/{schema}/agents/{agent_name}"
```
kaxil, line 309: These path segments go into the URL unencoded, and `requests` normalises the result before sending, so the request can land somewhere other than the arguments describe. With `agent_name="../../../schemas/OTHER/agents/VICTIM"` the prepared URL collapses to `/api/v2/databases/MY_DB/schemas/OTHER/agents/VICTIM`, and a name containing `?` turns the remainder into query params. Snowflake identifiers created double-quoted can legitimately contain `/`, `?` and `#`, so this is reachable without anything unusual. Wrapping each segment in `urllib.parse.quote(x, safe="")` fixes it, and it is worth doing for `describe_agent`, `list_agents` and the existing `run_agent` endpoint on line 205 in the same pass.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -77,7 +108,22 @@ def _request(
 
         response.raise_for_status()
 
-        return response.json()
+        if not response.content:
+            if response_type == "dict":
+                return {}
```
kaxil, line 113: Does this need to handle empty bodies at all? None of the three endpoints return one: per the Snowflake docs `DELETE .../agents/{name}` returns `{"status": "Request successfully completed"}`, describe returns the agent object, and `:run` with `stream: false` returns a single populated object. Where it does fire is `run_agent`, which used to raise `JSONDecodeError` on an empty 200 and now returns `{}` that `SnowflakeCortexAgentOperator` pushes to XCom as a successful task with no agent output. Since it keys on `response_type` rather than on the method, dropping it keeps that failure loud.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -77,7 +108,22 @@ def _request(
...
+
+        data = response.json()
+
+        if response_type == "dict":
+            if not isinstance(data, dict):
+                raise TypeError(f"Expected dict response, got {type(data).__name__}")
+            return data
+
+        if not isinstance(data, list) or not all(isinstance(item, dict) for item in data):
+            raise TypeError(f"Expected list[dict] response, got {type(data).__name__}")
```
kaxil, line 124: When the body is a list whose elements are not dicts, this reports `got list`, which reads as if a list was the wrong shape: `[1, 2, 3]` raises `Expected list[dict] response, got list`. Splitting the two conditions would let the element case say what actually failed. That half is also the only branch in the new validation without a test, since the parametrized case sends a dict and short-circuits on `not isinstance(data, list)`.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -164,6 +209,111 @@ def run_agent(
...
+        if like is not None:
+            params["like"] = like
+
+        if from_name is not None:
+            params["fromName"] = from_name
+
+        if show_limit is not None:
+            params["showLimit"] = show_limit
+
+        return self._request(
```
kaxil, line 280: Does this endpoint paginate? Snowflake's v2 spec defines a `Link` response header for list endpoints carrying `next` and `last` rels, and `_request` drops response headers entirely, so a caller has no way to tell a full result from a truncated first page. Since `list_agents` is new API, it seems worth deciding now whether to surface that or to document `from_name` as the way to page through.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: I have wrapped the various parts of each endpoint with `urllib.parse.quote(x, safe="")`

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: Removed lines 111-115 in the previous version of the file (before the last commit).

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: The conditions have been split in to separate branches. A new test case has been added to cover scenarios where the elements of the list are not of type `dict`.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: The pagination mechanism has now been explicitly documented in the docstring for `list_agents`.

shahar1 (approved):

PR #68259, opened 2026-06-09
Add SnowparkContainerJobOperator

justinpakzad:
---
**Description**

Snowpark Container Services lets users run containerized workloads directly within their Snowflake account. Things like ML training jobs, data processing pipelines, or any custom application packaged as a Docker image. Jobs are submitted via EXECUTE JOB SERVICE, which runs containers on a compute pool and terminates once the code exits. Currently there's no operator for orchestrating these jobs. Users would have to fall back to raw SQL via SQLExecuteQueryOperator and manually handle polling, log retrieval, error handling, and service teardown.

This PR adds SnowparkContainerJobOperator which:
  - Submits the job asynchronously via EXECUTE JOB SERVICE
  - Polls DESCRIBE SERVICE until the job reaches a terminal state (DONE, FAILED, etc.)
  - Fetches container logs via SYSTEM$GET_SERVICE_LOGS for all replicas and surfaces them in the Airflow task logs
  - Supports both stage-based specs (spec + spec_stage) and inline YAML (spec_text)
  - Supports fire-and-forget mode (wait_for_completion=False) for long-running jobs as well as other optional parameters (drop_on_completion, replicas, query_warehouse, etc.)

**Testing:**
- Unit tests covering parameter validation, SQL generation, polling logic, log retrieval, error handling for unexpected statuses, on_kill cleanup, and  execution paths.
- Manually tested end-to-end against a real Snowflake account with an active compute pool. Verified the success path (job completes, logs retrieved) and failure scenarios (job fails, error logs surfaced, operator raises appropriately). Also tested with various parameters such as replicas, spec_text vs spec/spec_stage, and wait_for_completion variations.

Added an example DAG and operator documentation. Running the example DAG requires some infrastructure to be provisioned beforehand (compute pool, image repository, pushed container image, etc.) those steps are covered in the operator's how-to guide. If approved, I plan on opening a follow-up PR to implement deferrable mode.

##### Was generative AI tooling used to co-author this PR?

- [x] Yes (Claude Opus 4.6)

---

* Read the **[Pull Request Guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-guidelines)** for more information. Note: commit author/co-author name and email in commits become permanently public when merged.
* For fundamental code changes, an Airflow Improvement Proposal ([AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/Airflow+Improvement+Proposals)) is needed.
[...]

PR #65590, opened 2026-04-21
Add on_kill() hook to BaseTrigger to handle user actions on triggers

amoghrajesh:
---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes:  Claude code

closes: https://github.com/apache/airflow/issues/36090

### Why + What
Deferred tasks interacting with external systems (BQ, DBX, etc.) have no standard way to cancel the external job when a user marks the task failed, clears it, or marks it succeeded. This leaves orphaned jobs running in the background.

I intend to fix that by adding `on_kill()` to `BaseTrigger` — a no-op by default that trigger authors override to cancel their external job when the task is acted on by a user.

The triggerer already passes trigger IDs through a `to_cancel` queue when a trigger is removed. Everything in that queue is a user action —> redistribution only happens when a triggerer is considered dead, at which point it is no longer processing its queue. We use `asyncio.Task.cancel("user-action")` to pass the reason through the existing cancellation path, so `run_trigger()` can check `e.args[0]` and `invoke on_kill()` without any extra API round-trip.

### Impact
1. Existing trigger subclasses
Purely additive. `on_kill()` defaults to a no-op so nothing breaks if you do not implement it.

2. Triggerer shutdown / redistribution
No performance impact. `on_kill()` is only invoked when `CancelledError` carries the "user-action" message. Shutdown fires `task.cancel("EOF - shutting down")` and redistribution goes through the `to_create` path — neither triggers `on_kill()`.

3. Provider authors
Override `on_kill()` to cancel external jobs. The framework guarantees it is only called for user initiated cancellations and you can implement it as needed.

### Testing

Trigger

```python

SENTINEL_FILE = "/tmp/on_cancel_was_called"


class VerifyOnCancelTrigger(BaseTrigger):
    """
    Trigger that sleeps forever.

    on_cancel() writes a sentinel file so we can verify it was called
    without needing a real external service like Databricks or BigQuery.
    """

    def serialize(self) -> tuple[str, dict]:
        return ("verify_on_cancel_trigger.VerifyOnCancelTrigger", {})

    async def run(self) -> AsyncIterator[TriggerEvent]:
        log.info("VerifyOnCancelTrigger running — mark this task as failed to test on_cancel()")
        await asyncio.sleep(99999)
        yield TriggerEvent("done")  # never reached

    async def on_kill(self) -> None:
        log.info("VerifyOnCancelTrigger.on_cancel() CALLED — simulated external job cancelled")
[...]

PR #70103, opened 2026-07-19
Add deferrable mode to SnowparkContainerJobOperator

justinpakzad:
---
**Description**

Follow-up to #68259 adding deferrable mode to SnowparkContainerJobOperator. Snowpark Container jobs are often long-running (ML training, batch processing, other containerized workloads that run for hours), which makes them a good fit for deferrable mode.

The SnowflakeHook is synchronous, so this follows a similar approach the Anthropic provider took. The trigger wraps each blocking hook call in asyncio.to_thread so a single poll doesn't stall the shared triggerer event loop. A timeout parameter is introduced for both modes. Cleanup is consistent across both modes, on timeout the service is dropped to stop billing on a job past its deadline. On poll errors and terminal failures it's left in place for inspection. 

The status Enum and terminal/non-terminal frozensets were moved to the trigger file so the operator can import them without a circular dependency.

**Testing**
- Unit tests for the trigger (serialization, terminal/unexpected/timeout/error events, cleanup) and the operator's defer and execute_complete paths.
- Manually tested end-to-end against a real Snowflake account: deferred success (job completes, logs retrieved, service dropped), terminal failure, both timeout types, on_kill, and the drop toggles.

- [x] Yes (Claude Opus 4.6)

---

### DIFF
--- a/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake.py
@@ -22,6 +22,7 @@ import os
 from collections.abc import Callable, Iterable, Mapping
 from contextlib import closing, contextmanager
 from datetime import datetime, timedelta
+from enum import Enum
 from functools import cached_property
 from io import StringIO
 from pathlib import Path
@@ -62,6 +63,39 @@ if TYPE_CHECKING:
 
     from airflow.providers.openlineage.extractors import OperatorLineage
     from airflow.providers.openlineage.sqlparser import DatabaseInfo
+
+
+class SnowparkContainerJobStatus(str, Enum):
+    """Statuses of a Snowpark Container Services job service."""
+
+    PENDING = "PENDING"
+    RUNNING = "RUNNING"
+    CANCELLING = "CANCELLING"
+    SUSPENDING = "SUSPENDING"
+    DELETING = "DELETING"
+    DONE = "DONE"
+    FAILED = "FAILED"
+    CANCELLED = "CANCELLED"
+    INTERNAL_ERROR = "INTERNAL_ERROR"
+
+
+CONTAINER_JOB_TERMINAL_STATUSES: frozenset[SnowparkContainerJobStatus] = frozenset(
+    {
+        SnowparkContainerJobStatus.DONE,
+        SnowparkContainerJobStatus.FAILED,
+        SnowparkContainerJobStatus.CANCELLED,
+        SnowparkContainerJobStatus.INTERNAL_ERROR,
+    }
+)
+CONTAINER_JOB_NON_TERMINAL_STATUSES: frozenset[SnowparkContainerJobStatus] = frozenset(
+    {
+        SnowparkContainerJobStatus.PENDING,
+        SnowparkContainerJobStatus.RUNNING,
+        SnowparkContainerJobStatus.CANCELLING,
+        SnowparkContainerJobStatus.SUSPENDING,
+        SnowparkContainerJobStatus.DELETING,
+    }
+)
 
 
 def _try_to_boolean(value: Any):
--- a/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
@@ -27,7 +27,7 @@ from airflow.providers.common.compat.sdk import conf
 from airflow.providers.common.compat.standard.operators import BaseOperator
 from airflow.providers.common.sql.hooks.handlers import fetch_one_handler
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
-from airflow.providers.snowflake.triggers.snowpark_containers_trigger import (
+from airflow.providers.snowflake.triggers.snowpark_containers import (
     NON_TERMINAL_STATUSES,
     TERMINAL_STATUSES,
     SnowparkContainerJobStatus,
--- a/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
@@ -19,49 +19,23 @@ from __future__ import annotations
 
 import time
 from collections.abc import Sequence
-from enum import Enum
+from datetime import timedelta
 from functools import cached_property
 from typing import TYPE_CHECKING, Any
 
+from airflow.providers.common.compat.sdk import conf
 from airflow.providers.common.compat.standard.operators import BaseOperator
 from airflow.providers.common.sql.hooks.handlers import fetch_one_handler
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+from airflow.providers.snowflake.triggers.snowpark_containers_trigger import (
+    NON_TERMINAL_STATUSES,
+    TERMINAL_STATUSES,
+    SnowparkContainerJobStatus,
+    SnowparkContainerJobTrigger,
+)
 
 if TYPE_CHECKING:
     from airflow.providers.common.compat.sdk import Context
-
-
-class SnowparkContainerJobStatus(str, Enum):
-    """Statuses of a Snowpark Container Services."""
-
-    PENDING = "PENDING"
-    RUNNING = "RUNNING"
-    CANCELLING = "CANCELLING"
-    SUSPENDING = "SUSPENDING"
-    DELETING = "DELETING"
-    DONE = "DONE"
-    FAILED = "FAILED"
-    CANCELLED = "CANCELLED"
-    INTERNAL_ERROR = "INTERNAL_ERROR"
-
-
-TERMINAL_STATUSES: frozenset[SnowparkContainerJobStatus] = frozenset(
-    {
-        SnowparkContainerJobStatus.DONE,
-        SnowparkContainerJobStatus.FAILED,
-        SnowparkContainerJobStatus.CANCELLED,
-        SnowparkContainerJobStatus.INTERNAL_ERROR,
-    }
-)
-NON_TERMINAL_STATUSES: frozenset[SnowparkContainerJobStatus] = frozenset(
-    {
-        SnowparkContainerJobStatus.PENDING,
-        SnowparkContainerJobStatus.RUNNING,
-        SnowparkContainerJobStatus.CANCELLING,
-        SnowparkContainerJobStatus.SUSPENDING,
-        SnowparkContainerJobStatus.DELETING,
-    }
-)
 
 
 class SnowparkContainerJobOperator(BaseOperator):
@@ -100,6 +74,14 @@ class SnowparkContainerJobOperator(BaseOperator):
         (default value: 10)
     :param snowflake_conn_id: Reference to
         :ref:`Snowflake connection id<howto/connection:snowflake>`
+    :param deferrable: Run the operator in deferrable mode. Only effective when
+        ``wait_for_completion`` is True. With ``wait_for_completion=False`` the
+        operator submits the job and returns immediately without deferring.
+        (default value: False)
+    :param timeout: Maximum seconds to wait for the job to reach a terminal
+        state. When it elapses the job service is dropped and the task fails.
+        A shorter ``execution_timeout`` preempts this cleanup, so the service
+        may be left running. (default value: 86400)
     :param database: name of database (will overwrite database defined
         in connection)
     :param schema: name of schema (will overwrite schema defined in
@@ -137,6 +119,8 @@ class SnowparkContainerJobOperator(BaseOperator):
         drop_on_completion: bool = True,
         poll_interval: int = 10,
         snowflake_conn_id: str = "snowflake_default",
+        deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
+        timeout: int = 24 * 60 * 60,
         database: str | None = None,
         schema: str | None = None,
         role: str | None = None,
@@ -158,12 +142,17 @@ class SnowparkContainerJobOperator(BaseOperator):
         self.drop_on_completion = drop_on_completion
         self.poll_interval = poll_interval
         self.snowflake_conn_id = snowflake_conn_id
+        self.deferrable = deferrable
+        self.timeout = timeout
         self.database = database
         self.schema = schema
         self.role = role
         self.warehouse = warehouse
         # Set after the job is submitted, parsed from the job submission response.
         self.job_name: str | None = None
+
+        if self.deferrable and not self.wait_for_completion:
+            self.log.warning("deferrable has no effect when wait_for_completion is False.")
 
     @cached_property
     def _hook(self) -> SnowflakeHook:
@@ -203,6 +192,7 @@ class SnowparkContainerJobOperator(BaseOperator):
 
     def _poll_for_status(self) -> str:
         """Poll until the job reaches a terminal state."""
+        end_time = time.time() + self.timeout
         while True:
             response = self._run_one(f"DESCRIBE SERVICE {self.job_name}", return_dictionaries=True)
             status = response.get("status")
@@ -210,6 +200,9 @@ class SnowparkContainerJobOperator(BaseOperator):
                 return status
             if status not in NON_TERMINAL_STATUSES:
                 raise RuntimeError(f"Job {self.job_name} returned unexpected status: {status}")
+            if time.time() > end_time:
+                self._drop_service()
+                raise TimeoutError(f"Job {self.job_name} did not reach a terminal status before the timeout.")
             time.sleep(self.poll_interval)
 
     def _log_container_output(self, status: str) -> None:
@@ -224,13 +217,25 @@ class SnowparkContainerJobOperator(BaseOperator):
             else:
                 self.log.info("Logs for instance_id %d:\n%s", instance_id, response)
 
+    def _drop_service(self) -> None:
+        """Best-effort drop of the job service."""
+        try:
+            self._hook.run(f"DROP SERVICE IF EXISTS {self.job_name}")
+        except Exception as e:
+            self.log.error("Error dropping service %s: %s", self.job_name, e)
+
     def on_kill(self) -> None:
         """Drop the running service on task kill."""
         if self.job_name:
-            try:
-                self._hook.run(f"DROP SERVICE IF EXISTS {self.job_name}")
-            except Exception as e:
-                self.log.error("Error dropping service %s: %s", self.job_name, e)
+            self._drop_service()
+
+    def _handle_final_status(self, status: str) -> None:
+        """Log container output, fail unless the job is DONE, and optionally drop the service on success."""
+        self._log_container_output(status)
+        if status != SnowparkContainerJobStatus.DONE:
+            raise RuntimeError(f"Job '{self.job_name}' finished with status: {status}")
+        if self.drop_on_completion:
+            self._drop_service()
 
     def execute(self, context: Context) -> str:
         """Submit and optionally wait for a Snowpark Container Services job."""
@@ -242,10 +247,35 @@ class SnowparkContainerJobOperator(BaseOperator):
             raise RuntimeError("Job name was not returned")
         if not self.wait_for_completion:
             return self.job_name
+        if self.deferrable:
+            self.defer(
+                trigger=SnowparkContainerJobTrigger(
+                    job_name=self.job_name,
+                    snowflake_conn_id=self.snowflake_conn_id,
+                    poll_interval=self.poll_interval,
+                    end_time=time.time() + self.timeout,
+                    database=self.database,
+                    schema=self.schema,
+                    role=self.role,
+                    warehouse=self.warehouse,
+                ),
+                # Pad past the trigger's end_time so its timeout event, which drops the service,
+                # fires before this hard backstop. A user-set execution_timeout takes precedence.
+                timeout=self.execution_timeout or timedelta(seconds=self.timeout + self.poll_interval + 60),
+                method_name="execute_complete",
+            )
         status = self._poll_for_status()
-        self._log_container_output(status)
-        if status != SnowparkContainerJobStatus.DONE:
-            raise RuntimeError(f"Job '{self.job_name}' finished with status: {status}")
-        if self.drop_on_completion:
-            self._hook.run(f"DROP SERVICE IF EXISTS {self.job_name}")
+        self._handle_final_status(status)
+        return self.job_name
+
+    def execute_complete(self, context: Context, event: dict[str, Any]) -> str:
+        """Resume after the trigger fires."""
+        self.job_name = event["job_name"]
+        status = event["status"]
+        if status == "timeout":
+            self._drop_service()
+            raise TimeoutError(event.get("message", f"Job '{self.job_name}' did not complete: {status}"))
+        if status == "error":
+            raise RuntimeError(event.get("message", f"Job '{self.job_name}' did not complete: {status}"))
+        self._handle_final_status(status)
         return self.job_name
--- a/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
@@ -80,8 +80,7 @@ class SnowparkContainerJobOperator(BaseOperator):
         (default value: False)
     :param timeout: Maximum seconds to wait for the job to reach a terminal
         state. When it elapses the job service is dropped and the task fails.
-        A shorter ``execution_timeout`` preempts this cleanup, so the service
-        may be left running. (default value: 86400)
+        (default value: 86400)
     :param database: name of database (will overwrite database defined
         in connection)
     :param schema: name of schema (will overwrite schema defined in
@@ -194,15 +193,15 @@ class SnowparkContainerJobOperator(BaseOperator):
         """Poll until the job reaches a terminal state."""
         end_time = time.time() + self.timeout
         while True:
+            if time.time() >= end_time:
+                self._drop_service()
+                raise TimeoutError(f"Job {self.job_name} did not reach a terminal status before the timeout.")
             response = self._run_one(f"DESCRIBE SERVICE {self.job_name}", return_dictionaries=True)
             status = response.get("status")
             if status in TERMINAL_STATUSES:
                 return status
             if status not in NON_TERMINAL_STATUSES:
                 raise RuntimeError(f"Job {self.job_name} returned unexpected status: {status}")
-            if time.time() > end_time:
-                self._drop_service()
-                raise TimeoutError(f"Job {self.job_name} did not reach a terminal status before the timeout.")
             time.sleep(self.poll_interval)
 
     def _log_container_output(self, status: str) -> None:
@@ -248,20 +247,29 @@ class SnowparkContainerJobOperator(BaseOperator):
         if not self.wait_for_completion:
             return self.job_name
         if self.deferrable:
+            now = time.time()
+            poll_buffer = timedelta(seconds=self.poll_interval + 60)
+            execution_deadline = None
+            defer_timeout = timedelta(seconds=self.timeout) + poll_buffer
+            if self.execution_timeout is not None:
+                # Hand the execution deadline to the trigger so it emits a timeout event that drops the
+                # service. The framework's defer timeout would otherwise kill the task with no cleanup.
+                execution_deadline = now + self.execution_timeout.total_seconds()
+                # Pad the backstop past that deadline so the trigger fires first.
+                defer_timeout = self.execution_timeout + poll_buffer
             self.defer(
                 trigger=SnowparkContainerJobTrigger(
                     job_name=self.job_name,
                     snowflake_conn_id=self.snowflake_conn_id,
                     poll_interval=self.poll_interval,
-                    end_time=time.time() + self.timeout,
+                    end_time=now + self.timeout,
+                    execution_deadline=execution_deadline,
                     database=self.database,
                     schema=self.schema,
                     role=self.role,
                     warehouse=self.warehouse,
                 ),
-                # Pad past the trigger's end_time so its timeout event, which drops the service,
-                # fires before this hard backstop. A user-set execution_timeout takes precedence.
-                timeout=self.execution_timeout or timedelta(seconds=self.timeout + self.poll_interval + 60),
+                timeout=defer_timeout,
                 method_name="execute_complete",
             )
         status = self._poll_for_status()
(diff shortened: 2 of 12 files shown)


justinpakzad:
Would appreciate a review on this. Thanks.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,179 @@
...
+)
+NON_TERMINAL_STATUSES: frozenset[SnowparkContainerJobStatus] = frozenset(
+    {
+        SnowparkContainerJobStatus.PENDING,
+        SnowparkContainerJobStatus.RUNNING,
+        SnowparkContainerJobStatus.CANCELLING,
+        SnowparkContainerJobStatus.SUSPENDING,
+        SnowparkContainerJobStatus.DELETING,
+    }
+)
```
SameerMesiah97, line 60: These statuses describe the Snowpark job/service rather than the trigger itself, and they're also consumed by the operator. Would it make more sense to keep the enum/status sets with the hook (similar to dbt Cloud), so the operator module does not need to import shared abstractions from the trigger module?

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -242,10 +247,35 @@ def execute(self, context: Context) -> str:
...
+                    poll_interval=self.poll_interval,
+                    end_time=time.time() + self.timeout,
+                    database=self.database,
+                    schema=self.schema,
+                    role=self.role,
+                    warehouse=self.warehouse,
+                ),
+                # Pad past the trigger's end_time so its timeout event, which drops the service,
+                # fires before this hard backstop. A user-set execution_timeout takes precedence.
+                timeout=self.execution_timeout or timedelta(seconds=self.timeout + self.poll_interval + 60),
```
SameerMesiah97, line 264: I don't think the timeout passed to `defer() `is necessarily consistent with the deadline used by the trigger. The trigger's `end_time` is always derived from self.timeout, whereas `defer(timeout=...) `uses self.execution_timeout when one is provided. For example, `execution_timeout=60s` with the `default timeout=86400s` gives the framework a 60s timeout while the trigger still has a 24h deadline. This may result in an orphaned job as the trigger will timeout before the timeout event can be emitted (which will be caught by `execute_complete` to trigger service cancellation). I know you added an on_kill method to potentially handle this situation but looking at the BaseTrigger class, it appears that this specific method is inert in timeout scenarios. 

We could pass the execution deadline to the trigger as well and let it determine the applicable deadline, (please see what has been done for the Airbyte provider).  The timeout passed to `defer()` can then be padded slightly so the trigger has an opportunity to emit the corresponding timeout event first.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,179 @@
...
+
+    :param job_name: name of the submitted job service to poll.
+    :param snowflake_conn_id: reference to the Snowflake connection id.
+    :param poll_interval: seconds to sleep between ``DESCRIBE SERVICE`` polls.
+    :param end_time: epoch deadline (``time.time()`` seconds) after which a ``timeout``
+        event is emitted.
+    :param database: (Optional) name of database.
+    :param schema: (Optional) name of schema.
+    :param role: (Optional) name of role.
+    :param warehouse: (Optional) name of warehouse.
```
SameerMesiah97, line 75: I think you should add the default value for each parameter to the docstring entries.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,179 @@
...
+                yield TriggerEvent(
+                    {
+                        "status": "error",
+                        "job_name": self.job_name,
+                        "message": f"Job {self.job_name} returned unexpected status: {status}",
+                    }
+                )
+                return
+
+            if time.time() > self.end_time:
```
SameerMesiah97, line 161: 1) Change this to `>=` to handle timeouts on the boundary.
2) Should the timeout check happen before polling/evaluating the job status? Currently, if the trigger wakes after `end_time` (or the `DESCRIBE SERVICE` call itself takes us past it) and the service returns a terminal status, we'll emit that terminal event even though the trigger deadline has already expired. If `end_time` is intended to be a hard deadline, I think it should be checked before starting another poll.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        return response.get("status") if response else None
+
+    async def run(self) -> AsyncIterator[TriggerEvent]:
+        """Poll the job status and yield exactly one terminal event."""
+        hook = self._get_hook()
+        while True:
+            try:
+                status = await self._describe_status(hook=hook)
+            except Exception as e:
```
SameerMesiah97, line 143: Do we want a single transient `DESCRIBE SERVICE `failure to terminate the trigger, given these jobs may run for hours, or should polling failures be retried up to the deadline?

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,179 @@
...
+        """Return the job's current status via ``DESCRIBE SERVICE``, or ``None`` if absent."""
+        # SnowflakeHook is synchronous. Run the blocking poll off the event loop so a
+        # single query does not stall every other trigger on this triggerer.
+        response: Any = await asyncio.to_thread(
+            hook.run,
+            f"DESCRIBE SERVICE {self.job_name}",
+            handler=fetch_one_handler,
+            return_dictionaries=True,
+        )
+        return response.get("status") if response else None
```
SameerMesiah97, line 135: Just checking if 'status' is guaranteed to be upper case in the response object.

providers/snowflake/tests/unit/snowflake/operators/test_snowpark_containers.py
```diff
@@ -280,3 +312,58 @@ def test_execute_skips_drop_when_disabled(self, mock_hook_cls, mock_submit, mock
...
         mock_hook.run.assert_not_called()
+
+    @mock.patch.object(SnowparkContainerJobOperator, "_submit_job", return_value=JOB_NAME)
+    def test_execute_defers_when_deferrable(self, mock_submit):
+        op = _make_operator(deferrable=True)
+        with pytest.raises(TaskDeferred) as exc:
+            op.execute(context=None)
+        assert isinstance(exc.value.trigger, SnowparkContainerJobTrigger)
+        assert exc.value.trigger.job_name == JOB_NAME
+        assert exc.value.method_name == "execute_complete"
```
SameerMesiah97, line 323: If you implement the deadline handling suggested above, could we add coverage for it here? In particular, an explicit `execution_timeout` should determine the trigger's execution deadline, while the timeout passed to `defer()` should include the additional polling buffer.

providers/snowflake/tests/unit/snowflake/triggers/test_snowpark_containers.py
```diff
@@ -0,0 +1,143 @@
...
+TRIGGER_PATH = "airflow.providers.snowflake.triggers.snowpark_containers"
+CLASSPATH = f"{TRIGGER_PATH}.SnowparkContainerJobTrigger"
+HOOK = f"{TRIGGER_PATH}.SnowflakeHook"
+
+JOB_NAME = "TEST_JOB"
+CONN_ID = "snowflake_default"
+POLL_INTERVAL = 1.0
+
+
+class TestSnowparkContainerJobTrigger:
```
SameerMesiah97, line 36: If you implement the suggested deadline handling in the trigger above, these tests will need updating as well to verify that the timeout takes precedence once the deadline has been reached, including over a terminal status.

SameerMesiah97 (commented):
This needs some work (especially the trigger). I have left some comments.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: I agree the trigger isn't the most obvious place for these, but I'm not sure the hook is the best place either. The `SnowflakeHook` doesn't directly reference them at all. In dbt Cloud the hook actually references the statuses. Happy to move them if you feel strongly though.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Apperciate the suggestion. I agree the execution deadline approach makes more sense here. I've updated the operator to pass an `execution_deadline` to the trigger and padded the `defer()` timeout so the trigger can emit its timeout event first. It now times out at whichever deadline (execution_deadline or end_time) is reached first. Added coverage for the new behavior as well.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: Done.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: I moved timeout checks above the job status evaluation so that the hard deadline is respected. I also updated the timeout check in the operator to align the two.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: I actually originally implemented a max consecutive failures approach here but I removed it from this PR as I felt that change was better suited for a separate PR. I think it should apply to the operator's polling too and not just the trigger. Planning on doing this in a follow up.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: Based on their documentation, yes.

providers/snowflake/tests/unit/snowflake/operators/test_snowpark_containers.py
justinpakzad: Yep, I added coverage for the new deadline handling.

providers/snowflake/tests/unit/snowflake/triggers/test_snowpark_containers.py
justinpakzad: Done.

justinpakzad:
@SameerMesiah97 I appreciate the review. I've addressed and replied to all of your comments. Let me know your thoughts. Thanks.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
SameerMesiah97: I do believe the hook is at a lower-level of abstraction so any 'primitives' such as enums used my multiple modules naturally belong there. I would not say it is a serious issue but importing it like this from the trigger does look a bit strange.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -248,20 +247,29 @@ def execute(self, context: Context) -> str:
         if not self.wait_for_completion:
             return self.job_name
         if self.deferrable:
```
SameerMesiah97, line 255: I would add a comment here explaining the interaction of `execution_timeout` and `timeout` i.e. that the former takes precedence over the latter if both are present.

SameerMesiah97 (commented):
Just one nit. Please check my response to the enum comment too.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: Fair, I've moved them to the hook. I've prefixed the status sets with CONTAINER_JOB_* since bare TERMINAL_STATUSES was now ambiguous about what it belonged to.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Added a comment.

justinpakzad:
> Just one nit. Please check my response to the enum comment too.

Addressed both of the comments.

SameerMesiah97 (approved):

justinpakzad:
Would appreciate a review on this. Thanks.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -242,10 +246,47 @@ def execute(self, context: Context) -> str:
...
+            # how long the job is polled, and execution_timeout, when set, enforces the task-level
+            # limit. The trigger times out on whichever is reached first.
+            now = time.time()
+            poll_buffer = timedelta(seconds=self.poll_interval + 60)
+            execution_deadline = None
+            defer_timeout = timedelta(seconds=self.timeout) + poll_buffer
+            if self.execution_timeout is not None:
+                # Hand the execution deadline to the trigger so it emits a timeout event that drops the
+                # service. The framework's defer timeout would otherwise kill the task with no cleanup.
+                execution_deadline = now + self.execution_timeout.total_seconds()
```
potiuk: `now` is `time.time()` at the moment of deferral, so everything before this point — connection setup, `_submit_job`, the `DESCRIBE` round trip — falls outside the window. `execution_timeout` is a task-level limit measured from task start, so a task with `execution_timeout=120` that spent 30s submitting gets 150s of wall clock. The comment two lines up says this "enforces the task-level limit", which is what makes it worth fixing rather than documenting. Deriving the deadline from the task's start instead would be exact:

```suggestion
                execution_deadline = (
                    context["ti"].start_date.timestamp() + self.execution_timeout.total_seconds()
                )
```

potiuk (commented):
Approach looks solid — the trigger runs the blocking `SnowflakeHook` through `asyncio.to_thread` rather than stalling the triggerer loop, `on_kill` is wired to the real `BaseTrigger` hook, and the tests are unusually thorough (parametrized, `autospec`, `time_machine`, deadline-precedence and boundary cases). Four observations inline, one of which I'd like resolved before this merges; none are blockers.

### Deadline measured from defer time (`operators/snowpark_containers.py:260`)

The deferred path starts the `execution_timeout` clock when the task defers rather than when it started, so total task runtime can exceed `execution_timeout` by however long submission took. Details and a suggested fix inline.

### Smaller observations

- `operators/snowpark_containers.py:30` — `TERMINAL_STATUSES` / `NON_TERMINAL_STATUSES` were public names in a released module; the rename drops them without an alias.
- `operators/snowpark_containers.py:197` — the timeout path drops the service before fetching container logs, and ignores `drop_on_completion=False`.
- `hooks/snowflake.py:68` — status enum placement in the SQL hook module; the trigger module would break the import cycle just as well.

Worth noting the `__init__` warning you added is in the right place — combination-of-arguments checks belong in `__init__` per apache/airflow#70296, not in `execute()`.

---

>

---

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -19,51 +19,25 @@
...
+from datetime import timedelta
 from functools import cached_property
 from typing import TYPE_CHECKING, Any
 
+from airflow.providers.common.compat.sdk import conf
 from airflow.providers.common.compat.standard.operators import BaseOperator
 from airflow.providers.common.sql.hooks.handlers import fetch_one_handler
-from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+from airflow.providers.snowflake.hooks.snowflake import (
+    CONTAINER_JOB_NON_TERMINAL_STATUSES,
```
potiuk: `TERMINAL_STATUSES` and `NON_TERMINAL_STATUSES` were public module-level names in this module and shipped in 6.15.0 (`Add SnowparkContainerJobOperator (#68259)`), still present in 6.16.1. Renaming them to `CONTAINER_JOB_*` and moving them to the hook turns `from airflow.providers.snowflake.operators.snowpark_containers import TERMINAL_STATUSES` into an `ImportError` on what is otherwise a feature release. `SnowparkContainerJobStatus` is unaffected — it's re-imported here — so it's just these two. Two aliases after the import block would keep the old spelling working.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -203,12 +191,16 @@ def _submit_job(self) -> str:
 
     def _poll_for_status(self) -> str:
         """Poll until the job reaches a terminal state."""
+        end_time = time.time() + self.timeout
         while True:
+            if time.time() >= end_time:
+                self._drop_service()
```
potiuk: On timeout the service is dropped before any container logs are fetched, and a job that hung is exactly the case where those logs are wanted — dropping the service destroys them. Consider `self._log_container_output(status)` before the drop here and in the `"timeout"` branch of `execute_complete`. Separately: this path drops unconditionally, ignoring `drop_on_completion=False`. It's documented on the `timeout` param so it may well be deliberate, but it's worth a second look, since a *failed* job is left in place while a *timed-out* one is not.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake.py
```diff
@@ -64,6 +65,39 @@
     from airflow.providers.openlineage.sqlparser import DatabaseInfo
 
 
+class SnowparkContainerJobStatus(str, Enum):
```
potiuk: These Snowpark Container Services statuses now live in the SQL hook module, which is otherwise `SnowflakeHook` plus its OpenLineage surface. I take it the move is to break the operator↔trigger import cycle — `triggers/snowpark_containers.py` would do that equally well, since the operator already imports the trigger and the trigger imports nothing from the operator. Non-blocking, just a placement thought.

justinpakzad: So I had originally placed them in the trigger as you mentioned, but it was pointed it out in one of the review comments above that it would make more sense to keep the enum/status sets with the hook as importing shared abstractions from the trigger module might look a bit strange. I think it's fine where it is but I'm happy to move em back to the trigger if you feel they are suited better there.

SameerMesiah97: > So I had originally placed them in the trigger as you mentioned, but it was pointed it out in one of the review comments above that it would make more sense to keep the enum/status sets with the hook as importing shared abstractions from the trigger module might look a bit strange. I think it's fine where it is but I'm happy to move em back to the trigger if you feel they are suited better there.

I think it is ultimately down to individual preferences. I do not feel strongly about the placement even though I made that comment. It’s up to you really

justinpakzad: While I was working on a follow up to implement retries (so a transient failure doesn't terminate the trigger/operator), I realized the SPCS trigger and operator will need some shared helpers beyond the statuses. So instead of putting them in the hook, I've moved them into a new util file in the existing `/utils` directory. This avoids the need to alias the status imports to maintain backwards compatibility and enables adding new helpers that are only used for SPCS.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Good call. Both timeout branches now log the container output. `_log_container_output` is now best effort (wrapped in try/except), since any failures fetching the logs would prevent the cleanup. As for your second point, I've added the guard to the timeout branches as well so the `drop_on_completion` is respected for timeouts. Tests have been added and updated accordingly.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: See my comment below. I've moved the statuses to a new utility file and reverted back to the original naming since they are not ambiguous anymore.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Good find. Updated.

justinpakzad:
Thanks for the review. I've addressed all of your comments. Please see my response in regards to the placement of the statuses as I have made some changes there. I also re-tested this e2e against my Snowflake account, specifically the new changes which log the container outputs on timeout. It's working as expected.

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
```diff
@@ -0,0 +1,166 @@
...
+                        "status": "error",
+                        "job_name": self.job_name,
+                        "message": f"Job {self.job_name} returned unexpected status: {status}",
+                    }
+                )
+                return
+
+            await asyncio.sleep(self.poll_interval)
+
+    async def on_kill(self) -> None:
```
potiuk, line 159: `BaseTrigger.on_kill()` only landed in Airflow 3.3.0 (apache/airflow#65590), and this provider's floor is `apache-airflow>=2.11.0`. On 2.11 through 3.2 nothing calls this method, so clearing or marking-failed a deferred SPCS task leaves the job service running and billing — and the operator's own `on_kill` doesn't cover that either, since a deferred task isn't on a worker.

The code itself is right and forward-compatible; it's the version gap that's worth surfacing. A sentence in the `deferrable` docstring or in `docs/operators/snowpark_containers.rst` saying that kill-time cleanup of a deferred job requires Airflow 3.3+ would save someone on 3.1 an unexplained orphaned service.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -212,25 +206,41 @@ def _poll_for_status(self) -> str:
...
-                self.log.error("Error dropping service %s: %s", self.job_name, e)
+            self._drop_service()
+
+    def _handle_final_status(self, status: str) -> None:
+        """Log container output, fail unless the job is DONE, and optionally drop the service on success."""
+        self._log_container_output(status)
+        if status != SnowparkContainerJobStatus.DONE:
+            raise RuntimeError(f"Job '{self.job_name}' finished with status: {status}")
+        if self.drop_on_completion:
+            self._drop_service()
```
potiuk, line 245: Small behaviour change worth being deliberate about. Before this PR the success-path drop was:

```python
if self.drop_on_completion:
    self._hook.run(f"DROP SERVICE IF EXISTS {self.job_name}")
```

— no `try`, so a drop that failed failed the task. Routing it through `_drop_service` (line 225) now logs at `error` level and lets the task succeed, which leaves a service running and billing with no failed task to alert on.

Swallowing may well be what you want here — the job did succeed, and failing an otherwise-good task on a cleanup error is harsh. But `on_kill` was best-effort because it runs during teardown, whereas this is the normal completion path, and the failure is now silent. Either keeping the raise on the completion path or a one-line note on why it's best-effort would make the choice legible.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -203,7 +190,14 @@ def _submit_job(self) -> str:
 
     def _poll_for_status(self) -> str:
         """Poll until the job reaches a terminal state."""
+        status = None
+        end_time = time.time() + self.timeout
```
potiuk: > - `time.monotonic()` for durations, not `time.time()`.

(`AGENTS.md` → Coding Standards)

This loop is entirely in-process, so `time.monotonic()` is the standard-compliant choice and is immune to a wall-clock step (NTP correction, manual clock change) silently shortening or extending the window. It'd be `time.monotonic()` on both this line and the `time.time() >= end_time` check below.

The trigger's `end_time` / `execution_deadline` are a different case and correctly stay on `time.time()` — those are absolute deadlines that get serialized and re-read in another process, where a monotonic value would be meaningless.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
```diff
@@ -137,6 +117,8 @@ def __init__(
         drop_on_completion: bool = True,
         poll_interval: int = 10,
         snowflake_conn_id: str = "snowflake_default",
+        deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
+        timeout: int = 24 * 60 * 60,
```
potiuk, line 121: This default also changes the existing non-deferrable path: `_poll_for_status` polled indefinitely in 6.15.0 / 6.16.1, and now a job still running at 24h fails the task and — with the default `drop_on_completion=True` — drops the service out from under it.

That's a sensible default, but it's a behaviour change for anyone already running SPCS jobs longer than a day, and it isn't visible from the PR title. Airflow's provider release managers regenerate `changelog.rst` from `git log`, so a line in `providers/snowflake/docs/changelog.rst` (or at minimum a sentence in the commit body) is the only place it'll get flagged for them.

potiuk (approved):
LGTM — all four points from my last pass are addressed: the deadline now derives from `ti.start_date`, `TERMINAL_STATUSES` / `NON_TERMINAL_STATUSES` keep their original names and stay importable from the operator module, both timeout paths log container output before dropping and respect `drop_on_completion`, and the statuses moved out of the SQL hook module into `utils/snowpark_containers.py`. CI is green and every thread is resolved. Four smaller observations inline, none blocking.

### Smaller observations

- `triggers/snowpark_containers.py:159` — `BaseTrigger.on_kill()` only exists from Airflow 3.3.0; across the 2.11–3.2 range this provider still supports, nothing calls it, so a killed deferred task leaves the service running. Worth a line in the docs.
- `operators/snowpark_containers.py:243` — the success-path drop used to be unguarded; routing it through `_drop_service` makes a failed drop silent.
- `operators/snowpark_containers.py:194` — `time.monotonic()` rather than `time.time()` for the in-process poll window (`AGENTS.md`); the trigger's epoch deadlines correctly stay on `time.time()`.
- `operators/snowpark_containers.py:121` — the new 24h default also caps the pre-existing synchronous path; worth a changelog line, since providers' changelogs are generated from `git log`.

---

>

---

providers/snowflake/src/airflow/providers/snowflake/triggers/snowpark_containers.py
justinpakzad: Good point. Added a note in the `snowpark_containers.rst`.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Good call out. The switch to best-effort was intentional and I'm aware it does change the current behavior but I think it's the better approach here. A job that already succeeded could be marked as failed (due to a transient error while dropping) and end up retrying an already successful job. I put a comment explaining why.

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Good call, I've updated `_poll_for_status` to use `time.monotonic()`. Also updated one of the tests since it was using time machine which doesn't seem to mock `time.monotonic()`

providers/snowflake/src/airflow/providers/snowflake/operators/snowpark_containers.py
justinpakzad: Good call. I've added a line to the change log.

justinpakzad:
> LGTM — all four points from my last pass are addressed: the deadline now derives from `ti.start_date`, `TERMINAL_STATUSES` / `NON_TERMINAL_STATUSES` keep their original names and stay importable from the operator module, both timeout paths log container output before dropping and respect `drop_on_completion`, and the statuses moved out of the SQL hook module into `utils/snowpark_containers.py`. CI is green and every thread is resolved. Four smaller observations inline, none blocking.


Appreciate the second round of review. All good callouts and they've been addressed.

providers/snowflake/tests/unit/snowflake/operators/test_snowpark_containers.py
```diff
@@ -150,6 +168,32 @@ def test_poll_waits_through_pending_then_done(self, mock_hook_cls, mock_sleep):
         assert op._poll_for_status() == "DONE"
         assert mock_sleep.call_count == 2
 
+    @pytest.mark.parametrize(
+        ("drop_on_completion", "drops"),
+        [(True, True), (False, False)],
+    )
+    @mock.patch("time.sleep")
+    @mock.patch("time.monotonic", side_effect=[0, 0, 20])
```
potiuk: `side_effect=[0, 0, 20]` supplies exactly the three values `_poll_for_status` consumes today: one for `end_time`, one for the check that must be under the deadline, one for the check that must be over it. It works, and the parametrized drop/no-drop split still exercises both cleanup paths.

The fragility is that `mock.patch("time.monotonic", ...)` patches the clock **process-wide** for the whole test body, not just the call under test. Anything else that reads the monotonic clock while the patch is active — `_make_operator`'s `BaseOperator.__init__`, a logging handler, or simply a fourth call if the loop ever grows one — raises `StopIteration` from somewhere with no obvious connection to the assertion.

An unbounded counter reads the same and cannot exhaust:

```suggestion
    @mock.patch("time.monotonic", side_effect=itertools.count(0, 20))
```

(with `import itertools` at the top). Same three values consumed, same behaviour, no coupling to the exact call count.

potiuk (approved):
LGTM — all four points from my last pass are addressed:

- `_poll_for_status` is on `time.monotonic()` for both the deadline and the check, while the trigger's serialized epoch deadlines correctly stay on `time.time()`.
- The deferrable `on_kill` version gap is documented in `snowpark_containers.rst`.
- The success-path drop keeps its best-effort behaviour with a comment giving the reason — a transient drop error would otherwise fail an already-successful job and retry it. That is the right call, and it is now legible to the next reader.
- The 24h default is called out as a breaking change directly below the `Changelog` header, which is exactly where that file's own instructions put breaking-change notes.

CI is green (56 passed, 28 skipped), no threads are open, and the branch merges clean.

One small observation, not blocking.

### Brittle clock mock (`providers/snowflake/tests/unit/snowflake/operators/test_snowpark_containers.py:176`)

Details inline. Dropping `time_machine` here was correct — it does not patch the monotonic clock, which is precisely why the test needed changing.

---

>

PR #70114, opened 2026-07-20
Fix ClickHouse test_connection unit test failing with newer clickhouse-connect

henry3260:
CI error : https://github.com/apache/airflow/actions/runs/29713514873

The assertion now checks only the SQL that gets sent, which is what the test is actually about, so it passes across the whole supported clickhouse-connect range.

---

### DIFF
--- a/providers/clickhousedb/tests/unit/clickhousedb/hooks/test_clickhouse.py
+++ b/providers/clickhousedb/tests/unit/clickhousedb/hooks/test_clickhouse.py
@@ -1147,10 +1147,13 @@ class TestClickHouseHookTestConnection:
         hook = ClickHouseHook(clickhouse_conn_id="clickhouse_test")
         hook.test_connection()
 
-        # Assert only on the SQL: the extra arguments clickhouse_connect's Cursor
-        # passes through vary between releases.
+        # Assert on the statement and the bound parameters only. Anything the
+        # driver appends beyond those (e.g. a settings kwarg) is its own concern
+        # and varies between releases.
         mock_client.query.assert_called_once()
-        assert mock_client.query.call_args.args[0] == "SELECT 1"
+        sql, parameters = mock_client.query.call_args.args[:2]
+        assert sql == "SELECT 1"
+        assert not parameters
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Lee-W (commented):
I'm semi-okay with this, but I'd like to see where this changed.

henry3260:
> I'm semi-okay with this, but I'd like to see where this changed.

- 1.4.x: ‎`query_result = self.client.query(operation, parameters)`

- 1.5.0: ‎`query_result = self.client.query(operation, parameters, settings=settings)`

FYI: https://github.com/ClickHouse/clickhouse-connect/releases/tag/v1.5.0

phanikumv (approved):
LGTM, because in every version, operation is passed positionally as the first arg
The test broke because the new API expects `settings` kwarg

Lee-W:
found it https://github.com/ClickHouse/clickhouse-connect/blob/47dbad3d09996438f701b68e48ffd8dd7da133f8/clickhouse_connect/dbapi/cursor.py#L62.

I think it would be better to check the second parameter as well, since it's technically controlled by Airflow. Even though we don't currently pass it to `test_connection`, we might do so in the future—or accidentally start doing so.

shahar1 (commented):
I wonder whether we should delete this test entirely, as it tests an internal behavior:

- The owned fact: `_test_connection_sql == "SELECT 1"` is already pinned directly by `test_test_connection_sql`.
- The behavior - "test_connection() succeeds end-to-end" is already pinned by `test_test_connection_success`.
- The current test's only distinct claim is "the literal SQL reaches the driver," which is unobservable without reaching into the internals (client.query).

What do you think?

henry3260:
> I wonder whether we should delete this test entirely, as it tests an internal behavior:
> 
> * The owned fact: `_test_connection_sql == "SELECT 1"` is already pinned directly by `test_test_connection_sql`.
> * The behavior - "test_connection() succeeds end-to-end" is already pinned by `test_test_connection_success`.
> * The current test's only distinct claim is "the literal SQL reaches the driver," which is unobservable without reaching into the internals (client.query).
> 
> What do you think?

Thanks for laying it out — I'd lean toward keeping it, tightened along the lines Wei suggested.
Per Wei's point, the first two positional arguments are ours — the SQL and the query parameters — so asserting on those is asserting our own contract, not the driver's. 
does that make sense to you?

eladkal:
cc @BentsiLeviav as the Clickhouse expert

shahar1 (approved):

eladkal (approved):

PR #69308, opened 2026-07-03
Add eval harness for testing AGENTS.md changes

RoyLee1224:
## What

Adds an eval harness (`dev/skill-evals/`) for testing AGENTS.md guidance against real scenarios. It answers: "does my AGENTS.md change actually affect agent behavior?"

```bash
uv run dev/skill-evals/eval.py --repeat 3
```

Compares the `main` branch AGENTS.md against your working tree. Each arm is a git worktree with the full repo, so the agent sees real source files. If AGENTS.md is unchanged, the working arm is skipped automatically.

No API key needed — authenticates via Claude Code OAuth (`claude /login`).

## Demo: testing the newsfragment golden rule
Following [the discussion in #release-management](https://apacheairflow.slack.com/archives/C03G9H97MM2/p1782814415838669)
Tested the golden rule (#67982) against real PRs where reviewers asked to remove newsfragments:

| Case | with AGENTS.md | without |
|------|:-:|:-:|
| Provider pod leak [#67333](https://github.com/apache/airflow/pull/67333#discussion_r3292590738) | 3/3 ✓ | 0/3 |
| API optimization #66696 | 3/3 ✓ | 1/3 |
| Scheduler fix [#64322](https://github.com/apache/airflow/pull/64322#discussion_r3032123197) | 2/3 | 0/3 |
| i18n cache fix [#65720](https://github.com/apache/airflow/pull/65720#discussion_r3138464983) | 1/3 | 0/3 |
| Worker connection testing [#62343](https://github.com/apache/airflow/pull/62343) (positive — should create) | 3/3 ✓ | 3/3 |
| Coordinator layer [#65958](https://github.com/apache/airflow/pull/65958) (positive — should create) | 3/3 ✓ | 3/3 |

Golden rule works for clear cases but struggles with ambiguous fixes where the model reasons "this bug affects users."

**Open question:** I'm not sure the current case selection is the right design. Would appreciate maintainers' thoughts on what cases are worth covering.

### cli demo
CleanShot 2026-07-02 at 00 57 34@2x

### UI demo(run `npx promptfoo@0.121.17 view`):
ui demo

## How it works

1. Creates git worktrees — one with `main`'s AGENTS.md, one with your working tree version. Both are full repo checkouts.
2. Generates a [promptfoo](https://github.com/promptfoo/promptfoo) config with `anthropic:claude-agent-sdk` provider and structured JSON output.
3. Runs each case against all arms in parallel, reports diff.
4. Worktrees cleaned up on exit.

## Files

[...]

PR #70120, opened 2026-07-20
Evaluate AGENTS.md and skills with Codex

RoyLee1224:
cc: @jason810496 

Follow up on #69308 by adding Codex as an alternative runtime for evaluating AGENTS.md and skills through promptfoo's built-in Codex SDK provider.

Run the evaluation with Codex:

  ```
  AGENT_RUNTIME=codex \
    prek run run-skill-eval --hook-stage manual --all-files
  ```

  The evaluation completed all six cases with no provider errors. Four behavioral assertions passed and two existing ambiguous newsfragment cases failed.

CleanShot 2026-07-20 at 17 26 29@2x
CleanShot 2026-07-20 at 17 26 48@2x

 

---

- [x] Yes gpt-5.6-sol

---

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @RoyLee1224** · by `@potiuk` · 2026-07-28 16:32 UTC
>
> Some review feedback from `jason810496` is waiting on you:
> - There are **5 unresolved review threads** on this PR from `jason810496`.
>
> **The ball is in your court** — you've been assigned to this PR. Push a fix or reply in each thread explaining why the feedback does not apply, then mark them resolved and ping the reviewer (`jason810496`) for a final look.
>
>
> **Note:** your branch is **225 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>


dev/skill-evals/eval.py
```diff
@@ -102,11 +109,17 @@ def find_sdk_modules() -> Path:
...
                 pf_pkg = pf_pkg.parent
+            package_parts = sdk_package.split("/")
             for candidate in (pf_pkg / "node_modules", pf_pkg.parent):
-                if (candidate / "@anthropic-ai" / "claude-agent-sdk").is_dir():
+                package_json = candidate.joinpath(*package_parts, "package.json")
+                try:
+                    installed_version = json.loads(package_json.read_text())["version"]
+                except (OSError, KeyError, ValueError):
+                    continue
+                if installed_version == expected_version:
```
jason810496: Does this means that we force the contributor to use the exact same version we pin?

.pre-commit-config.yaml
```diff
@@ -1019,7 +1019,10 @@ repos:
         entry: ./dev/skill-evals/eval.py
         language: node
         language_version: '22.22.0'
-        additional_dependencies: ['promptfoo@0.121.17', '@anthropic-ai/claude-agent-sdk@0.3.185']
+        additional_dependencies:
+          - 'promptfoo@0.121.17'
+          - '@anthropic-ai/claude-agent-sdk@0.3.185'
+          - '@openai/codex-sdk@0.144.6'
```
jason810496, line 1049: Or does it means we will use the Claude / Codex install by prek? Would it be possible to relex as to use system wise CLI first then fallback to prek installed?

.pre-commit-config.yaml
```diff
@@ -1019,7 +1019,10 @@ repos:
         entry: ./dev/skill-evals/eval.py
         language: node
         language_version: '22.22.0'
-        additional_dependencies: ['promptfoo@0.121.17', '@anthropic-ai/claude-agent-sdk@0.3.185']
```
jason810496, line 1028: I think it's fine for promptfoo as most of the user won't have it but I think we can assume most of the contributor have either Claude or Codex. Since we will pull ~310–350 MB Codex CLI binary, including Claude-only users now.

dev/skill-evals/eval.py
```diff
@@ -207,20 +220,55 @@ def remove_worktree(wt_dir: Path) -> None:
...
         "output_format": OUTPUT_FORMAT,
     }
+    if model:
+        config["model"] = model
     if skill_name:
         config["skills"] = [skill_name]
     return {"id": "anthropic:claude-agent-sdk", "label": label, "config": config}
 
 
+def build_codex_provider(
```
jason810496, line 247: Not a real comment but a interesting finding by claude.
Fortunately, our next steps is shirking the AGENTS.md haha.

---

Codex silently truncates AGENTS.md at 32 KiB, and Airflow's AGENTS.md is already 35,417 bytes, so a Codex-runtime eval can measure nothing while recording proof it ran (CONFIRMED)
Codex's DEFAULT_PROJECT_DOC_MAX_BYTES is 32,768 and codex-rs prefix-truncates project docs silently; build_codex_provider sets cli_config only for history persistence and never raises project_doc_max_bytes, even though promptfoo spreads cli_config straight into Codex config overrides. Since arms differ only in AGENTS.md, any guidance edit landing past the 32,768-byte boundary (the file's last ~2.6 KB, where new guidance is typically appended) produces byte-identical visible guidance in both arms — the exact "eval silently measures nothing" failure mode check_claude_md_symlink guards against on the Claude side, and the hash is still recorded as proof. Fix is one line: add project_doc_max_bytes to cli_config.

dev/skill-evals/eval.py
```diff
@@ -45,6 +46,11 @@
 from pathlib import Path
 
 PROMPTFOO_VERSION = "0.121.17"
+SDK_PACKAGES = {
+    "claude": ("@anthropic-ai/claude-agent-sdk", "0.3.185"),
+    "codex": ("@openai/codex-sdk", "0.144.6"),
+}
```
jason810496, line 54: Or another direction, we could define the lower bound for the CLI version. (Thought I have no idea how should we pin the version at this moment).

jason810496 (commented):
Thanks for adding the Codex support!

dev/skill-evals/eval.py
RoyLee1224: Yes, but prek installs the pinned version automatically, so contributors do not select or install it themselves.

I removed the version comparison. The code now selects the SDK installed by the hook instead of promptfoo's bundled copy.

.pre-commit-config.yaml
RoyLee1224: Good point. I split Codex into a separate manual hook. The existing Claude hook is unchanged, so Claude-only users get no additional download from this PR.

The Codex hook installs a newer SDK because promptfoo's bundled Codex binary is signed with a revoked certificate and macOS blocks it. This is also reported in [openai/codex #22135](https://github.com/openai/codex/issues/22135).

dev/skill-evals/eval.py
RoyLee1224: Nice catch, thanks! The Codex provider now raises `project_doc_max_bytes` so the complete `AGENTS.md` is included in the evaluation.

.pre-commit-config.yaml
RoyLee1224: It is technically possible, but prek would still install the fallback before the hook runs, so it would not reduce the download. It would also make the eval depend on an unpinned system CLI.

I prefer keeping the CLI managed by the separate manual prek hook for reproducible runs. Authentication still comes from the user's existing session.

dev/skill-evals/eval.py
RoyLee1224: I removed the comparison instead of adding a lower bound. A lower bound could still select promptfoo's bundled SDK, while `additional_dependencies` already defines the version used by the hook.

.pre-commit-config.yaml
```diff
@@ -1034,6 +1034,27 @@ repos:
...
+        # bundles its own @openai/codex-sdk, but that build ships a binary whose
+        # Developer ID certificate Apple has revoked — macOS kills it on exec —
+        # so a notarized version has to be installed explicitly.
+        entry: env AGENT_RUNTIME=codex ./dev/skill-evals/eval.py
+        language: node
+        language_version: '22.22.0'
+        additional_dependencies:
+          - 'promptfoo@0.121.17'
+          - '@openai/codex-sdk@0.144.6'
+        stages: ['manual']
```
jason810496, line 1050: `manual` is also the stage used by the repository-wide bulk command, so this hook still matches every `AGENTS.md` / `dev/skill-evals/` change. `prek` will therefore provision the ~300 MB Codex environment and then require Codex authentication for contributors who intended to run only Claude.

After rebasing onto `main`, please extend the existing bulk-manual skip list in the generated command source and its consumers:

```diff
-prek run --from-ref <target_branch> --stage manual --skip compile-ui-assets-dev --skip view-skill-eval
+prek run --from-ref <target_branch> --stage manual --skip compile-ui-assets-dev --skip view-skill-eval --skip run-skill-eval-codex
```

That means updating `contributing-docs/08_static_code_checks.rst`, the generated command in `AGENTS.md`, and the pre-push checklist command in `AGENTS.md`. The explicit `prek run run-skill-eval-codex ...` path documented in `dev/skill-evals/README.md` will remain opt-in.

scripts/tests/ci/prek/test_skill_eval.py
```diff
@@ -0,0 +1,174 @@
...
+        "apiKeyRequired": False,
+        "append_allowed_tools": ["Read", "Grep", "Glob"],
+        "model": "claude-test",
+        "output_format": skill_eval_module.OUTPUT_FORMAT,
+        "setting_sources": ["project"],
+        "skills": ["example-skill"],
+        "working_dir": str(tmp_path),
+    }
+
+
```
jason810496, line 118: This function exercises the pre-existing Claude branch unchanged by this PR. It still passes if the production change is reverted, so it conflicts with the repository requirement that each added test fail without the PR. Please remove it; the Codex/runtime tests above cover the new behavior.

```suggestion
```

jason810496 (commented):
LGTM after resolving the following comments, thanks.

.pre-commit-config.yaml
RoyLee1224: Thanks for the review! Rebased onto main and added `--skip run-skill-eval-codex` to the bulk manual command in all three places

jason810496 (approved):
Thanks.

PR #70124, opened 2026-07-20
Reject zero step in mapped XCom slices

### LINKED ISSUE #70123: Mapped XCom slice endpoint accepts a zero step

Airflow Core

### Apache Airflow version

`main`

The Execution API endpoint for reading a mapped XCom sequence slice accepts `step=0` and silently treats it as `step=1`. A zero slice step is invalid sequence semantics, so this hides malformed requests from Task SDK clients.

Steps to reproduce:

1. Start the Airflow development environment from the `main` branch.
2. Send the following request through an authenticated Execution API client:

   ```http
   GET /execution/xcoms/dag/runid/task/xcom_1/slice?step=0
   ```

3. Observe that the endpoint responds with HTTP 200 instead of rejecting the invalid step:

   ```json
   []
   ```

The behavior comes from using `params.step or 1`, which converts the valid integer value `0` to the default step of `1`.

The current Execution API version should reject `step=0` with HTTP 422 before querying XCom data. Older
released API versions should retain their existing behavior through the Execution API versioning layer.
Omitted, positive, and negative step values should remain unchanged.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

viiccwen:
closes #70123 

A mapped XCom slice with `step=0` is invalid, but the endpoint silently treated it as `step=1`, hiding malformed client requests.

This change returns HTTP 422 with an `invalid_step` response for every Execution API version. The default step of `1` is now applied only when the parameter is omitted, preserving positive and negative step behavior.

---

- [x] Yes - GPT-5.6 Sol

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
@@ -190,6 +190,12 @@ def get_mapped_xcom_by_slice(
     params: Annotated[GetXComSliceFilterParams, Query()],
     session: SessionDep,
 ) -> XComSequenceSliceResponse:
+    if params.step == 0:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail={"reason": "invalid_step", "message": "XCom slice step cannot be zero."},
+        )
+
     query = XComModel.get_many(
         run_id=run_id,
         key=key,
@@ -199,7 +205,7 @@ def get_mapped_xcom_by_slice(
     )
     query = query.order_by(None)
 
-    step = params.step or 1
+    step = params.step if params.step is not None else 1
 
     # We want to optimize negative slicing (e.g. seq[-10:]) by not doing an
     # additional COUNT query if possible. This is possible unless both start and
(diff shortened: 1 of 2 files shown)


viiccwen:
cc @jason810496, @henry3260. 🙌

henry3260 (changes requested):
Thanks for the fix!

IMO, I don't think this needs a new API version migration though. The old behavior (silently treating `step=0` as `step=1`) is exactly the bug being fixed — preserving it for clients declaring older API versions means their malformed requests keep being silently misinterpreted, which is what #70123 wants to surface.

viiccwen:
> Thanks for the fix!
> 
> IMO, I don't think this needs a new API version migration though. The old behavior (silently treating `step=0` as `step=1`) is exactly the bug being fixed — preserving it for clients declaring older API versions means their malformed requests keep being silently misinterpreted, which is what #70123 wants to surface.

Sure, that makes sense. I just updated.

viiccwen:
Hello @henry3260, politely tag for further review.

PR #67381, opened 2026-05-23
Add API endpoint for backfill dag run entries

shivaam:
Adds a new public API endpoint `GET /backfills/{backfill_id}/dag_runs` that returns
the `BackfillDagRun` entries for a given backfill with joined `DagRun` state. Users
can see what happened in a backfill: which dates ran, their states, and which slots
were skipped (with reason).

**Previous PR:** #64089 explored enriching `BackfillResponse` directly. Based on
feedback from @uranusjr, this PR takes the cleaner approach of a dedicated
sub-resource endpoint.

### Changes

- `BackfillDagRunResponse` / `BackfillDagRunCollectionResponse` Pydantic models
  with `AliasPath` for joined `DagRun` fields (matching existing `BackfillResponse` pattern)
- `list_backfill_dag_runs()` route using `joinedload(BackfillDagRun.dag_run)` — LEFT OUTER JOIN
  includes skipped slots where `dag_run_id` is NULL
- Pagination via `limit`/`offset`, default ordering by `sort_ordinal`
- 404 when backfill doesn't exist
- 8 unit tests: happy path, skipped slots, 404, pagination, empty backfill, ordering contract

### Response shape

```json
{
  "backfill_dag_runs": [
    {
      "id": 1,
      "backfill_id": 1,
      "dag_run_id": 42,
      "logical_date": "2024-01-01T00:00:00Z",
      "partition_key": null,
      "sort_ordinal": 1,
      "exception_reason": null,
      "dag_run_state": "success",
      "dag_run_run_id": "backfill__2024-01-01T00:00:00+00:00"
    },
    {
      "id": 2,
      "backfill_id": 1,
      "dag_run_id": null,
      "logical_date": "2024-01-02T00:00:00Z",
      "partition_key": null,
      "sort_ordinal": 2,
      "exception_reason": "already exists",
      "dag_run_state": null,
      "dag_run_run_id": null
    }
  ],
  "total_entries": 10
}
```

### Follow-up PRs

- UI: Progress bar on backfills table/banner + backfill detail page with Dag runs table
- CLI: `airflow backfill list` and `airflow backfill list-runs` commands

closes: #46250

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (claude-opus-4-6)

Generated-by: Claude Code (claude-opus-4-6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

[...]

PR #70126, opened 2026-07-20
Guard the backfill dag-runs list query count and tidy test params

pierrejeambrun:
Follow-up to #67381.

The `GET /backfills/{backfill_id}/dag_runs` endpoint added in #67381 relies on `joinedload` to keep its query count constant, but its tests didn't assert that — so a change that drops the eager-load could silently reintroduce an N+1.

- Add an `assert_queries_count(5)` guard on the happy-path test. Verified the count is constant: 2 dag runs and 6 dag runs both issue 5 queries, so nothing scales with the number of rows.
- Let the test client encode the query parameters from a native `dict` (`params={...}`) instead of hand-building `?limit=...&offset=...` / `?order_by=...` strings, so the client handles the encoding.

Test-only change.

---


vatsrahul1001 (approved):

vatsrahul1001:
LGTM! Thanks for updating test

Issue #39002, opened 2024-04-14
Allow passing airflow params as job parameter in databricks job


SubhamSinghal:
### Description

Currently if we need to pass any parameter to spark job we need to hardcode parameters in json config. But we want to pass parameters dynamically through airflow.
This feature allows to pass airflow config paramaters as job parameters in spark job.

### Use case/motivation

Dynamically update job parameter in spark job

### Related issues

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

Seoji:
I think you are using [DatabricksRunNowOperator](https://github.com/apache/airflow/blob/0f4884c3ca26c39bc6bd21967c950e718589fcd6/airflow/providers/databricks/operators/databricks.py#L687). 
If So, How about using notebook_params field in Operator's json parameter?
In databricks, you can use this parameter through dbutils.widgets.get.

potiuk:
@SubhamSinghal We are unassigning you from this issue as part of our updated [assignment policy](https://github.com/apache/airflow/blob/main/contributing-docs/04_how_to_contribute.rst#contribute-code-changes).

This is not meant to discourage your contribution — quite the opposite! You are still very welcome to work on this issue and submit a PR for it. Simply comment that you are working on it and open a PR when ready.

We found that formal assignments were not working well, as they often prevented others from contributing when the assignee was not actively working on the issue.

eladkal:
cc @moomindani not sure if this one is relevant but if is sounds like a very easy fix?

moomindani:
Hi @eladkal — taking this one on. Opened #66613 which builds on @SubhamSinghal's earlier work in #39007 (closed as stale). Picks up @dirrao's and @Lee-W's review feedback (list-comprehension refactor + use of `params.dump()` to resolve `Param` objects to their values) and extends the auto-injection from `DatabricksCreateJobsOperator` to `DatabricksRunNowOperator` (`job_parameters`) and `DatabricksSubmitRunOperator` (per-task dict-shaped slots) per @galafis's suggestion. Verified end-to-end against a real Databricks workspace.

PR #70130, opened 2026-07-20
Providers: Allow opting out of forwarding Dag-level parameters in DatabricksRunNowOperator

### LINKED ISSUE #70121: DatabricksRunNowOperator always pass parameters to Databricks Job

Providers

### Apache Airflow version

3.3.0

The change introduced by https://github.com/apache/airflow/pull/66613 (for https://github.com/apache/airflow/issues/39002) makes DatabricksRunNowOperator always include DAG-level params as job_parameters. Since apache-airflow-providers-databricks 7.16.0, users cannot opt out of passing those parameters, which creates compatibility issues for Databricks jobs whose entry point does not expect or accept them.

DatabricksRunNowOperator should provide a way to opt out of forwarding DAG-level params as job_parameters. Automatically injecting self.params changes the job payload and can break existing jobs that do not expect those parameters. The default behavior should either remain unchanged for backward compatibility or provide an explicit flag controlling parameter propagation.

### Operating System

_No response_

### Deployment

databricks

apache-airflow-providers-databricks 7.16.0

1.21.0

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

onlyarnav:
### Description

This PR introduces an option to opt out of forwarding Dag-level parameters in the `DatabricksRunNowOperator`. 

In PR #66613 (for issue #39002), parameter forwarding was introduced such that the operator's `params` dict is automatically forwarded as `job_parameters` when no `job_parameters` are specified. However, this changes the payload for Databricks jobs and breaks execution for jobs whose entry points do not expect or accept additional parameters.

To resolve this compatibility issue, this PR:
1. Adds a new boolean parameter `forward_dag_params` (default: `True`) to `DatabricksRunNowOperator`.
2. Checks `self.forward_dag_params` when constructing the payload inside `_build_run_now_payload` before merging `self.params`.
3. Adds unit tests to verify that Dag-level parameters are not forwarded when `forward_dag_params=False`.
4. Fixes minor Windows-specific issues (e.g. file encoding, `os.register_at_fork`, and a mock `fcntl`) to enable executing the local test suite on Windows development hosts.

closes: #70121

---

- [X] Yes (Google Antigravity)

Generated-by: Google Antigravity following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


providers/databricks/tests/conftest.py
jroachgolf84, line 1: What are these changes for? Why are they needed?

scripts/ci/prek/common_prek_utils.py
jroachgolf84, line 1: Should this change be part of this PR?

shared/observability/src/airflow_shared/observability/metrics/stats.py
jroachgolf84, line 1: Should this change be part of this PR?

jroachgolf84 (commented):
Please review my comments, specifically, there are a few changes in there I don't think should be in there.

providers/databricks/tests/conftest.py
```diff
@@ -16,4 +16,18 @@
...
+if os.name == "nt":
+    from types import ModuleType
+
+    fcntl_mock = ModuleType("fcntl")
+    fcntl_mock.LOCK_SH = 1
+    fcntl_mock.LOCK_EX = 2
+    fcntl_mock.LOCK_NB = 4
+    fcntl_mock.LOCK_UN = 8
+    fcntl_mock.flock = lambda fd, operation: None
+    sys.modules["fcntl"] = fcntl_mock
```
amoghrajesh: What does this do?

shared/observability/src/airflow_shared/observability/metrics/stats.py
```diff
@@ -48,7 +48,8 @@ def _reset_backend_after_fork() -> None:
     _backend = None
 
 
-os.register_at_fork(after_in_child=_reset_backend_after_fork)
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(after_in_child=_reset_backend_after_fork)
```
amoghrajesh: Same here.

amoghrajesh (dismissed):
@onlyarnav there is some unrelated changes in your PR diff, please revise it and keep the changes _only_ scoped to this PR.

shared/observability/src/airflow_shared/observability/metrics/stats.py
onlyarnav: This change was originally added as a local environment workaround to prevent startup crashes when executing the tests on a Windows host (since os.register_at_fork is POSIX-only). I have reverted it.

scripts/ci/prek/common_prek_utils.py
onlyarnav: This was also a local environment workaround to prevent UnicodeDecodeError when running pre-commit checks on a Windows machine. I have reverted it from this PR

shared/observability/src/airflow_shared/observability/metrics/stats.py
onlyarnav: Reverted

providers/databricks/tests/conftest.py
onlyarnav: Since fcntl is a POSIX-only module, airflow 3.0+ imports it during test execution setup, which crashes test collection on Windows. This mock was added so the test suite could be run and verified locally on a Windows machine. It was added by Antigravity, but I have reverted it.

providers/databricks/tests/conftest.py
onlyarnav: Since fcntl is a POSIX-only module, airflow 3.0+ imports it during test execution setup, which crashes test collection on Windows hosts. This mock was added so the test suite could be run and verified locally on a Windows machine. It was added by Antigravity, but I have reverted it.

jroachgolf84 (approved):
Thanks for the changes, LGTM.

eladkal:
cc @moomindani for a review from Databricks team

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
```diff
@@ -1315,7 +1319,7 @@ def _build_run_now_payload(self) -> dict[str, Any]:
             json["job_id"] = job_id
             del json["job_name"]
 
-        if not json.get("job_parameters") and self.params:
+        if self.forward_dag_params and not json.get("job_parameters") and self.params:
```
moomindani: Consider also skipping injection when a conflicting slot is already present.

Because `forward_dag_params` defaults to `True`, every Dag that is currently broken by #66613 stays broken until its author discovers this new flag — and the failure mode gives no hint that a new operator argument is the fix. The conflict is deterministic and fully knowable from the payload at this point, so it can be repaired automatically:

```python
_RUN_NOW_PARAM_SLOTS_CONFLICTING_WITH_JOB_PARAMETERS = (
    "notebook_params",
    "python_params",
    "jar_params",
    "spark_submit_params",
    "python_named_params",
    "dbt_commands",
)

if (
    self.forward_dag_params
    and not json.get("job_parameters")
    and self.params
    and not any(k in json for k in _RUN_NOW_PARAM_SLOTS_CONFLICTING_WITH_JOB_PARAMETERS)
):
    json["job_parameters"] = dict(self.params)
```

That fixes the regression for existing users without requiring them to change anything, and `forward_dag_params=False` then covers the remaining case the issue describes — a job whose entry point rejects extra params even when there is no slot conflict.

I'm happy to be argued out of this if maintainers prefer to keep the change minimal and purely opt-in; the silent-failure-by-default aspect is what makes me raise it.

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
```diff
@@ -1167,6 +1167,8 @@ class DatabricksRunNowOperator(ResumableJobMixin, BaseOperator):
         before polling begins so that a worker crash and retry reconnects to the existing run
         instead of triggering a duplicate run of the same job. Set to ``False`` to always trigger a
         fresh run on retry. Requires Airflow 3.3+; on earlier versions it is silently ignored.
+    :param forward_dag_params: Whether to forward Dag-level params as ``job_parameters``
```
moomindani, line 1180: The `.. note::` a few lines below still states forwarding happens unconditionally, so the two now contradict each other. Worth adding the opt-out to the note, e.g. append: "Set ``forward_dag_params=False`` to disable this."

Also worth noting there that `job_parameters` cannot be combined with `notebook_params` / `python_params` / `jar_params` / `spark_submit_params` / `python_named_params` / `dbt_commands` — the API rejects such a run, which is the main reason a user would reach for this flag.

providers/databricks/tests/unit/databricks/operators/test_databricks.py
```diff
@@ -2925,6 +2925,28 @@ def test_run_now_does_not_override_existing_job_parameters(self, db_mock_class):
         actual = db_mock.run_now.call_args.args[0]
         assert actual["job_parameters"] == {"explicit": "value"}
 
+    @mock.patch("airflow.providers.databricks.operators.databricks.DatabricksHook")
+    def test_run_now_does_not_inject_airflow_params_when_forward_dag_params_is_false(self, db_mock_class):
```
moomindani, line 2929: The test correctly covers the flag's mechanics (I verified it fails without the guard). What isn't covered is the scenario that actually motivates the change: Dag-level `params` combined with one of the legacy param slots, which is what the API rejects. Nothing in the existing suite exercises that combination.

Worth adding a case asserting that with `notebook_params` set and `forward_dag_params=False`, the payload contains `notebook_params` and no `job_parameters` — that pins the regression scenario rather than just the flag. If you take the auto-skip suggestion on `_build_run_now_payload`, the same case parametrized over the six conflicting slots would cover it directly.

moomindani (commented):
Thanks for the fix, and thanks for cleaning up the unrelated Windows changes — the diff is nicely scoped now.

I validated this against a real Databricks workspace, and the regression is more severe than the issue describes. The `run-now` endpoint rejects `job_parameters` outright when combined with any of the legacy param slots. All six confirmed:

```
job_parameters can not be used in combination with notebook_params
job_parameters can not be used in combination with python_params
job_parameters can not be used in combination with jar_params
job_parameters can not be used in combination with spark_submit_params
job_parameters can not be used in combination with python_named_params
job_parameters can not be used in combination with dbt_commands
```

Since Dag-level `params` propagate to the operator implicitly, a Dag like this has been failing since 7.16.0 with no user action:

```python
with DAG("d", params={"env": "prod"}):
    DatabricksRunNowOperator(task_id="t", job_id=1, notebook_params={"foo": "bar"})
    # payload -> {"job_id": 1, "notebook_params": {...}, "job_parameters": {"env": "prod"}}  # API rejects
```

I also checked whether an escape hatch already existed: `job_parameters={}` does not work, because `not json.get("job_parameters")` treats an empty dict as falsy and injects `params` anyway. So the new flag is genuinely necessary — the approach is right.

Verification I ran:

- Reverted the guard in `_build_run_now_payload` → the new test fails. Restored → full operator suite passes (201 tests).
- `prek run --stage pre-commit` clean on the diff.

The direction is good. I left three inline notes — the first one is the substantive one (default behaviour still leaves existing users broken); the others are the docstring and test coverage.

One more docs point I could not anchor inline, because the file is not part of this diff: `providers/databricks/docs/operators/run_now.rst` (the "Forwarding Airflow Dag params as Databricks job parameters" section, around line 60) still documents forwarding as unconditional. That guide is where users land when they hit this failure, so the opt-out should be discoverable there too — a sentence naming `forward_dag_params=False`, plus the mutual-exclusivity constraint with the legacy param slots.

Non-blocking, out of scope for this PR: #66613 added the same forwarding to `DatabricksCreateJobsOperator` (`parameters`) and `DatabricksSubmitRunOperator` (per-task dict slots), which have no opt-out. Scoping this to `RunNow` matches the issue, so no change needed here — just flagging it in case the flag should be applied consistently in a follow-up.

---

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
onlyarnav: Updated! I've updated the .. note:: in the DatabricksRunNowOperator docstring as well as the user documentation in providers/databricks/docs/operators/run_now.rst to explain forward_dag_params=False and document the Databricks API conflict with the legacy parameter slots

providers/databricks/tests/unit/databricks/operators/test_databricks.py
onlyarnav: Done! I've added a parametrized test test_run_now_skips_param_injection_with_legacy_param_slots covering all 6 conflicting parameter slots to verify that job_parameters is not injected when any of them are present

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
onlyarnav: I've updated _build_run_now_payload to define _RUN_NOW_PARAM_SLOTS_CONFLICTING_WITH_JOB_PARAMETERS and automatically skip auto-injecting job_parameters whenever any of the 6 legacy parameter slots are present in json

moomindani (approved):
All three addressed, verified by running `90e726b6` rather than reading the summary. The auto-skip is implemented as suggested, and the four paths behave correctly:

```
Dag params + notebook_params      -> {job_id, notebook_params}      # regression repaired, no user change
explicit job_parameters + slot    -> both preserved                 # user intent respected
Dag params, no slot               -> {job_id, job_parameters}       # forwarding still works
forward_dag_params=False          -> {job_id}                       # flag still honoured
```

Keeping an explicitly-passed `job_parameters` rather than silently dropping it is the right call — only the implicit forwarding is skipped, so a user who deliberately combined them still gets the API error instead of a silent behaviour change. 207 tests pass, and the `test_execute_does_not_mutate_json_template_field` edit is a necessary consequence of the skip logic with its assertions intact.

I went back to the workspace to check the approach itself, and found the constraint is bidirectional — worth knowing, though it does not change my assessment:

| Job config | Sent | Result |
|---|---|---|
| has job-level `parameters` | `notebook_params` only | rejected: *"Cannot use legacy parameters (...) because the job has job parameters configured"* |
| has job-level `parameters` | `job_parameters` only | accepted |
| no job-level `parameters` | `notebook_params` only | accepted |
| either | both | rejected: *"job_parameters can not be used in combination with notebook_params"* |

So for a job that *does* declare job-level parameters, skipping the injection trades one rejection for the other. But that combination was already impossible before #66613, so the set of broken Dags is unchanged — and the case your fix actually repairs (job without job-level parameters + legacy slot + Dag params) is a genuine #66613 regression. The skip condition maps correctly onto the API constraint, and `forward_dag_params` still earns its place for jobs whose entry point rejects extra params even with no slot conflict. Approach looks right to me.

One optional docs nit, non-blocking: the new note says `job_parameters` cannot be combined with the legacy slots, which is accurate but narrower than the API's actual rule. A reader may conclude "then I'll just use `notebook_params`" — which also fails if the job declares job-level parameters. A clause like "and the legacy slots cannot be used at all against a job that declares job-level parameters" would close that gap. Fine to leave for a follow-up.

---

eladkal (approved):

PR #44477, opened 2024-11-29
feat: automatically inject OL info into spark job in DataprocSubmitJobOperator

kacpermuda:
This PR introduces a completely new feature to OpenLineage integration. **It will NOT impact users that are not using OpenLineage or have not explicitly enabled this feature (False by default).**

## TLDR; 
When explicitly enabled by the user for supported operators, we will automatically inject parent job information into the Spark job properties. For example, when submitting a Spark job using the DataprocSubmitJobOperator, we will include details about the Airflow task that triggered it so that the OpenLineage Spark integration can include them in parentRunFacet.

## Why ?

To enable full pipeline visibility and track dependencies between jobs in OpenLineage, we utilize the parentRunFacet. This facet stores the identifier of the parent job that triggered the current job. This approach works across various integrations, f.e. you can pass Airflow’s job identifier to a Spark application if it was triggered by an Airflow operator. Currently, this process requires manual configuration by the user, such as leveraging [macros](https://airflow.apache.org/docs/apache-airflow-providers-openlineage/stable/macros.html):
```
DataprocSubmitJobOperator(
    task_id="my_task", 
    # ... 
    job={ 
        # ...
        "spark.openlineage.parentJobNamespace": "{{ macros.OpenLineageProviderPlugin.lineage_job_namespace() }}",
        "spark.openlineage.parentJobName": "{{ macros.OpenLineageProviderPlugin.lineage_job_name(task_instance) }}", 
        "spark.openlineage.parentRunId": "{{ macros.OpenLineageProviderPlugin.lineage_run_id(task_instance) }}"
    } 
)

```
Understanding how various Airflow operators configure Spark allows us to automatically inject parent job information.

## Controlling the Behavior

We provide users with a flexible control mechanism to manage this injection, combining per-operator enablement with a global fallback configuration. This design is inspired by the `deferrable` argument in Airflow.

```python
ol_inject_parent_job_info: bool = conf.getboolean(
    "openlineage", "spark_inject_parent_job_info", fallback=False
)
```
Each supported operator will include an argument like `ol_inject_parent_job_info`, which defaults to the global configuration value of `openlineage.spark_inject_parent_job_info`. This approach allows users to:

1. Control behavior on a per-job basis by explicitly setting the argument.
2. Rely on a consistent default configuration for all jobs if the argument is not set.

[...]

PR #70131, opened 2026-07-20
Link dbt Cloud runs to the triggering Airflow task via OpenLineage

mobuchowski:
dbt Cloud does not natively propagate the OpenLineage parent run of the Airflow task that triggered a job, so a triggered run has no stored link back to that task. Any tool that reads dbt Cloud runs on their own therefore sees them as isolated jobs, with no way to attach them to the Airflow run that started them. This adds an `openlineage_inject_parent_job_info` option to `DbtCloudRunJobOperator`. When enabled, it writes the task's OpenLineage parent and root run identifiers, as JSON, into the triggered run's `cause` field, so a consumer reading dbt Cloud runs can parse them and rebuild the `ParentRunFacet` that links the run to its Airflow task.

It is conceptually the same thing we've added in #44477 #64513 or #64807 but using different, dbt cloud specific mechanism.

The option defaults to the existing `openlineage.spark_inject_parent_job_info` config value, matching the convention already used by the Glue and EMR operators. Because dbt Cloud limits `cause` to 255 characters, the payload is trimmed to fit: the `parent` identifiers are always kept, the `root` block is dropped if the two together are too long, and injection is skipped if even the parent alone does not fit. When injection is enabled the `cause` is replaced entirely with these identifiers rather than the human-readable trigger reason, since the correlation is the point of the option.

### DIFF
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
@@ -234,7 +234,7 @@ class DbtCloudRunJobOperator(BaseOperator):
             )
 
         if self.openlineage_inject_parent_job_info:
-            self.log.info("Injecting OpenLineage parent job information into dbt Cloud run cause.")
+            self.log.info("Replacing dbt Cloud run cause with OpenLineage parent job information.")
             self.trigger_reason = inject_parent_job_information_into_dbt_cloud_cause(
                 self.trigger_reason, context["ti"]
             )
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
@@ -31,7 +31,10 @@ from airflow.providers.dbt.cloud.hooks.dbt import (
     JobRunInfo,
 )
 from airflow.providers.dbt.cloud.triggers.dbt import DbtCloudRunJobTrigger
-from airflow.providers.dbt.cloud.utils.openlineage import generate_openlineage_events_from_dbt_cloud_run
+from airflow.providers.dbt.cloud.utils.openlineage import (
+    generate_openlineage_events_from_dbt_cloud_run,
+    inject_parent_job_information_into_dbt_cloud_cause,
+)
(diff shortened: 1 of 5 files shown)


providers/dbt/cloud/src/airflow/providers/dbt/cloud/utils/openlineage.py
```diff
@@ -97,6 +99,62 @@ def _get_parent_run_metadata(task_instance):
...
+        log.warning(
+            "Could not import OpenLineage provider. Skipping the injection of OpenLineage "
+            "parent job information into the dbt Cloud run cause."
+        )
+        return cause
+
+    candidates = [
+        {"parent": facet["parent"], "root": facet["root"]},
+        {"parent": facet["parent"]},
+    ]
```
kacpermuda, line 144: I know that every character counts since the limit is low, but maybe we should add encapsulating `{"openlineage": {"parent"...}}` dict? Without the word `openlineage`, parent and root can be confusing for reader. But, if somebody sets it to True, they usually know what they're doing. Something to consider.

providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
```diff
@@ -219,6 +233,12 @@ def execute(self, context: Context):
                 f"Triggered via Apache Airflow by task {self.task_id!r} in the {self.dag.dag_id} DAG."
             )
 
+        if self.openlineage_inject_parent_job_info:
+            self.log.info("Injecting OpenLineage parent job information into dbt Cloud run cause.")
```
kacpermuda:
```suggestion
            self.log.info("Replacing dbt Cloud run cause with OpenLineage parent job information.")
```
Let's be explicit here, since we replace and not inject (because the char limit on cause is so low)

kacpermuda (approved):
Left two comments, lgtm overall

providers/dbt/cloud/src/airflow/providers/dbt/cloud/utils/openlineage.py
mobuchowski: I think we're too constrained here unfortunately :(

eladkal (approved):

PR #70132, opened 2026-07-20
Fix output type loss in AgentOperator's human-in-the-loop review

ColtenOuO:
## Summary

`AgentOperator`'s human-in-the-loop review path (`enable_hitl_review=True`) loses the original output type for any `output_type` that is neither `str` nor a Pydantic `BaseModel` (e.g. `list[str]`, `bool`, `dict`).

The output is round-tripped through a string while it's shown to a human reviewer: `HITLReviewMixin._to_string` serializes it before review, and `AgentOperator.execute()` deserializes it back into `output_type` after approval. The serialization side used `str(output)`, which produces a Python repr (e.g. `"['tag-a', 'tag-b']"`), not valid JSON.

The deserialization side then tried `json.loads()` on that repr, which always raised, and silently fell back to returning the raw repr *string* instead of the original list/bool/dict — a type change downstream tasks don't expect.

`regenerate_with_feedback()` (used when a reviewer requests changes) had the identical `str(output)` bug.

This is the same defect class already fixed for the `require_approval=True` path in #70075, which switched to `TypeAdapter(...).dump_json(...)`. That fix wasn't applied to the separate `enable_hitl_review=True` path, so it regressed here.

## Changes

- Both directions of the review round-trip now live next to each other in
  `utils/output_type.py`: the existing `rehydrate_pydantic_output()` and a new
  `dump_output_to_json()`.
- `dump_output_to_json()` replaces the three copies of the serialization block
  that had drifted apart — `HITLReviewMixin._to_string`,
  `AgentOperator.regenerate_with_feedback` and `LLMApprovalMixin.defer_for_approval`.
  That drift is how the original bug survived #70075.
- `AgentOperator.execute()`'s HITL-review branch reuses `rehydrate_pydantic_output`
  instead of a bespoke `json.loads`/`except` fallback, so both review paths behave
  consistently.

## Guarding an `output_type` pydantic cannot build a schema for

`AgentOperator` passes `output_type` straight through to `Agent(...)`, so pydantic-ai's
`[A, B]` multi-output lists, `ToolOutput`/`NativeOutput`/`PromptedOutput` markers and
output functions all reach these helpers. `TypeAdapter` raises
`PydanticSchemaGenerationError` (a `RuntimeError`, not a `ValidationError`) or
`AttributeError` on those, which neither side's `except` clause caught:

- `rehydrate_pydantic_output` now separates schema build from validation and falls
  back to plain `json.loads`. Without this an already-approved output is lost, after
  the model call and after the reviewer has signed off.
- `dump_output_to_json` falls back to `str(output)`. This one fires before the session
  XCom push, so without it the task dies before the reviewer is ever shown a review
  session. It is reachable through output functions, which only warn
  ("Falling back to unconstrained schema") for a non-schema-able return type and then
  run fine.

## Behaviour change

With the default `output_type=str`, the approved string is no longer parsed as JSON:

| Approved string | Before | After |
| --- | --- | --- |
| `'42'` | `42` | `'42'` |
| `'{"text": "hi"}'` | `{'text': 'hi'}` | `'{"text": "hi"}'` |
| `'true'` | `True` | `'true'` |
| `'null'` | `None` | `'null'` |

This matches the non-HITL path (which returns `result.output` unparsed) and the
`require_approval` path. `str` is the default `output_type` and `enable_hitl_review`
has shipped since 0.1.0, so a Dag relying on the implicit parse needs an explicit
`output_type` (e.g. `output_type=int`) or its own `json.loads()` downstream.

---

- [X] Yes — Claude Code (Sonnet 5, Opus 5)

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
@@ -22,8 +22,7 @@ import logging
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any, Protocol
 
-from pydantic import BaseModel, TypeAdapter
-
+from airflow.providers.common.ai.utils.output_type import dump_output_to_json
 from airflow.providers.common.compat.version_compat import AIRFLOW_V_3_3_PLUS
 
 if AIRFLOW_V_3_3_PLUS:
@@ -123,11 +122,8 @@ class LLMApprovalMixin:
         self.validate_approval_prompt()
 
         raw_output = output
-        if isinstance(output, BaseModel):
-            output = output.model_dump_json()
-        elif not isinstance(output, str):
-            # JSON round-trip: execute_complete validates the string back into output_type.
-            output = TypeAdapter(type(output)).dump_json(output).decode()
+        # JSON round-trip: execute_complete validates the string back into output_type.
+        output = dump_output_to_json(output)
 
         ti_id = context["task_instance"].id
 
--- a/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
@@ -21,8 +21,7 @@ import logging
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any, Protocol
 
-from pydantic import BaseModel, TypeAdapter
-
+from airflow.providers.common.ai.utils.output_type import dump_output_to_json
 from airflow.providers.common.compat.version_compat import AIRFLOW_V_3_3_PLUS
 
 if AIRFLOW_V_3_3_PLUS:
@@ -105,11 +104,8 @@ class LLMApprovalMixin:
                 "require_approval."
             )
 
-        if isinstance(output, BaseModel):
-            output = output.model_dump_json()
-        elif not isinstance(output, str):
-            # JSON round-trip: execute_complete validates the string back into output_type.
-            output = TypeAdapter(type(output)).dump_json(output).decode()
+        # JSON round-trip: execute_complete validates the string back into output_type.
+        output = dump_output_to_json(output)
 
         ti_id = context["task_instance"].id
 
--- a/providers/common/ai/src/airflow/providers/common/ai/mixins/hitl_review.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/mixins/hitl_review.py
@@ -21,8 +21,6 @@ import time
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any, Protocol
 
-from pydantic import BaseModel
-
 from airflow.providers.common.ai.exceptions import HITLMaxIterationsError
 from airflow.providers.common.ai.utils.hitl_review import (
     XCOM_AGENT_OUTPUT_PREFIX,
@@ -32,6 +30,7 @@ from airflow.providers.common.ai.utils.hitl_review import (
     HumanActionData,
     SessionStatus,
 )
+from airflow.providers.common.ai.utils.output_type import dump_output_to_json
 
 log = logging.getLogger(__name__)
 
@@ -268,8 +267,4 @@ class HITLReviewMixin:
 
     @staticmethod
     def _to_string(output: Any) -> str:
-        if isinstance(output, BaseModel):
-            return output.model_dump_json()
-        if not isinstance(output, str):
-            return str(output)
-        return output
+        return dump_output_to_json(output)
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
@@ -18,7 +18,6 @@
 
 from __future__ import annotations
 
-import json
 from collections.abc import Sequence
 from dataclasses import replace
 from datetime import timedelta
@@ -496,16 +495,11 @@ class AgentOperator(BaseOperator, HITLReviewMixin):
                 output,
                 message_history=result.all_messages(),
             )
-            if isinstance(self.output_type, type) and issubclass(self.output_type, BaseModel):
-                return rehydrate_pydantic_output(
-                    self.output_type,
-                    result_str,
-                    serialize_output=self._serialize_model_output,
-                )
-            try:
-                return json.loads(result_str)
-            except (ValueError, TypeError):
-                return result_str
+            return rehydrate_pydantic_output(
+                self.output_type,
+                result_str,
+                serialize_output=self._serialize_model_output,
+            )
 
         if self._serialize_model_output and isinstance(output, BaseModel):
             output = output.model_dump()
@@ -556,7 +550,4 @@ class AgentOperator(BaseOperator, HITLReviewMixin):
         result = agent.run_sync(feedback, message_history=messages, usage_limits=self.usage_limits)
         log_run_summary(self.log, result)
 
-        output = result.output
-        if isinstance(output, BaseModel):
-            output = output.model_dump_json()
-        return str(output), result.all_messages()
+        return self._to_string(result.output), result.all_messages()
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
@@ -18,7 +18,6 @@
 
 from __future__ import annotations
 
-import json
 from collections.abc import Sequence
 from dataclasses import replace
 from datetime import timedelta
@@ -493,16 +492,11 @@ class AgentOperator(BaseOperator, HITLReviewMixin):
                 output,
                 message_history=result.all_messages(),
             )
-            if isinstance(self.output_type, type) and issubclass(self.output_type, BaseModel):
-                return rehydrate_pydantic_output(
-                    self.output_type,
-                    result_str,
-                    serialize_output=self._serialize_model_output,
-                )
-            try:
-                return json.loads(result_str)
-            except (ValueError, TypeError):
-                return result_str
+            return rehydrate_pydantic_output(
+                self.output_type,
+                result_str,
+                serialize_output=self._serialize_model_output,
+            )
 
         if self._serialize_model_output and isinstance(output, BaseModel):
             output = output.model_dump()
@@ -553,7 +547,4 @@ class AgentOperator(BaseOperator, HITLReviewMixin):
         result = agent.run_sync(feedback, message_history=messages, usage_limits=self.usage_limits)
         log_run_summary(self.log, result)
 
-        output = result.output
-        if isinstance(output, BaseModel):
-            output = output.model_dump_json()
-        return str(output), result.all_messages()
+        return self._to_string(result.output), result.all_messages()
--- a/providers/common/ai/src/airflow/providers/common/ai/utils/output_type.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/utils/output_type.py
@@ -19,9 +19,15 @@
 from __future__ import annotations
 
 import json
+import logging
 from typing import Any
 
 from pydantic import BaseModel, TypeAdapter, ValidationError
+from pydantic_ai import NativeOutput, PromptedOutput, ToolOutput
+
+log = logging.getLogger(__name__)
+
+_OUTPUT_MARKERS = (ToolOutput, NativeOutput, PromptedOutput)
 
 
 def dump_output_to_json(output: Any) -> str:
@@ -38,6 +44,10 @@ def dump_output_to_json(output: Any) -> str:
     try:
         return TypeAdapter(type(output)).dump_json(output).decode()
     except Exception:
+        log.warning(
+            "Could not build a pydantic schema for %s; showing its repr to the reviewer instead.",
+            type(output),
+        )
         return str(output)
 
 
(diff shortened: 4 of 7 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
```diff
@@ -493,16 +492,11 @@ def execute(self, context: Context) -> Any:
...
-                return rehydrate_pydantic_output(
-                    self.output_type,
-                    result_str,
-                    serialize_output=self._serialize_model_output,
-                )
-            try:
-                return json.loads(result_str)
-            except (ValueError, TypeError):
-                return result_str
+            return rehydrate_pydantic_output(
```
kaxil, line 495: Dropping the `json.loads` fallback here regresses the exact case this PR targets. `rehydrate_pydantic_output` returns `raw` unchanged when `output_type` isn't a `BaseModel` subclass (its docstring notes the caller is expected to apply its own `json.loads`), so with `output_type=list[str]` this returns the JSON string `'["tag-a","tag-b"]'` instead of the list. The new `test_execute_with_hitl_rehydrates_non_base_model_output` fails on exactly this (`assert '["tag-a", "tag-b"]' == ['tag-a', 'tag-b']`). Keeping the previous `json.loads`/`except` fallback after the rehydrate call fixes it while preserving the `BaseModel` handling.

ColtenOuO: In PR #70075 (merged), `rehydrate_pydantic_output` was updated to handle any non-`str` output_type, not just `BaseModel` subclasses:

```python
if output_type is str:
    return raw
try:
    rehydrated = TypeAdapter(output_type).validate_json(raw)
except (ValidationError, ValueError, TypeError):
    return raw
```

```bash
uv run --project providers/common/ai python -c "tic_output
from airflow.providers.common.ai.utils.output_type import rehydrate_pydantic_outputtput=False)
result = rehydrate_pydantic_output(list[str], '[\"tag-a\",\"tag-b\"]', serialize_output=False)
print(repr(result), type(result))
"
```
Output:

```
['tag-a', 'tag-b'] <class 'list'>
```

So I don't think the `json.loads` fallback needs to come back (?

Let me know what you think, or if I might have missed anything!

Thanks for your review!

kaxil: You're right about #70075 -- it is an ancestor of this branch, and `rehydrate_pydantic_output(list[str], '["tag-a","tag-b"]', serialize_output=False)` does return `['tag-a', 'tag-b']`. My reasoning for `list[str]`/`bool`/`dict` was wrong.

The old `isinstance(self.output_type, type)` check was also guarding the case where `output_type` isn't a plain type at all. `AgentOperator` passes it straight through to `Agent(...)` (agent.py:342 -> hooks/pydantic_ai.py:279), so pydantic-ai's multi-output and marker forms land here, and `TypeAdapter` fails on them outside the helper's `except (ValidationError, ValueError, TypeError)`:

```
output_type=[Foo, Bar]      -> PydanticSchemaGenerationError   (PydanticUserError -> RuntimeError, so not caught)
output_type=ToolOutput(Foo) -> AttributeError: 'ToolOutput' object has no attribute '__mro__'
```

On pydantic-ai 2.13 / pydantic 2.13.4, `Agent(TestModel(), output_type=[Foo, Bar])` returns `Foo(a=0)` and `_to_string` gives `'{"a":0}'`. Before this PR that fell through to `json.loads` and came back as `{'a': 0}`; now it raises, after the model call and after the reviewer has already approved, so the approved output is lost.

Widening the helper's `except` doesn't get you back to the old result -- it returns `raw`, so callers get `'{"a":0}'` instead of the dict. Separating schema-build failure from validation does:

```python
try:
    adapter = TypeAdapter(output_type)
except Exception:
    # No pydantic schema for this output_type: ToolOutput/NativeOutput/PromptedOutput
    # markers, output functions, or a `[A, B]` union list. Fall back to plain JSON.
    try:
        return json.loads(raw)
    except (ValueError, TypeError):
        return raw
try:
    return adapter.validate_json(raw)
except (ValidationError, ValueError, TypeError):
    return raw
```

Putting it in `utils/output_type.py` rather than here also covers `LLMOperator`'s `require_approval` path, which calls the helper at llm.py:187.

providers/common/ai/src/airflow/providers/common/ai/mixins/hitl_review.py
```diff
@@ -271,5 +271,5 @@ def _to_string(output: Any) -> str:
         if isinstance(output, BaseModel):
             return output.model_dump_json()
         if not isinstance(output, str):
-            return str(output)
+            return TypeAdapter(type(output)).dump_json(output).decode()
```
kaxil: Same unguarded `TypeAdapter` on the serialize side: `str(output)` could never raise, `TypeAdapter(type(output))` can. It's reachable through pydantic-ai output functions, which only warn ("Falling back to unconstrained schema") for a non-schema-able return type and then run fine -- `Agent(TestModel(), output_type=make_widget)` where `make_widget` returns a plain class makes this raise `PydanticSchemaGenerationError` where it previously produced `'Widget(0)'`.

This one fires at line 111, before the session XCom push, so the task dies before the reviewer ever sees a session. `except Exception: return str(output)` restores the old degradation, and pairs with the schema-build guard on the rehydrate side.

providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
kaxil, line 495: This also changes the default `output_type=str`, and neither the description nor a test mentions it: an approved string of `'42'` used to come back as `42`, `'{"text": "hi"}'` as a dict, `'true'` as `True` and `'null'` as `None`; all four now stay strings. That's the right behaviour -- it makes the HITL path consistent with the non-HITL path, which returns `result.output` unparsed -- but `str` is the default `output_type` and `enable_hitl_review` has shipped in every release from 0.1.0 through 0.6.0, so it's worth a line in the description rather than arriving silently.

Running the six new tests against the pre-fix source gives 3 failed, 3 passed. `test_execute_with_hitl_rehydrates_non_base_model_output` is in the passing group: it patches `run_hitl_review` to return a hand-written `'["tag-a", "tag-b"]'`, so `_to_string` never runs and the `json.loads` removal is never exercised. Nothing currently covers the `execute()` half of this diff. A case with `output_type=str` and a JSON-parseable approved string would pin it.

providers/common/ai/src/airflow/providers/common/ai/operators/agent.py
```diff
@@ -556,4 +550,6 @@ def regenerate_with_feedback(self, *, feedback: str, message_history: Any) -> tu
         output = result.output
         if isinstance(output, BaseModel):
             output = output.model_dump_json()
-        return str(output), result.all_messages()
+        elif not isinstance(output, str):
+            output = TypeAdapter(type(output)).dump_json(output).decode()
+        return output, result.all_messages()
```
kaxil: `AgentOperator` already inherits `_to_string` from `HITLReviewMixin`, and these lines are now exactly what it does, so `return self._to_string(result.output), result.all_messages()` collapses it.

Worth noting the same block also lives in `LLMApprovalMixin.defer_for_approval` (approval.py:108-112), so that's three copies. The split is how the original bug arose in the first place: `approval.py` was fixed in #70075 while these two weren't. If you add the guard from the other comments, one shared helper next to `rehydrate_pydantic_output` fixes all three at once and gives the round-trip a matching pair.

ColtenOuO:
Thanks for reviewing, I have addressed all of the comments.

If there still have anything need to improve or change, feel free to let me know ><

providers/common/ai/src/airflow/providers/common/ai/utils/output_type.py
```diff
@@ -46,7 +65,19 @@ def rehydrate_pydantic_output(
     if output_type is str:
         return raw
     try:
-        rehydrated = TypeAdapter(output_type).validate_json(raw)
+        adapter = TypeAdapter(output_type)
```
kaxil: `TypeAdapter` doesn't raise for an output function -- it builds a `call` schema -- so this guard never fires for them, and `adapter.validate_json(raw)` below validates `raw` as the function's *arguments* and invokes it. Verified on both the provider's `pydantic-ai>=2.0.0` floor and 2.22.0, with

```python
def open_ticket(severity: int) -> Ticket:
    SIDE_EFFECTS.append(...)   # e.g. POST to Jira
    return Ticket(severity=severity)
```

`rehydrate_pydantic_output(open_ticket, '{"severity":0}')` returns `Ticket(severity=0)` **and runs `open_ticket` a second time**. If the reviewer edits the approved string it re-runs with the edited arguments (`'{"severity": 9999}'` gives a second call with `severity=9999`).

That is new here: `Agent(TestModel(), output_type=open_ticket)` is accepted by pydantic-ai, and at the merge-base `isinstance(open_ticket, type)` was False so `execute()` took `json.loads` and the function was never re-invoked (same probe against the pre-PR source: no new side effects). Output functions are pydantic-ai's tool-shaped final-output hook, so the ones with side effects now fire twice, after the reviewer has already approved.

The comment just below lists output functions as a case this branch handles, which is the tell: "did `TypeAdapter` raise" isn't the right test. A positive check works -- reject a non-class callable before adapting, and unwrap the markers instead of falling back for them (`ToolOutput(A).output`, `NativeOutput(A).outputs`, `PromptedOutput(A).outputs` all give `A`). That also fixes the smaller problem that `ToolOutput(A)` currently returns `{'x': 7}` where a bare `A` returns `A(x=7)`, with `serialize_output` ignored on that path, and it leaves `json.loads` serving only the genuinely un-buildable `[A, B]` union. Worth a test either way: `test_falls_back_to_json_when_output_type_has_no_schema` parametrizes only `[A, B]` and `ToolOutput(A)`, so the output-function case is asserted nowhere.

providers/common/ai/src/airflow/providers/common/ai/utils/output_type.py
```diff
@@ -18,11 +18,29 @@
...
+    The inverse of :func:`rehydrate_pydantic_output`: both sides of the review
+    round-trip live here so they cannot drift apart.
+    """
+    if isinstance(output, BaseModel):
+        return output.model_dump_json()
+    if isinstance(output, str):
+        return output
+    try:
+        return TypeAdapter(type(output)).dump_json(output).decode()
+    except Exception:
```
kaxil, line 46: This turns what used to be a crash into a silent repr. At the merge-base `approval.py` had no `try`/`except` here, so a non-serializable output raised before the reviewer saw anything; now `str(output)` runs and the reviewer is shown a Python object address -- `dump_output_to_json(Widget(3))` returns `'<__main__.Widget object at 0x1023430b0>'`.

Reachable on the path the new test comment already names: `Agent(TestModel(), output_type=make_widget)` builds on just a `UserWarning` ("Could not generate return schema for 'make_widget': unsupported return type ... Falling back to unconstrained schema") and returns a `Widget`. That string goes into the approval body and the `airflow_hitl_review_agent_output_N` XComs, and comes back out of `rehydrate_pydantic_output` as the same plain string -- the repr-instead-of-JSON failure this PR is fixing. Preferring the fallback over the crash is right, but it should say so in the log; the module has no logger, and `utils/file_analysis.py` in the same package already has one.

providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
```diff
@@ -105,11 +104,8 @@ def defer_for_approval(
                 "require_approval."
             )
 
-        if isinstance(output, BaseModel):
-            output = output.model_dump_json()
-        elif not isinstance(output, str):
-            # JSON round-trip: execute_complete validates the string back into output_type.
-            output = TypeAdapter(type(output)).dump_json(output).decode()
+        # JSON round-trip: execute_complete validates the string back into output_type.
+        output = dump_output_to_json(output)
```
kaxil, line 126: Heads-up for the rebase, since this is the only file that conflicts with `main`: #71046 added `raw_output = output` immediately above this block (`approval.py:125` on main) and reads it further down -- `isinstance(raw_output, list)` then `param_value = raw_output` -- to render the multi-select. Resolving the conflict by taking this side wholesale drops it and quietly breaks that path. The `pydantic` import removal is still correct after the rebase: `BaseModel` and `TypeAdapter` have no other use in the file.

ColtenOuO:
Thanks for catching this. I reproduced it myself and can confirm the output function re-invocation issue is real.

The latest commit addresses everything from this round:

- rehydrate_pydantic_output now unwraps ToolOutput/NativeOutput/PromptedOutput markers and gates on isinstance(x, type) instead of "did TypeAdapter raise" — an output function is never re-invoked during rehydrate.
- dump_output_to_json's fallback now logs a warning instead of degrading silently.
- Added tests covering marker unwrapping, the sequence-fallback case, and (most importantly) that an output function is never called during rehydrate.

Ready for another pass whenever you have time, thanks again for the thorough review!

providers/common/ai/src/airflow/providers/common/ai/utils/output_type.py
```diff
@@ -37,16 +65,38 @@ def rehydrate_pydantic_output(
...
+    if isinstance(output_type, _OUTPUT_MARKERS):
+        unwrapped = output_type.output if isinstance(output_type, ToolOutput) else output_type.outputs
+
+    if not isinstance(unwrapped, type):
+        # Not a plain class: an output function, a ``[A, B]`` union list, or a
+        # marker wrapping one. These reach us because the operator passes
+        # output_type straight to Agent(...). Fall back to plain JSON rather
+        # than risk building a schema that re-invokes an output function.
+        try:
+            return json.loads(raw)
```
aaron-y-chen, line 99: If `output_type=list[A]`, where `A` is a user-defined Pydantic model, `rehydrate_pydantic_output()` returns `list[dict]` instead of `list[A]`. 

I would expect the declared output type to be preserved, with the result being `[A(x=7)]`. Is this behavior intentional?

```console
uv run --project providers/common/ai python -c '
from pydantic import BaseModel
from airflow.providers.common.ai.utils.output_type import rehydrate_pydantic_output

class A(BaseModel):
    x: int

result = rehydrate_pydantic_output(
    list[A],
    "[{\"x\": 7}]",
    serialize_output=False,
)
print(result)
print(type(result[0]))
assert isinstance(result[0], A)
'

>>> [{'x': 7}]
>>> <class 'dict'>
>>> Traceback (most recent call last):
>>>   File "<string>", line 15, in <module>
>>> AssertionError
```

ColtenOuO: I think that is not intentional. Like a new bug cause this PR : (

`isinstance(unwrapped, type)` is `False` for a generic alias like `list[A]` (it's `types.GenericAlias`), so it was routed into
the plain `json.loads` fallback meant for output functions / `[A, B]` union lists, silently downgrading `list[A]` to `list[dict]`.

Fixed in 6280539 by also checking `typing.get_origin(unwrapped) is not None` so generic aliases still reach `TypeAdapter`. Note `callable(list[A])` is `True`, so a callable-based check can't distinguish it from an output
function — `get_origin()` is the right signal.

Also generalized the `serialize_output=True` dump step (`adapter.dump_python(rehydrated, mode="python")` instead of an `isinstance(rehydrated, BaseModel)` check), since a rehydrated `list[A]` is a list, not a `BaseModel`, and needs the same dict-shape dump its elements used to get.

ColtenOuO:
Thanks @aaron-y-chen's review!

I've addressed the problem in lastest commit, ready for the next round review!

PR #70137, opened 2026-07-20
Fix LLMSQLQueryOperator not stripping single-line markdown code fences

ColtenOuO:
## Summary

`LLMSQLQueryOperator._strip_llm_output` strips markdown code fences from LLM output before the generated SQL is validated/executed. The existing logic only handles fences where the LLM's response spans multiple lines
(e.g. ` ```sql\nSELECT 1\n``` `) — it splits on `"\n"` and requires at least 2 lines before attempting to drop the opening/closing fence.

When the whole fenced response is on a single line (e.g. ` ```SELECT 1``` `, no internal newline), `text.split("\n")` yields a
single element, the `len(lines) >= 2` guard is never satisfied, and the fence-stripping is skipped entirely — the backticks stay attached and the value is passed straight into `_validate_sql` (and potentially execution) as-is.

The operator's system prompt already asks the LLM not to use markdown at all (`"Return ONLY the SQL query, no explanation or markdown."`), but the existing multi-line fence-handling logic (and its test coverage) already assumes LLMs commonly ignore that instruction and wrap output in a fence anyway. 

This change extends the same defensive handling to the single-line variant of that same behavior, which the multi-line path doesn't cover.

## Changes

- `_strip_llm_output` gains an `elif` branch for the case where the fenced
  text has no internal newline: strip the leading/trailing ` ``` ` directly
  instead of relying on line-splitting. The existing multi-line branch is
  untouched, so behavior for every previously-tested case is unchanged.

## Test plan

- Added two regression cases to `TestStripLLMOutput`: a single-line fence
  with no language tag, and one wrapping a longer query.
- Full `test_llm_sql.py` suite passes (44 passed), including all
  pre-existing multi-line fence cases unchanged.
- `ruff format` / `ruff check`, `prek --stage pre-commit`, and
  `breeze run mypy` all clean on the changed file.

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
@@ -177,10 +177,13 @@ class LLMSQLQueryOperator(LLMOperator):
         text = raw.strip()
         if text.startswith("```"):
             lines = text.split("\n")
-            # Remove opening fence (```sql, ```, etc.) and closing fence
             if len(lines) >= 2:
+                # Remove opening fence (```sql, ```, etc.) and closing fence
                 end = -1 if lines[-1].strip().startswith("```") else len(lines)
                 text = "\n".join(lines[1:end]).strip()
+            elif text.endswith("```") and len(text) > 6:
+                # Whole fenced block on one line, e.g. "```SELECT 1```" -> "SELECT 1"
+                text = text[3:-3].strip()
         return text
 
     def _get_schema_context(self) -> str:
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
@@ -152,7 +152,7 @@ class LLMSQLQueryOperator(LLMOperator):
         )
         result = agent.run_sync(self.prompt, usage_limits=self.usage_limits)
         log_run_summary(self.log, result)
-        sql = self._strip_llm_output(result.output)
+        sql = self._strip_llm_output(result.output, dialect=self._resolved_dialect)
 
         if self.validate_sql:
             _validate_sql(sql, allowed_types=self.allowed_sql_types, dialect=self._resolved_dialect)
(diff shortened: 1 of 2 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
```diff
@@ -177,10 +177,12 @@ def _strip_llm_output(raw: str) -> str:
         text = raw.strip()
         if text.startswith("```"):
             lines = text.split("\n")
-            # Remove opening fence (```sql, ```, etc.) and closing fence
             if len(lines) >= 2:
+                # Remove opening fence (```sql, ```, etc.) and closing fence
                 end = -1 if lines[-1].strip().startswith("```") else len(lines)
                 text = "\n".join(lines[1:end]).strip()
+            elif text.endswith("```") and len(text) > 6:
```
Lee-W, line 184: let's add a short example as comment here. would be easier to understand

Lee-W (approved):

providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
ColtenOuO: Added! 

Updated the comment to include single-line fence examples like ```SELECT 1```. Thanks for the review!

providers/common/ai/src/airflow/providers/common/ai/operators/llm_sql.py
```diff
@@ -177,10 +177,13 @@ def _strip_llm_output(raw: str) -> str:
...
+                # Whole fenced block on one line, e.g. "```SELECT 1```" -> "SELECT 1"
+                text = text[3:-3].strip()
```
kaxil: Does this want to handle a single-line fence that still has a language tag, like ````sql SELECT 1````? `text[3:-3]` leaves `sql SELECT 1`, which then fails validation. The multi-line path strips the tag by dropping the first line, but there's no line to drop here. Fine to skip if you haven't seen models collapse a tagged fence onto one line -- just flagging since it's the same defensive path you're extending.

ColtenOuO: I think this defensive extension is worth handling — updated `_strip_llm_output` to also strip a same-line language tag on the single-line path — but only when the leading word is literally `sql` or the resolved dialect (the same value used in the system prompt's `dialect_label`), so a real query starting with a SQL keyword is never mistaken for a tag. 

Added test cases for `sql`, uppercase `SQL`, dialect match, and dialect mismatch.

PR #70138, opened 2026-07-20
Keep Dag run execution API endpoints working for older Task SDK clients

henry3260:
## Why

The `2026-06-30` execution API version added two new fields to the `DagRun` model: `team_name` (`AddTeamNameField`) and `partition_date` (`AddPartitionDateField`). Both version changes only registered response converters for the **nested** `dag_run` payload (e.g. inside `TIRunContext`), but not for endpoints that return a `DagRun` **directly** — `GET /dag-runs/{dag_id}/{run_id}` and `GET /dag-runs/previous`.

As a result, these endpoints break on the **server** side for a Task SDK client pinning `2026-04-06` (the only earlier version they serve — both routes were added in it). Both routes have return annotations, so FastAPI validates the response against the versioned `DagRun` response model, which inherits `StrictBaseModel` (`extra="forbid"`) and, per the version-change declarations, has no `team_name`/`partition_date` in `2026-04-06`. The un-stripped fields make response validation raise `fastapi.exceptions.ResponseValidationError` with `extra_forbidden` errors, and the worker receives a `500 Internal Server Error`. The response never reaches the client, so the symptom to grep for is `extra_forbidden` in api-server logs — not a client-side Task SDK validation error. Either way, calls that worked on `2026-04-06` now fail for pinned older clients — exactly what the versioning layer is supposed to prevent.

## What

- `airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py`: add `@convert_response_to_previous_version_for(DagRun)` converters to both `AddTeamNameField` and `AddPartitionDateField` so the new fields are stripped from direct `DagRun` responses. Because schema-based converters match the route's response model by identity, the `/dag-runs/previous` route (annotated `DagRun | None`) never matches `DagRun`, so it additionally gets path-based converters for `GET /dag-runs/previous`.
- `airflow-core/tests/unit/api_fastapi/execution_api/versions/v2026_06_30/test_dag_runs.py` (new): tests that requests with the previous API version (`2026-04-06`) get responses without `team_name`/`partition_date` from both endpoints, and that the `null` previous-run response still works.

Out of scope: the compat `GET /dag-runs/{dag_id}/previous` route (served only to clients pinning older than `2026-04-06`) has a pre-existing leak of newer `DagRun` fields (`note`, `partition_key`, `triggering_user_name`, and now these two) with a different failure mode — it has no return annotation, so the server responds 200 and the old SDK fails client-side on its own `extra="forbid"` model. That affects a disjoint set of versions and fields and needs its own fix; see the review discussion.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
@@ -99,6 +99,20 @@ class AddTeamNameField(VersionChange):
         if "dag_run" in response.body and isinstance(response.body["dag_run"], dict):
             response.body["dag_run"].pop("team_name", None)
 
+    @convert_response_to_previous_version_for(DagRun)  # type: ignore[arg-type]
+    def remove_team_name_from_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
+        """Remove the ``team_name`` field from responses returning a DagRun directly."""
+        if isinstance(response.body, dict):
+            response.body.pop("team_name", None)
+
+    # Schema-based converters are matched against the route's response model by identity, so a
+    # route annotated ``DagRun | None`` never matches ``DagRun`` and has to be addressed by path.
+    @convert_response_to_previous_version_for("/dag-runs/previous", ["GET"])  # type: ignore[arg-type]
+    def remove_team_name_from_previous_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
+        """Remove the ``team_name`` field from the previous-run response."""
+        if isinstance(response.body, dict):
+            response.body.pop("team_name", None)
+
 
 class AddAssetsByAliasEndpoint(VersionChange):
     """Add endpoint to resolve assets from an AssetAlias."""
@@ -141,3 +155,16 @@ class AddPartitionDateField(VersionChange):
         """Strip ``partition_date`` from the nested ``dag_run`` payload for older clients."""
         if "dag_run" in response.body and isinstance(response.body["dag_run"], dict):
             response.body["dag_run"].pop("partition_date", None)
+
+    @convert_response_to_previous_version_for(DagRun)  # type: ignore[arg-type]
+    def remove_partition_date_from_dag_run_response(response: ResponseInfo) -> None:  # type: ignore[misc]
+        """Strip ``partition_date`` from responses returning a DagRun directly."""
+        if isinstance(response.body, dict):
+            response.body.pop("partition_date", None)
+
+    # See remove_team_name_from_previous_dag_run: ``DagRun | None`` needs a path-based converter.
+    @convert_response_to_previous_version_for("/dag-runs/previous", ["GET"])  # type: ignore[arg-type]
+    def remove_partition_date_from_previous_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
+        """Strip ``partition_date`` from the previous-run response."""
+        if isinstance(response.body, dict):
+            response.body.pop("partition_date", None)
--- a/airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
@@ -100,7 +100,7 @@ class AddTeamNameField(VersionChange):
             response.body["dag_run"].pop("team_name", None)
 
     @convert_response_to_previous_version_for(DagRun)  # type: ignore[arg-type]
-    def remove_team_name_from_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
+    def remove_team_name_from_dag_run_response(response: ResponseInfo) -> None:  # type: ignore[misc]
         """Remove the ``team_name`` field from responses returning a DagRun directly."""
         if isinstance(response.body, dict):
             response.body.pop("team_name", None)
--- a/airflow-core/tests/unit/api_fastapi/execution_api/versions/v2026_06_30/test_dag_runs.py
+++ b/airflow-core/tests/unit/api_fastapi/execution_api/versions/v2026_06_30/test_dag_runs.py
@@ -1,0 +1,88 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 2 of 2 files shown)


henry3260:
@ashb Could I get your thoughts when you have time? 🙏

airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
```diff
@@ -99,6 +99,20 @@ def remove_team_name_field(response: ResponseInfo) -> None:  # type: ignore[misc
...
 
+    @convert_response_to_previous_version_for(DagRun)  # type: ignore[arg-type]
+    def remove_team_name_from_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
+        """Remove the ``team_name`` field from responses returning a DagRun directly."""
+        if isinstance(response.body, dict):
+            response.body.pop("team_name", None)
+
+    # Schema-based converters are matched against the route's response model by identity, so a
+    # route annotated ``DagRun | None`` never matches ``DagRun`` and has to be addressed by path.
+    @convert_response_to_previous_version_for("/dag-runs/previous", ["GET"])  # type: ignore[arg-type]
```
kaxil, line 110: There's a third endpoint that returns a `DagRun` directly: the compat `GET /dag-runs/{dag_id}/previous`, kept via `only_exists_in_older_versions` for clients pinning older than `2026-04-06`. It has no `response_model` annotation, so the `DagRun` schema converter won't match it, and the path converter here only covers `/dag-runs/previous`, so a worker pinning e.g. `2025-11-05` still gets `team_name`/`partition_date` from it and hits the same `extra="forbid"` failure. `note`/`partition_key` (added in `2026-04-06`) look like they already leak there for the same reason, so this may be pre-existing rather than something this PR must own. Is that compat path still reachable by old clients?

henry3260: > There's a third endpoint that returns a `DagRun` directly: the compat `GET /dag-runs/{dag_id}/previous`, kept via `only_exists_in_older_versions` for clients pinning older than `2026-04-06`. It has no `response_model` annotation, so the `DagRun` schema converter won't match it, and the path converter here only covers `/dag-runs/previous`, so a worker pinning e.g. `2025-11-05` still gets `team_name`/`partition_date` from it and hits the same `extra="forbid"` failure. `note`/`partition_key` (added in `2026-04-06`) look like they already leak there for the same reason, so this may be pre-existing rather than something this PR must own. Is that compat path still reachable by old clients?

Yes, still reachable — 2025-08-10 through 2025-11-05, and a released SDK did call it (at 35d23c3222 it pinned API_VERSION = "2025-08-10" and hit that path).

One correction though: it isn't a 500. That route has no return annotation, so response_model is None and FastAPI skips response validation — the server returns 200 with every field. The `extra="forbid"` failure lands on the client instead, when the old SDK does `PreviousDagRunResult(dag_run=resp.json())`. And you're right that it's pre-existing: `note, partition_key and triggering_user_name` already leak there today.

On scope — not because it's pre-existing (so is the bug here), but because the two don't overlap. Both routes this PR touches were added in 2026-04-06, so they return 404/405 below it; the only older version they ever serve is 2026-04-06, and the only fields added after that are the two this PR strips. The compat route is a disjoint set of versions and fields with a different failure mode, and fixing it needs a call on whether to give it a response model or add path converters for all five fields across four version files.

henry3260:
@kaxil Whenever you have a moment — could I get your thoughts on my reply about scope above? Thanks!

henry3260:
Hi @kaxil, checks are all green. Is there anything else I can address to help move this forward? Happy to adjust scope if needed :)

airflow-core/tests/unit/api_fastapi/execution_api/versions/v2026_06_30/test_dag_runs.py
```diff
@@ -0,0 +1,88 @@
...
+        state=DagRunState.SUCCESS,
+        logical_date=timezone.datetime(2025, 1, 1),
+        run_id="run1",
+    )
+    dag_maker.create_dagrun(
+        state=DagRunState.SUCCESS,
+        logical_date=timezone.datetime(2025, 1, 10),
+        run_id="run2",
+    )
+    session.commit()
```
kaxil, line 53: Neither fixture run sets `partition_date`, so both fields are `None` here and a value-gated converter (pop only when the value is falsy) would pass all three tests while still 500-ing a 2026-04-06 client on a partitioned Dag. `partition_date` is a real column, so setting it on one run before the commit closes that gap, the same way `test_old_version_strips_partition_date_from_dag_run` does in the sibling `test_task_instances.py`.

Worth noting `team_name` can't be exercised here at all -- it isn't a `DagRun` column, just a transient attribute the TIRunContext route assigns, so on these two routes it is always `None`.

airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
```diff
@@ -99,6 +99,20 @@ def remove_team_name_field(response: ResponseInfo) -> None:  # type: ignore[misc
         if "dag_run" in response.body and isinstance(response.body["dag_run"], dict):
             response.body["dag_run"].pop("team_name", None)
 
+    @convert_response_to_previous_version_for(DagRun)  # type: ignore[arg-type]
+    def remove_team_name_from_dag_run(response: ResponseInfo) -> None:  # type: ignore[misc]
```
kaxil: `remove_team_name_from_dag_run` reads like the nested-`dag_run` converter, which is what `remove_partition_date_from_dag_run` further down actually is. The new partition_date one dodged that with a `_response` suffix, so matching it here keeps the file readable for the next person adding a field.

airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_06_30.py
kaxil: Agreed on scope, and thanks for the correction -- I confirmed it. With the four converters reverted, both routes raise `fastapi.exceptions.ResponseValidationError` (`extra_forbidden` on `team_name`/`partition_date`), so a 2026-04-06 worker gets a 500 from the API server, not the client-side error the description describes. Worth fixing that in the body, since the signature an operator would grep for is `extra_forbidden` in api-server logs.

On the compat route, I checked the delta and you're right that it's separable: 3.2.0 shipped with no path converters in `v2026_04_06.py` and the route still un-annotated, so a `task-sdk` 1.1.8 worker pinning 2025-11-05 already fails there on `note`/`partition_key` -- two more fields don't change that, and it needs its own fix covering all of them.

kaxil (approved):

airflow-core/tests/unit/api_fastapi/execution_api/versions/v2026_06_30/test_dag_runs.py
henry3260: > Neither fixture run sets `partition_date`, so both fields are `None` here and a value-gated converter (pop only when the value is falsy) would pass all three tests while still 500-ing a 2026-04-06 client on a partitioned Dag. `partition_date` is a real column, so setting it on one run before the commit closes that gap, the same way `test_old_version_strips_partition_date_from_dag_run` does in the sibling `test_task_instances.py`.
> 
> Worth noting `team_name` can't be exercised here at all -- it isn't a `DagRun` column, just a transient attribute the TIRunContext route assigns, so on these two routes it is always `None`.

Applied!

PR #70142, opened 2026-07-20
UI: Make duration charts readable at a glance

### LINKED ISSUE #63582: Problems with rendering the task duration graph in airflow 3
### Apache Airflow version

Other Airflow 3 version (please specify below)

### If "Other Airflow 3 version" selected, which one?

3.1.7

### What happened?

The task duration graph in airflow 3 does not render that well in comparison with airflow 2. Can this be solved?

1. Chart size — Airflow 2 renders a large full-width chart; Airflow 3 shows a small card where bars are barely distinguishable
2. Y-axis unit — Airflow 2 uses minutes or hours (9, 12, 15…), Airflow 3 uses HH:MM:SS which is less intuitive at a glance
3. Median total duration line — the blue dashed reference line with a labeled value is completely absent in Airflow 3
4. Legend — explicitly shown in Airflow 2, not present in Airflow 3
5. Information density — in Airflow 2 everything is visible at once; in Airflow 3 you have to hover to see individual values, with no summary statistics shown

This is the UI in airflow 3

Image

this is the same UI in airflow 2

Image

_No response_

### How to reproduce

Click on a task to see the overall task duration graph.

### Operating System

python:3.12.0-slim-bookworm

apache-airflow-providers-microsoft-azure==12.5.0
apache-airflow-providers-databricks==6.5.0
apache-airflow-providers-standard==1.4.1

### Deployment

Other

### Deployment details

_No response_

### Anything else?

_No response_

### Code of Conduct

eddiesr93:
Redesign of the Dag Run / Task Instance duration chart on the overview pages, covering the five points raised in the issue.

- **Axis units** — durations render as `5m`, `1h 30m`, `1d 4h` instead of `00:05:00`, and ticks snap to steps people actually count in (30s, 1m, 5m, 15m, 1h, …) rather than Chart.js' decimal steps, which produced ticks like `26m 40s` / `33m 20s`.
- **Reference line** — the two mean lines are replaced by one dashed median-of-total line labelled `Median total: 4m 15s`. A single stuck run drags the mean well above where runs actually land, and the two old labels were both pinned to the same edge, so they overlapped each other and the most recent bars.
- **Legend** — enabled, so the queued and run series are named.
- **Chart size** — the card was pinned to 350px. It now flexes on the Dag overview and fills its row on the Task overview, which is what makes the bars comparable in the first place.
- **Stacking** — `y.stacked` was never set even though `x.stacked` was, so queued time was drawn *behind* run time rather than underneath it and was effectively invisible. It now sits at the base of each bar.

Median and tick-step selection are pure helpers (`src/utils/median.ts` and `getDurationTickStep`) with unit tests. Full UI suite passes: 88 files / 706 tests.

Screenshots below are the same 12 Dag Runs rendered through the real theme, one run deliberately stuck at ~32m so the median and the mean it replaces land in visibly different places.

**Before**

Duration chart before: HH:mm:ss axis, no legend, two overlapping mean labels, queued time hidden

**After**

Duration chart after: compact time axis on human steps, legend, single dashed median line, queued time stacked at the base

At the narrower width the card falls back to (when asset events share the row):

Duration chart after at 350px

closes: #63582

---

### DIFF
--- a/airflow-core/newsfragments/70142.improvement.rst
+++ b/airflow-core/newsfragments/70142.improvement.rst
@@ -1,1 +1,0 @@
-The Dag Run and Task Instance duration charts on the overview pages label the value axis in compact time units (``5m``, ``1h 30m``) on human-scale tick steps instead of ``HH:mm:ss``, mark the median total duration rather than the mean, show a legend, and stack queued time beneath run time so it is visible. The chart also sizes to the available width instead of a fixed 350px.
--- a/airflow-core/newsfragments/70142.improvement.rst
+++ b/airflow-core/newsfragments/70142.improvement.rst
@@ -1,0 +1,1 @@
+The Dag Run and Task Instance duration charts on the overview pages label the value axis in compact time units (``5m``, ``1h 30m``) on human-scale tick steps instead of ``HH:mm:ss``, mark the median total duration rather than the mean, show a legend, and stack queued time beneath run time so it is visible. The chart also sizes to the available width instead of a fixed 350px.
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
@@ -65,7 +65,7 @@
     "lastDagRun_other": "Last {{count}} Dag Runs",
     "lastTaskInstance_one": "Last Task Instance",
     "lastTaskInstance_other": "Last {{count}} Task Instances",
-    "medianTotalDuration": "Median total: {{duration}}",
+    "medianTotalDuration": "Total: {{duration}}",
     "queuedDuration": "Queued Duration",
     "runAfter": "Run After",
     "runDuration": "Run Duration"
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
@@ -65,6 +65,7 @@
     "lastDagRun_other": "Last {{count}} Dag Runs",
     "lastTaskInstance_one": "Last Task Instance",
     "lastTaskInstance_other": "Last {{count}} Task Instances",
+    "medianTotalDuration": "Median total: {{duration}}",
     "queuedDuration": "Queued Duration",
     "runAfter": "Run After",
     "runDuration": "Run Duration"
--- a/airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
+++ b/airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
@@ -103,8 +103,8 @@ export const Overview = () => {
           startDate={startDate}
         />
       </HStack>
-      <SimpleGrid columns={3} gap={5} my={5}>
-        <Box borderRadius={4} borderStyle="solid" borderWidth={1} p={2} width="350px">
+      <SimpleGrid columns={1} gap={5} my={5}>
+        <Box borderRadius={4} borderStyle="solid" borderWidth={1} maxWidth="900px" p={2}>
           {isLoadingTaskInstances ? (
             <Skeleton height="200px" w="full" />
           ) : (
--- a/airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
+++ b/airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { Box, HStack, Skeleton, SimpleGrid } from "@chakra-ui/react";
+import { Box, HStack, Skeleton } from "@chakra-ui/react";
 import dayjs from "dayjs";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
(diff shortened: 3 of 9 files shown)


airflow-core/newsfragments/70142.improvement.rst
bbovenzi, line 1: Let's remove the newsfragment

airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
```diff
@@ -65,6 +65,7 @@
     "lastDagRun_other": "Last {{count}} Dag Runs",
     "lastTaskInstance_one": "Last Task Instance",
     "lastTaskInstance_other": "Last {{count}} Task Instances",
+    "medianTotalDuration": "Median total: {{duration}}",
```
bbovenzi: I feel like "Median total" is a bit confusing. "Total:" is fine in my opinion

airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
```diff
@@ -103,8 +103,8 @@ export const Overview = () => {
           startDate={startDate}
         />
       </HStack>
-      <SimpleGrid columns={3} gap={5} my={5}>
-        <Box borderRadius={4} borderStyle="solid" borderWidth={1} p={2} width="350px">
+      <SimpleGrid columns={1} gap={5} my={5}>
+        <Box borderRadius={4} borderStyle="solid" borderWidth={1} maxWidth="900px" p={2}>
```
bbovenzi: Why is this different from the same Box in the Dag Overview?

bbovenzi: I'd also like to see a screenshot of how this is different and how the min and max sizes work for both overview pages.

bbovenzi (commented):
Thanks. This looks good. Can you provide a few more screenshots?

eddiesr93:
Pushed a follow-up commit after re-checking the layout against the `AssetEvents` sibling on the Dag overview.

Freeing the card width handed the chart height to Chart.js' default 2:1 aspect ratio, which the old fixed `350px` had been capping as a side effect. Measured on a 1400px row: the canvas came out **992×496**, and on a 2560px monitor it would have scaled past 1000px tall. The chart now renders into a fixed-height box with `maintainAspectRatio: false`, so it is 280px tall at every width, and the loading skeletons were bumped from 200px to match so there is no jump on load.

For the record, the sibling panel is unaffected — `AssetEvents` measures 380px wide with the old fixed width, with the flexed card, and with a capped variant, because it sizes to its content and there is spare room in the row either way.

The screenshots above have been regenerated against the fixed height.

---

airflow-core/newsfragments/70142.improvement.rst
eddiesr93: Removed.

---

airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
eddiesr93: Changed to `Total: {{duration}}`.

One thing worth flagging so it is a deliberate choice rather than a silent one: the line is plotted at the *median* of the per-run totals, not a sum, so `Total:` is shorter but slightly less literal. Happy to leave it as is — say the word if you would rather have something like `Median:`.

---

airflow-core/src/airflow/ui/src/pages/Task/Overview/Overview.tsx
eddiesr93: No good reason — that was an inconsistency on my side. The Task overview was left in its single-column `SimpleGrid` with only a `maxWidth`, while the Dag overview flexed.

Both pages now use the identical Box:

```tsx
<Box
  borderRadius={4}
  borderStyle="solid"
  borderWidth={1}
  flex="1 1 520px"
  maxWidth="900px"
  minWidth="320px"
  p={2}
>
```

The `SimpleGrid` is gone (it only ever wrapped one child) and the Task overview uses the same wrapping `HStack`. Adopting the cap on the Dag overview also fixes something the flex version had wrong on its own: without `maxWidth` the chart kept growing on an ultrawide monitor.

---

eddiesr93: Screenshots below, measured rather than eyeballed. Both pages now share one Box, so the only difference between them is whether Asset Events is in the row.

Canvas is **280px tall in every case** — that is the fixed height from the follow-up commit.

| Case | chart width | Asset Events |
|---|---|---|
| Dag overview, 1900px viewport | 882 (capped) | 380 |
| Dag overview, 1280px viewport | 820 | 380 |
| Dag overview, 900px viewport | 840 (row wrapped) | 380, below |
| Task overview, 1900px viewport | 882 (capped) | — |
| Task overview, 420px viewport | 360 (at `minWidth`) | — |

**Dag overview — wide (1900px):** chart stops at `maxWidth`, Asset Events keeps its natural width.



**Dag overview — medium (1280px):**



**Dag overview — narrow (900px):** the row wraps, chart takes the full width and Asset Events drops below.



**Task overview — wide (1900px):** same Box, no sibling, so the cap is what stops it spanning the page.



**Task overview — narrow (420px):** at `minWidth=320px`, the narrowest it renders.



Worth noting on Asset Events: it measures 380px in every variant, including against the old fixed-width card, because it sizes to its content and there is spare room in the row either way. The flexed chart does not squeeze it.

---

bbovenzi (approved):

PR #70144, opened 2026-07-20
Add Google API toolset for agents to access 1000+ APIs

gopidesupavan:
Add GoogleCloudToolset to the common.ai provider so agents can call explicitly allow-listed Google APIs through pydantic-ai tools.

The toolset exposes three tools for agents:

- `list_gcp_methods` to show the allowed Google API methods
- `describe_gcp_method` to inspect parameters and schemas before calling
- `call_gcp` to execute an allowed method and return JSON

## Design

The toolset is intentionally generic rather than service-specific. Instead of implementing separate wrappers for GCS, BigQuery, Pub/Sub, Cloud Run, and other Google services, it uses Google Discovery REST metadata to resolve method names, parameters, request schemas, and response schemas at runtime. This keeps the provider small while still allowing agents to work with the Google APIs that publish Discovery documents.

Access is deny-by-default. Dag authors must pass `allowed_methods`, for example `storage/v1:objects.list` or `bigquery/v2:jobs.query`. Method-level wildcards are supported, but the API and version must be explicit so an agent cannot move from an approved stable API to an alpha or beta surface by changing tool input.

Credentials, impersonation, and project selection come from the Airflow Google connection through `GoogleBaseHook`. They are not exposed as tool arguments, so the model can choose only the allowed API method and request parameters, not which credentials or project to use.

The toolset also adds guardrails around high-risk behavior:

- Credential-returning methods are not matched by wildcards and must be allow-listed verbatim.
- Media upload/download is rejected.
- Pagination is bounded with `max_pages`.
- Returned payload size is bounded with `max_output_bytes`.
- Project parameters can be pinned to the connection project by default.

---

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
@@ -67,6 +67,6 @@ def __getattr__(name: str):
         except ImportError as e:
             from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
 
-            raise AirflowOptionalProviderFeatureException(e)
+            raise AirflowOptionalProviderFeatureException() from e
         return GoogleCloudToolset
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
@@ -20,7 +20,14 @@ from __future__ import annotations
 
 from airflow.providers.common.ai.toolsets.hook import HookToolset
 
-__all__ = ["AWSToolset", "HookToolset", "MCPToolset", "SQLToolset", "airflow_toolset_to_langchain_tools"]
+__all__ = [
+    "AWSToolset",
+    "GoogleCloudToolset",
+    "HookToolset",
+    "MCPToolset",
+    "SQLToolset",
+    "airflow_toolset_to_langchain_tools",
+]
 
 
 def __getattr__(name: str):
@@ -54,4 +61,12 @@ def __getattr__(name: str):
 
             raise AirflowOptionalProviderFeatureException() from e
         return AWSToolset
+    if name == "GoogleCloudToolset":
+        try:
+            from airflow.providers.common.ai.toolsets.google import GoogleCloudToolset
+        except ImportError as e:
+            from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+
+            raise AirflowOptionalProviderFeatureException(e)
+        return GoogleCloudToolset
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
@@ -1,0 +1,784 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Curated Google toolset exposing allow-listed Discovery API methods as pydantic-ai tools."""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+from fnmatch import fnmatchcase
+from functools import cache
+from importlib import import_module
+from typing import TYPE_CHECKING, Any
+
+try:
+    import googleapiclient.discovery_cache
+    from googleapiclient.discovery import build
+except ImportError as e:
+    from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+
+    raise AirflowOptionalProviderFeatureException(e)
+
+from pydantic_ai.exceptions import ModelRetry
+from pydantic_ai.tools import ToolDefinition
+from pydantic_ai.toolsets.abstract import AbstractToolset, ToolsetTool
+from pydantic_core import SchemaValidator, core_schema
+
+from airflow.providers.common.ai.utils.tool_definition import return_schema_kwargs
+from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from pydantic_ai._run_context import RunContext
+
+_PASSTHROUGH_VALIDATOR = SchemaValidator(core_schema.any_schema())
+
+log = logging.getLogger(__name__)
+
+_DISCOVERY_DOC_URL = "https://www.googleapis.com/discovery/v1/apis/{api}/{version}/rest"
+
+
+def _normalize_method(method: str) -> str:
+    """Comparison key for method entries: case- and underscore-insensitive."""
+    return method.replace("_", "").casefold()
+
+
+def _require_string_arg(*, tool_args: dict[str, Any], key: str) -> str:
+    value = tool_args.get(key)
+    if not isinstance(value, str) or not value:
+        raise ValueError(f"The {key!r} argument is required and must be a non-empty string.")
+    return value
+
+
+def _format_project_parameter(*, project: str, param_def: dict[str, Any]) -> str:
+    pattern = param_def.get("pattern")
+    if not isinstance(pattern, str):
+        return project
+    try:
+        if re.fullmatch(pattern, project):
+            return project
+        resource_project = f"projects/{project}"
+        if re.fullmatch(pattern, resource_project):
+            return resource_project
+    except re.error:
+        return project
+    return project
+
+
+def _get_google_base_hook_class() -> type[Any]:
+    try:
+        return import_module("airflow.providers.google.common.hooks.base_google").GoogleBaseHook
+    except ImportError as e:
+        raise AirflowOptionalProviderFeatureException(e)
+
+
+# Methods that put credentials or decrypted secrets into the agent's context.
+# Keys are version-agnostic (``api:resource.method``) so v1/v1beta variants are
+# equally protected. A wildcard in ``allowed_methods`` never matches these --
+# each must be listed verbatim (with its version) to be callable. Defense in
+# depth, not exhaustive: least-privilege IAM on the connection's service
+# account remains the real boundary.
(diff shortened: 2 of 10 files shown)


gopidesupavan:
Screenshot 2026-07-20 at 19 52 59

Screenshot 2026-07-20 at 20 38 16

Screenshot 2026-07-20 at 20 38 23

Screenshot 2026-07-20 at 20 39 06

Screenshot 2026-07-20 at 20 39 14

gopidesupavan:
follows same approach here https://github.com/apache/airflow/pull/70087

providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
```diff
@@ -20,7 +20,13 @@
 
 from airflow.providers.common.ai.toolsets.hook import HookToolset
 
-__all__ = ["AWSToolset", "HookToolset", "MCPToolset", "SQLToolset", "airflow_toolset_to_langchain_tools"]
+__all__ = [
+    "GoogleCloudToolset",
```
kaxil, line 25: Why not just have it use `HookToolset` with GCPHook?

gopidesupavan: `HookToolset` is still useful when the workflow maps to existing hook methods. The gap here is that it only exposes Python methods that a hook already wraps, while this toolset exposes allowed Google API client methods directly.

For example, an incident triage agent might need to check recent BigQuery jobs, Dataflow jobs, and GCS object arrival. With `GoogleCloudToolset`, the Dag author can allow-list the exact API methods, such as `bigquery/v2:jobs.list`, `bigquery/v2:jobs.get`, `dataflow/v1b3:projects.locations.jobs.list`, and `storage/v1:objects.list`.

Doing the same with `HookToolset` depends on each service hook having the right method and returning data in a useful shape for the agent. Some APIs also have no dedicated hook coverage. For example, Cloud Monitoring time series reads can be exposed as `monitoring/v3:projects.timeSeries.list`, while the existing Stackdriver hook mainly wraps alert policy and notification channel operations.

So this is not replacing `HookToolset`; it is for cases where the agent should call selected Google REST API methods directly, instead of being limited to the Python methods currently wrapped by Airflow hooks

gopidesupavan: One example I tested was this prompt:

```
Pipeline is slow today in project `playground-s-11-af79665a`. Check active Cloud Monitoring alert policies, list recent BigQuery pipeline rows, and confirm raw files arrived in `gs://gcp-toolset-demo-af79665a/raw/`. Then compare Cloud Monitoring metrics for BigQuery query count and GCS request count for today versus yesterday.
```

With `HookToolset`, the agent handled the parts backed by existing hook methods: alert policies, BigQuery rows, and GCS object listing. But it could not answer the Monitoring time-series part because that metric-read operation is not exposed by the hook methods available to it.

With `GoogleCloudToolset`, the agent completed the full request by calling the relevant Google REST API methods directly, including `monitoring/v3:projects.timeSeries.list`.

This is just one example; in other multi-service troubleshooting questions the missing piece may be a different Google API method. ideally here it is not to replace `HookToolset`, but to support cases where we want to expose selected Google API methods without needing a dedicated Airflow hook wrapper for each one.

gopidesupavan:
I’d welcome feedback from the Google provider team @MaksYermak

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+import json
+import logging
+import os
+import re
+from fnmatch import fnmatchcase
+from functools import cache
+from importlib import import_module
+from typing import TYPE_CHECKING, Any
+
+try:
```
Lee-W: Let's move these imports to the button of import section for better readiblity

providers/common/ai/src/airflow/providers/common/ai/toolsets/__init__.py
```diff
@@ -54,4 +61,12 @@ def __getattr__(name: str):
 
             raise AirflowOptionalProviderFeatureException() from e
         return AWSToolset
+    if name == "GoogleCloudToolset":
+        try:
+            from airflow.providers.common.ai.toolsets.google import GoogleCloudToolset
+        except ImportError as e:
+            from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+
+            raise AirflowOptionalProviderFeatureException(e)
```
Lee-W:
```suggestion
            raise AirflowOptionalProviderFeatureException() from e
```

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
Lee-W: let's move it to the button of import section

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+    except re.error:
+        return project
+    return project
+
+
+def _get_google_base_hook_class() -> type[Any]:
+    try:
+        return import_module("airflow.providers.google.common.hooks.base_google").GoogleBaseHook
+    except ImportError as e:
+        raise AirflowOptionalProviderFeatureException(e)
```
Lee-W:
```suggestion
        raise AirflowOptionalProviderFeatureException() from e
```

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+        self._gcp_conn_id = gcp_conn_id
+        self._impersonation_chain = impersonation_chain
+        self._enforce_project = enforce_project
+        self._allow_remote_discovery = allow_remote_discovery
+        self._max_pages = max_pages
+        self._max_output_bytes = max_output_bytes
+        self._patterns = patterns
+        self._literal_methods = frozenset(literal_methods)
+        # Populated lazily at call time, so Dag parsing never touches network
+        # or credentials.
```
Lee-W:
```suggestion
        # Populated lazily at call time, so Dag parsing never touches network or credentials.
```

nit: I feel it's easier to read without breakline

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+                {
+                    "truncated": True,
+                    "max_output_bytes": self._max_output_bytes,
+                    "data": payload[: self._max_output_bytes],
+                }
+            )
+        return payload
+
+
+# ---------------------------------------------------------------------------
```
Lee-W: Do we need this comment block? it doesn't seem to provide much information

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+            payload["more_pages_available"] = True
+        return self._serialize_response(payload)
+
+    # ------------------------------------------------------------------
+    # Project pinning
+    # ------------------------------------------------------------------
+
+    def _apply_project_guard(self, *, method_doc: dict[str, Any], params: dict[str, Any]) -> None:
+        project = self._get_hook().project_id
+        if not project:
```
Lee-W:
```suggestion
        if not (project := self._get_hook().project_id):
```

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+        # through construction; re-checked here in case the bundle differs between
+        # the parsing and execution environments. Fetched lazily at call time,
+        # never at Dag parse time.
+        if not self._allow_remote_discovery:
+            raise ValueError(
+                f"API {api}/{version} has no bundled discovery document and "
+                "allow_remote_discovery is disabled."
+            )
+        if key not in self._remote_docs:
+            import requests
```
Lee-W: I think we can move it to the top

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+            for method_name in resource.get("methods", {}):
+                canonical = f"{prefix}{resource_name}.{method_name}"
+                paths[_normalize_method(canonical)] = canonical
+            if "resources" in resource:
+                walk(resource["resources"], f"{prefix}{resource_name}.")
+
+    walk(doc.get("resources", {}), "")
+
+    api_name = doc.get("name")
+    if api_name:
```
Lee-W:
```suggestion
    if (api_name := doc.get("name")):
```

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+            return out
+        if "additionalProperties" in node and depth > 0:
+            return {"<key>": _describe_schema(doc, node["additionalProperties"], depth - 1, seen)}
+        return type_name
+    if type_name == "array":
+        if depth > 0:
+            return [_describe_schema(doc, node.get("items", {}), depth - 1, seen)]
+        return type_name
+    enum = node.get("enum")
+    if enum:
```
Lee-W:
```suggestion
    if (enum := node.get("enum")):
```

Lee-W (commented):
some nits but overall looks good

guan404ming (approved):
Besides @Lee-W's comments, LGTM

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
```diff
@@ -0,0 +1,814 @@
...
+    pairs = set()
+    for filename in os.listdir(_bundled_docs_dir()):
+        if filename.endswith(".json"):
+            api, _, version = filename[:-5].partition(".")
+            if version:
+                pairs.add((api, version))
+    return frozenset(pairs)
+
+
+@cache
```
guan404ming: It worth caching the derived method-path map instead of the whole document, so the Dag parse stays cheap. The full document is only needed at call time on the worker.

guan404ming (commented):
I found that Google's documents are big. compute/v1, the docstring's own example, costs 68 ms and 10.2 MB resident, and @cache keeps it for the life of the parse process.

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
gopidesupavan: good catch updated it now

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
gopidesupavan: no removed

providers/common/ai/src/airflow/providers/common/ai/toolsets/google.py
guan404ming: Thanks!

gopidesupavan:
converting to draft for now to align few interface decisions. @vikramkoka converted this to draft.

Lee-W:
are we going to have a dev list thread for that?

vikramkoka:
> converting to draft for now to align few interface decisions. @vikramkoka converted this to draft.

Thanks @gopidesupavan

kaxil:
Closing this since we discussed the same with Pavan on the AWS API toolset PR

PR #70147, opened 2026-07-20
Support schema-qualified table names in db clean

seanmuth:
Celery result backends are commonly provisioned into their own database schema, separate from Airflow's metadata tables. `airflow db clean` has no way to reach `celery_taskmeta` / `celery_tasksetmeta` in that case, since it only ever looks in the connection's default schema, so those tables are silently skipped every time.

`_TableConfig`, `reflect_tables`, and the archive/export/drop machinery in `db_cleanup.py` now understand schema-qualified table names (`"schema.table"` dot notation). A new `[celery] result_backend_schema` config option lets a deployment point the built-in `celery_taskmeta` / `celery_tasksetmeta` configs at the schema they actually live in, without needing a fork or manual per-table config.

Verified end-to-end against a real Postgres instance: clean → archive (lands in the same schema as the source table) → export → drop, all schema-aware. New unit + integration tests cover schema parsing, the config-driven wiring, and the full cleanup cycle; existing test suite passes unmodified on both sqlite and Postgres.

---

### DIFF
--- a/airflow-core/src/airflow/utils/db_cleanup.py
+++ b/airflow-core/src/airflow/utils/db_cleanup.py
@@ -63,12 +63,22 @@ ARCHIVED_TABLES_FROM_DB_MIGRATIONS = [
 ]
 
 
+def _split_schema_table(qualified_name: str) -> tuple[str | None, str]:
+    """Split a possibly schema-qualified table name (``schema.table``) into ``(schema, table)``."""
+    schema, sep, name = qualified_name.partition(".")
+    if sep:
+        return schema, name
+    return None, qualified_name
+
+
 @dataclass
 class _TableConfig:
     """
     Config class for performing cleanup on a table.
 
-    :param table_name: the table
+    :param table_name: the table name. May be schema-qualified with dot notation
+        (e.g. ``"celery.celery_taskmeta"``) for a table that lives in a schema other than the
+        connection's default, such as a Celery result backend provisioned into its own schema.
     :param extra_columns: any columns besides recency_column_name that we'll need in queries
     :param recency_column_name: date column to filter by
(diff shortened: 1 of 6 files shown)


airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -142,6 +162,18 @@ def readable_config(self):
...
+# We need `fallback=None` because this is executed at top level code and provider configuration
+# (the celery provider) might not be loaded. When set, this points db clean at the schema the
+# Celery result backend tables (celery_taskmeta, celery_tasksetmeta) actually live in, for
+# deployments that provision the result backend into a schema other than the metadata DB's default
+# (a common pattern to keep Celery's own tables separate from Airflow's).
+_celery_result_backend_schema = conf.get("celery", "result_backend_schema", fallback=None)
+
+
+def _celery_table_name(name: str) -> str:
+    return f"{_celery_result_backend_schema}.{name}" if _celery_result_backend_schema else name
```
eladkal: I am not sure about celery related code from airflow core.
It's a pattern we try to do the other way.. extract all provider related stuff out

uranusjr:
I added a commit that removed the reading-provider-in-core part. This PR now only contains adding support for schema-qualified table names in db clean.

Once this is accepted, I will raise a second PR that implements the Celery table name cleanup with a cleaner approach.

uranusjr (approved):

eladkal:
> I added a commit that removed the reading-provider-in-core part. This PR now only contains adding support for schema-qualified table names in db clean.
> 
> Once this is accepted, I will raise a second PR that implements the Celery table name cleanup with a cleaner approach.

LGTM then

eladkal (approved):

PR #67449, opened 2026-05-24
Feat/kubernetes executor callback support

sjyangkevin:
Implements `supports_callbacks` on KubernetesExecutor by running each ExecuteCallback workload as its own pod, alongside the existing task-pod pipeline. Gated on `AIRFLOW_V_3_3_PLUS`.

* related: #62887
* related: #61153
* related: #62645 
* continue work on: #63454

### Approach

Use the same way as how Task is dispatched to Kubernetes but with a different configuration for pod label/annotation.

**Pod annotations: tasks vs callbacks**

For task pods, `PodGenerator.construct_pod()` writes these annotations so the watcher can reconstruct `TaskInstanceKey`:

```
dag_id=<dag_id>
task_id=<task_id>
try_number=<n>
run_id=<run_id>
map_index=<n>   (if mapped)
```

For callback pods, there is no `task_id`, `try_number`, or `map_index`. The relevant identity is the callback UUID. 

Proposed annotations:

```
callback_id=<uuid-str>   ← primary identity, used to reconstruct CallbackKey
dag_id=<dag_id>          ← from ExecuteCallback.log_path (executor_callbacks/<dag_id>/...)
run_id=<run_id>          ← same source, for observability/log correlation
```

A new pod label is also needed to let the watcher distinguish callback pods from task pods without relying on annotation presence:

```
airflow-workload-type=callback
```

**Pod construction for callbacks: no `PodGenerator.construct_pod()` changes**

`PodGenerator.construct_pod()` is specifically built for task identity (takes `dag_id`, `task_id`, `try_number`, etc.). **Callback pods should not go through `construct_pod()`**.  They need their own construction path that:

1. Starts from the same base pod template (`get_base_pod_from_template`)
2. Sets the correct command: `workload_to_command_args(workload)` for `ExecuteCallback`, which serializes workload to `--json-string`
3. Sets labels: `airflow-worker=<job_id>`, `kubernetes_executor=True`, `airflow-workload-type=callback`
4. Sets annotations: `callback_id`, `dag_id`, `run_id` (extracted from `workload.log_path` or `workload.callback.data`)
5. Sets `pod_id` from `create_unique_id("callback", callback_id_short)` to avoid name collisions with task pods

The command is identical because `execute_workload` dispatches on the workload type at runtime — the same subprocess entry point handles both `ExecuteTask` and `ExecuteCallback`.

**Differences from the task pod lifecycle:**

| Step | Task pod | Callback pod |
|---|---|---|
[...]

PR #70148, opened 2026-07-20
 allow deadline alert UUID references in serialized Dag schema

JH-A-Kim:
---

# Description

This pr is a continuation of PR #68195 
Related: PR #67449 

What happens: When a Dag defines a deadline alert, SerializedDagModel._generate_deadline_uuids rewrites the Dag's deadline field from a list of dicts into a list of UUID strings referencing rows in the deadline_alert table. This has a few different problems. 

1. schema.json only allowed deadline to be a dict, an array of dicts, or null so this would throw a validation error. 

2. _generate_deadline_uuids rewrote the passed-in dag.data in place. Any of the other code that still used the reference would then get something completely different from what they expected, that being a dict of deadlines as opposed to a list of UUIDs

3. One of the possible fixes to the above using a deepcopy would then create a separate issue wherein `test_dagrun_success_deadline` and `test_dagrun_deadline_variable_interval_stable` would be using the original version of the dag data that has deadline as an actual dict instead of the changed version with deadlines as a list of strings that are UUIDs.

The fix

1. Add an array of string option to the deadline schema, so it can accept the UUID.

2. deepcopy dag.data inside write_dag before writing any of the UUID information so we prevent mutation of the caller object.
 
3. In _write_dag, do session.flush() to get the updated information to the db and then from there just call down from the db as the single source of truth. Important thing to note is there is no commit so only that session sees the changes and no other sessions see it. (tldr of this fix is to make it so that it acts like a client server contract with video games wherein all the important information needs to be gathered from the source of truth and not locally.)

Generated-by:  Sonnet 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/src/airflow/models/serialized_dag.py
+++ b/airflow-core/src/airflow/models/serialized_dag.py
@@ -19,6 +19,7 @@
 
 from __future__ import annotations
 
+import copy
 import logging
 import zlib
 from collections.abc import Callable, Iterable, Iterator, Sequence
@@ -645,6 +646,12 @@ class SerializedDagModel(Base):
         name_updated = False
         reused_deadline_data: dict[str, dict] | None = None
         if dag.data.get("dag", {}).get("deadline"):
+            # The deadline handling below rewrites data["dag"]["deadline"] from a list of
+            # encoded dicts into a list of UUID references. Work on a copy so we never mutate
+            # the caller's LazyDeserializedDAG in place.
+            from airflow.serialization.serialized_objects import LazyDeserializedDAG
+
+            dag = LazyDeserializedDAG(data=copy.deepcopy(dag.data), last_loaded=dag.last_loaded)
             # Try to reuse existing deadline UUIDs if the deadline definitions haven't changed.
             # This preserves the hash and avoids unnecessary SerializedDagModel recreations.
             existing_serialized_dag = session.scalar(
--- a/airflow-core/src/airflow/models/serialized_dag.py
+++ b/airflow-core/src/airflow/models/serialized_dag.py
@@ -649,9 +649,8 @@ class SerializedDagModel(Base):
             # The deadline handling below rewrites data["dag"]["deadline"] from a list of
             # encoded dicts into a list of UUID references. Work on a copy so we never mutate
             # the caller's LazyDeserializedDAG in place.
-            from airflow.serialization.serialized_objects import LazyDeserializedDAG
 
-            dag = LazyDeserializedDAG(data=copy.deepcopy(dag.data), last_loaded=dag.last_loaded)
+            dag = dag.model_copy(update={"data": copy.deepcopy(dag.data)})
             # Try to reuse existing deadline UUIDs if the deadline definitions haven't changed.
             # This preserves the hash and avoids unnecessary SerializedDagModel recreations.
             existing_serialized_dag = session.scalar(
--- a/airflow-core/src/airflow/serialization/schema.json
+++ b/airflow-core/src/airflow/serialization/schema.json
@@ -201,11 +201,6 @@
                     "type": "array",
                     "items": { "$ref": "#/definitions/dict" }
                 },
-                {
-                    "$comment": "A Dag's deadline alerts are saved as rows in the deadline_alert table, so once stored the serialized Dag keeps only a list of UUID strings pointing at those rows (see SerializedDagModel._generate_deadline_uuids).",
-                    "type": "array",
-                    "items": { "type": "string" }
-                },
                 { "type": "null" }
             ]
         },
--- a/airflow-core/src/airflow/serialization/schema.json
+++ b/airflow-core/src/airflow/serialization/schema.json
@@ -201,6 +201,11 @@
                     "type": "array",
                     "items": { "$ref": "#/definitions/dict" }
                 },
+                {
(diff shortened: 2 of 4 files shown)


JH-A-Kim:
@ferruzzi may I have a review on this? Thank you!

airflow-core/tests/unit/models/test_serialized_dag.py
```diff
@@ -1155,3 +1156,99 @@ def test_deadline_reuse_skips_write_when_hash_matches(self, testing_dag_bundle,
...
+        assert original_deadline
+        assert all(isinstance(item, dict) for item in original_deadline)
+
+        SDM.write_dag(lazy_dag, bundle_name="testing", session=session)
+        session.commit()
+
+        assert lazy_dag.data["dag"]["deadline"] == original_deadline
+
+    def test_sync_dag_to_db_returns_db_normalized_deadline_ids(self, testing_dag_bundle, session):
+        """sync_dag_to_db must return a SerializedDAG with the DB-normalized deadline UUIDs. Basically just verify that the UUIDs returned by sync_dag_to_db match the persisted deadline_alert rows in the DB."""
```
ferruzzi:
```suggestion
        """
        sync_dag_to_db must return a SerializedDAG with the DB-normalized deadline UUIDs.  
        
        Verify that the UUIDs returned by sync_dag_to_db match the persisted deadline_alert rows in the DB.
        """
```

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -645,6 +646,12 @@ def write_dag(
         name_updated = False
         reused_deadline_data: dict[str, dict] | None = None
         if dag.data.get("dag", {}).get("deadline"):
+            # The deadline handling below rewrites data["dag"]["deadline"] from a list of
+            # encoded dicts into a list of UUID references. Work on a copy so we never mutate
+            # the caller's LazyDeserializedDAG in place.
+            from airflow.serialization.serialized_objects import LazyDeserializedDAG
```
ferruzzi: ~~I don't think this needs to be a local import; move it to the top if you can.~~

Edit:  Actually, see below.

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -645,6 +646,12 @@ def write_dag(
...
+
+            dag = LazyDeserializedDAG(data=copy.deepcopy(dag.data), last_loaded=dag.last_loaded)
```
ferruzzi: This is fragile.  If someone ever adds a new field to `LazyDeserializedDAG`, this will silently drop it and would be very hard to troubleshoot.  Instead, let's make a copy keeping the existing parameters and just update `data`:


```suggestion
            dag = dag.model_copy(update={"data": copy.deepcopy(dag.data)})
```

Some might say the deepcopy isn't strictly required here, but this is already an O(n) method since the hash calculation is already walking the full dag data anyway.  It's not adding any meaningful extra overhead for the safety it provides.  I like it.
 
In fact, if we do that, it also eliminates the new `LazyDeserializedDAG` import above.

ferruzzi (commented):
Left a couple small comments.

ferruzzi:
Pretty sure those failing CI tests were just fixed in main, give it a rebase and they should go green.

JH-A-Kim:
@ferruzzi i believe this is ready for review again!

airflow-core/src/airflow/serialization/schema.json
```diff
@@ -201,6 +201,11 @@
                     "type": "array",
                     "items": { "$ref": "#/definitions/dict" }
                 },
+                {
+                    "$comment": "Once persisted, a Dag's deadline alerts live as rows in the deadline_alert table and the serialized Dag keeps only a list of UUID strings referencing them (see SerializedDagModel._generate_deadline_uuids). This branch lets the stored form validate at any lifecycle stage, not only before the dict->UUID rewrite.",
```
ferruzzi, line 205: I don't see `$comment`s anywhere else, which makes this feel a bit out of place?

airflow-core/src/airflow/serialization/schema.json
```diff
@@ -201,6 +201,11 @@
...
+                    "type": "array",
+                    "items": { "type": "string" }
```
ferruzzi: As written, this accepts any string, we could tighten it by adding a regex pattern to validate that it is a UUID7:

```json
  "items": {
      "type": "string",
      "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
  }
```

ferruzzi (approved):
Approved with two suggestions/nits

airflow-core/src/airflow/serialization/schema.json
JH-A-Kim: For this I tried to preserve Kevins original code but I can remove the comments if you would like?

JH-A-Kim: And in the `schema.json` there are a few different places using `$comment` such as in `_is_mapped`, `operator`, `task_group` and a few other places although the comments they have are significantly shorter.

airflow-core/src/airflow/serialization/schema.json
JH-A-Kim: I added this in commit 834dcd6 !

ferruzzi:
LGTM, pending second approval.

PR #69148, opened 2026-06-30
UI: Allow scoping plugin tabs to specific Dags and tasks

dheerajturaga:
Plugin external views and React apps that target the Dag, Dag run, task, or task instance pages appear on every Dag, with no way to limit a tab to the Dags and tasks it is relevant to. A deployment running several Dag-specific plugins ends up with tabs that are dead weight on unrelated Dags, and a plugin cannot hide its own tab from inside itself.

This adds a single optional `applies_to` block:

```python
"applies_to": {
    "dag_tags": ["ml"],                      # Dag carries any of these tags
    "dag_ids": ["train_pipeline"],           # exact dag_id
    "task_ids": ["train_model"],             # exact task_id
    "operators": ["KubernetesPodOperator"],  # operator class name
    "operator_names": ["@task.bash"],         # UI/display name
}
```

All keys are optional; omitting `applies_to` shows the item everywhere, which is the existing behaviour. `operators` matches operator class names only. `operator_names` matches UI/display names (`custom_operator_name`), including decorator names such as `@task.bash`.

Criteria combine like Kubernetes label selectors — **OR within a key, AND across keys** — but the AND applies only across the keys the current page can *evaluate*. A `task_ids` criterion cannot be judged on a Dag-level page, so it is skipped there rather than failing the match. That lets one `applies_to` block be shared by a plugin's Dag- and task-level destinations instead of needing one per destination.

Scoping is also applied to the `dag_overview` and `task_overview` React app destinations, which filter locally rather than going through `usePluginTabs` — without that, a scoped plugin still rendered its panel on every Dag.

Plugin showing on all Dags:
all_dags_plugin

Plugins showing only for the tutorial Dag:
tutorial_only_plugin

Plugins showing for `example_*` Dags:
example_only_plugin

**On the feedback in this thread:**

- *API sprawl* — the schema gains exactly one field, ever. Future criteria (e.g. `asset_groups` for asset pages) become keys inside `PluginAppliesToResponse` rather than new top-level parameters. I've deliberately kept this PR to the Dag/task axis so the model can be reviewed on its own; assets are a natural follow-up.
- *Dependency direction* — `dag_tags` puts the opt-in on the Dag side: the Dag author tags their Dag to receive the tab. `operators` is the case where a plugin legitimately knows what it applies to. `dag_ids` and `task_ids` cover deployments where the Airflow operator, not the plugin author, knows the mapping.
[...]

PR #70149, opened 2026-07-21
UI: Pass Dag, DagRun, TaskInstance, and Asset objects to React app plugins

dheerajturaga:
Follows the pattern established in #69986, which enriched the props passed to React app plugins beyond bare route params by resolving assetUri from the UI's TanStack Query cache. This extends that pattern to the full context objects: React plugins now receive dag, dagRun, taskInstance, and asset as typed props alongside the existing route-id strings.

Each object is resolved from the query cache and gated on the route params it depends on, so it is served from the cache the parent details page already populated (no extra request) where present, and is undefined on routes or destinations that lack those ids (e.g. nav, base, dashboard).

Related: #69148 

Screenshots from dummy plugin:
image

image

---

Claude Code Opus 4.8

### DIFF
--- a/airflow-core/src/airflow/ui/src/pages/ReactPlugin.tsx
+++ b/airflow-core/src/airflow/ui/src/pages/ReactPlugin.tsx
@@ -20,18 +20,33 @@ import { Spinner } from "@chakra-ui/react";
 import { type FC, lazy, Suspense } from "react";
 import { useParams } from "react-router-dom";
 
-import { useAssetServiceGetAsset } from "openapi/queries";
-import type { ReactAppResponse } from "openapi/requests/types.gen";
+import {
+  useAssetServiceGetAsset,
+  useDagRunServiceGetDagRun,
+  useDagServiceGetDagDetails,
+  useTaskInstanceServiceGetMappedTaskInstance,
+} from "openapi/queries";
+import type {
+  AssetResponse,
+  DAGDetailsResponse,
+  DAGRunResponse,
+  ReactAppResponse,
+  TaskInstanceResponse,
+} from "openapi/requests/types.gen";
 
 import { ErrorPage } from "./Error";
 
 export type PluginProps = {
+  asset?: AssetResponse;
   assetId?: string;
   assetUri?: string;
+  dag?: DAGDetailsResponse;
   dagId?: string;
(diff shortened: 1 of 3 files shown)


airflow-core/src/airflow/ui/src/pages/ReactPlugin.tsx
```diff
@@ -85,12 +122,16 @@ export const ReactPlugin = ({ reactApp }: { readonly reactApp: ReactAppResponse
...
+        asset={asset}
         assetId={assetId}
         assetUri={assetUri}
+        dag={dag}
         dagId={dagId}
+        dagRun={dagRun}
         mapIndex={mapIndex}
         runId={runId}
         taskId={taskId}
+        taskInstance={taskInstance}
```
bbovenzi: Let's just make this a `const pluginProps = {}` and then we can pass the same content to plugin and lazy plugin. `<Plugin {...pluginProps} />` A little more DRY

airflow-core/src/airflow/ui/src/pages/ReactPlugin.tsx
```diff
@@ -67,15 +82,37 @@ const loadPlugin = (reactApp: ReactAppResponse): Promise<{ default: PluginCompon
...
+  const { data: taskInstance } = useTaskInstanceServiceGetMappedTaskInstance(
+    {
+      dagId: dagId ?? "",
+      dagRunId: runId ?? "",
+      mapIndex: mapIndex === undefined ? -1 : parseInt(mapIndex, 10),
+      taskId: taskId ?? "",
+    },
+    undefined,
+    { enabled: Boolean(dagId) && Boolean(runId) && Boolean(taskId) },
+  );
```
bbovenzi, line 114: These _should_ be fine  and not firing any extra queries since they should already be cached.

bbovenzi (commented):
Let's update the documentation to let plugin developers know this and that its only for react plugins of course.

airflow-core/src/airflow/ui/src/pages/ReactPlugin.tsx
dheerajturaga: Done!

bbovenzi (approved):

PR #70150, opened 2026-07-21
UI: Add Run ID filter to the Grid view

dheerajturaga:
The Dag Runs list page lets users narrow runs by Run ID, but the Grid view's filter bar offered no equivalent — forcing users to eyeball run columns to locate a specific run. This adds the same Run ID search to the Grid view so both surfaces filter consistently.

grid_view_run_id_filter

Additional Screenshots:
image

image

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
+++ b/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
@@ -1282,6 +1282,32 @@ paths:
           title: Triggering User Prefix
         description: Case-sensitive, index-friendly prefix match. See "Filtering with
           pattern parameters".
+      - name: run_id_pattern
+        in: query
+        required: false
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: "Case-insensitive substring match (SQL `ILIKE`). Slower than\
+            \ `run_id_prefix_pattern` on large tables \u2014 see \"Filtering with\
+            \ pattern parameters\"."
+          title: Run Id Pattern
+        description: "Case-insensitive substring match (SQL `ILIKE`). Slower than\
+          \ `run_id_prefix_pattern` on large tables \u2014 see \"Filtering with pattern\
+          \ parameters\"."
+      - name: run_id_prefix_pattern
+        in: query
+        required: false
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: Case-sensitive, index-friendly prefix match. See "Filtering
+            with pattern parameters".
+          title: Run Id Prefix Pattern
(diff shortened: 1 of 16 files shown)


bbovenzi (commented):
Could you show a screenshot of what the filter dropdown, and form look like too?

We're adding run_id pattern and the prefix pattern. We should either use both of them in the UI, with the prefix toggle button or drop the prefix pattern from the API.

dheerajturaga:
> Could you show a screenshot of what the filter dropdown, and form look like too?
> 
> We're adding run_id pattern and the prefix pattern. We should either use both of them in the UI, with the prefix toggle button or drop the prefix pattern from the API.

@bbovenzi, Ive added more screenshots in the description. I also dropped the prefix pattern from the pr, its not needed. Hope this looks good?

bbovenzi:
Screenshot 2026-07-22 at 11 57 44 AM


Ok with it removed now we need to disable the advanced toggle on the right side here.

dheerajturaga:
> Screenshot 2026-07-22 at 11 57 44 AM
> Ok with it removed now we need to disable the advanced toggle on the right side here.

Done! 
image

airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
```diff
@@ -1328,6 +1328,27 @@ paths:
...
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: "SQL LIKE expression \u2014 use `%` / `_` wildcards (e.g. `%customer_%`).\
+            \ Use the pipe `|` operator for OR logic (e.g. `dag1 | dag2`). Regular\
+            \ expressions are **not** supported. \n\n**Performance note:** this full-match\
+            \ pattern is evaluated as ``ILIKE '%term%'`` and most of the time prevents\
+            \ the database from using B-tree indexes, which can be very slow on large\
+            \ tables. Prefer the equivalent ``run_id_prefix_pattern`` parameter when\
```
bbovenzi: The docs are inaccurate without the prefix_pattern param. Thinking of it, I am sorry to make you go back. But I think its best to keep this pattern search consistent with all our others and to have a prefix_pattern on the grid endpoint too.

dheerajturaga: No worries! Let me wire this in and test it out. Happy to make this perfect regardless of how many back and forth it takes

dheerajturaga:
@bbovenzi, I added back the pattern match, let me know if this is good

image

image

image

bbovenzi (approved):
LGTM. Let's just rebase and rebuild the API docs

vatsrahul1001:
Moving to 3.4.0 this is more suitable for minor as its a new feature

PR #70152, opened 2026-07-21
Narrow AirflowException to specific exceptions in DataSync operator

haseebmalik18:
The AWS DataSync operator raised the catch-all `AirflowException` for ten unrelated conditions, so a DAG author could not distinguish an invalid constructor call from a failed task execution when handling errors.

This narrows each usage to a specific type:

- **Usage/validation errors** now raise `ValueError`.
- **Runtime domain failures** raise dedicated exceptions added to
  `aws/exceptions.py`:
  - `DataSyncTaskNotFoundError`
  - `DataSyncMultipleTasksError` / `DataSyncMultipleLocationsError`
  - `DataSyncLocationNotFoundError`
  - `DataSyncTaskCreationError`
  - `DataSyncTaskExecutionFailedError`

The new classes subclass `AirflowException`, so existing `except AirflowException` handlers keep working, with no breaking change.

Tests are updated to assert the specific types, with added coverage for the raise sites that were previously untested.

Part of the ongoing effort to reduce direct `AirflowException` usage. The `check-no-new-airflow-exceptions` allowlist entry for this file is tightened from 10 to 0.


providers/amazon/src/airflow/providers/amazon/aws/operators/datasync.py
```diff
@@ -181,7 +192,7 @@ def __init__(
         if self.source_location_uri and self.destination_location_uri:
             valid = True
         if not valid:
-            raise AirflowException(
+            raise ValueError(
```
Sanjays2402: switching this from AirflowException to ValueError is a behavior change for anyone catching AirflowException around operator construction, since ValueError isn't in that hierarchy. the raises in execute() all stay AirflowException subclasses, so this one's now the odd one out. intentional?

haseebmalik18: Yeah, intentional. The constructor check is argument validation, so ValueError fits it better, while the execute failures stay AirflowException subclasses. It's also at parse time where the DAG loader catches broad Exception, so a bad config still surfaces as a normal import error.

o-nikolas:
This needs a rebase @haseebmalik18, I think they may imply some changes on this one too.

haseebmalik18:
Rebased

haseebmalik18:
@o-nikolas

haseebmalik18:
@o-nikolas Following up

o-nikolas (approved):

providers/amazon/src/airflow/providers/amazon/aws/operators/datasync.py
o-nikolas: @ferruzzi and @vincbeck, this seems fine to me. Agreed?

vincbeck (approved):

providers/amazon/src/airflow/providers/amazon/aws/operators/datasync.py
vincbeck: Yes, it also can be considered a bug fix because it should have been `ValueError` at the first place

haseebmalik18:
Merge conflicts resolved

PR #70159, opened 2026-07-21
Show backfill Dag runs in the UI

shivaam:
Use the backfill Dag-run entries API added in #67381 to make the slots and Dag runs associated with a backfill inspectable in the UI.

## What changed

- Open a cover-sized dialog from a backfill in the Dag Backfills table.
- Give the open dialog a shareable `/dags/{dag_id}/backfills/{backfill_id}` URL with normal close and back navigation.
- Show a paginated slots table with either Logical Date or Partition Key, Dag run state, not-created reason, Run ID, and slot ordinal.
- Link created entries to their Dag run while keeping not-created slots visible with their recorded reason.
- Translate known not-created reasons and preserve unfamiliar backend values instead of hiding them as `Unknown`.
- Link the active-backfill banner back to the Backfills tab.

## Screenshot

Backfill slots shown in a cover-sized dialog

## Request behavior

Opening the dialog requests the selected backfill metadata and the currently visible page of slots. While the selected backfill is unfinished, both requests follow the configured auto-refresh interval.

After the backfill completes, the visible slot page continues refreshing only if it still contains a non-terminal created Dag run. This allows the table to receive its final state before polling stops. Closing the dialog disables both modal-specific queries.

Changing pages requests only the selected page. The UI does not sweep hidden pages and does not require an aggregation endpoint.

Related: #67381

## Testing

- `pnpm vitest run src/pages/Dag/Backfills/Backfills.test.tsx src/components/Banner/BackfillBanner.test.tsx` — 4 tests passed
- `pnpm lint`
- `pnpm build`
- Local production Airflow validation for direct links, close navigation, Logical Date and Partition Key backfills, translated and skipped slots, and active polling
- Live browser trace: 3 metadata polls and 3 slot-page polls while open; 0 modal requests after close

---

- [x] Yes — Codex (GPT-5)

---


bbovenzi:
I am skeptical that we need a full details page because we have the same info available in the Backfills table view. But we should make it easy to filter the dag runs list by backfill.

shivaam:
Thanks, that’s fair—the Details tab largely duplicates metadata already shown in the Backfills table and I’ll remove it.

I think the dedicated route is still useful for the complete set of backfill slots. Some slots have no Dag run because creation was skipped, so they have `dag_run_id = null` plus a reason such as `in flight`. Those entries cannot appear in the existing Dag Runs list.

I propose keeping the page focused on the slot list and existing backfill controls. A backfill filter on the general Dag Runs list could still be a useful complementary improvement.

bbovenzi:
You mean number of backfill slots that do not have a run yet?

Let's put that in a modal perhaps? Then we don't have to worry about confusing users about where they just navigated to

shivaam:
> You mean number of backfill slots that do not have a run yet?

When Airflow processes a backfill, it evaluates each logical slot in the requested date range. For each slot, it decides whether to create a Dag run. That decision is separate from executing the run:
- If Airflow creates a run, the slot links to it and exposes its current state.
- If Airflow does not create a run, the slot records the reason, such as `in flight` or `already exists`.
The latter is a completed creation decision, not a slot still waiting for a run.

> Let’s put that in a modal perhaps?

A modal makes sense for keeping users in context, but I’m concerned about putting a potentially large, paginated slot collection inside it. Could we instead keep a lightweight route containing only the slots table, reached through an explicit “View slots” action with clear back navigation?

That said, I’m also fine with a modal if that’s the preferred direction. The main thing I’d like to preserve is somewhere users can inspect both the created Dag runs and the slots for which no run was created.

bbovenzi:
I think a `cover` sized dialog modal would work well here https://www.chakra-ui.com/docs/components/dialog#cover

shivaam:
> I think a `cover` sized dialog modal would work well here https://www.chakra-ui.com/docs/components/dialog#cover

I tried it and it looks good. I have updated the pr with just the modal. Thanks for the suggestion.

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/BackfillDagRunsModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const [pageIndex, setPageIndex] = useState(0);
+  const tableState = {
+    pagination: {
+      pageIndex,
+      pageSize,
+    },
+    sorting: [],
+  } satisfies TableState;
+  const refetchInterval = useAutoRefresh({ dagId: backfill?.dag_id });
+  const shouldPoll = backfill?.completed_at === null && !backfill.is_paused;
```
bbovenzi: We need to add autoRefresh to the get backfills query in `Backfills.tsx`. Otherwise this will never change.

airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
```diff
@@ -7,9 +7,15 @@
     "backwards": "Run Backwards",
     "dateRange": "Date Range",
     "errorStartDateBeforeEndDate": "Start Date must be before the End Date",
+    "exceptionReason": {
+      "alreadyExists": "Already exists",
+      "inFlight": "In flight",
+      "unknown": "Unknown"
```
bbovenzi, line 13: Should we show unknown as an error or just show whatever the raw error message is?

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/Backfills.tsx
```diff
@@ -106,18 +117,18 @@ const getColumns = (translate: (key: string) => string): Array<ColumnDef<Backfil
 export const Backfills = () => {
   const { t: translate } = useTranslation();
   const { setTableURLState, tableURLState } = useTableURLState();
+  const [selectedBackfill, setSelectedBackfill] = useState<BackfillResponse>();
```
bbovenzi: To better support this as a page. We can store this in the url params instead. so if you link to `/dags/{dag_id}/backfills/{backfill_id}` then the cover modal is already open.

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/BackfillDagRunsModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+    case "in flight":
+      return translate("components:backfill.exceptionReason.inFlight");
+    default:
+      return translate("components:backfill.exceptionReason.unknown");
+  }
+};
+
+const getColumns = (translate: (key: string) => string): Array<ColumnDef<BackfillDagRunResponse>> => [
+  {
+    accessorKey: "logical_date",
```
bbovenzi: If it can only be either a partition or logical date per backfill then we should just toggle to show one column or the other instead of mixing it like this.

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/BackfillDagRunsModal.tsx
shivaam: Addressed in 9a965b6a92. The modal now fetches the selected backfill by URL ID and polls its metadata while unfinished. The visible slot page also keeps polling until any stale queued or running rows reach a terminal state, then both queries stop; closing the modal disables them.

airflow-core/src/airflow/ui/public/i18n/locales/en/components.json
shivaam: Updated. Known enum values remain translated (already exists, in flight, and unknown), while any unfamiliar backend value is displayed verbatim instead of being collapsed to Unknown.

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/Backfills.tsx
shivaam: Implemented in 9a965b6a92. Opening a backfill now navigates to /dags/{dag_id}/backfills/{backfill_id}, which opens the cover dialog and supports direct links and browser navigation. The modal fetches that backfill directly, so deep links do not depend on the currently paginated Backfills row.

airflow-core/src/airflow/ui/src/pages/Dag/Backfills/BackfillDagRunsModal.tsx
shivaam: Updated. The backend enforces one slot type per backfill, so the table now renders either a Partition Key column or a Logical Date column for the whole backfill.

bbovenzi (approved):

vatsrahul1001:
https://github.com/apache/airflow/pull/67381 is marked for 3.4.0 as its a new feature updating milestone for this as 3.4.0 as well

PR #70160, opened 2026-07-21
Add Oracle Cloud Infrastructure provider and Generative AI hook

AlejandroMorgante:
Adds a dedicated Oracle Cloud Infrastructure (OCI) provider so OCI cloud integrations remain separate from the Oracle Database provider.

The provider uses these identifiers:

- Display name: Oracle Cloud Infrastructure (OCI)
- Provider ID: `oci`
- Package: `apache-airflow-providers-oci`
- Namespace: `airflow.providers.oci`

## Scope

- Adds the `oci` connection type with API-key, OCI configuration-file, instance-principal, and resource-principal authentication.
- Adds `OciBaseHook` for OCI authentication, native client construction, connection testing, compartment resolution, and client caching.
- Adds `OciGenerativeAIHook`, exposing the native OCI `GenerativeAiClient` for Hosted Application, Hosted Deployment, and work-request operations.
- Adds provider metadata, documentation, connection UI metadata, unit tests, and a read-only system Dag.

## Security and packaging

- Custom service endpoints and worker-local credential-file selection remain explicit hook arguments controlled by Dag authors.
- Regions supplied through Connections must be OCI region identifiers containing only ASCII letters, digits, and hyphens; domain names and URLs are rejected.
- The OCI Python SDK remains an explicit provider extra because it transitively depends on the LGPL-licensed `crc32c` package, which ASF policy classifies as Category X.
- `pip install 'apache-airflow[oci]'` forwards to `apache-airflow-providers-oci[oci]`, so the Airflow extra installs usable OCI support while `apache-airflow[all]` and the base provider avoid the Category X dependency.

## Testing

- 35 OCI provider tests pass, including malicious-region rejection for all four authentication modes and behavior without the optional SDK.
- OCI provider documentation and metadata validation pass.
- Provider dependency metadata and `uv.lock` are synchronized.
- Breeze command images were regenerated.

## Provider adoption

The provider adoption proposal will be sent to the Airflow dev list with the PR as its implementation reference. The required stewards, sponsoring committer, continuous validation plan, and discussion period must be confirmed before merge.

No OCI resources were created, updated, or deleted during the live smoke test.

## Connection UI

This PR introduces a new Connection type, so there is no equivalent before state.

OCI Connection form

---

- [X] Yes — Codex (GPT-5)


AlejandroMorgante:
@SameerMesiah97 Could you please help me with a review? Thanks!

providers/oracle/src/airflow/providers/oracle/hooks/base_oci.py
```diff
@@ -0,0 +1,233 @@
...
+        from flask_appbuilder.fieldwidgets import BS3PasswordFieldWidget, BS3TextFieldWidget
+        from flask_babel import lazy_gettext
+        from wtforms import PasswordField, StringField
+
+        return {
+            "tenancy": StringField(lazy_gettext("Tenancy OCID"), widget=BS3TextFieldWidget()),
+            "fingerprint": StringField(lazy_gettext("Key Fingerprint"), widget=BS3TextFieldWidget()),
+            "key_content": PasswordField(
+                lazy_gettext("Private Key Content"), widget=BS3PasswordFieldWidget()
+            ),
```
shahar1: `key_content` extra holds the raw API signing key but matches no entry in `DEFAULT_SENSITIVE_FIELDS`, so it is not redacted by `GET /api/v2/connections/{id}` or `mask_secret`. format: password only masks the input widget. 
If you rename to `private_key_content` it gets auto-masked.

providers/oracle/src/airflow/providers/oracle/hooks/base_oci.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @cached_property
+    def connection(self) -> Connection:
+        """Return the configured Airflow connection."""
+        return self.get_connection(self.oci_conn_id)
+
+    def get_oci_config(self) -> tuple[dict[str, Any], OciSigner | None]:
+        """Build OCI SDK configuration and an optional signer from the Airflow connection."""
+        oci = _get_oci_sdk()
+        conn = self.connection
```
shahar1: `self.connection` read unconditionally, so `instance_principal`/`resource_principal` still require an `oci_default` row to exist. `AwsBaseHook.conn_config` catches this and carries on.

providers/oracle/src/airflow/providers/oracle/hooks/base_oci.py
```diff
@@ -0,0 +1,233 @@
...
+    :param config_file: OCI SDK configuration file selected by the Dag author.
+    :param profile: Profile to load from the OCI SDK configuration file.
+    :param service_endpoint: Optional service endpoint selected by the Dag author.
+    """
+
+    conn_name_attr = "oci_conn_id"
+    default_conn_name = "oci_default"
+    conn_type = "oci"
+    hook_name = "Oracle Cloud Infrastructure"
+    client_class: Callable[..., OciClient] | None = None
```
shahar1: `client_class` is unreachable by construction (any OCI client class needs the optional SDK at import time), so subclasses must override `_get_client_class()` instead. Two mechanisms, one usable.

providers/oracle/provider.yaml
```diff
@@ -91,6 +92,10 @@ integrations:
       - /docs/apache-airflow-providers-oracle/operators.rst
     logo: /docs/integration-logos/Oracle.png
     tags: [software]
+  - integration-name: Oracle Cloud Infrastructure
```
shahar1: nit: New integration has no `how-to-guide`, so `generative_ai.rst` isn't linked from the integrations index.

shahar1 (changes requested):
Reviewed using AI assistance (Apache Magpie), I went over the comments before posting

providers/oracle/docs/connections/oci.rst
```diff
@@ -0,0 +1,125 @@
...
+
+.. code-block:: bash
+
+    pip install 'apache-airflow-providers-oracle[oci]'
+
+The default connection ID is ``oci_default``.
+
+Service hooks reuse this connection for credentials and region while selecting their own OCI SDK
+client class. Each SDK client derives its endpoint from the configured region unless the Dag author
+passes a ``service_endpoint`` argument to the hook.
```
SameerMesiah97: I would change the first sentence to this as it is a bit more clear:

`Service hooks reuse the connection's credentials and region while instantiating the appropriate OCI SDK client.`

providers/oracle/docs/connections/oci.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Authentication types
+--------------------
+
+API key
+    This is the default. Configure ``User OCID`` in Login, the optional private key passphrase in
+    Password, and ``tenancy``, ``fingerprint``, ``region``, and ``key_content`` in Extra. A Dag
+    author may pass ``key_file`` to the hook instead of storing ``key_content`` in the connection.
```
SameerMesiah97: I think we could specify what fields are required and optional here. The phrasing could be better as well.

providers/oci/docs/generative_ai.rst
```diff
@@ -0,0 +1,143 @@
...
+        )
+    )
+    work_request_id = response.headers.get("opc-work-request-id")
+
+Agent invocation
+----------------
+
+This hook covers the Generative AI management API only. Invoking an active Hosted Application uses
+the Generative AI inference endpoint and a custom application path; it is intentionally outside this
+management hook's contract.
```
SameerMesiah97, line 86: This is more of a general comment about this piece of documentation: it seems like you are going quite deep into the OCI SDK rather than zooming out and focusing on the Airflow interface. My concern is that this effectively becomes a second API reference that needs to stay in sync with Oracle's SDK and REST documentation. If new methods are added or endpoints change, we'll need to update these docs even though the Airflow hook itself hasn't changed.

Maybe it would be bettter to briefly describe the connection/authentication model, explain that `conn`/`get_conn()` expose the native `GenerativeAiClient`, provide a couple of usage examples, and then link to the official OCI SDK documentation for the complete API surface.

providers/oracle/docs/generative_ai.rst
```diff
@@ -0,0 +1,143 @@
...
+                inbound_auth_config_type="IDCS_AUTH_CONFIG",
+                idcs_config=IdcsAuthConfig(
+                    domain_url="https://idcs-example.identity.oraclecloud.com",
+                    scope="agent.invoke",
+                    audience="https://agent.example.com",
+                ),
+            ),
+        )
+    )
+    work_request_id = response.headers.get("opc-work-request-id")
```
SameerMesiah97: I wonder if these examples could be simplified. Since the purpose of the hook is simply to expose the native OCI SDK client, I think the docs only need to communicate the intended usage rather than provide a fully runnable OCI SDK example. I would suggest a pattern like the below for these code blocks:

```
hook = OciGenerativeAIHook(oci_conn_id="oci_default")
response = hook.conn.create_hosted_application(...)
work_request_id = response.headers["opc-work-request-id"]
```

providers/oracle/docs/generative_ai.rst
```diff
@@ -0,0 +1,143 @@
...
+
+The OCI SDK derives the management endpoint as
+``https://generativeai.<region>.oci.oraclecloud.com`` and adds the ``20231130`` API base path.
+The hook exposes these operations without changing OCI retry, pagination, or concurrency-control
+arguments.
+
+==========================================================  ===================================================
+OCI SDK client method                                       REST operation
+==========================================================  ===================================================
+``create_hosted_application``                               ``POST /20231130/hostedApplications``
```
SameerMesiah97: Why is the version number for the API endpoint i.e. 20231130 hardcoded? Won't this change?

Also, I think this table might not be needed here. I go into more detail in a comment below but basically it appears that you are duplicating information that should belong in the OCI SDK documentation.

providers/oci/src/airflow/providers/oci/hooks/base.py
```diff
@@ -0,0 +1,233 @@
...
+    The hook supports API key, OCI configuration file, instance principal, and resource principal
+    authentication. API key credentials are read from the connection fields, while principal
+    authentication is delegated to the OCI SDK.
+
+    :param oci_conn_id: The :ref:`OCI connection id <howto/connection:oci>`.
+    :param auth_type: OCI authentication type selected by the Dag author.
+    :param key_file: API signing private key path selected by the Dag author.
+    :param config_file: OCI SDK configuration file selected by the Dag author.
+    :param profile: Profile to load from the OCI SDK configuration file.
+    :param service_endpoint: Optional service endpoint selected by the Dag author.
```
SameerMesiah97, line 81: I would add the default values for each parameter to this docstring.

providers/oracle/src/airflow/providers/oracle/hooks/base_oci.py
```diff
@@ -0,0 +1,233 @@
...
+    from airflow.sdk import Connection
+
+    OciSigner = Any
+
+OciClient = TypeVar("OciClient")
+
+OCI_AUTH_TYPE_API_KEY = "api_key"
+OCI_AUTH_TYPE_CONFIG_FILE = "config_file"
+OCI_AUTH_TYPE_INSTANCE_PRINCIPAL = "instance_principal"
+OCI_AUTH_TYPE_RESOURCE_PRINCIPAL = "resource_principal"
```
SameerMesiah97: This looks like a textbook scenario for an enum class. Please see the below:

```
class OciAuthType(str, Enum):
    API_KEY = "api_key"
    CONFIG_FILE = "config_file"
    INSTANCE_PRINCIPAL = "instance_principal"
    RESOURCE_PRINCIPAL = "resource_principal"
```

providers/oci/tests/unit/oci/hooks/test_base.py
```diff
@@ -0,0 +1,403 @@
...
+        self.set_connection(Connection(extra={"compartment_id": "connection-compartment"}))
+
+        assert self.hook.get_compartment_id() == "connection-compartment"
+
+    def test_get_compartment_id_requires_value(self):
+        self.set_connection(Connection())
+
+        with pytest.raises(ValueError, match="An OCI compartment OCID must be provided"):
+            self.hook.get_compartment_id()
+
```
SameerMesiah97, line 371: Minor nit: I think a few of these tests (for example `get_compartment_id` and possibly the `get_client `cases) could be parameterized to reduce a bit of duplication. Not a blocker though.

providers/oracle/tests/unit/oracle/hooks/test_base_oci.py
```diff
@@ -0,0 +1,403 @@
...
+            """
+import sys
+
+sys.modules["oci"] = None
+import airflow.providers.oracle.hooks.base_oci
+import airflow.providers.oracle.hooks.generative_ai
+""",
+        ],
+        check=True,
+    )
```
SameerMesiah97: I would import the below:

`from textwrap import dedent`

and then do this:

```
subprocess.run(
    [
        sys.executable,
        "-c",
        dedent(
            """
            import sys

            sys.modules["oci"] = None
            import airflow.providers.oracle.hooks.base_oci
            import airflow.providers.oracle.hooks.generative_ai
            """
        ),
    ],
    check=True,
)
```

SameerMesiah97 (commented):
Looks very clean overall. I have left some comments.

SameerMesiah97 (approved):
Looks good.

providers/oracle/pyproject.toml
```diff
@@ -60,14 +60,17 @@ requires-python = ">=3.10"
 # After you modify the dependencies, and rebuild your Breeze CI image with ``breeze ci-image build``
 dependencies = [
     "apache-airflow>=2.11.0",
-    "apache-airflow-providers-common-compat>=1.8.0",
+    "apache-airflow-providers-common-compat>=1.8.0",  # use next version
```
shahar1: `airflow.providers.common.compat.sdk` only exists since common-compat **1.10.0**, and only exports `BaseHook` / `AirflowNotFoundException` / `AirflowOptionalProviderFeatureException` since **1.12.0** — with compat 1.8–1.11 installed, importing the new hooks raises `ImportError` while pip considers the environment consistent. The `# use next version` marker is for a bound that isn't released yet (cf. `amazon`/`fab`/`keycloak`); 1.12.0 already shipped, so:

```suggestion
    "apache-airflow-providers-common-compat>=1.12.0",
```

providers/oracle/src/airflow/providers/oracle/hooks/base_oci.py
```diff
@@ -0,0 +1,263 @@
...
+            config, signer = self.get_oci_config()
+            client_kwargs = {"signer": signer} if signer is not None else {}
+            oci.identity.IdentityClient(config=config, **client_kwargs).list_regions()
+        except Exception as e:
+            return False, f"{type(e).__name__} error occurred while testing connection: {e}"
+        return True, "Connection successfully tested"
+
+    def get_compartment_id(self, compartment_id: str | None = None) -> str:
+        """Return an explicit compartment OCID or the default from the connection."""
+        resolved_compartment_id = compartment_id or self.connection.extra_dejson.get("compartment_id")
```
shahar1: With `oci_conn_id=None` (instance/resource principal), `self.connection` raises the misleading `ValueError: An OCI connection ID is required for API key authentication.`, and with a defined-but-missing connection this leaks `AirflowNotFoundException` instead of the compartment error below. The docs tell principal-auth users to call this method, so:

```suggestion
        resolved_compartment_id = compartment_id or self._get_optional_connection_extras().get("compartment_id")
```

Worth a small test for the `oci_conn_id=None` case too.

providers/oci/tests/unit/oci/hooks/test_base.py
```diff
@@ -0,0 +1,425 @@
...
+    AirflowOptionalProviderFeatureException,
+)
+from airflow.providers.oracle.get_provider_info import get_provider_info
+from airflow.providers.oracle.hooks.base_oci import (
+    OciAuthType,
+    OciBaseHook,
+    _get_oci_sdk,
+)
+
+GenerativeAiClient = pytest.importorskip("oci.generative_ai").GenerativeAiClient
```
shahar1, line 31: Nit: this module-level skip also skips `test_get_oci_sdk_requires_optional_extra` and `test_hook_modules_import_without_optional_oci_sdk` — the two tests whose purpose is missing-SDK behaviour — in environments without `oci` installed. Consider moving those two to a module without the skip. Not blocking.

shahar1 (changes requested):
Thanks for the thorough follow-up — all points from my previous review are addressed, and the security-boundary design (connection extras can't steer `key_file`/`config_file`/`service_endpoint`) plus the placeholder-sync test are genuinely nice. One packaging issue is blocking, one smaller bug inline.

### Blocking — `common-compat` minimum bound too low (`pyproject.toml:63`)

The hooks import `BaseHook`, `AirflowNotFoundException`, and `AirflowOptionalProviderFeatureException` from `airflow.providers.common.compat.sdk`. That module only exists since `apache-airflow-providers-common-compat` **1.10.0**, and exports these three symbols only since **1.12.0** — but the bound here stays `>=1.8.0`. On an environment with compat 1.8–1.11, `import airflow.providers.oracle.hooks.base_oci` fails with `ImportError` even though pip considers the environment consistent. Please bump to `>=1.12.0`. The `# use next version` marker is used when the required version isn't released yet — 1.12.0 (and 1.17.0) already are, so the marker should be dropped.

### `get_compartment_id()` breaks for principal auth without a connection (`base_oci.py:235`)

`self.connection.extra_dejson` is read directly, so with `oci_conn_id=None` (instance/resource principal) this raises the misleading `ValueError: An OCI connection ID is required for API key authentication.`, and with a missing connection row it leaks `AirflowNotFoundException` instead of the intended compartment error. The docs tell principal-auth users to call this method. Reusing `_get_optional_connection_extras()` fixes both paths — plus a small test for the no-connection case.

### Smaller observations

- `tests/.../test_base_oci.py:38` — the module-level `pytest.importorskip("oci.generative_ai")` also skips the two tests whose purpose is missing-SDK behaviour (`test_get_oci_sdk_requires_optional_extra`, `test_hook_modules_import_without_optional_oci_sdk`) when `oci` isn't installed. Consider hosting those two in a module without the skip. Not blocking.

---

>

---
Drafted-by: Claude Code (Fable 5); reviewed by @shahar1 before posting

providers/oracle/pyproject.toml
AlejandroMorgante: Airflow's provider dependency convention requires keeping the current lower bound with # use next version; cross-provider >= bumps are performed by Release Managers and are rejected by the selective check in regular PRs

shahar1: I've set a PR label that will allow bumping provider's dependency.
You could set now to `>=12.0` safely.

shahar1 (approved):
Good to go once common-compat version is settled

vikramkoka:
I am fully supportive of this work. 

However, don't you think it is better to put this into a new OCI (Oracle Cloud Infra) provider? 
Rather than putting it into the Oracle DB provider? 

cc: @eladkal @kaxil @potiuk @ashb

AlejandroMorgante:
Good point — I agree that OCI should be a separate provider, similar to microsoft.azure and microsoft.mssql. I suggest using oci as the provider ID. The alternatives oracle.cloud and oracle.oci would require restructuring the existing oracle provider, since oracle is already a provider rather than an umbrella namespace. We can still use “Oracle Cloud Infrastructure” as the public-facing name for clarity.

- Display name: **Oracle Cloud Infrastructure (OCI)**
- Provider ID: `oci`
- Package: **apache-airflow-providers-oci**
- Namespace: **airflow.providers.oci**

eladkal:
I think I am leaning towards 
apache-airflow-providers-oracle-cloud (OCI can sit within the cloud provider)
apache-airflow-providers-oracle-db (To have what we currently have in apache-airflow-providers-oracle)

and possible discontinue apache-airflow-providers-oracle or handle it as an empty shell that install all oracle related packages?


Migrating a provider name is something we did in the past with `apache-airflow-providers-jira` -> `apache-airflow-providers-atlassian-jira` (We did it with steps to avoid breaking changes)

vikramkoka:
> I think I am leaning towards apache-airflow-providers-oracle-cloud (OCI can sit within the cloud provider) apache-airflow-providers-oracle-db (To have what we currently have in apache-airflow-providers-oracle)
> 
> and possible discontinue apache-airflow-providers-oracle or handle it as an empty shell that install all oracle related packages?
> 
> Migrating a provider name is something we did in the past with `apache-airflow-providers-jira` -> `apache-airflow-providers-atlassian-jira` (We did it with steps to avoid breaking changes)

no issues with naming. 
But, I still think the oracle-cloud provider should have a quick dev list email and go into "incubation"

kaxil:
Agreed with Elad on `apache-airflow-providers-oracle-cloud` -- OCI can easily be confused with https://opencontainers.org/

eladkal:
> But, I still think the oracle-cloud provider should have a quick dev list email and go into "incubation"

It should follow the procedure of adding a new provider yes. I think we just need to figure out the mechanism here and how we phase out the oracle provider infavor of oracle-db and oracle-cloud

eladkal (changes requested):
Temp block till we alligned on the setup of the oracle folder structure

eladkal:
cc @kaxil @vikramkoka @AlejandroMorgante I started the first step of the migration https://github.com/apache/airflow/pull/71325

I think it would be better to solve `oracle -> oracle-oracledb` migration and only after all steps are finished introduce `oracle-oci` The plan is explained in https://github.com/apache/airflow/pull/71325

potiuk:
Agreeing with where this landed: `apache-airflow-providers-oracle-cloud` rather than `oci`, and after #71325 rather than before it.

The sequencing matters more than it may look. #71325 turns `oracle` into an umbrella namespace (`providers/oracle/oracledb/`), which is what makes room for this provider to live at `providers/oracle/cloud/` as `airflow.providers.oracle.cloud`. Landing it now as a top-level `providers/oci/` would mean moving every module, the package name, the connection type and the docs a second time, and users would be asked to migrate twice.

Concretely, what this PR needs before it can be reviewed for merge:

- wait for #71325 to land, then re-target to `providers/oracle/cloud/` with the package name `apache-airflow-providers-oracle-cloud` and module path `airflow.providers.oracle.cloud`;
- follow the new-provider procedure, including the announcement on the dev list — a new provider is a project-level commitment, not only a code review;
- rebase to resolve the current `uv.lock` conflict from the flit-4 change on main.

None of this is a comment on the code. The hook layer reads well — the auth types are enumerated, the OCI SDK import is guarded behind `AirflowOptionalProviderFeatureException`, and the connection form is properly specified. It is worth converting the PR to draft while it waits, so it drops out of the maintainer review queue and nobody re-reviews it in the interim.

---

Issue #70117, opened 2026-07-20
Task supervisor lingers indefinitely on CeleryExecutor when psutil.Process.wait() returns None (socket_cleanup_timeout never triggers)


PrafullPandey06:
### Under which category would you file this issue?

Task SDK

### Apache Airflow version

3.1.8

### What happened and how to reproduce it?

Task supervisor processes (`airflow worker -- <uuid>`) remain alive indefinitely after their task subprocess has finished and exited. The `workers.socket_cleanup_timeout` safety-net (introduced in 3.0.2 via https://github.com/apache/airflow/pull/51180) never triggers because the supervisor never detects the child process exit — `_process_exit_monotonic` is never set.

On our production Celery workers running for ~3 days:
- **worker-0:** 26 stuck supervisors, 357 zombie processes, 0 running tasks
- **worker-1:** 30 stuck supervisors, 417 zombie processes, 0 running tasks

Each stuck supervisor holds ~340 MB of RSS. With 26–30 orphaned supervisors, this accounts for ~8–10 GB of wasted memory per worker pod, causing OOM kills on containers with 6–8 Gi limits.

## How to reproduce

1. Deploy Airflow 3.1.8 with CeleryExecutor using `dumb-init` as container entrypoint (standard Helm chart setup):
   ```
   command: ["/usr/bin/dumb-init", "--"]
   args: ["bash", "-c", "exec airflow celery worker -H ${HOSTNAME}"]
   ```

2. Configure `worker_concurrency=16` (or any value)

3. Run a DAG with dynamically mapped tasks (using `expand()`), e.g. a task that produces many map instances

4. Wait for tasks to complete

5. Check for lingering supervisors:
   ```bash
   # Inside the worker container:
   count=0
   for pid in /proc/[0-9]*; do
     cmd=$(tr '\0' ' ' < "$pid/cmdline" 2>/dev/null)
     case "$cmd" in "airflow worker -- "*) count=$((count+1));; esac
   done
   echo "live supervisors: $count"
   ```

6. Expected: 0 supervisors when no tasks are running

7. Actual: supervisors accumulate over time, never exit

### Environment
- **Airflow:** 3.1.8
- **Executor:** CeleryExecutor with `worker_concurrency=16`
- **Container init:** `dumb-init` (PID 1) — standard in Airflow Helm chart
- **Worker uptime:** ~2 days 18 hours
- **DAG triggering the issue:** `generate_compliance_report` with dynamically mapped task (`map_index=152`)

[...]

PR #70163, opened 2026-07-21
Fix task supervisor hang by resetting SIGCHLD in CeleryExecutor

hooiv:
## Description
This PR resolves the memory leak and infinite hang observed in Airflow 3 Task SDK supervisors running on Celery workers. 
### Root Cause
When Airflow 3 runs a task via the `CeleryExecutor`, the `execute_workload` task is executed inside a Celery/billiard worker process. 
Because `billiard` workers are spawned via `fork()`, they inherit the global `SIGCHLD` signal handler installed by the Celery main pool manager. This inherited handler uses a non-blocking `waitpid(-1, WNOHANG)` loop.
When the Airflow supervisor (`WatchedSubprocess`) forks a child process to execute the actual task SDK payload, that child is a direct descendant of the Celery worker. When the task finishes, the child sends `SIGCHLD` to the worker, triggering the inherited `billiard` signal handler. The handler reaps the child process and discards the exit status.
When the Airflow supervisor subsequently attempts to retrieve the exit status via `psutil.Process.wait(timeout=0)`, it receives an `ECHILD` (No child processes) error. `psutil` catches this and returns `None`. The supervisor's polling loop did not expect a `None` return value, leaving `_exit_code` unset, which prevented the `socket_cleanup_timeout` from ever triggering. The supervisor hung indefinitely in `epoll_wait()`.
### The Fix
This PR introduces  fix:
1. **(Celery Executor):** Added a `SIGCHLD` reset in `celery_executor_utils.py::execute_workload`. By resetting `SIGCHLD` to `SIG_DFL` before invoking `supervise()`, we prevent the inherited billiard handler from stealing the exit status of the Task SDK subprocess.
2. **(Task SDK Supervisor):** Updated `_check_subprocess_exit` in `supervisor.py` to handle cases where `psutil.wait(timeout=0)` returns `None`. It now logs a warning, gracefully assigns an exit code of `-1`, and returns immediately, preventing the supervisor from hanging if an exit status is ever consumed externally (e.g. reparenting to `dumb-init`).
### Testing
- Added a dedicated unit test in `test_supervisor.py` to ensure `None` is mapped to `-1` and the loop exits cleanly.

### DIFF
--- a/providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py
+++ b/providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py
@@ -239,6 +239,13 @@ def execute_workload(input: str) -> None:
 
     log.info("[%s] Executing workload in Celery: %s", celery_task_id, workload)
 
+    # Reset SIGCHLD to default to prevent billiard's inherited handler from stealing
+    # the exit status of child processes spawned by the Task SDK supervisor.
+    import signal
+
+    if hasattr(signal, "SIGCHLD"):
+        signal.signal(signal.SIGCHLD, signal.SIG_DFL)
+
     try:
         BaseExecutor.run_workload(workload)
     except Exception as e:
--- a/task-sdk/src/airflow/sdk/execution_time/supervisor.py
+++ b/task-sdk/src/airflow/sdk/execution_time/supervisor.py
@@ -1141,7 +1141,14 @@ class WatchedSubprocess:
             if raise_on_timeout:
                 raise
         else:
-            self._process_exit_monotonic = time.monotonic()
+            if self._exit_code is None:
+                self.process_log.warning(
+                    "Process has already exited but its exit status was consumed by an external process. "
+                    "Setting exit code to -1."
+                )
+                self._exit_code = -1
+                self._process_exit_monotonic = time.monotonic()
+                return self._exit_code
 
             if expect_signal is not None and self._exit_code == -expect_signal:
                 # Bypass logging, the caller expected us to exit with this
--- a/task-sdk/tests/task_sdk/execution_time/test_supervisor.py
+++ b/task-sdk/tests/task_sdk/execution_time/test_supervisor.py
@@ -4465,3 +4465,31 @@ class TestMakeBufferedSocketReader:
         finally:
             r.close()
             w.close()
+
+
+class TestCheckSubprocessExit:
+    @mock.patch("airflow.sdk.execution_time.supervisor.time.monotonic", return_value=123.456)
+    def test_subprocess_exit_status_consumed_by_init(self, mock_time):
+        """
+        Test that if wait() returns None (e.g., child process was already reaped by an init process like dumb-init),
+        the exit code is set to -1 to prevent an infinite loop, and _process_exit_monotonic is recorded.
+        """
+        from airflow.sdk.execution_time.supervisor import WatchedSubprocess
+        from unittest.mock import MagicMock
+        import uuid
+        
+        watched_subprocess = WatchedSubprocess(
+            process_log=MagicMock(),
+            id=uuid.uuid4(),
+            pid=12345,
+            stdin=MagicMock(),
+            client=MagicMock(),
+            process=MagicMock(),
+        )
+        watched_subprocess._process.wait.return_value = None
+
+        exit_code = watched_subprocess._check_subprocess_exit()
+
+        assert exit_code == -1
+        assert watched_subprocess._exit_code == -1
+        assert watched_subprocess._process_exit_monotonic == 123.456


hooiv:
Updated the patch and description to match the confirmed failure path in #70117. psutil.Process.wait(timeout=0) can return None after the child wait status has already been consumed; _exit_code then stayed unset, so the supervisor loop never entered socket cleanup. The regression now covers that exact None result and maps it to the existing unknown exit code (-1). The earlier dumb-init/NoSuchProcess explanation was removed because it was not confirmed. Focused Linux test and Ruff checks pass.

hooiv:
I've pushed a completely new approach that fixes the actual root cause! 

In #70117, @Andrushika correctly pointed out that `dumb-init` couldn't be reaping the process while the supervisor was alive, but proved *something* was stealing the exit status.

It turns out it was **Celery's `billiard` library**. Because the supervisor runs inside the Celery worker process, it inherits `billiard`'s aggressive global `SIGCHLD` handler. When the Airflow task subprocess exits, the `billiard` handler intercepts the signal and calls `waitpid(-1, WNOHANG)`, silently eating the exit status before our `psutil.wait()` can see it! 

I've force-pushed a fix that simply resets `SIGCHLD` to `SIG_DFL` inside `celery_executor_utils.py` right before the supervisor takes over. This prevents `billiard` from interfering and perfectly resolves the memory leak. 

The PR description has been updated .

Andrushika:
Thanks for figuring it out! Would it be possible to give a minimal reproducible example for how "Celery's billiard library" triggers this bug? Please also include the necessary tests in your fix.
BTW, you can turn this PR into "open" when it's ready for review. It's draft currently.

hooiv:
Thanks for the review! I have pushed a commit that adds a unit test for this fix in `tests/providers/celery/executors/test_celery_executor.py`.

Regarding how `billiard` triggers this bug, Celery's multiprocessing fork (`billiard`) installs a global `SIGCHLD` handler upon initialization. This handler loops and calls `waitpid(-1, os.WNOHANG)` to aggressively reap *any* child process that exits, which intercepts the exit statuses of child processes spawned by the Airflow Task SDK.

Here is a minimal reproducible example demonstrating the issue:

`python
import os
import time
import psutil
from billiard.pool import Pool

if __name__ == "__main__":
    # 1. Starting a billiard pool installs a global SIGCHLD handler 
    # that aggressively calls waitpid(-1, WNOHANG) to reap ANY exiting children
    pool = Pool(1)
    
    # 2. Airflow Task SDK forks a child process to run the task
    pid = os.fork()
    if pid == 0:
        # Task subprocess does its work and exits
        os._exit(42)
        
    proc = psutil.Process(pid)
    
    # Give the child time to exit and billiard's signal handler time to wake up and reap it
    time.sleep(0.2)
    
    # 3. Airflow Task SDK tries to check the exit status, but wait() returns None 
    # because billiard already stole the exit status!
    print(f"Child PID {pid}")
    print(f"psutil.wait(timeout=0) returned: {proc.wait(timeout=0)}")
    
    pool.close()
    pool.join()
`

If you run this script, `proc.wait(timeout=0)` will return `None` (meaning the status has been consumed) instead of the actual exit code `42`. By resetting `SIGCHLD` to `SIG_DFL` before executing the workload, we prevent `billiard` from stealing the exit status of the Airflow tasks.

hooiv:
did a much deeper dive into the Celery/billiard source code and the Airflow 3 SDK architecture, and I have the complete and accurate picture. The reproducer I provided earlier was flawed because I explicitly injected a global handler into the script to force the failure, incorrectly assuming that's what `billiard` did under the hood.
Here is the *actual* sequence of events causing the memory leak/hang:
1. The Celery main process installs a global `SIGCHLD` handler to manage its worker pool using a `waitpid(-1, WNOHANG)` loop.
2. When Celery forks a worker process (via `billiard`), that worker **inherits** the `SIGCHLD` handler.
3. In Airflow 3, the Celery worker receives the `execute_workload` task and directly calls `supervise()`.
4. `supervise()` spawns the Task SDK subprocess via `WatchedSubprocess.start()` (using `os.fork()`). 
5. When the Task SDK subprocess finishes, it sends `SIGCHLD` to its parent (the Celery worker).
6. **The Race Condition:** The Celery worker's inherited `billiard` handler intercepts the signal, calls `waitpid(-1)`, and consumes the exit status before the supervisor's `psutil.wait(0)` loop gets a chance to see it!
7. The supervisor loop calls `psutil.wait(0)`, which receives `ECHILD` and gracefully returns `None`. Because our supervisor code didn't handle `None`, it hung infinitely in `epoll_wait()`.
I have force-pushed  implements a fix:
1. We now explicitly reset `SIGCHLD` to `SIG_DFL` inside `execute_workload()` in `celery_executor_utils.py` before `supervise()` is called. This safely uninstalls the inherited billiard handler for the duration of the workload execution. (Note: My previous commit put this reset in `_execute_in_fork`, which is only used by Airflow 2, completely missing the Airflow 3 execution path).
2.  I modified `_check_subprocess_exit` in `supervisor.py` to handle `psutil.wait()` returning `None` by setting the exit code to `-1` and returning immediately. This is a critical safety net that ensures the supervisor will never hang if an exit status is lost (e.g. if the supervisor crashes and the child actually does reparent to `dumb-init`).
I have also added a dedicated unit test in `test_supervisor.py` to cover the `None` scenario.

PR #70167, opened 2026-07-21
Align Dag Runs docs with Airflow 3 trigger timetable defaults

Tegh25:
The Data Interval and Catchup sections in `dag-run.rst` still described `@daily` and catchup as if `CronDataIntervalTimetable` were the default. That matches Airflow 2 / `create_cron_data_intervals=True`, not Airflow 3, where bare cron strings resolve to `CronTriggerTimetable` by default.

Update the Data Interval section to explain that interval semantics depend on the timetable: trigger timetables use zero-width intervals at the trigger time; data-interval timetables use contiguous windows and schedule after the
window ends. Link to the timetable comparison docs for `logical_date` and `run_id` differences.

Update the Catchup section and tutorial example to describe behavior under the Airflow 3 default (`CronTriggerTimetable`, `catchup=False`), and keep the prior data-interval story as the explicit opt-in case (`CronDataIntervalTimetable` or `create_cron_data_intervals=True`). Use timetable-neutral wording ("scheduled run times") in the catchup intro so it
applies to both timetable kinds.

No newsfragment: documentation-only clarification aligned with existing `authoring-and-scheduling/timetable.rst` content.

---

- [X] Yes — Cursor (Auto)

Generated-by: Cursor (Auto) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/docs/core-concepts/dag-run.rst
+++ b/airflow-core/docs/core-concepts/dag-run.rst
@@ -71,15 +71,16 @@ side-by-side comparison, including how ``logical_date`` and ``run_id`` differ.
 
 All dates in Airflow are tied to the data interval concept in some way. The
 "logical date" (also called ``execution_date`` in Airflow versions prior to 2.2)
-of a Dag run is defined by the timetable: for a trigger timetable it is the
-trigger time; for a data-interval timetable it is the start of the data
-interval, not when the Dag is actually executed.
+of a Dag run is defined by the timetable: for both timetable kinds it is
+``data_interval_start``. For the default (zero-width) trigger timetable that
+equals the trigger time. With a non-zero ``interval=`` on a trigger timetable
+it is ``trigger_time - interval``. For a data-interval timetable it is the
+start of the contiguous window, not when the Dag is actually executed.
 
 Similarly, since the ``start_date`` argument for the Dag and its tasks points to
 the same logical date, it marks the start of scheduling for the Dag (the first
 possible logical date), not when tasks in the Dag will start running. In other
-words, a Dag run will only be scheduled once the timetable advances past
-``start_date``.
+words, a Dag run will only be scheduled once the timetable reaches ``start_date``.
 
 .. tip::
 
--- a/airflow-core/docs/core-concepts/dag-run.rst
+++ b/airflow-core/docs/core-concepts/dag-run.rst
@@ -50,24 +50,36 @@ Data Interval
 -------------
 
 Each Dag run in Airflow has an assigned "data interval" that represents the time
-range it operates in. For a Dag scheduled with ``@daily``, for example, each of
-its data interval would start each day at midnight (00:00) and end at midnight
-(24:00).
+range it operates in. How that interval is defined depends on the Dag's
+timetable.
 
-A Dag run is usually scheduled *after* its associated data interval has ended,
-to ensure the run is able to collect all the data within the time period. In
-other words, a run covering the data period of 2020-01-01 generally does not
-start to run until 2020-01-01 has ended, i.e. after 2020-01-02 00:00:00.
+In Airflow 3, a Dag scheduled with a bare cron string such as ``@daily`` uses
+:ref:`CronTriggerTimetable` by default (``[scheduler] create_cron_data_intervals``
+is ``False``). For that timetable, ``data_interval_start`` and
+``data_interval_end`` are the same — the trigger time (for ``@daily``, midnight
+each day). The run is created to execute *at* that time.
+
+If you need a contiguous full-day window instead — for example each run covering
+from midnight to the next midnight — use a data-interval timetable such as
+:ref:`CronDataIntervalTimetable`, or set ``[scheduler] create_cron_data_intervals``
+to ``True``. With that setup, a Dag run is usually scheduled *after* its
+associated data interval has ended, so a run covering 2020-01-01 generally does
+not start until after 2020-01-02 00:00:00.
+
+See :ref:`Differences between "trigger" and "data interval" timetables` for a
+side-by-side comparison, including how ``logical_date`` and ``run_id`` differ.
 
 All dates in Airflow are tied to the data interval concept in some way. The
 "logical date" (also called ``execution_date`` in Airflow versions prior to 2.2)
-of a Dag run, for example, denotes the start of the data interval, not when the
-Dag is actually executed.
+of a Dag run is defined by the timetable: for a trigger timetable it is the
+trigger time; for a data-interval timetable it is the start of the data
+interval, not when the Dag is actually executed.
 
 Similarly, since the ``start_date`` argument for the Dag and its tasks points to
-the same logical date, it marks the start of *the Dag's first data interval*, not
-when tasks in the Dag will start running. In other words, a Dag run will only be
-scheduled one interval after ``start_date``.
+the same logical date, it marks the start of scheduling for the Dag (the first
+possible logical date), not when tasks in the Dag will start running. In other
+words, a Dag run will only be scheduled once the timetable advances past
+``start_date``.
 
 .. tip::
 
@@ -104,10 +116,10 @@ Dag run fails.
 Catchup
 -------
 
-An Airflow Dag defined with a ``start_date``, possibly an ``end_date``, and a non-asset schedule, defines a series of intervals which the scheduler turns into individual Dag runs and executes.
-By default, Dag runs that have not been run since the last data interval are not created by the scheduler upon activation of a Dag ( Airflow config ``scheduler.catchup_by_default=False``). The scheduler creates a Dag run only for the latest interval.
+An Airflow Dag defined with a ``start_date``, possibly an ``end_date``, and a non-asset schedule, defines a series of scheduled run times which the scheduler turns into individual Dag runs and executes.
+By default, missed scheduled run times are not created by the scheduler upon activation of a Dag (Airflow config ``scheduler.catchup_by_default=False``). The scheduler creates a Dag run only for the next (or most recently applicable) scheduled run time according to the timetable.
 
-If you set ``catchup=True`` in the Dag, the scheduler will kick off a Dag Run for any data interval that has not been run since the last data interval (or has been cleared). This concept is called Catchup.
+If you set ``catchup=True`` in the Dag, the scheduler will kick off a Dag Run for any scheduled run time that has not been run since the last run (or has been cleared). This concept is called Catchup.
 
 If your Dag is not written to handle its catchup (i.e., not limited to the interval, but instead to ``Now`` for instance.),
 then you will want to turn catchup off, which is the default setting or can be done explicitly by setting ``catchup=False`` in the Dag definition, if the default config has been changed for your Airflow environment.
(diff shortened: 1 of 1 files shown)


Tegh25:
Tagging @ferruzzi for visibility

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -138,20 +150,30 @@ then you will want to turn catchup off, which is the default setting or can be d
...
-
-If the ``dag.catchup`` value had been ``True`` instead, the scheduler would have created a Dag Run
-for each completed interval between 2015-12-01 and 2016-01-02 (but not yet one for 2016-01-02,
-as that interval hasn't completed) and the scheduler will execute them sequentially.
+2016-01-02 at 6 AM (or from the command line), with the Airflow 3 default of
+:ref:`CronTriggerTimetable` for ``@daily`` and ``catchup=False``, the scheduler
+does **not** create a run for the past midnight. The next Dag run is created at
+midnight on the morning of 2016-01-03, with ``data_interval_start`` and
+``data_interval_end`` both equal to that trigger time.
+
```
ferruzzi, line 165: Are you sure about this one?  Based on my understanding and what is already said above, I would expect the previous midnight to get run.

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -50,24 +50,36 @@ Data Interval
...
+See :ref:`Differences between "trigger" and "data interval" timetables` for a
+side-by-side comparison, including how ``logical_date`` and ``run_id`` differ.
 
 All dates in Airflow are tied to the data interval concept in some way. The
 "logical date" (also called ``execution_date`` in Airflow versions prior to 2.2)
-of a Dag run, for example, denotes the start of the data interval, not when the
-Dag is actually executed.
+of a Dag run is defined by the timetable: for a trigger timetable it is the
+trigger time; for a data-interval timetable it is the start of the data
```
ferruzzi: That's only true for zero-width intervals, otherwise it's `trigger_time-interval`, right?

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -50,24 +50,36 @@ Data Interval
...
+interval, not when the Dag is actually executed.
 
 Similarly, since the ``start_date`` argument for the Dag and its tasks points to
-the same logical date, it marks the start of *the Dag's first data interval*, not
-when tasks in the Dag will start running. In other words, a Dag run will only be
-scheduled one interval after ``start_date``.
+the same logical date, it marks the start of scheduling for the Dag (the first
+possible logical date), not when tasks in the Dag will start running. In other
+words, a Dag run will only be scheduled once the timetable advances past
```
ferruzzi: Nipick, it looks like it's a >=, so "reaches" might be technically more accurate than "advances past"?

airflow-core/docs/core-concepts/dag-run.rst
Tegh25: When writing this section, I was referencing this example on [the timetable doc](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timetable.html#the-time-when-a-dag-run-is-triggered).
> For example, suppose there is a cron expression @daily or 0 0 * * *, which is scheduled to run at 12AM every day. If you enable Dags using the two timetables at 3PM on January 31st, [CronTriggerTimetable](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timetable.html#crontriggertimetable) creates a new Dag run at 12AM on February 1st. The run_id timestamp is midnight, on February 1st.

In this case after enablement, the next Dag run is created on the following midnight, not the previous one.

airflow-core/docs/core-concepts/dag-run.rst
Tegh25: I see what you mean. You're referring to CronTriggerTimetable or DeltaTriggerTimetable with a non-zero  `interval=` argument, right?

You're right: ``DagRunInfo.logical_date`` is always ``data_interval.start``. For the default zero-width case (``interval=timedelta()``), that equals the trigger time. With a non-zero ``interval``, the run still fires at the cron tick ``T``, but the data interval is ``[T - interval, T]``, so ``logical_date`` is ``T - interval`` (i.e. ``data_interval_start``), not ``T``.

I can adjust that sentence so it doesn't exclude the non-default case.
e.g. that for a trigger timetable ``logical_date`` is ``data_interval_start``, which equals the trigger time when the interval is zero-width (the default).

airflow-core/docs/core-concepts/dag-run.rst
Tegh25: Agreed. I can make the change to "reaches".

airflow-core/docs/core-concepts/dag-run.rst
ferruzzi: Something isn't adding up.  If you check out the unit test `test_daily_cron_trigger_no_catchup_first_starts_at_next_schedule` (specifically the `first-run` case) [here](https://github.com/apache/airflow/blob/063d5f108dad0ff0167814e65899aa7b95bd8f12/airflow-core/tests/unit/timetables/test_trigger_timetable.py#L78) it seems to be testing this and asserting the opposite.  Any chance you are using out of date docs to update out of date docs?  LOL  Have a look at the test, maybe I'm just mis-reading it.

Tegh25: It seems like you're right and those lines are wrong for the example as written.

I had misread the first-run path. Because the tutorial Dag sets ``start_date``, ``TimeRestriction.earliest`` is also set. Therefore, ``catchup=False`` uses ``max(_align_to_prev(now), _align_to_next(earliest))`` and does **not** call ``_calc_first_run()`` (that's only when ``earliest is None``). This is indicated by the [catchup logic here](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/timetables/trigger.py#L97-L106).

In the example: Picked up at 2016-01-02 06:00 with ``@daily``, that means the first run is **2016-01-02 00:00** (the most recent past tick), not 2016-01-03. Same idea as ``test_daily_cron_trigger_no_catchup_first_starts_at_next_schedule`` with ``id="first-run"``, as you pointed out.

I'll fix the example explanation to say the scheduler creates a run for that morning's midnight (zero-width interval), and only later ticks wait for their schedule time. Thanks for the catch.

Although this logic in the code makes sense, from the perspective of a user of Airflow, this behavior for ``catchup=False`` seems unintuitive to me. I'd welcome anyone else to chime in on this conversation.

ferruzzi: Yeah, I've often found that the `catchup` behavior doesn't quite work the way I expect it to, but I usually figure that just because _I_ expect it to work a certain way doesn't mean that's what others expect.  I'm not really sure on this one.  I think the intent is that it is triggering the most recent run so a newly enabled Dag will always do _something_ right away.  Which also maybe feels like a reasonable expectation.  Maybe the behavior should be changed, but even if it does get changed, that's solidly out of scope of this doc update PR.  

Either way, nice deep-dive on the code.  Thanks for checking it.  It sounds like the `timetable.rst` doc also needs updating then, yeah?  There may be more going on in that doc that needs fixing as well.  Do you think you'll have time to update that as well (in a new PR), or shall we cut a new Issue so someone else can take it?

ferruzzi (approved):
Approved pending green CI

airflow-core/docs/core-concepts/dag-run.rst
Tegh25: Yes, I'll take another look at `timetable.rst` and update the doc according to what we determined here, in a new PR.

Regarding time: The MLH program ends next week, but I'm happy to continue contributing beyond the end of the program.

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -138,20 +151,33 @@ then you will want to turn catchup off, which is the default setting or can be d
...
+only created once that schedule time is reached.
+
+If instead the Dag used a data-interval timetable (for example
+:ref:`CronDataIntervalTimetable`, or ``[scheduler] create_cron_data_intervals=True``),
+the scheduler would immediately create a run for the most recently completed
+interval (2016-01-01 through 2016-01-02), and the next run would cover
+2016-01-02 through 2016-01-03 after that interval ends.
+
+Be aware that using a ``datetime.timedelta`` object as ``schedule`` follows the
+same default-vs-data-interval split via ``[scheduler] create_delta_data_intervals``.
```
kaxil: Nothing reads `[scheduler] create_delta_data_intervals`. The timedelta branch of `_create_timetable` gates on `create_cron_data_intervals` too ([dag.py#L149-L152](https://github.com/apache/airflow/blob/20b9a85c557ce0b10e22401a560bf50d856c97a5/task-sdk/src/airflow/sdk/definitions/dag.py#L149-L152)), so the delta option is inert and this line would send readers to a switch that does nothing.

Also worth keeping the concrete detail the old paragraph had. `DeltaMixin._align_to_prev` returns `current` unchanged ([_delta.py#L70-L71](https://github.com/apache/airflow/blob/20b9a85c557ce0b10e22401a560bf50d856c97a5/airflow-core/src/airflow/timetables/_delta.py#L70-L71)), so with `catchup=False` the first `DeltaTriggerTimetable` run lands at pickup time (6 AM in this example), not on a midnight tick. That is a bigger difference from the cron case than "the same split".

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -50,24 +50,37 @@ Data Interval
...
+equals the trigger time. With a non-zero ``interval=`` on a trigger timetable
+it is ``trigger_time - interval``. For a data-interval timetable it is the
+start of the contiguous window, not when the Dag is actually executed.
 
 Similarly, since the ``start_date`` argument for the Dag and its tasks points to
-the same logical date, it marks the start of *the Dag's first data interval*, not
-when tasks in the Dag will start running. In other words, a Dag run will only be
-scheduled one interval after ``start_date``.
+the same logical date, it marks the start of scheduling for the Dag (the first
+possible logical date), not when tasks in the Dag will start running. In other
```
kaxil: With a non-zero `interval=` on a trigger timetable, `start_date` is not the first possible logical date. `restriction.earliest` bounds `run_after`, and the data interval is `[run_after - interval, run_after]` ([trigger.py#L104-L112](https://github.com/apache/airflow/blob/20b9a85c557ce0b10e22401a560bf50d856c97a5/airflow-core/src/airflow/timetables/trigger.py#L104-L112)), so `start_date=2015-12-01` with `interval=timedelta(days=1)` gives a first logical date of 2015-11-30. Since the paragraph above now spells out that case, "the earliest trigger time" would be more accurate here.

The old "one interval after `start_date`" was also saying something true for data-interval timetables: the first run does not execute until its window closes. "Once the timetable reaches `start_date`" loses that.

airflow-core/docs/core-concepts/dag-run.rst
```diff
@@ -50,24 +50,37 @@ Data Interval
...
+each day). The run is created to execute *at* that time.
+
+If you need a contiguous full-day window instead — for example each run covering
+from midnight to the next midnight — use a data-interval timetable such as
+:ref:`CronDataIntervalTimetable`, or set ``[scheduler] create_cron_data_intervals``
+to ``True``. With that setup, a Dag run is usually scheduled *after* its
+associated data interval has ended, so a run covering 2020-01-01 generally does
+not start until after 2020-01-02 00:00:00.
+
+See :ref:`Differences between "trigger" and "data interval" timetables` for a
```
kaxil: Nit: a bare `:ref:` pulls in the whole section title, so this renders as `See Differences between "trigger" and "data interval" timetables for a side-by-side comparison`, quotes and all. The explicit-title form that the removed line used reads better:

```
:ref:`timetables comparison <Differences between "trigger" and "data interval" timetables>`
```

Same for lines 172-173.

kaxil (approved):

airflow-core/docs/core-concepts/dag-run.rst
Tegh25: Thanks for the feedback, I addressed the second point in this comment.

On `[scheduler] create_delta_data_intervals`, I agree that today the timedelta branch of `_create_timetable` gates on `create_cron_data_intervals`, so the delta config is unused. To me that looked like a bug rather than intended behavior, which will be fixed in https://github.com/apache/airflow/pull/69869 after it merges. Therefore this doc update keeps documenting the config as what I believe is the intended switch and does not remove the advice.

If you believe that this behavior is not a bug and should be documented as such, then I'll adjust the wording as you described. Let me know.

PR #70155, opened 2026-07-21
Document that Code view shows only Python source for stub tasks

jason810496:
- related: #70059

## Why

[ADR-0006](https://github.com/apache/airflow/pull/70059) records that the Code view shows only the Python Dag file for mixed-language (`@task.stub`) Dags, and its Consequences section relies on the user documentation stating this explicitly so the behavior is expected rather than reported as a bug.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes, with help of Claude Code Fable 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70170, opened 2026-07-21
Fix flaky static checks caused by migration-reference hook race

jason810496:
- related: #70155 (docs-only PR whose Static checks hit this flake)
- failed run: https://github.com/apache/airflow/actions/runs/29797370113/job/88533337550

## Why

The "CI image checks / Static checks" job fails intermittently with `RuntimeError: revision = not found in .../fab/migrations/versions/0000_1_4_0_create_ab_tables_if_missing.py`

The three `update-migration-references*` hooks live in three prek workspaces that run **in parallel**, yet each invocation of the shared `scripts/ci/prek/migration_reference.py` processed **all three apps** (`airflow`, `fab`, `edge3`) and rewrote every migration file and `migrations-ref.rst` even when nothing changed — so concurrent containers race on the same files, and a reader catching a file mid-write sees truncated content and fails.

## What

- Each hook's `files:` filter already scopes it to its own app's migration files/docs, so the all-apps loop was pure overlap: scope each hook invocation to its own app and the hooks no longer share any files — the race is structurally gone.
- Reproduced before the fix with CI's exact invocation (`prek --all-files` with the same `SKIP` list): the fab hook failed with the same `revision = not found` error while passing in isolation.

---

### DIFF
--- a/airflow-core/.pre-commit-config.yaml
+++ b/airflow-core/.pre-commit-config.yaml
@@ -262,6 +262,7 @@ repos:
         name: Update migration ref doc
         language: python
         entry: ../scripts/ci/prek/migration_reference.py
+        args: ["--app", "airflow"]
         pass_filenames: false
         files:
           (?x)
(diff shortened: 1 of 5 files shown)


jason810496:
This should fix the flaky CI issue pointed out in https://github.com/apache/airflow/pull/70155#issuecomment-5030338988 by @shahar1.

No need to backport, the docs live on main.

shahar1 (approved):

potiuk (approved):

PR #70176, opened 2026-07-21
Better log when spark k8s driver remains in Unknown Phase

amoghrajesh:
The retry loop that monitored spark driver pod status gave up after repeated "Unknown" pod phases with no explanation of what that phase means, leaving retry policies (and humans) with no way to tell if it was worth retrying. Kubernetes documents Unknown phase as an error communicating with the node the pod should be running on: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase which is a credible transient condition -- so
we should state that explicitly instead of leaving it opaque.

---

---


Lee-W (approved):

jason810496 (approved):
Nice, no harm to have more explicit error message, thanks.

PR #70180, opened 2026-07-21
Allow mounting the Airflow config as a directory to avoid the subPath mount race

diegoalvarez-clarity:
## Why
The chart mounts `airflow.cfg` (and `airflow_local_settings.py`) into every component and the KubernetesExecutor pod template via a `subPath` bind-mount of the `config` ConfigMap. `subPath` + ConfigMap hits a known Kubernetes race: on node/volume lifecycle events (kubelet volume reconstruction, node autoscaling) the pre-computed subPath source inode goes stale and runc fails the mount at container init:

    error during container init: error mounting
    ".../volume-subpaths/config/base/1" to rootfs at "/opt/airflow/airflow.cfg"

The task fails with `StartError` and no application log (the container never starts). It's intermittent, cross-component, multi-node, and runtime-version-independent.

## What
Add an opt-in `mountConfigAsDir` (default `false`; existing behaviour unchanged). When `true`, the `config` ConfigMap is mounted as a directory at `{airflowHome}/config` (no subPath) and `AIRFLOW_CONFIG` is set to `{airflowHome}/config/airflow.cfg`. A directory-projected ConfigMap volume has no pre-computed subPath bind source, so it's immune to the race. `airflow_local_settings.py` keeps its default path, so local-settings loading is unaffected. The separate `pod_template_file.yaml` subPath mount is out of scope.

## Tests
- `test_mount_config_as_dir` — flag on, across all ten templates that include the config mount (deployments, statefulsets, cronjobs, jobs): exactly one directory `config` mount, no `airflow.cfg` or `airflow_local_settings.py` subPath, `AIRFLOW_CONFIG` set.
- `test_config_mounted_via_subpath_by_default` — no values: historical `airflow.cfg` subPath, no local-settings subPath, no directory mount, no `AIRFLOW_CONFIG`.
- `test_local_settings_mounted_via_subpath_when_set` — `airflowLocalSettings` set: both subPath mounts.

## Notes
Backward compatible (default off).

This PR was created with the assistance of generative AI tools.


salvamadrid:
+1

Miretpl:
> The task fails with StartError and no application log (the container never starts). It's intermittent, cross-component, multi-node, and runtime-version-independent. See https://github.com/apache/airflow/discussions/33788.

The solution to the issue mentioned here is basically not to mount the `/opt/airflow` directory as a whole, and the solution is mentioned in the discussion by Jarek. If you have the same case as mentioned in the discussion, you can mount a volume at `/opt/airflow/config` (or a different location) and set the desired env variables/config options accordingly.

diegoalvarez-clarity:
Thanks @Miretpl — I think we're describing two different cases, so let me draw the line clearly.

The whole `/opt/airflow` scenario from #33788 is a user-supplied mount, and yes, mounting at `/opt/airflow/config` instead is the right fix there.

The race this PR targets is on the chart's own config mount. The shared `airflow_config_mount` helper unconditionally emits two subPath mounts — `airflow.cfg` and `airflow_local_settings.py` — onto the config volume. Those are what runc fails to bind under node/volume-lifecycle churn (kubelet volume reconstruction / node autoscaling):

`error mounting ".../volume-subpaths/config/base/1" to rootfs at "/opt/airflow/airflow.cfg"`

The container never starts, so the task hard-fails with no application log.

A values-level extraVolumeMounts at `/opt/airflow/config` is additive — it can't remove the helper's subPath mounts, so the race persists. There's no values-only way to drop them today, and that's the gap this PR closes.

`mountConfigAsDir`: true switches the shared helper from the two subPaths to a single directory mount at `/opt/airflow/config` + `AIRFLOW_CONFIG`. It's opt-in and default-off, so existing behavior is unchanged. helm template diff: subPath count 2 → 0, everything else byte-identical; `airflow_local_settings.py` stays at its default path via the directory mount.

diegoalvarez-clarity:
Hi @jedcunningham @hussein-awala @jscheffl @bugraoz93 @Miretpl 👋

Gentle nudge on this one when you get a chance. To recap where it stands: the PR adds an opt-in mountConfigAsDir flag (default false, so existing behavior is unchanged) that mounts the config ConfigMap as a directory instead of via subPath binds, which avoids the stale-inode StartError race we hit during node reconstruction / autoscaling.

@Miretpl — I followed up on your question about #33788 above: this targets the chart's own airflow.cfg / airflow_local_settings.py subPath mounts rather than user-supplied ones, since values-only extraVolumeMounts can't remove the helper's subPath mounts. Happy to expand or adjust if that framing doesn't fully address your concern.

Tests cover both the new directory-mount path and the preserved subPath behavior. I'm glad to rebase or make changes to help this move forward — thanks for taking a look! 🙏

chart/templates/_helpers.yaml
```diff
@@ -63,6 +63,10 @@ If release name contains chart name it will be used as a full name.
   # Hard Coded Airflow Envs
   - name: AIRFLOW_HOME
     value: {{ .Values.airflowHome }}
+  {{- if .Values.mountConfigAsDir }}
+  - name: AIRFLOW_CONFIG
+    value: {{ printf "%s/config/airflow.cfg" .Values.airflowHome | quote }}
```
Miretpl: It should rather user proposed `airflow_config_dir` helper.

chart/templates/_helpers.yaml
```diff
@@ -645,7 +649,16 @@ server_tls_key_file = /etc/pgbouncer/server.key
   {{- printf "%s-config" (include "airflow.fullname" .) }}
 {{- end }}
 
+{{- define "airflow_config_dir" -}}
+  {{- printf "%s/config" .Values.airflowHome | quote }}
+{{- end }}
+
 {{- define "airflow_config_mount" -}}
+{{- if .Values.mountConfigAsDir }}
```
Miretpl, line 657: Value `true` here affects rendering of the `airflow_local_settings.py` file.

chart/newsfragments/70180.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``mountConfigAsDir`` to mount the Airflow config ConfigMap as a directory (setting ``AIRFLOW_CONFIG``) instead of via ``subPath``, avoiding the Kubernetes ``subPath`` + ConfigMap stale-mount race that can fail containers at init with a ``runc`` ``StartError``.
```
Miretpl: Rather not needed newsfragment.

chart/values.schema.json
```diff
@@ -685,6 +685,12 @@
             "x-docsSection": "Common",
             "default": "See values.yaml"
         },
+        "mountConfigAsDir": {
+            "description": "Mount the Airflow config ConfigMap as a directory (at ``{airflowHome}/config``) and set ``AIRFLOW_CONFIG`` accordingly, instead of mounting ``airflow.cfg`` via ``subPath``. Avoids the Kubernetes ``subPath`` + ConfigMap stale-mount race that can fail containers at init with a ``runc`` ``StartError``. Opt-in; the default preserves the historical ``subPath`` mount.",
```
```suggestion
            "description": "Mount the Airflow config ConfigMap as a directory (at ``{airflowHome}/config``) and set ``AIRFLOW_CONFIG`` accordingly, instead of mounting ``airflow.cfg`` via ``subPath``.",
```
It is description of the config option, not the justification why it was added.

chart/values.yaml
```diff
@@ -305,6 +305,13 @@ airflowConfigAnnotations: {}
 # 'airflow_local_settings' file as a string (templated)
 airflowLocalSettings: ~
 
+# Mount the Airflow config ConfigMap as a directory (at ``{airflowHome}/config``) and set
+# ``AIRFLOW_CONFIG`` accordingly, instead of mounting ``airflow.cfg`` via ``subPath``. This avoids
+# the Kubernetes ``subPath`` + ConfigMap stale-mount race that can fail containers at init with a
+# ``runc`` ``StartError`` on node/volume lifecycle events. ``airflow_local_settings.py`` keeps its
+# default path. Opt-in; the default preserves the historical ``subPath`` mount.
```
```suggestion
# Mount the Airflow config ConfigMap as a directory (at ``{airflowHome}/config``) and set
# ``AIRFLOW_CONFIG`` accordingly, instead of mounting ``airflow.cfg`` via ``subPath``.
```

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
         for doc in docs:
             assert expected_mount in jmespath.search("spec.template.spec.initContainers[0].volumeMounts", doc)
 
+    def test_mount_config_as_dir(self):
+        components = [
```
Miretpl, line 420: We should test all components that use this mount config, not just the core.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+            "templates/workers/worker-deployment.yaml",
+            "templates/api-server/api-server-deployment.yaml",
+            "templates/triggerer/triggerer-deployment.yaml",
+            "templates/dag-processor/dag-processor-deployment.yaml",
+        ]
+        docs = render_chart(
+            values={"mountConfigAsDir": True, "airflowLocalSettings": "# local settings"},
+            show_only=components,
+        )
+        assert len(docs) == len(components)
```
Miretpl: It doesn't check anything, really. All of these components are rendered by default and if they will not, there is coverage in different tests.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+        dir_mount = {"name": "config", "mountPath": "/opt/airflow/config", "readOnly": True}
+        for doc in docs:
+            mounts = jmespath.search("spec.template.spec.containers[0].volumeMounts", doc)
```
Miretpl: We could have done that directly in the `jmespath`.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+            # airflow.cfg / airflow_local_settings.py are now served from the mounted directory
```
Miretpl:
```suggestion
```

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+            assert dir_mount in mounts
+            assert not [m for m in mounts if m.get("subPath") in ("airflow.cfg", "airflow_local_settings.py")]
```
Miretpl: We can do it with `jmespath`.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+            # AIRFLOW_CONFIG points Airflow at the config in the mounted directory
```
Miretpl:
```suggestion
```

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -416,6 +416,42 @@ def test_have_all_config_mounts_on_init_containers(self):
...
+            env = jmespath.search("spec.template.spec.containers[0].env", doc)
+            assert {"name": "AIRFLOW_CONFIG", "value": "/opt/airflow/config/airflow.cfg"} in env
+
+    def test_config_mounted_via_subpath_by_default(self):
+        docs = render_chart(
+            values={"airflowLocalSettings": "# local settings"},
```
Miretpl, line 477: If you are setting this value here, you don't really checking the default as it overrides default config for one of the affected by this change areas.

Miretpl (changes requested):
This PR lacks many things from logic verification to the test coverage. As it is stated that there was AI used, please for the future verify carefully the content of the PR.

Miretpl:
> @Miretpl — I followed up on your question about https://github.com/apache/airflow/discussions/33788 above: this targets the chart's own airflow.cfg / airflow_local_settings.py subPath mounts rather than user-supplied ones, since values-only extraVolumeMounts can't remove the helper's subPath mounts. Happy to expand or adjust if that framing doesn't fully address your concern.

So, as it is a different issue, the mentioned issue should not be mentioned in the description.

chart/templates/_helpers.yaml
diegoalvarez-clarity: Done in 1342d421ee. The helper now returns the unquoted path and both call sites quote it, so the config subdirectory is encoded once. Rendered output is unchanged.

chart/templates/_helpers.yaml
diegoalvarez-clarity: Correct. With the flag on the whole subPath branch is skipped, including the conditional airflow_local_settings.py mount. The file still lands at the same path because airflow_local_setting_path already resolves inside the mounted directory. test_mount_config_as_dir now asserts, on every component, exactly one directory config mount and no airflow.cfg or airflow_local_settings.py subPath (05572dab2e).

chart/newsfragments/70180.feature.rst
diegoalvarez-clarity: Removed in 8ef39ae7a5.

chart/values.schema.json
diegoalvarez-clarity: Reworded to describe the option only, in 8ef39ae7a5.

chart/values.yaml
diegoalvarez-clarity: Applied your wording in 8ef39ae7a5.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Now covers all ten templates that include airflow_config_mount: flower, both cleanup cronjobs and the two jobs added, with the CronJob pod path handled by kind (05572dab2e).

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: You are right. Removed.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Done, both assertions are jmespath filters now.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Done, both assertions are jmespath filters now.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Removed.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Removed.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
diegoalvarez-clarity: Split in 05572dab2e. test_config_mounted_via_subpath_by_default renders with no values and also asserts no directory mount and no AIRFLOW_CONFIG. The airflowLocalSettings case moved to test_local_settings_mounted_via_subpath_when_set.

diegoalvarez-clarity:
Thanks for the thorough review, @Miretpl. All comments are addressed in three commits, one per area: the helper reuse (1342d421ee), the values and schema wording plus the newsfragment removal (8ef39ae7a5), and the test rework (05572dab2e). The directory-mount test now covers all ten templates that include the config mount, and the default-behaviour test renders with no values. On the airflow_local_settings.py point I replied on the thread with why the path is unchanged and added an assertion for it.

I've also removed the misleading reference to the other issue. Sorry for the confusion.

Ready for another look.

PR #70178, opened 2026-07-21
Include spark submit canonical logs in failure exceptions

amoghrajesh:
### Why?

Airflow only surfaced spark-submit's exit code when the process failed, discarding the JVM's actual stdout/stderr (missing jar, bad main class, Kerberos errors, etc.). Humans and Retry policies evaluating these exceptions had nothing but an exit code to reason from. 

Some cases I encountered are when Spark's `SparkSubmitHook.submit()` raises `AirflowException` on three
failure paths: 

- a nonzero spark-submit exit code, 
- a Kubernetes exit-code mismatch, and 
- a standalone-cluster submit that produced no driver id.

In all three cases the exception message contained only the masked command line and an exit code -- never the spark-submit process's own stdout/stderr, even though that output already streams through `_process_spark_submit_log()` line by line and often contains the real cause (missing jar, bad main class, Kerberos failure, etc.).

This makes root-causing failures harder for humans, and gives automated/LLM-based retry policies nothing to work with beyond an exit code cos they can't distinguish a transient submit infra hiccup from a permanent misconfiguration.

### What

This PR adds a small rolling buffer of the last 20 lines of spark-submit's own output (`_last_submit_log_lines`, populated in
`_process_spark_submit_log`) and appends it, masked the same way the command line already is, to all three submit-failure exceptions via a new `_format_submit_log_tail()` helper. Example:

```shell
            "Cannot execute: spark-submit --master spark://spark-master:7077 --deploy-mode cluster "
            "/opt/spark/apps/does-not-exist.jar. Error code is: 1.\n"
            "Last spark-submit output:\n"
            "WARNING: Using incubator modules: jdk.incubator.vector\n"
            "26/07/21 07:03:03 WARN NativeCodeLoader: Unable to load native-hadoop library for your "
            "platform... using builtin-java classes where applicable\n"
            'Exception in thread "main" org.apache.spark.SparkException: Failed to get main class in '
            "JAR with error 'File file:/opt/spark/apps/does-not-exist.jar does not exist'.  Please "
            "specify one with --class.\n"
            "\tat org.apache.spark.deploy.SparkSubmit.error(SparkSubmit.scala:1065)\n"
            "\tat org.apache.spark.deploy.SparkSubmit.prepareSubmitEnvironment(SparkSubmit.scala:547)\n"
            "\tat org.apache.spark.deploy.SparkSubmit.doSubmit(SparkSubmit.scala:95)"
```

---

##### Was generative AI tooling used to co-author this PR?

[...]

PR #70183, opened 2026-07-21
Surface YARN diagnostics for spark on yarn clusters

amoghrajesh:
---

Continuation of the exception-enrichment work for spark like: [Better log when spark k8s driver remains in Unknown Phase](https://github.com/apache/airflow/pull/70176) and [Include spark submit canonical logs in failure exceptions](https://github.com/apache/airflow/pull/70178). Retry policies (LLM-based or deterministic) can only reason from what in the exception message, and Spark's YARN tracking
path was the thinnest of the three deploy modes -- `_start_yarn_application_status_tracking`
raised only `"ended with state: FAILED, final status: FAILED"`, with no indication of *why*.

## Solution

YARN has a "diagnostics" field to provide detailed diag bundle for failures: https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html

Rather than assume the RM API's `diagnostics` field is populated with anything useful, I triggered real jobs against a live cluster and captured the actual RM responses.

| Job tested | Diagnostics returned by RM | Message before | Message after |
|---|---|---|---|
| MapReduce job whose AM container fails to launch (missing `HADOOP_MAPRED_HOME` env) | `Application application_...0002 failed 2 times due to AM Container for appattempt_...000002 exited with exitCode: 1 ... Error: Could not find or load main class org.apache.hadoop.mapreduce.v2.app.MRAppMaster ...` | `YARN application application_...0002 ended with state: FAILED, final status: FAILED` | `YARN application application_...0002 ended with state: FAILED, final status: FAILED`<br>`Diagnostics: Application application_...0002 failed 2 times due to AM Container for appattempt_...000002 exited with exitCode: 1 ... Error: Could not find or load main class org.apache.hadoop.mapreduce.v2.app.MRAppMaster ...` |
| Long-running `pi` job killed mid-flight via `yarn application -kill` | `Application application_...0005 was killed by user root at 192.168.117.5` | `YARN application application_...0005 ended with state: KILLED, final status: KILLED` | `YARN application application_...0005 ended with state: KILLED, final status: KILLED`<br>`Diagnostics: Application application_...0005 was killed by user root at 192.168.117.5` |

Both captured payloads were fed into `LLMRetryPolicy` across runbook instruction variants and all consistently classified `FAIL`, including on bare defaults (`none`), once the message carried the real diagnostics text. The killed job case also resolved an open question: "was killed by user root at \<ip\>" is unambiguous, distinct from how YARN phrases
scheduler-preemption terminations -- so diagnostics alone is enough to tell an explicit kill from an infra-driven one.

## Proposed Change

- `_query_yarn_application_status` now returns `tuple[str, str, str]` (`state`, `finalStatus`, `diagnostics`), reading `app.get("diagnostics", "")`.
- Both failure raises in `_start_yarn_application_status_tracking` append `\nDiagnostics: {diagnostics}` when present; unchanged when absent.
- `query_yarn_application_status` (the normalized-status public method) updated to unpack the 3-tuple.

---

### DIFF
--- a/providers/apache/spark/src/airflow/providers/apache/spark/hooks/spark_submit.py
+++ b/providers/apache/spark/src/airflow/providers/apache/spark/hooks/spark_submit.py
@@ -907,7 +907,7 @@ class SparkSubmitHook(BaseHook, LoggingMixin):
         while True:
             self.log.debug("Polling YARN RM REST API for application %s", application_id)
             try:
-                state, final_status = self._query_yarn_application_status(application_id)
+                state, final_status, diagnostics = self._query_yarn_application_status(application_id)
             except RuntimeError as exc:
                 consecutive_failures += 1
                 if consecutive_failures > max_consecutive_failures:
@@ -933,16 +933,18 @@ class SparkSubmitHook(BaseHook, LoggingMixin):
             elif poll_count % heartbeat_interval == 0:
                 self.log.info("YARN application %s is still %s", application_id, state)
 
+            diagnostics_suffix = f"\nDiagnostics: {diagnostics}" if diagnostics else ""
             if state in self._YARN_FINAL_FAILURES:
                 raise RuntimeError(
                     f"YARN application {application_id} ended with state: {state}, "
-                    f"final status: {final_status}"
+                    f"final status: {final_status}{diagnostics_suffix}"
                 )
             if final_status == self._YARN_FINAL_SUCCESS:
                 return
             if final_status in self._YARN_FINAL_FAILURES:
                 raise RuntimeError(
                     f"YARN application {application_id} ended with final status: {final_status}"
+                    f"{diagnostics_suffix}"
                 )
             if final_status != self._YARN_FINAL_UNDEFINED:
                 raise RuntimeError(
@@ -1001,8 +1003,15 @@ class SparkSubmitHook(BaseHook, LoggingMixin):
 
         return None
 
-    def _query_yarn_application_status(self, application_id: str) -> tuple[str, str]:
-        """GET ``/ws/v1/cluster/apps/{id}`` once and return ``app.state`` and ``app.finalStatus``."""
+    def _query_yarn_application_status(self, application_id: str) -> tuple[str, str, str]:
+        """
+        GET ``/ws/v1/cluster/apps/{id}`` once.
+
+        Returns ``app.state``, ``app.finalStatus``, and ``app.diagnostics`` - diagnostics is
+        where YARN puts the actual human readable failure reason (AM launch error, container
+        OOM, explicit kill, etc.), so failure exceptions can include it instead of just the
+        two terminal-state enum values.
+        """
         url = f"{self._get_yarn_rm_base_url()}/ws/v1/cluster/apps/{application_id}"
         try:
             resp = requests.get(url, auth=self._resolved_yarn_rm_auth, timeout=self._HTTP_TIMEOUT)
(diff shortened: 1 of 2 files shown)


Lee-W (approved):

providers/apache/spark/tests/unit/apache/spark/hooks/test_spark_submit.py
```diff
@@ -1773,6 +1778,42 @@ def test_yarn_status_tracking_fails_on_failed_state_with_undefined_final_status(
 
         mock_sleep.assert_not_called()
 
+    @patch("airflow.providers.apache.spark.hooks.spark_submit.time.sleep")
+    @patch("airflow.providers.apache.spark.hooks.spark_submit.requests.get")
+    def test_yarn_status_tracking_includes_diagnostics_on_state_failure(self, mock_get, mock_sleep):
+        """RM state FAILED/KILLED -> raised message includes the RM's diagnostics field."""
+        mock_get.return_value = self._rm_status_resp(
+            "KILLED", diagnostics="Application application_1700000000000_0001 was killed by user root"
```
kaxil: This test is named `..._on_state_failure` and the docstring says "RM state FAILED/KILLED", but `_rm_status_resp("KILLED", ...)` sets `finalStatus=KILLED` with `state` defaulting to `FINISHED`. That lands in the `final_status in _YARN_FINAL_FAILURES` branch -- the same one `test_yarn_status_tracking_includes_diagnostics_on_final_status_failure` already covers. The `state in _YARN_FINAL_FAILURES` raise, where you added `{diagnostics_suffix}` to the `ended with state:` message, never actually gets exercised. Passing `state="KILLED"` (or `state="FAILED"`) here would cover that branch.

providers/apache/spark/src/airflow/providers/apache/spark/hooks/spark_submit.py
```diff
@@ -933,16 +933,18 @@ def _start_yarn_application_status_tracking(self, application_id: str) -> None:
...
+                    f"final status: {final_status}{diagnostics_suffix}"
                 )
             if final_status == self._YARN_FINAL_SUCCESS:
                 return
             if final_status in self._YARN_FINAL_FAILURES:
                 raise RuntimeError(
                     f"YARN application {application_id} ended with final status: {final_status}"
+                    f"{diagnostics_suffix}"
                 )
             if final_status != self._YARN_FINAL_UNDEFINED:
```
kaxil, line 949: This "returned unexpected final status" raise doesn't append `diagnostics_suffix` like the two failure raises above it. YARN's `FinalApplicationStatus` includes `ENDED`, which would fall through to here, and it's arguably where diagnostics helps most since the final status itself is unrecognized. Intentional, or worth appending the suffix here too?

kaxil (approved):

providers/apache/spark/tests/unit/apache/spark/hooks/test_spark_submit.py
amoghrajesh: Fixed by passing `state="KILLED"` explicitly so it exercises the state in `_YARN_FINAL_FAILURES` raise this test is actually meant to cover.

providers/apache/spark/src/airflow/providers/apache/spark/hooks/spark_submit.py
amoghrajesh: Not intentional, added {diagnostics_suffix} there too, plus a dedicated test. Agreed it's the branch where diagnostics helps most since the status itself is already unrecognized.

amoghrajesh: Handled in 34c1c26d49

providers/apache/spark/tests/unit/apache/spark/hooks/test_spark_submit.py
amoghrajesh: 34c1c26d49

amoghrajesh:
cc @nailo2c fyi

amoghrajesh:
Unrelated failures, merging.

Issue #70093, opened 2026-07-19
Add deferrable support to LLM operators in common.ai provider


guan404ming:
## Why 

LLM calls in LLMOperator / AgentOperator block a worker slot for the whole API call (30s to minutes). Airflow's deferral mechanism fits this perfectly.

## Proposal

Step 1 — LLMTrigger + deferrable LLMOperator

- Run the LLM call in the triggerer via async agent.run(), freeing the worker slot
- deferrable=True on LLMOperator, rehydrate output_type on resume
- Chain with require_approval (HITL) so the whole flow never holds a worker

Step 2 — Deferrable subclasses

- LLMBranchOperator (branch on resume)
- LLMSQLOperator, LLMSchemaCompareOperator

Step 3 — AgentTrigger for AgentOperator

- Multi-turn agent in the triggerer, message history serialized
- Toolsets rebuilt from serializable spec (conn_id + hook path)
- Durable-storage replay on triggerer restart, no re-billed LLM calls

## Use case

- Hundreds of concurrent LLM tasks without holding hundreds of worker slots
- Long-running reasoning/agent tasks (10+ min) become cheap to run

br413:
Working on Step 1 (LLMTrigger + deferrable LLMOperator) — will open a PR shortly.

kaxil:
Thanks for writing this up, and the motivation is real: LLM and agent tasks shouldn't hold a worker slot idle for the whole call. But running the LLM/agent work in the triggerer isn't the right mechanism, so I'd rather not take this direction.

The triggerer is for tasks that have nothing to do but wait on external state. A trigger's `run()` is a poll (is the job done, does the file exist, has the time passed): idempotent, side-effect-free, and safe to re-run. That last property is the one that matters here, and the docs call it out directly: "you should assume that a trigger instance can run more than once ... so you must be mindful about side effects" (`deferring.rst`). Triggers are re-created from their serialized form and re-run on any triggerer restart, redeploy, or HA redistribution.

[...]

PR #70184, opened 2026-07-21
Add deferrable LLMOperator with LLMTrigger

br413:
## Summary
- Add `LLMTrigger` to run pydantic-ai `agent.run()` in the triggerer
- Add `deferrable=True` on `LLMOperator` to free worker slots during LLM API calls
- Rehydrate `output_type` on resume; `require_approval` chains after deferrable completion

Step 1 of #70093

## Test plan
- [ ] `pytest providers/common/ai/tests/unit/common/ai/triggers/test_llm.py`
- [ ] `pytest providers/common/ai/tests/unit/common/ai/operators/test_llm.py -k Deferrable`
- [ ] `pytest providers/common/ai/tests/unit/common/ai/utils/test_output_type.py -k Serialization`
- [ ] CI provider-common-ai checks pass


providers/common/ai/src/airflow/providers/common/ai/triggers/llm.py
```diff
@@ -0,0 +1,112 @@
...
+    if usage_limits is None:
+        return None
+    return usage_limits.model_dump(exclude_defaults=True)
+
+
+def deserialize_usage_limits(data: dict[str, Any] | None) -> Any | None:
+    """Restore pydantic-ai ``UsageLimits`` from trigger persistence."""
+    if data is None:
+        return None
+    from pydantic_ai.usage import UsageLimits
```
Lee-W: why do we need to do inline import here

providers/common/ai/src/airflow/providers/common/ai/operators/llm.py
```diff
@@ -154,6 +164,25 @@ def llm_hook(self) -> PydanticAIHook:
         }
         return PydanticAIHook.get_hook(self.llm_conn_id, hook_params=hook_params)
 
+    def _finalize_output(self, output: Any) -> Any:
+        if self._serialize_model_output and isinstance(output, BaseModel):
+            return output.model_dump()
+        return output
+
+    def _defer_llm_call(self) -> None:
```
Lee-W: why do we need to make it a separate method

providers/common/ai/src/airflow/providers/common/ai/operators/llm.py
```diff
@@ -173,17 +206,31 @@ def execute(self, context: Context) -> Any:
...
+                self.output_type, output, serialize_output=self._serialize_model_output
+            )
 
-    def execute_complete(self, context: Context, generated_output: str, event: dict[str, Any]) -> Any:
-        """Resume after human review and restore the Pydantic model for XCom consumers."""
-        output = super().execute_complete(context, generated_output, event)
-        return rehydrate_pydantic_output(
-            self.output_type, output, serialize_output=self._serialize_model_output
+        if event.get("status") == "error":
+            raise AirflowException(event.get("message", "LLM call failed in deferrable mode"))
```
Lee-W: We decided not to raise AirflowException directly https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#don-t-raise-airflowexception-directly

providers/common/ai/src/airflow/providers/common/ai/triggers/llm.py
br413: Good catch — there was no need for a lazy import here. pydantic-ai is already a provider dependency, so I moved UsageLimits to the module level in 061ad47.

providers/common/ai/src/airflow/providers/common/ai/operators/llm.py
br413: Agreed — those helpers did not carry enough logic to justify the indirection. I inlined the defer call in execute() and the output serialization at each return path in 061ad47.

providers/common/ai/src/airflow/providers/common/ai/operators/llm.py
br413: Thanks for the pointer to the contributing guide. I added LLMOperatorException under providers/common/ai/exceptions.py (same pattern as OpenAIBatchJobException) and use that in execute_complete instead of raising AirflowException directly. Updated in 061ad47.

kaxil:
Thanks for the work here, and for turning the review comments around quickly.

I'm going to close this along with #70093. The short version: the triggerer is the wrong home for an LLM/agent call. A trigger's `run()` has to be idempotent and side-effect-free because it re-runs on any triggerer restart or redistribution, and a billed, non-deterministic `agent.run()` with tools executing in-process is neither. `LLMTrigger.run()` here re-issues the whole call from scratch on every relaunch, so a restart mid-flight re-bills it, which is the opposite of the "no re-billed calls" property the issue is after. I've left the full reasoning, and the primitives that do fit (3.2 async tasks for slot-free concurrency, `AgentOperator(durable=True)` for replay-safe long runs), on #70093.

Appreciate the effort, and sorry to send it back. This one's an architecture-direction call, not a comment on the code itself.

Issue #68661, opened 2026-06-17
Introduce RUN-level OpenLineage Events for dbt Cloud Jobs in Airflow Integration


mishrasangeeta87:
### Airflow DAG

```from airflow import DAG
from datetime import datetime
from airflow.providers.dbt.cloud.operators.dbt import DbtCloudRunJobOperator

default_args = {
    "start_date": datetime(2026, 6, 9),
    "schedule_interval": None,
}

with DAG("dbt_cloud_dag", default_args=default_args) as dag:

    dbt_cloud_job = DbtCloudRunJobOperator(
        task_id="dbt_cloud_job_run",
        job_id=70471823610018,
        dbt_cloud_conn_id="dbt_cloud_default",
        wait_for_termination=True,
        check_interval=10,
        timeout=300,
    )
```

### Airflow Logs

```Waiting for job run 70471883265346 to terminate.
Getting the status of job run 70471883265346.
Current status of job run 70471883265346:  STARTING
```

The operator successfully triggers a dbt Cloud Job and stores the run information in Airflow XCom. The dbt Cloud run ID can be retrieved successfully.

However:

- No OpenLineage START event is emitted for the dbt Cloud Job execution.
- No OpenLineage COMPLETE or FAIL event is emitted for the dbt Cloud Job execution.
- The dbt Cloud run ID (70471883265346) is not represented as an OpenLineage run.
- The emitted model/test/source/snapshot events directly reference the Airflow DAG run through the parent facet.
- There is no OpenLineage entity representing the dbt Cloud Job execution itself.

### Current Hierarchy

```
 Airflow Task Run
├── dbt Model
├── dbt Model
├── dbt Test
├── dbt Snapshot
└── dbt Source
```

The dbt Cloud Job layer is completely missing.

---

## Problem Statement

dbt Cloud introduces a Job abstraction that does not exist in dbt Core.

A dbt Cloud Job execution has:

- A unique Run ID
- Start time
- End time
- Status
- Execution metadata
- Parent orchestration context

Therefore, a dbt Cloud Job execution should be represented as a first-class OpenLineage RUN.

Currently, every model, test, source, and snapshot event emitted by dbt Cloud references the Airflow Job run as its parent. This results in an incorrect lineage hierarchy:

```
Airflow Task Run
├── dbt Model
├── dbt Test
├── dbt Snapshot
└── dbt Source
```

The actual execution hierarchy should be:

```
Airflow Task Run
└── dbt Cloud Job Run
[...]

PR #70185, opened 2026-07-21
Attach dbt Cloud job metadata to Airflow OpenLineage events

br413:
## Summary
- Add `_build_dbt_cloud_job_lineage` to expose `dbt_cloud_job_id`, `dbt_cloud_run_id`, and `dbt_cloud_account_id` as OpenLineage job tags on the Airflow task event
- Wire into `get_openlineage_facets_on_start` / `get_openlineage_facets_on_complete` for `DbtCloudRunJobOperator` and `DbtCloudJobRunSensor`
- Short-term approach per maintainer feedback on #68661 until [OpenLineage#4720](https://github.com/OpenLineage/OpenLineage/issues/4720) lands

Relates to #68661

## Test plan
- [ ] `pytest providers/dbt/cloud/tests/unit/dbt/cloud/utils/test_openlineage.py -k BuildDbtCloudJobLineage`
- [ ] CI provider-dbt-cloud checks pass

### DIFF
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/operators/dbt.py
@@ -351,20 +351,29 @@ class DbtCloudRunJobOperator(BaseOperator):
         """Returns DBT Cloud hook."""
         return DbtCloudHook(self.dbt_cloud_conn_id, **self.hook_params)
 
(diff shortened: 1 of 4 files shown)


kacpermuda:
Let's go for the proper long term approach here instead of patching it. After we introduce a proper top level events, we can introduce some custom dbt cloud run facet that will gold this information instead of using tags. This will give the OL consumers a clear way of consuming this information, rather than switching from this temp to a proper one later. Since this is a new feature, we can simply delay it a bit and implement it properly.

PR #70190, opened 2026-07-21
Add more provider module categories to the provider page

Lee-W:
## Why
Follow up of https://github.com/apache/airflow/pull/70122. Extra links, queues, plugins, auth managers, DB managers, and dialects were not browsable in the registry

## What
Add extra links, queues, plugins, auth managers, DB managers, and dialects. Each of them now gets it own category tab. The row now sorted by module count and tail collapses into a "More" menu so the page won't be too messy

image

image

 

---

---

### DIFF
--- a/dev/registry/extract_parameters.py
+++ b/dev/registry/extract_parameters.py
@@ -54,7 +54,12 @@ from pathlib import Path
 import yaml
 from extract_metadata import fetch_provider_inventory, read_inventory
 from registry_contract_models import validate_modules_catalog, validate_provider_parameters
-from registry_tools.types import BASE_CLASS_IMPORTS, CLASS_LEVEL_SECTIONS, MODULE_LEVEL_SECTIONS
+from registry_tools.types import (
+    BASE_CLASS_IMPORTS,
+    CLASS_LEVEL_CATEGORY_OVERRIDES,
+    CLASS_LEVEL_SECTIONS,
+    MODULE_LEVEL_SECTIONS,
+)
 
 AIRFLOW_ROOT = Path(__file__).parent.parent.parent
 SCRIPT_DIR = Path(__file__).parent
@@ -598,12 +603,6 @@ def discover_classes_from_provider(
                 continue
             cls = candidate
 
-            # Use section name as category for class-level entries
-            category_map = {
-                "notifications": "notifications",
-                "secrets-backends": "secrets",
-            }
-
             discovered.append(
                 make_entry(
                     cls,
@@ -611,7 +610,7 @@ def discover_classes_from_provider(
                     module_type,
                     class_path,
                     module_path,
-                    category=category_map.get(section_name, section_name),
+                    category=CLASS_LEVEL_CATEGORY_OVERRIDES.get(section_name, section_name),
                 )
             )
 
--- a/dev/registry/extract_parameters.py
+++ b/dev/registry/extract_parameters.py
@@ -629,57 +629,36 @@ def discover_classes_from_provider(
                 )
             )
 
-    # --- Plugins (dict entries with "name" + a class-path field, see types.py) ---
-    plugin_class_field = DICT_SHAPED_CLASS_LEVEL_SECTIONS["plugins"][1]
-    for plugin in provider_yaml.get("plugins", []):
-        if not isinstance(plugin, dict):
-            continue
-        if not (class_path := plugin.get(plugin_class_field, "")):
-            continue
-        if (resolved := _resolve_dotted_path(class_path)) is None:
-            continue
-        module_path, class_name, candidate = resolved
-        if candidate is None or not inspect.isclass(candidate):
-            log.warning("%s is not a class", class_path)
-            continue
+    # --- Dict-shaped class-level sections (plugins, dialects; each entry is a
+    # dict carrying an integration-name field plus a class-path field, see types.py) ---
+    for section_name, (
+        module_type,
+        class_field,
+        integration_field,
+    ) in DICT_SHAPED_CLASS_LEVEL_SECTIONS.items():
+        for entry in provider_yaml.get(section_name, []):
+            if not isinstance(entry, dict):
+                continue
+            if not (class_path := entry.get(class_field, "")):
+                continue
+            if (resolved := _resolve_dotted_path(class_path)) is None:
+                continue
+            module_path, class_name, candidate = resolved
+            if candidate is None or not inspect.isclass(candidate):
+                log.warning("%s is not a class", class_path)
+                continue
 
-        discovered.append(
-            make_entry(
-                candidate,
-                class_name,
-                "plugin",
-                class_path,
-                module_path,
-                integration=plugin.get("name", ""),
-                category=CLASS_LEVEL_CATEGORY_OVERRIDES.get("plugins", "plugins"),
+            discovered.append(
+                make_entry(
+                    candidate,
+                    class_name,
+                    module_type,
+                    class_path,
+                    module_path,
+                    integration=entry.get(integration_field, ""),
+                    category=CLASS_LEVEL_CATEGORY_OVERRIDES.get(section_name, section_name),
+                )
             )
-        )
-
-    # --- Dialects (dict entries with "dialect-type" + a class-path field, see types.py) ---
-    dialect_class_field = DICT_SHAPED_CLASS_LEVEL_SECTIONS["dialects"][1]
-    for dialect in provider_yaml.get("dialects", []):
-        if not isinstance(dialect, dict):
-            continue
-        if not (class_path := dialect.get(dialect_class_field, "")):
-            continue
-        if (resolved := _resolve_dotted_path(class_path)) is None:
-            continue
-        module_path, class_name, candidate = resolved
-        if candidate is None or not inspect.isclass(candidate):
-            log.warning("%s is not a class", class_path)
-            continue
-
-        discovered.append(
-            make_entry(
-                candidate,
-                class_name,
-                "dialect",
-                class_path,
-                module_path,
-                integration=dialect.get("dialect-type", ""),
-                category=CLASS_LEVEL_CATEGORY_OVERRIDES.get("dialects", "dialects"),
-            )
-        )
 
     # --- Task decorators (class-name key in each entry) ---
     for decorator in provider_yaml.get("task-decorators", []):
--- a/dev/registry/extract_parameters.py
+++ b/dev/registry/extract_parameters.py
@@ -359,9 +359,9 @@ def _should_skip_class(name: str) -> bool:
     return False
 
 
-def _get_first_docstring_line(cls: type) -> str | None:
+def _get_first_docstring_line(obj: object) -> str | None:
     """Return the first non-empty line of a class docstring, or None."""
-    doc = getattr(cls, "__doc__", None)
+    doc = getattr(obj, "__doc__", None)
     if not doc:
         return None
     for line in doc.strip().splitlines():
@@ -429,6 +429,27 @@ def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
         return False
 
     return "execute_resumable" in source
+
+
+def _resolve_dotted_path(class_path: str) -> tuple[str, str, object] | None:
+    """Split a dotted ``module.name`` path and import ``name`` from that module.
+
+    Returns ``None`` (without logging) if ``class_path`` has no dot, or ``None``
+    (after logging a warning) if the module import fails. Otherwise returns
+    ``(module_path, name, obj)``, where ``obj`` may be ``None`` if the module
+    has no such attribute — callers decide how to treat a missing attribute.
+    """
+    parts = class_path.rsplit(".", 1)
+    if len(parts) != 2:
+        return None
+    module_path, name = parts
+    try:
+        mod = importlib.import_module(module_path)
+        obj = getattr(mod, name, None)
+    except Exception:
+        log.warning("Could not import %s", class_path)
+        return None
+    return module_path, name, obj
 
 
 def discover_classes_from_provider(
@@ -588,16 +609,9 @@ def discover_classes_from_provider(
         for class_path in provider_yaml.get(section_name, []):
             if not class_path or not isinstance(class_path, str):
                 continue
-            parts = class_path.rsplit(".", 1)
-            if len(parts) != 2:
+            if (resolved := _resolve_dotted_path(class_path)) is None:
                 continue
-            module_path, class_name = parts
-            try:
-                mod = importlib.import_module(module_path)
-                candidate = getattr(mod, class_name, None)
-            except Exception:
-                log.warning("Could not import %s", class_path)
-                continue
+            module_path, class_name, candidate = resolved
             if candidate is None or not inspect.isclass(candidate):
                 log.warning("%s is not a class", class_path)
                 continue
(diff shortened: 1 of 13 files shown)


potiuk (approved):

dev/registry/extract_parameters.py
```diff
@@ -586,6 +586,10 @@ def make_entry(
                 "secrets-backends": "secrets",
                 "logging": "logging",
                 "executors": "executors",
+                "extra-links": "extra-links",
+                "queues": "queues",
+                "auth-managers": "auth-managers",
+                "db-managers": "db-managers",
```
phanikumv: Ideally an entry here would be useful when it renames something , something like `"secrets-backends": "secrets"`

these 4 entries are no-ops. The input and output are the same string, which is exactly what the fallback produces anyway. Delete those four lines and the behavior will be identical.

dev/registry/registry_tools/types.py
```diff
@@ -109,6 +109,48 @@
...
+        "suffixes": [],
+        "label": "Extra Links",
+        "icon": "I",
+    },
+    "queue": {
+        "yaml_key": "queues",
+        "level": "flat",
+        "suffixes": [],
+        "label": "Queues",
+        "icon": "Q",
```
kaxil, line 131: What's this?

dev/registry/registry_tools/types.py
```diff
@@ -109,6 +109,48 @@
...
+        "label": "DB Managers",
+        "icon": "D",
+    },
+    "dialect": {
+        "yaml_key": "dialects",
+        "level": "flat",
+        "suffixes": [],
+        "label": "Dialects",
+        "icon": "C",
+    },
```
kaxil, line 160: Queues and dialect feel weird

Lee-W: Queues look something like this:

image

And dialects look like this:

image

I'm a bit unsure about adding dialects, but I do think we should include queues.

I also considered removing all of these less common fields altogether, but in the end I decided to include them since they don't add much noise and may still be useful for some users.

dev/registry/registry_tools/types.py
Lee-W: consolidate the discussion in https://github.com/apache/airflow/pull/70190#discussion_r3627054316

dev/registry/extract_parameters.py
Lee-W: sounds good. also removed logging and executors above

dev/registry/registry_tools/types.py
```diff
@@ -139,11 +181,19 @@
...
+#
+# "plugins" and "dialects" are also flat/class-path sections, but each entry is a
+# dict (plugin-class / dialect-class-name) rather than a bare string, so they are
+# handled by their own loops in extract_parameters.py instead of this generic one.
 CLASS_LEVEL_SECTIONS: dict[str, str] = {
     "notifications": "notifier",
     "secrets-backends": "secret",
     "logging": "logging",
     "executors": "executor",
+    "extra-links": "extra_link",
```
phanikumv, line 202: `extract_versions.py` doesn't read these `CLASS_LEVEL_SECTIONS` — it has its own hardcoded `FQCN_SECTIONS` (line 304) that still stops at notifications/secrets/logging/executors. So backfilled version pages won't show any of the new categories, only latest pages will. Was that intentional? 

It'd be nice if `extract_versions.py` picked these up from here the way it already does with `MODULE_LEVEL_SECTIONS`.

dev/registry/registry_tools/types.py
kaxil: Aah **Message Queues** -- I think we should relabel it to "**Message Queues**" so folks aren't confused with things like "queue" argument on Base Operator. 

https://airflow.apache.org/docs/apache-airflow-providers-common-messaging/stable/index.html

Lee-W: Just updated it. Thanks!

dev/registry/registry_tools/types.py
Lee-W: Thanks for pointing out! It wasn't intentional. `extract_versions.py` now uses `CLASS_LEVEL_SECTIONS` in `types.py` instead of a separate hardcoded list, so backfilled version will show all the same categories . Also moved the category-override rule into a shared `CLASS_LEVEL_CATEGORY_OVERRIDES` constant so `extract_parameters.py` and `extract_versions.py` won't drift again.

Lee-W:
image

dev/registry/extract_versions.py
```diff
@@ -300,15 +320,13 @@ def process_module(module_path: str, module_type: str, integration: str, categor
...
-        "secrets-backends": ("secret", "secrets", "secrets backend"),
-        "logging": ("logging", "logging", "log handler"),
-        "executors": ("executor", "executors", "executor"),
-    }
-    for yaml_key, (mod_type, category, desc_suffix) in FQCN_SECTIONS.items():
+    # Class-level sections (full class paths, no source file parsing needed).
+    # Iterating CLASS_LEVEL_SECTIONS keeps this in sync with the single
+    # source of truth in types.py; a section added there without a matching
+    # FQCN_DESC_SUFFIXES entry raises KeyError instead of silently skipping.
+    for yaml_key, mod_type in CLASS_LEVEL_SECTIONS.items():
```
kaxil, line 330: Good that this tracks `CLASS_LEVEL_SECTIONS` now, but `plugins` and `dialects` are deliberately excluded from that dict (dict-shaped entries), and they only get their own loops in `extract_parameters.py`. So version pages built here still miss those two types while the latest page has them: postgres, common.sql and microsoft.mssql get a Dialects tab on latest and nothing on older versions, same for the seven providers with `plugins` (databricks, edge3, informatica, openlineage, apache.hive, apache.kafka, common.ai).

`test_all_class_level_sections_produce_covered_types` won't catch it either, since it iterates the same dict.

registry/src/_data/types.json
```diff
@@ -58,5 +58,35 @@
     "id": "toolset",
     "label": "Toolsets",
     "icon": "U"
+  },
+  {
+    "id": "extra_link",
```
kaxil, line 68: Each entry here feeds more CSS than `.tab-icon.<id>`. `_data/statsData.js` builds one stats card per types.json entry with `colorClass: t.id` and no zero-count filter, so `/stats/` needs `.type-icon.<id>` and `.share-bar.<id>`; `provider-card.njk`'s breakdown bar needs `.provider-card .modules .<id>`. All three families stop at `decorator` in main.css, so the six new ids get transparent slices in the listing-page breakdown bar (`.breakdown > div` has no default background, though widths still sum to 100% and tooltips still work), six 2.5rem chips on `/stats/` with no background or icon colour, and gray share-bars since `--meter-color` falls back to `--color-gray-500`.

The `--color-*` tokens are already added in tokens.css, so it's one line per id in each of the three blocks. `toolset` has the same gap from #70122.

registry/src/js/provider-detail.js
```diff
@@ -155,6 +157,44 @@
...
+    });
+
+    document.addEventListener('keydown', function(e) {
+      if (e.key === 'Escape') closeMoreMenu();
+    });
+
+    // Selecting a tab inside the menu also closes it (filtering itself is
+    // handled by the shared moduleTabs click listener above).
+    moduleTabMoreMenu.querySelectorAll('.module-tab').forEach(function(tab) {
+      tab.addEventListener('click', closeMoreMenu);
```
kaxil, line 199: The shared `moduleTabs` listener does `moduleTabs.forEach(t => t.classList.remove('active'))` and then marks the clicked tab, and the tab clicked here is inside the menu that just got hidden. So after picking Dialects from More, no visible tab is highlighted (All lost `active` too) and there's nothing showing what the list is filtered by. Adding `active` to the More button when the selected type is in `overflowTypes` would cover it.

registry/src/js/provider-detail.js
```diff
@@ -155,6 +157,44 @@
...
+    });
+
+    document.addEventListener('click', function(e) {
+      if (!moduleTabMoreMenu.contains(e.target) && e.target !== moduleTabMoreBtn) {
+        closeMoreMenu();
+      }
+    });
+
+    document.addEventListener('keydown', function(e) {
+      if (e.key === 'Escape') closeMoreMenu();
```
kaxil, line 193: If focus is inside the menu when Escape fires, setting `hidden` drops focus to `<body>`. A `moduleTabMoreBtn.focus()` inside `closeMoreMenu` keeps it on the trigger. Related: `role="menu"` / `role="menuitem"` implies arrow-key navigation, which isn't wired up, so dropping the roles may be the easier promise to keep.

dev/registry/registry_tools/types.py
```diff
@@ -139,11 +181,27 @@
...
+    "queues": "queue",
```
kaxil, line 203: Worth a follow-up outside this PR: `provider.yaml.schema.json` declares `queues` items as `{name, message-queue-class}`, the same shape as `plugins`, but all six providers that set it list a bare class path, which is what this treats them as. The `items` block has no `type` key, so neither shape is actually validated. Since this is the first consumer of the section, tightening it to `items: {type: string}` would stop a future provider following the documented dict shape from being silently dropped by the `isinstance(class_path, str)` guard in extract_parameters.py.

dev/registry/tests/test_extract_parameters.py
```diff
@@ -1086,3 +1157,31 @@ def test_extra_whitespace_is_tolerated(self):
...
+                "extract_parameters.ThisAttrDoesNotExist",
+                ("extract_parameters", "ThisAttrDoesNotExist", None),
+            ),
+            (
+                "extract_parameters.get_category",
+                ("extract_parameters", "get_category", get_category),
+            ),
+        ),
+    )
+    def test__resolve_dottted_path(self, class_path: str, expected: tuple[str, str, object]):
```
uranusjr:
```suggestion
    def test__resolve_dotted_path(self, class_path: str, expected: tuple[str, str, object]):
```

dev/registry/registry_tools/types.py
```diff
@@ -147,11 +189,39 @@
...
+}
+
+# Maps yaml section key -> (type id, class-path field name) for dict-shaped
+# class-level sections (each yaml entry is a dict, not a bare class-path
+# string), shared by extract_versions.py and extract_parameters.py so the
+# field name is defined exactly once.
+DICT_SHAPED_CLASS_LEVEL_SECTIONS: dict[str, tuple[str, str]] = {
+    "plugins": ("plugin", "plugin-class"),
+    "dialects": ("dialect", "dialect-class-name"),
+}
```
uranusjr, line 216: I think this can have a third component `name` and `dialect-type` instead, so `discover_classes_from_provider` only needs one loop instead of two.

dev/registry/registry_tools/types.py
```diff
@@ -116,6 +116,48 @@
...
+        "suffixes": [],
+        "label": "DB Managers",
+        "icon": "D",
+    },
+    "dialect": {
+        "yaml_key": "dialects",
+        "level": "flat",
+        "suffixes": [],
+        "label": "Dialects",
+        "icon": "C",
```
uranusjr: I wonder if we want to choose a different letter since C might be useful for connection. Or maybe db_manager can use M instead? Not sure.

dev/registry/extract_versions.py
Lee-W: `extract_versions.py` now walks `DICT_SHAPED_CLASS_LEVEL_SECTIONS` alongside `CLASS_LEVEL_SECTIONS`, so plugins and dialects show up on backfilled version pages as well.

The test is parametrized over that dict as well, so a new dict-shaped section will be covered.

registry/src/_data/types.json
Lee-W: Added the missing `.type-icon`, `.share-bar` and `.provider-card .modules` rules for the new ids, plus `toolset` and `retry_policy` which had the same gap. They use `rgb(from var(--color-X) r g b / 0.2)` like the existing `.tab-icon` block rather than hardcoded colours.

registry/src/js/provider-detail.js
Lee-W: "More" button is now `active` while the selected type is in the overflow menu, so it's still visible what the list is filtered by.

registry/src/js/provider-detail.js
Lee-W: `closeMoreMenu` now returns focus to the trigger when focus was inside the menu. I also dropped `role="menu"` / `role="menuitem"`.

dev/registry/registry_tools/types.py
Lee-W: Yep, will create a follow up PR for this.

dev/registry/tests/test_extract_parameters.py
Lee-W: fixed

dev/registry/registry_tools/types.py
Lee-W: sounds good, just updated it

dev/registry/registry_tools/types.py
Lee-W: Making 
* `dialect` -> `D`
* `db_manager` -> `M`

so `C` is left for connection.

Thought of whether we have better combinations, but this seems to be the best we can get

registry/src/_data/types.json
```diff
@@ -63,5 +63,35 @@
...
+  },
+  {
+    "id": "auth_manager",
+    "label": "Auth Managers",
+    "icon": "A"
+  },
+  {
+    "id": "db_manager",
+    "label": "DB Managers",
+    "icon": "M"
```
kaxil, line 90: The M here doesn't reach the module cards. `provider-version.njk:374` renders the card chip as `{{ module.type[0] | upper }}` instead of reading this file, so `db_manager` cards show D, colliding with `decorator` and `dialect` -- the collision `ff18815f` was written to avoid. Six of the 19 ids disagree with their curated letter that way (`transfer` X->T, `secret` K->S, `decorator` @->D, `toolset` U->T, `extra_link` I->E, `db_manager` M->D).

The same block also misses colour: `.provider-detail-page .modules .module .icon.<id>` covers 6 of 19 ids and the base `.icon` rule sets no background, so those cards get a transparent chip. Both halves already affect `secret`, `decorator` and `toolset`, so a follow-up is fine, but a small `typeIcon` lookup filter beside `sortTypesByCount` would let the card read from here the way the tabs already do.

dev/registry/tests/test_types.py
```diff
@@ -93,6 +95,14 @@ def test_class_level_sections_are_subset_of_flat(self):
             assert yaml_key in FLAT_LEVEL_SECTIONS
             assert FLAT_LEVEL_SECTIONS[yaml_key] == type_id
 
+    def test_class_level_category_overrides_are_subset_of_class_level_sections(self):
+        assert set(CLASS_LEVEL_CATEGORY_OVERRIDES.keys()) <= set(CLASS_LEVEL_SECTIONS.keys())
+
+    def test_dict_shaped_class_level_sections_are_subset_of_flat(self):
```
kaxil, line 101: This and the sibling above check the subset direction, but the failure mode from last round was the other one: a flat section that neither table consumes gets silently skipped by both extractors, and the `FQCN_DESC_SUFFIXES[mod_type]` KeyError can't catch it, since it only fires for keys already in a table. `set(FLAT_LEVEL_SECTIONS)` is exactly `set(CLASS_LEVEL_SECTIONS) | set(DICT_SHAPED_CLASS_LEVEL_SECTIONS) | {"transfers", "task-decorators"}` today, so asserting that equality passes now and fails loudly next time.

kaxil (approved):

registry/src/_data/types.json
Lee-W: Both fixed. A `typeIcon` filter next to `sortTypesByCount` gives the card its curated letter, and the card colour rules now cover all 19 ids instead of 6.

It now shows M for db_mangers

image

dev/registry/tests/test_types.py
Lee-W: Added, with `transfers` and `task-decorators` as a named exemption set so the reason they are excluded sits next to the assertion. Dropping a key from `CLASS_LEVEL_SECTIONS` now failed the test.

PR #69159, opened 2026-06-30
Improve dark mode switch contrast (#66759)

deepujain:
**Title:** Improve dark mode switch contrast (#66759)

## Summary
Unchecked switches can blend into dark surfaces, which makes the Dag pause toggle difficult to distinguish in dark mode. This adds a token-based dark-mode border to the shared switch control while keeping the light-mode off state transparent and the checked state on the existing color palette border.

## Changes
- **`airflow-core/src/airflow/ui/src/theme.ts`** -- add an emphasized border to the shared Chakra switch recipe so unchecked switches remain visible in dark mode.

## Screenshots / visual proof
Before, using the upstream `src/theme.ts` on a local Airflow 3 instance:

Before: Airflow DAG list dark mode

After, using this PR on the same local Airflow route:

After: Airflow DAG list dark mode

Close crop of the affected switch state:

PR 69159 switch crop comparison

## Test plan
- [x] `pnpm install --ignore-scripts`
- [x] `pnpm exec eslint src/theme.ts`
- [x] `pnpm exec tsc --p tsconfig.app.json --pretty false`
- [x] `pnpm build`
- [x] CI passes

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

Fixes #66759

PR #67406, opened 2026-05-24
Improve DAG toggle contrast in dark mode

bobu-putheeckal:
## What

Improve dark-mode switch contrast by adding a dark-mode border to the shared Chakra switch control. The checked state keeps using the existing color palette token, while the unchecked/off state gains a visible outline on dark surfaces.

This addresses the DAG pause toggle contrast reported in #66759.

## Why

In dark mode, the unchecked switch track can blend into the surrounding dark background, making paused/off DAG toggles hard to distinguish. A token-based border makes the switch shape clearer without introducing new colors or changing light-mode behavior.

## Tests

- `pnpm exec eslint src/theme.ts`
- `pnpm exec tsc --p tsconfig.app.json --pretty false`
- `pnpm build`

## Screenshots

I could not capture a full Airflow DAG-list screenshot from the raw Vite dev shell because that shell depends on Airflow backend-provided base/config values to render the app. The change is limited to the theme recipe and was verified with the focused static checks and production build above.

---

closes: #66759

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — OpenAI Codex

Generated-by: OpenAI Codex following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70203, opened 2026-07-21
UI: Make the Dag pause toggle distinguishable in dark mode

### LINKED ISSUE #66759: Dag toggle is very hard to distinguise in dark mode

Airflow Core

### Apache Airflow version

irflow 3.1.8+astro.1

It is very hard to distinguish between DAG that is toggled on and off in dark mode 
￼
￼

Image

Image

_No response_

### Operating System

_No response_

### Deployment

Astronomer

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

eddiesr93:
In dark mode the Dag pause toggle is hard to make out: its off-state track uses
`bg.emphasized`, which in this palette lands within 0.05 oklch lightness of the page
background (0.08 in light mode), so only the white thumb reads and the pill has no
visible edge.

This gives the switch control a 1px inset ring so the shape is legible in both modes.

### Why an inset shadow and not a border

Two earlier attempts, #67406 and #69159, added a `border` to the control and were closed
after review found the thumb was no longer centred and the horizontal layout shifted.
That is unavoidable with a border: the thumb's size and its checked travel are both
derived from `--switch-height` / `--switch-width`, so any box-model space the control
takes shrinks the content box the thumb sits in while its translate distance stays the
same. `box-shadow: inset` costs no layout space.

The thumb also stays concentric with the track's end cap — both are centred on (8, 8),
with radii 6.4 and 8. Measured in a running UI:

| | control | thumb | thumb gap T / B / leading |
|---|---|---|---|
| `main` | 32×16 | 12.8×12.8 | 1.60 / 1.60 / 1.60 |
| this PR | 32×16 | 12.8×12.8 | 1.60 / 1.60 / 1.60 |

### Before / after

Dags list, `breeze start-airflow --dev-mode`.

**Dark mode** — before, then after:

dark mode before

dark mode after

**Light mode.** The same problem is milder but present: against a striped row the
off-state track sits about 0.04 oklch from the row behind it, so the pill loses its edge
there too. That is why the ring is applied in both modes rather than only in dark.

light mode before

light mode after

closes: #66759

---

---

Drafted-by: Claude Code (Opus 4.8); reviewed by @eddiesr93 before posting

### DIFF
--- a/airflow-core/src/airflow/ui/src/theme.ts
+++ b/airflow-core/src/airflow/ui/src/theme.ts
@@ -382,6 +382,12 @@ const defaultAirflowTheme: ThemingConfig = {
     },
     switch: {
       slots: [],
+      base: {
+        control: {
+          shadow: "inset 0 0 0px 1px var(--shadow-color)",
+          shadowColor: { _dark: "gray.500", _light: "gray.400" },
+        },
+      },
       defaultVariants: { size: "sm" } as Record<string, string>,
     },
     // size="sm" gives px/py:2 on cell and columnHeader vs px/py:3 for "md".


dheerajturaga (approved):
I agree in dark more this is not easy to make out. However Id leave the decision to merge to the UI maintainers

Shrividya (approved):

choo121600 (approved):
Looks good! It was a bit hard to see in dark mode ;)
Thanks for improving it!

PR #70205, opened 2026-07-21
Allow separate object and pod labels for chart components

### LINKED ISSUE #53190: dagProcessor in helm chart doesn't have a labels section
### Description

All other parts of the helm chart have a labels section specific to the workload except for the dagProcessor. Can one be added please?

### Use case/motivation

I'd like the dagProcessor to have labels

### Related issues

_No response_

### Code of Conduct

vyncint:
Add `podLabels` so workload object labels and pod labels can be configured separately, for `apiServer`, `scheduler`, `workers.celery`, `triggerer`, `dagProcessor`, `flower`, `statsd`, `pgbouncer`, `redis` and `otelCollector`.

When `podLabels` is unset, existing behavior remains unchanged: `<component>.labels` stays on the pods, while the Deployment or StatefulSet receives only global labels. Setting `podLabels` (including `{}`) opts into separation: `<component>.labels` applies to the workload object and `<component>.podLabels` applies to pods. Component values retain precedence over global labels.

`otelCollector` is the one exception: its Deployment already merged component labels onto the object, so only its pod side gains the new knob and its object labels are unchanged.

closes: #53190

Behaviour matrix, verified by rendering every affected workload:

| `podLabels` | workload object labels | pod labels |
| --- | --- | --- |
| unset (default `~`) | global only — unchanged | component + global |
| set | component + global | podLabels + global |
| `{}` | component + global | global only |

Tests:

- `pytest helm_tests/airflow_aux/test_pod_labels.py helm_tests/dagprocessor` — 38 passed. The new module parametrizes all three states over every component; `dagProcessor` keeps its existing dedicated module.
- `pytest helm_tests/{dagprocessor,statsd,otel_collector,redis} helm_tests/airflow_core/test_{api_server,scheduler,triggerer,worker}.py helm_tests/other/test_{flower,pgbouncer}.py` — 779 passed, 1 failed.
- `pytest helm_tests/airflow_aux/test_{chart_quality,basic_helm_chart,annotations}.py` — 114 passed, 9 failed.
- Backward compatibility: `helm template` with every component's `labels` set and `podLabels` unset is **byte-identical to main across 3638 lines** (random secrets pinned so the render is deterministic).

The 10 local failures are pre-existing and unrelated: they assert on Helm's JSON-schema error wording (`value must be one of 'X'`) while local Helm 3.14.4 emits `must be one of the following: "X"`, and they cover `dags.persistence.accessMode`, `*.pullPolicy` and `apiServer.securityContext`. Re-running the same modules on a clean checkout reproduces exactly the same counts, so this change adds no new failures.

---

- [X] Yes — Codex (GPT-5) for the initial `dagProcessor` commit, Claude Opus 5 for this revision

Generated-by: Codex (GPT-5), Claude Opus 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/chart/newsfragments/53190.feature.rst
+++ b/chart/newsfragments/53190.feature.rst
@@ -1,0 +1,1 @@
+Allow Dag processor pod labels to be configured separately with ``dagProcessor.podLabels``.
--- a/chart/newsfragments/53190.feature.rst
+++ b/chart/newsfragments/53190.feature.rst
@@ -1,1 +1,1 @@
-Allow Dag processor pod labels to be configured separately with ``dagProcessor.podLabels``.
+Allow pod labels to be configured separately from object labels with ``podLabels``, for ``apiServer``, ``scheduler``, ``workers.celery``, ``triggerer``, ``dagProcessor``, ``flower``, ``statsd``, ``pgbouncer``, ``redis`` and ``otelCollector``. Leaving ``podLabels`` unset keeps the existing behaviour.
--- a/chart/templates/api-server/api-server-deployment.yaml
+++ b/chart/templates/api-server/api-server-deployment.yaml
@@ -30,6 +30,12 @@
 {{- $containerSecurityContext := include "containerSecurityContext" (list .Values.apiServer .Values) }}
 {{- $containerSecurityContextWaitForMigrations := include "containerSecurityContext" (list .Values.apiServer.waitForMigrations .Values) }}
 {{- $containerLifecycleHooks := or .Values.apiServer.containerLifecycleHooks .Values.containerLifecycleHooks }}
+{{- $deploymentLabels := .Values.labels }}
+{{- $podLabels := .Values.apiServer.labels }}
+{{- if ne .Values.apiServer.podLabels nil }}
+  {{- $deploymentLabels = mustMerge (deepCopy .Values.apiServer.labels) .Values.labels }}
+  {{- $podLabels = .Values.apiServer.podLabels }}
+{{- end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -40,7 +46,7 @@ metadata:
     release: {{ .Release.Name }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     heritage: {{ .Release.Service }}
-    {{- with .Values.labels }}
+    {{- with $deploymentLabels }}
       {{- toYaml . | nindent 4 }}
     {{- end }}
   {{- if .Values.apiServer.annotations }}
@@ -80,8 +86,8 @@ spec:
         tier: airflow
         component: api-server
         release: {{ .Release.Name }}
-        {{- if or .Values.labels .Values.apiServer.labels }}
-          {{- mustMerge .Values.apiServer.labels .Values.labels | toYaml | nindent 8 }}
+        {{- if or .Values.labels $podLabels }}
+          {{- mustMerge (deepCopy $podLabels) .Values.labels | toYaml | nindent 8 }}
         {{- end }}
       annotations:
         checksum/metadata-secret: {{ include (print $.Template.BasePath "/secrets/metadata-connection-secret.yaml") . | sha256sum }}
--- a/chart/templates/dag-processor/dag-processor-deployment.yaml
+++ b/chart/templates/dag-processor/dag-processor-deployment.yaml
@@ -31,6 +31,12 @@
 {{- $containerSecurityContextLogGroomerSidecar := include "containerSecurityContext" (list .Values.dagProcessor.logGroomerSidecar .Values) }}
 {{- $containerSecurityContextWaitForMigrations := include "containerSecurityContext" (list .Values.dagProcessor.waitForMigrations .Values) }}
 {{- $containerLifecycleHooks := or .Values.dagProcessor.containerLifecycleHooks .Values.containerLifecycleHooks }}
+{{- $deploymentLabels := .Values.labels }}
+{{- $podLabels := .Values.dagProcessor.labels }}
+{{- if ne .Values.dagProcessor.podLabels nil }}
+  {{- $deploymentLabels = mustMerge (deepCopy .Values.dagProcessor.labels) .Values.labels }}
+  {{- $podLabels = .Values.dagProcessor.podLabels }}
+{{- end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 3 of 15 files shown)


Miretpl:
Could you please check discussions from previous proposed PRs for the relevant change https://github.com/apache/airflow/pull/59290?

Miretpl (changes requested):
Requesting changes to block accidential merge

vyncint:
Thanks @Miretpl — I read through #59290 and understand the hold now. That PR made the same change for the webserver and stalled on a design question that was never settled, so rather than re-land it for dag-processor I'd like a direction first.

Today per-component labels (e.g. dagProcessor.labels) only land on the pod template (spec.template.metadata.labels); the Deployment's own metadata.labels gets global labels only. Two ways to close that gap:

A) Merge component labels onto the Deployment metadata directly (what this PR does now). Simple, but it changes where existing users' labels land, so it's a backward-compatibility concern.

B) Add a per-component podLabels section (like the existing annotations / podAnnotations split): labels go to the Deployment, podLabels to the pod, and when podLabels is unset, labels fall back to the pod level to preserve today's behavior. This is the approach Miretpl proposed in #59290.

On scope, #59290 also concluded this should be applied across all components rather than one at a time, and target apiServer rather than webserver (webserver is going away with Airflow 2.x). Happy to broaden this PR accordingly.

Could a committer confirm (1) approach A or B, and (2) whether you want the full multi-component change in one PR? I'll rework and rebase onto main once there's a direction. cc @jscheffl @potiuk @jedcunningham @romsharon98

Miretpl:
My idea https://github.com/apache/airflow/pull/59290#issuecomment-3667433158 after becoming a Committer (I wasn't back then) is still the same. Separation, like in annotations, gives more flexibility, and I see cases when it would be nice to separate labels between the pod and the deployment itself. And if there will be a user who will want to have the same labels, there will be a possibility to set the same value on both levels.

vyncint:
Thanks @Miretpl — agreed, the separation is the cleaner model. I'll go with podLabels:

- labels -> Deployment metadata.labels
- podLabels -> pod spec.template.metadata.labels
- backward compatible: when podLabels is unset, labels stay on the pod level (today's behavior) and the Deployment keeps only global labels.

I'll implement it on dag-processor here first as the reference pattern (values.yaml + template + tests), and once you're happy with it extend the same pattern to the other components, using apiServer rather than webserver. Prefer everything in one PR, or dag-processor first then follow-ups?

Miretpl:
> I'll implement it on dag-processor here first as the reference pattern (values.yaml + template + tests), and once you're happy with it extend the same pattern to the other components, using apiServer rather than webserver. Prefer everything in one PR, or dag-processor first then follow-ups?

In general small PRs are better, but whole thing will rather not be big, so I will adjust. Do as it will be more comfortable for you.

> using apiServer rather than webserver

The Webserver will also be needed, but for the 1.2x line (on main we have 2.0) only.

vyncint:
Implemented the `podLabels` split requested in the review and earlier discussion:

- Leaving `dagProcessor.podLabels` unset preserves the existing behavior: `dagProcessor.labels` applies to pods, while the Deployment receives global labels only.
- Setting `dagProcessor.podLabels` (including `{}`) opts into separation: `dagProcessor.labels` applies to the Deployment and `dagProcessor.podLabels` applies to pods.
- Component labels keep precedence over global labels, and `deepCopy` prevents `mustMerge` from mutating the source values.

I also added values/schema documentation, a news fragment, and regression coverage for the fallback, separated-label, collision, and explicit-empty-map cases. The branch is rebased on current `main`.

Validation completed:

- `breeze testing helm-tests --use-xdist` — 2095 passed
- `breeze ci selective-check --commit-ref HEAD`

---

vyncint:
Verified on the rebased branch:

- `podLabels` unset — unchanged behavior (Deployment: global only, pods: component + global)
- `podLabels` set — Deployment: component + global, pods: podLabels + global
- `podLabels: {}` — pods get global only
- collisions — component wins on the Deployment, podLabels wins on pods

Backward compatibility: rendering the whole chart with the same values against current `main` and against this branch gives byte-identical output.

Tests: 7 passed in `test_labels_deployment.py`, 125 passed across the dagprocessor, redis, statsd and otel_collector helm suites.

Ready for review @Miretpl.

vyncint:
@Miretpl extended the pattern to the remaining components in 177dc55: apiServer, scheduler, workers.celery, triggerer, flower, statsd, pgbouncer, redis and otelCollector. Title and description updated to match the wider scope.

otelCollector turned out to be the one exception. Its Deployment already merged component labels onto the object, so there is no backward-compatibility branch to preserve there — only its pod side gains the knob, and its object labels are unchanged. Every other component follows the dagProcessor shape exactly.

Verification:

- helm template with every component's labels set and podLabels unset is byte-identical to main across 3638 lines, with the random secrets pinned so the render is deterministic. That is the backward-compatibility proof.
- New parametrized module helm_tests/airflow_aux/test_pod_labels.py covers all three states (unset, set, {}) for every component: 38 passed together with the existing dagProcessor module. I put the new cases in one parametrized module rather than nine near-identical files, following test_annotations.py; happy to split them per component if you prefer that layout.
- Component suites for statsd, otel_collector, redis, api_server, scheduler, triggerer, worker, flower and pgbouncer: 779 passed, 1 failed.
- Chart quality, basic chart and annotations modules: 114 passed, 9 failed.

The 10 failures are pre-existing and unrelated. They assert on Helm's JSON-schema error wording, and local Helm 3.14.4 phrases it differently from what the tests expect; they cover dags.persistence.accessMode, pullPolicy and apiServer.securityContext. Running the same modules on a clean checkout of the base gives exactly the same counts, so nothing here is new. CI runs a pinned Helm, so I expect them green there.

vyncint:
Hi @Miretpl,

Following up on this. The podLabels split you asked for is in, and I extended it to the remaining components in 177dc55: apiServer, scheduler, workers.celery, triggerer, flower, statsd, pgbouncer, redis and otelCollector.

Could you take another look when you have a moment? The changes-requested review is still blocking merge.

Miretpl:
@vyncint Most of the people here are doing work voluntarily, so getting feedback may take some time, especially in the AI era where it is easy to generate a PR, but still people have to take a look at it if it is aligned with the project. Sooner or later, I or some other maintainer will take another look at the PR and give another round of feedback. Please do not mention people when the PR is stale for a day or two. If it is stale for around 2 weeks, I think it is a good time period to try to ping any maintainer for the review.

PR #69930, opened 2026-07-15
Add dedicated Glue crawler lifecycle operators

AlejandroMorgante:
Split the AWS Glue crawler lifecycle into explicit create, update, run, and delete operators while keeping the existing `GlueCrawlerOperator` backward compatible during its deprecation period.

## Motivation

`GlueCrawlerOperator` currently combines resource provisioning, configuration changes, and execution. This means a Dag intended only to run an existing crawler can also create or modify it. Operation-specific operators make each task's intent and required AWS permissions explicit.

## Operators

- `GlueCrawlerCreateOperator` creates a crawler from its Boto3 configuration.
- `GlueCrawlerUpdateOperator` updates an existing crawler without starting it.
- `GlueCrawlerRunOperator` starts an existing crawler and supports synchronous or deferrable completion waits through the existing `GlueCrawlerCompleteTrigger`.
- `GlueCrawlerDeleteOperator` deletes an existing crawler through the Boto3 client exposed by `GlueCrawlerHook`.

The existing `GlueCrawlerOperator` remains available for compatibility and emits an `AirflowProviderDeprecationWarning` directing new Dags to the operation-specific operators.

## System test

The Glue system test now validates the complete lifecycle:

1. Create the crawler.
2. Update its description and verify the change through Boto3.
3. Run it with `deferrable=True`.
4. Run it with `deferrable=False`.
5. Delete it during teardown with the new delete operator.

## Validation

Validated end-to-end against a real AWS environment. The system test created, updated, ran, and deleted the Glue crawler, then confirmed that the crawler, Glue database, Glue job, and S3 bucket no longer existed.

The system test can be reproduced from the Airflow checkout with `AWS_PROFILE`, `AWS_REGION`, and `ROLE_ARN` configured in `files/airflow-breeze-config/environment_variables.env`:

```bash
SYSTEM_TESTS_ENV_ID=<unique-id> \
breeze testing system-tests \
  --forward-credentials \
  --test-timeout 240 \
  providers/amazon/tests/system/amazon/aws/example_glue.py \
  -q
```

Latest result: `1 passed in 191.86s`. The focused Glue crawler operator and trigger tests also pass: `66 passed`, with 100% coverage of the modified operator module. Provider mypy, regular and manual prek checks, and the Amazon provider documentation build also pass.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Codex (GPT-5)

[...]

PR #70206, opened 2026-07-22
Remove hard-coded deferrable crawler run from example_glue system test

seanghaeli:
The `run_crawler_deferrable` task added in #69930 hard-codes `deferrable=True`, which hangs forever under `dag.test(use_executor=True)` (no triggerer runs in that path), timing out executor-based system test harnesses. The non-deferrable `run_crawler` task covers the operator; deferrable behavior is exercised by unit tests and the harness-level deferrable mode (`AIRFLOW__OPERATORS__DEFAULT_DEFERRABLE`).


providers/amazon/tests/system/amazon/aws/example_glue.py
AlejandroMorgante: Could we also remove the explicit `deferrable=False` from run_crawler?

AlejandroMorgante (commented):
Also, if we remove **run_crawler_deferrable**, we should update `providers/amazon/docs/operators/glue.rst`, since it still references the removed `howto_operator_glue_crawler_run_deferrable` markers through an exampleinclude.

o-nikolas (approved):

PR #70207, opened 2026-07-22
Preserve pinned version_data when a Dag version is still in use

o-nikolas:
## Problem
  
  When a Dag file's *behavior* changes but its *structure* doesn't (the very common case — e.g.
  editing a constant inside a `@task` function body, which is not serialized), `write_dag`
  does not create a new `DagVersion`. Instead it refreshes the latest row's bundle pointer
  **in place**:
  
      dag_version.bundle_version = bundle_version   # overwrite
      dag_version.version_data   = version_data     # overwrite
  
  For Git bundles this is harmless: a pinned run resolves its code from
  `dag_run.bundle_version` (a commit SHA, recorded on the run itself and never overwritten),
  and git's immutable object store can always check that SHA out.
  
  For bundles that rely on `version_data` (the json blob which represents the
    state of the code for that bundle version; for S3 this is a dictionary
    manifest containing the S3 version for each file in the bucket) the
  overwrite is destructive. The manifest stored in `DagVersion.version_data` is the **only**
  way to reconstruct the pinned code.
  
  Taking the upcoming S3 Bundle as an example: when the bucket content changes mid-run
 (for a change that doesn't modify the dag structure itself), the parse overwrites the row and the pinned run's
  manifest is gone:
  
      BEFORE (run pinned to v1)          AFTER re-parse of v2 (hash unchanged)
      ─────────────────────────          ─────────────────────────────────────
      DagRun.bundle_version: v1          DagRun.bundle_version: v1   (still!)
      DagVersion row:                    DagVersion row (same row):
        bundle_version: v1                 bundle_version: v2   ◄─ overwritten
        version_data:   manifest_v1        version_data:   manifest_v2 ◄─ v1 manifest DESTROYED
  
  A second, pre-existing symptom of the same overwrite (which in this case affects all versioned
  Bundles, including Git): triggering a run
  against a specific `bundle_version` resolves the serialized code via a `DagVersion` lookup keyed on
  `bundle_version`, so when the DagVersion is overwritten the `bundle_version`s are lost, `DagVersionNotFound`
  are raised. This will be possible to do now with these changes.
  
  ## Fix
  
   Now, when version_data is involved (e.g. a Bundle is being used which
    needs to store metadata in the DagVersion row AND versioning is
    enabled for that Bundle AND the DagVersion has active tasks) a new
    DagVersion is created instead of overwriting (mirroring the existing
    path we take when the dag has a structural change). The old row keeps
    serving pinned runs; the new row becomes the latest for new runs.
    Bundles without version_data (e.g. Git) keep the cheap in-place
    pointer refresh.

  
      BEFORE (run pinned to v1)          AFTER re-parse of v2 (hash unchanged)
      ─────────────────────────          ─────────────────────────────────────
      DagRun.bundle_version: v1          DagRun.bundle_version: v1
      DagVersion vN:                     DagVersion vN (untouched):
        bundle_version: v1                 bundle_version: v1
        version_data:   manifest_v1        version_data:   manifest_v1  ◄─ pinned runs keep it
                                         DagVersion vN+1 (new, latest):
                                           bundle_version: v2           ◄─ new runs inherit it
                                           version_data:   manifest_v2
  
  **Once a `(bundle_version, version_data)` pairing has been observed by task instances, it is
  effectively immutable.**
  
  Safety measures:
  
  - **Bundles without `version_data` (Git, Local): no change.** Either their `bundle_version` alone
    recovers the code or they don't support versioning at all. So the in-place DagVersion refresh
    behaviour remains.
  - **No task instances on the current version: no change.** Nothing is pinned to it, so the
    metadata (including `version_data`) is still refreshed in place.
  - New rows are created only when bucket content actually changes. Bucket updates are a deliberate
    action by the user and will not be as frequent as repository or local FS changes.
  - Rows are garbage collected by the existing `db_cleanup` handling for `dag_version`. 

 

---

---

### DIFF
--- a/airflow-core/src/airflow/models/serialized_dag.py
+++ b/airflow-core/src/airflow/models/serialized_dag.py
@@ -595,6 +595,21 @@ class SerializedDagModel(Base):
         }
 
     @classmethod
+    def _has_task_instances(cls, dag_id: str, dag_version_id: UUID, *, session: Session) -> bool:
+        """Check whether any task instance references the given DagVersion."""
+        return bool(
+            session.scalar(
+                select(
+                    exists().where(
+                        # Using dag_id filter to speed up query via the composite index.
+                        TaskInstance.dag_id == dag_id,
+                        TaskInstance.dag_version_id == dag_version_id,
+                    )
+                )
+            )
+        )
+
+    @classmethod
     @provide_session
     def write_dag(
         cls,
@@ -688,38 +703,44 @@ class SerializedDagModel(Base):
 
         new_dag_hash = cls.hash(dag.data)
 
+        preserve_pinned_version_data = False
         if serialized_dag_hash == new_dag_hash and dag_version and dag_version.bundle_name == bundle_name:
-            # Serialized content is unchanged, so we don't create a new DagVersion.
-            # But if the bundle advanced, refresh the latest version's pointer in place — tasks resolve
+            # Serialized content is unchanged, so we normally don't create a new DagVersion.
+            # If the bundle advanced, refresh the latest version's pointer in place — tasks resolve
             # their code from ``ti.dag_version.bundle_version`` at run time, so a stale
             # pointer makes runs execute an outdated commit.
             bundle_metadata_changed = (
                 dag_version.bundle_version != bundle_version or dag_version.version_data != version_data
             )
-            if bundle_metadata_changed:
-                dag_version.bundle_version = bundle_version
-                dag_version.version_data = version_data
-                session.merge(dag_version)
-                DagCode.update_source_code(dag_id=dag.dag_id, fileloc=dag.fileloc, session=session)
-            if name_updated or bundle_metadata_changed:
-                # A write occurred — a deadline alert name update and/or a bundle
-                # metadata refresh — so report True so callers know the DB changed.
-                return True
-            log.debug("Serialized DAG (%s) is unchanged. Skipping writing to DB", dag.dag_id)
-            return False
+            # The in-place refresh is safe only for bundles whose ``bundle_version`` alone can recover the
+            # code (e.g. Git: the commit SHA resolves against the immutable object store). When
+            # ``version_data`` is involved (e.g. For bundles like the S3 bundle a json manifest of S3 object
+            # versions), this row holds the only copy of the data which is needed for runs pinned to the old
+            # bundle_version to reconstruct their code. Overwriting it would break their reproducibility. In
+            # that case fall through and create a new DagVersion instead: the old row keeps serving pinned
+            # runs, the new row becomes the latest for new runs.
+            preserve_pinned_version_data = (
+                bundle_metadata_changed
+                and (dag_version.version_data is not None or version_data is not None)
+                and cls._has_task_instances(dag.dag_id, dag_version.id, session=session)
+            )
+            if not preserve_pinned_version_data:
+                if bundle_metadata_changed:
+                    dag_version.bundle_version = bundle_version
+                    dag_version.version_data = version_data
+                    session.merge(dag_version)
+                    DagCode.update_source_code(dag_id=dag.dag_id, fileloc=dag.fileloc, session=session)
+                if name_updated or bundle_metadata_changed:
+                    # A write occurred — a deadline alert name update and/or a bundle
+                    # metadata refresh — so report True so callers know the DB changed.
+                    return True
+                log.debug("Serialized DAG (%s) is unchanged. Skipping writing to DB", dag.dag_id)
+                return False
 
         has_task_instances: bool = False
         if dag_version:
-            has_task_instances = bool(
-                session.scalar(
-                    select(
-                        exists().where(
-                            # Using dag_id filter to speed up query via the composite index.
-                            TaskInstance.dag_id == dag.dag_id,
-                            TaskInstance.dag_version_id == dag_version.id,
-                        )
-                    )
-                )
+            has_task_instances = preserve_pinned_version_data or cls._has_task_instances(
+                dag.dag_id, dag_version.id, session=session
             )
 
         if dag_version and not has_task_instances:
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -688,38 +703,44 @@ def write_dag(
...
-                # metadata refresh — so report True so callers know the DB changed.
-                return True
-            log.debug("Serialized DAG (%s) is unchanged. Skipping writing to DB", dag.dag_id)
-            return False
+            # The in-place refresh is safe only for bundles whose ``bundle_version`` alone can recover the
+            # code (e.g. Git: the commit SHA resolves against the immutable object store). When
+            # ``version_data`` is involved (e.g. For bundles like the S3 bundle a json manifest of S3 object
+            # versions), this row holds the only copy of the data which is needed for runs pinned to the old
+            # bundle_version to reconstruct their code. Overwriting it would break their reproducibility. In
+            # that case fall through and create a new DagVersion instead: the old row keeps serving pinned
```
jedcunningham, line 720: I worry that creating new dagversions for _every_ single dag in the bundle when any file changes is going to get confusing/wasteful fast, especially at scale.

I think we should explore other designs to keep the "dagversion changes when we observe this specific dag changed" property. Maybe an optional bundleversion table to allow (bundle_name, version)->version_data mapping or something? Given the size, I also don't love shoving version_data down to each dagrun either - that has problems on a different axis.

o-nikolas: Hey @jedcunningham , thanks for having a look!

> I worry that creating new dagversions for every single dag in the bundle when any file changes is going to get confusing/wasteful fast, especially at scale.

I'm not sure I follow this. New DagVersions are only created if 1) the dag bundle being used needs data storage 2) bundle versioning is enabled for that bundle 3) the metadata has actually changed for that dag file 4) and that dag has currently running task instances (which need the data persisted).

Specifically 3 and 4 above above means that not _every_ single dag in the bundle will get a new version. The aperture for when new DagVersions would be created, beyond which are already created today, is still quite small. Not to mention S3 bucket deployments are a very intentional action by users, and I think a possible chance that a subset of dags will mint a single record each with a deployment to the bucket is not that much of an issue either, honestly. With documentation and existing garbage collection tooling, I think it is not such a crisis.

> Maybe an optional bundleversion table to allow (bundle_name, version)->version_data mapping or something

This was proposed initially and vetoed by the community. Airflow as a community is incredibly sensitive about adding new schema, never mind new tables (the amount of hoops we're having to jump through in multi-team land because of this would make your head spin). We also lose the FK graceful cleanup that we get by colocating the version_data along with the DagVersion it belongs to.

>  I also don't love shoving version_data down to each dagrun either - that has problems on a different axis.

I agree, and also considered this, it is probably the second least worst option IMHO, behind the one I propose in this PR. Which again for the reasons at the top of this comment (and in the PR description) I think have actually a fairly small blast radius and is something a user can decide if they would like to take on when choosing to enable versioning on the S3 bundle. 

Let me know what you think with all that in mind!

ephraimbuddy: Thanks for the detailed writeup here @o-nikolas, the problem is real and well explained. I replied on the dev list thread where the storage question was first discussed, with some context on what was declined there and an experiment implementing the external path. https://lists.apache.org/thread/0z6qn3wo4qqv2fylj94f6bhrbt7p3g4n
Happy to continue the direction discussion there.

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -688,38 +703,44 @@ def write_dag(
...
+            # runs, the new row becomes the latest for new runs.
+            preserve_pinned_version_data = (
+                bundle_metadata_changed
+                and (dag_version.version_data is not None or version_data is not None)
+                and cls._has_task_instances(dag.dag_id, dag_version.id, session=session)
+            )
+            if not preserve_pinned_version_data:
```
ephraimbuddy, line 727: This is where I disagree with point 3 from the list. This branch only runs when the serialized hash did not change. So the Dag itself did not change at all. What changed is version_data, and version_data covers the whole bucket. So one changed file anywhere in the bucket makes this true for every Dag in the bundle. All of them get a new DagVersion even though their own files were not touched. To keep the aperture small, the code would need to ask "did this Dag's own files change". The manifest is opaque here, so it cannot ask that.

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -594,6 +594,21 @@ def _prefetch_dag_write_metadata(
             for dag_id in dag_id_list
         }
 
+    @classmethod
+    def _has_task_instances(cls, dag_id: str, dag_version_id: UUID, *, session: Session) -> bool:
+        """Check whether any task instance references the given DagVersion."""
+        return bool(
+            session.scalar(
+                select(
+                    exists().where(
```
ephraimbuddy, line 603: This is point 4 from the list. There is no filter on task state here. Finished tasks count the same as running ones. So after the first run of the current version, this check is true. It stays true until db clean removes the old task instances. In practice it means "the Dag has ever run", not "the Dag has running tasks". Together with the other comment, every bucket deploy creates a new DagVersion, SerializedDagModel and DagCode row for every Dag that has ever run.

ephraimbuddy (commented):
Hey Niko, bringing the specifics here as you asked on the list. No rush at all, enjoy the vacation :)

You asked which of the five points I disagree with. It is 3 and 4. Details are inline, but the short version is, the changed branch only runs when the Dag itself did not change. The manifest covers the whole bucket, so a change to any file reaches every Dag in the bundle. And the task instance check has no state filter, so it means "has ever run", not "has running tasks". Put together, a bundle with 200 Dags deployed daily creates 600 new rows a day with the same content. Cleanup can only remove them after the old task instances themselves can be removed. Users already file issues about version growth like this. So I don't think the aperture is small for the users this feature is for.

On points 1, 2 and 5 I agree with you. I would only say that "unversioned users are unaffected" also means the cost lands fully on the users of the feature we are building.

This is why I lean toward keeping the manifest out of DagVersion. That can be the external store or the mapping table you ranked second. With either one, a bucket deploy never creates a DagVersion for a Dag that did not change.

Issue #68382, opened 2026-06-11
Connection port field does not validate that the value is a valid port number


amoghrajesh:
Airflow stores a `port` field on connections, but nowhere does it enforce that the value is a valid port number. A user can create a connection with `port = -1`, `port = 99999999`, or any other integer that is not a valid TCP/UDP port (0–65535), and Airflow will accept and persist it without complaint.

This is accepted through every entry point: the CLI, the public REST API, the execution API, and direct model construction. The value flows through to workers and is handed to provider hooks, which may silently misbehave or produce confusing errors far from the source of the bad data.

The connection port field should reject values outside the valid range at every layer where connections are accepted or created.

**Affected areas:**
- Core connection model (`airflow.models.connection`)
- Task SDK connection definition (`airflow.sdk.definitions.connection`)
- Public REST API connection request schema
- Execution API connection schema
- CLI connection create/update command
- JSON schema used for execution-time communication

**Why fix this:**
Port numbers have a well-defined valid range by the TCP/IP specification (0–65535). Accepting values outside this range produces connections that can never work, with no feedback to the user at the point of creation. Validation should be enforced at the boundary, not discovered at runtime inside a hook or provider.

ashb:
If port is optional then `0` should not be accepted either.

ashb:
And we probably need to think carefully about if we change the DB model, or even the response schema -- how should we handle existing data that might have now-invalid data in it?

amoghrajesh:
Thanks for adding that.

ch9tan-design:
Hi @amoghrajesh,

I'd like to work on this issue. Could you please assign it to me?

Thanks!

ashb:
@ch9tan-design Just submit a PR. But also before you work on it, how do you plan on tackling the complication I raised?

ch9tan-design:
Thanks for the guidance.

My initial thought is to identify all entry points where a connection port can be provided (Connection model, Task SDK, REST API, Execution API, CLI, and JSON schema) and determine whether there is already a shared validation layer that can be reused.

If possible, I would prefer to centralize the validation logic to avoid duplicating the port range check across multiple components. I'll investigate the existing validation patterns in these areas and submit a PR with the proposed approach.

Please let me know if there is a preferred validation layer I should follow.

ch9tan-design:
My current thought is to avoid enforcing validation in a way that breaks existing persisted data.

[...]

PR #70208, opened 2026-07-22
Fix Connection port field range validation (#68382)

dj-wise-ronin:
### Related Issue
Closes: #68382

### Description
Connection models in Apache Airflow accept integer values for the `port` field but do not enforce that the value is a valid network port in the range `[1, 65535]`.
This PR adds validation logic to connection ports at three layers:
1. **SQLAlchemy Connection model** validation via `@validates("port")` to ensure port numbers assigned are integers between 1 and 65535.
2. **Task SDK Connection model** validation in `__attrs_post_init__` to enforce range limits and format validation during standard instantiation and URI parsing.
3. **Pydantic API datamodel** validation using `Field(ge=1, le=65535)` on the REST API gateway connections schema.

Additionally, this PR refactors port validation logic into a shared validator in `airflow_shared.configuration.connection` and includes unit tests for both SQLAlchemy and Task SDK model implementations.

### PR Checklist
- [x] Related Issue: #68382
- [x] Unit tests added and passing
- [x] Code rebased on latest `upstream/main`

### DIFF
--- a/shared/configuration/src/airflow_shared/configuration/connection.py
+++ b/shared/configuration/src/airflow_shared/configuration/connection.py
@@ -1,4 +1,3 @@
-#
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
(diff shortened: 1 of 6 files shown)


pierrejeambrun (commented):
What I'm affraid is that if a user has a wrong connection in their current database, upgrading with this patch mean that the task sdk endpoint will fail at runtime. Basically a task could now fail because of this. (public API though is validating only for 'write' operations, which sounds good).

We should probably do the same on the task sdk, enforce validation on 'write' only and let 'read' as it is today.

I would keep validation from `3`, and delete `1` and `2` related code as a safer approach.

shared/configuration/src/airflow_shared/configuration/connection.py
```diff
@@ -0,0 +1,34 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+
+
+def parse_and_validate_port(port: int | str | None) -> int | None:
+    """Parse and validate connection port range."""
+    if port is None:
```
pierrejeambrun, line 22: 'connection' validation isn't related to `configuration` it probably shouldn't live in that module.

pierrejeambrun (commented):
Closing, stale, feel free to reopen when you have addressed above comments and fixed the CI.

PR #70212, opened 2026-07-22
fix ruff check docs command in FAQ

202201142:
Fixes the ruff check documentation command in the FAQ.


202201142 (commented):
The Problem: The doc stated that AIR302 checks for dynamic values in DAG/Task constructors. However, [Ruff's documentation](https://docs.astral.sh/ruff/rules/airflow3-moved-to-provider/) shows AIR302 is actually for modules moved to providers.

The Fix: Updated the reference to AIR304 (airflow3-dag-dynamic-value), which is the correct rule that [flags runtime-varying functions](https://docs.astral.sh/ruff/rules/airflow3-dag-dynamic-value/) like datetime.now().

henry3260 (approved):
Good

PR #70213, opened 2026-07-22
Helm: Scope KEDA_DB_CONN to worker and triggerer pods

pranayyelugam:
The KEDA database connection was emitted from the shared environment helper included by every component, so enabling KEDA autoscaling on workers (or the triggerer) injected the KEDA_DB_CONN secret into the scheduler, api-server, dag-processor, flower and the migration and user-creation jobs, none of which use it. Only the worker and triggerer scalers resolve this variable, so restricting to those pods.

 

---

---


chart/templates/_helpers.yaml
```diff
@@ -87,7 +87,8 @@ If release name contains chart name it will be used as a full name.
   {{- $triggererKedaEnabled := and .Values.triggerer.enabled .Values.triggerer.keda.enabled }}
   {{- $workersKedaNeedsDbConn := and .Values.workers.celery.keda.enabled (or (eq .Values.data.metadataConnection.protocol "mysql") (and .Values.pgbouncer.enabled (not .Values.workers.celery.keda.usePgbouncer))) }}
   {{- $triggererKedaNeedsDbConn := and $triggererKedaEnabled (or (eq .Values.data.metadataConnection.protocol "mysql") (and .Values.pgbouncer.enabled (not .Values.triggerer.keda.usePgbouncer))) }}
-  {{- if or $workersKedaNeedsDbConn $triggererKedaNeedsDbConn }}
+  {{- $component := .Component | default "" }}
+  {{- if or (and (eq $component "worker") $workersKedaNeedsDbConn) (and (eq $component "triggerer") $triggererKedaNeedsDbConn) }}
```
Miretpl: Variables defined under the `standard_airflow_environment` helper function, by definition, should be added to every component. Could we extract KEDA-related environments to a dedicated helper function which would be used in the proper components instead? I think it will make the implementation itself cleaner.

Miretpl (changes requested):

chart/templates/_helpers.yaml
pranayyelugam: I moved the KEDA connection out of standard_airflow_environment into a dedicated keda_airflow_environment helper, pulled in only by the worker and triggerer deployments.

On the helper's shape: I kept the logic (MySQL, or bypassing PgBouncer) and the worker-vs-triggerer selection inside the helper rather than making it dumb to avoid pushing that condition into both deployment templates and repeat it at each call site

pranayyelugam: @Miretpl Can you take a look when you get a chance?

Miretpl: I will. I'm going slowly from oldest to most recently updated. I will probably check tomorrow.

chart/templates/_helpers.yaml
```diff
@@ -158,6 +148,21 @@ If release name contains chart name it will be used as a full name.
   {{- end }}
 {{- end }}
 
+{{/* KEDA scaler database connection for the worker and triggerer autoscalers */}}
+{{- define "keda_airflow_environment" }}
+  {{- $triggererKedaEnabled := and .Values.triggerer.enabled .Values.triggerer.keda.enabled }}
+  {{- $workersKedaNeedsDbConn := and .Values.workers.celery.keda.enabled (or (eq .Values.data.metadataConnection.protocol "mysql") (and .Values.pgbouncer.enabled (not .Values.workers.celery.keda.usePgbouncer))) }}
+  {{- $triggererKedaNeedsDbConn := and $triggererKedaEnabled (or (eq .Values.data.metadataConnection.protocol "mysql") (and .Values.pgbouncer.enabled (not .Values.triggerer.keda.usePgbouncer))) }}
+  {{- $component := .Component | default "" }}
+  {{- if or (and (eq $component "worker") $workersKedaNeedsDbConn) (and (eq $component "triggerer") $triggererKedaNeedsDbConn) }}
```
Miretpl: I think that we could simplify this by:
1. Moving checking if keda is enabled for a particular component to the template files
2. Passing as an additional argument here the value of `.Values.triggerer.keda.usePgBouncer` or `.Values.workers.celery.keda.usePgBouncer`

and the rest of the logic is common.

chart/tests/helm_tests/other/test_keda.py
```diff
@@ -406,3 +406,49 @@ def test_overwrite_keda_max_replica_count(self):
         )
 
         assert jmespath.search("spec.maxReplicaCount", docs[0]) == 5
+
+    @staticmethod
+    def _env_names(doc):
+        return jmespath.search("spec.template.spec.containers[].env[].name", doc) or []
+
+    def test_worker_keda_db_conn_not_leaked_to_other_components(self):
+        """Worker KEDA_DB_CONN should render on the worker only, not on unrelated components."""
```
Miretpl:
```suggestion
```
Name of the test case is descriptive enough that we don't need additional comment I think.

chart/tests/helm_tests/other/test_keda.py
```diff
@@ -406,3 +406,49 @@ def test_overwrite_keda_max_replica_count(self):
...
+        )
+        worker, scheduler, api_server, dag_processor, triggerer = docs
+        assert "KEDA_DB_CONN" in self._env_names(worker)
+        assert "KEDA_DB_CONN" not in self._env_names(scheduler)
+        assert "KEDA_DB_CONN" not in self._env_names(api_server)
+        assert "KEDA_DB_CONN" not in self._env_names(dag_processor)
+        assert "KEDA_DB_CONN" not in self._env_names(triggerer)
+
+    def test_triggerer_keda_db_conn_not_leaked_to_other_components(self):
+        """Triggerer KEDA_DB_CONN should render on the triggerer only, not on the worker or others."""
```
Miretpl:
```suggestion
```
Same as above comment.

Miretpl (changes requested):

chart/templates/_helpers.yaml
pranayyelugam: @Miretpl Good suggestion. Updated.

pranayyelugam:
@Miretpl can I get a review on this? Thanks

Miretpl:
@pranayyelugam Most of the people here are doing work voluntarily, so getting the feedback may take some time, especially in the AI area where it is easy to generate a PR, but still people have to take a look at it if it is aligned with the project. Sooner or later, I will take another look at the PR and give another round of feedback. Please do not mention people in particular comments and/or get feedback when the PR is stale for a day or two. If it is stale for around 2 weeks, I think it is a good time period to try to ping any maintainer for the review.

Miretpl (approved):

Miretpl:
@pranayyelugam congrats and welcome!

Miretpl:
Could you do manual backport?

PR #70215, opened 2026-07-22
reject backslash-after-scheme urls in is_safe_url

Samin061:
`is_safe_url` in `airflow-core/src/airflow/api_fastapi/core_api/security.py` rejects a leading `//`, `/\`, `\/` and `\\`, but not a backslash that follows the scheme, so `https:\\evil.com` is parsed by urllib as a relative path, `urljoin` re-attaches the base netloc, and the netloc comparison passes. Browsers follow WHATWG, where `\` is `/` for special schemes, so the same string navigates to `https://evil.com` — `GET /auth/token/login?next=https:\\evil.com` returns it verbatim as the redirect location and `GET /auth/login?next=` forwards it to the auth manager login page, both without authentication. Normalising backslashes to forward slashes before the prefix check and the parse makes the helper agree with the parser that actually resolves the value; the second `unquote` goes away with it so the guard and the resolution look at the same string.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/security.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/security.py
@@ -1007,12 +1007,13 @@ def is_safe_url(target_url: str, request: Request | None = None) -> bool:
 
     # According to WHATWG for http/https /// is interpreted as // whereas urllib doesnt
     # this leads to an inconsistency where python returns a target url with /// as a valid url
-    # The same thing also happens with \ where under WHATWG \ are translated to /
-    target_url = unquote(target_url).strip()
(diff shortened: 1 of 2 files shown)


pierrejeambrun (approved):
Thanks, LGTM!

Samin061:
Thanks for the review and merge, glad this one landed.

Samin061:
Tested against 3.3.1rc1. The fix is in via the v3-3-test backport (#70515): `https:\\evil.com`, `https:/\evil.com`, `https:\/evil.com` and the percent-encoded `https%3A%5C%5Cevil.com` all return False against the default base URL, while same-origin paths like `/dags` and `dags/x` still resolve as safe. Working as expected for me.

PR #70218, opened 2026-07-22
Add deferrable mode to QuickSight operator and sensor

ozemin:
Adds `deferrable=True` support to `QuickSightCreateIngestionOperator` and `QuickSightSensor`, following the same waiter-based pattern as the rest of the Amazon provider. QuickSight has no official boto3 waiters (boto/boto3#2769), so this adds a custom `ingestion_complete` waiter and a `QuickSightIngestionCompletedTrigger`, and replaces the hook's hand-written polling loop, which held a worker slot with `time.sleep` and had no upper bound.

One thing without a direct precedent: `DescribeIngestion` requires `AwsAccountId`, so it is resolved on the worker before deferring and passed to the trigger — the triggerer never makes a sync STS call.

Behaviour changes: the synchronous wait is now bounded by `waiter_max_attempts` (default 60 attempts × 30s) instead of waiting forever; noted in the provider changelog. `QuickSightHook.wait_for_state` and the operator's `check_interval` are deprecated (still working) in favour of `wait_for_ingestion` / `waiter_delay`. Failures raise `QuickSightIngestionFailedError`, an `AirflowException` subclass.

---

- [X] Yes — Claude Code

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/hooks/quicksight.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/hooks/quicksight.py
@@ -18,10 +18,14 @@
 from __future__ import annotations
 
 import time
+import warnings
 
 from botocore.exceptions import ClientError
 
+from airflow.exceptions import AirflowProviderDeprecationWarning
+from airflow.providers.amazon.aws.exceptions import QuickSightIngestionFailedError
 from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
+from airflow.providers.amazon.aws.utils.waiter_with_logging import wait
 from airflow.providers.common.compat.sdk import AirflowException
 
 
@@ -52,6 +56,7 @@ class QuickSightHook(AwsBaseHook):
         wait_for_completion: bool = True,
         check_interval: int = 30,
         aws_account_id: str | None = None,
+        waiter_max_attempts: int = 60,
     ) -> dict:
         """
         Create and start a new SPICE ingestion for a dataset; refresh the SPICE datasets.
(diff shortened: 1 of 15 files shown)


o-nikolas (commented):
Looks like there are conflicts and build failures that need looking at. Also, I'm not sure of the breaking change (the new 30min timeout). I think large ingestions will easily pass that and fail dags for folks using this operator. We could make the default much higher, or go through a deprecation process where we log that the user should switch to setting a high default before we remove it. Or wait until a major release for the amazon provider package.

Curious what @vincbeck and @ferruzzi think

providers/amazon/docs/changelog.rst
```diff
@@ -26,6 +26,13 @@
 Changelog
 ---------
 
+.. warning::
```
vincbeck: Do not update changelog. This will be done by the release manager

providers/amazon/src/airflow/providers/amazon/aws/hooks/quicksight.py
```diff
@@ -66,31 +71,63 @@ def create_ingestion(
...
+                    "AwsAccountId": aws_account_id or self.account_id,
+                    "DataSetId": data_set_id,
+                    "IngestionId": ingestion_id,
+                },
+                failure_message="Amazon QuickSight SPICE ingestion failed.",
+                status_message="Status of Amazon QuickSight SPICE ingestion is",
+                status_args=["Ingestion.IngestionStatus", "Ingestion.ErrorInfo"],
+            )
+        except AirflowException as e:
+            raise QuickSightIngestionFailedError(str(e)) from e
```
vincbeck: Do we need a custom exception? `RuntimeError` suffice here no?

providers/amazon/src/airflow/providers/amazon/aws/operators/quicksight.py
```diff
@@ -71,22 +82,73 @@ def __init__(
...
         wait_for_completion: bool = True,
-        check_interval: int = 30,
+        waiter_delay: int = 30,
+        waiter_max_attempts: int = 60,
+        check_interval: int | None = None,
+        deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
         **kwargs,
     ):
         super().__init__(**kwargs)
+        if check_interval is not None:
```
vincbeck, line 91: This is kind of weird, you set a default value and then you check whether a value is assigned to it. In other words, anyone using this operator without setting `check_interval` will get a deprecation warning. I would leave the default value to `None` and keep following the deprecation warning process

ozemin: The default is `None`, so the warning only fires when `check_interval` is passed
explicitly. It's a few lines above the context collapsed here:

https://github.com/apache/airflow/blob/29658c9ec0f28caf97cc466cffb9f7a1c818799f/providers/amazon/src/airflow/providers/amazon/aws/operators/quicksight.py#L87-L99

Happy to change it if you meant something else.

ozemin:
> We could make the default much higher

Raised the default wait to 36 hours: 4320 attempts at the 30s delay, or 2160 at the sensor's 60s poke interval. That's how long AWS itself waits on a SPICE ingestion in the CloudFormation wait policy, where 36 hours is both the default and the maximum:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-quicksight-dataset-ingestionwaitpolicy.html

providers/amazon/src/airflow/providers/amazon/aws/hooks/quicksight.py
ozemin: Agreed, switched to `RuntimeError`.

vincbeck:
Static checks are failing

o-nikolas (approved):

PR #62922, opened 2026-03-05
AIP-104: Iterable Tasks and Batched Task Mapping

dabla:
---

##### Was generative AI tooling used to co-author this PR?

- [ x ] Yes (please specify the tool below)

Claude Code (Fable 5.1).

## Description

This PR is the initial implementation of **Iterable Tasks (IT)**, as discussed in the [devlist](https://lists.apache.org/list.html?dev@airflow.apache.org) and building upon the foundations of [AIP-104](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=421954527). (Originally prototyped as "Dynamic Task Iteration"; renamed to *Iterable Tasks* following review feedback to avoid confusion with Dynamic Task Mapping.)

For further context on the use cases and performance benefits of IT, see this [Medium Article](https://medium.com/apache-airflow/scaling-with-airflow-3-2-when-to-defer-and-when-to-use-native-async-operators-78a1ba35237d) and the new [`dynamic-task-mapping-vs-iteration.rst`](task-sdk/docs/dynamic-task-mapping-vs-iteration.rst) doc added in this PR, which compares IT with Dynamic Task Mapping (DTM) and Dynamic Task Batching in depth.

## The XCom Database Constraint Challenge

While porting our internal "monkey-patched" version of IT (used since Airflow 2.x) to the core, I've identified a significant technical hurdle regarding XCom handling.

Around Airflow 2.10/2.11, a change was introduced to the database constraints for the XCom table. Specifically:

- Current State: The DB prevents creating indexed XComs (`map_index >= 0`) unless a corresponding mapped TaskInstance exists in the `task_instance` table.
- The Conflict: IT is designed to process multiple indexed XComs within a single Task Instance. Because there is no 1-to-1 mapping of a sub-task index to a physical TI row, the DB constraint blocks the insertion of these results.

The drawback is that XComs wouldn't automatically be removed from the database when a TaskInstance is deleted, which is the purpose of that constraint. So appending the index to the XCom key would be a good enough solution for IT, but not for DTM.

### Current implementation in this PR

- **`XComIterable`** (`airflow.sdk.bases.xcom`) appends the sub-task index directly to the XCom key (`return_value_<index>`) to bypass the constraint, and exposes the results as a lazy `Sequence` (`__len__`/`__getitem__`/`__iter__`) so a downstream task can consume them the same way it would consume an `.expand()` result.
[...]

PR #70223, opened 2026-07-22
Add a `POST /xcoms/{dag_id}/{run_id}/{task_id}/keys` execution API endpoint

dabla:
Add a `POST /xcoms/{dag_id}/{run_id}/{task_id}/keys` execution API endpoint that
accepts a list of XCom keys and returns all matching values in a single database
query. Use it in `XComIterable` to reduce iteration from N round-trips to one.

### Motivation

`XComIterable` (introduced in [AIP-104 Task Iteration](https://github.com/apache/airflow/pull/62922)) stores per-index results
under distinct keys (`return_value_0`, `return_value_1`, …) with the same
`map_index`. The existing `GET …/slice` endpoint cannot be reused because it
ranges over `map_index` for a single key — the inverse structure. Iterating or
slicing a 1000-item result previously issued 1000 separate `XCom.get_one` calls.

### Changes

- **New endpoint** `POST /xcoms/{dag_id}/{run_id}/{task_id}/keys` — accepts
  `{"keys": [...]}`, returns values in request-key order (`null` for missing
  keys), filtered by `map_index` query param (default `-1`)
- **`has_xcom_access`** simplified: reads the optional `key` from
  `request.path_params` instead of requiring a `{key}` path segment, so a
  separate router/dependency is no longer needed
- **`XComKeysRequest`** Pydantic model added to execution API data models
- **`GetXComByKeys`** comms message, `XComOperations.get_by_keys` client method,
  `handle_get_xcom_by_keys` supervisor handler, and processor dispatch wiring
- **`XComIterable.__iter__`** and **`__getitem__` slice** now use `GetXComByKeys`
  — one round-trip for full iteration, deserialization kept lazy via generator
  expression in `__iter__`
- Tests for the new endpoint (`TestGetXComByKeys`) and for `XComIterable`
  (`TestXComIterable`)

---

- [ x  ] Yes (please specify the tool below)

Claude Sonnet 4.6 (GitHub Copilot CLI)

---


task-sdk/src/airflow/sdk/execution_time/schema/schema.json
```diff
@@ -2747,6 +2747,49 @@
       "title": "GetXCom",
       "type": "object"
     },
+    "GetXComByKeys": {
```
amoghrajesh, line 2848: Are these hand edited? Can you run the tooling for this instead? There should be a prek hook for this.

task-sdk/src/airflow/sdk/execution_time/lazy_sequence.py
amoghrajesh, line 1: Is this dead code as of this diff? It is never constructed by xcom_arg or any operator code, and it duplicates rather than replacing the actual class (LazyXComSequence) that has the N-round-trip problem described in the PR.

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -113,6 +114,7 @@ def has_xcom_access(
     dependencies=[Depends(has_xcom_access)],
 )
 
+
```
amoghrajesh, line 117:
```suggestion
```

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
```diff
@@ -356,6 +358,38 @@ def get_xcom(
...
+        "Optimised for XComIterable iteration, reducing N round-trips to one."
+    ),
+)
+def get_xcom_by_keys(
+    dag_id: str,
+    run_id: str,
+    task_id: str,
+    request_body: XComKeysRequest,
+    session: SessionDep,
+    map_index: Annotated[int, Query()] = -1,
```
amoghrajesh, line 375: Why? The PR states that map_index is not needed and this is fetching for a single map index.

amoghrajesh (commented):
I haven't followed up closely with AIP 104, but I have some qns from an initial look.

task-sdk/src/airflow/sdk/execution_time/lazy_sequence.py
dabla: No it's not a duplicate, as here is slices over keys instead of indices, it will be used in AIP-104.

airflow-core/src/airflow/api_fastapi/execution_api/routes/xcoms.py
dabla: It is not used for the slicing, but it can be used in combination with batch, which means each IterableOperator will have it's own instance of XComIterable and thus have a map_index associated with it.

task-sdk/src/airflow/sdk/execution_time/lazy_sequence.py
dabla: Maybe this PR should be merged after AIP-104 [PR](https://github.com/apache/airflow/pull/62922) is merged?

dabla:
This PR should only be merged once [AIP-104](https://github.com/apache/airflow/pull/62922) is merged.

PR #64610, opened 2026-04-01
Add partition_key and partition_key_regexp_pattern filters for asset events

hussein-awala:
## Summary

Add two query parameters for filtering asset events by partition key across the full stack:

- **`partition_key`** — exact-match filter using `==` equality; leverages the B-tree
  composite index `(asset_id, partition_key)` for fast lookups. Always available.
- **`partition_key_regexp_pattern`** — regular-expression filter using SQLAlchemy's
  `regexp_match` (PostgreSQL `~`, MySQL `REGEXP`). Opt-in (see below).

The two parameters can be combined; conditions are applied with AND logic.

Also includes:
- Composite index `(asset_id, partition_key)` on `asset_event` with an Alembic migration.
- `InletEventsAccessor` exposes `.partition_key(value)` and `.partition_key_regexp_pattern(pattern)`.
- Docs in `assets.rst` with REST and accessor usage examples.

## Regexp filtering is opt-in (ReDoS safeguard)

A user-supplied regular expression is evaluated by the database's own regex engine, which is a
ReDoS vector. It is therefore controlled by a single config, **`[api] regexp_query_timeout`**
(float seconds):

- `0` (default): regexp filtering is disabled — `partition_key_regexp_pattern` returns HTTP 400.
- any positive value: enables the filter and bounds each regexp query's runtime.

The bound is enforced as a transaction-local `statement_timeout` on PostgreSQL and as
`max_execution_time` on MySQL (previous value restored afterwards). It is applied automatically by
the filter's FastAPI dependency, so a query using this filter can never run unbounded. Invalid
patterns are rejected with HTTP 400 via `re.compile()`.

Exact-match `partition_key` is unaffected by this setting.

## Test plan

- [x] Core REST API and Execution API tests for exact match and regexp filtering (by-asset / by-alias)
- [x] Combining `partition_key` and `partition_key_regexp_pattern` (AND)
- [x] Regexp filter disabled by default -> HTTP 400; invalid pattern -> HTTP 400
- [x] Timeout applied on PostgreSQL/MySQL and restored; float (fractional-second) values
- [x] Task SDK client/supervisor/accessor forward both params

PR #54891, opened 2025-08-24
feat: Add lazy filtering support for inlet events by time range, ordering, and limit

hussein-awala:
This PR extends the asset events API client and inlet events accessor to support additional query options while keeping the accessor **lazy**, meaning events are only fetched when accessed.

## Changes
- Updated `InletEventsAccessors.__getitem__` to accept optional filtering arguments (`after`, `before`, `ascending`, `limit`) without breaking existing usage (new key type is `tuple[<existing key type>, dict]`)
- The accessor continues to **lazily fetch events** from `SUPERVISOR_COMMS` only when accessed, but it will only fetch the events that match the filters when they are provided.

## Impact
- Existing calls like `accessors[asset]` continue to work as before.
- New functionality allows **lazy, filtered queries** by time range, limit, and sort order.
- Enables more efficient and flexible retrieval of asset event history without eager fetching.

## Use Cases
- Useful when the asset is used as input and not as a trigger source.
- Easily retrieve the latest available value (`ascending=False` and `limit=1`).
- Filter by date range for backfill scenarios without fetching the full history and filtering locally.

PR #67941, opened 2026-06-03
Add asset partition sensor

Leondon9:
Adds an `AssetPartitionSensor` (and backing `AssetPartitionTrigger`) to the standard provider so a **time-scheduled** Dag can wait, at the task level, for a specific asset **partition** produced by another Dag — the use case from #67375.

### What it does

- `AssetPartitionSensor(asset=..., partition_key=..., after=..., deferrable=...)` succeeds once an asset event with the given `partition_key` exists.
- `after` (templated) bounds the lookup to the run's interval, so a stale event carrying a reused key from an earlier run doesn't satisfy the wait. `partition_key` is templated too.
- Deferrable mode releases the worker slot while waiting.
- The waited-for asset is registered as a task **inlet**, so the dependency stays visible in the asset/lineage graph even though the Dag stays time-scheduled (it is not asset-triggered).

### Core change

The only `airflow-core` change wires the triggerer so a deferred trigger can look up asset events: `GetAssetEventByAsset` / `GetAssetEventByAssetAlias` are added to the `ToTriggerSupervisor` union and dispatched through a shared request handler (also adopted by the worker supervisor, which becomes a net reduction). A trigger has no metadata-DB or Execution API access of its own, so this is required for deferrable mode — without it the deferred lookup fails the supervisor's frame validation. A regression test covers both the union membership and the dispatch.

Builds on the asset-event `partition_key` filters from #64610; this PR adds no API, DB, OpenAPI, or UI changes of its own.

### Tests & docs

- Unit tests for the sensor, trigger, shared request handlers, and the triggerer supervisor path.
- How-to guide (`sensors/asset.rst`) and an example Dag.

related: #67375

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Codex (GPT-5) and Claude Code (Opus 4.8)

Generated-by: Codex (GPT-5) and Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70225, opened 2026-07-22
Add AssetEventSensor to standard provider

hussein-awala:
## Summary

Adds a new `AssetEventSensor` to the `standard` provider that waits for **asset events** matching a set of filters to reach an expected count.

It builds on the asset-event query filters recently merged into the API/Task SDK:
- `partition_key` / `partition_key_regexp_pattern` (#64610)
- `extra` key/value filtering
- time range / ordering / limit (`after`, `before`, `ascending`, `limit`) from #54891

Because those filters are only available from Airflow 3.4, the sensor is gated to **Airflow 3.4+**.

## Features

- Target an `Asset` / `AssetAlias` (or raw `name` / `uri` / `alias_name`).
- Filters: `after`, `before`, `ascending`, `limit`, `partition_key`, `partition_key_regexp_pattern`, `extra`.
- `expected_count`: how many (processed) events are required to succeed. `-1` (default) means **at least one** (`count >= 1`); `0` means **exactly zero**; any other positive value requires an **exact match** (`count == expected_count`). Values `< -1` are rejected. Note: if `limit` is set below an exact `expected_count`, the condition can never be satisfied and the sensor waits until it times out.
- `process_result`: an optional callable (or dotted import path) to transform / deduplicate / filter the fetched events **before** the count check. The processed events are pushed to XCom.

## Scope: synchronous only

This PR intentionally ships a **synchronous sensor only**, per review feedback.

An earlier revision added an `AssetEventTrigger` and `deferrable=True`, but the execution API calls behind asset event fetching (`GetAssetEventByAsset` / `GetAssetEventByAssetAlias`) are not available in the triggerer, so that path could not work in practice. It has been removed rather than shipped broken.

Deferrable support can follow in separate PRs, roughly:

1. Triggerer changes needed to support asset event fetching.
2. Refactor `InletEventsAccessor` internals to ready it for a native async interface.
3. Add the trigger on top of the above.
4. Add `deferrable=True` to the sensor.

## Notes

- Uses `airflow.sdk.execution_time.context.InletEventsAccessor`, which is not (yet) a formally public API for arbitrary assets. It mirrors how `external_task.py` already imports from `airflow.sdk.execution_time.*`.

## Docs

- Adds a howto guide (`sensors/asset.rst`) covering basic usage, filtering / `expected_count`, and `process_result`, wired into the provider's `how-to-guide` reference.
- Adds an `example_asset_sensor` example DAG (version-gated to 3.4+) referenced by the guide.

## Tests

- Sensor unit tests are backed by a DB harness (a `SUPERVISOR_COMMS` stand-in in `conftest.py` that runs the real execution-API SQL/filter paths against the test DB), so `poke` is exercised against real asset events rather than mocks.
- Covers poke semantics (exact / at-least-one / exactly-zero / no-match), all filters incl. `partition_key`(_regexp)/`extra`/time-range/ordering/limit, `process_result` (callable + import-path, dedup/filter/expand/empty/raise), and argument validation.

---

Generated-by: Cursor (Claude Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


Lee-W (commented):
Thanks! I'll take a deeper look, but I'd first like to understand why we need a sensor like this.

I thought we could always use a custom `PartitionMapper` to achieve something similar. Are there any use cases I'm missing? Thanks!

hussein-awala:
> Thanks! I'll take a deeper look, but I'd first like to understand why we need a sensor like this.
> 
> I thought we could always use a custom `PartitionMapper` to achieve something similar. Are there any use cases I'm missing? Thanks!

@Lee-W Good point, but they're at different layers. `PartitionMapper` is scheduler-side: it decides whether/how a run is *created* from partition-aligned events, and its only count gate (`MinimumCount`/`WaitForAll`) works over a declared partition-key window on a `PartitionedAssetTimetable`.

`AssetEventSensor` is a task-level primitive: it waits *inside an already-running Dag* (cron/manual/any trigger) until events matching arbitrary filters reach a count, then passes them downstream via XCom. Two things a mapper can't do:

1. Mid-Dag gating in a non-asset-scheduled Dag (e.g. wait for ≥3 events on today's partition before continuing).
2. Filter on `extra`/time-range and reuse the payload (e.g. wait for N events with `extra={"status":"validated"}`, then `process_result` them for downstream tasks), mappers only see `partition_key`.

We can say it's a wrapper for the `inlet_events` but with the sensor features (wait for the events, max wait, soft fail, ...)

So complementary, not redundant. Can add this to the docs if useful.

Lee-W:
Thanks for answering in details. But I'm still not convinced this should be a task-level thing.

1. Why couldn't we just split them into two Dags and let the downstream one be asset-scheduled? If a Dag needs both a time and an asset schedule, we also have `AssetOrTimeSchedule`. And for the "wait for N" cases, we have `RollupMapper` with a `WaitPolicy` (`MinimumCount` / `WaitForAll`).
2. And why do we need `extra` here? `partition_key` is the field we designed to be the discriminator — it's the indexed one, while `extra` is free-form metadata the scheduler doesn't index. If a producer needs to distinguish events by something, that something belongs in the partition key, and a custom `PartitionMapper` can decode it into whatever shape the downstream Dag needs.

But my main concern is that this dependency is not known by Airflow. We can't schedule on it, and it doesn't show up in the lineage. The whole point of assets is that declaring a data dependency makes it visible to the platform. That works against what the feature is for.

nathadfield:
@Lee-W We hit this exact tension migrating some pipelines to partitioned assets, a concrete data point.

Take a simple linear DAG: task A runs, then task B needs an asset produced by a different DAG before it can proceed, then task C runs. Today that's A -> ExternalTaskSensor -> B -> C, and A runs immediately regardless of whether the other DAG's data is ready. Forcing that dependency onto the DAG's own schedule would delay A too, a real regression, since A never needed it.

Splitting A into its own DAG doesn't help either: A, B, and C are one logical unit of work that should be run, observed, and retried together, and splitting them apart only serves the scheduler's DAG-level-only asset constraint, not any real independence between them. A task-level sensor between A and B is the right fit, and this shape isn't rare.

On the visibility concern: could the sensor also set inlets for the asset(s) it waits on? inlets/outlets already drive the lineage graph independent of scheduling, so the dependency stays visible to the platform without tying it to what triggers the DAG run.

Leondon9:
I opened the original discussion behind this (#67375), so let me add the concrete case, since the "why task-level?" question is the crux.

The workload is a **time-scheduled** batch consumer (`@daily`, etc.) whose SQL is written against its own `data_interval_start` / `_end`. Mid-run it needs to wait for an upstream asset partition (e.g. `orders / dt=<interval>`) produced by a *different* Dag on its own cadence.

Why the existing mechanisms don't quite cover it:

- **Asset-schedule the consumer / `AssetOrTimeSchedule`:** asset-triggered runs get a *degenerate* data interval — `DataInterval.exact(run_after)`, pinned to the trigger instant — not the window the consumer's SQL is built on. Moving the wait onto the schedule changes the run's contract, even though the consumer still wants its own time interval.
- **`RollupMapper` + `WaitPolicy`:** scheduler-side — it gates whether a *run is created* from partition-aligned events. It can't gate *between tasks* in an already-running, time-scheduled Dag, which is where the wait belongs here.

On the visibility concern (the strongest objection, I think): I don't think satisfying it requires scheduling on the asset. Airflow already separates lineage from scheduling — inlets/outlets draw the dependency graph independently of what triggers a run. A task-level sensor can declare the awaited asset as an inlet, so the dependency stays visible to the platform (asset graph, lineage, OpenLineage) without coupling the run to asset-triggering. That keeps both: the consumer keeps its time schedule and data interval, and the data dependency stays first-class.

So I'd argue this is a genuine gap for time-partitioned batch consumers — the alternatives either change the run's contract or act at the wrong layer — and a task-level sensor is complementary to asset scheduling, not a replacement for it.

providers/standard/src/airflow/providers/standard/triggers/asset.py
```diff
@@ -0,0 +1,206 @@
...
+                uri=self.uri,
+                alias_name=self.alias_name,
+                after=self.after,
+                before=self.before,
+                ascending=self.ascending,
+                limit=self.limit,
+                partition_key=self.partition_key,
+                partition_key_regexp_pattern=self.partition_key_regexp_pattern,
+                extra=self.extra,
+            )
```
uranusjr, line 190: This will not work because ultimately the execution API calls use GetAssetEventByAsset and GetAssetEventByAssetAlias, both not available in the triggerer. The tests do not surface this because they mock the actual API calls.

Also, I would prefer having a natively async version instead of slapping on sync_to_async. Not just here, but essentially everywhere in the trigger.

providers/standard/src/airflow/providers/standard/triggers/asset.py
```diff
@@ -0,0 +1,206 @@
...
+    extra: dict[str, str] | None,
+) -> list[Any]:
+    """
+    Fetch asset events matching the given filters.
+
+    This uses the Task SDK :class:`InletEventsAccessor`, which lazily fetches events
+    from the supervisor. It is a blocking call and must be wrapped with
+    ``sync_to_async`` when used from within the triggerer.
+    """
+    from airflow.sdk.execution_time.context import InletEventsAccessor
```
uranusjr, line 64: Using InletEventsAccessor here seems unnecessary. Coupling with the async issue, some refactoring to the underlying implementation would be a good idea.

providers/standard/src/airflow/providers/standard/example_dags/example_asset_sensor.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    # [END example_asset_event_sensor_filtered]
+
+    # [START example_asset_event_sensor_async]
+    wait_for_asset_event_async = AssetEventSensor(
+        task_id="wait_for_asset_event_async",
+        asset=my_asset,
+        process_result=latest_event_only,
+        deferrable=True,
+    )
```
uranusjr, line 75: Aside from the deferrable issue, passing `process_result=latest_event_only` probably also won’t work? The sensor’s docstring says _a function defined in a DAG file will import on the worker but typically fail to import in the triggerer process_.

uranusjr (changes requested):
There are multiple issues with deferrable support. I’d suggest dropping this for now and only implement a synchronous sensor in this PR. After that’s merged, we can

1. Introduce triggerer changes needed to support asset event fetching.
2. Refactor InletEventsAccessor internals to ready it for an async interface.
3. Add the trigger to depend on the above changes.
4. Add `deferrable=True` to the sensor that only works on Airflow 3.4+.

Leondon9:
This sequencing makes sense. FWIW step 1 — the triggerer changes for asset-event fetching — is already implemented in #67941, which independently hit the exact issue you flagged: `GetAssetEventByAsset` / `GetAssetEventByAssetAlias` aren't handled in the triggerer, and a mocked `SUPERVISOR_COMMS` hides it. That PR adds them to the `ToTriggerSupervisor` union + dispatch, with a regression test driving the real supervisor path (it fails if the union entries are removed).

Its trigger also takes the shape you're describing: natively async (`await SUPERVISOR_COMMS.asend(GetAssetEventByAsset(...))`), no `sync_to_async` and no `InletEventsAccessor` — so for this narrower partition-only case it doesn't need the accessor async-refactor (that refactor is still separately useful if the generic sensor routes through the accessor).

Happy for #67941 to serve as that step-1 prerequisite, or for the relevant commits to be split/reviewed however fits the plan — whatever keeps this moving.

hussein-awala:
> There are multiple issues with deferrable support. I’d suggest dropping this for now and only implement a synchronous sensor in this PR. After that’s merged, we can
> 
> 1. Introduce triggerer changes needed to support asset event fetching.
> 2. Refactor InletEventsAccessor internals to ready it for an async interface.
> 3. Add the trigger to depend on the above changes.
> 4. Add `deferrable=True` to the sensor that only works on Airflow 3.4+.

@uranusjr Thanks, that's a fair catch on the triggerer,`GetAssetEventByAsset` / `GetAssetEventByAssetAlias` aren't available there, so the deferrable path couldn't have worked in practice, and my DB-backed tests hid it because the harness stands in for `SUPERVISOR_COMMS` and runs the execution-API SQL directly.

I've dropped deferrable support from this PR: `AssetEventTrigger` and its tests are gone, the shared fetch/count/serialize helpers moved into the sensor module, and the trigger is deregistered from `provider.yaml`. What's left is the synchronous sensor only. That also resolves your point on the example DAG — `process_result=latest_event_only` is fine now that it only ever runs on the worker.

providers/standard/src/airflow/providers/standard/sensors/asset.py
```diff
@@ -0,0 +1,225 @@
...
+        "partition_key_regexp_pattern",
+        "extra",
+        "after",
+        "before",
+    )
+
+    def __init__(
+        self,
+        *,
+        asset: Asset | AssetAlias | None = None,
```
uranusjr, line 148: I wonder if it make sense to make it either positional (not keyword-only), or maybe rename it e.g. `obj`. `asset=AssetAlias(...)` seems wrong to me.

providers/standard/src/airflow/providers/standard/sensors/asset.py
```diff
@@ -0,0 +1,225 @@
...
+def _count_satisfied(count: int, expected_count: int) -> bool:
+    """
+    Return whether the number of (processed) asset events satisfies the expectation.
+
+    ``expected_count == -1`` means "at least one" (``count >= 1``); any other value
+    requires an exact match (``count == expected_count``).
+    """
+    if expected_count == -1:
+        return count >= 1
+    return count == expected_count
```
uranusjr, line 46: This means we won’t be able to do “at least n” where n > 1. Instead of having a magic number `-1`, maybe we can add a separate “policy” flag?

Asset partitioning already introduced a similar concept with `MinimumCount`, and `WaitForAll`. Maybe this can use the same vocabulary?

providers/standard/src/airflow/providers/standard/sensors/asset.py
```diff
@@ -0,0 +1,225 @@
...
+    :param before: Only include events at or before this timestamp.
+    :param ascending: Whether events are returned in ascending timestamp order.
+    :param limit: Maximum number of events to fetch.
+    :param partition_key: Filter by exact partition key match.
+    :param partition_key_regexp_pattern: Filter by partition key regexp pattern.
+    :param extra: Filter by key/value pairs contained in the event ``extra`` field.
+    :param expected_count: The number of events required to succeed. ``-1`` (the default) means
+        "at least one" (``count >= 1``); ``0`` means "exactly zero"; any other positive value
+        requires an exact match. Note that if ``limit`` is set below an exact ``expected_count``
+        the condition can never be satisfied (the sensor will wait until it times out).
```
uranusjr, line 126: Instead of leaving this footgun, maybe we can add a check to make `limit < expected_count` error early?

providers/standard/tests/unit/standard/conftest.py
```diff
@@ -0,0 +1,185 @@
...
+ASSET_NAME = "my_asset"
+ASSET_URI = "s3://bucket/key"
+ALIAS_NAME = "my_alias"
+
+
+def event_timestamp(day: int) -> datetime:
+    return datetime(2024, 1, day, tzinfo=timezone.utc)
+
+
+class _DBBackedComms:
```
uranusjr, line 45: Please use `mock_supervisor_comms` instead and feed prepared `AssetEventsResult`s. The filter tests are asserting SQL semantics that are already covered by `test_asset_events.py`; what the sensor owes is it builds the right message, so assert on `send.assert_called_once_with(GetAssetEventByAsset(...))`. As a bonus, we’ll also be able to drop `db_test`.

uranusjr:
Also, the lineage point raised above makes sense to me; we should resolve that before moving forward. Let’s have the sensor declare the awaited asset as an inlet and read `context["inlet_events"][asset]`. That makes the dependency visible without coupling it to scheduling, and drops the `InletEventsAccessor` usage.

Leondon9:
I raised this use case in #67375, and I've got the deferrable path working in #67941 already — happy to pick up the follow-up (steps 1/3/4) once the sync sensor lands.

It's in the shape you described: triggerer support for `GetAssetEventByAsset`/`ByAssetAlias` (with a test on the real supervisor path, which is where the mocked-comms gap hides), a natively-async trigger with no `InletEventsAccessor`, and `deferrable` gated to 3.4+. Will rebase onto the merged sensor and reshape however fits the sequencing.

PR #70228, opened 2026-07-22
Parse additional Kafka, Redis & SQS options in `broker_transport_options`

stephen-bracken:
Parse additional dict options in `broker_transport_options`
---

The following options in `broker_transport_options` are processed as nested dictionaries in the celery config, they should be json parsed in the same way as `sentinel_kwargs` when creating the `default_celery.DEFAULT_CELERY_CONFIG`

Kafka:
- `kafka_admin_config`
- `kafka_common_config`
- `kafka_consumer_config`
- `kafka_producer_config`

Redis:
- `sentinel_kwargs`

SQS:
- `client-config`
- `fetch_message_attributes`
- `predefined_exchanges`
- `predefined_queues`
- `sqs-creation-attributes`

- [X] No

---


providers/celery/tests/unit/celery/executors/test_celery_executor.py
```diff
@@ -724,15 +724,24 @@ def test_celery_executor_with_no_recommended_result_backend(caplog):
...
-def test_sentinel_kwargs_loaded_from_string():
+@pytest.mark.parametrize(
+    (
+        "option",
+        "value",
+        "expected",
+    ),
+    [
+        ("sentinel_kwargs", '{"service_name": "mymaster"}', {"service_name": "mymaster"}),
+        ("kafka_common_config", '{"compression.type": "zstd"}', {"compression.type": "zstd"}),
```
Miretpl: Could you test any option which is valid for the parsing?

Miretpl (changes requested):
Could you update the provider documentation configuration?

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,10 +44,62 @@
...
+    "sentinel_kwargs",
+    "sqs-creation-attributes",
+]
+
+
 def _broker_supports_visibility_timeout(url):
     return url.startswith(("redis://", "rediss://", "sqs://", "sentinel://"))
 
 
+def _broker_transport_options(broker_url: str, conf) -> dict[str, Any]:
```
Miretpl: Could you add an annotation for `conf`?

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,10 +44,62 @@
...
+                "[celery_broker_transport_options] visibility_timeout = <seconds>"
+            )
+
+    # Parse dict options
+    for option in _BROKER_TRANSPORT_DICT_OPTIONS:
+        if option in broker_transport_options:
+            try:
+                option_value = json.loads(broker_transport_options[option])
+                if not isinstance(option_value, dict):
+                    raise ValueError
```
Miretpl: Could you specify the proper exception message here? It may be confusing for user what happened without any message.

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,10 +44,62 @@
...
+                broker_transport_options[option] = option_value
+            except Exception:
+                raise AirflowException(
```
Miretpl:
```suggestion
                raise ValueError(
```
We're moving away from `AirflowException`, and when we change something, we fix that place in the same change. You will also need to change the `generated/known_airflow_exceptions.txt` file.

providers/celery/src/airflow/providers/celery/get_provider_info.py
```diff
@@ -292,13 +292,69 @@ def get_provider_info():
             "celery_result_backend_transport_options": {
```
Miretpl, line 292: There is a missing `predefined_queues` here and in `provider.yaml`.

providers/celery/src/airflow/providers/celery/get_provider_info.py
```diff
@@ -292,13 +292,69 @@ def get_provider_info():
...
+                        "version_added": "3.24.0",
+                        "type": "string",
+                        "example": '{"connect_timeout": 5}',
+                        "default": None,
+                    },
+                    "fetch_message_attributes": {
+                        "description": "SQS fetch message attributes.\n",
+                        "version_added": "3.24.0",
+                        "type": "string",
+                        "example": '{\n  "MessageSystemAttributeNames": ["SenderId", "SentTimestamp"],\n  "MessageAttributeNames": ["S3MessageBodyKey"]\n}\n',
```
Miretpl:
```suggestion
                        "example": '{"MessageSystemAttributeNames": ["SenderId", "SentTimestamp"]',
```
and similarly for others. While I agree it would be more readable, I assume some users may think the formatted version will work in the configuration. It will not, and due to that, I believe that we should have a less formatted but more correct version in terms of usage. (I removed `MessageAttributeNames` as this is only an example, but if you think it should stay in, you can leave it - I didn't work with SQS, so it is hard for me to judge the correctness of the example itself without checking now).

Miretpl (changes requested):

providers/celery/src/airflow/providers/celery/executors/default_celery.py
stephen-bracken: This method is extracted from the existing broker_transport_options parse in `get_default_celery_config`, but sure I will improve this

stephen-bracken: also note that the `ValueError` here gets excepted on line 100 and re-raised

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,11 +49,68 @@
...
+                "[celery_broker_transport_options] visibility_timeout = <seconds>"
+            )
+
+    # Parse dict options
+    for option in _BROKER_TRANSPORT_DICT_OPTIONS:
+        if option in broker_transport_options:
+            try:
+                option_value = broker_transport_options[option]
+                if not isinstance(option_value, str):
+                    raise ValueError(f"broker_transport_option {option} is invalid: {option_value}")
```
Miretpl:
```suggestion
                    raise ValueError(f"broker_transport_option {option} is not string: {option_value}")
```
To make the message more precise.

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,11 +49,68 @@
...
+                option_json = json.loads(option_value)
+                if not isinstance(option_json, dict):
+                    raise ValueError(f"broker_transport_option {option} is invalid: {option_json}")
```
Miretpl:
```suggestion
                    raise ValueError(f"broker_transport_option {option} value is not dictionary: {option_json}")
```
To make the message more precise.

providers/celery/src/airflow/providers/celery/executors/default_celery.py
```diff
@@ -44,11 +49,68 @@
...
+                broker_transport_options[option] = option_json
+            except Exception as exc:
+                raise ValueError(
+                    f"Broker transport option {option} should be written in the correct dictionary format."
```
Miretpl:
```suggestion
                    f"Broker transport option {option} value should be written in the correct JSON format."
```

Miretpl (changes requested):
Small nits and we will be ready to merge. Thanks for taking care of AirflowException here!

Miretpl (approved):
Will leave it a day or two open if anyone would like to make review too.

PR #70229, opened 2026-07-22
Redact secrets from LLMRetryPolicy classification prompts

Lee-W:
## Why
LMRetryPolicy sends the failing task's exception message to a third-party LLM, which can leak connection strings, credentials, or PII.

## What
The message now passes through Airflow's secrets masker by default; set `redact_exception=False` to send the raw text when it is known to be safe.

 

---

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
@@ -28,6 +28,8 @@ from typing import TYPE_CHECKING
 
 from pydantic import BaseModel
 
+from airflow.providers.common.compat.sdk import redact
+
 try:
     from airflow.sdk.definitions.retry_policy import (
         ExceptionRetryPolicy,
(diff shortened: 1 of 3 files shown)


gopidesupavan (approved):

providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
```diff
@@ -109,12 +133,15 @@ def __init__(
         instructions: str | None = None,
         fallback_rules: list[RetryRule] | None = None,
         timeout: float = 30.0,
+        *,
+        redact_exception: bool = True,
```
kaxil, line 137: `providers/common/ai/docs/retry_policies.rst` isn't updated here: the Parameters table lists the five existing constructor args but not `redact_exception`, and "How it works" step 1 still reads "Sends the exception message to the configured LLM". The "Local LLM support" section right below that table is framed as the mitigation for "environments where exception data must not leave the infrastructure", which now understates the default behaviour. Once the page carries that guidance, the 15-line `.. warning::` block here could shrink to a pointer rather than restating the `:param redact_exception:` text.

kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
Lee-W: Yep, updated. I do think we should have a better way to keep these in sync, but I'd rather keep this as is for now and address that in a follow-up PR.

PR #70234, opened 2026-07-22
Fix KubernetesPodOperator dry_run requiring live Kubernetes API client

### LINKED ISSUE #45812: KubernetesPodOperator dry_run failure

cncf-kubernetes

apache-airflow-providers-cncf-kubernetes 4.3.0

### Apache Airflow version

2.3.4

### Operating System

Linux

### Deployment

### Deployment details

n/a

### What happened

We are upgrading from apache-airflow-providers-cncf-kubernetes 3.0.0 to 4.3.0 (going slowly through releases).

We have a custom script, that during docker image build of our airflow, tests all dags and all dag tasks in dry_run mode. Mostly to detect Python syntax errors, dag cycles duplicate tasks, wrong imports, ntemplating errors etc.

This was working all fine with our existing airflow, but we decided to upgrade airflow to newer version, and that also means updating airflow providers. After fixing bunch of other issues, I found the issues with KubernetedPodOperator dry run.

New dry_run added in https://github.com/apache/airflow/commit/d56ff765e15f9fcd582bc6d1ec0e83b0fedf476a invokes `KubernetesPodOperator` `build_pod_request_obj()` method which has a call to a property `self.hook.is_in_cluster`:

```python
        pod.metadata.labels.update(
            {
                'airflow_version': airflow_version.replace('+', '-'),
                'airflow_kpo_in_cluster': str(self.hook.is_in_cluster),
            }
        )
```

Unfortunately this property constructs a Kube API client object which requires kube client config / credentials to work.

```python
    @property
    def is_in_cluster(self):
        """Expose whether the hook is configured with ``load_incluster_config`` or not"""
        if self._is_in_cluster is not None:
            return self._is_in_cluster
        self.api_client  # so we can determine if we are in_cluster or not
        return self._is_in_cluster```
```

This causes dry_run to not able to execute in isolated test environment:

```
  File "<stdin>", line 97, in <module>
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/models/dag.py", line 2307, in cli
    args.func(args, self)
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/cli/cli_parser.py", line 51, in command
    return func(*args, **kwargs)
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/utils/cli.py", line 99, in wrapper
    return f(*args, **kwargs)
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/cli/commands/task_command.py", line 545, in task_test
    ti.dry_run()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1815, in dry_run
    self.task.dry_run()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/operators/kubernetes_pod.py", line 607, in dry_run
    pod = self.build_pod_request_obj()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/operators/kubernetes_pod.py", line 595, in build_pod_request_obj
    'airflow_kpo_in_cluster': str(self.hook.is_in_cluster),
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/hooks/kubernetes.py", line 283, in is_in_cluster
    self.api_client  # so we can determine if we are in_cluster or not
  File "/usr/local/lib/python3.9/functools.py", line 993, in __get__
    val = self.func(instance)
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/hooks/kubernetes.py", line 291, in api_client
    return self.get_conn()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/hooks/kubernetes.py", line 239, in get_conn
    config.load_kube_config(
  File "/home/airflow/.local/lib/python3.9/site-packages/kubernetes/config/kube_config.py", line 808, in load_kube_config
    loader = _get_kube_config_loader(
  File "/home/airflow/.local/lib/python3.9/site-packages/kubernetes/config/kube_config.py", line 767, in _get_kube_config_loader
    raise ConfigException(
kubernetes.config.config_exception.ConfigException: Invalid kube-config file. No configuration found.
```

We would like to continue using dry_run, but be able to run it without providing credentials or kube config. It does not need to be 100% accurate.

Two options:

* env var to bypass setting of `airflow_kpo_in_cluster` label in dry run mode, if user requests to do so.
* never populate it in dry_run mode. (change signature of `build_pod_request_obj` to have `dry_run: bool = False` kwarg and invoke it with `dry_run=True` in KubernetesPodOperator.dry_run()` method.

(or both) 

n/a

### How to reproduce

n/a

### Anything else

n/a

### Code of Conduct

Aaryan123456679:
`airflow dags test` (and calling `KubernetesPodOperator.dry_run()` directly) failed outside a real cluster because `build_pod_request_obj()` unconditionally read `hook.is_in_cluster` to set the `airflow_kpo_in_cluster` label. That property instantiates a Kubernetes API client, which requires kube credentials/config to be available — something a dry run must not need.

This adds a keyword-only `dry_run` parameter to `build_pod_request_obj()` (default `False`, so existing callers are unaffected) and has `dry_run()` skip the `is_in_cluster` lookup when set.

closes: #45812

### DIFF
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
@@ -1557,12 +1557,16 @@ class KubernetesPodOperator(BaseOperator):
             except kubernetes.client.exceptions.ApiException:
                 self.log.exception("Unable to delete pod %s", self.pod.metadata.name)
 
-    def build_pod_request_obj(self, context: Context | None = None) -> k8s.V1Pod:
+    def build_pod_request_obj(self, context: Context | None = None, *, dry_run: bool = False) -> k8s.V1Pod:
         """
         Return V1Pod object based on pod template file, full pod spec, and other operator parameters.
 
         The V1Pod attributes are derived (in order of precedence) from operator params, full pod spec, pod
         template file.
+
+        :param dry_run: if True, skip anything that requires a live Kubernetes API client
+            (e.g. determining whether the hook is running in-cluster), since dry runs must not
+            require kube credentials or config to be available.
         """
         self.log.debug("Creating pod for KubernetesPodOperator task %s", self.task_id)
 
@@ -1673,12 +1677,11 @@ class KubernetesPodOperator(BaseOperator):
         pod.metadata.labels.update(labels)
         # Add Airflow Version to the label
         # And a label to identify that pod is launched by KubernetesPodOperator
-        pod.metadata.labels.update(
-            {
-                "airflow_version": airflow_version.replace("+", "-"),
-                "airflow_kpo_in_cluster": str(self.hook.is_in_cluster),
-            }
-        )
+        pod.metadata.labels.update({"airflow_version": airflow_version.replace("+", "-")})
+        if not dry_run:
+            # self.hook.is_in_cluster instantiates a Kubernetes API client, which requires
+            # kube credentials/config to be available and must be skipped during a dry run.
+            pod.metadata.labels["airflow_kpo_in_cluster"] = str(self.hook.is_in_cluster)
         pod_mutation_hook(pod)
         return pod
 
@@ -1689,7 +1692,7 @@ class KubernetesPodOperator(BaseOperator):
         Does not include labels specific to the task instance (since there isn't
         one in a dry_run) and excludes all empty elements.
         """
-        pod = self.build_pod_request_obj()
+        pod = self.build_pod_request_obj(dry_run=True)
         print(yaml.dump(prune_dict(pod.to_dict(), mode="strict")))
 
     def process_duplicate_label_pods(self, pod_list: list[k8s.V1Pod]) -> k8s.V1Pod:
--- a/providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod.py
+++ b/providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod.py
@@ -472,6 +472,34 @@ class TestKubernetesPodOperator:
             "airflow_kpo_in_cluster": str(k.hook.is_in_cluster),
         }
 
+    @patch(HOOK_CLASS)
+    def test_build_pod_request_obj_dry_run_skips_live_kube_client(self, hook_mock):
+        """dry_run must not require a live Kubernetes API client (e.g. no kube config available)."""
+        type(hook_mock.return_value).is_in_cluster = mock.PropertyMock(
+            side_effect=RuntimeError("kube config not available")
+        )
+        k = KubernetesPodOperator(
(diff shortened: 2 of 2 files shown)


Miretpl:
There are couple of unrelated changes in the PR + conflics - could you resolve it?

Aaryan123456679:
Thanks for flagging this. The branch had two extra commits that were unrelated to this fix (a check_provider_yaml_files.py namespaced-provider crash fix and a Jira provider.yaml conn-field fix) — those were already merged into main separately via #70224 and #70261, which is what caused the conflicts. I dropped both commits and rebased the actual KubernetesPodOperator dry_run fix cleanly onto latest main. The branch now only contains the two files relevant to this PR (pod.py and its test) and is mergeable.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -1673,12 +1677,11 @@ def build_pod_request_obj(self, context: Context | None = None) -> k8s.V1Pod:
...
         # Add Airflow Version to the label
         # And a label to identify that pod is launched by KubernetesPodOperator
-        pod.metadata.labels.update(
-            {
-                "airflow_version": airflow_version.replace("+", "-"),
-                "airflow_kpo_in_cluster": str(self.hook.is_in_cluster),
-            }
-        )
+        pod.metadata.labels.update({"airflow_version": airflow_version.replace("+", "-")})
+        if not dry_run:
```
potiuk, line 1681: The side effect is that a dry run now produces a pod spec **missing the `airflow_kpo_in_cluster` label** — so the thing `dry_run()` prints is not quite the thing that will be submitted, which is the one property a dry run is supposed to have.

It's a defensible trade (a slightly incomplete spec beats an unusable command), but worth being deliberate: either note the omission in the `dry_run()` output so nobody diffs it against a real pod and gets confused, or set the label to a placeholder like `"unknown"` so the key is at least present in the rendered spec.

---

potiuk (approved):
Thanks — real usability bug. `dry_run()` existing to show what *would* be submitted shouldn't require working kube credentials, and `self.hook.is_in_cluster` instantiates an API client purely to compute a label value. Threading a keyword-only `dry_run` flag through `build_pod_request_obj` is the minimal way to fix it, and keyword-only keeps it backward compatible for the many subclasses that override or call this method.

The tests are the right shape: making `is_in_cluster` raise and asserting the dry-run path survives *and* the normal path still raises proves the flag actually gates the client instantiation, rather than just that dry_run happens to work.

One consequence worth deciding on inline.

---

PR #70237, opened 2026-07-22
Fix clearTaskInstances returning 500 instead of 422 on invalid body

SreeramaYeshwanthGowd:
`POST /dags/{dag_id}/clearTaskInstances` returns `500` instead of `422` when the request body is invalid. The `ClearTaskInstancesBody` validator raised `pydantic.ValidationError("...")`, which in Pydantic v2 isn't valid and re-raise a `TypeError`, and the API has no handler for it, so a bad request surfaces as a server error.

Raising `ValueError` lets Pydantic report it as a `422`, matching the sibling `PatchTaskInstanceBody` validator in the same file. Added a parametrized test covering each invalid body case.


bugraoz93 (commented):
Thanks for your first PR! 
I am wondering whether we could update this on the openapi auto exception generation/doc generation to include ValidationError too, otherwise we need to check each and every similar case as they should also have the same behavior

SreeramaYeshwanthGowd:
@bugraoz93 Thanks! Quick clarification on the mechanism: passing a plain string to Pydantic's `ValidationError` (`raise ValidationError("...")`) actually raises `TypeError`, not a validation error, because the class expects `line_errors`. Nothing handles that `TypeError`, so the response comes back as `500` instead of `422`. Since the real error is a `TypeError` and not a `ValidationError`, a central `ValidationError` handler would not catch these. Switching to `raise ValueError` fixes it, and FastAPI already adds `422` to the OpenAPI docs on its own (see `create_openapi_http_exception_doc`).

I also checked the tree, and the only other place using this pattern is `services/public/connections.py:162`, which is currently unreachable, so there is nothing else to fix here.

Would appreciate another look when you get a chance. Thanks again!

SreeramaYeshwanthGowd:
@bugraoz93 @ephraimbuddy  whenever you have a moment. Kindly review the PR. Thanks!

pierrejeambrun (approved):
LGTM, thanks.

SreeramaYeshwanthGowd:
@pierrejeambrun Thanks for the review. I will raise a follow up PR for core/src/airflow/api_fastapi/core_api/services/public/connections.py

pierrejeambrun:
> @pierrejeambrun Thanks for the review. I will raise a follow up PR for core/src/airflow/api_fastapi/core_api/services/public/connections.py

@SreeramaYeshwanthGowd I double checked after posting my suggestion and in `connection.py` the VallidationError is catch and reraised so I think we're fine, no need for an follow up PR.

PR #70238, opened 2026-07-22
Add securityContexts.disableDefaults flag for OpenShift SCC compatibility

### LINKED ISSUE #41630: Helm chart can't remove .securityContexts.runAsUser and securityContexts.fsGroup  for Openshift compatibility

1.15.0 (latest released)

### Apache Airflow version

2.9.3

### Kubernetes Version

v1.26.15+4818370

### Helm Chart configuration

```
securityContexts: {
  pod: {
    runAsUser: 
    fsGroup:  
  }
}
```

_No response_

### What happened

.securityContexts.pod.runAsUser and securityContexts.pod.fsGroup should allow us to remove the values entirely for openshift compatibility. Bitnami have done the same in their image https://github.com/bitnami/charts/blob/f7bd58e4b2842e0bf1bf2dcd0288beea98dd87a9/bitnami/airflow/values.yaml#L29

This is needed when users like myself do not have cluster admin rights to assign scc's to service accounts that can use any UID and GID. By default Openshift applies the restricted-v2 SCC, which doesn't allow any id for runAsUser and fsGroup.

Allow .runAsUser and fsGroup values to be omitted or allow .securityContexts block to be removed so Openshifts SCC policy will apply the defaults.

### How to reproduce

In an openshift cluster, try to deploy as a project admin.

### Anything else

_No response_

### Code of Conduct

Aaryan123456679:
OpenShift assigns `runAsUser`/`fsGroup` automatically via SCC and rejects pods that hard-code them, so users need to null these out in the rendered pod `securityContext`. The Helm chart's `values.schema.json` typed `securityContexts.pod.runAsUser`/`runAsGroup`/`fsGroup` as non-nullable integers (via the vendored `io.k8s.api.core.v1.PodSecurityContext` definition), so setting them to `null` failed schema validation before the templates ever rendered — the templates themselves already pass through explicit nulls correctly.

This introduces a new `AirflowPodSecurityContext` schema definition (not `io.k8s.*`-prefixed, so it survives `vendor_k8s_json_schema.py` regeneration) that allows `null` for these three fields, and points all 16 `securityContexts.pod` `$ref`s at it instead of the vendored definition.

closes: #41630

## Test plan

- [x] Added `test_main_pod_setting_runasuser_fsgroup_null` verifying `runAsUser`/`fsGroup` render as `None` when explicitly nulled
- [x] `breeze testing helm-tests --test-type security` — 188 passed
- [x] `breeze testing helm-tests --use-xdist` (full suite) — 2061 passed
- [x] Verified via `helm template` that the schema now accepts the null override and the rendered manifest's `spec.template.spec.securityContext` reflects the explicit nulls

---

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @Aaryan123456679** · by `@potiuk` · 2026-07-28 16:32 UTC
>
> Some review feedback from a maintainer is waiting on you:
> - There are **0 unresolved review threads** on this PR from a maintainer.
>
> **The ball is in your court** — you've been assigned to this PR. Push a fix or reply in each thread explaining why the feedback does not apply, then mark them resolved and ping the reviewer (a maintainer) for a final look.
>
>
> **Note:** your branch is **110 commits behind `main`** — please rebase and push again to get up-to-date CI results.
>


chart/values.schema.json
```diff
@@ -104,7 +104,7 @@
                 "pod": {
                     "description": "Default pod security context definition. The values in this parameter will be used when `securityContexts` is not defined for specific Pods.",
                     "type": "object",
-                    "$ref": "#/definitions/io.k8s.api.core.v1.PodSecurityContext",
+                    "$ref": "#/definitions/AirflowPodSecurityContext",
```
bugraoz93: Thanks for the PR! Would we need this change? From the doc you create I understood as it should be null as default which we also migrated toward different naming in definitions than others. I don’t think the problem should be in definition place as long as we set to null?

Aaryan123456679: The `default` here doesn't relax the type constraint — it's just documentation for tooling/UI, not a JSON Schema validation rule.

The actual problem is that the vendored `io.k8s.api.core.v1.PodSecurityContext` definition types `runAsUser`/`runAsGroup`/`fsGroup` as `"type": "integer"` only (non-nullable). So even with `default: {}` at the parent `pod` object, a user who explicitly sets `securityContexts.pod.runAsUser: null` in their own values file (exactly what's needed for OpenShift, since SCC assigns the UID/GID automatically) fails schema validation before the templates ever render.

We can't just relax the vendored definition in place, because `vendor_k8s_json_schema.py` regenerates every `io.k8s.*`-prefixed definition from the upstream Kubernetes OpenAPI spec — a hand edit there would be silently reverted on the next regen. That's why this introduces `AirflowPodSecurityContext`: identical to the vendored definition, except `runAsUser`/`runAsGroup`/`fsGroup` allow `["integer", "null"]`, and it's named outside the `io.k8s.*` namespace so it's untouched by regeneration.

Happy to adjust the approach if there's a lighter-weight way to express this that I'm missing — let me know if you had something else in mind.

Miretpl (changes requested):
I would see it more like adding a flag which would make all of the default options not added. Currently, in most cases (if not in all), if `securityContexts.*` is empty, in the helper function logic we are adding some values. I think they should not be added there, and if we want to have some default values in the chart for them, they should be in the `values.yaml` file (version 2.0).

In this PR, I would add a flag for not adding these default values in the helper function when, e.g. `securityContexts.pod` is `{}` (default in `values.yaml` file). This would be, IMHO, much more intuitive for the users; in the end will require less logic to handle and maintain and should be cleaner too.

WDYT?

Aaryan123456679:
Thanks for the suggestion — that's a cleaner approach. I've reworked the PR accordingly:

- Added a `securityContexts.disableDefaults` flag (default `false`). When set, the chart no longer fills in the default `runAsUser`/`fsGroup` (pod) or `runAsUser` (container) values wherever `securityContexts.pod` / `securityContexts.containers` (or the per-component overrides) are left empty — the `securityContext` key is omitted from the rendered manifest entirely, rather than emitting an explicit `null`.
- Reverted the `values.schema.json` change back to the vendored `io.k8s.api.core.v1.PodSecurityContext` ref (dropped the earlier `AirflowPodSecurityContext` nullable-fields workaround), since the flag makes that special-casing unnecessary.
- Explicit values in `securityContexts.pod` / `securityContexts.containers` (global or per-component) still take priority over `disableDefaults`, so existing fine-grained overrides are unaffected.
- Updated the newsfragment, `production-guide.rst`, and `test_security_context.py` to describe the flag instead of the null-based workaround.

Let me know if this matches what you had in mind, or if you'd adjust the flag's scope or naming.

chart/templates/_helpers.yaml
```diff
@@ -914,6 +923,9 @@ The template looks for `runAsUser` and `fsGroup` specifically, any other paramet
 Values are not accumulated meaning that if runAsUser is set to 10 in <node>.securityContexts.pod,
 any extra values set to securityContexts or uid+gid will be ignored.
 
+If securityContexts.disableDefaults is true and no explicit runAsUser/fsGroup is set anywhere,
+this template renders an empty string, since there is no meaningful id to chown to.
```
Miretpl:
```suggestion
this template renders an empty string, since there is no id to chown to.
```

chart/templates/_helpers.yaml
```diff
@@ -890,15 +896,18 @@ If no value is passed for <node>.securityContexts.container, defaults to UID in
     +-----------------------------------+      +------------+
 
 The template can be called like so:
-  include "localContainerSecurityContext" .Values.statsd
+  include "localContainerSecurityContext" (list .Values.dags.gitSync .Values)
```
Miretpl, line 928: Could you make it unified with other invocations of securityContext-related helper methods? In one place, we have `list .Values .Values.scheduler` order and in the other `list .Values.scheduler .Values` order. It would be easier for the developers and maintainers to have it in one order.

chart/newsfragments/70238.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added ``securityContexts.disableDefaults`` to the Helm chart. When set, the chart no longer fills in default ``runAsUser``/``fsGroup`` (pod) or ``runAsUser`` (container) values, letting platforms such as OpenShift assign them automatically via SCC instead of the chart hard-coding them.
```
Miretpl:
```suggestion
Added ``securityContexts.disableDefaults`` (default ``false``) to the Helm chart. When set, the chart no longer fills in the default ``runAsUser``/``fsGroup`` (pod) or ``runAsUser`` (container) values, allowing platforms such as OpenShift to assign them automatically via SCC instead of the chart hard-coding them.

Value ``false`` for ``securityContexts.disableDefaults`` is deprecated and will be ``true`` by default in the next major release. Please adjust your configuration accordingly.
```
and I would change the type from `bugfix` to `significant` and deprecate the value `false` for this parameter (with additional change in NOTES and values files with the `deprecated` information).

chart/values.schema.json
```diff
@@ -126,6 +126,12 @@
                             "allowPrivilegeEscalation": false
                         }
                     ]
+                },
+                "disableDefaults": {
+                    "description": "If true, the chart will not set any default `runAsUser`/`fsGroup` (pod) or `runAsUser` (container) values when `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component overrides) are left empty. Useful on platforms such as OpenShift, where the cluster assigns these values automatically via SCC and rejects pods/containers that hard-code them.",
```
```suggestion
                    "description": "If ``true``, the chart will not set any default ``securityContext`` values when ``securityContexts.pod`` / ``securityContexts.containers`` (or the equivalent per-component overrides) are left empty.",
```
+ deprecated message. Field description should be rather constraints to the only required information.

chart/values.yaml
```diff
@@ -47,6 +47,12 @@ gid: 0
 securityContexts:
   pod: {}
   containers: {}
+  # If true, the chart will not set any default `runAsUser`/`fsGroup` (pod) or `runAsUser`
+  # (container) values when securityContexts.pod / securityContexts.containers (or the
+  # equivalent per-component overrides) are left empty. Useful on platforms such as
+  # OpenShift, where the cluster assigns these values automatically via SCC and rejects
+  # pods/containers that hard-code them.
```
Miretpl: Same comment as in schema file.

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,120 @@ def test_main_pod_setting(self):
         for doc in docs[1:]:
             assert ctx_value == jmespath.search("spec.template.spec.securityContext", doc)
 
+    def test_disable_defaults_pod_and_container(self):
```
Miretpl, line 425: Could you add tests for:
1. pod template file
2. overrides when component level option is set
?

Miretpl (changes requested):

chart/templates/_helpers.yaml
Aaryan123456679: Updated the wording as suggested in d971b4fec7.

chart/templates/_helpers.yaml
Aaryan123456679: Done in d971b4fec7 — switched `airflowPodSecurityContextsIds` to the same `(list <node> .Values)` calling convention used by the other security-context helpers, and updated its sole caller in worker-deployment.yaml accordingly.

chart/newsfragments/70238.bugfix.rst
Aaryan123456679: Updated in d971b4fec7 — reworded per your suggestion, changed to `.significant.rst`, and added a note that the `false` default is deprecated and will flip to `true` in a future release.

chart/values.schema.json
Aaryan123456679: Updated in d971b4fec7 to just the required-information description plus a note that the `false` default is deprecated.

chart/values.yaml
Aaryan123456679: Updated in d971b4fec7 to match the schema file wording.

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Added both in d971b4fec7: `test_pod_security_context_disable_defaults` / `test_container_security_context_disable_defaults` / `test_pod_security_context_set_overrides_disable_defaults` in test_pod_template_file.py for the pod-template-file path, and `test_disable_defaults_component_level_override_still_applied` here for the component-level override case.

Aaryan123456679:
@Miretpl All the review feedback above has been addressed in d971b4fec7:

- Fixed the wording in `_helpers.yaml` and unified `airflowPodSecurityContextsIds`'s argument order with the other security-context helpers (updating its caller in `worker-deployment.yaml` accordingly).
- Reworded the newsfragment, switched it to `.significant.rst`, and added a note that the `disableDefaults: false` default is deprecated and will flip to `true` in a future release — mirrored in `NOTES.txt`, `values.yaml`, and `values.schema.json`.
- Added test coverage for `disableDefaults` on the pod-template-file path (`test_pod_template_file.py`) and for a component-level `securityContexts` override taking priority over a global `disableDefaults: true` (`test_security_context.py`).

All Helm test suites (`security`, `airflow_aux`) pass. Could you take another look when you have a chance?

chart/values.schema.json
```diff
@@ -126,6 +126,12 @@
                             "allowPrivilegeEscalation": false
                         }
                     ]
+                },
+                "disableDefaults": {
+                    "description": "If `true`, the chart will not set any default `securityContext` values when `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component overrides) are left empty. (Default `false` is deprecated and will change to `true` in a future release.)",
```
```suggestion
                    "description": "If `true`, the chart will not set any default `securityContext` values when `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component overrides) are left empty. (Default `false` is deprecated and will change to `true` in a future release)",
```

chart/values.yaml
```diff
@@ -47,6 +47,11 @@ gid: 0
 securityContexts:
   pod: {}
   containers: {}
+  # If `true`, the chart will not set any default `securityContext` values when
+  # securityContexts.pod / securityContexts.containers (or the equivalent per-component
+  # overrides) are left empty. (Default `false` is deprecated and will change to `true`
+  # in a future release.)
```
```suggestion
  # If `true`, the chart will not set any default `securityContext` values when
  # `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component
  # overrides) are left empty. (Default `false` is deprecated and will change to `true`
  # in a future release)
```

chart/tests/helm_tests/security/test_security_context.py
Miretpl: I don't see tests for point 2.

Miretpl (changes requested):

Miretpl:
> @Miretpl All the review feedback above has been addressed in https://github.com/apache/airflow/commit/d971b4fec75730eee4a103f21432ec870a29c905:

Just re-request the review - no need to mention people if only a couple of days have passed. Someone sooner or later will take a look. If it is stale for around 2 weeks, I think it is a good time period to try to ping any maintainer for the review.

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: The test for this already exists — `test_disable_defaults_component_level_override_still_applied` in `chart/tests/helm_tests/security/test_security_context.py` (line 539, added in d971b4fec7). It sets `securityContexts.disableDefaults: true` globally plus a component-level `securityContexts.pod` override on `scheduler` and `workers.celery`, then asserts the component-level override wins over the global setting for both components.

Miretpl: We modify all security contexts in **ALL** components, and in tests we have coverage only for two: `workers.celery` and `scheduler`. We should have coverage for all.

Miretpl (changes requested):

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Extended `test_disable_defaults_component_level_override_still_applied` into a parametrized test covering all 13 components that `disableDefaults` touches — `scheduler`, `apiServer`, `dagProcessor`, `triggerer`, `workers.celery`, `flower`, `statsd`, `pgbouncer`, `redis`, `createUserJob`, `migrateDatabaseJob`, `cleanup`, and `databaseCleanup` — matching every template this PR's diff modifies. `databaseCleanup` wasn't covered by any other test in the file either, so this closes that gap too.

Also split the volume-permissions init-container assertion into its own `test_disable_defaults_workers_volume_permissions_init_container_still_added` test, since it checks a different behavior (that `disableDefaults` doesn't suppress init-container injection) from the component-override-priority behavior the parametrized test now covers.

Full security test suite: 205 passed, 0 failed.

chart/values.schema.json
Miretpl: Not resolved.

chart/values.yaml
Miretpl: Not resolved - please, do not resolve comments which were not addressed by you and if you believe that something shouldn't be done, you can put a comment under suggestion to explain why do you think that,

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+        """
+        securityContexts.disableDefaults suppresses the chart's hard-coded runAsUser/fsGroup
+        and allowPrivilegeEscalation/capabilities defaults, e.g. for OpenShift SCC compatibility.
+        """
```
Miretpl:
```suggestion
```
I think that there is no value in having that comment next to test case.

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
         for doc in docs[1:]:
             assert ctx_value == jmespath.search("spec.template.spec.securityContext", doc)
 
+    def test_disable_defaults_pod_and_container(self):
```
Miretpl, line 425: We could separate job from deployments.
*(valid for more than only this test case)*

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+            ],
+        )
+
+        assert jmespath.search("spec.jobTemplate.spec.template.spec.securityContext", docs[0]) is None
+        assert (
+            jmespath.search("spec.jobTemplate.spec.template.spec.containers[0].securityContext", docs[0])
+            is None
+        )
+
+        for doc in docs[1:]:
```
Miretpl: Could you move it to `jmespath.search`?
*(valid for more than only this test case)*

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+                        "persistence": {"enabled": True, "fixPermissions": True},
+                    }
+                },
+            },
+            show_only=["templates/workers/worker-deployment.yaml"],
+        )
+
+        init_container_names = [
+            c["name"] for c in jmespath.search("spec.template.spec.initContainers", docs[0])
+        ]
```
Miretpl: Could you move this `for` logic to `jmespath`?
*(valid for more than only this test case)*

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+        assert "volume-permissions" not in init_container_names
+
+    def test_disable_defaults_explicit_override_still_applied(self):
+        """Explicit securityContexts overrides take priority over disableDefaults."""
```
Miretpl:
```suggestion
```
IMO, the name of the test means the same.

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+        """A component-level securityContexts override takes priority over global disableDefaults."""
+        docs = render_chart(
+            values={"securityContexts": {"disableDefaults": True}, **component_values},
+            show_only=[show_only],
+        )
+
+        assert jmespath.search(security_context_path, docs[0]) == {"runAsUser": 8000}
+
+    def test_disable_defaults_workers_volume_permissions_init_container_still_added(self):
+        """disableDefaults must not suppress the volume-permissions init container injection."""
```
Miretpl:
```suggestion
```
Name of the test case means that really.

chart/tests/helm_tests/security/test_security_context.py
```diff
@@ -422,6 +422,239 @@ def test_main_pod_setting(self):
...
+            "createUserJob",
+            "migrateDatabaseJob",
+            "cleanup",
+            "databaseCleanup",
+        ],
+    )
+    def test_disable_defaults_component_level_override_still_applied(
+        self, component_values, show_only, security_context_path
+    ):
+        """A component-level securityContexts override takes priority over global disableDefaults."""
```
Miretpl:
```suggestion
```
Name of the test case means that really.

Miretpl (changes requested):

chart/tests/helm_tests/security/test_security_context.py
Miretpl: I don't think there is a need to restate things which were done in the code. If some comment is done, you can simply write done and save the reviewer some time by not reading the same thing twice, especially if it is copied and pasted from the AI model.

chart/values.schema.json
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/values.yaml
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Done.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/tests/helm_tests/security/test_security_context.py
Aaryan123456679: Understood — will keep replies terse going forward.

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

chart/values.schema.json
```diff
@@ -126,6 +126,12 @@
                             "allowPrivilegeEscalation": false
                         }
                     ]
+                },
+                "disableDefaults": {
+                    "description": "If `true`, the chart will not set any default `securityContext` values when `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component overrides) are left empty. (Default `false` is deprecated and will change to `true` in a future release)",
```
```suggestion
                    "description": "If `true`, the chart will not set any default `securityContext` values when `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component overrides) are left empty (default `false` is deprecated and will change to `true` in a future release).",
```

chart/values.schema.json
```diff
@@ -10054,6 +10060,82 @@
         }
     },
     "definitions": {
+        "AirflowPodSecurityContext": {
```
Miretpl: Hmm, I think it was on the first iteration, was deleted after that, and suddenly it came back. It is not used anywhere - could you delete it?

chart/values.yaml
```diff
@@ -47,6 +47,11 @@ gid: 0
 securityContexts:
   pod: {}
   containers: {}
+  # If `true`, the chart will not set any default `securityContext` values when
+  # `securityContexts.pod` / `securityContexts.containers` (or the equivalent per-component
+  # overrides) are left empty. (Default `false` is deprecated and will change to `true`
+  # in a future release)
```
```suggestion
  # overrides) are left empty (default `false` is deprecated and will change to `true`
  # in a future release).
```

Miretpl (changes requested):

chart/values.schema.json
Aaryan123456679: Thanks for the headsup, I have done the necessary changes.

Miretpl (approved):

Miretpl:
@Aaryan123456679 could you do manual backport of it?

Aaryan123456679:
Raised another PR for this.

PR #70240, opened 2026-07-22
Fix config update --option/--ignore-option never matching options

SreeramaYeshwanthGowd:
`airflow config update` filtered changes by comparing the composed `section.option` key against the bare option name the user passes, so `--option` and `--ignore-option` never matched and silently did nothing. The sibling `airflow config lint` already matches on the bare option name, so this makes `update` consistent with it.

 

---

---


uranusjr:
This needs a bugfix newsfragment. Otherwise I think this is good.

SreeramaYeshwanthGowd:
Thanks @uranusjr! Added the bugfix newsfragment

potiuk (approved):
Thanks — good catch, and the fix is right. `update_config` was building `full_key = f"{section}.{option}"` and comparing it against whatever the user passed to `--option` / `--ignore-option`, so a bare option name could never match and the filters silently did nothing.

I checked the intended interface against the sibling `lint_config` command, whose docstring gives `--option check_slas`, `--ignore-option smtp_user` and `--section smtp --option smtp_user` — bare option names throughout. So comparing against `conf_option` makes `update` consistent with `lint`, rather than inventing a new convention. The CLI help ("The option name(s) to update") says the same.

Parametrizing the test across both `--option` and `--ignore-option` with a present/absent key pair is a nice way to prove the filter works in both directions, and the newsfragment is warranted since this is a user-visible CLI fix.

Nothing further from me.

---

PR #70246, opened 2026-07-22
Exclude uv/pip installer versions from generated constraints files

### LINKED ISSUE #70231: uv version discrepancy  between the constraints and the container

Airflow Core

### Apache Airflow version

3.3.0

In the container `apache/airflow:slim-3.3.0-python3.14`

uv --version ->  0.11.21

but in the https://raw.githubusercontent.com/apache/airflow/constraints-3.3.0/constraints-3.14.txt

uv==0.11.26

### Code of Conduct

bramhanandlingala:
closes: #70231

### What

`uv` and `pip` versions were being written into generated constraints
files via `pip freeze`, even though they're build tools, not Airflow
dependencies.

### Fix

Added `"uv=="` and `"pip=="` to the exclusion list in
`freeze_distributions_to_file()` (`scripts/in_container/run_generate_constraints.py`),
matching the same exclusion already used in `uninstall_all_packages()`
in the same file.

### Why this fixes the issue

Constraints files and published Docker images are refreshed on
different, independent cadences — a rebuild of one doesn't
necessarily happen alongside the other. Capturing `uv`'s version in
constraints caused it to drift from whatever version actually shipped
in the image, exactly as reported in #70231. Excluding it removes
that class of mismatch at the source.

- [X] Yes — Claude


raphaelauv:
Hey , thanks for the fix 

but about `  published Docker images are built once and never rebuilt` , this is not true, it happened several times 

example : docker image of 3.3.0 was re-built and re-publish

bramhanandlingala:
@raphaelauv 

Good catch, thanks — updated the description. Removed the "never rebuilt" claim and reworded it to focus on the actual point: constraints and image rebuilds happen on independent cadences, so there's no guarantee uv's captured version matches what ships in a given image.

shahar1 (changes requested):

potiuk:
This is not needed - see issue (closed) for explanation.

PR #70248, opened 2026-07-22
Fix Trigger Again showing empty config for the selected run

pierrejeambrun:
Re-triggering a Dag with the config of a selected run (the "Trigger DAG w/ config" / "Trigger Again with config" menu item) left the configuration empty — the JSON editor showed `{}` and the selected run's `conf` was not carried over.

The prefill only pushed the run's `conf` into the param store when the Dag's declared params were already present. For Dags with no declared params (or before the params query resolved), that branch was skipped, so the param store kept its default empty config and the sync effect reset the form back to `{}`, dropping the run's configuration. The conf is now always propagated to the param store on prefill; seeding the initial params (for stable section ordering) still waits until they are available.

Before / after (config carried over from the selected run):

## Before
https://github.com/user-attachments/assets/83555e55-c417-488f-b6ea-75560c13e6e0

## After
https://github.com/user-attachments/assets/ee7befab-6564-422a-83bf-6f4aacfe4644

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/TriggerDag/TriggerDAGForm.test.tsx
+++ b/airflow-core/src/airflow/ui/src/components/TriggerDag/TriggerDAGForm.test.tsx
@@ -18,7 +18,7 @@
  */
 import "@testing-library/jest-dom";
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { Wrapper } from "src/utils/Wrapper";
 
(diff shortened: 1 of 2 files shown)


ashb:
Does this also work if "arbitrary" json conf is provided?

ashb (approved):

pierrejeambrun:
> Does this also work if "arbitrary" json conf is provided?

Yes, an arbitrary config is propagated via the 'trigger again with config' button. It wasn't clear in my previous video because it looks like the dag had params, but the dag had no conf and it was added manually when starting the previous run. 

Here is a better example

https://github.com/user-attachments/assets/1a0965c4-65d1-414e-8abc-ded88feaccfb

PR #70249, opened 2026-07-22
airflowctl: fix datetime.datetime CLI parameters rejecting all input

### LINKED ISSUE #70232: airflowctl: datetime.datetime-typed CLI parameters don't accept any value

Airflow Core

### Apache Airflow version

airflow 3.2.2 / airflow-ctl 0.1.5 (also present on `main`)

`airflowctl dagrun list` (and any other auto-generated CLI command with a `datetime.datetime`-typed parameter, e.g. `--start-date`/`--end-date`) rejects every possible input string. No date format — plain date, full ISO-8601, anything — will parse.

**Reproduction:**
```bash
airflowctl dagrun list -e <env> --start-date "2026-07-01" --limit 1
# argument --start-date: invalid datetime value: '2026-07-01'

airflowctl dagrun list -e <env> --start-date "2026-07-01T00:00:00" --limit 1
# argument --start-date: invalid datetime value: '2026-07-01T00:00:00'
```
**Root cause:**

In `airflow-ctl/src/airflowctl/ctl/cli_config.py`, `_python_type_from_string` builds the argparse `type=` callable for CLI arguments straight from the OpenAPI-derived operation signatures:

```python
mapping: dict[str, type | Callable] = {
    "int": int,
    "float": float,
    "bool": bool,
    "str": str,
    "bytes": bytes,
    "list": list,
    "dict": json_dict_type,
    "tuple": tuple,
    "set": set,
    "datetime.datetime": datetime.datetime,
    "dict[str, typing.Any]": json_dict_type,
}
```

For any parameter annotated `datetime.datetime`, this passes the bare `datetime.datetime` class to argparse as `type=`. argparse then calls `type(value)` on
the raw CLI string, i.e. `datetime.datetime("2026-07-01")` — invoetime.datetime(year, month, day, ...)`), not a parser. Since theconstructor expects positional ints, passing a single string always raises a `TypeError`, which argparse reports as `invalid datetime value: '...'` regardless
of the input.

Notably, the exact same class of bug existed for `dict`-typed parxed by mapping to `json_dict_type` (a real parser using`json.loads`) instead of the bare `dict` constructor. The `datetime.datetime` entry was missed and still uses the bare type.

`datetime.datetime` should map to an actual parsing function, not the bare class, e.g.:

```python
def iso_datetime_type(val: str) -> datetime.datetime:
    """Parse an ISO-8601 datetime string."""
    try:
        return datetime.datetime.fromisoformat(val)
    except ValueError as e:
        raise argparse.ArgumentTypeError(f"invalid datetime value: {val!r}") from e
```

and:

```python
"datetime.datetime": iso_datetime_type,
```

This mirrors the existing fix pattern already used for `dict` → `e mapping.

### Operating System

macOS (also confirmed against a production Airflow instance, running Linux / Debian distroless)

### Deployment

Other

_No response_

Installed via `uv tool install apache-airflow-ctl`. This bug report is explicitly about the airflowctl tool.

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

Not Applicable

Not Applicable

### Anything else?

Confirmed present at:
- Tag `airflow-ctl/0.1.5` (commit `30e72a3`) — `airflow-ctl/src/airflowctl/ctl/cli_config.py:517`
- `main` (as of 2026-07-22) — same file, `cli_config.py:549`,  unchanged

Affects every argparse-generated CLI command whose underlying operation signature has a `datetime.datetime`-typed parameter (e.g. `dagrun list --start-date/--end-date`), not just this one command.

### Code of Conduct

bramhanandlingala:
closes: #70232

### What
`datetime.datetime`-typed CLI parameters in `airflowctl` (e.g.
`dagrun list --start-date/--end-date`) rejected every input value.

### Root cause
`_python_type_from_string()` in `cli_config.py` mapped
`datetime.datetime` to the bare class as the argparse `type=`
callable. argparse calls `type(value)` on the raw CLI string, so it
tried `datetime.datetime("2026-07-01")` — but that constructor expects
positional `(year, month, day, ...)` ints, not a string, so it always
raised `TypeError`, regardless of the input format.

### Fix
Added `iso_datetime_type()`, a real parser using
`datetime.datetime.fromisoformat()`, and mapped `datetime.datetime` to
it instead of the bare class. This mirrors the existing fix already
applied for `dict` → `json_dict_type`.

- [X] Yes — Claude


bramhanandlingala:
@ bugraoz93, @dheerajturaga, @henry3260 and @potiuk  @kaxil 
request you please review and approve for merge this fix

bramhanandlingala:
@ bugraoz93, @dheerajturaga, @henry3260 and @potiuk @kaxil @Lee-W @shahar1 

Did I miss anyone  in the list,  please advise
after a while this code fix would be stale ,  trying to get approval as soon as possible,  any approvals or guidance is motivation to improve my work, please help

airflow-ctl/tests/airflow_ctl/ctl/test_cli_config.py
```diff
@@ -364,6 +365,37 @@ def test_json_dict_type_rejects_non_object_json(self, value):
...
+        ],
+    )
+    def test_iso_datetime_type_parses_iso_string(self, value, expected):
+        """An ISO-8601 datetime string (date-only or full) is parsed into a datetime.
+
+        Regression test for https://github.com/apache/airflow/issues/70232: previously the
+        bare ``datetime.datetime`` class was used as the argparse ``type=`` callable, so
+        argparse called ``datetime.datetime(value)`` on the raw string, which always raised
+        a ``TypeError`` regardless of the input.
+        """
```
henry3260:
```suggestion
        """An ISO-8601 datetime string (date-only or full) is parsed into a datetime."""
```

airflow-ctl/src/airflowctl/ctl/cli_config.py
```diff
@@ -566,7 +576,7 @@ def _python_type_from_string(type_name: str | type) -> type | Callable:
             "dict": json_dict_type,
             "tuple": tuple,
             "set": set,
-            "datetime.datetime": datetime.datetime,
+            "datetime.datetime": iso_datetime_type,
```
henry3260, line 579: Could we add a test like this?
```
def test_command_factory_wires_iso_parser_to_datetime_params(self):
    """A generated datetime CLI arg parses an ISO date end to end."""
    command_factory = CommandFactory()
    dagrun_list_args = []
    for group in command_factory.group_commands:
        if group.name != "dagrun":
            continue
        for sub in group.subcommands:
            if sub.name == "list":
                dagrun_list_args = list(sub.args)
                break
    start_date_arg = next(a for a in dagrun_list_args if a.flags == ("--start-date",))
    assert start_date_arg.kwargs["type"]("2026-07-01") == datetime.datetime(2026, 7, 1)
```

henry3260 (commented):
Thanks for the fix!

shahar1 (changes requested):

bramhanandlingala:
Hi @henry3260 

Thanks for the detailed review and the helpful suggestion! I've updated the PR by simplifying the docstring and adding the end-to-end regression test to verify that CommandFactory wires iso_datetime_type for datetime CLI parameters. I'd appreciate it if you could take another look when you have a chance.

airflow-ctl/src/airflowctl/ctl/cli_config.py
```diff
@@ -637,7 +647,7 @@ def _python_type_from_string(type_name: str | type) -> type | Callable:
             "tuple": tuple,
             "set": set,
             "datetime.date": datetime.date,
```
potiuk: Same bug, one line up: `datetime.date("2026-07-01")` raises `TypeError` just like `datetime.datetime` did. Latent today because the live date params are served by a hand-written command, but the next generated `datetime.date` parameter inherits it. Worth fixing in the same PR:

```suggestion
            "datetime.date": iso_date_type,
            "datetime.datetime": iso_datetime_type,
```

…with an `iso_date_type()` alongside `iso_datetime_type()` using `datetime.date.fromisoformat()`.

airflow-ctl/src/airflowctl/ctl/cli_config.py
```diff
@@ -205,6 +205,16 @@ def json_dict_type(val: str | dict[str, Any]) -> dict[str, Any]:
     return parsed
 
 
+def iso_datetime_type(val: str | datetime.datetime) -> datetime.datetime:
+    """Parse an ISO-8601 datetime argument."""
+    if isinstance(val, datetime.datetime):
+        return val
+    try:
+        return datetime.datetime.fromisoformat(val)
```
potiuk, line 215: `fromisoformat()` only accepts a trailing `Z` from Python 3.11 onwards, and `airflow-ctl` supports `>=3.10`. So `--start-date 2026-07-01T12:00:00Z` parses on 3.11+ and raises on 3.10 — same command, same server, different interpreter. Since `Z` is the form the API itself emits, users will paste it back in:

```suggestion
        return datetime.datetime.fromisoformat(val.replace("Z", "+00:00") if val.endswith("Z") else val)
```

airflow-ctl/tests/airflow_ctl/ctl/test_cli_config.py
```diff
@@ -364,6 +366,31 @@ def test_json_dict_type_rejects_non_object_json(self, value):
         with pytest.raises(argparse.ArgumentTypeError, match="expected JSON object"):
             json_dict_type(value)
 
+    def test_iso_datetime_type_returns_datetime_input_unchanged(self):
+        """A datetime.datetime input is returned as-is without re-parsing."""
+        import datetime
```
potiuk: `datetime` is imported at module scope (line 21) by this same PR, so this local import can go.

potiuk (commented):
The fix itself is right. `argparse` calls `type(value)` on the raw string, so `datetime.datetime("2026-07-01")` could never work regardless of format, and routing through a real parser is the correct shape — following the `dict` → `json_dict_type` precedent already in that map rather than inventing a new one. The end-to-end wiring test that was requested in the earlier review is in place too, which is the test that would actually have caught the original bug.

Three small things, none of them blocking the approach.

### `datetime.date` immediately above has the identical bug

`"datetime.date": datetime.date` maps to the bare class, so `datetime.date("2026-07-01")` raises `TypeError` in exactly the same way. It's latent right now — the live date parameters (`partition_date_gte` / `partition_date_lte` in `operations.py`) are served by a hand-written command with its own `_parse_partition_date` — but the next generated `datetime.date` parameter walks straight into the bug this PR just fixed. Closing it costs a few lines while the file is open.

### `fromisoformat()` and the `Z` suffix differ by Python version

`airflow-ctl` supports `>=3.10`, and `datetime.fromisoformat()` only learned to accept a trailing `Z` in 3.11. So `--start-date 2026-07-01T12:00:00Z` parses on 3.11+ and fails on 3.10, for the same command against the same server. Normalising a trailing `Z` to `+00:00` before parsing removes that split, and `Z` is the form the API itself emits, so users will paste it back in.

### Smaller observations

- `tests/airflow_ctl/ctl/test_cli_config.py:371` — a function-local `import datetime`, while the same PR adds `import datetime` at module scope (line 21). Imports belong at the top of the file; the local one can go.

On the review state: the change request on this PR points at the gen-AI disclosure discussion rather than at the code, and the body now carries a disclosure paragraph. That block is `shahar1`'s to lift once they're satisfied — I'm not approving around it.

---

>

---

bramhanandlingala:
Thanks for catching that, @potiuk — removed the local import in [9a4bba9](https://github.com/bramhanandlingala/airflow/commit/9a4bba9). Also added the iso_date_type() fix and the trailing-Z handling you flagged in the other two comments, along with test coverage for both. Appreciate you taking the time to review this in detail.

potiuk (commented):
Thanks — all three points from my last round are in: `iso_date_type` for `datetime.date`, the trailing-`Z` normalisation with a test, and the local import removed. Together with the end-to-end wiring test, the code looks good to me.

One optional nit: `datetime.datetime` is a subclass of `datetime.date`, so `iso_date_type(some_datetime)` returns the datetime unchanged rather than a `date`. argparse only ever passes strings here, so it's harmless — `if type(val) is datetime.date` would make it exact if you want.

The remaining gate is the outstanding change request about the gen-AI disclosure, which is for its reviewer to lift. I'm not approving around it. The branch is also about 6 weeks old, so it'll need a rebase to get a fresh CI run before merge.

---

>

---

PR #69140, opened 2026-06-29
Honor async hook subclass overrides in get_async_connection

anxkhn:
`get_async_connection` resolved the connection through the literal `BaseHook`
class, so a hook that overrides `aget_connection`/`get_connection` was ignored in
the triggerer even though the sync path honors it on the worker.

The sync path dispatches through the instance: provider hooks call
`self.get_connection(...)` (e.g. `HttpHook`), so a subclass override wins. The
async path called the module helper, which hard-coded `BaseHook.aget_connection`
and only took a `conn_id`, so it could never see a subclass override. A user who
subclasses `HttpAsyncHook` and overrides `aget_connection` to inject
host/login/etc. gets correct fields on the worker but empty fields in the
triggerer.

This adds an optional `hook` parameter so the helper dispatches through the
calling hook (defaulting to `BaseHook`, so all existing callers are unchanged),
and passes the hook from `HttpAsyncHook`. Custom subclasses now behave the same
in the triggerer as on the worker.

This reopens the fix from #67440, which was closed only as inactive-draft
housekeeping (not on its merits); this PR is rebased on the latest `main` and
ready for review.

closes: #66845

---

##### Was generative AI tooling used to co-author this PR?

- [ ] Yes (please specify the tool below)

Issue #54079, opened 2025-08-04
Remove/Replace DSSKey references from Airflow


amoghrajesh:
### Body

As seen from: https://github.com/apache/airflow/pull/54078, which is a short term fix to workaround DSSKey
1. Remove DSSKey references entirely in the codebase
2. Bump minimum to paramiko 4.0+
3. Provide migration guide for users with DSA keys
4. Bump types-paramiko version that was limited in https://github.com/apache/airflow/pull/54173

Migration that could be required: 

Users would need to:
Generate other keys to be used in their connections (RSA, ECDSA, or Ed25519 keys) and update their Airflow connections to use those

### Committer

- [x] I acknowledge that I am a maintainer/committer of the Apache Airflow project.

joshuabvarghese:
I am interested in doing it

amoghrajesh:
Go for it @joshuabvarghese. Feel free to run through your approach on this if you aren't sure.
cc @Lee-W since you expressed interest too.

Lee-W:
@joshuabvarghese Feel free to reach out to me if you need PR review or clarification on this issue 🙂

developer-guy:
are there any updates on this 👀

Lee-W:
open PR WIP https://github.com/apache/airflow/pull/54165

pkhanpara:
Until this merges the workaround is to use env var  ` _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-apache-airflow-providers-ssh paramiko>=2.9.0,<4.0.0}`

rawwar:
I'm picking this up

PR #70256, opened 2026-07-22
Prepare providers release 2026-07-22

shahar1:
Regular provider release wave for 2026-07-22. Version bumps and changelog entries were classified per the `prepare-providers-documentation` workflow; all template-generated files were regenerated by `breeze release-management prepare-provider-documentation --reapply-templates-only`.

**Summary (50 providers changed, 42 skipped as tooling/template-only):**

- **Major (4):** microsoft.azure 14.0.0 (azure-mgmt-datafactory floor 2.0→10.0), ssh 6.0.0 / sftp 6.0.0 (paramiko 4, DSS keys removed), airbyte 6.0.0 (airbyte-api 0.x→1.x, `requests` no longer installed)
- **Minor (14):** amazon 9.33.0, google 22.3.0, cncf.kubernetes 10.20.0, apache.kafka 1.15.0, celery 3.23.0, common.ai 0.7.0, common.compat 1.17.0, databricks 7.18.0, edge3 4.2.0, openlineage 2.20.0, akeyless 0.3.0, anthropic 0.2.0, apache.hive 9.6.0, apache.spark 6.3.0
- **Patch (14):** http, fab, keycloak, openai, hashicorp, teradata, weaviate, docker, elasticsearch, opensearch, pgvector, clickhousedb, common.sql, atlassian.jira 3.3.5
- **Incremental (2):** snowflake 6.15.0 and postgres 7.0.0 — previously-prepared but unreleased versions; new commits folded into the existing sections rather than re-bumping
- **Doc-only (16):** `.latest-doc-only-change.txt` updated, no version bump

`# use next version` inter-provider pins resolved: sftp → ssh `>=6.0.0`, http → common.compat `>=1.17.0`.

Follow-up commits reconcile the wave against the commits actually being released:

- **Ordering.** Three changelog sections (apache.spark, postgres, snowflake) are restored to newest-merge-first — the order the generator produces from `git log`, which the incremental flow had drifted off by appending newer entries at the bottom.
- **Gaps.** common.compat was originally left out of the wave, which made http's `# use next version` marker resolve to the already-published 1.16.0 — but http 6.0.5 calls `get_async_connection(hook=...)`, added in #69140 and never released, so that pairing raises `TypeError` in the triggerer. airbyte was re-classified from patch to major: its only pending commit moves `airbyte-api` to the 1.x series and stops installing `requests`, excluding every previously supported version of a hard dependency. cncf.kubernetes and apache.spark were each missing an entry, and sftp/ssh cited the tracking issue (#54079) instead of the merge request that shipped the change (#69669).
- **Late addition.** atlassian.jira joined the wave after the initial cut. #70224 merged ~5h later and, behind a tooling subject ("Fix conn-fields check crash for nested provider packages"), declared a `Verify SSL` connection field and hidden-field behaviour in the provider's `provider.yaml` — metadata that ships to users via `get_provider_info.py`. It is bumped 3.3.4 → 3.3.5 (Misc). The incremental flow now sweeps every provider and prompts before adding or dropping one, so a provider that gains a user-facing change mid-wave can no longer slip out of the release silently (apache/airflow skill PR: shahar1:incremental-detect-providers-joining-wave).

---

- [X] Yes — Claude Code (Opus 4.8, 1M context)

Generated-by: Claude Code (Opus 4.8, 1M context) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


providers/keycloak/docs/changelog.rst
```diff
@@ -25,6 +25,26 @@
 Changelog
 ---------
 
+0.8.2
+.....
+
+Bug Fixes
+~~~~~~~~~
+
+* ``derive keycloak oauth redirect_uri from configured base url (#69801)``
```
vincbeck:
```suggestion
* ``Derive keycloak oauth redirect_uri from configured base url (#69801)``
```

providers/keycloak/docs/changelog.rst
```diff
@@ -25,6 +25,26 @@
...
+* ``Set 'secure' flag on keycloak login cookies behind a tls proxy (#69594)``
+
+Doc-only
+~~~~~~~~
+
+* ``Hide non-user-facing entries from ad-hoc provider release notes``
```
vincbeck: No PR on this one?

shahar1: It shouldn't have appeared at all, it seems like internal instruction for the changelog :)
Trying to figure out why it happened

vincbeck: There are other occurrences actually (cncf and keycloak)

vincbeck (approved):

providers/apache/kafka/docs/changelog.rst
```diff
@@ -27,6 +27,32 @@
...
+Features
+~~~~~~~~
+
+* ``Add a DagRun and TaskInstance event listener under the Kafka provider (#68082)``
+* ``Add Amazon MSK IAM (OAUTHBEARER) support to Apache Kafka provider (#69427)``
+
+Misc
+~~~~
+
+* ``Rename Kafka Listener to Kafka Event Producer and update the documentation (#70014)``
```
shahar1: Note to myself: as the Kafka Listener is only introduced, this entry doesn't make sense for someone who encounters it in the first time.

gopidesupavan:
@shahar1 can we exclude common-dataquality its only skelton at present merged, we are in discussions some of things on rules.. so the actual pr that is not merged yet.

shahar1:
> @shahar1 can we exclude common-dataquality its only skelton at present merged, we are in discussions some of things on rules.. so the actual pr that is not merged yet.

Good catch, I'll add an indication in the prepare-docs skill about proivders in ongoing development

eladkal:
> > @shahar1 can we exclude common-dataquality its only skelton at present merged, we are in discussions some of things on rules.. so the actual pr that is not merged yet.
> 
> Good catch, I'll add an indication in the prepare-docs skill about proivders in ongoing development

I assume the ready state in provider.yaml is wrong?

shahar1:
> > > @shahar1 can we exclude common-dataquality its only skelton at present merged, we are in discussions some of things on rules.. so the actual pr that is not merged yet.
> > 
> > 
> > Good catch, I'll add an indication in the prepare-docs skill about proivders in ongoing development
> 
> I assume the ready state in provider.yaml is wrong?

You're correct, PR for fixing it: https://github.com/apache/airflow/pull/70275

bugraoz93 (approved):

providers/common/sql/docs/changelog.rst
```diff
@@ -25,6 +25,18 @@
 Changelog
 ---------
 
+2.0.3
+.....
+
+Bug Fixes
+~~~~~~~~~
+
+* ``Make psycopg (v3) the default synchronous Postgres driver (#69526)``
```
Miretpl, line 34: In most of the cases, it is classified as `Misc` instead of `Bug Fixes`, and to me it is more like `Misc`. Is there a particular reason for the different classification in this provider?

providers/pgvector/docs/changelog.rst
```diff
@@ -20,6 +20,24 @@
 Changelog
 ---------
 
+1.7.3
+.....
+
+Bug Fixes
+~~~~~~~~~
+
+* ``Make psycopg (v3) the default synchronous Postgres driver (#69526)``
```
Miretpl, line 29: Same as the previous comment.

Miretpl (commented):
I don't know much about providers realease, but I tried to find something odd. Just two comments/questions.

providers/common/sql/docs/changelog.rst
shahar1: Good catch — reclassified to `Misc` in d5264b6.

You're right that there's no principled reason for the different call. The slice adds a psycopg (v3) / psycopg2 fallback in `get_sqlalchemy_engine()` for bare `postgresql://` URLs, which reads like a bug fix — but the break it guards against (psycopg2 not being installed) only exists because **this same wave** turns psycopg2 into an optional extra of the postgres provider. In the released state psycopg2 was a hard dependency, so nothing was actually broken; the slice adapts to an in-wave change rather than fixing released behaviour, which makes it `Misc` — the same call already made in amazon, google, and celery for this PR. Postgres keeps its `Breaking changes` entry, since that's where the default driver actually changes.

---

providers/pgvector/docs/changelog.rst
shahar1: Same reasoning as the common.sql thread — reclassified to `Misc` in d5264b6. The pgvector slice adds a `USE_PSYCOPG3` branch that adapts to the postgres provider's new in-wave default driver rather than fixing released behaviour, so it lines up with the `Misc` classification the same PR already has in amazon, google, and celery.

---

providers/common/sql/docs/changelog.rst
shahar1: Nice catch! Looks like non-determinism of the agentic skill that generated the docs. My hypothesis is that it attributed the patterns to a crash recovery, rather than compatibility.

Miretpl (approved):

providers/common/sql/docs/changelog.rst
Miretpl: > Postgres keeps its Breaking changes entry, since that's where the default driver actually changes.

That makes sense.

> Looks like non-determinism of the agentic skill that generated the docs. My hypothesis is that it attributed the patterns to a crash recovery, rather than compatibility.

Probably - yes

PR #68548, opened 2026-06-15
Add NodeCoordinator

shivaam:
This adds the Python-side NodeCoordinator needed to run TypeScript tasks through Airflow coordinator mode.

It pairs with the public TypeScript Task SDK interface in #67908. A Python Dag still defines the scheduling graph with stub tasks, while the task implementation can live in a TypeScript bundle. At execution time, Airflow routes the stub task to NodeCoordinator, which starts the Node.js bundle through the existing supervisor subprocess path.

The expected bundle layout is:

```text
/opt/airflow/ts-bundles/
  bundle.mjs
  airflow-metadata.yaml
```

`bundle.mjs` is the Node.js entrypoint for the TypeScript task runtime. `airflow-metadata.yaml` describes the bundle and provides the supervisor schema version the coordinator should use when talking to the runtime.

Example metadata:

```yaml
airflow_bundle_metadata_version: "1.0"
sdk:
  language: typescript
  version: "0.1.0-alpha.0"
  supervisor_schema_version: "2026-06-16"
source: src/airflow.ts
dags:
  sales_pipeline:
    tasks:
      - extract
      - transform
```

For this first PR, NodeCoordinator only needs `sdk.supervisor_schema_version`, but this metadata file follows the same general shape as Airflow executable bundle metadata so TypeScript bundle tooling can grow into it later.

Example Airflow config:

```ini
[sdk]
coordinators = {
    "ts": {
        "classpath": "airflow.sdk.coordinators.node.NodeCoordinator",
        "kwargs": {"bundles_root": ["/opt/airflow/ts-bundles"]}
    }
}
queue_to_coordinator = {"ts": "ts"}
```

This first version keeps bundle selection deliberately simple. `bundles_root` is an ordered fallback list of directories, and the first valid `bundle.mjs` with valid metadata is used. Dag/task-aware selection across multiple TypeScript bundles is left for follow-up work once the TypeScript bundle tooling exists.

### Testing

I ran a full local daemon E2E with `api-server`, `dag-processor`, and `scheduler` to test this locally with my other code. 

### Follow-up work

- Add the TypeScript SDK runtime PR with `startCoordinator()`.
- Add an official TypeScript pack/build tool that generates `bundle.mjs` and `airflow-metadata.yaml`.
- Add Dag/task-aware multi-bundle selection after bundle tooling exists.
- CI/CD tests

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes - OpenAI Codex

[...]

PR #70273, opened 2026-07-23
Remove airflow-metadata.yaml sidecar support from NodeCoordinator

jason810496:
- related: #69295 (`airflow-ts-pack`), #68548 (`NodeCoordinator`)

## Why

`airflow-ts-pack` **always** embeds the bundle metadata into `bundle.mjs` itself as a leading `//# airflowMetadata=<base64>` comment but the `NodeCoordinator` fallback that accepts a hand-written `airflow-metadata.yaml`. However, no tool produces the `airflow-metadata.yaml` file.

## What

- `NodeCoordinator` now treats the embedded metadata as the only accepted source: a `bundle.mjs` without the `//# airflowMetadata=` marker is rejected with `"bundle.mjs has no embedded airflow metadata.

---

- [x] Yes, with help of Claude Code (Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/task-sdk/src/airflow/sdk/coordinators/node/coordinator.py
+++ b/task-sdk/src/airflow/sdk/coordinators/node/coordinator.py
@@ -45,18 +45,17 @@ if TYPE_CHECKING:
 log: FilteringBoundLogger = structlog.get_logger(logger_name="coordinators.node")
 
 BUNDLE_FILENAME = "bundle.mjs"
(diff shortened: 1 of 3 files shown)


guan404ming (approved):
LGTM, verified locally and worked as expected.

PR #70277, opened 2026-07-23
Move ts-sdk prek hooks to its own config and add compile hook

jason810496:
## Why

The ts-sdk prek hooks lived in the root `.pre-commit-config.yaml` even though their scripts already live under `ts-sdk/scripts/ci/prek/` and every other workspace project (`task-sdk`, `airflow-core`, `go-sdk`, `shared/*`, ...) owns its hooks in a nested config; building the ts-sdk also required a locally installed pnpm.

## Verification

- `prek run ts-sdk-lint --all-files`, `prek run check-ts-sdk-supervisor-schema --all-files` — both pass from the new workspace config.
- `prek run compile-ts-sdk` — builds `ts-sdk/dist/` with prek-managed pnpm; excluded from default-stage runs (`prek run --dry-run` on a ts-sdk file does not list it), so ordinary commits do not trigger a build.
- Trigger parity: `prek run ts-sdk-lint --files ts-sdk/src/index.ts` runs the hook; `--files README.md` skips it.

---

### DIFF
--- a/ts-sdk/.pre-commit-config.yaml
+++ b/ts-sdk/.pre-commit-config.yaml
@@ -54,7 +54,12 @@ repos:
         entry: ./scripts/ci/prek/compile_ts_sdk.py
         language: node
         stages: [manual]
-        always_run: true
+        files: |
+          (?x)
+          ^src/.*\.ts$|
+          ^package\.json$|
+          ^pnpm-lock\.yaml$|
+          ^tsconfig(\.build)?\.json$
         additional_dependencies: ['pnpm@10.28.1']
         pass_filenames: false
         require_serial: true
--- a/ts-sdk/.pre-commit-config.yaml
+++ b/ts-sdk/.pre-commit-config.yaml
@@ -54,12 +54,7 @@ repos:
         entry: ./scripts/ci/prek/compile_ts_sdk.py
         language: node
         stages: [manual]
-        files: >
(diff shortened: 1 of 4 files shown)


ts-sdk/.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+          ^scripts/generate-supervisor\.mjs$
+        additional_dependencies: ['pnpm@10.28.1']
+        pass_filenames: false
+        require_serial: true
+      - id: compile-ts-sdk
+        name: Compile TypeScript SDK
+        entry: ./scripts/ci/prek/compile_ts_sdk.py
+        language: node
+        stages: [manual]
+        always_run: true
```
guan404ming: Could we use `files:` like `compile-ui-assets` in `airflow-core/.pre-commit-config.yaml` here?

guan404ming (approved):
Nice!

ts-sdk/.pre-commit-config.yaml
jason810496: Good call! I addressed in https://github.com/apache/airflow/pull/70277/commits/998c5514d899759968d143afe1c8a2f71992b527. Thanks.

guan404ming (approved):
Thanks for the update. LGTM!

Issue #41093, opened 2024-07-29
Add `execute_tasks_new_python_interpreter` argument to `PythonOperator` 


pedro-cf:
### Description

Currently from my knowledge it is only possible to force all python tasks to execute in a new 
interpreter for every execution globally via this setting: 

- https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#execute-tasks-new-python-interpreter

Would it be possible to add this argument to the `PythonOperator` so it is possible to control this behavior at the task level?

### Use case/motivation

- https://github.com/apache/airflow/discussions/40956

### Related issues

- https://github.com/apache/airflow/discussions/40956

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

potiuk:
Marked it as a good first issue for anyone to take - but feel free to propose a PR as well.

Prab-27:
Hi @potiuk, I would like to work on this issue as my first issue. May I ?

potiuk:
Sure

Prab-27:
@potiuk I have updated description I think it would solve your query.

potiuk:
@Prab-27 We are unassigning you from this issue as part of our updated [assignment policy](https://github.com/apache/airflow/blob/main/contributing-docs/04_how_to_contribute.rst#contribute-code-changes).

This is not meant to discourage your contribution — quite the opposite! You are still very welcome to work on this issue and submit a PR for it. Simply comment that you are working on it and open a PR when ready.

We found that formal assignments were not working well, as they often prevented others from contributing when the assignee was not actively working on the issue.

PR #70280, opened 2026-07-23
Allow PythonOperator to override task interpreter mode

fat-catTW:
## Why
`[core] execute_tasks_new_python_interpreter` currently applies globally, so enabling fresh Python interpreter startup affects every task in the deployment.

Some Python tasks need that behavior, for example to pick up plugin changes immediately, but forcing all tasks onto the slower execution mode is unnecessary.

relates: #41093

## Solution
Add `execute_tasks_new_python_interpreter` to `PythonOperator` as a task-level override.

When set to `True` or `False`, the task value is carried through Dag serialization, the `ExecuteTask` workload, and the task supervisor. When left as `None`, task execution keeps using the global `[core] execute_tasks_new_python_interpreter` configuration.

-`[X]`Yes (please specify the tool below)

---

### DIFF
--- a/airflow-core/newsfragments/70280.bugfix.rst
+++ b/airflow-core/newsfragments/70280.bugfix.rst
@@ -1,0 +1,1 @@
+Restored support for ``[core] execute_tasks_new_python_interpreter`` when running tasks with LocalExecutor and CeleryExecutor.
--- a/airflow-core/src/airflow/executors/base_executor.py
+++ b/airflow-core/src/airflow/executors/base_executor.py
@@ -724,6 +724,7 @@ class BaseExecutor(LoggingMixin):
                 log_path=workload.log_path,
                 subprocess_logs_to_stdout=subprocess_logs_to_stdout,
                 sentry_integration=getattr(workload, "sentry_integration", ""),
+                execute_tasks_new_python_interpreter=workload.execute_tasks_new_python_interpreter,
             )
         if isinstance(workload, ExecuteCallback):
             from airflow.sdk.execution_time.callback_supervisor import supervise_callback
--- a/airflow-core/src/airflow/executors/base_executor.py
+++ b/airflow-core/src/airflow/executors/base_executor.py
@@ -724,7 +724,6 @@ class BaseExecutor(LoggingMixin):
                 log_path=workload.log_path,
                 subprocess_logs_to_stdout=subprocess_logs_to_stdout,
                 sentry_integration=getattr(workload, "sentry_integration", ""),
-                execute_tasks_new_python_interpreter=workload.execute_tasks_new_python_interpreter,
             )
         if isinstance(workload, ExecuteCallback):
             from airflow.sdk.execution_time.callback_supervisor import supervise_callback
--- a/airflow-core/src/airflow/executors/workloads/task.py
+++ b/airflow-core/src/airflow/executors/workloads/task.py
@@ -27,6 +27,14 @@ from airflow.api_fastapi.execution_api.datamodels.taskinstance import TaskInstan
 from airflow.executors.workloads.base import BaseDagBundleWorkload, BundleInfo
 from airflow.utils.state import TaskInstanceState
 
+
+def _get_execute_tasks_new_python_interpreter(ti: object) -> bool | None:
+    value = getattr(getattr(ti, "task", None), "execute_tasks_new_python_interpreter", None)
+    if value is None or isinstance(value, bool):
+        return value
+    return None
+
+
 if TYPE_CHECKING:
     from airflow.api_fastapi.auth.tokens import JWTGenerator
     from airflow.models.taskinstance import TaskInstance as TIModel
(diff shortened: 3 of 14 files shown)


ignacioparicio:
Not a review, just something I ran into that seems relevant.

I think this PR does more than add a per-task override. The config fallback in `_should_use_exec_for_task` also makes `[core] execute_tasks_new_python_interpreter` work again for CeleryExecutor and LocalExecutor, which has had no effect since 3.0.

The Celery code that reads it sits behind an `if not AIRFLOW_V_3_0_PLUS:` gate, and as far as I can tell the Edge worker is the only thing still honouring it on 3.x.

I opened https://github.com/apache/airflow/pull/70901 in parallel to document that and get `airflow config lint` to flag it (happy to rework it if this lands first).

So if this merges it's a behaviour change for anyone still running Celery or Local with `execute_tasks_new_python_interpreter` set to `True`, which might deserve a newsfragment of its own.

fat-catTW:
Thanks, that makes sense. Since this PR restores the global ``[core] execute_tasks_new_python_interpreter`` behavior for LocalExecutor and CeleryExecutor, I’ll add a newsfragment for that. Your #70901 can then probably focus on the config lint/documentation side depending on which PR lands first.

uranusjr:
If I understand this correctly… this won’t work. `ExecuteTask.make()` reads `ti.task.execute_tasks_new_python_interpreter`, but `ti.task` is not guarenteed to be set at this point.

fat-catTW:
Good catch, thanks for review. 
I changed this so `ExecuteTask.make()` no longer reads from `ti.task`.
The scheduler now resolves the task from the serialized Dag during enqueue, reads `execute_tasks_new_python_interpreter` from that serialized task, and passes the value explicitly into `ExecuteTask.make()`. If the serialized Dag or task is unavailable, it falls back to `None`, so the existing global config fallback behavior is preserved.

uranusjr:
I don’t think this is going to fly either, at least as-is. `ExecuteTask.make()` is used in providers, you need to add this to more places. Unfortunately, providers also introduce complex version compatibility issues, so you’ll need to figure out a way to ensure things work correctly between different version combinations. And for the combinations where things can’t work, you’ll need to be able to do error handling, preferrably at parse time. The two easiest ways this can go down are either to silently make the argument do nothing (or drop it with a warning), or make the scheduler crash. Neither is acceptable IMO.

fat-catTW:
Thanks, that makes sense.
The PythonOperator-level override has broader provider/core compatibility implications than I initially accounted for.

While working on this, I noticed a related bug.
The existing global `[core] execute_tasks_new_python_interpreter` setting was not being honored by the task supervisor path. I narrowed this PR to only restore that fallback behavior, and removed the per-task override changes from this PR.

So this PR is now a smaller bugfix related to #41093, rather than closing it. The per-task override can be handled separately with a more complete design for provider executor compatibility.

potiuk (changes requested):
Thanks for narrowing this down, but the bug this PR now fixes has already landed on `main` in #72164, and the remaining hunk would undo how that PR scoped the option, so I'm closing this one as superseded rather than asking for a rebase.

### Superseded by #72164

```python
def _should_use_exec() -> bool:
    """Whether forked children should ``exec`` a fresh interpreter."""
    return sys.platform in _FORK_EXEC_PLATFORMS or conf.getboolean(   # ← already on main as _task_process_uses_exec()
        "core", "execute_tasks_new_python_interpreter", fallback=False
    )
```

#72164 ("Allow opting into fork+exec task processes via execute_tasks_new_python_interpreter", merged 2026-09-14) added `_task_process_uses_exec()` with exactly this `or conf.getboolean(...)` fallback, a `TestTaskProcessUsesExec` test, the `72164.significant.rst` newsfragment and an updated `config.yml` description. This PR's diff no longer applies to `main`, and the green CI from 2026-08-08 predates that merge.

### The hunk also changes the Dag processor and the triggerer

`_should_use_exec()` is not task-only; it is the shared platform gate:

```text
airflow-core/src/airflow/dag_processing/processor.py:619:  use_exec = target is _parse_file_entrypoint and supervisor._should_use_exec()
airflow-core/src/airflow/jobs/triggerer_job_runner.py:551: use_exec=supervisor._should_use_exec(),
```

Reading the config here means `execute_tasks_new_python_interpreter=True` execs a fresh interpreter for every Dag-file parse and for the triggerer too, which the option has never described. #72164 avoided exactly this by introducing a separate `_task_process_uses_exec()` and leaving `_should_use_exec()` platform-only; its newsfragment states that the Dag processor, the triggerer and task callbacks keep bare fork. If this PR were rebased with this hunk kept, it would silently revert that scoping.

### Smaller observations
- `test_should_use_exec_honors_platform_and_config` stubs `supervisor.conf.getboolean` with `lambda *args, **kwargs: config_value`, which returns the value for any section/key; `conf_vars` (already imported in this file) is the way to override config in tests.
- The title, body and labels still describe the per-task `PythonOperator` override that was removed on 2026-08-08, and `70280.bugfix.rst` duplicates the merged `72164.significant.rst` entry.
- If you want to pursue the original per-task override, that needs a new design that answers `uranusjr`'s points (providers construct `ExecuteTask` themselves; unsupported version combinations must fail at parse time), best as a fresh PR.

---

>

potiuk:
Closing as superseded by #72164 (see the review above). Thanks for the work on this; a fresh PR is the best route if you want to pursue the per-task override.

PR #70285, opened 2026-07-23
Bake Node.js and pnpm runtime into breeze CI image

jason810496:
## Why

Language-SDK work in the repo needs a Node.js/pnpm runtime for the `ts-sdk` (running integration/system tests, building the example bundle, etc.), but unlike Go and the JDK — which every breeze CI image already ships (`install_golang` / `install_jdk` in `scripts/docker/install_os_dependencies.sh`) — Node.js is absent, so there is no runtime available inside `breeze shell` today. This bakes Node.js + pnpm into the CI image the same unconditional way, so it is always present for lang-SDK integration/system testing.

## Verification

- Built the CI image end to end (`breeze ci-image build --python 3.10`) and confirmed inside the built image: `node --version` → `v22.23.1`, `npm --version` → `10.9.8`, `pnpm --version` → `11.16.0`, all resolving on the default `PATH`; Go 1.26.5 and OpenJDK 11 remain present (no regression to the existing baked-in toolchains).
- Ran the `install_nodejs` logic standalone in a bare `debian:bookworm-slim` container (the image's `BASE_IMAGE`) with a live checksum match against `nodejs.org`'s published `SHASUMS256.txt`.
- `prek run` passes on all changed files (Dockerfile lint, shell syntax, license headers, inlined-scripts sync).

---

- [x] Yes, with help of Claude Code Opus 4.8 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/scripts/docker/install_os_dependencies.sh
+++ b/scripts/docker/install_os_dependencies.sh
@@ -30,6 +30,9 @@ PYTHON_LTO=${PYTHON_LTO:-true}
 GOLANG_MAJOR_MINOR_VERSION=${GOLANG_MAJOR_MINOR_VERSION:-1.24.4}
 TEMURIN_VERSION=${TEMURIN_VERSION:-11}
 NODEJS_VERSION=${NODEJS_VERSION:-22.23.1}
+# Keep in sync with the "packageManager" pin in ts-sdk/package.json so the version corepack
+# resolves for ts-sdk is the one baked into the image.
+PNPM_VERSION=${PNPM_VERSION:-10.28.1}
 RUSTUP_DEFAULT_TOOLCHAIN=${RUSTUP_DEFAULT_TOOLCHAIN:-stable}
 RUSTUP_VERSION=${RUSTUP_VERSION:-1.29.0}
 COSIGN_VERSION=${COSIGN_VERSION:-3.0.5}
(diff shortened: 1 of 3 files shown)


scripts/docker/install_os_dependencies.sh
```diff
@@ -420,6 +421,33 @@ https://packages.adoptium.net/artifactory/deb ${DISTRO_CODENAME} main" \
...
+        echo "Unsupported architecture for nodejs: ${arch}"
+        exit 1
+    fi
+    curl --retry 3 --retry-delay 5 \
+        "https://nodejs.org/dist/v${NODEJS_VERSION}/node-v${NODEJS_VERSION}-${target}.tar.xz" \
+        -o /tmp/nodejs.tar.xz
+    echo "${nodejs_sha256}  /tmp/nodejs.tar.xz" | sha256sum --check
+    tar -xJf /tmp/nodejs.tar.xz --strip-components=1 -C /usr/local --no-same-owner
+    rm -f /tmp/nodejs.tar.xz
+    corepack enable --install-directory /usr/local/bin
```
guan404ming, line 451: `corepack enable` seems to only write shims, so pnpm is fetched from the registry rather than baked in and the verified `11.16.0` isn't what `ts-sdk`gets (it pins `pnpm@10.28.1`). Could add `corepack prepare pnpm@<version> --activate` to cache it in the image.

guan404ming (commented):
Thanks, left one comment otherwise looks great!

guan404ming (approved):
Thanks for the update!

Issue #51163, opened 2025-05-28
More concise error message in the api-server log when `execution_api_server_url` isn't correct


radumas:
### Description

When the `execution_api_server_url` config variable isn't correct, the consequences are:

- Tasks don't run successfully
- task state mismatch under Events tab in the UI https://github.com/apache/airflow/issues/50191

These don't explicitly identify the root cause.

### Use case/motivation

It should be easier to identify when something core in the configuration is misconfigured such that the different components of Airflow can't communicate with each other. This could be evident from:

* a configuration checker (like `airflow config check`)
* On the homepage of the UI? More details when one of the components has a big red X
* In the `api-server` log (or the scheduler log?)

### Related issues

The symptoms of misconfiguration were first identified here https://github.com/apache/airflow/issues/50191

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

amoghrajesh:
I guess we could probably have a means where we could atleast provide a mechanism to ping the execution API server that works without a token from the worker.

amoghrajesh:
Let me self-assign this for now. I will come back at it at a later point in time (maybe for 3.0.3 or 3.1)

pierrejeambrun:
@amoghrajesh Do you think you'll be able to get this in 3.1.0 or do we want to change milestone to 3.1+ considering other priorities?

amoghrajesh:
@pierrejeambrun thanks for the bump. Won't be able to make 3.1, moving to 3.1+

PR #70290, opened 2026-07-23
Add task log message when the Execution API is unreachable

karenhsieh75:
## Why

A supervisor connection failure to the Execution API only showed up in the executor's stdout. As a result, UI users only saw a failed task with an empty log, making it difficult to identify the root cause.

## How

- Catch `httpx.RequestError` in `supervise_task` and log the failure (with the relevant config to check) to the task's own log before re-raising
- Add a test covering the new log message

related: #51163

---


karenhsieh75:
@guan404ming Could you please help review this? Thanks!

PR #70291, opened 2026-07-23
Add support for operators that achieve durable execution manually in registry

amoghrajesh:
---

- [x] Yes: claude sonnet 4.6

---

### What

Follow-up to https://github.com/apache/airflow/pull/70289, which added `__supports_durable_execution: ClassVar[bool] = True` directly to `KubernetesPodOperator` and `AgentOperator` both of which implement crash-safe execution against `task_state_store` without inheriting `ResumableJobMixin`. That attribute wasn't read by anything yet; the Registry's `is_durable_capable()` only checked for `ResumableJobMixin` in the MRO, so both operators stayed invisible to the badge despite genuinely qualifying.

### Current behaviour

`is_durable_capable()` in `dev/registry/extract_parameters.py` only qualifies a class via the `ResumableJobMixin` structural check (inherits it, all abstract methods overridden, `execute_resumable` reachable from `execute()`). Classes that implement durable execution some other way have no path to qualify.

### Proposed change

Adds a second, independent qualifying path: `is_durable_capable()` now also checks for a class level `__supports_durable_execution = True` declaration. Because the attribute name uses a double leading underscore, Python name mangles it to `_{ClassName}__supports_durable_execution` on whichever class declares it. The check computes that mangled name per class being inspected (`f"_{cls.__name__}__supports_durable_execution"`) rather than a fixed string, so it only matches the exact class that wrote the declaration in its own body -- not any subclass that merely inherits from it.

### Changes of Note

This distinction matters concretely: `KubernetesPodOperator` has real subclasses (`SparkKubernetesOperator`, `KubernetesJobOperator`, `KubernetesStartKueueJobOperator`) that override `execute()` themselves. Nothing verifies those overrides preserve the `task_state_store` reconnect behavior yet, so the marker must not propagate to them automatically. 

### Testing

Manual validation after generating registry data shows:

```json
    {
      "id": "common-ai-agent-AgentOperator",
      "name": "AgentOperator",
      "type": "operator",
      "import_path": "airflow.providers.common.ai.operators.agent.AgentOperator",
      "module_path": "airflow.providers.common.ai.operators.agent",
      "short_description": "Run a pydantic-ai Agent with tools and multi-turn reasoning.",
      "docs_url": "https://airflow.apache.org/docs/apache-airflow-providers-common-ai/stable/_api/airflow/providers/common/ai/operators/agent/index.html#airflow.providers.common.ai.operators.agent.AgentOperator",
      "source_url": "https://github.com/apache/airflow/blob/providers-common-ai/0.7.0/providers/common/ai/src/airflow/providers/common/ai/operators/agent.py#L120",
      "category": "common-ai",
      "provider_id": "common-ai",
      "provider_name": "Common AI",
      "supports_durable_execution": true
    },
```

And

```json
    {
      "id": "cncf-kubernetes-pod-KubernetesPodOperator",
      "name": "KubernetesPodOperator",
      "type": "operator",
      "import_path": "airflow.providers.cncf.kubernetes.operators.pod.KubernetesPodOperator",
      "module_path": "airflow.providers.cncf.kubernetes.operators.pod",
      "short_description": "Execute a task in a Kubernetes Pod.",
      "docs_url": "https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/_api/airflow/providers/cncf/kubernetes/operators/pod/index.html#airflow.providers.cncf.kubernetes.operators.pod.KubernetesPodOperator",
      "source_url": "https://github.com/apache/airflow/blob/providers-cncf-kubernetes/10.20.0/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py#L142",
      "category": "kubernetes",
      "provider_id": "cncf-kubernetes",
      "provider_name": "Kubernetes",
      "supports_durable_execution": true
    },
```

---

### DIFF
--- a/dev/registry/tests/test_extract_parameters.py
+++ b/dev/registry/tests/test_extract_parameters.py
@@ -225,6 +225,24 @@ class PlainOperator:
         return None
 
 
+class ManuallyDurableOperator:
+    """Implements durable execution directly (e.g. via task_state_store), without
+    ResumableJobMixin -- mirrors KubernetesPodOperator/AgentOperator."""
+
+    __supports_durable_execution = True
+
+    def execute(self, context):
+        return None
+
+
+class ManuallyDurableSubclass(ManuallyDurableOperator):
+    """Overrides execute() itself -- must NOT inherit the parent's declaration,
+    since nothing here verifies it preserves the reconnect behavior."""
+
+    def execute(self, context):
+        return "something else entirely"
+
+
 class TestIsDurableCapable:
     def test_fully_implemented_and_wired_qualifies(self):
         assert is_durable_capable(FullyImplementedResumableOperator, FakeResumableJobMixin) is True
@@ -240,6 +258,12 @@ class TestIsDurableCapable:
 
     def test_mixin_unavailable_disqualifies(self):
         assert is_durable_capable(FullyImplementedResumableOperator, None) is False
+
+    def test_manual_durable_marker_qualifies_without_mixin(self):
+        assert is_durable_capable(ManuallyDurableOperator, None) is True
+
+    def test_subclass_not_redeclaring_marker_disqualifies(self):
+        assert is_durable_capable(ManuallyDurableSubclass, FakeResumableJobMixin) is False
 
 
 # ---------------------------------------------------------------------------
--- a/dev/registry/extract_parameters.py
+++ b/dev/registry/extract_parameters.py
@@ -386,11 +386,29 @@ def load_resumable_job_mixin() -> type | None:
 
 
 def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
-    """Return True if a class fully implements ResumableJobMixin's crash-recovery contract.
+    """Return True if a class implements durable/crash-safe execution.
 
-    Inheriting the mixin is not sufficient: a complete override is inert unless
-    execute() actually calls execute_resumable().
+    Two ways to qualify:
+    1. A class-level `__supports_durable_execution = True`
+    declaration (for operators that implement this directly against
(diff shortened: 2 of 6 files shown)


dev/registry/tests/test_extract_parameters.py
```diff
@@ -225,6 +225,24 @@ def execute(self, context):
         return None
 
 
+class ManuallyDurableOperator:
+    """Implements durable execution directly (e.g. via task_state_store), without
+    ResumableJobMixin -- mirrors KubernetesPodOperator/AgentOperator."""
+
+    __supports_durable_execution = True
```
Lee-W, line 232: After a second thought, do we really need `__`? Are we worrying whether the user will change it? Or is there any reason we need to use `__`?

Lee-W: since we're accessing this value through the registry script, i start to think maybe it's ok to make it public? or `_...`?

amoghrajesh: There's a genuine reason actually. Two things:

1. We do not want users to change it
2. We do not want the inherited operators to get the property by itself, for example - KPO has  `__supports_durable_execution` set but I wouldnt want SparkKPO to have it as well just cos it inherits KPO, we do not know if SparkKPO or any inherited operator _really supports_ durable execution

Lee-W: ok, the inherit parts make sense to me. i thought all the subclass will automatically have this capibility

amoghrajesh: Nope, better safe than sorry!

Lee-W (approved):

PR #70294, opened 2026-07-23
Highlight user-code frames in task log tracebacks

pierrejeambrun:
When a task fails, the frame that actually raised is almost always in the user's DAG code, but in the log traceback it looks identical to the surrounding framework frames from installed packages — so you have to scan every `File "..."` line to find where your code is.

This colors traceback frames by origin: frames in **user code** (DAG bundle, plugins, local files) stay prominent (`fg.info`, bold), while frames in installed packages (`site-packages` / `dist-packages`) are muted (`fg.muted`). The failing user frame now stands out at a glance.

Classification is a single rule — a frame is user code unless its path is under a `site-packages`/`dist-packages` directory — which is robust across bundle layouts (local folder, git, `/tmp/airflow/dag_bundles/...`) without hardcoding a bundle path. A `data-frame-source` attribute (`user`/`library`) is set on each frame for testability and as a styling hook.

## Before
error in user code only
Screenshot 2026-07-23 at 16 58 44
error in library
Screenshot 2026-07-23 at 16 59 14

## After
error in user code only
Screenshot 2026-07-23 at 16 52 45
error in library
Screenshot 2026-07-23 at 16 52 30

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/renderStructuredLog.test.tsx
+++ b/airflow-core/src/airflow/ui/src/components/renderStructuredLog.test.tsx
@@ -35,6 +35,47 @@ describe("tiContextFields", () => {
   });
 });
 
+describe("renderStructuredLog — traceback frame highlighting", () => {
+  const sitepkgFile = "/usr/local/lib/python3.14/site-packages/airflow/sdk/bases/operator.py";
+  const dagBundleFile = "/tmp/airflow/dag_bundles/astro/main/dags/non_alert/enrich_ticket.py";
+
+  const renderWithError = () => {
+    const result = renderStructuredLog({
+      index: 0,
+      logLink: "",
+      logMessage: {
+        error_detail: [
+          {
+            exc_notes: [],
+            exc_type: "TypeError",
+            exc_value: "fromisoformat: argument must be str",
+            frames: [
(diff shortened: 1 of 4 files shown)


bbovenzi:
How does this look in dark mode? 

How different do we want these links to look like? fg.info vs fg.muted in light mode is fairly subtle.

collinmcnulty:
The pieces I really want in color are the parts that are my code (so not site-packages and not airflow itself) and the actual error message (ParserError in this case). There's so much blue on the screen that its actually hard to pick those out. I think it would be fine to put the site-packages lines just in the normal black text color.

bbovenzi:
That makes sense to me, system file lines aren't important so we should reduce their highlight or remove it entirely

pierrejeambrun:
Updated. Now library line appear in normal text color:

Screenshot 2026-07-23 at 19 16 27
Screenshot 2026-07-23 at 19 16 58


Here airflow is considered 'user code' because that's a local development setup where airflow is actually installed in editable mode from source. Normally in prod environment it will be 'pip installed...' and land in site packages.

vatsrahul1001 (approved):

pierrejeambrun:
On a real install. With airflow installed from a wheel:

Screenshot 2026-07-24 at 11 28 33
Screenshot 2026-07-24 at 11 28 52

Issue #70265, opened 2026-07-23
Provider migration for decoupling remote logging from core


jason810496:
- related: #67056

## Why

#67056 decoupled remote logging from the hardcoded if/elif chain in
`airflow-core/src/airflow/config_templates/airflow_local_settings.py`. Core and the Task SDK now
share one resolver (`airflow_shared.logging.factory.resolve_remote_task_log`) that dispatches on the
URL scheme of `[logging] remote_base_log_folder` via `ProvidersManager` and instantiates the
provider class through a no-arg `from_config()` classmethod.

The legacy per-scheme chain in `airflow_local_settings.py` is transitional: it stays as a fallback
until every remote logging backend has been migrated to the new mechanism. This issue tracks that
migration, one sub-issue per backend.

## The pattern

Each migration is a small, self-contained PR with three parts (see the merged amazon examples:
s3 — #69817, cloudwatch — #69816):

1. **Add `from_config()`** — a no-arg classmethod on the provider's `RemoteLogIO` class that mirrors
   the corresponding legacy branch in `airflow_local_settings.py`, including the
   `[logging] remote_task_handler_kwargs` IO-kwargs merge (where the legacy branch applies it) and
   `expanduser` on `base_log_folder`, so behavior is unchanged for existing configs.
2. **Register the scheme** — add a `remote-logging:` section to the provider's `provider.yaml` and
   mirror it in `get_provider_info.py`:

   ```yaml
   remote-logging:
     - classpath: airflow.providers.amazon.aws.log.s3_task_handler.S3RemoteLogIO
       scheme: s3
   ```

3. **Tests** — mirror `TestS3RemoteLogIOFromConfig` in
   `providers/amazon/tests/unit/amazon/aws/log/test_s3_task_handler.py`: `from_config` field mapping,
   IO-kwargs filtering, scheme registration, and end-to-end `resolve_remote_task_log` dispatch.
4. **Real system test** — set up the actual backend yourself (a real service or a local
   equivalent), run a task with remote logging enabled, and verify logs are uploaded and read back
   through the new dispatch path. Include the setup and verification results in the PR description.

If `from_config()` raises on a bad config, the shared factory falls back to the legacy path — so
these migrations are non-breaking by construction.

## Backends

- [x] `s3` (amazon) — #69817
- [x] `cloudwatch` (amazon) — #69816
- [x] `gs` (google) — #70266
- [x] `stackdriver` (google) — #70267
[...]

PR #70295, opened 2026-07-23
Add OpensearchRemoteLogIO.from_config and register opensearch scheme

### LINKED ISSUE #70272: Add OpensearchRemoteLogIO.from_config and register opensearch scheme

## Why

`from_config()` classmethod. This issue migrates the OpenSearch backend.

## Design note (feedback welcome)

Unlike the object-storage backends, OpenSearch is currently selected by `[opensearch] host` being
set — `remote_base_log_folder` typically has no scheme for these deployments, so pure scheme
dispatch cannot reach it. The proposal (same as the sibling Elasticsearch issue #70271):

- Register an `opensearch` scheme and document setting
  `[logging] remote_base_log_folder = opensearch://` as the forward-looking configuration.
- Keep the legacy host-based selection working via the transitional fallback in
  `airflow_local_settings.py` until the chain is removed (tracked by the meta issue), so existing
  configs are unaffected.

## What

- [ ] Add `OpensearchRemoteLogIO.from_config()` in
      `providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py`, mirroring the
      legacy branch:
      https://github.com/apache/airflow/blob/104ad12e190db7197fa303d98a3ab68879eafd33/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L349-L379
      — reading the `[opensearch]` options the branch reads (`host`, `port`, `username`,
      `password`, `target_index`, `write_stdout`, `write_to_os`, `json_format`, `host_field`,
      `offset_field`, and `log_id_template` with its
      `{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}` fallback) plus
      `[logging] base_log_folder` (with `expanduser`) and `[logging] delete_local_logs`, so
      behavior is unchanged for existing configs. Note the legacy branch does *not* merge
      `remote_task_handler_kwargs` IO-kwargs for this backend — mirror that too.
- [ ] Register the `opensearch` scheme under a `remote-logging:` section in
      `providers/opensearch/provider.yaml` and mirror it in
      `providers/opensearch/src/airflow/providers/opensearch/get_provider_info.py`.
- [ ] Document the `remote_base_log_folder = opensearch://` form in the provider's logging docs.
      OpenSearch cluster or a local one, e.g. via Docker), run a task with remote logging enabled,
      and confirm logs are written and read back through the new dispatch path. Include the setup
      and verification results in the PR description.

## Reference

FrankYang0529:
## Why

#67056 decoupled remote logging from the hardcoded branches in `airflow_local_settings.py`: core and the Task SDK now resolve the handler via ProvidersManager dispatch on the `[logging] remote_base_log_folder` URL scheme. This migrates the `opensearch` scheme.

## How

- Add `OpensearchRemoteLogIO.from_config()` reading the `[opensearch]` section — a mirror of the legacy branch, so the user-facing config surface is unchanged.
- Register the `opensearch` scheme in `provider.yaml`.

## Verify

- `uv run --project providers/opensearch pytest providers/opensearch/tests/unit/opensearch/log/test_os_task_handler.py`
- end to end test

```
# 1. Create config
mkdir -p files/airflow-breeze-config
cat > files/airflow-breeze-config/environment_variables.env <<'EOF'
export AIRFLOW__LOGGING__REMOTE_LOGGING=True
export AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER=opensearch://
export AIRFLOW__OPENSEARCH__HOST=http://opensearch:9200
export AIRFLOW__OPENSEARCH__PORT=9200
export AIRFLOW__OPENSEARCH__WRITE_TO_OS=True
export AIRFLOW__OPENSEARCH__JSON_FORMAT=True
export AIRFLOW__OPENSEARCH__WRITE_STDOUT=False
export AIRFLOW__OPENSEARCH__TARGET_INDEX=airflow-logs
EOF

# 2. Start Airflow
breeze start-airflow --integration opensearch --backend postgres --load-example-dags

# 3. Run Dag
airflow dags unpause example_bash_operator
airflow dags trigger example_bash_operator
airflow dags list-runs example_bash_operator

# 4. Check log
curl -s localhost:9200/airflow-logs/_count ; echo
curl -s 'localhost:9200/airflow-logs/_search?q=log_id:example_bash_operator&size=3'
```

 

* closes: #70272
* related: #70265

---

- [X] Yes - Claude Code

---

### DIFF
--- a/providers/opensearch/docs/logging/index.rst
+++ b/providers/opensearch/docs/logging/index.rst
@@ -44,6 +44,22 @@ First, to use the handler, ``airflow.cfg`` must be configured as follows:
     username = <username>
     password = <password>
 
+On Airflow 3.3.0 or above you can also route remote logging to OpenSearch through the
+provider dispatch mechanism by adding an ``opensearch://`` scheme to
+``[logging] remote_base_log_folder``:
+
+.. code-block:: ini
+
+    [logging]
+    remote_logging = True
+    remote_base_log_folder = opensearch://
+
+    [opensearch]
+    host = <host>
+    port = <port>
+    username = <username>
+    password = <password>
+
 To output task logs to stdout in JSON format, the following config could be used:
 
 .. code-block:: ini
--- a/providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
+++ b/providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
@@ -873,6 +873,44 @@ class OpensearchRemoteLogIO(LoggingMixin):  # noqa: D101
 
     processors = ()
 
+    @classmethod
+    def from_config(cls) -> OpensearchRemoteLogIO:
+        """
+        Build the remote log IO from Airflow logging and ``[opensearch]`` configuration.
+
+        The ``opensearch://`` value in ``[logging] remote_base_log_folder`` is only a routing
+        marker, so every connection and behaviour parameter is read from the ``[opensearch]``
+        section here.
+
+        This does not merge ``[logging] remote_task_handler_kwargs`` IO-kwargs, matching the
+        legacy behavior for OpenSearch.
+        """
+        remote_task_handler_kwargs = conf.getjson("logging", "remote_task_handler_kwargs", fallback={})
+        if not isinstance(remote_task_handler_kwargs, dict):
+            raise ValueError(
+                "logging/remote_task_handler_kwargs must be a JSON object (a python dict), we got "
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # ``[opensearch] port`` declares an empty-string default, so the key is always present and
+        # ``conf.getint`` raises on ``int("")`` instead of falling back to 9200.
+        port = conf.get("opensearch", "port", fallback="")
+        return cls(
+            base_log_folder=os.path.expanduser(conf.get_mandatory_value("logging", "base_log_folder")),
+            delete_local_copy=conf.getboolean("logging", "delete_local_logs"),
+            host=conf.get("opensearch", "host", fallback=""),
+            port=int(port) if port else 9200,
+            username=conf.get_mandatory_value("opensearch", "username"),
+            password=conf.get_mandatory_value("opensearch", "password"),
+            write_stdout=conf.getboolean("opensearch", "write_stdout"),
+            write_to_opensearch=conf.getboolean("opensearch", "write_to_os"),
+            json_format=conf.getboolean("opensearch", "json_format"),
+            target_index=conf.get_mandatory_value("opensearch", "target_index"),
+            host_field=conf.get_mandatory_value("opensearch", "host_field"),
+            offset_field=conf.get_mandatory_value("opensearch", "offset_field"),
+            log_id_template=conf.get("opensearch", "log_id_template", fallback="")
+            or "{dag_id}-{task_id}-{run_id}-{map_index}-{try_number}",
+        )
+
     def __attrs_post_init__(self):
         self.host = _format_url(self.host)
         self.port = self.port if self.port is not None else (urlparse(self.host).port or 9200)
--- a/providers/opensearch/provider.yaml
+++ b/providers/opensearch/provider.yaml
@@ -96,6 +96,10 @@ connection-types:
 logging:
   - airflow.providers.opensearch.log.os_task_handler.OpensearchTaskHandler
 
+remote-logging:
+  - classpath: airflow.providers.opensearch.log.os_task_handler.OpensearchRemoteLogIO
+    scheme: opensearch
+
 config:
   opensearch:
     description: ~
--- a/providers/opensearch/src/airflow/providers/opensearch/get_provider_info.py
+++ b/providers/opensearch/src/airflow/providers/opensearch/get_provider_info.py
@@ -60,6 +60,12 @@ def get_provider_info():
             }
         ],
         "logging": ["airflow.providers.opensearch.log.os_task_handler.OpensearchTaskHandler"],
+        "remote-logging": [
+            {
+                "classpath": "airflow.providers.opensearch.log.os_task_handler.OpensearchRemoteLogIO",
+                "scheme": "opensearch",
+            }
+        ],
         "config": {
             "opensearch": {
                 "description": None,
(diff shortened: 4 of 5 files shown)


potiuk:
I reviewed this alongside the four sibling PRs doing the same port for other providers (#70525 elasticsearch, #70301 wasb, #70682 oss, #70549 stackdriver), and this one diverges from the legacy behaviour it is porting.

`airflow_local_settings.py` splits into two camps. The object-storage backends merge the IO half of `remote_task_handler_kwargs`:

```python
REMOTE_TASK_LOG = WasbRemoteLogIO(
    **cast("dict[str, Any]", { ... } | _io_kwargs)
)
```

The two search backends do not. `OpensearchRemoteLogIO` is constructed with explicit keyword arguments and no `| _io_kwargs` at all:

```python
REMOTE_TASK_LOG = OpensearchRemoteLogIO(
    host=OPENSEARCH_HOST,
    port=OPENSEARCH_PORT,
    ...
    log_id_template=OPENSEARCH_LOG_ID_TEMPLATE,
)
```

identical in shape to the `ElasticsearchRemoteLogIO` branch directly above it. This PR adds the merge anyway, so `from_config` forwards every non-`FileTaskHandler` key to `OpensearchRemoteLogIO.__init__`.

That is a behaviour change rather than a port. `remote_task_handler_kwargs` is one config knob whose keys are split between the file handler and the IO object, so an opensearch deployment can currently put arbitrary keys there for the file-handler side and nothing reaches the IO class. After this change those keys are passed to the constructor, and anything it does not accept raises `TypeError: unexpected keyword argument` at startup — on a configuration that works today.

Worth noting #70525 hit the same fork and went the other way, with the reasoning recorded in its docstring:

> Unlike the object-storage backends, this does not merge ``[logging] remote_task_handler_kwargs`` IO-kwargs, matching the legacy behavior for Elasticsearch.

Two ways forward, and I don't mind which:

- Drop the merge here so the port is faithful and opensearch stays consistent with elasticsearch.
- Keep it, if you think the legacy omission was itself an oversight — but then say so in the PR description and docstring, since it changes startup behaviour for existing deployments and shouldn't ride in as an incidental detail of a port.

Everything else here looks like a clean port of the legacy branch.

---

providers/opensearch/docs/logging/index.rst
```diff
@@ -44,6 +44,22 @@ First, to use the handler, ``airflow.cfg`` must be configured as follows:
     username = <username>
     password = <password>
 
+On Airflow 3.x you can also route remote logging to OpenSearch through the provider
+dispatch mechanism by adding an ``opensearch://`` scheme to
+``[logging] remote_base_log_folder``:
```
jason810496:
```suggestion
On Airflow 3.3.0 or above you can also route remote logging to OpenSearch through the provider dispatch mechanism by adding an ``opensearch://`` scheme to ``[logging] remote_base_log_folder``:
```

providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
```diff
@@ -873,6 +874,43 @@ class OpensearchRemoteLogIO(LoggingMixin):  # noqa: D101
...
+                f"{type(remote_task_handler_kwargs)}"
+            )
+        # remote_task_handler_kwargs mixes FileTaskHandler kwargs with IO kwargs; only the
+        # latter belong to this class (same split as airflow_local_settings.py).
+        fth_params = frozenset(inspect.signature(FileTaskHandler.__init__).parameters) - {
+            "self",
+            "base_log_folder",
+        }
+        io_kwargs = {k: v for k, v in remote_task_handler_kwargs.items() if k not in fth_params}
+        port = conf.get("opensearch", "port", fallback="")
```
jason810496, line 896:
```suggestion
        port = conf.getint("opensearch", "port", fallback=9200)
```

jason810496 (commented):
Thanks! I agreed with Jarek's comment overall, we need to follow the existing pattern in 

https://github.com/apache/airflow/blob/182919ac348badad1f235ba058fe52aa86339098/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L357-L387

providers/opensearch/docs/logging/index.rst
FrankYang0529: Updated it. Thanks for the suggestion.

providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
FrankYang0529: I prefer to use `port = conf.get("opensearch", "port", fallback="")`. The default value for port field is empty string. If we use `conf.getint("opensearch", "port", fallback=9200)`, it doesn't fallback to 9200 port.

https://github.com/apache/airflow/blob/4899e9041859db58fabeb87f231e454a5f65a418/providers/opensearch/provider.yaml#L110-L116

It shows error like following:

```
ImportError: Unable to load logging config from airflow.config_templates.airflow_local_settings.DEFAULT_LOGGING_CONFIG due to: AirflowConfigException:Failed to convert value to int. Please check "PORT" key in "opensearch" section. Current value: "".
```

The original `airflow_local_setting.py` path has similar problem. I can create another PR to fix it after we get conclusion here.

FrankYang0529:
> * Drop the merge here so the port is faithful and opensearch stays consistent with elasticsearch.

Thanks @potiuk and @jason810496's review. I prefer to keep behavior consistent with existing pattern.

providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
jason810496: How about using `port = conf.get("opensearch", "port", fallback="") or 9200` here and the `airflow_local_setting.py`.

Yes, let's create another PR to fix the same semantic error in `airflow_local_setting.py`, thanks.

jason810496 (commented):
Thanks for the update. Nice catch! I didn't know this.

providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
FrankYang0529: Thank you. I use 9200 as default in `port=int(port) if port else 9200` and keep `port = conf.get("opensearch", "port", fallback="")`.

jason810496 (approved):

providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
FrankYang0529: Followup PR: https://github.com/apache/airflow/pull/71170

Issue #67236, opened 2026-05-20
Surface resumable operators in the Airflow Registry


kaxil:
## Problem

Users browsing the [Airflow Registry](https://airflow.apache.org/registry/) cannot tell which operators support resumption -- i.e., operators that can pick up from where they left off after a worker crash, restart, or retry, rather than re-running from scratch.

This matters for users planning long-running or expensive workloads (data transfers, ML training, large queries) where re-running from scratch on retry is wasteful or unsafe.

## Proposal

Add a visible indicator on operator pages (and a filter on listing pages) in the Airflow Registry that marks operators as resumable.

Open questions to resolve in this issue:

- **Definition.** What qualifies as "resumable"? Candidates: deferrable operators (trigger-based resumption), operators with explicit checkpointing, idempotent transfer operators that can resume from offset/cursor, or some combination. Pick one definition or model it as multiple flags.
- **Source of truth.** How do we get this signal? Options: a class attribute on the operator (e.g. `resumable: ClassVar[bool] = True`), inference from `BaseOperator` subclass traits (e.g. presence of `defer()` / `execute_complete()`), or curated metadata in `provider.yaml`. A class attribute is the most discoverable and authoritative.
- **Registry surfacing.** Badge on the operator detail page + a filter chip on the operator list page. Tooltip explaining what "resumable" means with a link to docs.

Implementation likely spans two repos:
- `apache/airflow` -- metadata on the operators
- `apache/airflow-site` -- rendering in `landing-pages/` (registry)

## Out of scope

- Implementing resumability in operators that don't have it. This issue is about *surfacing* the capability, not adding it.
- Changing how deferrable operators work.

## Project tracking

https://github.com/orgs/apache/projects/612

PR #70298, opened 2026-07-23
Add durable/deferrable filters to the Airflow Registry

amoghrajesh:
---

- [x] Yes - claude sonnet 4.6

---

### Important Note

**What "durable" and "deferrable" mean in the PR:** there are two different reasons an operator can survive a crash, and they work in totally different ways. "Durable" means the operator's own code remembers where its job was and reconnects to it instead of starting over, and this operates using task state store in **synchronous mode**. "Deferrable" is running using a trigger, in **asynchronous mode**. The end result looks similar to a user (nothing gets redone), but one is a promise the operator has to keep and the other is just how the system already works for everyone.

### What

Follow-up to the merged `supports_durable_execution` badge PR (#67236). Now that we have badge support, we should also have a few filters for ease of navigation for two related-but-distinct capabilities: "durable" (anything that survives a crash without redoing work: `ResumableJobMixin`/manual-marker operators, or deferrable ones) and "deferrable" (the narrower, deferrable-only subset). The registry had no deferrable detection at all before this change.

### Current behaviour

The Registry only tracks `supports_durable_execution` (`ResumableJobMixin` or the manual `__supports_durable_execution` marker). Deferrable operators, a much more common capability across most cloud providers, have no signal anywhere in the pipeline, and there's no way to filter a provider's module list by either capability.

### Proposed change

- **Durable detection**: `is_durable_capable(cls)` in `dev/registry/extract_parameters.py` walks the resolved `execute()` chain from `cls`, following both `super().execute()` and an explicit `ParentClass.execute(self, context)` call (the pattern task decorators like `@task.agent`/`@task.kubernetes` use), to check whether a class that declares the durable marker, or genuinely implements `ResumableJobMixin`, still preserves that behavior after subclassing. Whole-line comments are stripped before matching so a comment that merely mentions delegation isn't mistaken for real code.
- **Deferrable detection**: `supports_deferrable(cls)` added to `dev/registry/extract_parameters.py`. Checking for a `deferrable` constructor parameter isn't enough, since a subclass can inherit the parameter while overriding `execute()` with code that never reads it, and an operator that always defers unconditionally has no such parameter at all. This walks outward from `execute()` into helper methods it calls via `self.<name>(...)`, and up the MRO through delegation, looking for an actual `self.defer(...)` call, a `self.deferrable` read, or a `TaskDeferred` raise. Code that's only reachable on a core older than the one being imported (e.g. a `defer()` fallback behind `if AIRFLOW_V_3_3_PLUS: raise ...`, as in the HITL and common.ai approval operators) is excluded, since it never actually runs on the core the extractor imports.
- **Schema**: `supports_deferrable: bool = False` added to `ModuleContract`, wired into `make_entry()` alongside `supports_durable_execution` (13 fields total, kept as an independent signal, not merged into the existing one).
- **Frontend**: two independent badges on module cards ("Durable", "Deferrable": teal/blue) shown per their own signal, since an operator can have either, both, or neither. Two filter toggles on a provider's module list: "Durable (includes deferrable)" and "Deferrable only".

### Changes of Note

- **The two filters are a deliberate union/subset pair, not independent booleans.** "Durable" is `supports_durable_execution OR supports_deferrable`, checking it shows every deferrable operator plus every `ResumableJobMixin`/manual-marker operator, even ones that aren't deferrable. "Deferrable only" is `supports_deferrable` alone, never OR'd. This means an operator that is both durable and deferrable appears under *both* filters, which is intentional and not confusing in practice, since the two independent badges on the card explain why.
- The checkbox labels ("Durable (includes deferrable)" / "Deferrable only") were chosen specifically to make that asymmetric containment relationship (deferrable is a subset of durable, never the reverse) visible at the point of clicking, rather than relying on a tooltip or a caption someone could miss.
- `is_deferrable` inheriting a parent's `__init__` unmodified still counts as deferrable (unlike the `supports_durable_execution` manual-marker case, which deliberately does *not* propagate to subclasses), since inheriting an unmodified constructor is a real functional guarantee the parameter still works, whereas inheriting a durable-execution claim says nothing about whether an overridden `execute()` still preserves it.
- This PR also widens `is_durable_capable()` itself, not just adds the new deferrable signal alongside it. That function's existing marker-lookup logic had three separate false-negative bugs (an MRO walk that lost track across multiple inheritance, a lookup that gave up on a class with no own `execute` override, and a name-mangling calculation that broke on classes with a leading underscore), all now fixed. This changes the already-shipped Durable badge's output for a few real operators, most notably `EksPodOperator` and `GKEStartPodOperator`, which were previously misreported as not durable-capable and now correctly show the badge. `SparkKubernetesOperator` also reports durable, correctly, but for a reason the detector doesn't actually check (see the docstring on `_delegates_execute_to`); its deferrable path never delegates via `super()` at all, and is durable only because `KubernetesPodOperator`'s `execute_sync`/`execute_async` both independently call the pod-reattach mechanism.

### Testing

- Built breeze to extract the provider data using: `breeze registry extract-data`

- Built the actual Eleventy site against this real data and confirmed in the rendered HTML: both badges appear on the right operator cards (e.g. Databricks' page shows both a "Durable" and "Deferrable" badge on `DatabricksRunNowOperator`), both filter checkboxes render with the intended labels, and the `data-durable`/`data-deferrable` attributes are correctly populated per the OR/non-OR logic on every module checked.

For current changes, this is the behaviour:

- **Databricks**: `DatabricksSubmitRunOperator` and `DatabricksRunNowOperator` are both durable *and* deferrable. `DatabricksNotebookOperator`, `DatabricksSQLStatementsOperator`, `DatabricksTaskBaseOperator`, `DatabricksTaskOperator`, and `DatabricksSQLStatementsSensor` are deferrable only. The rest of the provider's operators/hooks/sensors are neither.
- **Apache Spark**: `SparkSubmitOperator` is durable but *not* deferrable (no `deferrable` param) -- this is the single durable-only case across the entire registry.
- **Google/BigQuery**: `BigQueryInsertJobOperator` is both durable and deferrable, same pattern as Databricks.
- **Amazon/Redshift**: `RedshiftDataOperator` is both durable and deferrable.
- **Snowflake**: `SnowflakeSqlApiOperator` is both durable and deferrable.

Across the full dataset: 1779 modules total, 6 durable, 217 deferrable, 5 with both (the five listed above). 

### Manual verification (`breeze registry extract-data`, generated `dev/registry/modules.json`)

Every operator raised across all review rounds, checked against the freshly generated data:

| Operator | durable | deferrable | Verdict |
|---|---|---|---|
| `EksPodOperator` | True | True | Durable fix confirmed |
| `GKEStartPodOperator` | True | True | Durable fix confirmed |
| `SparkKubernetesOperator` | True | True | Durable fix confirmed (correct answer, coincidental reasoning per the `_delegates_execute_to` docstring note) |
| `KubernetesPodOperator` | True | True | Baseline unchanged |
| `AgentOperator` | True | False | Baseline unchanged |
| `HITLOperator` | False | False | `defer_for_approval` fix confirmed |
| `ApprovalOperator` | False | False | Fix confirmed |
| `HITLBranchOperator` | False | False | Fix confirmed |
| `HITLEntryOperator` | False | False | Fix confirmed |
| `LLMOperator` | False | False | Fix confirmed |
| `LLMBranchOperator` | False | False | Fix confirmed |
| `LLMSQLQueryOperator` | False | False | Fix confirmed |
| `LLMSchemaCompareOperator` | False | False | Fix confirmed |
| `LLMFileAnalysisOperator` | False | False | Fix confirmed |
| `DiscordWebhookOperator` | False | False | No false-positive regression |
| `HttpOperator` | False | True | Real deferral still detected |
| `DatabricksRunNowOperator` | True | True | Matches PR description example |
| `DatabricksSubmitRunOperator` | True | True | Matches PR description example |
| `KubernetesJobOperator` | False | True | Correctly not durable (own Job path) |
| `GKEStartJobOperator` | False | True | Correctly not durable |
| `SparkJDBCOperator` | False | False | Correctly not durable (drops `execute_resumable`) |

Some screenshots:

In this image, "Durable (includes deferrable)" is selected, and you can see both durable + deferrable (BigQueryInsertJobOperator) and just deferrable ones selected

image

"Deferrable only selection":

image

Operator like snowflake has only one operator with both deferrable and durable property, so its here:

image

Spark is one of few operators with only "durable", and no deferrable option:

image

### What's next

Interactive click-through verification (actually toggling the checkboxes in a running browser to confirm `filterModules()` behaves correctly) hasn't been done -- only the generated HTML/data have been checked, not runtime JS behavior. Registry-wide (cross-provider) discovery -- "which are all the durable/deferrable operators across the whole registry," via a Pagefind search filter facet -- was discussed and deliberately deferred to a separate follow-up.

---


dev/registry/extract_parameters.py
```diff
@@ -467,7 +473,7 @@ def make_entry(
         category: str = "",
         transfer_desc: str | None = None,
     ) -> dict:
-        """Build a full module entry dict with all 12 fields."""
+        """Build a full module entry dict with all 13 fields."""
```
ashb: Nit: lets remove the number of fields ;)

amoghrajesh: 🫡 ee0ee7b224

registry/src/provider-version.njk
```diff
@@ -340,6 +340,16 @@ eleventyComputed:
     <section class="modules">
       <div class="modules-header">
         <h2>Modules</h2>
+        <div class="capability-filter-toggles">
```
kaxil, line 388: These render on every version page, but only the `isLatest` branch gets modules from `modules.json`. Older versions use `pv.versionData.modules`, which `extract_versions.py` builds by AST and which only carries 7 fields per module: name, type, import_path, short_description, docs_url, source_url, category. I checked `registry/src/_data/versions/amazon/9.17.0/metadata.json`: 0 of 370 modules have `supports_durable_execution`, and `supports_deferrable` won't be there either. So on any page built from those payloads (1349 in the current `_data/versions` snapshot) both data attributes are `false` and ticking either box empties the whole grid with no explanation. The badge degraded gracefully because it just doesn't render, a filter doesn't. Could the toggles be gated on `pv.isLatest`, or on at least one module qualifying?

registry/src/css/main.css
```diff
@@ -3170,6 +3170,26 @@ main {
   max-width: 16rem;
 }
 
+.capability-filter-toggles {
```
kaxil, line 3301: `.modules-header` is a flex row with `justify-content: space-between`, no `flex-wrap`, and no breakpoint, and `.search-wrapper` is `flex: 1; min-width: 0` (line 1021). The two new labels are `white-space: nowrap` and come to roughly 340px between them, so all of that comes out of the module search box. At phone widths the search input collapses to about the width of its icon. `flex-wrap: wrap` on `.modules-header` would sort it.

registry/src/css/main.css
```diff
@@ -3263,10 +3284,18 @@ main {
...
+}
+
+.provider-detail-page .modules .module .content h3 .durable-badge {
   background: color-mix(in srgb, var(--color-teal-500) 15%, transparent);
   color: var(--color-teal-400);
 }
 
+.provider-detail-page .modules .module .content h3 .deferrable-badge {
+  background: color-mix(in srgb, var(--color-blue-500) 15%, transparent);
+  color: var(--color-blue-400);
```
kaxil: `--color-blue-400` (#60a5fa) on a 15% `--color-blue-500` wash is 6.1:1 in dark mode but 2.14:1 over a white card, at `--text-xs`. The durable badge has the same issue (1.62:1). Since you've already split the shared rule from the per-badge colors here, both could use `light-dark()` the way tokens.css and line 1700 do. `light-dark(var(--color-blue-600), var(--color-blue-400))` gets light mode to 4.35:1.

registry/src/provider-version.njk
amoghrajesh: Gated the whole block on `pv.isLatest` rather than "at least one qualifying module" to target exactly the reported failure mode without the extra complexity.

registry/src/css/main.css
amoghrajesh: Added `flex-wrap: wrap;` to `.modules`header. Confirmed nothing else in that container depends on a single-row layout.

registry/src/css/main.css
amoghrajesh: Nice, fixed both badges with `light-dark()`, matching the pattern at `.category-card.color-blue .provider-badge`. One addition needed: there was no `--color-teal-600` in tokens.css and I checked the math - `teal-500` alone only gets to `~2.5:1` against a white card, nowhere near AA. Added `--color-teal-600: #0d9488` so the durable badge could follow the same `light-dark(var(--color-teal-600)`, var(--color-teal-400))` pattern you used for blue.

amoghrajesh: Handled 5b1b609ddc

registry/src/css/main.css
amoghrajesh: Handled 5b1b609ddc

registry/src/provider-version.njk
amoghrajesh: Handled 5b1b609ddc

registry/src/provider-version.njk
```diff
@@ -383,6 +383,19 @@ eleventyComputed:
     <section class="modules">
       <div class="modules-header">
         <h2>Modules</h2>
+        {% if pv.isLatest %}
```
kaxil: The `isLatest` gate closes the older-version path, but there's a second route to the same dead toggle it doesn't cover, and this one hits the pages people actually land on.

The live catalogue is rebuilt incrementally. `publish-docs-to-s3.yml:655` passes specific provider ids into `registry-build.yml`, which pulls the published `modules.json` down from S3, and `merge_registry_data.py:83-84` keeps every non-targeted provider's module dicts verbatim. `_validate` (`registry_contract_models.py:233-235`) is `model_validate(payload); return payload`, so it never writes the Pydantic defaults back into the dict. So after the next provider release, every provider that wasn't in that build still has `pv.isLatest` true and no `supports_deferrable` key at all, and ticking either box blanks its grid.

It isn't only a stale-data problem either. From the current tree, 17 of 105 provider packages contain a class declaring a `deferrable` parameter, and the durable marker appears in exactly 2 (`pod.py:288`, `agent.py:226`), so 87 packages have neither. Amazon and Google alone account for most of the deferrable classes. Both toggles are dead controls on the large majority of provider pages even after a full rebuild.

Counting in the template and rendering each box only when its count is non-zero fixes the stale-data case and the genuinely-empty case together, and self-heals when a rebuild lands. The accumulate-in-a-loop pattern already works in this file (`totalModules` at :38-39), and the type tabs 28 lines down already gate themselves the same way (`{% if moduleCounts[t.id] > 0 %}` at :358). Putting the count in the label saves a click too. Worth pairing with the empty state from `provider-filters.js:81-94`, which already hides the grid and writes a `#filter-status` count for the providers list; the module grid has no equivalent, and the "no modules" banner at :454-465 is keyed off `totalModules` so client-side filtering never reaches it.

registry/src/provider-version.njk
```diff
@@ -383,6 +383,19 @@ eleventyComputed:
...
+        {# pv.versionData.modules (older versions) lacks these fields entirely. #}
+        <div class="capability-filter-toggles">
+          <label class="capability-filter-toggle" title="Reconnects on retry (ResumableJobMixin) or frees the worker slot while waiting (deferrable) instead of resubmitting or blocking.">
```
kaxil: "Reconnects on retry (ResumableJobMixin)" names a mechanism neither durable-marked operator actually uses. `KubernetesPodOperator` is `(BaseOperator)` (`pod.py:142`) and `AgentOperator` is `(BaseOperator, HITLReviewMixin)` (`agent.py:120`); both carry the manual marker with a comment right above it saying "supports durable execution directly, without ResumableJobMixin" (`pod.py:286-288`, `agent.py:224-226`). `is_durable_capable` has the two paths precisely so those operators qualify without the mixin, so the mixin is the one thing the currently-badged modules have in common only by not using it. The shipped badge copy at :418 describes the behaviour without naming an SDK class, which reads better for a site visitor anyway.

While you're in here, the Deferrable badge tooltip on :421 says "trigger state is persisted and reassigned automatically", which is a bit stronger than what happens: the row persists the classpath and the serialized kwargs, not progress, so a reassigned trigger re-runs `run()` from the start rather than resuming. `deferring.rst:170` tells trigger authors to assume exactly that ("a trigger instance can run *more than once*").

dev/registry/extract_parameters.py
```diff
@@ -432,6 +433,11 @@ def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
     return "execute_resumable" in source
 
 
+def supports_deferrable(cls: type) -> bool:
+    """Return True if a class exposes a `deferrable` constructor parameter."""
```
kaxil: `get_params_from_class` unions every ancestor's `__init__` params, so this answers "does some ancestor name `deferrable`" rather than "can this defer". Because the union ignores who actually *reads* the parameter, it currently publishes at least one capability claim that isn't true.

`DiscordWebhookOperator` is `class DiscordWebhookOperator(HttpOperator)` and is registered in the catalogue (`discord/provider.yaml:74-77`). Its own `__init__` declares no `deferrable` and forwards `**kwargs` up (`discord_webhook.py:57-70`), so the union picks the param up from `HttpOperator.__init__` (`http/operators/http.py:121`). But the only thing that reads it is `HttpOperator.execute`'s `if self.deferrable: self.execute_async(...)` (`http.py:164-168`), and `DiscordWebhookOperator.execute` overrides that outright with a synchronous `self.hook.execute()` (`discord_webhook.py:95-102`). So it gets the Deferrable badge, matches both new checkboxes, and ships `supports_deferrable: true` in `/api/modules.json`, while passing `deferrable=True` frees no worker slot at all. An inherited knob whose only reader has been overridden is dead.

It misses in the other direction too, which is what makes this a limit of the proxy rather than a patchable edge case. An operator that defers *unconditionally* has no flag to expose, so there is no param to find. `DateTimeSensorAsync` is the clearest: its docstring is "Deferring itself to avoid taking up a worker slot", its `execute` body is a bare `self.defer(...)`, and `deferrable` appears nowhere in the file (`sensors/date_time.py:83-143`, registered at `standard/provider.yaml:102`). I ran the predicate against `DateTimeSensorAsync`, `MSGraphAsyncOperator`, `AwaitMessageSensor`, `SSHRemoteJobOperator` and `HITLOperator`: all five return False while each one's own `execute` or `poke` calls `self.defer(`. So "Deferrable only" hides exactly the always-deferrable operators, and "Requires deferrable=True" on the :421 tooltip is unfollowable for them since there is no such param to set. A sweep suggests a handful of other always-deferring classes worth a spot check, though I only confirmed these five end to end.

`is_durable_capable` right above already takes the opposite position on inheritance, and says why: the marker is deliberately name-mangled so it is *not* inherited, because a subclass that overrides `execute()` "may not preserve the parent's task_state_store reconnect behavior, so the declaration must not be inherited". That reasoning applies verbatim here, and this predicate is the neighbour that ignores it.

I don't have a clean bound to propose. I tried the three that look obvious and each has a named counter-example in the tree, so this is probably worth deciding deliberately rather than patching:

- Grep the class's own `execute` for `self.defer(`, the way `is_durable_capable` greps for `execute_resumable`: false-negatives `TriggerDagRunOperator`. Its `execute` is `trigger_dagrun.py:233-302` and delegates, so the `if self.deferrable: self.defer(` sits in `_trigger_dag_af_2` at `:396-397`, one call away.
- Grep the whole class body instead: false-negatives `TimeDeltaSensorAsync` (`sensors/time_delta.py:154-167`, registered at `provider.yaml:103`), which inherits the param *and* the deferring `execute` from `TimeDeltaSensor`, so its own body contains neither `self.defer(` nor `self.deferrable`.
- Exclude classes that override `execute` at all: drops `HttpOperator` itself, which defines its own.

A per-class opt-in marker mirroring `__supports_durable_execution` would be consistent with the durable path and immune to all three, at the cost of provider-side annotation. Whichever way you go, the Discord shape is worth a regression test: forwards `**kwargs` to a deferrable parent, overrides `execute` synchronously, must classify as non-deferrable.

registry/src/css/main.css
```diff
@@ -3468,8 +3490,18 @@ main {
   font-size: var(--text-xs);
   font-weight: var(--font-medium);
   border-radius: var(--radius-lg);
+}
+
+.provider-detail-page .modules .module .content h3 .durable-badge {
   background: color-mix(in srgb, var(--color-teal-500) 15%, transparent);
-  color: var(--color-teal-400);
+  color: var(--color-teal-600);
+  color: light-dark(var(--color-teal-600), var(--color-teal-400));
```
kaxil: This is a real improvement, the light-mode durable badge was effectively unreadable at 1.62:1 before. It just doesn't quite clear the line: compositing the 15% `teal-500` wash onto the card's own `--color-card-bg`, which resolves to `#ffffff` in light mode (`tokens.css:250`), `teal-600` comes out at 3.26:1, against the 4.5:1 that 12px `--font-medium` text needs. `teal-700` (#0f766e) gets it to 4.77:1 on the same composite. For what it's worth my last comment under-called the blue one too: `blue-600` is 4.35:1, just under the line, where `blue-700` (#1d4ed8) would be 5.64:1. Dark mode is comfortable either way, 7.9:1 and 6.1:1.

registry/src/js/provider-detail.js
```diff
@@ -25,6 +25,9 @@
   var extraDepsList = document.getElementById('extra-deps-list');
   var extrasDataEl = document.getElementById('extras-data');
   var moduleSearch = document.getElementById('module-search');
+  var durableOnlyFilter = document.getElementById('durable-only-filter');
+  var deferrableOnlyFilter = document.getElementById('deferrable-only-filter');
+  var moduleTabs = document.querySelectorAll('.module-tab');
```
kaxil: This one looks like it came back in with the main merge rather than on purpose. #70190 narrowed the selector to `.module-tab[data-type]`, which is still there on line 33, so the branch's older copy landing above it leaves two `var moduleTabs` in the same scope. Nothing breaks today since the second assignment wins, but the dead line is the exact selector the comment on 31-32 exists to warn against: the natural instinct on seeing two near-identical declarations is to delete the second one, and then the More button joins `moduleTabs` and clicking it resets `currentType` to `'all'`.

registry/src/provider-version.njk
amoghrajesh: Good catch, the `isLatest` gate only covered one of the two failure modes you found. Switched to counting qualifying modules directly in the template (mirroring the existing `totalModules` accumulator and the `moduleCounts[t.id] > 0` gate the type tabs already use) and rendering each toggle only when its count is non-zero. This covers the stale data case (a field that's missing or false everywhere computes to a zero count) and the genuinely empty case with the same check, and self-heals after every rebuild without needing to special-case either. Put the count in each label too, per your suggestion.

Verified against real data: Databricks shows both toggles at `(7)`/`(7)`; Apache Spark shows only the Durable toggle at `(1)` with no Deferrable toggle in the DOM at all (it has zero deferrable operators); a provider with neither capability renders no toggle block whatsoever. Also confirmed the toggles still compose correctly with the existing type/category/search filters. (see image for DBX)

One piece from your comment I have not done yet: the empty state message for when a filter combination client side narrows to zero results (right now it just goes blank, same as before). Happy to add that here too if you'd rather not split it into a separate follow-up.

image


image

registry/src/provider-version.njk
amoghrajesh: Good catch on both. Fixed the "Durable" filter checkbox's tooltip to drop the `ResumableJobMixin` mention — 

Also reworded the "Deferrable" badge tooltip.

registry/src/css/main.css
amoghrajesh: Sure, my numbers were against a plain white background, not the actual composited card background, so they were off I guess.

Bumped both to `-700: --color-teal-700` & `--color-blue-700 (#1d4ed8, ~5.64:1)`, added to tokens.css and swapped into the `light-dark()` calls for both badges.

registry/src/js/provider-detail.js
amoghrajesh: My bad, fixed it

dev/registry/extract_parameters.py
amoghrajesh: Very good catch. I am working a solution in my head very similar to `is_durable_capable`

amoghrajesh: Replaced the whole approach now. Instead of checking whether a deferrable parameter exists anywhere in the class's ancestry, `supports_deferrable()` now resolves execute via getattr (the same MRO-following resolution is_durable_capable already uses) and checks that method's actual source for `self.deferrable` or `self.defer`. That directly answers "does the code that runs for this class use deferral" instead of proxying through whether a setting merely exists somewhere in the hierarchy.

Verified against every class you named, and a couple more from the same family, using the real code (not just fixtures): HttpOperator -> True, DiscordWebhookOperator -> False (the false positive is gone), DateTimeSensorAsync/TimeDeltaSensorAsync/AwaitMessageSensor/HITLOperator/MSGraphAsyncOperator/SSHRemoteJobOperator -> all True. TimeDeltaSensorAsync in particular is now handled correctly because it resolves through the real MRO rather than only checking each class's own body and that's also why it doesn't fall into the same trap as the "grep whole class body" option you ruled out.

`TriggerDagRunOperator` still can't be found by source grep alone, since its `self.defer` call is one hop away inside `_trigger_dag_af_2`. Added it to a small, explicitly commented `_DEFERRABLE_EXCEPTIONS` set rather than trying to chase call graphs. Went with this over the full per-class opt-in marker you mentioned, since the source grep now has zero known false positives and covers everything except this one exception.

Added the Discord shape regression test you asked for, plus five more covering each shape above. Ran it against every provider in the tree (1954 modules, zero errors) as a sanity check at scale.

amoghrajesh: Handled in 03b18d4da9

registry/src/js/provider-detail.js
amoghrajesh: Handled in [handling simple review comments](https://github.com/apache/airflow/pull/70298/commits/7d3cba4b39510a6badba4d44c7314bc10a05c4dd)

registry/src/css/main.css
amoghrajesh: Handled in [handling simple review comments](https://github.com/apache/airflow/pull/70298/commits/7d3cba4b39510a6badba4d44c7314bc10a05c4dd)

registry/src/provider-version.njk
amoghrajesh: Handled in [handling simple review comments](https://github.com/apache/airflow/pull/70298/commits/7d3cba4b39510a6badba4d44c7314bc10a05c4dd)

registry/src/provider-version.njk
amoghrajesh: Handled in [handling simple review comments](https://github.com/apache/airflow/pull/70298/commits/7d3cba4b39510a6badba4d44c7314bc10a05c4dd)

dev/registry/extract_parameters.py
```diff
@@ -432,6 +433,41 @@ def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
...
+
+    execute = getattr(cls, "execute", None)
+    if execute is None:
+        return False
+    try:
+        source = inspect.getsource(execute)
+    except (OSError, TypeError):
+        return False
+
+    return "self.deferrable" in source or "self.defer" in source
```
kaxil: Ran this against the tree (1292 registered operator, sensor and transfer classes): `DiscordWebhookOperator` stays False, so the false positive is gone, but the exception list covers one of four delegation shapes and eleven registered classes that defer come back False:

- `raise TaskDeferred(` with no `self.defer` anywhere: `VespaIngestOperator.execute` (`vespa_ingest.py:85`, `vespa/provider.yaml:53`). `self.defer` is already a prefix of `self.deferrable`, so the first clause on this line is dead; `"TaskDeferred" in source` is the one worth having.
- A child `execute` wrapping `super().execute(context)`: `EksPodOperator` (`eks.py:1244`; the branch is in KPO's `execute` at `pod.py:803`) and the deprecated `GlueCrawlerOperator` (`glue_crawler.py:317` into `GlueCrawlerRunOperator.execute` at `:182`). Following `super().execute(` up the MRO fixes both without touching Discord, whose `execute` never calls super.
- `self.defer` behind a helper: `DatabricksStartWarehouseOperator` and `DatabricksStopWarehouseOperator` (`warehouse.py:183`, `:209` into `_wait_or_defer` at `:117-118`); `DatabricksTaskBaseOperator` via `monitor_databricks_job` to `_defer_on_run` (`databricks.py:2197`, `:2185-2186`, `:2071`), inherited unchanged by `DatabricksNotebookOperator` and `DatabricksTaskOperator`; `TeradataComputeClusterProvision/Resume/SuspendOperator` via `_compute_cluster_execute` to `_handle_cc_status` (`teradata_compute_cluster.py:168`, unconditional).

With the count in the label the misses are visible: the Databricks page renders `(4)` at HEAD where your round-2 screenshot showed `(7)`, and the PR description's own "deferrable only" list (`DatabricksNotebookOperator`, `DatabricksTaskBaseOperator`, `DatabricksTaskOperator`) is False at HEAD. The description also still describes the constructor-parameter detector and `get_params_from_class`, so it wants a refresh once this settles.

A bounded walk from the resolved `execute` (follow `self.<name>(` through `getattr`, follow `super().execute(` to the next `execute` in the MRO, stop on `self.defer`/`self.deferrable`/`TaskDeferred`) goes from 243 to 254 True across those 1292 classes, the eleven above exactly, with `TriggerDagRunOperator` found without the list and Discord still False since its only self-call is `self.hook.execute()`. If the list stays instead, it needs the ten call-chain names and a test that fails when a listed class starts matching the grep. Separately, match `self.defer` on a word boundary: `self.defer_for_approval(` in the five common.ai LLM operators (`llm.py:169`) counts today by prefix accident. Whether they should carry the badge is its own call, since on 3.3+ that path raises `TaskAwaitingInput` with no trigger at all (`approval.py:168-172`) and only pre-3.3 cores fall back to `self.defer(HITLTrigger)`, so the Triggerer promise in the tooltip is wrong for them on current cores.

dev/registry/extract_parameters.py
```diff
@@ -463,7 +499,7 @@ def discover_classes_from_provider(
     """Discover classes from a single provider by importing its modules at runtime.
 
     Reads the provider.yaml to find which modules/classes to inspect, imports them,
-    and returns metadata for each discovered class with all 12 Module fields.
+    and returns metadata for each discovered class with all 13 Module fields.
```
kaxil, line 629: Ash's nit about the field count landed on `make_entry` but not here or on `test_has_all_13_fields` (`test_extract_parameters.py:362`); both bump every time a field is added.

registry/src/provider-version.njk
```diff
@@ -403,6 +427,9 @@ eleventyComputed:
               {% if module.supports_durable_execution %}
               <span class="durable-badge" title="Reconnects to an already-running job on retry instead of resubmitting. Requires durable=True (default) and depends on operator configuration.">Durable</span>
               {% endif %}
+              {% if module.supports_deferrable %}
+              <span class="deferrable-badge" title="Frees the worker slot while waiting. If the Triggerer crashes, another one picks it up using the same trigger class and arguments. Requires deferrable=True.">Deferrable</span>
```
kaxil: With the new predicate the badge now lands on operators that defer unconditionally and have no `deferrable` parameter, so "Requires deferrable=True" is an instruction that fails for them. `DateTimeSensorAsync` (`date_time.py:106-114`) and `HITLOperator` (`hitl.py:199-244`) forward `**kwargs` to BaseOperator, which raises `TypeError("Invalid arguments were passed ...")` on the leftover key (`operator.py:1103-1107`); `TimeSensorAsync` hard-codes `super().__init__(deferrable=True, **kwargs)` (`time.py:128`), so passing it again is a duplicate-keyword TypeError. Something like "Pass deferrable=True where the operator exposes it; some defer unconditionally" covers both, and the same phrase sits on the filter tooltip at :402.

registry/src/css/tokens.css
```diff
@@ -96,6 +97,8 @@
   --color-rose-600: #e11d48;
   --color-teal-400: #2dd4bf;
   --color-teal-500: #14b8a6;
+  --color-teal-600: #0d9488;
```
kaxil: `--color-teal-600` was added for the previous round and nothing references it now that the badge moved to `teal-700`; it can go.

registry/src/provider-version.njk
kaxil: A separate follow-up for the empty state is fine, no need to grow this PR for it. The count-gated toggles cover both cases I raised here.

amoghrajesh: Already fixed - the toggles now hide themselves when empty. Nothing to do.

dev/registry/extract_parameters.py
amoghrajesh: Rewrote the deferrable detector to actually trace through helper methods and parent classes instead of just text-matching. Tested it against real code in the repo it now correctly catches all the operators you listed, and still correctly excludes Discord.

dev/registry/extract_parameters.py
amoghrajesh: Found a test that would silently miss it if 2 fields got deleted. Fixed it to check for all 13.

registry/src/provider-version.njk
amoghrajesh: Fixed the tooltip text so it doesn't say "you must pass this" for operators where passing it would actually crash.

registry/src/css/tokens.css
amoghrajesh: Deleted the unused color leftover from an earlier round.

amoghrajesh: Handled in ef2d05c5bf

registry/src/provider-version.njk
amoghrajesh: Handled in ef2d05c5bf

dev/registry/extract_parameters.py
amoghrajesh: Handled in ef2d05c5bf

dev/registry/extract_parameters.py
amoghrajesh: Handled in ef2d05c5bf

dev/registry/extract_parameters.py
```diff
@@ -391,28 +392,138 @@ def load_resumable_job_mixin() -> type | None:
         return None
 
 
+# Matches an actual self.defer() call or self.deferrable attribute read, but not
+# self.defer_for_approval(). TaskDeferred catches operators that raise it directly instead of
```
kaxil, line 398: The regex is right, the outcome isn't. `self\.defer\(` no longer matches `self.defer_for_approval(`, but the `_SELF_CALL_RE` walk below follows that call into `LLMApprovalMixin.defer_for_approval`, whose fallback branch is a literal `self.defer(HITLTrigger(...))` at `approval.py:178`. I ran the detector at HEAD across the 1201 registered operator, sensor and transfer classes and logged the match site for each hit: all five common.ai LLM operators come back True on that line. `LLMOperator`, `LLMBranchOperator`, `LLMSQLQueryOperator` and `LLMSchemaCompareOperator` reach it as `execute -> self.defer_for_approval()`, and `LLMFileAnalysisOperator` one hop further through `LLMOperator.execute`.

That branch is dead where the extractor runs. `defer_for_approval` takes the `AIRFLOW_V_3_3_PLUS` path and raises `TaskAwaitingInput` with no trigger at all, and I get `AIRFLOW_V_3_3_PLUS is True` importing it from the core the extractor imports. The call also sits behind `if self.require_approval:` (`llm.py:167-169`). The five still ship `supports_deferrable: true`, a Deferrable badge whose tooltip promises another Triggerer picks the work back up, and a slot in both filter counts.

Whichever way you land on the badge itself, `test_defer_for_approval_does_not_false_match` (`test_extract_parameters.py:562`) is the part worth fixing first: `DefersForApprovalOnly.defer_for_approval` at :510 has a `return None` body, so it passes against the stub and cannot fail against the real shape. Giving that fixture's helper a `self.defer(` body, the way the Discord fixture puts the real behaviour in the overriding method, makes it test the thing it's named for.

registry/src/css/main.css
```diff
@@ -3297,6 +3298,26 @@ main {
   max-width: 16rem;
 }
 
+.capability-filter-toggles {
```
kaxil, line 3301: This group can't shrink, so it runs past the viewport on a phone. It's `display: flex` with no `flex-wrap`, its children are `white-space: nowrap` (:3314), and nothing sets `min-width: 0`, so its width is pinned to its content width however narrow the parent gets. I rendered the real bundle and measured it: 370.1px with single-digit counts, 384.5px at `(24)`/`(18)`, 394.3px at Amazon's scale, and still 384.5px when I force `.modules-header` down to 310px. `.container` is 358px wide inside a 390px viewport (`--space-4` each side), so it overflows on every provider page that renders these, worst on the big providers.

The `flex-wrap: wrap` from the earlier round is on `.modules-header`, which only buys the group its own line under the h2. It doesn't let the group itself break. The same property here would. Putting the counts in the label was my suggestion and it's what pushed the labels past the width where this stays invisible, so apologies for the round trip.

dev/registry/extract_parameters.py
```diff
@@ -391,28 +392,138 @@ def load_resumable_job_mixin() -> type | None:
...
-    `__supports_durable_execution` in its own body qualifies this way.
-
-    Inheriting the mixin alone is not sufficient for the second path: a
-    complete override is inert unless execute() actually calls
-    execute_resumable().
+    1. A class-level `__supports_durable_execution = True` declaration (for operators like
+    KubernetesPodOperator/AgentOperator that implement this directly against
+    task_state_store, without ResumableJobMixin). Inherited by a subclass that hasn't
+    replaced the declaring class's `execute()`, whether by not overriding it at all, or by
+    delegating back via `super().execute()`.
```
kaxil, line 544: This widens a badge that already shipped, and I can't find it called out anywhere in the PR. Running main's `is_durable_capable` and this one over the same 1201 classes, main marks 10 durable and this branch marks 13, with nothing lost. The three new ones are `EksPodOperator`, `GKEStartPodOperator` and `SparkKubernetesOperator`. The first two look like real false negatives getting fixed, since GKE defines no `execute` of its own and Eks ends every path in `super().execute(context)`.

`SparkKubernetesOperator` is the one I'd look at twice, because it's the class the old docstring named as the reason the marker must not be inherited. `_delegates_execute_to` reads any `super().execute(` in the source as delegation, but that `execute` returns out of `execute_async(context)` before reaching it whenever `self.deferrable` is set (`spark_kubernetes.py:357-361`). The badge is still correct for it, just not for the reason this check tests: what persists and reattaches the pod is `get_or_create_pod`, and both of KPO's `execute_sync` and `execute_async` call it. So the rule is looser than the docstring reads, and a subclass that branches away from `super()` without that backstop would inherit the claim anyway.

No objection to the widening, it fixes two real misses. Mostly asking whether it wants to be its own PR given it changes output for an already-released badge, and the description's "across the full dataset ... 6 durable" figures want a pass either way.

dev/registry/extract_parameters.py
kaxil: Widening this before you pick a fix, because a `defer_for_approval`-shaped fix does not close it.

The same dead-branch shape is in the standard provider. `HITLOperator.execute` raises `TaskAwaitingInput` under `if AIRFLOW_V_3_3_PLUS:` (`operators/hitl.py:234-241`) and falls through to `self.defer(trigger=HITLTrigger(...))` at :244 only on cores below 3.3. `airflow.providers.standard.operators.hitl` is a registered operators module (`provider.yaml:100`), so all four public classes are in the dataset.

Running HEAD's `supports_deferrable` directly against the installed providers, with `AIRFLOW_V_3_3_PLUS` imported from `airflow.providers.standard.version_compat` evaluating True:

```
supports_deferrable=True   HITLOperator
supports_deferrable=True   ApprovalOperator
supports_deferrable=True   HITLBranchOperator
supports_deferrable=True   HITLEntryOperator
supports_deferrable=True   LLMOperator
supports_deferrable=True   LLMBranchOperator
supports_deferrable=True   LLMSQLQueryOperator
supports_deferrable=True   LLMSchemaCompareOperator
supports_deferrable=True   LLMFileAnalysisOperator
```

Nine classes across two providers, all claiming Deferrable off a branch that cannot execute where the extractor runs.

The standard four are why a name-based exclusion does not close this: their `self.defer(` is in the class's own `execute` source, so `_DEFERRAL_TOKEN_RE.search(source)` matches on the first line of `_references_deferral`, before any walking happens. Whatever you land on has to reason about the branch, not about the name of the call site.

Two landings, and I would take the first:

1. Do not count a `self.defer(` that is only reachable on a core below the one the extractor is running on. Both sites sit behind an `AIRFLOW_V_3_3_PLUS`-style flag, so this is checkable rather than heuristic, and the registry generates each page against a single core anyway.
2. Or decide the badge means "defers on some supported core" and keep all nine, in which case the tooltip has to stop saying another Triggerer picks the work up, because on 3.3+ nothing does: the task parks in `AWAITING_INPUT` and the Core API response handler or the scheduler timeout sweep resumes it.

Either way `test_defer_for_approval_does_not_false_match` still needs the stub fixed, and a second case pinning the HITL shape would keep this from coming back.

kaxil:
Round 5 wrap-up, so everything outstanding is in one place. All of it is verified against `c2d04a7`.

### Blocking

1. The `defer_for_approval` thread, which I have just widened: it is nine classes across two providers, not five, and a fix scoped to `defer_for_approval` leaves the four standard HITL operators wrong.
2. `flex-wrap: wrap` on `.capability-filter-toggles`, which currently overflows 390px on every provider page that renders it.

### Still open from round 4

The field-count nit. The stronger assertion in `ef2d05c` is an improvement, but the nit was the count in the name, and at HEAD there are now three hardcoded copies rather than two: the `discover_classes_from_provider` docstring ("all 13 Module fields"), `test_has_all_13_fields`, and `test_all_13_fields_present` with its 13-element `required_fields` list. `Module` is a `@dataclass`, so `{f.name for f in fields(Module)}` collapses all three and cannot drift. If you would rather keep the literal, say so and I will resolve the thread.

### Satisfied, please resolve

These three are collapsed as outdated in the UI, so they are easy to miss, but they are done:

- The `provider-version.njk` tooltip. HEAD reads "Pass deferrable=True where the operator exposes it; some defer unconditionally", which is the fix.
- `--color-teal-600`. Gone from `tokens.css`, with no references left in `main.css`.
- The delegation-shapes thread. The rewrite covers all four shapes. I re-ran the detector at HEAD: `EksPodOperator`, `GlueCrawlerOperator`, and the Databricks task and warehouse operators all come back True, and `DiscordWebhookOperator` stays False.

### Description

Two statements in the body no longer match the code:

- "checks for a `deferrable` constructor parameter, reusing the existing `get_params_from_class()` machinery". That detector is gone. HEAD walks `execute()` outward through `self.<helper>()` calls and up the MRO through `super().execute()`, looking for deferral tokens.
- "14 fields total". `Module` and `ModuleContract` both carry 13 at HEAD, which is also what both tests assert.

The "1779 modules total, 6 durable, 217 deferrable, 5 with both" line predates both detector rewrites and will move again once the badge question lands, so it is worth re-running at the end rather than now. The Databricks examples still hold; I re-checked the four listed as deferrable-only.

### Not asking for either of these

Listing them so neither surfaces later as a surprise. The `title=` tooltips on the badges and on the new toggles are not keyboard-reachable, but that is a pattern this PR inherits rather than introduces. And the filter checkboxes have the same bfcache and form-restore exposure that `moduleSearch` already has today.

dev/registry/extract_parameters.py
amoghrajesh: Confirmed the wider scope: nine operators across `common.ai` and standard share the same shape, a real `self.defer(...)` gated by a version check that is True on this core. Fixed it so the code actually understands that this line only runs on an older Airflow version, and excludes it when deciding if something is deferrable. Also fixed a case where a decorator was hiding the real module the check needed to look at. Updated the `DefersForApprovalOnly` fixture to carry the real shape instead of a stub, and added a fixture for the direct-in-execute case plus one confirming the reachable case still works.

registry/src/css/main.css
amoghrajesh: Fixed, added `flex-wrap` so the checkboxes wrap onto a new line instead of overflowing.

dev/registry/extract_parameters.py
amoghrajesh: Replaced all three hardcoded counts with one derived from the Module dataclass fields, so they can never drift apart again.

amoghrajesh: re the widening scope, keeping this fix in the current PR since it fixes a real issue either way. Also worth noting: the reasoning for `SparkKubernetesOperator` is coincidental rather than solid; the answer is correct but not for the stated reason. Added a docstring note explaining that so it is not mistaken for a load-bearing invariant. Leaving the durable count in the description as is for now, since it will change again before this settles.

amoghrajesh: Handled in [final review from kaxil](https://github.com/apache/airflow/pull/70298/commits/44bd0106b938e83554f91a5646c1b994ed4b8f18)

registry/src/css/main.css
amoghrajesh: Handled in [final review from kaxil](https://github.com/apache/airflow/pull/70298/commits/44bd0106b938e83554f91a5646c1b994ed4b8f18)

dev/registry/extract_parameters.py
amoghrajesh: Handled in [final review from kaxil](https://github.com/apache/airflow/pull/70298/commits/44bd0106b938e83554f91a5646c1b994ed4b8f18)

dev/registry/extract_parameters.py
```diff
@@ -391,28 +394,159 @@ def load_resumable_job_mixin() -> type | None:
...
+            return base
+    return None
+
+
+def _delegates_execute_to(cls: type, target: type, depth: int) -> bool:
+    """Return True if `cls`'s resolved `execute()` chain reaches `target.execute`.
+
+    Covers a class that never overrides `execute` (inherits `target.execute` directly, e.g.
+    GKEStartPodOperator), one whose override ends in `super().execute(context)` (e.g.
+    EksPodOperator, SparkKubernetesOperator), and one that names the parent class directly
```
kaxil: The docstring note promised in the round-4 reply is not here, and the description points readers at it twice.

Reply 4012645933 said "Added a docstring note explaining that". At this commit `grep -niE "coincid|get_or_create_pod|backstop"` over `dev/registry/` returns nothing, and 44bd0106 did not touch this docstring. The description meanwhile says "for a reason the detector doesn't actually check (see the docstring on `_delegates_execute_to`)", and its table row reads "coincidental reasoning per the `_delegates_execute_to` docstring note". Both send a reader here for an explanation that does not exist.

The line above is the part worth changing. `SparkKubernetesOperator` is listed as an example of "one whose override ends in `super().execute(context)`", which is the reading round 4 questioned. At HEAD its `execute` (`spark_kubernetes.py:352-361`) is:

```python
if self.deferrable:
    self.execute_async(context)
    return

return super().execute(context)
```

On the deferrable path it returns at :359 and never reaches `super().execute(context)` at :361. `_delegates_execute_to` matches it regardless, because `_SUPER_EXECUTE_RE.search` at :462 is a text search with no view of the early return. The verdict is still right, for the reason you gave in the reply: `execute_sync` and `execute_async` reach the pod reattach independently.

Adding the note you described would close it, naming the failure mode it guards against: a subclass whose `execute` neither reaches the marker-declaring class nor goes through the pod reattach, yet inherits the durable claim. The example list above also wants `SparkKubernetesOperator` either dropped or marked as the case that matches for a reason the check does not test, since as written it teaches the opposite.

registry/src/css/main.css
```diff
@@ -3297,6 +3298,27 @@ main {
...
```
kaxil, line 3304: The `flex-wrap` landed here and it does fix this group, but the overflow moved to the sibling rather than going away.

Measured by rendering the built bundle at the widest real label counts the extractor produces (amazon, 115 and 115), with the pre-PR bundle as the baseline. This group's own `scrollWidth - clientWidth` is 0 at every width tested, and at 320, 360 and 390 the toggles wrap to two rows with no page overflow, so the round-4 ask is satisfied.

What changed is the search input beside it. `.module-search-wrapper` at :3297-3299 sets `max-width: 16rem` and no `min-width`, so it becomes the flex item that absorbs the squeeze once the toggles stop shrinking, while `.capability-filter-toggle` children are `white-space: nowrap` at :3316.

```
pre-PR         320-900px : no page overflow, search input 195-256px
HEAD 44bd0106  435-470px : page overflow up to 34px, search input 54px
               530-560px : page overflow up to 32px, search input 54px
```

At 54px the placeholder "Search modules..." no longer fits. These are large-phone-landscape and small-tablet widths rather than phone widths, which is why a check at 390px now reads clean.

Adding `min-width: 12rem` to `.module-search-wrapper` takes the narrowest search input to 193px and leaves no width between 320 and 900 with page scroll, measured the same way. One line if you want it.

dev/registry/extract_parameters.py
```diff
@@ -530,6 +770,7 @@ def make_entry(
             "provider_id": provider_id,
             "provider_name": provider_name,
             "supports_durable_execution": is_durable_capable(cls_or_obj, resumable_mixin),
+            "supports_deferrable": supports_deferrable(cls_or_obj),
```
kaxil, line 818: Decorator entries never receive either capability field, because this is the only path that goes through `make_entry`.

The decorator loop builds its own dict literal (:904-931 at this commit) instead of calling `make_entry`, so every `@task.*` entry ships 11 keys rather than 13 and falls back to the contract defaults. That is all 24 decorators the extractor discovers. Two matter for this feature: `@task.agent` wraps `AgentOperator` and `@task.kubernetes` wraps `KubernetesPodOperator`, both of which the detector reports durable, so those cards can never show a badge or match either filter. The docstrings added this round name `@task.agent` and `@task.kubernetes` as the motivating cases for `_EXPLICIT_EXECUTE_RE` and `_delegates_execute_to`, so the walk was extended for exactly this delegation shape and the result is then dropped.

`test_all_module_fields_present` (`test_extract_parameters.py:1002-1013`) cannot catch it. It derives the required set from `fields(Module)`, which is right, but its fixture sets `"task-decorators": []` at :821, so the decorator path is never exercised. Giving that fixture one decorator entry makes the test fail until the entry either goes through `make_entry` or gains the two fields.

On the description. "This is the single durable-only case across the entire registry" does not hold: your own verification table lists `AgentOperator` at durable True and deferrable False two sections below, which is a second durable-only case, and `SparkSubmitOperator` is not in the table at all. The per-class rows themselves check out, so it is this sentence and the aggregates that are off. For the aggregates I would regenerate rather than patch the numbers, and it is worth checking the run for import warnings while you do: the claimed durable figure is lower than what the pre-merge baseline produces, which points at a run that was dropping classes on import rather than one that is merely stale.

Noticed rather than requested, since it predates this PR and sits outside the diff: the durable badge tooltip says "Requires durable=True (default)". That holds for twelve of the thirteen badged classes, because `ResumableJobMixin` defaults `durable=True` (`resumablejobmixin.py:99`) and `KubernetesPodOperator` resolves `None` to `True` on 3.3+ (`pod.py:417-420`). `AgentOperator` declares `durable: bool = False` (`agent.py:267`), so for that one the tooltip tells a reader the opposite of what they need to do.

dev/registry/extract_parameters.py
```diff
@@ -421,15 +555,121 @@ def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
...
+        return None
     try:
-        source = inspect.getsource(execute)
+        source = inspect.getsource(method)
     except (OSError, TypeError):
+        return None
+
+    # unwrap() undoes a functools.wraps() decorator, whose __globals__ would otherwise
+    # point at the decorator's own module instead of the method's.
+    func = inspect.unwrap(getattr(method, "__func__", method))
```
kaxil, line 622: Half of the round-4 fix depends on this `unwrap`, and nothing tests it.

Neutering `inspect.unwrap` here leaves the whole suite green at 88 passed, yet it is the only reason four of the nine classes come back False. `BaseOperatorMeta._apply_defaults` wraps `HITLOperator.execute`, and the wrapper's `__globals__` does not carry `AIRFLOW_V_3_3_PLUS` while the unwrapped function's does. Both report the same `__module__`, which is why reading `__module__` would not have surfaced it. Without this line the guard flag resolves to nothing, the dead branch is not stripped, and `HITLOperator`, `ApprovalOperator`, `HITLBranchOperator` and `HITLEntryOperator` all go back to True. The five common.ai classes are unaffected, since `defer_for_approval` is not wrapped.

The reason no test covers it is structural rather than an oversight: the file contains no `functools`, no `wraps`, no `__globals__` and no `unwrap` anywhere, so every fixture is a plain method and none can reach this line. The two fixtures added this round do genuinely constrain the stripper, and reverting that half turns exactly `test_defer_for_approval_helper_with_dead_fallback_does_not_false_match` and `test_hitl_shaped_dead_fallback_does_not_false_match` red, so that half is protected. This half is not.

A fixture with a `functools.wraps` wrapper whose `__globals__` belong to a different module than the wrapped function would pin it, and it is worth having, because a silent regression here puts back the exact nine-class false positive this round was about.

Worth knowing that this is not specific to the 3.3 flag. `BaseOperatorMeta` wraps `execute` on nearly every registered class, so the `__globals__` you get back is usually not the method's own, and dozens of classes carry a version flag that only becomes visible after unwrapping, across several `AIRFLOW_V_3_*_PLUS` names rather than just this one. `_is_terminal_block` has the same shape of gap: forcing it to return True also leaves the suite green, though unlike this line it changes no verdict in the tree today.

dev/registry/extract_parameters.py
```diff
@@ -421,15 +555,121 @@ def is_durable_capable(cls: type, resumable_mixin: type | None) -> bool:
...
+
+def _references_deferral(
+    origin: type, current: type, source: str, visited: set[tuple[int, str]], depth: int
+) -> bool:
+    """Return True if `source` (the resolved `execute()` of `current`, called on `origin`) references deferral.
+
+    `origin` stays fixed across recursion so `super().execute()` hops resolve against its
+    real MRO, while `current` walks forward through the chain.
+    """
+    if _DEFERRAL_TOKEN_RE.search(source):
```
kaxil, line 641: Grouped into one comment, since none of these changes a badge today.

This search runs on source with comments and docstrings intact: `_get_reachable_method_source` strips dead version branches but not comments, while the delegation path strips them at :461 via `_strip_comment_lines`, whose docstring names this exact hazard. So a comment mentioning `self.defer()`, a docstring `:raises TaskDeferred:`, an `except TaskDeferred`, or a plain `self.deferrable = False` write all match. Instrumenting the regex to record whether the deciding hit sat inside a comment or string gives zero across the registered set, so nothing is misbadged. The shapes are in the tree, in `MSGraphAsyncOperator.execute_complete` (`msgraph.py:238`) and at `pod.py:1165`, but both sit in resume-side methods on classes that defer for real.

At :646 the helper is resolved against `current`, while `self.<name>()` dispatches on the instance's real class, which is `origin`. `current` advances as `super().execute()` hops are followed, so past the first hop the walk reads an ancestor's copy and cannot see a more derived override. The seeding also differs from the siblings: this walk starts at `_references_deferral(cls, cls, ...)` (:669) while `_delegates_execute_to` (:500) and `_execute_chain_calls_resumable` (:521) start at `_find_owner_of_execute(cls.__mro__, 0)`. Resolving helpers against `origin` across the dataset flips zero verdicts, and the obvious carriers miss it: `EksPodOperator` and `GKEStartPodOperator` both override `invoke_defer_method`, but `KubernetesPodOperator.execute` contains `if not self.deferrable:` (`pod.py:803`), so the match at :625 returns True on that frame before the helper loop runs.

`visited` is marked at :634 and :645 without recording the depth it was searched at, so a node first reached shallow is skipped when re-reached with more budget. That precondition fires 22 times on real data and still flips zero verdicts, because the walk never bottoms out at `depth <= 0`: `_MAX_DEFERRAL_WALK_DEPTH = 6` does not bind on any real chain. That is also why reproducing it requires forcing a chain deeper than 6, and why it becomes live if that constant is raised or a provider grows a deeper chain.

Fixes are one-liners each: call `_strip_comment_lines` inside `_get_reachable_method_source`, tighten `TaskDeferred\b` to `raise TaskDeferred` (which keeps the `VespaIngestOperator` case your comment at :396-398 cites, `vespa_ingest.py:85`), resolve helpers against `origin`, and key `visited` on depth. Worth one test either way: `TestIsDurableCapable` has `test_comment_mentioning_delegation_is_not_mistaken_for_delegation` at `test_extract_parameters.py:369`, and `TestSupportsDeferrable` (:584-651) has fourteen tests with no comment or docstring case.

dev/registry/tests/test_extract_parameters.py
```diff
@@ -266,12 +322,335 @@ def test_manual_durable_marker_qualifies_without_mixin(self):
...
+        be fooled by it into reporting capable."""
+
+        class CommentedNonDelegatingSubclass(ManuallyDurableOperator):
+            def execute(self, context):
+                # overrides execute rather than calling super().execute(),
+                return None
+
+        assert is_durable_capable(CommentedNonDelegatingSubclass, FakeResumableJobMixin) is False
+
+    def test_decoy_execute_call_does_not_shadow_real_delegation(self):
```
kaxil: This passes for a different reason than its name, so it does not currently bind the behaviour it describes.

The decoy is `cursor.execute("select 1")`, and `cursor` is a local variable rather than a class in the MRO. `_next_execute_hop` builds `mro_by_name` from `origin.__mro__` class names, so `mro_by_name.get("cursor")` returns None and the loop takes the `continue` at :468. It never reaches the try-every-match-in-order behaviour that the docstring at :456-458 describes, which is the thing the test is named for. Changing that loop to return on the first resolved match leaves the whole suite green.

A decoy that names a class which is in the MRO but owns no `execute`, placed before the real delegation, would make it bind.

dev/registry/extract_parameters.py
amoghrajesh: Added the note. It now says the `super().execute(` check is a text search with no view of control flow, names `SparkKubernetesOperator` as the live case that returns before reaching that call, explains the verdict is still right because `execute_sync` and `execute_async` both reach the pod reattach independently, and states the gap: a subclass that reaches neither the marker class nor that reattach would inherit the claim without earning it. Also dropped `SparkKubernetesOperator` from the `super().execute()` example list, since as written it taught the opposite.

registry/src/css/main.css
amoghrajesh: Added min-width: 12rem to `.module-search-wrapper`, so it stops being the item that absorbs the squeeze once the toggles stop shrinking.

dev/registry/extract_parameters.py
amoghrajesh: Fixed. The decorator entry now carries both fields, computed from the operator the decorator actually builds. Since the registered `class-name` points at the factory function rather than an operator, there was nothing to inspect; a new `_resolve_decorated_operator_class` reads the `decorated_operator_class=` the factory passes to `task_decorator_factory` and resolves it against the factory's own module. That convention holds for all 26 decorator modules in the tree.

Also gave the fixture a decorator entry, so `test_all_module_fields_present` now exercises that path, plus a test asserting the entry reports the built operator's real capability rather than just having the keys.

Fixed the durable tooltip too, since that line is in this diff: it now says the flag defaults to True on most but not all operators, instead of stating it as a flat default.

dev/registry/extract_parameters.py
amoghrajesh: Added one. The fixture builds a `functools.wraps` wrapper inside a separate module namespace, so its `__globals__` genuinely differ from the wrapped function's, which is what makes the flag invisible without unwrapping. Confirmed it binds: neutering the unwrap call turns exactly that test red and leaves the rest green. Also noted in the comment there that `BaseOperatorMeta` wraps execute on nearly every registered operator, so this is the common path rather than an edge case.

dev/registry/extract_parameters.py
amoghrajesh: Took all four. `_get_reachable_method_source` now strips comments, so that path matches the delegation path. `TaskDeferred\b` is now `raise TaskDeferred`, which keeps `VespaIngestOperator` and drops `except TaskDeferred` and `:raises TaskDeferred:`. Helpers resolve against origin instead of current, matching how self.<name>() actually dispatches. And visited keys now include the remaining depth, so a node first reached shallow is not skipped when re-reached with more budget.

Added a comment-only deferral fixture to TestSupportsDeferrable, and confirmed it binds: removing the comment stripping turns that test red.

dev/registry/tests/test_extract_parameters.py
amoghrajesh: Renamed the existing one to what it actually tests, which is the not a class in the MRO skip. Added a second one that binds the ordering behaviour: the decoy names a class that is in the MRO but owns no execute, with nothing after it that does, so the hop resolves to no owner and the walk has to continue to the real delegation below. Confirmed changing the loop to return on the first resolved match turns exactly that test red.

dev/registry/extract_parameters.py
amoghrajesh: Handled in 3f2104367f

registry/src/css/main.css
amoghrajesh: Handled in 3f2104367f

dev/registry/extract_parameters.py
amoghrajesh: Handled in 3f2104367f

dev/registry/extract_parameters.py
amoghrajesh: Handled in 3f2104367f

dev/registry/extract_parameters.py
amoghrajesh: Handled in 3f2104367f

dev/registry/tests/test_extract_parameters.py
amoghrajesh: Handled in 3f2104367f

kaxil (approved):
Round-6 items all verified at 4c6aacf6: the docstring note, the decorator capability fields (all 24 decorators resolve, `@task.agent` and `@task.kubernetes` now badge durable), and the three new tests each fail when the code they pin is reverted.

Two small things before merge, neither blocking. The description's aggregate counts (1779/6/217/5) and the "SparkSubmitOperator is the single durable-only case" line are out of date now that the numbers have settled (AgentOperator and `@task.agent` are durable-only too, and the delegation walk also flips SparkKubernetesOperator and SnowflakeNotebookOperator to durable, both correctly). Since this gets squash-merged under the body it is worth a refresh. Also, `test_module_contract_accepts_legacy_modules_without_supports_deferrable` asserts on the input dict because `_validate` returns its payload, so it passes even with the field deleted or its default flipped; `ModuleContract.model_validate(payload).supports_deferrable is False` is the binding form (the durable sibling has the same shape).

Optional follow-up material, not for this PR: the round-5 origin-resolution and depth-key fixes have no test that fails without them, and `_execute_chain_calls_resumable` still substring-matches unstripped source while the deferral path strips comments. The Durable badge tooltip ("Reconnects to an already-running job on retry") also predates this PR and does not describe AgentOperator or `@task.agent`, whose durable mode replays persisted step state rather than reconnecting to a job.

amoghrajesh:
@kaxil Both done.

Description refreshed against a fresh `breeze registry extract-data`: 1995 modules, 17 durable, 241 deferrable, 14 both. Dropped the "single durable-only case" line; durable-only is `SparkSubmitOperator`, `AgentOperator` and `@task.agent`. Also added `SnowflakeNotebookOperator` to the Snowflake bullet and the verification table, and
the two Databricks warehouse operators to the deferrable-only list, which the old "the rest are neither" sentence got wrong.

Both legacy-module tests now assert on the parsed model rather than the input dict: `ModuleContract.model_validate(payload).supports_deferrable is False`, and the same for the durable sibling. Flipping either default to `True` fails them.

Taking the rest as follow-up: tests for the round-5 origin-resolution and depth-key
fixes, comment stripping in `_execute_chain_calls_resumable`, and the Durable badge
tooltip wording for the replay case.

kaxil:
> @kaxil Both done.
> 
> Description refreshed against a fresh `breeze registry extract-data`: 1995 modules, 17 durable, 241 deferrable, 14 both. Dropped the "single durable-only case" line; durable-only is `SparkSubmitOperator`, `AgentOperator` and `@task.agent`. Also added `SnowflakeNotebookOperator` to the Snowflake bullet and the verification table, and the two Databricks warehouse operators to the deferrable-only list, which the old "the rest are neither" sentence got wrong.
> 
> Both legacy-module tests now assert on the parsed model rather than the input dict: `ModuleContract.model_validate(payload).supports_deferrable is False`, and the same for the durable sibling. Flipping either default to `True` fails them.
> 
> Taking the rest as follow-up: tests for the round-5 origin-resolution and depth-key fixes, comment stripping in `_execute_chain_calls_resumable`, and the Durable badge tooltip wording for the replay case.

Feel free to merge after CI 🎉

amoghrajesh:
The failing checks aren't related. Merging the PR.

PR #70299, opened 2026-07-23
Support cloud URI globs in DocumentLoaderOperator

guan404ming:
## Why

- `DocumentLoaderOperator` accepted glob patterns for local paths but rejected them for cloud storage URIs, so loading documents from `s3://bucket/logs/**/*.json` was impossible.
- Cloud sources were limited to a single object or one flat directory, forcing users to add an extra task just to list files.

## How

- Route cloud URIs containing wildcards through `ObjectStoragePath.glob()`, which supports recursive ** matching across directories.
- Reject wildcards in the scheme or bucket segment with a clear error, since those have no fixed root to glob from.
- Apply the same filtering rules as local globs, so behaviour matches between local and cloud sources.

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
@@ -68,9 +68,9 @@ class DocumentLoaderOperator(BaseOperator):
     :param source_path: A local path, glob pattern, or storage URI
         (``s3://``, ``gs://``, ``azure://``, ``file://``, ...). Cloud URIs
         go through :class:`~airflow.sdk.ObjectStoragePath` / fsspec.
-        ``**`` enables recursive matching for local globs. Cloud URIs
-        accept a single file or a directory; cross-directory globs in a
-        cloud URI are not supported in this version.
+        ``**`` enables recursive matching, for both local paths and cloud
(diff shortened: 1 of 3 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
```diff
@@ -219,18 +222,29 @@ def _resolve_remote_files(self, source_path: str) -> list[FilePathT]:
...
+            raise FileNotFoundError(f"Cloud URI '{source_path}' is neither a file nor a directory.")
 
         candidates = sorted(
             (p for p in root.iterdir() if not p.name.startswith(".")),
             key=str,
         )
         return self._filter_files([p for p in candidates if p.is_file()], is_directory_mode=True)
 
+    def _resolve_remote_glob(self, source_path: str) -> list[FilePathT]:
+        from airflow.sdk import ObjectStoragePath
```
gopidesupavan: can we use this from the compact provider?

gopidesupavan (approved):

providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
guan404ming: Sure, great catch thanks. Just updated

PR #70301, opened 2026-07-23
Add WasbRemoteLogIO.from_config and register wasb remote logging scheme

### LINKED ISSUE #70268: Add WasbRemoteLogIO.from_config and register wasb remote logging scheme

## Why

`from_config()` classmethod. This issue migrates the `wasb` scheme.

## What

- [ ] Add `WasbRemoteLogIO.from_config()` in
      `providers/microsoft/azure/src/airflow/providers/microsoft/azure/log/wasb_task_handler.py`,
      mirroring the legacy branch:
      https://github.com/apache/airflow/blob/104ad12e190db7197fa303d98a3ab68879eafd33/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L226-L248
      — including the `[logging] remote_task_handler_kwargs` IO-kwargs merge, `expanduser` on
      `base_log_folder`, reading `[azure_remote_logging] remote_wasb_log_container` (fallback
      `airflow-logs`) into `wasb_container`, and stripping the `wasb://` prefix from
      `remote_base_log_folder`, so behavior is unchanged for existing configs.
- [ ] Register the `wasb` scheme under a `remote-logging:` section in
      `providers/microsoft/azure/provider.yaml` and mirror it in
      `providers/microsoft/azure/src/airflow/providers/microsoft/azure/get_provider_info.py`.
      service or a local equivalent), run a task with remote logging enabled, and confirm logs are
      uploaded and read back through the new dispatch path. Include the setup and verification
      results in the PR description.

## Notes

The legacy branch matches any value *starting with* `wasb` — including non-URI values like
`wasb-logs`. Scheme dispatch only fires for proper `wasb://...` URIs (a value without `://` has no
URL scheme), so non-URI values keep resolving through the legacy fallback until the chain is
removed. The `wasb://` URI form should be documented as the supported form going forward; migrating
non-URI configs can be handled in the deprecation step tracked by the meta issue.

## Reference

The legacy branch derives a default connection id from `WasbHook.default_conn_name`; handle the
connection id the same way the merged amazon PRs do (the hook reads
`[logging] remote_log_conn_id` itself).

Andrushika:
Add WasbRemoteLogIO.from_config and register wasb remote logging scheme
---
## What
Core resolves remote log handlers by URL scheme through ProvidersManager dispatch, added in #67056.This PR adds `WasbRemoteLogIO.from_config()` and registers the `wasb` scheme, so Azure Blob logging is built by the provider instead of the hardcoded branch in `airflow_local_settings.py`. Existing `wasb://` configs resolve to an equivalent handler, and a `from_config` error falls back to the legacy path, so behaviour does not change.

Tested with a real task using `remote_base_log_folder = wasb://logs` against a local Azurite. The log uploaded to the `airflow-logs` container and read back. `ProvidersManager().remote_logging_handler_by_scheme("wasb")` returns `WasbRemoteLogIO`, so the resolver uses provider dispatch, not the legacy branch. Unit tests mirror `TestS3RemoteLogIOFromConfig`.

related: #70265
closes: #70268

Generated-by: Claude Opus 4.8 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/microsoft/azure/provider.yaml
+++ b/providers/microsoft/azure/provider.yaml
@@ -988,6 +988,10 @@ secrets-backends:
 logging:
   - airflow.providers.microsoft.azure.log.wasb_task_handler.WasbTaskHandler
 
+remote-logging:
+  - classpath: airflow.providers.microsoft.azure.log.wasb_task_handler.WasbRemoteLogIO
+    scheme: wasb
+
 extra-links:
   - airflow.providers.microsoft.azure.operators.data_factory.AzureDataFactoryPipelineRunLink
   - airflow.providers.microsoft.azure.operators.synapse.AzureSynapsePipelineRunLink
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/get_provider_info.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/get_provider_info.py
@@ -957,6 +957,12 @@ def get_provider_info():
         ],
         "secrets-backends": ["airflow.providers.microsoft.azure.secrets.key_vault.AzureKeyVaultBackend"],
         "logging": ["airflow.providers.microsoft.azure.log.wasb_task_handler.WasbTaskHandler"],
+        "remote-logging": [
+            {
+                "classpath": "airflow.providers.microsoft.azure.log.wasb_task_handler.WasbRemoteLogIO",
+                "scheme": "wasb",
+            }
+        ],
         "extra-links": [
             "airflow.providers.microsoft.azure.operators.data_factory.AzureDataFactoryPipelineRunLink",
             "airflow.providers.microsoft.azure.operators.synapse.AzureSynapsePipelineRunLink",
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/log/wasb_task_handler.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/log/wasb_task_handler.py
@@ -17,6 +17,7 @@
 # under the License.
 from __future__ import annotations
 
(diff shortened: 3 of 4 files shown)


potiuk (approved):
Faithful port. All four constructor arguments match the legacy wasb branch — `base_log_folder`, `remote_base` with the `wasb://` prefix stripped, `delete_local_copy`, `wasb_container` — and applying `| io_kwargs` is right here, since wasb is one of the object-storage backends where legacy merges. (I reviewed this alongside the sibling ports; the two search backends deliberately do not merge, and #70295 currently gets that wrong.)

The io/fth split test is the one that earns its keep, and I checked it isn't passing vacuously:

```python
("logging", "remote_task_handler_kwargs"): '{"delete_local_copy": true, "max_bytes": 1024}'
...
assert subject.delete_local_copy is True
assert not hasattr(subject, "max_bytes")
```

`max_bytes` is a real `FileTaskHandler.__init__` parameter, so this exercises both sides with genuine names — the IO kwarg reaches the class, the file-handler kwarg is filtered. If the split logic broke, `max_bytes` would be forwarded and construction would raise, so the test guards the mechanism and not just the outcome.

One inherited oddity, not something to fix here: `conf.get_mandatory_value(..., fallback="airflow-logs")` is contradictory — mandatory with a fallback. It comes across verbatim from `airflow_local_settings.py`, so the port is accurate; worth someone untangling in core at some point.

---

Andrushika:
Thanks for reviewing! The static check failure looks unrelated, could you please rerun it?

PR #70302, opened 2026-07-23
Support bzip2 and xz compressed inputs in LLM file analysis

guan404ming:
## Why

- LLMFileAnalysisOperator only accepted gzip, so archived logs and exports stored as `.log.xz` or `.csv.bz2` were rejected outright.
- The path suffixes were already recognised, but any codec other than gzip was refused, forcing users to decompress in a separate task.

## How

- Accept bzip2 and xz alongside gzip for .log, .json, and .csv inputs, using the standard library only.
- Replace the hardcoded gzip branch with a codec lookup, so all supported codecs share one read path.
- Keep snappy and zstd rejected, and keep compression unsupported for Parquet, Avro, image, and PDF inputs.

---


providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
```diff
@@ -18,11 +18,13 @@
 
 from __future__ import annotations
 
+import bz2
 import csv
 import gzip
 import io
 import json
 import logging
+import lzma
```
kaxil: `_bz2` and `_lzma` are optional CPython extensions, so on an interpreter built without them (the classic pyenv source build without liblzma-dev) this import now takes down the whole module, and with it `LLMFileAnalysisOperator` and `@task.llm_file_analysis`, even for the plain and gzip inputs that worked before this PR. Since `detect_file_format` already keys off `_DECOMPRESSORS` membership, registering bzip2/xz under `try`/`except ImportError` would degrade those builds to the existing "Compression ... is not supported" error instead (pandas guards these two imports the same way).

providers/common/ai/tests/unit/common/ai/utils/test_file_analysis.py
```diff
@@ -378,19 +382,29 @@ def test_detect_file_format_rejects_unsupported_compression(self, tmp_path):
...
-    @pytest.mark.parametrize("filename", ["sample.parquet.gz", "sample.avro.gz", "sample.png.gz"])
-    def test_detect_file_format_rejects_unsupported_gzip_format_combinations(self, tmp_path, filename):
+    @pytest.mark.parametrize(
+        "filename", ["sample.parquet.gz", "sample.avro.bz2", "sample.png.xz", "sample.pdf.gz"]
+    )
+    def test_detect_file_format_rejects_unsupported_compression_format_combinations(self, tmp_path, filename):
         path = tmp_path / filename
         path.write_bytes(b"content")
 
         with pytest.raises(LLMFileAnalysisUnsupportedFormatError, match="not supported for"):
```
kaxil: This `match` can't tell the two rejection branches apart: the codec-level message ("Compression 'bzip2' is not supported for file analysis.") contains the same substring, so the new `sample.avro.bz2` and `sample.png.xz` cases would still pass if bzip2/xz support were reverted. Something like `match=r"not supported for '\w+' file analysis"` pins them to the format-combination branch.

providers/common/ai/docs/operators/llm_file_analysis.rst
```diff
@@ -155,9 +155,10 @@ Supported Formats
 
 - Text-like: ``.log``, ``.json``, ``.csv``, ``.parquet``, ``.avro``
 - Multimodal: ``.png``, ``.jpg``, ``.jpeg``, ``.pdf`` when ``multi_modal=True``
-- Gzip-compressed text inputs are supported for ``.log.gz``, ``.json.gz``, and
-  ``.csv.gz``.
-- Gzip is not supported for ``.parquet``, ``.avro``, image, or PDF inputs.
+- ``gzip``, ``bzip2``, and ``xz`` compressed text inputs are supported for
+  ``.log``, ``.json``, and ``.csv`` (``.log.gz``, ``.csv.bz2``, ``.json.xz``, ...).
```
kaxil: The code also accepts compressed `.txt` here (`_COMPRESSION_SUPPORTED_FORMATS` includes `txt`, and `test_detect_file_format` covers `notes.txt.gz`), so `.txt` belongs in this list. The Text-like bullet above is missing `.txt` too since #70431 didn't touch this page, may as well fix both while you're here.

providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
guan404ming: Good catch, fixed. Imports now guarded; missing codecs fall back to the unsupported-compression error.

providers/common/ai/tests/unit/common/ai/utils/test_file_analysis.py
guan404ming: Fixed, tightened the match to pin the format-combination branch.

providers/common/ai/docs/operators/llm_file_analysis.rst
guan404ming: Fixed, added .txt to both lists.

providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
```diff
@@ -536,10 +552,10 @@ def _render_avro(path: ObjectStoragePath, *, sample_rows: int, max_content_bytes
 
 def _read_raw_bytes(path: ObjectStoragePath, *, compression: str | None, max_bytes: int) -> bytes:
     with path.open("rb") as handle:
-        if compression == "gzip":
-            with gzip.GzipFile(fileobj=handle) as gzip_handle:
-                return _read_limited_bytes(gzip_handle, path=path, max_bytes=max_bytes)
-        return _read_limited_bytes(handle, path=path, max_bytes=max_bytes)
+        if compression is None:
+            return _read_limited_bytes(handle, path=path, max_bytes=max_bytes)
+        with _DECOMPRESSORS[compression](handle) as decompressed:
```
kaxil, line 565: One asymmetry inside this dispatch: `gzip.open` raises `BadGzipFile` if anything follows a complete stream, but `bz2.open` and `lzma.open` silently return just the first stream, so `_read_limited_bytes` sees a normal EOF and nothing marks the content as incomplete. The spec-legal case is the interesting one: with 4 null bytes of xz Stream Padding between two streams, `xz -t` passes and `xz -dc` prints both, while `lzma.open` returns only the first. Pre-PR these inputs were rejected outright, so it's worth either treating non-null `unused_data` after `eof` as an error via `BZ2Decompressor`/`LZMADecompressor`, or pinning the current behavior in a test and noting the limitation in the docs.

providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
```diff
@@ -370,12 +386,12 @@ def detect_file_format(path: ObjectStoragePath) -> tuple[str, str | None]:
         raise LLMFileAnalysisUnsupportedFormatError(
             f"Unsupported file format {detected!r} for {path}. Supported formats: {', '.join(SUPPORTED_FILE_FORMATS)}."
         )
-    if compression and compression != "gzip":
+    if compression and compression not in _DECOMPRESSORS:
```
kaxil: On a build without `_lzma` this branch fires first and reports `Compression 'xz' is not supported for file analysis.`, the same message zstd gets, even though the docs now say xz is supported, so it points people at their file rather than at their Python build. It also makes the message for something like `sample.parquet.bz2` depend on the build, since the format-combination check below only runs when the codec module is present. Checking a static `{"gzip", "bzip2", "xz"}` set here and raising `AirflowOptionalProviderFeatureException` for a known-but-missing codec would match what `_render_parquet` and `_render_avro` already do lower down, and the rst could gain a clause saying bzip2/xz need an interpreter built with those modules.

providers/common/ai/tests/unit/common/ai/utils/test_file_analysis.py
```diff
@@ -378,19 +383,43 @@ def test_detect_file_format_rejects_unsupported_compression(self, tmp_path):
...
 
-    @pytest.mark.parametrize("filename", ["sample.parquet.gz", "sample.avro.gz", "sample.png.gz"])
-    def test_detect_file_format_rejects_unsupported_gzip_format_combinations(self, tmp_path, filename):
+    @pytest.mark.parametrize(("filename", "codec"), [("events.csv.bz2", "bzip2"), ("events.json.xz", "xz")])
+    def test_detect_file_format_rejects_compression_without_codec_module(self, tmp_path, filename, codec):
         path = tmp_path / filename
         path.write_bytes(b"content")
 
-        with pytest.raises(LLMFileAnalysisUnsupportedFormatError, match="not supported for"):
+        with patch.dict(_DECOMPRESSORS, {"gzip": gzip.open}, clear=True):
```
kaxil, line 425: This pins the rejection but not the half I was actually worried about, that plain and gzip inputs keep working while the codecs are missing, which is one more assert inside the same patched block. Related, `import bz2` and `import lzma` at the top of this module are unguarded, so on exactly those builds the module fails to collect, and `sample.avro.bz2` / `sample.png.xz` in the combination test would hit the codec-level message and fail its regex rather than skip. `pytest.importorskip` (already used for pyarrow and fastavro below) plus `.gz` params for the combination test covers it, and the expansion-bound tests at 262 and 279 are still gzip-only if you want to reuse the parametrize you added at 410.

kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
guan404ming: Pinned current behavior in tests and documented the limitation; decompressor-level detection deferred.

providers/common/ai/src/airflow/providers/common/ai/utils/file_analysis.py
guan404ming: Fixed: known-codec set, AirflowOptionalProviderFeatureException naming the missing module, docs note added.

providers/common/ai/tests/unit/common/ai/utils/test_file_analysis.py
guan404ming: Added plain/gzip asserts and importorskip, parametrized per-file expansion; total-limit test stays gzip-only.

guan404ming:
Merging, feel free to let me know if anything need refine.

PR #70307, opened 2026-07-23
Fill scheduler task batches after concurrency rejection

avolant:
## Why

We observed many runnable task instances remaining in the `SCHEDULED` state for a long time. Higher-priority task instances repeatedly appeared ahead of them in the candidate query, but could not be queued because of Dag-internal concurrency constraints. Those blocked candidates consumed the query limit and delayed lower-priority runnable tasks from being considered.

The scheduler candidate query applies `LIMIT max_tis`, then performs several concurrency and capacity checks in memory. If a page contains both a blocked high-priority task and a runnable task, the previous loop stopped as soon as it selected any executable task, leaving the batch partially filled. Repeating that pattern can starve lower-priority runnable tasks.

For example, with `max_tis=2` and candidates ordered as:

```text
A: blocked by task concurrency
B: runnable
C: runnable
```

The first query returns `[A, B]`. The scheduler rejects `A`, accepts `B`, and previously stopped with only one of two available positions filled. `C` is never examined during that scheduling pass.

## What changed

The scheduler now continues querying while the batch remains under `max_tis` and the previous iteration discovered a new starvation filter.

After the first query above:

1. The task-concurrency filter learned from `A` excludes it from the next query.
2. The selected TI ID for `B` is explicitly excluded because selected TIs remain `SCHEDULED` in the database until the loop finishes.
3. The next query uses `LIMIT 1`, which is the remaining batch capacity, and selects `C`.
4. The completed batch is `[B, C]`.

This intentionally does not use SQL `OFFSET`. Every iteration adds newly learned starvation filters, which changes and shrinks the ordered result set. Applying an offset to that changed result set could skip valid candidates. Re-querying with the learned filters and selected-TI exclusion provides stable iteration over this dynamic candidate set.

Executor slot availability is also initialized outside the query loop. Accepted tasks can now span multiple query iterations, so executor capacity must be consumed across the entire loop. Reinitializing it per page would allow later pages to reuse capacity already consumed by earlier pages.

## Validation

- Added a database-backed regression test without monkeypatching that fails before the fix and fills the batch after the fix.
- Full SQLite scheduler module: 389 passed, 3 expected skips.
- PostgreSQL regression and executor-capacity coverage: passed.
- Ruff, mypy, pre-commit, and manual hooks: passed.

---

- [X] Yes -- Codex (GPT-5)

---
Drafted-by: Codex (GPT-5); reviewed by @avolant before posting


avolant:
Here is my _humain_'s formulation of the problem:

In our production environment, we had a problem where some tasks stayed in `SCHEDULED` state for longer than expected. we tracked down the root cause of the problem to [this line](https://github.com/apache/airflow/blob/d81660a8d705a77b71d13000ffc77680e438c6e3/airflow-core/src/airflow/jobs/scheduler_job_runner.py#L974);
```py
is_done = executable_tis or len(task_instances_to_examine) < max_tis
```

To our understanding, `is_done` condition was met (meaning, that `is_done` was set to `True`), but only one `executable_tis` was transitionned from `SCHEDULED` to `QUEUED`:
We had 32 as `max_tis_per_query`, but 31 of the _studied_ tasks couldn't run (meaning, couldn't transition from `SCHEDULED` to `QUEUED`), because they had reached the internal DAG `max_tis_per_task`. The trouble is that each scheduler iterations were taking the same 31 tasks, and scheduling only task per loop which was highly inefficient, because the above `is_done` condition was always met. 

This PR attempt to solve this by making sure the scheduler loop continues to check the next batch of _schedulable_  tasks:
```py
if len(executable_tis) >= max_tis or not found_new_filters:
```

avolant:
Commit `eac3bf9` keeps refill queries at `max_tis` instead of shrinking them to the remaining batch capacity. This avoids one database query per blocked task when the batch is nearly full, while stopping selection as soon as the batch reaches `max_tis`. The regression test verifies the batch is filled in two queries instead of four.

avolant:
Tested it live in our Airflow deployed instances and got the following logs:
```
  Selected 1 of 32 task instances after iteration 1; retrying after excluding known blocked candidates.
  Selected 29 of 32 task instances after iteration 1; retrying after excluding known blocked candidates.
  Selected 31 of 32 task instances after iteration 1; retrying after excluding known blocked candidates. 
  ```

Stating that we [reached this line](https://github.com/apache/airflow/pull/70307/changes#diff-072241d81275cd4b4b867f51025e9dca800610069305cffeb4b77ad45e135557R992)

avolant:
This has been in our production datacenter for the last 2 weeks, and we haven't experienced any regression so far. Still looking for a review

avolant:
Hello @ashb / @XD-DENG would you have some time to review my PR? :)

avolant:
Hi @potiuk / @hussein-awala 👋 
Can I have a review on this PR please?

PR #69610, opened 2026-07-08
Fix TimeSensor start_from_trigger causing dag_version churn

bramhanandlingala:
closes: #69543

`TimeSensor` computed its target time using `datetime.now()` at DAG-parse
time. When `start_from_trigger=True`, this value got baked into the
serialized DAG, causing the `dag_version` to change on every parse that
crossed a day boundary.

This can't be fixed by computing the value differently, since the target
moment is inherently different each day but `start_trigger_args` is only
computed once per DAG version. So `start_from_trigger=True` now raises a
`ValueError` instead of silently producing a broken serialization.
`deferrable=True` still works as before and is unaffected.

Also fixed the docs, which incorrectly said the target time is evaluated
against `data_interval_end`/`run_after` — it never was.

Added a test for the new `ValueError`; all existing tests still pass.

**Gen-AI disclosure:** I used a generative AI tool to help identify the root
cause, write tests, and draft the PR description. I reviewed, tested, and
verified all changes locally before submitting.

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude

Generated-by: Claude following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #55068, opened 2025-08-29
Re-enable start_from_trigger feature with rendering of template fields

dabla:
This PR tries to fix the rendering of templated fields with start from trigger args without breaking architectural assumptions, which the initial [PR 53071](https://github.com/apache/airflow/pull/53071) did violate and was reverted in [PR 55037](https://github.com/apache/airflow/pull/55037).

This PR could be simplified without the need of parsing the DAG if the render_template_fields would be re-usable in the triggerers, as of now it's part of the BaseOperator and thus not re-usable in BaseTrigger. As discussed with @ashb , apparently @amoghrajesh would be working on this (e.g. making template renderers reusable).

As opposed to previous attempt, here the rendering of the templates isn't done while creating the trigger into the database within the TriggerRunnerSupervisor. Here, the TriggerRunnerSupervisor will only retrieve the serialized DAG from the DagModel, and pass it to the workloads RunTrigger as an extra parameter (e.g named dag_data), which in turn will be used by the TriggerRunner in the subprocess in which there it will have access to the XCom's, and thus, shouldn't raise the: ImportError: cannot import name 'SUPERVISOR_COMMS' from 'airflow.sdk.execution_time.task_runner'.

~~So, if template rendering would be reusable outside the BaseOperator, then we wouldn't need to parse the DAG from the DagBag to retrieve the task (e.g. BaseOperator) to be able to do the template rendering, which would make the solution even easier but also less heavy (e.g. more performant) for the triggerer subprocess.~~

After further reflection, also regarding the work I'm doing for [AIP-88](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=334760511), the current solution will still need to get the task (e.g. Operator) for the trigger when yielding multiple events.  Same for the rendering of templates, you need the context to be able to render those, and to get the context, you need at least a RuntimeTaskInstance, which of course, requires guess what?  a task (e.g. BaseOperator).  So even if the rendering of templates would be reusable across operator and triggerer, you will still need the context, thus, in this case, wouldn't change that much to the current solution.

As loading DAG code is prohibited in the triggerer, we load the serialized DAG version from the database.  As a consequence, callables aren't supported as kwargs when start_from_trigger is enabled.  Thus, when enabling an operator as start_from_trigger, we will automatically validate the kwargs and check if those don't contain callables, otherwise an AirflowException will be raised stating which kwargs has a callable.

image

Also the [test dag](https://github.com/apache/airflow/pull/55037) from @kaxil is working:

image
image

[...]

PR #70310, opened 2026-07-23
Fix DateTimeSensorAsync crashing Dag parsing with templated target_time

### LINKED ISSUE #70284: DateTimeSensorAsync crashes Dag parsing with templated target_time + start_from_trigger=True
### Apache Airflow version

### What happened

file, not just the task.

### Reproduction

```python
import pendulum
from airflow.sdk import DAG

    DateTimeSensorAsync(
        task_id="t",
        start_from_trigger=True,
    )
```

Raises at Dag-parse time:

```
```

### How to reproduce

### Operating System

### Deployment

Other

### Deployment details

### Anything else

Related: #69610, #69543

### Code of Conduct

haseebmalik18:
Fixes `DateTimeSensorAsync` crashing Dag parsing when `target_time` is a Jinja template and `start_from_trigger=True`.

With `start_from_trigger=True`, `target_time` gets parsed in `__init__` at Dag-parse time, before Jinja renders it. So passing a template (which is the field's documented use) blew up with a `pendulum` `ParserError` and took down the whole Dag file. Now a static `target_time` still works, and a templated one raises a clear `ValueError` instead.
  
closes: #70284

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -98,6 +98,9 @@ class DateTimeSensorAsync(DateTimeSensor):
 
     :param target_time: datetime after which the job succeeds. (templated)
     :param start_from_trigger: Start the task directly from the triggerer without going into the worker.
+        This requires a static ``target_time`` (a datetime or ISO-8601 string). A templated
+        ``target_time`` is not supported here because the trigger is created at Dag-parse time,
+        before Jinja templates are rendered.
     :param trigger_kwargs: The keyword arguments passed to the trigger when start_from_trigger is set to True
         during dynamic task mapping. This argument is not used in standard usage.
     :param end_from_trigger: End the task directly from the triggerer without going into the worker.
@@ -125,8 +128,14 @@ class DateTimeSensorAsync(DateTimeSensor):
 
         self.start_from_trigger = start_from_trigger
         if self.start_from_trigger:
+            try:
+                moment = timezone.parse(self.target_time)
+            except ValueError as e:
+                raise ValueError(
+                    f"start_from_trigger=True requires a static target_time, not a template: {self.target_time!r}"
+                ) from e
             self.start_trigger_args.trigger_kwargs = dict(
-                moment=timezone.parse(self.target_time),
+                moment=moment,
                 end_from_trigger=self.end_from_trigger,
             )
 
--- a/providers/standard/tests/unit/standard/sensors/test_date_time.py
+++ b/providers/standard/tests/unit/standard/sensors/test_date_time.py
@@ -24,7 +24,7 @@ import pytest
 
 from airflow import macros
 from airflow.models.dag import DAG
-from airflow.providers.standard.sensors.date_time import DateTimeSensor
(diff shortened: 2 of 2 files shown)


providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -98,6 +98,9 @@ class DateTimeSensorAsync(DateTimeSensor):
 
     :param target_time: datetime after which the job succeeds. (templated)
     :param start_from_trigger: Start the task directly from the triggerer without going into the worker.
+        This requires a static ``target_time`` (a datetime or ISO-8601 string). A templated
+        ``target_time`` is not supported here because the trigger is created at Dag-parse time,
```
shahar1, line 102: The trigger is not created at Dag-parse time, but either in the worker or the triggerer

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -125,8 +128,14 @@ def __init__(
...
         if self.start_from_trigger:
+            try:
+                moment = timezone.parse(self.target_time)
+            except ValueError as e:
+                raise ValueError(
+                    f"start_from_trigger=True requires a static target_time, not a template: {self.target_time!r}"
+                ) from e
             self.start_trigger_args.trigger_kwargs = dict(
-                moment=timezone.parse(self.target_time),
+                moment=moment,
```
shahar1, line 138: How does it solve the crashing? It seems to improve the error message rather than fix the issue.

shahar1 (changes requested):

providers/standard/src/airflow/providers/standard/sensors/date_time.py
jroachgolf84: @shahar1 - I think the goal was to replicate something like this: https://github.com/apache/airflow/pull/69610/changes

providers/standard/src/airflow/providers/standard/sensors/date_time.py
haseebmalik18: Yes you're right, I overlooked that. Will reword the docstring.

providers/standard/src/airflow/providers/standard/sensors/date_time.py
haseebmalik18: Fair point. As is, this just turns a cryptic ParserError into a clearer ValueError, the Dag file still fails to parse. And yes, the intent was to mirror #69610 

The difference is that for TimeSensor, start_from_trigger could never work, since it computes "today at target_time" from the wall clock at parse time. Here target_time is a template field, so there are I think two ways to go about it:

1. If target_time is not parseable at parse time, log a warning and fall back to deferring from the worker. The template renders as usual and the Dag parses fine on all supported versions.
2. On Airflow 3.3+, pass the raw target_time string through trigger_kwargs and let the triggerer render it before the trigger runs. Since #55068  the triggerer renders any trigger kwarg whose name matches an operator template field, which FileSensor already relies on for filepath. That would make start_from_trigger actually work with a templated target_time, but it means extending DateTimeTrigger to accept a target_time kwarg and gating on the Airflow version, falling back to (1) on older versions.

PR #70312, opened 2026-07-23
UI: Show owning team on Dag, Dag Run, and Task Instance detail pages

vincbeck:
Multi-team deployments need to see which team owns a Dag directly from the detail pages, not just from the list views. This surfaces the owning team on the Dag, Dag Run, and Task Instance detail pages as a linked row that filters the Dag list by that team, mirroring how the Owner field already behaves. It also adds it to the page detail header as a new column. **In the case of the Dag details page, it replaces the column `owner`.**

The team row/column is only shown when multi-team support is enabled and a team is resolved for the object, so single-team deployments see no change.

### Dag details page
Screenshot 2026-07-23 at 1 47 47 PM

### Dag run details page
Screenshot 2026-07-23 at 1 47 59 PM

### Task instance details page
Screenshot 2026-07-23 at 1 48 11 PM

 

---

Claude (Opus 4.8)

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
@@ -141,6 +141,7 @@ def get_dag_run(dag_id: str, dag_run_id: str, session: SessionDep) -> DAGRunResp
             status.HTTP_404_NOT_FOUND,
             f"The DagRun with dag_id: `{dag_id}` and run_id: `{dag_run_id}` was not found",
         )
+    attach_team_names([dag_run], session=session)
     return dag_run
 
 
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
@@ -42,7 +42,7 @@ from airflow.api_fastapi.common.db.dag_runs import (
     attach_dag_versions_to_runs,
     eager_load_dag_run_for_list,
 )
-from airflow.api_fastapi.common.db.dags import attach_team_names
+from airflow.api_fastapi.common.db.dags import eager_load_teams
 from airflow.api_fastapi.common.parameters import (
     FilterOptionEnum,
     FilterParam,
@@ -134,14 +134,15 @@ dag_run_at_dag_router = AirflowRouter(tags=["DagRun"], prefix="/dags/{dag_id}")
 )
 def get_dag_run(dag_id: str, dag_run_id: str, session: SessionDep) -> DAGRunResponse:
     dag_run = session.scalar(
-        select(DagRun).filter_by(dag_id=dag_id, run_id=dag_run_id).options(joinedload(DagRun.dag_model))
+        select(DagRun)
+        .filter_by(dag_id=dag_id, run_id=dag_run_id)
+        .options(joinedload(DagRun.dag_model), *eager_load_teams(DagRun.dag_model))
     )
     if dag_run is None:
         raise HTTPException(
             status.HTTP_404_NOT_FOUND,
             f"The DagRun with dag_id: `{dag_id}` and run_id: `{dag_run_id}` was not found",
         )
-    attach_team_names([dag_run], session=session)
     return dag_run
 
 
@@ -694,7 +695,6 @@ def get_dag_runs(
             has_next = has_more
 
         attach_dag_versions_to_runs(dag_runs, session=session)
-        attach_team_names(dag_runs, session=session)
 
         return DAGRunCollectionResponse(
             dag_runs=dag_runs,
@@ -714,7 +714,6 @@ def get_dag_runs(
     )
     dag_runs = list(session.scalars(dag_run_select))
     attach_dag_versions_to_runs(dag_runs, session=session)
-    attach_team_names(dag_runs, session=session)
 
     return DAGRunCollectionResponse(
         dag_runs=dag_runs,
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
@@ -390,6 +390,10 @@ class DagRun(StrictBaseModel):
             else:
                 values["note"] = None
 
+        # A property rather than a column, so the loop above never picks it up.
+        if not insp.detached:
+            values["team_name"] = data.team_name
+
         return values
 
 
--- a/airflow-core/src/airflow/models/dag.py
+++ b/airflow-core/src/airflow/models/dag.py
@@ -38,6 +38,7 @@ from sqlalchemy import (
     Text,
     case,
     func,
+    inspect as sa_inspect,
     or_,
     select,
 )
@@ -63,7 +64,7 @@ from airflow.models.asset import AssetDagRunQueue
 from airflow.models.base import Base, StringID
 from airflow.models.dagbundle import DagBundleModel
 from airflow.models.dagrun import DagRun
-from airflow.models.team import Team, TeamOwnedMixin
+from airflow.models.team import Team
 from airflow.serialization.definitions.assets import SerializedAssetUniqueKey
 from airflow.serialization.encoders import DAT, encode_deadline_alert
 from airflow.serialization.enums import Encoding
@@ -79,6 +80,7 @@ if TYPE_CHECKING:
     from typing import TypeAlias
 
     from dateutil.relativedelta import relativedelta
+    from sqlalchemy.orm.state import InstanceState
 
     from airflow.sdk import Context
     from airflow.serialization.definitions.assets import (
@@ -307,7 +309,7 @@ class DagOwnerAttributes(Base):
         return dag_links
 
 
-class DagModel(Base, TeamOwnedMixin):
+class DagModel(Base):
     """Table containing DAG properties."""
 
     __tablename__ = "dag"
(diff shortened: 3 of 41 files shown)


airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
```diff
@@ -141,6 +141,7 @@ def get_dag_run(dag_id: str, dag_run_id: str, session: SessionDep) -> DAGRunResp
             status.HTTP_404_NOT_FOUND,
             f"The DagRun with dag_id: `{dag_id}` and run_id: `{dag_run_id}` was not found",
         )
+    attach_team_names([dag_run], session=session)
```
pierrejeambrun: That pattern seems like a maintainability burden and that goes against how we do things for other related attribute.

Everywhere you need `team_name`, you need to manually call this `attach_team_names` function beforehand, or you will get wrong `None` value for team.

Instead can we move this at the relationship level on models or via a proxy, so `entity.team_name` always return the correct value. I think there's a cached per dag resolver  or something `DagModel.get_team_name`. 

Also if eager loading the realtionship that could remove the extra query all together.

pierrejeambrun (commented):
Thanks, one concern that I think is worth solving.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
vincbeck: The tradeoff with this is single-team deployments now pay for the bundle+teams eager-load on this endpoint. Are we okay with that?

vincbeck: @pierrejeambrun

pierrejeambrun: I think the eager loading on team doesn't cost anything cause there won't be any team to join on. 


For the eager loading of bundles, we can make this "raise" when not loaded just to be sure on the relationship, and only join the query (preload) if the condition is met from a config standpoint. Today this preloading is done in views or service most of the time. And condition this on the serialized if we need to. 

But that should cost too much I think

seanghaeli:
Is there a reason we can't show both Owner and Team simultaneously?

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
```diff
@@ -390,6 +390,10 @@ def safe_extract_from_orm(cls, data: Any) -> Any:
             else:
                 values["note"] = None
 
+        # A property rather than a column, so the loop above never picks it up.
+        if not insp.detached:
+            values["team_name"] = data.team_name
```
seanghaeli, line 395: The `team_name` here can be stale for up to 30 seconds:

The worker will now get `team_name` from here instead of from `ti_run()`. `data.team_name` falls back to this cache, which only updates every 30 seconds:

https://github.com/apache/airflow/blob/682cf06525cd3f7006b8c1b46c42c7102d84373c/airflow-core/src/airflow/models/dag.py#L108-L109

I verified this happens locally on breeze, and is fixed here:

https://github.com/apache/airflow/blob/7c3e41b63083ce9648235e0a5107b8ff236c7af9/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py#L261-L266

by adding the following line:

```.options(*eager_load_teams(DR.dag_model))```

**Comment by vincbeck** [comment]
> Is there a reason we can't show both Owner and Team simultaneously?

Space is limited but I guess we could. Though, I am wondering what the value of `owner` is at the first place

File: airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py (hunk 2 above)
Comment (vincbeck) [reply airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py:395]: I am trying to understand what the issue is then :) Are you saying this is currently good or we need a fix?

File: airflow-core/src/airflow/models/dag.py [hunk 3]
```diff
@@ -482,6 +482,15 @@ class DagModel(Base):
...
     )
+    # The Dag's bundle, which carries the owning team. ``lazy="raise"`` forbids implicit
+    # loading so a caller that forgets ``eager_load_teams`` cannot emit a silent N+1;
+    # ``TeamOwnedMixin.team_name`` only reads this once loader options have populated it.
+    bundle = relationship(
+        "DagBundleModel",
+        primaryjoin="DagModel.bundle_name == DagBundleModel.name",
+        viewonly=True,
+        lazy="raise",
+    )
```
pierrejeambrun: Why do we need this ?

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
pierrejeambrun: Yep my agend flagged this too. There's no eager loading of team there, so the team returned can be stale.

airflow-core/src/airflow/ui/src/pages/Dag/Header.tsx
```diff
@@ -112,10 +115,19 @@ export const Header = ({
...
+          label: translate("dagDetails.team"),
+          value: (
+            <RouterLink to={`/dags?teams=${encodeURIComponent(dag.team_name as string)}`}>
+              {dag.team_name}
+            </RouterLink>
+          ),
+        }
+      : {
+          label: translate("dagDetails.owner"),
+          value: <DagOwners ownerLinks={dag?.owner_links ?? undefined} owners={dag?.owners} />,
```
pierrejeambrun: We shouldn't mix those two. (owner and teams are different).

airflow-core/src/airflow/models/team.py
```diff
@@ -78,3 +79,48 @@ def get_all_team_names(cls, *, session: Session = NEW_SESSION) -> set[str]:
...
+                # Reuse this entity's own session. ``get_team_name`` is ``@provide_session``,
+                # and the session it would open is the *same* scoped session the caller is
+                # using, so closing it on exit detaches every object the caller still holds.
+                if state.session is not None:
+                    return DagModel.get_team_name(self.dag_id, session=state.session)
+                return DagModel.get_team_name(self.dag_id)
+            if (entity := getattr(entity, attribute)) is None:
+                return None
+        # A bundle maps to at most one team (unique index on dag_bundle_team.dag_bundle_name).
+        return entity[0].name if entity else None
```
pierrejeambrun: No idea why we used an association table for `dag_bundle_team_association_table` while we have a unique index directly on it. That would just come down to a normal N-1 relathionship, and would avoid this `[0]`.

Anyway out of this PR scope.

pierrejeambrun (commented):
Overall looking good to me

pierrejeambrun:
Actually something like:
the mixin's two ugly parts (the sa_inspect().unloaded walk and the generic string _team_path) both disappear if you (1) localize the teams[0]→scalar on the bundle and (2) let association_proxy do
  the chaining. Here's the shape I'd go for, no migration needed:
```
  # DagBundleModel — collapse the M-N collection to a scalar in ONE place
  @property
  def team_name(self) -> str | None:
      # unique index on dag_bundle_team.dag_bundle_name → at most one team
      return self.teams[0].name if self.teams else None

  # DagModel — the only place with real logic: the multi_team gate
  @property
  def team_name(self) -> str | None:
      if not conf.getboolean("core", "multi_team"):
          return None
      return self.bundle.team_name if self.bundle else None   # bundle lazy="raise" IS the guard

  # DagRun / TaskInstance — just chain, no logic
  team_name = association_proxy("dag_model", "team_name")
  team_name = association_proxy("dag_run", "team_name")
```

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/taskinstance.py
seanghaeli: @vincbeck We should fix it, which would simply be adding the following line:

`.options(*eager_load_teams(DR.dag_model))`

in line 265 of [airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py](https://github.com/apache/airflow/blob/7c3e41b63083ce9648235e0a5107b8ff236c7af9/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py#L261-L266)

airflow-core/src/airflow/models/team.py
vincbeck: We did that on purpose so that later if we want to have N-N relationship between dag bundles and teams, it is easier

airflow-core/src/airflow/ui/src/pages/Dag/Header.tsx
vincbeck: Alright, now I keep the owner regardless of multi-team

airflow-core/src/airflow/models/dag.py
vincbeck: `primaryjoin="DagModel.bundle_name == DagBundleModel.name",` was not needed but the relationship is needed

vincbeck:
PR is updated. Thanks for your feedbacks @seanghaeli and @pierrejeambrun . Please let me know if the new version is better

seanghaeli (approved):

airflow-core/src/airflow/ui/src/pages/Dag/Header.tsx
```diff
@@ -58,6 +59,8 @@ export const Header = ({
   const { t: translate } = useTranslation(["common", "dag"]);
   // We would still like to show the dagId even if the dag object hasn't loaded yet
   const { dagId } = useParams();
+  const multiTeamEnabled = Boolean(useConfig("multi_team"));
+  const showTeam = multiTeamEnabled && dag?.team_name !== undefined && dag.team_name !== null;
```
bbovenzi, line 63: We copy-pasted this showTeam logic in a lot of places. Let's just have a <TeamName />` component to handle this.

vincbeck: Good call, done

pierrejeambrun (approved):
LGTM, thanks

Issue #70296, opened 2026-07-23
Move template-field validation/transformation out of operator __init__ (exemption-list burn-down)


shahar1:
### Background

Template fields are rendered **after** an operator's constructor runs. Any logic applied to a template-field parameter's **value** inside `__init__` — validation, type checks, transformation, string interpolation — therefore operates on the un-rendered Jinja expression, not the real value.

This is documented in [Creating a custom Operator](https://airflow.apache.org/docs/apache-airflow/stable/howto/custom-operator.html#templating) and [contributing-docs/05_pull_requests.rst](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst), and it causes real bugs: #69813 (`SSHRemoteJobOperator` validated a templated `remote_base_dir` in `__init__`, so cleanup validation failed for any custom base) is a recent example.

> [!IMPORTANT]
> Checks that only ask **whether an argument was passed** are the exception and belong in `__init__`. Read [False positives](#false-positives) below before fixing an entry.

The `validate-operators-init` prek hook previously only verified that template fields are *assigned* verbatim at the top level of `__init__`; it could not see validation calls, conditionals, or transformations (and it did not cover sensors, `AwsBaseOperator[...]` subclasses, or `aws_template_fields(...)`-based classes at all). The hook has been extended to detect any non-sanctioned read of a template field in `__init__` (PR to follow).

### The exemption ratchet

Existing violations are listed in [`scripts/ci/prek/validate_operators_init_exemptions.txt`](https://github.com/apache/airflow/blob/main/scripts/ci/prek/validate_operators_init_exemptions.txt) as `path::ClassName` entries so the hook can enforce the rule on new code immediately.

- A PR that fixes a class **must remove its entry** in the same PR — the hook fails on stale exemptions, so this cannot be forgotten.
- New violations cannot be added: only files listed in the exemption file are suppressed.
- This issue is done when the exemption file is empty.

**How to fix a class:** move the validation / transformation from `__init__` into `execute()` (or the first method that runs after rendering), keep `__init__` down to plain `self.field = field` assignments (defaulting via `field or default` is fine). Don't forget the corresponding tests.

Please do **not** open sub-issues for individual entries — comment here or just open a PR referencing this issue.

### False positives

Not everything the hook flags should move. A check that only asks **whether an argument was passed** — the usual "exactly one of `a` or `b`" guard — belongs in `__init__` and must not be moved:

[...]

PR #70315, opened 2026-07-23
Fix template-field validation timing in ssh provider operators

bramhanandlingala:
Fixes the `ssh` provider's two entries from the #70296 exemption-list burn-down: `SSHOperator` and `SSHRemoteJobOperator` applied logic to template fields inside `__init__`, before Jinja templating runs.

- `SSHRemoteJobOperator`: `remote_base_dir` was validated in `__init__`, against the raw un-rendered template string. Moved validation to `execute()`.
- `SSHOperator`: when a pre-built `ssh_hook` was injected, `__init__` pushed the un-rendered `remote_host` onto it. Moved this to `execute()`.

Both `__init__` methods now only do plain assignments; the field-dependent logic runs in `execute()`, after templating.

Tests added for both operators covering the fixed behavior.

Related to #70296

- [X] Yes — Claude


shahar1 (commented):
Thanks for your contribution! 
1. Please fix static checks.
2. Make sure to remove the operator from the txt file.

bramhanandlingala:
@shahar1 

Thanks for the review!

Fixed the static check and removed both ssh entries from the exemption file. Let me know if anything else is needed.

shahar1 (approved):

providers/ssh/src/airflow/providers/ssh/operators/ssh.py
```diff
@@ -168,6 +166,9 @@ def execute(self, context=None) -> bytes | str:
         if self.command is None:
             raise AirflowException("SSH operator error: SSH command not specified. Aborting.")
 
+        if self.remote_host is not None:
```
shahar1: In `__init__` this assignment was nested inside `if ssh_hook and isinstance(ssh_hook, SSHHook):`, so it only ran when a hook was injected.
In `execute()` it runs unconditionally, which now includes the ssh_conn_id path - and an empty string passes `is not None`:

```python
if self.remote_host is not None:
    self.ssh_hook.remote_host = self.remote_host   # ← "" reaches here now
```

The ssh_hook cached property builds `SSHHook(remote_host=self.remote_host or "")`, and `SSHHook.__init__` falls back to `conn.host` when that is empty. This line then puts the empty value back, and `get_conn()` raises Missing required param: `remote_host`. Because remote_host is templated, `{{ dag_run.conf.get('host', '') }}` rendering to `""` is the realistic way to hit it - that worked before this PR.

```python
        if self.remote_host:
            self.ssh_hook.remote_host = self.remote_host
```

That also keeps the truthiness semantics consistent with the or "" already used in the cached property.

---

Drafted by AI-Assistant tool (Apache Magpie + Claude Code), confirmed and and reviewed by me

shahar1 (dismissed):

shahar1:
@bramhanandlingala noted an issue just before merging, please address this. Thank you! (please note that I merged `main` to your branch, so you might need to pull it first)

bramhanandlingala:
Thanks @shahar1! Fixed — switched to if self.remote_host: as suggested, and added tests covering both the __init__ no-mutation case and the execute() templated-value case. Ready for re-review.

shahar1:
> Thanks @shahar1! Fixed — switched to if self.remote_host: as suggested, and added tests covering both the **init** no-mutation case and the execute() templated-value case. Ready for re-review.

Please note that static checks currently fail

bramhanandlingala:
Thanks for flagging, @shahar1 ! Dug into it — the failure wasn't related to the SSH changes at all. It was a stale exemption entry: SnowparkContainerJobOperator had already been fixed and its exemption removed on main, but a stale copy of that entry survived in our branch's validate_operators_init_exemptions.txt through a few merge conflict resolutions. Removed it and pushed — all checks are passing now. Ready for re-review whenever you have a chance!

shahar1 (approved):

PR #70320, opened 2026-07-23
Fix DateTimeSensor crash when target_time renders to a datetime

1fanwang:
`target_time` is a template field, so it is only rendered after `__init__` runs. `DateTimeSensor` validated and normalized it in the constructor — acting on the un-rendered Jinja expression — and `poke()` parsed `self.target_time` directly with `timezone.parse()`. When `target_time` is templated and the Dag uses `render_template_as_native_obj=True`, the field renders to a `datetime`, and `poke()` crashed:

```
TypeError: argument 'input': 'DateTime' object cannot be cast as 'str'
```

The deferrable path already read the value through `_moment`, which handles both `str` and `datetime`, so only the synchronous sensor was affected. The constructor now stores `target_time` verbatim, and the datetime→ISO-string normalization it used to do is relocated into _moment rather than dropped; self.target_time itself is never mutated. `_moment` still parses and validates the rendered value, and both sensors route through it. `__init__` only assigns the template field now, in line with the burn-down in #70296.

related: #70296

Operator-level repro (real Dag, real render, real `poke`) — before the fix:

```
target_time after render: DateTime(2020, 1, 1, 0, 0, 0, tzinfo=Timezone('UTC')) DateTime
  ...
  return timezone.utcnow() > timezone.parse(self.target_time)
TypeError: argument 'input': 'DateTime' object cannot be cast as 'str'
```

After the fix:

```
target_time after render: DateTime(2020, 1, 1, 0, 0, 0, tzinfo=Timezone('UTC')) DateTime
poke -> True
```

`test_date_time.py`: 13 passed. The two normalization guards — `_moment` resolves a native-rendered and a constructor-passed datetime to an aware datetime — fail on the pre-fix source and pass after.

</details>

---

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -63,18 +63,20 @@ class DateTimeSensor(BaseSensorOperator):
 
     def __init__(self, *, target_time: str | datetime.datetime, **kwargs) -> None:
         super().__init__(**kwargs)
-        # target_time is a template field, so it is rendered after __init__ runs. Store it
-        # verbatim and defer validation/normalization to _moment, which sees the rendered value.
+        # target_time is a template field; store it verbatim and normalize once rendered.
         self.target_time = target_time
 
     def poke(self, context: Context) -> bool:
+        # Normalize a rendered datetime to an ISO string here (moved from __init__, which only
+        # saw the un-rendered Jinja) so the stored template field stays a string.
+        if isinstance(self.target_time, datetime.datetime):
+            self.target_time = self.target_time.isoformat()
         self.log.info("Checking if the time (%s) has come", self.target_time)
         return timezone.utcnow() > self._moment
 
     @property
     def _moment(self) -> datetime.datetime:
-        # target_time is a template field: after rendering it is usually a string, but with
-        # render_template_as_native_obj=True it can already be a datetime.
+        # After rendering target_time is usually a str; native rendering can yield a datetime.
         target_time: Any = self.target_time
         if isinstance(target_time, datetime.datetime):
             return target_time
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -63,30 +63,24 @@ class DateTimeSensor(BaseSensorOperator):
 
     def __init__(self, *, target_time: str | datetime.datetime, **kwargs) -> None:
         super().__init__(**kwargs)
-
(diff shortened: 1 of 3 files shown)


providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -63,30 +63,24 @@ class DateTimeSensor(BaseSensorOperator):
 
     def __init__(self, *, target_time: str | datetime.datetime, **kwargs) -> None:
         super().__init__(**kwargs)
-
-        # self.target_time can't be a datetime object as it is a template_field
-        if isinstance(target_time, datetime.datetime):
-            self.target_time = target_time.isoformat()
```
shahar1, line 69: The `isoformat()` is completely dropped in the new version - it should occur within the `execute()`

shahar1 (dismissed):

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -63,30 +63,26 @@ class DateTimeSensor(BaseSensorOperator):
...
-        # render_template_as_native_obj=True is used.
-        # In this case, the target_time is already a datetime object.
-        if isinstance(self.target_time, datetime.datetime):  # type:ignore[unreachable]
-            return self.target_time  # type:ignore[unreachable]
-
-        return timezone.parse(self.target_time)
+        # After rendering target_time is usually a str; native rendering can yield a datetime.
+        target_time: Any = self.target_time
+        if isinstance(target_time, datetime.datetime):
+            return target_time
```
shahar1: The `isoformat()` normalization moved to `poke()`, but `DateTimeSensorAsync ` overrides `execute()` and never pokes. So a naive datetime passed to the constructor no longer gets localized.

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -63,30 +63,26 @@ class DateTimeSensor(BaseSensorOperator):
...
-        else:
-            raise TypeError(
-                f"Expected str or datetime.datetime type for target_time. Got {type(target_time)}"
-            )
+        # target_time is a template field; store it verbatim and normalize once rendered.
+        self.target_time = target_time
 
     def poke(self, context: Context) -> bool:
+        # Normalize a rendered datetime to an ISO string here (moved from __init__, which only
+        # saw the un-rendered Jinja) so the stored template field stays a string.
```
shahar1: Could you please remove this comment?

shahar1 (commented):
Almost there,
Also - please ensure enough test coverage for all edge cases, thanks!

shahar1 (approved):

PR #70322, opened 2026-07-23
Fix templated validation in Azure operators

Revanth14:
Moves template-field-dependent logic out of `__init__` for three Microsoft Azure provider classes, so it runs against Jinja-rendered values rather than the raw `{{ ... }}` templates that `__init__` sees before rendering:

- `AzureVirtualMachineStateSensor` now validates the rendered `target_state` during execution.
- `GCSToAzureBlobStorageOperator` now checks rendered `match_glob` compatibility during execution.
- `OracleToAzureDataLakeOperator` now uses a value-preserving default assignment for `sql_params`.

As a result, validation for the sensor and the GCS-to-WASB operator now surfaces at task-execution time rather than at DAG-parse time an intentional consequence of not touching template fields before they are rendered.

The corresponding entries were removed from `validate_operators_init_exemptions.txt`.
Regression tests render actual Jinja expressions before execution.

Related: #70296

## Testing

​```console
uv run --project providers/microsoft/azure pytest \
  providers/microsoft/azure/tests/unit/microsoft/azure/sensors/test_compute.py \
  providers/microsoft/azure/tests/unit/microsoft/azure/transfers/test_gcs_to_wasb.py \
  providers/microsoft/azure/tests/unit/microsoft/azure/transfers/test_oracle_to_azure_data_lake.py \
  -xvs
​```


Revanth14:
@shahar1 This is for Azure Operators

providers/microsoft/azure/tests/unit/microsoft/azure/sensors/test_compute.py
```diff
@@ -43,14 +43,17 @@ def test_init(self):
         assert sensor.target_state == "running"
         assert sensor.azure_conn_id == CONN_ID
 
-    def test_init_invalid_target_state(self):
+    def test_invalid_templated_target_state_is_validated_after_rendering(self):
```
shahar1, line 46: Only the rejection path is tested; the behaviour the PR restores (valid templated value surviving to execution) is untested.

shahar1 (changes requested):
Thanks for your contribution!
Please add gen-AI disclosure section to the PR body.
Also, a minor comment.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/transfers/oracle_to_azure_data_lake.py
```diff
@@ -68,12 +68,10 @@ def __init__(
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
-        if sql_params is None:
-            sql_params = {}
         self.filename = filename
         self.oracle_conn_id = oracle_conn_id
         self.sql = sql
-        self.sql_params = sql_params
```
shahar1, line 76: Please restore the changes here (see #70505 for details)

PR #70323, opened 2026-07-23
Drop unused post_submit_commands copy from SparkSubmitOperator init

1fanwang:
`post_submit_commands` is a template field, so it is rendered after `__init__` runs. The constructor also kept a private `self._post_submit_commands = list(post_submit_commands) if post_submit_commands else []`, built from the un-rendered value. Nothing reads that copy: the operator forwards the rendered `self.post_submit_commands` to `SparkSubmitHook`, which builds its own list and runs the commands. Removing it stops the constructor reading the template field and clears its `validate-operators-init` exemption.

related: #70296

`validate_operators_init.py` on the operator, before (with the exemption removed):

```
SparkSubmitOperator's constructor applies logic to template fields. Template fields are
rendered after the constructor runs, so validation or transformation here acts on the
un-rendered Jinja expression and should move to execute():
  line 266: self._post_submit_commands = list(post_submit_commands) if post_submit_commands else []  (post_submit_commands)
```

After removing the line, the validator exits 0.

`test_spark_submit.py` (operator suite): 69 passed. The removed attribute had no readers, so behaviour is unchanged.

</details>

---

### DIFF
--- a/providers/apache/spark/src/airflow/providers/apache/spark/operators/spark_submit.py
+++ b/providers/apache/spark/src/airflow/providers/apache/spark/operators/spark_submit.py
@@ -263,7 +263,6 @@ class SparkSubmitOperator(ResumableJobMixin, BaseOperator):
         self._deploy_mode = deploy_mode
         self._hook: SparkSubmitHook | None = None
         self.post_submit_commands = post_submit_commands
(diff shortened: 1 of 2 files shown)


shahar1 (approved):
Nice catch!

PR #70325, opened 2026-07-23
Fix CohereEmbeddingOperator input_text normalization before rendering

MannXo:
Move the single-string normalization of `input_text` from `__init__` into `execute()`.

`input_text` is a template field, so it is rendered after the constructor runs. Wrapping a plain string into a list in `__init__` acts on the un-rendered Jinja expression, so a templated `input_text` (for example one that resolves to a list under native rendering) is normalized before it is ever rendered. Doing it in `execute()` operates on the rendered value.

Part of the template-field constructor burn-down; removes the `CohereEmbeddingOperator` entry from `scripts/ci/prek/validate_operators_init_exemptions.txt` in the same PR, as the hook requires.

related: #70296

---


shahar1 (approved):
Well done!
For next times - please ask your LLM to reduce amount of comments (the top comment in `execute` isn't necessary).
I resolved a conflict with the text file.

PR #70326, opened 2026-07-23
Fix WeaviateIngestOperator input_data validation before rendering

MannXo:
Move the `input_data is None` required-value check from `__init__` into `execute()`.

`input_data` is a template field, so it is rendered after the constructor runs. Validating it in `__init__` checks the un-rendered value. Running the check in `execute()` validates the value the task actually uses.

Small behavior change: constructing the operator with `input_data=None` no longer raises immediately; the `TypeError` is raised when the task runs, after templating.

Part of the template-field constructor burn-down; removes the `WeaviateIngestOperator` entry from `scripts/ci/prek/validate_operators_init_exemptions.txt` in the same PR, as the hook requires.

related: #70296

---


providers/weaviate/src/airflow/providers/weaviate/operators/weaviate.py
```diff
@@ -75,15 +75,16 @@ def __init__(
...
     @cached_property
     def hook(self) -> WeaviateHook:
         """Return an instance of the WeaviateHook."""
         return WeaviateHook(conn_id=self.conn_id, **self.hook_params)
 
     def execute(self, context: Context) -> None:
+        # input_data is a template field; validate it after rendering rather than in __init__,
+        # where the check would run against the un-rendered value.
+        if self.input_data is None:
+            raise TypeError("input_data is required")
```
shahar1, line 85: The direction is right, but for an `is None` check the "validates the un-rendered value" rationale doesn't apply: a templated `input_data` is a Jinja string or `XComArg` (never `None` at construction), and a literal `None` renders to `None`. The old `__init__` check had no false positives - moving it to `execute()` just delays the common mistake (omitting `input_data`, silent due to the `= None` default) from a parse-time import error to a runtime task failure that burns retries.

The `= None` default is vestigial from the removed input_json alternative (#44745). Suggest dropping it, i.e.,
```
input_data: list[dict[str, Any]] | pd.DataFrame,
```

That restores the parse-time `TypeError` via Python itself, still passes the hook (signature defaults are excluded from its scan), and lets mypy flag input_data=None. Keep the new `execute()` check - it's a real improvement for values that render to `None`, which previously crashed as 'NoneType' object is not iterable inside `hook.batch_data`. Your new test passes unchanged under this signature.

---

Drafted-by: Claude Code (Fable 5); reviewed by me before posting

providers/weaviate/src/airflow/providers/weaviate/operators/weaviate.py
```diff
@@ -75,15 +75,16 @@ def __init__(
...
-            raise TypeError("input_data is required")
-
     @cached_property
     def hook(self) -> WeaviateHook:
         """Return an instance of the WeaviateHook."""
         return WeaviateHook(conn_id=self.conn_id, **self.hook_params)
 
     def execute(self, context: Context) -> None:
+        # input_data is a template field; validate it after rendering rather than in __init__,
+        # where the check would run against the un-rendered value.
```
shahar1: Please remove these comments (they are not too useful)

shahar1 (dismissed):

providers/weaviate/src/airflow/providers/weaviate/operators/weaviate.py
MannXo: Removed.

providers/weaviate/src/airflow/providers/weaviate/operators/weaviate.py
MannXo: Reflected. Dropped the `= None `default on `input_data` -> now `input_data: list[dict[str, Any]] | pd.DataFrame` (required). Python raises `TypeError` at parse time if it's omitted; the `execute()` check stays to catch values that render to `None`.

shahar1 (approved):

PR #70328, opened 2026-07-23
Fix airflowctl trigger sending unrecognized fields to older API servers

### LINKED ISSUE #70327: Airflowctl dags trigger fails with extra_forbidden against Airflow < 3.2.0
### Airflow CTL Version

Latest stable version

### Airflow CTL Command

airflowctl dags trigger example_simplest_dag --dag-run-id <run_id> --env production

### Keyring Backend / Version

Keyring Backend/Headless

### Auth Type

Token

### What is the current behaviour?

`airflowctl dags trigger` fails against any Airflow deployment running a version earlier than 3.2.0, with:

{'type': 'extra_forbidden', 'loc': ['body', 'partition_key'], 'msg': 'Extra inputs are not permitted', 'input': None}

Root cause: `DagOperations.trigger()` in `airflow-ctl/src/airflowctl/api/operations.py` serializes the request body without `exclude_none=True`:

```python
json=trigger_dag_run.model_dump(mode="json")
```

`TriggerDAGRunPostBody` includes `partition_key` and `bundle_version`, both added in Airflow 3.2.0's model. Since these default to `None` and are never excluded from serialization, every trigger request includes them as `null` regardless of the target server's Airflow version. Any server running < 3.2.0 doesn't recognize these fields and rejects the request entirely.

`AssetsOperations.create_event()` in the same file already handles this correctly:
```python
json=asset_event_body.model_dump(mode="json", exclude_none=True)
```

Confirmed via the live OpenAPI schema on a 3.1.x deployment (`curl .../openapi.json`) that `partition_key`/`bundle_version` genuinely don't exist in that server's `TriggerDAGRunPostBody` schema, and confirmed directly against the `apache/airflow` source that these fields were added starting at the `3.2.0` tag (absent in `3.1.0` through `3.1.8`).

### What is the expected results?

`airflowctl dags trigger` should succeed against any Airflow API server version whose `TriggerDAGRunPostBody` schema it's compatible with, i.e., it shouldn't send fields the target server doesn't recognize just because the client's own model happens to declare them, when those fields were never explicitly set by the caller.

### Anything else?

Steps to reproduce:
1. Deploy Airflow 3.1.x (reproduced on 3.1.0 via Charmed Airflow on Kubernetes)
2. `airflowctl auth login --api-url <url> --env <env> --skip-keyring`
3. `airflowctl dags trigger <dag_id> --dag-run-id <run_id> --env <env>`
4. Observe the `extra_forbidden` error

Fix confirmed locally: adding `exclude_none=True` to `trigger()`'s `model_dump()` call resolves the issue, tested both with and without `--logical-date` passed explicitly. PR incoming with the fix plus a regression test.

### Code of Conduct

namita-can:
DagOperations.trigger() didn't exclude None fields when serializing the request, unlike create_event() in the same file. This meant every trigger request included partition_key and bundle_version (added in Airflow 3.2.0), which older API servers reject as unrecognized fields.

Adds exclude_none=True to match create_event()'s existing pattern, plus a regression test asserting unset fields are excluded from the request body.

Manually verified against a live Airflow 3.1.x deployment (Charmed Airflow on Kubernetes): confirmed dags trigger fails with the extra_forbidden error on the unpatched code, and succeeds cleanly — both with and without --logical-date passed explicitly, once the fix is applied.

* closes: #70327 

 

---

- [] Yes (please specify the tool below)

---

### DIFF
--- a/airflow-ctl/src/airflowctl/api/operations.py
+++ b/airflow-ctl/src/airflowctl/api/operations.py
@@ -603,7 +603,7 @@ class DagsOperations(BaseOperations):
             trigger_dag_run.conf = {}
         try:
             self.response = self.client.post(
-                f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json")
+                f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json", exclude_none=True)
             )
             return DAGRunResponse.model_validate_json(self.response.content)
         except ServerResponseError as e:
--- a/airflow-ctl/src/airflowctl/api/operations.py
+++ b/airflow-ctl/src/airflowctl/api/operations.py
@@ -503,8 +503,13 @@ class DagsOperations(BaseOperations):
         """Create a Dag run."""
         if trigger_dag_run.conf is None:
             trigger_dag_run.conf = {}
+        # partition_key and bundle_version were added in Airflow 3.2.0; older API
+        # servers (extra="forbid") reject them as unrecognized fields even when null.
+        # logical_date must stay in the body even when None - the server declares it
+        # without a default, making it a required (though nullable) field.
         self.response = self.client.post(
-            f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json")
+            f"dags/{dag_id}/dagRuns",
+            json=trigger_dag_run.model_dump(mode="json", exclude={"partition_key", "bundle_version"}),
         )
         return DAGRunResponse.model_validate_json(self.response.content)
 
--- a/airflow-ctl/tests/airflow_ctl/api/test_operations.py
+++ b/airflow-ctl/tests/airflow_ctl/api/test_operations.py
@@ -1157,6 +1157,36 @@ class TestDagOperations:
         response = client.dags.trigger(dag_id=self.dag_id, trigger_dag_run=self.trigger_dag_run)
         assert response == self.dag_run_response
 
+    def test_trigger_excludes_none_fields_from_request_body(self):
+        """Regression test: trigger() must exclude unset (None) fields from the
+        request body. Older API servers reject unrecognized fields (e.g.
+        partition_key, bundle_version, added in a later API version) with a
+        422 extra_forbidden error if they're present at all, even as null.
+        """
+        captured_body = {}
+
+        def handle_request(request: httpx.Request) -> httpx.Response:
+            nonlocal captured_body
+            captured_body = json.loads(request.content)
+            return httpx.Response(200, json=json.loads(self.dag_run_response.model_dump_json()))
+
+        client = make_api_client(transport=httpx.MockTransport(handle_request))
+        client.dags.trigger(dag_id=self.dag_id, trigger_dag_run=self.trigger_dag_run)
+
+        # conf is explicitly defaulted to {} by trigger() when unset, so it's expected.
+        assert captured_body == {"conf": {}}
+        for field in (
+            "dag_run_id",
+            "data_interval_start",
+            "data_interval_end",
+            "logical_date",
+            "run_after",
+            "note",
+            "partition_key",
+            "bundle_version",
+        ):
+            assert field not in captured_body, f"{field} should be excluded when unset"
+
 
 class TestDagRunOperations:
     dag_id = "dag_id"
--- a/airflow-ctl/tests/airflow_ctl/api/test_operations.py
+++ b/airflow-ctl/tests/airflow_ctl/api/test_operations.py
@@ -1260,6 +1260,28 @@ class TestDagOperations:
         response = client.dags.trigger(dag_id=self.dag_id, trigger_dag_run=self.trigger_dag_run)
         assert response == self.dag_run_response
 
+    def test_trigger_excludes_unsupported_fields_from_request_body(self):
+        """Regression test: trigger() must exclude partition_key/bundle_version
+        (added in Airflow 3.2.0) so older API servers don't reject the request
+        with extra_forbidden - but must still send logical_date (even as null),
+        since the server declares it as a required (non-defaulted) field.
+        """
+        captured_body = {}
+
+        def handle_request(request: httpx.Request) -> httpx.Response:
+            nonlocal captured_body
+            captured_body = json.loads(request.content)
+            return httpx.Response(200, json=json.loads(self.dag_run_response.model_dump_json()))
+
+        client = make_api_client(transport=httpx.MockTransport(handle_request))
+        client.dags.trigger(dag_id=self.dag_id, trigger_dag_run=self.trigger_dag_run)
+
+        for field in ("partition_key", "bundle_version"):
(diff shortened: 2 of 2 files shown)


airflow-ctl/src/airflowctl/api/operations.py
```diff
@@ -603,7 +603,7 @@ def trigger(
             trigger_dag_run.conf = {}
         try:
             self.response = self.client.post(
-                f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json")
+                f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json", exclude_none=True)
```
potiuk, line 606: `exclude_none=True` drops **every** null field, including `logical_date`, which the server requires to be present (see the body of my review). It also can't distinguish "never provided" from "explicitly set to null", so a user who deliberately passes a null value has it silently dropped.

The stated problem is narrower than this: older servers reject *fields they don't know about*. So the targeted fix is to omit only those, e.g. `exclude={"partition_key", "bundle_version"}` when talking to an older server, ideally gated on a negotiated API version rather than applied unconditionally.

If you'd rather keep it general, `exclude_unset=True` is at least semantically honest about "the caller didn't set this" — but it has the same `logical_date` problem, so it doesn't avoid the regression on its own.

---

airflow-ctl/tests/airflow_ctl/api/test_operations.py
```diff
@@ -1157,6 +1157,36 @@ def handle_request(request: httpx.Request) -> httpx.Response:
...
+            nonlocal captured_body
+            captured_body = json.loads(request.content)
+            return httpx.Response(200, json=json.loads(self.dag_run_response.model_dump_json()))
+
+        client = make_api_client(transport=httpx.MockTransport(handle_request))
+        client.dags.trigger(dag_id=self.dag_id, trigger_dag_run=self.trigger_dag_run)
+
+        # conf is explicitly defaulted to {} by trigger() when unset, so it's expected.
+        assert captured_body == {"conf": {}}
+        for field in (
```
potiuk, line 1178: This asserts `"logical_date" not in captured_body`, which codifies the behaviour I think is the bug — so the test would keep passing after a fix that reintroduces the field.

More broadly, `httpx.MockTransport` returns a canned 200 without validating the body against the server model, so no test here can catch "the server would have rejected this". A test that round-trips the body through `airflow.api_fastapi.core_api.datamodels.dag_run.TriggerDAGRunPostBody.model_validate(captured_body)` would have caught the missing required field immediately, and is worth adding whatever direction the fix takes.

---

potiuk (changes requested):
Thanks — the problem is real: newer optional fields like `partition_key` and `bundle_version` being serialised as `null` will trip `extra="forbid"` on an older API server. Worth fixing.

But `exclude_none=True` overshoots, and I think it breaks the **current** server while fixing the old one. In `airflow-core/src/airflow/api_fastapi/core_api/datamodels/dag_run.py:226` the server-side model declares:

```python
logical_date: AwareDatetime | None
```

— no default, so in Pydantic that is a **required** field that must be present in the body, even when null. Every sibling field has `= None` or a `default_factory`; `logical_date` deliberately does not, and the `@model_validator` doesn't supply one either.

So triggering a Dag run without an explicit logical date — the normal case — will now omit `logical_date` entirely and get a 422 "Field required" from a current server. That's a worse regression than the one being fixed.

Details and a suggested direction inline.

---

airflow-ctl/src/airflowctl/api/operations.py
namita-can: Good catch, @potiuk. Pushed a fix that switches from exclude_none=True to a targeted exclude={"partition_key", "bundle_version"} - this keeps logical_date in the body even when None (since the server requires it, just allows null), while still dropping the two fields older servers reject.

airflow-ctl/tests/airflow_ctl/api/test_operations.py
namita-can: Reworked the test to assert partition_key/bundle_version are excluded and logical_date is present.

airflow-ctl/src/airflowctl/api/operations.py
```diff
@@ -503,8 +503,13 @@ def trigger(
...
         if trigger_dag_run.conf is None:
             trigger_dag_run.conf = {}
+        # partition_key and bundle_version were added in Airflow 3.2.0; older API
+        # servers (extra="forbid") reject them as unrecognized fields even when null.
+        # logical_date must stay in the body even when None - the server declares it
+        # without a default, making it a required (though nullable) field.
         self.response = self.client.post(
-            f"dags/{dag_id}/dagRuns", json=trigger_dag_run.model_dump(mode="json")
+            f"dags/{dag_id}/dagRuns",
+            json=trigger_dag_run.model_dump(mode="json", exclude={"partition_key", "bundle_version"}),
```
potiuk, line 512: Excluding these unconditionally means they can never be sent, including to servers that support them — and both are user-settable today (`cli_config.py:714` generates a CLI flag for every model field bar `schema_`), so `--partition-key foo` is silently discarded rather than rejected.

Conditioning the exclusion on the value keeps both sides working:

```suggestion
        exclude = {
            field
            for field in ("partition_key", "bundle_version")
            if getattr(trigger_dag_run, field) is None
        }
        self.response = self.client.post(
            f"dags/{dag_id}/dagRuns",
            json=trigger_dag_run.model_dump(mode="json", exclude=exclude),
        )
```

An old server no longer sees the unknown null fields; a new server still receives values the user actually set. A user who sets `--partition-key` against an old server then gets a 422, which is the correct outcome — that server genuinely cannot honour the request, and saying so is better than pretending it worked.

---

potiuk (commented):
Thanks — the `logical_date` half is fixed correctly, and the comment explaining *why* it has to stay in the body (no default, so required-but-nullable) is exactly the kind of note that stops someone re-introducing `exclude_none` later. The test asserting its presence is right too.

Two things still open, one of them new.

### The unconditional exclusion silently drops user-supplied values

I checked whether this is reachable, and it is. `TriggerDAGRunPostBody` declares both fields for real, and `cli_config.py:714` generates a `--<field>` flag for every model field except `schema_` (`excluded_parameters = ["schema_"]`). So `airflowctl dags trigger --partition-key foo` is a valid command today: argparse accepts it, the model carries the value, and `trigger()` now strips it before sending. The Dag run is created without the partition key and nothing reports an error.

That is a worse failure than the one being fixed. A 422 from an old server is loud and diagnosable; this is a user watching their partition key disappear with a successful exit code.

Detail and a suggested fix inline.

### The round-trip test

This was in my last review and is worth restating: `httpx.MockTransport` returns a canned 200 without validating the body, so no assertion about dict keys can catch "the server would have rejected this". Validating the captured body through the server's own model would have caught both the original `logical_date` bug and this one:

```python
from airflow.api_fastapi.core_api.datamodels.dag_run import TriggerDAGRunPostBody as ServerBody
ServerBody.model_validate(captured_body)
```

Worth adding whichever way the exclusion goes.

---

PR #70331, opened 2026-07-23
Strip GlueDataQualityOperator ruleset after rendering

1fanwang:
`ruleset` is a template field, so it is rendered after `__init__` runs. The constructor did `self.ruleset = ruleset.strip()`, which strips the un-rendered value. A templated `ruleset` that renders with surrounding whitespace is therefore never stripped, and `validate_inputs()` rejects it because the rendered value no longer ends with `]`:

```
AttributeError: RuleSet must starts with Rules = [ and ends with ]
```

Store `ruleset` verbatim in the constructor and strip the rendered value at the start of `execute()`, before validation.

related: #70296

Operator-level repro (real Dag, real render, real `execute` with a mocked Glue conn). Before the fix, a templated ruleset rendering to `'  Rules = [...]  '` reaches `validate_inputs()` unstripped and raises:

```
AttributeError: RuleSet must starts with Rules = [ and ends with ]
```

After the fix, `execute()` strips first and the API receives the clean `'Rules = [...]'`. The new `test_execute_strips_rendered_ruleset` drives that render→execute path; it fails on the pre-fix source and passes after. `TestGlueDataQualityOperator`: 11 passed.

</details>

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
@@ -449,8 +449,7 @@ class GlueDataQualityOperator(AwsBaseOperator[GlueDataQualityHook]):
                 raise AttributeError("Target table must have DatabaseName and TableName")
 
     def execute(self, context: Context):
-        # ruleset is a template field; strip the rendered value here rather than in __init__,
-        # which only sees the un-rendered Jinja expression.
+        # ruleset is a template field; strip the rendered value here, not in __init__.
         self.ruleset = self.ruleset.strip()
         self.validate_inputs()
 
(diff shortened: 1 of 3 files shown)


providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -449,6 +449,8 @@ def validate_inputs(self) -> None:
                 raise AttributeError("Target table must have DatabaseName and TableName")
 
     def execute(self, context: Context):
+        # ruleset is a template field; strip the rendered value here, not in __init__.
```
shahar1, line 452: @1fanwang for next times - please ask your LLM to avoid adding such commentary :)

shahar1 (approved):

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
1fanwang: Thanks, will do @shahar1

PR #70333, opened 2026-07-23
Validate ProduceToTopicOperator topic after rendering

1fanwang:
`topic` is a template field, so it is rendered after `__init__` runs. The presence check `if not (self.topic and self.producer_function)` ran in the constructor, where a templated `topic` is still the un-rendered Jinja expression (always truthy). A `topic` that renders to empty therefore slipped past the guard, and `execute()` tried to produce to it. Move the check to the start of `execute()`, which runs after rendering.

related: #70296

Operator-level repro (real Dag, real render, real `execute`): a templated `topic` rendering to `""` reaches the producer on the pre-fix code (message delivery is attempted and times out) instead of being rejected. After the fix, `execute()` raises up front:

```
AirflowException: topic and producer_function must be provided. Got topic= and producer_function=...
```

The new `test_execute_rejects_empty_rendered_topic` drives that render→execute path; it fails on the pre-fix source and passes after. `test_produce.py`: 3 passed.

</details>

---


shahar1 (approved):

shahar1:
Failed tests seem unrelated

shahar1 (approved):

PR #70334, opened 2026-07-23
Move template-field checks out of __init__ in Hive sensors

1fanwang:
`table`/`partition` (HivePartitionSensor) and `partition_names` (NamedHivePartitionSensor) are template fields, so they are rendered after `__init__` runs. Two constructors read them:

- `HivePartitionSensor` defaulted `partition` with `if not partition: partition = "ds='{{ ds }}'"` — an imperative read of the field. Switched to the sanctioned `partition or "ds='{{ ds }}'"` idiom (behaviour identical).
- `NamedHivePartitionSensor` raised `TypeError` in `__init__` when `partition_names` was a `str`. A whole-list native template (`partition_names="{{ params.names }}"`) is a `str` before rendering, so the constructor rejected it outright. Moved the array-type check to `poke()`, which runs after rendering.

related: #70296

`validate_operators_init.py` exits 0 for both sensors after the change (RED before, flagging `if not partition:` and `if isinstance(partition_names, str):`).

New `test_native_templated_partition_names` drives the render→poke path with a native-templated list; it fails on the pre-fix source (`TypeError` at construction) and passes after. `test_named_hive_partition.py`: 6 passed, 4 skipped (need a live metastore). HivePartitionSensor's change is a behaviour-preserving refactor, covered by the existing suite.

</details>

---

### DIFF
--- a/providers/apache/hive/src/airflow/providers/apache/hive/sensors/named_hive_partition.py
+++ b/providers/apache/hive/src/airflow/providers/apache/hive/sensors/named_hive_partition.py
@@ -56,9 +56,6 @@ class NamedHivePartitionSensor(BaseSensorOperator):
         super().__init__(poke_interval=poke_interval, **kwargs)
 
         self.next_index_to_poke = 0
-        if isinstance(partition_names, str):
-            raise TypeError("partition_names must be an array of strings")
-
         self.metastore_conn_id = metastore_conn_id
         self.partition_names = partition_names
         self.hook = hook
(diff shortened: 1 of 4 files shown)


providers/apache/hive/src/airflow/providers/apache/hive/sensors/named_hive_partition.py
```diff
@@ -95,6 +92,8 @@ def poke_partition(self, partition: str) -> Any:
         return self.hook.check_for_named_partition(schema, table, partition)
 
     def poke(self, context: Context) -> bool:
+        if isinstance(self.partition_names, str):
+            raise TypeError("partition_names must be an array of strings")
```
shahar1, line 96: Could you please add a test for `TypeError` in its new site?

shahar1 (approved):

PR #70343, opened 2026-07-23
feat: Add optional static_url to BaseOperatorLink to skip XCom for constant links

### LINKED ISSUE #55432: Optional support for static Extra Links without XCom dependency
### Description

This feature request entails the addition of an opt-in feature for extra links which are not stored as an XCom, for static URLs and reduced metaDB bloat.

### Use case/motivation

As far as I'm aware, extra links are only supported via XComs. This makes sense for dynamic, task-instance specific / DAG-run specific URLs (e.g. a link to an external Spark job); however, XComs are _not_ an ideal mechanism for static URLs which are only specific to the given operator (rather than the task instance). For example, a very common use case for the Extra Links feature is to point to a static, operator-specific documentation page. Yet, storing that URL in an XCom for each task instance is clearly not ideal in that circumstance, and only adds bloat to the metaDB.

### Related issues

Only very tangentially related: https://github.com/apache/airflow/issues/54527

Caveat: I am willing to help with the backend implementation, but I have no understanding or experience regarding react / frontend development.

### Code of Conduct

iRAFEEK:
Closes #55432

## Problem

`BaseOperatorLink.get_link()` was an `@abstractmethod`, forcing every
subclass to implement it even when the link URL is a constant (e.g.,
a documentation link). This meant operators with static links still had
to push to XCom during `execute()` and read from the metadata database
on every task run, adding unnecessary overhead.

## Solution

This PR introduces an optional `static_url` property on
`BaseOperatorLink`:

- `static_url` defaults to `None` (no behavior change for existing subclasses)
- `get_link()` is no longer `@abstractmethod`; its default implementation
  returns `static_url` if set, otherwise raises `NotImplementedError`
  with a helpful message guiding the developer to override one or the other
- Subclasses with a constant URL can now override `static_url` instead
  of `get_link()`, with zero XCom involvement

## Backward Compatibility

Fully backward-compatible. All existing subclasses that implement
`get_link()` continue to work without any changes.

## Example

```python
class MyDocsLink(BaseOperatorLink):
    name = "Documentation"
    operators = [MyOperator]

    @property
    def static_url(self) -> str:
        return "https://myoperator.readthedocs.io"
    # No get_link() needed — no XCom push during execute()
```

## Tests

Added `task-sdk/tests/task_sdk/bases/test_operatorlink.py` covering:
- `static_url` defaults to `None`
- Subclass using `static_url` returns correct URL from `get_link()`
- Subclass overriding `get_link()` directly still works (existing behavior)
- Subclass implementing neither raises `NotImplementedError` with a clear message
- `name` remains abstract


aaron-y-chen (commented):
Hi, thanks for the PR! I wonder whether this would still write to XCom when `static_url` is set.

https://github.com/apache/airflow/blob/7217f84bd00c7efc074254f1c544510107163b67/task-sdk/src/airflow/sdk/execution_time/task_runner.py#L2261-L2263

It seems `_xcom_push_to_db()` is executed unconditionally after `get_link()`. Would you mind taking a look?

task-sdk/src/airflow/sdk/bases/operatorlink.py
```diff
@@ -44,6 +44,24 @@ class BaseOperatorLink(metaclass=ABCMeta):
...
+        without reading from XCom. This avoids unnecessary metadata database writes
+        and reads for links whose URL is constant across all task instances
+        (e.g. a link to the operator's documentation page).
+
+        Override this in subclasses where the link URL does not vary per task instance.
+        When ``static_url`` is set, there is no need to implement :meth:`get_link`
+        or push anything to XCom during task execution.
+
+        :return: A static URL string, or ``None`` to use the XCom-based :meth:`get_link`.
+        """
```
zach-overflow, line 62: It might be a good idea to add this information to the [relevant site doc](https://github.com/apache/airflow/blob/main/airflow-core/docs/howto/define-extra-link.rst) too.

iRAFEEK:
@aaron-y-chen good catch, you're right. `finalize()` was pushing to XCom unconditionally after `get_link()`, so a `static_url` link was still persisted every run. Fixed in 8349e44: skip the push when the link is static. Safe because the render path (`SerializedBaseOperator.get_extra_links`) calls `get_link()` directly and never reads that XCom back. Dynamic links are unchanged; added a test covering both.

iRAFEEK:
@zach-overflow good call, added a `static_url` note with an example to `airflow-core/docs/howto/define-extra-link.rst` in the same commit (8349e44). Thanks!

iRAFEEK:
Thanks again for the reviews @aaron-y-chen @zach-overflow! I've pushed:
- the XCom fix (skip the `_xcom_push_to_db()` call in `finalize()` for static links, since the render path resolves them via `get_link()` directly) + a test, and
- the docs note for `static_url`, and
- a small lint follow-up (ruff `PT012` + trailing newlines) after the newer ruff landed via main.

Everything's green locally (including `ruff==0.16.0`). Could a committer kick off / approve the CI workflows when you have a moment? Happy to address anything else. 🙏

iRAFEEK:
Gentle follow-up on this one 🙂

Both review comments have been addressed — the XCom-skip fix (+ test) for @aaron-y-chen's point, and the `static_url` docs note for @zach-overflow's — and lint is green locally (`ruff==0.16.0`). The CI workflows are still awaiting maintainer approval, so they haven't had a chance to run yet.

If a committer could approve the workflow run and take another look when you have a moment, I'd really appreciate it. No rush — thanks for your time and for maintaining Airflow! 🙏

zach-overflow:
@iRAFEEK you'll need to get approvals from the reviewers marked as codeowners for this PR (the reviewers with the badge next to their username). I only commented on this PR since I saw a notification due to it being linked to the issue I filed. LGTM, but my review doesn't have any impact on the merge-ability of this.

Thanks for working on this!

PR #70347, opened 2026-07-23
Fix PsrpOperator option checks

1fanwang:
`command`, `powershell` and `cmdlet` are mutually exclusive, and `arguments` / `parameters` are only valid alongside the latter two. All five are template fields, and every check was written as a truthiness test:

- `exactly_one` was handed a **set literal**, so passing the same value to two options deduped to one and passed validation.
- A provided-but-empty option (`command=""`, `arguments=[]`) read as absent, so valid combinations were rejected and invalid ones accepted.
- `execute()` dispatched on truthiness too, so a `command` that renders to `""` fell through and ran as `powershell`.

The checks stay in `__init__`. Per https://github.com/apache/airflow/pull/70505 the constructor is the only place that can answer "was this argument passed?", so they are rewritten with the `is None` polarity the docs now prescribe. Only the dispatch in `execute()` moves to `is not None`, where the rendered value is finally known.

The `cmdlet`-derived `task_id` default is out of scope here. Dropping it changes task identity, so it goes in a follow-up rather than riding along with a validation fix. That also keeps this one unblocked: https://github.com/apache/airflow/pull/70656 made the same validation fix without the `task_id` change and was closed in favour of this PR, so the burn-down now waits on this landing.

`PsrpOperator` therefore stays on the `validate-operators-init` exemption list, since the hook flags the `task_id` derivation rather than these checks. Restoring that entry also resolved the merge conflict, which was only ever that one line.

related: https://github.com/apache/airflow/issues/70296, https://github.com/apache/airflow/pull/70505

| # | Scenario | Result |
| --- | --- | --- |
| 1 | New/changed cases against the **pre-fix** operator | The new regression cases fail; the existing guards still pass |
| 2 | Full operator suite after the fix | The operator suite passes |

**Raw logs**

1. Red: new tests run against `upstream/main`'s `psrp.py`:

```console
$ git show upstream/main:providers/.../operators/psrp.py > providers/.../operators/psrp.py
$ uv run --project providers/microsoft/psrp pytest providers/microsoft/psrp/tests/unit/microsoft/psrp/operators/test_psrp.py \
    -q -k "invalid_option_combination or empty_option_counts or dispatches_as_command"
E   assert equals failed
E      -(None, )                   +(
E                                +  "cmd.exe /c @'\n\n'@",
E                                +)
=========================== short test summary info ============================
FAILED ...::test_invalid_option_combination[two-options-one-empty]
FAILED ...::test_invalid_option_combination[two-equal-options]
FAILED ...::test_empty_option_counts_as_provided[empty-command]
FAILED ...::test_empty_option_counts_as_provided[arguments]
FAILED ...::test_empty_option_counts_as_provided[parameters]
FAILED ...::test_command_rendering_to_empty_dispatches_as_command
============ 6 failed, 3 passed, 33 deselected, 1 warning in 15.40s ============
```

2. Green: full suite on the fixed operator:

```console
$ uv run --project providers/microsoft/psrp pytest providers/microsoft/psrp/tests/unit/microsoft/psrp/operators/test_psrp.py -q
======================== 42 passed, 1 warning in 3.44s =========================
```

</details>


providers/microsoft/psrp/docs/changelog.rst
```diff
@@ -27,6 +27,17 @@
 Changelog
 ---------
 
+Main
+....
+
+Breaking changes
+~~~~~~~~~~~~~~~~~
```
shahar1: Please remove these two, the release manager will put the note in the right place

providers/microsoft/psrp/src/airflow/providers/microsoft/psrp/operators/psrp.py
```diff
@@ -128,6 +118,12 @@ def __init__(
         self.psrp_session_init = psrp_session_init
 
     def execute(self, context: Context) -> list[Any] | None:
+        if not exactly_one(*{self.command, self.powershell, self.cmdlet}):
```
shahar1: This was originally wrong as it dedupes equaly values before `exactly_one` counts them. Quick fix:

```suggestion
        if not exactly_one(self.command, self.powershell, self.cmdlet):
```

providers/microsoft/psrp/docs/changelog.rst
1fanwang: ack, done in https://github.com/apache/airflow/pull/70347/commits/56b1e50dd0389fbb89d44c4337119e40465e8007

providers/microsoft/psrp/src/airflow/providers/microsoft/psrp/operators/psrp.py
1fanwang: I see, it's not supposed to be set here, fixed in https://github.com/apache/airflow/pull/70347/commits/56b1e50dd0389fbb89d44c4337119e40465e8007

providers/microsoft/psrp/src/airflow/providers/microsoft/psrp/operators/psrp.py
```diff
@@ -106,15 +105,6 @@ def __init__(
...
     ) -> None:
-        args = {command, powershell, cmdlet}
-        if not exactly_one(*args):
-            raise ValueError("Must provide exactly one of 'command', 'powershell', or 'cmdlet'")
-        if arguments and not (powershell or cmdlet):
-            raise ValueError("Arguments only allowed with 'powershell' or 'cmdlet'")
-        if parameters and not (powershell or cmdlet):
-            raise ValueError("Parameters only allowed with 'powershell' or 'cmdlet'")
-        if cmdlet:
-            kwargs.setdefault("task_id", cmdlet)
```
uranusjr, line 117: I don’t think this is correct. The checks are against argument usage, not the rendered results. Errors should be raised even if the template would eventually evaluate to a falsy value IMO.

shahar1: Thanks for the feedback, TP. The current definition in the pre-commit is indeed too coarse, so I'll try to fine-tune it so such cases of checking against argument usage won't be flagged. There were about 20% of false-positive cases like that have already been merged - I'll figure out how to handle them after fixing the pre-commit.

@1fanwang - please hold with this PR until I figure it out

1fanwang: Act, thanks the feedback / heads-up, will keep an eye out on the new issue/pr raised on addressing this at scale in repo.

shahar1: I resolved the conflict in the exemptions list, but this should be fixed before merging :)
See clarification in the "False positive" section: https://github.com/apache/airflow/issues/70296

1fanwang: Thanks Shahar! checking

providers/microsoft/psrp/src/airflow/providers/microsoft/psrp/operators/psrp.py
```diff
@@ -56,8 +56,7 @@ class PsrpOperator(BaseOperator):
     :param command: command to execute on remote host. (templated)
     :param powershell: powershell to execute on remote host. (templated)
     :param cmdlet:
-        cmdlet to execute on remote host (templated). Also used as the default
-        value for `task_id`.
+        cmdlet to execute on remote host (templated).
```
potiuk: This docstring edit is the visible half of the behaviour change — dropping "Also used as the default value for `task_id`" alongside removing the `setdefault`.

Documenting it here is right, but a docstring is not where users find out that their task_ids changed. That needs the newsfragment/changelog entry described in the review body.

---

potiuk (changes requested):
Thanks — the core fix is correct and worth having. `args = {command, powershell, cmdlet}` built a **set**, so equal values collapsed into one entry, and `exactly_one` then counted truthiness rather than presence — meaning an empty-string `command` read as "not provided". Switching to explicit `is not None` checks fixes both problems, and carrying the same correction into `execute()` keeps it consistent.

One blocking concern: this PR also **removes `kwargs.setdefault("task_id", cmdlet)`**, so a Dag that relied on the cmdlet supplying the task_id now gets a different task_id. That changes task identity — history, logs, XComs and UI links all key off it — so it isn't a refactor, it's a breaking change for those users.

Please either:

1. split the `task_id` removal into its own PR so the (uncontroversial) validation fix can land immediately, or
2. keep it here but add a newsfragment in `airflow-core/newsfragments/` explaining that PsrpOperator no longer defaults `task_id` to `cmdlet` and what users must do — plus a line in the provider changelog.

For context: #70656 makes the same validation fix without the `task_id` change. It was opened six days after this one, so I'm closing it in favour of this PR — which means the validation fix now depends on this one moving. Option 1 would unblock it fastest.

These are judgement calls rather than mechanical fixes — I'd welcome your own reasoning in reply. My review was AI-assisted and shouldn't be treated as settled.

---

providers/microsoft/psrp/src/airflow/providers/microsoft/psrp/operators/psrp.py
1fanwang: Reverted with the setdefault , so the docstring documents the task_id default again

1fanwang:
> Thanks — the core fix is correct and worth having. `args = {command, powershell, cmdlet}` built a **set**, so equal values collapsed into one entry, and `exactly_one` then counted truthiness rather than presence — meaning an empty-string `command` read as "not provided". Switching to explicit `is not None` checks fixes both problems, and carrying the same correction into `execute()` keeps it consistent.
> 
> One blocking concern: this PR also **removes `kwargs.setdefault("task_id", cmdlet)`**, so a Dag that relied on the cmdlet supplying the task_id now gets a different task_id. That changes task identity — history, logs, XComs and UI links all key off it — so it isn't a refactor, it's a breaking change for those users.
> 
> Please either:
> 
> 1. split the `task_id` removal into its own PR so the (uncontroversial) validation fix can land immediately, or
> 2. keep it here but add a newsfragment in `airflow-core/newsfragments/` explaining that PsrpOperator no longer defaults `task_id` to `cmdlet` and what users must do — plus a line in the provider changelog.
> 
> For context: #70656 makes the same validation fix without the `task_id` change. It was opened six days after this one, so I'm closing it in favour of this PR — which means the validation fix now depends on this one moving. Option 1 would unblock it fastest.
> 
> These are judgement calls rather than mechanical fixes — I'd welcome your own reasoning in reply. My review was AI-assisted and shouldn't be treated as settled.
> 

thanks! went with option 1, so this PR is validation-only now.

potiuk (approved):

PR #60688, opened 2026-01-17
In GCP Datafusion hook handle start_pipeline() when response is None / Invalid

chirodip98:
closes: #50387
Refference : https://cdap.atlassian.net/wiki/spaces/DOCS/pages/477560983/Lifecycle+Microservices#Start-a-Program

PR #70351, opened 2026-07-24
Replace AirflowException with ValueError in ProduceToTopicOperator

FrankYang0529:
Based on `CLAUDE.md`, the community is actively reducing AirflowException usage. Replace `raise AirflowException(...)` with `raise ValueError(...)` in ProduceToTopicOperator.

 

* related: #70296
---

- [X] Yes - Claude Code

---

### DIFF
--- a/generated/known_airflow_exceptions.txt
+++ b/generated/known_airflow_exceptions.txt
@@ -136,7 +136,6 @@ providers/apache/hive/src/airflow/providers/apache/hive/hooks/hive.py::9
 providers/apache/hive/src/airflow/providers/apache/hive/operators/hive_stats.py::1
 providers/apache/hive/src/airflow/providers/apache/hive/transfers/s3_to_hive.py::6
 providers/apache/kafka/src/airflow/providers/apache/kafka/operators/consume.py::2
-providers/apache/kafka/src/airflow/providers/apache/kafka/operators/produce.py::1
 providers/apache/kafka/src/airflow/providers/apache/kafka/triggers/await_message.py::1
 providers/apache/kylin/src/airflow/providers/apache/kylin/hooks/kylin.py::1
 providers/apache/kylin/src/airflow/providers/apache/kylin/operators/kylin_cube.py::5
--- a/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/produce.py
+++ b/providers/apache/kafka/src/airflow/providers/apache/kafka/operators/produce.py
@@ -23,7 +23,7 @@ from typing import Any
 
 from airflow.providers.apache.kafka.hooks.produce import KafkaProducerHook
 from airflow.providers.common.compat.module_loading import import_string
-from airflow.providers.common.compat.sdk import AirflowException, BaseOperator
+from airflow.providers.common.compat.sdk import BaseOperator
 
 local_logger = logging.getLogger("airflow")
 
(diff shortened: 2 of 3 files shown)


shahar1 (changes requested):
Apparently #70333 was opened earlier and addresses the problem more precisely.
One thing in mind - here you tried to address two issues in one PR (validate_operators_init_exemption.txt + known_airflow_execptions) - I prefer to handle each separtely.

Feel free to tackle the other operators (there are still plenty), or to turn this PR into addressing the known_airflow_exceptions part once the other is merged.

FrankYang0529:
I will turn this one into addressing the known_airflow_exceptions part after https://github.com/apache/airflow/pull/70333 is merged.

shahar1 (commented):
We might need to take an intermediate step of creating an exception class that derives from both AirflowException and ValueError, because it will break the retry mechanism that is based on detecting the exception type.
It had been discussed in the past but was put on hold, I'll try to revive this discussion.

jason810496 (commented):
I wonder should we introduced compatible layer for the exception (e.g. subclass of both AirflowException and ValueError) just in case users leverage the callback hook and catch against the `AirflowException`.

If the scenario I mentioned is valid, this is kind of a breaking change for the users.

xBis7:
What is the goal of changing the exception class? Is it to take a step closer to cleaning up the `AirflowException` type or is it to make sure that when we get an error, it's easier to debug it?

> because it will break the retry mechanism that is based on detecting the exception type.

I've looked at the code and I don't think that there is an issue there. I don't see any check for `AirflowException`.

But there is a good chance that users are wrapping the operator's `execute()` call with a `try - except AirflowException`. As far as users are concerned, this is a breaking change, just like @jason810496 said.

If we can determine the goal for this change, then we can figure out what direction to follow. For example, if this change is going to happen across all operators then we might want to create a new class type extending `AirflowException` as already suggested.

FrankYang0529:
Thanks for the review. This PR started as part of #70296, which moves template field validation out of `__init__`. #70333 fixed that part for this operator first, so I turned this PR into another change I found along the way: replacing `raise AirflowException(...)` with `raise ValueError(...)`. The change follows the [contributing guide](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#dont-raise-airflowexception-directly), and [AGENTS.md](https://github.com/apache/airflow/blob/1cf0d2c884c96e5f54e02630b93cd86e07ccb8af/AGENTS.md?plain=1#L142) says "When you touch code that already raises `AirflowException`, prefer narrowing it to a more specific exception rather than leaving or duplicating it."

Several merged PRs have already replaced `AirflowException` with a built-in exception in code that runs during a task. None of them added a class that extends `AirflowException`.

- #70831
- #70539
- #70540
- #69786
- #70218
- #66279
- #70537
- #70535
- #60688
- #54812
- #55462
- #56911

A user who wraps any of these in `try/except AirflowException` runs into the same change as with this PR. Should these also get a new class that extends `AirflowException`? If so, it seems better to decide that once for all providers rather than in this PR alone. I'm happy to add the class here if we go that way.

jason810496 (commented):
> A user who wraps any of these in try/except AirflowException runs into the same change as with this PR.

My concern is the user facing exception handler like on_failure_callback since 2.x or the retry policy introduced in 3.x.

> Should these also get a new class that extends AirflowException? If so, it seems better to decide that once for all providers rather than in this PR alone. I'm happy to add the class here if we go that way.

The best approach I can think of is introducing compatible exception in core and wait for the new core release, then add the compatible import of that new exception in the providers.

Or another direction is just keep the exception as-is and don't fix, since the swapping the exception for code style best practices might introduce bigger user impact than we expect.

PR #70354, opened 2026-07-24
feat(dag-processing): let bundles supply a per-bundle parse PYTHONPATH

1fanwang:
### Why

A DAG bundle defines its *source* (a path and a version) but not the *environment* its files are parsed in. When one DAG processor serves multiple bundles, every bundle is parsed in that processor's single Python environment. Bundles from different teams that need different dependencies (or different versions of the same one) can't each be parsed against their own — a DAG that imports a team-only or version-pinned package builds fine for the team but fails `DagBag` import when a shared processor parses it. Today the only isolation is running a DAG processor per environment (`--bundle-name` scoping), which costs a process per dependency set.

### What

Add an opt-in `BaseDagBundle.parse_pythonpath` (empty by default). When set, the DAG processor prepends those entries to `sys.path` in the per-file parsing subprocess for that bundle only, so a single processor can parse bundles against their own dependencies (for example each bundle's own virtualenv `site-packages`). Bundles that declare nothing parse exactly as before.

- `BaseDagBundle.parse_pythonpath` property (default `[]`).
- `DagFileParseRequest.bundle_pythonpath`, threaded from `DagFileProcessorManager._create_process` through `DagFileProcessorProcess.start`.
- `_parse_file` prepends it to `sys.path` before parsing.

### Tests

`_parse_file` parses a DAG importing a module present only on the bundle's `parse_pythonpath`, and reproduces the import failure without it; plus `parse_pythonpath` default and override.

---

Generated-by: GitHub Copilot CLI

### DIFF
--- a/airflow-core/src/airflow/dag_processing/bundles/base.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/base.py
@@ -360,6 +360,18 @@ class BaseDagBundle(ABC):
         After `initialize` has been called, all dag files in the bundle should be accessible from this path.
         """
 
+    @property
+    def parse_pythonpath(self) -> list[str]:
+        """
+        Extra ``sys.path`` entries to make importable when *parsing* this bundle's DAG files.
+
+        Override to return absolute paths (for example this bundle's own virtualenv ``site-packages``)
+        so that a single DAG processor can parse bundles that require different dependencies, each
+        against its own. The entries are prepended to ``sys.path`` in the parsing subprocess for this
+        bundle only. Empty by default, so parsing uses the DAG processor's environment unchanged.
+        """
+        return []
+
     @abstractmethod
     def get_current_version(self) -> str | BundleVersion | None:
         """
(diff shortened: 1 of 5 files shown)


1fanwang:
Closing this — on reflection it doesn't earn a core extension point.

Airflow already forks a fresh process per file to parse (the "prevent COW in forked process parsing dag file" path in the manager), and it already supports running a DAG processor per bundle via `--bundle-name`, each in its own environment. That gives true isolation, including bundles that need *conflicting* versions of the same package — which a shared-interpreter `sys.path` prepend fundamentally cannot (you can't import two versions of one top-level package in one interpreter). So `parse_pythonpath` only covers the additive "module missing from the base env" subset that the per-processor model already handles, at the cost of nudging users toward a shared interpreter over separate ones.

I don't have an end-to-end case showing it solves something the existing mechanism can't, so it's not the right shape to add to core. If a concrete need surfaces that per-processor isolation genuinely can't meet, I'll reopen with a reproducer. Thanks.

PR #70359, opened 2026-07-24
Keep S3DeleteObjectsOperator validation in __init__, narrow to ValueError

AmoghAtreya:
### Description
related: #70296 — burn-down of `__init__`-time validation exemptions.

#### Changes Made (`S3DeleteObjectsOperator`, `amazon`):
- Rewrote the `__init__` argument-combination guard with explicit `is not None`
  checks so the `validate-operators-init` hook recognizes it (the old
  comprehension-based check read as a false positive). The logic is unchanged —
  equivalent on every input combination.
- Narrowed the raised exception from `AirflowException` to `ValueError`
  (parse-time only; the `execute()`-time guard is deliberately retained
  unchanged as a safety net for templated `keys` rendering to `None`).
- Removed the `S3DeleteObjectsOperator` entry from
  `scripts/ci/prek/validate_operators_init_exemptions.txt` and updated the
  `generated/known_airflow_exceptions.txt` count (5 → 4).
- Updated the unit test to expect `ValueError`.

works on: #70296

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (Claude Sonnet)

Generated-by: Claude following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

PR #70361, opened 2026-07-24
Move operator-init exemptions list to generated/

shahar1:
`validate_operators_init_exemptions.txt` lived under `scripts/`, which is one of the
paths that earns a `backport-to-v3-3-test` label. Every PR that moves template-field
logic out of an operator's `__init__` and drops that operator's line from the list
therefore got flagged for a backport that never applies — the burn-down
([#70296](https://github.com/apache/airflow/issues/70296)) only ever happens on `main`.

Moving the file to `generated/` fixes that without touching the labeller: `generated/`
matches no rule in `boring-cyborg.yml`. It also puts the two prek allowlists in the
same place, next to `known_airflow_exceptions.txt`.

The hook itself is unchanged apart from the path, which now comes from
`AIRFLOW_ROOT_PATH` like the other prek scripts. The stale-exemption error message
prints the repo-relative path instead of the bare filename, since the file no longer
sits next to the script.

Verified: `prek run validate-operators-init --all-files` passes, and the loader still
reads all 80 exempted classes across 62 files from the new location.

---


shahar1:
I'll merge this PR once handling most of the open PRs in https://github.com/apache/airflow/pull/70359

shahar1:
At this rate we'll get done with the changes even before we get his merged, closing for now - I'll handle the labels manually.

PR #70364, opened 2026-07-24
Align Step Functions input template field assignment

daesung9511:
Align the Step Functions input template field with the corresponding constructor argument and instance attribute.

The operator previously declared `input` as a template field while accepting `state_machine_input` in `__init__` and assigning it to `self.input`. This caused transformed and missing assignment validation failures.

@shahar1 

This change:

* renames the template field from `input` to `state_machine_input`
* assigns the constructor argument directly to `self.state_machine_input`
* updates the execution logic and unit tests to use the aligned attribute name
* removes the operator from the initialization exemption list

Tests:

```bash
breeze run pytest \
  providers/amazon/tests/unit/amazon/aws/operators/test_step_function.py
```

related: #70296

---

* [ ] Yes (please specify the tool below)

---


shahar1 (commented):
Please fix static checks

daesung9511:
> Please fix static checks
Hi @shahar1 I've fixed the issue. Thank you for your time

providers/amazon/src/airflow/providers/amazon/aws/operators/step_function.py
```diff
@@ -97,7 +97,7 @@ def __init__(
         self.state_machine_arn = state_machine_arn
         self.name = name
         self.is_redrive_execution = is_redrive_execution
-        self.input = state_machine_input
+        self.state_machine_input = state_machine_input
```
shahar1, line 102: This change is a breaking change - we should slowly deprecate it

shahar1 (changes requested):

providers/amazon/src/airflow/providers/amazon/aws/operators/step_function.py
daesung9511: Thanks! I kept input as a deprecated alias for state_machine_input and added tests. Does this look like the right approach?

PR #64751, opened 2026-04-06
Deadline Alerts: Add dynamic interval resolution support via Variables

SameerMesiah97:
**Description**

This change allows deadline intervals to be configured dynamically via Airflow Variables by introducing support for interval objects (e.g. `VariableInterval`).

`VariableInterval` defers resolution until deadline evaluation time, where it retrieves the Variable value (interpreted as seconds) and converts it into a `timedelta`. This ensures that intervals are always materialized to `timedelta` before being used at runtime.

To support this, the `deadline_alert.interval` column is migrated from `Float` to `JSON`, allowing serialized interval objects to be stored instead of only numeric values.

**Rationale**

The primary motivation is to enable dynamic configuration of deadline intervals using Airflow Variables without requiring changes to DAG code. This allows operators to adjust deadline behavior externally, making the feature more flexible and easier to manage in production environments.

An explicit `VariableInterval` type is used instead of implicitly resolving Variables to avoid silent fallbacks and hidden side effects. This makes the behavior opt-in and predictable, ensuring that DAG authors are aware when interval values are dynamically sourced from Variables.

Resolving intervals at deadline evaluation time ensures that values are materialized to `timedelta` just before use, while remaining stable for the lifetime of a DAG Run. This approach avoids introducing additional execution-time mechanisms and keeps the implementation aligned with existing deadline evaluation behavior.

**Migrations**

The migration `0117_3_3_0_change_deadline_interval_to_json.py` updates the `deadline_alert.interval` column from `Float` to `JSON` to support storing serialized interval objects. Existing numeric values are preserved and continue to be supported during decoding for backwards compatibility. Serialized `VariableInterval` objects are skipped during downgrade as they do not contain interval values.

**Tests**

Added tests covering:

* `VariableInterval.resolve()` for valid and invalid Variable values.
*  Invalid `None` interval after deserialization.
* DAG Run behavior with `VariableInterval`, including:

  - Stable resolution within the same DAG Run.
  - Failure when the referenced Variable is missing.
  - Coverage added to an existing DAGRun success test for `VariableInterval`.

Other tests have been updated to accommodate the serialization of the `timedelta` object where necessary.

**Documentation**

* Modified 'how-to' docs to mention that `DeadlineAlert `intervals may be dynamically resolved via `VariableInterval`.
* Added docstring for `VariableInterval` with usage example and behavior notes.
* Added the migration to the "Reference for Database Migrations" documentation.

**Backwards Compatibility**

[...]

PR #70370, opened 2026-07-24
Fix TypeError clearing a DAG run when a queued-at deadline interval is stored as JSON

### LINKED ISSUE #70368: Clearing a DAG run/task instance with a DagRunQueuedAtDeadline fails with TypeError: unsupported type for timedelta seconds component: dict
### Apache Airflow version

3.3.0 and later

### What happened

Clearing a failed DAG run or task instance returns HTTP 500 ("Internal Service Error") in the UI when the DAG defines a deadline that references the DAG run's `queued_at` time (`DeadlineReference.DAGRUN_QUEUED_AT`). The API server logs:

```
TypeError: unsupported type for timedelta seconds component: dict
```

The clear should succeed, and the queued-at deadline should be recalculated against the new `queued_at`.

### Root cause

In 3.3.0, migration `0117_3_3_0_change_deadline_interval_to_json.py` (from #64751, which added `VariableInterval` support) changed the `deadline_alert.interval` column from FLOAT to JSON. Existing values were rewritten into the serde envelope, e.g. `{"__classname__": "datetime.timedelta", "__version__": 2, "__data__": 3600.0}`.

`_recalculate_dagrun_queued_at_deadlines()` in `airflow-core/src/airflow/models/taskinstance.py` (called from `clear_task_instances`) still reads the raw column and does `timedelta(seconds=deadline_alert.interval)`, passing the JSON dict straight into `timedelta`. Every other consumer (DagRun creation in `serialization/definitions/dag.py`, the model `__repr__`) decodes the JSON first. Only the clear path was missed.

### How to reproduce

1. Define a DAG with a deadline referencing `DeadlineReference.DAGRUN_QUEUED_AT`.
2. Let a run queue so a `Deadline` row is created.
3. Clear the DAG run or one of its task instances.

The API server returns 500 with the `TypeError` above.

### Workarounds

- Delete the task instance in the UI instead of clearing it.
- Temporarily remove the `deadline` parameter from the DAG.

Yes.

hkc-8010:
## Summary

Clearing a DAG run or task instance whose DAG defines a deadline referencing the run's queued-at time (`DeadlineReference.DAGRUN_QUEUED_AT`) raised `TypeError: unsupported type for timedelta seconds component: dict` and surfaced as an HTTP 500 in the API server.

Since 3.3.0 the `deadline_alert.interval` column is JSON (migration `0117_3_3_0_change_deadline_interval_to_json.py`, from #64751), holding a serialized `timedelta` or variable-backed interval, e.g. `{"__classname__": "datetime.timedelta", "__version__": 2, "__data__": 3600.0}`. `_recalculate_dagrun_queued_at_deadlines` (called from `clear_task_instances`) still passed that raw dict into `timedelta(seconds=...)`. Every other consumer (DagRun creation in `serialization/definitions/dag.py`, the model `__repr__`) decodes the JSON first; only the clear path was missed.

DagRun creation already had the decode plus the interval resolution written out inline, so rather than copy it into the clear path, both steps move into `serialization/decoders.py` and both call sites use them.

## Changes

- `airflow-core/src/airflow/serialization/decoders.py`: add `decode_deadline_alert_model()` (decode a `DeadlineAlert` ORM row) and `resolve_deadline_alert_interval()` (resolve a `SerializedVariableInterval` to a `timedelta`). `decode_deadline_alert` gets its missing return annotation.
- `airflow-core/src/airflow/models/taskinstance.py`: `_recalculate_dagrun_queued_at_deadlines` calls the two helpers instead of building the encoded dict itself. `SerializedVariableInterval.resolve()` raises `ValueError` when the Variable is missing or is not an integer, and `decode_deadline_alert` raises on a payload it refuses to decode, so on the clear path those are caught and logged and that deadline is left at its old time rather than failing the whole clear. DagRun creation still raises.
- `airflow-core/src/airflow/serialization/definitions/dag.py`: `_process_dagrun_deadline_alerts` was doing the same two steps inline and now calls the helpers, so its `DeadlineAlertFields`, `SerializedVariableInterval`, `DAT` and `Encoding` imports are gone.

## Session forwarding

#71968 added the `session` parameter to `SerializedVariableInterval.resolve()`, and #72651 started passing it from DagRun creation. That matters: this runs under the scheduler's `prohibit_commit` guard, and without an explicit session `Variable.get` falls through to `@provide_session`, which hands back the scheduler's own scoped session and then tries to commit and close it.

Moving the resolution into a helper means the helper has to carry that through, so `resolve_deadline_alert_interval()` takes a keyword-only `session` and both call sites pass theirs. Three tests cover it, and each one fails if the argument is dropped:

- `test_forwards_session_to_variable_lookup` (new `tests/unit/serialization/test_decoders.py`) on the helper itself.
- `test_dagrun_deadline_variable_interval_uses_callers_session` on the DagRun creation path, which had no coverage of this before.
- `test_clear_task_instances_recalculates_dagrun_queued_deadlines[variable_interval]` asserts `Variable.get` is called with the same session object the clear is running in, not just a non-`None` one.

## Tests

- `airflow-core/tests/unit/models/test_taskinstance.py`: the existing test stored the interval as a plain float, so it never exercised the JSON path. It now stores the serialized envelope and is parametrized over a fixed `timedelta` and a variable-backed interval. `test_clear_task_instances_skips_deadline_with_unresolvable_interval` is new and covers the skip. Both fixtures also had to start storing a serialized callback, since #72651 made callback decoding strict and the old raw `{"path": ...}` dict is no longer a valid `callback_def`.
- All of the above fail on `main` with the reported `TypeError`.

## SDK coupling

This PR adds no `airflow.sdk` import to core. #71802 and #72651 have both landed since the last review, so `resolve_deadline_alert_interval()` isinstance-checks the core `SerializedVariableInterval` and there is no runtime SDK import left in this path. `generated/known_sdk_imports_in_core.txt` is unchanged by this PR.

closes: #70368

## PR Checklist

- [x] Targeted at the `main` branch
- [x] Tests added (fixed interval, variable-backed interval, the unresolvable-interval skip, and session forwarding at both call sites)
- [x] `prek` hooks pass on the changed files, and `mypy airflow-core` is clean


airflow-core/src/airflow/models/taskinstance.py
```diff
@@ -248,10 +248,26 @@ def _recalculate_dagrun_queued_at_deadlines(
     if not results:
         return
 
+    # Local imports to avoid a circular import between models and serialization.
+    from airflow.sdk.definitions.deadline import VariableInterval
```
kaxil, line 252: I don't like the SDK import here

kaxil: It reintroduces coupling between Task SDK (client) and Scheduler/API-server (server)

amoghrajesh (changes requested):
This propagates the same coupling I raised in #64751 - `isinstance(interval, VariableInterval)` here imports `airflow.sdk.definitions.deadline.VariableInterval` into core airflow at runtime. See https://github.com/apache/airflow/pull/64751#pullrequestreview-4865354516 for details.

airflow-core/src/airflow/models/taskinstance.py
hkc-8010: Removed. `taskinstance.py` no longer imports anything from the SDK. The decode and the interval resolution now live in `serialization/decoders.py` as `decode_deadline_alert_model()` and `resolve_deadline_alert_interval()`, and `SerializedDAG._process_dagrun_deadline_alerts` was doing the same two steps inline, so it calls them too and drops its own `VariableInterval` import.

Net effect on the counts introduced in #71212: `definitions/dag.py` goes 2 to 1, `decoders.py` stays at 2, `taskinstance.py` stays at 1.

hkc-8010:
@amoghrajesh Agreed the counterpart is the real fix, and #71802 is already adding `SerializedVariableInterval` for it. I did not want this bugfix to depend on that landing, so here the SDK type is confined to one function, `resolve_deadline_alert_interval()` in `serialization/decoders.py`, which both DagRun creation and the deadline recalculation now call. That is the only `isinstance` against the SDK dataclass left in this path, so when #71802 lands it changes in one place and no call site moves.

The duplication you flagged is gone too. `taskinstance.py` has no SDK import, and `definitions/dag.py` dropped its `VariableInterval` import, so the hook tightened `generated/known_sdk_imports_in_core.txt` from 2 to 1 for that file.

airflow-core/src/airflow/serialization/decoders.py
```diff
@@ -224,6 +225,35 @@ def decode_deadline_alert(encoded_data: dict):
...
+    """
+    Resolve a decoded alert's interval to a ``timedelta``.
+
+    A ``VariableInterval`` reads its Airflow Variable here, so this is only called at the point
+    a deadline is actually calculated.
+
+    :meta private:
+    """
+    if isinstance(alert.interval, VariableInterval):
+        return alert.interval.resolve()
```
ferruzzi, line 253: resolve() can raise `ValueError` (for example VariableInterval.resolve() raises on a non-integer), so on the clear path this can abort the dagrun clear even though there is a valid deadline in the db.  Since dagrun create and dagrun clear both filter through here now, I think we should be catching and logging in the `taskinstance.py` where this gets called IF we're clearing.

airflow-core/src/airflow/serialization/decoders.py
```diff
@@ -24,6 +24,7 @@
 import dateutil.relativedelta
 
 from airflow._shared.module_loading import import_string
+from airflow.sdk.definitions.deadline import VariableInterval
```
ferruzzi, line 27: I'm not sure about this one.  We moved this sdk import to the module level but left the other sdk import below it as a local?

ferruzzi (commented):
@amoghrajesh looks like the SDK coupling you flagged has been addressed, and I think this is clear to move now if you agree.

`models/taskinstance.py` no longer imports from the SDK at all, and the single remaining `isinstance(..., VariableInterval)` is confined to `resolve_deadline_alert_interval()` in `serialization/decoders.py`.  Worth noting that `decoders.py` already imported `VariableInterval` (function-locally, inside `decode_deadline_alert`), so this consolidates an existing coupling site rather than adding one, and `serialization/definitions/dag.py` loses its own import in the process.  The core to SDK coupling is one file lower than before the PR.

#71802 then replaces that last isinstance in one place, so nothing here has to move again.

This PR is the only one of the three currently touching `_recalculate_dagrun_queued_at_deadlines` that is milestoned and backport labelled, so it is the only route this fix has to 3.3.x.  More context in my comments on #70714 and #71850.

airflow-core/src/airflow/serialization/decoders.py
hkc-8010: Done. The decode and the resolve are now wrapped in `try/except (ValueError, TypeError)` inside `_recalculate_dagrun_queued_at_deadlines`, which logs a warning naming the deadline and the dag run and moves on to the next one, so a missing or non-integer Variable leaves that deadline at its old time instead of failing the clear. `decode_deadline_alert` can also raise `TypeError` on an unknown interval type, so that is caught in the same place. Kept it at the call site rather than in the helper like you suggested, so DagRun creation still raises. New test for it: `test_clear_task_instances_skips_deadline_with_unresolvable_interval`.

airflow-core/src/airflow/serialization/decoders.py
hkc-8010: Gone now. #71802 merged since you wrote this, so `resolve_deadline_alert_interval()` isinstance-checks the core `SerializedVariableInterval` and the module-level SDK import is removed. `decode_deadline_alert` keeps its own function-local `VariableInterval` import exactly as it is on `main`. After the rebase this PR adds no `airflow.sdk` import to core anywhere, and `generated/known_sdk_imports_in_core.txt` is no longer part of the diff.

vatsrahul1001:
@ferruzzi can you review again?

ferruzzi:
@vatsrahul1001 - Amogh has the blocking review, we'll need him to lift that either way, but yes, a re-review on this one is on my list.

ferruzzi:
@hkc-8010 - Thanks for your patience on this one.  There are a bunch of PRs open which all edit the same portion of code.  I've been trying to get them merged in the order that creates the least number of messy merge conflicts, and you are up next (after I get #72651 merged).  It's important that you do this merge right since the previous one in the chain added `session` to this block and it's important we don't drop that in the rebase here.

`serialization/definitions/dag.py:763-766` on main now has:

```
interval = deserialized_deadline_alert.interval

if isinstance(interval, SerializedVariableInterval):
    interval = interval.resolve(session=session) 
```

Passing `session=session` there is a critical bug fix* despite it being an optional parameter in the call.  It would be easy to miss on your rebase, and nothing would necessarily tell you.  You've dropped that line in favor of the new helper, so we have to make sure you carry the fix forward in yours. 

# TLDR

 Here's my proposal for your fix:  Add `Session` to the `if TYPE_CHECKING` import block then:

```
def resolve_deadline_alert_interval(
    alert: SerializedDeadlineAlert, *, session: Session | None = None
) -> datetime.timedelta:
    if isinstance(alert.interval, SerializedVariableInterval):
        return alert.interval.resolve(session=session)
    return alert.interval
```

#71968 added `test_resolve_forwards_session` but it only checks `resolve()` itself, not what helpers are using, so nothing is protecting your particular change.  Maybe have a look at how that test works and see if you can think of a test that would guarantee a provided session gets passed through and is the one that is used (as opposed to creating a new session), that would be outstanding.


## * The technical reason, if you care:  

This code runs under the the scheduler's `prohibit_commit` guard.  When `Variable.get` is called without a `session`, it falls through to `@provide_session` in `MetastoreBackend.get_variable`.  Since `settings.Session` is a `scoped_session`, instead of opening a fresh `session` it just hands back the open `session` that the scheduler is holding and tries to commit and close it on the way out.  Except (most importantly) the guard prevents that commit and rolls back the session.  You can check out `get_variable_from_secrets` at/around `models/variable.py:502-505` if you want to see exactly what we're trying to avoid.  Either way, passing the `session` in avoids the `provide_session` wrapper and that bug, entirely.

ferruzzi:
@hkc-8010  72651merged this morning which makes this one the head of the snake with 5 more blocked behind it.   I'd love to get this one rebased, re-reviewed, and merged ASAP.   

@amoghrajesh - I think your 5 August concern is already retired.  You flagged the `isinstance(interval, VariableInterval)` runtime coupling; hkc pointed that to the core `SerializedVariableInterval` back on 10 September, and the only remaining runtime import of the SDK class was on main, which #72651 removed by mapping the classname onto `SerializedVariableInterval` instead.  That deletion should get pulled in with the next rebase.  I don't think there's any SDK coupling left in this path.

hkc-8010:
@ferruzzi Rebased onto main with #72651 in. The `session=session` is carried forward. Since the resolution moved into a helper, the helper takes it:

```python
def resolve_deadline_alert_interval(
    alert: SerializedDeadlineAlert, *, session: Session | None = None
) -> datetime.timedelta:
```

and both DagRun creation and the clear path pass theirs.

On the test, there are three, and each one goes red if the argument is dropped (I checked by deleting it and re-running):

- `test_forwards_session_to_variable_lookup`, in a new `tests/unit/serialization/test_decoders.py`, on the helper itself.
- `test_dagrun_deadline_variable_interval_uses_callers_session` in `test_dagrun.py`, on the DagRun creation path. Nothing covered that before, so this one is protecting your fix rather than mine.
- `test_clear_task_instances_recalculates_dagrun_queued_deadlines[variable_interval]` now asserts `Variable.get` is called with the same session object the clear is running in, not just a non-`None` one.

One other thing the rebase turned up, since you are sequencing these. #72651 made callback decoding strict, and the clear path decodes the callback now where before it only touched the interval. So the raw `{"path": ..., "kwargs": {}}` dict the test fixtures stored is no longer a valid `callback_def` and two tests failed with `Deadline callback does not name a class`. They store `serialize(AsyncCallback(...))` now, which is what `serialized_dag` actually writes. Worth flagging because the unresolvable-interval test kept passing straight through that, just for the wrong reason.

@amoghrajesh agreed on the coupling, there is nothing left in this path. The SDK import check exits clean and `generated/known_sdk_imports_in_core.txt` is not part of the diff.

airflow-core/tests/unit/models/test_taskinstance.py
```diff
@@ -4205,13 +4263,78 @@ def test_clear_task_instances_recalculates_dagrun_queued_deadlines(dag_maker, se
...
+        mock.patch("airflow.models.taskinstance.log") as mock_log,
+    ):
+        clear_task_instances(tis, session)
+
+    dag_run = session.scalar(select(DagRun).where(DagRun.id == dag_run.id))
+    assert dag_run.queued_at > original_queued_at
+
+    deadline = session.scalar(select(Deadline).where(Deadline.dagrun_id == dag_run.id))
+    assert deadline.deadline_time == original_deadline_time
+    assert mock_log.warning.call_count == 1
```
ferruzzi, line 4342: Non-blocking thought:   More than just "something threw an error", we can (should?) check that it's the error we expected (not some random breaking change later), and that it's _our_ deadline that threw it.  That last one maybe overkill, but doesn't hurt?

```suggestion
    assert mock_log.warning.call_count == 1
    msg, *args = mock_log.warning.call_args.args
    assert "interval could not be resolved" in msg
    assert args[0] == deadline.id
```

airflow-core/src/airflow/models/taskinstance.py
```diff
@@ -248,10 +248,29 @@ def _recalculate_dagrun_queued_at_deadlines(
...
+        # We can't use evaluate_with() since the new queued_at is not written to the DB yet, and
+        # interval is stored as JSON, so it has to be decoded rather than passed to timedelta().
+        try:
+            interval = resolve_deadline_alert_interval(
+                decode_deadline_alert_model(deadline_alert), session=session
+            )
+        except (ValueError, TypeError):
+            # A variable-backed interval resolves against an Airflow Variable that may be missing
+            # or non-numeric. Leave this deadline alone rather than failing the whole clear.
+            log.warning(
```
ferruzzi: Non-blocking thought:  Only worth it if you are making other changes already, but the message might be a little misleading.  The `try` is covering decode and resolve, so if the decoder refuses the callback, I'm pretty sure that ends up tripping this message too.

You have `exc_info=True`, so the actual stack trace gets bubbled up and they'll see the real cause, so it's not critical.  Something like "deadline alert could not be decoded" or drop the exact cause entirely and say "Error while recalculating deadline" might be more accurate than "interval could not be resolved".  (and the same update to the comment just before the `try` would be appropriate)

ferruzzi (approved):
Thanks for your patience on this while we untangle the PR knot.  Left two (final?) thoughts, neither are blocking and can be closed unless you are making other changes.

airflow-core/src/airflow/models/taskinstance.py
hkc-8010: Took your second option. It now reads "Error while recalculating deadline %s for DagRun %s.%s, leaving it unchanged" and the cause is left to exc_info, and I updated the comment above the try to match.

You are right that the decoder trips it, and it is not hypothetical. On the last rebase #72651 made callback decoding strict, the fixtures' hand-written callback_def stopped being valid, and this message reported an unresolvable interval while the interval was fine.

airflow-core/tests/unit/models/test_taskinstance.py
hkc-8010: Done, with one addition. Your message and deadline.id assertions are in, reworded for the new log line from the other thread, plus:

```python
mock_variable_get.assert_called_once_with("missing_deadline_interval_key", session=session)
```

That is the one with teeth, and it comes out of your other comment. Since the try covers decode too, this test passes when decode fails before the interval is ever read, which is what it was actually doing right after the #72651 rebase. I checked by putting the old raw-dict callback_def back: it now fails with "Expected 'get' to be called once. Called 0 times." where before it stayed green.

ferruzzi: Nice, I like it.

ferruzzi (approved):
Re-approving since there have been a few changes.  

~~Currently one failing CI test is flake, I'll re-trigger it when the suite finishes.~~

airflow-core/src/airflow/serialization/decoders.py
```diff
@@ -338,6 +341,43 @@ def decode_deadline_alert(encoded_data: dict):
...
+    """
+    Decode a ``DeadlineAlert`` ORM row into its serialized representation.
+
+    :meta private:
+    """
+    return decode_deadline_alert(
+        {
+            DeadlineAlertFields.REFERENCE: deadline_alert.reference,
+            DeadlineAlertFields.INTERVAL: deadline_alert.interval,
+            DeadlineAlertFields.CALLBACK: deadline_alert.callback_def,
```
amoghrajesh, line 354: This helper is named as a general "decode a DeadlineAlert ORM row", but it skips the row's `name`. The model has a `name` column and `decode_deadline_alert` reads it (line 337 just above), so `SerializedDeadlineAlert.name` comes back as None even when the row has one.

Nothing reads `.name` on either of your two callers, so this isn't a bug as of today but its a trap for the next caller. Worth adding:
```
    DeadlineAlertFields.NAME: deadline_alert.name,
```

airflow-core/src/airflow/serialization/decoders.py
```diff
@@ -338,6 +341,43 @@ def decode_deadline_alert(encoded_data: dict):
...
+        }
+    )
+
+
+def resolve_deadline_alert_interval(
+    alert: SerializedDeadlineAlert, *, session: Session | None = None
```
amoghrajesh, line 360: The signature makes `session` optional, but the docstring right below spends three lines explaining that a caller who omits it gets its session rolled back under `prohibit_commit`. Both callers bypass it.

Can this be `*, session: Session` with no default? That turns a silent misuse into a TypeError at the caller, which is the whole point of the three tests you added for this.

airflow-core/src/airflow/serialization/decoders.py
```diff
@@ -338,6 +341,43 @@ def decode_deadline_alert(encoded_data: dict):
...
+    Resolve a decoded alert's interval to a ``timedelta``.
+
+    A ``SerializedVariableInterval`` reads its Airflow Variable here, so this is only called at
+    the point a deadline is actually calculated. It raises ``ValueError`` if the Variable is
+    missing or is not an integer number of seconds.
+
+    :param alert: The decoded alert whose interval should be resolved.
+    :param session: Existing SQLAlchemy Session. Both callers run under the scheduler's
+        ``prohibit_commit`` guard, so the open session has to reach ``Variable.get`` instead of
+        ``provide_session`` handing back the same scoped session and rolling it back on exit.
```
amoghrajesh, line 372: Small thing: this session rationale is written out four times - here, and again in test_dagrun.py:1568, test_taskinstance.py:4248, and test_decoders.py:41. Keep it here as the source of truth and let the tests point at it

airflow-core/tests/unit/models/test_taskinstance.py
```diff
@@ -4130,8 +4130,28 @@ async def empty_callback_for_deadline():
...
+    """Test that clearing tasks recalculates all (and only) DAGRUN_QUEUED_AT deadlines.
+
+    Since Airflow 3.3 the ``deadline_alert.interval`` column is JSON (a serialized ``timedelta``
+    or ``VariableInterval``), so the recalculation must decode it instead of passing the raw value
+    to ``timedelta()``. Storing the interval via ``serialize`` here mirrors production and covers
+    both interval kinds.
+    """
+    from airflow.models.variable import Variable
+    from airflow.sdk.definitions.deadline import VariableInterval
+    from airflow.sdk.serde import serialize
```
amoghrajesh, line 4150: Top level import pls.

airflow-core/tests/unit/models/test_taskinstance.py
```diff
@@ -4205,13 +4263,88 @@ def test_clear_task_instances_recalculates_dagrun_queued_deadlines(dag_maker, se
...
+def test_clear_task_instances_skips_deadline_with_unresolvable_interval(dag_maker, session):
+    """A variable-backed interval that cannot be resolved must not abort the clear.
+
+    ``SerializedVariableInterval.resolve()`` raises ``ValueError`` when the Airflow Variable is
+    missing or is not an integer, and that happens while the DAG run is being cleared. The clear
+    should still go through, leaving the unresolvable deadline at its old time.
+    """
+    from airflow.models.variable import Variable
+    from airflow.sdk.definitions.deadline import VariableInterval
+    from airflow.sdk.serde import serialize
```
amoghrajesh, line 4281: Top level import pls.

airflow-core/src/airflow/serialization/decoders.py
ferruzzi: I am embarrassed that I missed this.  Thank you, Amogh.  This is exactly the bug chain I've been tangling with for the last several days.

airflow-core/src/airflow/serialization/decoders.py
hkc-8010: Fixed, along with a test so it stays fixed.

You are right that nothing reads `.name` today, and that is exactly what made this worth a test rather than just a one-line patch: with no caller and no coverage, the next refactor drops it again and nothing goes red. `TestDecodeDeadlineAlertModel` in `test_decoders.py` now asserts the round trip. I checked it is a real differential by reverting the added line, which fails with `-None +'my_deadline'`.

airflow-core/src/airflow/serialization/decoders.py
hkc-8010: Done, it is now `*, session: Session` with no default.

Worth noting this costs nothing at the call sites: all three (`definitions/dag.py`, `models/taskinstance.py`, and the tests) already passed `session=` explicitly, so this is a pure tightening. Anyone who gets it wrong in future now gets a `TypeError` at the call site instead of a silent rollback, which is the point.

Separately, while you are here: the only thing still blocking this PR is your `CHANGES_REQUESTED` from 2026-08-05 about the SDK coupling. That coupling is gone since #71802 and #72651 landed. `models/taskinstance.py` no longer imports from the SDK at all, and `known_sdk_imports_in_core.txt` is not in the diff. Would you mind re-reviewing so it can move?

airflow-core/src/airflow/serialization/decoders.py
hkc-8010: Agreed, that was four copies of the same paragraph drifting apart. The full rationale now lives only in the `resolve_deadline_alert_interval()` docstring, and the three tests point at it by name.

airflow-core/tests/unit/models/test_taskinstance.py
hkc-8010: Hoisted.

airflow-core/tests/unit/models/test_taskinstance.py
hkc-8010: Hoisted.

PR #70385, opened 2026-07-24
Sanitize error message in DatabricksExecutionTrigger 

fabbuc-gyg:
## What

Strip characters that Postgres `jsonb` cannot store (NUL bytes and unpaired UTF-16 surrogates) from a failed Databricks task's error text, in both `extract_failed_task_errors` and `extract_failed_task_errors_async`.

## Why

`DatabricksExecutionTrigger` puts each failed task's error into its `TriggerEvent` which gets persisted in the deferred task instance's `next_kwargs`, which is a Postgres `jsonb` column. Since Databricks output can contain NUL bytes, postgres rejects this with an error like this:

```
sqlalchemy.exc.DataError: (psycopg2.errors.UntranslatableCharacter) unsupported Unicode escape sequence
DETAIL: \u0000 cannot be converted to text.
```

Because the invalid event is retried, the triggerer enters a `CrashLoopBackOff` which makes it unhealthy.

## How

A small `make_jsonb_safe()` helper regex-strips `\x00` and `\ud800-\udfff` from the error string (non-string values pass through unchanged). It is applied to the extracted error in both the sync and async helpers, so the emitted trigger event is always persistable.

## Tests

Added unit tests in `providers/databricks/tests/unit/databricks/utils/test_databricks.py`:
- direct `make_jsonb_safe` cases (NUL strip, unpaired-surrogate strip, clean-text passthrough, non-string passthrough)
- `extract_failed_task_errors` and `extract_failed_task_errors_async` sanitize error output

---

- [X] Yes - Cursor was used to help draft this change; the author reviewed it.

Generated-by: Cursor

### DIFF
--- a/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
@@ -28,13 +28,6 @@ _JSONB_INVALID_CHARS = re.compile(r"[\x00\ud800-\udfff]")
 def make_jsonb_safe(error: str | int) -> str | int:
     """
     Strip characters that cannot be stored in a Postgres ``jsonb`` column from error text.
-
-    Databricks run output is arbitrary external text and can contain NUL bytes or
-    unpaired UTF-16 surrogates (for example when a task emits binary data). These are
-    rejected by Postgres ``jsonb``, and since a failed task's error is persisted into
-    the deferred task instance's ``next_kwargs`` via the trigger event, an unsanitised
-    value crashes the triggerer when it writes the event. Non-string values are returned
-    unchanged.
     """
     if isinstance(error, str):
         return _JSONB_INVALID_CHARS.sub("", error)
--- a/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
@@ -17,8 +17,28 @@
 # under the License.
 from __future__ import annotations
 
+import re
+
 from airflow.providers.common.compat.sdk import AirflowException, XComArg
 from airflow.providers.databricks.hooks.databricks import DatabricksHook, RunState
+
+_JSONB_INVALID_CHARS = re.compile(r"[\x00\ud800-\udfff]")
+
+
+def make_jsonb_safe(error: str | int) -> str | int:
+    """
+    Strip characters that cannot be stored in a Postgres ``jsonb`` column from error text.
+
+    Databricks run output is arbitrary external text and can contain NUL bytes or
+    unpaired UTF-16 surrogates (for example when a task emits binary data). These are
+    rejected by Postgres ``jsonb``, and since a failed task's error is persisted into
+    the deferred task instance's ``next_kwargs`` via the trigger event, an unsanitised
+    value crashes the triggerer when it writes the event. Non-string values are returned
+    unchanged.
+    """
+    if isinstance(error, str):
+        return _JSONB_INVALID_CHARS.sub("", error)
+    return error
 
 
 def normalise_json_content(content, json_path: str = "json") -> str | bool | list | dict | XComArg:
@@ -75,7 +95,9 @@ def extract_failed_task_errors(
                     error = run_output["error"]
                 else:
                     error = run_state.state_message
-                failed_tasks.append({"task_key": task_key, "run_id": task_run_id, "error": error})
+                failed_tasks.append(
+                    {"task_key": task_key, "run_id": task_run_id, "error": make_jsonb_safe(error)}
+                )
     return failed_tasks
 
 
@@ -101,7 +123,9 @@ async def extract_failed_task_errors_async(
                     error = run_output["error"]
                 else:
                     error = run_state.state_message
-                failed_tasks.append({"task_key": task_key, "run_id": task_run_id, "error": error})
+                failed_tasks.append(
+                    {"task_key": task_key, "run_id": task_run_id, "error": make_jsonb_safe(error)}
+                )
     return failed_tasks
 
 
--- a/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/utils/databricks.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import re
+from typing import Any
 
 from airflow.providers.common.compat.sdk import AirflowException, XComArg
 from airflow.providers.databricks.hooks.databricks import DatabricksHook, RunState
@@ -25,10 +26,8 @@ from airflow.providers.databricks.hooks.databricks import DatabricksHook, RunSta
 _JSONB_INVALID_CHARS = re.compile(r"[\x00\ud800-\udfff]")
 
 
-def make_jsonb_safe(error: str | int) -> str | int:
-    """
-    Strip characters that cannot be stored in a Postgres ``jsonb`` column from error text.
-    """
+def make_jsonb_safe(error: Any) -> Any:
+    """Strip characters that cannot be stored in a Postgres ``jsonb`` column from error text."""
     if isinstance(error, str):
         return _JSONB_INVALID_CHARS.sub("", error)
     return error
--- a/providers/databricks/tests/unit/databricks/utils/test_databricks.py
+++ b/providers/databricks/tests/unit/databricks/utils/test_databricks.py
@@ -27,6 +27,7 @@ from airflow.providers.databricks.hooks.databricks import RunState
 from airflow.providers.databricks.utils.databricks import (
     extract_failed_task_errors,
     extract_failed_task_errors_async,
+    make_jsonb_safe,
     normalise_json_content,
     validate_trigger_event,
 )
@@ -94,6 +95,18 @@ class TestDatabricksOperatorSharedFunctions:
         event = {}
         with pytest.raises(AirflowException):
             validate_trigger_event(event)
+
+    def test_make_jsonb_safe_strips_nul_bytes(self):
+        assert make_jsonb_safe("PK\x03\x04binary\x00error") == "PK\x03\x04binaryerror"
+
+    def test_make_jsonb_safe_strips_unpaired_surrogates(self):
+        assert make_jsonb_safe("bad\ud800surrogate\udfff") == "badsurrogate"
+
+    def test_make_jsonb_safe_leaves_clean_text_unchanged(self):
+        assert make_jsonb_safe(ERROR_MESSAGE) == ERROR_MESSAGE
+
+    def test_make_jsonb_safe_passes_through_non_strings(self):
+        assert make_jsonb_safe(123) == 123
 
 
 class TestExtractFailedTaskErrors:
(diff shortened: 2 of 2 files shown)


eladkal:
cc @moomindani for Databricks team review

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
```diff
@@ -17,9 +17,22 @@
...
+import re
+
 from airflow.providers.common.compat.sdk import AirflowException, XComArg
 from airflow.providers.databricks.hooks.databricks import DatabricksHook, RunState
 
+_JSONB_INVALID_CHARS = re.compile(r"[\x00\ud800-\udfff]")
+
+
+def make_jsonb_safe(error: str | int) -> str | int:
+    """
```
moomindani, line 29: `ruff` fails here with `D200 One-line docstring should fit on one line`, which is what the CI Static checks job is failing on. `main` passes this hook, so it comes from this PR.

```python
def make_jsonb_safe(error: str | int) -> str | int:
    """Strip characters that cannot be stored in a Postgres ``jsonb`` column from error text."""
```

As a first-time contributor you may not have the hooks installed yet — `uv tool install prek && prek install` will run these locally before each commit and catch this class of failure before pushing. See [static code checks](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
```diff
@@ -17,9 +17,22 @@
...
 
+import re
+
 from airflow.providers.common.compat.sdk import AirflowException, XComArg
 from airflow.providers.databricks.hooks.databricks import DatabricksHook, RunState
 
+_JSONB_INVALID_CHARS = re.compile(r"[\x00\ud800-\udfff]")
+
+
+def make_jsonb_safe(error: str | int) -> str | int:
```
moomindani, line 28: Minor, non-blocking: the annotation `error: str | int` is narrower than what the function actually accepts — it passes through any non-string unchanged, and callers here only ever pass `run_output["error"]` or `state_message`, both of which are strings in practice. Since the guard exists to be defensive about whatever Databricks returns, `error: Any -> Any` (or just `str` if you want to assert the contract) would describe the behaviour more honestly than `str | int`, which suggests `int` is a meaningful case when it is only incidental.

moomindani (commented):
Thanks for this, and welcome — this is a real bug with a real operational impact, and the diagnosis is correct.

I verified it end-to-end against a local Postgres 16.13 rather than reasoning about it, writing the actual payload shapes into a `jsonb` column:

| Case | Result |
|---|---|
| `errors[].error` with NUL, unsanitized | rejected: `UntranslatableCharacter: unsupported Unicode escape sequence` — exactly the error you reported |
| same, via `make_jsonb_safe` | stored OK |
| unpaired surrogate, unsanitized | rejected: `InvalidTextRepresentation: invalid input syntax for type json` |
| same, sanitized | stored OK |

So both classes you strip are genuinely rejected by Postgres, and the fix genuinely resolves them. The triggerer CrashLoopBackOff consequence follows from the event being retried, which makes this worth fixing rather than working around.

I also checked the regex is not over-broad: `\x01`, `\x1f`, newlines, tabs, and non-BMP characters (emoji) are all accepted by `jsonb`, and your pattern preserves all of them. It strips exactly the two rejected classes and nothing else — that scoping is right.

One thing I want to record because I initially suspected it was a gap and it turned out not to be: the same `TriggerEvent` carries `run_state`, which embeds Databricks-controlled `state_message`. That is *not* a second hole, because `RunState.to_json()` stringifies it first, so a NUL becomes a literal `\u0000` escape inside a JSON string value and Postgres accepts it. I confirmed that stores fine. Scoping the fix to `errors` is correct — no need to widen it.

One blocker before this can merge: a ruff failure, flagged inline. CI's Static checks job is already failing on it.

---

eladkal:
@fabbuc-gyg can you please fix the open items?

fabbuc-gyg:
Hi @eladkal, So sorry for the delay. Will check this out today.

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
fabbuc-gyg: Fixed — docstring is now a single line. Also installed `prek` locally so this class of failure gets caught before pushing. Thanks for the pointer.

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
fabbuc-gyg: Agreed, changed to `error: Any -> Any`. `str | int` was misleading — `int` was never a meaningful case, the guard is just there to be defensive about whatever Databricks puts in `run_output["error"]`.

fabbuc-gyg:
Thanks @moomindani for the thorough verification, and sorry for the slow turnaround @eladkal.

Both open items are addressed, and the branch is rebased on current `main`:

- **D200 (blocker):** docstring is now a single line. Verified with the pinned `ruff==0.16.4` from `.pre-commit-config.yaml` — `ruff check` and `ruff format --check` both pass on the two touched files. `prek` is installed locally now.
- **Annotation:** `error: str | int -> str | int` is now `error: Any -> Any`. You were right that `str | int` implied `int` was meaningful when it was only incidental; the guard is there to be defensive about whatever Databricks returns in `run_output["error"]`, and `Any` says that honestly. Behaviour is unchanged, so the existing tests (including the non-string passthrough case) still cover it.

No changes to the sanitizer logic or its scoping — appreciated the confirmation that `run_state.state_message` is already safe via `RunState.to_json()` and that the character class is not over-broad.

CI on the new commit is sitting in `action_required`, so it needs a committer to approve the run.

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
fabbuc-gyg: Ok thanks :-)

providers/databricks/src/airflow/providers/databricks/utils/databricks.py
fabbuc-gyg: Thanks, I adjusted it

fabbuc-gyg:
Should be fine now :-)

moomindani (approved):
Both points from my July review are closed on `f1bb073570`. LGTM.

Worth recording because it is not visible from the diff: I mutation-tested the sanitization tests, since tests like these easily pass either way. Making `make_jsonb_safe` return its argument unchanged fails exactly the four stripping tests and leaves the clean-text and non-string cases green — they pin the behaviour.

#72313 touches the same two `failed_tasks.append(...)` sites, so whichever lands second needs a trivial rebase; the changes compose.

---

seanmuth (commented):
Thanks for tracking this down — we hit this exact crash in production yesterday: Airflow 3.1.8, triggerer stuck in `CrashLoopBackOff` for ~10h on a `DatabricksExecutionTrigger` whose failed task's `error` output carried a raw NUL byte from a Spark stack trace (`"col" was called from <command--1>` — a column-not-found `AnalysisException`). Confirmed `make_jsonb_safe()` on `extract_failed_task_errors`/`_async` would have prevented that specific crash. Nice, tight fix.

One gap I'd flag before merge, though: `RunState.to_json()` (`hooks/databricks.py:147`) does `json.dumps(self.__dict__)` on `state_message` with no sanitization, and `DatabricksExecutionTrigger` embeds that string verbatim as the `"run_state"` key of the *same* `TriggerEvent` — at `triggers/databricks.py:155`, and again at `:181`/`:208` in the workflow-task retry/terminal branches. `state_message` is the exact same Databricks free-text field this PR already treats as untrusted (it's the `error` fallback here) — so a NUL byte or lone surrogate landing in `state_message` instead of `run_output["error"]` reproduces the identical crash, entirely bypassing `make_jsonb_safe()`:

```python
>>> import json
>>> json.dumps({"state_message": "boom \u0000 more"})
'{"state_message": "boom \\u0000 more"}'
```

That `\u0000` escape survives nested inside the `"run_state"` string, and hits the same Postgres `UntranslatableCharacter` once the outer payload lands in `next_kwargs` (jsonb) — same failure mode, different field, and it ships unfixed even after this PR merges.

Worth noting this isn't unique to `RunState` either — `ClusterState.to_json()` and `SQLStatementState.to_json()` (same file) follow the identical `state_message` → unsanitized `json.dumps(self.__dict__)` pattern, so the same class of bug likely reaches other Databricks triggers too.

Given that, I'd suggest sanitizing at the source — inside `to_json()` itself (or wherever `state_message` gets assigned) — rather than only at the two `extract_failed_task_errors` call sites, so every caller gets the protection instead of just this one. Doesn't need to block this PR if you'd rather land the reported-crash fix now and follow up separately, but I'd hate to see this merge as "the NUL byte fix" and have the next crash report be the same root cause from a different field.

---
Drafted-by: Claude Code (Sonnet 5); reviewed by @seanmuth before posting

moomindani:
The production case is useful — an independent report with a real `AnalysisException` payload is better evidence for this fix than my synthetic one.

The second hole is not there, though, and I checked rather than reasoned. I replayed both payload shapes into a real Postgres 16.13 `jsonb` column the way SQLAlchemy writes them (`json.dumps` text cast to `jsonb`), with the same dirty string — a NUL plus a lone surrogate — in each:

| payload | result |
|---|---|
| `errors[].error`, unsanitized | rejected: `UntranslatableCharacter: unsupported Unicode escape sequence` |
| `run_state = RunState.to_json()` carrying that `state_message` | **stored** |
| `run_state` as a nested dict instead of a `to_json()` string | rejected |

The double encoding is what saves it. `json.dumps` turns a NUL into the six ASCII characters `\u0000`; serializing the event payload for the column escapes that backslash in turn, so Postgres sees literal text rather than an escape it must interpret. Rejection needs a *live* `\u0000` in the wire JSON, and only the raw `errors[]` path produces one — the third row is what that would take.

I suspect the repro is what led there: in Python `"boom \^@ more"` is a literal backslash-caret-at rather than a NUL, since `\^` is not an escape sequence — `\^@` is how psql renders a NUL, not how Python writes one. A real `\x00` comes out as `\u0000`. Every state in these triggers goes through `to_json()` (`triggers/databricks.py:155`, `:181`, `:208`, `:320`, `:331`, `:424`, and `operators/databricks.py:280`), so `ClusterState` and `SQLStatementState` are covered by the same stringification.

One thing does remain, and it is not this: `from_json()` revives the real NUL, so if a revived `state_message` is ever put into a payload unstringified, that is the hole. Today it only reaches exception text and logs. Sanitizing inside `to_json()` would also change the string that `databricks_repair_reason_new_settings` matching reads (`operators/databricks.py:210`, `:226`) — an argument for keeping this at the persistence boundary, where the PR already has it.

So this looks right to me as it stands, and my approval holds.

---

fabbuc-gyg:
Is there anything blocking this from a merge? :-)

moomindani:
Nothing from my side — the approval stands, and the rebase I flagged in my merge-order note is not needed.

#72313 landed since then, and the two changes sit on different lines: it rewrote the loop header (`_latest_attempts(...)`), while this one rewrites the `failed_tasks.append(...)` payload. I merged current `main` into `f1bb073570` locally to be sure — no conflict, and the combined `utils/test_databricks.py` is 24 passed, with both the attempt-dedup tests and the sanitization tests green.

So it is waiting on a committer to merge.

---

PR #70390, opened 2026-07-24
UI: Hide "run with latest bundle version" clear option for non-versioned bundles

### LINKED ISSUE #70371: Hide "Run with latest bundle version" checkbox for non-versioned bundles when clearing
### Description

Apache Airflow version

3.2.2

When clearing a past Dag Run, a "Run with latest bundle version" checkbox is shown.

Image

Its current visibility condition is whether the Dag version numbers differ(latest_dag_version.version_number != the run's latest dag_version):

**The problem is that this checkbox also appears for bundles that do not support versioning, such as LocalDagBundle. In that case, toggling it has no effect on the executed result.**

Because dag_run.bundle_version is always None for LocalDagBundle, the serialized Dag resolved at run time is always the latest, regardless of the checkbox.
The checkbox appears to offer an "old version vs latest version" choice, but for LocalDagBundle there is no pinned old version to choose from.

```
# airflow-core/src/airflow/models/dagbag.py
def _version_from_dag_run(dag_run, *, session):
    if not dag_run.bundle_version:      # always True for LocalDagBundle
        if dag_version := DagVersion.get_latest_version(dag_id=dag_run.dag_id, session=session):
            return dag_version.id        # → always falls back to the latest
    return dag_run.created_dag_version_id
```

**Proposal: Change the visibility condition from "Dag version numbers differ" to "the run is actually pinned to a specific version"**

- Versioning-capable bundles (e.g. GitDagBundle) -> show the checkbox
- Non-versioning bundles(e.g. LocalDagBundle) -> hide the checkbox

### Use case/motivation

- LocalDagBundle users see the checkbox and assume they can "re-run with the old code," but runs always execute the latest code. Presenting a non-existent choice erodes trust in the UI.
- The word "bundle version" is itself meaningless for LocalDagBundle (supports_versioning = False, get_current_version() -> None), adding to the confusion.
- Gating visibility on actual pinning makes the checkbox appear only where it has a real effect(versioning-capable bundles), making the UI honest.

### Related issues

_No response_

### Code of Conduct

MannXo:
When clearing a Dag run, the "run with latest bundle version" checkbox was shown whenever the run's Dag version differed from the latest. For bundles that do not support versioning (e.g. LocalDagBundle) the run's `bundle_version` is always null and the serialized Dag resolved at run time is always the latest, so the checkbox offered a choice that had no effect.

The checkbox now also requires the run to be pinned to a specific bundle version, so it appears only for versioning-capable bundles (e.g. GitDagBundle) where toggling it actually changes the outcome. The visibility logic is extracted into a small pure helper with unit tests, mirroring the existing `TaskInstance/runOnLatestVersion.ts`.

closes: #70371

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
+++ b/airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
@@ -24,6 +24,7 @@ import { CgRedo } from "react-icons/cg";
 import { useDagServiceGetDagDetails } from "openapi/queries";
 import type { DAGRunResponse } from "openapi/requests/types.gen";
 import { ActionAccordion } from "src/components/ActionAccordion";
+import { shouldShowRunOnLatestVersionForRun } from "src/components/Clear/Run/runOnLatestVersion";
 import { useRerunWithLatestVersion } from "src/components/Clear/useRerunWithLatestVersion";
 import { Checkbox, Dialog } from "src/components/ui";
 import SegmentedControl from "src/components/ui/SegmentedControl";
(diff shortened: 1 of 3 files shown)


bbovenzi:
Duplicate of #70427. As that one is out of draft, we'll close this one.

PR #70345, opened 2026-07-23
Read HITLOperator subject/body into the summary lazily

1fanwang:
`subject` and `body` are template fields, rendered after `__init__` runs. `hitl_summary` was built in the constructor from `self.subject`/`self.body`, so it captured the **un-rendered** Jinja. OpenLineage reads `hitl_summary` when it builds the task START event, so a templated `subject`/`body` was emitted as raw `{{ ... }}` instead of the resolved value.

Make `hitl_summary` a property that reads `subject`/`body` at access time — after rendering — so listeners and the OpenLineage START event see the resolved values. Runtime and subclass additions (`timeout_datetime`, `approved`, `branches_to_execute`, …) go through a private `_hitl_summary_extra` dict that the property merges in; the summary keys are otherwise unchanged.

related: #70296

<details><summary>Testing Done</summary>

New `test_summary_reflects_rendered_subject_body` renders a templated `subject`/`body` and asserts the summary exposes the resolved values. RED on the pre-fix source, GREEN after the property:

```
# pre-fix: subject/body snapshotted un-rendered in __init__
assert op.hitl_summary["subject"] == "Review for 2020-01-01"
E   AssertionError: assert equals failed
E      -'Review for {{ ds }}'     +'Review for 2020-01-01'

# after: full suite
81 passed, 1 skipped
```

The `example_openlineage_hitl_dag` system DAG (all four HITL variants) drives the START/COMPLETE events end-to-end in CI and checks each event's `hitl_summary` carries the rendered `subject`. `validate_operators_init.py` on the operator exits 0.

</details>

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (please specify the tool below)

Generated-by: GitHub Copilot CLI following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70391, opened 2026-07-24
Make the HITLOperator summary extension point public

shahar1:
related: #70345

## Human Summary
I had some concerns regarding `_hitl_summary_extra` being a private dictionary after I merged the above.
As the docs state that subclasses can extend HITLOperator, it should function better as a public attribute.

## AI Summary

#70345 turned `hitl_summary` into a computed property (so templated `subject`/`body` are read after rendering) and moved runtime/subclass additions into a private `_hitl_summary_extra` dict. That left `HITLOperator` subclasses without a supported way to extend the summary — the pattern the old attribute explicitly documented ("subclasses can extend it") now silently mutates a throwaway snapshot — and it leaks the private attribute name into OpenLineage `include_full_task_info` events (which serialize the operator's `__dict__`, where a property never appears).

This renames the dict to a public `hitl_summary_extra` and documents the contract on the property: each access builds a fresh snapshot, mutation of the returned dict has no effect, and extensions go through `hitl_summary_extra`. Since #70345 has not shipped in a provider release yet, the rename is free of compatibility cost — waiting until after the next provider wave would turn it into a real break.

---

### DIFF
--- a/providers/standard/src/airflow/providers/standard/operators/hitl.py
+++ b/providers/standard/src/airflow/providers/standard/operators/hitl.py
@@ -133,7 +133,7 @@ class HITLOperator(BaseOperator):
         self.validate_defaults()
 
         # Runtime/subclass additions to the summary; config-derived entries live in the property.
-        self._hitl_summary_extra: dict[str, Any] = {}
+        self.hitl_summary_extra: dict[str, Any] = {}
 
     @property
     def hitl_summary(self) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Agreed with the concern that prompted this. Once `hitl_summary` became a property, each access builds a fresh dict, so the documented "subclasses can extend it" pattern silently stopped working — anything written to the returned dict lands in a throwaway snapshot. Leaving the only working route behind a private name wasn't a real extension point.

The docstring addition is the part I value most here: writing down that the snapshot is rebuilt per access is what stops the next person rediscovering this the hard way.

Checked that the rename is complete — all ten internal sites, and nothing outside `hitl.py` referenced the private name. No compat exposure either, since `_hitl_summary_extra` was private and hasn't appeared in a released changelog entry.

One thought for later, not for this PR: a public mutable dict means subclasses have to call `super().__init__()` before touching it or they get an `AttributeError`. Both current subclasses do. An overridable hook would be order-independent, but the dict matches how the internals already use it and is the smaller change.

---

PR #70394, opened 2026-07-24
Support setting an S3 object ACL for remote task logs

### LINKED ISSUE #8917: Support specify s3 object ACL args for remote logging
**Description**
Support specify s3 object ACL args for remote logging, like 'bucket-owner-full-control'

**Use case / motivation**
We are deploying an airflow server in the client's AWS server, using EC2 instance role(cross account) send logging to our S3 bucket.  We want to specify ACL arg "bucket-owner-full-control" to grant owner permission on logs.

**Related Issues**

FrankYang0529:
### Why

- When Airflow runs under one AWS account but writes logs to a bucket owned by another account, S3 (ACL-enabled) makes the writing account the object owner, so the bucket owner cannot read or manage the logs.
- Setting the `bucket-owner-full-control` ACL on upload is the standard remedy, but the S3 remote log handler previously had no way to set any ACL.

### How

- Add an `acl_policy` field to `S3RemoteLogIO`; when set, `write()` passes it as the `ACL` argument to the underlying `put_object` call.

### Verification

- Real two-account AWS e2e: A bucket under account A set to Bucket owner preferred. A task run under account B writing logs to account A's bucket:
  - Round 1 — no ACL: account A (bucket owner) gets `AccessDenied` reading the uploaded log.
  - Round 2 — `s3_task_handler_acl_policy = bucket-owner-full-control`: account A can read the log, and object ownership transfers to the bucket owner.

 

* closes: #8917
---

- [X] Yes - Claude Code

---

### DIFF
--- a/providers/amazon/docs/logging/s3-task-handler.rst
+++ b/providers/amazon/docs/logging/s3-task-handler.rst
@@ -43,6 +43,26 @@ To enable this feature, ``airflow.cfg`` must be configured as follows:
 
 In the above example, Airflow will try to use ``S3Hook(aws_conn_id='my_s3_conn')``.
 
+Setting an object ACL for cross-account logging
+'''''''''''''''''''''''''''''''''''''''''''''''
+
+When Airflow runs under one AWS account but writes logs to a bucket owned by a different
+account, S3 makes the writing account the owner of each uploaded log object. As a result the
+bucket owner cannot read or manage the logs. Applying the ``bucket-owner-full-control`` ACL
+on upload grants the bucket owner full control over the log objects.
+
+Set the ACL with the ``[aws] s3_task_handler_acl_policy`` option:
+
+.. code-block:: ini
+
+    [aws]
+    # ACL applied to every task log object uploaded to S3, e.g. for cross-account buckets.
+    s3_task_handler_acl_policy = bucket-owner-full-control
+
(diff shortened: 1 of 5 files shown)


potiuk (approved):
Useful capability, and the verification in the description is the strongest I've seen in this queue — a real two-account setup showing `AccessDenied` for the bucket owner without the ACL and success with it. That's the actual failure demonstrated rather than asserted.

The wiring is thorough: config, `remote_task_handler_kwargs`, and the `S3TaskHandler` kwarg all reach `acl_policy`, with kwarg-over-config precedence covered. `test_write_with_acl_policy` is the one that matters most, since it asserts `put_object` actually received `ACL=...` and that the body still round-trips — proving the value reaches boto3 rather than just landing on the object.

Collapsing an empty config value to `None` with `or None`, and only setting `extra_args["ACL"]` when truthy, means the unset path is genuinely unchanged.

One correction I pushed rather than sending back: `version_added` said `9.33.0`, but that version is already released — both `providers-amazon/9.33.0rc1` and the final `providers-amazon/9.33.0` tags exist — so it cannot contain this option. Changed to `9.34.0` in `provider.yaml` and the generated `get_provider_info.py`, and confirmed the two stay consistent via the `update-providers-dependencies` hook.

---

PR #69161, opened 2026-06-30
Preserve Dataform workflow invocation config

deepujain:
**Title:** Preserve Dataform workflow invocation config

## Summary

`DataformHook.create_workflow_invocation` now normalizes dict-based workflow invocation payloads into a `WorkflowInvocation` proto before passing the request to the generated Dataform client.

This keeps the public operator input unchanged while making nested proto fields explicit. In particular, `invocation_config.fully_refresh_incremental_tables_enabled` is now covered by a regression test so full-refresh workflow invocation configs are not dropped at the hook boundary.

## Changes

- **`providers/google/src/airflow/providers/google/cloud/hooks/dataform.py`** -- coerce dict workflow invocation payloads into the generated `WorkflowInvocation` proto before calling the Dataform client.
- **`providers/google/tests/unit/google/cloud/hooks/test_dataform.py`** -- add a regression that passes the issue's dict-shaped `invocation_config` and verifies `fully_refresh_incremental_tables_enabled=True` reaches the client request.

## Evidence it works

- The regression follows the issue's reported operator input shape: a dict workflow invocation with nested `invocation_config`.
- The test asserts the generated client receives a proto request with the nested full-refresh flag preserved.

## Test plan

- [x] `uvx --from uv==0.11.21 uv run --project providers/google ruff format --check providers/google/src/airflow/providers/google/cloud/hooks/dataform.py providers/google/tests/unit/google/cloud/hooks/test_dataform.py`
- [x] `uvx --from uv==0.11.21 uv run --project providers/google ruff check providers/google/src/airflow/providers/google/cloud/hooks/dataform.py providers/google/tests/unit/google/cloud/hooks/test_dataform.py`
- [x] `AIRFLOW_HOME=/private/tmp/airflow-test-53843 uvx --from uv==0.11.21 uv run --project providers/google pytest providers/google/tests/unit/google/cloud/hooks/test_dataform.py::TestDataformHook::test_create_workflow_invocation providers/google/tests/unit/google/cloud/hooks/test_dataform.py::TestDataformHook::test_create_workflow_invocation_preserves_invocation_config_from_dict -xvs --with-db-init`
- [x] CI passes

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

Fixes #53843

Issue #53843, opened 2025-07-29
Full Refresh arg in invocation_config does not work via DataformCreateWorkflowInvocationOperator


rsenseman:
I'm trying to run Dataform in "Full Refresh" mode with this Airflow Operator definition and it straight up doesn't work. It only runs the repo in non-full refresh mode. This leaves us with effectively no way to run our production DAG via "Full Refresh". I tried sorting through the code but I can't see where the arg is getting dropped. 

```
        create_workflow_invocation_full_refresh = DataformCreateWorkflowInvocationOperator(
            task_id="create_workflow_invocation_full_refresh",
            project_id=PROJECT_ID,
            region=DATAFORM_REGION,
            repository_id=REPOSITORY_ID,
            
            workflow_invocation={
                "compilation_result": "{{ task_instance.xcom_pull('create_compilation_result')['name'] }}",
                "invocation_config": {
                    "fully_refresh_incremental_tables_enabled": True,
                }
            }
        )
```

Class Definition: https://github.com/apache/airflow/blob/34ed71e52c1d8356194d34cb5018ff4032d66e2f/providers/google/src/airflow/providers/google/cloud/operators/dataform.py#L187

shahar1:
I'm not entirely sure that #69161 actually fixed this issue - might have been a Dataform SDK issue, and in any case more details are needed to actually reproduce it.

PR #70395, opened 2026-07-24
Revert Dataform workflow invocation dict normalization (#69161)

shahar1:
Reverts #69161. The change is a provably wire-identical no-op: proto-plus already converts nested dicts recursively, so the pre- and post-change `CreateWorkflowInvocationRequest` serialize byte-identically and the dict path never dropped `invocation_config`. Unknown fields raise `ValueError` in proto-plus, so a silent drop at this layer was never possible.

The symptom reported in #53843 was caused outside Airflow: dataform-core 3.0.10–3.0.16 defaulted incremental tables to `protected: true` (dataform-co/dataform#1914), which silently disables full refresh server-side; fixed upstream in core 3.0.17 (dataform-co/dataform#1942). Repos pin `dataformCoreVersion`, which is why the symptom persisted after the upstream fix. #53843 has been closed as not planned accordingly.

The accompanying test only pinned third-party proto-plus conversion behavior, which the project's testing standards exclude.

related: #53843

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/hooks/dataform.py
+++ b/providers/google/src/airflow/providers/google/cloud/hooks/dataform.py
@@ -187,8 +187,6 @@ class DataformHook(GoogleBaseHook):
         """
         client = self.get_dataform_client()
         parent = f"projects/{project_id}/locations/{region}/repositories/{repository_id}"
-        if isinstance(workflow_invocation, dict):
-            workflow_invocation = WorkflowInvocation(workflow_invocation)
         return client.create_workflow_invocation(
             request={"parent": parent, "workflow_invocation": workflow_invocation},
             retry=retry,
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Verified the load-bearing claim rather than taking it on trust — building the request both ways and comparing:

```
wire-identical:                             True
invocation_config preserved via dict path:  True
unknown field raises ValueError:            Unknown field for WorkflowInvocation: bogus_field
```

So the removed normalization was a genuine no-op: proto-plus already converts nested dicts recursively when the client builds `CreateWorkflowInvocationRequest`, `invocation_config` was never dropped on the dict path, and a silent drop was impossible since unknown fields raise.

The test removal is the more valuable half. Asserting that proto-plus converts a nested dict is pinning third-party behaviour, which our testing standards exclude, and it gave false confidence that we were fixing something we weren't.

Clean revert too — `WorkflowInvocation` is still imported and used for the `State` enum in both files, so nothing dangles.

I didn't independently check the upstream dataform-core attribution, but the revert doesn't rest on it.

---

Issue #70355, opened 2026-07-24
Status of testing Providers that were prepared on July 22, 2026


shahar1:
I have a kind request for all the contributors to the latest provider distributions release.
Could you please help us to test the RC versions of the providers?

The guidelines on how to test providers can be found in

[Verify providers by contributors](https://github.com/apache/airflow/blob/main/dev/README_RELEASE_PROVIDERS.md#verify-the-release-candidate-by-contributors)

Let us know in the comments, whether the issue is addressed.

These are providers that require testing as there were some substantial changes introduced:

## Provider [airbyte: 6.0.0rc1](https://pypi.org/project/apache-airflow-providers-airbyte/6.0.0rc1)
   - [ ] [Migrate Airbyte provider to `airbyte-api` library 1.0 and `httpx` (#68882)](https://github.com/apache/airflow/pull/68882): @aaronsteers
## Provider [akeyless: 0.3.0rc1](https://pypi.org/project/apache-airflow-providers-akeyless/0.3.0rc1)
   - [x] [Add cloud-based auth (aws_iam, gcp, azure_ad) to Akeyless secrets backend (#69772)](https://github.com/apache/airflow/pull/69772): @baraka-akeyless
## Provider [amazon: 9.33.0rc1](https://pypi.org/project/apache-airflow-providers-amazon/9.33.0rc1)
   - [ ] [Add OpenLineage parent info to EMR Spark steps (#70182)](https://github.com/apache/airflow/pull/70182): @mobuchowski
     Linked issues:
       - [Linked Issue #66816](https://github.com/apache/airflow/pull/66816): @rahul-madaan
       - [Linked Issue #64513](https://github.com/apache/airflow/pull/64513): @rahul-madaan
       - [Linked Issue #47508](https://github.com/apache/airflow/pull/47508): @mobuchowski
       - [Linked Issue #64807](https://github.com/apache/airflow/pull/64807): @rahul-madaan
   - [x] [Add S3RemoteLogIO.from_config and register s3 remote logging scheme (#69817)](https://github.com/apache/airflow/pull/69817): @jason810496
     Linked issues:
       - [Linked Issue #67056](https://github.com/apache/airflow/pull/67056): @jason810496
   - [x] [Add CloudWatchRemoteLogIO.from_config and register cloudwatch scheme (#69816)](https://github.com/apache/airflow/pull/69816): @jason810496
     Linked issues:
[...]

Issue #222, opened 2015-08-05
Add Kerberos support


garthcn:
It'd be great if Airflow could add Kerberos support to hooks (e.g. HDFS, etc). The `snakebite` lib Airflow depends on just added Kerberos support in this PR: https://github.com/spotify/snakebite/pull/153

mistercrunch:
I think that an `impersonate` flag in BaseOperator could be a decent generic way to help around security and attribution of "who is running what", even if Airflow is triggering the job.

The way `impersonate` would work is that the job on the worker would `sudo su {task.owner}` prior to running it. This assumes unix based roles and security to be deployed on workers. Of course many systems (take mysql for instance) don't support unix-type authentication.

Impersonate could be forced to `True` for given operators using the undocumented `policy` feature described in `settings.py`

txominpelu:
Hello @garthcn and @mistercrunch , I would be really interested in having this feature in airflow. Right now the lack of support for kerberos is the main thing that is blocking the adoption of airflow in my project. 

I am willing to help in the implementation. Has anybody already started working on it?

mistercrunch:
Can you describe which part(s) of Airflow you are thinking about integrating with Kerberos? Specific operators? The whole platform!?

txominpelu:
I'd like to modify some operators (e.g Hive) so that they can be scheduled and run with the right user in a hadoop cluster in secure mode .

pharoson:
#229 is a similar issue. It includes the current workaround I'm using to run workflows as a different user.

mistercrunch:
@txominpelu , I'm not sure if you read the impersonation section in here:
http://www.cloudera.com/content/cloudera/en/documentation/cdh4/v4-3-0/CDH4-Security-Guide/cdh4sg_topic_9_1.html

You could have a model where the user on their dev box need to provide their kerberos creds, and in production the server has superuser creds and impersonates the owner of the task.

bolkedebruin:
I did the work in snakebite. I am also interested in having Kerberos in. And although I need to wrap my mind around how airflow works and how it could be integrated I am
Willing to spend some time on it. I would think of something like:

Airflow has its own principal and keytab. This allows it to renew tickets on behalf of users, so it allows users to use their own credentials for "submitting". I'm not sure yet though how the submitting part would work. Or in other words how to make sure scripts executed under the users credentials are secured
And someone else cannot impersonate the user. Any ideas?

bolkedebruin:
[...]

PR #111, opened 2015-07-08
Use a bare raise so the original exception gets propagated.

jbalogh:
With this change I get full stack traces in the logs. Without it the exception appears to originate from the line with `raise e`.

PR #70396, opened 2026-07-24
Detect linked issues at the end of PR bodies in provider testing issue

### LINKED ISSUE #53843: Full Refresh arg in invocation_config does not work via DataformCreateWorkflowInvocationOperator

shahar1:
The provider testing issue generator (`breeze release-management generate-issue-content-providers`) silently missed linked issues whose reference is the last thing in the PR body — a common shape, since `Fixes #NNNNN` often sits on the final line. `ISSUE_MATCH_IN_BODY` (`r" #([0-9]+)[^0-9]"`) requires a non-digit character *after* the number, which doesn't exist at end of string after the body's lines are joined.

Concrete occurrence: in the google 22.3.0rc1 testing issue (#70355), PR #69161 (body ending with `Fixes #53843`) was rendered with no linked issues, and the release manager had to add #53843 by hand.

Replacing the trailing `[^0-9]` with a negative lookahead `(?![0-9])` matches the same references without requiring a trailing character. Verified against the real #69161 body: the old pattern finds nothing, the new one finds 53843; mid-body references behave identically.

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
@@ -256,7 +256,7 @@ MY_DIR_PATH = os.path.dirname(__file__)
 SOURCE_DIR_PATH = str(AIRFLOW_ROOT_PATH)
 PR_PATTERN = re.compile(r".*\(#([0-9]+)\)")
 PR_REFERENCE_PATTERN = re.compile(r"#([0-9]+)")
-ISSUE_MATCH_IN_BODY = re.compile(r" #([0-9]+)[^0-9]")
+ISSUE_MATCH_IN_BODY = re.compile(r" #([0-9]+)(?![0-9])")
 # Release-management commits (provider documentation / release preparation) are pure
 # release-process noise: they are not user-facing changes and existing providers already
 # exclude them (the release tooling parks them in the changelog's excluded section). Match
--- a/dev/breeze/tests/test_release_management_commands.py
+++ b/dev/breeze/tests/test_release_management_commands.py
@@ -23,6 +23,7 @@ import pytest
 
 from airflow_breeze.commands import release_management_commands
 from airflow_breeze.commands.release_management_commands import (
+    ISSUE_MATCH_IN_BODY,
     _ensure_default_python_for_reproducible_client,
     _is_initial_provider_release,
     _should_include_provider_in_issue,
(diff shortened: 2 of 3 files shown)


dev/breeze/tests/test_release_management_commands.py
```diff
@@ -292,3 +293,18 @@ def test_get_package_version_possibly_from_stable_txt_for_java_sdk(
...
     assert get_package_version_possibly_from_stable_txt("java-sdk") == expected_version
+
+
+@pytest.mark.parametrize(
+    ("body", "expected"),
+    [
+        ("Some description closes: #12345 and more text", [12345]),
+        # reference at the very end of the body (e.g. "Fixes #53843" as the last line)
+        ("Generated-by: some agent Fixes #53843", [53843]),
+        ("related: #111, also see #222", [111, 222]),
```
potiuk, line 304: nit, optional — the fix also cures adjacent references, since the old `[^0-9]` consumed the separator that the next match needed. A case pinning that behaviour down:

```suggestion
        ("related: #111, also see #222", [111, 222]),
        ("adjacent references see #111 #222", [111, 222]),
```

dev/breeze/src/airflow_breeze/commands/release_management_commands.py
```diff
@@ -256,7 +256,7 @@
 SOURCE_DIR_PATH = str(AIRFLOW_ROOT_PATH)
 PR_PATTERN = re.compile(r".*\(#([0-9]+)\)")
 PR_REFERENCE_PATTERN = re.compile(r"#([0-9]+)")
-ISSUE_MATCH_IN_BODY = re.compile(r" #([0-9]+)[^0-9]")
+ISSUE_MATCH_IN_BODY = re.compile(r" #([0-9]+)(?![0-9])")
```
potiuk, line 259: nit — `dev/assign_cherry_picked_prs_with_milestone.py:50` still carries a byte-identical copy of the *old* pattern. I checked: it's dead code there — defined but never referenced in that file — so nothing is actually broken. But it now diverges from this one, and it's the first hit anyone gets greping for `ISSUE_MATCH_IN_BODY`. Deleting it in this PR would save the next person the confusion.

potiuk (approved):
Good catch — the old `[^0-9]` needed a trailing character to exist, so a reference as the last thing in the body was silently dropped. Lookahead is the right fix. I traced both call sites (`release_management_commands.py:2925` and `:4749`), and the parametrized cases genuinely fail against the old pattern, so the test earns its place.

Worth noting the fix quietly cures a second bug the description doesn't mention: because `[^0-9]` *consumed* a character, adjacent references lost the separator the next match needed — on ` #111 #222` the old pattern matched `#111`, ate the space, and then missed `#222` entirely.

Two nits inline, neither blocking; approving.

---

> will follow up.*
>

---

PR #70413, opened 2026-07-24
UI: Add Select Recent Configurations dropdown to Trigger Dag form

### LINKED ISSUE #69820: Add in-form "Select Recent Configurations" dropdown to the Trigger DAG form (navigation QoL, coexisting with "Trigger again with this config")
### Description

Airflow 3.2.0 restored config reuse via the "Trigger again with this config"
menu item (#56254 / #56406). That flow requires the user to first navigate to
the DAG run list/grid, visually find the right past run, select it, and only
then trigger from it.

This request is to ALSO add back the Airflow 2 "Select Recent Configurations"
dropdown *inside the trigger form itself*. Picking an entry pre-fills the
Configuration JSON without ever leaving the form.

The point of this request is the navigation quality-of-life, not config reuse in
the abstract. The two features are complementary:
- "Trigger again with this config": for when you already know and can find the
  exact run you want to copy.
- In-form dropdown: for when you just want to grab a recent config while
  composing a run, without breaking flow to go hunt through the run list.

Proposal: keep the menu item as-is and add the dropdown to the trigger form.
Both write the same conf field, so there is no conflict.

### Use case/motivation

Many DAGs are triggered manually with configs that repeat or differ by only a
few parameters. The friction is navigation, not the reuse itself.

Current 3.2.0 flow to reuse a config:
  open DAG -> go to runs list/grid -> scan for the run with the config I want
  -> select it -> Trigger button -> "Trigger again with this config" -> adjust
  -> run.

When a DAG has run many times and runs look visually similar, finding "the run
that had the config I want" is slow and error-prone, and it pulls the user out
of the trigger flow entirely.

Airflow 2 flow (what this restores):
  open trigger form -> pick a recent config from the dropdown -> adjust -> run.

The in-form dropdown removes the detour into the run list. It surfaces recent
configs right where the user is already composing the run. This is the specific
quality-of-life that the menu item does not provide, since the menu item is
built around first locating and selecting a specific run.

Multiple users have asked for exactly this (Discussion #56038). The dropdown and
the menu item can coexist with minimal added surface area, since the dropdown
just populates the existing conf field.

### Related issues

- #56254  Trigger form missing "Select Recent Configurations" from airflow 2 (implemented as the menu item)
- #56406  PR that added "Trigger again with this config" (3.2.0); a dropdown was built here but removed in favor of the menu item
- #56038  Discussion: Missing dropdown "Select Recent Configurations" in DagRun UI Airflow v3 (users still asking for this)
- #27714  Original Airflow 2 feature: Re-use recent DagRun JSON-configurations
- #36878  Airflow 2 PR: configurable number of recent configs in the dropdown

### Code of Conduct

josix:
### Context 
According to the #69820's description, reusing a past run's config currently requires leaving the trigger flow to hunt for the right run in the run list and use "Trigger again with this config". In this PR, I'm going to restore the Airflow 2 in-form dropdown lets users pick any distinct config from the recent runs in a scrollable list while composing a run. The dropdown reuses the existing dagRuns list endpoint and the same conf-only prefill path as the menu item, so no new API surface is added.

closes: #69820

### Verification

image

image

 

---

- [x] Yes

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
+++ b/airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
@@ -1,0 +1,117 @@
+/*!
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { createListCollection, Flex, Select, type SelectValueChangeDetails, Text } from "@chakra-ui/react";
+import { useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
+
+import { useDagRunServiceGetDagRuns } from "openapi/queries";
+import type { DAGRunResponse } from "openapi/requests/types.gen";
+
+import Time from "../Time";
+
+type RecentConfigOption = {
+  run: DAGRunResponse;
+  value: string;
+};
+
+type RecentConfigSelectProps = {
+  readonly dagId: string;
+  readonly onSelectConf: (conf: Record<string, unknown>) => void;
+  readonly open: boolean;
+};
+
+const RecentConfigSelect = ({ dagId, onSelectConf, open }: RecentConfigSelectProps) => {
+  const { t: translate } = useTranslation("components");
+  const [selectedValue, setSelectedValue] = useState<Array<string>>([]);
+  const { data, isLoading } = useDagRunServiceGetDagRuns(
+    { dagId, limit: 25, orderBy: ["-run_after"] },
+    undefined,
+    { enabled: open },
+  );
+
+  const options = useMemo(() => {
+    const seenConfs = new Set<string>();
+
+    return (data?.dag_runs ?? []).reduce<Array<RecentConfigOption>>((items, run) => {
+      const hasConf = run.conf !== null && Object.keys(run.conf).length > 0;
+      const confKey = hasConf ? JSON.stringify(run.conf) : undefined;
+      const isNewConf = confKey !== undefined && !seenConfs.has(confKey);
+
+      if (isNewConf) {
+        seenConfs.add(confKey);
+        items.push({ run, value: run.dag_run_id });
+      }
+
+      return items;
+    }, []);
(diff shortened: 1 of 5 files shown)


airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+    items: options,
+    itemToString: (item: RecentConfigOption) => item.run.dag_run_id,
+  });
+
+  const handleValueChange = ({ items, value }: SelectValueChangeDetails<RecentConfigOption>) => {
+    const [selected] = items;
+
+    setSelectedValue(value);
+    if (selected?.run.conf) {
+      onSelectConf(selected.run.conf);
```
pierrejeambrun, line 75: This will reset user defined logicalDate, runId, and note to defaults.
But here it fires interactively: a user who sets a logical date / custom run ID / manual data interval / note and then picks a recent config to reuse its params loses all of those to defaults. The dropdown is meant to apply a past run's conf, not reset the run's scheduling fields.

Either fix that or move this config selector at the top of the form, so we know that all fields are affected and will be reset to

airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+      disabled={isLoading || options.length === 0}
+      onValueChange={handleValueChange}
+      size="sm"
+      value={selectedValue}
+    >
+      <Select.Label fontSize="xs">{translate("triggerDag.recentConfig")}</Select.Label>
+      <Select.Control>
+        <Select.Trigger>
+          <Select.ValueText placeholder={translate("triggerDag.recentConfigPlaceholder")} />
+        </Select.Trigger>
```
pierrejeambrun, line 96: When we run through 'trigger again with this conf button', this drop down probably shouldn't be there at all. 

It's confusing to have it (with nothing selected), but actually a conf is passed down. And if it's to change the run conf to another run, there's no point running through that 'trigger again with this conf' button.

pierrejeambrun (commented):
Thanks a lot, 

A couple suggestions before this can move forward :)

airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
josix: Thanks for catching this, I've fixed the issue. Selecting a recent configuration now only swaps `conf` via `reset((prev) => ({ ...prev, conf }))` and leaves the logical date, run ID, note, and data interval as the user set them. The prefill path still performs a full reset.
Here is the recording for this fix: https://jam.dev/c/02f5eba8-bc69-440d-975e-ca233b620f9b

airflow-core/src/airflow/ui/src/components/TriggerDag/RecentConfigSelect.tsx
josix: Yeah, thanks for pointing it out. The form already retains a DAG run's `conf` when using the “Trigger Again with This Config” button, so I think we don't need the dropdown in this case. I've removed the selector from the form for this flow.

josix:
Hello @pierrejeambrun, thank you for the kind review. I've addressed the issues you pointed out. Please take a look when you have chance Thank you so much!

PR #70416, opened 2026-07-24
Feature expand health endpoint reporting

JH-A-Kim:
---

# Description
These changes act as an implementation for the suggested changes in the design proposal:

[Design propsoal](https://docs.google.com/document/d/1yWO-0YL6TNpABLwXoKy3YGVFZGnr4CE0Q1iZ_rpNmBo/edit?tab=t.0)

and from the Discussion from the following thread and lazy consensus on the dev list:

[DISCUSS](https://lists.apache.org/thread/l1b1wkxs69omjsb8cwdryr2oydjmn3yg)
[LAZY CONSENSUS](https://lists.apache.org/thread/9t7k3vnkw7n9chxrp53wvzy363jcnwom)

These changes were co authored by @JH-A-Kim and @Tegh25 

The changes are the following:

- extended monitor.py to support instances and detailed status reporting in the JSON output for the health endpoint
- added helper functions to get information from the metadata database for running components, aggregate statuses from a list of component instances, and a legacy health reporter.
- extended get_airlfow_health to include lists of components and report their aggregated status.
- added aggregated status reporting in a three tier model for detailed_status, healthy > degraded > down. These are all instances are healthy, at least one instance is healthy and at least one is unhealthy, and then down is where all instances are down. 

Side Note: There was a slight change from using `unhealthy` to `down` in `detailed_status`for clarity.

Generated-by: [Sonnet 5] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


ferruzzi:
Could you include a screenshot of the rendered health page after these changes?

JH-A-Kim:
Screenshot 2026-08-05 at 8 59 12 AM

airflow-core/src/airflow/jobs/job.py
```diff
@@ -101,6 +101,10 @@ class Job(Base, LoggingMixin):
     executor_class: Mapped[str | None] = mapped_column(String(500))
     hostname: Mapped[str | None] = mapped_column(String(500))
     unixname: Mapped[str | None] = mapped_column(String(1000))
+    team_name: Mapped[str | None] = mapped_column(
+        String(50), ForeignKey("team.name", ondelete="SET NULL"), nullable=True
+    )
+    bundle_names: Mapped[list[str] | None] = mapped_column(ExtendedJSON, nullable=True)
```
vincbeck, line 107: I dont understand this.
1. Why do we need `bundle_names`? 
2. If we need it, why having a list of bundle names? A Dag processor can be associated to only one Dag bundle, am I wrong?

Tegh25: `bundle_names` is the Dag-processor counterpart to `team_name` on triggerers. Both are scope metadata on the job row so health / job APIs can show what subset of work a given instance owns, not only that “some” instance is alive.

A Dag processor is not limited to a single bundle. `airflow dag-processor --bundle-name` is appendable, and `DagFileProcessorManager.bundle_names_to_parse` is a `list[str] | None`. So the column is a list to match that existing CLI/manager contract, e.g. one processor for bundle-a + bundle-b, or several processors each with their own subset in multi-team setups.

There are some test cases that also indicate that Dag processors can have multiple bundles: [test_manager.py, lines 2783-2799](https://github.com/apache/airflow/blob/main/airflow-core/tests/unit/dag_processing/test_manager.py/#L2783-L2799)

Happy to rename/clarify in docs if “list” doesn't feel intuitive. the shape follows the existing `--bundle-name` / `bundle_names_to_parse` behavior rather than introducing a new one-bundle-per-processor constraint.

vincbeck: What do we use `bundle_names` for? It is to fetch the associated team names later? If so, we should then use `team_name` only

airflow-core/src/airflow/api/common/airflow_health.py
```diff
@@ -16,77 +16,165 @@
...
+
+@provide_session
+def get_jobs_health(job_runner_class, *, session: Session = NEW_SESSION) -> list[Job]:
+    """Return all running jobs for the runner class, ordered by latest heartbeat."""
+    return list(
+        session.scalars(
+            select(Job)
+            .where(
+                Job.job_type == job_runner_class.job_type,
+                Job.state == JobState.RUNNING,
```
ferruzzi: It looks like there may be a bug with this which we will need to fix separately, but if you look at `Trigger.assign_unassigned` in `models/trigger.py:402-406`, they use `end_date IS NULL` and the time since the last heartbeat to determine if it's "alive" and they specifically don't use `state`.  We should do similar here, but it may not be so easy... Without some kind of upper time range, once a triggerer goes "DEGRADED" there is no way for that failed job to fall off the end of the list so the triggerer will never recover unless that job row is removed from the db, which is also wrong....

You would then also have to modify the same concept in the dag-processor around L152 and triggerer around L135 for the same reason, and some of your new tests may need to be adjusted accordingly.

The issue is that if the job quits unexpectedly (process killed, for example), we don't have a job reaper and that will stay RUNNING for the rest of time.  We should likely add that reaper, but that's definitely out of scope for this project and the proposed solution is already in use elsewhere so it should be fine here too I think.  I need to think the reaper idea through a bit, but I'll cut an Issue for it when I have a better idea of the actual issue there.

airflow-core/tests/unit/api/common/test_airflow_health.py
```diff
@@ -16,127 +16,246 @@
...
-@patch("airflow.api.common.airflow_health.TriggererJobRunner.most_recent_job", return_value=None)
-@patch("airflow.api.common.airflow_health.DagProcessorJobRunner.most_recent_job", return_value=None)
-def test_get_airflow_health_only_metadatabase_healthy(
-    latest_scheduler_job_mock,
-    latest_triggerer_job_mock,
-    latest_dag_processor_job_mock,
-):
+@patch("airflow.api.common.airflow_health.get_jobs_health")
+def test_get_airflow_health_only_metadatabase_healthy(mock_get_jobs_health):
+    mock_get_jobs_health.side_effect = [[], [], []]
```
ferruzzi, line 130: Every test in here patches `get_jobs_health`, so the filter inside it never runs against real job rows in the same test as the legacy status derivation.  I think that's why the suite passes despite the change I flagged over in `airflow_health.py`.

If you add a DB-backed case where a triggerer has one row which is non-RUNNING and assert `status == "unhealthy"`, I bet it would fail because the empty result from `get_jobs_health` falls into the `else: triggerer_status = None` at L135.  On main that same DB state reports UNHEALTHY, since `most_recent_job()` ignores state and
`is_alive()` then returns False.

That's different from the killed-process case in my other comment.  A clean exit leaves the row non-RUNNING, so it disappears from the report; a `kill -9` leaves it RUNNING so it sticks around as a zombie.  But both of those failure cases come from the same assumption of using the raw state as the deciding factor, which is why I think they can both be fixed in one place.

airflow-core/src/airflow/jobs/job.py
Tegh25: `bundle_names` isn’t used to fetch `team_name` in this PR, it describes the Dag processor’s scope and is returned directly. Health returns those names so users can see which bundles an instance owns.

We can resolve bundle to team via `DagBundleModel.get_team_names` (the manager already does that for metrics), but a processor can own multiple bundles, bundles can be teamless/shared, and multi-team may be off, so a single `team_name` on the job can’t replace the list of `bundle_names`. Triggerers, for example, are team-scoped by design, but Dag processors are bundle-scoped.

Our original proposal may have been unclear with exactly what we planned to do with Dag processor `bundle_names`, I apologize for that. I believe returning `bundle_names` is the best approach here, but we can discuss `team_name` retrieval logic if you want to select just a single `team_name`.

airflow-core/src/airflow/api/common/airflow_health.py
Tegh25: I agree filtering `state == RUNNING` can cause issues for the reasons you described. We also should not reuse `_is_alive()`'s heartbeat threshold in `get_jobs_health()`, otherwise the query would only return alive jobs and stale instances would never appear.

*Proposed filter:*
`job_type`, `end_date IS NULL`, and `latest_heartbeat > now - retention` where `retention` is a multiple of `health_check_threshold`. Inside that window `is_alive()` still marks fresh vs stale, and it adds some kind of upper time range so degraded can recover without a reaper. 

What about the empty-list path at L135/L152? Clean exit of a job sets `end_date`, so the filtered list becomes empty and we currently report None instead of main's `unhealthy`.

airflow-core/tests/unit/api/common/test_airflow_health.py
Tegh25: Yep, I can definitely work on adding a DB-backed test that does not mock `get_jobs_health()` once we decide on an approach for job filtering.

airflow-core/src/airflow/jobs/job.py
vincbeck: Sure, thanks for the explanations

airflow-core/src/airflow/jobs/triggerer_job_runner.py
```diff
@@ -213,6 +213,8 @@ def __init__(
             raise ValueError(f"Capacity number {capacity!r} is invalid")
         self.queues = queues
         self.team_name = team_name
+        if job.team_name is None:
+            job.team_name = team_name
```
vincbeck: When can that happen? Also, I am wondering if having `self.team_name` is necessary since now we have it in `job`. WDYT?

Tegh25: Only one case I can think of, the `if job.team_name is None` branch covers callers that construct a bare `Job()` and only pass `team_name` into the runner, which is something some of the tests do.

Looking at it now, `self.team_name` is redundant if we treat `job.team_name` as the source of truth. If we want to keep the constructor arg, we can set `job.team_name = team_name` indefinitely. Or we drop the constructor arg and require callers to set `team_name` on Job.

Which resolution do you prefer, and would removing `self.team_name` be in the scope of this PR?

vincbeck: > Or we drop the constructor arg and require callers to set team_name on Job. Which resolution do you prefer, and would removing self.team_name be in the scope of this PR?

I would go with that in this PR yes

airflow-core/src/airflow/api/common/airflow_health.py
JH-A-Kim: Yeah I did notice recently while testing the endpoint that if i ended the service by killing the container, that if i restarted the container and looked at the endpoint their would be previous instances of the components still left over that were not cleaned up because their was nothing to reap them. It would list them as down but running.

ferruzzi: @JH-A-Kim Yup, that's the exact case.  It's not part of this PR, but there should be a reaper to clean those up.  Do you have time to open an Issue with the details so someone gets to it?

ferruzzi: @Tegh25  Good discussion, but I think section 5 of your SoW document already settles this and we don't need a new mechanism.

The acceptance table has this row:


> | 3 jobs, all stale | unhealthy | 3 unhealthy |


`latest_heartbeat > now - retention` can't satisfy that.  If all three are stale they're all outside the window, the query returns nothing, and we can't report "3 unhealthy".  The same row rules out the current `state == RUNNING` filter from the other side, since a cleanly stopped job leaves RUNNING and vanishes from the list.

So the filter should be `job_type` (plus `end_date IS NULL` if we want to drop cleanly-exited jobs) and nothing else.  No state, no time window.  Then `is_alive()` decides healthy vs unhealthy per instance, which is what the table describes, and the aggregate follows: all instances alive = healthy, some alive = degraded, none alive = down.

On your L135/L152 question: empty list must not map to None.  Today `most_recent_job()` has no filter at all beyond `job_type`, so `status: null` means "this component has never run" and nothing else.  Any component that has ever run reports healthy or unhealthy.  If we let an empty filtered list produce `null`, then a triggerer that's been dead for an hour reports `"status": null`, and anyone alerting on `status != "healthy"` silently stops getting paged exactly when it matters most.  Empty means DOWN, and DOWN means legacy status = "unhealthy".  Same for the scheduler, which is already correct since it has no `else: None` branch.

That also removes the need for a retention knob: the leftover rows from killed containers that @JH-A-Kim hit are a reaper's job, not the read path's.  That's tracked separately, and until it exists those rows (correctly) show as unhealthy instances rather than disappearing.

One doc fix while we're here: Phase 1 says "Query all RUNNING jobs per type (or recent jobs within a window)". That parenthetical is what led us here and it contradicts section 5.  Likely worth updating that.

With that decided you should be unblocked on the DB-backed test you offered, and the whole section 5 table is a good source for cases.

JH-A-Kim: > @JH-A-Kim Yup, that's the exact case. It's not part of this PR, but there should be a reaper to clean those up. Do you have time to open an Issue with the details so someone gets to it?

yeah I should in a week be able to work on setting up an issue for that. Working on some stuff for the GitHub team for the past week that needs to be done by the end of next week Friday but I should be able to finish my work before then and then I can get to making an issue!

airflow-core/tests/unit/api/common/test_airflow_health.py
Tegh25: @ferruzzi please see [f498018](https://github.com/apache/airflow/pull/70416/commits/f498018f46d4524838f18735804f0589a6ec34bc). I added the filtering logic and removed the empty list branches as we discussed, also updated the tests and included DB tests. I'll rebase once we're satisfied with the changes.

airflow-core/src/airflow/jobs/triggerer_job_runner.py
Tegh25: Please see [0f9ae91](https://github.com/apache/airflow/pull/70416/commits/0f9ae912a44ee37158557ac63923ed6f7463dcac), I removed the `team_name` arg as discussed and updated some of the tests.

airflow-core/tests/unit/api/common/test_airflow_health.py
ferruzzi: Yeah, I think that looks right.   One side note, the docs are now out of date, you mention `null` a handful of times; make a pass on those when you get a sec.

ferruzzi: I think it looks right.   Just make a pass on the docs and see which of those need updating, I see a few references to a `null` value which I don't think are accurate now with that change.

Tegh25: Should be good to go now. GitHub doesn't let me resolve this comment.

vincbeck (approved):
LGTM. Would love to have @bbovenzi and @pierrejeambrun opinions on this one

vincbeck:
Could you work on the conflicts please?

Tegh25:
> Could you work on the conflicts please?

Resolved 🙂

ferruzzi (approved):
Nice work, thanks to both of you for seeing this through to the end, it's been a long ride.

JH-A-Kim:
Thank you for being an incredible mentor! I'll see you both around the repo and slack (and eventually when I come back from my internship lol) (and also on my other pr that I haven't had time to get back to)

Issue #61452, opened 2026-02-04
Add support for Kubernetes Gateway API (HTTPRoute) in Helm Chart


prakharcode:
### Description

Currently, the official Apache Airflow Helm chart only supports the standard Kubernetes Ingress API. Many modern Kubernetes environments are migrating to the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/).

I would like the Helm chart to include optional templates for Gateway API resources, specifically:

HTTPRoute: Support for routing traffic to the Webserver, Flower, and API Server.

Configuration Values: New sections in values.yaml (e.g., gateway.enabled, gateway.web.hostnames, gateway.web.parentRefs) similar to the existing ingress configuration.

The Gateway API is now GA and is becoming the standard for networking in Kubernetes 1.24+.

### Related issues

The only related discussion I have found: https://github.com/apache/airflow/discussions/51967

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

abhijeets25012-tech:
I would like to contribute

prakharcode:
Hi @abhijeets25012-tech, I would like to take it up already, will start working on a PR soon, you're welcome to help with reviews and testing!

WhiteDiamondz:
Hey @prakharcode !

Have you had any chance to work on this ? I recently encountered the need for HTTPRoute support and could work on a merge request if necessary.

Thanks !

Miretpl:
In general, anyone can work on this issue, and if at the same time, we have more than one PR, we will choose the best one, so I encourage anyone to work on this. For more details, you can also check the https://github.com/apache/airflow/blob/main/contributing-docs/04_how_to_contribute.rst#contribute-code-changes doc, as the handling of the issues was changed not so long ago.

Just a slight note too - the Helm Chart itself is during the refurbish discussion, so there is a chance that this feature will be delayed until e.g. Helm Chart 2.0, which currently does not have a release date.

somaz94:
Working on this — draft PR opened: https://github.com/apache/airflow/pull/67675

Miretpl:
[...]

PR #68552, opened 2026-06-15
Improve API server HTTPRoute config based on review feedback

somaz94:
Follow-up to #67675 addressing post-merge review feedback from @Miretpl. The
original PR has not been released yet, so these refinements land before the
upcoming chart release as requested.

Changes:
- Move the HTTPRoute config from top-level `httpRoute.apiServer.*` to
  `apiServer.httpRoute.*` — it is an API-server setting, not a global one
  (values.yaml, values.schema.json, template, docs, tests).
- Fail template rendering with a clear message when `apiServer.httpRoute.enabled`
  is true but the Gateway API HTTPRoute CRD is not installed (via
  `.Capabilities.APIVersions.Has`).
- Fail when both `ingress.apiServer.enabled` and `apiServer.httpRoute.enabled`
  are set — HTTPRoute is an alternative to the API server Ingress.
- Strengthen tests: assert full `spec.parentRefs` / `spec.rules` structures, add
  asserts to the validation test, inline a single-use variable, and split the
  parametrized creation test into separate cases (no inner `if`).
- Add an `api_versions` parameter to the `render_chart` helper so
  Capabilities-gated templates render offline.

Validation:
- 48 apiserver helm unit tests pass (incl. new CRD-missing and ingress-conflict cases)
- `helm template` renders for enabled/disabled and fails as expected for both guards
- chart-quality (values↔schema) test passes; ruff clean

related: #67675

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70425, opened 2026-07-25
Add Webserver HTTPRoute support to the Helm chart 1.2x line

aaron-y-chen:
closes: #61452

## Summary

This adds Helm chart support for exposing the **Webserver** with Kubernetes Gateway API `HTTPRoute` on the 1.2x chart line (Airflow 2.x).

- adds `webserver.httpRoute` values, schema, template, tests, and docs;
- routes the default backend to the Webserver service on `ports.airflowUI`;
- supports custom labels, annotations, hostnames, `parentRefs`, path settings, and full `rules` override;
- renders only for Airflow 2 (`airflowVersion` < 3.0.0), where the Webserver exists — on Airflow 3 the API Server replaces it;
- fails fast when Webserver Ingress and Webserver HTTPRoute are both enabled (guarded from both the HTTPRoute and the Ingress template);
- extracts the shared `httpRoute` schema into `definitions.httpRoute` so `apiServer`, `flower`, and `webserver` reference one definition instead of three near-identical copies.

## Validation

E2E screenshot through the Gateway API path (Envoy Gateway, Airflow 2.11):

61452_webserver_httproute

The port-forward targets the **Envoy Gateway dataplane service** and the URL host `airflow.localhost` is the Gateway listener hostname, so the request path is:

```text
browser -> Envoy Gateway service -> Gateway listener -> Webserver HTTPRoute -> airflow-webserver service
```

---

Generated-by: [Claude Opus 4.8] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/chart/docs/production-guide.rst
+++ b/chart/docs/production-guide.rst
@@ -450,7 +450,6 @@ is configured in an Airflow 2 deployment:
 
 Flower HTTPRoute resources are only created when Flower itself is created, so ``flower.enabled`` must be
 ``true`` and the executor must include ``CeleryExecutor`` or ``CeleryKubernetesExecutor``.
-Webserver HTTPRoute resources are only created for Airflow 2 deployments (``airflowVersion`` < 3.0.0).
 
 For fine-grained routing, supply ``apiServer.httpRoute.rules``, ``flower.httpRoute.rules``, or
 ``webserver.httpRoute.rules`` directly — the entry mirrors the upstream ``HTTPRouteRule`` schema and
--- a/chart/docs/production-guide.rst
+++ b/chart/docs/production-guide.rst
@@ -399,7 +399,7 @@ Gateway API (HTTPRoute)
 ^^^^^^^^^^^^^^^^^^^^^^^
 
 As an alternative to ``Ingress``, the chart can create
-`Kubernetes Gateway API <https://gateway-api.sigs.k8s.io/>`_ ``HTTPRoute`` resources for the API server and Flower.
+`Kubernetes Gateway API <https://gateway-api.sigs.k8s.io/>`_ ``HTTPRoute`` resources for the API server, Flower, and the Webserver.
 This requires the Gateway API CRDs to be installed in the cluster and a ``Gateway`` to already exist —
 the chart only creates the ``HTTPRoute`` resources and attaches them to the Gateway via ``parentRefs``.
 
@@ -429,13 +429,35 @@ the chart only creates the ``HTTPRoute`` resources and attaches them to the Gate
        hostnames:
          - flower.example.com
 
+The Webserver exists only on Airflow 2 (the API server replaces it on Airflow 3), so its ``HTTPRoute``
+is configured in an Airflow 2 deployment:
+
+.. code-block:: yaml
+   :caption: values.yaml (Airflow 2)
+
+   defaultAirflowTag: "2.11.0"
+   airflowVersion: "2.11.0"
+
+   webserver:
+     httpRoute:
+       enabled: true
+       parentRefs:
+         - name: main-gateway
+           namespace: gateway-system
+           sectionName: https
+       hostnames:
+         - airflow.example.com
+
 Flower HTTPRoute resources are only created when Flower itself is created, so ``flower.enabled`` must be
 ``true`` and the executor must include ``CeleryExecutor`` or ``CeleryKubernetesExecutor``.
+Webserver HTTPRoute resources are only created for Airflow 2 deployments (``airflowVersion`` < 3.0.0).
 
-For fine-grained routing, supply ``apiServer.httpRoute.rules`` or ``flower.httpRoute.rules`` directly —
-the entry mirrors the upstream ``HTTPRouteRule`` schema and overrides the default rule generated from
-the corresponding ``apiServer.httpRoute.path`` and ``apiServer.httpRoute.pathType``, or
-``flower.httpRoute.path`` and ``flower.httpRoute.pathType`` values.
+For fine-grained routing, supply ``apiServer.httpRoute.rules``, ``flower.httpRoute.rules``, or
+``webserver.httpRoute.rules`` directly — the entry mirrors the upstream ``HTTPRouteRule`` schema and
+overrides the default rule generated from the corresponding
+``apiServer.httpRoute.path`` and ``apiServer.httpRoute.pathType``,
+``flower.httpRoute.path`` and ``flower.httpRoute.pathType``, or
+``webserver.httpRoute.path`` and ``webserver.httpRoute.pathType`` values.
 
 .. note::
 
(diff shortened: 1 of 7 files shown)


bujjibabukatta (approved):
changes looks good .

chart/docs/production-guide.rst
```diff
@@ -429,13 +429,35 @@ the chart only creates the ``HTTPRoute`` resources and attaches them to the Gate
...
+         - airflow.example.com
+
 Flower HTTPRoute resources are only created when Flower itself is created, so ``flower.enabled`` must be
 ``true`` and the executor must include ``CeleryExecutor`` or ``CeleryKubernetesExecutor``.
-
-For fine-grained routing, supply ``apiServer.httpRoute.rules`` or ``flower.httpRoute.rules`` directly —
-the entry mirrors the upstream ``HTTPRouteRule`` schema and overrides the default rule generated from
-the corresponding ``apiServer.httpRoute.path`` and ``apiServer.httpRoute.pathType``, or
-``flower.httpRoute.path`` and ``flower.httpRoute.pathType`` values.
+Webserver HTTPRoute resources are only created for Airflow 2 deployments (``airflowVersion`` < 3.0.0).
```
Miretpl:
```suggestion
```
It is already stated a couple of lines above, so it should not be needed here. Also, I think that below can be just added to the paragraph above.

chart/values.schema.json
```diff
@@ -12519,6 +12339,101 @@
         }
     },
     "definitions": {
+        "httpRoute": {
```
Miretpl, line 12367: Could you also do a similar change for the main (extraction of httpRoute def to definitions)?

chart/values.yaml
```diff
@@ -2529,6 +2530,52 @@ apiServer:
 webserver:
   enabled: true
 
+  # Kubernetes Gateway API (HTTPRoute) configuration for the Webserver.
+  # Only applies to Airflow 2 (`airflowVersion` < 3.0.0), where the Webserver exists.
```
```suggestion
```
We have this info couple lines above.

helm-tests/tests/helm_tests/airflow_aux/test_basic_helm_chart.py
```diff
@@ -384,35 +384,41 @@ def test_network_policies_are_valid(self, airflow_version, executor):
...
-                id="flower-ingress",
+                id="ingress",
             ),
             pytest.param(
                 {
-                    "ingress": {
-                        "web": {"enabled": True},
-                        "apiServer": {"enabled": True},
-                    },
+                    "ingress": {"apiServer": {"enabled": True}},
```
Miretpl, line 412: Why we are deleting `web` option?

helm-tests/tests/helm_tests/webserver/test_httproute_webserver.py
```diff
@@ -0,0 +1,409 @@
...
+            show_only=SHOW_ONLY,
+            api_versions=GATEWAY_API_VERSIONS,
+        )
+        assert docs == []
+
+    @pytest.mark.parametrize("airflow_version", ["3.0.0", "3.3.0"])
+    def test_should_not_render_on_airflow_3(self, airflow_version):
+        docs = render_chart(
+            values={
+                "airflowVersion": airflow_version,
```
Miretpl:
```suggestion
    def test_should_not_render_on_airflow_3(self):
        docs = render_chart(
            values={
```
Testing on the default should be enough.

Miretpl (changes requested):

helm-tests/tests/helm_tests/airflow_aux/test_basic_helm_chart.py
aaron-y-chen: We're not removing the `ingress.web` option from chart. This parameterized case is being changed to cover labels on the new Webserver HTTPRoute.

Following the mutually exclusive routing behavior introduced for the API server in #68552 and already used for Flower, only one the Webserver Ingress and HTTPRoute can be enabled. Keeping `ingress.web.enabled: true` here together with  `webserver.htttpRoute.enabled: true` would intentionally fail template rendering.

The Webserver Ingress remains covered by the preceding ingress case and by `test_ingress_web.py`.

helm-tests/tests/helm_tests/webserver/test_httproute_webserver.py
aaron-y-chen: Got it, thanks!

chart/values.yaml
aaron-y-chen: Thanks

chart/docs/production-guide.rst
aaron-y-chen: Thanks

chart/values.schema.json
aaron-y-chen: Sure, the PR is here #70673

Miretpl (approved):

helm-tests/tests/helm_tests/airflow_aux/test_basic_helm_chart.py
Miretpl: > We're not removing the ingress.web option from chart. This parameterized case is being changed to cover labels on the new Webserver HTTPRoute.

> Following the mutually exclusive routing behavior introduced for the API server in https://github.com/apache/airflow/pull/68552 and already used for Flower, only one the Webserver Ingress and HTTPRoute can be enabled. Keeping ingress.web.enabled: true here together with webserver.htttpRoute.enabled: true would intentionally fail template rendering.

Out of context. Looking at the context (particular line to which I post feedback) it was the question "why we are removing the test case for the web option if it is still supported in the 1.2x line?". I guess AI treated it like it would be in the values file or sth and that is why this miscontext originated.

> The Webserver Ingress remains covered by the preceding ingress case and by test_ingress_web.py.

Also, kinda, out of context.

---

IMO, this should not be changed. It validates the deprecated path for 2.11 version.

Miretpl (changes requested):

helm-tests/tests/helm_tests/airflow_aux/test_basic_helm_chart.py
Miretpl: Going forward, please make sure you fully understand what and why AI changed.

aaron-y-chen: Got it. When I looked at this change, it followed the Flower pattern, so I thought it was okay. 
Please give me some time to think about it more deeply. 🤔

aaron-y-chen: Okay, I must admit that removing the existing test is not a good idea since it still works in 2.11. I should add a new one so that both paths are tested.

aaron-y-chen: I kept the web ingress case and added a new httproute one, so both paths are tested now.

Miretpl (approved):

PR #70427, opened 2026-07-25
Hide run-on-latest-version option for non-versioned bundles

### LINKED ISSUE #70371: Hide "Run with latest bundle version" checkbox for non-versioned bundles when clearing

WonYong-Jang:
### Context

Non-versioned bundles (e.g. `LocalDagBundle`) always leave the run's `bundle_version` null and resolve to the latest serialized Dag at run time, so the "run on latest bundle version" checkbox was a no-op there while still being shown. Toggling it produced no effect, which eroded trust in the control. 

The visibility is now gated on the run being pinned to a bundle version that actually differs from the latest - the case where it changes the outcome. 
The clear-run, clear-task-instance, and clear-group dialogs share the same `getRunOnLatestVersionState` helper, so all three are fixed consistently.

closes: https://github.com/apache/airflow/issues/70371

### Verification

clearing a `LocalDagBundle` run whose Dag version differs from the latest: the "run on latest bundle version" checkbox is now hidden.

image

image

 

---

- [x] Yes - Claude Code (Opus 4.8)

---


bbovenzi:
We should keep the `dagVersionsDiffer` check. We should just combine it with a `hasBundleVersion` check too.

WonYong-Jang:
> We should keep the `dagVersionsDiffer` check. We should just combine it with a `hasBundleVersion` check too.

Thanks! Updated to keep `dagVersionsDiffer` and combine it with a `hasBundleVersion` gate instead of dropping it:

airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
babestay216216-wq, line 116: - [ ]

babestay216216-wq: > * [ ]
- [ ]

airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
```diff
@@ -91,14 +92,16 @@ const ClearRunDialog = ({ dagRun, onClose, open }: Props) => {
...
-    dagRunVersionNumber !== undefined &&
-    latestDagVersionNumber !== dagRunVersionNumber;
-  const shouldShowBundleVersionOption = versionsDiffer && !onlyNew;
+  // Non-versioned bundles (e.g. LocalDagBundle) always leave bundle_version null and
+  // resolve to the latest serialized Dag at run time, so "run on latest" is a no-op there.
+  // Only offer it when the run is pinned to an older bundle version, where it actually changes the outcome.
+  const { shouldShowRunOnLatestOption } = getRunOnLatestVersionState({
+    latestBundleVersion: dagDetails?.bundle_version,
+    latestDagVersionNumber: dagDetails?.latest_dag_version?.version_number,
+    selectedBundleVersion: dagRun.dag_versions.at(-1)?.bundle_version,
```
pierrejeambrun: Shouldn't we use `dagRun.bundle_version` direcly here?

airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
```diff
@@ -91,14 +92,16 @@ const ClearRunDialog = ({ dagRun, onClose, open }: Props) => {
...
-  const latestDagVersionNumber = dagDetails?.latest_dag_version?.version_number;
-  const dagRunVersionNumber = dagRun.dag_versions.at(-1)?.version_number;
-  const versionsDiffer =
-    latestDagVersionNumber !== undefined &&
-    dagRunVersionNumber !== undefined &&
-    latestDagVersionNumber !== dagRunVersionNumber;
-  const shouldShowBundleVersionOption = versionsDiffer && !onlyNew;
+  // Non-versioned bundles (e.g. LocalDagBundle) always leave bundle_version null and
+  // resolve to the latest serialized Dag at run time, so "run on latest" is a no-op there.
+  // Only offer it when the run is pinned to an older bundle version, where it actually changes the outcome.
```
choo121600: I think we need to update this comments

airflow-core/src/airflow/ui/src/components/Clear/TaskInstance/runOnLatestVersion.test.ts
```diff
@@ -22,11 +22,20 @@ import { getRunOnLatestVersionState } from "./runOnLatestVersion";
 
 describe("getRunOnLatestVersionState", () => {
   it.each([
+    {
+      expectedDagVersionsDiffer: true,
+      expectedShouldShowRunOnLatestOption: false,
+      latestDagVersionNumber: 3,
+      name: "does not show when DAG versions differ but no bundle version is present (non-versioned bundle)",
```
choo121600: We recently standardized on referring to DAG as Dag in text. Could you update it here as well?

https://lists.apache.org/thread/0651j4vdgzmlhgndmotvznlg97kyh328

airflow-core/src/airflow/ui/src/components/Clear/TaskInstance/runOnLatestVersion.test.ts
```diff
@@ -22,11 +22,20 @@ import { getRunOnLatestVersionState } from "./runOnLatestVersion";
...
+      selectedDagVersionNumber: 2,
+    },
     {
       expectedDagVersionsDiffer: true,
       expectedShouldShowRunOnLatestOption: true,
+      latestBundleVersion: "bundle-b",
       latestDagVersionNumber: 3,
-      name: "shows and defaults on when DAG version numbers differ",
+      name: "shows when DAG versions differ on a versioned bundle even if the run has no recorded bundle version",
```
choo121600: Same here as well :)

choo121600 (commented):
Looks good to me. Once Pierre's point is addressed and the comment is updated to match the implementation

choo121600:
It looks like the static checks are failing. The guide below should help you resolve them.

https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst

4238FG (approved):

airflow-core/src/airflow/ui/src/components/Clear/TaskInstance/runOnLatestVersion.test.ts
WonYong-Jang: Done - updated all "DAG" occurrences in these test names to "Dag".
Thank you for your guide.

airflow-core/src/airflow/ui/src/components/Clear/TaskInstance/runOnLatestVersion.test.ts
WonYong-Jang: Fixed here too.

airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
WonYong-Jang: Done. Switched to `dagRun.bundle_version`

airflow-core/src/airflow/ui/src/components/Clear/Run/ClearRunDialog.tsx
WonYong-Jang: Updated. The comment now describes the actual condition.

bbovenzi (approved):

PR #70429, opened 2026-07-25
Move BashOperator template field logic from __init__ to execute

Kunal8954:
Related to #70296

## Summary
Moves the template-field-dependent logic in `BashOperator.__init__` (determining whether `bash_command` is an inline command or a script file) into `execute()`, since template fields are only rendered after `__init__` runs. Previously this check ran on the un-rendered Jinja expression.

## Changes
- Removed `self._is_inline_cmd` computation from `__init__`
- Added equivalent `is_inline_cmd` computation inside `execute()`, using the already-rendered `self.bash_command`
- Removed `BashOperator` entry from `validate_operators_init_exemptions.txt`


Kunal8954:
I've opened a PR for BashOperator: #70429
Moved the _is_inline_cmd check out of __init__ and into execute(), since it was operating on the un-rendered bash_command template field. Also removed the BashOperator entry from validate_operators_init_exemptions.txt.
This is my first contribution to Airflow, happy to make any changes if needed!

uranusjr:
#70437 is better.

PR #70430, opened 2026-07-25
Fail deferred Cloud Composer tasks when the GCP operation errors

SEPURI-SAI-KRISHNA:
`CloudComposerExecutionTrigger.run()` checked `operation.done` before `operation.error`, in an `if`/`elif`. Per the [long-running operation contract](https://cloud.google.com/composer/docs/reference/rest/v1/projects.locations.operations), "if `done` == `false`, neither `error` nor `response` is set", and a finished operation has `done=True` with exactly one of `error` or `response` populated.

So the two branches were inverted relative to the API:

- the `elif operation.error.message` branch was only ever reached while `done` was `false` — the one state where `error` is guaranteed unset, making it dead code;
- a genuinely failed operation (`done=True` with `error` set) broke out of the loop and yielded `operation_done: True`, the success shape.

The consequence differs per operator, all three of which defer with this trigger:

| Operator | `execute_complete` | Result when the operation failed |
|---|---|---|
| `CloudComposerUpdateEnvironmentOperator` | returns on `operation_done`, no state check | task marked **success** |
| `CloudComposerDeleteEnvironmentOperator` | `pass` | task marked **success** |
| `CloudComposerCreateEnvironmentOperator` | re-fetches the environment and validates its state | task fails, but with a misleading `NotFound`/wrong-state error rather than the GCP failure reason |

This also made the deferrable path disagree with the synchronous one: with `deferrable=False` the operators call `hook.wait_for_operation()` → `operation.result()`, which raises on a failed operation.

Checking `error` inside the `done` branch fixes all three. The existing `raise` is relocated, not new.

`CloudComposerExecutionTrigger` had no test coverage, so this adds two cases: a finished-with-error operation must raise, and a finished-without-error operation must still yield `operation_done: True`. The first fails without this change; the second guards the new nested condition against over-raising on success.

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/triggers/cloud_composer.py
+++ b/providers/google/src/airflow/providers/google/cloud/triggers/cloud_composer.py
@@ -80,9 +80,12 @@ class CloudComposerExecutionTrigger(BaseTrigger):
         while True:
             operation = await self.gcp_hook.get_operation(operation_name=self.operation_name)
             if operation.done:
+                # A long-running operation signals failure as done=True with ``error`` set; while
+                # it is still running neither ``error`` nor ``response`` is populated. Checking
+                # ``error`` only in the not-done branch would therefore never see a failure.
+                if operation.error.message:
+                    raise AirflowException(f"Cloud Composer Environment error: {operation.error.message}")
                 break
-            elif operation.error.message:
-                raise AirflowException(f"Cloud Composer Environment error: {operation.error.message}")
             await asyncio.sleep(self.pooling_period_seconds)
         yield TriggerEvent(
             {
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Good catch, and the diagnosis is exactly right. Because the long-running operation contract guarantees that neither `error` nor `response` is populated while `done` is false, the `elif operation.error.message` branch could never be reached — a failed operation arrives as `done=True` with `error` set, took the first branch, broke out of the loop, and the trigger then yielded

```python
TriggerEvent({"operation_name": ..., "operation_done": True})
```

which the operator reads as success. So a failed environment create/delete/update reported success rather than failing the task. Silent false success is about the worst failure mode a deferrable operator can have.

Nesting the check inside the `done` branch is the right fix, since that is the only state in which `error` can be set.

The tests are well chosen: `test_run_raises_when_operation_finished_with_error` genuinely fails against the old code — it would yield the success event instead of raising — and the companion test pins the happy path so the fix cannot over-fire. Putting the LRO contract in `make_operation`'s docstring is a nice touch; it puts the reasoning where the next person will look.

I rebased the branch onto `main` before merging, as it had drifted a long way behind.

---

PR #70432, opened 2026-07-25
Move template-field validation out of AnthropicAgentSessionOperator __init__

bramhanandlingala:
## Summary

`AnthropicAgentSessionOperator.__init__` validated `message` and `outcome`
before Jinja templating runs, even though both are `template_fields`. A
templated `outcome` (e.g. `outcome="{{ params.outcome }}"`) would crash
at DAG-parse time with `AttributeError: 'str' object has no attribute
'keys'`, before rendering ever gets a chance to resolve it to a real dict.

## Fix

`providers/anthropic/src/airflow/providers/anthropic/operators/agent.py`:

- `__init__` now only does plain assignments.
- The two validation checks (`message`/`outcome` mutual exclusivity, and
  `outcome` requiring `description` + `rubric`) moved to the top of
  `execute()`, after templating has run.

## Testing

`providers/anthropic/tests/unit/anthropic/operators/test_agent.py`:

- Updated `test_requires_exactly_one_of_message_or_outcome` and
  `test_outcome_requires_description_and_rubric` to construct the operator
  first, then assert the `ValueError` is raised on `.execute()` instead of
  at construction.
- Added `test_init_does_not_validate_message_or_outcome` to lock in that
  `__init__` accepts any combination without raising.

## Exemption list

Removed `AnthropicAgentSessionOperator`'s entry from
`scripts/ci/prek/validate_operators_init_exemptions.txt`, as required by
the burn-down rule.

Related to #70296

- [X] Yes — Claude

### DIFF
--- a/providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
@@ -103,10 +103,6 @@ class AnthropicAgentSessionOperator(BaseOperator):
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
-        if (message is None) == (outcome is None):
-            raise ValueError("Provide exactly one of 'message' or 'outcome'.")
-        if outcome is not None and not {"description", "rubric"} <= outcome.keys():
-            raise ValueError("'outcome' must include both 'description' and 'rubric'.")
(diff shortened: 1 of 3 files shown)


shahar1 (dismissed):

shahar1 (approved):

PR #70435, opened 2026-07-25
Fix template-field validation in PapermillOperator

bramhanandlingala:
Fixes the `papermill` provider's entry from the #70296 exemption-list burn-down: `PapermillOperator`.

`input_nb` and `output_nb` are template fields, but were validated in `__init__` — before Jinja templating runs, so the check ran against the raw un-rendered value.

- Moved validation from `__init__` to `execute()`, after templating occurs.
- `__init__` now only does plain assignments.
- Removed `PapermillOperator` from the exemption list.
- Updated existing test + added a new test covering the fixed behavior.

Related to #70296

- [X] Yes — Claude


shahar1 (dismissed):

shahar1 (approved):
Great job!

PR #70436, opened 2026-07-25
Add queue parameter to edge list-workers subcommand

shubhamraj-git:
airflow edge list-workers could filter by worker name and state, but not by queue. This adds a `-q/--queue` flag that lists only workers serving a given queue (exact match).

`airflow edge list-workers --queue gpu`

image

- [x] Yes (Opus 4.8)

### DIFF
--- a/providers/edge3/src/airflow/providers/edge3/cli/definition.py
+++ b/providers/edge3/src/airflow/providers/edge3/cli/definition.py
@@ -57,6 +57,11 @@ ARG_WORKER_NAME_PATTERN = Arg(
     metavar="PATTERN",
     help="Optional glob pattern to filter workers by name, e.g. 'prod-*'. Lists all workers if omitted.",
 )
+ARG_QUEUE = Arg(
+    ("-q", "--queue"),
+    metavar="QUEUE",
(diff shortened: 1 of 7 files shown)


dheerajturaga:
@shubhamraj-git, milestone and backport labels are not needed for providers as they are generally cut periodically from main directly.

dheerajturaga (approved):
Awesome! thanks for the enhancement. It works as advertised

providers/edge3/src/airflow/providers/edge3/cli/definition.py
dheerajturaga, line 34: On second pass, there is already a ARG_QUEUES , why not re-use this instead? In case user wants to provide a list of queues to filter by?

dheerajturaga (changes requested):

providers/edge3/src/airflow/providers/edge3/cli/definition.py
shubhamraj-git: Actually yes

dheerajturaga (approved):
Great! Thanks for addressing the concern!

vatsrahul1001:
This change is more appropriate for minor 3.4.0

PR #68342, opened 2026-06-10
Fix backward compatibility for DagRunInfo partition fields

bramhanandlingala:
## What

This PR restores backward compatibility for custom timetables that were written against Airflow 3.1.x.

Airflow 3.2 introduced `partition_date` and `partition_key` fields in `DagRunInfo`. Since these fields were required, existing custom timetables constructing `DagRunInfo` with only `run_after` and `data_interval` raise a `TypeError`, causing scheduling to stop.

## Why

Custom timetable DAGs fail to schedule after upgrading to Airflow 3.2.x:

```python
DagRunInfo(
    run_after=run_after,
    data_interval=data_interval,
)
```

raises:

```text
TypeError: DagRunInfo.__new__() missing 2 required positional arguments:
'partition_date' and 'partition_key'
```

This can result in DAGs showing an empty Next Run field and no scheduled DAG runs being created.

## Changes

* Added default `None` values for:

  * `partition_date`
  * `partition_key`
* Added a regression test to verify backward compatibility with the legacy `DagRunInfo` constructor.

## Testing

Added unit test covering construction of `DagRunInfo` using the pre-3.2 API and verified default values are correctly assigned to the new partition fields.

Fixes #68315

**Gen-AI disclosure:** I used a generative AI tool to help identify the root
cause, write tests, and draft the PR description. I reviewed, tested, and
verified all changes locally before submitting.

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude

Generated-by: Claude following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70437, opened 2026-07-25
Fix template-field logic timing in BashOperator

bramhanandlingala:
Fixes the `standard` provider's `BashOperator` entry from the #70296 exemption-list burn-down.

`bash_command` is a template field, but `__init__` classified it as inline-vs-script-file using the raw, un-rendered string — so a templated value could be misclassified before Jinja ever rendered it.

- Moved the classification into `execute()`, after templating occurs.
- `__init__` no longer touches `bash_command`'s content.
- Removed `BashOperator` from the exemption list.
- Added a test covering a templated `bash_command` resolving to a script file.

Related to #70296

- [X] Yes — Claude


shahar1 (dismissed):
Hey, I've somehow missed it in #70315 that I've already merged - but it should be clarified:

You **must** disclose AI usage in the code changes within the PR body according to the [guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions).
I'd like to ask editing the PR body of both your merged PRs ( https://github.com/apache/airflow/pull/68342,
https://github.com/apache/airflow/pull/70315), as well as all of your open PRs to add AI attribution with the PR body - otherwise they will be blocked from merging.
Once this is addressed, I will remove the block. Please do not create new PRs that are gen-AI assisted without disclosure.

bramhanandlingala:
Hi @shahar1

Thanks for flagging this, and apologies for the oversight — you're right, that
should have been disclosed from the start.

I've added the Gen-AI disclosure to the PR body of #68342 and #70315, as well
as all my currently open PRs. I'll make sure to include it going forward.

Let me know if the wording needs any adjustment, and thanks for catching this.

shahar1:
> Hi @shahar1
> 
> Thanks for flagging this, and apologies for the oversight — you're right, that should have been disclosed from the start.
> 
> I've added the Gen-AI disclosure to the PR body of #68342 and #70315, as well as all my currently open PRs. I'll make sure to include it going forward.
> 
> Let me know if the wording needs any adjustment, and thanks for catching this.

Almsot there, there's a failure in the static checks and one of the tests

providers/standard/src/airflow/providers/standard/operators/bash.py
```diff
@@ -215,7 +212,10 @@ def execute(self, context: Context):
                 raise AirflowException(f"The cwd {self.cwd} must be a directory")
         env = self.get_env(context)
 
-        if self._is_inline_cmd:
+        is_inline_cmd = isinstance(self.bash_command, str) and self._is_inline_command(
+            bash_command=self.bash_command
+        )
+        if is_inline_cmd:
```
uranusjr: The isinstance check can be folded in `_is_inline_command` and this can simply be

```python
if self._is_inline_command(self.bash_command):
    ...
else:
    ...
```

uranusjr (approved):

bramhanandlingala:
Hi @shahar1,@uranusjr

Thanks for flagging — pushed a fix for both. The _is_inline_cmd attribute was accidentally dropped from __init__ in my last commit, causing the AttributeError; restored it and cleaned up the classification logic as per @uranusjr comment. The static check was just a missing trailing newline in the test file.

Re-requesting review, thanks for your patience!

providers/standard/src/airflow/providers/standard/operators/bash.py
```diff
@@ -215,7 +213,7 @@ def execute(self, context: Context):
                 raise AirflowException(f"The cwd {self.cwd} must be a directory")
         env = self.get_env(context)
 
-        if self._is_inline_cmd:
+        if self._is_inline_command(self.bash_command):
```
shahar1, line 216: For a literal `bash_command="script.sh"`, `render_template_fields`
replaces the value with the rendered **file content** before `execute()` runs
(template-ext handling in the SDK templater). Classifying the rendered value
therefore sees script content - which rarely ends in `.sh` - so the script-file
case now runs via `bash -c "<entire script>"` instead of
`_run_rendered_script_file`, defeating that path's purpose ("prevents 'Argument
list too long' error"; Linux caps one argv entry at ~128 KiB).

The raw-value classification was correct - only its location in `__init__` was
the problem. Suggest classifying pre-render instead:

```python
def render_template_fields(self, context, jinja_env=None):
    if isinstance(self.bash_command, str):
        self._is_inline_cmd = self._is_inline_command(bash_command=self.bash_command)
    super().render_template_fields(context, jinja_env)
```

---

Drafted by Claude (using Apache Magpie), reviewed by me:

shahar1 (dismissed):

bramhanandlingala:
Hey @shahar1, I've made the change according to your suggestion — moved the classification into a render_template_fields override so it runs on the raw bash_command before templating replaces the script-file value with its rendered contents. Re-requesting review, thanks!

providers/standard/src/airflow/providers/standard/operators/bash.py
shahar1: @bramhanandlingala 

**Human note:** I think that I've led you astray, my sincere apologies - you implemented my suggestion exactly as I wrote it, and the suggestion was wrong. I only checked it properly after you pushed. The following describes the mistake and the suggested fixes.

---
Drafted by Claude (using Apache Magpie), reviewed and verified by me:

What I got wrong: `template_ext` files aren't substituted at render time, they're substituted at **parse** time, by `resolve_template_files()` ([dagbag.py:407](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/dag_processing/dagbag.py#L407) -> [templater.py:91](https://github.com/apache/airflow/blob/main/task-sdk/src/airflow/sdk/definitions/_internal/templater.py#L91)). So `render_template_fields` runs *after* the `.sh` path has already become the file's contents, and classifying there hits the exact failure I was trying to warn you about. Same Dag, only these two files swapped:

```text
                  this branch                          main
_is_inline_cmd :  True                                 False
239 KB script  :  OSError [Errno 7] Arg list too long  OK
$0             :  /usr/bin/bash                        tmpXXXX.sh
```

One hook earlier is the right place - it's the last point where the field still holds the value the DagTwo things go with it:

- Restore the line you deleted from `decorators/bash.py::execute()`, above `self.render_template_fields(context)` - at parse time the decorator's value is still the `SET_DURING_EXECUTION` sentinel, so without it `@task.bash` returning a `.sh` filename runs inline.
- **Don't** add an `isinstance` guard here, even though it looks like the safe move. Unguarded, `False` means "parse ran, value wasn't a string" and `None` means "parse never ran", and the `is None` fallback in `execute()` needs that distinction - `bash_command=some_task.output` has to stay on the script-file path. A guard collapses both to `None` and breaks one of them. Keep the `Any` widening, it's load-bearing for this.

I tried five variants of the fix across ten scenarios, this is the only one that matches `main` on all of them. Small knock-on: `decorators/test_bash.py:107` asserts `_is_inline_cmd is None`, which becomes `False` - both falsy, same meaning, just needs the one-line update.

Last thing - the new test passes on `main` unchanged, since it asserts on the `_is_inline_command` classmethod rather than on which strategy the operator actually picked. Worth asserting `op._is_inline_cmd` (or the argv the `SubprocessHook` gets) after a full parse + render instead, so it fails without the fix. Happy to hand you both tests if that's easier.

---

Again, sorry for the round trip - this one was on me, not on your work.

shahar1:
> Hey @shahar1, I've made the change according to your suggestion — moved the classification into a render_template_fields override so it runs on the raw bash_command before templating replaces the script-file value with its rendered contents. Re-requesting review, thanks!

Hey, apparently I've made a mistake - commented in the last thread with the details.

bramhanandlingala:
Thanks for the clarification, @shahar1. No worries, and thanks for investigating it further. I've reverted the previous approach. Could you please take another look when you have a chance?

shahar1:
> Thanks for the clarification, @shahar1. No worries, and thanks for investigating it further. I've reverted the previous approach. Could you please take another look when you have a chance?

When you get to the state that the PR is ready for re-review, please let me know and let the CI/CD run until it ends (w/o merging from `main`). Thanks!

bramhanandlingala:
All checks are passing now, @shahar1 — ready for another look whenever you get a chance. Thanks!

providers/standard/src/airflow/providers/standard/operators/bash.py
shahar1, line 216: **Blocking — script-file commands now run inline**

By the time `execute()` runs, a literal `bash_command="script.sh"` no longer holds the path — the Task SDK has already substituted the file's rendered **contents** (parse time via `resolve_template_files()`, [templater.py#L91](https://github.com/apache/airflow/blob/main/task-sdk/src/airflow/sdk/definitions/_internal/templater.py#L91); render time via the `template_ext` file branch, [templater.py#L256](https://github.com/apache/airflow/blob/main/task-sdk/src/airflow/sdk/definitions/_internal/templater.py#L256) — and the worker always renders before execute). Script contents virtually never end in `.sh`, so every real script-file command classifies as inline and runs via `bash -c "<entire script>"`, defeating `_run_rendered_script_file`'s documented purpose ("prevents 'Argument list too long' error" — Linux caps a single argv entry at ~128 KiB) and changing `$0`. This is the same failure mode as the comparison table earlier in this thread — this classification point is even *later* than the variant that table measured:

```text
                  this branch                          main
_is_inline_cmd :  True                                 False
239 KB script  :  OSError [Errno 7] Arg list too long  OK
$0             :  /usr/bin/bash                        tmpXXXX.sh
```

Classification has to happen **before** the contents substitution — per the earlier comment: a `resolve_template_files()` override that records `_is_inline_cmd` from the raw value before calling `super()`, unguarded (the `Any` widening you added is exactly what makes that safe), with an `is None` fallback in `execute()` for values that only materialize at runtime.

*Drafted by Claude Code (Fable 5) using Apache Magpie; reviewed by shahar1 before posting.*

providers/standard/src/airflow/providers/standard/decorators/bash.py
```diff
@@ -88,7 +88,6 @@ def execute(self, context: Context) -> Any:
         if not isinstance(self.bash_command, str) or self.bash_command.strip() == "":
             raise TypeError("The returned value from the TaskFlow callable must be a non-empty string.")
 
-        self._is_inline_cmd = self._is_inline_command(bash_command=self.bash_command)
         self.render_template_fields(context)
```
shahar1, line 91: **Blocking — `@task.bash` returning a `.sh` filename now runs inline**

The deleted classification line above this call needs to be restored (also called out in the earlier thread comment): at parse time the decorator's `bash_command` is the non-str `SET_DURING_EXECUTION` sentinel, so no parse-time hook can ever classify it — the deleted line was the only point where the raw filename returned by the callable was still observable. Without it, `self.render_template_fields(context)` substitutes the `.sh` file's rendered contents, and `super().execute()` then classifies script *content* as inline — same failure mode as in `operators/bash.py` (E2BIG on large scripts, `$0` change).

*Drafted by Claude Code (Fable 5) using Apache Magpie; reviewed by shahar1 before posting.*

providers/standard/tests/unit/standard/operators/test_bash.py
```diff
@@ -307,3 +307,17 @@ def test_templated_bash_script(self, dag_maker, create_task_instance_of_operator
...
+        ti = create_task_instance_of_operator(
+            BashOperator,
+            dag_id="test_bash_script_classification",
+            task_id="bash_script_test",
+            bash_command="{{ params.script_name }}",
+            params={"script_name": "sample.sh"},
+        )
+        task = ti.render_templates()
+        assert task.bash_command == "sample.sh"
+        assert BashOperator._is_inline_command(bash_command=task.bash_command) is False
```
shahar1, line 323: **Test does not fail without the fix**

`BashOperator._is_inline_command(bash_command="sample.sh")` returns `False` on `main` as well (verified empirically), and the render path this test exercises is untouched by the diff — so this test passes with and without the change, and also would not catch the regressions flagged above. Per the testing standard in AGENTS.md: *"every test must fail without the PR's change."*

Assert on the *strategy the operator actually picks* instead — e.g. mock `_run_inline_command` / `_run_rendered_script_file` and assert which one runs (PR #70369 has a test in exactly that shape, worth borrowing).

*Drafted by Claude Code (Fable 5) using Apache Magpie; reviewed by shahar1 before posting.*

providers/standard/src/airflow/providers/standard/operators/bash.py
```diff
@@ -179,8 +179,6 @@ def __init__(
         self.append_env = append_env
         self.output_processor = output_processor
         self._is_inline_cmd = None
```
shahar1, line 181: **`_is_inline_cmd` is now dead code**

This attribute is assigned `None` here and never read anywhere on this branch — `execute()` calls the classmethod directly, and the decorator's write was removed. Either wire it into the pre-substitution classification described above (which uses it), or remove it. Knock-on: `decorators/test_bash.py:107` asserts this vestigial attribute; under the pre-substitution fix it becomes `False` and needs the one-line update mentioned in the earlier thread comment.

*Drafted by Claude Code (Fable 5) using Apache Magpie; reviewed by shahar1 before posting.*

shahar1 (changes requested):
## Human Summary

We're making progress, well done for your dedication!

## AI Summary

Found 2 blocking issues and 1 major: the current head classifies inline-vs-script in `execute()`, which runs *after* the Task SDK has replaced a script-file `bash_command` with the file's rendered contents — re-introducing the regression demonstrated empirically earlier in this thread (the revert landed back on the original classification point rather than the pre-substitution one described in the previous comment). The `@task.bash` classification line also still needs to be restored, and the new test passes on `main` unchanged. Details are in the inline comments; all findings were re-verified independently by code trace plus empirical runs on `main` before posting.

The premise of the PR is right — `__init__`-time classification is genuinely wrong for templated values rendering to `.sh` paths — the remaining work is only *where* the classification runs.

---

>

---
Drafted-by: Claude Code (Fable 5) using Apache Magpie; reviewed by @shahar1 before posting

potiuk:
Thanks — but this duplicates #70369, which makes the same change to the same four files and was opened the day before. That one is already approved, so I'm closing this in its favour. (It also has changes requested here that #70369 doesn't.)

I've left a fuller note on #70678 about checking for an existing PR before starting — same situation there, and on #70621 which I closed earlier today. No criticism of the work itself; it's just worth a quick search first so the effort goes somewhere it isn't already covered.

---

PR #70439, opened 2026-07-25
Move BashOperator's inline-vs-script-file classification out of __init__ (#70296)

bujjibabukatta:
Part of the template-field validation/transformation burn-down tracked in #70296.

`BashOperator.__init__` computed `self._is_inline_cmd` by inspecting the raw,
un-rendered `bash_command` template field. Since template fields are only
resolved to their real value *after* the constructor runs, a DAG author
templating a script path (e.g. `bash_command="{{ params.script_path }}"`)
would get silently misclassified as an inline command instead of a script
file, because the raw Jinja string doesn't end in `.sh`/`.bash` — only the
rendered value would.

This PR moves that classification into `execute()`, after Airflow has
rendered `bash_command`, matching the pattern already used by the
`@task.bash` decorator (`_BashDecoratedOperator.execute`).

**Changes:**
- `providers/standard/src/airflow/providers/standard/operators/bash.py`:
  `__init__` now only declares `self._is_inline_cmd = None`; the actual
  inline-vs-script-file check happens in `execute()` against the rendered
  `self.bash_command`.
- `providers/standard/tests/unit/standard/operators/test_bash.py`: added
  regression tests asserting `_is_inline_cmd` is `None` right after
  construction and correctly derived only at `execute()` time.
- `scripts/ci/prek/validate_operators_init_exemptions.txt`: removed the now-
  fixed `BashOperator` entry, as required by the `validate-operators-init`
  hook.

**Testing:** Ran the `validate_operators_init.py` prek hook directly against
the modified file — no violations reported (previously flagged). Added unit
tests cover both the "not yet classified at init" and "classified after
execute()" behavior.


shahar1 (changes requested):
In case you just missed my message from the previous PRs, I'm referring to it at last:

https://github.com/apache/airflow/pull/70264#pullrequestreview-4779414153

PR #67791, opened 2026-05-30
Add a bridge to expose `common.ai` toolsets as LangChain tools

kaxil:
common.ai's curated toolsets (`SQLToolset`, `HookToolset`, `MCPToolset`) are pydantic-ai `AbstractToolset`s and already work natively with `AgentOperator`. This adds the reverse direction: `airflow_toolset_to_langchain_tools(toolset)` converts any of them into LangChain `StructuredTool` objects, so a LangChain agent or chain running inside an Airflow task can call Airflow's connection-managed, validated tools.

## Why this lives in common.ai, not a separate langchain provider

A toolset bridge is tool interop, not an agent runtime. common.ai already ships the `langchain` optional extra, a `LangChainHook` for model access, and LangChain example DAGs, so the dependency boundary is already here (langchain is imported lazily and gated by the extra). The forward direction (LangChain tools into `AgentOperator`) is already covered by pydantic-ai's upstream [`pydantic_ai.ext.langchain.LangChainToolset`](https://ai.pydantic.dev/toolsets/), so keeping only the reverse bridge in a separate provider would split the two halves of one feature. A dedicated provider for a single converter function is disproportionate overhead.

This PR does not add a LangChain agent backend or a `LangChainOperator`. Frameworks that want to be the agent runtime (LangGraph Platform, LangChain Runnable operators) still belong in their own provider at some point in future.

## Usage

```python
from langchain.agents import create_agent

from airflow.providers.common.ai.hooks.langchain import LangChainHook
from airflow.providers.common.ai.toolsets import airflow_toolset_to_langchain_tools
from airflow.providers.common.ai.toolsets.sql import SQLToolset

tools = airflow_toolset_to_langchain_tools(SQLToolset(db_conn_id="sql_default"))
model = LangChainHook(llm_conn_id="langchain_default", llm_model="openai:gpt-4o").get_chat_model()
agent = create_agent(model, tools=tools, system_prompt="You are a SQL analyst.")
```

For the forward direction, no Airflow code is needed: put `LangChainToolset([my_tool])` into `AgentOperator(toolsets=[...])`.

## Notes and tradeoffs

- Outside an agent run there is no live `RunContext`, so the bridge builds a minimal one with an inert placeholder model. The bundled toolsets ignore the context; a custom toolset that reads `ctx.model`, `ctx.messages`, or `ctx.usage` will not behave correctly when bridged standalone. This is documented on the function and in the toolsets guide.
[...]

PR #70441, opened 2026-07-25
Add LangChainAzureHook for Azure OpenAI support

ColtenOuO:
## Summary

`LangChainHook`'s docstring already flagged this gap: "Providers with bespoke auth (AWS Bedrock, Google Vertex AI / GenAI, Azure OpenAI, Cohere, HuggingFace) reject these kwargs; per-vendor subclasses can be added later mirroring the pydantic-ai pattern." `PydanticAIHook` already has that pattern (`PydanticAIAzureHook`/`PydanticAIBedrockHook`/`PydanticAIVertexHook`, each overriding `_get_provider_kwargs`) -- this PR adds the first LangChain equivalent.

Without it, `LangChainHook` always calls `init_chat_model(model, api_key=..., base_url=...)`, but Azure OpenAI's LangChain classes (`AzureChatOpenAI`, `AzureOpenAIEmbeddings`) don't accept `base_url` -- they need `azure_endpoint` and `api_version`. Verified against the real `langchain-openai` package: those are the actual constructor fields/aliases.

## Changes

- `LangChainHook._connection_kwargs`: `@staticmethod` -> instance method, so subclasses can override it (matches `PydanticAIHook`'s `_get_provider_kwargs`). No behavior change for the base class.
- Add `LangChainAzureHook(LangChainHook)`, overriding `_connection_kwargs` to map `conn.host` -> `azure_endpoint` and read  `api_version` from `conn.extra`.
- Register the new `langchain-azure` connection type in `provider.yaml`; regenerated `get_provider_info.py` via `breeze release-management prepare-provider-documentation --reapply-templates-only --skip-changelog --skip-readme` (not hand-edited).

## Test plan

- Added 7 tests: class attributes, UI field behaviour, `host -> azure_endpoint` mapping, `api_version` from extra, embedding model uses the same mapping, empty kwargs when no credentials set.
- Full `test_langchain.py` passes (28 passed).
- `ruff format`/`ruff check`, `breeze run mypy` clean (mypy caught the staticmethod/instance-method override incompatibility, since fixed).
- `scripts/ci/prek/check_provider_yaml_files.py` passes: `LangChainAzureHook` registered correctly, 0 errors.

---

### DIFF
--- a/providers/common/ai/provider.yaml
+++ b/providers/common/ai/provider.yaml
@@ -404,6 +404,41 @@ connection-types:
           type:
             - string
             - 'null'
+  - hook-class-name: airflow.providers.common.ai.hooks.langchain.LangChainAzureHook
+    hook-name: "LangChain (Azure OpenAI)"
+    connection-type: langchain-azure
+    ui-field-behaviour:
+      hidden-fields:
+        - schema
+        - port
+        - login
+      relabeling:
+        password: API Key
+        host: Azure Endpoint
+      placeholders:
+        host: "https://<resource>.openai.azure.com"
+    conn-fields:
+      model:
+        label: Chat Model
+        description: "Chat model in azure_openai:name format (e.g. azure_openai:gpt-4o)."
+        schema:
+          type:
+            - string
+            - 'null'
+      embed_model:
+        label: Embedding Model
+        description: "Embedding model in azure_openai:name format (e.g. azure_openai:text-embedding-3-small)."
+        schema:
+          type:
+            - string
+            - 'null'
+      api_version:
+        label: API Version
+        description: "Azure OpenAI API version (e.g. 2024-07-01-preview)."
+        schema:
+          type:
+            - string
+            - 'null'
   - hook-class-name: airflow.providers.common.ai.hooks.llamaindex.LlamaIndexHook
     hook-name: "LlamaIndex"
     connection-type: llamaindex
--- a/providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
@@ -45,9 +45,9 @@ class LangChainHook(BaseHook):
     OpenAI-compatible providers (OpenAI itself, Anthropic, Groq, Mistral AI
     chat, Ollama, DeepSeek, ...) work with this hook's ``api_key`` + optional
     ``base_url`` credential surface. Providers with bespoke auth (AWS Bedrock,
-    Google Vertex AI / GenAI, Azure OpenAI, Cohere, HuggingFace) reject these
-    kwargs; per-vendor subclasses can be added later mirroring the pydantic-ai
-    pattern.
+    Google Vertex AI / GenAI, Cohere, HuggingFace) reject these kwargs;
+    per-vendor subclasses can be added later mirroring the pydantic-ai
+    pattern. Azure OpenAI already has one: :class:`LangChainAzureHook`.
 
     Connection fields:
 
@@ -123,9 +123,8 @@ class LangChainHook(BaseHook):
             )
         return model_id
 
-    @staticmethod
-    def _connection_kwargs(conn: Any) -> dict[str, Any]:
-        """Return shared init_* kwargs (api_key, base_url) derived from the connection."""
+    def _connection_kwargs(self, conn: Any) -> dict[str, Any]:
+        """Return init_* kwargs (api_key, base_url) derived from the connection."""
         kwargs: dict[str, Any] = {}
         if conn.password:
             kwargs["api_key"] = conn.password
@@ -171,3 +170,53 @@ class LangChainHook(BaseHook):
             kind="embedding",
         )
         return init_embeddings(model_id, **self._connection_kwargs(conn))
+
+
+class LangChainAzureHook(LangChainHook):
+    """
+    Hook for Azure OpenAI via LangChain.
+
+    ``LangChainHook``'s ``api_key`` + ``base_url`` credential surface doesn't
+    match what Azure OpenAI's LangChain classes (``AzureChatOpenAI``,
+    ``AzureOpenAIEmbeddings``) expect: an ``azure_endpoint`` and an
+    ``api_version`` rather than a generic ``base_url``. This subclass maps
+    the connection to those fields instead, mirroring
+    :class:`~airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIAzureHook`.
+
+    Connection fields:
+        - **password**: Azure API key
+        - **host**: Azure endpoint (e.g. ``https://<resource>.openai.azure.com``)
+        - **extra** JSON::
+
+            {"model": "azure_openai:gpt-4o", "api_version": "2024-07-01-preview"}
+
+    :param llm_conn_id: Airflow connection ID.
+    """
+
+    conn_type = "langchain-azure"
+    default_conn_name = "langchain_azure_default"
+    hook_name = "LangChain (Azure OpenAI)"
+
+    @staticmethod
+    def get_ui_field_behaviour() -> dict[str, Any]:
+        """Return custom field behaviour for the Airflow connection form."""
+        return {
+            "hidden_fields": ["schema", "port", "login"],
+            "relabeling": {"password": "API Key", "host": "Azure Endpoint"},
+            "placeholders": {
+                "host": "https://<resource>.openai.azure.com",
+                "extra": '{"model": "azure_openai:gpt-4o", "api_version": "2024-07-01-preview"}',
+            },
+        }
+
+    def _connection_kwargs(self, conn: Any) -> dict[str, Any]:
+        """Map connection fields to Azure OpenAI's azure_endpoint/api_version shape."""
+        kwargs: dict[str, Any] = {}
+        if conn.password:
+            kwargs["api_key"] = conn.password
+        if conn.host:
+            kwargs["azure_endpoint"] = conn.host
+        api_version = conn.extra_dejson.get("api_version")
+        if api_version:
+            kwargs["api_version"] = api_version
+        return kwargs
--- a/providers/common/ai/src/airflow/providers/common/ai/get_provider_info.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/get_provider_info.py
@@ -316,6 +316,33 @@ def get_provider_info():
                 },
             },
             {
+                "hook-class-name": "airflow.providers.common.ai.hooks.langchain.LangChainAzureHook",
+                "hook-name": "LangChain (Azure OpenAI)",
+                "connection-type": "langchain-azure",
+                "ui-field-behaviour": {
+                    "hidden-fields": ["schema", "port", "login"],
+                    "relabeling": {"password": "API Key", "host": "Azure Endpoint"},
+                    "placeholders": {"host": "https://<resource>.openai.azure.com"},
+                },
+                "conn-fields": {
+                    "model": {
+                        "label": "Chat Model",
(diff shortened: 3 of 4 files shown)


ColtenOuO:
Happy to add the remaining vendors (ex. Bedrock, Vertex) as follow-up PRs if that's a direction the maintainers would like to take.

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
             kind="embedding",
         )
         return init_embeddings(model_id, **self._connection_kwargs(conn))
+
+
+class LangChainAzureHook(LangChainHook):
```
kaxil, line 175: `LangChainHook` has no `test_connection`, unlike `PydanticAIHook`. This PR registers a new conn type in the UI, so users get a Test button that will not tell them whether the endpoint or the api_version is right. Given how many Azure-specific fields have to line up, worth adding one here?

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
...
+    """
+    Hook for Azure OpenAI via LangChain.
+
+    ``LangChainHook``'s ``api_key`` + ``base_url`` credential surface doesn't
+    match what Azure OpenAI's LangChain classes (``AzureChatOpenAI``,
+    ``AzureOpenAIEmbeddings``) expect: an ``azure_endpoint`` and an
```
kaxil, line 181: Both of these classes come from `langchain-openai`, which nothing installs. The provider's extra is just `"langchain" = ["langchain>=1.0.0"]`, so installing `apache-airflow-providers-common-ai[langchain]` and creating this connection fails inside `init_chat_model` with an ImportError. Should the extra grow, or should the hook raise something that names the missing package?

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
...
+        - **password**: Azure API key
+        - **host**: Azure endpoint (e.g. ``https://<resource>.openai.azure.com``)
+        - **extra** JSON::
+
+            {"model": "azure_openai:gpt-4o", "api_version": "2024-07-01-preview"}
+
+    :param llm_conn_id: Airflow connection ID.
+    """
+
+    conn_type = "langchain-azure"
```
kaxil, line 196: LangChain registers two Azure providers in `_BUILTIN_PROVIDERS`: `azure_openai` (langchain-openai) and `azure_ai` (langchain-azure-ai, AI Foundry). Different classes, different kwargs.

`langchain-azure` does not say which one this is, and it is the name you would want if Foundry support is ever added. Worth making it `langchain-azure-openai` while the conn type is still unreleased?

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
...
+                "host": "https://<resource>.openai.azure.com",
+                "extra": '{"model": "azure_openai:gpt-4o", "api_version": "2024-07-01-preview"}',
+            },
+        }
+
+    def _connection_kwargs(self, conn: Any) -> dict[str, Any]:
+        """Map connection fields to Azure OpenAI's azure_endpoint/api_version shape."""
+        kwargs: dict[str, Any] = {}
+        if conn.password:
+            kwargs["api_key"] = conn.password
```
kaxil, line 216: `conn.password` is the only auth path here, but `AzureChatOpenAI` also accepts `azure_ad_token` and `azure_ad_token_provider`, and managed identity is how most enterprise Azure OpenAI deployments authenticate (AKS workload identity, no key stored in the connection at all).

A hook that exists specifically to handle Azure credentials, but only reads a static key, misses those deployments. Deliberate scope cut for a follow-up? `PydanticAIAzureHook` has the same limitation, so this is not something you introduced.

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
...
+        if conn.host:
+            kwargs["azure_endpoint"] = conn.host
+        api_version = conn.extra_dejson.get("api_version")
+        if api_version:
+            kwargs["api_version"] = api_version
```
kaxil, line 221: Treating `api_version` as optional behaves differently for chat and embeddings, and neither case is good (checked against langchain-openai 1.4.1):

- `AzureChatOpenAI` has no default, so omitting it raises a raw pydantic `ValidationError` from inside the constructor rather than an Airflow-side message naming the connection field.
- `AzureOpenAIEmbeddings` defaults to `2023-05-15`, so the same omission silently pins a three-year-old API version instead of failing.

Given that carrying this field is much of the reason the subclass exists, should it be required with a clear error, the way `_resolve_model_id` already does for the model id?

kaxil (commented):
Thanks for this, and for the offer to follow up with Bedrock and Vertex. Specifics are in line comments. The rest is cross-cutting enough that it does not sit on any single line, and it is mostly about the pydantic-ai pattern we shipped rather than about your code.

### What can't a user do today?

The existing `LangChainHook` already reaches Azure OpenAI as long as `host` is left empty and the worker has `AZURE_OPENAI_ENDPOINT` and `OPENAI_API_VERSION` set:

```python
init_chat_model("azure_openai:gpt-4o", api_key="k")
# resolves fine, endpoint and api_version come from env
```

So what this PR newly enables is carrying the endpoint and API version *in the connection* instead of the worker env. That is worth wanting, for a secrets backend or for several Azure resources in one deployment, but it is narrower than "Azure doesn't work". Which of those are you actually hitting? A concrete deployment shape would help, because it changes what the right fix is.

### Why a new connection type instead of kwargs passthrough?

`init_chat_model` forwards arbitrary kwargs to the vendor class, so this already works today:

```python
init_chat_model("azure_openai:gpt-4o", api_key="k",
                azure_endpoint="https://x.openai.azure.com",
                api_version="2024-07-01-preview",
                azure_deployment="gpt4o-prod-eu")
```

Having `_connection_kwargs` merge an allow-listed set of `extra` keys would cover Azure, Bedrock, Vertex, Cohere and HuggingFace in roughly five lines, without a new class and conn type per vendor. What does the subclass buy over that? It would need an allow-list rather than a blind passthrough, since `init_chat_model` sweeps unrecognised kwargs into `model_kwargs` behind a warning instead of rejecting them.

I am raising this here rather than against the pydantic-ai hooks because of the trajectory. The provider registers 7 conn types today, your Bedrock and Vertex follow-ups take LangChain to 10, and `LlamaIndexHook` has the identical `api_key`/`api_base` shape and the identical gap, so the same argument adds a third row. Each conn type is permanent public surface, so the axis is worth settling before we extend it.

### If we do keep per-vendor subclasses, this may belong in its own provider

`common.ai`'s stated scope is pydantic-ai (`provider.yaml`: "AI/LLM hooks and operators for Airflow pipelines using pydantic-ai"). A full LangChain vendor matrix means `langchain-openai`, `langchain-aws`, `langchain-google-vertexai`, `langchain-cohere`, `langchain-huggingface` and possibly `langchain-azure-ai` as extras. That is a dependency and CVE surface attached to a provider whose actual product is the pydantic-ai operators and toolsets, and LangChain 1.x moves fast enough that its churn would start gating `common.ai` releases. Airflow already ships per-vendor providers (`anthropic`, `cohere`, `openai`, `pinecone`, `qdrant`, `weaviate`), so `apache-airflow-providers-langchain` fits that shape better.

Now is the cheap moment to decide. `common.ai` is 0.x and `lifecycle: incubation`, and the `langchain` conn type has only been out since 0.4.0. The toolset bridge from #67791 would stay here regardless, since it exposes `common.ai` toolsets as LangChain tools.

None of this should turn into you standing up a new provider before anything can merge. If that is where it lands, it gets scoped separately. The code here is clean and well tested.

providers/common/ai/src/airflow/providers/common/ai/hooks/langchain.py
```diff
@@ -171,3 +170,53 @@ def get_embedding_model(self) -> Embeddings:
...
+        return kwargs
```
kaxil, line 222: Nothing here can set `azure_deployment`. Azure deployment names are user-chosen, so they routinely differ from the model name, and the SDK routes on it:

```
azure_deployment=None            -> .../openai/deployments/gpt-4o/chat/completions
azure_deployment='gpt4o-prod-eu' -> .../openai/deployments/gpt4o-prod-eu/chat/completions
```

So anyone whose deployment is not named exactly after the model cannot use this hook. Deliberate?

providers/common/ai/provider.yaml
```diff
@@ -404,6 +404,41 @@ connection-types:
...
+        - login
+      relabeling:
+        password: API Key
+        host: Azure Endpoint
+      placeholders:
+        host: "https://<resource>.openai.azure.com"
+    conn-fields:
+      model:
+        label: Chat Model
+        description: "Chat model in azure_openai:name format (e.g. azure_openai:gpt-4o)."
```
kaxil, line 423: Nothing enforces this format. Putting `openai:gpt-4o` on a `langchain-azure` connection does not raise: `init_chat_model` transfers `azure_endpoint` and `api_version` into `model_kwargs` behind a `UserWarning` and hands back a plain `ChatOpenAI`, so it fails later at request time with an error that points nowhere useful. Worth defaulting the prefix, or validating it in the hook?

ColtenOuO:
Thanks a lot for the review, and for putting this much thought into the feedback!

To be honest, what got me started on this PR was this part of the docstring:

> Providers with bespoke auth (AWS Bedrock, Google Vertex AI / GenAI, Azure OpenAI, Cohere, HuggingFace) reject these kwargs; per-vendor subclasses can be added later mirroring the pydantic-ai pattern.

So I followed the pydantic-ai pattern and put together an Azure version, hoping to help extend it.

---

## What can't a user do today?

Honestly — nothing is broken. Users really can use Azure OpenAI today without an Azure hook.

As I mentioned above, because the docstring brings this up, I took it to mean that adding per-vendor subclasses was a planned direction for the future. (Or maybe I misread what that comment meant — please let me know if that's the case ><)

But what I can confirm is that there's no problem with the current usage at all.

---

## Why a new connection type instead of kwargs passthrough?

You're right that a plain allow-list of per-vendor keys would be considerably easier to maintain.

The advantage I see in the subclasses is on the UI side: when a user picks a vendor, we can show the fields that vendor actually needs, which makes the experience friendlier and gives them an easy way to see what LangChain vendors are supported right now.

image

As in the screenshot above — if we pick Azure as the vendor, both the Standard Fields and the Extra Fields immediately adapt to it and hint at the keys the user should fill in.

Maybe there's a way to ask the user to choose the vendor first when they select LangChain, and then list the corresponding Standard Fields and Extra Fields? If that were possible, I think the user experience would be even friendlier ><

That said, it would probably mean touching things inside `airflow-core`, so it may not be a good approach either.

I may well be overthinking this (sorry, I do that a lot xD) — if we set the UI experience aside, I agree that a plain allow-list is the better way to maintain this.

---

## If we do keep per-vendor subclasses, this may belong in its own provider

I agree on all three points.

---

I've also gone through all the code-level suggestions and I'm happy to handle them.

Before that, though, I'd like to wait until the direction on subclasses is settled before moving on to the next step.

Thanks again for raising these questions and suggestions!

ColtenOuO:
Since progress on common-ai is moving forward, this PR isn't adding much value anymore. Closing it to keep the Airflow PR list clean.

PR #70446, opened 2026-07-25
Add airflow dags stability check CLI command

202201142:
Add a new CLI command to detect non-deterministic DAG serialization by parsing and serializing DAGs twice and comparing the resulting hashes.

Non-deterministic DAG definitions can cause unnecessary `dag_version` changes, making deployments and code reviews harder. This command helps identify such DAGs and can be used in CI/CD to catch serialization instability before merging.

If the serialized hash differs between parses for the DAG, the command reports the differences between the serialized representations and exits with status code `1`.

 

---

---


bhavaniravi:
Is a CLI command really necessary?

I feel this belongs to lint rules. Eg., Ruff [AIR304](https://docs.astral.sh/ruff/rules/airflow3-dag-dynamic-value/) captures this exactly

r-richmond:
This CLI command is much more robust. And would be a welcome addition. 

As an anecdote we just got burned by a subtle issue that caused DAG version inflation. This was missed by the ruff rule but this CLI command would have caught it.

Personally I plan on using it in our CICD. Ty for the contribution

dipb-sl:
> Is a CLI command really necessary?
> 
> I feel this belongs to lint rules. Eg., Ruff [AIR304](https://docs.astral.sh/ruff/rules/airflow3-dag-dynamic-value/) captures this exactly

AIR304 is AST based checking so it will not cover all the cases. Simple example is: 
If you store `datetime.now()` in variable and then assign it to `start_time` Ruff check will not catch this issue if  assign `datetime.now` directly then only it can catch you can verify it.

PR #70448, opened 2026-07-25
Validate EmrAddStepsOperator job flow arguments after rendering

haseebmalik18:
`job_flow_id` and `job_flow_name` on `EmrAddStepsOperator` are template fields, so Jinja renders them after `__init__` runs. The mutual-exclusion check lived in `__init__`, which meant it validated the raw `{{ ... }}` expressions instead of the actual values. This moves the check into `execute()` so it runs after rendering.

Part of #70296.

---


shahar1:
Closing as false positive according to the up-to-date definition, see details here:
https://github.com/apache/airflow/pull/70505
Feel free to take on the remaining operators :)

PR #70449, opened 2026-07-25
Emit GCSToGCSOperator deprecation warnings after rendering

haseebmalik18:
`source_object`, `source_objects`, and `delimiter` on `GCSToGCSOperator` are template fields, so Jinja renders them after `__init__` runs. The wildcard/delimiter deprecation warnings were emitted in `__init__`, so they checked the raw `{{ ... }}` expressions instead of the rendered values. This moves them into a helper that runs at the start of `execute()`.
 
Part of #70296.

---

### DIFF
--- a/providers/google/tests/unit/google/cloud/transfers/test_gcs_to_gcs.py
+++ b/providers/google/tests/unit/google/cloud/transfers/test_gcs_to_gcs.py
@@ -24,7 +24,7 @@ import pytest
 
 from airflow.exceptions import AirflowProviderDeprecationWarning
 from airflow.providers.common.compat.openlineage.facet import Dataset
-from airflow.providers.common.compat.sdk import AirflowException
+from airflow.providers.common.compat.sdk import DAG, AirflowException
 from airflow.providers.google.cloud.transfers.gcs_to_gcs import WILDCARD, GCSToGCSOperator
 
 TASK_ID = "test-gcs-to-gcs-operator"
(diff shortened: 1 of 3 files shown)


providers/google/tests/unit/google/cloud/transfers/test_gcs_to_gcs.py
```diff
@@ -1082,22 +1077,41 @@ def test_execute_returns_list_of_destination_uris_single_file(self, mock_hook):
...
+                destination_bucket=DESTINATION_BUCKET,
+                destination_object=DESTINATION_OBJECT_PREFIX,
+                dag=dag,
+            )
+
+        operator.render_template_fields({"params": {"source_object": SOURCE_OBJECT_WILDCARD_FILENAME}})
+        assert operator.source_object == SOURCE_OBJECT_WILDCARD_FILENAME
+
+        with pytest.warns(AirflowProviderDeprecationWarning, match="Usage of wildcard"):
+            operator.execute(None)
```
potiuk: `execute(None)` works here because nothing reads the context before the warnings fire, but `mock.MagicMock()` or `{}` is more robust — if the warning block ever moves below something context-dependent, this fails with an unrelated `AttributeError` rather than the assertion you meant.

Also, on line 25: `from airflow import DAG` in a provider test. Provider tests run against multiple Airflow versions, where the usual import is via `tests_common` compat (or `airflow.sdk` on 3.x). Worth checking this doesn't trip the compat runs against older Airflow.

---

potiuk (approved):
Thanks — correct fix, and the best-tested version of this pattern I've seen in the current batch. `source_object`, `source_objects` and `delimiter` are all template fields, so emitting these deprecation warnings from `__init__` checked the raw `"{{ ... }}"` strings: a templated wildcard never warned, and a literal one warned at parse time for every Dag processor loop rather than once per task run.

`test_wildcard_deprecation_warning_uses_rendered_source_object` is the part that matters — it builds a real DAG and asserts the warning fires on the *rendered* value. That's the actual contract, and it's what distinguishes this from a mechanical move.

Pulling the three warnings into `_warn_on_deprecated_template_fields()` rather than inlining them in `execute` also keeps `execute` readable.

Note there's a near-identical PR in #70621, opened three days after this one. I've suggested there that it be closed in favour of this one — this came first and has the stronger test. No action needed from you.

Two small things inline.

---

haseebmalik18:
@potiuk Made requested changes

PR #70451, opened 2026-07-25
UI: Highlight selected Grid task square

dheerajturaga:
Highlight the selected task instance square in Grid view so it is easier to identify the currently open task.

The highlight is scoped to the selected Dag run and task, so selecting a task no longer highlights the same task across every visible Dag run column. Dark mode uses a lighter outline color for better contrast.

Tests added for selected task highlighting, dark-mode outline token selection, different-task behavior, and same-task/different-run behavior.

before: (Hard to easily see when you have a lot of tasks)
image

After:
image

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/airflow-core/src/airflow/ui/src/layouts/Details/Grid/GridTI.test.tsx
+++ b/airflow-core/src/airflow/ui/src/layouts/Details/Grid/GridTI.test.tsx
@@ -78,6 +78,10 @@ describe("GridTI", () => {
     renderGridTI(`/dags/example_dag/runs/${SELECTED_RUN_ID}/tasks/selected_task`);
 
     expect(screen.getByTestId("task-state-badge")).toHaveAttribute("data-selected", "true");
+    expect(screen.getByTestId("task-state-badge").closest("[data-task-id='selected_task']")).toHaveAttribute(
+      "data-selected",
+      "true",
+    );
   });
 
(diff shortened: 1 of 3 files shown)


shubhamraj-git (commented):
Thanks for the changes..

The dark mode looks good, but the light mode is not that different from the original if we have lot of grids, maybe we need to increase the border.

Image

dheerajturaga:
> Thanks for the changes..
> 
> The dark mode looks good, but the light mode is not that different from the original if we have lot of grids, maybe we need to increase the border.
> 
> Image

Make it thicker 2px
image

bbovenzi:
Adding a focus highlight makes sense. But why get rid of the row highlight?

dheerajturaga:
> Adding a focus highlight makes sense. But why get rid of the row highlight?

Good catch, That wasnt my intent. Let me fix this

dheerajturaga:
> Adding a focus highlight makes sense. But why get rid of the row highlight?

Done!
image

vatsrahul1001:
Moving to 3.4.0 as this new addition suits minot better

dheerajturaga:
@bbovenzi , @shubhamraj-git ok to merge?

bbovenzi (approved):

PR #70455, opened 2026-07-26
Speed up Celery task publishing with Redis pipelines

### LINKED ISSUE #8854: Send task to queue in bulk - Celery Executor
**Description**

Hello,
I recently took care of CeleryExecutor. I managed to optimize the status retrieval by using bulk operations. Instead of fetching the status for each task using a separate query, one is sent for all tasks. This has accelerated this process more than 100 times in many cases.
https://github.com/apache/airflow/pull/7542
However, we still use single requests in many processes to send tasks to the queue. This is very effective because of network latency.
https://github.com/apache/airflow/blob/f1dc2e0b0e358582c1df0cc07a5cc95fa721dc44/airflow/executors/celery_executor.py#L196-L206
It would be nice if it could be done as a bulk request in a single request.  For Redis, this means using Pipeline.
https://github.com/andymccurdy/redis-py#pipelines

Can it be done easily in Celery?

Best regards,
Kamil

shivaam:
CeleryExecutor currently publishes each task to Redis in a separate network
round trip. This makes publication time grow with both the number of scheduled
tasks and broker latency.

This change lets Celery and Kombu prepare each task message normally, then
buffers the final Redis queue writes in a non-transactional pipeline and
flushes them together. Callback workloads and non-Redis brokers keep the
existing publication path.

Redis pipelining is enabled by default and can be disabled with
`[celery] redis_pipelined_publish_enabled = False`. The implementation mirrors
Kombu's private Redis transport methods, so a focused contract test will fail
if a future Kombu update changes that integration point.

Local validation:

- Celery executor unit module: 74 passed, 6 skipped.
- Linux multiprocessing regression: passed 250 repeated dispatch cycles.
- PostgreSQL/Celery integration suite: 10 passed with real Redis and RabbitMQ
  brokers and Celery workers.
- Real daemon load: scheduler, API server, DAG processor, triggerer, PostgreSQL,
  Redis, and a four-process Celery worker completed 1,000/1,000 Bash tasks and
  four/four DAG runs successfully. Redis recorded exactly 1,000 queue writes
  and drained to zero.
- Changed source files pass `mypy`; regular `prek` checks pass. The
  provider-wide manual `mypy` hook was terminated by the local Docker resource
  limit (exit 137/143).

closes: #8854

---

- [X] Yes — Codex (GPT-5)

---


eladkal:
Can you clarify if this change is only for Redis or did you mention Redis because this is the broker you used for testing? Would that work with other brokers like SQS or RabbitMQ?

shivaam:
This optimization is Redis-only because it relies on Redis pipelines. RabbitMQ and SQS continue using the existing publication path, so their behavior remains unchanged.

---
Drafted-by: Codex (GPT-5); reviewed by @shivaam before posting

eladkal:
> This optimization is Redis-only because it relies on Redis pipelines. RabbitMQ and SQS continue using the existing publication path, so their behavior remains unchanged.
> 
> Drafted-by: Codex (GPT-5); reviewed by @shivaam before posting

Lets maybe add a table to the doc that clarifies which broker supports this

PR #70462, opened 2026-07-26
Expand $env:TEMP in Windows SSH remote job paths

### LINKED ISSUE #70438: SSHRemoteJobOperator: the default Windows remote_base_dir is never expanded, so the job directory cannot be created

Providers

### Apache Airflow version

main (`cd7235aff3`), `apache-airflow-providers-ssh` 6.0.0

`SSHRemoteJobOperator` cannot run against a Windows remote when `remote_base_dir` is
left at its default, because every generated PowerShell command embeds the default
base directory in a **single-quoted** string, where PowerShell does not expand
`$env:TEMP`.

`WINDOWS_DEFAULT_BASE_DIR` is defined as a PowerShell expression:

https://github.com/apache/airflow/blob/main/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py#L30

```python
WINDOWS_DEFAULT_BASE_DIR = "$env:TEMP\\airflow-ssh-jobs"
```

but `build_windows_wrapper_command` interpolates the resulting path into a
single-quoted assignment:

https://github.com/apache/airflow/blob/main/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py#L289-L290

```python
    wrapper = f"""$jobDir = '{job_dir}'
New-Item -ItemType Directory -Force -Path $jobDir | Out-Null
```

Single quotes are literal in PowerShell, so `$jobDir` holds the literal text
`$env:TEMP\airflow-ssh-jobs\...`. PowerShell then parses the leading `$env:` as a
PSDrive qualifier and fails.

**Reproduce without a remote host** — decode what the provider generates:

```python
from airflow.providers.ssh.utils.remote_job import RemoteJobPaths, build_windows_wrapper_command
import base64

paths = RemoteJobPaths(job_id="af_dag_task_run_try1_abcdef", remote_os="windows")
cmd = build_windows_wrapper_command(command="echo hello", paths=paths)
print(base64.b64decode(cmd.split("-EncodedCommand ")[1]).decode("utf-16-le"))
```

Output:

```powershell
$jobDir = '$env:TEMP\airflow-ssh-jobs\af_dag_task_run_try1_abcdef'
New-Item -ItemType Directory -Force -Path $jobDir | Out-Null
$log = '$env:TEMP\airflow-ssh-jobs\af_dag_task_run_try1_abcdef\stdout.log'
'' | Set-Content -Path $log
...
```

Running those two lines in Windows PowerShell 5.1 (Windows 11):

```
New-Item : Cannot find drive. A drive with the name '$env' does not exist.
    + CategoryInfo          : ObjectNotFound: ($env:String) [New-Item], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound,Microsoft.PowerShell.Commands.NewItemCommand
```

Nothing is created — neither a literal `$env:TEMP` folder nor anything under the real
`%TEMP%`. `Set-Content -Path $log` fails identically, so the job directory and log file
never exist and the job cannot start.

Cleanup is affected too, but silently: `build_windows_cleanup_command` produces

```powershell
Remove-Item -Recurse -Force -Path '$env:TEMP\airflow-ssh-jobs\...' -ErrorAction SilentlyContinue
```

which suppresses the same `DriveNotFoundException` and removes nothing.

Setting `remote_base_dir` to a literal path (for example `C:\Temp\airflow-ssh-jobs`)
avoids all of this, which is presumably why the default has not been hit before.

The default should resolve to the remote user's temporary directory.

The paths are used as literal data in single-quoted PowerShell strings, so the
expansion cannot happen at string-build time on the Airflow side — the value is only
known on the remote host. Two directions:

1. Resolve the temporary directory on the remote as part of the existing OS detection
   step (`build_windows_os_detection_command` already runs a probe) and use the
   returned literal path as the default base dir.
2. Or emit the assignment double-quoted (`$jobDir = "..."`) so PowerShell expands it,
   and quote the remainder appropriately. This is more invasive, since every path is
   currently treated as literal data and a user-supplied `remote_base_dir` containing
   `$` would then be interpreted rather than used verbatim.

Direction 1 looks safer to me because it keeps every emitted path literal and confines
the change to how the default is obtained.

### Operating System

Remote: Windows. Verified the PowerShell behaviour on Windows 11 with Windows PowerShell 5.1.

### Deployment

Other

### Deployment details

Not deployment specific — reproducible from the provider's command builders alone.

### Anything else?

Introduced with the operator in #60297. Occurs on every run against a Windows remote
that does not set `remote_base_dir`.

I could not find an existing issue or open PR covering this. The open PRs touching
`WINDOWS_DEFAULT_BASE_DIR` (#69834, #69885, #70098, #70233) all address #69813, which
is a different problem — that the cleanup guard compares against the default base dir
rather than the configured one.

### Code of Conduct

xFrone:
Closes: #70438

`WINDOWS_DEFAULT_BASE_DIR` is a PowerShell expression rather than a real path:

```python
WINDOWS_DEFAULT_BASE_DIR = "$env:TEMP\\airflow-ssh-jobs"
```

but every builder interpolated the resulting paths into **single-quoted** PowerShell
strings, where nothing is expanded. PowerShell then parses the leading `$env:` as a
PSDrive qualifier and fails:

```
New-Item : Cannot find drive. A drive with the name '$env' does not exist.
```

So `SSHRemoteJobOperator` could not create its job directory on a Windows remote at all
unless `remote_base_dir` was set to a literal path. Cleanup failed the same way, but
silently, because of `-ErrorAction SilentlyContinue`.

## Approach

The sentinel stays opaque on the Python side and is expanded only where a path is
rendered into PowerShell, through one helper:

```python
def _powershell_path(path: str) -> str:
    if path == WINDOWS_TEMP_PREFIX or path.startswith(WINDOWS_TEMP_PREFIX + "\\"):
        relative = path[len(WINDOWS_TEMP_PREFIX) :].lstrip("\\")
        if not relative:
            return WINDOWS_TEMP_PREFIX
        return f"(Join-Path $env:TEMP '{relative.replace(chr(39), chr(39) * 2)}')"
    return f"'{path.replace(chr(39), chr(39) * 2)}'"
```

so `$jobDir = '{job_dir}'` now emits

```powershell
$jobDir = (Join-Path $env:TEMP 'airflow-ssh-jobs\af_dag_task_run_try1_abcdef')
```

I picked this over resolving the temp directory remotely, or switching to expanding
quotes, because:

- **Every user-supplied path stays inert data.** Only the exact default prefix is
  treated as an expression, so a `$` inside a user's `remote_base_dir` is still used
  verbatim.
- **`_validate_job_dir` is untouched.** Both the job dir and the default base dir keep
  the sentinel on the Python side, so the existing prefix comparison still holds.
- **The operator, the trigger and the event payload are untouched.** Cleanup runs from
  `execute_complete` after deferral, where `_paths` is `None`, so a remotely-resolved
  base directory would have had to be threaded through the trigger event.

The change is confined to the six `build_windows_*` builders. The POSIX builders,
`RemoteJobPaths`, `WINDOWS_DEFAULT_BASE_DIR` and the operator's logic are unchanged.

Also corrected the `remote_base_dir` docstring, which documented the Windows default as
`C:\Windows\Temp\airflow-ssh-jobs` — a value that has never been the default and that
the operator's own `_validate_base_dir` warns about, since `C:\Windows` is in its
`dangerous_patterns` list.

## Verification

Ran the generated scripts in Windows PowerShell 5.1 (Windows 11).

Before, `New-Item` raised `DriveNotFoundException` and nothing was created anywhere —
not even a literal `$env:TEMP` directory. After, the emitted script creates
`C:\Users\<user>\AppData\Local\Temp\airflow-ssh-jobs\af_test_job_abcdef` and its
`stdout.log`, and the emitted cleanup command removes it.

Every generated script was also checked with
`[System.Management.Automation.Language.Parser]::ParseInput` and by binding the emitted
expressions to `Set-Content -Path`, `Move-Item -Path`/`-Destination` and
`Remove-Item -Path` on a real host.

## Tests

There are twelve path interpolation sites across the six builders, and four of them live
in a *second* base64 payload nested inside the wrapper. Tests that decode only the outer
layer pass even when those four are broken, so the new tests decode both layers and
sweep the four polling builders explicitly. I verified this by mutation: reverting any
individual site to the old single-quoted form makes a test fail.

Covered: the wrapper and the cleanup command expand the default; a custom base dir stays
a plain literal; a near-miss prefix such as `$env:TEMPORARY\jobs` is *not* expanded; and
a single quote is escaped correctly in both branches.

## Notes

- `$env:TEMP` is now resolved per SSH session — submission, polling, cleanup and
  `on_kill` each open their own connection. That is stable for ordinary Windows OpenSSH,
  and it is what makes the default work at all, but it is worth being aware of.
- Only `$env:TEMP` is special-cased. The `remote_base_dir` docstring now states that the
  value must be a literal path, so `remote_base_dir="$env:LOCALAPPDATA\jobs"` is not
  silently treated as an expression.
- No newsfragment: `providers/ssh` has no `newsfragments/` directory.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @xFrone** · by `@potiuk` · 2026-08-13 12:55 UTC
>
> - :x: **Merge conflicts**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/10_working_with_git.rst).
>
>
>

### DIFF
--- a/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
@@ -65,7 +65,10 @@ class SSHRemoteJobOperator(BaseOperator):
     :param remote_host: Override the host from the connection (templated)
     :param environment: Environment variables to set for the command (templated)
     :param remote_base_dir: Base directory for job artifacts (templated).
-        Defaults to /tmp/airflow-ssh-jobs on POSIX, C:\\Windows\\Temp\\airflow-ssh-jobs on Windows
+        Defaults to /tmp/airflow-ssh-jobs on POSIX, $env:TEMP\airflow-ssh-jobs on Windows.
+        Must be a literal path: it is emitted as a single-quoted PowerShell/shell string
+        and is not subject to environment-variable or other shell expansion (the default
+        above is the one built-in exception, expanded internally via ``Join-Path``).
     :param poll_interval: Seconds between status polls (default: 5)
     :param log_chunk_size: Max bytes to read per poll (default: 65536)
     :param timeout: Hard timeout in seconds for the entire operation
--- a/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
@@ -65,10 +65,9 @@ class SSHRemoteJobOperator(BaseOperator):
     :param remote_host: Override the host from the connection (templated)
     :param environment: Environment variables to set for the command (templated)
     :param remote_base_dir: Base directory for job artifacts (templated).
-        Defaults to /tmp/airflow-ssh-jobs on POSIX, $env:TEMP\airflow-ssh-jobs on Windows.
-        Must be a literal path: it is emitted as a single-quoted PowerShell/shell string
-        and is not subject to environment-variable or other shell expansion (the default
-        above is the one built-in exception, expanded internally via ``Join-Path``).
+        Defaults to ``/tmp/airflow-ssh-jobs`` on POSIX and
+        ``$env:TEMP\airflow-ssh-jobs`` on Windows. Custom values must be
+        literal paths and are not expanded by the remote shell.
     :param poll_interval: Seconds between status polls (default: 5)
     :param log_chunk_size: Max bytes to read per poll (default: 65536)
     :param timeout: Hard timeout in seconds for the entire operation
--- a/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
@@ -50,7 +50,7 @@ def _powershell_path(path: str) -> str:
     if path == WINDOWS_TEMP_PREFIX or path.startswith(WINDOWS_TEMP_PREFIX + "\\"):
         relative = path[len(WINDOWS_TEMP_PREFIX) :].lstrip("\\")
         if not relative:
-            return "$env:TEMP"
+            return WINDOWS_TEMP_PREFIX
         escaped_relative = relative.replace("'", "''")
         return f"(Join-Path $env:TEMP '{escaped_relative}')"
     escaped_path = path.replace("'", "''")
--- a/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
@@ -29,6 +29,33 @@ from typing import Literal
 POSIX_DEFAULT_BASE_DIR = "/tmp/airflow-ssh-jobs"
 WINDOWS_DEFAULT_BASE_DIR = "$env:TEMP\\airflow-ssh-jobs"
 
+# The default Windows base dir above is a PowerShell *expression*, not a literal path:
+# it must be expanded via Join-Path. Every other path is emitted as a single-quoted
+# PowerShell literal, where "$env:TEMP" is inert text -- PowerShell would try to resolve
+# a nonexistent "$env" PSDrive and fail (#70438). WINDOWS_TEMP_PREFIX lets
+# ``_powershell_path`` recognize that one sentinel and expand only its prefix, while a
+# user-supplied ``remote_base_dir`` (a real path) always stays a literal.
+WINDOWS_TEMP_PREFIX = "$env:TEMP"
+
+
+def _powershell_path(path: str) -> str:
+    """
+    Render a remote path as a PowerShell expression.
+
+    Paths are emitted as single-quoted literals so that nothing in them is interpreted.
+    The one exception is the default Windows base directory, which is a PowerShell
+    expression (``$env:TEMP``) rather than a real path: its prefix is expanded with
+    ``Join-Path`` while the rest stays literal.
+    """
+    if path == WINDOWS_TEMP_PREFIX or path.startswith(WINDOWS_TEMP_PREFIX + "\\"):
+        relative = path[len(WINDOWS_TEMP_PREFIX) :].lstrip("\\")
+        if not relative:
+            return "$env:TEMP"
+        escaped_relative = relative.replace("'", "''")
+        return f"(Join-Path $env:TEMP '{escaped_relative}')"
+    escaped_path = path.replace("'", "''")
+    return f"'{escaped_path}'"
+
 
 def _validate_job_dir(job_dir: str, remote_os: Literal["posix", "windows"]) -> None:
     """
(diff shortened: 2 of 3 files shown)


providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
```diff
@@ -65,7 +65,10 @@ class SSHRemoteJobOperator(BaseOperator):
     :param remote_host: Override the host from the connection (templated)
     :param environment: Environment variables to set for the command (templated)
     :param remote_base_dir: Base directory for job artifacts (templated).
-        Defaults to /tmp/airflow-ssh-jobs on POSIX, C:\\Windows\\Temp\\airflow-ssh-jobs on Windows
+        Defaults to /tmp/airflow-ssh-jobs on POSIX, $env:TEMP\airflow-ssh-jobs on Windows.
+        Must be a literal path: it is emitted as a single-quoted PowerShell/shell string
+        and is not subject to environment-variable or other shell expansion (the default
+        above is the one built-in exception, expanded internally via ``Join-Path``).
```
SameerMesiah97: This is too long. I would suggest the below:

```
Defaults to ``/tmp/airflow-ssh-jobs`` on POSIX and
    ``$env:TEMP\airflow-ssh-jobs`` on Windows. Custom values must be
    literal paths and are not expanded by the remote shell.
```

providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
```diff
@@ -29,6 +29,33 @@
...
+
+def _powershell_path(path: str) -> str:
+    """
+    Render a remote path as a PowerShell expression.
+
+    Paths are emitted as single-quoted literals so that nothing in them is interpreted.
+    The one exception is the default Windows base directory, which is a PowerShell
+    expression (``$env:TEMP``) rather than a real path: its prefix is expanded with
+    ``Join-Path`` while the rest stays literal.
+    """
```
SameerMesiah97, line 48: This docstring is slightly too long as well. Please see the below:

```
  """
  Render a path as a PowerShell expression.
  
  Most paths are emitted as literals. Paths under the built-in
  ``$env:TEMP`` base directory are emitted as expressions so the
  environment variable is expanded at runtime.
  """
```

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
```diff
@@ -176,6 +176,129 @@ def test_with_environment(self):
...
+class TestWindowsDefaultBaseDirExpansion:
+    """Regression tests for #70438.
+
+    ``$env:TEMP\\airflow-ssh-jobs`` is a PowerShell expression used as the default
+    Windows base directory, but every path is interpolated into a *single-quoted*
+    PowerShell string where ``$env:`` is not expanded -- it is emitted literally and
+    PowerShell fails trying to resolve a ``$env`` PSDrive. Only the default-base-dir
+    sentinel should be expanded (via ``Join-Path``); a user-supplied ``remote_base_dir``
+    must stay a literal, unexpanded string.
+    """
```
SameerMesiah97: Same here. I would just keep it one line:

`"""Regression tests for Windows default base directory expansion (#70438)."""`

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
```diff
@@ -176,6 +176,129 @@ def test_with_environment(self):
...
+
+    def _decode(self, cmd: str) -> str:
+        encoded_script = cmd.split("-EncodedCommand ")[1]
+        return base64.b64decode(encoded_script).decode("utf-16-le")
+
+    @staticmethod
+    def _decode_child(outer_script: str) -> str:
+        """Decode the *child* script nested inside a wrapper's outer script.
```
SameerMesiah97: I would keep only this line in the docstring. Extra detail should be kept in comments.

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
```diff
@@ -176,6 +176,129 @@ def test_with_environment(self):
...
+        """A base_dir that merely starts with the sentinel string as a substring --
+        ``$env:TEMPORARY`` is not ``$env:TEMP`` -- must NOT be mistaken for the sentinel.
+        It must stay a plain single-quoted literal, not be run through Join-Path."""
+        paths = RemoteJobPaths(job_id="test_job", remote_os="windows", base_dir="$env:TEMPORARY\\jobs")
+        wrapper = build_windows_wrapper_command("C:\\scripts\\test.ps1", paths)
+        decoded = self._decode(wrapper)
+
+        assert "Join-Path" not in decoded
+        assert "$jobDir = '$env:TEMPORARY\\jobs\\test_job'" in decoded
+
```
SameerMesiah97: These tests are solid but they can be improved further. Here is what I would suggest:

1) `test_custom_base_dir_stays_literal` and `test_near_miss_prefix_stays_literal` can be parametrized.

2) The docstrings in general are too long for these tests (and their helpers). It would be best for the docstring to be one line that expresses what the test is for. Keep in mind that not every single test needs a docstring, especially those where the intent is clear from the naming.

SameerMesiah97 (commented):
Looks fine to me. CI needs to be triggered. I have left some comments.

xFrone:
Pushed the docstring trims and the parametrized test.

CI is still blocked on the first-time-contributor workflow approval -- all four runs are
sitting at `action_required`, so a maintainer needs to trigger them.

providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
xFrone: Applied your wording verbatim.

providers/ssh/src/airflow/providers/ssh/utils/remote_job.py
xFrone: Applied your wording verbatim.

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
xFrone: Applied.

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
xFrone: Trimmed to the summary line. The detail about the nested child script moved into a comment above the split -- it explains why a second decode is needed at all, which is not obvious from the code.

providers/ssh/tests/unit/ssh/utils/test_remote_job.py
xFrone: Folded `test_near_miss_prefix_stays_literal` into `test_custom_base_dir_stays_literal` as a parametrized case, with ids `custom-base-dir` and `near-miss-prefix`, and cut the remaining docstrings in the class to one line.

Kept a short comment on the near-miss param: why `$env:TEMPORARY` matters is not obvious from the value alone.

PR #70464, opened 2026-07-26
Fix EcsRunTaskOperator deferred-logs region and template field init

fat-catTW:
Move `EcsRunTaskOperator` template-field decisions out of `__init__` so templated values are rendered before they affect ECS log handling.

This removes the operator from the `validate_operators_init` exemption list.

Besides moving the template-field logic out of ` __init__` (so templated awslogs_region/region_name values are rendered before they take effect), this also fixes a real bug in the deferrable path: `execute_complete` built its `AwsLogsHook` with `region_name=self.region_name`, so when `awslogs_region` was explicitly set to a different region than the task's region, the final log fetch after deferral queried CloudWatch in the wrong region and found no log stream. The new `resolve_awslogs_region()` helper is now the single place the log region is resolved (explicit `awslogs_region`, falling back to `region_name`), and both `execute_complete` and the task log fetcher use it. Note this covers only the post-deferral fetch — log forwarding during deferral still uses the task region inside `TaskDoneTrigger`; that remaining half is tracked in https://github.com/apache/airflow/issues/70465.

related: #70296

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/operators/ecs.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/operators/ecs.py
@@ -491,9 +491,6 @@ class EcsRunTaskOperator(EcsBaseOperator):
         self.reattach = reattach
         self.number_logs_exception = number_logs_exception
 
-        if self.awslogs_region is None:
-            self.awslogs_region = self.region_name
-
         self.arn: str | None = None
         self.container_name: str | None = container_name
         self._started_by: str | None = None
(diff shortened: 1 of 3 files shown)


shahar1 (approved):

shahar1:
Nice! You might haven't notice, but you probably fixed another bug as well :)

fat-catTW:
Thanks! 
I missed that at first too, but yes, this also makes `execute_complete()` honor `awslogs_region` in deferrable mode instead of always using `region_name` when fetching the final CloudWatch log event.

PR #70469, opened 2026-07-26
Fix DateTimeSensorAsync crash on templated target_time with start_from_trigger

### LINKED ISSUE #70284: DateTimeSensorAsync crashes Dag parsing with templated target_time + start_from_trigger=True

burakcoleman:
`DateTimeSensorAsync.__init__` computes `start_trigger_args.trigger_kwargs["moment"]` by calling
`timezone.parse(self.target_time)` at Dag-parse time, before Jinja templating has run. `target_time`
is a documented template field (e.g. `"{{ data_interval_end.tomorrow().replace(hour=1) }}"`), so a
templated value raises a raw `pendulum.parsing.exceptions.ParserError` there, crashing parsing of the
entire Dag file, not just the one task.

This can't be fixed by moving the parse into `execute()`: when `start_from_trigger=True`, the scheduler
defers the task straight from `TaskInstance.defer_task()` using the pre-computed `self.task.start_trigger_args`
— `execute()` never runs for this path. So the value genuinely has to be resolved in `__init__`, the same
constraint `TimeSensor` hit in #69610.

Fix: catch the parse failure and raise a clear `ValueError` explaining that `start_from_trigger=True`
requires a static `target_time`, instead of letting the raw `ParserError` propagate and take down the
whole Dag file. Static/ISO `target_time` values are unaffected and keep working exactly as before.

Added tests covering both the static-time (still works) and templated-time (now raises a clear,
task-scoped error instead of crashing Dag parsing) cases.

closes: #70284
Related: #69610

- [x] Yes — Claude Code

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -21,6 +21,8 @@ import datetime
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, NoReturn
 
+from pendulum.parsing.exceptions import ParserError
+
 from airflow.providers.common.compat.sdk import BaseSensorOperator, timezone
 from airflow.providers.standard.triggers.temporal import DateTimeTrigger
 from airflow.providers.standard.version_compat import AIRFLOW_V_3_0_PLUS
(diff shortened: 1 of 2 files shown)


raphaelauv:
same than https://github.com/apache/airflow/pull/70310 ?

burakcoleman:
Good catch, thanks — #70310 already fixes the same issue (#70284) the same way. Closing this one as a duplicate.

PR #70470, opened 2026-07-26
UI: Add a time-window scope to the Dags list run state filter

yuseok89:
Follow-up to #69875, the run-state filter UX refactor that reviewers asked to split into its own PR.

The Dags list had two separate run-state filters ("Last run" and "Any run") side by side. This merges them into a single Run state control with a lookback selector, and adds an optional time window so the run-state filters can target a recent period instead of an unbounded run history.

  related: #67094

### What changed

- UI: ~~The two run-state selects become one~~ The Dags list's separate `Last run` and `Any run` pills become one `Run state` ~~select~~ pill plus a lookback: `Latest run` / `Last 24 hours` / `Last 7 days` / `Last 30 days` / `Any time`. Picking a state defaults the lookback to Latest run, so today's behavior is unchanged. The lookback maps onto the existing `last_dag_run_state` / `dag_run_state` params; a time window adds `dag_run_state_within_hours`.
- Backend: `_AnyDagRunStateFilter` gains an optional `dag_run_state_within_hours` bound, applied inside the existing correlated `EXISTS`.

### Performance

I benchmarked the `dag_run_state` + `within_hours` filter on Postgres against 10M `dag_run` rows (5,000 Dags × 2,000 `success` runs each, `run_after` spread uniformly over the last ~2 years), measuring the filter's `EXISTS` query with `EXPLAIN (ANALYZE)` on a warm cache while sweeping the window from "no window" through 7 / 30 / 90 / 365 days.

Only the **"last 7 days"** window was slow — about **1.3 s**; every other window ran in **under 100 ms**. The narrower the window, the more runs the query has to scan per Dag to find one recent enough, and `run_after` isn't covered by the current index, so each of those is a heap lookup.

A covering index `(dag_id, state, run_after)` turns that scan into a direct seek and brings the "last 7 days" case down to about **45 ms**. On every other window it made **no meaningful difference** — those were already under 100 ms, and the small deltas are within run-to-run noise.

This PR does not add that index.

### Demo

- 200 Dags × 5,000 runs → ~1M `dag_run` rows
- ~30% failed / ~70% success
- `run_after` spread over the last year

https://github.com/user-attachments/assets/35fe0b83-6f60-4609-a42b-0aa9a58cf0fb

### After the rebase onto #71731
The two run-state pills are now a single one, with the lookback inside it.

https://github.com/user-attachments/assets/a78c7fdf-635b-44ba-83b4-1897c5ffa0e3

---

  - Opus 4.8

---


airflow-core/src/airflow/ui/src/pages/DagsList/DagsList.tsx
```diff
@@ -314,6 +316,7 @@ export const DagsList = () => {
     dagDisplayNamePattern: Boolean(dagDisplayNamePattern) ? dagDisplayNamePattern : undefined,
     dagRunsLimit,
     dagRunState,
+    dagRunStateWithinHours: Boolean(dagRunState) && withinHours !== null ? Number(withinHours) : undefined,
```
bbovenzi: Let's make sure that `Number(withinHours)` is returning a valid number.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/RunStateScopeSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const SCOPE_OPTIONS: ReadonlyArray<{ labelKey: string; value: RunStateScope }> = [
+  { labelKey: "latestRun", value: "latest" },
+  { labelKey: "last24Hours", value: "24" },
+  { labelKey: "last7Days", value: "168" },
+  { labelKey: "last30Days", value: "720" },
+  { labelKey: "anyTime", value: "any" },
+];
+
+export const RunStateScopeSelect = ({ dataTestId, onChange, triggerProps, value }: Props) => {
```
bbovenzi: Nit. I think this variable name is confusing. Let's do `RunStateLookbackSelect`

airflow-core/src/airflow/ui/public/i18n/locales/en/dags.json
```diff
@@ -24,6 +24,15 @@
...
     "runIdPatternFilter": "Search Dag Runs",
+    "runScope": {
+      "anyTime": "Any time",
+      "in": "in",
+      "last7Days": "Last 7 days",
+      "last24Hours": "Last 24 hours",
+      "last30Days": "Last 30 days",
+      "latestRun": "Latest run"
+    },
+    "runState": "Run state",
```
bbovenzi: We have a lot of these translation keys already. The dashboard page has a time range selector.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/DagsFilters.tsx
```diff
@@ -67,14 +70,20 @@ export const DagsFilters = () => {
   const showPaused = searchParams.get(PAUSED_PARAM);
   const showFavorites = searchParams.get(FAVORITE_PARAM);
   const needsReview = searchParams.get(NEEDS_REVIEW_PARAM);
-  const state = searchParams.get(LAST_DAG_RUN_STATE_PARAM);
-  const activeRunState = searchParams.get(DAG_RUN_STATE_PARAM);
+  const lastRunState = searchParams.get(LAST_DAG_RUN_STATE_PARAM);
+  const anyRunState = searchParams.get(DAG_RUN_STATE_PARAM);
+  const anyRunStateWindow = searchParams.get(DAG_RUN_STATE_WITHIN_HOURS_PARAM);
```
bbovenzi: Let's call everything a "window", a "period", or a "lookback" but let's not mix our variable names or "scope".

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/runStateFilter.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  return { dagRunState: state, dagRunStateWithinHours: scope };
+};
+
+export const runScopeFor = (
+  lastRunState: string | null,
+  anyRunState: string | null,
+  anyRunStateWindow: string | null,
+): RunStateScope =>
+  lastRunState !== null || anyRunState === null ? "latest" : ((anyRunStateWindow ?? "any") as RunStateScope);
```
bbovenzi: Let's make sure we validate the run state window is a correct number here too.

airflow-core/src/airflow/api_fastapi/common/parameters.py
```diff
@@ -1473,19 +1473,29 @@ def depends(cls, has_pending_actions: bool | None = Query(None)) -> _PendingActi
...
-            .where(DagRun.dag_id == DagModel.dag_id, DagRun.state == self.value)
-            .exists()
-        )
+        # EXISTS seeks the (dag_id, state) index per Dag rather than scanning the whole table; the
+        # optional run_after bound is not covered by that index, so it is filtered within each Dag's
+        # matching rows (still bounded per Dag, not a full scan).
+        conditions = [DagRun.dag_id == DagModel.dag_id, DagRun.state == self.value]
+        if self.within_hours is not None:
+            conditions.append(DagRun.run_after >= timezone.utcnow() - timedelta(hours=self.within_hours))
+        has_run_in_state = sql_select(DagRun.dag_id).where(*conditions).exists()
```
bbovenzi: Without  new index and for time ranges, I think we should invert this subquery to look at the dag runs in the range first instead of checking the history of every each dag.

```
        bound = timezone.utcnow() - timedelta(hours=self.within_hours)
        dag_ids_in_window = (
            sql_select(DagRun.dag_id)
            .where(DagRun.run_after >= bound, DagRun.state == self.value)
            .distinct()
        )
```

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/RunStateLookbackSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+  readonly onChange: (value: RunStateScope) => void;
+  readonly triggerProps?: ChakraSelect.TriggerProps;
+  readonly value: RunStateScope;
+};
+
+const SCOPE_OPTIONS: ReadonlyArray<{ labelKey: string; value: RunStateScope }> = [
+  { labelKey: "latestRun", value: "latest" },
+  { labelKey: "last24Hours", value: "24" },
+  { labelKey: "last7Days", value: "168" },
+  { labelKey: "last30Days", value: "720" },
```
bbovenzi, line 40: We should try to reuse the time ranges that the dashboard and overview page already use. I could see a world where we move it into the airflow config so deployment managers can decide based on how often their runs fire. Some teams do every minute, others every day.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsList.tsx
yuseok89: Thanks.
Added a guard so only a valid positive number is used.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/RunStateScopeSelect.tsx
yuseok89: Thanks.
Renamed to `RunStateLookbackSelect`, and standardized on "lookback" everywhere.

airflow-core/src/airflow/ui/public/i18n/locales/en/dags.json
yuseok89: Thanks.
Moved the labels into the shared `common:timeRange` namespace.
Reused the existing `last24Hours` and added the missing `last7Days`/`last30Days`.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/DagsFilters.tsx
yuseok89: Thanks.
Everything is on "lookback" now.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/runStateFilter.ts
yuseok89: Thanks.
`runLookbackFor` now only accepts a known lookback value; anything else falls back to "any".
Added tests for the invalid cases.

airflow-core/src/airflow/ui/src/pages/DagsList/DagsFilters/RunStateLookbackSelect.tsx
yuseok89: The time-range labels are now reused from the shared `common:timeRange` namespace.
Making the ranges configurable is a nice idea. I was thinking it might be better as a separate follow-up rather than part of this PR. What do you think?

airflow-core/src/airflow/api_fastapi/common/parameters.py
yuseok89: Thanks.
I did a quick benchmark of the inverted approach on Postgres (10M `dag_run` rows, warm cache). Rough numbers:

| within_hours | current | inverted (run_after-first) |
|---|---:|---:|
| none | < 50 ms | < 50 ms |
| 7 days | ~1 s | ~1 s |
| 30 days | < 100 ms | ~4 s |
| 90 days | < 50 ms | ~7 s |
| 365 days | < 50 ms | ~5 s |

At least in this setup, the inverted version scans the whole `run_after` slice, so it seemed to get worse as the window widens.

The one thing that held up across every window (tens of ms) was a covering index `(dag_id, state, run_after)`.

I know you wanted to avoid a new index.
Just flagging this in case it changes the trade-off.
What do you think?

bbovenzi:
Great. Let's rebase after https://github.com/apache/airflow/pull/71731 was merged

yuseok89:
Rebased onto main.
#71731 changed the filter architecture quite a bit, so I reworked the run state filter as a single FilterBar pill. I also dropped the filter configs and i18n keys that became unused, let me know if that cleanup feels too aggressive.
I will update the PR description to match.

Issue #53410, opened 2025-07-16
Sensitive data from extra fields is visible in Edit Connection window


CarbonRK:
### Apache Airflow version

3.0.3

### What happened?

After upgrading Airflow from version 3.0.2 to 3.0.3 sensitive data in extra fields and extra fields JSON is visible in edit connection page.

### What you think should happen instead?

Sensitive data should be masked with ***

### How to reproduce

Enter Edit Connection window in Airflow UI.

### Operating System

Ubuntu 24.04

### Deployment

Other Docker-based deployment

### Deployment details

Airflow deployed with Docker Swarm.
Settings in airflow.cfg:

- hide_sensitive_var_conn_fields = True

I also added the following after realising the issue occurred:: 
- sensitive_var_conn_names = private_key,private_key_passphrase
However, this did not change anything.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

potiuk:
I think we could potentially mask sensitive data in fields, but masking them in JSON is likely not a good idea. Not a bug BTW. This is a new feature. When you edit connections, you should be able to see all fields (including sensitive ones) - password is only masked in order to avoid someone standing behind your back read the password - but other than that you can reveal and see it anyway - the settings you refer to are only for masking thing in logs. But yeah, I guess in the UI we could also mask sensitive fields to avoid this scenario of someone behind your back looking at your screen while you are editing the connection. New feature. not a bug. Marking as "good first issue".

CarbonRK:
Oh, that's interesting what you wrote, @potiuk. When we had Airflow 3.0.2 deployed, the fields 'private_key' and 'private_key_passphrase' were actually masked with '***', so I thought it was a new feature compared to Airflow 2.8.4, which we had deployed earlier. However, after upgrading to 3.0.3, the field masking disappeared, so I thought it was a bug in the newer Airflow version.
I'll try deploying a test instance of 3.0.2 to see what happens.

potiuk:
Interesting. Yes

CarbonRK:
> Interesting. Yes

Hey @potiuk, just a quick update: I've deployed the test instance of Airflow 3.0.2, and here's what the Edit Connection window looks like after adding a new connection.
Now, a question: Is hiding those fields a feature or a bug in version 3.0.2? :P
[...]

PR #70473, opened 2026-07-26
Add password field type to FlexibleForm

Ed3110:
Extra fields that a provider declares with `format: "password"` are rendered as
plain text in the connection form. `FieldSelector` has no branch for that format,
so they fall through to `FieldString`, which renders an Input with no type.
The standard `password` field has been masked with a toggle for a while;
declared extra fields never got the same treatment.

This adds a `FieldPassword` component to `FlexibleForm` and the matching branch in
`FieldSelector`, reusing the (`FiEye`/`FiEyeOff`) toggle pattern from
ConnectionStandardFields. `FieldSelector` is shared with the Dag trigger form, so
params declared with `format: "password"` are masked there too.

Verified on a Salesforce connection — Security Token, Consumer Secret and Private
Key mask correctly and the toggle works, other extra fields are unaffected — and
on a Dag param form. Tests cover the masked input, value storage, clearing, and
the toggle.

This does not cover free-form keys a provider doesn't declare, 
such as `private_key` on an SSH connection, which is what the reporter hit.
Those have no schema and no format, so the UI has no signal to go on; 
masking them would need the backend to mark sensitive keys. 
Happy to look at that separately.

before:
image
image

after:
image

related: #53410

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
+++ b/airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { Input } from "@chakra-ui/react";
+import { Input, InputGroup } from "@chakra-ui/react";
 import { useState } from "react";
 
 import { PasswordToggle } from "src/components/PasswordToggle";
@@ -39,7 +39,11 @@ export const FieldPassword = ({ name, namespace = "default", onUpdate }: Flexibl
   };
 
   return (
-    <div style={{ position: "relative", width: "100%" }}>
+    <InputGroup
+      endElement={
+        <PasswordToggle isVisible={showPassword} onToggle={() => setShowPassword(!showPassword)} />
+      }
+    >
       <Input
         autoComplete="new-password"
         disabled={disabled}
@@ -54,14 +58,6 @@ export const FieldPassword = ({ name, namespace = "default", onUpdate }: Flexibl
         type={showPassword ? "text" : "password"}
         value={(param.value ?? "") as string}
       />
-      <PasswordToggle
-        isVisible={showPassword}
-        onToggle={() => setShowPassword(!showPassword)}
-        position="absolute"
-        right={2}
-        top="50%"
-        transform="translateY(-50%)"
-      />
-    </div>
+    </InputGroup>
   );
 };
--- a/airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
+++ b/airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
@@ -18,7 +18,6 @@
  */
 import { Input } from "@chakra-ui/react";
 import { useState } from "react";
-import { useTranslation } from "react-i18next";
 
 import { PasswordToggle } from "src/components/PasswordToggle";
 import { paramPlaceholder, useParamStore } from "src/queries/useParamStore";
@@ -27,7 +26,6 @@ import type { FlexibleFormElementProps } from ".";
 
 export const FieldPassword = ({ name, namespace = "default", onUpdate }: FlexibleFormElementProps) => {
   const [showPassword, setShowPassword] = useState(false);
-  const { t: translate } = useTranslation("components");
   const { disabled, paramsDict, setParamsDict } = useParamStore(namespace);
   const param = paramsDict[name] ?? paramPlaceholder;
   const handleChange = (value: string) => {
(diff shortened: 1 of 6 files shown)


Ed3110:
Thanks for kicking off CI @potiuk

The Firefox UI e2e job failed at the `Setup pnpm` step (`self-installer
exits with code 1`) before any code ran. 
The Chromium and WebKit jobs running the same suite both passed, so it doesn't look related to this PR. 
Could it be re-run when convenient?

potiuk:
> Thanks for kicking off CI @potiuk
> 
> The Firefox UI e2e job failed at the `Setup pnpm` step (`self-installer exits with code 1`) before any code ran. The Chromium and WebKit jobs running the same suite both passed, so it doesn't look related to this PR. Could it be re-run when convenient?

Re-run it - but if this is still failing rebasing to latest main (and waiting for maintainer review when it's green) is the right next step for you

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    setParamsDict(paramsDict);
+    onUpdate(value);
+  };
+
+  return (
+    <>
+      <div style={{ position: "relative", width: "100%" }}>
+        <Input
+          disabled={disabled}
```
bbovenzi:
```suggestion
          autoComplete="new-password"
          disabled={disabled}
```

We need to prevent browsers from trying to autofill

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
```diff
@@ -0,0 +1,89 @@
...
+            cursor: "pointer",
+            position: "absolute",
+            right: "10px",
+            top: "50%",
+            transform: "translateY(-50%)",
+          }}
+          type="button"
+        >
+          {showPassword ? <FiEye size={15} /> : <FiEyeOff size={15} />}
+        </button>
```
bbovenzi: This is a copy of ConnectionStandardFields.tsx. Let's extract that out into a shared component instead of copying code. Also, we should probably move this to a chakra Button so that it can pick up our theming.

Ed3110: Thanks for the review, I'll get on it.

Ed3110: I think I've got it. The toggle is now in `src/components/PasswordToggle.tsx`
and both FieldPassword and StandardFields use it. I built it on `ui/IconButton`,
which seemed like the right wrapper, and it also gives a tooltip and an
aria-label.

I left `showPassword` in the callers since the `Input`'s `type` needs it, but
let me know if you'd rather it lived somewhere else.

I also moved `hidePassword`/`showPassword` out of `flexibleForm` into a new
`passwordToggle` section in `components.json`. That file looked like it's keyed
by component, but I wasn't sure, so happy to move them.

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+      {param.schema.examples ? (
+        <datalist id={`list_${name}`}>
+          {param.schema.examples.map((example) => (
+            <option key={example} value={example}>
+              {example}
+            </option>
+          ))}
+        </datalist>
+      ) : undefined}
```
bbovenzi: Remove this.

We do not need example passwords.

airflow-core/src/airflow/ui/src/components/PasswordToggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const PasswordToggle = ({ isVisible, onToggle }: Props) => {
+  const { t: translate } = useTranslation("components");
+
+  return (
+    <IconButton
+      label={isVisible ? translate("passwordToggle.hide") : translate("passwordToggle.show")}
+      onClick={onToggle}
+      position="absolute"
```
bbovenzi: Let's remove this absolute positioning. It should be handled by the parent comonent, via extending the icon button props `type Props {} & IconButtonProps;`

airflow-core/src/airflow/ui/src/pages/Connections/ConnectionStandardFields.tsx
bbovenzi, line 60: While we are at it. Can we also turn off autocomplete here?

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
Ed3110: Done.

airflow-core/src/airflow/ui/src/pages/Connections/ConnectionStandardFields.tsx
Ed3110: Done, thanks!

airflow-core/src/airflow/ui/src/components/PasswordToggle.tsx
Ed3110: Done — `Props` now extends `IconButtonProps` and spreads the rest onto `IconButton`. Both call sites pass the positioning themselves now.

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
```diff
@@ -0,0 +1,67 @@
...
+      // "undefined" values are removed from params, so we set it to null to avoid falling back to DAG defaults.
+      paramsDict[name].value = value === "" ? null : value;
+    }
+
+    setParamsDict(paramsDict);
+    onUpdate(value);
+  };
+
+  return (
+    <div style={{ position: "relative", width: "100%" }}>
```
bbovenzi: Oh we can also use chakra's `InputGroup` here like:

```
<InputGroup endElement={<PasswordToggle isVisible={showPassword} onToggle={…} />}>
  <Input autoComplete="new-password" type={showPassword ? "text" : "password"} … />
</InputGroup>
```

Then we don't need all of this styling

Ed3110: Done, thanks!

bbovenzi (approved):

airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldPassword.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    setParamsDict(paramsDict);
+    onUpdate(value);
+  };
+
+  return (
+    <InputGroup
+      endElement={
+        <PasswordToggle isVisible={showPassword} onToggle={() => setShowPassword(!showPassword)} />
+      }
```
bbovenzi:
```suggestion
      endElement={<PasswordToggle isVisible={showPassword} onToggle={() => setShowPassword(!showPassword)} />}
```

PR #70474, opened 2026-07-26
Fix EcsRunTaskOperator deferred logs read from the wrong region

### LINKED ISSUE #70465: EcsRunTaskOperator deferrable mode fetches CloudWatch logs from wrong region when awslogs_region differs from region_name

Providers

### Apache Airflow version

main (development)

When `EcsRunTaskOperator` runs in deferrable mode with `awslogs_region` set to a different region than `region_name`, the CloudWatch logs forwarded **during** deferral are fetched from the wrong region.

`EcsRunTaskOperator.execute` passes only `region_name=self.region_name` to `TaskDoneTrigger`:

https://github.com/apache/airflow/blob/main/providers/amazon/src/airflow/providers/amazon/aws/operators/ecs.py#L553-L564

and `TaskDoneTrigger.run` uses that single region for **both** the ECS waiter client and the CloudWatch `get_log_events` client:

https://github.com/apache/airflow/blob/main/providers/amazon/src/airflow/providers/amazon/aws/triggers/ecs.py#L190-L223

So `_forward_logs` queries `awslogs_group`/`log_stream` in the task's region instead of the logs region. The stream is not found there, and the trigger logs "Tried to get logs from stream ... but it didn't exist (yet)" for the whole task duration; no task logs are forwarded while deferred.

**Steps to reproduce:**

1. Create an ECS task definition whose `awslogs` driver ships logs to a CloudWatch log group in region B, while the task runs in region A.
2. Run an `EcsRunTaskOperator` with `deferrable=True`, `region_name="<region A>"`, `awslogs_region="<region B>"`, `awslogs_group` and `awslogs_stream_prefix` set.
3. Observe that no container logs are relayed during deferral, only the "didn't exist (yet)" warnings.

The trigger should read logs from `awslogs_region` (falling back to `region_name` when unset), matching the non-deferrable path.

This is the remaining half of the log-region resolution fixed in #70464: that PR made the post-deferral fetch in `execute_complete` use `resolve_awslogs_region()`, but the in-deferral forwarding in `TaskDoneTrigger` still uses the ECS region. The likely fix is a separate log-region parameter on `TaskDoneTrigger` (the single `region_name` is used for two different clients), defaulted for backward compatibility with in-flight serialized triggers.

### Operating System

Not Applicable

### Deployment

Other

amazon

apache-airflow-providers-amazon (main)

Not Applicable

### Anything else?

Found while reviewing #70464.

### Code of Conduct

---

ColtenOuO:
When `EcsRunTaskOperator` runs with `deferrable=True` and the task definition ships its container logs to a different region than the one the task runs in (`awslogs_region` != `region_name`), no logs are forwarded during deferral -- the trigger just repeats "Tried to get logs from stream ... but it didn't exist (yet)" for the whole task duration.

`TaskDoneTrigger` was given a single `region_name` and used it for two different clients: the ECS waiter (correct) and the CloudWatch `get_log_events` client (wrong -- the log group lives in the other region).

The operator never passed the logs region to the trigger, so the trigger had no way to tell them apart.

This adds a `log_region_name` parameter to `TaskDoneTrigger`, used only for the `AwsLogsHook`, and has the operator pass `resolve_awslogs_region()` when deferring. The `EcsHook` keeps using `region_name`.

The parameter defaults to `None` and falls back to `region_name`, so triggers that were already serialized before this change deserialize without the key and keep their current behaviour.

This is the remaining half of the log-region resolution: #70464 fixed the post-deferral fetch in `execute_complete`, and the non-deferrable path already used `resolve_awslogs_region()` via `_get_task_log_fetcher()`.

closes: #70465

---

- [X] Yes — Claude Code (Opus 5) for writing Tests

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/triggers/ecs.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/triggers/ecs.py
@@ -128,6 +128,9 @@ class TaskDoneTrigger(BaseTrigger):
         Will fail after that many unsuccessful attempts.
     :param aws_conn_id: The Airflow connection used for AWS credentials.
     :param region_name: The AWS region where the cluster is located. Used to build the hook.
+    :param log_region_name: The AWS region where the CloudWatch logs are stored. Defaults to
+        ``region_name`` when not set, which is correct unless the task definition ships its logs
+        to another region.
     :param verify: Whether or not to verify SSL certificates. Used to build the hook.
     :param botocore_config: Configuration dictionary for the botocore client. Used to build the hook.
     :param region: (deprecated) use ``region_name`` instead.
@@ -146,6 +149,7 @@ class TaskDoneTrigger(BaseTrigger):
         verify: bool | str | None = None,
         botocore_config: dict | None = None,
         region: str | None = None,
+        log_region_name: str | None = None,
     ):
         if region is not None:
             warnings.warn(
@@ -167,6 +171,7 @@ class TaskDoneTrigger(BaseTrigger):
 
         self.log_group = log_group
         self.log_stream = log_stream
+        self.log_region_name = log_region_name
 
     def serialize(self) -> tuple[str, dict[str, Any]]:
         return (
@@ -178,6 +183,7 @@ class TaskDoneTrigger(BaseTrigger):
                 "waiter_max_attempts": self.waiter_max_attempts,
                 "aws_conn_id": self.aws_conn_id,
                 "region_name": self.region_name,
+                "log_region_name": self.log_region_name,
                 "log_group": self.log_group,
                 "log_stream": self.log_stream,
                 "verify": self.verify,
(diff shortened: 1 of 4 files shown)


providers/amazon/src/airflow/providers/amazon/aws/triggers/ecs.py
```diff
@@ -167,6 +171,9 @@ def __init__(
 
         self.log_group = log_group
         self.log_stream = log_stream
+        # Triggers serialized before this parameter existed deserialize with None, keeping the
+        # previous behaviour of reading logs from the cluster region.
+        self.log_region_name = log_region_name if log_region_name is not None else region_name
```
SameerMesiah97: Since `log_region_name` is immediately resolved to `region_name` when omitted, `self.log_region_name` is never actually `None`. Would it make more sense to perform this fallback when constructing the `AwsLogsHook` instead, so the original parameter value is preserved? I think either approach works, but handling the fallback at the point of use avoids one parameter implicitly altering the value of another during initialization

SameerMesiah97 (commented):
Looks good to me. I just have one comment but its minor.

providers/amazon/src/airflow/providers/amazon/aws/triggers/ecs.py
ColtenOuO: Makes sense, I've moved the fallback to where the `AwsLogsHook` is built, so `log_region_name` now keeps the value the caller passed.

Thanks for pointing that!

shahar1 (approved):

PR #63454, opened 2026-03-12
feat(kubernetes): add executor callback support to KubernetesExecutor

sjyangkevin:
related: #62887 

### Summary

This PR implements executor callback support for KubernetesExecutor, continuing the work started in #61153 and #62645. Synchronous callbacks are routed to an executor to run as workloads.

### Approach

Callbacks run as Kubernetes pods, the same isolation model used for tasks. The workload (an ExecuteCallback JSON payload) is passed to the `execute_workload` entrypoint that tasks use, and the pod's exit code determines success or failure.

The implementation threads callbacks through the existing executor pipeline with annotation-based discrimination to distinguish callback pods from task pods:

### 1. Scheduling

The scheduler job calls `queue_workload(ExecuteCallback, ...)`. `_process_workloads` moves it onto
`task_queue` as a `KubernetesJob`, sets `event_buffer[key] = CallbackState.QUEUED`, and
adds the key to `running`.

### 2. Pod creation

`sync()` pops the job from `task_queue` and calls `kube_scheduler.run_next()`. When
`run_next` detects an `ExecuteCallback` workload, it routes to `_run_next_callback`, which:

- Builds a pod using the standard base template
- Annotates it with `{"callback_id": <uuid>}` — the sole identifier the watcher uses to
  correlate the pod back to the callback
- Labels it with `airflow-worker=<scheduler_job_id>` so `KubernetesJobWatcher` discovers it
- Applies the user's `pod_mutation_hook` (same as task pods)
- Serializes the `ExecuteCallback` workload to JSON and passes it as the container command
  via `execute_workload`

### 3. Execution inside the pod

The pod runs `python -m airflow.sdk.execution_time.execute_workload --json-string <...>`.
`execute_workload` deserializes the workload (Pydantic discriminated union handles both
`ExecuteTask` and `ExecuteCallback`) and calls `BaseExecutor.run_workload`, which dispatches
to `supervise_callback()`. The supervisor forks a subprocess that imports and runs the
callback function at the configured import path.

### 4. Result collection

`KubernetesJobWatcher` streams pod events from the K8s API. When it sees a callback pod
(identified by the presence of `callback_id` in annotations), it builds
`{"callback_id": <uuid>}` as the annotation dict instead of the task annotation dict.
`annotations_to_key` returns the UUID string directly as the `WorkloadKey`. The watcher
puts a `KubernetesWatch` on `watcher_queue`.

`AirflowKubernetesScheduler.sync()` converts it to `KubernetesResults` and puts it on
[...]

PR #63491, opened 2026-03-12
Unify executor workload queues

anishgirianish:
---

##### Was generative AI tooling used to co-author this PR?

- [ ] Yes (please specify the tool below)

  ## Summary
Refactors `BaseExecutor` workload queue management so that adding a new workload type is a small.

Follows the direction proposed by @ferruzzi  https://github.com/apache/airflow/pull/62343#discussion_r2127826554.

## Problem

Adding a new workload type (like ExecuteCallback or TestConnection) required touching ~6 places in BaseExecutor: a new queue dict, a new `supports_*` flag,slots calculation, an isinstance branch in queue_workload, a dedicated scheduling method, and isinstance branches in dequeue/trigger logic. Each provider executor that overrode `queue_workload` also needed updating. This made extending the executor interface unnecessarily painful.

## What this does

Replaces the per-type queue dicts and boolean capability flags with three primitives:

- `executor_queues`: one `defaultdict(dict)` keyed by `WorkloadType`, replacing `queued_tasks`, `queued_callbacks` and `queued_connection_tests`.
- `supported_workload_types`: a `frozenset[WorkloadType]` replacing `supports_callbacks` and `supports_connection_test`.
- `WORKLOAD_TYPE_PRIORITY` plus a per-schema `sort_key`: the type registry decides ordering across types, the schema decides ordering within a type.

`queue_workload` in the base class is now generic: validate the type, store by key. `trigger_tasks` and `trigger_connection_tests` are merged into `trigger_workloads`, which drains all queue types through a single per-heartbeat budget.

## Adding a new workload type after this refactor

1. Declare the `WorkloadType` member and place it in `_workload_type_priority_order`.
2. Add the schema to the `All` and `ExecutorWorkload` discriminated unions, and override `sort_key` if it needs ordering within its group.
3. Add a branch to `BaseExecutor.run_workload` and to `state_class_for_key`.
4. Add the type to `supported_workload_types` on supporting executors and handle it in `_process_workloads`.

---

* Read the **[Pull Request Guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-guidelines)** for more information. Note: commit author/co-author name and email in commits become permanently public when merged.
* For fundamental code changes, an Airflow Improvement Proposal ([AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/Airflow+Improvement+Proposals)) is needed.
[...]

PR #70475, opened 2026-07-26
Add callback support on kubernetes executors

JH-A-Kim:
---

# Description
Related PR's: #67449 
#63454
Related bug fix PR: #70148 

To reiterate the main points this pr aims to implement support for callbacks on the Kubernetes executor because it currently only knows how to handle task instances. This PR makes it so it can also handle `ExecuteCallback` workloads. These are keyed by `CallbackKey` so that callbacks can execute as their own pod instead of requiring another mechanism.

The core changes are the following:

- New Generic types in kubernetes_executor_types.py where WorkloadKey/WorkloadState/WorkloadCommand type aliases help to widen the executors internal types from always being a task to now being able to be a task or a callback and is gated behind `AIRFLOW_V_3_3_PLUS` so that behaviour older versions of Airflow can remain unchanged
- The routing in `queue_workload` and `process_workloads` now branch on the the workload is either a `ExecuteTask` or `ExecuteCallback`
- Pods now have another build path for `ExecuteCallback` called `_run_next_callback` which builds and submits the pod itself and returns `None`.
- Now with this there is a new pod constructor called `PodGenerator.construct_callback_pod` which builds a minimal pod for a callback and tags it with a `callback_id` id annotation and a `airflow-workload-type: callback` label so that the watcher can tell it apart from task pods
- For the watcher when the pod events annotation contains a `callback_id` the watcher reconstructs a `CallbackKey` instead of a `TaskInstanceKey` and state resolution treats a callback with no error as `CallbackState.SUCCESS` rather than querying the (nonexistent) `TaskInstance` row.
- For pod adoption after a scheduler restart, it just re-labels it with a new schedulers worker ID, since callback pods dont participate in the normal orphan recovery/retry logic that task pods do.

Generated-by: Sonnet 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


JH-A-Kim:
@ferruzzi Currently one of the big bug fixes was in `_change_state()` where there would be a timeout because callbacks would enter into the task retry logic for the specific case of before execution a failure would occur and would then check for `dag_id` which it does not have and continuously add it back to the retry queue. Which would cause tests to timeout. But do you think its worth adding logic to restart the callback pod if that were to occur? Currently my check just makes sure entrance into that particular bit of logic is only for tasks. But if the pod were to fail before running I think it might be necessary for it to try and restart in that case automatically.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -805,9 +852,26 @@ def _change_state(
...
+                filter_for_tis = TaskInstance.filter_for_tis([key])
+                if filter_for_tis is not None:
+                    state = session.scalar(select(TaskInstance.state).where(filter_for_tis))
+                else:
+                    state = None
+                state = TaskInstanceState(state) if state else None
+            elif AIRFLOW_V_3_3_PLUS:
+                from airflow.utils.state import CallbackState
+
+                state = CallbackState.SUCCESS
```
ferruzzi, line 872: Shouldn't this be wrapped in a state check if `Callback.state` is still `QUEUED`?  Assuming SUCCESS feels dangerous.

ferruzzi:
@JH-A-Kim It seems worth doing, but maybe in a different PR, this one is already pretty chonky and I don't believe any execujtors currently retry callbacks.  I think the right answer there is to convert `_get_task_instance_state` to `_get_workload_state` which checks `Callback.state` directly; `Callback.state` already has `ACTIVE_STATES` and `TERMINAL_STATES`` defined which you can check against.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
JH-A-Kim: From how I understand it the callback state is never updated during the running so it will always be in the `QUEUED` state so adding the check wouldnt really change the behaviour of the code as far as i am aware

JH-A-Kim: And there is no state row in callback states for the db because there is a lot more cases that tasks can be than callbacks with callbacks only having 6 potential states and tasks having 12ish potential states.

JH-A-Kim: But if we do add the check it could help to catch anything in the future?

JH-A-Kim:
> @JH-A-Kim It seems worth doing, but maybe in a different PR, this one is already pretty chonky and I don't believe any execujtors currently retry callbacks. I think the right answer there is to convert `_get_task_instance_state` to `_get_workload_state` which checks `Callback.state` directly; `Callback.state` already has `ACTIVE_STATES` and `TERMINAL_STATES`` defined which you can check against.

yeah that makes sense ill create an issue when I get the time to ask about potentially adding retry logic to callbacks, or should I create a discussion on the devlist?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
ferruzzi: Alright, I've been thinking about this.   It looks like this is fine how it is for now, but it may need some more thinking later.  Something feels off with the callback state flow here, but I can't place it right now.   

Go ahead and resolve this when you read it.  If you do go on to do the callback-retry work, maybe spend some time thinking this through... it seems like there's a step missing here with the state updates compared to the task state updates.

ferruzzi:
> yeah that makes sense ill create an issue when I get the time to ask about potentially adding retry logic to callbacks, or should I create a discussion on the devlist?

An Issue is fine for now, but mark is as dependent on this PR and on #63491 both being merged, that way we can reduce merge conflicts all around.

kubernetes-tests/tests/kubernetes_tests/test_kubernetes_executor.py
```diff
@@ -239,3 +244,296 @@ def test_pod_failure_logging_non_failed_state(self, mock_log):
...
+        self._delete_airflow_pod("scheduler")
+        self.ensure_resource_health("airflow-scheduler")
+        print(f"[{dag_id}] Scheduler restarted; waiting for callback pod to complete.")
+
+        if pre_restart_phase in ("Succeeded", "Deleted"):
+            # Pod already completed before the restart — the test goal (pod completion
+            # despite scheduler lifecycle) is still met.
+            return
+
+        # The slow callback sleeps for 30s; allow plenty of time for completion.
```
ferruzzi: Shouldn't we also assert that the Callback table is updated along the way?

airflow-core/src/airflow/example_dags/example_deadline_callback.py
```diff
@@ -0,0 +1,127 @@
...
+            callback=SyncCallback("airflow.example_dags.example_deadline_callback.failing_deadline_callback"),
+        ),
+    ) as dag_failing:
+
+        @task
+        def dummy_task_failing():
+            """Placeholder task for the failing-callback DAG."""
+            print("dummy_task_failing executed")
+
+        dummy_task_failing()
```
SameerMesiah97, line 127: Given that this DAG covers multiple scenarios rather than 'standard usage', I wonder whether this belongs in example dags at all. Could we keep the example DAG focused on demonstrating Kubernetes executor callback support and move the scheduler-restart/failure fixtures into the relevant integration test setup?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -1034,6 +1098,19 @@ def adopt_launched_task(
...
+
+            try:
+                kube_client.patch_namespaced_pod(
+                    name=pod.metadata.name,
+                    namespace=pod.metadata.namespace,
+                    body={"metadata": {"labels": {"airflow-worker": new_worker_id_label}}},
+                )
+            except ApiException as e:
+                self.log.info("Failed to adopt pod %s. Reason: %s", pod.metadata.name, e)
+            return
```
SameerMesiah97, line 1113: This seems inconsitent with what has been done for the other executors (for e.g. `LambdaExecutor`) wrt adding callback support. . Do we actually want callback workloads to participate in scheduler adoption here, or should the lifecycle semantics be consistent across executors?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor.py
```diff
@@ -1186,7 +1262,13 @@ def _adopt_completed_pods(self, kube_client: client.CoreV1Api) -> None:
...
                 continue
 
-            ti_id = annotations_to_key(pod.metadata.annotations)
+            ti_id: TaskInstanceKey | CallbackKey
+            if AIRFLOW_V_3_3_PLUS and CALLBACK_POD_ANNOTATION_KEY in pod.metadata.annotations:
+                from airflow.models.callback import CallbackKey
+
+                ti_id = CallbackKey(id=pod.metadata.annotations[CALLBACK_POD_ANNOTATION_KEY])
+            else:
+                ti_id = annotations_to_key(pod.metadata.annotations)
```
SameerMesiah97, line 1271: The above comment on adopting callbacks applies here too.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor_types.py
```diff
@@ -45,8 +69,8 @@ class FailureDetails(TypedDict, total=False):
 class KubernetesResults(NamedTuple):
     """Results from Kubernetes task execution."""
```
SameerMesiah97, line 70:
```suggestion
    """Results from Kubernetes workload execution."""
```

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor_utils.py
```diff
@@ -576,46 +594,76 @@ def _health_check_kube_watchers(self):
...
         kube_image = next_job.kube_image or self.kube_config.kube_image
 
+        # Callback workloads follow a separate pod-construction path.
+        if AIRFLOW_V_3_3_PLUS and len(command) == 1:
+            from airflow.executors.workloads import ExecuteCallback
+
+            workload = command[0]
+            if isinstance(workload, ExecuteCallback):
+                self._run_next_callback(workload.key, workload, pod_template_file)
+                return None
```
SameerMesiah97, line 642: Why does `_build_pod_request` now also submit the pod for callback workloads? This previously had a clear side-effect-free contract where it only constructed and returned the pod, with submission handled by the caller. Could `_run_next_callback` just construct and return the callback pod so we retain that separation and callbacks can follow the same pod creation path as tasks?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor_utils.py
```diff
@@ -728,6 +776,42 @@ def _close_async_pod_client(self) -> None:
...
+    ) -> None:
+        """Build and submit a callback pod for an ExecuteCallback workload."""
+        base_worker_pod = self._get_base_worker_pod(pod_template_file)
+
+        # Extract dag_id and run_id from the standardised log_path:
+        # "executor_callbacks/<dag_id>/<run_id>/<callback_id>"
+        log_parts = (workload.log_path or "").split("/")
+        dag_id = log_parts[1] if len(log_parts) >= 4 else ""
+        run_id = log_parts[2] if len(log_parts) >= 4 else ""
+
```
SameerMesiah97, line 790: It seems rather odd to parse critical metadata from logs. Should we be deriving `dag_id` and `run_id` by parsing `workload.log_path` here? what if the format changes? Is this approach not too brittle? If the callback pod requires this metadata, should it be exposed directly by the workload instead?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/pod_generator.py
```diff
@@ -461,6 +461,81 @@ def construct_pod(
...
+            CALLBACK_POD_ANNOTATION_KEY,
+            CALLBACK_WORKLOAD_TYPE_KEY,
+        )
+
+        # Derive a k8s-safe pod name from the callback UUID.
+        short_id = callback_id.replace("-", "")[:8]
+        pod_id = add_unique_suffix(name=f"callback-{short_id}", rand_len=8, max_len=POD_NAME_MAX_LENGTH)
+
+        try:
+            image = base_worker_pod.spec.containers[0].image
```
SameerMesiah97, line 489: Is the precedence here intentional? `kube_image` is effectively only a fallback since we copy the image from `base_worker_pod` into the dynamic pod whenever it exists. This also seems different from the normal task pod construction where `kube_image` is passed into the dynamic pod and the normal reconciliation determines the final result. Could we follow the same approach here?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/pod_generator.py
```diff
@@ -461,6 +461,81 @@ def construct_pod(
...
+            if not image:
+                image = kube_image
+        except Exception:
```
SameerMesiah97, line 492: Exception here is too broad. Not necessarily unjustified but I wish to know why a narrower exception was not used/

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/executors/kubernetes_executor_utils.py
```diff
@@ -165,16 +171,25 @@ def _run(
...
-                "run_id": annotations.get("run_id"),
-                "try_number": annotations["try_number"],
-            }
-            map_index = annotations.get("map_index")
-            if map_index is not None:
-                task_instance_related_annotations["map_index"] = map_index
+            if AIRFLOW_V_3_3_PLUS and CALLBACK_POD_ANNOTATION_KEY in annotations:
+                # Callback pod: forward only the annotations that process_watcher_task needs.
+                # Callback pods carry callback_id instead of task_id/try_number.
+                task_instance_related_annotations = {
```
SameerMesiah97, line 177: nit: could be `workload_related_annotations` to be more accurate.

kubernetes-tests/tests/kubernetes_tests/test_kubernetes_executor.py
```diff
@@ -239,3 +244,296 @@ def test_pod_failure_logging_non_failed_state(self, mock_log):
...
+        phase = self._wait_for_pod_phase(pod_name, ["Succeeded", "Failed", "Deleted"], timeout=180)
+        assert phase in ("Succeeded", "Deleted"), (
+            f"Callback pod {pod_name!r} reached {phase!r} after scheduler restart (expected Succeeded/Deleted)"
+        )
```
SameerMesiah97, line 539: Not sure if this tests actually covers what it is supposed to. Consdier a situation where the new scheduler never adopts the callback pod. The pod would continue running independently and eventually reach `Succeeded` anyway, so this test would still pass. Should we explicitly verify that the callback pod is adopted by the new scheduler and that its terminal state is processed?

That's if we decide that callbacks should be adopted.

kubernetes-tests/tests/kubernetes_tests/test_kubernetes_executor.py
```diff
@@ -239,3 +244,296 @@ def test_pod_failure_logging_non_failed_state(self, mock_log):
...
+        print(f"[{dag_id}] dag_run_id={dag_run_id}")
+
+        pod = self._wait_for_callback_pod(dag_run_id, timeout=120)
+        pod_name = pod["metadata"]["name"]
+        print(f"[{dag_id}] callback pod appeared: {pod_name}")
+
+        try:
+            # Failed pods are NOT auto-deleted (delete_worker_pods_on_failure=False default).
+            phase = self._wait_for_pod_phase(pod_name, ["Failed"], timeout=120)
+            assert phase == "Failed", f"Expected Failed phase, got {phase!r}"
```
SameerMesiah97, line 460: The test description says we're verifying that the watcher emits FAILED, but asserting the Kubernetes pod phase only verifies that the callback container failed. Is there a way to assert that the failed result was actually processed by the executor?

SameerMesiah97 (commented):
I have left some comments.

aaron-y-chen (commented):
Hi, I think there's a potential bug that might needs your help to fix. 🙂

On `KubernetesExecutor` (Airflow 3.4.0, k8s v1.30.13), if the scheduler restarts while an `ExecuteCallback` pod from `example_deadline_callback.py` is still running, the scheduler afterwards crashes on every start and never recovers on its own.

### How to reproduce

1. Deploy Airflow on K8s with `KubernetesExecutor`.
2. Raise the sleep in `slow_deadline_callback` from `30` to `300` (or more longer) so there is time to act.
3. Trigger `example_deadline_callback_slow`.
4. Wait until the callback pod is `Running`
   (`kubectl get pods -l airflow-workload-type=callback`).
5. While it is still running, restart the scheduler
   (`kubectl delete pod -l component=scheduler`).
6. Wait for the callback pod to reach `Succeeded`, then allow up to one
   `orphaned_tasks_check_interval` (default 300s), the crash does not start immediately.

From then on the scheduler crash-loops: every restart dies within ~20s and it never recovers on its own. The only way out was deleting the leftover callback pod by hand, and nothing in the logs or the UI points at which pod to delete.

Image

The scheduler log shows:

```
File ".../executors/kubernetes_executor.py", line 1271, in _adopt_completed_pods
    ti_id = annotations_to_key(pod.metadata.annotations)
File ".../kubernetes_helper_functions.py", line 171, in annotations_to_key
    try_number = int(annotations["try_number"])
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

The callback also never reaches a terminal state. Its pod logs `"event":"Callback executed successfully"`, but `callback.state` stays `queued` indefinitely:

Image

Note the DAG run and its task instance are both `success`, so this is not visible from the UI, only the callback result is lost.

Image

JH-A-Kim:
Ok so for those of you that are invested in this pr I have recently arrived back home from a conference and am moving for work so I dont have a lot of bandwith at the current moment. I will get to this as soon as I can but it will likely be around a week and a half to two weeks before I can really hunker down and get back into this and resolve some of the comments. 😢

JH-A-Kim:
Ok well update time! UPS lost my router so I don't have internet :( the SOONEST my internet provider can get me a new router is the 11TH of September (currently surviving on data and a dream)

ferruzzi:
Any update?  It's the last executor left :smile:

PR #70480, opened 2026-07-26
WaitSensor - time_to_wait templated

raphaelauv:
time_to_wait should be templated so we can do

```python
from airflow.sdk import DAG, Param
from airflow.providers.standard.sensors.time_delta import WaitSensor
with DAG(
        dag_id="my_dag",
        schedule=None,
        catchup=False,
        params={
            "duration_minutes": Param(
                default=30,
                type="integer",
                minimum=1,
                description="Duration in minutes to wait.",
            )
        },
):

    WaitSensor(
        task_id="wait_for_duration",
        time_to_wait="{{ params.duration_minutes }}",
        deferrable=True
    )

```

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/time_delta.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/time_delta.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import warnings
+from collections.abc import Sequence
 from datetime import datetime, timedelta
 from time import sleep
 from typing import TYPE_CHECKING, Any
@@ -177,6 +178,8 @@ class WaitSensor(BaseSensorOperator):
     :param deferrable: Run sensor in deferrable mode
     """
 
+    template_fields: Sequence[str] = ("time_to_wait",)
+
     def __init__(
         self,
         time_to_wait: timedelta | int,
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Thanks — and this one lands on the right side of a distinction that's been catching people out.

Making `time_to_wait` templated requires moving the `int -> timedelta` conversion out of `__init__`, because at construction time the value may still be `"{{ ... }}"`. That's a **value** transformation, so deferring it to execute is correct — unlike a provision check ("was this argument passed?"), which per #70296 must stay in the constructor. Worth saying explicitly since several PRs in that campaign have moved the wrong kind.

I checked the one thing that can quietly break when a class declares `template_fields`: `BaseSensorOperator` doesn't define any, and neither does `TimeDeltaSensor` in the same module, so this is purely additive — nothing inherited gets shadowed.

`_resolve_time_to_wait` handling `timedelta` passthrough and `int(value)` for both native ints and rendered strings is the right shape, and the parametrisation over `timedelta(minutes=1)`, `1`, and `"{{ 1*2 }}"` covers all three input forms that can now reach it.

One small thought, not blocking: `test_wait_sensor_templating_error` pins Python's raw message, `invalid literal for int() with base 10: 'nothing'`. That's accurate but it's CPython's wording rather than ours, so the test is coupled to something we don't control. Wrapping the conversion to raise something like `ValueError(f"time_to_wait must be a number of minutes or a timedelta, got {value!r}")` would give users a better message *and* give the test something stable to assert.

---

PR #70482, opened 2026-07-27
Validate AppflowBaseOperator source after template rendering

mitre88:
`source` is one of `AppflowBaseOperator`'s template fields, but it was validated against `SUPPORTED_SOURCES` inside `__init__`, which runs before Jinja rendering — so a templated `source` was rejected at Dag parse time with the raw expression in the error message. This moves the check to the top of `execute()`, where the rendered value is available, and removes the class from the exemption list.

related: #70296

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/operators/appflow.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/operators/appflow.py
@@ -84,8 +84,6 @@ class AppflowBaseOperator(AwsBaseOperator[AppflowHook]):
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
-        if source is not None and source not in SUPPORTED_SOURCES:
-            raise ValueError(f"{source} is not a supported source (options: {SUPPORTED_SOURCES})!")
         self.filter_date = filter_date
         self.flow_name = flow_name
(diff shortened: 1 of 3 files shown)


uranusjr:
The test should also validate a templated value actually passes the test. The current test also passed without the change to implementation.

mitre88:
Thanks for the review! Added `test_templated_source_passes_validation_after_rendering`: the operator is constructed with `source="{{ var.value.appflow_source }}"` (which raised `ValueError` at parse time before this change) and executes successfully once the field holds the rendered value. Verified the new test fails against the previous implementation and passes with this one.

One note on scope: the `AppflowRun{Full,Before,After,Daily}Operator` subclasses still have their own `source`/`filter_date` checks in `__init__` (they are not in the exemption list since they inherit `template_fields` rather than declaring them), so a templated `source` still fails at parse time through those. Happy to move those in a follow-up PR if that is welcome.

---

mitre88:
Closing — #70440 merged a more complete version of this: it moves the base-class check to execute time and also converts the `AppflowRun*` subclass checks to overridable `_validate_source`/`_validate_filter_date` hooks (the follow-up gap noted above). Nothing left for this PR to add.

---

PR #70483, opened 2026-07-27
Check EmrAddStepsOperator cluster arguments after template rendering

mitre88:
`job_flow_id` and `job_flow_name` are template fields of `EmrAddStepsOperator`, but the "exactly one of" mutual-exclusion check ran in `__init__`, before Jinja rendering. This moves the check to the top of `execute()` (relocated verbatim), updates the existing test to assert the failure at execute time, and removes the class from the exemption list.

related: #70296

---


uranusjr:
This is checking against any argument usages, not the templated results.

mitre88:
Understood — since the check guards any usage of the argument rather than the rendered value, it belongs in `__init__`. Thanks for the review!

---

PR #70484, opened 2026-07-27
Check GCSToS3Operator match_glob support after template rendering

mitre88:
`match_glob` is a template field of `GCSToS3Operator`, but the guard that rejects it on `apache-airflow-providers-google<10.3.0` ran in `__init__`, before Jinja rendering. A templated `match_glob` therefore failed at Dag parse time even when the rendered value would be empty. The provider-version probe stays in `__init__` (it does not touch template fields); only the raise moves to `execute()`, relocated verbatim. Removes the class from the exemption list.

related: #70296

---


uranusjr:
This should still be in `__init__`. The check is against _any_ usage of the argument, not the templated result.

mitre88:
Makes sense — the version-support check applies to any usage of `match_glob`, not the rendered value. Thanks for the review!

---

PR #70487, opened 2026-07-27
Resolve AwsToAwsBaseOperator destination connection after template rendering

mitre88:
`source_aws_conn_id` and `dest_aws_conn_id` are template fields of `AwsToAwsBaseOperator`, but the fallback ("dest not set → use source") was resolved in `__init__`, before Jinja rendering. `__init__` now keeps plain assignments (also dropping a duplicated `source_aws_conn_id` assignment) and the fallback lives in a `resolved_dest_aws_conn_id` property, which `DynamoDBToS3Operator` (the only in-repo consumer) reads at execute time. Removes the class from the exemption list.

related: #70296

---


uranusjr (approved):

vincbeck (approved):

mitre88:
Closing — #70445 merged an equivalent fix for this class (a `dest_aws_conn_id` property with the fallback in the getter, which also keeps existing call sites working), removed the exemption entry, and covers the fallback with tests. Nothing left for this PR to add. Thanks for the reviews!

---

PR #70488, opened 2026-07-27
Resolve GCSFileTransformOperator destination fallbacks after rendering

mitre88:

`GCSFileTransformOperator` resolves `destination_bucket`/`destination_object` fallbacks (`destination_bucket or source_bucket`) in `__init__`, where the source fields still hold un-rendered Jinja expressions. This moves the fallback resolution to `execute()` (and mirrors it in `get_openlineage_facets_on_start`, which runs before `execute`), so the fallback applies to the rendered values.

Scope notes from review:
- The `delimiter` deprecation warning in `GCSListObjectsOperator` stays in `__init__` — deprecation applies to any usage of the argument, not the rendered value.
- The `objects`/`prefix` provision checks in `GCSDeleteObjectsOperator` stay in `__init__` — they are false positives under the relaxed validator definition proposed in #70505. Both classes keep their exemption entries; only the `GCSFileTransformOperator` entry is removed.

Note: `TestGCSDeleteObjectsOperator::test_get_openlineage_facets_on_start[objects]` fails on a pristine `main` checkout as well — pre-existing and unrelated to this change.

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/gcs.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/gcs.py
@@ -603,8 +603,8 @@ class GCSFileTransformOperator(GoogleCloudBaseOperator):
         super().__init__(**kwargs)
         self.source_bucket = source_bucket
         self.source_object = source_object
-        self.destination_bucket = destination_bucket or self.source_bucket
-        self.destination_object = destination_object or self.source_object
+        self.destination_bucket = destination_bucket
+        self.destination_object = destination_object
 
         self.gcp_conn_id = gcp_conn_id
         self.transform_script = transform_script
@@ -612,6 +612,8 @@ class GCSFileTransformOperator(GoogleCloudBaseOperator):
         self.impersonation_chain = impersonation_chain
 
     def execute(self, context: Context) -> None:
+        self.destination_bucket = self.destination_bucket or self.source_bucket
+        self.destination_object = self.destination_object or self.source_object
         hook = GCSHook(gcp_conn_id=self.gcp_conn_id, impersonation_chain=self.impersonation_chain)
 
         with NamedTemporaryFile() as source_file, NamedTemporaryFile() as destination_file:
@@ -658,8 +660,8 @@ class GCSFileTransformOperator(GoogleCloudBaseOperator):
             name=self.source_object,
         )
         output_dataset = Dataset(
-            namespace=f"gs://{self.destination_bucket}",
-            name=self.destination_object,
+            namespace=f"gs://{self.destination_bucket or self.source_bucket}",
+            name=self.destination_object or self.source_object,
         )
 
         return OperatorLineage(inputs=[input_dataset], outputs=[output_dataset])
--- a/scripts/ci/prek/validate_operators_init_exemptions.txt
+++ b/scripts/ci/prek/validate_operators_init_exemptions.txt
@@ -17,7 +17,6 @@ providers/google/src/airflow/providers/google/cloud/operators/cloud_storage_tran
 providers/google/src/airflow/providers/google/cloud/operators/dataproc.py::DataprocCreateClusterOperator
 providers/google/src/airflow/providers/google/cloud/operators/dataproc.py::DataprocSubmitJobOperator
 providers/google/src/airflow/providers/google/cloud/operators/functions.py::CloudFunctionDeployFunctionOperator
-providers/google/src/airflow/providers/google/cloud/operators/gcs.py::GCSFileTransformOperator
 providers/google/src/airflow/providers/google/cloud/sensors/bigquery_dts.py::BigQueryDataTransferServiceTransferRunSensor
 providers/google/src/airflow/providers/google/cloud/sensors/cloud_composer.py::CloudComposerExternalTaskSensor
 providers/google/src/airflow/providers/google/cloud/transfers/azure_fileshare_to_gcs.py::AzureFileShareToGCSOperator


providers/google/src/airflow/providers/google/cloud/operators/gcs.py
```diff
@@ -229,18 +229,18 @@ def __init__(
         super().__init__(**kwargs)
         self.bucket = bucket
         self.prefix = prefix
-        if delimiter:
-            warnings.warn(
-                "Usage of 'delimiter' is deprecated, please use 'match_glob' instead. Planned removal date: October 5, 2026.",
-                AirflowProviderDeprecationWarning,
-                stacklevel=2,
-            )
```
uranusjr: I think this should still be in `__init__`.

mitre88: Agreed — the deprecation applies to any usage of the argument, not the rendered value. Moved the warning back to `__init__` as it is on main, restored the `GCSListObjectsOperator` exemption entry, and dropped the execute-time warning test. The other two changes (objects/prefix validation in `GCSDeleteObjectsOperator` and destination fallbacks in `GCSFileTransformOperator`) are unchanged.

---

providers/google/src/airflow/providers/google/cloud/operators/gcs.py
```diff
@@ -311,25 +311,20 @@ def __init__(
...
-            err_message = "(Task {task_id}) Either objects or prefix should be set. Both are None.".format(
-                **kwargs
-            )
-            raise ValueError(err_message)
-        if objects is not None and prefix is not None:
-            err_message = "(Task {task_id}) Objects or prefix should be set. Both provided.".format(**kwargs)
-            raise ValueError(err_message)
-
-        super().__init__(**kwargs)
-
```
shahar1, line 333: Please restore the changes here, as it will be flagged as false-positive (see #https://github.com/apache/airflow/pull/70505 for more details)

providers/google/src/airflow/providers/google/cloud/operators/gcs.py
```diff
@@ -311,25 +311,20 @@ def __init__(
...
     def execute(self, context: Context) -> None:
+        if self.objects is None and self.prefix is None:
+            raise ValueError(f"(Task {self.task_id}) Either objects or prefix should be set. Both are None.")
+        if self.objects is not None and self.prefix is not None:
+            raise ValueError(f"(Task {self.task_id}) Objects or prefix should be set. Both provided.")
+
```
shahar1, line 331: Please revert these changes

scripts/ci/prek/validate_operators_init_exemptions.txt
```diff
@@ -25,8 +25,6 @@ providers/google/src/airflow/providers/google/cloud/operators/cloud_storage_tran
 providers/google/src/airflow/providers/google/cloud/operators/dataproc.py::DataprocCreateClusterOperator
 providers/google/src/airflow/providers/google/cloud/operators/dataproc.py::DataprocSubmitJobOperator
 providers/google/src/airflow/providers/google/cloud/operators/functions.py::CloudFunctionDeployFunctionOperator
-providers/google/src/airflow/providers/google/cloud/operators/gcs.py::GCSDeleteObjectsOperator
```
shahar1: Please restore this line

shahar1 (dismissed):
Please focus this PR only on GCSFileTransformOperator (see #https://github.com/apache/airflow/pull/70505 for more details)

mitre88:
Done — reverted the `GCSDeleteObjectsOperator` changes (code, test, and exemption entry restored to match main) per the relaxed provision-check definition in #70505. The PR now only moves the `GCSFileTransformOperator` destination fallbacks to execute time; title and description updated accordingly.

---

mitre88:
@shahar1 Done — the PR is now scoped to `GCSFileTransformOperator` only and rebased on current `main` (which includes #70505).

To confirm the remaining change is consistent with the #70505 distinction: `destination_bucket or self.source_bucket` is a truthiness/value read of the unrendered template parameter, **not** an argument-provision check (`is None`/`is not None`), so it is not covered by the new sanctioned pattern — the exemption entry only stays valid if the fallback moves out of `__init__`. That's what this PR does:

- `__init__` now stores `destination_bucket`/`destination_object` verbatim (no read of the unrendered value).
- The `or source_*` fallback resolves in `execute()`, after rendering.
- `get_openlineage_facets_on_complete` (which can run outside `execute`) resolves the fallbacks locally.
- The `GCSFileTransformOperator` exemption entry is removed; `validate_operators_init.py` passes on the file (exit 0).

Everything else from the earlier revision (`GCSDeleteObjectsOperator`, `GCSListObjectsOperator`, their exemption entries) was dropped.

---

shahar1 (commented):
Could you please add unit tests?

potiuk (approved):
The move of the `or self.source_*` fallback from `__init__` to `execute()` (mirrored in `get_openlineage_facets_on_start`) is correct and clears the `validate-operators-init` check for this class: running the validator's AST check, `main` reports two constructor reads of template fields on this operator and the patched source reports none, so dropping the exemption entry is required. Per the #70296 guidance this is a value read rather than a provision check, so `execute()` is the sanctioned home.

Since `shahar1` asked for tests and this had gone quiet, I pushed them on top (and rebased the branch on `main`): both construct the operator without a destination, change `source_bucket`/`source_object` after construction the way template rendering does, and assert the destination follows the rendered values in `execute()` and in the OpenLineage output dataset. Both fail against the previous `__init__`-time fallback.

### Smaller observations
- The practical effect is narrow: `destination_*` are template fields too, so the copied expression was rendered anyway on `main`; the difference is visible only for non-idempotent templates (e.g. `{{ macros.uuid.uuid4() }}`) or when `source_*` changes after construction. Worth keeping in mind when reading the description, which is still written for the earlier multi-operator revision.
- `destination_bucket`/`destination_object` are now `None` on the operator until `execute()` runs (e.g. in the Rendered Templates view); `get_openlineage_facets_on_start` is covered, and I found no other pre-execute reader in the repo.
- The `backport-to-v3-3-test` label looks accidental for a providers change (providers release from `main`).

---

> will follow up.*
>

PR #70491, opened 2026-07-27
Rename Neptune operator db_cluster_id argument to cluster_id

mitre88:

`NeptuneStartDbClusterOperator` and `NeptuneStopDbClusterOperator` declare `cluster_id` in `template_fields` but expose the value as the `db_cluster_id` constructor argument, assigning `self.cluster_id = db_cluster_id` in `__init__` — which the `validate-operators-init` check flags. Per review feedback, this renames the **argument** to `cluster_id` (rather than renaming the attribute as in the first iteration of this PR), so the templated attribute and `template_fields` entry stay exactly as they are on main and existing Dags templating the attribute are unaffected.

**Breaking change**: Dags passing `db_cluster_id=` to these two operators must switch to `cluster_id=`, and now fail loudly with a `TypeError` at construction time. A note was added to the provider changelog. The `db_cluster_id` keys in the operators' XCom return values and in the Neptune trigger/hook APIs are unchanged.

Both classes are removed from the exemption list and the `validate-operators-init` check passes locally, as do the unit tests.

---

### DIFF
--- a/providers/amazon/docs/changelog.rst
+++ b/providers/amazon/docs/changelog.rst
@@ -26,6 +26,12 @@
 Changelog
 ---------
 
+.. note::
+    The ``db_cluster_id`` argument of ``NeptuneStartDbClusterOperator`` and
+    ``NeptuneStopDbClusterOperator`` is renamed to ``cluster_id`` so that it matches the
+    operators' templated ``cluster_id`` attribute. Update Dags that pass ``db_cluster_id=``
+    to use ``cluster_id=`` instead.
+
 9.34.0
 ......
 
--- a/providers/amazon/docs/changelog.rst
+++ b/providers/amazon/docs/changelog.rst
@@ -26,6 +26,12 @@
 Changelog
 ---------
 
+.. note::
+    The ``db_cluster_id`` argument of ``NeptuneStartDbClusterOperator`` and
+    ``NeptuneStopDbClusterOperator`` is renamed to ``cluster_id`` so that it matches the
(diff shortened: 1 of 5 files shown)


uranusjr:
Since this is breaking either way, maybe it’s easier to rename the argument instead?

mitre88:
Good point — reworked the PR to rename the argument instead: the constructor now takes `cluster_id`, the templated `cluster_id` attribute and `template_fields` stay exactly as on main, and a wrong keyword fails loudly with a `TypeError`. Updated the unit tests, the system test example, and added a breaking-change note to the provider changelog. PR title and description updated accordingly.

---

vincbeck (approved):
This can be treated as a bug fix, I dont think we need a major version bump for this one

o-nikolas (commented):
To me it feels more correct to change the template field names, not the argument itself in the constructor. But it I wont die on that hill if everyone else things otherwise

eladkal:
> To me it feels more correct to change the template field names, not the argument itself in the constructor. But it I wont die on that hill if everyone else things otherwise

I believe this is more AWS decision how you want to play this. Probably the merge click is waiting for the team to decide if you are happy with it as is or would you like to ask author to change approach.

vincbeck:
> To me it feels more correct to change the template field names, not the argument itself in the constructor. But it I wont die on that hill if everyone else things otherwise

I agree with Niko, can we change the template field instead?

o-nikolas:
@mitre88 Can you revisit this?

PR #70492, opened 2026-07-27
Use common.compat.sdk for timezone imports in providers

FrankYang0529:
## Why

- `airflow.utils.timezone` is now resolved at runtime by a module-level `__getattr__`. mypy cannot see dynamically produced attributes, so `from airflow.utils import timezone` fails with `Module "airflow.utils" has no attribute "timezone" [attr-defined]`.
- The `mypy-providers` hook only checks files included in the current push diff, so these pre-existing failures cannot be detected until an unrelated change happens to touch an affected file.

## How

- Replace the imports with `from airflow.providers.common.compat.sdk import timezone`.

 

---

- [X] Yes - Claude Code

---

### DIFF
--- a/providers/standard/tests/unit/standard/triggers/test_temporal.py
+++ b/providers/standard/tests/unit/standard/triggers/test_temporal.py
@@ -23,11 +23,13 @@ from unittest import mock
 import pendulum
 import pytest
 
+from airflow.providers.common.compat.sdk import timezone
 from airflow.providers.standard.triggers.temporal import DateTimeTrigger, TimeDeltaTrigger
 from airflow.triggers.base import TriggerEvent
-from airflow.utils import timezone
 from airflow.utils.state import TaskInstanceState
-from airflow.utils.timezone import utcnow
+
+# Bound at import time so tests patching ``temporal.timezone.utcnow`` still read a real clock here.
+utcnow = timezone.utcnow
 
 
 def test_input_validation():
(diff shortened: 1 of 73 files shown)


kaxil (approved):

providers/standard/tests/unit/standard/triggers/test_temporal.py
```diff
@@ -23,11 +23,13 @@
...
 
+from airflow.providers.common.compat.sdk import timezone
 from airflow.providers.standard.triggers.temporal import DateTimeTrigger, TimeDeltaTrigger
 from airflow.triggers.base import TriggerEvent
-from airflow.utils import timezone
 from airflow.utils.state import TaskInstanceState
-from airflow.utils.timezone import utcnow
+
+# Bound at import time so tests patching ``temporal.timezone.utcnow`` still read a real clock here.
+utcnow = timezone.utcnow
```
Lee-W, line 32: or can we just do `from airflow.providers.common.compat.sdk.timezone import utcnow`

FrankYang0529: I tried it, but it shows `ModuleNotFoundError: No module named 'airflow.providers.common.compat.sdk.timezone'; 'airflow.providers.common.compat.sdk' is not a package`.

Lee-W: yep, your're right, i just checked. it's not something we can directly use in the old way. cool, then let's merge it

Lee-W (approved):

PR #70493, opened 2026-07-27
Delay BigQueryToMsSqlOperator source table parsing

fat-catTW:
Move `BigQueryToMsSqlOperator` parsing of `source_project_dataset_table` out of `__init__` so templated values are rendered before validation and dataset/table extraction.

This also removes the corresponding `validate_operators_init` exemption and adds regression coverage for rendered source table values.

related: #70296

- `[X]`  Yes (please specify the tool below)

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
+++ b/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
@@ -83,7 +83,6 @@ class BigQueryToMsSqlOperator(BigQueryToSqlBaseOperator):
 
         super().__init__(
             target_table_name=target_table_name,
-            dataset_table="__airflow_unrendered_dataset.__airflow_unrendered_table",
             **kwargs,
         )
         self.mssql_conn_id = mssql_conn_id
--- a/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
+++ b/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
@@ -81,15 +81,9 @@ class BigQueryToMsSqlOperator(BigQueryToSqlBaseOperator):
 
             target_table_name = mssql_table
 
-        try:
-            _, dataset_id, table_id = source_project_dataset_table.split(".")
-        except ValueError:
-            raise ValueError(
-                f"Could not parse {source_project_dataset_table} as <project>.<dataset>.<table>"
-            ) from None
         super().__init__(
             target_table_name=target_table_name,
-            dataset_table=f"{dataset_id}.{table_id}",
+            dataset_table="__airflow_unrendered_dataset.__airflow_unrendered_table",
             **kwargs,
         )
         self.mssql_conn_id = mssql_conn_id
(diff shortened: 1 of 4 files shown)


providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
```diff
@@ -81,15 +81,9 @@ def __init__(
...
-        try:
-            _, dataset_id, table_id = source_project_dataset_table.split(".")
-        except ValueError:
-            raise ValueError(
-                f"Could not parse {source_project_dataset_table} as <project>.<dataset>.<table>"
-            ) from None
         super().__init__(
             target_table_name=target_table_name,
-            dataset_table=f"{dataset_id}.{table_id}",
+            dataset_table="__airflow_unrendered_dataset.__airflow_unrendered_table",
```
potiuk: I understand why this is here — `BigQueryToSqlBaseOperator.__init__` declares `dataset_table: str` as a required keyword and unconditionally does `self.dataset_id, self.table_id = dataset_table.split(".")`, so it needs *something* dot-separated at construction time. But a magic sentinel is a fragile way to satisfy that.

It's observable: until `execute` runs, `self.dataset_id` / `self.table_id` hold `__airflow_unrendered_dataset` / `__airflow_unrendered_table`, and the parent builds `self.source_project_dataset_table = f"{project_id}.{self.dataset_id}.{self.table_id}"` from them. Anything reading the operator before execution — the UI, a log line, a task that fails during rendering — surfaces those strings to users with no explanation.

The parent already accepts `dataset_id` and `table_id` as explicit optional parameters, so the cleaner fix is probably to give `dataset_table` a default of `None` in the base class and guard the split. That's a change to shared code, so happy for it to be a follow-up rather than this PR — but if the sentinel stays, please add a comment here saying it's a placeholder overwritten in `execute`, so nobody spends an afternoon grepping for where those strings come from.

---

potiuk (approved):
Thanks — and the test here is the strongest in this whole batch. `test_execute_uses_rendered_source_project_dataset_table` calls `operator.render_template_fields({...})` and *then* `execute`, asserting on the resulting `list_rows` and `BigQueryTableLink.persist` arguments. That exercises the real rendering path rather than simulating it by assigning attributes, so it would catch a field being dropped from `template_fields` — several of the sibling PRs in this batch fake the rendering step instead. I've pointed #70530 at this as the pattern to copy.

Extracting `_get_source_project_dataset_table_parts()` rather than duplicating the split in two places is right, and keeping the `ValueError` with `from None` preserves the clean error message.

One question inline about the placeholder.

---

providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_mssql.py
fat-catTW: Thanks for the detailed note. 
I went with the cleaner base-class fix you suggested.

`BigQueryToSqlBaseOperator` now accepts `dataset_table=None` and only splits it when provided, so `BigQueryToMsSqlOperator` no longer needs to pass the synthetic `__airflow_unrendered_*` placeholder. The MSSQL operator still derives `dataset_id` / `table_id` from the rendered `source_project_dataset_table` during `execute()`.

I also added a regression assertion to the rendered-source test to confirm the placeholder values are not present before execution.

PR #69469, opened 2026-07-06
Core: Make psycopg (v3) the default synchronous Postgres driver

Dev-iL:
related: 
- #68453
- #68496
- #69526 

Mirrors the async default switch: a bare `postgresql://` or legacy `postgres://` / `postgres+psycopg2://` `sql_alchemy_conn` is now rewritten to `postgresql+psycopg://` instead of `postgresql+psycopg2://`. An explicit `postgresql+psycopg2://` URL is never rewritten and keeps working as long as `apache-airflow-providers-postgres[psycopg2]` is installed.

### What changed

- `_upgrade_postgres_metastore_conn` (`airflow-core/src/airflow/configuration.py`): `good_scheme` changed from `postgresql+psycopg2` to `postgresql+psycopg` **when the `psycopg` (v3) package is actually importable** (`find_spec("psycopg") is not None`), mirroring the exact guard `settings.py` already uses for the async driver default (`_USE_PSYCOPG3`); falls back to `postgresql+psycopg2` otherwise, so environments that don't have psycopg3 installed (e.g. a min-supported-Airflow-version test matrix, or a provider release pinned before psycopg3 support existed) keep working instead of hitting `sqlalchemy.exc.NoSuchModuleError`. `bad_schemes` is unchanged.
- `prepare_engine_args` (`airflow-core/src/airflow/settings.py`): the psycopg2-specific `executemany_mode`/`executemany_batch_page_size` tuning now follows the **actually-configured** `SQL_ALCHEMY_CONN` scheme, not whether the `psycopg` package happens to be importable.
- Two pre-existing migration bugs, invisible under psycopg2 but broken under psycopg3, found by running this branch's CI and then replaying the full upgrade/downgrade/upgrade chain against a real Postgres locally:
  - `airflow-core/src/airflow/migrations/versions/0017_2_9_2_fix_inconsistency_between_ORM_and_migration_files.py`: `literal("0")` (a Python `str`) bound to the `Integer` column `dag.max_consecutive_failed_dag_runs`. psycopg2 let Postgres coerce the untyped bind param silently; psycopg3's dialect renders it with an explicit `::VARCHAR` cast, causing `psycopg.errors.DatatypeMismatch` on every fresh migration. Changed to `literal(0)`.
[...]

PR #70496, opened 2026-07-27
Breeze: Make postgres driver explicit and version-aware

shahar1:
Since #69469 moved Breeze's `backend-postgres.yml` to a bare `postgresql://` scheme (so that core's autodetection selects psycopg3), every Breeze postgres session running Airflow from sources emits a `FutureWarning` at config load — `main` now lists the bare scheme in `bad_schemes`.

Making the driver explicit has to be version-aware: the migration tests install a released Airflow into the same container, no released version ships psycopg (v3) in its default constraints, and releases before 3.2.0 can run on SQLAlchemy 1.4, which has no `postgresql+psycopg` dialect at all — an explicit psycopg3 URL crashes them with `sqlalchemy.exc.NoSuchModuleError` (this is exactly what failed the Postgres core jobs on this PR's first revision).

Breeze now derives `POSTGRES_DRIVER` from `--use-airflow-version` — psycopg2 when a released version is installed, psycopg for sources/wheel/sdist/branch/PR installs — and the compose file interpolates it into both `sql_alchemy_conn` and the Celery result backend URL, defaulting to psycopg.

Verified locally: the Airflow 2.11.0 migration leg resolves `postgresql+psycopg2://` and completes, and a from-sources session resolves `postgresql+psycopg://` with no `FutureWarning`.

---

- [X] Yes — Claude Code (Fable 5, Opus 5)

Generated-by: Claude Code (Fable 5, Opus 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/dev/breeze/src/airflow_breeze/params/shell_params.py
+++ b/dev/breeze/src/airflow_breeze/params/shell_params.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import os
+import re
 import sys
 from base64 import b64encode
 from copy import deepcopy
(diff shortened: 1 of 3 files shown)


Dev-iL (approved):
Doesn't look like this setting has any backward compatibility concerns, so the fix should be fine.

I wonder why this was skipped in https://github.com/apache/airflow/pull/69469. Perhaps it was a matter of ordering...

potiuk:
There are some errors to fix ?

potiuk:
@Dev-iL - maybe you can take a look ?

Dev-iL:
Ok so here's the deal:
- The warning fires on 2.11.0 which ships an older version of `airflow-core/src/airflow/configuration.py` that predates the psycopg v3 autodetection.
- It should be possible fix the old-airflow `breeze shell` invocations by specifying `--custom-db-url` as `"postgresql://postgres:airflow@postgres/airflow"`

Dev-iL:
@shahar1 @potiuk merge?

shahar1:
> @shahar1 @potiuk merge?

I will merge once approved, still investigating the root cause of the issue :)

Dev-iL:
Huh, CI was green earlier

shahar1:
> Huh, CI was green earlier



> Huh, CI was green earlier

Unrelated to this PR, main is currently broken. Trying to work on a fix :)

potiuk (approved):

PR #70497, opened 2026-07-27
Show which LLM providers each Common AI connection type reaches

Lee-W:
## Why
Common AI provider page shows `pydanticai, pydanticai-azure, pydanticai-bedrock, langchain`, etc. but not upstream service (Anthropic, AWS Bedrock, Ollama, ...) each one actually talks to without reading source code.

## What
Add an optional external-integrations list to the connection-types schema and populate it in the Common AI provider.yaml from the existing connection docs, thread it through the registry metadata extractor and contract model, and render it as a small badge list under each connection chip on the provider detail page.

image

 

---

---


providers/common/ai/provider.yaml
```diff
@@ -407,6 +430,13 @@ connection-types:
   - hook-class-name: airflow.providers.common.ai.hooks.llamaindex.LlamaIndexHook
     hook-name: "LlamaIndex"
     connection-type: llamaindex
+    external-integrations:
```
kaxil: This list doesn't match what the connection can reach. `LlamaIndexHook.get_llm()` / `get_embedding_model()` only ever construct llama-index's `OpenAI` / `OpenAIEmbedding` classes, and the hook docstring explicitly routes Cohere, Bedrock, Vertex and HuggingFace to "instantiate the LlamaIndex class directly in your `@task`", which bypasses the hook and this connection entirely. Unlike `pydanticai` there is no env-var fallback path here that could make them work, so I think the accurate list is OpenAI plus OpenAI-compatible endpoints (Ollama, vLLM, per the hook's own placeholder), and Anthropic is questionable for the same reason. The mirrored list in `get_provider_info.py` needs the same edit.

dev/registry/extract_metadata.py
```diff
@@ -782,12 +784,14 @@ def main():
...
             conn_type = conn.get("connection-type", "")
             hook_class = conn.get("hook-class-name", "")
+            external_integrations = conn.get("external-integrations", [])
             if conn_type:
                 connection_types.append(
                     {
                         "conn_type": conn_type,
                         "hook_class": hook_class,
                         "docs_url": resolve_connection_docs_url(conn_type, conn_url_map, base_docs_url),
+                        "external_integrations": external_integrations,
```
kaxil: `extract_versions.py` builds the same `connection_types` dicts (around line 394) and wasn't given this field, so per-version metadata will never carry it. The template splits on `pv.isLatest`, which means the table shows on a release while it's latest and then silently disappears once it's superseded. `ConnectionTypeContract` is already reused by `ProviderVersionMetadataContract`, so mirroring this one line there is all it takes.

providers/common/ai/provider.yaml
```diff
@@ -147,6 +147,16 @@ connection-types:
   - hook-class-name: airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIHook
     hook-name: "Pydantic AI"
     connection-type: pydanticai
+    external-integrations:
+      - OpenAI
+      - Anthropic
+      - Google
+      - AWS Bedrock
+      - Groq
+      - Mistral
```
kaxil: Same vendor is "Mistral" here and "Mistral AI" in the langchain list below; both render in the same table, so it reads as two different upstreams. Worth picking one spelling. On the same note, the hook docstring names DeepSeek in this hook's covered set and the langchain list includes it, so it probably belongs here too.

airflow-core/src/airflow/provider.yaml.schema.json
```diff
@@ -447,6 +447,14 @@
                         },
                         "additionalProperties": false
                     },
+                    "external-integrations": {
```
kaxil: Two thoughts on the field definition. The name sits one key away from the existing top-level `integrations`, which describes a different axis (framework integrations driving docs/logos/tags), so authors skimming the schema may conflate them; something like `upstream-services` would be harder to misread. And whatever name it lands on, it deserves a subsection in contributing-docs/23_provider_hook_migration_to_yaml.rst next to `ui-field-behaviour` and `conn-fields`, since that's the one place provider authors learn the connection-types fields.

dev/registry/registry_contract_models.py
```diff
@@ -52,6 +52,7 @@ class ConnectionTypeContract(BaseModel):
     conn_type: str
     hook_class: str = ""
     docs_url: str | None = None
+    external_integrations: list[str] = Field(default_factory=list)
```
kaxil: Nothing in dev/registry/tests exercises the new field; deleting the `conn.get("external-integrations", [])` line in extract_metadata.py wouldn't fail any test. test_registry_contract_models.py already has the legacy connection_types payload test (around line 111) this extends naturally: assert the field defaults to `[]` on old payloads and round-trips when present.

providers/common/ai/provider.yaml
Lee-W: Sure, it's now trimmed the list to OpenAI, Ollama and vLLM. Dropped Anthropic for the same reason.

providers/common/ai/provider.yaml
Lee-W: Unified it as `Mistral AI` and added DeepSeek.

airflow-core/src/airflow/provider.yaml.schema.json
Lee-W: It doesn't feel like `upstream` to me so I choose the term `external-services` instead. Also documneted it in  `contributing-docs/23_provider_hook_migration_to_yaml.rst`.

dev/registry/registry_contract_models.py
Lee-W: Added the test cases to `test_registry_contract_models.py`. It validates `external_services`. If provided, the value should be there. If not, no error should be raised.

providers/common/ai/provider.yaml
```diff
@@ -374,6 +391,13 @@ connection-types:
   - hook-class-name: airflow.providers.common.ai.hooks.langchain.LangChainHook
     hook-name: "LangChain"
     connection-type: langchain
+    external-services:
```
kaxil, line 393: The three lists now apply three different rules to OpenAI-compatible self-hosted endpoints: pydanticai gets Ollama, vLLM and LM Studio, llamaindex gets Ollama and vLLM, langchain gets Ollama only. `langchain.py:55` documents `host` as "custom endpoints, Ollama, vLLM" and `_connection_kwargs` passes it through as `base_url` unchanged, so vLLM belongs here too (and in `get_provider_info.py:313`). Worth settling whether LM Studio is in all three or none, since it reaches the same surface.

providers/common/ai/provider.yaml
```diff
@@ -147,6 +147,17 @@ connection-types:
   - hook-class-name: airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIHook
     hook-name: "Pydantic AI"
     connection-type: pydanticai
+    external-services:
```
kaxil, line 150: This renders as a bare "External services" column, so it reads as the complete set, but the base hook just maps `conn.password` to `api_key` for whichever provider the model id names, and 27 of pydantic-ai's provider classes take `api_key` at the version resolving today (Cohere, OpenRouter, Together, Fireworks, Cerebras and HuggingFace among them). Ten listed out of 27 means a reader concludes the rest are unsupported, and the list needs an edit every time pydantic-ai adds a provider. Either label the column and the schema description as representative rather than exhaustive, or keep enumerations for the single-vendor types where they are complete by construction.

registry/src/css/main.css
```diff
@@ -3556,6 +3556,40 @@ main {
   border-color: var(--color-green-400);
 }
 
+/* Per-connection upstream integration matrix */
```
kaxil: The rename did not reach the CSS: this comment still says "upstream integration matrix" and the classes are `conn-integrations-table/-header/-row/-list`, used at `provider-version.njk:256-268`. They are new in this PR so nothing consumes them yet, which makes renaming free now and dead-name debt later.

kaxil (approved):

providers/common/ai/provider.yaml
Lee-W: langchain now lists vLLM — `langchain.py` documents `host` as "custom endpoints, Ollama, vLLM" and passes it straight through as `base_url`, so it belongs there. LM Studio is gone from pydanticai, so the lists now follow what the hooks document rather than adding a vendor nothing covers.

registry/src/css/main.css
Lee-W: Renamed to `conn-external-services-{table,header,row,list}` and reworded the
comment.

providers/common/ai/provider.yaml
Lee-W: Went with labelling here. The column now reads "External services (examples)", and both the schema description and the `external-services` section in `contributing-docs/23_provider_hook_migration_to_yaml.rst` say the list only names representative ones instead of all of them.

PR #70498, opened 2026-07-27
Match providers to Explore categories by integration name, not just id

Lee-W:
## Why
The Explore page and each provider's category chips only matched a provider to a category by substring-matching the category's keywords against the provider's id. Providers whose id doesn't happen to   contain a relevant keyword never showed up, even when their
`provider.yaml` declares integrations that obviously belong there.
e.g., `common-ai` (wraps `pydantic-ai`) ships hooks named "Pydantic AI", "MCP Server", "LangChain", and "LlamaIndex", each tagged ai, but none of those words are substrings of the id "common-ai", so it never appeared under AI & Machine Learning or in a search for "LangChain"/"MCP".

## What
Match against each provider's declared integration names (already extracted as provider.categories[].name) in addition to its id, and add the ai-ml category keywords needed for common-ai's integrations. Also scan connection_types[].external_integrations when present so a future field populating that data benefits automatically.

Note on scope: matching by integration name also widens the umbrella providers, since they
declare many integrations — `amazon` goes from 1 category to 4 (Databases, Data Processing, Workflow), `google` 1 to 3, `alibaba` and `cncf-kubernetes` 1 to 2. These memberships are accurate; the id-only matcher just couldn't see them. Expect the same effect from future keyword additions.

---

Generated-by: [Tool Name] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/registry/src/_data/exploreCategories.js
+++ b/registry/src/_data/exploreCategories.js
@@ -55,7 +55,7 @@ module.exports = [
     name: 'AI & Machine Learning',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9.663 17h4.673M12 3v1m6.364 1.636l-.707.707M21 12h-1M4 12H3m3.343-5.657l-.707-.707m2.828 9.9a5 5 0 117.072 0l-.548.547A3.374 3.374 0 0014 18.469V19a2 2 0 11-4 0v-.531c0-.895-.356-1.754-.988-2.386l-.548-.547z" /></svg>',
     color: 'amber',
-    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector', 'langchain', 'llamaindex', 'mcp', 'pydantic ai', 'pydantic-ai'],
+    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector', 'langchain', 'llamaindex', 'mcp', 'pydantic ai'],
     description: 'OpenAI, vector DBs, and ML platforms',
   },
   {
--- a/registry/src/_data/exploreCategories.js
+++ b/registry/src/_data/exploreCategories.js
@@ -55,7 +55,7 @@ module.exports = [
     name: 'AI & Machine Learning',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9.663 17h4.673M12 3v1m6.364 1.636l-.707.707M21 12h-1M4 12H3m3.343-5.657l-.707-.707m2.828 9.9a5 5 0 117.072 0l-.548.547A3.374 3.374 0 0014 18.469V19a2 2 0 11-4 0v-.531c0-.895-.356-1.754-.988-2.386l-.548-.547z" /></svg>',
     color: 'amber',
-    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector'],
+    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector', 'langchain', 'llamaindex', 'mcp', 'pydantic ai', 'pydantic-ai'],
     description: 'OpenAI, vector DBs, and ML platforms',
   },
   {
--- a/registry/src/_data/exploreCategories.js
+++ b/registry/src/_data/exploreCategories.js
@@ -31,7 +31,7 @@ module.exports = [
     name: 'Databases',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M4 7v10c0 2.21 3.582 4 8 4s8-1.79 8-4V7M4 7c0 2.21 3.582 4 8 4s8-1.79 8-4M4 7c0-2.21 3.582-4 8-4s8 1.79 8 4" /></svg>',
     color: 'blue',
-    keywords: ['postgres', 'mysql', 'mongo', 'redis', 'neo4j', 'elasticsearch', 'cassandra', 'couchbase', 'influxdb', 'sqlite', 'odbc', 'jdbc', 'common-sql'],
+    keywords: ['postgres', 'mysql', 'mongo', 'redis', 'neo4j', 'elasticsearch', 'cassandra', 'couchbase', 'influxdb', 'sqlite', 'odbc', 'jdbc', 'common-sql', 'mssql', 'opensearch', 'cloudant', 'arangodb'],
     description: 'SQL, NoSQL, and time-series databases',
   },
   {
@@ -39,7 +39,7 @@ module.exports = [
     name: 'Data Warehouses',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M19 11H5m14 0a2 2 0 012 2v6a2 2 0 01-2 2H5a2 2 0 01-2-2v-6a2 2 0 012-2m14 0V9a2 2 0 00-2-2M5 11V9a2 2 0 012-2m0 0V5a2 2 0 012-2h6a2 2 0 012 2v2M7 7h10" /></svg>',
     color: 'purple',
-    keywords: ['snowflake', 'databricks', 'teradata', 'vertica', 'trino', 'presto', 'dbt'],
+    keywords: ['snowflake', 'databricks', 'teradata', 'vertica', 'trino', 'presto', 'dbt', 'exasol', 'clickhouse'],
     description: 'Snowflake, Databricks, and analytics platforms',
   },
   {
@@ -47,7 +47,7 @@ module.exports = [
     name: 'Messaging & Notifications',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 12h.01M12 12h.01M16 12h.01M21 12c0 4.418-4.03 8-9 8a9.863 9.863 0 01-4.255-.949L3 20l1.395-3.72C3.512 15.042 3 13.574 3 12c0-4.418 4.03-8 9-8s9 3.582 9 8z" /></svg>',
     color: 'green',
-    keywords: ['slack', 'sendgrid', 'smtp', 'discord', 'telegram', 'pagerduty', 'opsgenie', 'twilio'],
+    keywords: ['slack', 'sendgrid', 'smtp', 'discord', 'telegram', 'pagerduty', 'opsgenie', 'twilio', 'dingtalk', 'apprise'],
     description: 'Slack, email, SMS, and alerting services',
   },
   {
@@ -71,7 +71,7 @@ module.exports = [
     name: 'Workflow & Orchestration',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M4 4v5h.582m15.356 2A8.001 8.001 0 004.582 9m0 0H9m11 11v-5h-.581m0 0a8.003 8.003 0 01-15.357-2m15.357 2H15" /></svg>',
     color: 'rose',
-    keywords: ['standard', 'celery', 'docker', 'kubernetes', 'ssh', 'http', 'ftp', 'sftp'],
+    keywords: ['standard', 'celery', 'docker', 'kubernetes', 'ssh', 'http', 'ftp', 'sftp', 'grpc'],
     description: 'Core operators, executors, and connectivity',
   },
 ];
--- a/registry/src/_data/exploreCategoryProviders.js
+++ b/registry/src/_data/exploreCategoryProviders.js
@@ -33,6 +33,13 @@ module.exports = function () {
         }
       }
     }
+    // explore.njk shows only the first six as badges, so rank before slicing
+    // (as its Top/Incubating rows already do) — otherwise the row is whatever
+    // providers.json happened to list first, and widening a category's
+    // membership silently pushes the well-known names off it.
+    matched.sort(
+      (a, b) => (b.pypi_downloads?.monthly || 0) - (a.pypi_downloads?.monthly || 0),
+    );
     map[category.id] = matched;
   }
   return map;
--- a/registry/src/_data/exploreCategoryProviders.js
+++ b/registry/src/_data/exploreCategoryProviders.js
@@ -19,6 +19,7 @@
 
 const providersData = require("./providers.json");
 const exploreCategories = require("./exploreCategories");
+const { providerMatchesKeyword } = require("./providerKeywordMatch");
 
 module.exports = function () {
   const map = {};
@@ -26,10 +27,7 @@ module.exports = function () {
     const matched = [];
     for (const provider of providersData.providers) {
       for (const keyword of category.keywords) {
-        if (
-          provider.id.includes(keyword) ||
-          keyword.includes(provider.id)
-        ) {
+        if (providerMatchesKeyword(provider, keyword)) {
           matched.push(provider);
           break;
         }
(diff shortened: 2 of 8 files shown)


registry/src/_data/providerKeywordMatch.js
```diff
@@ -0,0 +1,62 @@
...
+// "Pydantic AI"), and, once populated, each connection type's declared
+// external integrations. The latter field doesn't exist in the generated
+// data yet, so this reads as an empty list until a future change in
+// dev/registry/extract_metadata.py starts populating it — no further
+// changes needed here when that lands.
+function collectSearchableValues(provider) {
+  const values = [provider.id];
+  for (const category of provider.categories || []) {
+    if (category.name) {
+      values.push(category.name);
```
kaxil, line 48: I ran the new matcher over all 105 in-tree `provider.yaml` files. Beyond the intended additions it also pulls the umbrella providers into categories they weren't in before: `amazon` goes 1 category to 4 (Databases via "Amazon OpenSearch Serverless", Data Processing via "Amazon Managed Service for Apache Flink", Workflow via "Amazon Elastic Kubernetes Service (EKS)"), `google` 1 to 3, `alibaba` and `cncf-kubernetes` 1 to 2. The counts on the Explore cards move too: Databases 12 to 17, Data Processing 8 to 12, Workflow 8 to 11.

Is that intended? The commit message lists the small providers but not these, and amazon has ~50 integrations, so most future keyword additions will pull it in as well.

registry/src/_data/providerKeywordMatch.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+  for (const connectionType of provider.connection_types || []) {
+    for (const externalIntegration of connectionType.external_integrations || []) {
```
kaxil: This can't ever be populated as written. `ConnectionTypeContract` in `dev/registry/registry_contract_models.py` sets `extra="forbid"` and declares only `conn_type`, `hook_class`, `docs_url`, and `extract_metadata.py` validates providers.json against it before writing. Adding the field upstream fails validation first, so the contract change and these four lines would land together anyway. I'd drop the loop and the comment above it for now.

registry/src/_data/providerKeywordMatch.js
```diff
@@ -0,0 +1,62 @@
...
+
+// Two arbitrary strings "match" a keyword if either contains the other
+// (case-insensitively), mirroring the original id-only substring check.
+function fuzzyIncludes(value, keyword) {
+  if (!value) {
+    return false;
+  }
+  const normalizedValue = value.toLowerCase();
+  const normalizedKeyword = keyword.toLowerCase();
+  return normalizedValue.includes(normalizedKeyword) || normalizedKeyword.includes(normalizedValue);
```
kaxil: The `normalizedKeyword.includes(normalizedValue)` half was safe when the left side was always a slug, but it now runs against free-text integration names, and the tree has 3-4 char ones ("Tpt", "Ttu", "YDB", "Bteq") that a longer keyword could swallow. I checked forward-only matching across all 105 `provider.yaml` files and category membership comes out identical, so nothing depends on the reverse direction today. Worth dropping it, or keeping it scoped to `provider.id`.

registry/src/_data/exploreCategories.js
```diff
@@ -31,31 +31,31 @@ module.exports = [
...
+    keywords: ['slack', 'sendgrid', 'smtp', 'discord', 'telegram', 'pagerduty', 'opsgenie', 'twilio', 'dingtalk', 'apprise'],
     description: 'Slack, email, SMS, and alerting services',
   },
   {
     id: 'ai-ml',
     name: 'AI & Machine Learning',
     icon: '<svg fill="none" stroke="currentColor" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9.663 17h4.673M12 3v1m6.364 1.636l-.707.707M21 12h-1M4 12H3m3.343-5.657l-.707-.707m2.828 9.9a5 5 0 117.072 0l-.548.547A3.374 3.374 0 0014 18.469V19a2 2 0 11-4 0v-.531c0-.895-.356-1.754-.988-2.386l-.548-.547z" /></svg>',
     color: 'amber',
-    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector'],
+    keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector', 'langchain', 'llamaindex', 'mcp', 'pydantic ai', 'pydantic-ai'],
```
kaxil: `'pydantic-ai'` matches nothing. `fuzzyIncludes` compares raw strings, so neither direction connects the hyphen to the space in the integration name "Pydantic AI", and `'pydantic ai'` alone already covers it. Normalizing `[-_\s]` inside `fuzzyIncludes` would let one keyword handle both spellings and save the next person the same double entry.

registry/src/_data/providerCategoryMap.js
```diff
@@ -19,14 +19,15 @@
...
+const { providerMatchesKeyword } = require("./providerKeywordMatch");
 
 module.exports = function () {
   const map = {};
   for (const provider of providersData.providers) {
     const cats = [];
     for (const category of exploreCategories) {
       for (const keyword of category.keywords) {
-        if (provider.id.includes(keyword) || keyword.includes(provider.id)) {
+        if (providerMatchesKeyword(provider, keyword)) {
```
kaxil, line 30: This fixes the category dropdown on /providers/ (through `data-categories`), but the search box on that page still matches `data-name` only (`matchesSearch` in `src/js/provider-filters.js`), so typing "langchain" there is still a miss. If the search half of the description is in scope, `providers.njk` needs the integration names in a data attribute too.

kaxil (approved):

registry/src/_data/providerKeywordMatch.js
Lee-W: Yes, it's intended. For example, Amazon really does ship OpenSearch Serverless and Managed Service for Apache Flink — the id-only matcher just couldn't see them. Someone searching for those and not finding Amazon isn't ideal.

And yes, this scales with the keyword list rather than with the provider. I don't think capping is the right answer though — excluding a provider because it's large would make the listing less accurate. If it starts getting too noisy, we can look at a new design instead of dropping accurate memberships.

registry/src/_data/providerKeywordMatch.js
Lee-W: Checked and removed. Thanks!

registry/src/_data/providerKeywordMatch.js
Lee-W: Agreed, removed. `fuzzyIncludes` is now forward-only.

registry/src/_data/exploreCategories.js
Lee-W: Agreed — removed, and `fuzzyIncludes` normalizes now.

registry/src/_data/providerCategoryMap.js
Lee-W: `providers.njk` now emits the integration names as `data-integrations` and `matchesSearch` tests it, so "langchain" hits on /providers/.

image

registry/src/js/provider-filters.js
```diff
@@ -40,9 +40,11 @@
       const lifecycle = item.dataset.lifecycle;
       const name = item.dataset.name || '';
       const categories = item.dataset.categories || '';
+      const integrations = item.dataset.integrations || '';
 
       const matchesLifecycle = currentLifecycle === 'all' || lifecycle === currentLifecycle;
-      const matchesSearch = name.includes(currentSearch.toLowerCase());
+      const search = currentSearch.toLowerCase();
+      const matchesSearch = name.includes(search) || integrations.includes(search);
```
kaxil: The build-time matcher collapses `[-_\s]+` to a single space on both sides specifically so `'pydantic-ai'` matches "Pydantic AI", but this comparison is a raw `includes`, so the two halves of the same feature now disagree. I simulated `matchesSearch` against the rendered `data-integrations` for all 105 in-tree providers: `pydantic ai`, `langchain`, `mcp` and `llamaindex` all find `common-ai`, while `pydantic-ai`, `pydantic_ai` and `mcp-server` find nothing, and `speech to text` misses `google` because the integration is named "Google Cloud Speech-to-Text". Running `normalize()` on both sides here closes all of those and picks up `common-ai` as a query for free; a browser IIFE can't `require` the module, so a comment pointing back at `_data/providerKeywordMatch.js` is probably the best you can do about the duplication.

registry/src/providers.njk
```diff
@@ -63,12 +63,19 @@ mainClass: providers-page
       {% set lc = provider.lifecycle or "production" %}
       {% set lcDisplay = "incubation" if lc == "incubation" else ("deprecated" if lc == "deprecated" else "stable") %}
       {% set cats = providerCategoryMap[provider.id] or [] %}
+      {% set integrations = [] %}
+      {% for cat in provider.categories or [] %}
+        {% if cat.name %}
+        {% set integrations = (integrations.push(cat.name | lower), integrations) %}
```
kaxil: `push` mutates in place and Nunjucks `{% set %}` inside a `{% for %}` writes to the enclosing frame, so the `(push(), integrations)` comma expression isn't buying anything over a bare push. `{{ (provider.categories or []) | selectattr('name') | join(',', 'name') | lower }}` inline on line 78 renders byte-identical output on nunjucks 3.2.4 (checked against a normal provider, a missing `categories` key, an empty list, an entry with no name, and the one integration name containing `&`), and lets lines 66-71 go away.

registry/src/providers.njk
```diff
@@ -63,12 +63,19 @@ mainClass: providers-page
...
+        {% endif %}
+      {% endfor %}
       <li class="provider-item"
           data-lifecycle="{{ lcDisplay }}"
           data-name="{{ provider.name | lower }}"
           data-downloads="{{ provider.pypi_downloads.monthly }}"
           data-updated="{{ provider.last_updated or '' }}"
-          data-categories="{{ cats | join(',') }}">
+          data-categories="{{ cats | join(',') }}"
+          data-integrations="{{ integrations | join(',') }}">
```
kaxil: Now that the search box matches integration names, a card can match on text it never renders: searching `bedrock` or `athena` returns the Amazon card, whose only prose is "Amazon integration (including Amazon Web Services (AWS))", and `llamaindex` returns Common AI with no mention of LlamaIndex, which reads more like a filtering bug than a hit. Since `data-integrations` is already on the `<li>`, `provider-filters.js` could surface the matched name on the card at filter time without any new build data. Follow-up rather than a blocker, but worth deciding.

registry/src/js/provider-filters.js
Lee-W: Both sides collapse the same way now. `normalize()` is duplicated into the IIFE with a comment pointing back at `_data/providerKeywordMatch.js`.

I swept every provider name and integration name as a query against the rendered attributes: no existing match is lost, and `pydantic-ai`, `pydantic_ai`, `mcp-server`, `common-ai` and `speech to text` all resolve now.

Lee-W: image

registry/src/js/provider-filters.js
```diff
@@ -33,16 +33,27 @@
   let currentSearch = '';
   let debounceTimer;
 
+  // Kept in sync by hand with normalize() in src/_data/providerKeywordMatch.js:
```
Lee-W, line 36: I think it's ok to use comment here. I don't expect this will change often

registry/src/providers.njk
Lee-W: Dropped the loop — `data-integrations` is the inline `selectattr('name') | join(',', 'name') | lower` now.

registry/src/providers.njk
Lee-W: Agree it looks like a bug. Let me do it as a follow up

registry/src/_data/exploreCategoryProviders.js
```diff
@@ -19,17 +19,15 @@
...
   for (const category of exploreCategories) {
     const matched = [];
     for (const provider of providersData.providers) {
       for (const keyword of category.keywords) {
-        if (
-          provider.id.includes(keyword) ||
-          keyword.includes(provider.id)
-        ) {
+        if (providerMatchesKeyword(provider, keyword)) {
           matched.push(provider);
```
kaxil, line 31: `matched` comes out in `providers.json` order and `explore.njk:32` renders `| slice(0, 6)` with no sort, so widening membership changes which six badges show. Running both matchers over all 105 in-tree providers and applying the same slice: Messaging & Notifications drops `slack` and `smtp` from the row while the card's own copy still reads "Slack, email, SMS, and alerting services"; Databases drops `mongo`/`jdbc`/`influxdb`, Data Warehouses `teradata`/`trino`, Data Processing `apache-livy`/`apache-kylin`, Workflow `http`/`sftp`, AI & Machine Learning `qdrant`.

This isn't the capping you rejected, nothing loses membership, the badge row just shows a different six. Sorting `matched` here by `pypi_downloads.monthly` before it reaches the template would keep the well-known names in the row; `explore.njk:68` and `:108` already rank their rows before slicing instead of taking the first N.

registry/src/_data/providerKeywordMatch.js
```diff
@@ -0,0 +1,58 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Shared by providerCategoryMap.js and exploreCategoryProviders.js so the
+// two directions of the same provider <-> explore-category matching (used
+// respectively by the per-provider page's category chips and the Explore
```
kaxil: The prose describing this rule didn't move with it. `README.md:259` still calls `keywords` a "List of substrings matched against `provider.id`" and 262-264 repeats "matches (substring) the provider's ID". That's what the next person reads before adding a keyword, and it now understates the match surface.

The PR body also still promises the `connection_types[].external_integrations` scan that came out last round, and this comment credits "the per-provider page's category chips" when `providerCategoryMap`'s only consumer is `providers.njk:65` feeding the /providers/ dropdown (`provider-detail.njk` has no `categor` match at all).

registry/src/providers.njk
```diff
@@ -68,7 +68,8 @@ mainClass: providers-page
           data-name="{{ provider.name | lower }}"
           data-downloads="{{ provider.pypi_downloads.monthly }}"
           data-updated="{{ provider.last_updated or '' }}"
-          data-categories="{{ cats | join(',') }}">
+          data-categories="{{ cats | join(',') }}"
+          data-integrations="{{ (provider.categories or []) | selectattr('name') | join(',', 'name') | lower }}">
```
kaxil, line 73: The client mirror of `collectSearchableValues` is missing the id half: the build-time matcher searches `provider.id` plus the integration names, but the box only sees `data-name` (the display name) and this attribute. Sweeping all 105 in-tree providers as queries, not one is findable by the `apache-airflow-providers-<id>` string `provider-card.njk:20` prints under the title, and `clickhousedb`, `cncf-kubernetes`, `common-dataquality`, `edge3`, `microsoft-mssql`, `microsoft-psrp` and `microsoft-winrm` don't resolve from their bare id either.

Same card-text-vs-search-text mismatch as https://github.com/apache/airflow/pull/70498#discussion_r3677876974, so happy for it to ride along in that follow-up rather than here.

kaxil (approved):

registry/src/_data/exploreCategoryProviders.js
Lee-W: Sorted `matched` by `pypi_downloads.monthly` now. Also matches how the Top Providers and Incubating rows on the same page already work.

registry/src/_data/providerKeywordMatch.js
Lee-W: All three updated. The `keywords` bullet and the paragraph under it now describe the id plus integration-name surface and the separator collapsing, and point at `providerKeywordMatch.js` as the matcher. The comment now credits the category dropdown on /providers/ rather than the per-provider page, and the `external_integrations` line is out of the PR body.

registry/src/providers.njk
Lee-W: `data-id` is on the `<li>` and `matchesSearch` tests it, and the Pagefind record now carries the id too.

I index `provider.id` instead of distribution name, and drop the shared prefix off the query instead. Matching the distribution name directly makes `apache` go from 18 cards to all 105, `airflow` from 2 to 105 and `providers` from 0 to 105, since every provider's name starts with `apache-airflow-providers-`.

Sweeping all 105 names, ids and distribution names plus every integration name: nothing previously findable is lost, and queries that returned nothing drop from 114 to 2 (`providers`, and the bare prefix with no trailing separator — both already returned nothing).

PR #70499, opened 2026-07-27
Add a Retry Policies category to the provider registry

Lee-W:
## Why
RetryPolicy is a Dag-author-facing pluggable component (task-sdk AIP-105) with a concrete implementation, common-ai's LLMRetryPolicy, but it had no presence in the provider registry -- no module type to classify it under and no provider.yaml section to declare it, so it was undiscoverable when browsing the registry.

## What
Make it searchable in the registry

image

 

---

---

### DIFF
--- a/airflow-core/src/airflow/provider.yaml.schema.json
+++ b/airflow-core/src/airflow/provider.yaml.schema.json
@@ -397,6 +397,30 @@
                 ]
             }
         },
+        "retry-policies": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "integration-name": {
+                        "type": "string",
+                        "description": "Integration name. It must have a matching item in the 'integration' section of any provider."
+                    },
+                    "python-modules": {
+                        "description": "List of python modules containing the retry policies.",
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    }
+                },
+                "additionalProperties": false,
+                "required": [
+                    "integration-name",
+                    "python-modules"
+                ]
+            }
+        },
         "connection-types": {
             "type": "array",
             "description": "Array of connection types mapped to hook class names",
--- a/dev/registry/registry_tools/types.py
+++ b/dev/registry/registry_tools/types.py
@@ -109,6 +109,13 @@ MODULE_TYPES: dict[str, dict] = {
         "label": "Toolsets",
         "icon": "U",
     },
+    "retry_policy": {
+        "yaml_key": "retry-policies",
+        "level": "module",
+        "suffixes": ["RetryPolicy"],
+        "label": "Retry Policies",
+        "icon": "R",
+    },
 }
 
 # Runtime base class imports for issubclass checks (extract_parameters.py).
@@ -121,6 +128,7 @@ BASE_CLASS_IMPORTS: list[tuple[str, str]] = [
     ("bundle", "airflow.dag_processing.bundles.base.BaseDagBundle"),
     ("operator", "airflow.sdk.bases.operator.BaseOperator"),
     ("toolset", "pydantic_ai.toolsets.abstract.AbstractToolset"),
+    ("retry_policy", "airflow.sdk.definitions.retry_policy.RetryPolicy"),
 ]
 
 # Derived lookups used by extraction scripts.
--- a/registry/src/_data/types.json
+++ b/registry/src/_data/types.json
@@ -58,5 +58,10 @@
     "id": "toolset",
     "label": "Toolsets",
     "icon": "U"
+  },
+  {
+    "id": "retry_policy",
+    "label": "Retry Policies",
+    "icon": "R"
   }
 ]
--- a/scripts/in_container/run_provider_yaml_files_check.py
+++ b/scripts/in_container/run_provider_yaml_files_check.py
@@ -495,8 +495,9 @@ def check_hook_class_name_entries_in_connection_types(yaml_files: dict[str, dict
 
 @run_check("Checking that python-modules in retry-policies exist and belong to provider package")
 def check_retry_policy_modules_exist_and_belong_to_package(yaml_files: dict[str, dict]) -> tuple[int, int]:
-    # retry-policies modules live under policies/, not retry-policies/, so parse_module_data's
-    # glob-based completeness assertion (which assumes directory name == resource_type) does not apply here.
+    # parse_module_data's glob-based completeness check assumes a category's modules live in a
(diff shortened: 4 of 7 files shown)


registry/src/_data/types.json
```diff
@@ -58,5 +58,10 @@
     "id": "toolset",
     "label": "Toolsets",
     "icon": "U"
+  },
+  {
+    "id": "retry_policy",
```
kaxil, line 63: Every other id in this file has a matching `--color-<id>` in `registry/src/css/tokens.css` and a `.tab-icon.<id>` rule in `main.css`, and neither exists for `retry_policy`, so the badge falls through to the bare `.tab-icon` rule with no background. It's visible in your own screenshot: the O/H/@/U pills are colored, R is plain text. #70122 added both lines when it introduced `toolset`.

airflow-core/src/airflow/provider.yaml.schema.json
```diff
@@ -397,6 +397,30 @@
...
             }
         },
+        "retry-policies": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "integration-name": {
+                        "type": "string",
+                        "description": "Integration name. It must have a matching item in the 'integration' section of any provider."
```
kaxil, line 407: Nothing enforces this for `retry-policies`. `check_invalid_integration` in `scripts/in_container/run_provider_yaml_files_check.py` iterates a hardcoded `["sensors", "operators", "hooks", "triggers", "bundles", "toolsets"]`, and #70122 added `toolsets` to it. That same list gates the `python-modules` existence check via `check_if_objects_exist_and_belong_to_package`, so a typo in the module path silently yields an empty category instead of a CI failure. One caveat if you add it: the module-completeness half won't port over verbatim, since `parse_module_data` globs `**/{resource_type}/*.py` and common.ai's policy lives in `policies/retry.py`, not `retry-policies/`.

dev/registry/registry_tools/types.py
```diff
@@ -121,6 +128,7 @@
     ("bundle", "airflow.dag_processing.bundles.base.BaseDagBundle"),
     ("operator", "airflow.sdk.bases.operator.BaseOperator"),
     ("toolset", "pydantic_ai.toolsets.abstract.AbstractToolset"),
+    ("retry_policy", "airflow.sdk.definitions.retry_policy.RetryPolicy"),
```
kaxil, line 131: The provider.yaml validator keeps its own copy of this mapping. `check_all_provider_classes_are_registered` has a `base_class_resource_map` with no `RetryPolicy` entry, and it builds `registered_modules` from a tuple that omits `retry-policies`. So a provider can ship a `RetryPolicy` subclass, forget to declare it, and CI stays green, which is the discoverability gap this PR is closing. #70122 added `(AbstractToolset, "toolsets")` plus the tuple entry. `LLMRetryPolicy` is the only subclass in-tree right now, so adding both shouldn't turn anything red.

registry/src/_data/types.json
Lee-W: Added `--color-retry_policy` to `tokens.css` and the matching `.tab-icon.retry_policy` rule to `main.css`. Thanks!

airflow-core/src/airflow/provider.yaml.schema.json
Lee-W: Added `retry-policies` to the resource-type lists in `check_invalid_integration` and `check_duplicates_in_integrations_names_of_hooks_sensors_operators`, plus a new `check_retry_policy_modules_exist_and_belong_to_package` that runs the `python-modules` existence check.

I did not add `retry-policies` to the list in `check_correctness_of_list_of_sensors_operators_hook_trigger_modules`, so `glob`-based completeness is deliberately not asserted for this section.

dev/registry/registry_tools/types.py
Lee-W: Added `(RetryPolicy, "retry-policies")` to `base_class_resource_map` and `"retry-policies"` to the tuple that builds `registered_modules`. Both were needed — with only the first, the class is found but never counted as registered.

Now if we remove the declaration from `common/ai/provider.yaml`, `LLMRetryPolicy` will be reported as unregistered.

scripts/in_container/run_provider_yaml_files_check.py
```diff
@@ -489,6 +493,25 @@ def check_hook_class_name_entries_in_connection_types(yaml_files: dict[str, dict
     return num_connection_types, num_errors
 
 
+@run_check("Checking that python-modules in retry-policies exist and belong to provider package")
+def check_retry_policy_modules_exist_and_belong_to_package(yaml_files: dict[str, dict]) -> tuple[int, int]:
```
potiuk, line 497: The comment here is doing important work — retry policies live under `policies/retry.py` while every other registry category's modules sit in a directory matching the yaml key (`operators/`, `hooks/`, `sensors/`, `triggers/`), which is why `parse_module_data`'s glob-based completeness check can't be reused and this bespoke check exists.

Not something to change in this PR — the feature already shipped with that layout and moving it would be a breaking import path. But it's worth knowing that adding a second provider with retry policies will hit the same divergence, and the next person adding a registry category may copy this bespoke check rather than the standard one without realising why it's different.

A line in the comment saying "unlike other categories, retry policies intentionally live under `policies/`" would make the *why* explicit for that next reader, rather than just the *what*.

---

potiuk (approved):
Thanks — coherent end-to-end addition: schema, registry tooling, provider data, site data/styling, and a validation check all move together, which is what makes a new registry category actually usable rather than half-wired.

I verified the references resolve rather than assuming: `airflow.sdk.definitions.retry_policy.RetryPolicy` exists (the ABC at line 92), and `providers/common/ai/policies/retry.py` contains `LLMRetryPolicy(RetryPolicy)`, matching the declared `suffixes: ["RetryPolicy"]`. So the registry entry describes something real.

Adding `retry-policies` to the duplicate-detection list alongside the other categories is easy to forget and you didn't — good.

One observation inline about the directory convention.

---

kaxil (approved):

amoghrajesh (approved):
LGTM thanks @Lee-W

scripts/in_container/run_provider_yaml_files_check.py
Lee-W: Reworded to state the rule instead of enumerating categories: `parse_module_data`'s glob assumes a category's modules live in a directory named after its yaml key, and retry policies intentionally live under `policies/` instead — so that check can't be reused and this existence-only check stands in for it.

PR #70504, opened 2026-07-27
Enable provider-based GCS remote logging resolution

### LINKED ISSUE #70266: Add GCSRemoteLogIO.from_config and register gs remote logging scheme

## Why

`from_config()` classmethod. This issue migrates the `gs` scheme.

## What

- [ ] Add `GCSRemoteLogIO.from_config()` in
      `providers/google/src/airflow/providers/google/cloud/log/gcs_task_handler.py`, mirroring the
      legacy branch:
      https://github.com/apache/airflow/blob/104ad12e190db7197fa303d98a3ab68879eafd33/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L207-L224
      — including the `[logging] remote_task_handler_kwargs` IO-kwargs merge, `expanduser` on
      `base_log_folder`, and reading `[logging] google_key_path` (fallback `None`) into
      `gcp_key_path`, so behavior is unchanged for existing configs.
- [ ] Register the `gs` scheme under a `remote-logging:` section in
      `providers/google/provider.yaml` and mirror it in
      `providers/google/src/airflow/providers/google/get_provider_info.py`.
      service or a local equivalent), run a task with remote logging enabled, and confirm logs are
      uploaded and read back through the new dispatch path. Include the setup and verification
      results in the PR description.

## Reference

The legacy branch derives a default connection id from `GCSHook.default_conn_name`; handle the
connection id the same way the merged amazon PRs do (the hook reads
`[logging] remote_log_conn_id` itself).

aaron-y-chen:
closes: #70266
related: #70265
related: #67056

## Summary

Migrate the `gs` remote logging scheme to the provider dispatch introduced by #67056:

- add `GCSRemoteLogIO.from_config()`, preserving the legacy GCS config mapping and IO kwargs;
- register `gs` in the Google provider metadata generated from `provider.yaml`.

Core and the Task SDK can now resolve GCS logging through `ProvidersManager`. The existing legacy path remains as a compatibility fallback.

## Validation

Tested against a real GCS bucket from Breeze using forwarded user ADC and:

```ini
remote_logging = True
remote_base_log_folder = gs://airflow-70266-<random>/airflow-70266
delete_local_logs = True
```

No `remote_log_conn_id` or `google_key_path` was set, so the handler used ADC.

- `gs` resolved to `airflow.providers.google.cloud.log.gcs_task_handler.GCSRemoteLogIO`;
- the task emitted `AIRFLOW_70266_GCS_REMOTE_LOG_E2E_OK`, and `gcloud storage cat` returned the same marker from `attempt=1.log`;
- the corresponding local log no longer existed after upload;
- the Airflow UI reported the `gs://.../attempt=1.log` source and displayed the marker.

Airflow UI reading the task log from GCS after local cleanup:

The uploaded `attempt=1.log` in the real GCS bucket:

---

Generated-by: [GPT 5.6 Sol] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/google/provider.yaml
+++ b/providers/google/provider.yaml
@@ -1489,5 +1489,9 @@ logging:
   - airflow.providers.google.cloud.log.gcs_task_handler.GCSTaskHandler
   - airflow.providers.google.cloud.log.stackdriver_task_handler.StackdriverTaskHandler
 
+remote-logging:
+  - classpath: airflow.providers.google.cloud.log.gcs_task_handler.GCSRemoteLogIO
+    scheme: gs
+
 queues:
   - airflow.providers.google.event_scheduling.events.pubsub.PubSubMessageQueueEventTriggerContainer
--- a/providers/google/src/airflow/providers/google/cloud/log/gcs_task_handler.py
+++ b/providers/google/src/airflow/providers/google/cloud/log/gcs_task_handler.py
@@ -17,6 +17,7 @@
 # under the License.
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 import shutil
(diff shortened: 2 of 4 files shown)


shahar1 (approved):
LGTM — a faithful port of the `gs` scheme onto the #67056 dispatch, matching the merged s3 (#69817) and cloudwatch (#69816) migrations line for line, with real-backend e2e evidence in the description. CI is green.

What I checked beyond the diff itself:

- **Config parity with the legacy branch.** `from_config()` reproduces [`airflow_local_settings.py` L207–L224](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L207-L224) exactly — same four keys, `expanduser` on `base_log_folder`, `google_key_path` → `gcp_key_path`, and the same `| io_kwargs` merge order. Existing `gs://` configs resolve to an identical `GCSRemoteLogIO`.
- **`google_key_path` unset (ADC) doesn't trip `get_mandatory_value`.** `[logging] google_key_path` defaults to `""` rather than `None` in `config.yml`, so the mandatory getter returns a falsy string instead of raising — same as the legacy path, and consistent with the ADC-only run in your description.
- **Dropping the derived default conn id is fine.** The legacy branch called `_default_conn_name_from(..., "GCSHook")`; provider dispatch deliberately doesn't ([`factory.py`](https://github.com/apache/airflow/blob/main/shared/logging/src/airflow_shared/logging/factory.py#L106-L111): *"Provider dispatch does not produce a default conn id … providers that want a backend-specific default can read from their own hook inside `from_config`"*). `GCSRemoteLogIO.hook` already reads `[logging] remote_log_conn_id` itself and falls back to ADC, and the supervisor's `_remote_logging_conn` prefetch only fires when that key is explicitly set — so no behaviour change, and it matches what #69817 shipped.
- **Backward compatibility of the new `provider.yaml` key.** `provider_info.schema.json` does not set `additionalProperties: false` in 2.11.1 (or on `main`), so the `remote-logging` block is ignored rather than rejected on older Airflow. The Compat 2.11.1 / 3.0.6 / 3.1.8 / 3.2.2 / 3.3.0 jobs confirm it.
- **The test guards earn their keep.** The `hasattr(manager, "remote_logging_handler_by_scheme")` skip and `pytest.importorskip("airflow._shared.logging.factory")` read as defensive at first glance, but they're exactly what lets this class run under the provider-compat jobs against Airflow versions predating #67056.

### Smaller observations

- `providers/google/src/airflow/providers/google/cloud/log/gcs_task_handler.py` — this makes three verbatim copies of the ~18-line `remote_task_handler_kwargs` parse-and-split block (s3, cloudwatch, now gcs), with several backends still to migrate under #70265. Factoring it out is awkward while providers still support Airflow 2.11 — `airflow._shared.logging.factory` isn't importable there — so I'd leave this PR as-is and raise the dedup question on the umbrella issue instead. Purely a note, not a change request.

---

> will follow up.*
>

aaron-y-chen:
Thanks shahar! 😁

PR #70506, opened 2026-07-27
Fire failure listeners when TriggerDagRunOperator API call fails

### LINKED ISSUE #63089: UnboundLocalError on TriggerDagRunOperator triggering a dag that doesn't exist
### Apache Airflow version

main (development)

### If "Other Airflow 3 version" selected, which one?

_No response_

### What happened?

When TriggerDagRunOperator operator raises `DagRunTriggerException` in task-sdk it's handled to trigger the dag but when the target dag is invalid then API server returns 404 which is re-raised. Since the Exception is raised under a block that handles the DagRunTriggerException state variable is not set with a value causing finally block to give UnboundLocalError. When TriggerDagRunOperator tries to rerun a dag with same run_id then the conflict status code response from API server is handled. 404 can also be handled in similar manner but this feels like not something limited to TriggerDagRunOperator but in turn could be caused by others too. The other approach would be to initialize state with a sentinel value to emit stats only when it's defined.

_handle_trigger_dag_run raising a ServerResponseError

https://github.com/apache/airflow/blob/fabc5c35da0219f904817e54ee4393f1ae07b101/task-sdk/src/airflow/sdk/execution_time/task_runner.py#L1265-L1266

finally block accessing state that is not initialized

https://github.com/apache/airflow/blob/fabc5c35da0219f904817e54ee4393f1ae07b101/task-sdk/src/airflow/sdk/execution_time/task_runner.py#L1323-L1329

```
::group::Log message source details sources=["/home/karthikeyan/airflow/logs/dag_id=test_non_existent_triggerdagrun/run_id=manual__2026-03-07T17:50:19.584252+00:00/task_id=trigger_dag/attempt=4.log"] 
::endgroup::
[2026-03-07T18:05:25.663612Z] INFO - DAG bundles loaded: dags-folder-1, example_dags
[2026-03-07T18:05:25.664236Z] INFO - Filling up the DagBag from /home/karthikeyan/airflow/dags/non_existent_dagrun.py
[2026-03-07T18:05:25.717248Z] INFO - Task instance is in running state
[2026-03-07T18:05:25.717448Z] INFO -  Previous state of the Task instance: TaskInstanceState.QUEUED
[2026-03-07T18:05:25.717663Z] INFO - Current task name:trigger_dag
[2026-03-07T18:05:25.717759Z] INFO - Dag name:test_non_existent_triggerdagrun
[2026-03-07T18:05:25.718240Z] INFO - Triggering Dag Run. trigger_dag_id=nonexistent_dag_id 
[2026-03-07T18:05:25.734854Z] ERROR - Top level errorUnboundLocalError: cannot access local variable 'state' where it is not associated with a value
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1827 in main
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1325 in run
,AirflowRuntimeError: API_SERVER_ERROR: {'status_code': 404, 'message': 'Server returned error', 'detail': {'detail': {'reason': 'not_found', 'message': "Dag with dag_id: 'nonexistent_dag_id' not found"}}}
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1266 in run
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1396 in _handle_trigger_dag_run
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/comms.py, line 218 in send
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/comms.py, line 312 in _get_response
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/comms.py, line 299 in _from_frame
,DagRunTriggerException: 
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1236 in run
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/execution_time/task_runner.py, line 1653 in _execute_task
    File /home/karthikeyan/stuff/python/airflow/task-sdk/src/airflow/sdk/bases/operator.py, line 443 in wrapper
    File /home/karthikeyan/stuff/python/airflow/providers/standard/src/airflow/providers/standard/operators/trigger_dagrun.py, line 269 in execute
    File /home/karthikeyan/stuff/python/airflow/providers/standard/src/airflow/providers/standard/operators/trigger_dagrun.py, line 297 in _trigger_dag_af_3
[2026-03-07T18:05:25.745036Z] WARNING - Process exited abnormally exit_code=1 
```

_No response_

### How to reproduce

1. Trigger a dag with TriggerDagRunOperator and a target dag that doesn't exist.

```python

from airflow.providers.standard.operators.trigger_dagrun import TriggerDagRunOperator
from airflow.sdk import dag

@dag()
def test_non_existent_triggerdagrun():

    trigger_dag = TriggerDagRunOperator(
        task_id="trigger_dag", trigger_dag_id="nonexistent_dag_id", wait_for_completion=True
    )

test_non_existent_triggerdagrun()
```

test case in tests/task_sdk/api/test_client.py

```python
def test_trigger_dag_not_found(self):
    """Test that if the dag is not found, the client returns an error"""

    def handle_request(request: httpx.Request) -> httpx.Response:
        if request.url.path == "/dag-runs/test_nonexistent_dag/test_run_id":
            return httpx.Response(
                status_code=404,
                json={
                    "detail": {
                        "reason": "not_found",
                        "message": "Dag with dag_id: 'test_nonexistent_dag' not found",
                    }
                },
            )
        return httpx.Response(status_code=422)

    client = make_client(transport=httpx.MockTransport(handle_request))
    result = client.dag_runs.trigger(dag_id="test_nonexistent_dag", run_id="test_run_id")

    assert result == ErrorResponse(error=ErrorType.DAG_NOT_FOUND)
```

### Operating System

Ubuntu 20.04

_No response_

### Deployment

Virtualenv installation

### Deployment details

_No response_

### Anything else?

_No response_

### Code of Conduct

ABiscuitttt:
When `TriggerDagRunOperator` targets a nonexistent Dag (or the API server otherwise rejects the trigger with 403/5xx), the supervisor reports the error by raising `AirflowRuntimeError` from the `TriggerDagRun` comms round-trip. That raise happens inside `run()`'s `except DagRunTriggerException` handler, and an exception raised in an except handler is not caught by the sibling except clauses — it escaped `run()` entirely. `main()` then never reached `finalize()`, the only place `on_task_instance_failed` / `on_failure_callback` listeners run. The task still showed FAILED in the UI (via the supervisor's exit-code fallback), but alerting plugins were silently skipped for exactly this failure mode.

The fix routes the error through the normal failure path (`_handle_current_task_failed`), so the task returns a terminal state, `finalize()` fires the listeners, and retry policy is honored (`UP_FOR_RETRY` when retries remain).

Relationship to prior work on #63089: the reported `UnboundLocalError` symptom was already fixed by #67955 (the `state` variable is now initialized and the `ti.finish` stats emission is guarded), and #63207 was an earlier, unmerged attempt at the full fix. What remains — and what this PR addresses — is the API error escaping `run()` entirely, which silently skips failure listeners, failure callbacks, and retry handling.

Regression tests cover both the no-retry (`FAILED`) and retry (`UP_FOR_RETRY`) paths, asserting the terminal state reaches the supervisor instead of the error escaping `run()`.

closes: #63089

---

- [X] Yes — Kimi Code CLI

Generated-by: Kimi Code CLI following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/newsfragments/70506.bugfix.rst
+++ b/airflow-core/newsfragments/70506.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``on_task_instance_failed`` and ``on_failure_callback`` not firing when ``TriggerDagRunOperator`` fails to trigger a Dag (e.g. target Dag not found). The API error is now routed through the normal task failure path, so listeners fire and retry policy is honored.
--- a/task-sdk/src/airflow/sdk/execution_time/task_runner.py
+++ b/task-sdk/src/airflow/sdk/execution_time/task_runner.py
@@ -1626,7 +1626,7 @@ def run(
         try:
             msg, state = _handle_trigger_dag_run(drte, context, ti, log)
         except AirflowRuntimeError as e:
-            # The supervisor reports API-server errors (target DAG not found, permission
+            # The supervisor reports API-server errors (target Dag not found, permission
             # denied, ...) by raising here. Route through the normal failure path so
             # finalize() fires failure listeners; re-raising would escape run() entirely
             # and skip finalize().
--- a/task-sdk/src/airflow/sdk/execution_time/task_runner.py
+++ b/task-sdk/src/airflow/sdk/execution_time/task_runner.py
@@ -1623,7 +1623,16 @@ def run(
         msg, state = _handle_current_task_success(context, ti)
     except DagRunTriggerException as drte:
         log.info("::group::Post Execute")
-        msg, state = _handle_trigger_dag_run(drte, context, ti, log)
+        try:
+            msg, state = _handle_trigger_dag_run(drte, context, ti, log)
+        except AirflowRuntimeError as e:
+            # The supervisor reports API-server errors (target DAG not found, permission
+            # denied, ...) by raising here. Route through the normal failure path so
+            # finalize() fires failure listeners; re-raising would escape run() entirely
+            # and skip finalize().
+            log.exception("Triggering Dag Run failed")
+            msg, state = _handle_current_task_failed(ti, e, log, context)
+            error = e
     except TaskDeferred as defer:
         log.info("::group::Post Execute")
         msg, state = _defer_task(defer, ti, log)
--- a/task-sdk/tests/task_sdk/execution_time/test_task_runner.py
+++ b/task-sdk/tests/task_sdk/execution_time/test_task_runner.py
@@ -133,6 +133,7 @@ from airflow.sdk.execution_time.comms import (
     PreviousTIResult,
     PrevSuccessfulDagRunResult,
     RescheduleTask,
+    RetryTask,
     SetAssetStateStoreByName,
     SetAssetStateStoreByUri,
     SetRenderedFields,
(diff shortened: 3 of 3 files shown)


ABiscuitttt:
Some context on how we ran into this: we run a global failure-alert listener plugin that pages our on-call channel (Feishu/Lark) through the `on_task_instance_failed` hook when a task instance reaches its final `FAILED` state:

```python
@hookimpl
def on_task_instance_failed(previous_state, task_instance, error):
    if task_instance.state != TaskInstanceState.FAILED:
        return  # still up_for_retry; only alert on final failure
    send_alert(task_instance, error)  # internal notification
```

When a `TriggerDagRunOperator` points at a non-existent Dag, the task dies from the API error without this hook ever firing, so no alert goes out — from the on-call perspective the task fails silently. That is what led to this fix: routing the API error through the normal failure path makes `finalize()` fire the listener (and honor the retry policy) as expected.

ABiscuitttt:
Closing as superseded by #70719, which addresses the same `TriggerDagRunOperator` 404 failure-listener issue more generally.

Issue #70503, opened 2026-07-27
Restore argument-provision checks to __init__ in operators fixed by the burn-down


shahar1:
### Background

`validate-operators-init` flags every read of a template-field parameter in `__init__`. A companion PR narrows that: a `field is None` / `field is not None` check reads whether the argument was *passed*, not its value, and belongs in the constructor. (With `render_template_as_native_obj=True` a provided field renders to `None`, so the same check in `execute()` reports a supplied argument as missing; and raising at construction surfaces a static authoring mistake as a Dag import error rather than per task instance.)

Replaying both checker versions over every exemption entry removed so far (31 PRs, 48 entries) shows **8 entries across 5 PRs** where the old rule forced a change that was not needed. Those PRs are merged and green — nothing is broken — but the files now carry comments and regression tests asserting the placement the docs discourage, and contributors working #70296 learn the rule by reading them.

This issue tracks putting them back. Each is independent; separate PRs are fine.

**Blocked on #70505.** Until it merges, `validate-operators-init` still flags every read of a template-field parameter in `__init__`, so a restored check fails the hook — a revert landed earlier would have to re-add the exemption entry it just removed.

### Revert

- [x] `providers/common/ai/.../operators/agent.py` (#70338) — `message_history is not None and enable_hitl_review` is pure provision. Restore to `__init__`, revert `test_agent.py::test_message_history_with_hitl_review_raises` to constructor-raises. The comment `# message_history is a template field; validate the combination after rendering.` states the rule wrongly and must go.
- [x] `providers/databricks/.../operators/databricks_repos.py` (#70341) — all four `branch`/`tag`/`repo_id`/`repo_path` checks in `DatabricksReposCreate/Update/DeleteOperator` are pure provision. Restore to `__init__`; revert the two tests docstringed *"rejected at execute, after rendering"*.
- [x] `providers/neo4j/.../operators/neo4j.py` (#70348) — `if cypher is None: raise` returns to `__init__`. The deprecation-warning placement is a separate question; #70373's rationale stands independently.

### Split, don't revert

- [x] `providers/common/ai/.../operators/document_loader.py` (#70338) — the `source_path`/`source_bytes` exclusivity is provision and returns to `__init__`. The third check, `source_bytes is not None and file_type is None`, backs the downstream `assert self.file_type is not None`; `file_type` is a template field, so that assumption only holds after rendering and it stays in `execute()`. Update the comment to say which is which — this is the canonical example of the boundary.

[...]

PR #70508, opened 2026-07-27
Fail Neo4jOperator argument mistakes at Dag parse time

ColtenOuO:
> [!NOTE]
> Blocked on #70505. The `validate-operators-init` hook on `main` still flags
> `is None` provision checks in `__init__`, so CI will be red until that merges.
> Kept as a draft until then — verified clean against the checker from #70505
> (exit 0, no exemption entry needed).

Moves the two argument-provision checks of `Neo4jOperator` back to `__init__`, one of the reverts tracked in #70503.

Whether `cypher` or `sql` was passed is a property of how the Dag was written, not of the rendered value. `execute()` cannot answer it: with `render_template_as_native_obj=True` a supplied field can render to `None`, so the check there reports an argument the author did provide as missing. Checking at construction also turns a static authoring mistake into a Dag import error rather than a failure on every task instance.

**Both** provision checks move, not only `cypher is None`. After #70373 the operator keeps `sql` and `cypher` as separate fields, so `Neo4jOperator(sql=...)` is valid with `cypher` unset — the constructor check has to be `cypher is None and sql is None`. Once it is written that way, leaving the mutual-exclusivity check in `execute()` would split two checks that ask the same question across two phases.

The `execute()` guard is kept deliberately: `Neo4jHook.run` takes a `str`, and the constructor check does not narrow `self.cypher` to `str` (it is legitimately `None` when `sql` was used). A passed field can also render to `None`, which is exactly the case the guard covers. `cast` would hide that and `assert` is not allowed in production code.

<details><summary>Testing done</summary>

- `providers/neo4j/tests/unit/neo4j/operators/test_neo4j.py`: 6 passed. The two
  reverted tests assert construction-time failure again; a new test covers the
  `execute()` guard by rendering a supplied field to `None`.
- Checker from #70505 on the operator: exit 0 — no exemption entry needs re-adding.
- Checker on `main`: exit 1, as expected before #70505 merges.
- `breeze run mypy` on the operator: clean.

</details>

related: #70503, #70505

---

Drafted-by: Claude Code (Opus 5); reviewed by @ColtenOuO before posting

### DIFF
--- a/providers/neo4j/src/airflow/providers/neo4j/operators/neo4j.py
+++ b/providers/neo4j/src/airflow/providers/neo4j/operators/neo4j.py
@@ -57,6 +57,10 @@ class Neo4jOperator(BaseOperator):
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
+        if sql is not None and cypher is not None:
+            raise ValueError("Cannot provide both `sql` and `cypher`. Use `cypher` only.")
+        if cypher is None and sql is None:
+            raise ValueError("Parameter `cypher` is required.")
         self.neo4j_conn_id = neo4j_conn_id
         self.cypher = cypher
         self.sql = sql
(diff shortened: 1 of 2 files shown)


ColtenOuO:
Hi @shahar1,

Sorry for the late update, I've changed Pull Request as ready for review.

Thanks again!

shahar1 (approved):
Well done!

PR #70517, opened 2026-07-27
Support a Unix Domain Socket for the metrics clients

### LINKED ISSUE #70666: Support a Unix Domain Socket for the metrics clients
### Description

Airflow submits metrics through a network connection defined by host and port. However, the existing backends also support using an UDS socket but there is no way to configure that at the moment.

A new `[metrics] statsd_socket_path` will be very useful to unlock this.

### Use case/motivation

The motivation is to unlock the advantages of UDS sockets for this:
- Lower overhead
- Better security posture: local-only exposure, filesystem permissions
- No dropped packages

### Related issues

_No response_

### Code of Conduct

sortega:
Adds a \`[metrics] statsd_socket_path\` option so Airflow can submit metrics over a Unix Domain Socket instead of host/port — for both the StatsD and DataDog backends.

Today \`get_statsd_logger\` / \`get_dogstatsd_logger\` only pass \`statsd_host\` / \`statsd_port\`, so deployments that expose the daemon over a UDS — e.g. the Datadog Kubernetes operator, which mounts a socket and sets \`DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket\` — have no way to use it from Airflow. Both \`datadog.DogStatsd\` and the \`statsd\` library (\`UnixSocketStatsClient\`) support a socket path; this exposes it through Airflow config.

\`statsd_host\` and \`statsd_port\` keep their existing defaults (\`localhost\` / \`8125\`). When \`statsd_socket_path\` is set it takes precedence and host/port are ignored:
- **DataDog backend** (\`statsd_datadog_enabled = True\`): builds \`DogStatsd(socket_path=...)\`. Accepts a plain path or a \`unix://\` / \`unixgram://\` / \`unixstream://\` URL.
- **StatsD backend**: builds \`statsd.UnixSocketStatsClient\` instead of the UDP \`StatsClient\`. Expects a plain filesystem path.

When \`statsd_socket_path\` is unset, behavior is identical to before.

- New \`[metrics] statsd_socket_path\` config key (\`default: ~\`).
- \`get_dogstatsd_logger\` and \`get_statsd_logger\` (shared) gain a \`socket_path\` kwarg; when set, the socket-capable client is selected automatically.
- airflow-core and task-sdk wrappers read the new key and select \`UnixSocketStatsClient\` / pass \`socket_path\` to \`DogStatsd\` when it is set.
- Unit tests covering socket_path passthrough for both backends.

Closes #70666

---

- [X] Yes — Claude Code (Opus 4.8 1M)

Generated-by: Claude Code (Opus 4.8 1M) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -1214,6 +1214,18 @@ metrics:
       type: integer
       example: ~
       default: "8125"
+    statsd_socket_path:
```
uranusjr, line 1219: Matching the other two configurations, maybe this should be `statsd_path` instead?

uranusjr (commented):
It is too implicit to me that the path silently takes precedence over host+port. An exception should be raised probably when both are set.

sortega:
> It is too implicit to me that the path silently takes precedence over host+port. An exception should be raised probably when both are set.

We can fix that by making the host/port nullable and removing the default values. But then we need to reintroduce the default values at the metrics loggers because we don't want to break existing configurations that just enable metrics but don't specify host and port.

I'll do that

airflow-core/src/airflow/config_templates/config.yml
xBis7: I think that the name shouldn't change. This is how `statsd` defines it

```python
class UnixSocketStatsClient(StreamClientBase):
    """Unix domain socket version of StatsClient."""

    def __init__(self, socket_path, prefix=None, timeout=None):
```

The variable is actually called `socket_path`.

xBis7:
Having configured all of the options `host`, `port` and `socket_path` isn't fatal.

For datadog, the same class accepts all of them and then silently gives precedence to the `socket_path`.

For statsd, each config option is used for a different class but all of them could co-exist. If both classes are instantiated, the same metric will be emitted twice which isn't optimal or desired but still there won't be an issue for the system.

IMO, we shouldn't set `host` and `port` to None when `socket_path` is also provided. For statsd, we should just give precedence to the `socket_path` and log a warning while for datadog, we should just log a warning.

sortega:
@xBis7, makes sense. No exception anymore.

airflow-core/src/airflow/config_templates/config.yml
sortega: I agree. I'm keeping `statsd_socket_path`

airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -1195,11 +1195,14 @@ metrics:
...
-        Specifies the host address where the StatsD daemon (or server) is running
+        Specifies the host address where the StatsD daemon (or server) is running. Left
+        unset by default so the DataDog client can fall back to its own environment
+        variables; the StatsD client defaults to ``localhost``. If ``statsd_socket_path`` is also
+        set, the socket takes precedence and a warning is logged.
       version_added: 2.0.0
       type: string
       example: ~
-      default: "localhost"
+      default: ~
```
xBis7: Now all three, host, port and socket path, default to empty. Is this the desired behavior? It seems to me that it will break existing deployments.

I think we should still keep the default for host and port to not affect any users that rely on them but don't specify a particular value and make the new addition fallback to empty if not provided. If provided, then we can check host and port and log a warning if they have a value.

xBis7: From the user perspective, their understanding should be that the behavior is the same but now there is a new option which will take precedence if specified, otherwise nothing changes.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -31,11 +31,54 @@
...
+    socket_path = conf.get("metrics", "statsd_socket_path", fallback=None)
+    host = (
+        conf.get("metrics", "statsd_host", fallback=None)
+        if conf.has_option("metrics", "statsd_host")
+        else None
+    )
+    port = (
+        conf.getint("metrics", "statsd_port", fallback=None)
+        if conf.has_option("metrics", "statsd_port")
+        else None
```
uranusjr: This would crash with the default (not explicitly configured) since *None* still makes `has_option` return True, but that wouldn’t be parsable by `getint`. (This would not be an issue if the suggestions in https://github.com/apache/airflow/pull/70517/changes#r3721048824 are applied.)

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
```diff
@@ -34,23 +35,25 @@ def get_statsd_logger() -> SafeStatsdLogger:
...
             raise AirflowConfigException(
-                "Your custom StatsD client must extend the statsd.StatsClient in order to ensure "
-                "backwards compatibility."
+                "Your custom StatsD client must extend the statsd.StatsClient or "
+                "statsd.UnixSocketStatsClient in order to ensure backwards compatibility."
             )
         log.info("Successfully loaded custom StatsD client")
 
     else:
         stats_class = StatsClient
```
uranusjr: This probably should be changed to match? https://github.com/apache/airflow/pull/70517/changes#diff-8a528bc2492e4fcc8361aeb91b6014d77b09d1ad190c3f75a7d9b8f087ddc6bcR51

(This also raises the issue whether we need some mechanism to keep the two sides in sync. Maybe out of scope of this PR, or maybe can be fixed right here right now.)

airflow-core/src/airflow/config_templates/config.yml
sortega: OK, back to having statsd_socket_path having priority over host and port

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: OK, back to having statsd_socket_path having priority over host and port

airflow-core/newsfragments/70517.feature.rst
uranusjr, line 1: This now contradicts the implementation in multiple parts. We might not even need it at all since the new feature is not that major.

The PR description also needs updating.

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
sortega: Absolutely

airflow-core/newsfragments/70517.feature.rst
sortega: Updated

sortega:
@uranusjr, @xBis7, I think it should be fine now. I'm keeping the default values and make it explicit that the new setting has more priority

airflow-core/src/airflow/observability/metrics/statsd_logger.py
```diff
@@ -31,26 +31,28 @@
...
+    socket_path = conf.get("metrics", "statsd_socket_path", fallback=None)
 
     # no need to check for the scheduler/statsd_on -> this method is only called when it is set
     # and previously it would crash with None is callable if it was called without it.
-    from statsd import StatsClient
+    from statsd import StatsClient, UnixSocketStatsClient
 
     if stats_class:
-        if not issubclass(stats_class, StatsClient):
+        if not (issubclass(stats_class, StatsClient) or issubclass(stats_class, UnixSocketStatsClient)):
```
uranusjr:
```suggestion
        if not issubclass(stats_class, (StatsClient, UnixSocketStatsClient)):
```

More importantly, passing this check doesn't mean the class works in socket mode. A custom client that subclasses `StatsClient` is accepted, and `get_statsd_logger` then calls it as `stats_class(socket_path=..., prefix=...)`. Combining `statsd_custom_client_path` with `statsd_socket_path` is an unhandled `TypeError` at startup, with nothing pointing at the two settings that conflict.

Could we require `UnixSocketStatsClient` specifically when `socket_path` is set, and raise `AirflowConfigException` otherwise?

airflow-core/src/airflow/observability/metrics/statsd_logger.py
```diff
@@ -31,26 +31,28 @@
...
-                "Your custom StatsD client must extend the statsd.StatsClient in order to ensure "
-                "backwards compatibility."
+                "Your custom StatsD client must extend the statsd.StatsClient or "
+                "statsd.UnixSocketStatsClient in order to ensure backwards compatibility."
             )
         log.info("Successfully loaded custom StatsD client")
 
     else:
-        stats_class = StatsClient
+        stats_class = StatsClient if socket_path is None else UnixSocketStatsClient
```
uranusjr: This line, the `issubclass` check above it, and the error message are now byte-identical copies of Task SDK's `get_statsd_logger`. 

`get_statsd_logger` in `shared` already receives `socket_path`. Could `stats_class` default to `None` there and let the shared builder pick `UnixSocketStatsClient` vs `StatsClient`? That puts the choice (and the compatibility check from my other comment) in one place, and it also makes the PR
description accurate.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -171,12 +173,15 @@ def get_statsd_logger(
...
 ) -> SafeStatsdLogger:
     """Return logger for StatsD."""
-    statsd = stats_class(
-        host=host,
-        port=port,
-        prefix=prefix,
-        ipv6=ipv6,
-    )
+    if socket_path is not None:
+        statsd = stats_class(socket_path=socket_path, prefix=prefix)
```
uranusjr: We should add a comment in `config.yml` that `ipv6` is silently dropped in this mode.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -31,11 +31,12 @@
 )
 
 if TYPE_CHECKING:
-    from statsd import StatsClient
+    from statsd import StatsClient, UnixSocketStatsClient
 
     from .protocols import DeltaType
     from .validators import ListValidator
 
+
```
uranusjr:
```suggestion
```

airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -1209,11 +1210,24 @@ metrics:
...
+      description: |
+        Path to a Unix Domain Socket used to submit metrics, instead of ``statsd_host``
+        and ``statsd_port``. Applies to both the StatsD and DataDog backends. Takes
+        precedence over ``statsd_host`` / ``statsd_port`` when set. For the DataDog
+        backend (``statsd_datadog_enabled = True``)
+        a plain path or a ``unix://`` / ``unixgram://`` / ``unixstream://`` URL is
+        accepted; for the StatsD backend a plain filesystem path is expected.
+      version_added: 3.4.0
+      type: string
+      example: "unix:///var/run/datadog/dsd.socket"
```
uranusjr: This example only works for the DataDog backend because Datadog’s `_get_uds_socket` knows to strip the prefix. The `statsd` library does not. The description below explains the difference, but the example is the part people
copy.

```suggestion
      example: "/var/run/datadog/dsd.socket"
```

shared/observability/tests/observability/metrics/test_stats.py
```diff
@@ -71,6 +71,17 @@ def test_passes_ipv6_as_keyword(self, mock_stats_client):
         statsd_logger.get_statsd_logger(stats_class=mock_stats_client, ipv6=True)
         assert mock_stats_client.call_args.kwargs["ipv6"] is True
 
+    def test_socket_path_constructs_stats_class_with_socket_path(self):
+        # In socket mode the builder constructs the given stats_class with socket_path/prefix
+        # (host/port are omitted). The caller is responsible for passing a socket-capable class.
```
uranusjr:
```suggestion
```

airflow-core/src/airflow/observability/metrics/statsd_logger.py
sortega: Absolutely. I've also moved the check to `metrics/statsd_logger.py`

airflow-core/src/airflow/observability/metrics/statsd_logger.py
sortega: Done that

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: Documented as such

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: Removed

airflow-core/src/airflow/config_templates/config.yml
sortega: 👍

sortega:
I've addressed those comments, @uranusjr

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -171,12 +173,31 @@ def get_statsd_logger(
...
-        host=host,
-        port=port,
-        prefix=prefix,
-        ipv6=ipv6,
-    )
+    if socket_path is not None:
+        if stats_class is None:
+            stats_class = UnixSocketStatsClient
+        elif isinstance(stats_class, type) and not issubclass(stats_class, UnixSocketStatsClient):
+            raise ValueError(
```
uranusjr: Shouldn’t this still raise AirflowConfigException?

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -22,6 +22,8 @@
 from functools import wraps
 from typing import TYPE_CHECKING, TypeVar, cast
 
+from statsd import StatsClient, UnixSocketStatsClient
```
uranusjr, line 25: These need fallbacks because statsd may not be available if the user opts for datadag.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -158,9 +159,10 @@ def timer(
 
 def get_statsd_logger(
     *,
-    stats_class: type[StatsClient],
+    stats_class: type[StatsClient] | type[UnixSocketStatsClient] | None = None,
```
uranusjr: I wonder if this should also accept `Callable` instead since non-type inputs is explicitly allowed (there are checks against this).

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
```diff
@@ -16,41 +16,22 @@
...
-    # and previously it would crash with None is callable if it was called without it.
-    from statsd import StatsClient
-
-    if stats_class:
-        if not issubclass(stats_class, StatsClient):
-            raise AirflowConfigException(
-                "Your custom StatsD client must extend the statsd.StatsClient in order to ensure "
-                "backwards compatibility."
-            )
-        log.info("Successfully loaded custom StatsD client")
```
uranusjr, line 45: We should still emit this log.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: I think that's not importable from this module

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: ~~The current code has a `issubclass` check. Allowing `Callable` will be moot as we require the subclassing. Do you want to drop that check?~~

Allowing the callable was a mistake

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
sortega: That's right. I've restored it.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: Move them back to `if TYPE_CHECKING` and used a lazy import in the function

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: As I understand it, we have these options:
- Make shared/observability depend on shared/configuration and use AirflowConfigException
- Use a different exception, either generic like `ValueError`
- Introduce a specific one like `InvalidCustomStatsClient`

sortega:
@uranusjr, do you want me to go for the observability lib depending on the config lib?

xBis7:
@sortega Thanks for updating the PR!

If you need a new exception type, then it needs to be added here instead of adding a new dependency

https://github.com/sortega/airflow/blob/sortega/DNAWF-4497_dd_dogstats_url/shared/observability/src/airflow_shared/observability/exceptions.py

But, I don't think we should do that. `airflow-core` and `task-sdk` both initialize their own config parser and then re-use some parts from`shared/configuration`. Because the `conf` object is different for these main packages, config is read twice under `airflow-core` and `task-sdk` and then passed to `shared`. E.g.

https://github.com/apache/airflow/pull/70517/changes#diff-8a528bc2492e4fcc8361aeb91b6014d77b09d1ad190c3f75a7d9b8f087ddc6bcR33

https://github.com/apache/airflow/pull/70517/changes#diff-fba7485cf1bc00467f34ed32e234b40b630fe7f0ff52df832d27f4c10b679418R34

So, if you need to do a validation which will throw an `AirflowConfigException`, then that should happen in the factory methods and then the validated value should be passed to the shared code.

You don't have to move the entire thing, just do a quick validation regarding the config value and the rest stays the same.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
uranusjr: You don’t need to really “depend” on anything since all shared libs are, well, shared across Airflow components, and `configuration` is already available everywhere `observability`. All you need to do is to import like this

```python
from ...configuration import AirflowConfigException
```

There are existing cases using this pattern in other shared libs.

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
```diff
@@ -23,32 +23,49 @@
...
-        if not issubclass(stats_class, StatsClient):
+    if custom_class is not None:
+        if socket_path is not None:
+            if not (isinstance(custom_class, type) and issubclass(custom_class, UnixSocketStatsClient)):
+                raise AirflowConfigException(
+                    "Your custom StatsD client must extend the statsd.UnixSocketStatsClient "
+                    "when using a socket path in order to ensure backwards compatibility."
+                )
+        elif not (
+            isinstance(custom_class, type) and issubclass(custom_class, (StatsClient, UnixSocketStatsClient))
```
uranusjr: I don’t think this is right? When we reach here, `socket_path` is not None, so UnixSocketStatsClient should not be checked here?

Also, the `isinstance(custom_class, type)` part seems excessive. A non-type input here would simply cause a `TypeError` on `issubclass`, which would be clean enough for users to know they should have passed in a type. This is also documented anyway.

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
```diff
@@ -23,32 +23,49 @@
...
+        return statsd_logger.get_socket_statsd_logger(
+            stats_class=custom_class,
+            socket_path=socket_path,
+            prefix=conf.get("metrics", "statsd_prefix"),
+            influxdb_tags_enabled=conf.getboolean("metrics", "statsd_influxdb_enabled", fallback=False),
+            statsd_disabled_tags=conf.get("metrics", "statsd_disabled_tags", fallback=None),
+            metrics_allow_list=conf.get("metrics", "metrics_allow_list", fallback=None),
+            metrics_block_list=conf.get("metrics", "metrics_block_list", fallback=None),
+            stat_name_handler=conf.getimport("metrics", "stat_name_handler"),
+            statsd_influxdb_enabled=conf.getboolean("metrics", "statsd_influxdb_enabled", fallback=False),
```
uranusjr: Using the same relative import technique, we should be able to move all these into `get_socket_statsd_logger` instead so they are not repeated.

Alternatively, at least do

```python
logger_kwargs = {
    "prefix": conf.get("metrics", "statsd_prefix"),
    ...
}
if socket_path is None:
    return statsd_logger.get_udp_statsd_logger(**kwargs)
return statsd_logger.get_socket_statsd_logger(socket_path=socket_path, **kwargs)
```

sortega: Sure, nice refactor

task-sdk/src/airflow/sdk/observability/metrics/statsd_logger.py
sortega: Yes, removing the type check and the UnixSocketStatsClient

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: I managed to keep the check in the other file

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
uranusjr:
```suggestion
```

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -42,6 +42,36 @@ then add the following lines to your configuration file e.g. ``airflow.cfg``
...
+    [metrics]
+    statsd_on = True
+    statsd_socket_path = /var/run/statsd/statsd.sock
+    statsd_prefix = airflow
+
+When ``statsd_socket_path`` is set, ``statsd_host``, ``statsd_port``, and
+``statsd_ipv6`` are ignored.
+
+The standard StatsD backend uses a stream Unix socket. When the DataDog backend
+is enabled, both stream and datagram Unix sockets are supported:
```
uranusjr, line 62: This “datagram” is probably being caught as a spelling error and failing CI. Add it to `spelling_wordlist.txt` to fix the error.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -42,6 +42,36 @@ then add the following lines to your configuration file e.g. ``airflow.cfg``
...
+
+.. code-block:: ini
+
+    [metrics]
+    statsd_datadog_enabled = True
+    statsd_socket_path = /var/run/datadog/dsd.socket
+    statsd_prefix = airflow
+
+For maximum compatibility, configure a plain filesystem path. The DataDog
```
uranusjr: We currently use “DataDog” and “Datadog” in different parts of the documentation. Worth unifying.

uranusjr (approved):
A few minor issues this need to fix before the PR is mergeable.

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: Done

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
sortega: Done

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
sortega: Datadog is the right one: https://www.datadoghq.com/about/resources/
Done

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -170,21 +191,47 @@ def get_statsd_logger(
...
 ) -> SafeStatsdLogger:
-    """Return logger for StatsD."""
-    statsd = stats_class(
-        host=host,
-        port=port,
-        prefix=prefix,
-        ipv6=ipv6,
+    """Return a UDP StatsD logger."""
+    # Local import to avoid requiring statsd when other backends are used (e.g. Datadog)
+    from statsd import StatsClient as _StatsClient
```
xBis7: Isn't the import under the type checking block enough?

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
```diff
@@ -170,21 +191,47 @@ def get_statsd_logger(
...
     statsd_influxdb_enabled: bool = False,
 ) -> SafeStatsdLogger:
-    """Return logger for StatsD."""
-    statsd = stats_class(
-        host=host,
-        port=port,
-        prefix=prefix,
-        ipv6=ipv6,
+    """Return a UDP StatsD logger."""
+    # Local import to avoid requiring statsd when other backends are used (e.g. Datadog)
```
xBis7: If we end up in this file, then the user has already enabled statsd in the config. I would expect statsd to be available, but even if for some reason it isn't and it's configured to be used, then it should fail. Isn't it a valid scenario for a failure?

xBis7 (approved):
I have 2 non-blocking questions. LGTM!

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: No, because it's used it line 198.

Repro:
```python
from typing import TYPE_CHECKING


if TYPE_CHECKING:
    import os

module = os
```

xBis7: ok, thanks for checking!

shared/observability/src/airflow_shared/observability/metrics/statsd_logger.py
sortega: I suppose it's too conservative. I've added a commit with the direct import

sortega:
@uranusjr, @xBis7 good to merge?

airflow-core/newsfragments/70517.feature.rst
hussein-awala, line 1: Could you remove this please? it's reserved for significant changes, this PR introduces a new feature.

airflow-core/src/airflow/observability/metrics/datadog_logger.py
```diff
@@ -30,6 +30,7 @@ def get_dogstatsd_logger() -> SafeDogStatsdLogger:
         tags_in_string=conf.get("metrics", "statsd_datadog_tags"),
         host=conf.get("metrics", "statsd_host"),
         port=conf.getint("metrics", "statsd_port"),
```
hussein-awala, line 32: should we add a fallback here since they are optional now (when `statsd_socket_path` is not None)? same for task-sdk

airflow-core/src/airflow/observability/metrics/statsd_logger.py
```diff
@@ -23,40 +23,48 @@
...
-    # no need to check for the scheduler/statsd_on -> this method is only called when it is set
-    # and previously it would crash with None is callable if it was called without it.
-    from statsd import StatsClient
+    socket_path = conf.get("metrics", "statsd_socket_path", fallback=None)
+    custom_class = conf.getimport("metrics", "statsd_custom_client_path", fallback=None)
 
-    if stats_class:
-        if not issubclass(stats_class, StatsClient):
+    if custom_class is not None:
+        if socket_path is not None:
```
hussein-awala, line 40: I think this will not work as expected if someone deactivate the feature using env vars: `AIRFLOW__METRICS__STATSD_SOCKET_PATH=""`, you can normalize blank/whitespace-only values to None.

hussein-awala (commented):
Overall the change looks good, add a few comments, LGTM once they are addressed.

airflow-core/src/airflow/observability/metrics/datadog_logger.py
sortega: These have non-null defaults (localhost and 8125)

airflow-core/src/airflow/observability/metrics/statsd_logger.py
sortega: Yep. I've added an `or None` to both places so we normalize `""` into `None`

airflow-core/newsfragments/70517.feature.rst
sortega: Removed

hussein-awala (approved):

PR #70529, opened 2026-07-27
Validate storage transfer job body after template rendering

mitre88:

`CloudDataTransferServiceCreateJobOperator` lists `body` in `template_fields` but deep-copies and validates it in `__init__`. With a fully templated `body` (a Jinja expression or XComArg), the validator ran against the un-rendered expression, so `TransferJobValidator`'s checks — the AWS-credential restriction and the single-data-source rule — were silently bypassed. The deep copy and validation now run at the start of `execute()`, against the rendered value, right before `TransferJobPreprocessor` mutates the body.

Added a test constructing the operator with a templated `body` that renders to a body embedding AWS credentials — with the previous implementation the credential check never fires and the job is created; now it raises before calling the hook. The class is removed from the exemption list and the `validate-operators-init` check passes locally.

Per the discussion in #70505 this is a genuine value read (validation of the rendered dict), not an argument-provision check.

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/cloud_storage_transfer_service.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/cloud_storage_transfer_service.py
@@ -257,19 +257,19 @@ class CloudDataTransferServiceCreateJobOperator(GoogleCloudBaseOperator):
     ) -> None:
         super().__init__(**kwargs)
         self.body = body
-        if isinstance(self.body, dict):
-            self.body = deepcopy(body)
         self.aws_conn_id = aws_conn_id
         self.gcp_conn_id = gcp_conn_id
         self.api_version = api_version
         self.project_id = project_id
         self.google_impersonation_chain = google_impersonation_chain
-        self._validate_inputs()
 
     def _validate_inputs(self) -> None:
         TransferJobValidator(body=self.body).validate_body()
 
     def execute(self, context: Context) -> dict:
+        if isinstance(self.body, dict):
+            self.body = deepcopy(self.body)
+        self._validate_inputs()
         TransferJobPreprocessor(body=self.body, aws_conn_id=self.aws_conn_id).process_body()
         hook = CloudDataTransferServiceHook(
             api_version=self.api_version,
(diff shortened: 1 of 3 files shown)


providers/google/src/airflow/providers/google/cloud/operators/cloud_storage_transfer_service.py
```diff
@@ -257,19 +257,19 @@ def __init__(
...
         self.project_id = project_id
         self.google_impersonation_chain = google_impersonation_chain
-        self._validate_inputs()
 
     def _validate_inputs(self) -> None:
         TransferJobValidator(body=self.body).validate_body()
 
     def execute(self, context: Context) -> dict:
+        if isinstance(self.body, dict):
+            self.body = deepcopy(self.body)
```
potiuk, line 271: The deepcopy is load-bearing and its purpose isn't obvious from here — `TransferJobPreprocessor.process_body()` calls `_inject_aws_credentials()` and `_reformat_schedule()`, which mutate the body **in place**. Without the copy, the operator would write AWS credentials into the caller's own dict.

Two things worth doing:

1. **A one-line comment** saying why the copy exists. The original had the copy sitting next to the assignment in `__init__` where the reader could at least see it was defensive; now it's a bare reassignment in `execute` that looks removable. This is precisely the kind of line someone "simplifies" away later.
2. **A test for it.** The new test covers the validation-raises path, but nothing asserts the behaviour the deepcopy is there for. Something like: build the operator with a dict containing an `awsAccessKey`, run a successful `execute`, and assert the dict you passed in is unchanged. Without that, the copy could be dropped in a future refactor and the whole suite would stay green while the operator started mutating user data.

---

potiuk (approved):
Thanks — correct move. `body` is in `template_fields`, so `_validate_inputs()` in `__init__` was running `TransferJobValidator` against the un-rendered value; a templated body could never pass validation and a bad one broke Dag parsing instead of failing the task. Deferring both the deepcopy and the validation to `execute` fixes that, and removing the exemptions entry closes it out.

Good that the test asserts `create_transfer_job.assert_not_called()` — that shows it bails before touching the API.

One coverage gap inline.

---

shahar1 (approved):

PR #70531, opened 2026-07-27
Validate Cloud Function deploy body after template rendering

mitre88:

`CloudFunctionDeployFunctionOperator` lists `body`, `location`, and `api_version` in `template_fields`, but `__init__` validated the body, ran `ZipPathPreprocessor.preprocess_body()` (which both validates and mutates the body), and constructed `GcpBodyFieldValidator` pinned to the un-rendered `api_version`. A fully templated `body` crashed with `AttributeError` at parse time inside the preprocessor. All of this now runs at the start of `execute()` against the rendered values.

The missing-`location`/`body` truthiness checks and the zip-path exclusivity rules depend on rendered values (an expression rendering to an empty string must count as missing), so per the discussion in #70505 these are genuine value reads, not provision checks.

Tests: converted the four construction-time raise tests to execute-time, and added `test_templated_body_deploys_after_rendering`, which constructs the operator with a templated `body` (previously a parse-time crash) and deploys once the field holds the rendered value — it fails against the previous implementation. The class is removed from the exemption list and the `validate-operators-init` check passes locally.

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/functions.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/functions.py
@@ -171,12 +171,9 @@ class CloudFunctionDeployFunctionOperator(GoogleCloudBaseOperator):
         self.gcp_conn_id = gcp_conn_id
         self.api_version = api_version
         self.zip_path = zip_path
-        self.zip_path_preprocessor = ZipPathPreprocessor(body, zip_path)
+        self.validate_body = validate_body
         self._field_validator: GcpBodyFieldValidator | None = None
         self.impersonation_chain = impersonation_chain
-        if validate_body:
-            self._field_validator = GcpBodyFieldValidator(CLOUD_FUNCTION_VALIDATION, api_version=api_version)
-        self._validate_inputs()
         super().__init__(**kwargs)
 
     def _validate_inputs(self) -> None:
@@ -227,6 +224,12 @@ class CloudFunctionDeployFunctionOperator(GoogleCloudBaseOperator):
         }
 
     def execute(self, context: Context):
+        self.zip_path_preprocessor = ZipPathPreprocessor(self.body, self.zip_path)
+        if self.validate_body:
+            self._field_validator = GcpBodyFieldValidator(
+                CLOUD_FUNCTION_VALIDATION, api_version=self.api_version
+            )
+        self._validate_inputs()
         hook = CloudFunctionsHook(
             gcp_conn_id=self.gcp_conn_id,
             api_version=self.api_version,
--- a/providers/google/tests/unit/google/cloud/operators/test_functions.py
+++ b/providers/google/tests/unit/google/cloud/operators/test_functions.py
@@ -80,10 +80,25 @@ class TestGcfFunctionDeploy:
             op.execute(None)
 
     def test_body_empty(self):
+        op = CloudFunctionDeployFunctionOperator(
+            project_id="test_project_id", location="test_region", body={}, task_id="id"
+        )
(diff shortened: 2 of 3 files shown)


providers/google/src/airflow/providers/google/cloud/operators/functions.py
```diff
@@ -227,6 +224,12 @@ def extra_links_params(self) -> dict[str, Any]:
         }
 
     def execute(self, context: Context):
+        self.zip_path_preprocessor = ZipPathPreprocessor(self.body, self.zip_path)
```
potiuk, line 227: `self.zip_path_preprocessor` is now only assigned here, so between construction and the first `execute` the attribute doesn't exist at all — any access raises `AttributeError` rather than returning something sensible. That affects subclasses, tests, and anything introspecting the operator between parse and run.

Note you've already handled the sibling case correctly: `self._field_validator: GcpBodyFieldValidator | None = None` is still initialised in `__init__` and merely *reassigned* in `execute`. Doing the same for the preprocessor — declare it as `None` in `__init__`, build it here — would make the two consistent and keep the operator's attribute surface stable from construction onwards.

---

potiuk (approved):
Thanks — correct move. `body` is a template field, so `GcpBodyFieldValidator` in `__init__` was validating the un-rendered value: a templated body could never pass validation, and a literal bad one broke Dag parsing instead of failing the task. Constructing the validator and the `ZipPathPreprocessor` in `execute` and running `_validate_inputs()` there is the right shape.

`test_templated_body_deploys_after_rendering` is a good addition — asserting `create_new_function.assert_called_once()` after a rendered body proves the happy path still works, not just that the failure path raises.

One robustness point inline about an attribute that no longer exists until `execute` runs.

---

providers/google/tests/unit/google/cloud/operators/test_functions.py
```diff
@@ -80,10 +80,25 @@ def test_missing_fields(self, mock_hook, body, message):
...
+            project_id="test_project_id", location="test_region", body={}, task_id="id"
+        )
         with pytest.raises(AirflowException):
-            CloudFunctionDeployFunctionOperator(
-                project_id="test_project_id", location="test_region", body={}, task_id="id"
-            )
+            op.execute(None)
+
+    @mock.patch("airflow.providers.google.cloud.operators.functions.CloudFunctionsHook")
+    def test_templated_body_deploys_after_rendering(self, mock_hook):
```
shahar1, line 90: Test passes on unmodified main, which means that it doesn't check an actual regression

shahar1 (approved):

shahar1:
**Edit:** few corrections I made after review (drafted by Claude Code & Codex).

- body: The code expected a dictionary, but received text such as "{{ var.value.body }}". Checking whether that text contains sourceUploadUrl usually returns false, so the operator incorrectly decides it does not need to upload the zip. If the template’s variable name happens to contain a Cloud Functions field name, the same checks can instead produce a TypeError or a misleading conflict error. The original claim that this always caused an AttributeError was wrong.
location: A template expression is a nonempty string, so it passes the original “is a location provided?” check—even if its eventual value is empty.
- api_version: The validator receives the literal template expression instead of something like "v1beta2". It therefore skips checks that apply to that particular API version.
- The test list is also stale. `test_templated_body_deploys_after_rendering` passed against
unmodified `main` and was replaced by three tests that each fail without this change:
`test_templated_body_with_zip_path_uploads_after_rendering`,
`test_templated_location_rendered_empty_raises` and
`test_templated_api_version_validates_after_rendering` — one per templated field.

PR #70533, opened 2026-07-27
Emit GCSListObjectsOperator delimiter deprecation warning after rendering

yuseok89:
related: #70296

The delimiter template field was inspected in __init__, so the deprecation warning fired on the unrendered Jinja value instead of the resolved one.
Move the check to execute() so it reflects the value the task actually runs with.

---

  - Opus 4.8

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/gcs.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/gcs.py
@@ -229,18 +229,21 @@ class GCSListObjectsOperator(GoogleCloudBaseOperator):
         super().__init__(**kwargs)
         self.bucket = bucket
         self.prefix = prefix
-        if delimiter:
-            warnings.warn(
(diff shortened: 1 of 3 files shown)


potiuk (approved):
Thanks — clean and minimal. `delimiter` is a template field, so warning about it in `__init__` checked the raw `"{{ ... }}"` string: a templated delimiter never triggered the deprecation notice, and a literal one warned on every Dag-processor parse rather than once per task run.

Two things I appreciated: you used the same `_warn_on_deprecated_template_fields()` helper name as #70449 and #70542, which makes the pattern greppable across the campaign rather than each operator inventing its own shape; and the test wraps the existing `execute` call in `pytest.warns` with `context=mock.MagicMock()` rather than `None` — several other PRs in this batch pass `None` and only get away with it because nothing reads the context before the warning.

Nothing further from me.

---

PR #70536, opened 2026-07-27
Keep Grid scroll position when opening a task.

Ei-Sandi:
## Summary

In the Dag details view, the Grid on the left keeps its own vertical scroll position. Clicking a task that is scrolled below the fold currently snaps the Grid back to the top the **first** time you do it — losing your place. This PR keeps the scroll position across that navigation, while still resetting to the top when you leave the Dag entirely and come back.

## Problem

On a Dag with enough tasks to overflow the Grid, scroll down and click a task that is off the top of the viewport. The list jumps back to `scrollTop = 0` (the first task) even though the layout looks unchanged. Clicking a *second* task (task → task) does **not** reset the scroll — it stays put. So the reset happens exactly once, on the first click, and it gets more painful the longer the list is (e.g. when task groups are expanded). 

#### Original 

https://github.com/user-attachments/assets/f4eec01c-efe5-4e1a-9f2d-286633942a5e

#### Current

https://github.com/user-attachments/assets/f44ec326-7e56-4fb7-89be-6f555381dd5a

## Root cause

The Grid lives inside `DetailsLayout`, and **every** top-level page renders its own `DetailsLayout`:

- `dags/:dagId` → `<Dag>`
- `dags/:dagId/tasks/:taskId` → `<Task>`
- `dags/:dagId/runs/:runId` → `<Run>`
- `dags/:dagId/runs/:runId/tasks/:taskId` → `<TaskInstance>`
- `<GroupTaskInstance>`, `<MappedTaskInstance>`

These are **sibling routes** in `router.tsx`, not nested under a shared layout.

- **First click** (`<Dag>` → `<Task>`): the matched route element changes, so React unmounts `<Dag>` (destroying its `DetailsLayout`, its `Grid`, and the scroll container DOM node) and mounts a fresh `<Task>`. The new scroll container starts at `scrollTop = 0`, and `useVirtualizer` reads `0` on mount and renders the top rows.
- **Subsequent clicks** (`<Task>` → `<Task>`): the matched element type is unchanged; only the `:taskId` param updates. React keeps the instance and its DOM, so the scroll position survives.

That asymmetry — cross-element on the first hop, same-element afterwards — is why it resets exactly once. It is not task-group specific; groups just make the list longer so more scroll is lost.

## Solution (this PR)

Remember the Grid's `scrollTop` across the remount and restore it, scoped to a single Dag visit.

- **Persist + restore** — `useGridScrollRestoration` saves `scrollTop` (in a module-scoped `Map` keyed by `dagId`, which survives the remount) on scroll, and restores it once, after the rows exist, in a `useLayoutEffect` (before
  paint, no flash). Covers both the plain-Grid and the shared Grid + Gantt scroll containers.
- **Reset on leave** — `useResetGridScrollOnLeave` (called once from the always-mounted `BaseLayout`) clears the store whenever the route is not a Dag detail path (`/dags/<id>…`). So scroll is kept only *within* a Dag; leaving to
  the Dags list / home / assets and returning starts at the top.

| Navigation | Scroll |
|---|---|
| Dag → Task → Run → Task Instance (same `dagId`) | preserved |
| Dag → Dags list / home / assets → back to Dag | reset to top |
| First visit to a Dag | starts at top |

Files: `useGridScrollRestoration.ts` (+ test), `Grid.tsx` (wire-up), and one line in `BaseLayout.tsx` (reset-on-leave). Routing and the panel layout are untouched.

## Alternative considered (not taken)

**Make `DetailsLayout` a single shared parent route** so the Grid mounts once and persists across Dag ↔ Run ↔ Task ↔ TaskInstance (the "correct" React Router structure), with per-page content flowing through the existing `<Outlet/>`.

Rejected as disproportionate: it is a large refactor with real regression risk.

## Testing

New unit tests cover:
  - **Persist/restore:** save-on-scroll, restore-once-rows-exist, deferred restore when rows arrive later, restore fires only once (does not clobber later user scrolling), no-op when there is no saved offset / a zero offset / a missing scroll element, and listener cleanup on unmount.
  - **Reset-on-leave:** `isDagDetailsPath` for detail vs non-detail paths, `clearGridScrollOffsets` drops every entry, and the hook clears on a non-detail path but keeps offsets while inside a Dag (via `MemoryRouter`).

---

Co-authored-by: [Claude Opus 5] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/layouts/Details/Grid/useGridScrollRestoration.ts
+++ b/airflow-core/src/airflow/ui/src/layouts/Details/Grid/useGridScrollRestoration.ts
@@ -1,0 +1,103 @@
+/*!
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { useEffect, useLayoutEffect, useRef } from "react";
+import { useLocation } from "react-router-dom";
+
+/**
+ * Grid vertical scroll offset per ``dagId``, kept at module scope so it survives the
+ * remount that happens when navigating between the sibling ``<Dag>`` / ``<Run>`` /
+ * ``<Task>`` / ``<TaskInstance>`` routes — each mounts its own ``Grid``, which would
+ * otherwise start at ``scrollTop = 0``.
+ */
+const gridScrollOffsetByDagId = new Map<string, number>();
+
+// Matches a specific Dag (`/dags/<id>...`) but not the `/dags` list.
+const DAG_DETAILS_PATH = /^\/dags\/(?<dagId>[^/]+)/u;
+
+export const extractDagIdFromPath = (pathname: string): string | undefined =>
+  DAG_DETAILS_PATH.exec(pathname)?.groups?.dagId;
+
+export const saveGridScrollOffset = (dagId: string, offset: number): void => {
+  if (dagId !== "") {
+    gridScrollOffsetByDagId.set(dagId, offset);
+  }
+};
+
+export const readGridScrollOffset = (dagId: string): number | undefined => gridScrollOffsetByDagId.get(dagId);
+
+export const clearGridScrollOffsets = (): void => gridScrollOffsetByDagId.clear();
+
+type Params = {
+  readonly dagId: string;
+  readonly getScrollElement: () => HTMLElement | null;
+  /** Number of rendered rows — restore waits until the list has rows (and thus height). */
+  readonly rowCount: number;
+};
+
+/**
+ * Preserves the grid's vertical scroll position across the route transitions that
+ * remount the grid. Saves ``scrollTop`` on every scroll and restores it once, after
+ * the rows exist, so the remount is invisible to the user.
+ */
+export const useGridScrollRestoration = ({ dagId, getScrollElement, rowCount }: Params): void => {
+  const hasRestoredRef = useRef(false);
+
+  useEffect(() => {
+    const element = getScrollElement();
+
+    if (element === null || dagId === "") {
+      return undefined;
+    }
+
+    const handleScroll = () => saveGridScrollOffset(dagId, element.scrollTop);
+
+    element.addEventListener("scroll", handleScroll, { passive: true });
+
+    return () => element.removeEventListener("scroll", handleScroll);
+  }, [dagId, getScrollElement]);
+
+  // Layout effect (not a plain effect) so the position is set before paint, avoiding
+  // a visible flash at the top of the list on remount.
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/ui/src/layouts/Details/Grid/useGridScrollRestoration.ts
```diff
@@ -0,0 +1,103 @@
...
+ * ``<Task>`` / ``<TaskInstance>`` routes — each mounts its own ``Grid``, which would
+ * otherwise start at ``scrollTop = 0``.
+ */
+const gridScrollOffsetByDagId = new Map<string, number>();
+
+// Matches a specific Dag (`/dags/<id>...`) but not the `/dags` list.
+const DAG_DETAILS_PATH = /^\/dags\/(?<dagId>[^/]+)/u;
+
+export const extractDagIdFromPath = (pathname: string): string | undefined =>
+  DAG_DETAILS_PATH.exec(pathname)?.groups?.dagId;
```
bbovenzi, line 34: We can just call `const { dagId } = useParams();`

bbovenzi (changes requested):
Thanks! Actually I think the single shared parent is an interesting route to go. It could help further reduce re-renders. 

But let's not store a scrollTop but instead only try to scroll to a task when there is one selected with something like `rowVirtualizer.scrollToIndex()`. I think we already have the taskId+runId exist as element ids. Then we have no additional state and its possible to deep-link. Also, the scrollTop approach would easily lose sync with task groups expanding/collapsing.

airflow-core/src/airflow/ui/src/layouts/Details/Grid/useGridScrollRestoration.ts
Ei-Sandi: @bbovenzi Hi, I looked into the option, `rowVirtualizer.scrollToIndex()` won't work as we have to choose start, center, end or auto and we cannot get back the exact position for the task as when we clicked. 

Is there any alternative would you like to suggest ?

dsuhinin: @bbovenzi do you think we can try to move with the current approach?

bbovenzi: Shouldn't we know where in the grid's structure the task in question sits and which task groups are open or closed? Then all the rows are a fixed height so we can calculate the height offset. "center" is fine after that

The problem with storing a px top offset is that if the dag structure changes between versions or the expanded task groups change then it breaks again.

dsuhinin: Ill try to catch this changes and a bit refactor. thank you!

dsuhinin: hey @bbovenzi! I believe I've fixed it here -> https://github.com/apache/airflow/pull/73123. Can you please take a look when you will have some time. thank you!

dsuhinin: hey @bbovenzi ! Can you please take a look into the new approach -> https://github.com/apache/airflow/pull/73123. We don't store a raw pixel top-offset as the source of truth. On the remount that opening a task triggers, we restore the last scrollTop only to avoid the Grid jumping; and because that remount never changes the structure, the offset still lines up. If the structure does change (new version, or groups expanded/collapsed), the restored offset no longer matches, so we fall back to locating the     selected task by id in the flattened rows and, since rows are fixed-height, use the index to center it — no reliance on the stored pixels. So the failure mode you flagged is handled by the index path.

bbovenzi:
Closing in favor of https://github.com/apache/airflow/pull/73123

PR #70537, opened 2026-07-27
Use built-in exceptions for Postgres hook input validation

mitre88:

`PostgresHook` raised `AirflowException` for two input-validation failures: a `sqlalchemy_query` connection extra that is not a dict (now `TypeError`), and requesting `realdictcursor` with psycopg3 (now `ValueError`, matching the invalid-cursor branch immediately below it that already raises `ValueError`). The `known_airflow_exceptions.txt` entry for the file drops from 2 to 0, and the existing test now asserts the specific exception type and message.

---

### DIFF
--- a/generated/known_airflow_exceptions.txt
+++ b/generated/known_airflow_exceptions.txt
@@ -376,7 +376,6 @@ providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py::1
 providers/opensearch/src/airflow/providers/opensearch/operators/opensearch.py::9
 providers/pagerduty/src/airflow/providers/pagerduty/hooks/pagerduty.py::1
 providers/pagerduty/src/airflow/providers/pagerduty/hooks/pagerduty_events.py::2
-providers/postgres/src/airflow/providers/postgres/hooks/postgres.py::2
 providers/presto/src/airflow/providers/presto/hooks/presto.py::1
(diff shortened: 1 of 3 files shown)


potiuk (approved):
Thanks — and the type selection here is better than most of this migration.

Splitting the two raises by kind rather than defaulting both to `ValueError` is the right call: `sqlalchemy_query` not being a dict is a **type** error, while `realdictcursor` is a valid-type-but-unsupported **value**. That distinction usually gets flattened. The `realdictcursor` change also now matches its immediate neighbour, which already raises `ValueError(f"Invalid cursor passed {_cursor}...")` — so the two adjacent failure paths finally agree.

Removing the ratchet entry outright is right too: it was `::2` and both occurrences are gone, so the line goes rather than being decremented to a stale count. The `match=` on the updated test is a genuine strengthening.

One small gap: the `realdictcursor` branch has **no test asserting the new type**. The only `realdict` reference in `test_postgres.py` is a `monkeypatch.setattr(..., "RealDictCursor", None)` at line 69 — nothing exercises the raise. So half this PR is covered and half isn't. It's a message-preserving type change so the risk is low, and a two-line `pytest.raises(ValueError, match="realdictcursor is not supported")` around `hook._get_cursor("realdictcursor")` would close it whenever you next touch the file. Not blocking.

---

PR #70539, opened 2026-07-27
Use ValueError when no ArangoDB collection operation is specified

mitre88:

`ArangoDBCollectionOperator.execute` raised `AirflowException` when none of the insert/update/replace/delete operations was specified — a plain input-validation error, now a `ValueError`. The `known_airflow_exceptions.txt` entry for the file drops from 1 to 0. The raise had no test coverage, so the PR adds one asserting the new type and that no hook operation is invoked.

---


potiuk (approved):
Thanks — best of your three exception PRs in this batch.

`ValueError` is right: "at least one operation must be specified" is a caller-argument problem, not an Airflow-runtime condition. Import cleaned up, ratchet entry removed outright rather than decremented.

What sets this one apart is that it **adds a test that didn't exist**. #70540 and #70537 updated assertions on paths already covered; here `test_no_operation_fails` is new coverage for a branch that had none — with `match=` pinning the message and `insert_documents.assert_not_called()` proving the operator bails before touching the hook rather than merely raising somewhere. That last assertion is the one people skip.

---

PR #70540, opened 2026-07-27
Use ValueError for unsupported recursive glob in ADLS upload

mitre88:

`LocalFilesystemToADLSOperator.execute` raised `AirflowException` when `local_path` contains an unsupported `**` recursive glob — a plain input-validation error, now a `ValueError`. The `known_airflow_exceptions.txt` entry for the file drops from 1 to 0, and the existing test asserts the new type.

---


potiuk (approved):
Thanks — textbook version of this migration. An unsupported glob pattern is a caller-supplied argument error, which is exactly what `ValueError` is for, and the message is unchanged so nothing user-facing shifts except the type.

Three things done right that often get missed in this campaign: the now-unused `AirflowException` import is removed rather than left dangling; the `known_airflow_exceptions.txt` entry is deleted outright (count 1 -> 0) rather than decremented to a stale zero; and the test gains a `match=` argument, so it now pins the message as well as the type — strictly stronger than before.

Worth noting for completeness: `ValueError` isn't a subclass of `AirflowException`, so a Dag wrapping this operator in `except AirflowException` stops catching it. That's inherent to the migration rather than anything wrong here, and the ratchet file exists precisely to track it — just flagging it's a real behaviour change, not a pure refactor.

---

PR #70544, opened 2026-07-27
Rename StepFunctionStartExecutionOperator's input field to match its constructor argument

bramhanandlingala:
Fixes the amazon provider's StepFunctionStartExecutionOperator entry from the #70296 exemption-list burn-down.

`state_machine_input` is a template field, but `__init__` assigned it to `self.input` — a name that doesn't match either the constructor argument (`state_machine_input`) or the declared template field, so the `validate-operators-init` prek hook flagged it as a missing/invalid assignment.

Renamed the internal attribute (and the corresponding entry in `template_fields`) from `input` to `state_machine_input` so it matches the constructor argument directly. No transformation or validation was involved — this was a pure naming mismatch, not a rendering-order bug — and the public constructor signature is unchanged, so this isn't a breaking change for existing DAGs.

- Renamed `self.input` → `self.state_machine_input` in `__init__` and `execute()`.
- Updated `template_fields` accordingly.
- Updated the one test assertion that referenced `op.input`.

Verified locally that `scripts/ci/prek/validate_operators_init.py` reports zero findings for this class after the change, and that the existing unit tests in `test_step_function.py` pass unmodified aside from the one updated attribute assertion.

Related to #70296

- [X] Yes [Claude]


airflow-core/src/airflow/jobs/scheduler_job_runner.py
vincbeck: Seems very unrelated

vincbeck (changes requested):

vincbeck (approved):

PR #67881, opened 2026-06-02
Drain result_queue in LocalExecutor to prevent process join deadlock

SakshamKapoor2911:
closes: #67870

### Description
This PR fixes a deadlock in the `LocalExecutor` during executor shutdown (`end()`).

Currently, the executor calls `proc.join()` on active worker processes. However, if a worker has written enough results to the `result_queue` to fill the OS-level pipe buffer (typically 64KB), the worker process blocks indefinitely on its `put()` call. 
Because the parent scheduler process is blocked on the unbounded `proc.join()` and not reading from `result_queue`, a classic multiprocessing deadlock occurs. The scheduler hangs indefinitely, stopping heartbeats and preventing systemd/Kubernetes from restarting the process (as it never exits).

### Changes
1. **Draining during Join:** Updated the shutdown loop in `LocalExecutor.end()` to continuously drain the `result_queue` using `_read_results()` while waiting for worker processes to join with a small timeout.
2. **Robust Queue Reading:** Wrapped the queue drainage loop in `_read_results()` with a `try/except (OSError, EOFError)` block to gracefully handle cases where pipes are already broken or closed during task exit.
3. **Forced Terminate:** Implemented the `LocalExecutor.terminate()` method to forcefully kill any remaining workers when a hard stop is requested.

### Context
This contribution was created as part of the required capstone of the CodePath AI301 course, where students learn how to make responsible and effective use of generative AI tools for open-source software contributions.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (Antigravity IDE powered by Gemini)

Generated-by: Antigravity IDE (Gemini) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

---

> [!NOTE]
> **✅ Ready for review** · @SakshamKapoor2911 → `@potiuk` · 2026-06-12 13:30 UTC
>
> Thanks @SakshamKapoor2911 — all checks are green and this PR is marked **ready for maintainer review**. The ball is with the maintainers now; a maintainer will take the next look.
>
> <sub>_Automated triage — may be imperfect._</sub>

PR #70545, opened 2026-07-27
Drain result_queue before dispatch to prevent runtime deadlock in LocalExecutor

### LINKED ISSUE #70526: Scheduler deadlocks in pipe write during task dispatch (LocalExecutor) — runtime variant of #67881

Airflow Core

### Apache Airflow version

3.3.0

**Issue Description**

The scheduler permanently deadlocks during **normal task dispatch** (not shutdown), blocked in a pipe `write()` to the LocalExecutor's activity queue. All dispatch halts; the SchedulerJob heartbeat stops while the systemd service stays "active". ~10 minutes later, every running task's supervisor kills its task ("Server indicated the task shouldn't be running anymore"), burning retry attempts.

This appears to be the **runtime/dispatch-path variant of the shutdown deadlock fixed in #67881** (merged 2026-07-02, not yet released — 3.3.0 still ships the pre-fix `end()` and empty `terminate()`).

Occurred **4 times in ~14 hours** on one production host once dispatch volume rose (~150–250 concurrent tasks). Identical signature each time.

**Evidence captured mid-deadlock:**
- `/proc/<scheduler-pid>/stack` → blocked in `pipe_write`
- `/proc/<scheduler-pid>/syscall` → `write(fd=9)`; fd 9 → `pipe:[<inode>]`
- Scanning `/proc/*/fd`: the **same pipe inode held (read fd 7 + write fd 9) by the scheduler and every LocalExecutor worker process**
- Host healthy: 100+ GB free RAM, idle CPU, no OOM, clean kernel log
- One occurrence happened with **zero** api-server errors/timeouts, ruling out API-side causes

**Steps to reproduce:**

Race condition — probabilistic and load-correlated, so not minimally deterministic, but reliably triggered in our environment within minutes-to-hours:

1. LocalExecutor, `parallelism = 256`, default `max_tis_per_query` (512), PostgreSQL backend
2. Generate dispatch bursts of 150–250 mostly lightweight, long-running tasks (ours poll remote Spark jobs), with tasks completing (returning results) concurrently with new dispatch
3. Scheduler eventually blocks forever in `pipe_write`; heartbeat stops

Notably there is no clean load threshold: the same host survived a 30-minute soak at 256 running tasks, then deadlocked at ~40 running during a later dispatch wave — consistent with a timing race on the shared queues, not resource exhaustion.

Suggested synthetic repro: a DAG with ~300 mapped `time.sleep(300)` tasks, cleared repeatedly to force dispatch bursts while earlier tasks are completing.

**Mitigations tried (none sufficient):**

| Change | Result |
|---|---|
| `[api] workers` 4 → 12 | fixed unrelated api saturation; deadlock recurred with 0 timeouts |
| `default_pool` 300 → 150 | reduced frequency; recurred |
| `[scheduler] max_tis_per_query` 512 → 32 | reduced frequency; recurred |
| Pausing the high-volume DAGs | stable (current workaround, at large throughput cost) |

The scheduler should never block indefinitely on executor queue writes — a full result/activity queue should trigger draining (or a bounded/non-blocking put with retry), so a slow consumer can't freeze the entire scheduling loop.

**Suspected mechanism** — same as #67881 describes, but hit from the dispatch side:

1. Workers fill `result_queue`'s OS pipe buffer (~64KB) and block on `put()`
2. Blocked workers stop draining `activity_queue`
3. Scheduler blocks in `_process_workloads()` → `self.activity_queue.put(workload)` (`airflow/executors/local_executor.py`)
4. Being single-threaded at that point, the scheduler never reaches `_read_results()`, so `result_queue` never drains → circular wait, permanent

#67881 adds result-queue draining during `end()`/join, but `_process_workloads()` still calls a blocking `put()` with no draining fallback, so the runtime deadlock remains possible on `main` as well.

### Operating System

Amazon Linux 2023

### Deployment

Virtualenv installation

_No response_

n/a — core issue (apache-airflow-task-sdk 1.3.0 for reference)

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

Standalone deployment: systemd units for scheduler (LocalExecutor), api-server, and dag-processor on a single EC2 host; PostgreSQL metadata DB.

Happy to provide full `/proc` captures, journald excerpts of the supervisor mass-terminations, and timing data from all 4 occurrences on request.

Mainly using Claude to diagnose & type this  report out. 
- Looked myself for existing GitHub issues, and the one I linked was the closest thing I could find to ours. 
- Looked myself at our Airflow server, and found running tasks randomly killed (with the 
```
[2026-07-27 13:41:09] INFO - spark app is running...check in again in 120
[2026-07-27 13:43:15] INFO - spark app is running...check in again in 120
[2026-07-27 13:45:20] INFO - spark app is running...check in again in 120
[2026-07-27 13:47:24] INFO - spark app is running...check in again in 120
[2026-07-27 13:49:29] INFO - spark app is running...check in again in 120
[2026-07-27 13:50:40] ERROR - Server indicated the task shouldn't be running anymore. Terminating process detail={"reason":"not_found","message":"Task Instance not found, it may have been moved to the Task Instance History table"`)
```
message). This is what Claude diagnoses as an effect of a scheduler deadlock

### Code of Conduct

SakshamKapoor2911:
Fixes a runtime deadlock in LocalExecutor where the scheduler blocks indefinitely on activity_queue.put() during task dispatch, a gap left by #67881 which only addressed the shutdown path.

Workers fill the result_queue pipe buffer and block on put(). Blocked workers stop consuming activity_queue, so the scheduler blocks on activity_queue.put() and never reaches _read_results() to drain the pipe — a circular deadlock.

The fix: drain result_queue via _read_results() before each activity_queue.put() in _process_workloads, using the same drain-before-write pattern already applied to end().

closes: #70526

---

- [X] Yes — OpenCode (DeepSeek v4 Pro)

Generated-by: OpenCode (DeepSeek v4 Pro) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/tests/unit/executors/test_local_executor.py
+++ b/airflow-core/tests/unit/executors/test_local_executor.py
@@ -369,43 +369,15 @@ class TestLocalExecutor:
 
         assert len(executor.event_buffer) == result_count
 
-    def test_process_workloads_drains_results_before_put(self):
-        executor = LocalExecutor(parallelism=1)
-        executor.activity_queue = mock.MagicMock()
-        executor.result_queue = mock.MagicMock()
-        executor._unread_messages = mock.MagicMock()
-        executor._check_workers = mock.MagicMock()
-        workload = mock.MagicMock()
-        workload.key = TaskInstanceKey("test_dag", "test_task", "test_run")
-        executor.queued_tasks = {workload.key: workload}
-
-        call_order = []
-        orig_read_results = executor._read_results
-        orig_put = executor.activity_queue.put
-
-        def tracking_read_results():
-            call_order.append("_read_results")
-            return orig_read_results()
-
-        def tracking_put(*args, **kwargs):
-            call_order.append("put")
-            return orig_put(*args, **kwargs)
-
-        executor._read_results = tracking_read_results
-        executor.activity_queue.put = tracking_put
-
-        executor._process_workloads([workload])
-
-        assert call_order == ["_read_results", "put"]
-
-    def test_process_workloads_drains_results_for_multiple_workloads(self):
+    @pytest.mark.parametrize("num_workloads", [1, 3])
+    def test_process_workloads_drains_results_before_put(self, num_workloads):
         executor = LocalExecutor(parallelism=1)
         executor.activity_queue = mock.MagicMock()
         executor.result_queue = mock.MagicMock()
         executor._unread_messages = mock.MagicMock()
         executor._check_workers = mock.MagicMock()
         workloads_list = []
-        for i in range(3):
+        for i in range(num_workloads):
             workload = mock.MagicMock()
             workload.key = TaskInstanceKey("test_dag", f"test_task_{i}", "test_run")
             workloads_list.append(workload)
@@ -428,7 +400,7 @@ class TestLocalExecutor:
 
         executor._process_workloads(workloads_list)
 
-        assert call_order == ["_read_results", "put", "_read_results", "put", "_read_results", "put"]
+        assert call_order == ["_read_results", "put"] * num_workloads
 
     @pytest.mark.parametrize(
         ("conf_values", "expected_server"),
--- a/airflow-core/tests/unit/executors/test_local_executor.py
+++ b/airflow-core/tests/unit/executors/test_local_executor.py
@@ -369,6 +369,67 @@ class TestLocalExecutor:
 
         assert len(executor.event_buffer) == result_count
 
+    def test_process_workloads_drains_results_before_put(self):
+        executor = LocalExecutor(parallelism=1)
+        executor.activity_queue = mock.MagicMock()
+        executor.result_queue = mock.MagicMock()
+        executor._unread_messages = mock.MagicMock()
+        executor._check_workers = mock.MagicMock()
+        workload = mock.MagicMock()
+        workload.key = TaskInstanceKey("test_dag", "test_task", "test_run")
+        executor.queued_tasks = {workload.key: workload}
+
+        call_order = []
+        orig_read_results = executor._read_results
+        orig_put = executor.activity_queue.put
+
+        def tracking_read_results():
+            call_order.append("_read_results")
+            return orig_read_results()
+
+        def tracking_put(*args, **kwargs):
+            call_order.append("put")
+            return orig_put(*args, **kwargs)
+
+        executor._read_results = tracking_read_results
+        executor.activity_queue.put = tracking_put
+
+        executor._process_workloads([workload])
+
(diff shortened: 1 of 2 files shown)


SakshamKapoor2911:
@eladkal @amoghrajesh — all CI checks pass. Fixes a LocalExecutor shutdown deadlock. Ready for maintainer review.

vatsrahul1001 (approved):

vatsrahul1001:
LGTM! should be good to merged post code owner review

vatsrahul1001:

SakshamKapoor2911:
@vatsrahul1001 thanks for the approval and for moving this to the 3.3.2 milestone.

@XD-DENG @ashb @o-nikolas @pierrejeambrun @hussein-awala @dheerajturaga CI is green; PR needs a code owner sign-off to merge. It is a relatively small fix; could one of you take a look when you have a moment?

airflow-core/tests/unit/executors/test_local_executor.py
```diff
@@ -369,6 +369,67 @@ def test_end_drains_result_queue_to_avoid_join_deadlock(self):
...
+            call_order.append("put")
+            return orig_put(*args, **kwargs)
+
+        executor._read_results = tracking_read_results
+        executor.activity_queue.put = tracking_put
+
+        executor._process_workloads(workloads_list)
+
+        assert call_order == ["_read_results", "put", "_read_results", "put", "_read_results", "put"]
+
```
pierrejeambrun: nit: Tests are very similar, can we merge them into a parameterized test.

pierrejeambrun (approved):

airflow-core/tests/unit/executors/test_local_executor.py
SakshamKapoor2911: Done. I unified both cases into a single `@pytest.mark.parametrize("num_workloads", [1, 3])` test.

darren-cheung-sa:
> Moving to 3.3.2 as this is still pending code owner review and do not want to rush on merging this as it not critical

Seems it wasn't included in 3.3.2. And looks like there are no deadlines for 3.3.3 or 3.4.0

For our system (using LocalExecutor), it is a bit urgent since it means there is always a risk of a deadlock happening (stopping all our DAGs)

darren-cheung-sa:
@vatsrahul1001 CC on [this](https://github.com/apache/airflow/pull/70545#issuecomment-5187505961)

Issue #70267, opened 2026-07-23
Add StackdriverRemoteLogIO.from_config and register stackdriver scheme


jason810496:
Part of #70265 (related: #67056).

## Why

#67056 decoupled remote logging from the hardcoded branches in `airflow_local_settings.py`: core and
the Task SDK now resolve the handler via `ProvidersManager` dispatch on the
`[logging] remote_base_log_folder` URL scheme, instantiating the provider class through a no-arg
`from_config()` classmethod. This issue migrates the `stackdriver` scheme.

## What

- [ ] Add `StackdriverRemoteLogIO.from_config()` in
      `providers/google/src/airflow/providers/google/cloud/log/stackdriver_task_handler.py`,
      mirroring the legacy branch:
      https://github.com/apache/airflow/blob/104ad12e190db7197fa303d98a3ab68879eafd33/airflow-core/src/airflow/config_templates/airflow_local_settings.py#L250-L268
      — including the `[logging] remote_task_handler_kwargs` IO-kwargs merge and `expanduser` on
      `base_log_folder`, so behavior is unchanged for existing configs.
- [ ] Register the `stackdriver` scheme under a `remote-logging:` section in
      `providers/google/provider.yaml` and mirror it in
      `providers/google/src/airflow/providers/google/get_provider_info.py`.
- [ ] Add tests mirroring `TestS3RemoteLogIOFromConfig` in
      `providers/amazon/tests/unit/amazon/aws/log/test_s3_task_handler.py`.
- [ ] Verify end to end with a real system test: set up the actual backend yourself (a real
      service or a local equivalent), run a task with remote logging enabled, and confirm logs are
      uploaded and read back through the new dispatch path. Include the setup and verification
      results in the PR description.

## Notes

- The legacy branch parses the log name from the URL path: `stackdriver:///airflow-tasks` →
  `gcp_log_name="airflow-tasks"` (`urlsplit(...).path[1:]`). Mirror this in `from_config`, with a
  clear `ValueError` when no log name can be derived (see how #69816 handles `log_group_arn`
  parsing for `cloudwatch`).
- It also reads `[logging] google_key_path` (fallback `None`) into `gcp_key_path`, and passes no
  `remote_base` — unlike the other object-storage backends.

## Reference

Merged examples to follow: #69817 (s3), #69816 (cloudwatch). If `from_config` raises on a bad
config, the shared factory falls back to the legacy path, so this is not a breaking change.

aaron-y-chen:
[...]

PR #70525, opened 2026-07-27
Add ElasticsearchRemoteLogIO.from_config and register elasticsearch scheme

yuseok89:
closes: #70271
related: #70265
related: #67056

Migrates the Elasticsearch remote-logging backend to the provider dispatch mechanism: adds `ElasticsearchRemoteLogIO.from_config()` and registers an `elasticsearch` scheme, so `[logging] remote_base_log_folder = elasticsearch://` routes through the provider.

The legacy host-based selection still works via the fallback in `airflow_local_settings.py`, so existing configs are unaffected.

### End-to-end verification
`breeze start-airflow --integration elasticsearch`, `remote_base_log_folder = elasticsearch://`, ran a task with remote logging enabled.

**Written to Elasticsearch:**
```
[Breeze:3.10.20] root@e01884b83d1e:/opt/airflow$ curl -s "http://elasticsearch:9200/_cat/indices?v" | grep airflow-logs
yellow open   airflow-logs lejo5BB3RJa-3h0J0fzmBA   1   1         10            0     20.7kb         20.7kb       20.7kb
[Breeze:3.10.20] root@e01884b83d1e:/opt/airflow$ curl -s "http://elasticsearch:9200/airflow-logs/_search?size=1&pretty" | head -40
{
  "took" : 1104,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 10,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "airflow-logs",
        "_id" : "urj5o58BYIM3Bhrt3lZ7",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : "2026-07-27T14:27:55.643862Z",
          "level" : "info",
          "event" : "::group::Pre Execute",
          "logger" : "task",
          "filename" : "task_runner.py",
          "lineno" : 2356,
          "log_id" : "example_xcom-push_sample_xcoms-manual__2026-07-27T14:27:53.127874+00:00--1-1",
          "offset" : 1
        }
      }
    ]
  }
}
```

**Read back in the UI:**

image

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes (please specify the tool below)
  - Opus 4.8

---

[...]

PR #70549, opened 2026-07-27
Register Stackdriver remote task logging handler

fat-catTW:
Register Stackdriver task remote logging with the provider remote logging dispatch registry.

This lets `remote_base_log_folder = stackdriver:///...` resolve through the `ProvidersManager` scheme dispatch path to `StackdriverRemoteLogIO.from_config()`, instead of relying on the legacy logging config fallback.

The change also teaches `StackdriverRemoteLogIO.from_config()` to build itself from Airflow logging configuration, including the Stackdriver log name derived from `remote_base_log_folder`.

related: #70267

---

### DIFF
--- a/providers/google/provider.yaml
+++ b/providers/google/provider.yaml
@@ -1495,6 +1495,8 @@ logging:
 remote-logging:
   - classpath: airflow.providers.google.cloud.log.gcs_task_handler.GCSRemoteLogIO
     scheme: gs
+  - classpath: airflow.providers.google.cloud.log.stackdriver_task_handler.StackdriverRemoteLogIO
+    scheme: stackdriver
 
 queues:
   - airflow.providers.google.event_scheduling.events.pubsub.PubSubMessageQueueEventTriggerContainer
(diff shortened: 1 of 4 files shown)


potiuk (approved):
Faithful on all four fields, including the `urlsplit(...).path[1:]` derivation that turns `stackdriver:///airflow-tasks` into `airflow-tasks`, and the `| io_kwargs` merge that legacy applies for this backend.

The guard is a genuine improvement rather than just a port:

```python
if not log_name:
    raise ValueError(
        "Cannot derive a Stackdriver log name from "
        f"logging/remote_base_log_folder: {remote_base_log_folder!r}"
    )
```

Legacy had nothing here, so `remote_base_log_folder = stackdriver://` with no path silently produced an empty `gcp_log_name` and a handler that could not work. Turning that into a startup error that names the offending config value is the right call, and parametrizing the rejection test over the bad forms is the right way to pin it.

Reviewing this alongside the sibling ports, the cluster made opposite choices on superficially similar "config is empty" cases — #70525 falls back to a class default host, this one raises. Both are correct for their situation: an empty Elasticsearch host has a sensible default, an empty Stackdriver log name does not. Worth noting only because the divergence is deliberate rather than accidental.

Adding a second entry under the existing `remote-logging` block alongside `gs` also confirms multiple schemes per provider work, which is useful to have demonstrated.

---

PR #70550, opened 2026-07-27
Move keycloak JWT tokens to separate cookies

### LINKED ISSUE #61771: Keycloak provider login fails if access token is too large because of cookie size limit

keycloak

apache-airflow-providers-keycloak==0.4.1

### Apache Airflow version

3.1.5

### Operating System

Ubuntu 22.04.1 LT

### Deployment

Virtualenv installation

### Deployment details

Packages airflow-core and apache-airflow-providers-keycloak installed via pip in virtualenv. Keycloak config is pointing to Keycloak instance deployed on OpenShift. The Keycloak access token is quite large because multiple realm roles are included in access token.

### What happened

After logging in via Keycloak SSO the Airflow UI enters into a 401 unauthorized loop. After successfull auth code flow, the cookie with access token + refresh token is not passed to Airflow UI because they exceed the cookie size limit of 4KB.

Image

After Keycloak login, the Airflow home page should display without any 401 unauthorized errors.

### How to reproduce

Login with a Keycloak user having a large access token, e.g., a user with 15+ roles, so that access token + refresh token exceed cookie size limit of 4kB.

### Anything else

_No response_

### Code of Conduct

stephen-bracken:
Move Keycloak JWT tokens out of Airflow JWT
---

closes: #61771

The Keycloak `access_token` and `refresh_token` fields of the Airflow JWT make the cookie set in the browser very large. Most browsers use a cookie size limit of 4096 bytes by default, so with a modest number of claims on the Keycloak JWTs the Airflow JWT will easily fill this limit. If we convert the `KeycloakAuthManagerUser` into a pydantic model, FastAPI can automatically insert the values of cookies into the appropriate fields when the user is being requested.

https://fastapi.tiangolo.com/tutorial/cookie-param-models/#cookies-with-a-pydantic-model

- [X] No

---

### DIFF
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
@@ -25,3 +25,11 @@ CONF_REALM_KEY = "realm"
 CONF_SERVER_URL_KEY = "server_url"
 CONF_REQUESTS_POOL_SIZE_KEY = "requests_pool_size"
 CONF_REQUESTS_RETRIES_KEY = "requests_retries"
+
+# Extra Cookie names
+COOKIE_NAME_ACCESS_TOKEN = "access_token"
+COOKIE_NAME_ID_TOKEN = "_id_token"
+COOKIE_NAME_NAME = "name"
+COOKIE_NAME_OAUTH_STATE = "_oauth_state"
+COOKIE_NAME_REFRESH_TOKEN = "refresh_token"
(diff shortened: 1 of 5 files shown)


vincbeck (approved):
Code looks pretty solid!

stephen-bracken:
@vincbeck I need to do a followup PR on this, I don't think the cookie is being retrieved correctly here unfortunately

vincbeck:
Please do. Please also do some testing before, that would avoid this kind of situation

PR #70341, opened 2026-07-23
Validate Databricks Repos operators' template fields after rendering

1fanwang:
`branch`, `tag` and `repo_path` are template fields, so they are rendered after `__init__` runs. The Create, Update and Delete Repos operators enforced their mutual-exclusivity and presence combinations in the constructor, acting on the un-rendered Jinja expressions — so a templated `branch`/`tag`/`repo_path` couldn't resolve before the check ran. Move those checks into `execute()`, preserving order. The `git_url`/`git_provider` detection reads no template field and stays in `__init__`.

related: #70296

<details><summary>Testing Done</summary>

Updated each operator's `test_init_exception` to construct first, then assert the `AirflowException` surfaces from `execute()` (covering all combinations: branch+tag, neither; repo_id+repo_path, neither). All three fail on the pre-fix source (the constructor raises first) and pass after. `test_databricks_repos.py`: 11 passed. `validate_operators_init.py` exits 0; the `raise AirflowException` count is unchanged.

</details>

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (please specify the tool below)

Generated-by: GitHub Copilot CLI following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70551, opened 2026-07-27
Restore Dag-parse-time validation for Databricks Repos operators #70503

haseebmalik18:
Moves the `branch`/`tag`/`repo_id`/`repo_path` provision checks in `DatabricksReposCreate/Update/DeleteOperator` back to `__init__` (reverting #70341) so bad argument combos fail when the Dag is parsed instead of at execute time. The `repo_path` regex check stays in `execute()` since it reads the rendered value.

related: #70503  
> Depends on #70505, will rebase once that merges.


shahar1 (approved):
Great!

Issue #53052, opened 2025-07-08
[Filters] Add filters to Search - Assets


cmarteepants:
### View  
Search – Assets

<img width="1450" height="815" alt="Image" src="https://github.com/user-attachments/assets/0468e77b-53c8-4be5-9614-d4436f0cb0e0" />

### Filters to Add

- Filter by **Asset Group**  
  ↪ Narrow results to a logical grouping of assets (e.g., `revenue_reporting`, `customer_metrics`)

- Filter by **Consuming DAG** and / or **Producing Task**  
  ↪ Quickly locate assets by their upstream producer or downstream consumer

- Filter by **Last Event Timestamp (Range)**  
  ↪ Identify stale or recently updated assets

- Filter by **Is Alias (Yes/No)**  
  ↪ Distinguish between assets vs. placeholders

- Filter by **Has Events (Yes/No)**  
  ↪ Distinguish between assets with active data lineage vs inactive definitions

### API Support  

⚠️ Partial

- The `assets` API supports filtering by:
  - `dag_id`, `task_id`, and sometimes `group`
- Currently lacks:
  - Filtering by event presence (`has_events`)
  - Range filtering by last event timestamp

### UI Impact  

- Add a filter panel above the asset list table
- Useful in large deployments with hundreds or thousands of assets

### Notes  

- Enables users to answer critical questions like:
  - “What assets are stale?”
  - “Which DAGs produce assets in this domain?”
- Especially helpful when debugging broken pipelines or validating asset freshness

jroachgolf84:
Something similar here: https://github.com/apache/airflow/issues/47164.

jroachgolf84:
> Filter by Asset Group
> ↪ Narrow results to a logical grouping of assets (e.g., revenue_reporting, customer_metrics)

@cmarteepants, is there currently a way to assign an Asset to a "group"?

cmarteepants:
@jroachgolf84 there is an arg for it, but it's not currently being reflected in the UI. Probably 3.2 when we revamp it.

nikhil25803:
Hey @cmarteepants and @jroachgolf84, I'd like to work on this. Can you please assign it to me? 
Additionally, issue #53053 is somewhat related. Can I also pick that up, along with this issue?

vatsrahul1001:
@nikhil25803, how are we progressing on this?

nikhil25803:
[...]

PR #70552, opened 2026-07-27
Add last_asset_event_timestamp sort option to GET /assets

yuting1008:
## Why

- Users could see last_asset_event in the list but couldn't sort by it, making it hard to find recently-active (or long-stale) assets in a large asset list.

## How

- Created a helper function `generate_last_asset_event_query()` to query assets with the latest event timestamp.
- Added `last_asset_event_timestamp` as a sortable attribute on GET /assets via SortParam's `to_replace`.
- Wired the Assets list UI's "Last Asset Event" column to the new field and enabled sorting on it.

https://github.com/user-attachments/assets/c3a5faca-183b-4ae8-ab07-403b6a54f532

related: https://github.com/apache/airflow/pull/50279, https://github.com/apache/airflow/issues/53052

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/common/db/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/common/db/assets.py
@@ -38,7 +38,7 @@ def generate_last_asset_event_query() -> Subquery:
     return (
         select(
             AssetEvent.asset_id,  
-            func.max(AssetEvent.id).label("last_asset_event_id"),  # The ID of the last AssetEvent
+            func.max(AssetEvent.id).label("last_asset_event_id"),  
             func.max(AssetEvent.timestamp).label("last_asset_event_timestamp"),
         )
         .join(
--- a/airflow-core/src/airflow/api_fastapi/common/db/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/common/db/assets.py
@@ -37,7 +37,7 @@ def generate_last_asset_event_query() -> Subquery:
 
     return (
         select(
-            AssetEvent.asset_id,  # The ID of the Asset, which we'll need to JOIN to the AssetModel
+            AssetEvent.asset_id,  
             func.max(AssetEvent.id).label("last_asset_event_id"),  # The ID of the last AssetEvent
             func.max(AssetEvent.timestamp).label("last_asset_event_timestamp"),
         )
--- a/airflow-core/src/airflow/api_fastapi/common/db/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/common/db/assets.py
@@ -1,0 +1,53 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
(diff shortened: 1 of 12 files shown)


airflow-core/src/airflow/api_fastapi/common/db/assets.py
```diff
@@ -0,0 +1,53 @@
...
+    """Build a subquery yielding the ID and timestamp of the latest AssetEvent per asset."""
+    last_asset_event_per_asset = (
+        select(AssetEvent.asset_id, func.max(AssetEvent.timestamp).label("last_timestamp"))
+        .group_by(AssetEvent.asset_id)
+        .subquery()
+    )
+
+    return (
+        select(
+            AssetEvent.asset_id,  # The ID of the Asset, which we'll need to JOIN to the AssetModel
```
henry3260:
```suggestion
            AssetEvent.asset_id,  
```

airflow-core/src/airflow/api_fastapi/common/db/assets.py
```diff
@@ -0,0 +1,53 @@
...
+            func.max(AssetEvent.id).label("last_asset_event_id"),  # The ID of the last AssetEvent
```
henry3260:
```suggestion
            func.max(AssetEvent.id).label("last_asset_event_id"),

File: airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py [hunk 3]
```diff
@@ -151,42 +155,24 @@ def get_assets(
...
-        .group_by(AssetEvent.asset_id)
-        .subquery()
-    )
-
     assets_select_statement = select(
         AssetModel,
-        asset_event_query.c.last_asset_event_id,  # This should be the AssetEvent.id
-        asset_event_query.c.last_asset_event_timestamp,
-    ).outerjoin(asset_event_query, AssetModel.id == asset_event_query.c.asset_id)
+        _last_asset_event_query.c.last_asset_event_id,  # This should be the AssetEvent.id
```
henry3260:
```suggestion
        _last_asset_event_query.c.last_asset_event_id,  
```

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
```diff
@@ -497,6 +497,20 @@ def test_order_by_raises_400_for_invalid_attr(self, test_client, session):
         msg = "Ordering with 'fake' is disallowed or the attribute does not exist on the model"
         assert response.json()["detail"] == msg
 
+    def test_order_by_last_asset_event_timestamp(self, test_client, session):
```
henry3260, line 519: The sort directions are asserted, but I don't think this test can distinguish the new sort key from the primary-key fallback.

`_create_assets_events(varying_timestamps=True)` assigns
`timestamp=DEFAULT_DATE + timedelta(days=i - 1)` to `asset_id=i`, so timestamp order is identical to id order. `SortParam._resolve()` always appends the primary key as a tiebreaker (in the same direction as the first key), so `order_by=id` / `-id` produce exactly the same `[1, 2, 3]` / `[3, 2, 1]`. If the `to_replace` mapping pointed at the wrong column — or were dropped entirely — these assertions would still pass.

Could we make the timestamps non-monotonic asset id, so the expected
order differs from both id order and its reverse? e.g. offsets `[2, 0, 1]`, giving `[2, 3, 1]` ascending and `[1, 3, 2]` descending.

henry3260 (commented):
Thanks for your first contribution!

yuting1008:
@henry3260 Thanks for reviewing! I added a `timestamp_day_offsets` to set each event's offset explicitly.

henry3260 (dismissed):
LGTM from my side

bbovenzi:
I have a PR open for the same thing but to move to a custom get ui/assets endpoint https://github.com/apache/airflow/pull/69940 Given that we want to do more to customized the assets list from what the public API can do I think we should mark this a a duplicate

henry3260:
> I have a PR open for the same thing but to move to a custom get ui/assets endpoint #69940 Given that we want to do more to customized the assets list from what the public API can do I think we should mark this a a duplicate

Ah, sorry — I didn’t notice #69940 was already open for this.

PR #70559, opened 2026-07-27
Speed up constraints-version-check by resolving the baseline once

potiuk:
The `Finalize tests / Deps` matrix is a serial tail on every canary run — it starts only once the rest of the run has finished, then adds another 40–70 minutes of wall clock. About half of that was the command recomputing an answer it already had.

`explain_package_upgrade` needs two resolutions per outdated package: the unpinned workspace, and the workspace with that package pinned to its latest version. Only the second depends on the package. The first is identical for every package in a run — each explanation restores `pyproject.toml` and `uv.lock` before the next one starts, and the remaining inputs (`python_version`, `airflow_constraints_mode`, `github_repository`) are fixed for the whole command. It was re-run per package anyway.

Measured on the `Deps 3.11:constraints` job of run [29839957350](https://github.com/apache/airflow/actions/runs/29839957350): 692 packages checked, 63 outdated, 23.8 min of the 32.4 min step spent in the explain phase. That is 126 full `uv sync --resolution highest --refresh` invocations where 64 would do.

The baseline is now resolved once per command run — lazily, so a run with nothing outdated (or without `--explain-why`) does no resolution at all — and shared by every explanation. `sync_and_freeze`, `get_additional_sync_args` and `preserve_files` move to module level so both paths share them; a dead nested `preserve_pyproject_file` in `process_packages` is dropped.

Explanations and their classification branches are unchanged. Resolving once also removes a latent non-determinism: every baseline sync passed `--refresh`, so an index change part-way through a run could leave two packages compared against different baselines.

The module had no tests; this adds coverage for the new contract — baseline resolved exactly once across N outdated packages, not resolved when nothing is outdated or `--explain-why` is off, and `explain_package_upgrade` performing only the pinned resolution.

Not addressed here, noted for anyone looking at the same job:

- The PyPI metadata loop is 692 sequential `urllib.request.urlopen` calls with no connection reuse. Usually 0.4–1.9 min, but 8.7 min in the run above.
- `dependency-upgrade-summary` produces a diagnostic report and gates nothing — no job `needs:` it — so it could run as its own scheduled workflow instead of extending every canary run's critical path.

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/utils/constraints_version_check.py
+++ b/dev/breeze/src/airflow_breeze/utils/constraints_version_check.py
@@ -446,14 +446,6 @@ def process_packages(
     github_repository: str | None,
     cooldown_days: int = 4,
 ) -> tuple[int, int, list[str], dict[str, int]]:
-    @contextmanager
-    def preserve_pyproject_file(pyproject_path: Path):
-        original_content = pyproject_path.read_text()
-        try:
-            yield
-        finally:
-            pyproject_path.write_text(original_content)
-
     def fetch_pypi_data(pkg: str) -> dict:
         pypi_url = f"https://pypi.org/pypi/{pkg}/json"
         with urllib.request.urlopen(pypi_url) as resp:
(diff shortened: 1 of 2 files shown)


shahar1 (approved):
NICE!

PR #70561, opened 2026-07-27
Restore the lang-SDK Go bundle build

potiuk:
`main` is currently red on `Kubernetes tests / K8S Lang-SDK` for every PR, at the "Build Go bundle" step:

```
go: updates to go.mod needed; to update it:
	go mod tidy
lang-SDK provisioning failed in: Build Go bundle
subprocess.CalledProcessError: Command '['go', 'tool', 'airflow-go-pack', ...]' returned non-zero exit status 1.
```

`kubernetes-tests/lang_sdk/go_example/` is a separate module that resolves the Go SDK from the in-repo sources:

```
replace github.com/apache/airflow/go-sdk => ../../../go-sdk
```

so it carries its own copy of the SDK's indirect requirements. #70226 bumped `google.golang.org/grpc` 1.79.3 → 1.82.1 inside `/go-sdk` only, leaving the example module pinned to the old versions. Go refuses to build an inconsistent module, so the job fails on every pull request rather than only on the one that did the bump.

This is `go mod tidy` in that module — the three indirect requirements realign with the SDK, plus `go.sum`:

```diff
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
-	google.golang.org/grpc v1.79.3 // indirect
-	google.golang.org/protobuf v1.36.10 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260414002931-afd174a4e478 // indirect
+	google.golang.org/grpc v1.82.1 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
```

Verified by running exactly what CI runs, in the example module:

```
$ go tool airflow-go-pack --output /tmp/lang_sdk_check .
Wrote bundle /tmp/lang_sdk_check (sdk=go/(devel), dags=1)
```

Worth a follow-up: nothing tidies the dependent module when Dependabot bumps a dependency in `/go-sdk`, so the next Go bump breaks this again the same way. A `go mod tidy -diff` check over the example module would catch it on the bump PR itself.

---


shahar1 (approved):
Good catch!

jason810496 (commented):
Thank Jarek!
> main is currently red on Kubernetes tests / K8S Lang-SDK for every PR, at the "Build Go bundle" step:

And sorry that I didn't notice the main is failing.

PR #70567, opened 2026-07-27
UI: Wire advanced-search toggle for Grid Triggering User filter

dheerajturaga:
The Grid endpoint accepts both the substring ``triggering_user`` and the index-friendly ``triggering_user_prefix`` variants of the Triggering User filter, and the pill on the Grid's filter bar already renders the substring/prefix advanced-search toggle. But the Grid query hooks were only ever sending the substring param regardless of toggle state — so the toggle effectively did nothing on the Grid.

Route the value through ``useAdvancedSearchArg`` in ``useGridRuns`` and ``useGridStructure`` so the toggle actually selects between the two variants on each request, matching the Run ID filter wiring landed in #70150 and the Dag Runs page.

image

image

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/queries/useGridRuns.ts
+++ b/airflow-core/src/airflow/ui/src/queries/useGridRuns.ts
@@ -56,6 +56,15 @@ export const useGridRuns = ({
     value: runIdPattern,
   });
 
+  // Advanced-search toggle picks between the substring ``triggeringUser`` and the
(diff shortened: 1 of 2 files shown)


bbovenzi (approved):

vatsrahul1001:
Fixes issue introduced by https://github.com/apache/airflow/pull/70150, moved 70150 to 3.4.0 so moving this one as well

PR #70568, opened 2026-07-27
Catch lang-SDK Go example module drift before it reaches main

potiuk:
Follow-up to #70561, which cleaned up the drift by hand. This stops the next Go bump from doing it again.

`kubernetes-tests/lang_sdk/go_example` is a separate Go module that resolves the SDK through a `replace` onto `../../../go-sdk`, so it carries its own copy of the SDK's indirect requirements. Nothing re-tidies it when a dependency moves inside `/go-sdk`, and Dependabot bumps exactly one module per PR — #70226 moved `google.golang.org/grpc` 1.79.3 → 1.82.1 in `/go-sdk` only.

What makes it worth guarding is the blast radius: Go refuses to build an inconsistent module, so once such a bump merges, `Kubernetes tests / K8S Lang-SDK` fails at the "Build Go bundle" step on **every** pull request until someone notices — not just on the PR that caused it.

Config can't prevent this. Dependabot *security* updates ignore `.github/dependabot.yml` entirely (which today has no `gomod` entry at all), and a second Dependabot PR for the example module would merge at a different time and leave `main` red in between. The drift has to fail the bump PR itself.

The check is `go mod tidy -diff` in the example module — exactly the question the failing CI step asks, it never writes to the working tree, and it exits non-zero when the module is untidy. It runs only when either module's `go.mod`/`go.sum` changes, which is sufficient: any new requirement in the SDK necessarily lands in `go-sdk/go.mod` first.

Verified against the real regression. On `main` as it stands:

```
$ prek run check-go-example-mod-tidy --files go-sdk/go.mod
ERROR: kubernetes-tests/lang_sdk/go_example is not tidy.
...
Fix it in this PR by running:

    (cd kubernetes-tests/lang_sdk/go_example && go mod tidy)
```

and the reported diff is byte-for-byte the change #70561 makes. After applying that tidy, the hook exits 0.

### This PR's own static checks will be red until #70561 merges

That is the hook working, not a defect — `main` is currently untidy, so the check it adds correctly fails. The commit was made with `SKIP=check-go-example-mod-tidy` for the same reason. Once #70561 lands, this goes green with no change here; happy to rebase onto it instead if you'd rather have one green PR.

Not addressed: the existing `go-mod-tidy` hook in `go-sdk/.pre-commit-config.yaml` is `types: [go]`, so it does not fire on `go.mod`/`go.sum` changes either. That has not bitten us because Dependabot tidies the module it bumps, but it is the same gap one level up.

---


shahar1 (approved):

jason810496 (commented):
Yes, that is exactly what I thought about to prevent further drift, thanks.

PR #70569, opened 2026-07-28
Add `DmsReloadTablesOperator`

AlejandroMorgante:
Add `DmsReloadTablesOperator` to reload selected target tables for an active AWS DMS replication task.

The operator wraps the AWS DMS `ReloadTables` API and supports:

- `data-reload` and `validate-only` reload options.
- Synchronous and deferrable completion waits.
- Configurable waiter delay and maximum attempts.
- Templated replication task, table, and reload-option arguments.

The DMS system test now creates a `full-load-and-cdc` task with PostgreSQL logical replication enabled, waits for the initial load, and exercises a deferrable table reload against AWS.

Testing:

  - System test validated end-to-end against a real AWS environment.
  - DMS unit tests: 131 passed.

References:

  - [Boto3 `reload_tables` API](https://docs.aws.amazon.com/boto3/latest/reference/services/dms/client/reload_tables.html)
  - [AWS DMS table reload requirements and limitations](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.ReloadTables.html)

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/hooks/dms.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/hooks/dms.py
@@ -212,6 +212,32 @@ class DmsHook(AwsBaseHook):
         dms_client = self.get_conn()
         dms_client.stop_replication_task(ReplicationTaskArn=replication_task_arn)
 
+    def reload_tables(
+        self,
+        replication_task_arn: str,
+        tables_to_reload: list[dict[str, str]],
+        reload_option: str = "data-reload",
+    ) -> str:
+        """
+        Reload target tables for a running replication task.
+
+        .. seealso::
+            - :external+boto3:py:meth:`DatabaseMigrationService.Client.reload_tables`
+
+        :param replication_task_arn: Replication task ARN
+        :param tables_to_reload: Tables to reload, including schema and table names
+        :param reload_option: Use ``data-reload`` to reload data and revalidate it when validation is
+            enabled. Use ``validate-only`` to revalidate without reloading data; this option only applies
+            when validation is enabled.
+        :return: Replication task ARN
+        """
+        response = self.get_conn().reload_tables(
+            ReplicationTaskArn=replication_task_arn,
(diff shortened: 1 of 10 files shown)


providers/amazon/src/airflow/providers/amazon/aws/hooks/dms.py
```diff
@@ -212,6 +212,32 @@ def stop_replication_task(self, replication_task_arn):
         dms_client = self.get_conn()
         dms_client.stop_replication_task(ReplicationTaskArn=replication_task_arn)
 
+    def reload_tables(
```
vincbeck: We try to avoid methods in hooks that adds no real value. This is the case for this one. You can boto3 API directly in operator instead

providers/amazon/src/airflow/providers/amazon/aws/operators/dms.py
```diff
@@ -419,6 +420,134 @@ def execute(self, context: Context):
...
+                WaiterConfig={
+                    "Delay": self.waiter_delay,
+                    "MaxAttempts": self.waiter_max_attempts,
+                },
+            )
+
+    def execute(self, context: Context) -> str:
+        """Start reloading target tables for an AWS DMS replication task."""
+        if self.wait_for_completion and self.reload_option != "data-reload":
+            raise ValueError("wait_for_completion is only supported when reload_option is 'data-reload'.")
```
vincbeck: Instead of failing whouls we just ignore `wait_for_completion` when `reload_option` is not `data-reload`?

AlejandroMorgante: **Good catch.** I overlooked that validate-only has its own asynchronous completion state in `ValidationState`. I’ll add a **dedicated** **waiter** path selected by `reload_option`, so `wait_for_completion` works consistently for both strategies

vincbeck (approved):

o-nikolas (approved):

PR #70574, opened 2026-07-28
Check GCSToAzureBlobStorageOperator match_glob support after template…

bramhanandlingala:
Fixes the microsoft/azure provider's GCSToAzureBlobStorageOperator entry from the #70296 exemption-list burn-down.

`match_glob` is a template field, but `__init__` validated it against the installed `apache-airflow-providers-google` version and raised immediately if unsupported — so this check ran on the un-rendered Jinja expression instead of the actual rendered value.

Moved the compatibility check from `__init__` into `execute()`, right before `match_glob` is used to build the GCS list call. `__init__` now only does a plain assignment. `_is_match_glob_supported` itself stays computed in `__init__`, since it depends on the installed provider version, not on any template field.

- Removed the raise from `__init__`; `self.match_glob = match_glob` is now a plain assignment.
- Added the same check in `execute()`, guarded by `_is_match_glob_supported`.
- Updated the existing test to construct the operator successfully and assert the error on `.execute()` instead of construction.

Verified locally that `scripts/ci/prek/validate_operators_init.py` reports zero findings for this class after the change, and that the existing unit tests in `test_gcs_to_wasb.py` pass with the updated test.

Related to #70296

---

- [X] Yes - Claude


providers/microsoft/azure/tests/unit/microsoft/azure/transfers/test_gcs_to_wasb.py
```diff
@@ -49,14 +49,17 @@ def test_init_defaults(self):
...
+            match_glob="**/*.csv",
+        )
         with pytest.raises(ValueError, match="match_glob"):
-            GCSToAzureBlobStorageOperator(
-                task_id=TASK_ID,
-                gcs_bucket=GCS_BUCKET,
-                container_name=CONTAINER,
-                match_glob="**/*.csv",
-            )
+            op.execute(context=None)
```
potiuk: `context=None` works only because the raise happens before anything touches the context. It's a bit fragile as documentation of intent — if the guard later moves below something that reads `context`, this fails with an unrelated `AttributeError` rather than the assertion you meant.

`mock.MagicMock()` (as #70529 uses for the same situation) or `{}` would be more robust. Also worth adding `mock_gcs_hook.return_value.list.assert_not_called()` — #70723's equivalent test does that, and it's what actually proves the operator bailed out before doing any work rather than just raising somewhere.

---

potiuk (approved):
Thanks — same fix as #70723 does for the Amazon `GCSToS3Operator`, and correct for the same reason: `match_glob` is in `template_fields`, so checking it in `__init__` validated the un-rendered Jinja string and turned a bad value into a Dag *parse* failure rather than a task failure. Keeping the `_is_match_glob_supported` probe in `__init__` is right, since that depends on the installed Google provider rather than anything templated.

Using `ValueError` here is the better choice — I've suggested on #70723 that it match yours rather than the other way round, since Airflow is trying to reduce direct `AirflowException` raises. Worth the two of you syncing so the sibling changes land consistent.

One test nit inline, not blocking.

---

potiuk (changes requested):

`if not self._is_match_glob_supported and match_glob:` asks whether `match_glob` was passed (combined with an environment capability that `__init__` can already answer). It's a provision check, so it should be rewritten in place rather than relocated:

```python
# in __init__ — keep it here
if not self._is_match_glob_supported and match_glob is not None:
    raise ValueError(
        "The 'match_glob' parameter requires 'apache-airflow-providers-google>=10.3.0'."
    )
```

Note the `is not None` polarity matters: `if match_glob:` is a truthiness test on the un-rendered Jinja string, which is a third question that matches neither intent. You already use `ValueError` here, which is right.

This is the twin of #70723 — worth the two of you landing the same shape.

---

bramhanandlingala:
@potiuk Thanks for catching this — restored the check to __init__ using
match_glob is not None, removed the duplicate from execute(), and
updated the test to assert the raise on construction instead of execute().

potiuk (approved):

PR #70576, opened 2026-07-28
Fix DateTimeSensorAsync crash on templated target_time with start_from_trigger (real fix, closes #70284)

### LINKED ISSUE #70284: DateTimeSensorAsync crashes Dag parsing with templated target_time + start_from_trigger=True

nanjeshramesh:
`DateTimeSensorAsync.__init__` calls `timezone.parse(self.target_time)` at Dag-parse time
when `start_from_trigger=True`, before Jinja rendering has happened. `target_time` is a
documented template field (e.g. `"{{ data_interval_end.tomorrow().replace(hour=1) }}"`), so
passing a template — the field's normal, documented use case — raises a raw
`pendulum.parsing.exceptions.ParserError` right there, crashing parsing of the **entire Dag
file**, not just the one task.

## Relation to #70310 and #70469

Both prior attempts at this issue (#70310 by @haseebmalik18, #70469 by @burakcoleman) catch
the parse failure in `__init__` and raise a clearer `ValueError` instead of the raw
`ParserError`. That's a real improvement to the error message, but as reviewers on both PRs
pointed out, it doesn't fix the crash — the Dag still fails to parse whenever
`start_from_trigger=True` and `target_time` is a template.

This PR takes the approach @haseebmalik18 outlined in the #70310 review discussion: defer
resolving `target_time` until the triggerer can render it, the same way `FileSensor` already
handles a templated `filepath`.

## What this PR does

On Airflow >= 3.3, instead of parsing `target_time` at Dag-parse time, the raw (unrendered)
template string is passed straight through to `DateTimeTrigger` via
`start_trigger_args.trigger_kwargs` under the key `"target_time"` — which matches
`DateTimeSensor.template_fields`. `BaseTrigger`'s `task_instance` setter already picks up any
`trigger_kwargs` key that matches an operator template field (see
`airflow.triggers.base.BaseTrigger`), and `TriggererJobRunner.run_trigger` calls
`render_template_fields()` on the trigger before invoking `run()` — this is exactly the
mechanism introduced in #55068 and that `FileSensor` already relies on for `filepath`.

`DateTimeTrigger` now accepts either a resolved `moment` (unchanged, existing behavior) or a
raw `target_time` string. In the latter case, `moment` starts as `None` and is lazily parsed
from `target_time` the first time it's needed (`run()` or `serialize()`), by which point the
triggerer has already rendered the template in place.

On Airflow < 3.3, that triggerer-side rendering mechanism doesn't exist yet, so a templated
`target_time` can never be resolved via `start_from_trigger`. Rather than crash Dag parsing on
every parse cycle, `start_from_trigger` is disabled with a `log.warning`, and the task falls
back to deferring from the worker via `execute()` instead — at that point `target_time` has
already been rendered normally by the scheduler/worker, so it works correctly, just without
the trigger-only optimization.

Static/ISO `target_time` values are completely unaffected on any version.

## Testing

- Added unit tests in `test_temporal.py` covering: `DateTimeTrigger` accepting an unrendered
  `target_time` at construction without raising, resolving it once rendered, raising a clear
  error if it's still unrendered when needed, rejecting both/neither of `moment`/`target_time`,
  and a full simulated triggerer flow (construct with raw template → `render_template_fields`
  → `run()`) that fires at the correct time.
- Added unit tests in `test_date_time.py` covering `DateTimeSensorAsync.__init__` branching on
  `AIRFLOW_V_3_3_PLUS` for a templated `target_time`, and confirming a static `target_time`
  behaves identically regardless of that flag.
- Ran the exact reproduction script from #70284: confirmed it raises
  `pendulum.parsing.exceptions.ParserError` on `main`, and parses cleanly with this fix.
- Full existing test suites for both files pass (32/32), no regressions.
- `ruff check` / `ruff format --diff` clean on all changed files.

closes: #70284
Related: #70310, #70469, #55068, #69610

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -24,7 +24,7 @@ from typing import TYPE_CHECKING, Any, NoReturn
 
 from airflow.providers.common.compat.sdk import BaseSensorOperator, timezone
 from airflow.providers.standard.triggers.temporal import DateTimeTrigger
-from airflow.providers.standard.version_compat import AIRFLOW_V_3_0_PLUS
+from airflow.providers.standard.version_compat import AIRFLOW_V_3_0_PLUS, AIRFLOW_V_3_3_PLUS
 from airflow.triggers.base import StartTriggerArgs
 
 if TYPE_CHECKING:
@@ -89,6 +89,10 @@ class DateTimeSensorAsync(DateTimeSensor):
 
     :param target_time: datetime after which the job succeeds. (templated)
     :param start_from_trigger: Start the task directly from the triggerer without going into the worker.
+        This requires either a static ``target_time`` (a datetime or ISO-8601 string) or, on
+        Airflow >= 3.3, a templated ``target_time`` that the triggerer can render before the trigger
+        runs. On earlier Airflow versions a templated ``target_time`` cannot be resolved this way, so
+        ``start_from_trigger`` is disabled with a warning and the task defers from the worker instead.
     :param trigger_kwargs: The keyword arguments passed to the trigger when start_from_trigger is set to True
         during dynamic task mapping. This argument is not used in standard usage.
     :param end_from_trigger: End the task directly from the triggerer without going into the worker.
@@ -116,16 +120,53 @@ class DateTimeSensorAsync(DateTimeSensor):
 
         self.start_from_trigger = start_from_trigger
         if self.start_from_trigger:
+            try:
+                moment = self._moment
+            except ValueError:
+                # target_time couldn't be parsed as a static datetime at Dag-parse time. This is
+                # the normal, documented case of target_time being a Jinja template, e.g.
+                # "{{ data_interval_end.tomorrow().replace(hour=1) }}" -- not necessarily bad input.
+                moment = None
+
             # Replaced rather than mutated: ``start_trigger_args`` is a class attribute, so
             # assigning through it would overwrite the arguments of every other task built
             # from this operator.
-            self.start_trigger_args = dataclasses.replace(
-                self.start_trigger_args,
-                trigger_kwargs=dict(
-                    moment=self._moment,
-                    end_from_trigger=self.end_from_trigger,
-                ),
-            )
+            if moment is not None:
+                self.start_trigger_args = dataclasses.replace(
+                    self.start_trigger_args,
+                    trigger_kwargs=dict(
+                        moment=moment,
+                        end_from_trigger=self.end_from_trigger,
+                    ),
+                )
+            elif AIRFLOW_V_3_3_PLUS:
+                # Hand the raw template string to the trigger under the same name as this
+                # operator's template field ("target_time"). The triggerer renders any
+                # start_trigger_args kwarg whose name matches an operator template field before
+                # running the trigger (see BaseTrigger.task_instance / render_template_fields),
+                # the same mechanism FileSensor relies on for a templated `filepath`.
+                self.start_trigger_args = dataclasses.replace(
+                    self.start_trigger_args,
+                    trigger_kwargs=dict(
+                        target_time=self.target_time,
+                        end_from_trigger=self.end_from_trigger,
+                    ),
+                )
+            else:
+                # Airflow < 3.3 triggerers can't render template fields on a trigger before it
+                # runs, so a templated target_time can never be resolved via start_from_trigger.
+                # Falling back to the normal worker-deferred path (execute()) avoids crashing Dag
+                # parsing on every parse cycle; execute() runs after the scheduler/worker has
+                # already rendered target_time normally.
+                self.log.warning(
+                    "start_from_trigger=True requires a static target_time on Airflow < 3.3, but "
+                    "%r looks like a template for task %r. Disabling start_from_trigger and "
+                    "deferring from the worker instead. Upgrade to Airflow >= 3.3 to defer "
+                    "directly from the triggerer with a templated target_time.",
+                    self.target_time,
+                    self.task_id,
+                )
+                self.start_from_trigger = False
 
     def execute(self, context: Context) -> NoReturn:
         self.defer(
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -122,15 +122,11 @@ class DateTimeSensorAsync(DateTimeSensor):
                 moment = self._moment
             except ValueError:
                 if not self._looks_like_template(self.target_time):
-                    # Not a template, and not parseable either -- this is genuinely invalid
-                    # input (e.g. "not-a-date"), so fail fast instead of silently deferring to
-                    # a trigger that could never resolve it either.
+                    # genuinely invalid input (e.g. "not-a-date"), not a template: fail fast
                     raise
                 moment = None
 
-            # Replaced rather than mutated: ``start_trigger_args`` is a class attribute, so
-            # assigning through it would overwrite the arguments of every other task built
-            # from this operator.
+            # start_trigger_args is a class attribute, so replace it rather than mutate it
             if moment is not None:
                 self.start_trigger_args = dataclasses.replace(
                     self.start_trigger_args,
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -88,11 +88,9 @@ class DateTimeSensorAsync(DateTimeSensor):
     It is a drop-in replacement for DateTimeSensor.
 
     :param target_time: datetime after which the job succeeds. (templated)
-    :param start_from_trigger: Start the task directly from the triggerer without going into the worker.
-        This requires either a static ``target_time`` (a datetime or ISO-8601 string) or, on
-        Airflow >= 3.3, a templated ``target_time`` that the triggerer can render before the trigger
-        runs. On earlier Airflow versions a templated ``target_time`` cannot be resolved this way, so
-        ``start_from_trigger`` is disabled with a warning and the task defers from the worker instead.
+    :param start_from_trigger: Start the task directly from the triggerer instead of a worker.
+        Supports static ``target_time`` values and, on Airflow >= 3.3, templated
+        ``target_time`` values.
     :param trigger_kwargs: The keyword arguments passed to the trigger when start_from_trigger is set to True
         during dynamic task mapping. This argument is not used in standard usage.
     :param end_from_trigger: End the task directly from the triggerer without going into the worker.
@@ -123,9 +121,11 @@ class DateTimeSensorAsync(DateTimeSensor):
             try:
                 moment = self._moment
             except ValueError:
-                # target_time couldn't be parsed as a static datetime at Dag-parse time. This is
-                # the normal, documented case of target_time being a Jinja template, e.g.
-                # "{{ data_interval_end.tomorrow().replace(hour=1) }}" -- not necessarily bad input.
+                if not self._looks_like_template(self.target_time):
+                    # Not a template, and not parseable either -- this is genuinely invalid
+                    # input (e.g. "not-a-date"), so fail fast instead of silently deferring to
+                    # a trigger that could never resolve it either.
+                    raise
                 moment = None
 
             # Replaced rather than mutated: ``start_trigger_args`` is a class attribute, so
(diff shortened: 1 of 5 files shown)


providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -115,10 +119,44 @@ def __init__(
...
-                moment=self._moment,
-                end_from_trigger=self.end_from_trigger,
-            )
+            try:
+                moment = self._moment
+            except ValueError:
+                # target_time couldn't be parsed as a static datetime at Dag-parse time. This is
+                # the normal, documented case of target_time being a Jinja template, e.g.
+                # "{{ data_interval_end.tomorrow().replace(hour=1) }}" -- not necessarily bad input.
+                moment = None
```
SameerMesiah97, line 128: This catches all `ValueErrors `and treats them as if the value were simply "not yet renderable". Is that intentional? For example, an invalid static value like "not-a-date" would now follow the same path as an unresolved Jinja template. Should those cases be distinguished so genuinely invalid input still fails eagerly?

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -88,6 +88,10 @@ class DateTimeSensorAsync(DateTimeSensor):
 
     :param target_time: datetime after which the job succeeds. (templated)
     :param start_from_trigger: Start the task directly from the triggerer without going into the worker.
+        This requires either a static ``target_time`` (a datetime or ISO-8601 string) or, on
+        Airflow >= 3.3, a templated ``target_time`` that the triggerer can render before the trigger
+        runs. On earlier Airflow versions a templated ``target_time`` cannot be resolved this way, so
+        ``start_from_trigger`` is disabled with a warning and the task defers from the worker instead.
```
SameerMesiah97: This can be more concise. Please see the below:

```
:param start_from_trigger: Start the task directly from the triggerer instead of a worker.
    Supports static ``target_time`` values and, on Airflow >= 3.3, templated
    ``target_time`` values.
```

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -115,10 +119,44 @@ def __init__(
...
+                self.start_trigger_args.trigger_kwargs = dict(
+                    moment=moment,
+                    end_from_trigger=self.end_from_trigger,
+                )
+            elif AIRFLOW_V_3_3_PLUS:
+                # Hand the raw template string to the trigger under the same name as this
+                # operator's template field ("target_time"). The triggerer renders any
+                # start_trigger_args kwarg whose name matches an operator template field before
+                # running the trigger (see BaseTrigger.task_instance / render_template_fields),
+                # the same mechanism FileSensor relies on for a templated `filepath`.
```
SameerMesiah97: The comment is too long. Please see the below;

```
# Pass the unresolved template to the trigger. On Airflow >= 3.3 the
# triggerer renders trigger kwargs that correspond to template fields before
# starting the trigger.
```

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -115,10 +119,44 @@ def __init__(
...
+                self.start_trigger_args.trigger_kwargs = dict(
+                    target_time=self.target_time,
+                    end_from_trigger=self.end_from_trigger,
+                )
+            else:
+                # Airflow < 3.3 triggerers can't render template fields on a trigger before it
+                # runs, so a templated target_time can never be resolved via start_from_trigger.
+                # Falling back to the normal worker-deferred path (execute()) avoids crashing Dag
+                # parsing on every parse cycle; execute() runs after the scheduler/worker has
+                # already rendered target_time normally.
```
SameerMesiah97: Is this comment needed? Looks like the warning message communicates the necessary information. I think this comment can be removed or trimmed.

providers/standard/src/airflow/providers/standard/triggers/temporal.py
```diff
@@ -37,24 +37,77 @@ class DateTimeTrigger(BaseTrigger):
...
     :param moment: when to yield event
+    :param target_time: a Jinja-templated ``target_time`` string that has not yet been rendered.
+        Used instead of ``moment`` only when this trigger is created via ``start_from_trigger`` with
+        a ``target_time`` that could not be resolved to a concrete datetime at Dag-parse time (i.e.
+        it is a template).
+        Mutually exclusive with ``moment``. ``target_time`` is one of ``DateTimeSensor``'s
+        ``template_fields``, so the triggerer renders it in place (see
+        ``BaseTrigger.render_template_fields``) before ``run()`` is invoked; it is then parsed into
+        ``moment`` on first use. This mirrors how ``FileSensor`` defers rendering of a templated
+        ``filepath`` to the triggerer.
```
SameerMesiah97: This is too long as well. Please see the below:

```
:param target_time: Templated ``target_time`` value to resolve when the trigger
    runs. Used instead of ``moment`` when the target time cannot be determined
    during operator initialization. Mutually exclusive with ``moment``.
```

providers/standard/src/airflow/providers/standard/triggers/temporal.py
```diff
@@ -37,24 +37,77 @@ class DateTimeTrigger(BaseTrigger):
...
-        self.moment: pendulum.DateTime = timezone.convert_to_utc(moment)
-        self.end_from_trigger = end_from_trigger
+        self.moment = timezone.convert_to_utc(moment)
+
+    def _resolve_moment(self) -> pendulum.DateTime:
+        """Return ``moment``, parsing it from a (by now rendered) ``target_time`` if needed."""
+        if self.moment is not None:
+            return self.moment
+        if not self.target_time:
+            raise TypeError("DateTimeTrigger requires either 'moment' or 'target_time' to be set")
```
SameerMesiah97, line 82: Why duplicate the validation logic in the constructor here? Will it not be guaranteed that `moment` or `target_time` is set by the time `_resolve_moment` is called?

providers/standard/src/airflow/providers/standard/triggers/temporal.py
```diff
@@ -37,24 +37,77 @@ class DateTimeTrigger(BaseTrigger):
...
+        return self.moment
 
     def serialize(self) -> tuple[str, dict[str, Any]]:
+        # `target_time` is deliberately not round-tripped here: it is folded into a concrete
+        # `moment` by `_resolve_moment()` before serialization, the same way `TimeDeltaTrigger`
+        # folds `delta` into `moment` (see the exclusion for this class in
+        # scripts/ci/prek/check_trigger_serialize_init.py). By the time serialize() is called,
+        # target_time has either already been rendered by the triggerer (see
+        # BaseTrigger.render_template_fields) or resolution raises a clear error -- there is no
+        # unrendered template left to preserve.
```
SameerMesiah97: Not sure if this comment is needed. The code is self-explanatory.

providers/standard/tests/unit/standard/sensors/test_date_time.py
```diff
@@ -157,3 +157,54 @@ def test_async_start_from_trigger_localizes_naive_datetime(self):
...
+    def test_async_start_from_trigger_static_target_time_pre_3_3_unaffected(self):
+        """A static target_time must keep working identically regardless of AIRFLOW_V_3_3_PLUS."""
+        op = DateTimeSensorAsync(
+            task_id="async_static_pre_3_3",
+            target_time="2020-01-01T00:00:00+00:00",
+            start_from_trigger=True,
+            dag=self.dag,
+        )
+        assert op.start_from_trigger is True
+        assert op.start_trigger_args.trigger_kwargs["moment"] == pendulum.parse("2020-01-01T00:00:00+00:00")
```
SameerMesiah97, line 210: Would it be worth adding a test for a non-templated bt invalid `target_time` (e.g. "not-a-date") so the expected behaviour for genuinely invalid input is explicit?

SameerMesiah97 (commented):
Looks good to me overall. But I would go through the diff again and see if any docstrings/comments could be shortened/removed. I have pointed a few of them that I felt were a bit too verbose but there are certainly others that can be more concise.

CI needs to be triggered

nanjeshramesh:
Thanks for the thorough review!

**On the `except ValueError` catch-all** — good catch, that was a real gap. A genuinely invalid static `target_time` (e.g. `"not-a-date"`) was being treated the same as an unrendered Jinja template and silently deferred instead of failing at Dag-parse time like before. Fixed by checking for actual Jinja delimiters (`{{`/`{%`) before assuming "unrendered template" — anything else re-raises immediately. Added `test_async_start_from_trigger_invalid_static_target_time_fails_fast` to cover it.

**On `_resolve_moment`'s validation check** — `__init__` does guarantee one of `moment`/`target_time` is set, so this is only a defensive guard against `target_time` being cleared after construction (rather than surfacing a confusing `AttributeError` later). Added a comment explaining that instead of removing it.

**Docstrings/comments** — trimmed all of them to roughly your suggested wording (the `start_from_trigger` param doc, the template-hand-off comment, the pre-3.3 fallback comment, the `target_time` param doc on `DateTimeTrigger`, and the `serialize()` comment).

Pushed as three commits: fail-fast fix + trimmed docs, further comment trims, and the new test.

nanjeshramesh:
Thanks for taking another look. Trimmed the two you called out: removed the `_resolve_moment` defensive-guard explanation (agreed, the code speaks for itself there) and shortened the invalid-input comment to one line. Let me know if there's anything else that still reads as excessive.

Also flagging: CI needs a maintainer to re-approve since this is a fork PR (that label seems to get cleared on every push).

nanjeshramesh:
Just checking in, it's been a few days since my last update. Happy to make any further changes if there's more feedback, and flagging again in case it got lost: the fork-PR CI approval seems to still be pending, so the checks haven't actually run since the last push.

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -116,16 +118,49 @@ def __init__(
...
-                trigger_kwargs=dict(
-                    moment=self._moment,
-                    end_from_trigger=self.end_from_trigger,
-                ),
-            )
+            try:
+                moment = self._moment
+            except ValueError:
+                if not self._looks_like_template(self.target_time):
+                    # genuinely invalid input (e.g. "not-a-date"), not a template: fail fast
```
Lee-W:
```suggestion
```

Lee-W: I don't think we need this comment

providers/standard/src/airflow/providers/standard/sensors/date_time.py
```diff
@@ -116,16 +118,49 @@ def __init__(
...
+                    "directly from the triggerer with a templated target_time.",
+                    self.target_time,
+                    self.task_id,
+                )
+                self.start_from_trigger = False
+
+    @staticmethod
+    def _looks_like_template(target_time: Any) -> bool:
+        """Whether ``target_time`` contains unrendered Jinja delimiters."""
+        return isinstance(target_time, str) and ("{{" in target_time or "{%" in target_time)
```
Lee-W: let's also check }} and %}

providers/standard/src/airflow/providers/standard/sensors/date_time.py
nanjeshramesh: Agreed, removed.

providers/standard/src/airflow/providers/standard/sensors/date_time.py
nanjeshramesh: Good catch, added. Also added a test covering a target_time with only a closing delimiter (no opening one) to prove it's still treated as template-shaped rather than genuinely invalid.

nanjeshramesh:
@Lee-W I've addressed your three comments: removed the extra comment and added `}}` and `%}` to the template check, with a test for a closing-only delimiter. Since this is a fork PR, CI needs a maintainer to approve the run before it will execute. Could you trigger it when you have a moment?

nanjeshramesh:
Closing this as superseded: #72659 merged on Sep 20 and fixes #70284 by falling back to the worker path for templated targets, which covers the crash this PR was written for. Thanks @Lee-W and @SameerMesiah97 for the reviews.

The one thing this PR did beyond that was keep `start_from_trigger` working for templated `target_time` on Airflow 3.3+ by handing the template to the trigger to render. If that is something the project wants, I am happy to open a small follow-up on top of #72659, but I do not want to leave a conflicting PR open for it, so I will only do that if someone asks.

PR #70581, opened 2026-07-28
Fix DbtCloudRunJobOperator having false failures during deferred polling

### LINKED ISSUE #70398: Deferred DbtCloudRunJobOperator (and similar triggers) fail spuriously via mask_secret() cross-event-loop RuntimeError in _get_connection()
## Apache Airflow version

3.0.6

## What happened?

Deferred `DbtCloudRunJobOperator` tasks fail almost immediately with a misleading `DbtCloudJobRunException: Job run <id> has failed.` even though the real dbt Cloud job is still running (and later completes successfully, sometimes hours later). The task fails within ~10 seconds of being deferred — far too fast to reflect a real terminal job state for a job that in fact took several more hours to finish.

I traced this down to a real bug in Airflow core (Task SDK), not in the dbt Cloud provider itself, and not specific to any one secrets backend implementation. Full trace below.

### Environment

- Deployment: Amazon MWAA
- `secrets.backend`: `airflow.providers.amazon.aws.secrets.secrets_manager.SecretsManagerBackend`
- `secrets.backend_kwargs`: `{"config_prefix":"airflow/config","connections_prefix":"airflow/connections","variables_prefix":"airflow/variables"}`
- Connection `dbt_cloud` (conn_type `dbt_cloud`) has `password` populated (used by the provider as the dbt Cloud API token) and no `extra`.
- The task is `DbtCloudRunJobOperator(..., deferrable=True)` using the connection above.

### Observed log sequence

**Worker (synchronous `execute()`, before deferring) — succeeds fine:**
```
{"logger": "airflow.hooks.base", "event": "Connection Retrieved 'dbt_cloud'", "level": "info"}
{"logger": "...DbtCloudHook", "event": "Getting the status of job run <job_run_id>.", "level": "info"}
{"logger": "...DbtCloudHook", "event": "Current status of job run <job_run_id>: QUEUED", "level": "info"}
{"event": "Pausing task as DEFERRED.", "level": "info"}
```

**Triggerer (async `DbtCloudRunJobTrigger.run()` poll cycle) — crashes:**
```
{"event": "trigger <dag_id>/<run_id>/dbt_cloud_task/-1/1 (ID <n>) starting", "level": "info"}
{"logger": "...DbtCloudHook", "event": "Getting the status of job run <job_run_id>.", "level": "info"}
{
  "logger": "task",
  "error_detail": [{
    "exc_type": "RuntimeError",
    "exc_value": "You cannot use AsyncToSync in the same thread as an async event loop - just await the async function directly.",
    "frames": [
      {"filename": ".../airflow/sdk/execution_time/context.py", "lineno": 131, "name": "_get_connection"},
      {"filename": ".../airflow/secrets/base_secrets.py", "lineno": 79, "name": "get_connection"},
      {"filename": ".../airflow/secrets/base_secrets.py", "lineno": 65, "name": "deserialize_connection"},
      {"filename": ".../airflow/models/connection.py", "lineno": 557, "name": "from_json"},
      {"filename": "<string>", "lineno": 4, "name": "__init__"},
      {"filename": ".../sqlalchemy/orm/state.py", "lineno": 481, "name": "_initialize_instance"},
      {"filename": ".../sqlalchemy/util/langhelpers.py", "lineno": 70, "name": "__exit__"},
      {"filename": ".../sqlalchemy/util/compat.py", "lineno": 211, "name": "raise_"},
      {"filename": ".../sqlalchemy/orm/state.py", "lineno": 479, "name": "_initialize_instance"},
      {"filename": ".../airflow/models/connection.py", "lineno": 177, "name": "__init__"},
      {"filename": ".../airflow/sdk/execution_time/secrets_masker.py", "lineno": 134, "name": "mask_secret"},
      {"filename": ".../airflow/jobs/triggerer_job_runner.py", "lineno": 740, "name": "send"},
      {"filename": ".../asgiref/sync.py", "lineno": 186, "name": "__call__"}
    ]
  }],
  "event": "Task failed with exception",
  "level": "error"
}
{"event": "Unable to retrieve connection from secrets backend (SecretsManagerBackend). Checking subsequent secrets backend.", "level": "error"}
{"result": "TriggerEvent<{'status': 'error', 'message': 'You cannot use AsyncToSync in the same thread as an async event loop - just await the async function directly.', 'run_id': <job_run_id>}>", "event": "Trigger fired event", "level": "info"}
{"event": "trigger completed", "level": "info"}
```

This `TriggerEvent(status='error', ...)` is then consumed by `DbtCloudRunJobOperator.execute_complete()`, which raises `DbtCloudJobRunException: Job run <job_run_id> has failed.` — a message that implies the *dbt Cloud job* failed, when in reality Airflow never actually observed its terminal state at all.

**Confirmed independently via the dbt Cloud Admin API** that the underlying run continued after Airflow marked the task failed, and later reached a real terminal state (`Success`) several hours afterward — proving the "has failed" exception was not based on any actual observed job outcome.

### Root cause (traced against `apache/airflow` tag `3.0.6`)

This is not a dbt Cloud provider bug, and not specific to `SecretsManagerBackend`. It's in `_get_connection()`, `task-sdk/src/airflow/sdk/execution_time/context.py`:

```python
def _get_connection(conn_id: str) -> Connection:
    backends = ensure_secrets_backend_loaded()
    for secrets_backend in backends:
        try:
            conn = secrets_backend.get_connection(conn_id=conn_id)
            if conn:
                if conn.password:
                    mask_secret(conn.password)   # <-- crashes here
                if conn.extra:
                    mask_secret(conn.extra)
                return conn
        except Exception:
            log.exception(
                "Unable to retrieve connection from secrets backend (%s). "
                "Checking subsequent secrets backend.",
                type(secrets_backend).__name__,
            )
    ...
    msg = SUPERVISOR_COMMS.send(GetConnection(conn_id=conn_id))  # <-- same crash, uncaught here
```

1. `secrets_backend.get_connection(conn_id)` **succeeds** — the connection is fetched and deserialized fine.
2. `mask_secret(conn.password)` is called next. `mask_secret()` (`task-sdk/src/airflow/sdk/execution_time/secrets_masker.py:132-134`) does:
   ```python
   if comms := getattr(task_runner, "SUPERVISOR_COMMS", None):
       comms.send(MaskSecret(value=secret, name=name))
   ```
3. Inside the triggerer, `SUPERVISOR_COMMS` is the triggerer's comms decoder (per `airflow/jobs/triggerer_job_runner.py`). Calling `.send()` on it from inside a coroutine that is *already* running the active event loop conflicts with the sync→async bridge (`asgiref.sync.AsyncToSync`) it uses internally, raising the `RuntimeError`.
4. This exception is caught by the `except Exception` in the loop (logged as "Checking subsequent secrets backend"), but since only one backend is configured, the loop ends. Execution falls through to the **unguarded** fallback call `SUPERVISOR_COMMS.send(GetConnection(conn_id=conn_id))` at line 160 — the exact same `.send()` primitive, called from the exact same broken async context — which raises the identical `RuntimeError`, this time uncaught, propagating up into the trigger's `run()` coroutine and surfacing as a generic `TriggerEvent(status='error')`.

Key point: **this has nothing to do with which secrets backend is configured, or which Connection field holds the secret.** Any deferred operator that (a) uses a Connection with `password` or `extra` populated, and (b) re-fetches that connection from inside the triggerer's async `run()` loop (as `DbtCloudRunJobTrigger` does on every poll, via `provide_account_id`'s `sync_to_async(self.get_connection)(...)` call), will hit this. It happens to surface here via `apache-airflow-providers-dbt-cloud`, but the defect is in Airflow core / Task SDK, not the provider.

### Related reports

- #53447 — same `RuntimeError: You cannot use AsyncToSync...` signature, reported for `BigQueryHook` in a custom trigger (Airflow 3.0.3). Closed as "custom trigger code," but a later comment on that thread (2026-02-19) points out the identical crash occurs inside the **official** `DbtCloudRunJobTrigger.run()` — this wasn't spun out into its own issue at the time.
- #64213 — same signature via `TriggerCommsDecoder.send()` cross-loop conflict, this time through the amazon provider's async AWS hook. Closed by the reporter citing possible triggerer instability rather than a confirmed deterministic defect, due to lack of an isolated repro.
- #69739 (open) — adds trigger-event payload validation to `DbtCloudRunJobOperator`/`DbtCloudJobRunSensor`. Useful, but addresses a different symptom (malformed/unrecognized event payloads) and does not touch this root cause.

This report differs from the above in that the trace above shows the exact same failure mode reproduced independently three times in the same ~20 second window (three separate deferred `DbtCloudRunJobOperator` tasks polling around the same time), each with a matching stack trace ending in `secrets_masker.py:134 mask_secret` → `triggerer_job_runner.py:740 send` → `asgiref/sync.py:186 AsyncToSync.__call__`, and independently confirmed against the downstream API that the reported "job run has failed" was not based on a real observed terminal state for at least two of the three runs.

## What you think should happen instead?

At minimum, `mask_secret()` failures inside `_get_connection()` should not be allowed to abort connection retrieval / masquerade as the underlying job/task failing — masking is a logging side-effect and shouldn't be able to fail the primary operation. More fundamentally, `SUPERVISOR_COMMS.send()` (specifically the triggerer's comms decoder implementation) needs to correctly bridge calls made from within its own already-running event loop (e.g. via `run_coroutine_threadsafe` against the stored main loop) instead of raising when `AsyncToSync` detects it's already inside that loop.

## How to reproduce

1. Configure any secrets backend (e.g. `SecretsManagerBackend`) with a Connection that has `password` and/or `extra` populated.
2. Use that connection with a deferrable operator whose trigger re-fetches the connection during its `run()` poll loop (e.g. `DbtCloudRunJobOperator(deferrable=True)`, whose trigger calls `provide_account_id`'s `sync_to_async(self.get_connection)(...)` on each status check).
3. Trigger the DAG and let the task defer.
4. Observe the triggerer log for the task: it fails within seconds with `RuntimeError: You cannot use AsyncToSync in the same thread as an async event loop`, surfaced by the operator as a misleading `DbtCloudJobRunException: Job run <id> has failed.`
5. Separately query the actual job status via the downstream API — it will still be in progress or will complete successfully well after Airflow reported failure.

## Operating System

Amazon Linux 2 (MWAA managed runtime, Python 3.12)

## Versions of Apache Airflow Providers

```
apache-airflow-providers-amazon==9.12.0
apache-airflow-providers-dbt-cloud==4.4.2
asgiref==3.9.1
```
(per the `constraints-3.0.6/constraints-3.12.txt` constraint file; MWAA pre-installs/version-locks these and does not allow pinning them independently.)

## Deployment

Amazon Managed Workflows for Apache Airflow (MWAA)

## Deployment details

MWAA environment running Airflow 3.0.6 on Python 3.12, with `secrets.backend` configured to `SecretsManagerBackend`. Same crash signature reproduced across three separate DAGs (each a single deferrable `DbtCloudRunJobOperator` task) within the same ~20 second window.

## Anything else?

Happy to provide additional sanitized log excerpts if useful. I did not include specific connection IDs, hostnames, account identifiers, or credential values in this report.

## Are you willing to submit PR?

## Code of Conduct

amoghrajesh:
---

- [x] Yes

closes: #70398

### Why?

Deferred `DbtCloudRunJobOperator` tasks failed within seconds of deferring with a misleading `DbtCloudJobRunException: Job run <id> has failed`, even though the dbt Cloud job kept running and completed successfully hours later.

### What 

`DbtCloudHook.get_headers_tenants_from_connection()` and `get_job_details()` resolved the connection through the hook's synchronous `connection` cached_property while running inside the triggerer's live event loop. That
path masks the connection's secret via a synchronous comms call, which `asgiref` rejects when called from a thread already running an event loop the resulting `RuntimeError` propagated up through the trigger and was surfaced by the operator as a job failure instead of the real (transient) error it was.

Both call sites now resolve the connection through the existing async safe `get_async_connection()` helper, sharing a cache slot with the sync `connection` cached_property so neither path performs a duplicate connection lookup. Also switched the file's `Connection` type import to the `airflow.providers.common.compat.sdk` compat shim (matching what
`get_connection()`/`get_async_connection()` actually return under Airflow 3), which let a stale `# type: ignore[return-value]` on the `connection` property be removed.

---

### DIFF
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
@@ -351,6 +351,16 @@ class DbtCloudHook(HttpHook):
 
     @cached_property
     def connection(self) -> Connection:
+        """
+        Resolve and cache the dbt Cloud connection (sync).
+
+        Do not read this property from async code running inside the triggerer's
+        event loop — it calls the synchronous ``get_connection()``, whose secret-masking
+        step raises ``RuntimeError`` when invoked from a thread that's already running an
+        event loop. Use ``_resolve_connection_async()`` instead; it shares this property's
+        cache slot so the connection is still only looked up once per hook instance
+        regardless of which path is used first.
+        """
         return self._require_password(self.get_connection(self.dbt_cloud_conn_id))
 
     async def _resolve_connection_async(self) -> Connection:
--- a/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
+++ b/providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
@@ -41,7 +41,7 @@ from airflow.providers.http.hooks.http import HttpHook
 if TYPE_CHECKING:
     from requests.models import PreparedRequest, Response
 
-    from airflow.models import Connection
+    from airflow.providers.common.compat.sdk import Connection
 
 DBT_CAUSE_MAX_LENGTH = 255
 
@@ -254,12 +254,13 @@ class DbtCloudHook(HttpHook):
 
     async def get_headers_tenants_from_connection(self) -> tuple[dict[str, Any], str]:
         """Get Headers, tenants from the connection details."""
+        conn = await self._resolve_connection_async()
         headers: dict[str, Any] = {}
-        tenant = self._get_tenant_domain(self.connection)
+        tenant = self._get_tenant_domain(conn)
         package_name, provider_version = _get_provider_info()
         headers["User-Agent"] = f"{package_name}-v{provider_version}"
         headers["Content-Type"] = "application/json"
-        headers["Authorization"] = f"Token {self.connection.password}"
+        headers["Authorization"] = f"Token {conn.password}"
         return headers, tenant
 
     def _log_request_error(self, attempt_num: int, error: str) -> None:
@@ -307,7 +308,8 @@ class DbtCloudHook(HttpHook):
         endpoint = f"{account_id}/runs/{run_id}/"
         headers, tenant = await self.get_headers_tenants_from_connection()
         url, params = self.get_request_url_params(tenant, endpoint, include_related)
-        proxies = self._get_proxies(self.connection) or {}
+        conn = await self._resolve_connection_async()
+        proxies = self._get_proxies(conn) or {}
         proxy = proxies.get("https") if proxies and url.startswith("https") else proxies.get("http")
         extra_request_args = {}
 
@@ -341,13 +343,30 @@ class DbtCloudHook(HttpHook):
         job_run_status: int = response["data"]["status"]
         return job_run_status
 
+    @staticmethod
+    def _require_password(conn: Connection) -> Connection:
+        if not conn.password:
+            raise AirflowException("An API token is required to connect to dbt Cloud.")
+        return conn
+
     @cached_property
     def connection(self) -> Connection:
-        _connection = self.get_connection(self.dbt_cloud_conn_id)
-        if not _connection.password:
-            raise AirflowException("An API token is required to connect to dbt Cloud.")
+        return self._require_password(self.get_connection(self.dbt_cloud_conn_id))
 
-        return _connection  # type: ignore[return-value]
+    async def _resolve_connection_async(self) -> Connection:
+        """
+        Resolve and cache the dbt Cloud connection (async).
+
+        Shares the ``connection`` cached_property's cache slot so a connection
+        fetched from either the sync or async path is not looked up twice on
+        the same hook instance, and so the async path never touches the sync
+        ``get_connection()``/``mask_secret()`` path from inside a running
+        event loop (which raises in the triggerer).
+        """
+        if "connection" not in self.__dict__:
+            conn = await get_async_connection(self.dbt_cloud_conn_id)
+            self.__dict__["connection"] = self._require_password(conn)
+        return self.__dict__["connection"]
 
     def get_conn(self, *args, **kwargs) -> Session:
         tenant = self._get_tenant_domain(self.connection)
--- a/providers/dbt/cloud/tests/unit/dbt/cloud/hooks/test_dbt.py
+++ b/providers/dbt/cloud/tests/unit/dbt/cloud/hooks/test_dbt.py
@@ -356,6 +356,107 @@ class TestDbtCloudHook:
             assert mock_get_connection.call_count == 1
             assert mock_get_async_connection.call_count == 0
 
+    @pytest.mark.asyncio
+    async def test_get_headers_tenants_from_connection_does_not_use_sync_get_connection(self):
+        hook = DbtCloudHook(ACCOUNT_ID_CONN)
+
+        with (
+            patch.object(DbtCloudHook, "get_connection") as mock_get_connection,
+            patch(
+                "airflow.providers.dbt.cloud.hooks.dbt.get_async_connection",
+                new=AsyncMock(
+                    return_value=Connection(
+                        conn_id=ACCOUNT_ID_CONN,
+                        conn_type=DbtCloudHook.conn_type,
+                        login=str(DEFAULT_ACCOUNT_ID),
+                        password=TOKEN,
+                        host=SINGLE_TENANT_DOMAIN,
+                    )
+                ),
+            ) as mock_get_async_connection,
+        ):
+            headers, tenant = await hook.get_headers_tenants_from_connection()
+
+            assert tenant == SINGLE_TENANT_DOMAIN
+            assert headers["Authorization"] == f"Token {TOKEN}"
+            mock_get_connection.assert_not_called()
+            assert mock_get_async_connection.call_count == 1
+
+    @pytest.mark.asyncio
+    async def test_resolve_connection_cached_async(self):
+        hook = DbtCloudHook(ACCOUNT_ID_CONN)
+
+        with patch(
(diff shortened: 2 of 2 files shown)


eladkal (approved):

providers/dbt/cloud/src/airflow/providers/dbt/cloud/hooks/dbt.py
```diff
@@ -341,13 +343,30 @@ async def get_job_status(
...
         return job_run_status
 
+    @staticmethod
+    def _require_password(conn: Connection) -> Connection:
+        if not conn.password:
+            raise AirflowException("An API token is required to connect to dbt Cloud.")
+        return conn
+
     @cached_property
     def connection(self) -> Connection:
```
Lee-W, line 353: I think we'll at least need some docstring here or someone else will use this again and encounter the same issue. Are we encountering such behavior in other hooks as well? or is it a dbt specific thing?

amoghrajesh: This is a general issue, not dbt specific and any hook that lazily resolves a `Connection` via a sync `cached_property` and gets called from inside an async triggers run() loop can hit it. tasl SDK fixed the general case in #55812, and this same "sync cached_property read from async code". 

dbt cloud `_resolve_account_id_async()` had already been fixed similarly (#56791) but the connection headers path was missed thats what this PR closes. Added a docstring on connection cross referencing `_resolve_connection_async()` so this does not get missed again on this hook.

amoghrajesh: Handled in dcac1a869f

Lee-W (approved):

PR #70583, opened 2026-07-28
Add `get_display_name` to `BaseUser` interface 

pierrejeambrun:
Follow-up to #68833.

That PR exposed `owner_display_name` in the audit log, but the action-logging decorator stored `user.get_name()` in **both** `owner` and `owner_display_name`, so the new column always matched the existing owner column — no visible benefit.

This adds a `get_display_name()` to the auth manager user model so the two can differ:

- `BaseUser.get_display_name()` is **not** abstract and defaults to `get_name()`, so auth managers written before this method existed keep working unchanged.
- The FAB user returns its full name (`get_full_name()`).
- `action_logging` now records `get_display_name()` in `owner_display_name` while `owner` keeps the raw identifier (`get_name()`), so the audit log can show a friendlier name where the auth manager provides one (and falls back to `owner` everywhere else).

The simple auth manager only knows a username, so it intentionally has no override and uses the inherited fallback.

For `SimpleAuthManager`, nothing changes. (display_name = name), new method isn't overriden:
Screenshot 2026-07-28 at 11 26 59

For `FabAuthManager`, which implements `get_display_name` as a `get_full_name`:
Screenshot 2026-07-28 at 11 33 01

Filtering 
Screenshot 2026-07-28 at 11 57 15

---

### DIFF
--- a/airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
+++ b/airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
@@ -16,14 +16,18 @@
 # under the License.
 from __future__ import annotations
 
+import asyncio
 import json
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
+from airflow.api_fastapi.auth.managers.models.base_user import BaseUser
 from airflow.api_fastapi.logging.decorators import (
     _mask_connection_fields,
     _mask_variable_fields,
     _sanitize_for_stdlib_log,
+    action_logging,
 )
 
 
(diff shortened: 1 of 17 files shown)


pierrejeambrun:
Also adjust the front-end filter so we are now searching for 'owner_display_name' instead. This way 'Thor' will work.

airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
```diff
@@ -122,3 +126,33 @@ def test_masks_val_alias(self):
...
+            def get_display_name(self) -> str:
+                return "Jane Doe"
+
+        request = MagicMock()
+        request.headers = {}
+        request.method = "GET"
+        request.query_params = {}
+        request.path_params = {}
+        request.body = AsyncMock(return_value=b"")
+        session = MagicMock()
```
amoghrajesh: Can we add a spec or autospec please?

amoghrajesh (approved):
LGTM with one nit.

airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
pierrejeambrun: I've rewritten this path to use a real Request object instead which is cleaner and added `spec` to Session mock object.

PR #63185, opened 2026-03-09
Guard 2 to 3+ migration against custom Dag bundle configs using startup-based initialization

jason810496:
- related: #47592
- closes: #63323

## Why

Some users encounter a flaky error during 2.x → 3.1+ upgrades. If the deployment configures any bundle other than the default `dags-folder`, the `0082_3_1_0_make_bundle_name_not_nullable` migration writes a single hard-coded bundle name that no longer matches the user's runtime config, so triggering a DagRun raises:

https://github.com/apache/airflow/blob/450a08e78499fbd6d21d2d6dc2392dff9dbfda98/airflow-core/src/airflow/dag_processing/bundles/manager.py#L333-L345

Reproducible edge cases:

**Case 1**: Dag bundle name is not `dags-folder`

```bash
AIRFLOW__DAG_PROCESSOR__DAG_BUNDLE_CONFIG_LIST='[{"name": "main", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle", "kwargs": {"path": "/files/dags", "refresh_interval": 1}}]'
```

**Case 2**: Multiple custom Dag bundles

```bash
AIRFLOW__DAG_PROCESSOR__DAG_BUNDLE_CONFIG_LIST='[{"name": "main1", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle", "kwargs": {"path": "/files/dags", "refresh_interval": 1}}, {"name": "main2", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle", "kwargs": {"path": "/files/dags", "refresh_interval": 1}}]'
```

In both cases the migration inserts `dags-folder` rather than the configured bundle, leaving the Dag rows pointing at a `bundle_name` that the running deployment cannot resolve.

## How

- At `DagFileProcessorManager` startup, after `sync_bundles_to_db` flushes the latest bundle state, scan `DagModel` for legacy-candidate rows (NULL `relative_fileloc` and no `DagVersion` for that `dag_id`) and route them to the most-specific configured bundle whose absolute `path` contains the Dag's `fileloc`. `relative_fileloc` is written at the same time so fileloc-based stale-detection works later.
[...]

PR #70586, opened 2026-07-28
Fix deactivation of stale zip-packaged DAGs

SamWheating:
#### The Issue:

When deactivating DAGs, airflow compares the `DagFileStat` generated by the manager to a DAG's fileloc in order to identify DAGs which were not seen on the most recent file parse. 

However, when processing zip-packaged DAGs, the DAG's relative_fileloc will be the value of `__file__` reported by zipfile, which includes the inner path within the zip archive (ex: `path/to/some/dag_package.zip/my_dag.py`).

Meanwhile the parsing stats used to index the `last_parsed` dictionary use the path of the archive file itself (ex:  `path/to/some/dag_package.zip`) since there is a single entry per file-parsed.

As a result, when looking at zip-packaged DAGs the `last_parsed.get()` lookup here will never return a value:
https://github.com/apache/airflow/blob/1438ea3587031417cc85d74323235cf087a058fb/airflow-core/src/airflow/dag_processing/manager.py#L481-L489

And thus zip-packaged DAGs are never properly deactivated. 

This issue doesn't surface too often, since most DAGs will also be caught by the `deactivate_deleted_dags` function which properly evaluates them based on inner paths. However for files which produce multiple DAGs ("dag factories") or DAGs which are renamed within a file, this second mechanism won't work either, since DAGs can be removed without the source file being removed.

As a result, certain DAGs within a zip file will never be removed and airflow will continue to attempt to schedule and run them. I have reproduced this issue in breeze on the latest version of `main`.

#### The Fix:

Within the `deactivate_stale_dags` function, we can just remove the inner part of the fileloc for zip-packaged DAGs to obtain the actual filepath of the archive itself. I've added this as a private method on the class, but it could also be added as a property on the `DAG` model itself.

---

No

### DIFF
--- a/airflow-core/src/airflow/dag_processing/manager.py
+++ b/airflow-core/src/airflow/dag_processing/manager.py
@@ -452,20 +452,6 @@ class DagFileProcessorManager(LoggingMixin):
         """Clean up stale DAG bundle version usage records."""
         BundleUsageTrackingManager().remove_stale_bundle_versions()
 
-    @staticmethod
-    def _file_name_from_fileloc(fileloc: str) -> str:
-        """
-        If a python file comes from a zip module, return just the path to the zip file.
-
-        file parsing stats are keyed by the top-level filename (path/to/archive.zip).
-        in order to correctly link DAGs to their relevant parsing stats, we need to normalize
-        fileloc to the archive file location, if applicable.
-        """
-        if ".zip" in fileloc:
-            return fileloc[: fileloc.index(".zip") + 4]
-
-        return fileloc
-
     @provide_session
     def deactivate_stale_dags(
         self,
@@ -520,9 +506,11 @@ class DagFileProcessorManager(LoggingMixin):
             # When the Dag's last_parsed_time is more than the stale_dag_threshold older than the
             # Dag file's last_finish_time, the Dag is considered stale as has apparently been removed from the file,
             # This is especially relevant for Dag files that generate Dags in a dynamic manner.
-            file_info = DagFileInfo(
-                rel_path=Path(self._file_name_from_fileloc(dag.relative_fileloc)), bundle_name=dag.bundle_name
-            )
+            rel_path = Path(dag.relative_fileloc)
+            file_info = DagFileInfo(rel_path=rel_path, bundle_name=dag.bundle_name)
+            if file_info not in last_parsed:
+                # Zip-packaged dags are keyed by the archive path, not the inner file, so try the parent as well
+                file_info = DagFileInfo(rel_path=rel_path.parent, bundle_name=dag.bundle_name)
             if last_finish_time := last_parsed.get(file_info, None):
                 if dag.last_parsed_time + timedelta(seconds=self.stale_dag_threshold) < last_finish_time:
                     self.log.info(
--- a/airflow-core/src/airflow/dag_processing/manager.py
+++ b/airflow-core/src/airflow/dag_processing/manager.py
@@ -452,6 +452,20 @@ class DagFileProcessorManager(LoggingMixin):
         """Clean up stale DAG bundle version usage records."""
         BundleUsageTrackingManager().remove_stale_bundle_versions()
 
+    @staticmethod
+    def _file_name_from_fileloc(fileloc: str) -> str:
+        """
+        If a python file comes from a zip module, return just the path to the zip file.
+
+        file parsing stats are keyed by the top-level filename (path/to/archive.zip).
+        in order to correctly link DAGs to their relevant parsing stats, we need to normalize
+        fileloc to the archive file location, if applicable.
+        """
+        if ".zip" in fileloc:
+            return fileloc[: fileloc.index(".zip") + 4]
+
+        return fileloc
+
     @provide_session
     def deactivate_stale_dags(
         self,
@@ -506,7 +520,9 @@ class DagFileProcessorManager(LoggingMixin):
             # When the Dag's last_parsed_time is more than the stale_dag_threshold older than the
             # Dag file's last_finish_time, the Dag is considered stale as has apparently been removed from the file,
             # This is especially relevant for Dag files that generate Dags in a dynamic manner.
-            file_info = DagFileInfo(rel_path=Path(dag.relative_fileloc), bundle_name=dag.bundle_name)
+            file_info = DagFileInfo(
+                rel_path=Path(self._file_name_from_fileloc(dag.relative_fileloc)), bundle_name=dag.bundle_name
+            )
             if last_finish_time := last_parsed.get(file_info, None):
                 if dag.last_parsed_time + timedelta(seconds=self.stale_dag_threshold) < last_finish_time:
                     self.log.info(
--- a/airflow-core/tests/unit/dag_processing/test_manager.py
+++ b/airflow-core/tests/unit/dag_processing/test_manager.py
@@ -1048,7 +1048,8 @@ class TestDagFileProcessorManager:
     def test_scan_stale_dags_deactivates_zip_packaged_dags(self, session, test_zip_path):
         """
         Ensure that zip-packaged DAGs are marked inactive when the file is parsed but the
-        DagModel.last_parsed_time is not updated.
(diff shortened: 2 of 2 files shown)


airflow-core/src/airflow/dag_processing/manager.py
```diff
@@ -445,6 +445,20 @@ def cleanup_stale_bundle_versions(self) -> None:
...
+    @staticmethod
+    def _file_name_from_fileloc(fileloc: str) -> str:
+        """
+        If a python file comes from a zip module, return just the path to the zip file.
+
+        file parsing stats are keyed by the top-level filename (path/to/archive.zip).
+        in order to correctly link DAGs to their relevant parsing stats, we need to normalize
+        fileloc to the archive file location, if applicable.
+        """
+        if ".zip" in fileloc:
```
kaxil: This matches `.zip` anywhere in the string, so any fileloc that merely contains those four characters gets truncated: `dags/report.zip.py` becomes `dags/report.zip`, and a file under a directory named `extracted.zip/` loses its filename too. The truncated key won't be in `last_parsed`, so the staleness check is silently skipped for such files (the same failure mode this PR fixes), and if the truncated path collides with a real archive's stats key, a live DAG inherits the archive's `last_finish_time` and can be wrongly deactivated.

One way to avoid the string surgery entirely is to try the exact key first and fall back to the parent on a miss. Zip members are only imported from the archive root (`_load_modules_from_zip` skips entries with `len(zip_path.parts) > 1`), so the parent of a zip-inner fileloc is always the archive itself:

```python
rel_path = Path(dag.relative_fileloc)
file_info = DagFileInfo(rel_path=rel_path, bundle_name=dag.bundle_name)
if file_info not in last_parsed:
    # Zip-packaged dags are keyed by the archive path, not the inner file.
    file_info = DagFileInfo(rel_path=rel_path.parent, bundle_name=dag.bundle_name)
```

This also covers archives not named `*.zip`, which do get parsed today since discovery and routing are content-based (`zipfile.is_zipfile`) but would never match the `.zip` check here. Reusing `correct_maybe_zipped` isn't an option either, its `zipfile.is_zipfile` gate would resolve the bundle-relative path against the processor's cwd and return the fileloc unchanged.

airflow-core/tests/unit/dag_processing/test_manager.py
```diff
@@ -1046,6 +1046,72 @@ def test_scan_stale_dags(self, session):
...
+        manager = DagFileProcessorManager(
+            max_runs=1,
+            processor_timeout=10 * 60,
+        )
+        bundle = MagicMock()
+        bundle.name = "testing"
+        manager._dag_bundles = [bundle]
+
+        test_dag_path = DagFileInfo(
+            rel_path=Path(test_zip_path),
```
kaxil: Since `bundle_path` isn't set here, `test_dag_path.bundle_path` is None, `DagBag._get_relative_fileloc` falls through, and the DB row ends up with an absolute `relative_fileloc`. Production always stores bundle-relative paths, and the sibling `test_scan_stale_dags` above sets `bundle_path=TEST_DAGS_FOLDER` with a true relative path. Using `rel_path=Path("test_zip.zip"), bundle_path=Path(test_zip_path).parent` (and building the `DagBag` from `test_dag_path.absolute_path`) would exercise the same key shapes as production, and stops the test depending on pytest's tmp dir name not containing `.zip`.

airflow-core/tests/unit/dag_processing/test_manager.py
```diff
@@ -1046,6 +1046,72 @@ def test_scan_stale_dags(self, session):
...
+        # Add DAG to the file_parsing_stats
+        stat = DagFileStat(
+            num_dags=1,
+            import_errors=0,
+            last_finish_time=timezone.utcnow() + timedelta(hours=1),
+            last_duration=1,
+            run_count=1,
+            last_num_of_db_queries=1,
+        )
+        manager._files = [test_dag_path]
```
kaxil: I don't think either `manager._files` or the `MagicMock` bundle above is read on this code path. `_files` isn't an attribute the manager itself uses (only tests assign it), and `deactivate_stale_dags` reads bundle state from the DB, which the `testing_dag_bundle` fixture provides. Looks carried over from the sibling test, these lines can be dropped.

airflow-core/src/airflow/dag_processing/manager.py
SamWheating: Oh this is a really good suggestion - will update!

Just thinking about whether or not this will introduce other weird edge cases and collisions between archives etc. I think its fine as long as users aren't creating zip-files ending in `.py` or anything super weird like that.

airflow-core/tests/unit/dag_processing/test_manager.py
SamWheating: Yeah, shameless copy+paste from other tests, will clean this up 👍

SamWheating: > _files isn't an attribute the manager itself uses (only tests assign it)

Actually looking into this a bit more, maybe the other tests don't use it either 🤔 Will see if its safe to remove

SamWheating:
Test failure seems unrelated (timeout from dockerhub) could someone re-run when you have a chance? Thanks!

airflow-core/src/airflow/dag_processing/manager.py
```diff
@@ -477,10 +477,11 @@ def deactivate_stale_dags(
...
                 to_deactivate.add(dag.dag_id)
                 continue
-            # When the Dag's last_parsed_time is more than the stale_dag_threshold older than the
-            # Dag file's last_finish_time, the Dag is considered stale as has apparently been removed from the file,
-            # This is especially relevant for Dag files that generate Dags in a dynamic manner.
-            file_info = DagFileInfo(rel_path=Path(dag.relative_fileloc), bundle_name=dag.bundle_name)
+            rel_path = Path(dag.relative_fileloc)
+            file_info = DagFileInfo(rel_path=rel_path, bundle_name=dag.bundle_name)
+            if file_info not in last_parsed:
+                # Zip-packaged dags are keyed by the archive path, not the inner file, so try the parent as well
```
kaxil, line 512: Since this hunk now conflicts with main (#63185 rewrote it and kept these lines), can the rebase also restore the original three-line comment explaining the staleness heuristic? The threshold comparison below is still the non-obvious part, and the new comment only covers the zip keying.

airflow-core/tests/unit/dag_processing/test_manager.py
```diff
@@ -1046,6 +1044,72 @@ def test_scan_stale_dags(self, session):
         # SerializedDagModel gives history about Dags
         assert serialized_dag_count == 1
 
+    @pytest.mark.usefixtures("testing_dag_bundle")
+    def test_scan_stale_dags_deactivates_zip_packaged_dags(self, session, test_zip_path):
+        """
+        Ensure that zip-packaged DAGs are marked inactive when the file is parsed but the
```
kaxil, line 1050: Could the docstring say what this test pins down? Zip dags store `relative_fileloc` as `<archive>.zip/<inner>.py` while `_file_stats` is keyed by the archive path, so without the parent fallback the lookup misses and the dag is never deactivated. As written it reads the same as the sibling test's docstring.

airflow-core/src/airflow/dag_processing/manager.py
kaxil, line 512: On the PR body's idea of a property on the DAG model: +1 to giving this mapping a name eventually. This lookup is the reverse of what `_get_observed_filelocs` does for `deactivate_deleted_dags` (archive to inner paths there, fileloc to parse-unit key here), and nothing links the two sites today. If the importer's one-level rule (`_load_modules_from_zip` skips entries with `len(zip_path.parts) > 1`) ever changes, the deletion path would keep working while this fallback silently stops matching. A small helper or `DagFileInfo` classmethod would keep them in step. Fine as a follow-up rather than in this PR.

kaxil (approved):

amoghrajesh:
Conflicts on this need to be handled.

PR #70588, opened 2026-07-28
Respect [fab] update_fab_perms when syncing per-DAG permissions during parsing

dor-bernstein:
During DAG parsing, `_serialize_dag_capturing_errors()` calls `_sync_dag_perms()` for every DAG whenever the FAB auth manager is active — regardless of the `[fab] update_fab_perms` setting. `sync_perm_for_dag()` creates/refreshes a per-DAG resource on every parse, which is avoidable DB work for deployments that manage DAG access at the role level and have disabled automatic FAB permission management.

This makes the per-DAG permission sync honor `[fab] update_fab_perms`:

- Default is `True`, so **there is no behavior change for existing configurations**.
- When set to `False` (operator has opted out of automatic FAB permission management), the per-DAG sync is skipped, avoiding the extra per-parse writes.

The flag's documented purpose is "Update FAB permissions and sync security manager roles", so honoring it in this path is consistent with its intent.


dor-bernstein:
Superseded by https://github.com/apache/airflow/pull/70589 (reopened from a personal fork). Closing this one.

PR #70593, opened 2026-07-28
Revoke JWT token on provider-level /auth/logout routes

securylight:
Core's `/auth/logout` (#67289) revokes the current JWT before redirecting to the auth manager's logout URL. FabAuthManager's and KeycloakAuthManager's own `/auth/logout` routes don't do the same before redirecting to the IdP, so a token stays valid if that route is reached directly. This adds the same `revoke_token()` call to both, for consistency.

## Test plan
- [x] Added `test_logout_revokes_token` / `test_logout_without_token_cookie_does_not_revoke` for both providers
- [x] Existing logout tests still pass

---


vincbeck (changes requested):
This is done even if the auth manager has its custom logout route. Why doing it twice? https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/api_fastapi/core_api/routes/public/auth.py#L65

PR #70594, opened 2026-07-28
Stop auto-scrolling task logs while the user is selecting text

Andrushika:
UI: Stop auto-scrolling task logs while the user is selecting text
---
When a task is still running, the task log view scrolls to the bottom every time new lines arrive. If the user is reading or selecting text near the bottom, the jump interrupts them. 
This PR skips the auto-scroll while a pointer drag is in progress, and restores it once the drag ends.

Before:

https://github.com/user-attachments/assets/235f0e37-b8fe-441b-8824-45375714160c

After:

https://github.com/user-attachments/assets/44d38897-4bda-4713-9a20-231d7c7247a6

---


pierrejeambrun (commented):
Thanks for the PR, good idea.

Instead of stopping the auto scroll when a pointer is up / down / selecting. 

We should instead stop / resume the auto scroll feature depending on if the users is at the bottom (very end of the scroll) or not.

If someone scrolled up a little bit to see something, the auto scroll should also stop.

Even after text selection, if someone scroll back down to the very end of logs, auto scroll should resume.

Andrushika:
Hi @pierrejeambrun, thanks for reviewing!

Your idea sounds great. There’s just one edge case I’d like to confirm: what if the user is already at the bottom of the log but is selecting some log text?

IIUC, the expected behavior would be to keep auto-scrolling even while the user is selecting text. But would it be better to pause auto-scrolling in this case?

Andrushika:
> IIUC, the expected behavior would be to keep auto-scrolling even while the user is selecting text. But would it be better to pause auto-scrolling in this case?

I assume this is correct and have updated it into this pattern. Happy to fix if it does not meet the expectation!

After:

https://github.com/user-attachments/assets/968e3e7f-3221-4585-90f3-4d43ef046635

Andrushika:
The only failed CI is unrelated to this PR.

bbovenzi (approved):

PR #69269, opened 2026-07-02
cncf-kubernetes: fix potential race condition in trigger_reentry flow caused by `is_istio_enabled`

noah-gil:
# Problem
When using the deferred `KubernetesPodOperator`, there is a potential race condition in the `trigger_reentry` success flow. After a pod has completed and the worker resumes the workload from the triggerer, it reads the pod from the Kubernetes API, checking that it still exists:
```py
try:
    self.pod = self.hook.get_pod(pod_name, pod_namespace)
except ApiException as e:
    if e.status != 404:
        raise
```
It then continues down the happy path until it reaches this part in `_clean`:
```py
if self.pod is None:
    return

istio_enabled = self.is_istio_enabled(self.pod)
```
Then it `is_istio_enabled`, it attempts a second pod read from the Kubernetes API:
```py
remote_pod = self.pod_manager.read_pod(pod)
```
At this point, it is possible that the pod has now been deleted by some external factor during the time between the first and second pod read. Should this occur, the task fails due to a raised `ApiException` error. Even though the pod succeeded, it is marked as error due to this failed API read, exactly what the first read tried to check for.

# Solution
I chose to mimic the existing case of returning `False` when there is no pod provided in the function's argument. It narrowly catches the 404 and returns false, while still re-raising an exception for any other API error. I am open for discussion on other approaches for this fix. As this is quite a small patch, do let me know if I should create any tests to test for this exact scenario.

Thank you for your time!

---

##### Was generative AI tooling used to co-author this PR?

- [ ] Yes (please specify the tool below)

---

PR #70595, opened 2026-07-28
Add `--min-completed-minutes` to `cleanup-pods` to prevent KPO race condition

noamst-monday:
## Problem

`KubernetesPodOperator` in synchronous (`deferrable=False`) mode calls `PodManager.await_pod_completion` to confirm the pod reached a terminal phase. This method polls `read_pod` every **2 seconds** ([`pod_manager.py:839`](https://github.com/apache/airflow/blob/main/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/utils/pod_manager.py#L839), hardcoded `time.sleep(2)`).

When a pod transitions to `Succeeded`, there is up to a 2-second window before `PodManager.await_pod_completion` observes the terminal phase. The `airflow kubernetes cleanup-pods` command currently deletes Succeeded/Failed/Evicted pods **immediately** — there is no minimum-age guard for terminal states.

If the cleanup job fires during that window, `PodManager.await_pod_completion`'s next `read_pod` call returns 404 and the task is marked **FAILED**, even though the pod completed successfully (exit code 0).

A `--min-pending-minutes` guard already exists for Pending pods (default 30 m, minimum 5 m). No equivalent exists for terminal states.

### Real-world reproduction

Confirmed via Kubernetes API server audit logs on a production EKS cluster running `apache-airflow-providers-cncf-kubernetes==10.19.0` on Airflow 3.2.2. The cleanup CronJob was set to run every 5 minutes.

Timeline of a representative failure:

```
13:15:12Z  PodManager polls pod       → phase Running
13:15:14Z  Container exits            → exit code 0 (pod phase → Succeeded)
13:15:17Z  cleanup-pods deletes pod   → 3 s after completion
13:15:18Z  PodManager polls pod       → 404 Not Found → task FAILED
```

Airflow task log excerpt:
```
[INFO]  Pod run-o1sxc2on has phase Running
[ERROR] Task failed with exception
ApiException: (404) Reason: Not Found
HTTP response body: {"message":"pods \"run-o1sxc2on\" not found","reason":"NotFound","code":404}
```

Audit log at deletion time confirmed `phase: Succeeded`, `containerStatuses[0].state.terminated.exitCode: 0`, `finishedAt: 13:15:14Z`.

### Related

PR #69269 fixed a similar 404 race in `is_istio_enabled` for the deferrable `trigger_reentry` path (10.20.0). This PR addresses the complementary gap: preventing the race at source by not deleting recently-completed pods.

## Solution

Add `--min-completed-minutes` (default `1`) to the `cleanup-pods` command. When set, Succeeded/Failed/Evicted pods are skipped unless their completion time exceeds the threshold. Set to `0` to restore the previous immediate-deletion behaviour.

Completion time is derived from the latest `finished_at` across both `container_statuses` **and** `init_container_statuses`. Falls back to the latest `conditions[*].last_transition_time` (updated at the terminal transition), then to `creation_timestamp` as a last resort.

## Changes

| File | Change |
|---|---|
| `cli/definition.py` | Add `ARG_MIN_COMPLETED_MINUTES` (default `1`, `allow_zero=True`) |
| `cli/kubernetes_command.py` | Add `_get_pod_completion_time()`; gate terminal-pod deletion by age |
| `tests/.../test_kubernetes_command.py` | `TestGetPodCompletionTime` with real k8s model objects; 4 integration-style tests for the new flag |
| `docs/changelog.rst` | Not modified — maintained by release manager |

CLI reference docs (`cli-ref.rst`) use `.. argparse::` and pick up the new flag automatically.

## Testing

```bash
pytest providers/cncf/kubernetes/tests/unit/cncf/kubernetes/cli/test_kubernetes_command.py -q
# 19 passed
```

## Usage

```bash
# Helm chart — keep default of 1 minute:
cleanup:
  args: ["bash", "-c", "exec airflow kubernetes cleanup-pods --namespace {{ .Release.Namespace }}"]

# Or set explicitly:
airflow kubernetes cleanup-pods --namespace airflow --min-completed-minutes 5
```

---

- [X] Yes (Claude Sonnet 4.6, Opus 5)

Generated-by: Claude Sonnet 4.6 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/cncf/kubernetes/docs/changelog.rst
+++ b/providers/cncf/kubernetes/docs/changelog.rst
@@ -33,7 +33,7 @@ Changelog
 New features
 ~~~~~~~~~~~~
 
-* ``Add --min-completed-minutes to cleanup-pods to prevent KPO race condition (#XXXXX)``
+* ``Add --min-completed-minutes to cleanup-pods to prevent KPO race condition (#70595)``
 
 10.20.0
 .......
--- a/providers/cncf/kubernetes/docs/changelog.rst
+++ b/providers/cncf/kubernetes/docs/changelog.rst
@@ -27,14 +27,6 @@
 Changelog
 ---------
 
-10.21.0
-.......
-
-New features
-~~~~~~~~~~~~
-
-* ``Add --min-completed-minutes to cleanup-pods to prevent KPO race condition (#70595)``
-
 10.20.0
 .......
 
--- a/providers/cncf/kubernetes/docs/changelog.rst
+++ b/providers/cncf/kubernetes/docs/changelog.rst
@@ -27,6 +27,14 @@
 Changelog
 ---------
 
+10.21.0
+.......
+
+New features
+~~~~~~~~~~~~
+
+* ``Add --min-completed-minutes to cleanup-pods to prevent KPO race condition (#XXXXX)``
+
 10.20.0
 .......
 
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
@@ -68,6 +68,19 @@ ARG_MIN_PENDING_MINUTES = Arg(
     ),
 )
 
+ARG_MIN_COMPLETED_MINUTES = Arg(
+    ("--min-completed-minutes",),
+    default=0,
+    type=positive_int(allow_zero=True),
+    help=(
+        "Minimum age in minutes of a completed (Succeeded/Failed/Evicted) pod before it is deleted. "
+        "Defaults to 0 (delete immediately, preserving current behaviour). "
+        "Set this to a value greater than the KubernetesPodOperator poll interval (~2 s) to prevent "
+        "a race where the cleanup job removes a pod before KPO observes its terminal phase, "
+        "causing a spurious task failure despite the pod having succeeded."
+    ),
+)
+
 ARG_TEAM = Arg(
     ("--team",),
     default=None,
@@ -84,7 +97,7 @@ KUBERNETES_COMMANDS = (
             "in evicted/failed/succeeded/pending states"
         ),
         func=lazy_load_command("airflow.providers.cncf.kubernetes.cli.kubernetes_command.cleanup_pods"),
-        args=(ARG_NAMESPACE, ARG_MIN_PENDING_MINUTES, ARG_VERBOSE),
+        args=(ARG_NAMESPACE, ARG_MIN_PENDING_MINUTES, ARG_MIN_COMPLETED_MINUTES, ARG_VERBOSE),
     ),
     ActionCommand(
         name="generate-dag-yaml",
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
@@ -70,14 +70,11 @@ ARG_MIN_PENDING_MINUTES = Arg(
 
 ARG_MIN_COMPLETED_MINUTES = Arg(
     ("--min-completed-minutes",),
-    default=0,
+    default=1,
     type=positive_int(allow_zero=True),
     help=(
         "Minimum age in minutes of a completed (Succeeded/Failed/Evicted) pod before it is deleted. "
-        "Defaults to 0 (delete immediately, preserving current behaviour). "
-        "Set this to a positive value to prevent a race condition where the cleanup job removes a "
-        "just-completed pod before KubernetesPodOperator has polled its terminal phase, "
-        "causing a spurious task failure despite the pod having succeeded."
+        "Default is 1. Set to 0 to delete immediately."
     ),
 )
 
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
@@ -75,8 +75,8 @@ ARG_MIN_COMPLETED_MINUTES = Arg(
     help=(
         "Minimum age in minutes of a completed (Succeeded/Failed/Evicted) pod before it is deleted. "
         "Defaults to 0 (delete immediately, preserving current behaviour). "
-        "Set this to a value greater than the KubernetesPodOperator poll interval (2 s, see ``await_pod_completion``) to prevent "
-        "a race where the cleanup job removes a pod before KPO observes its terminal phase, "
+        "Set this to a positive value to prevent a race condition where the cleanup job removes a "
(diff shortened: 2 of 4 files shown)


providers/cncf/kubernetes/docs/changelog.rst
eladkal, line 1: Please rollback changes to this file.
Changelog is automatically generated by release manager

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
```diff
@@ -68,6 +68,19 @@
     ),
 )
 
+ARG_MIN_COMPLETED_MINUTES = Arg(
+    ("--min-completed-minutes",),
+    default=0,
+    type=positive_int(allow_zero=True),
+    help=(
+        "Minimum age in minutes of a completed (Succeeded/Failed/Evicted) pod before it is deleted. "
+        "Defaults to 0 (delete immediately, preserving current behaviour). "
```
jedcunningham: Lets set a default age to avoid this race. No reason to maintain a race condition :)

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
```diff
@@ -68,6 +68,19 @@
...
+        "Set this to a positive value to prevent a race condition where the cleanup job removes a "
+        "just-completed pod before KubernetesPodOperator has polled its terminal phase, "
+        "causing a spurious task failure despite the pod having succeeded."
```
jedcunningham: We can probably shorten or remove this completely.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/definition.py
```diff
@@ -68,6 +68,19 @@
     ),
 )
 
+ARG_MIN_COMPLETED_MINUTES = Arg(
+    ("--min-completed-minutes",),
+    default=0,
```
noamst-monday: Do we want a different default here? Maybe allow >3 seconds grace by default to be safe from KPO polling interval?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/kubernetes_command.py
```diff
@@ -119,6 +119,21 @@ def generate_pod_yaml(args):
...
+
+    Uses the latest ``finished_at`` timestamp across all container statuses so that pods
+    with multiple containers (e.g. an init container + a base container) are judged by
+    the time the *last* container finished, not by when the pod was created.
+    """
+    times = []
+    for status in pod.status.container_statuses or []:
+        if status.state and status.state.terminated and status.state.terminated.finished_at:
+            times.append(status.state.terminated.finished_at)
+    return max(times) if times else pod.metadata.creation_timestamp
```
noamst-monday: Not sure about this one, open for suggestions

providers/cncf/kubernetes/docs/changelog.rst
noamst-monday: thanks, reverted.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/kubernetes_command.py
```diff
@@ -119,6 +119,21 @@ def generate_pod_yaml(args):
...
+def _get_pod_completion_time(pod):
+    """
+    Return the time the pod entered a terminal state, or its creation time as fallback.
+
+    Uses the latest ``finished_at`` timestamp across all container statuses so that pods
+    with multiple containers (e.g. an init container + a base container) are judged by
+    the time the *last* container finished, not by when the pod was created.
+    """
+    times = []
+    for status in pod.status.container_statuses or []:
```
jedcunningham: Init container statuses live in a separate field, `pod.status.init_container_statuses`, so they're never seen.

That matters because of the fallback direction. When an init container fails, the pod goes Failed/restartPolicy: Never, but the base container never starts, so it has no terminated state and times comes back empty. Then we return creation_timestamp, which is always earlier than real completion, so the age comes out inflated and the pod is deleted immediately. Same for pods evicted before containers started. In other words the guard silently no-ops in some of the exact terminal states the flag advertises, and it fails in the unsafe direction (delete too early) rather than the safe one.

Two changes:

1. Scan init container statuses too:

`statuses = [*(pod.status.container_statuses or []), *(pod.status.init_container_statuses or [])]`

2. Pick a fallback that isn't strictly older than completion. `max(c.last_transition_time for c in pod.status.conditions)` is at-or-after the terminal transition (the Ready condition flips on eviction) and is always present for a scheduled pod.

Also, no test currently covers either behavior the helper exists for — every case builds exactly one container status, so `max()` over multiple is unexercised, and because _make_pod uses a bare `MagicMock()`, `status.state` and `status.state.terminated` are auto-truthy and the None guards can't be exercised as false. A small parametrized unit test on `_get_pod_completion_time` with real `k8s.V1ContainerStatus` objects would cover both plus the fallback. Or, cover it with more comprehensive coverage of the cli command directly would work too.

jedcunningham (changes requested):

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/kubernetes_command.py
noamst-monday: Thanks! Appreciate your review, I'll take a look and update.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/cli/kubernetes_command.py
```diff
@@ -173,15 +200,18 @@ def cleanup_pods(args):
...
-                    > timedelta(minutes=min_pending_minutes)
-                )
+            )
+            is_terminal_old_enough = is_terminal and (
+                min_completed_minutes == 0
+                or current_time - _get_pod_completion_time(pod) > timedelta(minutes=min_completed_minutes)
+            )
+            if is_terminal_old_enough or (
+                pod_phase == pod_pending
+                and current_time - pod.metadata.creation_timestamp > timedelta(minutes=min_pending_minutes)
```
jedcunningham, line 214: Lets pull this into a separate flag too while we are here.

noamst-monday: Hope I understood your intention, if not let me know

jedcunningham: Exactly, ty!

jedcunningham (approved):

eladkal (approved):

Issue #70587, opened 2026-07-28
Get provider registry build off the docs publish critical path


shahar1:
## Human Summary

Since the providers registry build was introduced into the docs release process, the release
process for the providers' docs became longer by about 20-45 minutes (median ~23).
The big issue is that this process is composed of sequential runs of different GitHub Actions
workflows in different repositories (airflow, airflow-site, and airflow-site-archive), and this
process shouldn't be interrupted. More run time means:
1. More time spent on watching the entire process.
2. More chance that things will go wrong in the critical path, and we might need to do things all
   over again.

In my opinion, the provider registry is not as critical a component as the docs - which means that
it would be better separated into another workflow that will be initiated separately.

related: #50648, #45641, #44373, #32491, #50570, #66436

<details>
<summary>Detailed investigation - measurements, code references and two open questions (AI-generated)</summary>

During a provider release wave the release manager runs `breeze workflow-run publish-docs` and has
to keep the terminal open and the machine awake until it returns — the whole chain is driven from
the release manager's laptop, hence the warning at
[`dev/README_RELEASE_PROVIDERS.md:1769-1772`](https://github.com/apache/airflow/blob/379d4ac118071af99e62278fcaf0b1ba5b4f2c45/dev/README_RELEASE_PROVIDERS.md#L1769-L1772):

> Make sure to NOT close the terminal while the command is running and to keep your computer from
> sleep. This command will run several (3!) workflows from your terminal and this is important to
> keep it running until completion.

On the 2026-07-22 wave the `publish-docs-to-s3.yml` run alone took 90 minutes, and the docs had
been live on S3 since minute 47. The rest was the provider registry build, which the release
manager gains nothing by waiting for. Two questions below about getting it off that path.

#### Measured

Per-job timings, three recent wave runs:

| Job | [30335779728](https://github.com/apache/airflow/actions/runs/30335779728) | [30042123097](https://github.com/apache/airflow/actions/runs/30042123097) | [29107680148](https://github.com/apache/airflow/actions/runs/29107680148) |
|---|---|---|---|
| Build Info | 0.7m | 0.7m | 0.7m |
| Build documentation | 41.0m | 34.2m | 41.8m |
| Publish documentation to S3 | 4.3m | 4.7m | 4.5m |
| — *docs are live from here* — | | | |
| Registry: Build CI image | **32.5m** | **13.3m** | **31.8m** |
| Registry: Build & Publish | 10.5m | 9.3m | 9.7m |
[...]

PR #70618, opened 2026-07-28
Build the docs CI image once and let the registry share it

potiuk:
Addresses Question 1 of #70587, and takes Question 2's parallel option.

A provider docs publish built **two** CI images from the same sources — the docs job inline at `publish-docs-to-s3.yml:301-318` (python3.10) and the registry's own `build-ci-image` job (python3.12) — and only started the registry once the docs were already live on S3. Per #70587 the registry sub-workflow medians 23m, of which ~13m is that second image build, and it is 100% additive because it is strictly serial after the publish.

### What changes

* **One image per run.** A `build-ci-image` job in `publish-docs-to-s3.yml` builds from `inputs.ref` and stashes it; `build-docs` restores it instead of building inline, and `update-registry` passes `ci-image-already-built: true` so `registry-build.yml` skips its own build. A standalone registry dispatch has no such image and still builds one.
* **Both on one Python.** `default-python-version` moves 3.10 → 3.12, which is what the registry has always extracted on. This removes the second image that differed only by interpreter. Note `dev/registry/extract_metadata.py:47` and `extract_versions.py:49` fall back to `tomli` below 3.11 — that path is `# pragma: no cover`, so 3.12 is the better-exercised side.
* **Docs and registry run side by side.** `update-registry` moves from `needs: [publish-docs-to-s3, build-info]` to `needs: [build-info, build-ci-image]`. The registry reads nothing the docs publish produces.
* **The no-cache fallback is preserved**, moved into the build step it belongs to. `ci-image-build.yml` gains a `retry-without-cache` input (**default `false`**, so every other caller is unaffected); when the cached build fails it retries once with `DOCKER_CACHE=disabled`. This keeps the behaviour of the inline `breeze ci-image build || docker buildx build --pull` it replaces, for old refs whose `pyproject.toml` has drifted from main.
* **Mount cache retained.** Both image builds now set `upload-mount-cache-artifact: "true"`, so the next publish on the branch restores it rather than re-downloading the dependency set.
* **Cache source unchanged** — `docker-cache: registry` against `branch: main`, the cache the regular Test workflow keeps warm.

Resulting graph:

```
build-info -> build-ci-image -> { build-docs, update-registry }
```

### Consequences worth reviewing

1. **The failed-docs gate is gone.** `needs: [publish-docs-to-s3]` previously *gated* the registry, so a failed docs build skipped it. Running in parallel removes that: the registry can publish `/stable/` links to docs that never landed. #70587 measures the exposure window at median ~3 min, up to ~21 min, negative on 5 of 11 sampled runs. If that is not acceptable, the gate should move onto the S3 sync step rather than back onto the whole job.
2. **The registry publish job re-asserts the committer allowlist.** It previously inherited it by being `needs:` a job the allowlist skipped; now that job is also skipped on the shared-image path, so the check is stated explicitly on `build-and-publish-registry`.
3. **3.10 → 3.12 is exercised by the registry but new for the docs build.** Nothing in the docs path looked version-specific, but this is the change least verifiable without a real release run.

Not done here, and worth a separate look: restoring the canary image instead of building at all. It does not work through the stash — the Tests (AMD) run stashes only the default Python (`ci-image-save-v3-linux_amd64-3.10-…`), the stash name is branch-suffixed with no input to select a branch (current + base/default only), and artifacts expire in ~2 days. The cross-run-viable route would be pulling `ghcr.io/apache/airflow/<branch>/ci/python<ver>:latest`, which `push-image-cache.yml` publishes for every canary Python.

---

### DIFF
--- a/.github/workflows/publish-docs-to-s3.yml
+++ b/.github/workflows/publish-docs-to-s3.yml
@@ -254,39 +254,13 @@ jobs:
       upgrade-to-newer-dependencies: "false"
       docker-cache: "registry"
       disable-airflow-repo-cache: "false"
-
-  # Building against main's registry cache can fail outright when the ref's pyproject.toml has
-  # diverged from main — the case the docs job's inline `breeze ci-image build || docker buildx
-  # build` fallback used to cover. Retry once with the cache off, which is slow but always
-  # correct. Skipped whenever the cached build worked, so a fresh cache costs nothing.
-  build-ci-image-no-cache:
-    name: "Build CI image (no cache)"
-    needs: [build-info, build-ci-image]
-    if: always() && needs.build-ci-image.result == 'failure'
-    uses: ./.github/workflows/ci-image-build.yml
-    permissions:
-      contents: read
-      packages: write
-    with:
-      runners: '["ubuntu-22.04"]'
-      platform: "linux/amd64"
-      checkout-ref: ${{ inputs.ref }}
-      push-image: "false"
-      upload-image-artifact: "true"
-      upload-mount-cache-artifact: "true"
-      python-versions: ${{ format('["{0}"]', needs.build-info.outputs.default-python-version) }}
-      branch: "main"
-      constraints-branch: "constraints-main"
-      use-uv: "true"
-      upgrade-to-newer-dependencies: "false"
-      docker-cache: "disabled"
-      disable-airflow-repo-cache: "false"
+      # Building against main's cache can fail outright when the ref's pyproject.toml has
+      # diverged from main — the case the docs job's inline `breeze ci-image build || docker
+      # buildx build` fallback used to cover.
+      retry-without-cache: "true"
 
   build-docs:
-    needs: [build-info, build-ci-image, build-ci-image-no-cache]
-    if: >
-      !cancelled() &&
-      (needs.build-ci-image.result == 'success' || needs.build-ci-image-no-cache.result == 'success')
+    needs: [build-info, build-ci-image]
     timeout-minutes: 150
     name: "Build documentation"
     runs-on: ubuntu-latest
@@ -658,12 +632,8 @@ jobs:
   update-registry:
     # Runs alongside the docs build rather than after it: the registry reads nothing the docs
     # publish produces, so waiting only added its whole duration to the release manager's wait.
-    needs: [build-info, build-ci-image, build-ci-image-no-cache]
-    if: >
-      !cancelled() &&
-      (needs.build-ci-image.result == 'success' || needs.build-ci-image-no-cache.result == 'success') &&
-      (needs.build-info.outputs.registry-providers != '' ||
-      needs.build-info.outputs.registry-full-build == 'true')
+    needs: [build-info, build-ci-image]
+    if: needs.build-info.outputs.registry-providers != '' || needs.build-info.outputs.registry-full-build == 'true'
     name: "Update Provider Registry"
     permissions:
       contents: read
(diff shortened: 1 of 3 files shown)


shahar1 (approved):

.github/workflows/publish-docs-to-s3.yml
```diff
@@ -606,7 +630,9 @@ jobs:
             --destination-location "${SCHEMAS_DESTINATION}" "${args[@]}"
 
   update-registry:
-    needs: [publish-docs-to-s3, build-info]
+    # Runs alongside the docs build rather than after it: the registry reads nothing the docs
+    # publish produces, so waiting only added its whole duration to the release manager's wait.
+    needs: [build-info, build-ci-image]
```
kaxil, line 635: On the gate question: the sync steps look like the cheap place for it. Measured on run 30335779728, everything the registry job does after the image is 4m30s, and only 1m21s of that is `Sync registry to S3` plus `Publish version metadata` (which is host-side, no CI image needed). The site already goes up as the `registry-site` artifact, so those two steps could move into a job that needs `publish-docs-to-s3`, and the docs branch (~27m of Sphinx plus 4m publish) is long enough that the registry would be idle waiting for it anyway, so the wall clock does not change.

That also retires the #70587 caveat that the registry becomes the critical path on 5 of 11 runs. That was only true while it built its own image.

.github/workflows/publish-docs-to-s3.yml
```diff
@@ -226,8 +228,39 @@ jobs:
...
+      contents: read
+      packages: write
+    with:
+      runners: '["ubuntu-22.04"]'
+      platform: "linux/amd64"
+      # Built from the docs ref so the image matches the sources being documented, and
+      # cached against main's registry cache, which the regular Test workflow keeps warm.
+      checkout-ref: ${{ inputs.ref }}
+      push-image: "false"
+      upload-image-artifact: "true"
```
kaxil, line 245: Splitting the build into its own job means the docs path now pays an image round-trip it did not before. On run 30335779728, `Export CI docker image` plus `Stash CI docker image` was 4m26s, the consumer side (`Prepare breeze & CI image` in the registry job, minus the breeze install and disk cleanup it shares) about 5m, and the new job repeats the checkout / free-disk / breeze preamble for another ~1m50s. For a provider wave that is a clear net win against a 13-32m second build, but `build-ci-image` has no `if:`, so an `apache-airflow`-only publish skips `update-registry` and pays the whole ~11m for a stash that only `build-docs` reads. Per #70587 that cohort medians 21m, so it is a large relative hit. Worth gating the job on `registry-providers != '' || registry-full-build == 'true'` and leaving the inline build in `build-docs` for that case?

.github/workflows/registry-build.yml
```diff
@@ -97,6 +116,27 @@ jobs:
     timeout-minutes: 45
     name: "Build & Publish Registry"
     needs: [build-ci-image]
+    # `build-ci-image` is skipped when the caller stashed the image, so this cannot simply
+    # inherit its result — but it must keep enforcing the same committer allowlist that
+    # skipping that job used to enforce for us on a dispatch.
+    if: >
+      !cancelled() && needs.build-ci-image.result != 'failure' && (
+      github.event_name == 'workflow_call' ||
```
kaxil, line 124: `github.event_name` in a called workflow is the caller's event (`workflow_dispatch` here), never `workflow_call`, so this disjunct never fires and the job ends up gated on `github.event.sender.login` on both paths. Same net behaviour as before, since the old `needs: [build-ci-image]` inherited the identical check, but it does mean this allowlist is load-bearing for the shared-image path and it has drifted from the one on `build-info` in `publish-docs-to-s3.yml`, which also lists `vatsrahul1001`. A wave dispatched by them would publish docs and silently skip the registry.

PR #70619, opened 2026-07-28
Stop recommending psycopg3 connection strings on older Airflow

shahar1:
## Human Summary
This PR fixes psycopg connection-string guidance for older Airflow versions than 3.2.
It replaces psycopg3 connection strings with psycopg2 for compatibility with those versions.

## AI Summary
<details>
<summary>Click here</summary>
The amazon executor setup guides and the Celery `result_backend` example tell users to configure a `postgresql+psycopg://` (psycopg3) connection string. Both providers declare `apache-airflow>=2.11.0`, and releases before Airflow 3.2.0 (the first with a `sqlalchemy>=2.0` floor) can run on SQLAlchemy 1.4, which has no `postgresql+psycopg` dialect at all — a user following those instructions gets `sqlalchemy.exc.NoSuchModuleError` at startup, with nothing in the docs indicating the advice was version-specific.

The amazon guides are step-by-step deployment instructions where the connection string is pasted straight into an AWS console field, so the wrong value is especially easy to pick up.

This switches the documented value to `postgresql+psycopg2://`, which is valid on every Airflow version these providers support, and moves the version guidance into prose so the code block stays copy-pasteable. The three amazon executor guides had the same block duplicated inline; they now include a single shared block from `general.rst`, matching how the rest of those files already share content. The celery `result_backend` note also scopes the psycopg3 default to what actually governs it: the derivation applies to a driverless `postgresql://` URL when SQLAlchemy 2.0 and psycopg are both installed.

Verified with `breeze build-docs amazon celery` (build and spellcheck) — both packages pass and all affected pages render correctly.
</details>

---


uranusjr:
If this is for older Airflow, should this be filed to versioned branches instead of main?

potiuk:
> If this is for older Airflow, should this be filed to versioned branches instead of main?

It's not - providers currently (cc: @eladkal) support Airlfow 2.11+ from main - until we make a decision they don't

eladkal:
If specific providers will benfit from bumping min airflow version lets talk.

potiuk:
> If specific providers will benfit from bumping min airflow version lets talk.

I think that's about 100 conversation fior you to have, But yeah, I am interesting to see how many extra conversations and failures to handle it will bring.  👀

providers/amazon/docs/executors/general.rst
```diff
@@ -312,6 +312,24 @@ You will be prompted to enter the password if the connection is successful.
...
+   postgresql+psycopg2://<username>:<password>@<endpoint>/<database_name>
+
+The ``psycopg2`` driver above is valid on every Airflow version these executors support. From Airflow
+3.4.0, a driverless URL resolves to ``psycopg`` (psycopg3) where psycopg3 is installed and to
+``psycopg2`` otherwise. You may instead
+give an explicit ``postgresql+psycopg://`` URL from Airflow 3.2.0, provided psycopg3 is installed —
+whether it is depends on your ``apache-airflow-providers-postgres`` version. Do not use
+``postgresql+psycopg://`` before Airflow 3.2.0 at all: those releases can run on SQLAlchemy 1.4, which
+has no ``postgresql+psycopg`` dialect, so Airflow fails to start with
+``sqlalchemy.exc.NoSuchModuleError``.
```
Dev-iL: Rephrased suggestion, feel free to adjust.

```suggestion
The ``psycopg2`` driver is valid on every Airflow version these executors support. 

Support for the ``psycopg`` (v3) driver was added in Airflow 3.2.0. To use this driver, ensure it is installed, then specify ``postgresql+psycopg://``. Do not use ``postgresql+psycopg://`` before Airflow 3.2.0 at all: those releases typically run on SQLAlchemy 1.4, which has no ``postgresql+psycopg`` dialect, so Airflow will fail to start with ``sqlalchemy.exc.NoSuchModuleError``.

A driverless ``postgresql://`` URL resolves to ``psycopg2`` before Airflow 3.4.0, and to ``psycopg`` after.
```

providers/amazon/docs/executors/general.rst
```diff
@@ -312,6 +312,21 @@ You will be prompted to enter the password if the connection is successful.
...
 
+.. BEGIN SQL_ALCHEMY_CONN
+
+.. code-block:: bash
+
+   postgresql+psycopg2://<username>:<password>@<endpoint>/<database_name>
+
+.. note::
+   ``psycopg2`` works on every Airflow version these executors support. From Airflow 3.2.0 you may use
+   ``postgresql+psycopg://`` instead, provided psycopg3 is installed. Airflow before 3.2.0 does not
```
Dev-iL:
```suggestion
   ``postgresql+psycopg://`` instead, provided ``psycopg`` (v3) is installed. Airflow before 3.2.0 does not
```

I think it's better not to use `psycopg3` because this is not what users need to pip-install (the pypi entry looks like a placeholder last updated in 2013).

~~I think we can only mention "(v3)" the first time psycopg is mentioned in the document, no need to add that everywhere.~~ Disregard, I thought the repetitions were in the same file, but they aren't.

Dev-iL (approved):
Small suggestion, otherwise lgtm

shahar1:
> If specific providers will benfit from bumping min airflow version lets talk.

I don't think that specific providers will benefit from that -  we'll have to bump to 3.2+ and handle dropping 2.11-3.1 just for a doc example.

henry3260 (approved):

Issue #70548, opened 2026-07-27
Batch and memoize upstream map-length lookups for mapped stub arg bindings


jason810496:
### Background

Since https://github.com/apache/airflow/pull/69757, ti_run derives per-map-index arg bindings for mapped `@task.stub` tasks via `SchedulerDictOfListsExpandInput.resolve_expansion_sub_indexes` → `_get_map_lengths` (`airflow-core/src/airflow/models/expandinput.py`). `_get_map_lengths` carries a pre-existing TODO:

```python
# TODO: This initiates one database call for each XComArg. Would it be
# more efficient to do one single db call and unpack the value here?
```

Upstream map lengths are immutable once populated for a DagRun, yet every ti_run of every map index re-queries them: a stub expanded over K XComArg kwargs producing N map indexes issues N×K identical `get_task_map_length` queries per DagRun, on the hot ti_run route. (Stubs expanding over a single kwarg short-circuit and skip the lookups; literal kwargs cost `len()` only.)

### What needs to happen

1. Batch the per-kwarg length lookups into a single query (addressing the pre-existing `_get_map_lengths` TODO).
2. Consider memoizing resolved lengths per `(dag_id, run_id, task_id)` so repeated ti_run calls for sibling map indexes stop re-querying.

### Acceptance criteria

- One ti_run of a mapped stub issues at most one map-length query.
- Existing sub-index decomposition tests (`airflow-core/tests/unit/api_fastapi/execution_api/versions/head/test_task_instances.py`) stay green.

### Context

- Originating PR: https://github.com/apache/airflow/pull/69757 (review finding)
- Caller: `airflow-core/src/airflow/api_fastapi/execution_api/services/task_instances.py` (`_resolve_mapped_stub_arg_bindings`)

PR #70620, opened 2026-07-28
Batch upstream map-length lookups when resolving mapped kwargs

rjgoyln:
## Summary

Resolving mapped task kwargs previously issued one database query per `XComArg`, causing N+1 queries for expanded kwargs (especially with nested `zip()`/`concat()`). This change replaces those repeated lookups with a bulk prefetch, addressing the long-standing TODO in `SchedulerDictOfListsExpandInput._get_map_lengths`.

Upstream map lengths are now resolved in a fixed number of queries, improving both scheduler expansion and execution API argument binding.

## Changes

- Add `prefetch_map_lengths()` to resolve all referenced upstream map lengths in bulk.
- Add an optional `lengths` parameter to `get_task_map_length()` and use prefetched results when available.
- Prefetch once in `SchedulerDictOfListsExpandInput` and `SchedulerListOfDictsExpandInput`.
- Add tests covering query counts and behavior for mapped, unmapped, zipped, and unfinished upstreams.

## Testing

- Added query-count tests to verify the optimization.
- Existing scheduler and mapping tests pass.
- `mypy-airflow-core` and pre-commit checks pass.

* related: #70548
* related: #69757
-->

---

---

### DIFF
--- a/airflow-core/src/airflow/serialization/definitions/xcom_arg.py
+++ b/airflow-core/src/airflow/serialization/definitions/xcom_arg.py
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterable, Iterator, Mapping, Sequence
 from functools import singledispatch
 from typing import TYPE_CHECKING, Any
 
@@ -31,12 +31,21 @@ from airflow.serialization.definitions.notset import NOTSET, is_arg_set
 from airflow.utils.db import exists_query
 from airflow.utils.state import State
 
-__all__ = ["SchedulerXComArg", "deserialize_xcom_arg", "get_task_map_length"]
+__all__ = [
+    "SchedulerXComArg",
+    "deserialize_xcom_arg",
+    "get_task_map_length",
+    "prefetch_map_lengths",
+]
 
 if TYPE_CHECKING:
     from airflow.serialization.definitions.dag import SerializedDAG
     from airflow.serialization.definitions.mappedoperator import Operator
     from airflow.typing_compat import Self
+
+# Map length of each referenced task, keyed by ``(dag_id, task_id)``. A task whose
+# length is not known yet (upstream unfinished) is absent rather than mapped to None.
+MapLengths = Mapping[tuple[str, str], int]
 
 
 class SchedulerXComArg:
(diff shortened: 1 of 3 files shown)


uranusjr:
This only covers the batching part, not memoizing, so “closing” is wrong.

rjgoyln:
> This only covers the batching part, not memoizing, so “closing” is wrong.

OK. I got it. Thanks.

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,14 +154,102 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+    queries constant no matter how many arguments -- and how many tasks nested inside
+    ``zip()``/``concat()`` arguments -- have to be resolved.
+
+    Tasks whose length is not known yet are absent from the result, mirroring the
+    ``None`` that :func:`get_task_map_length` returns for them.
+    """
+    from airflow.models.taskinstance import TaskInstance
+    from airflow.models.taskmap import TaskMap
+    from airflow.models.xcom import XComModel
+    from airflow.serialization.definitions.mappedoperator import is_mapped
```
uranusjr: Do these need to be imported locally?

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,14 +154,102 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+    lengths: dict[tuple[str, str], int] = {}
+    if unmapped:
+        rows = session.execute(
+            select(TaskMap.dag_id, TaskMap.task_id, TaskMap.length).where(
+                TaskMap.run_id == run_id,
+                TaskMap.map_index < 0,
+                *_match_referenced_tasks(TaskMap, unmapped),
+            )
+        )
+        lengths.update({(dag_id, task_id): length for dag_id, task_id, length in rows})
```
uranusjr:
```suggestion
        lengths.update((dag_id, task_id): length for dag_id, task_id, length in rows)
```

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,14 +154,102 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+                    *_match_referenced_tasks(TaskInstance, mapped),
+                    # Special NULL treatment is needed because 'state' can be NULL.
+                    # The "IN" part would produce "NULL NOT IN ..." and eventually
+                    # "NULl = NULL", which is a big no-no in SQL.
+                    or_(
+                        TaskInstance.state.is_(None),
+                        TaskInstance.state.in_(s.value for s in State.unfinished if s is not None),
+                    ),
+                )
+                .distinct()
```
uranusjr, line 204: Do we need this distinct if the result is fed into set anyway? (Removing might be better from a performance perspective; I don’t know, this probably needs to be benchmarked.)

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,14 +154,102 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+                        XComModel.map_index >= 0,
+                        XComModel.key == XCOM_RETURN_KEY,
+                        *_match_referenced_tasks(XComModel, finished),
+                    )
+                    .group_by(XComModel.dag_id, XComModel.task_id)
+                )
+            }
+            # A finished mapped task that pushed nothing has no row to group, but its
+            # length is a known zero rather than an unresolved value.
+            lengths.update({key: counts.get(key, 0) for key in finished})
```
uranusjr:
```suggestion
            lengths.update((key, counts.get(key, 0)) for key in finished)
```

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,14 +154,102 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+    return lengths
+
+
 @singledispatch
-def get_task_map_length(xcom_arg: SchedulerXComArg, run_id: str, *, session: Session) -> int | None:
+def get_task_map_length(
+    xcom_arg: SchedulerXComArg, run_id: str, *, session: Session, lengths: MapLengths | None = None
```
uranusjr: Please always put `session` last.

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
rjgoyln: No — I checked and there is no cycle. I moved all four to the top, and dropped the same now-redundant local
imports from `get_task_map_length` also.

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
rjgoyln: Got it!

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
rjgoyln: Benchmarked it — `DISTINCT` should stay. `set()` deduplicates after every
row has crossed the connection and been materialized in Python; `DISTINCT`
does it in the database before the transfer. The duplication factor here is
the number of map indexes of the mapped upstream, so it is exactly the
workload this PR targets.

PostgreSQL 16, probing 3 mapped upstreams:

| map indexes | rows scanned | DISTINCT | plain   |
|------------:|-------------:|---------:|--------:|
| 1           | 3            | 0.451ms  | 0.441ms |
| 100         | 300          | 0.517ms  | 0.778ms |
| 1000        | 3,000        | 1.514ms  | 6.009ms |
| 5000        | 15,000       | 4.508ms  | 22.593ms |

SQLite shows the same shape (6.2x at 15k rows). Break-even is in the
single-digit-rows range, where the difference is ~0.01ms — noise. Above
that `DISTINCT` wins by 4-6x.

---

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
```diff
@@ -145,22 +157,103 @@ def iter_references(self) -> Iterator[tuple[Operator, str]]:
...
+def _match_referenced_tasks(model: Any, keys: Iterable[tuple[str, str]]) -> tuple[Any, Any]:
+    """
+    Build filters selecting rows of ``model`` for any of the given ``(dag_id, task_id)``.
+
+    Matching the two columns independently rather than as a row value keeps the query
+    portable. That is wider than the key set only if the keys span several Dags, which
+    the callers never do, and callers read results back by exact key regardless.
+    """
+    dag_ids, task_ids = zip(*keys)
+    return model.dag_id.in_(set(dag_ids)), model.task_id.in_(set(task_ids))
```
uranusjr:
```suggestion
    return model.dag_id.in_(dag_ids), model.task_id.in_(task_ids)
```

uranusjr:
Is it possible to further reduce the query with `UNION`? (SQLAlchemy `union_all`) I feel there’s a chance.

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
uranusjr: Actually, I think this is not strictly correct…? We should instead do something like

```python
tuple_(model.dag_id, mode.task_id).in_(keys))
```

airflow-core/tests/unit/serialization/definitions/test_xcom_arg.py
uranusjr, line 1: We should probably also have a query count test case for `expand_kwargs` (SchedulerListOfDictsExpandInput) as well.

rjgoyln:
> Is it possible to further reduce the query with `UNION`? (SQLAlchemy `union_all`) I feel there’s a chance.

Thanks for the suggestion. I benchmarked the `UNION ALL` approach.

**Advantages**
- Improves performance by about **35%** (~0.8 ms) on PostgreSQL 16.

**Disadvantages**
- Removes short-circuiting, requiring the database to compute XCom counts for all upstreams and Python to discard unfinished results afterward.
- Makes the query and unpacking logic significantly more complex.

I have the prototype ready. Let me know if you think the 0.8ms speedup is worth the added complexity, and I can push it!

airflow-core/tests/unit/serialization/definitions/test_xcom_arg.py
rjgoyln: Got it. I have Added the `expand_kwargs` query-count case.

airflow-core/src/airflow/serialization/definitions/xcom_arg.py
rjgoyln: Done. Switched to tuple_(...).in_(...), removed the helper, and passed sorted(...) to ensure deterministic SQL generation.

PR #70622, opened 2026-07-28
Fix malformed GitDagBundle error message for missing repository path

WonYong-Jang:
The error message shown when the repository path is missing was broken.
In `GitDagBundle._clone_repo_if_required`, the code was:

```
raise AirflowException("Repository path: %s not found", self.bare_repo_path)
```

This looks like it should put the path where `%s` is, but `AirflowException` does not do that.
When you pass the message and the value as two separate arguments, python just keeps both of them, so the error came out as:

```
 ('Repository path: %s not found', '/path/to/bare')
```

instead of a clear message.

The fix uses an f-string, which puts the path into the text directly. Now the message reads:

```
Repository path: /path/to/bare not found
```

There was also a test that checked for the old, broken text, so it was passing even though the code was wrong. The test now checks for the correct message.

 

---

- [x] Yes - Claude Code (Opus 4.8)

---


potiuk:
Good catch — this is a real bug. `AirflowException("...%s...", value)` doesn't interpolate; both arguments are kept, so the message rendered as a tuple rather than a path. The f-string is the right fix, and tightening the test to an exact-equality assertion against the real path is exactly what it needed.

One thing before this goes in. The project is actively reducing direct `AirflowException` usages, and the guidance is that when you touch a line that raises one, narrow it rather than leave it behind. Here the `except` clause tells you the specific failure — the path isn't there — so `FileNotFoundError` fits:

```python
except NoSuchPathError as e:
    # Protection should the bare repo be removed manually
    raise FileNotFoundError(f"Repository path: {self.bare_repo_path} not found") from e
```

The same method already does this a few lines above, so it stays consistent with its neighbours:

```python
raise RuntimeError("Error cloning repository") from e
raise RuntimeError(f"Invalid git repository at {self.bare_repo_path}") from e
```

Two follow-on edits if you take it:

- the test's `pytest.raises(AirflowException)` becomes `pytest.raises(FileNotFoundError)`
- `generated/known_airflow_exceptions.txt` has `providers/git/src/airflow/providers/git/bundles/git.py::5` — drop it to `::4`, otherwise the ratchet check fails

---

WonYong-Jang:
> Good catch — this is a real bug. `AirflowException("...%s...", value)` doesn't interpolate; both arguments are kept, so the message rendered as a tuple rather than a path. The f-string is the right fix, and tightening the test to an exact-equality assertion against the real path is exactly what it needed.
> 
> One thing before this goes in. The project is actively reducing direct `AirflowException` usages, and the guidance is that when you touch a line that raises one, narrow it rather than leave it behind. Here the `except` clause tells you the specific failure — the path isn't there — so `FileNotFoundError` fits:
> 
> ```python
> except NoSuchPathError as e:
>     # Protection should the bare repo be removed manually
>     raise FileNotFoundError(f"Repository path: {self.bare_repo_path} not found") from e
> ```
> 
> The same method already does this a few lines above, so it stays consistent with its neighbours:
> 
> ```python
> raise RuntimeError("Error cloning repository") from e
> raise RuntimeError(f"Invalid git repository at {self.bare_repo_path}") from e
> ```
> 
> Two follow-on edits if you take it:
> 
> * the test's `pytest.raises(AirflowException)` becomes `pytest.raises(FileNotFoundError)`
> * `generated/known_airflow_exceptions.txt` has `providers/git/src/airflow/providers/git/bundles/git.py::5` — drop it to `::4`, otherwise the ratchet check fails
> 
> Happy to merge once that's in.
> 

Thank you for the review! Applied all three changes.

Miretpl:
@WonYong-Jang could you resolve a conflicts?

WonYong-Jang:
> @WonYong-Jang could you resolve a conflicts?

@Miretpl @potiuk 
Resolved conflicts. Thanks

Miretpl (approved):

Issue #1032, opened 2016-02-18
Grayed out UI shown on all Pop-ups in the Webserver UI


r39132:
<img width="1085" alt="screenshot 2016-02-18 16 54 41" src="https://cloud.githubusercontent.com/assets/581734/13153346/8750e636-d66b-11e5-9233-f41316a53fb9.png">

I'm curious if others are seeing this!

mistercrunch:
Isn't this fixed in master and airbnb_prod?

r39132:
Hmn..unclean install!

PR #70623, opened 2026-07-28
Pin apache/infrastructure-actions to its released tags

### LINKED ISSUE #1007: Use Cassandra as backend to store task results (from Celery)
Using Cassandra as `celery_result_backend` doesn't work, notably because those config keys are not passed to Celery in `celery_executor.py` :

```
CASSANDRA_SERVERS
CASSANDRA_KEYSPACE
CASSANDRA_TABLE
CASSANDRA_READ_CONSISTENCY
CASSANDRA_WRITE_CONSISTENCY
CASSANDRA_ENTRY_TTL
```

Additionally, `CASSANDRA_SERVERS` is a python list of hosts, and `configuration.py` doesn't have a `getList` method.

I hacked this a little to make it work, but I still finally got, when starting a worker,

`[2016-02-15 17:19:09,071: WARNING/Worker-14] Cassandra error: InvalidRequestException(why="(String didn't validate.) [celery][tasks][traceback] failed validation"). Retrying...`

Any clues ?

potiuk:
[apache/infrastructure-actions#1032](https://github.com/apache/infrastructure-actions/pull/1032) added tag-based, per-action releases to that repository — each action is now tagged under its own prefix (`allowlist-check/vX.Y.Z`, `save/vX.Y.Z`, `restore/vX.Y.Z`, plus moving `…/v1` majors). Its recommended consumption is a SHA pin carrying the tag as a comment, which is what Dependabot needs to offer bumps.

Our pins predate that and have drifted apart in the meantime. The 28 uses sat on three different commits, none of them a release:

| commit | uses | how it got there |
|---|---|---|
| `b69225a8` | 18 | Dependabot walking the branch tip, most recently in #71302 |
| `0ff9972b` | 6 | added with `build_ci_image_with_cache` |
| `49df447b` | 4 | the original pin |

That spread is the symptom: with no version comment, Dependabot tracks whatever `main` happens to be, so pins move whenever a group update touches the file and stay put otherwise. `allowlist-check` carried a bare `# main` comment, saying so explicitly.

This moves all 28 to `61dcea11f19e2bbe1263f14d72235e8da17d3ad0`, the commit the `v1.0.0` (and `v1`) tags of all three actions point at, and names the tag in the comment:

```yaml
uses: apache/infrastructure-actions/stash/restore@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # restore/v1.0.0
uses: apache/infrastructure-actions/stash/save@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # save/v1.0.0
uses: apache/infrastructure-actions/allowlist-check@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # allowlist-check/v1.0.0
```

### Moving back to the tag is not a downgrade in behaviour

`b69225a8` is 61 commits ahead of `v1.0.0`, so this looks like it walks the pins backwards. It does not, for the code we actually run — over that range the only files touched under `stash/` are `pyproject.toml` and `uv.lock` (a ruff dev-dependency bump), and `allowlist-check/` is untouched. The other 59 commits are allowlist reviews and Dependabot updates elsewhere in the repository.

```
$ gh api repos/apache/infrastructure-actions/compare/61dcea11...b69225a8 \
      --jq '[.files[].filename] | map(select(startswith("stash/") or startswith("allowlist-check/")))'
["stash/pyproject.toml","stash/uv.lock"]
```

### The zizmor suppression goes away

`.github/zizmor.yml` suppressed `ref-version-mismatch` for `asf-allowlist-check.yml` on the grounds that the action is branch-tracked with no version tags. That has not been true since #1032, and a real tag resolves, so the suppression is deleted and that workflow is audited like every other one.

Verified rather than assumed — zizmor resolves the prefixed tags correctly. With the suppression removed:

```
$ zizmor .github/workflows/asf-allowlist-check.yml \
         .github/actions/prepare_single_ci_image/action.yml \
         .github/actions/prepare_breeze_and_image/action.yml
No findings to report. Good job. (2 suppressed)
```

and, as a control that the audit really was running rather than silently skipping, the same pin with a deliberately bogus tag is still caught:

```
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  - uses: apache/infrastructure-actions/stash/save@61dcea11…  # save/v9.9.9
                                                                ^^^^^^^^^^^ points to unknown ref
```

zizmor reads tags from the git refs API and peels annotated tags, so `save/v1.0.0` resolves to `61dcea11…` — exactly the pinned commit. Note this audit only runs in online mode; a local run without a token reports "No findings" for both the correct and the bogus pin.

### Line-length exemptions

Fifteen pins needed `# yamllint disable-line rule:line-length` — every `stash/restore` use plus `allowlist-check`. A 40-character SHA plus the action path already runs to ~95 columns, so a trailing `# restore/v1.0.0` crosses the 110 limit; `# save/v1.0.0` is short enough to fit. These lines passed before only because yamllint's `allow-non-breakable-words` exempts a line that is a single unbreakable token — adding the comment introduces a space and forfeits that. Keeping the comment on the same line is deliberate: Dependabot only reads it there.

### Unrelated, but noticed while verifying

zizmor picks its *suggested* tag with `max_by_key(|t| t.name.len())` over every tag at the commit, with no awareness of which action inside the repo is used. For this monorepo that means a `stash/save` pin is told it "is pointed to by tag `allowlist-check/v1.0.0`" — a sibling action's version — and that is offered as an auto-fix. Nothing in this PR depends on it, but do not accept that particular auto-fix if you see it.

---

### DIFF
--- a/.github/actions/build_ci_image_with_cache/action.yml
+++ b/.github/actions/build_ci_image_with_cache/action.yml
@@ -48,7 +48,8 @@ runs:
     # cache. The commit it was built from tells the two apart and is stashed on its own, so
     # deciding costs a few bytes rather than the image the decision may make unnecessary.
     - name: "Restore the commit the stashed CI image was built from"
-      uses: apache/infrastructure-actions/stash/restore@0ff9972b5872e19c9f4555c9159c2fea4f794355
+      # yamllint disable-line rule:line-length
+      uses: apache/infrastructure-actions/stash/restore@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # restore/v1.0.0
       with:
         key: "ci-image-commit-v3-${{ inputs.platform }}-${{ inputs.python }}\
           -${{ inputs.image-stash-ref }}"
@@ -73,7 +74,8 @@ runs:
       if: steps.restore-commit.outputs.stash-hit == 'true'
     # Restored ahead of the caches that feed the build, so a build made unnecessary skips them too.
     - name: "Restore the CI image stashed for this ref"
-      uses: apache/infrastructure-actions/stash/restore@0ff9972b5872e19c9f4555c9159c2fea4f794355
+      # yamllint disable-line rule:line-length
+      uses: apache/infrastructure-actions/stash/restore@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # restore/v1.0.0
       with:
         key: "ci-image-save-v3-${{ inputs.platform }}-${{ inputs.python }}\
           -${{ inputs.image-stash-ref }}"
@@ -108,7 +110,8 @@ runs:
     # Scoped to the ref for the same reason the image is: the mount cache holds the dependency set
     # the sources resolve to, and a ref's and the branch tip's are exactly what differ.
     - name: "Restore the mount cache stashed for this ref"
-      uses: apache/infrastructure-actions/stash/restore@0ff9972b5872e19c9f4555c9159c2fea4f794355
+      # yamllint disable-line rule:line-length
+      uses: apache/infrastructure-actions/stash/restore@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # restore/v1.0.0
       with:
         key: "ci-cache-mount-save-v3-${{ inputs.platform }}-${{ inputs.python }}\
           -${{ inputs.image-stash-ref }}"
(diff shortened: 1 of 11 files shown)


potiuk:
Rebased onto main. The conflicts were all the same shape — main had moved these pins to newer untagged commits, this change points them at the released tags instead, which is the whole point of the PR. I re-checked upstream and `61dcea11f19e2bbe1263f14d72235e8da17d3ad0` is still exactly `save/v1.0.0`, `restore/v1.0.0` and `allowlist-check/v1.0.0`, so the pins name real releases and Dependabot can offer bumps for them.

@vatsrahul1001 you review a lot of the CI workflow changes — would you mind a look?

---

gopidesupavan (approved):

PR #70628, opened 2026-07-28
Restore Dag-parse-time validation for common.ai operator arguments

baha-bouali:
Restores two `__init__` argument-provision checks that moved to `execute()` under the
old `validate-operators-init` rule, now that #70505 narrowed the rule to allow
`field is None` / `field is not None` provision checks back in `__init__`.

**`AgentOperator`** :  straight revert of #70338's `message_history`/`enable_hitl_review`
check back to `__init__`, with `test_agent.py::test_message_history_with_hitl_review_raises`
reverted to assert at construction time (and its unneeded `skipif` removed).

**`DocumentLoaderOperator`** : partial revert. `source_path`/`source_bytes` exclusivity is
pure provision and moves back to `__init__`. The `file_type` check stays in `execute()`,
since it backs an `assert` on the *rendered* value of `file_type` (itself a template
field). So, checking it in `__init__` would validate the unrendered template string instead.
Tests split to match: the two exclusivity tests revert to constructor-time, the two
`file_type` tests stay at `.execute()`.

## Testing

- `uv run --script scripts/ci/prek/validate_operators_init.py` on both files -> clean,
  no exemption entries needed under the narrowed rule.
- `pytest providers/common/ai/tests/unit/common/ai/operators/test_agent.py
  providers/common/ai/tests/unit/common/ai/operators/test_document_loader.py` -> passing.

related: #70503

---


providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
```diff
@@ -148,11 +152,10 @@ def __init__(
...
     def execute(self, context: Context) -> list[dict[str, Any]]:
-        # source_path/file_type are template fields; validate after rendering, not in __init__.
-        if self.source_path is not None and self.source_bytes is not None:
-            raise ValueError("Provide exactly one of 'source_path' or 'source_bytes', not both.")
-        if self.source_path is None and self.source_bytes is None:
-            raise ValueError("Provide exactly one of 'source_path' or 'source_bytes'.")
+        # source_path/source_bytes provision is checked in __init__ (that's just "was an
+        # argument passed"). file_type is different: it backs the assert below, and since
+        # file_type is itself a template field, whether it was *actually* supplied is only
+        # knowable after rendering -- so this one check has to stay here, not in __init__.
```
kaxil: The rationale here is inverted -- it repeats the claim #70505 refuted. Provision is *not* knowable after rendering: with `render_template_as_native_obj=True` a supplied field renders to `None`, which is precisely why the rule was narrowed to allow `is None` back in `__init__`.

What actually keeps this one in `execute()` is that it isn't a provision check at all, it's a value check: `_parse_bytes` needs a real extension string, so it has to see the rendered value. The PR body words it correctly ("checking it in `__init__` would validate the unrendered template string instead") -- worth pulling that wording into the comment, since #70503 makes this file a reference others will copy.

providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
```diff
@@ -148,11 +152,10 @@ def __init__(
...
         if self.source_bytes is not None and self.file_type is None:
             raise ValueError("'file_type' is required when using 'source_bytes' (e.g. '.pdf').")
```
kaxil, line 161: `file_type` keeps a rendered-value guard here, but `source_path` no longer has one, and the `assert self.source_path is not None` below was leaning on the check that just moved to `__init__`.

A supplied `source_path` that renders to `None` (the native-obj case #70505 is built around) now falls into the `else` branch, and `_resolve_files(None)` raises `TypeError: argument of type 'NoneType' is not iterable` instead of the previous clear `ValueError`. Worth keeping a `self.source_path is None` guard here next to the `file_type` one.

baha-bouali: I think it is fixed now. I reworded the comment and added back the source_path guard, thanks for catching this! Let me know if something is still wrong

providers/common/ai/src/airflow/providers/common/ai/operators/document_loader.py
```diff
@@ -148,13 +152,15 @@ def __init__(
...
-            raise ValueError("Provide exactly one of 'source_path' or 'source_bytes'.")
+        # file_type and source_path can each be *supplied* (as non-None argument) yet still
+        # render to None. These aren't provision checks (that already happened in __init__); 
+        # they guard the rendered value itself, since _parse_bytes/_resolve_files need a real 
+        # value to work with. Checking this in __init__ would validate the unrendered template 
+        # string instead of the value actually used here.
         if self.source_bytes is not None and self.file_type is None:
             raise ValueError("'file_type' is required when using 'source_bytes' (e.g. '.pdf').")
+        if self.source_bytes is None and self.source_path is None:
+            raise ValueError("Provide exactly one of 'source_path' or 'source_bytes'.")
```
shahar1, line 163: Could you please add a unit test for this case?

baha-bouali: Done! I added the unit test right in `test_document_loader.py`.

baha-bouali:
Hey @shahar1 just gentle ping. Could you take another look when you have a chance? I’ve addressed all the comments and the static checks are finally passing.Thanks!

baha-bouali:
Hey @kaxil, could you please take a look when you have a chance? Thank you for your time.

baha-bouali:
cc @potiuk

shahar1 (approved):

shahar1:
> Hey @shahar1 just gentle ping. Could you take another look when you have a chance? I’ve addressed all the comments and the static checks are finally passing.Thanks!

Sorry for the delay, was a on break from the open-source.
Overall LGTM, good work!

PR #66696, opened 2026-05-11
fix(api): replace COALESCE with index-friendly OR conditions in datetime range filters

hkc-8010:
## Summary

- \`datetime_range_filter_factory\` wrapped \`start_date\` and \`end_date\` columns in
  \`COALESCE(column, now())\` to treat currently-running tasks (NULL date) as "now".
  This form prevents PostgreSQL from using btree indexes, causing parallel sequential
  scans across the full \`task_instance\` table even for narrow date windows.
- Replace with explicit OR conditions that are semantically equivalent:
  \`(col >= X) OR (col IS NULL AND now() >= X)\`
  PostgreSQL can use a btree index on each OR branch via BitmapOr. Running tasks
  (NULL \`end_date\` / \`start_date\`) continue to match date-range queries correctly.
- Adds \`NullableDatetimeRangeFilter\` subclass of \`RangeFilter\` for the new logic.
  \`datetime_range_filter_factory\` returns this subclass for \`start_date\` / \`end_date\`
  and a plain \`RangeFilter\` for all other filter names.

## Performance

Measured against a 4.6M-row \`task_instance\` table with an \`end_date\` btree index:

| | Plan | Cost (first 1000 rows) |
|---|---|---|
| Before | Parallel Index Scan + Gather Merge (COALESCE wraps column) | 19,523 |
| After | Single Index Scan, stops early via LIMIT (OR branches) | 2,658 |

**7.4x reduction in query cost.**

Note: vanilla Airflow does not ship an \`end_date\` index by default. The OR form is
the correct architectural fix regardless. Any deployment that adds an \`end_date\` index
(via a migration) will immediately benefit. A follow-up migration to add the index is
recommended but is outside the scope of this PR.

## Changes

- \`airflow-core/src/airflow/api_fastapi/common/parameters.py\`: add
  \`NullableDatetimeRangeFilter\`; update \`datetime_range_filter_factory\`
- \`airflow-core/tests/unit/api_fastapi/common/test_parameters.py\`: add
  \`TestDatetimeRangeFilterFactory\` (7 tests covering type dispatch, SQL shape,
  no-COALESCE assertion, and NULL-branch presence)

## PR Checklist

- [x] My PR is targeted at the \`main\` branch
- [x] Tests added (31 pass locally and in Breeze core-tests, Python 3.10 / SQLite)
- [x] \`prek\` hooks all pass (ruff, ruff-format, mypy)
- [x] No newsfragment needed (internal performance fix, no user-visible behavior change)

closes: #66335

PR #70629, opened 2026-07-28
fix(api): scope NULL branch in NullableDatetimeRangeFilter for start_date lower bounds [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #70627: dagRuns API: start_date_gte filter matches terminal (failed/success) runs with a NULL start_date regardless of age

Airflow Core

### Apache Airflow version

3.2.2

### What happened

`GET /dags/~/dagRuns?start_date_gte=<recent-timestamp>` returns dag runs whose `start_date` is `NULL`, no matter how old the run's `logical_date`/`end_date` is — even when the run is in a terminal state (`failed`, `success`).

Found while querying recent dag runs against `/dags/~/dagRuns` with `start_date_gte` set to a recent cutoff: the response included dag runs from **2021–2024** (`state: failed`, `start_date: null`, populated `end_date`) as if they had just occurred, alongside genuinely recent runs. Roughly 99% of a `-logical_date`-ordered page were these stale null-`start_date` rows, drowning out real recent activity.

### Root cause

`NullableDatetimeRangeFilter.to_orm` (added in #66696 to fix #66335's COALESCE index-scan performance issue) builds the lower-bound predicate as:

```python
select = select.where(or_(self.attribute >= x, self.attribute.is_(None)))
```

The docstring justifies this as: "For lower bounds the NULL branch passes unconditionally — a not-yet-started/ended task will eventually satisfy any past lower bound." That assumption only holds while the row is *non-terminal* (`running`/`queued`, genuinely pending a future start). It's false for a dag run that is already `failed`/`success` with a persisted `NULL` `start_date` — that row will never "eventually" start, yet it now matches every `start_date_gte` filter forever, regardless of the query's cutoff.

### How to reproduce

1. Have a dag run in a terminal state (`failed`/`success`) whose `start_date` is `NULL` (e.g. an old manual trigger that failed before the scheduler set `start_date`).
2. Query `/dags/~/dagRuns?start_date_gte=<any-recent-timestamp>&order_by=-logical_date`.
3. Observe the terminal, null-`start_date` row is included in the response regardless of how far in the past its `logical_date`/`end_date` actually is.

The NULL-passes-unconditionally branch should be scoped to non-terminal states (or the filter should additionally require `end_date IS NULL` alongside `start_date IS NULL`, since a genuinely-pending run shouldn't have an `end_date` either). A terminal dag run with a null `start_date` should not satisfy a `start_date_gte` lower bound at all — it never had a `start_date` within the window, or any window.

### Operating System

distroless image base

### Deployment

_No response_

N/A

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

Not Applicable

Not Applicable (to this issue)

### Anything else?

This is the opposite failure mode from #66047 (UI not passing the filter param at all) and a different bug from #66335 (COALESCE hurting index usage) — #66696, which fixed #66335, introduced this NULL-passthrough regression as a side effect and marked itself "no user-visible behavior change."

### Code of Conduct

waterWang:
## Description

`NullableDatetimeRangeFilter` was added in #66696 to replace `COALESCE(column, now())` with index-friendly `OR` predicates. For lower bounds (`gte`/`gt`), the NULL branch passes unconditionally:

```python
select = select.where(or_(self.attribute >= x, self.attribute.is_(None)))
```

This is correct for **non-terminal** rows (a not-yet-started task will eventually start). But it's wrong for **terminal** dag runs whose `start_date` is `NULL` — a failed/success run that never started will never have a `start_date`, yet it matches every `start_date_gte` filter forever.

### Root cause

When querying `GET /dags/~/dagRuns?start_date_gte=<recent-timestamp>`, the response includes terminal dag runs from **years ago** (`state: failed`, `start_date: null`, populated `end_date`) as if they had just occurred. Roughly 99% of a `-logical_date`-ordered page are these stale null-`start_date` rows.

### Fix

Add an `extra_null_condition` parameter to `NullableDatetimeRangeFilter`. For `start_date` lower bounds, the factory passes `model.end_date.is_(None)`, so the NULL branch becomes:

```python
or_(start_date >= x, and_(start_date IS NULL, end_date IS NULL))
```

A genuinely pending run has `start_date IS NULL AND end_date IS NULL`. A terminal run with `NULL start_date` has `end_date IS NOT NULL` and is correctly excluded.

### Changes

- `parameters.py`: Add `extra_null_condition` to `NullableDatetimeRangeFilter`; update `datetime_range_filter_factory` to pass it for `start_date`
- `test_parameters.py`: Update existing lower-bound test to verify `end_date IS NULL`; add test for terminal-run exclusion

### Related issues

Fixes #70627

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -1197,7 +1197,15 @@ class NullableDatetimeRangeFilter(RangeFilter):
     started/ended task will eventually satisfy any past lower bound. For upper bounds the NULL branch
     is ``col IS NULL AND now() <= x``, preserving the COALESCE(col, now()) semantics without the
     function-wrap index penalty.
+
+    When *extra_null_condition* is provided (e.g. ``end_date.is_(None)`` for ``start_date`` lower
+    bounds), the NULL branch also requires that condition — preventing terminal rows whose column
+    happens to be ``NULL`` from matching every lower-bound filter indefinitely.
     """
+
+    def __init__(self, range_val, attr, extra_null_condition=None):
+        super().__init__(range_val, attr)
+        self.extra_null_condition = extra_null_condition
 
     def to_orm(self, select: Select) -> Select:
         if self.skip_none is False:
@@ -1208,10 +1216,16 @@ class NullableDatetimeRangeFilter(RangeFilter):
 
         if self.value.lower_bound_gte:
             x = self.value.lower_bound_gte
-            select = select.where(or_(self.attribute >= x, self.attribute.is_(None)))
+            null_condition = self.attribute.is_(None)
+            if self.extra_null_condition is not None:
+                null_condition = and_(null_condition, self.extra_null_condition)
+            select = select.where(or_(self.attribute >= x, null_condition))
         if self.value.lower_bound_gt:
             x = self.value.lower_bound_gt
-            select = select.where(or_(self.attribute > x, self.attribute.is_(None)))
+            null_condition = self.attribute.is_(None)
+            if self.extra_null_condition is not None:
+                null_condition = and_(null_condition, self.extra_null_condition)
+            select = select.where(or_(self.attribute > x, null_condition))
         if self.value.upper_bound_lte:
             x = self.value.upper_bound_lte
             select = select.where(or_(self.attribute <= x, and_(self.attribute.is_(None), func.now() <= x)))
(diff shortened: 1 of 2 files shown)


ColtenOuO:
The suffix of the PR title needs to be changed.

There have some garbled text.

waterWang:
Hi maintainers, just a friendly ping — this PR has been open for a while. Could you take a look when you have a moment? Happy to address any feedback. Thanks!

PR #70635, opened 2026-07-28
fix(amazon): include deferred-task trigger logs when reading from Clo…

### LINKED ISSUE #70317: Logs produced by deferrable tasks disapper from Airflow UI after completion of the task

Airflow Core

### Apache Airflow version

3.3.0

**Issue Description**

This issue is related to https://github.com/apache/airflow/issues/70314. Even though I was able to implement a workaround that sends logs produced by deferrable triggers to AWS CloudWatch, Airflow UI doesn't seem to be able to display them after completion of the task.

The way it works at the moment in 3.3.0 with remote logging configured to send logs to AWS CloudWatch is that Airflow UI pulls trigger logs for deferrable tasks from the triggerer machine via HTTP while the task is running in deferred state. However, once the task completes these logs disappear from Airflow UI.

**Issue Analysis**

I don't know much about internals of Airflow, but this is what my AI assistant and I were able to identify while debugging this problem (note that it assumes that trigger logs were successfully sent to AWS CloudWatch via the workaround listed in https://github.com/apache/airflow/issues/70314).

> When the UI fetches logs for a completed deferrable task, [`FileTaskHandler._read_remote_logs()`](https://github.com/apache/airflow/blob/3.3.0/airflow-core/src/airflow/utils/log/file_task_handler.py#L1002-L1005) calls `CloudWatchRemoteLogIO.stream(path, ti)` with the base task log path (e.g. `dag_id=.../attempt=1.log`). However, trigger logs are written to a separate CloudWatch stream with a [`.trigger.{job_id}.log` suffix](https://github.com/apache/airflow/blob/3.3.0/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L861), and [`CloudWatchRemoteLogIO.stream()`](https://github.com/apache/airflow/blob/3.3.0/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py#L199-L213) reads exactly one stream — the one it's given — with no discovery of related trigger streams. The local-file equivalent doesn't have this problem because [`_read_from_local()`](https://github.com/apache/airflow/blob/3.3.0/airflow-core/src/airflow/utils/log/file_task_handler.py#L884) uses `glob(worker_log_path.name + "*")` which naturally picks up both the base log and any `.trigger.*.log` files. CloudWatch has no glob, so `stream()` needs an explicit `DescribeLogStreams` prefix scan to find trigger streams.

**Steps to reproduce**

I've created a docker compose simulation that can be used to reproduce the problem.

**Prerequisites:** the simulation expects one to have an AWS account with privileges needed to write logs to a CloudWatch group.

**File: ./Dockerfile**
```
FROM apache/airflow:3.3.0-python3.13

USER airflow

RUN pip install --no-cache-dir \
    apache-airflow-providers-amazon==9.31.0 \
    apache-airflow-providers-standard==1.16.0
```

**File: ./docker-compose.yml**
```
x-airflow-common:
  &airflow-common
  build: .
  environment:
    &airflow-common-env
    AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres:5432/airflow
    AIRFLOW__CORE__EXECUTOR: LocalExecutor
    AIRFLOW__CORE__FERNET_KEY: 'olFLXR4nU_6WBvIsSDPIDCKWjh3OCWSMfWjt6XU0VfM='
    AIRFLOW__CORE__EXECUTION_API_SERVER_URL: http://airflow-webserver:8080/execution/
    AIRFLOW__CORE__SIMPLE_AUTH_MANAGER_ALL_ADMINS: 'True'
    AIRFLOW__API__SECRET_KEY: reproduction-secret-key-not-for-production
    AIRFLOW__API_AUTH__JWT_SECRET: reproduction-jwt-secret-not-for-production
    AIRFLOW__API_AUTH__JWT_ALGORITHM: HS512
    AIRFLOW__LOGGING__REMOTE_LOGGING: 'True'
    AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER: 'cloudwatch://arn:aws:logs:${AWS_REGION}:${AWS_ACCOUNT_ID}:log-group:${AWS_LOG_GROUP_NAME}'
    AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID: aws_default
    AIRFLOW__LOGGING__DELETE_LOCAL_LOGS: 'True'
    AIRFLOW__LOGGING__LOGGING_LEVEL: INFO
    AWS_DEFAULT_REGION: ${AWS_REGION}
    AWS_REGION: ${AWS_REGION}
    AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID}
    AWS_SECRET_ACCESS_KEY: ${AWS_SECRET_ACCESS_KEY}
    AWS_SESSION_TOKEN: ${AWS_SESSION_TOKEN:-}
    PYTHONPATH: '/opt/airflow/config:/opt/airflow/dags'
    PYTHONUNBUFFERED: '1'
  volumes:
    - ./dags:/opt/airflow/dags
    - ./config:/opt/airflow/config
  depends_on:
    postgres:
      condition: service_healthy

services:
  postgres:
    image: postgres:16
    environment:
      POSTGRES_USER: airflow
      POSTGRES_PASSWORD: airflow
      POSTGRES_DB: airflow
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U airflow"]
      interval: 10s
      timeout: 5s
      retries: 5

  airflow-init:
    <<: *airflow-common
    entrypoint: /bin/bash
    command:
      - -c
      - |
        set -ex
        airflow db migrate
        airflow connections create-default-connections
        echo "Airflow init complete."
    environment:
      <<: *airflow-common-env
      TASK_NAME: init

  airflow-scheduler:
    <<: *airflow-common
    command: airflow scheduler
    environment:
      <<: *airflow-common-env
      TASK_NAME: scheduler
    depends_on:
      airflow-init:
        condition: service_completed_successfully

  airflow-webserver:
    <<: *airflow-common
    command: airflow api-server
    ports:
      - "8080:8080"
    environment:
      <<: *airflow-common-env
      TASK_NAME: webserver
    depends_on:
      airflow-init:
        condition: service_completed_successfully

  airflow-dag-processor:
    <<: *airflow-common
    command: airflow dag-processor
    environment:
      <<: *airflow-common-env
      TASK_NAME: dag-processor
    depends_on:
      airflow-init:
        condition: service_completed_successfully

  airflow-triggerer:
    <<: *airflow-common
    command: airflow triggerer
    environment:
      <<: *airflow-common-env
      TASK_NAME: triggerer
    depends_on:
      airflow-init:
        condition: service_completed_successfully
```

**File: ./dags/test_triggerer_logging.py**
```

from airflow.sdk import DAG
from airflow.providers.standard.operators.python import PythonOperator
from airflow.providers.standard.sensors.time_delta import TimeDeltaSensor


with DAG(
    dag_id="test_triggerer_logging",
    schedule=None,
    start_date=datetime(2020, 1, 1),
    catchup=False
) as dag:

    before = PythonOperator(
        task_id="before_deferral",
        python_callable=lambda: print("Pre-deferral task log."))

    # Logs produced by this task will be missing in CloudWatch
    test_triggerer = TimeDeltaSensor(
        task_id="test_triggerer",
        delta=timedelta(seconds=120),
        deferrable=True)

    after = PythonOperator(
        task_id="after_deferral",
        python_callable=lambda: print("Post-deferral task log."))

    before >> test_triggerer >> after
```

**File: ./config/airflow_local_settings.py**
```
# Note: workaround from #70314 to send trigger logs to AWS CloudWatch

import os


def _patch_dictconfig_for_watchtower():
    import logging
    import logging.config
    import watchtower

    _original_dictConfig = logging.config.dictConfig

    def _dictConfig_hide_watchtower(config):
        saved = []
        remaining = []
        for ref in logging._handlerList:
            h = ref()
            if h is not None and isinstance(h, watchtower.CloudWatchLogHandler):
                saved.append(ref)
            else:
                remaining.append(ref)

        logging._handlerList[:] = remaining
        try:
            _original_dictConfig(config)
        finally:
            logging._handlerList.extend(saved)

    logging.config.dictConfig = _dictConfig_hide_watchtower


if os.environ.get("TASK_NAME") == "triggerer":
    _patch_dictconfig_for_watchtower()
```

**File: ./.env**
```
# AWS region and account ID
AWS_REGION=change_me
AWS_ACCOUNT_ID=change_me

# AWS credentials — ECS injects these via task role; Docker needs them explicitly.
# Populate via: aws configure export-credentials --profile change_me --format env
AWS_ACCESS_KEY_ID=change_me
AWS_SECRET_ACCESS_KEY=change_me
AWS_SESSION_TOKEN=change_me
AWS_LOG_GROUP_NAME=change_me
```

Trigger execution of the `test_triggerer_logging` DAG, then look at the logs tab of the `test_triggerer` task. You'll see that logs produced by the trigger gradually appear over time in Airflow UI while the task is running. However, once execution of the task is completed these logs disappear from Airflow UI. Logs produced by non-deferrable tasks remain in place after completion of those tasks.

Checking AWS CloudWatch reveals that trigger logs have been preserved with correct stream name right next to logs produced by non-deferrable tasks. So the issue is that Airflow UI isn't pulling trigger logs from AWS CloudWatch when it is configured to use remote logging.

Logs produced by triggers in deferrable tasks should remain visible in Airflow UI after the deferrable task is complete.

### Operating System

Docker with apache/airflow:3.3.0-python3.13 image

### Deployment

Other

_No response_

apache-airflow-providers-amazon==9.31.0
apache-airflow-providers-standard==1.16.0

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

See the supplied `Dockerfile` and `docker-compose.yml` in the "how to reproduce it" section.

### Anything else?

I was able to find a workaround that works via monkey-patching. Even though it seems to work, it is pretty fragile as it can break at any point if Airflow decides to change something in future versions.

**Add** the following code to the `./config/airflow_local_settings.py` file and re-start Docker Compose:
```
def _patch_cloudwatch_stream_for_trigger_logs():
    from airflow.providers.amazon.aws.log.cloudwatch_task_handler import CloudWatchRemoteLogIO

    _original_stream = CloudWatchRemoteLogIO.stream

    def _stream_with_trigger_logs(self, relative_path, ti):
        sources, logs = _original_stream(self, relative_path, ti)

        if getattr(ti, "state", None) == TaskInstanceState.DEFERRED:
            return sources, logs

        try:
            trigger_prefix = relative_path.replace(":", "_") + ".trigger."
            response = self.hook.conn.describe_log_streams(
                logGroupName=self.log_group,
                logStreamNamePrefix=trigger_prefix,
            )
            for stream_info in response.get("logStreams", []):
                stream_name = stream_info["logStreamName"]
                trigger_sources, trigger_logs = _original_stream(self, stream_name, ti)
                sources.extend(trigger_sources)
                logs.extend(trigger_logs)
        except Exception:
            pass

        return sources, logs

    CloudWatchRemoteLogIO.stream = _stream_with_trigger_logs


if os.environ.get("TASK_NAME") == "webserver":
    _patch_cloudwatch_stream_for_trigger_logs()
```

The idea of this patch is to use `describe_log_streams` as a replacement for `glob()` to find all AWS CloudWatch streams that are related to the task. So that Airflow UI can display all of them, and not only those produced by non-deferrable tasks.

### Code of Conduct

bujjibabukatta:
Title

fix(amazon): include deferred-task trigger logs when reading from CloudWatch (#70317)

Body

Closes #70317.

Root cause

CloudWatchRemoteLogIO.stream() (and the legacy CloudwatchTaskHandler._read_remote_logs()) only ever read the task's own CloudWatch stream. A deferred task's triggerer logs are written to a separate stream, <task log stream>.trigger.<triggerer job id>.log (FileTaskHandler.add_triggerer_suffix). While the task is DEFERRED, the UI tails those logs live from the triggerer over HTTP, which is why they're visible during the deferral — but once the task finishes, that live path goes away and nothing ever looked in the separate CloudWatch stream. The local-file backend doesn't have this problem because FileTaskHandler._read_from_local() uses glob(worker_log_path.name + "*"), which picks up attempt=1.log and attempt=1.log.trigger.*.log for free. CloudWatch has no glob equivalent, so the trigger stream has to be discovered explicitly.

Fix
Added AwsLogsHook.describe_log_streams(): a synchronous, paginated wrapper around DescribeLogStreams with a prefix filter (mirrors the existing async variant's ResourceNotFoundException → [] handling for a not-yet-existing log group).
CloudWatchRemoteLogIO.stream() now looks up any <stream>.trigger.*.log streams via that prefix scan and reads each one the same way it reads the base stream, merging them into the returned log groups. Skipped while the task is actively DEFERRED, since the live HTTP path already covers that state and the extra API call would just be repeated on every UI poll for a potentially long-running deferral.
Multiple trigger streams (a task deferred, resumed, and deferred again — even by different triggerer processes) are sorted numerically by job id, not lexicographically — job id 7 must sort before job id 200. (Caught this via test — an initial plain-string sort got it backwards.)
Applied the identical fix to the legacy CloudwatchTaskHandler._read_remote_logs() path (still registered in provider.yaml for older-style log-handler configs). Also removed a pre-existing dead call there (self.io.read(...), whose result was immediately discarded and redone manually right below it — wasting a CloudWatch API call on every read).
Trigger-stream discovery failures (e.g. a permissions issue) are reported as an extra message rather than raising, consistent with the existing error handling on the base stream.
Backward compatibility

When no trigger streams exist (the common, non-deferred case), behavior and API call count are unchanged — verified via test that the discovery call only fires when needed and returns [] cleanly for a non-existent/empty-prefix case.

Testing
12 new tests added to test_cloudwatch_task_handler.py, covering: base+trigger merge, correct numeric ordering across multiple trigger streams, the DEFERRED-state skip, graceful handling of a DescribeLogStreams failure, the legacy handler path, correct colon-replacement in the discovery prefix, and hook-level pagination (tested with 54+ streams) and missing-log-group handling.
All 42 tests in the file pass (33 existing + 12 new — some existing tests exercise overlapping paths).
ruff check, ruff format --check, and mypy all pass clean on the three changed files.
Trade-off worth flagging for reviewers

This adds one DescribeLogStreams API call per log read for tasks that are not currently DEFERRED (previously: zero). For a task that was never deferred, this returns an empty list almost instantly, but it is still an extra round-trip. Happy to gate this further (e.g. only when try_number matches a TaskInstanceHistory row known to have gone through DEFERRED) if that trade-off is a concern — went with the simpler, always-check approach first since it's the correct and easy default and mirrors what the issue reporter's own tested workaround did.

AI Disclosure
 This contribution used AI assistance.
Model(s) used: Claude

AI was used for:

Helping write this PR description.
Running the test suite locally to make sure everything passes before submitting.
Reviewing the implementation from a performance perspective.
The implementation and code changes were developed and validated by me.

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
@@ -40,6 +40,7 @@ from airflow.providers.amazon.aws.utils import datetime_to_epoch_utc_ms
 from airflow.providers.common.compat.sdk import conf
 from airflow.utils.log.file_task_handler import FileTaskHandler
 from airflow.utils.log.logging_mixin import LoggingMixin
+from airflow.utils.state import TaskInstanceState
 
 if TYPE_CHECKING:
     import structlog.typing
@@ -271,7 +272,63 @@ class CloudWatchRemoteLogIO(LoggingMixin):  # noqa: D101
         except Exception as e:
             messages.append(str(e))
 
+        # A deferred task's triggerer logs are stored under a *separate* stream
+        # (`<task log stream>.trigger.<job id>.log`) rather than the task's own stream --
+        # see FileTaskHandler.add_triggerer_suffix / TriggerLoggingFactory. The local-file
+        # reader picks these up for free via `glob(worker_log_path.name + "*")`
+        # (FileTaskHandler._read_from_local); CloudWatch has no glob equivalent, so they
+        # have to be discovered explicitly via a ListLogStreams-style prefix scan.
+        #
+        # Skip this while the task is actively DEFERRED: the UI already tails those logs
+        # live from the triggerer over HTTP in that state (FileTaskHandler
+        # ._read_from_logs_server), so the extra CloudWatch API call would be both redundant
+        # and, for a long-running deferral, repeated on every UI poll.
+        if getattr(ti, "state", None) != TaskInstanceState.DEFERRED:
+            try:
+                trigger_stream_names = self._get_trigger_stream_names(relative_path)
+            except Exception as e:
+                messages.append(f"Could not list trigger log streams for {relative_path}: {e}")
+                trigger_stream_names = []
+
+            for trigger_stream_name in trigger_stream_names:
+                messages.append(
+                    f"Reading remote log from Cloudwatch log_group: {self.log_group} "
+                    f"log_stream: {trigger_stream_name}"
+                )
+                try:
+                    trigger_gen: RawLogStream = (
+                        self._parse_log_event_as_dumped_json(event)
+                        for event in self.get_cloudwatch_logs(trigger_stream_name, ti)
+                    )
+                    logs.append(trigger_gen)
+                except Exception as e:
+                    messages.append(str(e))
+
         return messages, logs
+
+    def _get_trigger_stream_names(self, relative_path: str) -> list[str]:
+        """
+        Return the names of any trigger log streams associated with ``relative_path``.
+
+        A task instance can be deferred and resumed more than once -- even by different
+        triggerer processes -- so there may be more than one ``.trigger.<job id>.log``
+        stream for a single attempt. Sorted numerically by job id so multiple deferrals of
+        the same attempt read back in the order they actually happened (job ids are assigned
+        from a monotonically increasing DB sequence); anything that doesn't parse as an
+        integer job id sorts after, by name, rather than raising.
+        """
+        prefix = f"{relative_path.replace(':', '_')}.trigger."
+        streams = self.hook.describe_log_streams(log_group=self.log_group, log_stream_name_prefix=prefix)
+        names = [name for s in streams if (name := s.get("logStreamName"))]
+
+        def _sort_key(name: str) -> tuple[int, int | str]:
+            tail = name[len(prefix) :]
+            job_id_str = tail[: -len(".log")] if tail.endswith(".log") else tail
+            if job_id_str.isdigit():
+                return (0, int(job_id_str))
+            return (1, name)
+
+        return sorted(names, key=_sort_key)
 
     def get_cloudwatch_logs(
         self, stream_name: str, task_instance: RuntimeTI
@@ -417,17 +474,36 @@ class CloudwatchTaskHandler(FileTaskHandler, LoggingMixin):
         self, task_instance, try_number, metadata=None
     ) -> tuple[LogSourceInfo, LogMessages]:
         stream_name = self._render_filename(task_instance, try_number)
-        messages, logs = self.io.read(stream_name, task_instance)
 
         messages = [
             f"Reading remote log from Cloudwatch log_group: {self.io.log_group} log_stream: {stream_name}"
         ]
+        logs: LogMessages = []
         try:
             events = self.io.get_cloudwatch_logs(stream_name, task_instance)
-            logs = ["\n".join(self._event_to_str(event) for event in events)]
+            logs.append("\n".join(self._event_to_str(event) for event in events))
         except Exception as e:
-            logs = []
             messages.append(str(e))
+
+        # See CloudWatchRemoteLogIO.stream() for why deferred-task trigger logs need a
+        # separate lookup -- the same gap applies to this (legacy) handler class.
+        if getattr(task_instance, "state", None) != TaskInstanceState.DEFERRED:
+            try:
+                trigger_stream_names = self.io._get_trigger_stream_names(stream_name)
+            except Exception as e:
+                messages.append(f"Could not list trigger log streams for {stream_name}: {e}")
+                trigger_stream_names = []
+
+            for trigger_stream_name in trigger_stream_names:
+                messages.append(
+                    f"Reading remote log from Cloudwatch log_group: {self.io.log_group} "
+                    f"log_stream: {trigger_stream_name}"
+                )
+                try:
+                    events = self.io.get_cloudwatch_logs(trigger_stream_name, task_instance)
+                    logs.append("\n".join(self._event_to_str(event) for event in events))
+                except Exception as e:
+                    messages.append(str(e))
 
         return messages, logs
 
--- a/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
@@ -24,11 +24,11 @@ import json
 import logging
 import os
 import shutil
-from collections.abc import Generator
+from collections.abc import Callable, Generator
 from datetime import date, datetime, timedelta, timezone
 from functools import cached_property
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeVar
 from urllib.parse import urlsplit
 
 import attrs
@@ -41,6 +41,8 @@ from airflow.providers.common.compat.sdk import conf
 from airflow.utils.log.file_task_handler import FileTaskHandler
 from airflow.utils.log.logging_mixin import LoggingMixin
 from airflow.utils.state import TaskInstanceState
+
+T = TypeVar("T")
 
 if TYPE_CHECKING:
     import structlog.typing
@@ -272,41 +274,18 @@ class CloudWatchRemoteLogIO(LoggingMixin):  # noqa: D101
         except Exception as e:
             messages.append(str(e))
 
-        # A deferred task's triggerer logs are stored under a *separate* stream
-        # (`<task log stream>.trigger.<job id>.log`) rather than the task's own stream --
-        # see FileTaskHandler.add_triggerer_suffix / TriggerLoggingFactory. The local-file
-        # reader picks these up for free via `glob(worker_log_path.name + "*")`
-        # (FileTaskHandler._read_from_local); CloudWatch has no glob equivalent, so they
-        # have to be discovered explicitly via a ListLogStreams-style prefix scan.
-        #
-        # Skip this while the task is actively DEFERRED: the UI already tails those logs
-        # live from the triggerer over HTTP in that state (FileTaskHandler
-        # ._read_from_logs_server), so the extra CloudWatch API call would be both redundant
-        # and, for a long-running deferral, repeated on every UI poll.
-        if getattr(ti, "state", None) != TaskInstanceState.DEFERRED:
-            try:
-                trigger_stream_names = self._get_trigger_stream_names(relative_path)
-            except Exception as e:
-                messages.append(f"Could not list trigger log streams for {relative_path}: {e}")
-                trigger_stream_names = []
+        def _read_trigger_stream(name: str) -> RawLogStream:
+            return (
+                self._parse_log_event_as_dumped_json(event) for event in self.get_cloudwatch_logs(name, ti)
+            )
 
-            for trigger_stream_name in trigger_stream_names:
-                messages.append(
-                    f"Reading remote log from Cloudwatch log_group: {self.log_group} "
-                    f"log_stream: {trigger_stream_name}"
-                )
-                try:
-                    trigger_gen: RawLogStream = (
-                        self._parse_log_event_as_dumped_json(event)
-                        for event in self.get_cloudwatch_logs(trigger_stream_name, ti)
-                    )
-                    logs.append(trigger_gen)
-                except Exception as e:
-                    messages.append(str(e))
+        extra_messages, extra_logs = self.merge_trigger_logs(relative_path, ti, _read_trigger_stream)
+        messages.extend(extra_messages)
+        logs.extend(extra_logs)
 
         return messages, logs
 
-    def _get_trigger_stream_names(self, relative_path: str) -> list[str]:
+    def get_trigger_stream_names(self, relative_path: str) -> list[str]:
         """
         Return the names of any trigger log streams associated with ``relative_path``.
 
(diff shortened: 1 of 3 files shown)


providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
```diff
@@ -417,18 +474,37 @@ def _read_remote_logs(
...
             events = self.io.get_cloudwatch_logs(stream_name, task_instance)
-            logs = ["\n".join(self._event_to_str(event) for event in events)]
+            logs.append("\n".join(self._event_to_str(event) for event in events))
         except Exception as e:
-            logs = []
             messages.append(str(e))
 
+        # See CloudWatchRemoteLogIO.stream() for why deferred-task trigger logs need a
+        # separate lookup -- the same gap applies to this (legacy) handler class.
+        if getattr(task_instance, "state", None) != TaskInstanceState.DEFERRED:
```
vincbeck: This code looks very similar to the block above, any chance to factorize?

ferruzzi:
> This adds one DescribeLogStreams API call per log read for tasks that are not currently DEFERRED 

Not sure if this is an issue worth making changes, but maybe worth keeping an eye on if we go with this plan:   DescribeLogStreams is a fairly low-quota API at 25 TPS per account per region.  I don't know how likely a large environment is to hit that but it seems like a low quota to hit?

providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
```diff
@@ -271,8 +272,64 @@ def stream(self, relative_path: str, ti: RuntimeTI) -> StreamingLogResponse:
...
+        # (`<task log stream>.trigger.<job id>.log`) rather than the task's own stream --
+        # see FileTaskHandler.add_triggerer_suffix / TriggerLoggingFactory. The local-file
+        # reader picks these up for free via `glob(worker_log_path.name + "*")`
+        # (FileTaskHandler._read_from_local); CloudWatch has no glob equivalent, so they
+        # have to be discovered explicitly via a ListLogStreams-style prefix scan.
+        #
+        # Skip this while the task is actively DEFERRED: the UI already tails those logs
+        # live from the triggerer over HTTP in that state (FileTaskHandler
+        # ._read_from_logs_server), so the extra CloudWatch API call would be both redundant
+        # and, for a long-running deferral, repeated on every UI poll.
```
ferruzzi: This may be an edge case, but what happens if the task is deferred again?  `_get_log_retrieval_url` builds the served-log URL from `ti.triggerer_job.id` so the UI is only tailing the logs of the current deferred task.  Do we need some way to persist the logs from previous deferrals into the main logs?

providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
```diff
@@ -417,18 +474,37 @@ def _read_remote_logs(
...
+            logs.append("\n".join(self._event_to_str(event) for event in events))
         except Exception as e:
-            logs = []
             messages.append(str(e))
 
+        # See CloudWatchRemoteLogIO.stream() for why deferred-task trigger logs need a
+        # separate lookup -- the same gap applies to this (legacy) handler class.
+        if getattr(task_instance, "state", None) != TaskInstanceState.DEFERRED:
+            try:
+                trigger_stream_names = self.io._get_trigger_stream_names(stream_name)
```
ferruzzi: Calling a "private" method from another class is a code smell, should `io._get_trigger_stream_names` be renamed to drop the leading underscore?  It also looks like a genuinely useful helper that may be reused elsewhere if it wasn't private.

providers/amazon/tests/unit/amazon/aws/log/test_cloudwatch_task_handler.py
```diff
@@ -594,6 +594,168 @@ def _raise_access_denied(*args, **kwargs):
...
+            log_stream_name_prefix=f"{raw_path.replace(':', '_')}.trigger.",
+        )
+
+    @pytest.mark.parametrize(
+        ("stream_names", "expected_error_code"),
+        [
+            pytest.param([], "ResourceNotFoundException", id="missing_log_group_returns_empty"),
+        ],
+    )
+    def test_describe_log_streams_missing_log_group_returns_empty(self, stream_names, expected_error_code):
```
ferruzzi: Parameterizing with only one param case?  It also doesn't look like `expected_error_code` is used int he test unless I'm missing something

bujjibabukatta:
Hey @ferruzzi, thanks for digging into this so carefully — went through all four:

The re-deferral thing was a genuine bug, glad you caught it. Fixed it so we only skip the current triggerer's stream while DEFERRED, not the whole lookup — so logs from an earlier deferral don't disappear anymore. Added a test to lock that in.
Renamed _get_trigger_stream_names to get_trigger_stream_names (dropped the underscore), and did the same for the new shared helper.
Cleaned up the parametrize test too — it's just a plain test now, no more unused param.
And on the quota thing — heads up that the re-deferral fix actually makes DescribeLogStreams get called a bit more often now, not less, since it runs during DEFERRED too. Not opposed to gating it down the line (caching, or keying off TaskInstanceHistory) if it ends up mattering in practice, just didn't want to over-engineer it without real numbers.

providers/amazon/src/airflow/providers/amazon/aws/log/cloudwatch_task_handler.py
```diff
@@ -271,8 +274,98 @@ def stream(self, relative_path: str, ti: RuntimeTI) -> StreamingLogResponse:
...
+            messages.append(f"Could not list trigger log streams for {relative_path}: {e}")
+            return messages, results
+
+        if getattr(ti, "state", None) == TaskInstanceState.DEFERRED:
+            current_job_id = getattr(getattr(ti, "triggerer_job", None), "id", None)
+            if current_job_id is not None:
+                live_suffix = f".trigger.{current_job_id}.log"
+                trigger_stream_names = [
+                    name for name in trigger_stream_names if not name.endswith(live_suffix)
+                ]
```
ferruzzi, line 355: So many stacked `getattrs` here makes me think there should be a better way.  Can you try this, and see if it works?  It may need a little tweaking, but should be cleaner, I think:

```python
if (ti.state == TaskInstanceState.DEFERRED) and (job := getattr(ti, "triggerer_job", None)):
    live_stream = f"{relative_path.replace(':', '_')}.trigger.{job.id}.log"
    trigger_stream_names = [name for name in trigger_stream_names if name != live_stream]
```

Plus, hardcoding the prefix assembly here makes me think we should likely move that builder into a shared helper at some point, but that can be done in a different PR.

PR #70636, opened 2026-07-28
Wait for Job pods before failing KubernetesJobOperator discovery

### LINKED ISSUE #70585: KubernetesJobOperator.get_pods() fails the task instead of waiting for the Job controller to create pods

Providers

### Apache Airflow version

2.11.2

`get_pods()` issues `discover_pods_retry_number + 1` (default 4) `list_namespaced_pod` calls back to back with **no delay**, then raises `No pods running with labels ...` if the pods are not visible yet. Raising `discover_pods_retry_number` only adds more immediate GETs — there is no way to wait.

Production trace for one occurrence:

| time (UTC) | event |
|---|---|
| `00:00:23.223` | `job.py` — `Building job job-<task>-<suffix>` |
| `00:00:23.360` | task failed — `AirflowException: No pods running with labels ...` |
| `00:00:23` | Job controller — `SuccessfulCreate: Created pod job-<task>-<suffix>-<id>` |
| `00:01:28` | Job controller — `Job completed` |

`kube_job_created` and `kube_pod_created` are both `00:00:23`: the pod existed in the same second the operator gave up, 137 ms after the Job was built. The workload then ran to success unmonitored. ~20 occurrences in 5 days across 8+ unrelated DAGs.

Consequences:

1. The task is marked failed while its Job succeeds. Nothing waits on, monitors or cleans up the Job; `on_kill()` does not fire.
2. `self.pods` is never assigned, so `_cleanup_monitoring_pods()` in the `finally` block no-ops.
3. A retry or manual clear starts a **second** Job while the first is still running or has already committed its writes — silent double execution for non-idempotent tasks.

The message is also inaccurate: the pods are not "not running", they do not exist yet.

Wait for the pods, bounded by a timeout. No new parameters needed — `KubernetesJobOperator` extends `KubernetesPodOperator`, so `schedule_timeout_seconds` (deadline, default 120 via `startup_timeout_seconds`) and `startup_check_interval_seconds` (interval, default 5) are already inherited and documented. The first probe stays immediate, so tasks whose pods appear at once are unaffected.

Also: break on `>=` parallelism rather than `==`, and deprecate `discover_pods_retry_number` as a no-op.

### Operating System

_No response_

### Deployment

_No response_

 cncf provider at 10.13.0 but related code is unchanged in current main

1.19.0

### Kubernetes Version

1.35.6

### Helm Chart configuration

Not Applicable

Not Applicable

### Anything else?

The issue is Inherited by `KubernetesStartKueueJobOperator`, `GKEStartJobOperator`, `GKEStartKueueJobOperator` — i.e. exactly the Kueue case where pod creation is slowest.

The issue is known and was raised in https://github.com/apache/airflow/pull/53368 but rejected as out of scope for this PR.

### Code of Conduct

cipheraxat:
`KubernetesJobOperator.get_pods()` used to hammer `list_namespaced_pod` a few times with no delay, then fail if the Job controller had not created pods yet. In practice that race marks the task failed while the Job still runs to success, and a retry can start a second Job.

This change waits up to `schedule_timeout_seconds`, polling every `startup_check_interval_seconds` (both already inherited from `KubernetesPodOperator`). The first probe stays immediate. Discovery also stops once `len(pods) >= parallelism`, and `discover_pods_retry_number` is deprecated as a no-op.

closes: #70585

---

- [X] Yes — Cursor Grok 4.5

Generated-by: Cursor Grok 4.5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/job.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/job.py
@@ -154,10 +154,6 @@ class KubernetesJobOperator(KubernetesPodOperator):
                 AirflowProviderDeprecationWarning,
                 stacklevel=2,
             )
-        # Kept for backward compatibility with code that reads the attribute.
-        self.discover_pods_retry_number = (
-            discover_pods_retry_number if discover_pods_retry_number is not None else 3
-        )
         self.unwrap_single = unwrap_single
 
     @property
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/job.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/job.py
@@ -23,6 +23,7 @@ import json
 import logging
 import os
 import sys
+import time
 import warnings
 from collections.abc import Sequence
 from functools import cached_property
@@ -97,8 +98,8 @@ class KubernetesJobOperator(KubernetesPodOperator):
         'Background' - allow the garbage collector to delete the dependents in the background;
         'Foreground' - a cascading policy that deletes all dependents in the foreground.
         Default value is 'Foreground'.
-    :param discover_pods_retry_number: Number of time list_namespaced_pod will be performed to discover
-        already running pods.
+    :param discover_pods_retry_number: Deprecated and ignored. Pod discovery waits up to
+        ``schedule_timeout_seconds``, polling every ``startup_check_interval_seconds``.
     :param unwrap_single: Unwrap single result from the pod. For example, when set to `True` - if the XCom
         result should be `['res']`, the final result would be `'res'`. Default is True to support backward
         compatibility.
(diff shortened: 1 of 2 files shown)


providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/job.py
```diff
@@ -145,7 +146,18 @@ def __init__(
...
+                "`discover_pods_retry_number` is deprecated and has no effect. "
+                "Pod discovery waits up to `schedule_timeout_seconds`, polling every "
+                "`startup_check_interval_seconds`.",
+                AirflowProviderDeprecationWarning,
+                stacklevel=2,
+            )
+        # Kept for backward compatibility with code that reads the attribute.
+        self.discover_pods_retry_number = (
+            discover_pods_retry_number if discover_pods_retry_number is not None else 3
+        )
```
mo-rieger: AFAICS there is no code that reads `discover_pods_retry_number` when this fix is applied.

cipheraxat: Removed the unused `self.discover_pods_retry_number` assignment. The kwarg is still accepted and emits the deprecation warning; the test now only asserts that warning. Also rebased onto latest `main`.

PR #68919, opened 2026-06-23
Make deadline reads and serialization robust to dynamic/malformed intervals

seanghaeli:
`DeadlineAlert.interval` is a JSON column (serialized `timedelta` or dynamic `VariableInterval`), but the `/ui/dags/{dag_id}/deadlineAlerts` response declared `interval: float` — any alert with a dict-shaped interval failed validation and the endpoint returned 500, breaking the run-page deadline badge.

Fix: `interval` becomes `float | None` with a validator (seconds for fixed intervals, `None` for dynamic), and `interval` is dropped from `order_by` (sorting a JSON column sorts by structure, not duration).

Robustness fixes from earlier revisions split out to #70421 per review.

Closes: https://github.com/apache/airflow/issues/69245

PR #70637, opened 2026-07-28
Regenerate UI client and handle deadline alerts with no fixed interval

shahar1:
`main` currently fails `Compile / format / lint UI` for every PR that triggers a full static-check run.

#68919 made `DeadlineAlertResponse.interval` nullable (`float | None`) for dynamic intervals — a `VariableInterval` whose value is only resolved at scheduler evaluation time — but the generated TypeScript client was not regenerated. The stale client still declared `interval: number`, which masked a real type error: once the client is regenerated, `tsc` fails at five call sites that pass the value straight to `dayjs.duration(...)`.

This regenerates the client and makes those call sites handle a null interval.

A dynamic interval has no duration to display, so those sites render a dedicated wording instead — "Must complete within a dynamically resolved interval of {reference}" — via a new key in `en/dag.json` that other locales fall back to per the UI i18n policy (no other locale files touched). The shared `humanizeSeconds` helper keeps the null handling in one place next to the existing null-tolerant `renderDuration`, the "Met" badge tooltip is disabled rather than rendered empty when there are no rules to show, and a component test pins the dynamic-interval rendering.

Reproduce on `main` with no changes: `prek run ts-compile-lint-ui --all-files`.

related: #68919

> [!NOTE]
> `v3-3-test` will need the same fix. #70625 backports the same datamodel change and touches only backend files, so merging it as-is reproduces this breakage there. Folding these changes into that PR before it merges avoids landing the breakage and cherry-picking a follow-up.

---


shahar1:
Drafting as I'm still optimizing this PR.
My main concern - how the error didn't surface in static checks of https://github.com/apache/airflow/pull/68919

seanghaeli:
Thanks for putting this up. Seems that `ts-compile-lint-ui` pre-commit hook was skipped during https://github.com/apache/airflow/pull/68919, which is why this bug was not surfaced. Quick fix could be to make sure it runs whenever any yaml under `core_api/openapi/` changes, what do you think?

shahar1:
> Thanks for putting this up. Seems that `ts-compile-lint-ui` pre-commit hook was skipped during #68919, which is why this bug was not surfaced. Quick fix could be to make sure it runs whenever any yaml under `core_api/openapi/` changes, what do you think?

On it right now :) It seems that indeed there's a misconfiguration in the selective checks

bbovenzi (approved):

potiuk (approved):

seanghaeli (approved):

amoghrajesh (approved):

pierrejeambrun (commented):
Thanks for fixing this.

I'm pretty sure the CI was green, not sure how this did not trigger any error.

PR #70638, opened 2026-07-28
Add aws_conn_id to template_fields in Amazon S3 operators (#35259)

### LINKED ISSUE #35259: Allow jinja templating connection ids for all third party operators
### Description

We use private staging and prod S3s(Ceph clusters for example) in our office.
So there are often cases where DAGs are running with only connection ids changed.
We prefer to use <code>Param</code> rather than to use hardcoded connection ids to make our code reusable.
I only gave an example for Amazon operator, but templating connection ids is required for other operators too.

Why is it needed?
Code reusability

### Use case/motivation

```Python3
with DAG(
    dag_id="example_s3",
    params={
        "aws_conn_id": Param("", type="string"),
    },
    ...
) as dag:

    create_object = S3CreateObjectOperator(
        task_id="create_object",
        s3_bucket=bucket_name,
        s3_key=key,
        data=DATA,
        replace=True,
        aws_conn_id="{{ params.aws_conn_id }}", # Params enable us to provide runtime configuration
    )
```

### Related issues

_No response_

### Code of Conduct

divijaksolves:
closes: #35259

### Description
This PR adds `aws_conn_id` to `template_fields` in `S3CreateObjectOperator` to enable Jinja templating support for dynamic connection IDs.

### Testing
Tested locally via Breeze with:
`pytest providers/amazon/tests/unit/amazon/aws/operators/test_s3.py`
All 79 tests passed successfully.

---


SameerMesiah97 (commented):
Since this changes `template_fields`, should `test_template_fields` be updated to include a templated `aws_conn_id`?

divijaksolves:
Thanks for the review! Updated `test_template_fields` in `test_s3.py` to include a templated `aws_conn_id="aws_{{ ds }}"` and verified that all 79 tests pass locally via Breeze.

potiuk (approved):

potiuk:
Thanks for picking this up, and sorry it's not the outcome you were hoping for.

This change is a no-op. `aws_template_fields()` unions its arguments with a fixed set that already contains `aws_conn_id`:

```python
return tuple(sorted({"aws_conn_id", "region_name", "verify"} | set(template_fields)))
```

So the resulting tuple is byte-identical before and after:

```
('aws_conn_id', 'data', 'region_name', 's3_bucket', 's3_key', 'verify')
```

`S3CreateObjectOperator` already supports a templated `aws_conn_id` — every operator built on `AwsBaseOperator` does. You can verify with the example from #35259; passing `aws_conn_id="{{ params.aws_conn_id }}"` works on `main` today.

#35259 asks for templated connection ids across *all* third-party operators, and Amazon is the provider where it already works. The remaining work is in the providers that don't have an equivalent of `aws_template_fields` — that's where a fix would land.

Closing this one, but the issue is still open and worth taking on from that angle.

---

PR #70639, opened 2026-07-28
Let providers contribute tables to 'db clean'

uranusjr:
This is the follow-up mentioned in #70147.

A 'db-cleanup-tables' key in provider.yaml is added to specify a callable returning cleanup configs, discovered via ProvidersManager. The 'db clean' command merges them into the list of tables to clean at runtime.

Celery uses it to register 'celery_taskmeta' and 'celery_tasksetmeta' tables, using the '[celery] result_backend_schema' config that schema-qualifies them when the result backend is in its own schema. This allows them to always be correctly found by the 'db clean' command. Both are removed from Core's hardcoded list.

Possible future candidates (no Core change needed, deferred to their maintainers): FAB's session table and the edge3 tables.

### DIFF
--- a/providers/celery/src/airflow/providers/celery/get_provider_info.py
+++ b/providers/celery/src/airflow/providers/celery/get_provider_info.py
@@ -45,6 +45,7 @@ def get_provider_info():
             "airflow.providers.celery.executors.celery_kubernetes_executor.CeleryKubernetesExecutor",
         ],
         "cli": ["airflow.providers.celery.cli.definition.get_celery_cli_commands"],
+        "db-cleanup-tables": ["airflow.providers.celery.cleanup.get_db_cleanup_table_configs"],
         "config": {
             "celery_kubernetes_executor": {
                 "description": "This section only applies if you are using the ``CeleryKubernetesExecutor`` in\n``[core]`` section above\n",
@@ -131,6 +132,13 @@ def get_provider_info():
                         "type": "string",
                         "sensitive": True,
                         "example": "db+postgresql+psycopg://postgres:airflow@postgres/airflow",
+                        "default": None,
+                    },
+                    "result_backend_schema": {
+                        "description": "The database schema that the Celery result backend tables (``celery_taskmeta``,\n``celery_tasksetmeta``) live in, when they are provisioned into a schema other than\nthe one used by ``sql_alchemy_conn``. When set, ``airflow db clean`` (and the\narchived-table export/drop commands) operate on these tables in this schema instead\nof silently skipping them.\n",
+                        "version_added": None,
+                        "type": "string",
+                        "example": "celery",
                         "default": None,
                     },
                     "result_backend_sqlalchemy_engine_options": {
--- a/airflow-core/src/airflow/cli/cli_config.py
+++ b/airflow-core/src/airflow/cli/cli_config.py
@@ -520,7 +520,7 @@ ARG_DB_TABLES = Arg(
     help=lazy_object_proxy.Proxy(
         lambda: (
             f"Table names to perform maintenance on (use comma-separated list).\n"
-            f"Options: {import_string('airflow.cli.commands.db_command.all_tables')}"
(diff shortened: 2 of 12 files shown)


dheerajturaga:
Thanks for teasing this apart — I think the direction is right and worth landing.

The thing that makes this feel load-bearing rather than tidying: #70147 shipped schema support but no user-reachable knob for the two tables it was actually motivated by, so deployments with a separate result-backend schema get a silently-skipped `db clean` for Celery today. The obvious shortcut — adding `[celery] result_backend_schema` in core — recapitulates the FAB `session` hack exactly (core reading a provider's config at module import with a `fallback` because provider config isn't loaded yet). This PR is the cheap way out of that, and it reuses the `db-managers` shape rather than introducing a new pattern.

A few interface points I'd raise, in the order they bother me:

1. **Nameable contract.** The provider callable returns `list[dict]` splatted into `_TableConfig(**spec)` — a leading-underscore dataclass. Any field rename in core silently breaks installed providers with no version negotiation, and `except Exception: logger.warning(...)` swallows it. Either a `TypedDict` for the payload or promoting `_TableConfig` to a public `TableConfig` would give the extension point a stable, type-checkable shape.
2. **Don't silently skip when the user was explicit.** `--tables celery_taskmeta` failing quietly is exactly the failure mode this PR exists to fix. Log-and-skip is reasonable for a bare `db clean`; an explicit table name that resolves to a broken provider callable should be a hard error.
3. **Silent name collisions.** `itertools.chain` puts providers last, so a provider declaring `table_name="task_instance"` overrides core's config with no diagnostic. One `if name in existing: raise` closes it.
4. **`test_no_models_missing` narrows.** With `celery_taskmeta`/`celery_tasksetmeta` gone from `config_list`, the "every ORM model is either cleaned or excluded" guard no longer sees provider-contributed tables. Worth extending it to walk the merged set so a future orphan provider model still trips the check.

On the two candidates:

- **FAB `session`** — I'd move it. That `fallback="database"` comment in `db_cleanup.py` is the artifact this PR exists to delete; not moving it is the odd choice. Same table, same recency column, resolved at the right layer.
- **Edge3** — worth flagging that the Edge executor already purges its own tables in the hot path (`_purge_jobs` in `edge_executor.py`), so `db clean` here is a safety net for orphans (executor not running, team removed, executor swapped out), not the primary reaper. `edge_logs` is the cleanest fit. `edge_job` needs a terminal-state guard via `extra_columns=["state"]` + `extra_filters` — the pattern `connection_test_request` already uses — and `queued_dttm` reads to me as the better recency column since `last_update` is nullable. `edge_worker` I'd leave alone: it's a current-state registry keyed by `worker_name`, not an age-based cleanup target.

Neither is urgent; both fall out cleanly once the extension point is in.

---
Drafted-by: Claude Code (Opus 5); reviewed by @dheerajturaga before posting

Miretpl:
Hi, I like the idea and I think it is good direction. Regarding the future candidates - I would move all provider-related tables to providers without any exception as it is the direction in which we are heading with decoupling of Airflow.

eladkal (approved):

providers/celery/src/airflow/providers/celery/get_provider_info.py
```diff
@@ -133,6 +134,13 @@ def get_provider_info():
                         "example": "db+postgresql+psycopg://postgres:airflow@postgres/airflow",
                         "default": None,
                     },
+                    "result_backend_schema": {
+                        "description": "The database schema that the Celery result backend tables (``celery_taskmeta``,\n``celery_tasksetmeta``) live in, when they are provisioned into a schema other than\nthe one used by ``sql_alchemy_conn``. When set, ``airflow db clean`` (and the\narchived-table export/drop commands) operate on these tables in this schema instead\nof silently skipping them.\n",
+                        "version_added": None,
```
Miretpl, line 139: Could we specify `version_added` here?

Miretpl (commented):
One nit, the rest of the code looks good to me.

PR #66979, opened 2026-05-15
Enable ruff B008 (function-call-in-default-argument) and fix violations

shahar1:
## Summary

Enables ruff rule **B008** (`function-call-in-default-argument`) which catches function calls used as default argument values — a common source of bugs where mutable objects or config reads are shared across all calls instead of being fresh per call.

- Adds `[tool.ruff.lint.flake8-bugbear] extend-immutable-calls` to exempt FastAPI DI callables (`Depends`, `Query`, `Path`, `Body`, `Security`) and the stateless `cryptography` SHA256 descriptor, which are intentionally used in argument defaults
- Fixes **13 real violations** across core, task-sdk, shared libs, and providers:
  - **task-sdk `BaseSensorMixin`**: `conf.getfloat("sensors", "default_timeout")` was read once at import time; now read at instantiation
  - **task-sdk `Resources`**: `conf.getint(...)` for `cpus`/`ram`/`disk`/`gpus` read at import time; now read at instantiation
  - **providers/google `S3ToGCSOperator`**: `conf.getboolean("operators", "default_deferrable")` read at import time
  - **providers/google Vertex AI Ray hook and operator**: `resources.Resources()` mutable default shared across all calls
  - **shared/observability** (`SafeStatsdLogger`, `SafeDogStatsdLogger`, `SafeOtelLogger`): `PatternAllowListValidator()` mutable instances shared across all logger instances
  - **openlineage system test**: `setup_jinja()` mutable Jinja `Environment` shared across operator instances
  - **api_fastapi `LimitFilter.depends`**: `conf.getint("api", "fallback_page_limit")` in FastAPI dependency method moved to module-level constant `_FALLBACK_PAGE_LIMIT`, preserving the OpenAPI schema default value while making the intent explicit
  - **providers/fab test conftest**: `types.SimpleNamespace(...)` mutable default in context manager helper

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Sonnet 4.6 (claude-sonnet-4-6)

Generated-by: Claude Sonnet 4.6 (claude-sonnet-4-6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #66978, opened 2026-05-15
Enable ruff PLE1205 to catch logging calls with too many arguments

shahar1:
When a `logging` call has more positional arguments than the format string has `%`-placeholders, the extra arguments are silently dropped and the intended information never appears in the log. Enable `PLE1205` (`logging-too-many-args`) to make this a lint error, and fix all 12 existing instances.

## Changes

- **\`pyproject.toml\`**: add `PLE1205` to `extend-select`.
- **\`airflow-core\`** (1): add missing `%s` placeholder for the caught `OSError` in `file_task_handler.py`.
- **\`provider:amazon\`** (4):
  - `redshift_data.py`: collapse two-string call into `"Invalid poll_interval: %s", poll_interval`
  - `ssm.py`: merge accidentally split message strings into one format string with `%s`
  - `task_log_fetcher.py` (×2): add `%s` for the captured exception
- **\`provider:google\`** (7):
  - `compute.py` (×2): add `%s` so `self.body` is actually logged
  - `drive.py`: add `%s` for the `GoogleApiClientError`
  - Four system-test example files using `{}` placeholders (logging uses `%s`, not `.format()` syntax)

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Sonnet 4.6)

Generated-by: Claude Code (Sonnet 4.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #66977, opened 2026-05-15
Enable ruff B015 to catch silent no-op comparisons in tests

shahar1:
Bare `actual == expected` expressions in tests silently evaluate and discard their result — the test passes regardless of correctness. This is exactly what happened in #66894 (four `test_serialize` methods in the Vertex AI triggers had been vacuously passing for years).

## Changes

- **`pyproject.toml`**: add `B015` (`useless-comparison`) to `extend-select`. This rule fires on any bare comparison expression used as a statement.
- **16 test files**: add the missing `assert` to all 26 existing violations found by the rule across `airflow-core`, `task-sdk`, and providers `amazon`, `cncf/kubernetes`, `google`).

With this in place, any future bare comparison in a test is a lint error caught before CI runs.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Sonnet 4.6)

Generated-by: Claude Code (Sonnet 4.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #66960, opened 2026-05-14
Add static check ensuring trigger `__init__()` and `serialize()` stay in sync

shahar1:
A trigger's `__init__` and `serialize()` are written as a pair: any `__init__` parameter that `serialize()` does not return is silently dropped when the triggerer re-instantiates the trigger, falling back to the parameter's default. This kind of mismatch is easy to introduce and hard to spot in review.

This adds an AST-based prek static check over provider trigger modules (`*/src/airflow/providers/*/triggers/*.py`) that flags such mismatches.

Inspiration for this PR: https://github.com/apache/airflow/pull/66917

## Details

- Resolves the `__init__`/`serialize()` pair through **in-file base classes**, so a class that inherits one and overrides the other is still checked. Also follows `**super().serialize()[1]` spreads recursively.
- When neither the signature nor the serialize dict can be resolved statically (dynamic dict construction, `.update()`, base class in another file), it **skips** rather than guessing — no false positives.
- 8 pre-existing failures were found and reviewed:
  - **5 genuine violations** are listed in `KNOWN_VIOLATIONS` and excluded for now; they will be fixed in a follow-up PR (`DatabricksExecutionTrigger`, `DatabricksSQLStatementExecutionTrigger`, `BigQueryIntervalCheckTrigger`, `DataFusionStartPipelineTrigger`, `LivyTrigger`).
  - **3 by-design cases** are permanently excluded in `BY_DESIGN_EXCLUSIONS` with documented reasons — a deprecated/aliased parameter (`delta` → `moment`, `pod_name` → `pod_names`) is folded into its replacement at construction time, so it does not need to round-trip.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Sonnet 4.6)

Generated-by: Claude Code (Sonnet 4.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70640, opened 2026-07-28
Enable ruff B023 (function-uses-loop-variable) and fix violations

dkranchii:
B023 catches the late-binding closure-over-loop-variable footgun where a function defined inside a loop captures the loop variable by reference, so every function in the resulting list sees the same (final) value — a classic silent-bug source in Python. See the [Python FAQ entry](https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result).

The one user-visible fix is in `providers/standard/.../triggers/file.py`, where the `FileTrigger`'s `os.walk` lambda was dispatched to a worker thread via `anyio.to_thread.run_sync` while the outer glob iteration could advance, potentially walking the wrong path.

The rest are pre-existing latent-bug or false-positive sites fixed by binding the loop-derived variable as a default argument on the inner function or lambda:

- `airflow-core/.../0101_3_2_0_ui_improvements_for_deadlines.py` (migration; 7 sites in one nested function)
- `dev/breeze/.../ui_commands.py` (two i18n helpers)
- `devel-common/src/sphinx_exts/providers_extensions.py` (two class-extras lambdas)
- `providers/apache/hive/tests/unit/.../test_s3_to_hive.py`
- `providers/exasol/tests/unit/exasol/hooks/test_sql.py`
- `providers/google/tests/system/.../example_cloud_sql_query.py`
- `providers/smtp/src/airflow/providers/smtp/hooks/smtp.py` (OAuth2 auth callback)
- `shared/secrets_masker/.../secrets_masker.py` (subclass compat shim)

Follows the pattern of #66977 (B015), #66978 (PLE1205), #66979 (B008), and #66960 (trigger init/serialize static check).

---

- [X] Yes — Cursor


airflow-core/src/airflow/migrations/versions/0101_3_2_0_ui_improvements_for_deadlines.py
```diff
@@ -594,7 +594,13 @@ def _migrate_deadline_alerts() -> None:
             dags_with_deadlines.add(dag_id)
             deadline_alerts = dag_deadline if isinstance(dag_deadline, list) else [dag_deadline]
 
-            def _migrate_dag_deadlines(dag_conn: Connection) -> Iterable[str]:
+            def _migrate_dag_deadlines(
```
potiuk: Please revert this file and use `# noqa: B023` instead.

This is a **released migration** (0101, shipped in 3.2.0) and the change is semantically a no-op — the closure is called at line 688 inside the same iteration where it's defined here, so no value ever leaked across iterations.

Rewriting a shipped migration to satisfy a linter means a user who already ran it has different source than a fresh install, for zero behavioural gain. A `# noqa: B023` with a one-line comment ("closure is invoked within the same iteration") leaves the historical file intact and documents *why* it's safe, which is more useful to the next reader than five extra default arguments.

---

shared/secrets_masker/src/airflow_shared/secrets_masker/secrets_masker.py
```diff
@@ -223,8 +223,8 @@ def __init_subclass__(cls, **kwargs):
                 f = cls._redact
 
                 @functools.wraps(f)
-                def _redact(*args, replacement: str = "***", **kwargs):
-                    return f(*args, **kwargs)
+                def _redact(*args, replacement: str = "***", _f=f, **kwargs):
```
potiuk: Binding via a default argument leaks `_f` into the function's **keyword signature**, and this one is wrapped in `functools.wraps` — so introspection now advertises a parameter that isn't part of the contract, and a caller who happens to pass `_f` in `**kwargs` would silently substitute a different function. For a secrets-masking code path that's a sharper edge than it looks.

A closure factory avoids exposing anything:

```python
def _make_redact(f):
    @functools.wraps(f)
    def _redact(*args, replacement: str = "***", **kwargs):
        return f(*args, **kwargs)
    return _redact
```

Same effect, nothing added to the signature.

---

potiuk (changes requested):
Thanks — B023 is a worthwhile rule to turn on; late-bound closures over loop variables are a genuinely nasty bug class because they fail silently and only under specific timing. Adding it to `pyproject.toml` alongside the fixes is the right way to land it.

I traced the two most interesting cases rather than assuming:

- **`triggers/file.py`** is the nicest fix in the set — rewriting `run_sync(lambda: list(os.walk(path)))` as `run_sync(lambda p: list(os.walk(p)), path)` passes the value explicitly instead of relying on a default-argument trick. Worth using that form elsewhere where the callee accepts arguments; it reads as intent rather than as lint appeasement.
- **The migration** (`0101_3_2_0_ui_improvements_for_deadlines.py`) is a false positive in practice: `_migrate_dag_deadlines` is defined at line 597 and called at line 688 within the same loop iteration, so late binding was never observable there.

Two changes requested inline before this lands — one about rewriting a shipped migration for no behavioural gain, one about a binding that leaks into a public keyword signature in the secrets-masking path. Neither is hard to address.

These are judgement calls rather than mechanical fixes, so I'd like your own reasoning in reply rather than just an autofix — my review here was AI-assisted and shouldn't be treated as settled.

---

shared/secrets_masker/src/airflow_shared/secrets_masker/secrets_masker.py
dkranchii: @potiuk  Good catch on the signature leak. The bit that pushed it over the edge for me was your `**kwargs` collision point - `_redact` recurses through arbitrary user data (dicts, lists, tuples, sets), and any of those paths that route kwargs through would let a caller sneak `_f` in and substitute a different function *inside* the redaction walk. Not a place I want a sharp edge.
Switched to the factory form you suggested. Kept a short comment at the site so the next person doesn't try to "simplify" it back to a default-arg binding.

airflow-core/src/airflow/migrations/versions/0101_3_2_0_ui_improvements_for_deadlines.py
dkranchii: @potiuk Fair point on shipped-migration hygiene. I hadn't thought about the drift between a user's already-applied 3.2.0 source and `main` for the same revision ID. Reverted the function signature to exactly what shipped.
One small deviation from your suggestion: ruff flags B023 at the seven reads of `dag_id` / `serialized_dag_id` / `deadline_alerts` / `created_at` *inside* the closure body, not on the `def` line. So a single `# noqa: B023` on the def doesn't suppress anything, and seven per-line noqas would have added exactly the kind of body-level noise we're trying to avoid in a shipped migration. Went with a file-level `# ruff: noqa: B023` at the top of the module plus a short comment explaining the invariant (closure is invoked and fully drained via `list(_migrate_dag_deadlines(dag_conn))` inside the same iteration). Same intent - historical body byte-identical to 3.2.0, one centralized "why safe" explanation. Happy to switch to per-line noqas if you'd rather have it that way.

potiuk (approved):
Both points addressed, and on the migration you were right to push back on my suggestion.

`# noqa: B023` on the `def` line would indeed have suppressed nothing — ruff flags the reads inside the closure body, not the definition, so it would have taken seven per-line noqas to do what I asked for, in a shipped migration where body noise is exactly what we were trying to avoid. The module-level `# ruff: noqa: B023` with the invariant written out, plus the explicit "do not rewrite the body for lint hygiene" note, is a better answer than the one I proposed. Thanks for checking rather than just applying it.

The secrets-masker factory is right too, and your reasoning is the reason it matters: `_redact` recurses through arbitrary user data, so a `**kwargs` path that let a caller slip in `_f` would substitute a different function *inside* the redaction walk. The comment at the site should stop someone collapsing it back to a default-arg binding.

I have updated the branch from `main`. The one red check — `Postgres 18 / Py3.14 · API…CLI` — was from 4 August, ran for six minutes, and produced no pytest summary at all, so it never reached the test stage; that is an infrastructure or expired-image failure rather than anything in this diff. A lint rule plus behaviour-preserving rewrites would not break a single matrix cell of core API/CLI DB tests and nothing else.

Approving on that basis. If the fresh run comes back red on something real, ping me and I will look again rather than leaving you to guess at it.

---

potiuk (approved):

PR #70641, opened 2026-07-28
Drop connection headers on cross-host redirect in HttpAsyncHook (#70164)

### LINKED ISSUE #70164: HttpAsyncHook replays connection extra headers across a cross-host redirect

http

### What happened

`HttpAsyncHook.config` copies every non-reserved key of an HTTP Connection's `extra` field into the `aiohttp.ClientSession` headers, and the provider docs endorse that field as a place to carry credentials ("Login and Password authentication can be used along with any authentication method using headers. Headers can be given in json format in the Extras field").

`aiohttp` follows redirects by default and strips only the literal `Authorization` header when the redirect changes host. A secret carried under any other header name (`X-API-Key`, `Private-Token`, `apikey`, ...) is a plain session header and is replayed verbatim to every host in the redirect chain. Verified against aiohttp 3.14.1: `Authorization` is dropped on a cross-host redirect, `X-API-Key` is forwarded.

The destination does not have to be attacker-controlled for this to leak - an API that 302s downloads to a CDN or object store receives the connection's key.

Connection-supplied headers should get the same treatment `aiohttp` already gives `Authorization`: dropped once the redirect leaves the original host, preserved on same-host redirects.

### Follow-up work

https://github.com/apache/airflow/pull/70000 fixes the synchronous `HttpHook` by subclassing `requests.Session` and extending `rebuild_auth`, which is the hook `requests` exposes for exactly this. `aiohttp` has no equivalent per-redirect callback, so the async side needs a different approach - most likely disabling automatic redirect following in `HttpAsyncHook.run` and walking the chain manually so headers can be recomputed per hop, or moving the connection headers off the session and onto each request.

### Acceptance criteria

- A cross-host redirect from `HttpAsyncHook` does not forward Connection `extra` headers.
- A same-host redirect still forwards them.
- Regression test covering both directions, mirroring `test_connection_header_is_only_forwarded_on_a_same_host_redirect` in `providers/http/tests/unit/http/hooks/test_http.py`.

waterWang:
### Summary

`HttpAsyncHook` forwards Connection `extra` headers (e.g. `X-API-Key`, `Private-Token`) across cross-host redirects because `aiohttp` only strips the `Authorization` header. This is a security issue — a secret held under any other header name is replayed verbatim to the redirect target.

PR #70000 fixed the synchronous `HttpHook` by subclassing `requests.Session` and extending `rebuild_auth`. This PR fixes the async side.

### Approach

`aiohttp` has no per-redirect callback equivalent to `requests.Session.rebuild_auth()`, so the fix disables automatic redirect following (`allow_redirects=False`) and walks the redirect chain manually. When the redirect target changes host, connection-supplied headers are dropped from subsequent requests. Same-host redirects are unaffected.

### Changes

1. **`providers/http/src/airflow/providers/http/hooks/http.py`** — `AsyncHttpSession.run()` now handles redirects manually:
   - Tracks which headers originated from the connection
   - Disables `aiohttp`'s auto-redirect
   - On cross-host redirect: drops connection headers, keeps caller-supplied headers
   - On same-host redirect: retains all headers
   - Max 5 redirects followed (matching `requests` default)

2. **`devel-common/src/tests_common/test_utils/aiohttp.py`** — Added `headers` and `release()` to `MockAiohttpClientResponse` to support redirect tests.

3. **`providers/http/tests/unit/http/hooks/test_http.py`** — Added `test_async_connection_header_is_only_forwarded_on_a_same_host_redirect` parametrized for cross-host and same-host redirects.

### Verification

- Cross-host redirect: `X-API-Key` present on first request, absent on second
- Same-host redirect: `X-API-Key` present on both requests
- Existing tests pass unchanged

### Related

- Fixes #70164
- Sync hook fix: #70000

### DIFF
--- a/devel-common/src/tests_common/test_utils/aiohttp.py
+++ b/devel-common/src/tests_common/test_utils/aiohttp.py
@@ -40,6 +40,7 @@ class MockAiohttpClientResponse:
         method: str = "GET",
         url: str = "http://example.com",
         reason: str | None = None,
+        headers: dict[str, str] | None = None,
     ) -> None:
         self.status = status
         self._payload = payload
@@ -47,6 +48,11 @@ class MockAiohttpClientResponse:
         self._method = method
         self._url = url
         self._reason = reason
+        self.headers = headers or {}
+
+    async def release(self) -> None:
+        """No-op for unit tests; aiohttp frees the underlying connection."""
+        pass
 
     @property
     def reason(self) -> str:
--- a/providers/http/src/airflow/providers/http/hooks/http.py
+++ b/providers/http/src/airflow/providers/http/hooks/http.py
@@ -512,19 +512,63 @@ class AsyncHttpSession(LoggingMixin):
         url = _url_from_endpoint(self.base_url, endpoint)
         merged_headers = {**(self.headers or {}), **(headers or {})}
         extra_options = {**(self.extra_options or {}), **(extra_options or {})}
+        # Track which headers came from the connection so they can be dropped
+        # on a cross-host redirect (aiohttp only strips Authorization).
+        connection_header_names = set(self.headers.keys()) if self.headers else set()
 
         async def request_func() -> ClientResponse:
-            response = await self._request(
-                url,
-                params=data if self.method == "GET" else None,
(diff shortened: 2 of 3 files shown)


kaxil:

PR #70643, opened 2026-07-28
Run UI compile checks in CI when only an OpenAPI spec changes

shahar1:
#68919 changed `_private_ui.yaml` without regenerating the UI TypeScript client and CI stayed green: selective checks decide whether to skip the `ts-compile-lint-*` hooks from file groups that did not cover the UI codegen input specs, while the hooks' own `files:` triggers match any yaml under the openapi directories. The stale committed client then masked a type error that broke `main` for every PR running full static checks (fixed by #70637). The same gap applied to `v2-simple-auth-manager-generated.yaml`: a spec-only change also skipped the simple auth manager UI hook.

This tracks the codegen-input specs in their own `UI_OPENAPI_FILES` group, used only in the hook-skip rule — deliberately **not** in `API_CODEGEN_FILES`, which would force the full test matrix for every UI-endpoint change. It also removes the `ts-compile-lint-ui` trigger for the simple auth manager spec: that pattern was copied in when the hook was created (#51725), a month after the spec was renamed v1→v2 (#50705), so it has never matched a file — #53636 fixed the sibling hook's copy but missed this one — and the main UI codegen does not consume that spec (`ui/openapi-merge.json`).

Verified: replaying #68919's exact file list now runs both hooks with `full-tests-needed=false`; differential `SelectiveChecks` runs across 50+ file-set/event/branch scenarios show the only changed output anywhere is `skip-prek-hooks`, and only for the spec files. The regression test pins each group pattern individually (mutation-tested), plus a targeted `not in` assertion that survives mechanical refreshes of the skip-list constants.

related: #68919

> [!NOTE]
> Land #70637 first: this PR itself triggers full static checks (`dev/breeze` is in `ENVIRONMENT_FILES`), which run the UI hook against the currently stale client on `main`.

Review notes (three independent review rounds; remaining items are cosmetic, taken or left at maintainer discretion): the doc parenthetical "a generated spec still does" could name the REST API spec explicitly; "superset of the first hook's triggers" in the group comment could carry an extra "openapi" qualifier; a prek hook asserting the hook triggers stay ⊆ `UI_FILES ∪ UI_OPENAPI_FILES` would guard this drift class permanently and is left as a possible follow-up.

---

### DIFF
--- a/airflow-core/.pre-commit-config.yaml
+++ b/airflow-core/.pre-commit-config.yaml
@@ -205,8 +205,7 @@ repos:
         files: |
           (?x)
           ^src/airflow/ui/.*\.(js|ts|tsx|yaml|css|json)$|
-          ^src/airflow/api_fastapi/core_api/openapi/.*\.yaml$|
-          ^src/airflow/api_fastapi/auth/managers/simple/openapi/v1.*\.yaml$
+          ^src/airflow/api_fastapi/core_api/openapi/.*\.yaml$
         exclude: |
           (?x)
           ^src/airflow/ui/node-modules/.*|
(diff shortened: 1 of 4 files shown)


shahar1:
CC: @seanghaeli

bbovenzi (approved):

seanghaeli (approved):

amoghrajesh (approved):

pierrejeambrun (approved):
Thanks for fixing it

PR #70651, opened 2026-07-28
Support require_approval in LLMBranchOperator

guan404ming:
## Why

- `require_approval=True` was rejected on `LLMBranchOperator`; this implements it.
- Branching skips tasks irreversibly, so worth a human gate.

## How

- Pause for HITL review; run `do_branch` only after approval.
- Validate reviewer-modified branches against downstream task IDs.
- Reject non-string prompts with `require_approval` before agent runs.

E2E verified in breeze: pause, modified approval, correct skip.

---

### DIFF
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -75,6 +75,30 @@ With multiple branches:
     :start-after: [START howto_decorator_llm_branch_multi]
     :end-before: [END howto_decorator_llm_branch_multi]
 
+Human-in-the-Loop Approval
+--------------------------
+
+Set ``require_approval=True`` to pause the task after the LLM chooses the
+branch(es) and wait for a human reviewer to approve the choice before any
+downstream task is skipped. When ``allow_modifications=True``, the reviewer
+can also change the choice — the modified branch(es) are validated against
+the downstream task IDs before branching. With
+``allow_multiple_branches=True`` the reviewed value is a JSON list of task
+IDs (e.g. ``["task_a", "task_b"]``):
+
+.. code-block:: python
+
+    branch = LLMBranchOperator(
+        task_id="route",
+        prompt="Route this support ticket: {{ dag_run.conf['message'] }}",
+        llm_conn_id="openai_default",
+        require_approval=True,
+        allow_modifications=True,
+    )
+
+``approval_timeout`` and the rest of the approval behaviour are inherited
+from :ref:`LLMOperator <howto/operator:llm>`.
+
 How It Works
 ------------
 
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -92,6 +92,14 @@ against the downstream task IDs before branching:
     :start-after: [START howto_operator_llm_branch_approval]
     :end-before: [END howto_operator_llm_branch_approval]
 
+Rejecting the review, or letting ``approval_timeout`` expire, **fails** the
+task (``HITLRejectException`` / ``HITLTimeoutError``), so downstream tasks
+end up ``upstream_failed`` rather than skipped.
+
+``require_approval=True`` requires a string prompt: a decorated callable
+returning a ``Sequence[UserContent]`` raises ``TypeError`` before the LLM
+call.
+
 ``approval_timeout`` and the rest of the approval behaviour are inherited
 from :ref:`LLMOperator <howto/operator:llm>`.
 
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -80,21 +80,17 @@ Human-in-the-Loop Approval
 
 Set ``require_approval=True`` to pause the task after the LLM chooses the
 branch(es) and wait for a human reviewer to approve the choice before any
-downstream task is skipped. When ``allow_modifications=True``, the reviewer
-can also change the choice — the modified branch(es) are validated against
-the downstream task IDs before branching. With
-``allow_multiple_branches=True`` the reviewed value is a JSON list of task
-IDs (e.g. ``["task_a", "task_b"]``):
+downstream task is skipped. The review form shows the LLM's choice and the
+valid downstream task IDs. When ``allow_modifications=True``, the reviewer
+can also change the choice — rendered as a dropdown of the downstream task
+IDs, or a free-text JSON list of task IDs (e.g. ``["task_a", "task_b"]``)
+with ``allow_multiple_branches=True``. The reviewed branch(es) are validated
+against the downstream task IDs before branching:
 
-.. code-block:: python
-
-    branch = LLMBranchOperator(
-        task_id="route",
-        prompt="Route this support ticket: {{ dag_run.conf['message'] }}",
-        llm_conn_id="openai_default",
-        require_approval=True,
-        allow_modifications=True,
-    )
+.. exampleinclude:: /../../ai/src/airflow/providers/common/ai/example_dags/example_llm_branch.py
+    :language: python
+    :start-after: [START howto_operator_llm_branch_approval]
+    :end-before: [END howto_operator_llm_branch_approval]
 
 ``approval_timeout`` and the rest of the approval behaviour are inherited
 from :ref:`LLMOperator <howto/operator:llm>`.
@@ -123,6 +119,12 @@ Parameters
   task ID. When ``True`` the LLM may return one or more task IDs.
 - ``agent_params``: Additional keyword arguments passed to the pydantic-ai ``Agent``
   constructor (e.g. ``retries``, ``model_settings``). Supports Jinja templating.
+- ``require_approval``: If ``True``, the task pauses after the LLM chooses the
+  branch(es) and waits for human review before branching.  Default ``False``.
+- ``approval_timeout``: Maximum time to wait for a review (``timedelta``).  ``None``
+  means wait indefinitely.  Default ``None``.
+- ``allow_modifications``: If ``True``, the reviewer can change the chosen
+  branch(es) before approving.  Default ``False``.
 
 Logging
 -------
--- a/providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
@@ -108,6 +108,11 @@ class LLMBranchOperator(LLMOperator, BranchMixIn):
         else:
             branches = str(output)
 
+        if not branches:
+            raise ValueError(
+                f"LLM selected no branches for {self.task_id!r}, which would skip every downstream task."
+            )
+
         if self.require_approval:
             choices = sorted(self.downstream_task_ids)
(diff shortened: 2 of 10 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -95,4 +111,38 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
         else:
             branches = str(output)
 
+        if self.require_approval:
+            self.defer_for_approval(context, branches)  # type: ignore[misc]
```
kaxil: The reviewer never sees the branches they are choosing between. The default body is ```Prompt: <prompt>\n\n<chosen>``` and the modification field is free text (`{"type": "string"}`), so with `allow_modifications=True` they have to type a task ID from memory. I tried `refund` against an operator with `{refunds, billing}` downstream and got `ValueError: Reviewed branch(es) ['refund'] are not downstream tasks` at resume, after the approval was already recorded, so the retry burns a second LLM call and needs a fresh review. `defer_for_approval` accepts `body=`, so passing one that lists `sorted(self.downstream_task_ids)` covers most of it. Constraining the param instead with `"enum": sorted(self.downstream_task_ids)` would be better (`Param.resolve` runs jsonschema, and `FlexibleForm` renders enum params as a dropdown), but that needs a params override on the mixin.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -95,4 +111,38 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
...
+            return output
+        try:
+            branches = json.loads(output)
+        except json.JSONDecodeError as e:
+            raise ValueError(
+                f"Reviewed output {output!r} is not valid JSON. With "
+                f"allow_multiple_branches=True the reviewed output must be a "
+                f'JSON list of task IDs, e.g. ["task_a", "task_b"].'
+            ) from e
+        if not isinstance(branches, list) or not all(isinstance(b, str) for b in branches):
```
kaxil, line 156: This catches the wrong type and unknown task IDs, but not `[]`. A reviewer editing the field to `[]` passes both checks, and `do_branch(context, [])` lands in `skip_all_except` with an empty branch set, which skips every downstream task. I ran it to confirm the empty list flows straight through to `do_branch`. Given the premise that irreversible skipping is what needs the human gate, that seems like the one value to reject here.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -60,12 +70,18 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         kwargs.pop("output_type", None)
-        if kwargs.get("require_approval"):
-            raise ValueError("require_approval=True is not supported by LLMBranchOperator.")
         super().__init__(**kwargs)
         self.allow_multiple_branches = allow_multiple_branches
 
     def execute(self, context: Context) -> str | Iterable[str] | None:
+        if self.require_approval and not isinstance(self.prompt, str):
```
kaxil: This is now the third verbatim copy of the check: `LLMOperator.execute` has the same block, and `LLMApprovalMixin.defer_for_approval` raises its own version. The operator-level copies exist to fail before the LLM call, so would it make sense to put the pre-flight on the mixin and call it from both `execute()` bodies?

providers/common/ai/docs/operators/llm_branch.rst
```diff
@@ -75,6 +75,30 @@ With multiple branches:
...
+
+Set ``require_approval=True`` to pause the task after the LLM chooses the
+branch(es) and wait for a human reviewer to approve the choice before any
+downstream task is skipped. When ``allow_modifications=True``, the reviewer
+can also change the choice — the modified branch(es) are validated against
+the downstream task IDs before branching. With
+``allow_multiple_branches=True`` the reviewed value is a JSON list of task
+IDs (e.g. ``["task_a", "task_b"]``):
+
+.. code-block:: python
```
kaxil: Every other section on this page uses `exampleinclude` from `example_llm_branch.py`, and both `example_llm.py` and `example_llm_sql.py` already carry a `require_approval=True` example. Adding one there and including it here would keep the page consistent and give the feature example-DAG coverage. The Parameters section below also doesn't list `require_approval` / `approval_timeout` / `allow_modifications`, which `llm.rst` does.

kaxil:
Screenshot/video of how it works e2e would be ideal

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Done — body now lists valid branches; single-branch mode renders an enum dropdown via `modification_schema`.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Make sense to me, the empty reviewed lists now raise `ValueError` before `do_branch`, so nothing gets skipped silently.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Done, moved the check to `LLMApprovalMixin.validate_approval_prompt`, called from all three `execute()` bodies.

providers/common/ai/docs/operators/llm_branch.rst
guan404ming: Done, added `require_approval` example DAG, switched to `exampleinclude`, and documented the approval parameters.

guan404ming:
Attached an e2e demo: trigger, review with branch dropdown, approve modified choice, downstream branches run/skip.

[llm_branch_approval_demo.webm](https://github.com/user-attachments/assets/4ed3f230-ecf3-4986-bb5f-5c0453fd1911)

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
kaxil: The guard closes the reviewer path, thanks. It does not cover the LLM returning `[]` itself, though: `output_type = list[downstream_tasks_enum]` carries no minimum length, so an empty selection serializes to `"[]"`, writes a HITL detail reading `Chosen branch(es): []`, and pauses. Every reviewer action then fails the task, Approve via this new guard and Reject via `HITLRejectException`, and with `allow_modifications=False` they cannot fix it at all. Checking emptiness right after `branches` is computed (line 109) would cover both paths, and would also close the `require_approval=False` case, where `do_branch(context, [])` reaches `skip_all_except` with an empty set and skips every downstream task.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
kaxil: The single-branch dropdown is exactly it. Multi-branch mode still lands on free text though -- `modification_schema` is `None` when `allow_multiple_branches=True`, so the reviewer types raw JSON and a typo fails at resume the same way. The UI can already render the better form: `isFieldMultiSelect` in `airflow-core/src/airflow/ui/src/components/FlexibleForm/FieldSelector.tsx:82` selects `FieldMultiSelect` for `{"type": "array", "enum": [...]}`. What blocks it is two lines in the mixin -- `"value": output` stores the JSON string, and `execute_complete` rejects a non-string `params_input["output"]` -- so an array param needs the mixin to accept a list first. Reasonable to leave as a follow-up if you'd rather not widen the mixin contract here.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -95,4 +108,54 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
         else:
             branches = str(output)
 
+        if self.require_approval:
+            choices = sorted(self.downstream_task_ids)
+            chosen = branches if isinstance(branches, str) else json.dumps(branches)
+            body = (
+                f"```\nPrompt: {self.prompt}\n\nChosen branch(es): {chosen}\n```\n\n"
+                f"Valid branches: {', '.join(f'`{c}`' for c in choices)}"
```
kaxil: `prompt` is a template field, so this body routinely carries rendered external data inside the fence, while the `Valid branches:` list sits outside it. A prompt containing a closing fence ends the block early and can print its own `Valid branches:` line, forging the list the reviewer is meant to trust. The enum dropdown and the `downstream_task_ids` check still constrain what can actually be submitted, so this misleads rather than mis-routes, but the mixin's default body keeps everything fenced (`approval.py:132`), and moving the list above the fence would keep it that way.

providers/common/ai/docs/operators/llm_branch.rst
```diff
@@ -75,6 +75,26 @@ With multiple branches:
...
+IDs, or a free-text JSON list of task IDs (e.g. ``["task_a", "task_b"]``)
+with ``allow_multiple_branches=True``. The reviewed branch(es) are validated
+against the downstream task IDs before branching:
+
+.. exampleinclude:: /../../ai/src/airflow/providers/common/ai/example_dags/example_llm_branch.py
+    :language: python
+    :start-after: [START howto_operator_llm_branch_approval]
+    :end-before: [END howto_operator_llm_branch_approval]
+
+``approval_timeout`` and the rest of the approval behaviour are inherited
```
kaxil, line 103: Two things missing from this section. Reject and a timeout both fail the task, because `LLMApprovalMixin.execute_complete` raises `HITLRejectException` / `HITLTimeoutError`, so declining the LLM's routing leaves downstream tasks `upstream_failed` rather than skipped, the opposite of `ApprovalOperator`, whose docstring calls failing on reject "generally discouraged" for mixing human decision outcomes with task states. And line 67 tells the reader the callable may return a `Sequence[UserContent]` directly above a section telling them to set `require_approval=True`, with no note that the combination raises `TypeError`, which `llm.rst:162-164` does carry. `do_branch(context, None)` already means skip-all-downstream in `BranchMixIn` if you want the reject path to match.

kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Make sense to me, just moved the list above the fence.

providers/common/ai/docs/operators/llm_branch.rst
guan404ming: Fixed, documented both. Kept reject failing the task for now and will follow up to align with ApprovalOperator.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Fixed, empty selection now raises right after branches are computed, covering both paths.

guan404ming:
Thanks for the review, I'll merge this first and will follow up to refine this.

PR #66633, opened 2026-05-09
Fix remote processor injection happening before dictConfig runs in configure_logging

korex-f:
## Problem

In `configure_logging()`, `getattr(remote, "processors")` was called before `dictConfig()` ran. `dictConfig()` calls `_clearExistingHandlers()` which closes every handler registered in `logging._handlerList`, including  the watchtower handler built moments earlier. This caused CloudWatch/Watchtower logs to be silently dropped when using the Task SDK with remote logging.

## Fix

Move the remote processor injection to after `dictConfig()` has run. The handler is now built into a clean logging state and cannot be killed prematurely.

Also gate the injection on `not sending_to_supervisor` to avoid unmasked events from the task subprocess, and add a `None` default to `getattr` to handle third-party `RemoteLogIO` objects missing the `processors` attribute.

## Related

Provider-side fix: #68779

closes: #66475

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @korex-f** · by `@potiuk` · 2026-07-02 17:46 UTC
>
> Some review feedback from `@jason810496` is waiting on you:
> - 6 unresolved review thread(s) from `@jason810496` need a reply or a fix.
>
> **The ball is in your court** — you've been assigned to this PR. Reply or push a fix in each thread, then mark them resolved.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #70653, opened 2026-07-29
Add resumable_job.reconnect_failure metric for failed reconnects

1fanwang:
A ResumableJobMixin retry increments resumable_job.reconnect_attempt and then one
of reconnect_success / already_succeeded / terminal_resubmit, but only once the
prior job's status has been fetched and classified. When that status check raises
(external system unreachable, expired id, and so on) the attempt resolves to no
outcome: the failure is silent and reconnect_attempt no longer reconciles with its
outcome counters. Reconnection failure is the signal an operator most wants to
alert on, yet today it cannot be measured.

resumable_job.reconnect_failure is emitted when the reconnect decision raises, so
every reconnect_attempt resolves to exactly one outcome.

## Testing

Unit: a parametrized regression test covers the status-fetch and status-classification
raise paths (fails on main, passes here); the test_resumablejobmixin.py suite covers
the regression.

E2E: drove the real ResumableJobMixin against a real StatsD -> statsd-exporter ->
Prometheus -> Grafana stack. A scenario of 4 active + 3 succeeded + 2 terminal + 5
raising reconnects, plus 6 fresh submits, produced in Prometheus:

    airflow_resumable_job_reconnect_attempt   14
    airflow_resumable_job_reconnect_success    4
    airflow_resumable_job_already_succeeded    3
    airflow_resumable_job_terminal_resubmit    2
    airflow_resumable_job_reconnect_failure    5
    airflow_resumable_job_fresh_submit         6

reconnect_attempt (14) == reconnect_success + already_succeeded + terminal_resubmit
+ reconnect_failure (4 + 3 + 2 + 5). Without this change the sum is 9 and the
attempts do not reconcile.


task-sdk/src/airflow/sdk/bases/resumablejobmixin.py
```diff
@@ -149,42 +149,50 @@ def execute_resumable(self, context: Context) -> Any:
...
+                            # Job is in a terminal failed state, fall through and submit a new job.
+                            span.set_attribute("resumable.decision", "terminal_resubmit")
+                            stats.incr("resumable_job.terminal_resubmit", tags=stats_tags)
+                            self.log.warning(
+                                "Prior job in terminal state, resubmitting fresh",
+                                external_id_key=self.external_id_key,
+                                external_id=external_id,
+                                status=status,
+                            )
+                    except Exception:
```
potiuk, line 189: The `try` starts at line 152 and wraps the branch bodies, not just the decision — so an outcome counter and this one can both fire for a single attempt.

Each branch increments and *then* logs: `:161` → `:162`, `:172` → `:173`, `:182` → `:183`. If one of those logging calls raises, the outcome has already been counted and this `except` adds `reconnect_failure` on top, giving `reconnect_attempt` 1 and outcomes 2. That's the reconciliation your E2E table demonstrates, and it's the thing this PR exists to establish. Line 193 has the same problem in the trace: it overwrites a `resumable.decision` that was already set to `reconnect` / `already_succeeded` / `terminal_resubmit`.

Worth saying why I don't think this is academic: `self.log` is structlog, and remote log processors are injected into the global chain — #66633 is currently open about exactly that path. A logging call failing because a remote sink is unreachable is the same class of event this metric is meant to catch.

Narrowing the `try` to the fetch and classification, then branching on the result outside it, makes the exclusivity structural:

```python
try:
    status = self.get_job_status(external_id, context)
    if self.is_job_active(status):
        decision = "reconnect"
    elif self.is_job_succeeded(status):
        decision = "already_succeeded"
    else:
        decision = "terminal_resubmit"
except Exception:
    span.set_attribute("resumable.decision", "reconnect_failure")
    stats.incr("resumable_job.reconnect_failure", tags=stats_tags)
    raise
```

task-sdk/tests/task_sdk/bases/test_resumablejobmixin.py
```diff
@@ -296,6 +296,26 @@ def test_terminal_resubmit_fires_when_job_failed(self):
         assert "resumable_job.reconnect_success" not in called_names
         assert "resumable_job.fresh_submit" not in called_names
 
+    @pytest.mark.parametrize("failing_method", ["get_job_status", "is_job_active"])
```
potiuk, line 299: Both parameters here raise *before* any outcome counter fires, so they're the two positions where double-counting can't happen — the test passes for the same reason the bug hides.

Two additions worth having: `"is_job_succeeded"` (the third classification call, at `resumablejobmixin.py:169`) is one more token in this list; and a case where the *log* call raises after an outcome was counted would pin the point in my comment on line 189 — it fails today, and passes once the `try` is narrowed.

potiuk (commented):
Nice piece of work — the gap is real, and the E2E reconciliation table is better evidence than most metrics PRs bring. Four active, three succeeded, two terminal, five raising, and the attempt counter balancing only with the new metric is exactly the right way to demonstrate it.

One thing I'd like resolved, and it's aimed squarely at that invariant: **the `try` is wider than it needs to be**, so a logging call failing after an outcome counter has already fired would emit `reconnect_failure` as well, and `reconnect_attempt` would no longer reconcile. Details and a suggested narrowing inline on `resumablejobmixin.py:189`. The same widening makes the span's `resumable.decision` attribute overwrite a decision that was already recorded.

The test can't currently catch this: both parametrized cases raise before any outcome counter fires — see `test_resumablejobmixin.py:299`.

Nothing else from me. The YAML entry matches its siblings and the registry hook is green; no newsfragment needed, since none of the five existing `resumable_job.*` metrics has one; and the test follows the file's own patching convention, so I've no style notes. Note the branch is 876 commits behind `main`, so it'll want a rebase before it merges.

---

>

---

PR #70656, opened 2026-07-29
Fix PsrpOperator constructor validation to use provision checks inste…

bujjibabukatta:
## Summary

Part of the burn-down tracked in #70296.

`PsrpOperator.__init__` validated `command`/`powershell`/`cmdlet`/`arguments`/`parameters`
using truthiness checks instead of `is not None` checks. Per the guidance in #70296,
"was this provided" checks are allowed to stay in `__init__`, but must use `is not None`
polarity rather than truthiness — so this is a fix in place, not a move to `execute()`.

## What was wrong

- `command`/`powershell`/`cmdlet` were collected into a `set` before being passed to
  `exactly_one()`. If two of them happened to render to the same string, the set would
  collapse them and validation would incorrectly pass.
- `if arguments and ...`, `if parameters and ...`, `if cmdlet:` treated an explicitly
  provided empty list/dict/string as "not provided", silently skipping validation.

## Fix

Replaced truthiness checks with `is not None` checks, and pass fields to `exactly_one()`
directly instead of via a `set`. `__init__` is now limited to this provision check plus
plain `self.field = field` assignments.

## Testing

Added three regression tests in `test_psrp.py` covering: two duplicate field values,
an empty `arguments` list, and an empty `parameters` dict — all now correctly raise
`ValueError`.

Related: #70296

- [X] Yes - Claude


shahar1:
Please note that it's a competitor PR of https://github.com/apache/airflow/pull/70347 (you may keep this open, but I might end up merging the other - depending which provides the best match first)

scripts/in_container/run_generate_constraints.py
eladkal, line 1: Why was this file modified?

potiuk:
Thanks — your diagnosis is right, and the fix is correct: `args = {command, powershell, cmdlet}` built a set (collapsing equal values) and `exactly_one` counted truthiness rather than presence, so the `is not None` rewrite is the proper correction.

It duplicates #70347 though, which makes the identical change to the same two files and was opened on 23 July, six days before this one. That PR also carries the fix through `execute()` and removes the operator from `validate_operators_init_exemptions.txt`, so it's the more complete version — I'm closing this in its favour and have asked there for the one extra behavioural change it contains to be split out or documented.

One unrelated observation while I'm here: this PR also touches `scripts/in_container/run_generate_constraints.py`, which doesn't appear connected to the PsrpOperator fix. If that was an accidental inclusion it's worth checking your branch; if it's a real fix it deserves its own PR where it'll actually get reviewed rather than riding along unnoticed.

Worth a quick `gh pr list --search "<operator>"` before starting on one of these — several of the `validate_operators_init_exemptions.txt` entries have had two people working on them independently this week.

---

PR #69750, opened 2026-07-11
Add Task SDK, Go and Java SDK execution architecture diagrams

potiuk:
Adds graphviz-generated diagrams to the architecture overview
(core-concepts/overview.rst) explaining how a task actually runs across
Airflow's SDKs, in a new "Task execution architecture" section:

- Python Task SDK — the supervised path (Supervisor + forked task process over a
  msgpack socket) vs the native in-process path (dag.test()), plus a numbered
  message-flow sequence.
- Go SDK — standalone airflow-go-edge-worker pulling from the Edge API, running
  the compiled bundle as a go-plugin gRPC subprocess and calling the Execution
  API directly.
- Java (JVM) SDK — the new Coordinator layer (CoordinatorManager -> JavaCoordinator
  -> _JavaActivitySubprocess) reusing the Python Supervisor, driving a JVM
  subprocess over loopback TCP, with an architecture diagram and a numbered
  execution-workflow sequence.

All diagrams are generated by the existing generate-airflow-diagrams prek hook
(committed as .py + .png + .md5sum).

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70657, opened 2026-07-29
Split task execution architecture docs: concise overview + dev guide

jason810496:
- Follow-up to #69750; addresses review feedback (https://github.com/apache/airflow/pull/70657#issuecomment-5115032700).

## Why

The task execution architecture material added in #69750 mixed two audiences on the user-facing `core-concepts/overview.rst` page: a high-level picture useful to deployment managers, and deep worker internals that only contributors need.

## How

- Keep a concise, deployment-oriented **Task execution architecture** section in `core-concepts/overview.rst` with a single lightweight diagram showing that **each TaskInstance on the worker is a brand new subprocess**.
- Move the full developer-level detail (per-runtime execution paths, the Go edge worker, and the sequence diagrams) into a new dev-facing page `contributing-docs/31_task_execution_architecture.rst`, and extend the `generate-airflow-diagrams` prek hook to cover `contributing-docs/images/`.

---

- [x] Yes, with help of Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


uranusjr (changes requested):
I don’t think we want to put coordinator first. I intentionally put the concept to the back because the user does not need to care how it works, only how to use it. The new writeup emphasis on internals too much. It should instead focus on introducing how the coordinator can be used (by configuration) instead.

jason810496:
> I don’t think we want to put coordinator first. I intentionally put the concept to the back because the user does not need to care how it works, only how to use it. The new writeup emphasis on internals too much. It should instead focus on introducing how the coordinator can be used (by configuration) instead.

Hmmm, then I will need to move the whole paragraph introduce in https://github.com/apache/airflow/pull/69750 to somewhere else.

All the diagrams and descriptions are quite verbose for the users IMO -- http://apache-airflow-docs.s3-website.eu-central-1.amazonaws.com/docs/apache-airflow/stable/core-concepts/overview.html#non-python-language-sdks-go-and-java

jason810496:
Then would it be better to move them all (starting from "Task execution architecture" section) to `airflow-core/docs/core-concepts/task-execution-archciture.rst` as a new page instead of overwhelming the `core-concepts/overview.rst` page?

uranusjr:
I think they do not belong in user-facing documentation. It would fit if we have a section for advanced concepts or something, but we currently don’t. They do fit in dev-facing docs though (`contributing-docs`).

jason810496:
I just moved them all to `contributing-docs/31_task_execution_architecture.rst` (a new page under contributing-docs that don't make the new task-sdk one too large).

potiuk:
To be honest, I do not think those are "contributing" docs. Those are "deployment" docs that explain what is the interaction between different components, what processes are running and what to expect.

One of the big issues it explains is what is teh resource usage connected and what processes you can expect to see running when you use the SDK. I think the deployment managers should know that in some cases they can just run a long running process that does everything (golang) , and it some other cases they will have python processes taking 100s of MBs for each task run - this is far more deployment information than contributing documentation IMHO,

potiuk (dismissed):
I think we should first agree who is the audience of that docs.

IMHO. Deployment Managers - not contributors.

jason810496 (commented):
Hi @potiuk, @uranusjr 

I added a new `airflow-core/docs/img/diagram_task_execution_architecture.png` diagram that only focus on the worker process, how the coordinator interact with the Python / Lang-SDKs subprocesses without too much internal detail back to the overview architecture for the deployment manager perspative.

Then move all the existing diagrams that focus on the internal details to `contributing-docs/31_task_execution_architecture.rst`.

Please let me know what do you all think, thanks.

airflow-core/docs/core-concepts/overview.rst
```diff
@@ -192,90 +192,35 @@ code is never executed in the context of the *scheduler*.
...
+touches the database.
+
+Inside the supervisor, a ``CoordinatorManager`` selects a *coordinator* by matching the
+task's ``TaskInstance.queue``. Ordinary Python tasks run in the built-in coordinator under
+the same supervisor, with no separate language process. Non-Python tasks are routed to a
+language coordinator — a **Java** coordinator for JVM bundles, or an **Executable**
+coordinator for self-contained native bundles (for example, Go) — which launches the task's
+runtime as a **language SDK subprocess**. That subprocess talks back to the supervisor over
+a **MsgPack comm frame** protocol, so from the *Execution API*'s point of view a
+non-Python task behaves exactly like a Python one.
```
uranusjr: I’m still not sure this belongs to user-facing documentation. Nobody cares about these, and they shouldn’t need to.

airflow-core/docs/img/diagram_task_execution_architecture.png
uranusjr, line 1: This does not look right. The default behavior is to use Python, so “other queue” is either incorrect or poorly expressed.

airflow-core/docs/core-concepts/overview.rst
```diff
@@ -192,90 +192,35 @@ code is never executed in the context of the *scheduler*.
...
+
+When a *worker* runs a task it does not execute the user's code directly. It first starts a
+**supervisor** process. The supervisor is the only side that holds the task's short-lived
+credentials and talks to the *Execution API* — it proxies every Connection, Variable and
+XCom lookup and every heartbeat on the task's behalf, so the code that runs the task never
+touches the database.
+
+Inside the supervisor, a ``CoordinatorManager`` selects a *coordinator* by matching the
+task's ``TaskInstance.queue``. Ordinary Python tasks run in the built-in coordinator under
+the same supervisor, with no separate language process. Non-Python tasks are routed to a
```
uranusjr: Python tasks still has a separate process (forked), so this is wrong. But again, the entire description is unnecessary.

jason810496 (commented):
Thanks for the review, I rephrased again in https://github.com/apache/airflow/pull/70657/changes/206206440d36ad6e31aeabf46695ad5ece08509b.

I zoom out the whole paragraph and the diagram again to focus on the concept of "For every task instance it starts a **new subprocess**". The user-facing docs no longer mentioned the queue, coordinator, or the msg pack at all.

contributing-docs/README.rst
```diff
@@ -143,6 +143,10 @@ You can also dive deeper into more specific areas that are important for contrib
 * `Architecture diagram <17_architecture_diagrams.rst>`__ describes how to create and
   update the architecture diagrams embedded in Airflow documentation.
 
+* `Task execution architecture <31_task_execution_architecture.rst>`__ explains what happens inside a
+  worker when a task runs: the Python Supervisor and task runner, the Coordinator layer shared by the Java
+  and Go SDKs, and the standalone Go edge worker.
```
uranusjr: Do we want to move this lower? The numbers referenced around here are 17, **31**, 19, 20.

airflow-core/docs/core-concepts/overview.rst
```diff
@@ -192,90 +192,30 @@ code is never executed in the context of the *scheduler*.
...
-As above, the **Supervisor** is the central lifeline, so the JVM ↔ Supervisor round-trip over loopback TCP is
-drawn as arrows going back and forth to its neighbours:
+.. image:: ../img/diagram_task_execution_architecture.png
 
-.. image:: ../img/diagram_java_sdk_execution_sequence.png
+Nothing is pooled or reused between task instances, so *N* concurrent task instances cost *N* subprocesses plus anything the task code itself spawns.
+Sizing a worker therefore means budgeting for the peak number of task instances it runs at once, not for the worker process alone.
 
-.. note::
+The worker process is also the only side that holds the task's short-lived credentials and talks to the *Execution API*, so the subprocess running the task code never reaches the metadata database directly, in any language.
```
uranusjr, line 216: Not sure if we want to add a note here saying the Edge Worker is another story. (It does not follow the same rules even in Python.)

contributing-docs/README.rst
jason810496: Good point, I add a new Task Execution Lifecycle section at end of the doc in https://github.com/apache/airflow/pull/70657/changes/26e5dc78b4cb39472ec7c16cc420e0068aa158ac, thanks.

airflow-core/docs/core-concepts/overview.rst
jason810496: As discussed with Ash offline that we will remove Go Edge Worker before Go SDK GA so I think it's fine to not mention the Edge Worker at all.

uranusjr (approved):

PR #70658, opened 2026-07-29
Align timetable docs with Airflow 3 data interval and catchup behaviour

Tegh25:
Update the trigger vs data-interval comparison in `timetable.rst` so it matches Airflow 3 defaults and actual catchup / `next_dagrun_info` behaviour (aligned with the corrected Dag Run docs from #70167).

- Document that bare `schedule="@daily"` / `timedelta(...)` resolve to trigger timetables when `create_cron_data_intervals` / `create_delta_data_intervals` are False.
- Describe trigger timetables as zero-width data intervals by default (optional non-zero interval=), instead of as “no data interval.”
- Clarify `catchup=False` picks the most recent applicable tick / completed interval instead of backfilling missed runs; catchup=True creates every missed run sequentially.
- Fix first-enable and pause/re-enable @daily examples for both timetable kinds.
- Correct logical_date vs run_id: logical_date is data_interval_start; the run_id timestamp comes from run_after.

This PR comes after a discussion with @ferruzzi on catchup behaviour.

---

- [X] Yes — Cursor (Auto)

Generated-by: Cursor (Auto) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/docs/authoring-and-scheduling/timetable.rst
+++ b/airflow-core/docs/authoring-and-scheduling/timetable.rst
@@ -390,8 +390,8 @@ January 31st,
   tick — midnight on January 31st. ``logical_date``, ``data_interval_start``,
   ``data_interval_end``, and the ``run_id`` timestamp are all that midnight.
 - `CronDataIntervalTimetable`_ immediately creates a Dag run for the most
-  recently completed interval (midnight January 30th through midnight
-  January 31st). ``logical_date`` / ``data_interval_start`` are January 30th;
+  recently completed interval (midnight January 30 through midnight
+  January 31st). ``logical_date`` / ``data_interval_start`` are January 30;
   ``run_after`` and the ``run_id`` timestamp are midnight on January 31st.
 
 The following is another example showing the difference when skipping Dag runs:
--- a/airflow-core/docs/authoring-and-scheduling/timetable.rst
+++ b/airflow-core/docs/authoring-and-scheduling/timetable.rst
@@ -313,10 +313,17 @@ Airflow has two sets of timetables for cron and delta schedules:
 
 In Airflow 3, a bare cron string such as ``@daily`` in ``schedule=`` resolves to
 CronTriggerTimetable_ by default (``[scheduler] create_cron_data_intervals`` is
-``False``). A bare ``timedelta`` resolves to DeltaTriggerTimetable_ when
-``[scheduler] create_delta_data_intervals`` is ``False``. Set either flag to
-``True``, or pass an explicit data-interval timetable class, to get contiguous
-windows instead.
+``False``). A bare ``timedelta`` resolves to DeltaTriggerTimetable_ by default
+(``[scheduler] create_delta_data_intervals`` is also ``False``). Pass an explicit
+data-interval timetable class, or set ``create_cron_data_intervals`` to ``True``,
+to get contiguous windows instead.
+
+.. note::
+
+    ``[scheduler] create_delta_data_intervals`` is intended to control timedelta
+    schedules independently, but is not currently consulted. Those schedules
+    follow ``create_cron_data_intervals`` as well, so setting only
+    ``create_delta_data_intervals=True`` still yields DeltaTriggerTimetable_.
 
 - A trigger timetable (CronTriggerTimetable_ or DeltaTriggerTimetable_) represents
   each run as a point in time: by default ``data_interval_start`` and
@@ -345,7 +352,7 @@ of the contiguous window. ``logical_date`` is ``data_interval_start`` for both
 kinds.
 
 *Catchup* behavior
-^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~
 
 By default, ``catchup`` is ``False`` (Airflow config
 ``[scheduler] catchup_by_default``). Missed scheduled run times between
@@ -359,10 +366,11 @@ scheduled run time:
 - For a data-interval timetable, the most recently completed interval whose end
   is not after "now".
 
-If you set ``catchup=True``, the scheduler creates a Dag run for every scheduled
-run time (or completed interval) between ``start_date`` and "now" that has not
-yet run (or has been cleared). Runs are created in chronological order and may
-run concurrently up to the Dag's ``max_active_runs`` (defaulted from :ref:`config:core__max_active_runs_per_dag`).
+If you set ``catchup=True``, the scheduler creates Dag runs from the latest
+automated run (or ``start_date`` if none) forward to "now", one scheduled run
+time (or completed interval) at a time in chronological order. Those runs may
+execute concurrently up to the Dag's ``max_active_runs`` (defaulted from
+:ref:`config:core__max_active_runs_per_dag`).
 
 Catchup also applies when you pause a Dag for a period and then re-enable it.
 
@@ -376,7 +384,9 @@ The time when a Dag run is triggered
 Both trigger and data interval timetables can create the first Dag run
 immediately when ``catchup=False`` and ``start_date`` is in the past. What
 differs is *which* run is selected and how ``logical_date`` / ``run_id`` are
-derived.
+derived. Without a ``start_date`` (optional when ``catchup=False``), a trigger
+timetable with the default ``run_immediately=False`` waits for the next future
+tick instead — midnight on February 1st in the example below.
 
 ``logical_date`` is always ``data_interval_start``. The timestamp embedded in
 ``run_id`` comes from ``run_after`` (when the run is eligible to start). For a
--- a/airflow-core/docs/authoring-and-scheduling/timetable.rst
+++ b/airflow-core/docs/authoring-and-scheduling/timetable.rst
@@ -353,8 +353,9 @@ By default, ``catchup`` is ``False`` (Airflow config
 re-enabled. The timetable instead selects the most recently applicable
 scheduled run time:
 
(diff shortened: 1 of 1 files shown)


ferruzzi:
@kaxil had some good insight on the related PR and may want a look at this one, too.

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,57 +311,108 @@ Airflow has two sets of timetables for cron and delta schedules:
...
+  ``data_interval_end`` are the same (the trigger time). You can optionally pass a
+  non-zero ``interval=`` so the data interval ends at the trigger time and spans
+  that duration. A data-interval timetable (CronDataIntervalTimetable_ or
+  DeltaDataIntervalTimetable_) always uses a contiguous non-zero window between
+  consecutive schedule boundaries.
+- ``logical_date`` and the timestamp used in ``run_id`` differ between the two
+  kinds based on how they handle the data interval, as described in
+  :ref:`timetables_run_id_logical_date`.
 
 Whether taking care of *Data Interval*
```
ferruzzi: Section header is both awkwardly phrased and now kind of misleading.  What do you think of "Data Interval Shape" or something like that?

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,57 +311,108 @@ Airflow has two sets of timetables for cron and delta schedules:
...
+- For a trigger timetable, typically the latest cron tick (or delta boundary)
+  that is not after "now" and not before ``start_date``.
+- For a data-interval timetable, the most recently completed interval whose end
+  is not after "now".
 
-In these scenarios, the ``logical_date`` in the ``run_id`` are based on how the timetable handles the data
-interval.
+If you set ``catchup=True``, the scheduler creates a Dag run for every scheduled
+run time (or completed interval) between ``start_date`` and "now" that has not
+yet run (or has been cleared), and executes them sequentially.
```
ferruzzi: " and executes them sequentially."

Can you verify this? There is a `max_active_runs_per_dag` config option which implies they are concurrent, not sequential, unless catchup jobs are handled differently from other run types?

Tegh25: You're right, sequentially is not the right word to use here.
I'll also update `dag-run.rst` since I made the same mistake in the other PR.

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,68 +311,121 @@ Airflow has two sets of timetables for cron and delta schedules:
...
-
-A trigger timetable *does not* include *data interval*. This means that the value of ``data_interval_start``
-and ``data_interval_end`` are the same; the time when a Dag run is triggered.
-
-For a data interval timetable, the value of ``data_interval_start`` and ``data_interval_end`` are different.
-``data_interval_end`` is the time when a Dag run is triggered, while ``data_interval_start`` is the start of the interval.
+In Airflow 3, a bare cron string such as ``@daily`` in ``schedule=`` resolves to
+CronTriggerTimetable_ by default (``[scheduler] create_cron_data_intervals`` is
+``False``). A bare ``timedelta`` resolves to DeltaTriggerTimetable_ when
+``[scheduler] create_delta_data_intervals`` is ``False``. Set either flag to
```
kaxil: `create_delta_data_intervals` isn't read anywhere. `_create_timetable` checks `create_cron_data_intervals` in both branches, including the `timedelta`/`relativedelta` one: https://github.com/apache/airflow/blob/0bda2ffece97df4b90484a70dd4c62ff35993fda/task-sdk/src/airflow/sdk/definitions/dag.py#L149-L152. So setting `create_delta_data_intervals=True` on its own still gives you `DeltaTriggerTimetable`, and "set either flag to `True`" doesn't hold. The defaults happen to agree, so the default behaviour you describe is right. It looks like the flag was never wired up (2.11 hardcoded `DeltaDataIntervalTimetable` for deltas), so a code fix is probably the real answer, but the docs shouldn't promise the flag works in the meantime.

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,68 +311,121 @@ Airflow has two sets of timetables for cron and delta schedules:
...
+  not before ``start_date``. For DeltaTriggerTimetable_, pickup time itself —
+  a delta has no wall-clock tick to snap to.
+- For a data-interval timetable, the most recently completed interval whose end
+  is not after "now".
 
-In these scenarios, the ``logical_date`` in the ``run_id`` are based on how the timetable handles the data
-interval.
+If you set ``catchup=True``, the scheduler creates a Dag run for every scheduled
+run time (or completed interval) between ``start_date`` and "now" that has not
+yet run (or has been cleared). Runs are created in chronological order and may
```
kaxil: "or has been cleared" doesn't hold. The next run is derived by advancing from the last automated run (`calculate_dagrun_date_fields(..., reference_run=created_run)`, https://github.com/apache/airflow/blob/0bda2ffece97df4b90484a70dd4c62ff35993fda/airflow-core/src/airflow/jobs/scheduler_job_runner.py#L2601), so clearing an existing run's tasks never makes the scheduler create a run, and a run deleted from the middle of the sequence isn't refilled either. Catchup closes the gap between the latest automated run (or `start_date`) and now, nothing else.

Separately, the `max_active_runs` sentence now contradicts the page you link three lines down: `dag-run.rst` still says "the scheduler will execute them sequentially", and it's also where the "(or has been cleared)" wording comes from. Folding that in here, or a follow-up?

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,68 +311,121 @@ Airflow has two sets of timetables for cron and delta schedules:
...
-Suppose there are two running Dags with a cron expression ``@daily`` or ``0 0 * * *`` that use the two different timetables. If you pause the Dags at 3PM on January 31st and re-enable them at 3PM on February 2nd,
-
-- `CronTriggerTimetable`_ skips the Dag runs that were supposed to trigger on February 1st and 2nd. The next Dag run will be triggered at 12AM on February 3rd.
-- `CronDataIntervalTimetable`_ skips the Dag runs that were supposed to trigger on February 1st only. A Dag run for February 2nd is immediately triggered after you re-enable the Dag.
-
-In these examples, you see how a trigger timetable creates Dag runs more intuitively and similar to what
-people expect a workflow to behave, while a data interval timetable is designed heavily around the data
-interval it processes, and does not reflect a workflow's own properties.
+Both trigger and data interval timetables can create the first Dag run
+immediately when ``catchup=False`` and ``start_date`` is in the past. What
```
kaxil, line 385: This is true with a `start_date`, but `start_date` is optional in Airflow 3 (only required when `catchup=True`). With no `start_date` on the Dag or any task, `restriction.earliest` is `None` and the trigger timetable falls into the `_calc_first_run()` branch instead (https://github.com/apache/airflow/blob/0bda2ffece97df4b90484a70dd4c62ff35993fda/airflow-core/src/airflow/timetables/trigger.py#L98-L106), which with the default `run_immediately` picks the next future tick. For the example below that's midnight February 1st, not January 31st, which is the answer the old text gave. A sentence covering the no-`start_date` case would keep both readers right.

airflow-core/docs/authoring-and-scheduling/timetable.rst
```diff
@@ -310,68 +311,121 @@ Airflow has two sets of timetables for cron and delta schedules:
...
+earlier.
+
+For a data interval timetable, the values of ``data_interval_start`` and
+``data_interval_end`` are different. ``data_interval_end`` is the time when a
+Dag run is triggered (``run_after``), while ``data_interval_start`` is the start
+of the contiguous window. ``logical_date`` is ``data_interval_start`` for both
+kinds.
 
 *Catchup* behavior
 ^^^^^^^^^^^^^^^^^^
```
kaxil: Not from your change, but you're reshaping this hierarchy anyway: this underline is `^`, the same level as "Differences between "trigger" and "data interval" timetables", so catchup isn't a subsection of the comparison, and the two `~` sections after it ("The time when a Dag run is triggered" and "Switching between trigger and data interval timetables") end up nested under Catchup in the TOC. Dropping this to `~` puts all four at one level under the comparison section.

airflow-core/docs/authoring-and-scheduling/timetable.rst
Tegh25: I believe it is worth documenting the intended behavior, especially since the PR with the code fix is open and may be merged soon. As a compromise, I reworded the section slightly to not promise anything when `create_delta_data_intervals` is set, only stating its default. I also added the note stating that `create_delta_data_intervals` is currently not read.

airflow-core/docs/authoring-and-scheduling/timetable.rst
Tegh25: I removed "or has been cleared" and tightened the `max_active_runs` sentence. `dag-run.rst` no longer states "the scheduler will execute them sequentially", the "(or has been cleared)" statement in `dag-run.rst` can be followed up after this PR is resolved.

ferruzzi (approved):
Thanks for the changes

Tegh25:
Just need one more review 🙏

PR #70665, opened 2026-07-29
reject mixed-separator path traversal in imap attachment names

Samin061:
`_is_escaping_current_directory` in the imap hook rejects a traversing attachment name with `f"..{os.sep}" in name`, which is tied to the host separator. On Windows `os.sep` is a backslash but a forward slash is also a valid path separator, so an attachment named `../../evil` slips past the check and, once joined onto the output directory in `_correct_path`, resolves outside it, letting whoever sent the mail write an arbitrary file. This normalises both separators and rejects any `..` path component instead.

---

---

### DIFF
--- a/providers/imap/src/airflow/providers/imap/hooks/imap.py
+++ b/providers/imap/src/airflow/providers/imap/hooks/imap.py
@@ -332,7 +332,11 @@ class ImapHook(BaseHook):
         return os.path.islink(self._correct_path(name, local_output_directory))
 
     def _is_escaping_current_directory(self, name: str) -> bool:
-        return f"..{os.sep}" in name
+        # Windows also accepts "/" as a path separator, so a "../" attachment name
+        # traverses there even though os.sep is a backslash and the old
+        # f"..{os.sep}" check missed it. Normalise both separators and reject any
+        # ".." path component.
+        return any(part == ".." for part in name.replace("\\", "/").split("/"))
 
     def _correct_path(self, name: str, local_output_directory: str) -> str:
         return (
--- a/providers/imap/src/airflow/providers/imap/hooks/imap.py
+++ b/providers/imap/src/airflow/providers/imap/hooks/imap.py
@@ -332,11 +332,15 @@ class ImapHook(BaseHook):
         return os.path.islink(self._correct_path(name, local_output_directory))
 
     def _is_escaping_current_directory(self, name: str) -> bool:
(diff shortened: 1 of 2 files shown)


providers/imap/src/airflow/providers/imap/hooks/imap.py
```diff
@@ -322,7 +322,11 @@ def _is_symlink(self, name: str) -> bool:
         return os.path.islink(name)
 
     def _is_escaping_current_directory(self, name: str) -> bool:
-        return f"..{os.sep}" in name
+        # Windows also accepts "/" as a path separator, so a "../" attachment name
+        # traverses there even though os.sep is a backslash and the old
+        # f"..{os.sep}" check missed it. Normalise both separators and reject any
+        # ".." path component.
+        return any(part == ".." for part in name.replace("\\", "/").split("/"))
```
uranusjr: Non-Windows platforms treat backslashes differently, so this unconditional replacement is wrong. Use `os.altsep` and `os.sep` instead.

Samin061: Good point, backslashes are legal in POSIX filenames so the unconditional replace was wrong. Switched to splitting on os.sep plus os.altsep, so backslashes stay ordinary chars on POSIX and both separators get caught on Windows. Added separate POSIX and Windows parametrized tests to lock in the platform difference.

potiuk (approved):
Good fix. Tying the check to `os.sep` meant that on Windows, where `/` is equally valid as a separator, `../../evil` walked straight past the guard — and attachment names come from email, so that input is attacker-controlled whenever the mailbox receives external mail.

The revision addresses the earlier review point properly: using `os.sep` plus `os.altsep` keeps backslashes as ordinary characters on POSIX while catching both separators on Windows, and the two parametrized tests lock that platform difference in. `("..\\..\\test1.csv", False)` on POSIX is exactly the case worth pinning.

I checked the obvious follow-on question — absolute names like `/etc/passwd` — and they are covered, though somewhat by accident: `_correct_path` concatenates strings rather than using `os.path.join`, so an absolute name becomes `/out//etc/passwd` and stays inside the directory. Worth knowing that this is load-bearing, since switching that concatenation to `os.path.join` later would reintroduce an escape that this check does not cover.

Splitting on components also catches a bare `..` that the old substring test missed, which is a nice bonus even though it isn't exploitable on its own.

---

Samin061:
Thanks for the review and the merge, glad this one landed.

PR #70670, opened 2026-07-29
Raise the datamodel-code-generator floor for the codegen toolchain

potiuk:
The `codegen` dependency groups in `task-sdk` and `airflow-ctl` pinned
`datamodel-code-generator==0.41.0`. A number of published advisories affect
releases in that range; the highest fix floor among them is 0.64.0, so this
requires `>=0.71.0` to resolve clear of all of them with headroom rather than
landing on the boundary release.

Regenerating the models with the newer generator changes their shape: it no
longer emits a spurious `= None` default for fields the OpenAPI spec lists as
`required` but nullable. The regenerated models now match the spec instead of
silently accepting payloads that omit those fields.

**No Cadwyn migration is involved.** No datamodel, route, or spec file is
touched — the server-side contract is byte-identical. This is purely a change
in how the generator renders the same spec. The client pins
`airflow-api-version`, so Cadwyn shapes every response to the version the
client was generated against.

Test fixtures that relied on the old leniency now pass those fields
explicitly. Two of them were silently dropping a value by passing `schema_=`
instead of the `schema` alias — the old implicit default masked the bug.

The supervisor schema snapshot (`schema/schema.json`) is regenerated as a
consequence, since the comms models inherit from the generated datamodels.

Supersedes the `datamodel-code-generator` half of #70660, which bumps the same
pin but does not regenerate the models and so does not carry the fixture
updates that change requires.

---


potiuk:
Pushed [4fa3fcf](https://github.com/apache/airflow/commit/4fa3fcf21a8168d86cae6410c1a94b13490aaac5),
which clears the CI failures from the model regeneration. The generated models no longer carry a
spurious `= None` for fields the spec marks `required` but nullable, so the remaining call sites
have to name them:

- `test_rotate_fernet_key_command.py` — `schema_=` → `schema=`. A third instance of the alias
  bug already fixed twice in this PR: the value was being dropped, and the old implicit default
  hid it.
- `providers/amazon/.../test_base_aws.py` — names the six now-required `ConnectionResult` fields.
- `providers/openlineage/.../test_listener.py` — see below.
- `ts-sdk/src/generated/supervisor.ts` — regenerated; it derives from the same supervisor schema
  snapshot this PR already regenerates.

cc @mobuchowski @kacpermuda for the OpenLineage change — 26 tests in
`TestOpenLineageListenerAirflow3` were failing on `SdkDagRun` validation. Two points worth a look:

- `data_interval_start` / `data_interval_end` are passed as `None`, not datetimes. The
  runtime-TI tests assert `nominal_start_time=None` / `nominal_end_time=None`, so a real
  datetime would trade one failure for another. `None` reproduces exactly what the old implicit
  default produced.
- The three new keys (`data_interval_start`, `data_interval_end`, `partition_key`) go through
  the same `if field in SdkDagRun.model_fields` guard the existing `state` key already used, so
  the file keeps working against the older SDKs the provider supports (`apache-airflow>=2.11.0`).

No production code changed here — test fixtures plus one generated file.

The `Helm tests / Release Helm` failure is unrelated to this PR: `chart/Chart.yaml` pulls
`postgresql` from the retired `charts.bitnami.com/bitnami` index, so it fails on any PR that
runs helm tests.

---

mobuchowski (approved):

PR #70679, opened 2026-07-29
Allow templating GCS and BigQuery URIs in Vertex AI batch prediction

osintalex:
Allow callers to template both gcs, bq inputs and outputs for batch jobs. I am adding this in after swapping from using the vertex batch ai job operator to writing my own custom operator due to the inability to template destinations and outputs to both gcs and bq.

This seems like normal airflow behaviour to me to allow templating for different inputs/outputs, i.e. in dev environment you would inject a dev bigquery project from the config and then template that to build `<dev-project>.<dataset>` whereas in prod you would have `<prod-project>.<dataset>`. Lacking the ability to do this in my experience pretty much breaks this operator for multi-stage work environments where you want the same code to run across dev/staging/prod.

 

---

No

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
@@ -168,7 +168,6 @@ class CreateBatchPredictionJobOperator(GoogleCloudBaseOperator):
         "gcs_source",
         "bigquery_source",
         "gcs_destination_prefix",
-        "gcs_destination_prefix",
         "bigquery_destination_prefix",
     )
     operator_extra_links = (VertexAIBatchPredictionJobLink(),)
--- a/providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
@@ -159,7 +159,18 @@ class CreateBatchPredictionJobOperator(GoogleCloudBaseOperator):
     :param poll_interval: Interval size which defines how often job status is checked in deferrable mode.
     """
 
-    template_fields = ("region", "project_id", "model_name", "impersonation_chain", "job_display_name")
+    template_fields = (
+        "region",
+        "project_id",
+        "model_name",
+        "impersonation_chain",
+        "job_display_name",
+        "gcs_source",
+        "bigquery_source",
+        "gcs_destination_prefix",
+        "gcs_destination_prefix",
+        "bigquery_destination_prefix",
+    )
     operator_extra_links = (VertexAIBatchPredictionJobLink(),)
 
     def __init__(


providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
```diff
@@ -159,7 +159,18 @@ class CreateBatchPredictionJobOperator(GoogleCloudBaseOperator):
...
+    template_fields = (
+        "region",
+        "project_id",
+        "model_name",
+        "impersonation_chain",
+        "job_display_name",
+        "gcs_source",
+        "bigquery_source",
+        "gcs_destination_prefix",
+        "gcs_destination_prefix",
```
shahar1: This appears twice

shahar1 (changes requested):
1. Could you please fix the PR title? (it should very specific to what you patch in this PR, without conventional commits prefixes like `fix:`)
2. To get this approved and merged we'll need stronger justifications than just "it looks normal" - did you, or anyone else, actually need it on real-production cases? Do we need it for all currently suggested fields, or only few would be sufficient?

Template fields, at least in their current design, might incur a technical debt if misused in the operator - so we should be very picky regarding new additions. 
I raised up #70296 after improving the pre-commit that checks for more forms of misuse, and there were about 60 justified cases where it applied. It might not be the case here (I'll let the CI/CD run to judge that), but I want to ensure that they have indeed a valid case, because we cannot easily break it afterwards.

I would like to note that if you have a valid case and you need this feature, you could always subclass and override the existing operator.

providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py
osintalex: Ty will catch.

Regarding this https://github.com/apache/airflow/issues/70296

I don't think I really understand the implications of this. From what I can tell in the operator the `__init__` method simply assigns these attributes onto the class https://github.com/apache/airflow/blob/main/providers/google/src/airflow/providers/google/cloud/operators/vertex_ai/batch_prediction_job.py#L165. 

I'm struggling to see how that would cause some sort of issue or antipattern here, it doesn't look like any anything happens in the constructor here apart from the assignment, which I believe is in alignment with your guide here https://airflow.apache.org/docs/apache-airflow/stable/howto/custom-operator.html#limitations. Let me know if I've misunderstood though.

As for 2, yes I ran into this in production when running a batch inference job with sources in bigquery and destination in bigquery. It made in my experience the operator hard to understand and use since the bigquery source and destinations were not templatable, which I found counterintuitive. By comparison these sorts of fields are typically templatable across airflow, e.g. on bigquery insert job operators. It was also very difficult to debug since the job just mysteriously fails and you don't get any error feedback on the source/destination being invalid.

This is what I was trying to articulate by 'normal', i.e. allowing these fields to be templatable in my opinion is much more consistent with most other airflow operators, where source/destination outputs to GCS or BQ are typically templatable. 

Lacking the ability to template the operator makes it, in my opinion, quite unusable in most professional production environments where you would have dev/staging/prod environments that all need different input/output configurations but should run the same code.

While I could create a new class that inherits from it, that doesn't feel so good to me since I would argue most people using this operator in production systems will want the same thing as me, i.e. ability to template these fields dynamically based on environment. Seems to me like something which should be supported by default.

shahar1: Apologies for the delay, and thanks for the detailed answer!
I'm satisfied with your current explanation. The implications are mainly regaulations around what you can do with template fields in the constructor - once a field is templated, you cannot make any validations for its value in the operator's constructor. 
In this case it should be safe, so it's a LGTM from me.

shahar1 (approved):

PR #45931, opened 2025-01-22
Add config option [secrets]backends_order

moiseenkov:
Introduce a new configuration option for specifying secret backends load order:
```
[secrets]backends_order = custom,environment_variable,metastore
```
The default value represents current behavior, thus nothing will change for existing users.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @moiseenkov** · by `@potiuk` · 2026-07-15 16:00 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed (see our [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria)):
> - :x: **Merge conflicts** — rebase on `main`. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/10_working_with_git.rst).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #70681, opened 2026-07-29
Secrets backends order updated

lubimow-xwf:
Add configurable secrets backend lookup order in Core and Workers

Airflow now allows configuring the lookup order of secret backends via two new configuration options:
- ``[secrets] backends_order``: Controls lookup order for core components (supports ``custom``, ``environment_variable``, and ``metastore``).
- ``[workers] backends_order``: Controls lookup order on Task SDK workers (supports ``custom``, ``environment_variable``, and ``execution_api``).

The configured order can also be inspected in the UI on the Variables page and via the ``GET /ui/backends_order`` endpoint. Default values preserve existing behavior (``custom,environment_variable,metastore`` and ``custom,environment_variable,execution_api``).

Yes, Gemini


potiuk:
Assigning to myself - will take a look - but irmt might not be very quick

potiuk:
In the meantime - please tak a look at potential CI issues

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/config.py
```diff
@@ -74,3 +81,32 @@ def get_configs() -> ConfigResponse:
...
+    responses={
+        **create_openapi_http_exception_doc(
+            [
+                status.HTTP_404_NOT_FOUND,
+                status.HTTP_406_NOT_ACCEPTABLE,
+            ]
+        ),
+    },
+    response_model=Config,
+    dependencies=[Depends(requires_authenticated())],
```
potiuk: This handler is `get_config_value` from `routes/public/config.py` with both of its guards removed: `requires_access_configuration("GET")` became `requires_authenticated()`, and `_check_expose_config()` is gone.

`[api] expose_config` defaults to `"False"`, and `_check_expose_config()` raises 403 "Your Airflow administrator chose not to expose the configuration, most likely for security reasons." So on a default install, `GET /api/v2/config/section/secrets/option/backends_order` returns 403 while this returns 200 with the value to the lowest-privileged authenticated user.

`backends_order` isn't a secret, so I'm not calling this a vulnerability — but it is a control the deployment manager chose, and the same value now has two different authorization answers.

The neighbouring `/ui/config` endpoint is authenticated-only because it serves a fixed curated allowlist (`API_CONFIG_KEYS` plus `instance_name`, `theme`, …), not arbitrary config reads. Folding `backends_order` into that `get_configs()` payload would match the established pattern; otherwise please restore both guards.

Either way `TestGetBackendsOrder` should gain an `unauthorized_test_client` case and an `expose_config=False` case — neither is covered today.

airflow-core/src/airflow/secrets/__init__.py
```diff
@@ -29,12 +29,11 @@
 
 from airflow.utils.deprecation_tools import add_deprecated_classes
 
-__all__ = ["BaseSecretsBackend", "DEFAULT_SECRETS_SEARCH_PATH"]
+__all__ = [
```
potiuk: `DEFAULT_SECRETS_SEARCH_PATH` was in `__all__` and re-exported here, so `from airflow.secrets import DEFAULT_SECRETS_SEARCH_PATH` was public API. After this change it raises `AttributeError`.

The sibling symbol in this same file, `DEFAULT_SECRETS_SEARCH_PATH_WORKERS`, gets a `__getattr__` shim with a `DeprecatedImportWarning` and a back-compat fallback. This one gets nothing. Nothing in-tree still imports it, but out-of-tree code may.

Please either add the same deprecation shim, or keep the removal and document it in a `.significant.rst` newsfragment.

airflow-core/src/airflow/configuration.py
```diff
@@ -740,35 +739,103 @@ def get_custom_secret_backend(worker_mode: bool = False) -> BaseSecretsBackend |
...
     Initialize secrets backend.
 
     * import secrets backend classes
     * instantiate them and return them in a list
     """
-    backend_list = []
     worker_mode = False
-    if default_backends != DEFAULT_SECRETS_SEARCH_PATH:
+    search_section = "secrets"
+    environment_variable_args: str | None = (
```
potiuk: These three class paths are the ones `scripts/ci/prek/check_secrets_search_path_sync.py` exists to keep in sync (registered at `.pre-commit-config.yaml:588`) — it compares core's `DEFAULT_SECRETS_SEARCH_PATH` against task-sdk's `_SERVER_DEFAULT_SECRETS_SEARCH_PATH`.

This PR stops using both constants and embeds the same literals here and again in `task-sdk/src/airflow/sdk/configuration.py`. That's four copies, of which the hook only checks two — so the two new ones can drift silently, which is the exact failure the hook was written to prevent.

Cleanest fix is to build `backends_map` from the existing constants. Failing that, please extend the hook to cover these literals too.

task-sdk/src/airflow/sdk/configuration.py
```diff
@@ -340,3 +410,6 @@ def __getattr__(name: str):
         globals()[name] = val
         return val
     raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
+
+
+secrets_backend_list = initialize_secrets_backends()
```
potiuk: This runs at module import, and `airflow.sdk.configuration` is imported at module level by `sdk/bases/sensor.py`, `sdk/definitions/asset/__init__.py` and `sdk/definitions/_internal/abstractoperator.py` — so it executes on every SDK import, i.e. every Dag parse. Previously this was computed lazily inside `ensure_secrets_loaded()`.

Two effects worth weighing:

- With the default `[secrets] backends_order`, this imports and constructs `airflow.secrets.metastore.MetastoreBackend` — a core, metadata-DB-backed backend — inside the Dag File Processor and on workers, purely as an import side effect.
- A misconfigured `backends_order` now raises `AirflowConfigException` at import time, so it surfaces as a Dag import error rather than a config error at first secrets lookup.

I realise core does the same at module scope and this is probably consistency with it — but core is the heavyweight side, and this module's `conf` is deliberately lazy behind `__getattr__`, which this line forces eagerly. Could this stay lazy in the SDK?

airflow-core/src/airflow/ui/src/pages/Variables/BackendsOrderModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [backendsOrder, setBackendsOrder] = useState<Array<string> | string>();
+  const { data, error } = useConfigServiceGetBackendsOrderValue();
+
+  const onOpenChange = () => {
+    onClose();
+  };
+
+  useEffect(() => {
+    setBackendsOrder(data?.sections[0]?.options[0]?.value ?? "");
+  }, [data, open]);
```
potiuk: `backendsOrder` is derived entirely from `data`, so this is the `useState + useEffect` sync pattern the review guidelines call out:

> Avoid `useState + useEffect` to sync derived state. Use nullish coalescing or nullable override patterns instead.

— `.github/instructions/code-review.instructions.md` § UI Code

```suggestion
  const { data, error } = useConfigServiceGetBackendsOrderValue();
  const backendsOrder = data?.sections[0]?.options[0]?.value ?? "";

  const onOpenChange = () => {
    onClose();
  };
```

That also lets the `useState`/`useEffect` imports go, along with the `Array<string> | string` type (the value is only ever a string).

Unrelated but nearby: this modal is always mounted by `BackendsOrderCard`, so the query fires on every Variables page render even when the dialog is closed — `enabled: open` would avoid that.

airflow-core/src/airflow/ui/src/pages/Variables/BackendsOrderButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onClick,
+  state = null,
+}: {
+  readonly colorScheme: string;
+  readonly icon?: React.ReactNode;
+  readonly isLoading?: boolean;
+  readonly label: string;
+  readonly link?: string;
+  readonly onClick?: () => void;
+  readonly state?: TaskInstanceState | null;
```
potiuk: This component is a near-verbatim copy of `src/components/StatsCard.tsx` with `count` and `isRTL` removed. It keeps `state?: TaskInstanceState | null`, the `StateBadge` and the `link`/`RouterLink` branch — none of which the single call site in `BackendsOrderCard` uses, and task-instance state has no meaning for a settings button.

It also drops `StatsCard`'s RTL handling (`isRTL ? <FiChevronLeft /> : <FiChevronRight />`), so the chevron points the wrong way in the `ar` and `he` locales.

Could this just use `StatsCard` directly? If a separate component is genuinely needed, please drop the unused props and keep the RTL branch.

potiuk (dismissed):
Thanks for picking this up. The core idea is sound and the defaults are right — I checked that `custom,environment_variable,metastore` reproduces today's precedence exactly (`DEFAULT_SECRETS_SEARCH_PATH` is `[env, metastore]` and the custom backend was prepended), so no existing deployment silently reorders. The validation and its error messages are clear, and the config-level tests are thorough.

Some context on this review: I know this is a takeover of #45931, which carries a long review history. I reviewed the code as it stands on this branch rather than assuming those discussions still apply to this snapshot, so if any of the below was already argued out over there, say so and I'll drop it.

Four things I'd like resolved before this merges, and a few smaller ones.

### The new UI endpoint drops both guards of the route it copies

See the inline comment on `routes/ui/config.py`. `get_backends_order_value` mirrors the public `get_config_value` but without `requires_access_configuration("GET")` and without `_check_expose_config()`. Since `[api] expose_config` defaults to `False`, the public route answers 403 on a default install while this one answers 200 to any authenticated user.

### `DEFAULT_SECRETS_SEARCH_PATH` is removed from the public API with no deprecation

See the inline comment on `secrets/__init__.py`.

### The class paths guarded by `check-secrets-search-path-sync` are now hardcoded in two unguarded places

See the inline comment on `configuration.py`. Related: this branch also deletes `test_metastore_backend_in_server_chain` from `task-sdk/tests/task_sdk/execution_time/test_context.py`, which was the remaining coverage of `_SERVER_DEFAULT_SECRETS_SEARCH_PATH` — a constant that still exists and is still enforced by that hook. If the constant is genuinely dead now, remove it and the hook together; if it isn't, the test should stay.

### The Task SDK now builds every configured secrets backend at import time

See the inline comment on `task-sdk/src/airflow/sdk/configuration.py`.

### Smaller observations

- See the inline comments on `BackendsOrderModal.tsx` and `BackendsOrderButton.tsx`.
- **No newsfragment.** Two new config options, a new API endpoint, a new UI feature and a public-API removal are all user-facing; `airflow-core/newsfragments/` has no entry. Given the `DEFAULT_SECRETS_SEARCH_PATH` removal, `.significant.rst` is probably the right type.
- **The PR body is a single link.** The template is gone, including the generative-AI disclosure section. A short description of what changed and why would help reviewers who haven't followed #45931 — right now the only way to understand the change is to read the whole diff.
- **~75 lines are duplicated verbatim** between `airflow-core/src/airflow/configuration.py` and `task-sdk/src/airflow/sdk/configuration.py` — the `Backends` enum, `get_importable_secret_backend`, `backends_map`, and both validation blocks. `shared/configuration` exists for exactly this kind of shared logic.
- `backend = callback(*args) if args else callback()` — the `else` is unreachable; every `backends_map` entry carries a non-empty `args` tuple.

For what it's worth, several things I went looking for turned out fine: the regenerated `_private_ui.yaml` and `openapi-gen/*.ts` are in sync, the `conf` self-import inside `initialize_secrets_backends` follows the pattern already in that file, and the `feat:` prefix on the second commit is harmless since CI validates the PR title rather than commit subjects.

---

> confirmed by an Airflow maintainer. After you've
> addressed the points above and pushed an update, an Airflow
>

---

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/config.py
lubimow-xwf: Fixed.

airflow-core/src/airflow/secrets/__init__.py
lubimow-xwf: Restored to the original shape.

task-sdk/src/airflow/sdk/configuration.py
lubimow-xwf: Restored to the original shape.

airflow-core/src/airflow/configuration.py
lubimow-xwf: Restored to the original shape.

airflow-core/src/airflow/ui/src/pages/Variables/BackendsOrderButton.tsx
lubimow-xwf: Replaced with StatsCard.

airflow-core/src/airflow/ui/src/pages/Variables/BackendsOrderModal.tsx
lubimow-xwf: Applied both fixes.

lubimow-xwf:
Hey @potiuk 

I addressed all of your comments. 

I've made significant changes to both `configuration.py` files. I believe I simplified the changes, now sorting does not change the existing implementation too much, it only extends it. I apologize I did it now but this wasn't my PR originally and it was approved before, so I thought it was a matter of a couple of minor fixes.

Please take a look again and let me know what it looks like now, thanks

Btw, just quickly regarding UI changes, right now this change adds a button to show a modal to show a single line of text. Was it discussed before where and what should it look like? If it was me, I would just add a note in the footer of the Variables table, without any additional clicks. But that's just my five cents, right now it's unchanged.

airflow-core/src/airflow/configuration.py
```diff
@@ -740,6 +741,55 @@ def get_custom_secret_backend(worker_mode: bool = False) -> BaseSecretsBackend |
...
+class Backend(enum.Enum):
+    """Known secrets backends."""
+
+    ENVIRONMENT_VARIABLE = "environment_variable"
+    EXECUTION_API = "execution_api"
+    CUSTOM = "custom"
+    METASTORE = "metastore"
+
+    @classmethod
+    def from_module(cls, default_backend: str) -> Backend:
```
potiuk: These three literals are exactly the ones `scripts/ci/prek/check_secrets_search_path_sync.py` exists to keep in sync (registered at `.pre-commit-config.yaml:588`) — but that hook only extracts and compares the two list constants, `DEFAULT_SECRETS_SEARCH_PATH` in core and `_SERVER_DEFAULT_SECRETS_SEARCH_PATH` in task-sdk. It has no visibility into `from_module`.

Restoring the constants addressed the first half of my earlier comment, but the coupling moved rather than went away — and it now exists in two copies. This block is duplicated verbatim in `task-sdk/src/airflow/sdk/configuration.py` (lines 272-321): 50 lines identical except `from airflow.configuration import conf` vs `from airflow.sdk.configuration import conf`.

The concrete failure: add a fourth backend to `DEFAULT_SECRETS_SEARCH_PATH`. The hook stays green, because it only checks that the two constants match each other and they both moved. Then `initialize_secrets_backends` raises `ValueError: Unknown module provided: <class path>` on every start, in both core and the SDK.

`shared/configuration/` is the natural home for `Backend`, `from_module` and `_get_secrets_backend_order` — one definition instead of two, and the class-path mapping ends up next to the constants the hook already guards.

task-sdk/src/airflow/sdk/configuration.py
```diff
@@ -267,6 +269,55 @@ def get_custom_secret_backend(worker_mode: bool = False):
     return conf._get_custom_secret_backend(worker_mode=worker_mode)
 
 
+class Backend(enum.Enum):
```
potiuk: This block is identical to `airflow-core/src/airflow/configuration.py` lines 744-793 apart from the `conf` import — I diffed the two. See the comment there for the drift hazard; moving this into `shared/configuration/` would let both sides import a single definition.

airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -1589,6 +1589,22 @@ secrets:
       sensitive: true
       example: ~
       default: ""
+    backends_order:
+      description: |
+        .. note:: |experimental|
+
+        Comma-separated list of secret backends. These backends will be used in the order they are specified.
+        Please note that the `environment_variable` and `metastore` are required values and cannot be removed
```
potiuk: Single backticks are RST's default role rather than literal markup, so these render as italics instead of code. The `workers.backends_order` description immediately below uses double backticks for the same two values.

```suggestion
        Please note that the ``environment_variable`` and ``metastore`` are required values and cannot be removed
```

potiuk (commented):
All four blocking items from my previous review are properly fixed — the endpoint now carries both guards with tests covering each 403 path, the public constants are restored, the SDK is lazy again, and the UI reuses `StatsCard` instead of copying it. Thanks for working through them.

Two things left, one of which is my earlier finding resurfacing in a new location.

### The hardcoded class paths moved rather than went away

See the inline comment on `configuration.py`. `Backend.from_module` re-embeds the three literals that `check-secrets-search-path-sync` exists to keep in sync, in two files the hook doesn't inspect — and those 50 lines are now duplicated verbatim between core and task-sdk.

### No newsfragment

Two new config options, a new API endpoint and a new UI feature are user-facing; `airflow-core/newsfragments/` still has no entry.

### Smaller

- See the inline comment on `config.yml` (backtick style).
- The PR body is still a single link to #45931. A short description of what changed and why — plus the generative-AI disclosure section from the template — would help reviewers who haven't followed that PR.

---

>

---
Drafted-by: Claude Opus 5 (1M context); reviewed by @potiuk before posting

potiuk:
cc: @amoghrajesh @ephraimbuddy @kaxil -> I would also love to see your review here.

potiuk (approved):
Approve with some comments to complete.

airflow-core/src/airflow/configuration.py
lubimow-xwf: Quickly regarding duplication. I know it's duplicated, I copied it myself :) I wasn't aware there is a shared space, I thought you wanted core and sdk to be fully decoupled. By the way, why don't you move DEFAULT_SECRETS_SEARCH_PATH to shared space too? Wouldn't it be easier than having a script to compare them?

> The concrete failure: add a fourth backend to DEFAULT_SECRETS_SEARCH_PATH. The hook stays green, because it only checks that the two constants match each other and they both moved. Then initialize_secrets_backends raises ValueError: Unknown module provided: <class path> on every start, in both core and the SDK.

That's not a failure. Adding a new backend to that list and not extending the enum would be a bug. I think it's best to fail fast and let the developer fix the issues.

task-sdk/src/airflow/sdk/configuration.py
lubimow-xwf: Addressed in the other thread.

lubimow-xwf:
Hey @potiuk 

Please have a look again, I did move the ordering code to the shared module.

Regarding

> Backend.from_module re-embeds the three literals that check-secrets-search-path-sync exists to keep in sync

Well you had it hardcoded all around the project, I tried to address this issue too, I hope its fine now.

potiuk:
> Please have a look again, I did move the ordering code to the shared module.

LGTM. Thanks for being receptive.

> Well you had it hardcoded all around the project, I tried to address this issue too, I hope its fine now.

Yeah. We always try to leave the project better with every change, not worse, so thanks for that !

cc: @amoghrajesh @ephraimbuddy @kaxil - I would love if you take a look before I merge it finally.

PR #70684, opened 2026-07-29
Sensors should be consistent about whether or not they return a value

SamWheating:
Maybe more of a discussion topic here - but included one example change.

Many sensors will define their own `execute()` method, which then conditionally calls out to the deferable / non-deferable path. 

However, we're pretty inconsistent about whether or not that value should be returned or not. Some sensors will return the value of `execute()`:

https://github.com/apache/airflow/blob/cd3a7e1a5f6bbe9f6536ab593b06a5728deb0271/providers/http/src/airflow/providers/http/sensors/http.py#L160-L162

but most sensors will not, implicitly returning `None`: 
https://github.com/apache/airflow/blob/8dd76f1624bc28fdef8630684a06914891526578/providers/standard/src/airflow/providers/standard/sensors/filesystem.py#L120-L122

This makes for a really confusing experience when trying to subclass or extend an existing sensor. For example, trying to subclass the ExternalTaskSensor to add a delayed timestamp XCOM output:

```python
class ExternalTaskSensorWithCompletionDelay(ExternalTaskSensor):
    """Return a stable delay target when the upstream task is first observed."""

    def poke(self, context):
        is_done = super().poke(context)
        delay_target = timezone.utcnow() + timedelta(minutes=30).isoformat()
        
        return PokeReturnValue(is_done=bool(is_done), xcom_value=delay_target)
 ```
 
The XCOM value here will not actually be written, because the parent's class execute method ignores the return value and `task.execute()` returns nothing.
 
Is this expected behaviour? Or should we always be returning the value of `super().execute()`? I am open to discussion here but in my mind this sort of thing quietly breaks the extensibility and flexibility of builtin or provided operators.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @SamWheating** · by `@potiuk` · 2026-08-13 12:55 UTC
>
> - :x: **Merge conflicts**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/10_working_with_git.rst).
>
>
>

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/external_task.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/external_task.py
@@ -465,57 +465,56 @@ class ExternalTaskSensor(BaseSensorOperator):
     def execute(self, context: Context) -> None:
         """Run on the worker and defer using the triggers if deferrable is set to True."""
         if not self.deferrable:
-            super().execute(context)
+            return super().execute(context)
+        # Determine the timeout to use: prefer timeout parameter, fallback to execution_timeout
+        timeout_value = self.timeout
+        if not timeout_value and self.execution_timeout:
+            timeout_value = self.execution_timeout.total_seconds()
+
+        dttm_filter = self._get_dttm_filter(context)
+        if AIRFLOW_V_3_0_PLUS:
+            self.defer(
+                timeout=datetime.timedelta(seconds=timeout_value) if timeout_value else None,
+                trigger=WorkflowTrigger(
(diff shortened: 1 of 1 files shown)


SamWheating:
It looks like the one test failure is unrelated /  maybe a flakey test

SamWheating:
cc @potiuk - conflicts have been fixed. I am mostly interested in your thoughts on this issue / the inconsistency of the interface around sensors. Do you think that this is something we should be opinionated on?

PR #70685, opened 2026-07-29
Fail task instances whose stored next_kwargs cannot be processed

potiuk:
`handle_event_submit` decoded a task instance's stored `next_kwargs` and then
assumed the result was a dict. The decode caught only four exception types, so
anything the `BaseSerialization` compat fallback raised escaped; and the
`isinstance` check was under `if TYPE_CHECKING:`, so it never ran at runtime.

Both surface as exceptions from a function whose callers process every waiting
task instance in a single pass — the scheduler's `AWAITING_INPUT` timeout sweep
and two API routes — so one task instance with an unusable stored payload took
the whole batch down with it.

This decodes through a helper that validates its own result, and guards decode,
event insertion and re-encode as one unit. A task instance whose payload cannot
be processed is re-queued to fail via the existing `__fail__` path, so retries
and `on_failure` callbacks run normally, rather than being left parked for the
next sweep to trip over again.

### Behaviour change

The HITL response endpoint and the execution API's `AWAITING_INPUT` branch
previously returned HTTP 500 for such a payload; they now succeed, with the task
routed to fail. No request shape that previously worked is newly rejected.

### Test plan

- [x] `test_handle_event_submit_fails_task_with_unusable_next_kwargs` — parametrized over an undecodable payload and one that decodes to a non-dict
- [x] `test_awaiting_input_timeout_sweep_survives_unusable_next_kwargs` — end-to-end through the sweep, with a malformed and a healthy task instance in the same batch
- [x] Both fail on `main` without this change
- [x] `ruff check` / `ruff format` / `mypy` clean

### Deferred

Failing a task instance rather than raising is not free for the Human-in-the-loop path: a response whose `params_input` carries a serde-reserved key (`__classname__`, `__schema_id__`) is now recorded and discarded rather than rejected, and the `response_received` guard blocks a corrected resubmission. The fix belongs on the write side in `hitl.py`; tracked in #71036, and linked from a comment on `handle_event_submit`.

### DIFF
--- a/airflow-core/src/airflow/models/trigger.py
+++ b/airflow-core/src/airflow/models/trigger.py
@@ -511,6 +511,8 @@ def _decode_next_kwargs(next_kwargs_raw: Any) -> dict[str, Any]:
     what comes back out of the stored payload.
 
     :raise ValueError: The payload did not decode to a dict.
+    :raise Exception: Whatever the two decoders raise on a payload they cannot read -- the stored
+        blob is arbitrary, so the set is open and callers have to treat it as such.
     """
     from airflow.sdk.serde import deserialize
 
@@ -526,7 +528,9 @@ def _decode_next_kwargs(next_kwargs_raw: Any) -> dict[str, Any]:
     return next_kwargs
 
 
-def _fail_unresumable_task_instance(task_instance: TaskInstance, reason: str, *, session: Session) -> None:
+def _fail_unresumable_task_instance(
+    task_instance: TaskInstance, reason: str, exc: BaseException, *, session: Session
+) -> None:
     """
     Re-queue a task instance that cannot be resumed, so that a worker fails it.
 
@@ -534,12 +538,17 @@ def _fail_unresumable_task_instance(task_instance: TaskInstance, reason: str, *,
     fail the task immediately, which runs its normal failure handling (retries and callbacks
     included). Leaving the task instance parked instead would strand it there, as the event that
     should have resumed it is already gone.
+
+    The traceback travels in ``next_kwargs`` because that is the only channel that reaches the
+    task log; the exception is otherwise only logged where this runs, which the Dag author may
+    have no access to. It has to stay the list ``format_exception`` returns -- the runtime joins it.
     """
     task_instance.next_method = TRIGGER_FAIL_REPR
-    task_instance.next_kwargs = {"error": reason}
-    # Remove ourselves as its trigger
+    task_instance.next_kwargs = {
+        "error": reason,
+        "traceback": format_exception(type(exc), exc, exc.__traceback__),
+    }
     task_instance.trigger_id = None
-    # Finally, mark it as scheduled so it gets re-queued
     task_instance.state = TaskInstanceState.SCHEDULED
     task_instance.scheduled_dttm = timezone.utcnow()
     session.flush()
@@ -554,34 +563,61 @@ def handle_event_submit(event: TriggerEvent, *, task_instance: TaskInstance, ses
     as well as its state to scheduled. It also adds the event's payload
     into the kwargs for the task.
 
-    A task instance whose stored kwargs cannot be decoded (or re-encoded with the payload added)
-    is failed rather than resumed. That work happens in the scheduler and API processes, which
-    handle every parked task instance in turn, so a single unusable payload must not be able to
-    abort the caller.
+    A task instance whose stored kwargs cannot be decoded, or which the event payload cannot be
+    encoded into, is failed rather than resumed. This runs in the triggerer, the scheduler and the
+    API processes, each of which handles every waiting task instance in one pass, so a single
+    unusable payload must not be able to abort the caller. The triggerer had the worst of it: an
+    event whose submit raised was left unconfirmed and redelivered indefinitely.
+
+    Failing the task instance is not free for every caller: a Human-in-the-loop response whose
+    ``params_input`` serde cannot encode is now recorded and discarded rather than rejected, which
+    the submitter cannot retry. That wants validating on the write side; tracked at
+    https://github.com/apache/airflow/issues/71036
 
     :param task_instance: The task instance to handle the submit event for.
     :param session: The session to be used for the database callback sink.
     """
     from airflow.sdk.serde import serialize
-    from airflow.utils.state import TaskInstanceState
 
     next_kwargs_raw = task_instance.next_kwargs or {}
 
+    # Decoding and re-encoding fail for different reasons and are reported separately: blaming the
+    # stored kwargs for a payload the trigger just yielded would point the author at DB state that
+    # was never the problem.
     try:
         next_kwargs = _decode_next_kwargs(next_kwargs_raw)
-        # Add event to the plain dict, then serialize everything together so nested
-        # non-primitive values get proper serde encoding.
-        next_kwargs["event"] = event.payload
+    except Exception as exc:
+        log.exception(
+            "Could not decode the stored next_kwargs of %s; failing it instead of resuming it",
+            task_instance,
+        )
+        _fail_unresumable_task_instance(
+            task_instance,
+            "Could not resume the task: its stored next_kwargs could not be decoded "
+            f"({type(exc).__name__}: {exc})",
+            exc,
+            session=session,
+        )
+        return
+
+    # Add event to the plain dict, then serialize everything together so nested
+    # non-primitive values get proper serde encoding.
+    next_kwargs["event"] = event.payload
+    try:
         # Re-serialize using serde. The Execution API version converter
         # (ModifyDeferredTaskKwargsToJsonValue) handles converting this to
         # BaseSerialization format when serving old workers.
         serialized_next_kwargs = serialize(next_kwargs)
     except Exception as exc:
-        log.exception("Could not process next_kwargs of %s; failing it instead of resuming it", task_instance)
+        log.exception(
+            "Could not serialize the event payload for %s; failing it instead of resuming it",
+            task_instance,
+        )
         _fail_unresumable_task_instance(
             task_instance,
-            "Could not resume the task: its stored next_kwargs could not be processed "
+            f"Could not resume the task: the event payload could not be serialized "
             f"({type(exc).__name__}: {exc})",
+            exc,
             session=session,
         )
         return
--- a/airflow-core/src/airflow/models/trigger.py
+++ b/airflow-core/src/airflow/models/trigger.py
@@ -532,16 +532,11 @@ def _fail_unresumable_task_instance(
     task_instance: TaskInstance, reason: str, exc: BaseException, *, session: Session
 ) -> None:
     """
-    Re-queue a task instance that cannot be resumed, so that a worker fails it.
+    Route through ``__fail__`` (mirrors `Trigger.submit_failure`) so a worker fails the
+    task normally instead of leaving it stranded with no event left to resume it.
 
-    Mirrors :meth:`Trigger.submit_failure`: the special ``__fail__`` next_method makes the worker
-    fail the task immediately, which runs its normal failure handling (retries and callbacks
-    included). Leaving the task instance parked instead would strand it there, as the event that
-    should have resumed it is already gone.
-
-    The traceback travels in ``next_kwargs`` because that is the only channel that reaches the
-    task log; the exception is otherwise only logged where this runs, which the Dag author may
-    have no access to. It has to stay the list ``format_exception`` returns -- the runtime joins it.
+    Traceback goes into ``next_kwargs`` as a list -- the only channel reaching the task log --
+    since ``format_exception`` returns it that way and the runtime joins it.
     """
     task_instance.next_method = TRIGGER_FAIL_REPR
     task_instance.next_kwargs = {
--- a/airflow-core/src/airflow/models/trigger.py
+++ b/airflow-core/src/airflow/models/trigger.py
@@ -500,6 +500,51 @@ class Trigger(Base):
         return result
 
 
+def _decode_next_kwargs(next_kwargs_raw: Any) -> dict[str, Any]:
+    """
+    Decode the stored ``next_kwargs`` of a task instance into a plain dict.
+
+    Deserialize with serde first to provide a compat layer if there are mixed serialized
+    (BaseSerialisation and serde) data, which can happen if a deferred task resumes after upgrade.
+
+    The result is checked here rather than assumed, so callers never have to trust the shape of
+    what comes back out of the stored payload.
+
+    :raise ValueError: The payload did not decode to a dict.
+    """
+    from airflow.sdk.serde import deserialize
+
+    try:
+        next_kwargs = deserialize(next_kwargs_raw)
+    except (ImportError, KeyError, AttributeError, TypeError):
+        from airflow.serialization.serialized_objects import BaseSerialization
+
+        next_kwargs = BaseSerialization.deserialize(next_kwargs_raw)
+
+    if not isinstance(next_kwargs, dict):
+        raise ValueError(f"next_kwargs decoded to {type(next_kwargs).__name__}, expected a dict")
+    return next_kwargs
+
+
+def _fail_unresumable_task_instance(task_instance: TaskInstance, reason: str, *, session: Session) -> None:
+    """
+    Re-queue a task instance that cannot be resumed, so that a worker fails it.
+
+    Mirrors :meth:`Trigger.submit_failure`: the special ``__fail__`` next_method makes the worker
+    fail the task immediately, which runs its normal failure handling (retries and callbacks
+    included). Leaving the task instance parked instead would strand it there, as the event that
+    should have resumed it is already gone.
+    """
+    task_instance.next_method = TRIGGER_FAIL_REPR
+    task_instance.next_kwargs = {"error": reason}
+    # Remove ourselves as its trigger
+    task_instance.trigger_id = None
+    # Finally, mark it as scheduled so it gets re-queued
+    task_instance.state = TaskInstanceState.SCHEDULED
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/models/trigger.py
```diff
@@ -500,6 +500,51 @@ def get_sorted_triggers(
...
+    Re-queue a task instance that cannot be resumed, so that a worker fails it.
+
+    Mirrors :meth:`Trigger.submit_failure`: the special ``__fail__`` next_method makes the worker
+    fail the task immediately, which runs its normal failure handling (retries and callbacks
+    included). Leaving the task instance parked instead would strand it there, as the event that
+    should have resumed it is already gone.
+    """
+    task_instance.next_method = TRIGGER_FAIL_REPR
+    task_instance.next_kwargs = {"error": reason}
+    # Remove ourselves as its trigger
```
amoghrajesh:
```suggestion
```

airflow-core/src/airflow/models/trigger.py
```diff
@@ -500,6 +500,51 @@ def get_sorted_triggers(
...
+    task_instance.trigger_id = None
+    # Finally, mark it as scheduled so it gets re-queued
```
amoghrajesh:
```suggestion
```

airflow-core/tests/unit/jobs/test_scheduler_job.py
```diff
@@ -8077,6 +8077,60 @@ def test_awaiting_input_timeout_without_defaults_fails(self, dag_maker):
         assert ti.next_method == "execute_complete"
         assert ti.next_kwargs["event"]["error_type"] == "timeout"
 
+    def test_awaiting_input_timeout_sweep_survives_unusable_next_kwargs(self, dag_maker):
```
amoghrajesh, line 8216: Docstring and inline comments are narrating same thing, remove either.

airflow-core/tests/unit/models/test_trigger.py
```diff
@@ -256,6 +257,45 @@ def test_submit_event_no_n_plus_one_for_assets(_, session, asset_count, expected
...
+        # trips over the missing legacy keys.
+        pytest.param(
+            {"__classname__": "not.allowed.Thing", "__version__": 1, "__data__": {}},
+            id="undecodable",
+        ),
+        # Decodes cleanly, but not into a dict: legacy encoding of a bare datetime.
+        pytest.param({"__type": "datetime", "__var": 1735689600.0}, id="not-a-dict"),
+    ],
+)
+def test_handle_event_submit_fails_task_with_unusable_next_kwargs(
```
amoghrajesh, line 274: Same comment as above here.

amoghrajesh (approved):
Thanks @potiuk. I do not see any backcompat concerns too here, but @kaxil if you could take another look.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -509,33 +554,39 @@ def handle_event_submit(event: TriggerEvent, *, task_instance: TaskInstance, ses
...
-        assert isinstance(next_kwargs, dict)
-    next_kwargs["event"] = event.payload
+        next_kwargs = _decode_next_kwargs(next_kwargs_raw)
+        # Add event to the plain dict, then serialize everything together so nested
+        # non-primitive values get proper serde encoding.
+        next_kwargs["event"] = event.payload
+        # Re-serialize using serde. The Execution API version converter
+        # (ModifyDeferredTaskKwargsToJsonValue) handles converting this to
+        # BaseSerialization format when serving old workers.
+        serialized_next_kwargs = serialize(next_kwargs)
```
kaxil, line 605: One case this unit newly consumes: serde's `serialize` raises `AttributeError` for any dict containing its reserved keys (`__classname__` / `__schema_id__`), at any depth, and the HITL route stores `params_input` verbatim with no key validation, so a response whose params contain such a key blows up here. On main that request 500s, nothing is recorded, and the client can fix and resubmit. With this change the response is recorded, the task is routed to `__fail__`, the route returns 200 echoing the response, and the `response_received` 409 guard blocks a corrected resubmission. A write-side 400 for serde-reserved keys in `params_input`, next to the existing option checks in hitl.py, would keep that path recoverable (fine as a follow-up). Relatedly, should this function report back whether it resumed or failed the task? The HITL route currently has no way to know it just returned 200 for a response that was effectively discarded.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -509,33 +554,39 @@ def handle_event_submit(event: TriggerEvent, *, task_instance: TaskInstance, ses
...
+    except Exception as exc:
+        log.exception("Could not process next_kwargs of %s; failing it instead of resuming it", task_instance)
+        _fail_unresumable_task_instance(
+            task_instance,
+            "Could not resume the task: its stored next_kwargs could not be processed "
```
kaxil: The try also spans `serialize` after the event payload is added, so a failure caused by the incoming payload (a trigger yielding something serde can't encode, or the reserved-keys case above) gets reported as the stored next_kwargs being unreadable, pointing the author at DB state that was fine. Splitting the decode from the inject-and-serialize step, with a payload-specific message for the latter, would keep the diagnosis accurate.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -500,6 +500,51 @@ def get_sorted_triggers(
...
+    """
+    Re-queue a task instance that cannot be resumed, so that a worker fails it.
+
+    Mirrors :meth:`Trigger.submit_failure`: the special ``__fail__`` next_method makes the worker
+    fail the task immediately, which runs its normal failure handling (retries and callbacks
+    included). Leaving the task instance parked instead would strand it there, as the event that
+    should have resumed it is already gone.
+    """
+    task_instance.next_method = TRIGGER_FAIL_REPR
+    task_instance.next_kwargs = {"error": reason}
```
kaxil: Worth also setting `"traceback": format_exception(type(exc), exc, exc.__traceback__)` here like `submit_failure` does (means passing the exception into this helper); `resume_execution` already renders that key into the task log. As written the task log only shows the one-line reason while the full traceback lands in the scheduler/triggerer/API log, which the DAG author may not be able to see. It has to stay the list `format_exception` returns, since the runtime does `"\n".join(traceback)`.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -509,33 +554,39 @@ def handle_event_submit(event: TriggerEvent, *, task_instance: TaskInstance, ses
     as well as its state to scheduled. It also adds the event's payload
     into the kwargs for the task.
 
+    A task instance whose stored kwargs cannot be decoded (or re-encoded with the payload added)
+    is failed rather than resumed. That work happens in the scheduler and API processes, which
```
kaxil: This also runs in the triggerer, via `Trigger.submit_event`, and that caller arguably had the worst pre-fix failure mode: `handle_events` leaves the event unconfirmed on a raise, so the poison event is redelivered indefinitely. Maybe "the triggerer, scheduler and API processes". On the same theme, the `:raise ValueError:` note on `_decode_next_kwargs` documents only the isinstance failure; serde and BaseSerialization decode errors propagate from it too, which is exactly why this except has to stay broad. One sentence there would stop a future reader narrowing it back down.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -509,33 +554,39 @@ def handle_event_submit(event: TriggerEvent, *, task_instance: TaskInstance, ses
...
+    handle every parked task instance in turn, so a single unusable payload must not be able to
+    abort the caller.
+
     :param task_instance: The task instance to handle the submit event for.
     :param session: The session to be used for the database callback sink.
     """
-    from airflow.sdk.serde import deserialize, serialize
+    from airflow.sdk.serde import serialize
     from airflow.utils.state import TaskInstanceState
```
kaxil: This local import now shadows the module-level one from line 43 (`_fail_unresumable_task_instance` above uses the module-level name). Since the line above is already touched, may as well drop it.

airflow-core/tests/unit/jobs/test_scheduler_job.py
```diff
@@ -8077,6 +8077,60 @@ def test_awaiting_input_timeout_without_defaults_fails(self, dag_maker):
...
+                subject="approve?",
+                defaults=["Approve"],
+                multiple=False,
+                params={},
+            )
+        )
+        session.flush()
+
+        self.job_runner = SchedulerJobRunner(job=Job())
+        self.job_runner.check_awaiting_input_timeouts(session=session)
```
kaxil: Related, though outside this diff: now that `handle_event_submit` can route a TI to `__fail__` instead of resuming it, the sweep's response and defaults branches still count that TI under `num_resolved`, so the summary log reports it as "resolved (response/defaults)". Checking `ti.next_method == TRIGGER_FAIL_REPR` after the call and counting those separately would keep that log line honest.

kaxil:
A few notes on the description text rather than the code:

1. "the execution API's `AWAITING_INPUT` branch previously returned HTTP 500" -- it didn't. That route catches everything escaping the state-update helper and silently marks the TI FAILED with a direct UPDATE ([task_instances.py, lines 465-476 on main](https://github.com/apache/airflow/blob/b316afb44dc9686f5c39aab678a9a89cb21d53d2/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py#L465-L476)), bypassing retries and worker-run `on_failure` callbacks. So the PR improves that caller more than the body claims: the task now goes through the real `__fail__` path instead of a hard FAILED write. The HITL half of that sentence is accurate.

2. The caller list ("the scheduler's `AWAITING_INPUT` timeout sweep and two API routes") omits the triggerer: `Trigger.submit_event` also lands in `handle_event_submit`, and pre-fix it had the worst failure mode of the four -- `handle_events` only confirms an event's persist seq after `submit_event` returns, so an escaping decode error meant the poison event was redelivered indefinitely ([triggerer_job_runner.py, lines 764-776](https://github.com/apache/airflow/blob/b316afb44dc9686f5c39aab678a9a89cb21d53d2/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L764-L776)).

3. One tradeoff worth a sentence in the behaviour-change section: when decode fails for environment reasons rather than bad data (say an `allowed_deserialization_classes` mismatch or a provider missing on the scheduler mid rolling-upgrade), waiting tasks now fail and consume retries, where main halted and preserved them until the environment was fixed. Failing is the right call, but the body currently reads as if nothing previously-working is newly affected, and this is the one case where the old crash-loop was accidentally protective.

airflow-core/src/airflow/models/trigger.py
potiuk: Done. `_fail_unresumable_task_instance` now takes the exception and packs `format_exception(type(exc), exc, exc.__traceback__)` into `next_kwargs`, matching `submit_failure`. Kept as the list `format_exception` returns, since `operator.py` does `"\n".join(traceback)` when it renders it.

Your framing of why is the part worth keeping, so I put it in the docstring: `next_kwargs` is the only channel that reaches the task log, and the process log where this runs is frequently not the Dag author's to read.

---

airflow-core/src/airflow/models/trigger.py
potiuk: Agreed, and split. Decoding and re-encoding are now separate `try` blocks with their own messages — "its stored next_kwargs could not be decoded" versus "the event payload could not be serialized" — so a payload the trigger just yielded is no longer reported as unreadable database state.

Added `test_handle_event_submit_fails_task_when_the_event_payload_cannot_be_serialized`, which uses the reserved-key case from your other comment to drive the second branch.

---

airflow-core/src/airflow/models/trigger.py
potiuk: Both taken. The docstring now says "the triggerer, the scheduler and the API processes" and calls out that the triggerer had the worst of it — `handle_events` leaving the event unconfirmed meant the poison event came back indefinitely, which is a materially worse failure than one aborted sweep pass.

Also added the note to `_decode_next_kwargs`: it documents that the decoders' own failures propagate too and that the set is open, which is exactly why the caller's `except` is broad. That was the reader-narrows-it-back-down risk you were pointing at.

---

airflow-core/src/airflow/models/trigger.py
potiuk: Dropped — `TaskInstanceState` is imported at module level on line 43 and `_fail_unresumable_task_instance` was already relying on that, so the local import was pure shadowing.

---

airflow-core/tests/unit/jobs/test_scheduler_job.py
potiuk: Fixed here rather than left outside the diff — the log line only becomes untrue *because* of this change, so it belongs with it.

The sweep now tracks `num_unresumable` separately: after each `handle_event_submit` it checks `ti.next_method == TRIGGER_FAIL_REPR`, which is reachable in all three branches and means the task was neither resolved nor timed out. The summary reads "%i resolved (response/defaults), %i failed, %i could not be resumed". `test_awaiting_input_timeout_sweep_survives_unusable_next_kwargs` now asserts the counts are `(1, 0, 1)` rather than the previous `(1, 0)` with the bad one hidden inside "resolved".

---

airflow-core/src/airflow/models/trigger.py
potiuk: This is the sharpest of the batch, and I want to be straight about it rather than wave it through: for that specific case the change does trade one bad outcome for a different bad outcome. A HITL response whose `params_input` carries `__classname__` or `__schema_id__` currently 500s and can be corrected and resubmitted; after this change it is recorded, the task is routed to `__fail__`, the route answers 200, and the `response_received` guard makes the corrected resubmission a 409. That is worse for that user, and it is caused by this PR.

I am still landing it as is, because the alternative it replaces is worse in aggregate: the same raise in the triggerer leaves the event unconfirmed and redelivers it forever, and in the sweep it aborts the whole batch, taking unrelated healthy task instances with it. Trading "one user must contact an admin" for "one payload wedges a component" is the right way round. The reserved-key case also needs a params key literally named `__classname__` or `__schema_id__`, which is narrow.

The write-side 400 in `hitl.py` is the right fix and I agree it belongs there rather than here — validating what is accepted, not compensating after storing it.

On whether this function should report its outcome: it now does, without a signature change. `_fail_unresumable_task_instance` sets `next_method` to `TRIGGER_FAIL_REPR`, so a caller can check `ti.next_method == TRIGGER_FAIL_REPR` immediately after the call — which is exactly what the sweep now does for its summary counts. The HITL route could use the same check to answer something other than 200. I have not made that change here because it alters a public endpoint's response, which is not something I want to slip into a 3.3.1 backport, but the mechanism is in place for it.

---

airflow-core/src/airflow/models/trigger.py
potiuk: Removed — both restated the line directly beneath them.

---

airflow-core/src/airflow/models/trigger.py
potiuk: Removed — both restated the line directly beneath them.

---

airflow-core/tests/unit/jobs/test_scheduler_job.py
potiuk: Trimmed — the docstring keeps the intent in one line and the inline narration is gone.

---

airflow-core/tests/unit/models/test_trigger.py
potiuk: Trimmed — the docstring keeps the intent in one line and the inline narration is gone.

---

potiuk:
The diff has changed materially since both of your reviews, so flagging for a fresh look rather than letting the existing approval carry it.

@amoghrajesh — your approval predates @kaxil's review. Since then:

- the traceback now travels in `next_kwargs` via `format_exception`, the way `submit_failure` already does it, because that is the only channel that reaches the task log;
- decoding and re-encoding are separate `try` blocks with distinct messages, so a payload the trigger just yielded is no longer reported as unreadable stored kwargs;
- the docstrings now say this runs in the triggerer too, and record why the `except` has to stay broad;
- the shadowing local `TaskInstanceState` import is gone;
- the `AWAITING_INPUT` sweep counts a task it could not resume separately rather than reporting it as resolved, and the test asserts `(1, 0, 1)`.

@kaxil — the one finding I did not implement is the HITL write-side validation. It is filed as #71036 and linked from a comment on `handle_event_submit`; the reasoning is in that thread. Everything else is addressed above.

None of this is a trivial rewording, so I would rather both of you look again before it goes in.

---

airflow-core/src/airflow/models/trigger.py
```diff
@@ -500,6 +500,60 @@ def get_sorted_triggers(
...
+
+    Mirrors :meth:`Trigger.submit_failure`: the special ``__fail__`` next_method makes the worker
+    fail the task immediately, which runs its normal failure handling (retries and callbacks
+    included). Leaving the task instance parked instead would strand it there, as the event that
+    should have resumed it is already gone.
+
+    The traceback travels in ``next_kwargs`` because that is the only channel that reaches the
+    task log; the exception is otherwise only logged where this runs, which the Dag author may
+    have no access to. It has to stay the list ``format_exception`` returns -- the runtime joins it.
+    """
```
amoghrajesh, line 540:
```suggestion
    """
    Route through ``__fail__`` (mirrors `Trigger.submit_failure`) so a worker fails the
    task normally instead of leaving it stranded with no event left to resume it.

    Traceback goes into ``next_kwargs`` as a list -- the only channel reaching the task log --
    since ``format_exception`` returns it that way and the runtime joins it.
    """
```

amoghrajesh (commented):
Thanks, looks fine to me.

PR #70687, opened 2026-07-29
Add a user settings page for cross-page UI defaults

pierrejeambrun:
Adds a browser-local **Settings** page (reachable from the user menu) that gathers cross-page UI defaults in one place. Several of these behaviors had no way to set a personal default before: graph layout direction and dependency scope are remembered per-Dag, and the clear / mark-as dialog selections reset to a fixed choice on every action, so users kept re-applying the same options.

Each preference is stored in `localStorage` only (no server-side profile) and is consumed as a **fallback** — a per-Dag or per-action choice still overrides it.

Settings included:
- **Graph** — default graph layout direction
- **Clearing** — default run-clear selection; default task-clear selection; "prevent clearing running tasks"
- **Marking** — default mark-as selection (single task instance)

The redundant "Default to graph/grid view" quick-toggle is removed from the user menu — that preference (`default_dag_view`) is already set by the on-page grid/graph switcher, so the menu item was duplicative.

Screenshots:
Screenshot 2026-07-29 at 17 13 11
Screenshot 2026-08-10 at 12 52 58

---

### DIFF
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
@@ -88,8 +88,6 @@
   "dagRunId": "Dag Run ID",
   "dagRunState": "Dag Run State",
   "dagWarnings": "Dag warnings/errors",
-  "defaultToGraphView": "Default to graph view",
-  "defaultToGridView": "Default to grid view",
   "delete": "Delete",
   "diff": "Diff",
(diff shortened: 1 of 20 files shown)


bbovenzi:
This is great!

We do have light/dark mode, timezone and language exist outside of this. i think it still makes sense to make those quickly accessible and this is for "Advanced" settings.

bbovenzi (approved):

PR #70688, opened 2026-07-29
Fix Helm Chart regression with Kerberos in Kubernetes

Nataneljpwd:
---

- [x] No

Fixed helm chart regression introduced in #38815 
---
Due to the PR mentioned above, any task that used kerberos and ran kinit by itself has failed due to permission errors, which were caused due to a readonly mount for the kerberos cache on k8s pod template.
The pod template has been adjusted so that when there is a kerberos init container, the mount is readonly for the worker, otherwise, when no kerberos sidecar / init container is set, allow the base container to write to the cache directory

Also added a patch that we use locally for celery to allow for manual kinit per task.


Miretpl (changes requested):
I would add the `readOnlyKerberosCache` option under both the `workers.celery` and `workers.kubernetes` sections (+ closer to the Kerberos container options). I don't really see the reason for having different handling of it for the pod template and workers themselves.

chart/values.yaml
```diff
@@ -609,6 +609,8 @@ kerberos:
 # Airflow Worker Config
 workers:
   celery:
+    # Should the kerberos-ccache mount be readonly for Celery Workers
+    readonlyKerberosCache: true
```
Miretpl: Can you move it closer to the Kerberos-related configs under `workers.celery`?

chart/values.yaml
```diff
@@ -1109,6 +1111,7 @@ workers:
       containerLifecycleHooks: {}
 
     # Resource configuration for pods created with pod-template-file
+    readonlyKerberosCache: true
```
Miretpl, line 1115: Same comment as above.

Miretpl (changes requested):

Miretpl (changes requested):
Could you add test cases for the change?

chart/values.yaml
Nataneljpwd: done, it is under the kerberos related configurations

chart/values.yaml
Nataneljpwd: done, it is under the kerberos related configurations

Miretpl (changes requested):
Tests are failing - could you take a look?

Nataneljpwd:
> Tests are failing - could you take a look?

Hello, fixed th tests

Miretpl (approved):

Miretpl:
@Nataneljpwd could you make a manual backport?

Nataneljpwd:
> @Nataneljpwd could you make a manual backport?

Sure, will do

PR #70690, opened 2026-07-29
Fix CeleryExecutor fork deadlock that hangs the scheduler

dheerajturaga:
`CeleryExecutor` could stop scheduling indefinitely, with `/health` returning 503
until the scheduler was restarted manually.

Publishing workloads forked a `ProcessPoolExecutor` from the scheduler, which is
multi-threaded. `fork()` copies the state of every mutex but only the calling
thread, so a worker could inherit a lock held by a thread that was never copied
and block forever acquiring it. The parent then waited on that worker forever,
the scheduling loop stopped heartbeating, and the job never recovered on its own.

`py-spy dump` of a wedged scheduler:

```
Thread (idle): "MainThread"
    __enter__ (multiprocessing/synchronize.py:95)      <-- SemLock.acquire(), never returns
    get (multiprocessing/queues.py:100)                <-- Queue.get(): "with self._rlock:"
    _process_worker (concurrent/futures/process.py:242)
    ...
    _send_workloads_to_celery (celery_executor.py:256)
    _send_workloads (celery_executor.py:203)
    trigger_tasks (base_executor.py:406)
    heartbeat (base_executor.py:311)
    _run_scheduler_loop (scheduler_job_runner.py:1693)
```

CPython already warns about this from the same code path — running the existing
(quarantined) `test_send_workloads_to_celery_hang` on `main` emits
`This process is multi-threaded, use of fork() may lead to deadlocks in the child`.

The existing `OPERATION_TIMEOUT` guard cannot catch it. That `timeout()` lives
*inside* `send_workload_to_executor`, but the deadlock happens earlier, in the
pool worker's bootstrap (`call_queue.get()`), before the target function is
entered. A `SIGALRM` also cannot interrupt a process blocked in a C-level
semaphore wait.

### Why remove the pool instead of changing the start method

The obvious fix is to pass a non-`fork` `mp_context`. Measured against a live
Redis broker (4 workers warm, `sync_parallelism` auto = 11), that is much worse
than the status quo, because the pool is rebuilt on every publish:

| workloads | inline | fork pool | forkserver pool | persistent fork |
| --- | --- | --- | --- | --- |
| 5 | **3.2ms** | 51.2ms | 392.7ms | 3.6ms |
| 13 | **9.0ms** | 62.8ms | 2899ms | 6.4ms |
| 26 | **15.6ms** | 87.0ms | **24433ms** | 8.7ms |
| 128 | **72.4ms** | 186.4ms | 25981ms | 57.6ms |
| 512 | 364ms | 280ms | 42443ms | **152ms** |

Publishing a workload is a single short broker round-trip (~0.6ms), so the pool
costs more to start than the sends it parallelizes. Inline beats the fork pool at
every size up to 512 — and `[scheduler] max_tis_per_query` defaults to 16, so
real publish batches are far below where the pool could ever win. A per-batch
`forkserver`/`spawn` pool would turn an intermittent hang into a guaranteed
multi-second stall on every publish.

So publishing now happens in the scheduler process. That removes the hazard
rather than making the fork safe, and deletes the pool lifecycle, chunk sizing,
and `BrokenProcessPool` handling that a persistent pool would have required.

`[celery] sync_parallelism` now applies only to fetching task state, which is
what it documents ("How many processes CeleryExecutor uses to sync task state").
Deployments that set it to `1` to work around this hang can drop that override.

### Bulk state fetch

`BulkStateFetcher._get_many_using_multiprocessing` has the identical fork hazard
but genuinely parallelizes blocking network reads, so it keeps its pool and asks
for a non-`fork` start method instead. It honours `[celery] mp_start_method` then
`[core] mp_start_method` so an operator who has deliberately pinned one keeps
control; an explicit `fork` is respected with a warning rather than overridden.
Unset, it selects `forkserver`, else `spawn`.

Note this path is only reached when `[celery] result_backend` is neither a
key-value store nor a database — Redis and DB backends route to
`_get_many_from_kv_backend` / `_get_many_from_db_backend` and never fork.

### Tests

`test_send_workloads_to_celery_hang` was quarantined *and* failing on `main`: its
fixture passed a workload object in the tuple's `team_name` slot, so
`send_workload_to_executor` raised `AttributeError: 'MockWorkload' object has no
attribute 'upper'`. In the forked pool that crash happened out of sight, so the
test was quarantined rather than diagnosed. The fixture is fixed and the test
un-quarantined, since publishing no longer forks.

Added: publishing never starts a worker process (patched `os.fork` and
`ProcessPoolExecutor.__init__` both assert-fail), the state-fetch context
resolution chain including `[celery]` precedence over `[core]` and both fallback
warnings, and that `BulkStateFetcher` passes a non-`fork` context. Each new test
fails without the source change.

related: #8854 (a bulk publish API would remove this pool entirely)
related: #70163 (adjacent CeleryExecutor hang, different cause)

---


providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py
```diff
@@ -487,6 +484,46 @@ def fetch_celery_task_state(async_result: AsyncResult) -> tuple[str, str | Excep
...
+        else:
+            if configured == "fork":
+                log.warning(
+                    "mp_start_method is set to 'fork' for the Celery state-fetch pool. Forking the "
+                    "multi-threaded scheduler can deadlock a worker on an inherited lock and stall "
+                    "scheduling; prefer 'forkserver' or 'spawn'."
+                )
+            return multiprocessing.get_context(configured)
+
+    return multiprocessing.get_context("forkserver" if "forkserver" in available else "spawn")
```
kaxil, line 525: The forkserver context here has no preload set, and `_get_many_using_multiprocessing` builds a fresh pool on every `sync()`. Children are forked from a forkserver that never imported this module, so each pool pays a full airflow + celery import in every worker, and unlike fork it doesn't get cheaper on later pools. Measured with a stand-in module whose import costs 0.8s, 7 workers, 26 items, new pool per call:

```
      fork: sync1=26.2ms   sync2=17.1ms   sync3=9.9ms
forkserver: sync1=873.3ms  sync2=838.4ms  sync3=841.4ms
```

Calling `multiprocessing.set_forkserver_preload([...])` first turns that into 901ms / 19ms / 16ms. That is what `[celery] mp_forkserver_preload` and its `[core]` fallback exist for, and `set_component_mp_start_method` already applies them, but building the context directly here skips that. `rpc://` and `mongodb://` result backends are neither KV nor DB, so they land in this pool on every heartbeat, which is the same per-pool startup cost the forkserver column in your table rejects for publishing. Setting the preload, or hoisting the pool out of the per-sync `with`, would avoid it.

providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py
```diff
@@ -487,6 +484,46 @@ def fetch_celery_task_state(async_result: AsyncResult) -> tuple[str, str | Excep
...
+    block forever acquiring it, which stalls the scheduling loop until the scheduler is
+    restarted. ``forkserver``/``spawn`` workers start without the parent's locks.
+
+    Honours ``[celery] mp_start_method`` (then ``[core] mp_start_method``) so an operator who
+    has deliberately pinned a method keeps control, and falls back to ``forkserver`` then
+    ``spawn``.
+    """
+    available = multiprocessing.get_all_start_methods()
+    configured = None
+    if AIRFLOW_V_3_3_PLUS:
```
kaxil: The provider still declares `apache-airflow>=2.11.0`, so on 2.11 and 3.0 to 3.2 `configured` stays `None` and the pool is pinned to forkserver/spawn with no way to ask for `fork` back. The `provider.yaml` change in this PR tells operators `[celery] mp_start_method` selects this pool's start method, which won't be true on those versions. `resolve_mp_start_method` is only `conf.get(section, ...) or conf.get("core", ...)`, so inlining that lookup with `fallback=None` would keep the escape hatch on every version the provider supports.

providers/celery/src/airflow/providers/celery/executors/celery_executor.py
```diff
@@ -241,22 +229,13 @@ def _send_workloads(self, workload_tuples_to_send: Sequence[WorkloadInCelery]):
...
-                send_pool.map(send_workload_to_executor, workload_tuples_to_send, chunksize=chunksize)
-            )
-        return key_and_async_results
+        # Publish in this process rather than in a worker pool. Publishing a workload is a single
+        # short broker round-trip, so a pool costs more to start than the sends it parallelizes at
+        # any batch size the scheduler produces ([scheduler] max_tis_per_query defaults to 16).
+        # Forking the pool from the multi-threaded scheduler was also a deadlock risk: a worker
+        # could inherit a mutex held by a thread fork() did not copy and block forever acquiring
+        # it, stalling the scheduling loop with no way to recover short of a restart.
+        return list(map(send_workload_to_executor, workload_tuples_to_send))
```
kaxil, line 238: The benchmarks all assume a healthy broker. Each send arms a `[celery] operation_timeout` alarm (default 1.0s), so a serial batch's worst case is `len(batch) * 1.0s` on the scheduler loop, where the pool divided it by `sync_parallelism` (batch of 32 over 11 processes was chunksize 3, so ~3s). `trigger_tasks` passes the executor's open slots, so batch tracks `[core] parallelism` (default 32), and goes well past that for anyone who raised `max_tis_per_query`. 32s without a heartbeat is over `[scheduler] scheduler_health_check_threshold` (30), which is the 503 this PR is fixing, and `task_publish_max_retries` (default 3) puts the timed-out keys back for the next heartbeat to stall on again. Worth bounding total publish time per heartbeat, or at least calling out the tradeoff?

providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py
```diff
@@ -407,13 +408,9 @@ def send_workload_to_executor(
...
     Send workload to executor (serialized and executed as a Celery task).
 
-    This function runs either inline in the long-lived scheduler process (single-workload or
-    sync_parallelism=1 path) or in short-lived ProcessPoolExecutor subprocesses (multi-workload
-    path). To avoid pickling issues with team-specific Celery apps, we pass the team_name and
-    create the app at call time. The cached app lives for the duration of the caller process, so
-    the main benefit is the scheduler-inline path where the cache persists across publish cycles.
-    In the ProcessPoolExecutor path, each subprocess is recreated per publish batch and the cache
-    only lasts for that single batch.
+    This runs inline in the long-lived caller process (the scheduler, when publishing). The
```
kaxil, line 411: `_get_celery_app_for_workload`'s docstring a few functions up still has the old framing ("may run either inline in the scheduler process or in a publisher subprocess"). Worth updating alongside this one.

dheerajturaga:
@kaxil Thanks for the careful review — all four are fair. Pushed fixes for three; the fourth is a design question I'd like your steer on.

**Version gate (`resolve_mp_start_method`).** Fixed. You're right that gating on 3.3+ silently dropped the `mp_start_method` config on 2.11–3.2 and made the `provider.yaml` text false there. Now reads `conf.get("celery", ...) or conf.get("core", ...)` with `fallback=None` directly, so the escape hatch works on every supported version.

**forkserver with no preload / fresh pool per `sync()`.** Fixed by making the pool persistent on `BulkStateFetcher` — built lazily, reused across syncs, shut down in `end()`/`terminate()`, and recreated if a worker death breaks it. One thing worth flagging: in my measurements `set_forkserver_preload` alone did *not* amortize when a fresh pool was built each sync — each worker still re-imported (I confirmed via a per-worker PID marker that the preloaded module wasn't inherited). Hoisting the pool out of the per-`sync()` scope is what actually removes the cost, so I went with that. Happy to be shown otherwise if you've seen preload help across distinct pools.

**Stale docstring.** Fixed.

**Bounding publish time per heartbeat.** This is the one I'd like your call on. You're right that the batch tracks `[core] parallelism` (default 32) via `open_slots`, not `max_tis_per_query`, so an unresponsive broker gives ~`parallelism * operation_timeout` ≈ 32s serial, over the 30s health threshold, with `task_publish_max_retries` re-queuing the timed-out keys. I confirmed that returning only a published prefix is safe: `_get_workloads_to_schedule` only reads `queued_tasks`, and the pop happens per-returned-result in `_send_workloads`, so an un-sent remainder stays queued and is retried next heartbeat without consuming slots. My inclination is a wall-clock publish budget checked between sends (each send is already bounded by its own `operation_timeout` alarm), with the remainder deferred — defaulting to something like 15s so healthy brokers never hit it. Would you prefer that, a simpler per-heartbeat count cap, or just documenting the tradeoff? And if the budget, do you want it protective-by-default or opt-in?

---
Drafted-by: Claude Code (Opus 5); reviewed by @dheerajturaga before posting

CreeperBeatz:
Independent production reproduction of this, on released versions, in case a real-world
data point is useful while the PR is under review.

**Versions:** Airflow 3.3.0, `apache-airflow-providers-celery` 3.21.0, celery 5.6.3,
kombu 5.6.2, redis-py 6.4.0, Python 3.12, 6 cores, redis broker local to the scheduler,
Postgres result backend. `[celery] sync_parallelism` unset, so it resolved to
`max(1, cpu_count()-1)` = 5.

The scheduler stopped scheduling at 08:20:09 UTC and stayed wedged for 23 minutes until
restarted by hand. Its last log line is the enqueue itself:

```
08:20:09 Trying to enqueue tasks: [...] for executor: CeleryExecutor(parallelism=256)
```

Rather than py-spy I captured `/proc/<pid>/stack` from the host, which shows the same
thing from the kernel side and may be a useful complement to the dump in the description:

```
child 2747820 -- Threads: 1
  [<0>] futex_do_wait+0x3a/0x70      <-- waiting on a lock, but the process has ONE thread,
  [<0>] __futex_wait+0x99/0x100          so no thread in it can ever release that lock
  [<0>] futex_wait+0x72/0x120

child 2747819 -- Threads: 1
  [<0>] ksys_read+0xc6/0xf0          <-- blocked reading the result pipe

parent 2774404 -- Threads: 2
  [<0>] futex_do_wait+0x3a/0x70      <-- waiting on the pool
```

Both children were forked at 08:20:10, matching the last log line exactly. The single-thread
child blocked in `futex_wait` is, I think, the clearest possible signature of the inherited-lock
mechanism described here: there is no owning thread left in that process to release it.

Supporting evidence that nothing else was involved:

- redis healthy: 2.6 MB used, every queue at depth 0, `rejected_connections:0` - the
  workloads were never published
- Postgres: no lock waits, no long-running queries, connection count well under the limit
- `SchedulerJob` heartbeat age 1036 s, while `TriggererJob` and `DagProcessorJob` on the same
  host kept beating every 5-7 s
- the host itself was not resource constrained - CPU load and free memory both well within
  normal for this deployment

Two notes that may be worth folding into the docs part of this change:

**1. Nothing can restart a scheduler in this state** 

The process never exits, so a container restart policy never
fires and the orchestrator reports a perfectly healthy container throughout. Meanwhile
`[scheduler] enable_health_check` defaults to `False`, so there is no scheduler-side
liveness signal either, and the webserver's `/api/v2/monitor/health` kept reporting the
scheduler healthy right up until we queried the `Job` heartbeat directly. The failure is
therefore invisible to every automatic recovery path at once, and only manual intervention
ended it. A line in the docs pointing out that this hang cannot be caught without
`enable_health_check` (or an `airflow jobs check --job-type SchedulerJob` probe) would have
saved us the outage.

**2. `[celery] sync_parallelism = 1` is an effective mitigation on the released versions above, and is what we deployed.** 

It takes the existing in-process branch and never forks,
so the hazard is gone rather than made less likely. Worth noting the cost is close to zero
in a setup like ours: publishing is one short round-trip to a local broker,
`[scheduler] max_tis_per_query` defaults to 16 so batches are far below where a pool could
pay back its own startup, and with a `DatabaseBackend` result backend `BulkStateFetcher`
already uses the bulk `SELECT ... WHERE task_id IN (...)` path, so lowering it has no effect
on state fetching at all. That last point may be worth stating explicitly for anyone on an
affected version who wants to mitigate before this lands.

Issue #50708, opened 2025-05-16
Small memory leak from the dag-processor in Airflow 3.x


zachliu:
### Apache Airflow version

3.x

### If "Other Airflow 2 version" selected, which one?

N/A

### What happened?

There was a significant memory leak (1 Gb/hour depending on DAGs), tracked in https://github.com/apache/airflow/issues/49887 and https://github.com/apache/airflow/issues/50097, which has been addressed by https://github.com/apache/airflow/pull/50558 or via this patch https://github.com/apache/airflow/issues/49887#issuecomment-2868630859. 

However, one (maybe more) much smaller leak(s) remain—approximately 20 MB/hour—in the `dag-processor`

![Image](https://github.com/user-attachments/assets/fc5e660b-0253-42e7-9744-55f67347f01e)

### What you think should happen instead?

Memory usage of the `dag-processor` shall be stable.

### How to reproduce

Option 1: Run Airflow with only the example DAGs (active or pause, doesn't matter) and observe the growth of memory usage.

Option 2: Run `airflow version` command in a loop in any airflow container, and then observe the growth of memory usage.

```bash
#!/bin/bash

while true; do
  airflow version
  sleep 0.5
done
```

### Operating System

LM 21.3 with docker python:3.12-slim-bookworm

### Deployment

Docker-Compose

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

tirkarthi:
Does checking fd give any clue like are these sockets? It will be helpful if it can be reproduced by a standalone simple dag like last one related to `Variable`/`Connection`.

```
watch "pgrep -f dag-processor | xargs -I{} ls -l /proc/{}/fd/ >> fd.log"
```

Thanks

zachliu:
@tirkarthi i think have identified at least two leaks and they don't seem to show up in fd:
1. caused by these two library imports. this leak majorly affects the `worker`. it has little to no impact in other services unless you're using `["CMD-SHELL", 'airflow ...'] ` in docker-compose `healthcheck`
    ```python
    from importlib import metadata
    from sqlalchemy import create_engine
    ```

    to reproduce, run this in any airflow docker container
    ```bash
    #!/bin/bash

    while true; do
      # both of the following can show the leak
      # airflow version
      python test.py  # test.py contains the above two imports
[...]

PR #70694, opened 2026-07-29
Bump structlog>=26.1.0 and croniter>=6.2.2 to fix memory leaks

jason810496:
- related issue: #50708
- related comment: https://github.com/apache/airflow/issues/50708#issuecomment-4750514092
- upstream fixes: hynek/structlog#807 (released in structlog 26.1.0), pallets-eco/croniter#185 (released in croniter 6.1.0, fully effective from 6.2.0; leap-year `get_prev` fix in 6.2.1)

## Why

The dag-processor memory leak in #50708 has two library-level root causes that are already fixed upstream: structlog leaks one lock and file-handle entry in `WRITE_LOCKS` per Dag-file log reopen on every parse cycle (fixed in 26.1.0), and croniter leaks on every cron schedule evaluation. The current minimums (`structlog>=25.4.0`, `croniter>=2.0.2`) still admit the leaky versions, so constraint-less installs and release-branch constraints can resolve them. Every released Airflow to date ships a leaky structlog (3.2.2 constraints pin 25.5.0).

The croniter floor is 6.2.2 rather than 6.1.0: the 6.1.0 wheel still ships the `EXPRESSIONS` global cache (removed in 6.2.0 despite the 6.1.0 changelog claim), and 6.2.1 fixed `get_prev` skipping Feb 29 on leap years, which feeds `CronMixin._get_prev` via `infer_manual_data_interval` and `_align_to_prev`. `constraints-3.2.2` and `constraints-3.3.0` already pin `croniter==6.2.2`.

Verified empirically: on 3.2.2-era versions with example Dags only, the standalone stack leaks ~34 MB/h of anonymous memory concentrated in the dag-processor, while a main-based image already running structlog 26.1.0 and croniter 6.2.x is near-flat (~9 MB/h residual across 8 processes). `constraints-main` already resolves the fixed versions, so this bump changes nothing for main CI and only raises the floor.

## What

- Bump `structlog>=25.4.0` to `structlog>=26.1.0` in all 8 declaring distributions: airflow-core, task-sdk, airflow-ctl, and shared dagnode, listeners, logging, observability, providers_discovery.
- Bump `croniter>=2.0.2` to `croniter>=6.2.2` in airflow-core.
- Remove workarounds in `shared/logging` that are dead code at the new structlog floor: the `WRITE_LOCKS` WeakKeyDictionary monkeypatch in `configure_logging` (structlog 26.1.0 ships it natively) and the `_BYTES_LOGGER_HAS_NAME` fallback in `NamedBytesLogger` (`BytesLogger` always has the `name` slot now). The `_WRITE_LOGGER_HAS_NAME` detection stays, since `WriteLogger` has not received the upstream `name` slot yet.
- Regenerate `uv.lock` (requirement metadata only, locked versions unchanged: structlog 26.1.0, croniter 6.2.4).

---


airflow-core/pyproject.toml
```diff
@@ -85,7 +85,7 @@ dependencies = [
     "cadwyn>=6.1.1",
     "colorlog>=6.8.2",
     "cron-descriptor>=1.2.24",
-    "croniter>=2.0.2",
+    "croniter>=6.1.0",
```
kaxil: `6.1.0` picks up the `TIMESTAMP_TO_DT_CACHE` removal, but the `EXPRESSIONS` global dict cache is still in the 6.1.0 wheel. It only went away in 6.2.0 (the 6.1.0 changelog section claims that fix, but the shipped code disagrees).

6.2.1 then fixed `get_prev` skipping Feb 29 on leap years for day-of-month expressions, and `CronMixin._get_prev` feeds straight into `infer_manual_data_interval` and `_align_to_prev`:

```
# croniter 6.1.0
croniter("0 0 29 * *", start_time=datetime(2024, 3, 5)).get_prev(datetime) -> 2024-01-29
# croniter 6.2.1+                                                          -> 2024-02-29
```

`constraints-3.2.2` and `constraints-3.3.0` already pin `croniter==6.2.2`, so could the floor be `>=6.2.2` instead?

shared/logging/pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 dependencies = [
     "msgspec>=0.19.0",
     "pygtrie>=2.5.0",
-    "structlog>=25.4.0",
+    "structlog>=26.1.0",
```
kaxil, line 29: With the floor at 26.1.0, the `WRITE_LOCKS` monkeypatch in `configure_logging` (`shared/logging/src/airflow_shared/logging/structlog.py`, added in #65121) becomes dead code: 26.1.0 defines `WRITE_LOCKS` as a `WeakKeyDictionary` at module level, so the `isinstance(..., dict)` guard never fires. Its own comment says `TODO: drop once structlog ships the upstream fix (tracked for 26.1.0)`, so this is the PR that gets to remove it. The `_BYTES_LOGGER_HAS_NAME` fallback in the same module is dead too, but `_WRITE_LOGGER_HAS_NAME` is not, since 26.1.0's `WriteLogger` still has no `name` slot.

airflow-core/pyproject.toml
potiuk: As a croniter maintainer who released it - I agree.

kaxil (approved):

airflow-core/pyproject.toml
jason810496: Done in 736d432526

shared/logging/pyproject.toml
jason810496: Done in 736d432526

potiuk (approved):

PR #70699, opened 2026-07-29
UI: Preserve Dag tab when switching Dags

### LINKED ISSUE #57625: Dag view does not remember tab when switching between dags
### Apache Airflow version

3.1.1

_No response_

### What happened?

When you switch between dags the tab resets back to task instances in the grid view. This makes it harder to compare details between dags. This is also a regression from AF2

https://github.com/user-attachments/assets/17bc648a-3419-4116-a434-2a948e498765

_No response_

### How to reproduce

Select a tab that isn't "Task Instances" in the grid view and then switch to a different dag.

### Operating System

Debian

_No response_

### Deployment

Astronomer

### Deployment details

_No response_

### Anything else?

_No response_

### Code of Conduct

shivaam:
Closes: #57625

## What changed

When selecting another DAG from the DAG search dialog, preserve the current DAG-level tab when the destination supports it.

- Preserve standard DAG tabs such as Code, Details, Runs, Tasks, Calendar, Audit Log, and Overview.
- Preserve nested plugin-tab routes.
- Reset task-, run-, and task-instance-specific routes because those entities may not exist in the destination DAG.
- Reset Backfills when the destination DAG is not backfillable.
- Drop query parameters and fragments that belong to the previous DAG.

Backfills need a separate guard because tab availability is DAG-specific. A backfillable source DAG can expose `/backfills`, while the selected destination DAG may report `is_backfillable = false`.

https://github.com/user-attachments/assets/b6172157-195b-47ea-a00a-6f741ef35723

## Testing

- Added focused unit and component regression coverage.
- 64 targeted UI tests pass.
- Verified in Breeze with the real DAG picker across the standard tabs, nested routes, Backfills, query/hash cleanup, same-DAG selection, and browser Back/Forward navigation.

## Screenshots

Before/after screenshots will be attached to this draft.

---

- [X] Yes (Codex)

Generated-by: [Codex](https://github.com/openai/codex) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


choo121600:
Could you add before/after screenshots? It would help reviewers better understand the changes.

shivaam:
> Could you add before/after screenshots? It would help reviewers better understand the changes.

The change is just in the URL. I am not sure if a screenshot will help here as it is navigation change. I can probable add a video or something

shivaam:
Added a screenshot.

airflow-core/src/airflow/ui/src/utils/links.ts
```diff
@@ -54,6 +54,26 @@ export const getRedirectPath = (targetPath: string): string => {
 export const getNextHref = (location: Pick<Location, "hash" | "pathname" | "search">): string =>
   `${location.pathname}${location.search}${location.hash}`;
 
+export const getDagAdditionalPath = (pathname: string, tabValues: Array<string>): string => {
```
bbovenzi: We shouldn't need regex for this. Let's use react-router to more deterministically figure out what dag or task route we are on.

https://reactrouter.com/how-to/using-handle we coudl store both the entity in question and the tab like `handle: { entity: "dag", tab: "runs" }` so then if our new page isn't a dag then we don't do anything

shivaam: Thanks that simplified thinks. I replaced the regex with react router

airflow-core/src/airflow/ui/src/utils/links.ts
bbovenzi, line 1: The point of the route handle is that we don't need to do any url string parsing.

bbovenzi: We should eventually replace all of the get__AdditionalPaths with a `getTabPath` where we pass the entit(ies) and the tab string.

shivaam: Updated this to use route handles for regular DAG tabs. Plugin routes now reset to the destination DAG overview because the same plugin page may not exist for the new DAG. Adding compatibility checks for every plugin route would be brittle

shivaam: > We should eventually replace all of the get__AdditionalPaths with a `getTabPath` where we pass the entit(ies) and the tab string.

Yes, I will potentially pick this up as a follow up or will create an issue to track it once this pr is merged

airflow-core/src/airflow/ui/src/router.tsx
```diff
@@ -185,21 +185,22 @@ export const routerConfig = [
...
-          { element: <Events />, path: "events" },
-          { element: <Code />, path: "code" },
-          { element: <DagDetails />, path: "details" },
+          { element: <Events />, handle: { entity: "dag", tab: "events" }, path: "events" },
+          { element: <Code />, handle: { entity: "dag", tab: "code" }, path: "code" },
+          { element: <DagDetails />, handle: { entity: "dag", tab: "details" }, path: "details" },
           pluginRoute,
         ],
         element: <Dag />,
+        handle: { entity: "dag" },
```
bbovenzi:
```suggestion
```

Dead code

airflow-core/src/airflow/ui/src/utils/links.ts
```diff
@@ -54,6 +54,49 @@ export const getRedirectPath = (targetPath: string): string => {
 export const getNextHref = (location: Pick<Location, "hash" | "pathname" | "search">): string =>
   `${location.pathname}${location.search}${location.hash}`;
 
+type RouteMatch = {
+  readonly handle: unknown;
+  readonly params: Record<string, string | undefined>;
```
bbovenzi: We're never reading `match.params`

airflow-core/src/airflow/ui/src/utils/links.ts
```diff
@@ -54,6 +54,49 @@ export const getRedirectPath = (targetPath: string): string => {
...
+};
+
+type TabRouteHandle = {
+  readonly entity: string;
+  readonly tab?: string;
```
bbovenzi: These should probably be enums for better type safety.

bbovenzi (approved):

PR #70703, opened 2026-07-29
Add create and update support to Snowflake Cortex Agent hook

SameerMesiah97:
**Description**

This change extends `SnowflakeCortexAgentHook` with support for creating and updating Snowflake Cortex Agent Objects.

The hook now exposes `create_agent()` and `update_agent()` methods using the Snowflake Cortex Agent REST API. A shared helper constructs agent definition payloads while omitting unset optional fields, reducing duplication between the two operations.

**Rationale**

The existing hook already supports executing, describing, listing and deleting Cortex Agents, but the Snowflake Cortex Agent REST API also provides endpoints for creating and updating Cortex Agent Objects. Exposing these operations completes the core Cortex Agent management lifecycle and allows users to provision and modify agents directly from Airflow without invoking the REST API themselves.

**Notes**

Although the Snowflake documentation describes the update endpoint as returning a JSON status object, the endpoint can return a successful response with an empty body. The internal request helper now supports empty responses when explicitly enabled, and `update_agent()` opts into this behavior. Other hook methods continue to reject unexpected empty responses.

**Tests**

- `create_agent()` sends the expected request for all supported `CreateMode` values and omits unset optional fields from the request payload.
- `create_agent()` and `update_agent()` URL-encode dynamic endpoint path segments.
- `update_agent()` sends the expected request and returns an empty dictionary for a successful response with an empty body.
- Empty update responses are handled without attempting JSON deserialization.

**Backward Compatibility**

This change is fully backward compatible.

The new hook methods are additive and do not modify the behavior or signatures of existing public APIs. Existing consumers of `run_agent()`, `describe_agent()`, `list_agents()` and `delete_agent()` are unaffected.

Generated-by: [GPT 5.5] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
+++ b/providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
@@ -17,11 +17,21 @@
 
 from __future__ import annotations
 
+from enum import Enum
+from json.decoder import JSONDecodeError
 from typing import Any, cast
 
 import requests
 
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+
+
+class CreateMode(str, Enum):
+    """Resource creation modes for Cortex Agents."""
+
+    ERROR_IF_EXISTS = "errorIfExists"
+    OR_REPLACE = "orReplace"
+    IF_NOT_EXISTS = "ifNotExists"
 
 
 class SnowflakeCortexAgentHook(SnowflakeHook):
@@ -70,16 +80,64 @@ class SnowflakeCortexAgentHook(SnowflakeHook):
             timeout=timeout,
         )
 
-        if response.status_code >= 400:
+        if not response.ok:
             self.log.error(
                 "Snowflake Cortex Agent request failed with status %s: %s",
                 response.status_code,
                 response.text,
             )
 
+            try:
+                error = response.json()
+                raise RuntimeError(
+                    f"Snowflake Cortex Agent API request failed: {error.get('message', response.text)}"
+                )
+            except JSONDecodeError:
+                response.raise_for_status()
+
+        if not response.content:
+            return {}
+
         response.raise_for_status()
 
         return response.json()
+
+    def _build_agent_payload(
+        self,
+        *,
+        comment: str | None = None,
+        profile: dict[str, Any] | None = None,
+        models: dict[str, Any] | None = None,
+        instructions: dict[str, Any] | None = None,
+        orchestration: dict[str, Any] | None = None,
+        tools: list[dict[str, Any]] | None = None,
+        tool_resources: dict[str, Any] | None = None,
+    ) -> dict[str, Any]:
+        """Build a Cortex Agent request payload."""
+        payload: dict[str, Any] = {}
+
+        if comment is not None:
+            payload["comment"] = comment
+
+        if profile is not None:
+            payload["profile"] = profile
+
+        if models is not None:
+            payload["models"] = models
+
+        if instructions is not None:
+            payload["instructions"] = instructions
+
+        if orchestration is not None:
+            payload["orchestration"] = orchestration
+
+        if tools is not None:
+            payload["tools"] = tools
+
+        if tool_resources is not None:
+            payload["tool_resources"] = tool_resources
+
+        return payload
 
     def run_agent(
         self,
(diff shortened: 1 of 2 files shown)


SameerMesiah97:
Depends on PR #70101. Please do not review this until that is merged.

A separate PR was opened for these 2 methods as I wanted to gather feedback on the original management methods before extending the hook further with these less commonly used management APIs.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -215,6 +267,114 @@ def run_agent(
...
+        tools: list[dict[str, Any]] | None = None,
+        tool_resources: dict[str, Any] | None = None,
+        timeout: int | None = 600,
+    ) -> JsonDict:
+        """
+        Update a Snowflake Cortex Agent.
+
+        :param database: Database containing the agent.
+        :param schema: Schema containing the agent.
+        :param agent_name: Name of the Cortex Agent.
```
potiuk, line 349: `update_agent` accepts `comment`, `profile`, `models`, `instructions`, `orchestration`, `tools` and `tool_resources`, but the docstring goes straight from `agent_name` to `timeout` — seven of the eleven parameters are undocumented. `create_agent` a few lines up documents all of them, and these render in the published provider API reference, so `update_agent` would ship with seven undocumented arguments next to a sibling that documents them.

```suggestion
        :param agent_name: Name of the Cortex Agent.
        :param comment: Optional comment. Optional. Defaults to ``None``.
        :param profile: Agent profile configuration. Optional. Defaults to ``None``.
        :param models: Model configuration. Optional. Defaults to ``None``.
        :param instructions: Agent instructions. Optional. Defaults to ``None``.
        :param orchestration: Orchestration configuration. Optional. Defaults to ``None``.
        :param tools: Agent tools. Optional. Defaults to ``None``.
        :param tool_resources: Tool resource configuration. Optional. Defaults to ``None``.
```

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -215,6 +267,114 @@ def run_agent(
...
+        endpoint = (
+            f"/api/v2/databases/{quote(database, safe='')}"
+            f"/schemas/{quote(schema, safe='')}"
+            f"/agents/{quote(agent_name, safe='')}"
+        )
+
+        return self._request(
+            method="PUT",
+            endpoint=endpoint,
+            payload=self._build_agent_payload(
```
potiuk, line 364: This is the one I'd like confirmed before the PR lands.

`_build_agent_payload` omits every unset field, which makes `update_agent(comment="new comment")` *look* like a partial patch. But `PUT /agents/{name}` is REST-conventionally a full replace — if Snowflake implements it that way, that call silently clears the agent's `tools`, `models`, `instructions` and `orchestration`. I couldn't find anything in [Snowflake's Cortex Agents REST API docs](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-agents-rest-api) that states which semantics apply.

That's the difference between "change the comment" and "delete the agent's tool configuration", and the current signature actively encourages the dangerous reading. Could you check against a live account and write the answer into the docstring? If it turns out to be a replace, this probably wants an explicit warning in the docstring, or a `describe_agent`-then-merge path so callers can't lose configuration by omission.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -215,6 +267,114 @@ def run_agent(
...
+        schema: str,
+        agent_name: str,
+        comment: str | None = None,
+        profile: dict[str, Any] | None = None,
+        models: dict[str, Any] | None = None,
+        instructions: dict[str, Any] | None = None,
+        orchestration: dict[str, Any] | None = None,
+        tools: list[dict[str, Any]] | None = None,
+        tool_resources: dict[str, Any] | None = None,
+        create_mode: CreateMode = CreateMode.ERROR_IF_EXISTS,
```
potiuk, line 283: `create_mode.value` below means `create_agent(create_mode="orReplace")` fails with `AttributeError: 'str' object has no attribute 'value'`.

`CreateMode` subclasses `str` precisely so the raw API strings are usable, and a plain string is what a config-driven or templated caller naturally produces. Widening the annotation and normalising at the call site accepts both, and turns a typo into a clear `ValueError` instead of a malformed `createMode` query param that Snowflake rejects:

```suggestion
        create_mode: CreateMode | str = CreateMode.ERROR_IF_EXISTS,
```

…paired with `params={"createMode": CreateMode(create_mode).value}` in the `_request` call below.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
```diff
@@ -124,6 +139,43 @@ def _request(
 
         return data
 
+    def _build_agent_payload(
```
potiuk, line 142: Nit, take it or leave it: this never touches `self`, so it could be a `@staticmethod`.

providers/snowflake/tests/unit/snowflake/hooks/test_snowflake_cortex_agent.py
```diff
@@ -381,6 +384,157 @@ def test_get_text_response(
     ):
         assert SnowflakeCortexAgentHook.get_text_response(response) == expected
 
+    @pytest.mark.parametrize(
+        ("create_mode", "comment", "profile", "instructions", "expected_params", "expected_payload"),
```
potiuk, line 388: `models`, `orchestration`, `tools` and `tool_resources` are never passed non-`None` in either `test_create_agent` or `test_update_agent`, so four of the seven branches in the new `_build_agent_payload` are never executed.

From [`AGENTS.md` § Testing Standards](https://github.com/apache/airflow/blob/main/AGENTS.md#testing-standards):

> Target exactly 100% coverage of what the PR changes — no more, no less. Every changed or added behaviour must have a test; every test must fail without the PR's change.

Adding those four to the existing `default` param case (argnames tuple plus the one case's values and `expected_payload`) covers them without a new test.

potiuk (commented):
Thanks for splitting this out — the shape matches the existing hook methods well, and I checked the endpoint details against Snowflake's REST API docs: `createMode` as a query param with those three values, and `name` correctly absent from the PUT body, are both right.

Five things, one of which I'd like your input on before this lands.

The one that matters: **what are the update semantics?** `_build_agent_payload` omits unset fields, so `update_agent(comment="x")` reads like a partial patch — but `PUT /agents/{name}` is conventionally a full replace, which would mean that call silently wipes the agent's `tools`, `models` and `instructions`. Snowflake's public docs don't say which it is. Could you confirm against a live account and record the answer in the docstring? If it *is* a replace, the method probably needs a louder warning, or a `describe`-then-merge path.

The rest are in the inline comments: the `update_agent` docstring is missing seven of its parameters (`create_agent` three lines up documents all of them), `create_mode.value` breaks if a caller passes the plain string `"orReplace"`, and four of the seven new payload branches aren't exercised by the tests.

Also: your "don't review until #70101 is merged" comment is now stale — that one landed, and the stale bot has since flagged this PR. Worth a rebase/comment to keep it alive.

---

>

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: Done. I have not applied the exact same suggestion but docstring coverage has been increased as requested.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: I tested the API and PUT supports partial updates so for e.g. updating the only comment field is not going to overwrite the other fields.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: Applied.

providers/snowflake/src/airflow/providers/snowflake/hooks/snowflake_cortex_agent.py
SameerMesiah97: Applied.

providers/snowflake/tests/unit/snowflake/hooks/test_snowflake_cortex_agent.py
SameerMesiah97: The test for the create method now includes all the parameters you mentioned

PR #70704, opened 2026-07-29
Do not import the stored path when rebuilding a Callback from serialized data

potiuk:
`Callback.get_callback_path` imports the module named by a dotted-path string in
order to check that it resolves to a callable. That check is deliberately best
effort — the `ImportError` is logged and swallowed and the path is returned
either way, because the callable may only exist on the host that will eventually
run it.

Rebuilding a `Callback` from its serialized form went through that same path, so
deserializing one imported the module named in the stored data. The path was
already checked when the `Callback` was first created, and reconstruction happens
in components that never invoke the callback themselves, so importing there is
neither needed nor wanted — it runs module-level code of whatever the stored
string happens to name.

This carries a path read back from serialized data in a private `str` subclass
and returns it unchanged instead of resolving it a second time.

Construction from a Dag author's callable or dotted path is unaffected, including
the existing import-based callable check. The dot-path shape check still applies
to stored paths too, so a malformed stored path is still rejected.

### Test plan

- [x] `test_init_imports_the_module_named_by_the_path` — positive control: Dag-author construction still resolves the path
- [x] `test_deserialize_does_not_import_the_module_named_by_the_stored_path` — and the `AsyncCallback` equivalent
- [x] `test_serde_deserialize_does_not_import_the_module_named_by_the_stored_path` — full route through serde
- [x] `test_deserialize_rejects_a_path_which_is_not_a_dot_path` — shape check preserved on stored paths
- [x] `test_deserialize_keeps_a_stored_path_which_no_longer_points_at_a_callable`
- [x] Tests use a real fixture module whose import has an observable side effect, rather than mocking `import_string`

### DIFF
--- a/task-sdk/src/airflow/sdk/definitions/callback.py
+++ b/task-sdk/src/airflow/sdk/definitions/callback.py
@@ -28,6 +28,20 @@ from airflow.sdk._shared.module_loading import import_string, is_valid_dotpath
 log = structlog.getLogger(__name__)
 
 
+class _SerializedCallbackPath(str):
+    """
+    A callback path which was read back from serialized data.
+
+    The path of a ``Callback`` is checked when the ``Callback`` is first created. Rebuilding
+    that ``Callback`` from its serialized form carries the path in this wrapper so that
+    :meth:`Callback.get_callback_path` keeps it as-is instead of resolving it a second time.
+
+    :meta private:
+    """
+
+    __slots__ = ()
+
+
 class Callback(ABC):
     """
     Base class for Deadline Alert callbacks.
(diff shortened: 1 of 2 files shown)


task-sdk/src/airflow/sdk/definitions/callback.py
```diff
@@ -66,6 +80,14 @@ def get_callback_path(cls, _callback: str | Callable) -> str:
 
         stripped_callback = _callback.strip()
 
+        if isinstance(_callback, _SerializedCallbackPath):
+            # This path was already checked when the Callback it belongs to was created, so
+            # rebuilding that Callback keeps it as it is. Reconstruction happens in components
+            # which never call the callback themselves, and resolving the path there is neither
+            # needed nor wanted: the module it names is not necessarily importable in that
+            # process, and importing it has no bearing on the path we return either way.
+            return stripped_callback
```
amoghrajesh, line 87: Both the docstring and usage almost narrate the same thing, suggest removing in one of them

amoghrajesh (approved):

task-sdk/src/airflow/sdk/definitions/callback.py
potiuk: Right — it was narrated three times: the class docstring, the branch in `get_callback_path`, and again in `deserialize`. Kept it at `get_callback_path`, where the decision is actually made and where the non-obvious part lives (the module named by a stored path is not necessarily importable in the process doing the reconstruction). The class docstring is now one line pointing there, and the `deserialize` comment is gone.

---

PR #70711, opened 2026-07-30
UI: Keep task try history consistent when switching tasks

### LINKED ISSUE #70696: inconsistencies on showing previous tries in the airflow ui

Airflow Core

### Apache Airflow version

3.3.0

I have a dag that always run task `start` and then it randomly runs eitther `task_1` or `task_2` (see dag below)
I then restart the same dag run 4 times.
I see this for `start` as expected
Image

If i click around on the other tasks a few times i.e. click task_1 then 2 then 1 then click back on start, i now see:
where i see try number 2 4 times?

Image

Refreshing the page makes it go back to normal.
if i then want to see `task_1` i see only 3 tries even though i should see 4, and two of them are task 2 repeated?

Image

if i then want to see `task_2` i see only 3 tries instead of 4 and include try number 0, which shouldn't even exist

Image

example dag:
```

import random

@dag
def example_dag():
    @task.branch
    def start():
        return f"task_{random.randint(1, 2)}"

    @task
    def task_1():
        print("Task 1 executed")

    @task
    def task_2():
        print("Task 2 executed")

    @task
    def end():
        print("End task executed")
        
    start() >> [task_1(), task_2()] >> end()


example_dag()
```

All 3 tasks should display 4 tries. with no repeat try numbers, and no try number 0.

### Operating System

docker 

### Deployment

Docker-Compose

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

shivaam:
Task try history responses can contain duplicate try numbers, non-attempt try `0` records, or records that do not match the task currently displayed. The UI previously rendered the response directly, allowing attempts to appear duplicated or mixed after navigating between tasks.

This change scopes history to the active Dag run, task, and map index; removes nonpositive attempts; deduplicates by try number; and treats the current task instance as authoritative. Query keys and caching behavior remain unchanged.

A focused component regression test covers repeated task switching, duplicate and zero tries, mapped task indices, current-attempt precedence, and attempt selection.

Validation:

- Focused Vitest: 2 tests passed
- UI ESLint: passed
- UI TypeScript check: passed
- Airflow pre-commit hooks: passed
- Real UI verification with the issue's Dag and repeated-run workflow

closes: #70696

after-start-tries

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/TaskTrySelect.tsx
+++ b/airflow-core/src/airflow/ui/src/components/TaskTrySelect.tsx
@@ -71,10 +71,16 @@ export const TaskTrySelect = ({ onSelectTryNumber, selectedTryNumber, taskInstan
   const logAttemptDropdownLimit = 10;
   const showDropdown = finalTryNumber > logAttemptDropdownLimit;
 
-  // For some reason tries aren't sorted by try_number
-  const sortedTries = [...(tiHistory?.task_instances ?? [])].sort(
-    (tryA, tryB) => tryA.try_number - tryB.try_number,
+  const triesByNumber = new Map(
+    (tiHistory?.task_instances ?? []).filter((ti) => ti.try_number > 0).map((ti) => [ti.try_number, ti]),
   );
+
+  if (finalTryNumber > 0 && state !== "up_for_retry" && state !== null) {
+    // The current task instance is authoritative when it is also present in history.
+    triesByNumber.set(finalTryNumber, taskInstance);
+  }
+
+  const sortedTries = [...triesByNumber.values()].sort((tryA, tryB) => tryA.try_number - tryB.try_number);
 
   const tryOptions = createListCollection({
     items: sortedTries.map((ti) => ({
--- a/airflow-core/src/airflow/ui/src/components/TaskTrySelect.test.tsx
+++ b/airflow-core/src/airflow/ui/src/components/TaskTrySelect.test.tsx
@@ -18,8 +18,8 @@
  */
 import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import { render, screen } from "@testing-library/react";
-import type { PropsWithChildren } from "react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import type { PropsWithChildren, ReactNode } from "react";
 import { MemoryRouter } from "react-router-dom";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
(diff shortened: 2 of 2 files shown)


Andrushika:
Thanks! The current fix focuses on frontend filtering, and the description mentions that "task try history responses can contain duplicate try numbers". Is that intended behavior from the backend endpoint? If not, would it make more sense to fix it at the backend instead of the frontend?

shivaam:
> Thanks! The current fix focuses on frontend filtering, and the description mentions that "task try history responses can contain duplicate try numbers". Is that intended behavior from the backend endpoint? If not, would it make more sense to fix it at the backend instead of the frontend?

The endpoint intentionally returns both the current task instance and its history. We have unit tests that validate this behavior so believe it is expected.. Changing that would alter the public API and could affect other consumers.

The UI only needs unique, selectable try numbers, so normalizing them in the frontend seems safer for this UI-specific issue.

airflow-core/src/airflow/ui/src/components/TaskTrySelect.tsx
```diff
@@ -70,11 +70,23 @@ export const TaskTrySelect = ({ onSelectTryNumber, selectedTryNumber, taskInstan
...
+          ti.dag_run_id === dagRunId &&
+          ti.task_id === taskId &&
+          ti.map_index === mapIndex &&
+          ti.try_number > 0,
+      )
+      .map((ti) => [ti.try_number, ti]),
   );
 
+  // The current task instance is authoritative when it is also present in history.
+  triesByNumber.set(finalTryNumber, taskInstance);
```
bbovenzi: Let's also wrap this in finalTryNumber > 0 check

airflow-core/src/airflow/ui/src/components/TaskTrySelect.tsx
```diff
@@ -70,11 +70,23 @@ export const TaskTrySelect = ({ onSelectTryNumber, selectedTryNumber, taskInstan
...
-  const sortedTries = [...(tiHistory?.task_instances ?? [])].sort(
-    (tryA, tryB) => tryA.try_number - tryB.try_number,
+  const triesByNumber = new Map(
+    (tiHistory?.task_instances ?? [])
+      .filter(
+        (ti) =>
+          ti.dag_id === dagId &&
+          ti.dag_run_id === dagRunId &&
+          ti.task_id === taskId &&
+          ti.map_index === mapIndex &&
```
bbovenzi: These 4 filters should already be true

airflow-core/src/airflow/ui/src/components/TaskTrySelect.tsx
bbovenzi: Oh and we should also add a check to not set the final number when a TI.state is `up_for_retry` or `null` since they are just placeholders

bbovenzi (changes requested):
Let's rebase now that #70789 is merged

bbovenzi (approved):

Issue #70614, opened 2026-07-28
Scheduler stalls: 3.2 exceeds_max_non_backfill write contends with DAG-processor SELECT … FOR UPDATE on the dag table


dor-bernstein:
### Apache Airflow version

3.2.0+ (observed on 3.2.2)

### What happened

The scheduler intermittently stalls: `scheduler.critical_section_duration` climbs while scheduler CPU drops and task scheduling starves. Inspecting the metadata DB during a stall shows scheduler statements of the form `UPDATE dag SET exceeds_max_non_backfill = … WHERE dag.dag_id = …` blocked with `wait_event_type = Lock` / `wait_event = transactionid`, waiting on the DAG processor's parse transaction.

### Root cause

Before 3.2, the scheduler evaluated `max_active_runs` in memory and did not write the `dag` row during dagrun scheduling. **#60006 ("Separate 'next dag run' from 'max active runs'", merge commit `32bc0119b408c9a37841edd021b646324a02745e`, shipped in 3.2.0)** added the `dag.exceeds_max_non_backfill` column (migration `0097_3_2_0_add_exceeds_max_runs_flag_to_dag_model`) and made the scheduler **write it to the `dag` row** during dagrun scheduling:

- Write: `airflow-core/src/airflow/jobs/scheduler_job_runner.py::_set_exceeds_max_active_runs` (called from `_create_dag_runs` and `_start_queued_dagruns`).

That write now competes with a pre-existing lock held by the DAG processor for the entire per-file parse transaction:

- Lock: `airflow-core/src/airflow/dag_processing/collection.py::DagModelOperation.find_orm_dags` → `with_row_locks(select(DagModel) …)` i.e. `SELECT … FOR UPDATE` on every `dag` row for a parsed file, held through `update_dags` + serialization + permission sync until the transaction commits.

So the scheduler's `UPDATE dag SET exceeds_max_non_backfill = …` blocks on the DAG processor's `FOR UPDATE`. This is amplified when a single parsed file defines many DAGs (e.g. dynamically generated DAGs), because that parse transaction can run for minutes — holding the `dag`-row locks the whole time and stalling the scheduler loop.

### How to reproduce it (conceptually)

1. Use the FAB auth manager and generate many DAGs from a single DAG file.
2. While the DAG processor is parsing that file (holding `SELECT … FOR UPDATE` on its `dag` rows), have the scheduler attempt to update `exceeds_max_non_backfill` for one of those DAGs.
[...]

PR #70712, opened 2026-07-30
Skip redundant dag.exceeds_max_non_backfill writes to reduce scheduler/DAG-processor lock contention

bujjibabukatta:
## Summary
`_set_exceeds_max_active_runs` in `scheduler_job_runner.py` now only
writes `dag_model.exceeds_max_non_backfill` when the value actually
changes, instead of writing it unconditionally on every call.

## Root Cause
As described in #70614, this write can block behind the DAG
processor's `SELECT ... FOR UPDATE` lock (`find_orm_dags` in
`collection.py`), which is held for the whole per-file parse
transaction. Since the flag only flips to `True` when a DAG is at its
`max_active_runs` limit, most dagrun completions don't actually change
it — skipping the write in that case avoids the lock contention for
the common case.

This reduces how often the contention happens; it doesn't remove it
entirely, since the write (and the possible block) still happens when
the flag genuinely transitions. The DAG processor's lock scope itself
is untouched — that's a separate, bigger fix.

related: #70614

- [X] Yes - Claude

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -4212,7 +4212,9 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             )
             active_non_backfill_runs = runs_dict.get(dag_model.dag_id, 0)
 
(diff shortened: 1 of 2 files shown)


steveahnahn (commented):
Thanks for the work but I couldn't reproduce the write this removes, notes inline, happy to approve if i'm mistaken.

PR #70714, opened 2026-07-30
Pass Dag runs directly to Deadline references

### LINKED ISSUE #70710: Deadline reference evaluation should receive the DagRun instead of dag_id/run_id
### Body

Custom and built-in deadline references are evaluated via `evaluate_with(session=..., interval=..., dag_id=..., run_id=...)` at Dag run creation.  The two identifiers exist only so the built-in references can pull the Dag run out of the database, but the caller already has it.

`airflow-core/src/airflow/serialization/definitions/dag.py:764` has a TODO noting this (added in #58248):

```python
# TODO : Pretty sure we can drop these last two; verify after testing is complete
dag_id=self.dag_id,
run_id=orm_dagrun.run_id,
```

`DagRunLogicalDateDeadline` and `DagRunQueuedAtDeadline` pass those straight to `_fetch_from_db` (serialization/definitions/deadline.py:361), which queries `SELECT <column> FROM dag_run WHERE dag_id = ? AND run_id = ?`.  That is a query for a column of `orm_dagrun`, which is already loaded in the same session.  `AverageRuntimeDeadline` genuinely needs to query historical runs, but it only needs `dag_id`, which is already available as `orm_dagrun.dag_id`.

There is a second symptom. models/taskinstance.py:252 cannot use `evaluate_with()` at all:

```
# We can't use evaluate_with() since the new queued_at is not written to the DB yet.
```

and replicates the deadline recalculation instead.  That is the DB-read approach failing where you would expect: the in-memory object is ahead of the database, so the `SELECT` returns stale data.  Passing the object directly would let that code path use the normal evaluation logic.

### Proposal

Pass the DagRun (or a small evaluation-context object) into `evaluate_with` instead of loose identifiers:

- Built-in references read attributes directly; `_fetch_from_db` is no longer needed for the two DagRun references
- One fewer query per DagRun-type deadline alert per Dag run creation
- taskinstance.py can use `evaluate_with()` rather than duplicating it
- A context object leaves room to expose more without another signature change

### Consequence for required_kwargs

`required_kwargs` exists solely to declare which of these loose kwargs a reference wants forwarded, and the available pool is exactly `{dag_id, run_id}`.  If the identifiers go away, `required_kwargs` has nothing left to select and becomes vestigial.

It is also the most confusing part of the custom-reference API.  It reads like a way to pass your own configuration in, but declaring anything outside that pool raises ValueError on every evaluation.  The docs shipped with that exact mistake (fix is in #70709).

### Compatibility

This is a breaking change to a public extension point: any custom reference using `kwargs["dag_id"]` would break.  It needs a deprecation path rather than a straight removal, probably passing both forms for one release, deprecating `required_kwargs`, then removing it.

### Related

- #70706, #70708, #70709 (papercuts found in the same area)
- #58248 (introduced the TODO)

### Committer

imrichardwu:
Preserve the Dag run evaluation context for custom references that accept additional keyword arguments, so they continue to evaluate correctly.

 

Pass the in-memory Dag run to Deadline references so built-in references avoid redundant database reads and custom references can evaluate the current state.

Deprecate `required_kwargs` while preserving it for existing custom references. Use the shared evaluator when recalculating queued-at deadlines.

closes: #70710

---

---


imrichardwu:
@ferruzzi can I get a review?

airflow-core/docs/howto/deadline-alerts.rst
```diff
@@ -441,24 +441,17 @@ choose a different time.
     class MyCustomDecoratedReference(BaseDeadlineReference):
         """A custom reference evaluated when Dag runs are created."""
 
-        def _evaluate_with(self, *, session: Session, **kwargs) -> datetime:
-            # Add your business logic here
-            return your_datetime
+        def _evaluate_with(self, *, session: Session, dagrun) -> datetime:
+            return dagrun.logical_date
```
ferruzzi: For these two examples and the two down in the `deadline_reference` docstring, can we add back the `# Add your business logic here` comments and not return the dagrun.logical_date directly?  We should be showing that dagrun.logical_date is available to use in their business logic, not that it is or will be returned.  

Maybe something like this gets the point across?

```suggestion
        def _evaluate_with(self, *, session: Session, dagrun) -> datetime:
            my_datetime = my_business_logic(dagrun.logical_date)
            return my_datetime
```

airflow-core/src/airflow/models/deadline.py
```diff
@@ -320,34 +343,18 @@ def get_reference_class(cls, reference_name: str) -> type[BaseDeadlineReference]
...
-
-            base_time = self._evaluate_with(session=session, **filtered_kwargs)
+            """Evaluate this deadline with the supplied context."""
+            evaluation_kwargs = _get_evaluation_kwargs(self, self._evaluate_with, kwargs)
+            base_time = self._evaluate_with(session=session, **evaluation_kwargs)
             return base_time + interval if base_time is not None else None
 
         @abstractmethod
-        def _evaluate_with(self, *, session: Session, **kwargs: Any) -> datetime | None:
+        def _evaluate_with(self, *, session: Session, dagrun: Any) -> datetime | None:
```
ferruzzi: Rather than using Any, I think a cleaner solution might be to type `dagrun` as a `DagRunProtocol` and add `queued_at` to that protocol.

airflow-core/src/airflow/models/deadline.py
```diff
@@ -57,6 +59,27 @@
...
+        )
+        kwargs = {key: value for key, value in kwargs.items() if key in required_kwargs}
+        if missing_kwargs := required_kwargs - kwargs.keys():
+            raise ValueError(
+                f"{reference.__class__.__name__} is missing required parameters: {', '.join(missing_kwargs)}"
+            )
+        return kwargs
+
+    parameters = signature(evaluator).parameters
+    return {key: value for key, value in kwargs.items() if key in parameters}
```
ferruzzi, line 84: With the way you wrote it, if the user's custom `_evaluate_with` accepts `**kwargs`, they would all get dropped when they very likely intend kwargs to mean "give me anything", which is a standard Python understanding.  This way if `_evaluate_with` claims to accept anything, then give it everything:

```suggestion
    parameters = signature(evaluator).parameters
    if any(param.kind is param.VAR_KEYWORD for param in parameters.values()):
        return kwargs

    return {key: value for key, value in kwargs.items() if key in parameters}
```

It might mean changing some of your tests as well; `test_legacy_serialized_custom_reference_ignores_evaluation_context` for sure will need an adjustment in the asserts.

airflow-core/src/airflow/models/deadline.py
```diff
@@ -381,7 +388,7 @@ class FixedDatetimeDeadline(BaseDeadlineReference):
 
         _datetime: datetime
 
-        def _evaluate_with(self, *, session: Session, **kwargs: Any) -> datetime | None:
+        def _evaluate_with(self, *, session: Session, dagrun: Any) -> datetime | None:
```
ferruzzi: Rather than passing `dagrun` and not using it, I think we can widen the contract here by replacing it with `**kwargs` in any case where dagrun isn't actually used (this is the only one out of the built-in references):


```suggestion
        def _evaluate_with(self, *, session: Session, **kwargs) -> datetime | None:
```

task-sdk/src/airflow/sdk/definitions/deadline.py
```diff
@@ -362,9 +360,11 @@ def serialize_reference(self) -> dict:
         # Optionally, specify when it is calculated by providing a DeadlineReference.TYPES value.
         @deadline_reference(DeadlineReference.TYPES.DAGRUN_QUEUED)
         class MyQueuedRef(BaseDeadlineReference):
-            def _evaluate_with(self, *, session: Session, **kwargs) -> datetime:
-                # Put your business logic here
-                return some_datetime
+            # Optionally, you can specify when you want it calculated by providing a DeadlineReference.TYPES
```
ferruzzi: This comment duplicates the one just above the decorator.  The one outside feels like a better place.

airflow-core/src/airflow/serialization/definitions/deadline.py
```diff
@@ -153,7 +138,7 @@ class FixedDatetimeDeadline(SerializedBaseDeadlineReference):
 
         _datetime: datetime
 
-        def _evaluate_with(self, *, session: Session, **kwargs: Any) -> datetime | None:
+        def _evaluate_with(self, *, session: Session, dagrun: Any) -> datetime | None:
```
ferruzzi: Same as before; dagrun is never used, so replacing it with `**kwargs` is a bit cleaner.

airflow-core/docs/howto/deadline-alerts.rst
```diff
@@ -530,8 +525,5 @@ followed by a more urgent escalation if the Dag is still running.
   and give every field a default value.
 * **Plugin Placement**: One convenient place for custom references is in the plugins directory.
 * **API Server Restart**: Restart the Airflow API Server after adding or modifying custom references.
-* **Required Parameters**: ``required_kwargs`` declares which Dag run context values Airflow should
```
ferruzzi, line 567: Since Required Parameters is still available until 4.0, we should keep it here and just add a note that it is deprecated

airflow-core/src/airflow/models/deadline.py
ferruzzi: It doesn't look like this was actually addressed, was it?   You would at east need to add `queued_at` to `DagRunProtocol` for that to work and it's not showing in the diff.

imrichardwu: something like this `def _evaluate_with(self, *, session: Session, dagrun: DagRunProtocol)`?

ferruzzi: Your `DeadlineDagRunProtocol` was the better answer.

airflow-core/src/airflow/models/deadline.py
```diff
@@ -397,23 +415,14 @@ def deserialize_reference(cls, reference_data: dict):
     class DagRunLogicalDateDeadline(BaseDeadlineReference):
         """A deadline that returns a DagRun's logical date."""
 
-        required_kwargs = {"dag_id", "run_id"}
-
-        def _evaluate_with(self, *, session: Session, **kwargs: Any) -> datetime | None:
-            from airflow.models import DagRun
-
-            return _fetch_from_db(DagRun.logical_date, session=session, **kwargs)
+        def _evaluate_with(self, *, session: Session, dagrun: DeadlineDagRunProtocol) -> datetime | None:
```
Prab-27, line 421: Do we need to remove this unnecessary `session` params ?

imrichardwu: I personally think we should, since it’s dead code

imrichardwu: @ferruzzi thoughts? Originally, I thought about getting rid of it, but since the session uses the abstract method and references like AverageRuntimeDeadline that query the database, I think it makes sense to keep it for consistency.

ferruzzi: I'd keep it for your exact reasoning.  This method gets inherited and extended by user code plus and some internal places are hardcoding the `session=session` so keeping it in the signature feels like the right call and removing it feels like it would break user code.

airflow-core/src/airflow/models/taskinstance.py
```diff
@@ -249,9 +247,16 @@ def _recalculate_dagrun_queued_at_deadlines(
         return
 
     for deadline, deadline_alert in results:
-        # We can't use evaluate_with() since the new queued_at is not written to the DB yet.
-        deadline_interval = timedelta(seconds=deadline_alert.interval)
-        new_deadline_time = new_queued_at + deadline_interval
+        new_deadline_time = decode_deadline_reference(deadline_alert.reference).evaluate_with(
+            session=session,
+            interval=timedelta(seconds=deadline_alert.interval),
```
ferruzzi: deadline_alert.interval is a json value, you need to be deserializing the deadline.  Which also means your tests are hiding something if this is passing.  You also have to account for VariableIntervals now.

You'll need to do something like (untested code)

```python
  for deadline, deadline_alert in results:
      decoded_alert = decode_deadline_alert(
          {
              DeadlineAlertFields.REFERENCE: deadline_alert.reference,
              DeadlineAlertFields.INTERVAL: deadline_alert.interval,
              DeadlineAlertFields.CALLBACK: deadline_alert.callback_def,
          }
      )

      interval = decoded_alert.interval
      if isinstance(interval, VariableInterval):
          interval = interval.resolve()
  
      new_deadline_time = decoded_alert.reference.evaluate_with(
          session=session,
          interval=interval,
          dagrun=dagrun,
          dag_id=dagrun.dag_id,
          run_id=dagrun.run_id,
      )
```

airflow-core/tests/unit/models/test_taskinstance.py
```diff
@@ -4108,8 +4108,28 @@ async def empty_callback_for_deadline():
...
+                    "dag_id": dag_id,
+                    "run_id": run_id,
+                }
+            )
+            return dagrun.queued_at + interval
+
+    monkeypatch.setattr(
+        "airflow.models.taskinstance.decode_deadline_reference",
+        lambda reference: QueuedDeadlineReference(),
+    )
```
ferruzzi: Yeah, I think this patch is hiding something on you.  You are just passing whatever through it rather than checking if it's the right type.  I think these new changes can (and should?) be a new test anyway, you are testing that context is passed and that clearing a TI recalculates the deadline in the same test.

airflow-core/tests/unit/models/test_taskinstance.py
ferruzzi: This line is also catching you up now.  If this was a VariableInterval, this would blow up.  You should replace with with `interval=serialize(interval),` for a more accurate test.

ferruzzi:
Heads up on sequencing: this PR now overlaps two others in `_recalculate_dagrun_queued_at_deadlines`.

- #70370 (opened 24 July) is the targeted fix for the JSON-interval `TypeError` on that function.  It is milestoned **3.3.2** and carries `backport-to-v3-3-test`.

- #71850 (opened 19 August) persists the evaluation timing so custom queued-anchored references get re-anchored on clear.  It also conflicts with this PR across sevreral files.

This PR fixes the same `TypeError` incidentally, and I want to be explicit that **that does not make #70370 redundant**.  This is a refactor with a `significant.rst` newsfragment and a deprecation, so it is not backport material; #70370 is the only route the fix has to 3.3.x, where it  currently affects every queued-at deadline.  Please do not let this PR be used as a reason to close that one.

The cheapest merge order looks like it will be to merge #70370 first, then rebase here.  #70370 adds `decode_deadline_alert_model()` and `resolve_deadline_alert_interval()` to `serialization/decoders.py`, which is exactly what this PR hand-rolls at `taskinstance.py:255-260`.  Rebasing onto it lets you delete those lines.

ferruzzi:
Adding a fourth PR to the picture since it breaks something here concretely.

#71802 adds a core-side `SerializedVariableInterval`, makes `decode_deadline_alert` return that instead of `VariableInterval`, and deletes `resolve()` from the SDK entirely.

That will break `test_clear_task_instances_recalculates_variable_interval_deadline`, which patches `airflow.sdk.definitions.deadline.Variable.get`.  Worth knowing now rather than after a rebase.

Also worth noting, #71850's diff still has `required_kwargs: set[str] = set()` on 
`SerializedBaseDeadlineReference`, which this PR removes, and it adds an `evaluation_timing` class attribute immediately below it.  So be careful when rebasing, whichever lands first.

airflow-core/src/airflow/serialization/definitions/deadline.py
```diff
@@ -169,23 +155,14 @@ def deserialize_reference(cls, reference_data: dict):
...
         """A deadline that returns a DagRun's logical date."""
 
-        required_kwargs = {"dag_id", "run_id"}
-
-        def _evaluate_with(self, *, session: Session, **kwargs: Any) -> datetime | None:
-            from airflow.models import DagRun
-
-            return _fetch_from_db(DagRun.logical_date, session=session, **kwargs)
+        def _evaluate_with(self, *, session: Session, dagrun: DeadlineDagRunProtocol) -> datetime | None:
+            return dagrun.logical_date
```
ferruzzi, line 159: There are a few ways that `dagrun.logical_date` can be `None`, causing `dag.py:769` to silently skip the insert.  `_fetch_from_db` used to log in that case, and removing that method call was necessary, but now we don't have anything logging it.  It's worth adding a log here to replace that one, something like 

```
if dagrun.logical_date is None:
    self.log.warning("No deadline created for %s: the Dag run has no logical date.", dagrun.dag_id)
return dagrun.logical_date
```

The same applies below on `return dagrun.queued_at`.

airflow-core/tests/unit/models/test_deadline.py
```diff
@@ -335,138 +326,68 @@ def test_handle_miss_persists_executor_callback_routing_data(self, dagrun, sessi
...
-                    result = reference.evaluate_with(session=session, interval=interval, dag_id=DAG_ID)
-                    mock_fetch.assert_not_called()
-                    # Should return None when no DAG runs exist
-                    assert result is None
-            else:
-                result = reference.evaluate_with(session=session, interval=interval)
-                mock_fetch.assert_not_called()
-                assert result == DEFAULT_DATE + interval
+        assert reference.evaluate_with(session=session, interval=timedelta(), dagrun=dagrun) is None
+        assert caplog.messages == [message]
```
ferruzzi: Sorry, I know it's frustrating to get follow-up comments like this instead of getting them all at once.  I missed this earlier; we're not supposed to use caplog.messages anymore.  this should be `{"event": ...} in caplog` since we're using structlog now.

ferruzzi (commented):
Final nit, sorry.

airflow-core/src/airflow/serialization/definitions/deadline.py
ferruzzi: @imrichardwu 

Sorry for the churn; I know you did exactly what I asked for but #71767 ended up covering those log messages at the call site instead, which is arguably a better location and can include more useful information.  That one also needs to get merged before we can merge this one anyway.  
 
Could you take both of these back to a bare return, matching the `ReferenceModels` twins in `models/deadline.py`, and we'll let the other PR's log message handle them?

Sorry again, I'm juggling a bunch of overlapping PRs and trying to make the merge conflicts as seamless as I can.

airflow-core/src/airflow/models/taskinstance.py
```diff
@@ -249,9 +248,27 @@ def _recalculate_dagrun_queued_at_deadlines(
...
+        decoded_alert = decode_deadline_alert(
+            {
+                DeadlineAlertFields.REFERENCE: deadline_alert.reference,
+                DeadlineAlertFields.INTERVAL: deadline_alert.interval,
+                DeadlineAlertFields.CALLBACK: deadline_alert.callback_def,
+            }
+        )
+        interval = decoded_alert.interval
+        if not isinstance(interval, timedelta):
+            interval = interval.resolve()
```
seanghaeli: Looks good and I think it's mostly ready to merge, there's just one more thing that needs to be taken care of since new code has merged into main. #71968 added a session argument to `SerializedVariableInterval.resolve()`, so your `interval.resolve()` must become `interval.resolve(session=session)`.

Your call site runs inside `clear_task_instances`, which is midway through a batch of database changes it saves all at once at the end. Without the session, the Variable lookup opens the same connection and saves that half-finished batch early. #71968 already converted the other call site to pass it; this new one should follow.

seanghaeli:
Oh and also a rebase is needed

seanghaeli:
Looks good pending CI, @ferruzzi

PR #64929, opened 2026-04-08
Reduce per-DAG queries during DAG serialization with bulk prefetch

ephraimbuddy:
Replaces 3 SELECTs per DAG in write_dag (update interval check, hash comparison, version fetch) with 2 bulk queries via a new _prefetch_dag_write_metadata classmethod. Also fixes DagCode.update_source_code to reuse the caller's session and eagerly loads dag_owner_links to prevent N+1 queries.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes (please specify the tool below)
Claude Opus 4.6

Issue #56471, opened 2025-10-08
Non-deterministic DAG serialization for Dinamically generated dags leads to excessive versions


wolvery:
### Apache Airflow version

3.1.0

### What happened?

We are experiencing an issue where the dag processor creates a new version of a DAG in the serialized_dag table on nearly every parsing cycle, even when the underlying DAG file is functionally unchanged. 

The root cause appears to be a non-deterministic serialization process. The order of dictionary keys and list elements in the resulting JSON column of serialized_dag table is inconsistent between parses. This leads to different versions for logically identical DAGs, only the order of keys inside of the JSON are changing.

Example:
```
--- version_1
+++ version_2
@@ -1,13 +1,13 @@
-"dag_id": "test",
-"max_consecutive_failed_dag_runs": 7,
-"timetable": { ... },
-"relative_fileloc": "revision_dags/test.py",
-"task_group": { ... },
-"fileloc": "/opt/airflow/dags/revision_dags/test.py",
-"timezone": "UTC",
-"default_args": { ... },
-"description": "DAG for [domain='test', data_product='polaroid_input_features', pipeline='main']",
-"max_active_runs": 1,
-"tags": [ ... ],
-"start_date": 1640995200.0
+"max_consecutive_failed_dag_runs": 7,
+"task_group": { ... },
+"timezone": "UTC",
+"max_active_runs": 1,
+"fileloc": "/opt/airflow/dags/revision_dags/test.py",
+"timetable": { ... },
+"start_date": 1640995200.0,
+"description": "DAG for [domain='test', data_product='polaroid_input_features', pipeline='main']",
+"default_args": { ... },
+"tags": [ ... ],
+"relative_fileloc": "revision_dags/test.py",
+"dag_id": "test"
```

### What you think should happen instead?

It should sort the keys internally to perform the comparison and avoid the creation of a new version.

### How to reproduce

Creating a simple dynamic dag and importing in the global seems to lead to the problem.

### Operating System

airflow:3.1.0 python 3.10 image

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

Official Apache Airflow Helm Chart

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

anshuksi282-ksolves:
[...]

PR #70715, opened 2026-07-30
Eliminate duplicate json.dumps in SerializedDagModel

ahujaanmol1289:
## What this does

`SerializedDagModel.__init__()` and `hash()` each independently called
`_sort_serialized_dag_dict()` followed by `json.dumps(sort_keys=True)`.
The two JSON outputs differ only by whether `fileloc` and `bundle_name`
are present — every DAG write paid for two full recursive sorts and two
full JSON serializations when one of each suffices.

This PR adds `_compute_hash_and_storage_json()` that:
- Calls `_sort_serialized_dag_dict()` **once**
- Pops `fileloc`/`bundle_name`, generates hash JSON, computes MD5
- Restores the popped fields, generates storage JSON
- Returns `(dag_hash, storage_json_bytes, sorted_data)` in one pass

`__init__()` and `hash()` both delegate to it instead of doing their
own independent sort + serialize.

## Why

The duplicate work was unnecessary — the hash computation and storage
JSON generation operate on the same sorted dict and differ only by
two fields. Merging them removes one full recursive deep-sort and one
full `json.dumps()` call per DAG per parse cycle.

## Benchmark (50-task DAG, 100 iterations)

| Metric | Before | After |
|--------|--------|-------|
| Mean   | 1.83 ms | 1.61 ms |
| Median | 1.79 ms | 1.56 ms |

~12% improvement on `_compute_hash_and_storage_json` in isolation.

This runs on every DAG on every parse cycle, so the improvement
compounds linearly with DAG count.

## What does NOT change

- The hash value produced is **bit-identical** to before (verified across
  PYTHONHASHSEED=42, 123, 999)
- The stored JSON is **identical**
- Compression behavior (`compress_serialized_dags`) is unchanged
- `write_dag()` logic and control flow is untouched
- `hash()` remains a public classmethod with the same signature and
  return value (backward compatible)
- All 86 existing `test_serialized_dag` tests pass unmodified

## Testing

8 new tests in `TestComputeHashAndStorageJson`:
- Hash equivalence with the legacy implementation
- `fileloc` and `bundle_name` excluded from hash, present in storage JSON
- Hash changes when actual DAG content changes
- Input dict is not mutated by the computation
- `__data_cache` and `_data`/`_data_compressed` behavior preserved
  (parametrized: compressed + uncompressed)

Validated across: sqlite, postgres, mysql. Hash-stable across 3
PYTHONHASHSEED values. mypy clean. Full Core suite (3321 tests) passes.

Related: #56471, #64929

---

- []  Yes (please specify the tool below)

### DIFF
--- a/airflow-core/src/airflow/models/serialized_dag.py
+++ b/airflow-core/src/airflow/models/serialized_dag.py
@@ -347,10 +347,8 @@ class SerializedDagModel(Base):
     def __init__(self, dag: LazyDeserializedDAG) -> None:
         self.dag_id = dag.dag_id
         dag_data = dag.data
-        self.dag_hash = SerializedDagModel.hash(dag_data)
-
-        # partially ordered json data
-        dag_data_json = json.dumps(dag_data, sort_keys=True).encode("utf-8")
+        dag_hash, dag_data_json, _ = SerializedDagModel._compute_hash_and_storage_json(dag_data)
+        self.dag_hash = dag_hash
 
         if _COMPRESS_SERIALIZED_DAGS:
             self._data = None
@@ -367,18 +365,36 @@ class SerializedDagModel(Base):
         return f"<SerializedDag: {self.dag_id}>"
 
     @classmethod
+    def _compute_hash_and_storage_json(cls, dag_data: dict) -> tuple[str, bytes, dict]:
+        """
+        Compute the Dag hash and storage JSON in a single pass.
+
+        Sorts the serialized dict once, generates hash JSON (without fileloc/bundle_name),
+        then restores those fields and generates storage JSON.
+
+        :return: (dag_hash, storage_json_bytes, sorted_data)
+        """
+        sorted_data = cls._sort_serialized_dag_dict(dag_data)
+        dag_section = sorted_data["dag"]
+        saved_fileloc = dag_section.pop("fileloc", None)
+        saved_bundle_name = dag_section.pop("bundle_name", None)
+
+        hash_json = json.dumps(sorted_data, sort_keys=True).encode("utf-8")
+        dag_hash = md5(hash_json).hexdigest()
+
+        if saved_fileloc is not None:
+            dag_section["fileloc"] = saved_fileloc
+        if saved_bundle_name is not None:
+            dag_section["bundle_name"] = saved_bundle_name
+
+        storage_json = json.dumps(sorted_data, sort_keys=True).encode("utf-8")
+        return dag_hash, storage_json, sorted_data
+
+    @classmethod
     def hash(cls, dag_data):
         """Hash the data to get the dag_hash."""
-        dag_data = cls._sort_serialized_dag_dict(dag_data)
-        data_ = dag_data.copy()
-        # Remove fileloc from the hash so changes to fileloc
-        # does not affect the hash. In 3.0+, a combination of
-        # bundle_path and relative fileloc more correctly determines the
-        # dag file location.
-        data_["dag"].pop("fileloc", None)
-        data_["dag"].pop("bundle_name", None)
-        data_json = json.dumps(data_, sort_keys=True).encode("utf-8")
-        return md5(data_json).hexdigest()
+        dag_hash, _, _ = cls._compute_hash_and_storage_json(dag_data)
+        return dag_hash
 
     @classmethod
     def _sort_serialized_dag_dict(cls, serialized_dag: Any):
--- a/airflow-core/tests/unit/models/test_serialized_dag.py
+++ b/airflow-core/tests/unit/models/test_serialized_dag.py
@@ -1155,3 +1155,92 @@ class TestSerializedDagModel:
         alert = session.scalar(select(DAM).where(DAM.serialized_dag_id == orig_serdag.id))
         assert alert is not None
         assert alert.id == orig_alert.id
+
+
+class TestComputeHashAndStorageJson:
+    """Tests for SerializedDagModel._compute_hash_and_storage_json."""
+
+    @staticmethod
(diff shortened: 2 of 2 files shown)


airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -366,19 +364,37 @@ def __init__(self, dag: LazyDeserializedDAG) -> None:
...
-        data_ = dag_data.copy()
-        # Remove fileloc from the hash so changes to fileloc
-        # does not affect the hash. In 3.0+, a combination of
-        # bundle_path and relative fileloc more correctly determines the
-        # dag file location.
-        data_["dag"].pop("fileloc", None)
-        data_["dag"].pop("bundle_name", None)
-        data_json = json.dumps(data_, sort_keys=True).encode("utf-8")
-        return md5(data_json).hexdigest()
+        dag_hash, _, _ = cls._compute_hash_and_storage_json(dag_data)
```
kaxil, line 396: This makes the common path slower. `write_dag` calls `hash()` on every serialization pass to decide whether anything changed, and unchanged DAGs (the usual case) now pay for building the storage JSON that gets thrown away here. On a 50-task dict I measure standalone `hash()` about 26% slower. The constructor doesn't get faster either: on main, `__init__` only sorts inside `hash()`, the storage dump is a plain `json.dumps(dag_data, sort_keys=True)` on the unsorted dict, so it was already one sort plus two dumps, exactly what this helper does. The duplication that does exist is `write_dag` hashing and then `cls(dag)` hashing again when the DAG changed; passing the precomputed hash into the constructor would remove that without touching the unchanged-DAG path.

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -366,19 +364,37 @@ def __init__(self, dag: LazyDeserializedDAG) -> None:
...
+
+        hash_json = json.dumps(sorted_data, sort_keys=True).encode("utf-8")
+        dag_hash = md5(hash_json).hexdigest()
+
+        if saved_fileloc is not None:
+            dag_section["fileloc"] = saved_fileloc
+        if saved_bundle_name is not None:
+            dag_section["bundle_name"] = saved_bundle_name
+
+        storage_json = json.dumps(sorted_data, sort_keys=True).encode("utf-8")
```
kaxil, line 390: Dumping `sorted_data` here changes what gets stored when `compress_serialized_dags` is enabled. On main the stored JSON is dumped from the unsorted dict, so list order is preserved; `_sort_serialized_dag_dict` reorders tasks by task_id and alphabetizes every all-string list, e.g. `op_args: ["zeta", "alpha"]` round-trips from compressed storage as `["alpha", "zeta"]`, and `template_fields`/`tags` get the same treatment. I verified the bytes differ on such a dict. It also means `_data_compressed` no longer decompresses to the same lists as `__data_cache` or the uncompressed `_data`, which keep the original order, so the "stored JSON is identical" claim only holds for dicts whose lists are already sorted. (A `fileloc: None` entry would also be dropped from storage entirely, since the restore checks `is not None`.)

airflow-core/tests/unit/models/test_serialized_dag.py
```diff
@@ -1155,3 +1155,92 @@ def test_deadline_reuse_skips_write_when_hash_matches(self, testing_dag_bundle,
...
+            pytest.param(False, id="uncompressed"),
+            pytest.param(True, id="compressed"),
+        ],
+    )
+    def test_init_data_cache_and_storage(self, compress, monkeypatch):
+        import zlib
+
+        monkeypatch.setattr("airflow.models.serialized_dag._COMPRESS_SERIALIZED_DAGS", compress)
+        dag_data = self._make_dag_data()
+        lazy_dag = mock.MagicMock()
```
kaxil, line 1231: Repo convention is to always spec mocks: `mock.MagicMock(spec=LazyDeserializedDAG)`, otherwise attribute typos pass silently. The `import copy` / `import zlib` inside the tests should move to the top of the file too, and `SDM(lazy_dag)` does the same as the `__new__` + `__init__` pair below.

ahujaanmol1289:
Thanks for the thorough review — all three points are correct, and I appreciate
the detailed analysis.

**On the `hash()` regression**: You're right. I misread the original code — `__init__`
dumps the unsorted dict via `json.dumps(dag_data, sort_keys=True)`, so the sort
only happens once (inside `hash()`), not twice. Making `hash()` delegate to
`_compute_hash_and_storage_json()` forces the unchanged-DAG path to generate and
discard storage JSON it never needed. That's a net regression on the hot path.

**On the storage bytes change:** Also correct. `_sort_serialized_dag_dict()` reorders
lists (tasks by task_id, all-string lists alphabetically), which `json.dumps(sort_keys=True)`
alone does not. Dumping from the sorted dict changes the stored bytes for any DAG
with unsorted lists. The `fileloc: None` edge case is a real bug too — the
`is not None` guard would silently drop it from storage. I should have caught both
of these.

**On the test style:** Will fix — `spec=LazyDeserializedDAG` on mocks, imports to
file top, and simplifying the `__new__` + `__init__` pattern.

Your suggestion to pass the precomputed hash into the constructor instead is the
right fix — it removes the actual duplication (write_dag hashing and then cls(dag)
hashing again when the DAG changed) without touching the unchanged-DAG path or
altering storage bytes.

I'll rework the PR to take that approach. Should I push a revised commit to this
branch, or would you prefer I close this and open a fresh PR?

ahujaanmol1289:
Reworked this based on your feedback — new PR at https://github.com/apache/airflow/pull/71222/changes. Only passes the precomputed hash into the constructor, no changes to hash() or storage bytes. Also handles the deadline UUID mutation case by recomputing the hash after the mutation before passing it in.

PR #70717, opened 2026-07-30
Add Language SDK conformance spec to contributor guide

jason810496:
### Why

We should define the spec of "What" is a valid "Lang-SDK". Currently, the contributor guide only mention "how" to add a new Lang-SDK. Since all the SDK is developed independently and they might mature in different rate.

I will add "compatibility matrix" (a table that what does this SDK support) for each Lang-SDK as follow-up.

---

- [x] Yes (please specify the tool below) With help of Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/contributing-docs/30_new_language_sdk.rst
+++ b/contributing-docs/30_new_language_sdk.rst
@@ -21,7 +21,10 @@ Creating a new Language SDK
 Starting from 3.3, the standard Airflow workers can run task code implemented in
 languages other than Python, using a foreign Language SDK. This document
 describes how a new Language SDK can be contributed, so Airflow can execute
-tasks implemented in the language.
+tasks implemented in the language. It is also the normative *conformance
+specification* a Language SDK is expected to meet: the `Conformance`_ section
+defines the exact TaskInstance states and capabilities an SDK MUST, SHOULD, or
+MAY support.
 
 Two components are needed for Airflow to understand how to execute such a task:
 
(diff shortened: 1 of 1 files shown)


potiuk (approved):

contributing-docs/30_new_language_sdk.rst
```diff
@@ -333,6 +336,162 @@ Error handling
...
+
+``mixed-lang-stub-target`` (MUST)
+    The SDK can execute a task declared in a Python Dag with the ``@task.stub``
+    decorator. This is the primary execution path for every Language SDK.
+
+``task-logging`` (MUST)
+    The SDK forwards the task's log output — stdout / stderr and any structured log
+    records — to the supervisor over the ``--logs`` socket, so it surfaces in the task
+    log exactly like a Python task. Remote log *storage* (S3, GCS, …) is handled
+    uniformly by the supervisor of this and is not part of SDK conformance.
```
uranusjr:
```suggestion
    log exactly like a Python task. A remote log *storage* (S3, GCS, …) is handled
    uniformly by the supervisor and thus not part of SDK conformance.
```

contributing-docs/30_new_language_sdk.rst
```diff
@@ -333,6 +336,162 @@ Error handling
...
+
+``asset-inlets-outlets`` (MAY †)
+    Native tasks can declare asset inlets and outlets.
+
+``asset-scheduling`` (MAY †)
+    A native Dag can be scheduled on asset events.
+
+``object-store`` (MAY)
+    The SDK exposes an object-storage API (an ``ObjectStoragePath`` equivalent) usable from
+    Dag code.
```
uranusjr: Is this one also conditional but missed the dagger, or correctly marked but misplaced?

contributing-docs/30_new_language_sdk.rst
```diff
@@ -333,6 +336,162 @@ Error handling
...
+Conformance
+-----------
+
+The sections above describe *how* to build an SDK. This section defines *what* an
+SDK must do to be considered conformant: which TaskInstance states its runtime is
+expected to report and which optional capabilities it may offer. Because SDKs mature
+at different rates, they are not required to be identical; each one declares the
+subset it actually supports.
+
+The key words MUST, SHOULD, and MAY are used as described in :rfc:`2119`.
```
uranusjr, line 348: Does this work? I don’t think contributing docs are built with Sphinx? This might need to be a hyperlink instead.

jason810496: It's a feature from GitHub side that render it as https://datatracker.ietf.org/doc/html/rfc2119.html. Here's the screenshot of hovering the link.

Screenshot 2026-07-31 at 7 37 33 PM

contributing-docs/30_new_language_sdk.rst
jason810496: I didn't catch this, it should be tagged with dagger. I addressed in https://github.com/apache/airflow/pull/70717/commits/37a0207c2ac0b75576b76a06e3d8136d4425884c, thanks.

PR #70723, opened 2026-07-30
Check GCSToS3Operator match_glob support after template rendering

atikulmunna:
Rewrites the `match_glob` support check in `GCSToS3Operator.__init__` so it reads whether the argument was passed, and narrows it to `ValueError`.

`match_glob` is a template field, so the constructor only ever sees the un-rendered Jinja expression. The old check, `if not self.__is_match_glob_supported and match_glob:`, was a truthiness test on that string, which asks a third question matching neither intent.

It stays in `__init__`. It is a provision check, not a value check: it asks whether `match_glob` was supplied, combined with an environment capability the constructor can already answer. Per the false positives section of #70296 those belong in the constructor and get rewritten in place rather than moved, because with `render_template_as_native_obj=True` a provided field can render to `None`, and raising at construction surfaces a static authoring mistake as a Dag import error instead of once per task instance and per retry.

Since #70505 narrowed `validate-operators-init` to sanction `is None` / `is not None` reads, the rewrite passes the hook and the exemption entry still goes.

This change:

* rewrites the check to `match_glob is not None`, which is strictly stricter than the truthiness test since it also rejects `match_glob=""`
* narrows `AirflowException` to `ValueError`, matching #70359, which also drops `AirflowException` from the `common.compat.sdk` import as nothing else in the file used it
* removes the file's now-stale `gcs_to_s3.py::1` entry from `generated/known_airflow_exceptions.txt`, since narrowing removed its only `raise AirflowException`
* adds `test_match_glob_requires_recent_google_provider`, parametrised over a glob and the empty string so the `is not None` polarity is pinned, where nothing covered this error before
* removes `GCSToS3Operator` from `scripts/ci/prek/validate_operators_init_exemptions.txt`

An earlier revision of this PR moved the check into `execute()` instead, mirroring #70574. That was wrong for the reasons above, and @potiuk's review corrected it; the current diff is the in-place rewrite.

Verified locally:

```bash
python scripts/ci/prek/validate_operators_init.py \
  providers/amazon/src/airflow/providers/amazon/aws/transfers/gcs_to_s3.py
python scripts/ci/prek/check_new_airflow_exception_usage.py \
  providers/amazon/src/airflow/providers/amazon/aws/transfers/gcs_to_s3.py
ruff check && ruff format --check   # on the changed files
pytest providers/amazon/tests/unit/amazon/aws/transfers/test_gcs_to_s3.py
```

The test suite runs in a Linux container rather than natively; `_shared/observability/metrics/stats.py` calls `os.register_at_fork` at import, which is POSIX only.

related: #70296

---

---


providers/amazon/src/airflow/providers/amazon/aws/transfers/gcs_to_s3.py
```diff
@@ -184,6 +180,10 @@ def execute(self, context: Context) -> list[str]:
         }
         if self.__is_match_glob_supported:
             list_kwargs["match_glob"] = self.match_glob
+        elif self.match_glob:
+            raise AirflowException(
```
potiuk: Worth narrowing this to `ValueError` while you're moving it.

Airflow is actively reducing direct `AirflowException` raises (there's a `check-no-new-airflow-exceptions` prek hook enforcing it). Relocating an existing one verbatim during a refactor is explicitly allowed, so this isn't a violation — but this is a plain "caller passed an argument this environment can't support", which is exactly what `ValueError` is for.

The clincher is that **#70574 makes the identical change for `GCSToAzureBlobStorageOperator` and uses `ValueError`**. Same bug, same fix, two providers, two different exception types would be an odd thing to land in the same week. Worth syncing with that PR on which one wins — I'd suggest `ValueError` for both.

---

potiuk (approved):
Thanks — correct fix. `match_glob` is in `template_fields`, so validating it in `__init__` meant a templated value was checked before rendering, and a failure there breaks Dag *parsing* rather than failing the task. Moving it into `execute` alongside the `_is_match_glob_supported` branch is the right place, and the version probe itself correctly stays in `__init__` since it depends on the installed provider, not on the rendered value. Dropping the entry from `validate_operators_init_exemptions.txt` completes it.

Test is good — asserting `list.assert_not_called()` proves it fails before doing any work.

One consistency point inline.

---

potiuk (changes requested):

`if not self.__is_match_glob_supported and self.match_glob:` asks whether `match_glob` was passed (combined with an environment capability that `__init__` can already answer). It's a provision check, so it should be rewritten in place rather than relocated:

```python
# in __init__ — keep it here
if not self.__is_match_glob_supported and match_glob is not None:
    raise ValueError(
        "The 'match_glob' parameter requires 'apache-airflow-providers-google>=10.3.0'."
    )
```

Note the `is not None` polarity matters: `if match_glob:` is a truthiness test on the un-rendered Jinja string, which is a third question that matches neither intent.

My earlier point about narrowing `AirflowException` to `ValueError` still stands and is worth doing in the same edit — #70503 makes the same call for `dms.py`.

---

atikulmunna:
Thanks for the correction, and no need to apologise for the churn. You're right. I mirrored #70574 and treated it as settled precedent instead of checking it against the false positives section of #70296, which is exactly the failure mode you describe in your sweep.

Rewritten in place:

* the check stays in `__init__`, now `match_glob is not None` rather than the truthiness test
* narrowed to `ValueError`, which also drops `AirflowException` from the `common.compat.sdk` import since nothing else in the file used it
* the test asserts the raise at construction instead of from `execute()`
* exemption entry still removed, and I confirmed `validate-operators-init` passes with it gone, so #70505 does what you said

One thing that wasn't in your review: narrowing to `ValueError` removes this file's only `raise AirflowException`, which left a stale `gcs_to_s3.py::1` entry in `generated/known_airflow_exceptions.txt`. I dropped that too, same as #70359 did. Running `--cleanup` also wanted to remove an unrelated `mlengine.py` entry, which I put back so the diff stays on topic.

Also worth correcting my own PR description: I said there I couldn't run the tests. I've since got them running in a Linux container and the file is green, 31 passed. The Windows blocker was `os.register_at_fork` in `_shared/observability/metrics/stats.py`, which is POSIX only, so it isn't something a local install can work around.

The version probe stays in `__init__` untouched, since it depends on the installed provider rather than a rendered value.

shahar1 (commented):
The rewrite lands where `#70296` says it should, and it matches the `#70359`
precedent one-for-one. Two small things below, neither blocking — I'm leaving
this as a comment rather than an approval only because the earlier
`CHANGES_REQUESTED` is still the standing review state.

Things I checked rather than assumed:

- `match_glob is not None` is the shape the hook sanctions —
  `_collect_sanctioned_uses()` in `scripts/ci/prek/validate_operators_init.py`
  explicitly whitelists `is None` / `is not None` comparisons on template
  fields, which the old truthiness test was not. Removing the
  `validate_operators_init_exemptions.txt` entry is correct.
- The `generated/known_airflow_exceptions.txt` deletion is what the generator
  itself would produce: `AllowlistManager.check()` does `del allowlist[rel]`
  rather than tightening to `::0` when a file's count reaches zero. Keeping the
  unrelated `mlengine.py` entry out of the diff was the right call — that file
  no longer exists, and cleaning it up belongs in its own PR.
- Shape matches `f781f8b878` (`#70359`, `S3DeleteObjectsOperator`) exactly: same
  four files, same polarity, same `ValueError` narrowing, no changelog entry.
  Per `providers/AGENTS.md` § "Changelog — never use newsfragments", nothing
  further is needed.

### Smaller observations

- **PR description no longer matches the diff.** It still describes the first
  approach — check removed from `__init__`, raised from `execute()`,
  `AirflowException` deliberately kept, test named
  `test_execute__match_glob_requires_recent_google_provider`. All four are now
  false; the second commit reversed them. It also still states the unit tests
  couldn't be run locally, which you corrected in a comment. The review
  criteria call this out directly:

  > **Description doesn't match code**: PR description describes something
  > different from what the code actually does.

  — [`.github/instructions/code-review.instructions.md`](https://github.com/apache/airflow/blob/main/.github/instructions/code-review.instructions.md).
  The commit messages are accurate, so this is only the body; worth a quick
  rewrite so the next reader isn't chasing an `execute()` change that isn't
  there.

- `providers/amazon/tests/unit/amazon/aws/transfers/test_gcs_to_s3.py:79` — the
  new test doesn't lock in the `is not None` polarity; see the inline comment.

---

>

---

providers/amazon/tests/unit/amazon/aws/transfers/test_gcs_to_s3.py
```diff
@@ -75,6 +75,18 @@ def test_execute__match_glob(self, mock_hook):
...
+    def test_match_glob_requires_recent_google_provider(self):
+        with pytest.raises(ValueError, match="match_glob"):
+            GCSToS3Operator(
+                task_id=TASK_ID,
+                gcs_bucket=GCS_BUCKET,
+                prefix=PREFIX,
+                dest_aws_conn_id="aws_default",
+                dest_s3_key=S3_BUCKET,
+                match_glob=f"**/*{DELIMITER}",
+            )
```
shahar1, line 89: The test passes `match_glob=f"**/*{DELIMITER}"`, which is truthy, so it would still pass against the old `if not ... and match_glob:` condition — the only thing failing it on `main` is the `AirflowException` → `ValueError` swap, not the polarity change. Since `is not None` vs. truthiness was the substance of the review round, worth parametrising so the empty-string case is pinned:

```suggestion
    @mock.patch("airflow.providers.google.__version__", "10.2.0")
    @pytest.mark.parametrize("match_glob", [f"**/*{DELIMITER}", ""])
    def test_match_glob_requires_recent_google_provider(self, match_glob):
        with pytest.raises(ValueError, match="match_glob"):
            GCSToS3Operator(
                task_id=TASK_ID,
                gcs_bucket=GCS_BUCKET,
                prefix=PREFIX,
                dest_aws_conn_id="aws_default",
                dest_s3_key=S3_BUCKET,
                match_glob=match_glob,
            )
```

`AGENTS.md` asks for `@pytest.mark.parametrize` when tests differ only in input, so this stays one test rather than two.

atikulmunna:
Both fixed in 3c8d638003 plus a description rewrite.

Test: you're right, it wasn't testing the polarity at all. The truthy glob would have passed the old condition too, so the `ValueError` swap was doing all the work. Parametrised over the glob and the empty string as suggested. `match_glob=""` is the case that separates the two conditions. Both pass, file is at 32.

Description: it was still describing the first revision, including the bit about not being able to run the tests locally. Rewritten to match the diff, with a note that the earlier `execute()` version was corrected in review.

atikulmunna:
@potiuk @shahar1 ping on this one.

Both review points are addressed and the branch is up to date with main after a conflict in the exemptions file. The standing `CHANGES_REQUESTED` is from the first revision, which was rewritten in `454597cc6d`.

CI hasn't run since the merge commit either, the workflow runs are sitting at `action_required` and need an approval to start.

atikulmunna:
Merged main to pick up #71118.

The 8 failures on the previous run were all provider distribution builds failing on `flit build --no-setup-py`. That run was queued on 1 August and only started once CI got approved, so it built with the tooling from before #71118 dropped the flag. Not related to this change.

Diff is unchanged, same 4 files.

PR #70725, opened 2026-07-30
Bump ruff to 0.16.0 and pin the base lint rule set

Lee-W:
## Why

Ruff 0.16 widened its default rule set from ~250 to ~510 enabled rules. Airflow's config uses `extend-select`, which layers on top of whatever Ruff currently enables by default, so bumping the pin would have silently pulled in 261 unreviewed rules and produced sweeping autofix churn across the repo.

## What

- Pin `select` in `pyproject.toml` to the rule groups Airflow has always effectively had enabled (`E4`, `E7`, `E9`, `F`), so future Ruff releases cannot change the lint surface out from under us. Everything intentionally on is still listed in `extend-select`, unchanged.
- Bump the pinned Ruff version to `0.16.0`: `.pre-commit-config.yaml`, `devel-common/pyproject.toml`, and the two prek script headers (`scripts/ci/prek/check_imports_in_providers.py`, `scripts/ci/prek/ruff_format.py`).
- Two rules fall newly in scope under the pin because they sit inside sections Airflow selects wholesale. Both are deferred via `ignore` rather than fixed here — a version bump should not also carry semantic changes, and each needs per-site judgement that deserves its own reviewed change:
  - `ISC004` (implicit string concat in a collection literal) — 109 sites, unsafe autofix.
  - `LOG004` (`.exception()` outside an exception handler) — 13 sites; each call site can only be classified by tracing whether an exception isgenuinely in flight there.
- `task-sdk/src/airflow/sdk/execution_time/schema/AGENTS.md`: correct the bundle-registration snippet to match the actual `get_bundle()` / `VersionBundle` shape instead of a bare `versions/__init__.py` list.

 

---

---


dheerajturaga (approved):
LGTM

PR #70740, opened 2026-07-30
Move AzureFileShareToGCSOperator directory_name alias out of __init__

Andrushika:
Resolve AzureFileShareToGCSOperator directory_name alias after rendering
---

## Why

`directory_name` and `directory_path` are both template fields, rendered after `__init__` runs. The old constructor resolved the deprecated `directory_name` -> `directory_path` alias there, on the un-rendered values, which #70296 flags.

## What

Move the alias copy into `render_template_fields`, not `execute`. The alias needs to know if `directory_path` was passed, and with `render_template_as_native_obj=True` a supplied `directory_path` can render to `None`. So the check runs before rendering, where the raw values are still visible, and the copy runs after. The deprecation warning stays in `__init__`.

related: #70296

---

---


Andrushika:
cc @shahar1

providers/google/src/airflow/providers/google/cloud/transfers/azure_fileshare_to_gcs.py
```diff
@@ -127,6 +128,14 @@ def __init__(
         else:
             self.return_gcs_uris = return_gcs_uris
 
+    def render_template_fields(self, context: Context, jinja_env: jinja2.Environment | None = None) -> None:
```
shahar1: This override won't run for mapped tasks. `MappedOperator.render_template_fields` calls `unmapped_task._do_render_template_fields(...)` directly rather than the operator's own `render_template_fields` — see the comment at [`mappedoperator.py:844-847`](https://github.com/apache/airflow/blob/main/task-sdk/src/airflow/sdk/definitions/mappedoperator.py#L844). I confirmed it with a scratch DAG: a subclass override fires for a plain task and not for a `.partial().expand()` one.

So `AzureFileShareToGCSOperator.partial(directory_name="dir").expand(...)` regresses: `__init__` no longer aliases, the override never fires, and `execute` hits `RuntimeError: The directory_name must be set!.`. That path works on `main` today — `unmap()` goes through `__init__`, which aliased the raw template string into `directory_path`, and `_do_render_template_fields` then rendered it correctly. (Which also means the un-rendered-alias bug this PR fixes only ever affected non-mapped tasks.)

Decide in `__init__` and apply in `execute` — that keeps your pre-render decision, which is the right call, and `execute` runs on both the mapped and unmapped paths:

```python
# in __init__, replacing the render_template_fields override
self._use_directory_name = directory_path is None and directory_name is not None

# at the top of execute()
if self._use_directory_name:
    self.directory_path = self.directory_name
```

`_use_directory_name` isn't a template field, so `validate_operators_init` stays satisfied and the exemption removal still holds. If you'd rather keep it at render time, overriding `_do_render_template_fields` covers both paths too, but it's a private API — I'd lean toward the flag.

Either way, please add a mapped-task test (`.partial(directory_name=...)` → assert `directory_path` after execute), since that's the case that regressed.

providers/google/tests/unit/google/cloud/transfers/test_azure_fileshare_to_gcs.py
```diff
@@ -56,6 +61,96 @@ def test_init(self):
...
+            operator.execute(None)
+
+    @pytest.mark.parametrize(
+        "extra_kwargs",
+        [
+            pytest.param({"directory_path": AZURE_FILESHARE_DIRECTORY_PATH}, id="directory_path-supplied"),
+            pytest.param({}, id="neither-supplied"),
+        ],
+    )
+    def test_no_directory_name_deprecation_without_directory_name(self, extra_kwargs):
```
shahar1: This test passes unchanged on `main` — the `__init__` condition is semantically identical before and after (`self.directory_path`/`self.directory_name` were assigned unmodified from the parameters just above), so it exercises pre-existing behaviour rather than anything this PR changes.

[AGENTS.md Testing Standards](https://github.com/apache/airflow/blob/main/AGENTS.md#testing-standards): *"Target exactly 100% coverage of what the PR changes — no more, no less... every test must fail without the PR's change. Do not add tests for pre-existing logic that was already present before the PR."*

Worth dropping, unless you're deliberately backfilling a gap — in which case a one-line note saying so would help.

shahar1 (changes requested):
Thanks for taking this on — moving the alias out of `__init__` is the right direction, and the pre-render/post-render split is a genuinely subtle point that the second test documents well.

One blocking issue: the `render_template_fields` override never runs for mapped tasks, so this turns a currently-working case into a runtime failure. Details inline.

Smaller observations:

- `test_no_directory_name_deprecation_without_directory_name` passes unchanged on `main` — it exercises pre-existing `__init__` behaviour rather than anything this PR changes. Per [AGENTS.md Testing Standards](https://github.com/apache/airflow/blob/main/AGENTS.md#testing-standards): *"every test must fail without the PR's change... Do not add tests for pre-existing logic that was already present before the PR."* (`test_native_directory_path_rendering_to_none_is_not_aliased` is fine — it's an explicit behaviour-preservation guard, which is the documented exception.)
- Nit: the class docstring still says *"Note that `share_name`, `directory_path`, `prefix`, and `dest_gcs` are templated"* — worth adding `directory_name`, since making its templating actually work is what this PR is about.

---

>

providers/google/src/airflow/providers/google/cloud/transfers/azure_fileshare_to_gcs.py
Andrushika: Fixed with the flag approach you suggested. Added `test_mapped_task_aliases_directory_name` that goes through `partial()` -> `unmap()` -> `execute()`. I checked it fails on the previous version of this PR and passes on main. Thanks for catching this!

providers/google/tests/unit/google/cloud/transfers/test_azure_fileshare_to_gcs.py
Andrushika: Dropped, thanks for pointing it out!

Andrushika:
Thanks for reviewing, Shahar! I've addressed the requested changes. Could you take another look when you have time?

shahar1 (approved):

PR #70754, opened 2026-07-30
Allow configuring the default landing page

dsuhinin:
Some deployments barely use the Home dashboard — people open Airflow and immediately click through to the Dag list, every single time. This adds a small knob so the root URL can open the page they actually work in.                                                                                                    
                                                                                                                                                                                                                                                                                                                            
  New `[api] default_landing_page` config option, defaulting to `home` — nothing changes unless you opt in. Setting it to `dags` makes `/` redirect to the Dag list, while the Home dashboard stays reachable at `/home`, so nobody loses anything. The value is exposed through the existing UI config endpoint, and a tiny `LandingPage` component picks between rendering the dashboard and redirecting, so there's no server-side routing involved.                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
  Tests cover both the config endpoint exposing the new option and the landing component's behaviour for each value.


dsuhinin:
hey @bbovenzi . can you please take a look when you will have some time?

bbovenzi:
We need to update the url for the Home button in the NavBar so that you can still get to the dashboard.

dsuhinin:
@bbovenzi going to close this task.  The same change are here and now based on `user settings` page.

PR #70755, opened 2026-07-30
Hide team scoped values of options registered as sensitive

potiuk:
Options are registered as sensitive under their **base** section, but a team scoped
override lives elsewhere:

- config file — a `[<team>=<section>]` section
- environment — `AIRFLOW__<TEAM>___<SECTION>__<KEY>`

Every sensitivity decision was a direct membership test against the registered base
pairs, and the masking pass iterated those pairs and looked each section up
verbatim. A team scoped section was therefore never visited, and its value was
returned in clear text while the same option in the base section was correctly
hidden.

The environment spelling misses by a further step: `_include_envs` splits
`AIRFLOW__TEAM_A___CELERY__BROKER_URL` on `__` with `maxsplit=2`, yielding section
`team_a` and key `_celery__broker_url` — a pair that is neither the base nor the
team scoped section name, so no base keyed loop could reach it either.

### Approach

Sensitivity decisions now resolve the team scoped spelling back to the base option.

- `team_section_name()` builds the config file section name, and is used at **both**
  construction sites so the two representations cannot drift
- `base_section_name()` recovers the base section from it
- `is_sensitive_option()` tests a pair directly, then via the base section, then via
  the tail an environment variable contributes — so it can only ever recognise
  *more* options as sensitive, never fewer

Two deliberate choices:

**The team name is never parsed out of an environment variable name.** A team name
may contain underscores, so splitting `AIRFLOW__<TEAM>___<SECTION>__<KEY>` is
ambiguous. The name is matched against the tail each registered sensitive option
contributes instead. No section ends with `_` or `.` and no key starts with `_`
anywhere in core or providers, so this cannot collide with a global variable.

**The config file section is split on the *last* separator.** Base section names
never contain `=` (checked across all sections in `config.yml` and every `config:`
block in every `provider.yaml`), so the base section is recovered even for a team
name that contains the separator itself. `airflow teams sync` creates teams from
dag bundle config, whose `team_name` field is not validated against the CLI's
charset, so such a name is conceivable.

`routes/public/config.py` needed its own two line change: `get_config_value` never
goes through `as_dict`, so the parser change cannot reach it.

`_get_cmd_option_from_config_sources` and `_get_secret_option_from_config_sources`
are intentionally left alone. They are only called with base pairs, and making them
team aware would mean executing team scoped commands and fetching team secrets,
which the parser deliberately refuses. Those fallbacks are hidden instead.

### Behaviour changes

1. Team scoped values of sensitive options now read `< hidden >` in `as_dict`,
   `GET /config`, `GET /config/section/{section}/option/{option}` and
   `write(hide_sensitive=True)` — symmetric with how the base option already
   behaved. Anyone using the API to read a team's real value will now get the mask.
2. Team scoped `_cmd` / `_secret` entries are replaced with `< hidden >` in place
   rather than resolved and deleted as in a base section, since resolution is not
   supported for a team. The command string itself is no longer shown.
3. `write()` masking now covers team sections, so `airflow config list` is included.
4. Additive public API on the shared parser; no signature changes.
5. No change for non team configuration, and `display_sensitive=True` still returns
   real values.

`as_dict` cost 0.62 ms versus 0.48 ms baseline at 750 options with 40 sensitive
values.

### Test plan

- [x] `shared/configuration/tests/configuration/test_parser.py` — 6 tests: `base_section_name` parametrized including a separator inside the team name; `is_sensitive_option` asserting spoofing fails in both directions; `as_dict` for a file backed team section, with `display_source` and `display_sensitive=True` variants; team `_cmd`/`_secret` fallbacks; team environment variable; `write()`
- [x] 68/68 pass in that module; the behavioural ones fail against the unmodified parser
- [x] `airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py` — `TestTeamScopedOptionMasking`, covering the single option path masked and a non sensitive team scoped option still readable
- [x] `ruff` check and format clean; `mypy` reports the same single pre-existing error as baseline

### Deferred

The synthetic per-key secrets-backend options in `api_fastapi/core_api/services/public/config.py` match literal section names only, so they stay unaware of a team scoped spelling. Nothing leaks today because `_get_custom_secret_backend` is not team aware; tracked in #71037, and linked from a comment at that site.

### DIFF
--- a/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py
+++ b/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py
@@ -22,6 +22,8 @@ from unittest.mock import patch
 
 import pytest
 
+from airflow.configuration import conf
+
 from tests_common.test_utils.config import conf_vars
 
 pytestmark = pytest.mark.db_test
@@ -559,3 +561,66 @@ class TestPerKeyBackendKwargMasking(TestConfigEndpoint):
         )
         assert response.status_code == 200
         assert response.json()["sections"][0]["options"][0]["value"] == OPTION_VALUE_SENSITIVE_HIDDEN
+
+
+TEAM_NAME = "team_a"
+SECTION_TEAM_DATABASE = f"{TEAM_NAME}={SECTION_DATABASE}"
+SECTION_TEAM_CORE = f"{TEAM_NAME}={SECTION_CORE}"
+OPTION_VALUE_TEAM_SQL_ALCHEMY_CONN = "postgresql://team_a_user:team_a_password@team-a-db/airflow"
+OPTION_VALUE_TEAM_PARALLELISM = "512"
+
+
+class TestTeamScopedOptionMasking(TestConfigEndpoint):
+    """A team scoped override of a sensitive option lives in a ``[<team>=<section>]`` section rather
+    than in the section the option is registered under. The single-option endpoint has to resolve
+    that spelling back to the registered option, so that it redacts a team scoped value exactly as
+    it redacts the base one."""
+
+    @pytest.fixture(autouse=True)
+    def setup_team_scoped(self) -> Generator[None, None, None]:
+        team_config = {
+            SECTION_TEAM_DATABASE: {OPTION_KEY_SQL_ALCHEMY_CONN: OPTION_VALUE_TEAM_SQL_ALCHEMY_CONN},
+            SECTION_TEAM_CORE: {OPTION_KEY_PARALLELISM: OPTION_VALUE_TEAM_PARALLELISM},
+        }
+
+        # Anything that is not one of the team scoped options under test keeps being read from the
+        # real configuration - the endpoint reads [api] expose_config through the same object.
+        real_get = conf.get
+
+        def _mock_has_option(section: str, option: str, **_) -> bool:
+            return option in team_config.get(section, {})
+
+        def _mock_get(section: str, option: str, *args, **kwargs) -> str:
+            if option in team_config.get(section, {}):
+                return team_config[section][option]
+            return real_get(section, option, *args, **kwargs)
+
+        with (
+            conf_vars(AIRFLOW_CONFIG_ENABLE_EXPOSE_CONFIG),
+            patch(
+                "airflow.api_fastapi.core_api.routes.public.config.conf.has_option",
+                new=_mock_has_option,
+            ),
+            patch(
+                "airflow.api_fastapi.core_api.routes.public.config.conf.get",
+                new=_mock_get,
+            ),
+        ):
+            yield
+
+    def test_get_config_value_masks_team_scoped_sensitive_option(self, test_client):
+        response = test_client.get(
+            f"/config/section/{SECTION_TEAM_DATABASE}/option/{OPTION_KEY_SQL_ALCHEMY_CONN}",
+            headers=HEADERS_JSON,
+        )
+        assert response.status_code == 200
+        assert response.json()["sections"][0]["options"][0]["value"] == OPTION_VALUE_SENSITIVE_HIDDEN
+
+    def test_get_config_value_keeps_team_scoped_non_sensitive_option(self, test_client):
+        """An option that is not registered as sensitive stays readable under a team scoped section."""
+        response = test_client.get(
+            f"/config/section/{SECTION_TEAM_CORE}/option/{OPTION_KEY_PARALLELISM}",
+            headers=HEADERS_JSON,
+        )
+        assert response.status_code == 200
+        assert response.json()["sections"][0]["options"][0]["value"] == OPTION_VALUE_TEAM_PARALLELISM
--- a/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py
+++ b/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py
@@ -617,7 +617,6 @@ class TestTeamScopedOptionMasking(TestConfigEndpoint):
         assert response.json()["sections"][0]["options"][0]["value"] == OPTION_VALUE_SENSITIVE_HIDDEN
 
     def test_get_config_value_keeps_team_scoped_non_sensitive_option(self, test_client):
-        """An option that is not registered as sensitive stays readable under a team scoped section."""
         response = test_client.get(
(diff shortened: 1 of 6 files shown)


shared/configuration/src/airflow_shared/configuration/parser.py
```diff
@@ -562,6 +590,68 @@ def sensitive_config_values(self) -> set[tuple[str, str]]:
...
+                return True
+            # A team scoped ``_cmd`` / ``_secret`` fallback is not resolved into its value, so it
+            # stays in the output as configured and has to be recognised on its own.
+            for fallback_suffix in ("_cmd", "_secret"):
+                if not key.endswith(fallback_suffix):
+                    continue
+                if (base_section, key.removesuffix(fallback_suffix)) in self.sensitive_config_values:
+                    return True
+        # A team scoped environment variable is reported under the section and key its name splits
+        # into, which is neither the base nor the team scoped section name.
```
amoghrajesh, line 652: Same explanation repeated at `_include_envs`. keep any one.

shared/configuration/tests/configuration/test_parser.py
```diff
@@ -1124,6 +1125,152 @@ def test_team_env_var_format(self):
...
+        assert test_conf.is_sensitive_option("test", "sensitive_key")
+        assert test_conf.is_sensitive_option("team_a=test", "sensitive_key")
+        assert test_conf.is_sensitive_option("team-a=test", "sensitive_key")
+        # A team name is never able to hide the base section it overrides, not even when it
+        # contains the separator itself.
+        assert test_conf.is_sensitive_option("a=team=test", "sensitive_key")
+        # The section and key an AIRFLOW__<TEAM>___<SECTION>__<KEY> variable name splits into.
+        assert test_conf.is_sensitive_option("team_a", "_test__sensitive_key")
+
+        # An option that is not registered as sensitive stays readable under any spelling.
```
amoghrajesh, line 1153: Same comment verbatim at line 1234, and a near duplicate at 1182. Same point three times, we can drop some.

amoghrajesh (commented):
Separate issue found while reviewing: https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/api_fastapi/core_api/services/public/config.py#L40-L59 only matches literal secrets/workers section names, not their team scoped spelling. Same root cause, different mechanism, untouched here. `_get_custom_secret_backend` is not team aware either so nothing real leaks today but worth a follow up (cc: @vincbeck / @o-nikolas)

Although I am ok with this PR, would like either @vincbeck or @o-nikolas to review as well.

shared/configuration/src/airflow_shared/configuration/parser.py
potiuk: Trimmed — `_include_envs` now just says it covers the cmd/secret variants too and points at `is_sensitive_option`, which is where the reasoning belongs.

---

shared/configuration/tests/configuration/test_parser.py
potiuk: Dropped two of the three; the remaining one sits on the first test that makes the point.

---

potiuk:
Both threads addressed and resolved — duplicated rationale trimmed to one site in each case.

The separate gap you spotted in `services/public/config.py` is filed as #71037 and linked from a comment at that site; it is unreachable today since the secrets backend is not team aware, so it is recorded rather than fixed here.

@amoghrajesh

---

shared/configuration/tests/configuration/test_parser.py
```diff
@@ -1124,6 +1125,150 @@ def test_team_env_var_format(self):
...
+    def test_is_sensitive_option_resolves_team_scoped_names(self):
+        """A team scoped spelling of a registered sensitive option is recognised as sensitive."""
+        test_conf = AirflowConfigParser()
+        test_conf.sensitive_config_values.add(("test", "sensitive_key"))
+
+        assert test_conf.is_sensitive_option("test", "sensitive_key")
+        assert test_conf.is_sensitive_option("team_a=test", "sensitive_key")
+        assert test_conf.is_sensitive_option("team-a=test", "sensitive_key")
+        # A team name is never able to hide the base section it overrides, not even when it
+        # contains the separator itself.
```
amoghrajesh:
```suggestion
```

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_config.py
```diff
@@ -559,3 +561,66 @@ def test_get_config_value_masks_per_key_secrets_backend_kwarg(self, test_client)
...
+    def test_get_config_value_masks_team_scoped_sensitive_option(self, test_client):
+        response = test_client.get(
+            f"/config/section/{SECTION_TEAM_DATABASE}/option/{OPTION_KEY_SQL_ALCHEMY_CONN}",
+            headers=HEADERS_JSON,
+        )
+        assert response.status_code == 200
+        assert response.json()["sections"][0]["options"][0]["value"] == OPTION_VALUE_SENSITIVE_HIDDEN
+
+    def test_get_config_value_keeps_team_scoped_non_sensitive_option(self, test_client):
+        """An option that is not registered as sensitive stays readable under a team scoped section."""
```
amoghrajesh:
```suggestion
```

shared/configuration/src/airflow_shared/configuration/parser.py
```diff
@@ -77,6 +77,34 @@ def _collect_kwarg_env_vars(prefix: str) -> dict[str, str]:
...
+    Build the config file section name that holds the team scoped overrides of ``section``.
+
+    :param team_name: name of the team the overrides belong to
+    :param section: base section name that is being overridden
+    :return: the team scoped section name, e.g. ``team_a=celery``
+    """
+    return f"{team_name}{TEAM_SECTION_SEPARATOR}{section}"
+
+
+def base_section_name(section: str) -> str:
```
amoghrajesh, line 95: Do these have to be in the public API surface?

amoghrajesh (approved):
I have a qn and few nits.

shared/configuration/src/airflow_shared/configuration/parser.py
potiuk: I believe so.

potiuk: They are the reusable, shared configuration-related small util methods, so .. it checks all the boxes IMHO

Issue #70388, opened 2026-07-24
Support OpenTelemetry Auto-Instrumentation, Distros & Configurators


tailg8nj:
## Background

In attempting to leverage the auto-instrumentation mechanism with a vendored distribution , I have run into a few issues. I believe with a few small tweaks this could be much polished and remove the need for workarounds.

## Setup

Auto instrumentation has two ways of being invoked, one using the command line wrapper `opentelemetry-instrument` and the other explicitly invoking it in code. Some of the issues identified later in this document force the latter approach.

Placing the following at the top of `airflow_local_settings.py` achieves autoinstrumentation:

```python
from opentelemetry.instrumentation import auto_instrumentation

auto_instrumentation.initialize()

from airflow._shared.observability.traces import OverrideableRandomIdGenerator
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.trace import get_tracer_provider

if isinstance((tracer_provider := get_tracer_provider()), TracerProvider):
    tracer_provider.id_generator = OverrideableRandomIdGenerator()
```

## Issue 1: OverrideableRandomIdGenerator

*Change Difficulty: Easy*

This class was introduced in #63839. My understanding is this a key piece in the task span propagation strategy, but also looks alot like an anti-pattern: creating a span, just to throw it away, and re-use it's IDs later. This is not a blocker. See my comment requesting a more detailed explanation https://github.com/apache/airflow/pull/63839#discussion_r3544019341.

The workaround is to explicitly override any Distro's ID Generator, e.g.

```python
tracer_provider.id_generator = OverrideableRandomIdGenerator()
```

This class is marked private.

OpenTelemetry offers an environment variable to choose your ID generator, so long as the class is a defined entry point.

Could it be exposed using a Python Entry Point so that others could enforce it via ENV VAR such that I wouldn't have to reference it directly?

  * [Upstream ID Generator Creation](https://github.com/open-telemetry/opentelemetry-python/blob/634cec5f2a2fecb40cb9d8216888c7b8865b845a/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L627)
  * [Upstream ID Generator Environment Variable](https://github.com/open-telemetry/opentelemetry-python/blob/634cec5f2a2fecb40cb9d8216888c7b8865b845a/opentelemetry-api/src/opentelemetry/environment_variables/__init__.py#L49)
[...]

PR #70758, opened 2026-07-30
Support OpenTelemetry auto-instrumentation and vendor distros

rjgoyln:
## Summary

Airflow derives Dag run and task run span ids from the metadata database through a custom OpenTelemetry `IdGenerator`, so spans from different components join into one trace. That generator was only ever installed on a `TracerProvider` Airflow built itself — under `opentelemetry-instrument` or a vendor distro the provider already exists, OpenTelemetry refuses to replace it, and the linkage is lost without a word.

## Change

- Register `OverrideableRandomIdGenerator` under the `opentelemetry_id_generator` entry point group, selectable as `OTEL_PYTHON_ID_GENERATOR=airflow`.
- Patch the generator onto a provider Airflow did not build, and warn, instead of discarding it.
- Document the setup, the fallback's limits, and that `[traces]` exporter settings do not reach a borrowed provider.

The fallback is best-effort by construction — a `Tracer` keeps the generator it was built with, so only later tracers are linked — which is why the environment variable is the documented path.

The entry point is declared in `airflow-core` because `shared/observability` is vendored into consumers rather than published, so it has no installed metadata for OpenTelemetry to discover. Workers ride propagated `traceparent` values and need no override.

## Scope

Issue 1 of #70388 only; issues 2 and 3 remain open.

* related: #70388

---


airflow-core/pyproject.toml
```diff
@@ -211,6 +211,11 @@ dependencies = [
 [project.scripts]
 airflow = "airflow.__main__:main"
 
+# Lets OpenTelemetry auto-instrumentation and vendor distros select Airflow's id generator
+# with OTEL_PYTHON_ID_GENERATOR=airflow, instead of reaching into a private module.
+[project.entry-points.opentelemetry_id_generator]
```
tailg8nj, line 216: I'm not well versed in how this project is composed w.r.t the shared observability code, but should this entry point be listed here https://github.com/apache/airflow/blob/main/shared/observability/pyproject.toml ?

tailg8nj (commented):
This is a quick turn around! Overall this now allows a distro to set this environment variable (removing the need for custom code in local settings) if it detects it is operating on an airflow env. :+1:

airflow-core/pyproject.toml
rjgoyln: It can't be registered in `shared/` because that folder is not a published distribution, it's just forcefully vendored into the consumer builds at compile time. 
Therefore, the entry point must be declared in the actual installed package `apache-airflow-core` so OpenTelemetry can discover it at runtime.

potiuk (commented):
Nice piece of work. Registering the generator as a standard entry point is the right shape for this, and the caveat you documented is accurate — I checked the SDK rather than taking it on trust: `TracerProvider.get_tracer()` passes `self.id_generator` into `Tracer.__init__` (`sdk/trace/__init__.py:1454` → `:1119-1130`) and the tracer then uses its own captured reference, so patching after the fact really does affect only tracers built later. Precisely what `traces.rst` says.

Two design decisions I'd like written down before this lands. Neither is a defect.

### The adoption runs before `otel_on` is read

```python
def configure_otel(conf: ConfigParser):
    if _adopt_existing_tracer_provider():
        return
    otel_on = conf.getboolean("traces", "otel_on", fallback=False)
```

So a deployment running under `opentelemetry-instrument` with `otel_on=False` — tracing deliberately switched off in Airflow — gets its `TracerProvider.id_generator` swapped anyway, plus a warning on every component start advising it to set `OTEL_PYTHON_ID_GENERATOR=airflow`. That's guidance about a feature the operator has turned off, repeated at every startup. Before this change, `configure_otel` with `otel_on=False` did nothing at all.

Your test parametrizes `otel_on=[True, False]` and asserts the patch in both, so this looks deliberate — could you say why in a comment? My instinct is that the adoption should sit behind the `otel_on` check.

### The early return silently retires the `[traces]` exporter settings

When a provider already exists and `otel_on=True`, the `return` skips all exporter and span-processor setup. I think that's *correct* — with a vendor distro you don't want Airflow bolting a second OTLP exporter onto a provider the distro already configured — but nothing tells the operator. The warning mentions only span ids, and the new docs section covers only the generator caveat.

Someone who has `otel_host` / `otel_port` configured and then adopts auto-instrumentation gets no signal that those settings stopped applying. One sentence in the warning and one in the docs would cover it. If you considered `provider.add_span_processor(...)` on the borrowed provider and rejected it, that reasoning is worth a line in the docs too.

### Smaller things

The entry point hard-codes `airflow._shared.observability.traces`, but that source is importable by three paths — `airflow._shared…`, `airflow.sdk._shared…` (task-sdk symlinks the same directory) and `airflow_shared…` — each a distinct module object with a distinct class object. So the `isinstance` check in `_adopt_existing_tracer_provider` only matches when both sides arrived via the same path. It holds today, since `settings.py:42` is the sole caller and uses `airflow._shared…`. But if task-sdk ever calls `configure_otel` it would silently fail the check and double-patch. A duck-typed check, or a comment recording the coupling, would make that safe.

The description says the newsfragment is `70388.improvement.rst`; commit `2062b425` renamed it to `70758.improvement.rst` but the body still names the old one.

Finally, the Gen-AI disclosure box is unchecked here while your other two open PRs declare Claude Code — just confirming that's deliberate.

### Worth a second look from

This touches the span-id propagation design, so I'd like input from the people who built it:

- @xBis7 — authored the trace-linkage work in `shared/observability/.../traces/` and part of `traces.rst`; best placed to say whether adopting a foreign `TracerProvider` preserves the Dag-run/task-run linkage in practice.
- @ferruzzi — deepest recent history on Airflow's OTel metrics and telemetry config; the `otel_on` question above is really about how Airflow's telemetry switches should behave when someone else owns the provider.

No rush on either — this has been open since July and the branch is 1271 commits behind `main`, so it needs a rebase before it can merge regardless.

---

>

---

shared/observability/src/airflow_shared/observability/traces/__init__.py
```diff
@@ -248,7 +248,36 @@ def _load_exporter_from_env() -> SpanExporter:
...
+            "A TracerProvider was already installed; patched in %s so Airflow span ids "
+            "keep propagating. Set OTEL_PYTHON_ID_GENERATOR=airflow to have it applied "
+            "when the provider is built instead.",
+            OverrideableRandomIdGenerator.__name__,
+        )
+    return True
+
+
 def configure_otel(conf: ConfigParser):
+    if _adopt_existing_tracer_provider():
```
potiuk, line 287: This runs before `otel_on` is read on line 281, so the adoption happens whether or not Airflow tracing is enabled.

The case I'd want to be sure about: a deployment under `opentelemetry-instrument` that has deliberately set `otel_on=False`. Today `configure_otel` does nothing for them. After this change their `TracerProvider.id_generator` is replaced and every component start logs the warning from line 268, recommending an environment variable for a feature they have switched off.

`test_patches_id_generator_onto_existing_provider` parametrizes `otel_on=[True, False]` and asserts the patch in both, so this reads as intentional rather than accidental — could you record the reasoning in a comment here? My instinct is that the adoption belongs behind the `otel_on` check, but you have thought about this more than I have.

Separately, this `return` also skips the entire exporter and span-processor setup below. I think that is the right call — a vendor distro has already configured its own exporters and a second one would be worse — but an operator with `otel_host` / `otel_port` in their config gets no indication those settings have stopped applying. Worth a sentence in the warning on line 268 and in the new docs section.

airflow-core/pyproject.toml
potiuk, line 226: Worth a note here about a coupling that isn't visible from this file.

The same source is importable by three paths — `airflow._shared.observability.traces`, `airflow.sdk._shared.observability.traces` (task-sdk symlinks the same directory) and `airflow_shared.observability.traces` — and Python treats each as a separate module with its own `OverrideableRandomIdGenerator` class object. So the `isinstance` check in `_adopt_existing_tracer_provider` only succeeds when the entry-point-loaded class and the checking code came through the same path.

That holds right now: `settings.py:42` is the only caller of `configure_otel` and imports `airflow._shared…`, matching this entry point exactly. But if task-sdk ever calls it, the check fails silently and a second generator gets patched in over a perfectly good one, with a misleading warning.

Either a duck-typed check on the generator side, or a comment here recording that this path must match whatever `configure_otel` imports, would keep that from being a surprise later.

rjgoyln:
Thanks for the review. Rebased onto `main` and pushed as `f55fedb`.

### On `otel_on`

This is deliberate. `otel_on` only controls whether Airflow creates its own provider; spans can still reach a distro-owned provider when it is false. Gating this on `otel_on` would leave those spans with random IDs. Added a comment and parametrised test to cover this.

### On exporter settings

Agreed — documented in the warning, `traces.rst`, and newsfragment. The settings were already ineffective with a foreign provider, so the early return avoids constructing a provider that cannot be installed.

### On import paths

Confirmed. The task SDK already imports this module, so the check stays at the `configure_otel` entry point, with the coupling documented there.

Happy to switch to duck typing if preferred.

PR #70760, opened 2026-07-30
Fix missing _team_name on DagRun before some listener calls

kacpermuda:
In multi-team deployments the scheduler fires `on_dag_run_running / success / failed` listener hooks from several code paths. Before this change, `DagRun._team_name` was not guaranteed to be set when those hooks fired, so OpenLineage (and any other plugin reading `_team_name`) would miss the team on some events.

`_team_name` is a private  attribute — it is not part of any public API and is not guaranteed to be present on every `DagRun` object a listener receives. Since the value is already cached in `self._dag_id_to_team_name` for metrics purposes, setting it consistently before each hook fires adds no cost. Listeners that wish to use it should treat it as best-effort and guard accordingly.

This PR ensures `_team_name` is stamped on every ORM `DagRun` object before `notify_dagrun_state_changed()` is called, covering all five listener code paths. The stamping is cheap: `_get_team_names_for_dag_ids` caches results in `self._dag_id_to_team_name` for the lifetime of a scheduler loop, so most calls are a dict read with no DB query. I've also added comments to the existing stamping paths, just to make sure it's clear why it's done before listener call.

---

- [X] Yes — Claude Code (Sonnet 4.6)

Generated-by: Claude Code (Sonnet 4.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -1734,6 +1734,10 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                     .group_by(DagRun)
                 )
             )
+            # Stamp _team_name before update_state() calls notify_dagrun_state_changed(),
+            # which fires on_dag_run_success/failed listeners.
+            # _get_team_names_for_dag_ids caches results in self._dag_id_to_team_name, so
+            # this is usually a dict read with no DB query.
             if self._multi_team and paused_runs:
                 paused_dag_ids = {dr.dag_id for dr in paused_runs}
                 paused_team_mapping = self._get_team_names_for_dag_ids(paused_dag_ids, session)
@@ -1996,6 +2000,12 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 )
             )
 
+            # Stamp _team_name before _schedule_all_dag_runs calls notify_dagrun_state_changed()
+            # (on_dag_run_success/failed listeners). Runs that just transitioned QUEUED→RUNNING
+            # in _start_queued_dagruns may already have _team_name set (SQLAlchemy identity map
+            # returns the same Python objects); overwriting with the same value is harmless.
+            # _get_team_names_for_dag_ids caches results in self._dag_id_to_team_name, so this
+            # is usually a dict read with no DB query.
             if self._multi_team and dag_runs:
                 unique_dag_ids = {dr.dag_id for dr in dag_runs}
                 dr_team_mapping = self._get_team_names_for_dag_ids(unique_dag_ids, session)
@@ -2844,6 +2854,15 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             partial(self.scheduler_dag_bag.get_dag_for_run, session=session)
         )
 
+        # Stamp _team_name on each queued DagRun before the listener fires.
+        # Results are usually already cached in self._dag_id_to_team_name from the
+        # current or previous scheduler loop, so this is typically a dict read with no DB query.
+        if self._multi_team and dag_runs:
+            queued_team_map = self._get_team_names_for_dag_ids({dr.dag_id for dr in dag_runs}, session)
+            for dr in dag_runs:
+                if team := queued_team_map.get(dr.dag_id):
+                    dr._team_name = team
+
         for dag_run in dag_runs:
             dag_id = dag_run.dag_id
             run_id = dag_run.run_id
@@ -2981,6 +3000,12 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 execute=False,
             )
 
+            # dag_run was reloaded from DB above, so _team_name set on the original object is lost.
+            # Re-stamp it before the listener fires; the result is almost certainly cached in
+            # self._dag_id_to_team_name already, so this is a dict read with no DB query.
+            if self._multi_team:
+                if team := self._get_team_names_for_dag_ids([dag_run.dag_id], session).get(dag_run.dag_id):
+                    dag_run._team_name = team
             dag_run.notify_dagrun_state_changed(msg="timed_out")
             if dag_run.end_date and dag_run.start_date:
                 duration = dag_run.end_date - dag_run.start_date
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -458,6 +458,25 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         # Ensure all requested dag_ids are in the result (with None for those not found)
         return {dag_id: self._dag_id_to_team_name.get(dag_id) for dag_id in dag_ids}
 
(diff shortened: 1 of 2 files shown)


kacpermuda:
cc @ferruzzi , FYI as I see that you worked on adding this before

mobuchowski (approved):

uranusjr:
I would introduce a helper (say `_stamp_team_names`) to do the stamping instead of repeating the same logic multiple times. It would also help clean up the verbose comment added on every stamping block.

kacpermuda:
> I would introduce a helper (say `_stamp_team_names`) to do the stamping instead of repeating the same logic multiple times. It would also help clean up the verbose comment added on every stamping block.

@uranusjr Extracted the logic to a separate helper, ptal

airflow-core/tests/unit/jobs/test_scheduler_job.py
```diff
@@ -9635,6 +9635,205 @@ def test_dag_timeout_notifies_with_timed_out_msg(self, mock_get_listener_manager
...
 
+    @conf_vars({("core", "multi_team"): "true"})
+    @mock.patch("airflow.models.dagrun.get_listener_manager")
+    def test_dag_start_notifies_listener_with_team_name(self, mock_get_listener_manager, dag_maker, session):
+        """Test that on_dag_run_running receives dag_run with _team_name set."""
+        mock_listener_manager = MagicMock()
+        mock_get_listener_manager.return_value = mock_listener_manager
+
+        clear_db_teams()
+        clear_db_dag_bundles()
```
ferruzzi: Clearing the db at the beginning of the tests should be a code smell.  I believe this issue was fixed in #69093, if not, then we need to figure out what tests are not cleaning up after themselves and fix them.

airflow-core/tests/unit/jobs/test_scheduler_job.py
```diff
@@ -9635,6 +9635,205 @@ def test_dag_timeout_notifies_with_timed_out_msg(self, mock_get_listener_manager
...
+
+        team = Team(name="test_team")
+        session.add(team)
+        session.flush()
+
+        bundle = DagBundleModel(name="test_bundle")
+        bundle.teams.append(team)
+        session.add(bundle)
+        session.flush()
```
ferruzzi: These tests all have a pretty big repetitive block at the top, maybe move the setup into a shared fixture if none of the existing fixtures do what you need?  look at the `testing_team` fixture though, I think that covers most/all of this boilerplate.

vatsrahul1001:
@kacpermuda can you resolve comments?

airflow-core/tests/unit/jobs/test_scheduler_job.py
kacpermuda: Correct, I should have just used the fixtures currently present. Adjusted the code.

airflow-core/tests/unit/jobs/test_scheduler_job.py
kacpermuda: Correct, used `testing_team` and `testing_dag_bundle` fixtures instead, and linked them together in a new small fixture that I reused in all tests. Should be clean now, thanks for spotting that.

kacpermuda:
@vatsrahul1001 addressed all the comments

vatsrahul1001:
LGTM!, can be merged after code owners review

vatsrahul1001 (approved):

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -458,6 +458,25 @@ def _get_team_names_for_dag_ids(
         # Ensure all requested dag_ids are in the result (with None for those not found)
         return {dag_id: self._dag_id_to_team_name.get(dag_id) for dag_id in dag_ids}
 
+    def _stamp_team_names(self, dag_runs: Iterable[DagRun], session: Session) -> None:
```
jason810496:
```suggestion
    def _stamp_team_names(self, dag_runs: Collection[DagRun], session: Session) -> None:
```

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -458,6 +458,25 @@ def _get_team_names_for_dag_ids(
...
+        practice this means the first call per loop issues one batched query; subsequent calls
+        for the same dag_ids are pure dict reads with no DB round-trip.
+        """
+        if not self._multi_team:
+            return
+        runs = list(dag_runs)
+        if not runs:
+            return
+        team_map = self._get_team_names_for_dag_ids({dr.dag_id for dr in runs}, session)
+        for dr in runs:
```
jason810496: With the `Collection[DagRun]` annotation, the `list()` materialisation is no longer needed.

```suggestion
        if not dag_runs:
            return
        team_map = self._get_team_names_for_dag_ids({dr.dag_id for dr in dag_runs}, session)
        for dr in dag_runs:
```

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2981,6 +2995,9 @@ def _schedule_dag_run(
                 execute=False,
             )
 
+            # dag_run was reloaded from DB above, so _team_name set on the original object is lost.
+            # Team name should be added before listeners are called in notify_dagrun_state_changed()
```
jason810496: The first line does not hold: `_team_name` is not a mapped attribute (`DagRun.stats_tags` reads it with `getattr(self, "_team_name", None)`), and the reload above is a `select()` by primary key on the *same* session, so SQLAlchemy's identity map returns the object that is already loaded, i.e. `dag_run_reloaded is dag_run`. Even if it were expired or refreshed, that only touches mapped columns, never a plain instance attribute. So nothing is lost by the reload, and in the production path (`_do_scheduling` stamps, then `_schedule_all_dag_runs` -> `_schedule_dag_run`) the run arrives here already stamped.

Suggest dropping the inaccurate sentence and keeping the same wording as the other three call sites:

```suggestion
            # Team name should be added before listeners are called in notify_dagrun_state_changed()
```

---
Drafted-by: Claude Code (Opus 5); reviewed by @jason810496 before posting

jason810496 (commented):
Thanks, LGTM overall.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
kacpermuda: applied !

airflow-core/src/airflow/jobs/scheduler_job_runner.py
kacpermuda: applied !

airflow-core/src/airflow/jobs/scheduler_job_runner.py
kacpermuda: applied !

vatsrahul1001:

airflow-core/tests/unit/jobs/test_scheduler_job.py
ferruzzi: Much cleaner.  Thanks.

PR #70771, opened 2026-07-30
Add asset state store commands to airflowctl

justinpakzad:
---
This PR adds asset state store commands to airflowctl. This would allow users to get, list, delete, clear, and set key/values in an asset state store. All of the new commands correspond to existing endpoints that are available via the REST API. Unit and integration tests were added for each command and they were also tested against an actual state store via breeze. 

Note - another way to structure this would be by introducing new operations/command groups (e.g., StateStoreOperations or AssetsStateStoreOperations), I felt grouping them under `assets` made the most sense. Open to suggestions though.

---


justinpakzad:
Would appreciate a review on this. Thanks.

justinpakzad:
Requesting review on this. Thanks.

airflow-ctl/src/airflowctl/api/operations.py
```diff
@@ -326,6 +329,39 @@ def delete_queued_event(self, dag_id: str, asset_id: str) -> str | ServerRespons
...
+
+    def get_state_store(self, asset_id: str, key: str) -> AssetStateStoreResponse | ServerResponseError:
+        """Get a single asset state store entry."""
+        self.response = self.client.get(f"assets/{asset_id}/state-store/{key}")
+        return AssetStateStoreResponse.model_validate_json(self.response.content)
+
+    def set_state_store(self, asset_id: str, key: str, value: str) -> str | ServerResponseError:
+        """Set an asset state store value. Creates or overwrites the key."""
+        try:
+            parsed_value = json.loads(value)
```
SameerMesiah97, line 346: Is the implicit JSON coercion here intentional? Values such as `"123"`, `"true"`, and `"null"` will be stored as an integer, boolean, and null respectively, whereas `"hello"` remains a string. This could be surprising to users expecting the CLI argument to be preserved as text. If supporting JSON values is intended, could this behaviour be made clear in the command help?

airflow-ctl-tests/tests/airflowctl_tests/test_airflowctl_commands.py
```diff
@@ -56,6 +56,11 @@ def date_param():
     "assets list",
     "assets get 1",
     "assets create-event --asset-id=1",
+    "assets set-state-store 1 test_key test_value",
+    "assets get-state-store 1 test_key",
+    "assets list-state-store 1",
+    "assets delete-state-store 1 test_key",
+    "assets clear-state-store 1",
```
SameerMesiah97, line 64: The integration command sequence deletes `test_key` before running `clear-state-store`, so the clear command is only exercised against an already-empty state store. Could we set another key before clearing it so that the command is tested with an existing entry?

SameerMesiah97 (commented):
Just 2 minor issues. Looks good otherwise.

airflow-ctl/src/airflowctl/api/operations.py
justinpakzad: Yes, it's intentional. The declared type in the Pydantic model (which comes from the API) for the state store `value` field  is `JsonValue` so it's meant to accept any JSON type. This the same pattern used in the `XComOperations` (add and edit commands). Your point about user expectation is still valid though. I modified the help text so it's a bit more clear.

airflow-ctl-tests/tests/airflowctl_tests/test_airflowctl_commands.py
justinpakzad: Yep, I missed that. I've added another set before the clear command.

justinpakzad:
> Just 2 minor issues. Looks good otherwise.

Appreciate the review as always. Addressed both comments.

SameerMesiah97 (approved):

airflow-ctl/src/airflowctl/api/operations.py
```diff
@@ -326,6 +329,39 @@ def delete_queued_event(self, dag_id: str, asset_id: str) -> str | ServerRespons
         self.client.delete(f"dags/{dag_id}/assets/{asset_id}/queuedEvents")
         return asset_id
 
+    def list_state_store(self, asset_id: str) -> AssetStateStoreCollectionResponse | ServerResponseError:
+        """List all state store entries for an asset."""
+        return super().execute_list(
+            path=f"/assets/{asset_id}/state-store", data_model=AssetStateStoreCollectionResponse
```
henry3260:
```suggestion
            path=f"assets/{asset_id}/state-store", data_model=AssetStateStoreCollectionResponse
```

henry3260 (commented):
Sorry for the late review
PR description lists "list, delete, clear, and set", but the diff also adds `get-state-store`. Might be worth mentioning it so the description matches the commands.

airflow-ctl/src/airflowctl/ctl/help_texts.yaml
```diff
@@ -28,6 +28,11 @@ assets:
   delete-queued-events: "Delete all queued events for a given asset"
   delete-dag-queued-events: "Delete all queued asset events for a given Dag"
   delete-queued-event: "Delete a specific queued asset event for a given Dag and asset"
+  list-state-store: "List all state store entries for an asset"
+  get-state-store: "Retrieve a single asset state store entry by its ID and key"
+  set-state-store: "Set an asset state store value. Parsed as JSON if valid, otherwise stored as a string"
```
henry3260, line 33: The `json.loads` fallback turns the literal string `null` into `None`, which the client-side `AssetStateStoreBody` accepts but the core validator rejects with a 422 ("value cannot be null"). The CLI surfaces it as a server error and exits 1, so nothing crashes, but the help text ("Parsed as JSON if valid, otherwise stored as a string") could note that `null` is not allowed.

justinpakzad: Fair point, I had a look and the server also rejects NaN/Infinity as well as oversized values (larger than `max_value_storage_bytes`). I'm not sure it's necessary to call them out in the help texts since the client-side parsing does in fact parse those successfully, it's the server that ends up rejecting them (as you mentioned) and it surfaces a clear error message. That said, happy to add a note if you feel strongly about it.

henry3260: > Fair point, I had a look and the server also rejects NaN/Infinity as well as oversized values (larger than `max_value_storage_bytes`). I'm not sure it's necessary to call them out in the help texts since the client-side parsing does in fact parse those successfully, it's the server that ends up rejecting them (as you mentioned) and it surfaces a clear error message. That said, happy to add a note if you feel strongly about it.

Agreed

airflow-ctl/src/airflowctl/api/operations.py
justinpakzad: Fixed.

justinpakzad:
> Sorry for the late review PR description lists "list, delete, clear, and set", but the diff also adds `get-state-store`. Might be worth mentioning it so the description matches the commands.

Thanks for the review! I've addressed the comments and added "get" to the list of commands specified in the description.

henry3260 (approved):
Thanks

justinpakzad:
Seems the test failure was a transient error. Would appreciate if someone can give it a rerun.

PR #70452, opened 2026-07-25
Reuse existing ORM session when resolving OpenLineage dag_team_name

SameerMesiah97:
**Description**

This is a follow-up to PR #69109, which introduced the `dag_team_name` field to the OpenLineage `airflowDagRun` run facet.

This change updates the OpenLineage provider to reuse the existing SQLAlchemy session associated with the `DagRun` when resolving the `dag_team_name` field.

**Rationale**

During manual testing, it was discovered that creating a new managed session triggered Airflow's HA lock protection by performing an unexpected commit within the scheduler callback. Reusing the existing `DagRun` session ensures the lookup participates in the surrounding transaction while continuing to emit the `dag_team_name` field correctly.

**Tests**

Updated the existing `test_dag_run_team_name` unit test to verify that the existing SQLAlchemy session associated with the `DagRun` is reused when resolving `dag_team_name`.

###### Was generative AI tooling used to co-author this PR?

- [X] Yes (please specify the tool below)

Generated-by: [GPT 5.5] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70772, opened 2026-07-30
Fix openlineage team name handling, add dag_version_data and TI note

kacpermuda:
The previous implementation called `DagBundleModel.get_team_name()` without an explicit session. This triggered `@provide_session` to open a new `create_session()`, which — for ORM `DagRun` objects in the scheduler loop — resolves to the same scoped session and commits on exit. That commit fires Airflow's `prohibit_commit` HA lock guard, producing errors in multi-scheduler setups.

The fix switches to `object_session(dagrun)` to obtain the session the ORM object is already attached to, then passes it explicitly to `DagModel.get_team_name(dag_id, session=session)`, bypassing `@provide_session`'s `create_session()` path entirely. A similar fix was proposed in #70452; this PR simplifies it further by going directly through `DagModel` instead of the bundle.

Once #70760 is in, no DB call will be required for scheduler-side events in Airflow 3.3.1: the stamp is already in memory. However, the `_team_name` attribute is private with no stability guarantee; the check is guarded with `hasattr` and an `isinstance` check; so we might keep the db call here under try/except with fallback to None.

Two new fields added as well:

`DagVersion.version_data` (added in Airflow 3.3) carries arbitrary DAG-level metadata set at parse time. This PR surfaces it as `dag_version_data` in the OpenLineage `DagRunInfo` facet. No logic — just reading the attribute and attaching it to the event, guarded by `AIRFLOW_V_3_3_PLUS`.

`TaskInstance.note` is set when an operator or a user manually changes task state in the UI. Attaching it to OpenLineage task events makes manual interventions observable in lineage.

---

- [X] Yes — Claude Code (Sonnet 4.6)

Generated-by: Claude Code (Sonnet 4.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
+++ b/providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
@@ -982,6 +982,9 @@ class DagRunInfo(InfoJsonEncodable):
         "dag_bundle_version": lambda dagrun: DagRunInfo.dag_version_info(dagrun, "bundle_version"),
         "dag_version_id": lambda dagrun: DagRunInfo.dag_version_info(dagrun, "version_id"),
         "dag_version_number": lambda dagrun: DagRunInfo.dag_version_info(dagrun, "version_number"),
+        "dag_version_data": lambda dagrun: (
+            DagRunInfo.dag_version_info(dagrun, "version_data") if AIRFLOW_V_3_3_PLUS else None
+        ),
         "dag_team_name": lambda dagrun: DagRunInfo.team_name(dagrun) if AIRFLOW_V_3_3_PLUS else None,
         "deadlines": lambda dagrun: DagRunInfo.deadlines(dagrun),
     }
@@ -1039,7 +1042,7 @@ class DagRunInfo(InfoJsonEncodable):
         return {"alerts": result} if result else None
 
     @classmethod
-    def dag_version_info(cls, dagrun: DagRun, key: str) -> str | int | None:
+    def dag_version_info(cls, dagrun: DagRun, key: str) -> str | int | dict | None:
         """Extract DAG version info for given key, sourced from DagRun (on scheduler)."""
         # AF2 DagRun and AF3 DagRun SDK model (on worker) do not have this information
         dag_versions = safe_getattr(dagrun, "dag_versions", [])
@@ -1057,6 +1060,10 @@ class DagRunInfo(InfoJsonEncodable):
             return str(version_id) if version_id is not None else None
         if key == "version_number":
             return safe_getattr(current_version, "version_number")
+        if key == "version_data":
+            if not AIRFLOW_V_3_3_PLUS:
+                return None
+            return safe_getattr(current_version, "version_data")
         raise ValueError(f"Unsupported key: {key}`")
 
     @classmethod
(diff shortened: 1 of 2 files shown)


mobuchowski (approved):

providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
```diff
@@ -1073,18 +1080,27 @@ def team_name(cls, dagrun: DagRun) -> str | None:
...
+        # so guard with hasattr and an isinstance check.
+        if hasattr(dagrun, "_team_name"):
+            return dagrun._team_name if isinstance(dagrun._team_name, str) else None
+
         try:
-            bundle_name = cls.dag_version_info(dagrun, "bundle_name")
-            if not isinstance(bundle_name, str):
+            from sqlalchemy.orm import object_session
+
+            session = object_session(dagrun)
```
SameerMesiah97, line 1094: Doing this in fairly atypical in the provider so it's best to add a comment like this to explain why:

```
# Reuse the existing ORM session associated with the DagRun. Creating a new
# session here (via @provide_session) can trigger an unexpected commit within
# the scheduler callback, which breaks HA lock semantics.
```

providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
```diff
@@ -1073,18 +1080,27 @@ def team_name(cls, dagrun: DagRun) -> str | None:
...
+
+            if session is None:
                 return None
 
-            from airflow.models.dagbundle import DagBundleModel
+            from airflow.models.dag import DagModel
 
-            return DagBundleModel.get_team_name(bundle_name)
+            return DagModel.get_team_name(dagrun.dag_id, session=session)
```
SameerMesiah97, line 1101: I am curious about why youare using `get_team_name()` on `DagModel` rather than `DagBundleModel`? Since the ownership relationship is Team -> DagBundle -> DagModel, I'm wondering whether the bundle model is a more natural home for this lookup.

SameerMesiah97 (commented):
I agree that #70452 can be closed in favour of this. I just have 2 comments for you to address. Also, I would remind you to do a quick manual  test before merge to verify this works.

providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
kacpermuda: Added a comment there, thanks

providers/openlineage/src/airflow/providers/openlineage/utils/utils.py
kacpermuda: The only reason was to avoid getting bundle_name, this method takes dag_id rather than bundle_name, so it's one less call that we have to make and less code to maintain.

eladkal:
tests are failing

uranusjr:
Why do we need to bundle three things in one PR? All three look unrelated to me.

kacpermuda:
Yes, the team_name fix is the main part here, and two other attributes are just small additions to the ol events, just added them as I've noticed while doing the main fix that they are not included in OL events now. I can separate this into two PRs is it's problematic, but given the small size of this change I just put them here. Is that okay ?

mobuchowski:
I agree it seems small enough?

PR #70774, opened 2026-07-30
Add regression test for SERVER_TERMINATED precedence over retry

henry3260:
## Why

`ActivitySubprocess.final_state` resolves a task's final state through an ordered
chain of checks. The `SERVER_TERMINATED` check deliberately sits before the
`_should_retry` check: once the API server has told the supervisor to stop
(heartbeat returning 404 / 409 / 410), that TaskInstance is no longer the worker's
to decide on, so the worker must not independently mark it `UP_FOR_RETRY` even
when retries are configured.

That ordering is load-bearing, but nothing currently guards it. `TestSignalRetryLogic`
covers the retry and no-retry outcomes, and a separate integration test covers
`SERVER_TERMINATED`, but no test sets `_terminal_state = SERVER_TERMINATED` together
with `_should_retry = True` and a non-zero exit code — the only input combination
that distinguishes the current ordering from the reverse one. Reordering those two
branches would change behaviour silently.

## What

Adds `test_server_terminated_takes_precedence_over_retry` to `TestSignalRetryLogic`
in `task-sdk/tests/task_sdk/execution_time/test_supervisor.py`, asserting that
`final_state` returns `SERVER_TERMINATED` for a non-zero exit code with
`_should_retry = True`. Also imports `SERVER_TERMINATED` from
`airflow.sdk.execution_time.supervisor`.

---

### DIFF
--- a/task-sdk/tests/task_sdk/execution_time/test_supervisor.py
+++ b/task-sdk/tests/task_sdk/execution_time/test_supervisor.py
@@ -160,6 +160,7 @@ from airflow.sdk.execution_time.comms import (
     _ResponseFrame,
 )
 from airflow.sdk.execution_time.supervisor import (
+    SERVER_TERMINATED,
     ActivitySubprocess,
     InProcessSupervisorComms,
     InProcessTestSupervisor,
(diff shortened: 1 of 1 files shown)


henry3260:
The ‎`if` conditions here are easy to reorder accidentally, so a regression test is worth adding.

vatsrahul1001 (approved):

vatsrahul1001:
LGTM, @amoghrajesh can you take a look as well?

PR #70783, opened 2026-07-30
Make `JWTRefreshMiddleware` extensible via `BaseAuthManager.get_jwt_refresh_middleware()`

### LINKED ISSUE #70720: Make request cookies available to AuthManager methods
### Description

Add an interface for Auth Managers to access the request cookies from an auth request, in order to enable reading from the user's cookies at auth time.

### Use case/motivation

In order to separate out the JWTs for Keycloak and Airflow, we need to be able to retrieve the Keycloak JWT cookies from cookies inside the auth manager. To do this we currently need to either add the keycloak JWTs to the airflow JWT, which is sub-optimal because it creates a very large cookie, risking rejection on most browsers, or we need to modify the refresh token flow to also pass through a request object with the cookies included.

### Related issues

#70707

### Code of Conduct

stephen-bracken:
closes: #70720

Make JWTRefreshMiddleware extensible via BaseAuthManager.get_jwt_refresh_middleware()
---
Refactor the initialisation flow for the `JWTRefreshMiddleware` to give greater control to auth managers over how this middleware gets initialised.

The current behaviour is preserved by setting the `BaseAuthManager.use_jwt_middleware` attribute to `True` (default `True`).

By moving the `JWTRefreshMiddleware` initialisation to `BaseAuthManager.get_jwt_refresh_middleware()`, this allows auth managers to override the JWTRefreshMiddleware behaviour by inheriting from it and using the inherited class in get_fastapi_middlewares instead.

By factoring out the `_set_new_token()` method from `dispatch()`, this gives an interface for the inherited token refresh middleware to alter the behaviour of the middleware when setting the tokens. The interface has access to the response object to set any cookies as needed.

Changed the `_refresh_user()` method to accept a `request` object to allow accessing any cookies or state information from the request.

Also adds the `airflow.api_fastapi.app.request_cookie_is_secure()` helper to standardise setting HTTP secure cookies

- [X] No

---


airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
```diff
@@ -68,35 +69,65 @@ async def dispatch(self, request: Request, call_next):
...
+                path="/",
+                httponly=True,
+                secure=secure,
+                samesite="lax",
+            )
+        return response
+
     @staticmethod
-    async def _refresh_user(current_token: str) -> tuple[BaseUser | None, BaseUser | None]:
+    async def _refresh_user(current_token: str, request: Request) -> tuple[BaseUser | None, BaseUser | None]:
```
vincbeck: Why passing the request as parameter if you do not use it?

airflow-core/src/airflow/api_fastapi/core_api/app.py
```diff
@@ -172,12 +172,15 @@ def init_config(app: FastAPI) -> None:
...
 
-    app.add_middleware(JWTRefreshMiddleware)
+    auth_manager = get_auth_manager()
 
-    for middleware_cls, middleware_kwargs in get_auth_manager().get_fastapi_middlewares():
+    jwt_refresh_middleware, jwt_refresh_middleware_kwargs = auth_manager.get_jwt_refresh_middleware()
+
+    app.add_middleware(jwt_refresh_middleware, **jwt_refresh_middleware_kwargs)
+
+    for middleware_cls, middleware_kwargs in auth_manager.get_fastapi_middlewares():
```
vincbeck: Why having 2 kind of middlewares? Why not adding the JWT refresh middleware as part of `get_fastapi_middlewares` by default?

stephen-bracken: That was my original idea, this is to maintain backwards compatibility with `SimpleAuthManager` and `FabAuthManager`. Merging `get_jwt_refresh_middleware` into `get_fastapi_middlewares` could be done in another PR

airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
stephen-bracken: Passing the request allows sub classes to extend this method and inject information from the request cookies into the user model, see #70707 - https://github.com/apache/airflow/pull/70707/changes#diff-535b9f1fdb4aba7e6827a416e15095fa2d6c25fc62afb660bf1978f321a469ceR172

vincbeck: I see

airflow-core/src/airflow/api_fastapi/core_api/app.py
vincbeck: If we do it, we should do it in this PR. Any another PR?

stephen-bracken: sure, I'll give it a go here

vincbeck:
CI is failing

stephen-bracken:
I've updated `_refresh_user` to get the `current_token` from the request object, so now the `request` object is no longer a redundant argument.

edit: updated `_set_new_token` to generate the new jwt from the user, so the `new_user` argument is no longer redundant

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -180,7 +180,17 @@ def get_fastapi_middlewares(self) -> list[tuple[_MiddlewareFactory[Any], dict[st
         unauthenticated requests when public access is configured) should override this
         method.
         """
-        return []
+        return [self.get_jwt_refresh_middleware()]
+
+    def get_jwt_refresh_middleware(self) -> tuple[_MiddlewareFactory[Any], dict[str, Any]]:
```
vincbeck: Should be private

```suggestion
    def _get_jwt_refresh_middleware(self) -> tuple[_MiddlewareFactory[Any], dict[str, Any]]:
```

airflow-core/tests/unit/api_fastapi/auth/middlewares/test_refresh_token.py
```diff
@@ -146,15 +146,15 @@ async def test_dispatch_with_refreshed_user(
             pytest.param("http", "", False, id="http-no-local-ssl-cert"),
         ],
     )
+    @patch("airflow.api_fastapi.auth.middlewares.refresh_token.request_cookie_is_secure")
     @patch("airflow.api_fastapi.auth.middlewares.refresh_token.get_auth_manager")
     @patch("airflow.api_fastapi.auth.middlewares.refresh_token.resolve_user_from_token")
-    @patch("airflow.api_fastapi.auth.middlewares.refresh_token.conf")
     @pytest.mark.asyncio
     async def test_dispatch_cookie_secure_flag(
```
pierrejeambrun, line 156: `mock_request_cookie_is_secure` is mocking the secure flag. `"scheme", "ssl_cert"` parametrization isn't doing anything now, can be removed.

providers/fab/src/airflow/providers/fab/auth_manager/fab_auth_manager.py
```diff
@@ -373,11 +373,14 @@ def build_public_user(self, *, session: Session = NEW_SESSION) -> AnonymousUser
 
     def get_fastapi_middlewares(self) -> list[tuple[_MiddlewareFactory[Any], dict[str, Any]]]:
         """Register the FAB public-access middleware when public access is configured."""
+        middleware = super().get_fastapi_middlewares() if AIRFLOW_V_3_3_PLUS else []
```
pierrejeambrun, line 376: Condition isn't good there, 3.3.1 was just released, this will be shipped in 3.3.2 I guess.

providers/fab/tests/unit/fab/auth_manager/test_fab_auth_manager.py
```diff
@@ -328,8 +329,12 @@ def test_get_fastapi_middlewares_disabled(self, flask_app, auth_manager_with_app
         """No middleware is registered when public access is not configured."""
         previous = flask_app.config.get("AUTH_ROLE_PUBLIC")
         flask_app.config["AUTH_ROLE_PUBLIC"] = None
+        base_middleware = []
+        if AIRFLOW_V_3_4_PLUS:
```
pierrejeambrun, line 333: Why AIRFLOW_V_3_4_PLUS here?

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -180,7 +180,17 @@ def get_fastapi_middlewares(self) -> list[tuple[_MiddlewareFactory[Any], dict[st
...
-        return []
+        return [self._get_jwt_refresh_middleware()]
+
+    def _get_jwt_refresh_middleware(self) -> tuple[_MiddlewareFactory[Any], dict[str, Any]]:
+        """
+        Return the JWTRefreshMiddleware to refresh the Airflow JWT token.
+
+        :important: The JWTRefreshMiddleware should be included in get_fastapi_middlewares()
+        """
+        from airflow.api_fastapi.auth.middlewares.refresh_token import JWTRefreshMiddleware
```
pierrejeambrun, line 192: Why a local import?

airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
```diff
@@ -45,58 +50,114 @@ class JWTRefreshMiddleware(BaseHTTPMiddleware):
...
-                    secure=secure,
-                    samesite="lax",
-                    max_age=0 if new_token == "" else None,
-                )
-                # Clear any stale _token cookie at root path "/".
-                # Older Airflow instances may have set the cookie there;
-                # without this, the root-path cookie keeps being sent on
-                # every request, causing an infinite redirect loop.
-                if cookie_path != "/":
-                    response.delete_cookie(
```
pierrejeambrun, line 87: delete cookie was replaced with a call to `set_cookie`. It's not exactly the same, can we keep `delete_cookie`. (expires isn't passed down)

pierrejeambrun (commented):
I would keep the PR focused around the target issue.

That's also updating/refactoring a whole bunch, making the PR harder to review and surface to test bigger.

  1. Moves generate_jwt from before call_next to inside _set_new_token after call_next — introduces the fail-hard-vs-fail-soft semantic
  2. Reorganizes the outer try scope so the "if current_token" gate lives inside _refresh_user — makes the flow harder to read for anyone tracing "when is _refresh_user called".
  3. Changes the outer condition from if new_token is not None to if new_user or new_token is not None. Under a minimal refactor, the original condition
  still works.
  4. Introduces the if new_user: else new_token = "" dead branch inside _set_new_token — dead because dispatch never calls it with new_user=None, but present because the extraction was over-scoped.
  5. Swaps delete_cookie for set_cookie(max_age=0) — drops the expires=0 attribute. Small, but again: not needed for extensibility.
  6. @classmethod async def on _set_new_token despite using neither cls nor await.

providers/fab/src/airflow/providers/fab/auth_manager/fab_auth_manager.py
stephen-bracken: The `get_fastapi_middlewares()` method was added in 3.3.0

providers/fab/tests/unit/fab/auth_manager/test_fab_auth_manager.py
stephen-bracken: The `FabAuthManager` change is for compatibility. I am assuming that this change will be released in 3.4.0 (as that is what `airflow.__version__` is set to on main), which will add the changes to `BaseAuthManager.get_fastapi_middlewares()` and the `_get_jwt_refresh_middleware()` method - can correct the guard as necessary.

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
stephen-bracken: This method is supposed to be overrideable in sub classes to swap out the base behaviour `JWTRefreshMiddleware` for some other middleware, see https://github.com/apache/airflow/blob/90c0c8cf2e4b0e1f905b9425154fbde53e89286f/providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py#L62

providers/fab/src/airflow/providers/fab/auth_manager/fab_auth_manager.py
pierrejeambrun: My bad, read that too fast. Thanks

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
pierrejeambrun: > This method is supposed to be overrideable in sub classes to swap out the base behaviour JWTRefreshMiddleware for some other middleware, see

I don't understand how this answer the question above. Couldn't the local import be moved at the top of the files similarly to other imports?

stephen-bracken: This would work, but the `airflow.api_fastapi.auth.middlewares.refresh_token` import is circular

airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py
stephen-bracken: under the hood starlette is calling `set_cookie` with `max_age=0`, but I agree its an unnecessary change

airflow-core/tests/unit/api_fastapi/auth/middlewares/test_refresh_token.py
stephen-bracken: removed the mock, its not necessary

stephen-bracken:
@pierrejeambrun 
> 1. Moves generate_jwt from before call_next to inside _set_new_token after call_next — introduces the fail-hard-vs-fail-soft semantic

When `call_next` is called, the `request.state.user` attributes should be populated in the same way, allowing auth to take place via `get_user()`. If the user model was refreshed, the JWT cookies will be set on the response object from `call_next` like they normally would.
> 2. Reorganizes the outer try scope so the "if current_token" gate lives inside _refresh_user — makes the flow harder to read for anyone tracing "when is _refresh_user called".

Restored the `if current_token` gate
> 3. Changes the outer condition from if new_token is not None to if new_user or new_token is not None. Under a minimal refactor, the original condition
> still works.

In the new flow, `new_token` should only ever be `None` or `""`. The first condition checks whether we need to set any cookies and calls the necessary `get_cookie_path` and `request_cookie_is_secure` helpers, then clears the JWT cookie `if new_token == ""`, or passes the user model and response to `_set_new_token()` to set the cookies if it is populated.
> 4. Introduces the if new_user: else new_token = "" dead branch inside _set_new_token — dead because dispatch never calls it with new_user=None, but present because the extraction was over-scoped.

Removed dead branch
> 5. Swaps delete_cookie for set_cookie(max_age=0) — drops the expires=0 attribute. Small, but again: not needed for extensibility.

Restored `delete_cookie`
> 6. `@classmethod` async def on _set_new_token despite using neither cls nor await.

Swapped to `@staticmethod` async def to match `_refresh_user()`

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -180,7 +180,17 @@ def get_fastapi_middlewares(self) -> list[tuple[_MiddlewareFactory[Any], dict[st
...
         method.
         """
-        return []
+        return [self._get_jwt_refresh_middleware()]
+
+    def _get_jwt_refresh_middleware(self) -> tuple[_MiddlewareFactory[Any], dict[str, Any]]:
+        """
+        Return the JWTRefreshMiddleware to refresh the Airflow JWT token.
+
+        :important: The JWTRefreshMiddleware should be included in get_fastapi_middlewares()
```
pierrejeambrun: Nit:
`:important:` This isn't an rst directive.

Maybe use `.. warning::` instead, even if this is private function.

airflow-core/src/airflow/api_fastapi/app.py
```diff
@@ -60,6 +60,22 @@ def get_cookie_path() -> str:
     return API_ROOT_PATH or "/"
 
 
+def request_cookie_is_secure(request) -> bool:
```
pierrejeambrun: nit: param type?

pierrejeambrun (commented):
How does this work for old provider version (fab) and new core (removing the hardcoded JWTRefreshMiddleware to move it to `BaseAuthManager` ?
New fab calls `super().get_fastapi_middlewares()`, old fab do not so the refresh token middleware disappears?

stephen-bracken:
>How does this work for old provider version (fab) and new core (removing the hardcoded JWTRefreshMiddleware to move it to BaseAuthManager ?
New fab calls super().get_fastapi_middlewares(), old fab do not so the refresh token middleware disappears?

In 3.3.1 and prior versions of airflow the `JWTRefreshMiddleware` will be injected in the hardcoded initialisation in `airflow.api_fastapi.core_api.app.init_middlewares`:

https://github.com/apache/airflow/blob/d4581aecace0d7874fef1091cff08ca8edfe042d/airflow-core/src/airflow/api_fastapi/core_api/app.py#L178

In Airflow v3.3.0 and v3.3.1 `BaseAuthManager.get_fastapi_middlewares()` returns a `[]`, but after this change it will return `[(JWTRefreshMiddleware,{})]`.

The `super().get_fastapi_middlewares()` is there to include any middlewares we add to `BaseAuthManager` in `FabAuthManager`. This will need to be included in any auth managers that alter `get_fastapi_middlewares()` if they also use a `JWTRefreshMiddleware`. the `if AIRFLOW_V_3_3_PLUS` guard is there because `get_fastapi_middlewares()` was added in 3.3.0

pierrejeambrun (commented):
In your latest reply, you answer "old core <-> new provider", but my question was more about "old provider <-> new core".

I believe this is breaking the because core no longer hardcode `JWTRefreshMiddleware`, but old provider doesn't call `super().get_fastapi_middlewares()`.

Can you try this setup loacally and report back.

airflow-core/tests/unit/api_fastapi/auth/middlewares/test_refresh_token.py
```diff
@@ -118,10 +122,10 @@ async def test_dispatch_with_refreshed_user(
         mock_get_auth_manager.return_value = mock_auth_manager
         mock_auth_manager.refresh_user.return_value = refreshed_user
         mock_auth_manager.generate_jwt.return_value = "new_token"
-        mock_conf.get.return_value = ""
 
         call_next = AsyncMock(return_value=Response())
-        response = await middleware.dispatch(mock_request, call_next)
+        with conf_vars({("api", "base_url"): ""}):
```
pierrejeambrun: conf_vars({("api", "base_url"): ""}) at lines 127, 207, 235 is a no-op for what these tests exercise. request_cookie_is_secure reads ("api", "ssl_cert"), and API_BASE_URL in api_fastapi/app.py:49 is captured at import time so a runtime conf_vars can't shift get_cookie_path() either. Tests pass because ssl_cert defaults to "" anyway — but this looks like a leftover from the removed mock_conf.get.return_value = "" wildcard mock. Either target the right key or drop the block.

stephen-bracken: updated conf var

stephen-bracken:
@pierrejeambrun 
> I believe this is breaking the because core no longer hardcode JWTRefreshMiddleware, but old provider doesn't call super().get_fastapi_middlewares().

I have added `BaseAuthManager.use_jwt_middleware` (default True) that will trigger the old refresh behaviour where necessary. This allows auth managers to control the order in which the middleware from `get_jwt_refresh_middleware()` gets initialised without breaking old auth flows.

PR #70787, opened 2026-07-30
Fix DateTimeSensorAsync crash when target_time is a template with start_from_trigger=True

YoinkingFishy:
When `DateTimeSensorAsync` is used with `start_from_trigger=True` and `target_time` is a Jinja template string (e.g. `"{{ data_interval_end.tomorrow().replace(hour=1) }}"`), the operator crashes at DAG-parse time because `__init__` eagerly accesses `self._moment`, which calls `timezone.parse()` on the unresolved template string.

This fix detects template strings (containing `{{`) in `target_time` and disables `start_from_trigger` for that task instance, falling back to the normal `execute()` path. Template resolution happens at execution time in the worker, after which the task defers to the triggerer correctly.

- [X] Yes (specify the tool below)

Generated-by: OpenCode

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/date_time.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/date_time.py
@@ -115,10 +115,13 @@ class DateTimeSensorAsync(DateTimeSensor):
 
         self.start_from_trigger = start_from_trigger
         if self.start_from_trigger:
-            self.start_trigger_args.trigger_kwargs = dict(
-                moment=self._moment,
-                end_from_trigger=self.end_from_trigger,
-            )
+            if isinstance(self.target_time, str) and "{{" in self.target_time:
+                self.start_from_trigger = False
+            else:
+                self.start_trigger_args.trigger_kwargs = dict(
+                    moment=self._moment,
+                    end_from_trigger=self.end_from_trigger,
+                )
 
     def execute(self, context: Context) -> NoReturn:
         self.defer(


potiuk (changes requested):
Thanks for this — the crash is real and hits the class's own documented example, which uses `target_time="{{ data_interval_end.tomorrow().replace(hour=1) }}"`.

The approach is also right. `start_from_trigger` needs its trigger kwargs at scheduling time from the serialized Dag, and templates aren't rendered until the worker runs, so there's no way to resolve `_moment` early. Falling back to the normal `execute()` path is the correct answer rather than a workaround.

Two things needed before merge.

**A test.** Only the source file changed here. Constructing the sensor with a templated `target_time` and `start_from_trigger=True` should raise nothing and leave `start_from_trigger` false:

```python
def test_start_from_trigger_disabled_for_templated_target_time():
    sensor = DateTimeSensorAsync(
        task_id="ts",
        target_time="{{ data_interval_end }}",
        start_from_trigger=True,
    )
    assert sensor.start_from_trigger is False
```

That fails on `main` with the `timezone.parse()` error, so it pins exactly what you fixed. Worth a companion test asserting a non-templated `target_time` still sets `trigger_kwargs`, so the fallback can't silently widen later.

**Say something when the fallback kicks in.** Right now a user who asked for `start_from_trigger=True` silently gets a worker slot consumed instead, with nothing explaining why. An `self.log.info(...)` naming the reason would save a confusing debugging session:

```python
if isinstance(self.target_time, str) and "{{" in self.target_time:
    self.log.info(
        "target_time is templated and cannot be resolved before execution; "
        "starting from the worker instead of the triggerer."
    )
    self.start_from_trigger = False
```

One nit: the `"{{" in target_time` check misses Dags that override the Jinja delimiters via `jinja_environment_kwargs`. I don't think that's worth engineering around, but a short comment noting the assumption would help whoever hits it.

---

PR #70790, opened 2026-07-30
Do not show @timestamp in logs in UI

k-popov:
When:
- airflow3 is deployed into k8s with community helm chart
- airflow is configured to read logs from elasticsearch where fluentbit is pushing them by reading output of task pods
- configuration is the following (partial)
```
[core]
remote_logging = True
[elasticsearch]
json_format = True
write_stdout = True
```
after a task is **already finished** an additional `@timestamp` field is displayed in web UI which duplicates information in the main timestamp (at the beginning of line):
before

The patch is not including the `@timestamp` field itself but still performs mapping it into `timestamp` field is the latter is not present. With the proposed patch there is no duplicated information on timestamp, the log looks cleaner:
after

---

---

### DIFF
--- a/providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
+++ b/providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
@@ -103,12 +103,12 @@ def _format_error_detail(error_detail: Any) -> str | None:
 
 
 def _build_log_fields(hit_dict: dict[str, Any]) -> dict[str, Any]:
-    """Filter an OpenSearch hit to ``TASK_LOG_FIELDS`` and ensure compatibility with StructuredLogMessage."""
-    fields = {k: v for k, v in hit_dict.items() if k.lower() in TASK_LOG_FIELDS or k == "@timestamp"}
+    """Filter an ES hit to ``TASK_LOG_FIELDS`` and ensure compatibility with StructuredLogMessage."""
+    fields = {k: v for k, v in hit_dict.items() if k.lower() in TASK_LOG_FIELDS}
 
-    # Map @timestamp to timestamp
-    if "@timestamp" in fields and "timestamp" not in fields:
-        fields["timestamp"] = fields.pop("@timestamp")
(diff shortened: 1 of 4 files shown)


providers/opensearch/src/airflow/providers/opensearch/log/os_task_handler.py
```diff
@@ -103,12 +103,12 @@ def _format_error_detail(error_detail: Any) -> str | None:
 
 
 def _build_log_fields(hit_dict: dict[str, Any]) -> dict[str, Any]:
-    """Filter an OpenSearch hit to ``TASK_LOG_FIELDS`` and ensure compatibility with StructuredLogMessage."""
-    fields = {k: v for k, v in hit_dict.items() if k.lower() in TASK_LOG_FIELDS or k == "@timestamp"}
+    """Filter an ES hit to ``TASK_LOG_FIELDS`` and ensure compatibility with StructuredLogMessage."""
```
potiuk: This file is the OpenSearch handler, so the docstring should keep naming OpenSearch — `main` has "Filter an OpenSearch hit" here. Looks like it came across with the copy from the elasticsearch side.

```suggestion
    """Filter an OpenSearch hit to ``TASK_LOG_FIELDS`` and ensure compatibility with StructuredLogMessage."""
```

I'll push this one myself so you don't need another round-trip.

---

potiuk (approved):
Good catch, and precisely diagnosed. The old filter kept `@timestamp` in the output set and only removed it via `.pop()` when `timestamp` was absent — so whenever a hit carried both, which is the normal case once a shipper adds `@timestamp` alongside the `timestamp` the supervisor already wrote, the `pop` never ran and the field leaked into the UI.

Reading it from `hit_dict` as a fallback source instead of filtering it in is the right shape, and the `@timestamp`-only path still maps correctly since `TASK_LOG_FIELDS` contains `timestamp`. Fixing both providers symmetrically with the same both-present regression test in each is exactly right.

I pushed the docstring fix myself so this didn't need another round-trip from you — the OpenSearch handler's docstring had picked up "ES" from the elasticsearch copy.

---

PR #70793, opened 2026-07-30
Fix log template sync failing when a custom SQL Alchemy schema is configured

### LINKED ISSUE #47300: LogTemplate table not seeded in custom schema causes TypeError when triggering DAG
### Apache Airflow version

Other Airflow 2 version (please specify below)

2.10.4

### What happened?

What happened:
When using a custom schema for the Airflow metadata database, the log_template table is not automatically seeded with default data. As a result, when triggering a DAG, Airflow attempts to set the log_template_id using the following code:

run.log_template_id = int(session.scalar(select(func.max(LogTemplate.__table__.c.id))))
Because the custom schema’s log_template table is empty, the query returns None, and attempting to convert None to an integer raises a TypeError:
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
In contrast, when using the default public schema, the table is seeded with default data, and this error does not occur.

The log_template table should be seeded with default data even when using a custom schema—just as it is for the public schema. Alternatively, the code should handle an empty log_template table gracefully (for example, by defaulting to 0 when no rows exist).

### How to reproduce

Create a custom schema in your Airflow metadata database (e.g., my_custom_schema).
Update the Airflow configuration (in airflow.cfg or via the environment variable **sql_alchemy_schema = ${SQL_ALCHEMY_SCHEMA}**
to use your custom schema.

- Start Airflow.
- Run db migrate ( this will add permisssions and tables inside the schema)
- Trigger a DAG via the UI or API.
- Observe that the DAG trigger fails with the following error:
- TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
  because the log_template table in the custom schema is empty.

Image

Image

### Operating System

docker container- aws ecs

2.10.4 ( docker image)

### Deployment

Other Docker-based deployment

### Deployment details

we used the offciail docker image from airflow and on top of that we run few commands and deploy it in aws ecs service

### Anything else?

_No response_

### Code of Conduct

Aaryan123456679:
When `sql_alchemy_schema` is set, `reflect_tables()` derived the schema to reflect a mapped class's table by splitting its bare `__tablename__` on a dot, which discards the schema entirely for classes like `LogTemplate` whose table name has no dot. `synchronize_log_template()` then looked up the reflected table by that same bare name, which also does not match how SQLAlchemy keys schema-qualified tables in the reflected metadata. Together these left the `log_template` table unrecognized under a custom schema, causing triggering a Dag to fail with a `TypeError`.

This derives the schema from the mapped class's own configured table instead of parsing its name, and looks the reflected table up by its schema-qualified key.

closes: #47300

---

### DIFF
--- a/airflow-core/newsfragments/70793.bugfix.rst
+++ b/airflow-core/newsfragments/70793.bugfix.rst
@@ -1,0 +1,1 @@
+Fix log template sync failing when a custom SQL Alchemy schema is configured
--- a/airflow-core/newsfragments/70793.bugfix.rst
+++ b/airflow-core/newsfragments/70793.bugfix.rst
@@ -1,1 +1,0 @@
-Fix log template sync failing when a custom SQL Alchemy schema is configured
--- a/airflow-core/src/airflow/utils/db.py
+++ b/airflow-core/src/airflow/utils/db.py
@@ -1019,7 +1019,7 @@ def synchronize_log_template(*, session: Session = NEW_SESSION) -> None:
     from airflow.models.tasklog import LogTemplate
 
     metadata = reflect_tables([LogTemplate], session)
-    log_template_table: Table | None = metadata.tables.get(LogTemplate.__tablename__)
+    log_template_table: Table | None = metadata.tables.get(LogTemplate.__table__.key)
 
     if log_template_table is None:
         log.info("Log template table does not exist (added in 2.3.0); skipping log template sync.")
@@ -1104,11 +1104,20 @@ def reflect_tables(tables: list[MappedClassProtocol | str] | None, session, sche
     else:
         for tbl in tables:
             try:
-                table_name = tbl if isinstance(tbl, str) else tbl.__tablename__
                 tbl_schema: str | None
-                tbl_schema, sep, name = table_name.partition(".")
-                if not sep:
-                    tbl_schema, name = None, table_name
+                if isinstance(tbl, str):
+                    tbl_schema, sep, name = tbl.partition(".")
+                    if not sep:
+                        tbl_schema, name = None, tbl
+                else:
+                    # A mapped class already carries its configured schema (e.g. via
+                    # ``sql_alchemy_schema``) on its table; a bare ``__tablename__`` would
+                    # discard it and fall back to the connection's default schema.
+                    # ``__table__`` isn't declared on ``MappedClassProtocol`` (SQLAlchemy sets
+                    # it dynamically, invisible to mypy's static checks here), so read it via
+                    # ``getattr`` instead of a direct attribute access.
+                    name = tbl.__tablename__
+                    tbl_schema = getattr(tbl, "__table__").schema
                 metadata.reflect(
                     bind=bind, schema=tbl_schema, only=[name], extend_existing=True, resolve_fks=False
                 )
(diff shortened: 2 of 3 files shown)


airflow-core/newsfragments/70793.bugfix.rst
eladkal, line 1: There is no need for this file. Please remove it

Aaryan123456679: Done

airflow-core/src/airflow/utils/db.py
```diff
@@ -1104,11 +1104,20 @@ def reflect_tables(tables: list[MappedClassProtocol | str] | None, session, sche
...
+                        tbl_schema, name = None, tbl
+                else:
+                    # A mapped class already carries its configured schema (e.g. via
+                    # ``sql_alchemy_schema``) on its table; a bare ``__tablename__`` would
+                    # discard it and fall back to the connection's default schema.
+                    # ``__table__`` isn't declared on ``MappedClassProtocol`` (SQLAlchemy sets
+                    # it dynamically, invisible to mypy's static checks here), so read it via
+                    # ``getattr`` instead of a direct attribute access.
+                    name = tbl.__tablename__
+                    tbl_schema = getattr(tbl, "__table__").schema
```
Dev-iL, line 1120: Consider using the `local_table` property available via [SQLA's inspect api](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#accessing-table-and-metadata) instead of `getattr`. I am not 100% sure about this, but prrhaps it could satisfy mypy, so type safety is retained.

airflow-core/src/airflow/utils/db.py
```diff
@@ -1104,11 +1104,20 @@ def reflect_tables(tables: list[MappedClassProtocol | str] | None, session, sche
...
                 tbl_schema: str | None
-                tbl_schema, sep, name = table_name.partition(".")
-                if not sep:
-                    tbl_schema, name = None, table_name
+                if isinstance(tbl, str):
+                    tbl_schema, sep, name = tbl.partition(".")
+                    if not sep:
+                        tbl_schema, name = None, tbl
+                else:
+                    # A mapped class already carries its configured schema (e.g. via
```
Dev-iL, line 1113: I would trim the entire comment down to a few words or even remove it compeletely and let users refer to the PR for more context.

Dev-iL (commented):
Couple of suggestions.

Other than that, if AI was used in making this PR, please add a disclosure section to the PR description, per the accepted format.

airflow-core/src/airflow/utils/db.py
Aaryan123456679: Done — switched to inspect(tbl, raiseerr=True).local_table.schema. Mapper.local_table is typed as FromClause, which declares schema: Optional[str] directly, so this is fully typed — no getattr needed.

One nuance: inspect() on this Protocol-typed argument resolves mypy to an overload returning Any | None by default. Passing raiseerr=True explicitly (its existing default — no behavior change) steers mypy to the correct overload, so no cast/assert was needed either.

This also caught a bug in the test for this: it used a plain class with __table__/__tablename__ set as bare attributes, never actually SQLAlchemy-mapped, so inspect() on it raises and the test was silently asserting on a no-op. Fixed it to build the test class via a real declarative_base(metadata=...) mapping.

airflow-core/src/airflow/utils/db.py
Aaryan123456679: Done — trimmed to a single line pointing readers at this PR for context.

subhramit (approved):
Looks good

Dev-iL (approved):

PR #70795, opened 2026-07-30
Fix dynamic task mapping timeout leaving unfinished tasks marked successful

### LINKED ISSUE #37332: timeout on Dynamic Task mapping : skipped inner task , task status is still success
### Apache Airflow version

2.8.1

_No response_

### What happened?

Screenshot from 2024-02-10 22-05-57

this is confusing and not obvious to find the problem when just looking at the grid view on a complex dag

maybe we should mark the task as skipped or failed 

### How to reproduce

```python
from datetime import timedelta

from airflow.operators.python import PythonOperator
from pendulum import today
from airflow import DAG

dag = DAG(
    dag_id="test_dag",
    schedule_interval=None,
    dagrun_timeout=timedelta(seconds=5),
    start_date=today("UTC").add(days=-1)
)

with dag:
    def something(arg):
        import time
        time.sleep(int(arg))

    PythonOperator.partial(
        task_id="toto",
        python_callable=something,
    ).expand(op_args=[("10",), ("5",), ("4",)])

```

### Operating System

ubuntu 22.04

_No response_

### Deployment

Docker-Compose

### Deployment details

_No response_

### Anything else?

_No response_

### Code of Conduct

Aaryan123456679:
When a Dag run times out while a mapped task is still expanding, the scheduler marked its unfinished mapped task instances SKIPPED. Because SKIPPED does not count as a failure, the mapped task itself could still be reported as successful even though the Dag run that owned it timed out, misleading users about whether the run actually completed its work. Mark these task instances FAILED instead, so a timed-out Dag run is reflected consistently in the state of the tasks it left unfinished.

closes: #37332

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2948,7 +2948,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 default=None,
             )
             for task_instance in unfinished_task_instances:
-                task_instance.state = TaskInstanceState.SKIPPED
+                task_instance.state = TaskInstanceState.FAILED
                 session.merge(task_instance)
             session.flush()
             self.log.info("Run %s of %s has timed-out", dag_run.run_id, dag_run.dag_id)
--- a/airflow-core/tests/unit/jobs/test_scheduler_job.py
+++ b/airflow-core/tests/unit/jobs/test_scheduler_job.py
@@ -6620,7 +6620,7 @@ class TestSchedulerJob:
         run1 = session.merge(run1)
         session.refresh(run1)
         assert run1.state == State.FAILED
-        assert run1_ti.state == State.SKIPPED
+        assert run1_ti.state == State.FAILED
         session.flush()
         # Run relevant part of scheduling again to assert run2 has been scheduled
         self.job_runner._start_queued_dagruns(session)
--- a/airflow-core/newsfragments/70795.bugfix.rst
+++ b/airflow-core/newsfragments/70795.bugfix.rst
@@ -1,0 +1,1 @@
+Fix dynamic task mapping timeout leaving unfinished tasks marked successful


Aaryan123456679:
Hi Team, Friendly ping to get this PR reviewed.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2948,7 +2948,7 @@ def _schedule_dag_run(
                 default=None,
             )
             for task_instance in unfinished_task_instances:
-                task_instance.state = TaskInstanceState.SKIPPED
+                task_instance.state = TaskInstanceState.FAILED
```
subhramit: Correct me if I'm wrong, but won't this affect every DAG run timeout and not only mapped tasks?
That would cause every unfinished TI of every timed-out run become FAILED, including TIs that never started.

subhramit: This would also change teardown tasks as an unfinished teardown now would show FAILED, although it never ran.
"Mark run failed" does not change teardowns, sets running TIs to FAILED, and sets pending TIs to SKIPPED:
https://github.com/apache/airflow/blob/675d3e31675ee08e3f67fc99e2f679f7be2b2470/airflow-core/src/airflow/api/common/mark_tasks.py#L300-L318

Could we maybe use the same rule here? That would also fix the original issue, because the running mapped TIs become FAILED.

airflow-core/tests/unit/jobs/test_scheduler_job.py
```diff
@@ -6620,7 +6620,7 @@ def test_do_schedule_max_active_runs_dag_timed_out(self, dag_maker, session):
         run1 = session.merge(run1)
         session.refresh(run1)
         assert run1.state == State.FAILED
-        assert run1_ti.state == State.SKIPPED
+        assert run1_ti.state == State.FAILED
```
subhramit: This test has no mapped task. Only a single `BashOperator`.
Could we add a regression test for the issue (a mapped task with some TIs running and some pending at timeout), and one for a teardown task?

subhramit (changes requested):
Some comments.

PR #70796, opened 2026-07-30
Add `type` field sort prek hook for Helm `values.schema.json` file

Miretpl:
Adds a pre-hook which will sort all of the `type` fields in the `values.schema.json` file used for validation of Helm `values.yaml`. Part of #68596, which aims to fix `values.schema.json` validation and make it more precise.

For proper matching during the validation where we are looking at the object type in the schema, we need to have exact type matches and e.g. `["null", "string"]` and `["string", "null"]` are not matched by one rule, e.g. `["null", "string"]`. In theory, we could just add all currently existing options, but in the end I think that it will be more efficient when we have as few matching rules as possible.

As a follow-up, I will do another pre-hook which will check if all type fields defined in `values.schema.json` are defined in the matching rule for validation of the `values.schema.json` file.

---

---

### DIFF
--- a/scripts/ci/prek/sort_helm_values_schema_file_type_field.py
+++ b/scripts/ci/prek/sort_helm_values_schema_file_type_field.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


Miretpl:
As on main (Helm 2.0) we have couple of things deleted already, I will do manual backport for Helm 1.2x line after this will be merged.

Miretpl:
Strickly Helm-related so backport rather not needed.

scripts/ci/prek/sort_helm_values_schema_file_type_field.py
```diff
@@ -0,0 +1,63 @@
...
+_VALUES_SCHEMA_FILE = AIRFLOW_ROOT_PATH / "chart/values.schema.json"
+
+
+if __name__ != "__main__":
+    raise SystemExit(
+        "This file is intended to be executed as an executable program. You cannot use it as a module."
+    )
+
+
+def _sort(data: dict) -> None:
```
jedcunningham: This never recurses into `items`, so array-item type unions (e.g. command.items, service.ports.items) stay unsorted. Similar story with `additionalProperties`.

Miretpl: I've forgotten about `items`, and good catch regarding `additionalProperties`. I fixed the logic and looking at the changes in `values.schema.json` - it works correctly now.

jedcunningham (approved):

PR #70800, opened 2026-07-30
Add KeycloakJWTMiddleware to KeycloakAuthManager

stephen-bracken:
related: #70550 - adds `access_token` and `refresh_token` to user correctly

Add KeycloakJWTMiddleware to KeycloakAuthManager
---

- Add `KeycloakJWTMiddleware` to the `KeycloakAuthManager` to extract Keycloak JWT tokens from the request cookies and attach them to the user during auth requests.
- Pass Keycloak JWTs to get_user_from_token() in `KeycloakJWTMiddleware`, skip refreshing user when called in `JWTRefreshMiddleware` without the `access_token` and `refresh_token` parameters

Rename Keycloak JWT session cookies from `access_token` and `refresh_token` to `_access_token` and `_refresh_token` for consistency. Since the JWT cookies change is not yet released, this is safe to change.

Note: This change will logout any existing sessions for the `KeycloakAuthManager`, which should then be handled gracefully by issuing a new Airflow JWT token from the Keycloak login route.

- [X] No

---

### DIFF
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
@@ -27,9 +27,7 @@ CONF_REQUESTS_POOL_SIZE_KEY = "requests_pool_size"
 CONF_REQUESTS_RETRIES_KEY = "requests_retries"
 
 # Extra Cookie names
-COOKIE_NAME_ACCESS_TOKEN = "access_token"
+COOKIE_NAME_ACCESS_TOKEN = "_access_token"
 COOKIE_NAME_ID_TOKEN = "_id_token"
-COOKIE_NAME_NAME = "name"
 COOKIE_NAME_OAUTH_STATE = "_oauth_state"
-COOKIE_NAME_REFRESH_TOKEN = "refresh_token"
-COOKIE_NAME_USER_ID = "user_id"
+COOKIE_NAME_REFRESH_TOKEN = "_refresh_token"
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
@@ -1,0 +1,246 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, cast
+
+from fastapi import HTTPException, status
+from fastapi.responses import JSONResponse
+from jwt import ExpiredSignatureError, InvalidTokenError
+from starlette.middleware.base import BaseHTTPMiddleware
+
+from airflow.api_fastapi.app import get_auth_manager
+from airflow.api_fastapi.auth.managers.base_auth_manager import COOKIE_NAME_JWT_TOKEN
+from airflow.api_fastapi.core_api import security as core_api_security
+from airflow.providers.common.compat.sdk import conf
+from airflow.providers.keycloak.auth_manager.constants import (
+    COOKIE_NAME_ACCESS_TOKEN,
+    COOKIE_NAME_REFRESH_TOKEN,
+)
+from airflow.providers.keycloak.version_compat import AIRFLOW_V_3_1_8_PLUS
+
+try:
+    from airflow.api_fastapi.auth.managers.exceptions import AuthManagerRefreshTokenExpiredException
+except ImportError:
+
+    class AuthManagerRefreshTokenExpiredException(Exception):  # type: ignore[no-redef]
+        """In case it is using a version of Airflow without ``AuthManagerRefreshTokenExpiredException``."""
+
+        pass
+
+
+if AIRFLOW_V_3_1_8_PLUS:
+    from airflow.api_fastapi.app import get_cookie_path
+else:
+
+    def get_cookie_path() -> str:
+        return "/"
+
+
+if TYPE_CHECKING:
+    from fastapi import Request, Response
+
+    from airflow.providers.keycloak.auth_manager.keycloak_auth_manager import KeycloakAuthManager
+    from airflow.providers.keycloak.auth_manager.user import KeycloakAuthManagerUser
+
+
+class KeycloakJWTMiddleware(BaseHTTPMiddleware):
+    """
+    Attach the Keycloak JWT tokens to the user.
+
+    Gets the Keycloak JWT tokens from the request cookies
+    and attaches them to the user. If the token is expired,
+    attempt to refresh it using the refresh token.
+    """
+
+    async def dispatch(self, request: Request, call_next):
+        user = None
+        new_token = None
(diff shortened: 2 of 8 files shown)


vincbeck:
Why do we need a config here?

stephen-bracken:
I thought I would make the new JWT management an opt-in feature initially, I can take out the config if you don't think its necessary

vincbeck:
Yeah I think it would be better without config. That creates complexity

potiuk (commented):
Reviewed the middleware end to end. The refresh flow reads well, the version-compat guard is used consistently in both source and tests, and the thirteen tests cover the cases I would want — refresh, expiry, missing Airflow JWT, missing Keycloak token, and root-cookie clearing.

I also checked the earlier discussion about the config: that has been handled. There is no `provider.yaml` change, no `CONF_*_KEY`, and no toggle wording left, so the opt-in has genuinely gone rather than moved somewhere else.

Three things below, one of which I would like resolved before merge.

Separately, and not inline-able: the PR description is still the unedited template. For a change of this size that renames session cookies in an auth manager, the reasoning behind the rename is exactly what a reviewer needs, and this body becomes the squash commit message and the raw material for the changelog. Worth a few lines.

---

providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
```diff
@@ -27,9 +27,7 @@
 CONF_REQUESTS_RETRIES_KEY = "requests_retries"
 
 # Extra Cookie names
-COOKIE_NAME_ACCESS_TOKEN = "access_token"
+COOKIE_NAME_ACCESS_TOKEN = "_access_token"
```
potiuk, line 30: Renaming these leaves credentials behind in the browser that logout no longer clears.

`access_token` becomes `_access_token` and `refresh_token` becomes `_refresh_token`, while `COOKIE_NAME_USER_ID` and `COOKIE_NAME_NAME` are dropped entirely. Anyone with a live session is holding cookies under the old names, and after this change nothing writes or deletes them — `logout_callback` only deletes the new names now, and the `delete_cookie` calls for `user_id` and `name` are removed outright.

A refresh token is a credential, so "log out" leaving one in the browser is the wrong behaviour even for a short window. They are session cookies, with no `max_age` or `expires`, so they do go when the browser closes — that bounds the exposure but does not fix logout.

Keeping the legacy names deleted in `logout_callback` for one release would close it.

Worth noting the upgrade behaviour too, since nothing currently mentions it: an existing session sends the old cookie names, `_refresh_user` reads the new ones, gets `None`, raises 401, and `dispatch` clears the Airflow JWT. Every logged-in user is signed out by the upgrade. That is handled gracefully rather than erroring, but it should be a deliberate, documented choice rather than a side effect.

providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
```diff
@@ -0,0 +1,243 @@
...
+                            path="/",
+                            httponly=True,
+                            secure=secure,
+                            samesite="lax",
+                            max_age=0,
+                        )
+                else:
+                    response = await self._set_new_token(new_user, secure, response, cookie_path)
+
+        except HTTPException as exc:
```
potiuk, line 126: This `except` is scoped wider than the code it needs to protect — the `try` also wraps `await call_next(request)`.

If an `HTTPException` escapes a downstream handler, this middleware converts it into its own `JSONResponse`, bypassing FastAPI's registered exception handlers and changing the error shape for responses that have nothing to do with authentication.

I am not certain it is reachable: FastAPI's `ExceptionMiddleware` normally handles `HTTPException` inside `call_next`, so whether anything escapes depends on middleware ordering. So treat this as a question rather than a defect — but narrowing the `try` to just the refresh block removes the doubt entirely, and makes the intent obvious to the next reader.

providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
```diff
@@ -0,0 +1,243 @@
...
+    class AuthManagerRefreshTokenExpiredException(Exception):  # type: ignore[no-redef]
+        """In case it is using a version of Airflow without ``AuthManagerRefreshTokenExpiredException``."""
+
+        pass
+
+
+if AIRFLOW_V_3_1_8_PLUS:
+    from airflow.api_fastapi.app import get_cookie_path
+else:
+    get_cookie_path = lambda: "/"
```
potiuk: Minor: a `def` reads better than a lambda bound to a name here, and it is what ruff's E731 asks for where enabled. Same pattern appears in `routes/login.py`.

```python
def get_cookie_path() -> str:
    return "/"
```

providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
stephen-bracken: Since we haven't released a version that uses the `access_token` and `refresh_token` cookies yet, I don't think this is necessary.

providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
stephen-bracken: The `HTTPException` except pattern is used in the original `JWTRefreshMiddleware` that this is based on:
https://github.com/apache/airflow/blob/b2d1d8183dcea1813a92e2ef5a94aebac4c143e8/airflow-core/src/airflow/api_fastapi/auth/middlewares/refresh_token.py#L94

The pattern could be wrong, but I am being consistent with the existing JWT middleware here

providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
stephen-bracken: Agree it is deliberate to logout existing sessions, as any existing sessions will have the `access_token` and `refresh_token` in the Airflow JWT claims, which we want to move to the `_access_token` and `_refresh_token` JWT cookies.

potiuk: You're right, and I checked it properly: those names were never released.

At `providers-keycloak/0.8.2` the provider's `constants.py` contains only the `CONF_*` entries — no cookie names at all. `login.py` defined just `COOKIE_NAME_ID_TOKEN = "_id_token"` and `COOKIE_NAME_OAUTH_STATE = "_oauth_state"` locally (both unchanged by this PR) and imported `COOKIE_NAME_JWT_TOKEN` from core; the Keycloak `access_token`/`refresh_token` lived in the Airflow JWT claims, not in cookies. All six `COOKIE_NAME_*` constants arrived with #70550, which is not an ancestor of the 0.8.2 tag.

So there are no legacy cookies in anyone's browser and nothing for `logout_callback` to miss. Withdrawing that.

The upgrade behaviour is the part I would still like written down. A 0.8.2 session carries the tokens in the JWT claims, so after upgrading `_refresh_user` finds no `_access_token`/`_refresh_token` cookie, returns 401, and `dispatch` clears the Airflow JWT — every logged-in user is signed out once. I agree that is the right call; it just deserves a line in the keycloak changelog for whichever release ships it, so operators are not surprised by a fleet-wide logout.

---

providers/keycloak/src/airflow/providers/keycloak/auth_manager/middleware.py
potiuk: Fair — I checked core and you are right. `JWTRefreshMiddleware.dispatch` has the same shape: the `try` opens before `response = await call_next(request)` and the `except HTTPException` closes after it, so the wide scope is inherited rather than introduced here.

That settles it for this PR — being consistent with the middleware you are subclassing is the right default, and I would rather not have the two drift. If the concern is real it applies to core first, so it belongs there rather than in a copy of it. Not blocking on this.

---

providers/keycloak/src/airflow/providers/keycloak/auth_manager/constants.py
potiuk: Update: keycloak is now included in the 2026-08-01 release wave (#70932), and the forced-logout note is written into the keycloak 0.9.0 changelog there — so this is documented without any change needed on this PR. Resolving.

---

potiuk (approved):

PR #70802, opened 2026-07-31
fix(ssh): validate job dir cleanup against configured remote_base_dir

### LINKED ISSUE #69813: SSHRemoteJobOperator: inconsistent "remote_base_dir" validation

mmaxjr:
Fixes #69813

## Problem
`SSHRemoteJobOperator` lets users set a custom `remote_base_dir`, and it's
correctly validated for safety (`_validate_base_dir`) and used to build
`job_dir` via `RemoteJobPaths`. Execution and monitoring both work fine
with it.

But `_cleanup_remote_job` calls `build_posix_cleanup_command`/
`build_windows_cleanup_command`, which always validated `job_dir` against
the *hardcoded default* base directory (`_validate_job_dir`), not the one
actually configured. So cleanup for any job using a custom
`remote_base_dir` always raised:

### DIFF
--- a/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
+++ b/providers/ssh/src/airflow/providers/ssh/operators/ssh_remote_job.py
@@ -452,9 +452,9 @@ class SSHRemoteJobOperator(BaseOperator):
         """
         self.log.info("Cleaning up remote job directory: %s", job_dir)
         if remote_os == "posix":
-            cleanup_cmd = build_posix_cleanup_command(job_dir)
+            cleanup_cmd = build_posix_cleanup_command(job_dir, base_dir=self.remote_base_dir)
         else:
-            cleanup_cmd = build_windows_cleanup_command(job_dir)
+            cleanup_cmd = build_windows_cleanup_command(job_dir, base_dir=self.remote_base_dir)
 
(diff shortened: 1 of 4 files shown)


potiuk:
Thanks — this is a close call, and sorry to close it. Three PRs independently fixed this bug; I am going with #69885, open since 14 July.

Yours is good work and better in two respects: the docstring actually explains *why* a caller with a custom base dir must pass it, and you linked the originating issue #69813. #69885 has the edge on coverage though — it pins the rejection case on Windows as well as POSIX, whereas this covers the negative path for POSIX only. On a validation guard the refusals are the part most worth locking down.

Also worth knowing: #70091 has since rewritten these same command builders on `main` to use `shlex.quote`, so all three branches conflicted. I resolved that on #69885 rather than sending it back.

I would rather not lose your docstring — if you would like to carry it and the issue link over as a follow-up once #69885 lands, that would genuinely improve the result.

---

PR #70803, opened 2026-07-31
UI: Add button to find run on Grid from run header

suhaas-vaddadi:
**Button Placement and Icon**
**Run View**
image
**Task View**
image

**Effect of Pressing Button**
image

Adds a "Show in Grid View" button to the DagRun header. Clicking it filters the Grid view down to the current run. This functions even past the window of 50 runs normally visible on the grid view.

The button writes the current `dag_run_id` into the `run_id_pattern` URL search param. The Details/Grid layout already reads that same param, so the co-rendered Grid immediately filters to this run. This reuses the existing search-param filtering convention — no new API calls, just a URL-state update.

Reuses functionality added from [PR #70150](https://github.com/apache/airflow/pull/70150)

Note: `run_id_pattern` is a substring filter, not an exact match. The button sets it to the full run ID, so the target run always matches. If another run's ID happened to contain this one as a substring, both would appear.

 

---

---

### DIFF
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
@@ -72,7 +72,7 @@
       "mode": "Mode"
     },
     "expectedDuration": "Expected Duration",
-    "findOnGrid": "Find this run on the Grid",
+    "showInGrid": "Show in Grid View",
     "lastSchedulingDecision": "Last Scheduling Decision",
     "mappedPartitionKey": "Mapped Partition Key",
     "partitionDate": "Partition Date",
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
@@ -72,7 +72,6 @@
       "mode": "Mode"
     },
     "expectedDuration": "Expected Duration",
-    "showInGrid": "Show in Grid View",
     "lastSchedulingDecision": "Last Scheduling Decision",
     "mappedPartitionKey": "Mapped Partition Key",
     "partitionDate": "Partition Date",
@@ -80,6 +79,7 @@
     "queuedAt": "Queued At",
     "runAfter": "Run After",
     "runType": "Run Type",
+    "showInGrid": "Show in Grid View",
     "sourceAssetEvent": "Source Asset Event",
     "triggeredBy": "Triggered By",
     "triggeringUser": "Triggering User"
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
@@ -72,6 +72,7 @@
       "mode": "Mode"
     },
     "expectedDuration": "Expected Duration",
+    "findOnGrid": "Find this run on the Grid",
     "lastSchedulingDecision": "Last Scheduling Decision",
     "mappedPartitionKey": "Mapped Partition Key",
     "partitionDate": "Partition Date",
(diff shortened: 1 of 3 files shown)


airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
```diff
@@ -72,6 +72,7 @@
       "mode": "Mode"
     },
     "expectedDuration": "Expected Duration",
+    "findOnGrid": "Find this run on the Grid",
```
dheerajturaga: I think you can shorten the name
```suggestion
    "showInGrid": "Show in Grid View",
```

dheerajturaga (commented):
I feel the name is too long? I think we can shorten it

airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
suhaas-vaddadi: converted!

dheerajturaga:
@suhaas-vaddadi, Can you update the description with the changes made?

suhaas-vaddadi:
> @suhaas-vaddadi, Can you update the description with the changes made?

Should be now up to date!

dheerajturaga:
@suhaas-vaddadi, can you also fix the CI?

dheerajturaga (commented):
@suhaas-vaddadi, This works as expected. I feel for completeness, it would be great if this were extended to the task instances as well? 

Image

dheerajturaga (approved):
Awesome! works as expected. Would like another pair of eyes from UI experts. Good to merge if CI passes

shubhamraj-git:
Clicking "Show in Grid" in the run or task-instance header only sets the run_id_pattern filter, but the left panel keeps showing whichever view is active (Graph/Gantt), so from those views the button appears to do nothing. Maybe, have the action also switch the panel to Grid so the filtered run is actually visible after the click.

bbovenzi:
Wait if we're on a run, then the grid should paginate over to show the run. if that is broken then we should fix the route cause not add another button

dheerajturaga:
> Wait if we're on a run, then the grid should paginate over to show the run. if that is broken then we should fix the route cause not add another button

This is not happening today. If there are more than 50+ runs for example (UI limits to 50), the dag run doesn't show in the grid view

airflow-core/src/airflow/ui/src/pages/TaskInstance/Header.tsx
```diff
@@ -70,6 +87,9 @@ export const Header = ({ taskInstance }: { readonly taskInstance: TaskInstanceRe
       <HeaderCard
         actions={
           <>
+            <IconButton label={translate("dagRun.showInGrid")} onClick={showInGrid}>
+              <LuScanSearch />
+            </IconButton>
```
bbovenzi, line 92: Also, if we even do this then let's make this its own component. No need to copy past the exact logic.

bbovenzi (changes requested):
The actually functionality is "Filter this run" and remove all other runs in the grid view. In that case, we should move it inside the grid view instead of the header and make it a "Quick Filter" option.

shubhamraj-git (changes requested):
Added comment above, missed to make it part of review.

pierrejeambrun (commented):
> Wait if we're on a run, then the grid should paginate over to show the run. if that is broken then we should fix the route cause not add another button

Brent is right. 

> This is not happening today. If there are more than 50+ runs for example (UI limits to 50), the dag run doesn't show in the grid view

I think we need to fix that then. This button doesn't really make sense for me:
- Either we clicked through the grid, and we have the run/task already selected.
- Either we landed here through a link, and the grid should be focus around that run, with that run selected. This was fixed in the past I believe, we shouldn't have a grid where the selected run is actually outside the grid window. (makes it impossible to navigate history and runs after/before that specific run / task)

bbovenzi:
I am testing right now with a backfill dag. I selected a run, but more runs are being created. The selected run was originally in the first X runs, but eventually it wasnt and the pagination was automatically applied to keep my selected run in the page.

Screenshot 2026-08-21 at 3 44 26 PM

PR #52177, opened 2025-06-24
Add run_on_latest_version support for backfill and clear operations

ephraimbuddy:
With this option, users are able to choose the dag version they want
to run their dag/task after clearing or when running backfill. This
only applies to versioned bundles as non-versioned bundles run with
the latest dag version.

When the user choose the run with latest version, the bundle_version
associated with the dagrun is updated to the latest and the associated
serialized dag version updated to the latest. Choosing not to run
with latest version which is the default means that the bundle version
and serialized dag version that the dag ran with initially would be used
in running it again.

For backfill, there's now --run-on-latest-version flag that makes it run
with the latest version, otherwise it will run with the original bundle
the dagrun was created with. Note that it's only useful when rerunning
a dagrun using backfill. The default behaviour is using the initial bundle/version
and this is intentional otherwise running backfill will fail if there was
task rename in the latest version.

Summary of changes:

- Use SchedulerDagBag instead of DagBag for execution API
- Add run_on_latest_version field to DAGRunClearBody and ClearTaskInstancesBody models
- Add --run-on-latest-version CLI flag for backfill command
- Update backfill.py to support running tasks with latest DAG version
- Add UI checkbox for "Run with latest version" in clear dialogs
- Update SchedulerDagBag to handle latest version parameter
- Update API endpoints to support run_on_latest_version parameter

closes: #49007, closes: #49047

PR #70804, opened 2026-07-31
Limit deadline alerts endpoint to the latest serialized Dag row

### LINKED ISSUE #70779: UI deadlineAlerts endpoint loads all historical serialized DAG rows

Airflow-core

### Apache Airflow version

3.3.0

`GET /ui/dags/{dag_id}/deadlineAlerts` loads every historical `serialized_dag` row for the DAG.

In [`airflow.api_fastapi.core_api.routes.ui.deadlines.get_dag_deadline_alerts`](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/deadlines.py#L176-L180), the lookup is ordered descending but has no SQL `LIMIT`:

```python
serialized_dag = session.scalar(
    select(SerializedDagModel)
    .where(SerializedDagModel.dag_id == dag_id)
    .order_by(SerializedDagModel.id.desc())
)
```

For a DAG with 6,725 historical serialized DAG versions totaling about 4 GiB, PostgreSQL returned all rows and the API process exceeded its memory limit and was OOM-killed.

Add historical rows for one `dag_id`, then request:

```text
/ui/dags/<dag_id>/deadlineAlerts
```

The query should retrieve only the newest serialized DAG row.

The query should include `.limit(1)` before calling `session.scalar()`:

```python
.order_by(SerializedDagModel.id.desc())
.limit(1)
```

This returns only the latest serialized DAG and prevents memory usage from scaling with historical DAG-version count.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

imrichardwu:
The `GET /ui/dags/{dag_id}/deadlineAlerts` endpoint queries `serialized_dag` for the latest serialized DAG by ordering rows in descending order, but the query never applies a `LIMIT`. Although `session.scalar()` returns a single object, it does **not** limit the SQL query itself. The database and ORM still materialize the full result set before discarding every row except the first.

Since every `serialized_dag` row contains the full serialized DAG blob, this becomes a serious scalability issue for DAGs with long version histories. A DAG with thousands of historical versions can force the API server to load gigabytes of serialized data into memory just to return the newest entry. In one reported case, a DAG with roughly 6,725 versions caused the API process to consume about 4 GiB of memory and exceed its memory limit.

Adding `.limit(1)` ensures the database returns only the newest row. This reduces memory usage from scaling with a DAG's version history to a constant amount, while also allowing the query planner to stop after the first matching index entry instead of processing every matching row.

This also makes the implementation consistent with the existing `_get_latest_serdag` helper in `grid.py`, which already performs the same lookup with `LIMIT 1`.

closes: #70779

---

---


eladkal (approved):

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/deadlines.py
```diff
@@ -177,6 +177,7 @@ def get_dag_deadline_alerts(
         select(SerializedDagModel)
         .where(SerializedDagModel.dag_id == dag_id)
         .order_by(SerializedDagModel.id.desc())
+        .limit(1)
```
eladkal: I think limit 1 is right but is it right to always use the latest? I think there are backfil usecases where user ask to run older versions isn't it?

imrichardwu: I had the same question at first, but I think this is a separate concern.

This endpoint has always fetched the latest serialized DAG. Even before this PR, it queried serialized_dag ordered by descending ID and passed the result to session.scalar(), which returned the first row. The only thing this PR changes is adding .limit(1), which tells the database to stop after finding that first row instead of materializing the entire result set. The selected DAG version is exactly the same.

Historical-version backfills are already supported, but through a different mechanism. That behavior predates this PR (introduced in [#52177](https://github.com/apache/airflow/pull/52177)). The backfill CLI exposes --run-on-latest-version (and its inverse), and the final behavior is resolved from the run or global configuration. That's run-level version selection.

This endpoint is different. It's DAG-level and doesn't receive a DAG run or version identifier, so it has no way to know which historical serialized DAG it should use.

If we want deadline alerts to reflect the DAG definition for a specific historical run, that would require a separate, version-aware API change. I think that's a worthwhile discussion, but it's orthogonal to this PR. This change is purely a performance optimization and doesn't change the existing semantics.

ferruzzi: I think it looks safe and pretty sure that is code I actually wrote. 😅   I thought scalar() was enough, but you make a good point with the efficiency.

I would add that I have since learned that we have `SerializedDagModel.get()` which might actually replace this whole query with (untested)

```python
serialized_dag = SerializedDagModel.get(dag_id, session=session)
```

If that works, you'd have to refactor some of the tests accordingly, but it might be a nice clean-up.  Feel free to ignore that though, it's a non-blocking suggestion.

ferruzzi (approved):
Left a non-blocking suggestion in Elad's comment thread, otherwise LGTM

imrichardwu:
I updated it based on your suggestion @ferruzzi

PR #70805, opened 2026-07-31
Support DagBundle based deployment for Coordinator

### LINKED ISSUE #66334: Java-SDK artifact deployment strategies
## Why

How should user deploy their JARs?
We should know the full user-story ourself.

## What

Whether to bind DagBundle with JavaCoordinator config.

jason810496:
- closes: https://github.com/apache/airflow/issues/66334

## Why

Language-SDK subprocess coordinators (Java, Executable, Node) required a hand-managed filesystem root (`jars_root` / `executables_root` / `bundles_root`) for compiled artifacts. Users should be able to deploy those artifacts by existing DagBundle mechanism as well.

## How

- Add an optional `dag_bundle_name` kwarg on the `SubprocessCoordinator` base, giving three ways to locate artifacts:
  - explicit root set: scan that path (unchanged legacy behavior).
  - `dag_bundle_name` set: scan that configured bundle's path.
  - neither set: scan the task's own bundle, co-located with the Python stub Dag.
- Resolve roots in the base and expose them through `_get_scan_roots()` while keeping the existing `_build_execute_task_command(what)` hook signature, so a subclass only declares its explicit-root kwarg through `_explicit_artifact_roots`.
- Materialize the bundle through `airflow.sdk.execution_time.bundles.initialize_ti_bundle`, a new module split out of `task_runner` so starting a language-SDK task does not import the task runner into the supervisor.
- Validate at construction that an explicit root and `dag_bundle_name` are not both set, and that a named bundle is configured, via a new `DagBundlesManager.is_bundle_configured` that reads configured names without importing any bundle class.

## Notes

- Version semantics: a co-located bundle is pinned to the run's `bundle_info.version` because it is the same artifact the run was created from; a separately named artifact bundle resolves to the version current when the task starts. Either way the bundle is materialized at a concrete version, so `BundleVersionLock` protects the same tree the subprocess scans.

---

- [x] Yes, with help of Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/dag_processing/bundles/manager.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/manager.py
@@ -18,12 +18,13 @@ from __future__ import annotations
 
 import functools
 import importlib
+import json
 import logging
 import os
 import warnings
 from collections import defaultdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 from itsdangerous import URLSafeSerializer
 from pydantic import BaseModel, ValidationError
@@ -108,18 +109,54 @@ def _bundle_item_exc(msg):
     )
 
 
+class _BundleConfigSnapshot(NamedTuple):
+    """The configured Dag bundles, as names only and as full configs."""
+
+    configs: tuple[_ExternalBundleConfig, ...]
+    names: frozenset[str]
+
+
+_EMPTY_BUNDLE_CONFIG_SNAPSHOT = _BundleConfigSnapshot(configs=(), names=frozenset())
+
+
 @functools.cache
-def _configured_bundle_names() -> frozenset[str]:
-    """Return configured Dag bundle names, parsed by name only (no class import)."""
+def _parse_bundle_config_snapshot(config_json: str, load_examples: bool) -> _BundleConfigSnapshot:
+    """
+    Build the snapshot for one configuration, without importing any bundle class.
+
+    Keyed on the configuration rather than cached outright, so every reader of the
+    same configuration shares one snapshot while a configuration change is still
+    picked up.
+    """
+    bundle_config_list = _parse_bundle_config(json.loads(config_json))
+    if load_examples:
+        _add_example_dag_bundle(bundle_config_list)
+        _add_provider_example_dags_to_bundle(bundle_config_list)
+    return _BundleConfigSnapshot(
+        configs=tuple(bundle_config_list),
+        names=frozenset(cfg.name for cfg in bundle_config_list),
+    )
+
+
+def _load_bundle_config_snapshot() -> _BundleConfigSnapshot:
+    """
+    Read and validate the configured Dag bundles.
+
+    Shared by :meth:`DagBundlesManager.parse_config`, which goes on to import each
+    bundle class, and by :meth:`DagBundlesManager.is_bundle_configured`, which only
+    needs the names, so both resolve to the same snapshot.
+    """
     config_list = conf.getjson("dag_processor", "dag_bundle_config_list")
     if not config_list:
-        return frozenset()
+        return _EMPTY_BUNDLE_CONFIG_SNAPSHOT
     if not isinstance(config_list, list):
         raise AirflowConfigException(
             "Section `dag_processor` key `dag_bundle_config_list` "
             f"must be list but got {config_list.__class__}"
         )
-    return frozenset(cfg.name for cfg in _parse_bundle_config(config_list))
+    return _parse_bundle_config_snapshot(
+        json.dumps(config_list, sort_keys=True), conf.getboolean("core", "LOAD_EXAMPLES")
+    )
 
 
 def _parse_bundle_config(config_list) -> list[_ExternalBundleConfig]:
@@ -289,18 +326,9 @@ class DagBundlesManager(LoggingMixin):
         if self._bundle_config:
             return
 
-        config_list = conf.getjson("dag_processor", "dag_bundle_config_list")
-        if not config_list:
+        bundle_config_list = _load_bundle_config_snapshot().configs
+        if not bundle_config_list:
             return
-        if not isinstance(config_list, list):
-            raise AirflowConfigException(
-                "Section `dag_processor` key `dag_bundle_config_list` "
-                f"must be list but got {config_list.__class__}"
-            )
-        bundle_config_list = _parse_bundle_config(config_list)
-        if conf.getboolean("core", "LOAD_EXAMPLES"):
-            _add_example_dag_bundle(bundle_config_list)
-            _add_provider_example_dags_to_bundle(bundle_config_list)
 
         for bundle_config in bundle_config_list:
             if bundle_config.team_name and not conf.getboolean("core", "multi_team"):
@@ -670,8 +698,13 @@ class DagBundlesManager(LoggingMixin):
 
     @classmethod
     def is_bundle_configured(cls, name: str) -> bool:
-        """Return whether *name* is a configured Dag bundle, by config name only (no class import)."""
-        return name in _configured_bundle_names()
+        """
+        Return whether *name* is a configured Dag bundle.
+
+        Deliberately reads configured names only: a caller validating a bundle name
+        must not depend on every *other* configured bundle's class being importable.
+        """
+        return name in _load_bundle_config_snapshot().names
 
     def get_all_dag_bundles(self) -> Iterable[BaseDagBundle]:
         """
--- a/airflow-core/src/airflow/dag_processing/bundles/manager.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/manager.py
@@ -16,6 +16,7 @@
 # under the License.
 from __future__ import annotations
 
+import functools
 import importlib
 import logging
 import os
@@ -105,6 +106,20 @@ def _bundle_item_exc(msg):
     return AirflowConfigException(
         "Invalid config for section `dag_processor` key `dag_bundle_config_list`. " + msg
     )
+
+
+@functools.cache
+def _configured_bundle_names() -> frozenset[str]:
+    """Return configured Dag bundle names, parsed by name only (no class import)."""
+    config_list = conf.getjson("dag_processor", "dag_bundle_config_list")
+    if not config_list:
+        return frozenset()
+    if not isinstance(config_list, list):
+        raise AirflowConfigException(
+            "Section `dag_processor` key `dag_bundle_config_list` "
+            f"must be list but got {config_list.__class__}"
+        )
+    return frozenset(cfg.name for cfg in _parse_bundle_config(config_list))
 
 
 def _parse_bundle_config(config_list) -> list[_ExternalBundleConfig]:
@@ -655,8 +670,8 @@ class DagBundlesManager(LoggingMixin):
 
     @classmethod
     def is_bundle_configured(cls, name: str) -> bool:
-        """Return whether *name* is a configured Dag bundle, without constructing the bundle."""
-        return name in cls()._bundle_config
+        """Return whether *name* is a configured Dag bundle, by config name only (no class import)."""
+        return name in _configured_bundle_names()
 
     def get_all_dag_bundles(self) -> Iterable[BaseDagBundle]:
         """
--- a/airflow-core/src/airflow/dag_processing/bundles/manager.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/manager.py
@@ -653,6 +653,11 @@ class DagBundlesManager(LoggingMixin):
             name=name, version=version, version_data=version_data, **cfg_bundle.kwargs
         )
 
+    @classmethod
+    def is_bundle_configured(cls, name: str) -> bool:
+        """Return whether *name* is a configured Dag bundle, without constructing the bundle."""
+        return name in cls()._bundle_config
+
     def get_all_dag_bundles(self) -> Iterable[BaseDagBundle]:
         """
         Get all DAG bundles.
--- a/airflow-core/src/airflow/dag_processing/bundles/manager.py
+++ b/airflow-core/src/airflow/dag_processing/bundles/manager.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 
 import functools
 import importlib
-import json
 import logging
 import os
 import warnings
@@ -116,46 +115,24 @@ class _BundleConfigSnapshot(NamedTuple):
     names: frozenset[str]
 
 
-_EMPTY_BUNDLE_CONFIG_SNAPSHOT = _BundleConfigSnapshot(configs=(), names=frozenset())
-
-
 @functools.cache
-def _parse_bundle_config_snapshot(config_json: str, load_examples: bool) -> _BundleConfigSnapshot:
-    """
-    Build the snapshot for one configuration, without importing any bundle class.
-
-    Keyed on the configuration rather than cached outright, so every reader of the
-    same configuration shares one snapshot while a configuration change is still
-    picked up.
-    """
-    bundle_config_list = _parse_bundle_config(json.loads(config_json))
-    if load_examples:
-        _add_example_dag_bundle(bundle_config_list)
-        _add_provider_example_dags_to_bundle(bundle_config_list)
-    return _BundleConfigSnapshot(
-        configs=tuple(bundle_config_list),
-        names=frozenset(cfg.name for cfg in bundle_config_list),
-    )
-
-
 def _load_bundle_config_snapshot() -> _BundleConfigSnapshot:
-    """
-    Read and validate the configured Dag bundles.
-
-    Shared by :meth:`DagBundlesManager.parse_config`, which goes on to import each
-    bundle class, and by :meth:`DagBundlesManager.is_bundle_configured`, which only
-    needs the names, so both resolve to the same snapshot.
-    """
+    """Read and validate the configured Dag bundles, without importing their classes."""
     config_list = conf.getjson("dag_processor", "dag_bundle_config_list")
     if not config_list:
-        return _EMPTY_BUNDLE_CONFIG_SNAPSHOT
+        return _BundleConfigSnapshot(configs=(), names=frozenset())
     if not isinstance(config_list, list):
         raise AirflowConfigException(
             "Section `dag_processor` key `dag_bundle_config_list` "
             f"must be list but got {config_list.__class__}"
         )
-    return _parse_bundle_config_snapshot(
-        json.dumps(config_list, sort_keys=True), conf.getboolean("core", "LOAD_EXAMPLES")
+    bundle_config_list = _parse_bundle_config(config_list)
+    if conf.getboolean("core", "LOAD_EXAMPLES"):
+        _add_example_dag_bundle(bundle_config_list)
+        _add_provider_example_dags_to_bundle(bundle_config_list)
+    return _BundleConfigSnapshot(
+        configs=tuple(bundle_config_list),
+        names=frozenset(cfg.name for cfg in bundle_config_list),
     )
 
 
--- a/task-sdk/src/airflow/sdk/coordinators/_subprocess.py
+++ b/task-sdk/src/airflow/sdk/coordinators/_subprocess.py
@@ -37,7 +37,7 @@ import signal
 import socket
 import subprocess
 import time
-from typing import TYPE_CHECKING, TypeVar, cast
+from typing import TYPE_CHECKING, ClassVar, TypeVar, cast
 
 import attrs
 import psutil
@@ -54,6 +54,7 @@ if TYPE_CHECKING:
     from structlog.typing import FilteringBoundLogger
     from typing_extensions import Self
 
+    from airflow.dag_processing.bundles.base import BaseDagBundle  # noqa: SDK002
     from airflow.sdk.api.client import Client
     from airflow.sdk.api.datamodels._generated import BundleInfo, TaskInstance
 
@@ -408,6 +409,12 @@ class SubprocessCoordinator(BaseCoordinator):
     task_startup_timeout: float = 10.0
     dag_bundle_name: str | None = None
 
+    # Name of the subclass's explicit-root kwarg, used only in error messages.
+    # Subclasses that expose an explicit root set this and override
+    # :meth:`_explicit_artifact_roots`; a subclass that does neither lands on the
+    # task-bundle default instead of failing at execute time.
+    _root_kwarg: ClassVar[str] = "root"
+
     # Classified once at construction by :meth:`_classify_artifact_source` and
     # dispatched on by :meth:`_init_root_source` at execute time.
     _artifact_source: _ArtifactSource | None = attrs.field(init=False, default=None)
@@ -419,12 +426,24 @@ class SubprocessCoordinator(BaseCoordinator):
     # co-located artifacts.
     _active_bundle_info: BundleInfo | None = attrs.field(init=False, default=None)
 
+    @property
+    def _explicit_artifact_roots(self) -> Sequence[pathlib.Path]:
+        """Subclass's explicit artifact roots; empty (the default) selects task-bundle mode."""
+        return []
+
+    def _validate_artifact_source(self) -> None:
+        """Run subclass-specific validation after classification; default no-op."""
+
+    def __attrs_post_init__(self) -> None:
+        self._classify_artifact_source(self._explicit_artifact_roots, root_kwarg=self._root_kwarg)
+        self._validate_artifact_source()
+
     def _classify_artifact_source(self, configured: Sequence[pathlib.Path], *, root_kwarg: str) -> None:
         """
         Classify and validate how this coordinator locates artifacts (construction time).
 
-        Subclasses call this from ``__attrs_post_init__`` with their own root
-        field. It rejects setting both an explicit root and ``dag_bundle_name``,
+        Called from :meth:`__attrs_post_init__` with the subclass's explicit
+        roots. It rejects setting both an explicit root and ``dag_bundle_name``,
         fails fast when ``dag_bundle_name`` names a bundle that is not configured,
         and records the resulting :class:`_ArtifactSource` and explicit root.
         """
@@ -457,23 +476,23 @@ class SubprocessCoordinator(BaseCoordinator):
             details["configured_roots"] = [str(root) for root in self._configured_roots]
         log.debug("Coordinator artifact source selected", **details)
 
-    def _init_root_source(self) -> list[pathlib.Path]:
+    def _init_root_source(
+        self, logger: FilteringBoundLogger
+    ) -> tuple[list[pathlib.Path], BaseDagBundle | None]:
         """
-        Resolve the directories to scan for artifacts for the current task.
+        Resolve the directories to scan for artifacts, dispatched on the classified mode.
 
-        Dispatches on the :class:`_ArtifactSource` classified at construction.
-        An explicit root is returned as-is (no Dag bundle is resolved); otherwise
-        the root is a Dag bundle's materialized path — the named
-        ``dag_bundle_name`` bundle at its latest version, or the task's own
-        bundle pinned to the run's version. Called by :meth:`execute_task`, which
-        forwards the result to :meth:`_build_execute_task_command`.
+        Returns ``(roots, bundle)``: an explicit root yields no bundle (``None``);
+        a Dag-bundle mode returns the materialized path and the resolved bundle so
+        :meth:`execute_task` can hold a version lock over it. *logger* is the task
+        logger, so materialization failures surface in the task log.
         """
         if self._artifact_source is _ArtifactSource.EXPLICIT_ROOT:
-            return self._configured_roots
+            return self._configured_roots, None
+
+        from airflow.sdk.api.datamodels._generated import BundleInfo
 
         if self._artifact_source is _ArtifactSource.NAMED_BUNDLE:
-            from airflow.sdk.api.datamodels._generated import BundleInfo
-
             # NAMED_BUNDLE implies dag_bundle_name is set.
             target = BundleInfo(name=cast("str", self.dag_bundle_name))
         elif self._artifact_source is _ArtifactSource.TASK_BUNDLE:
@@ -489,11 +508,11 @@ class SubprocessCoordinator(BaseCoordinator):
         # load would risk an import cycle through the supervisor.
         from airflow.sdk.execution_time.task_runner import initialize_ti_bundle
 
-        bundle = initialize_ti_bundle(target, log)
+        bundle = initialize_ti_bundle(target, logger)
         path = bundle.path
         if not path.exists():
             raise FileNotFoundError(f"Dag bundle {target.name!r} resolved to {path}, which does not exist.")
-        return [path]
+        return [path], bundle
 
     def _build_execute_task_command(
         self, *, what: TaskInstance, roots: list[pathlib.Path]
@@ -539,8 +558,22 @@ class SubprocessCoordinator(BaseCoordinator):
         subprocess_logs_to_stdout: bool,
         **kwargs,
     ) -> BaseCoordinator.ExecutionResult:
-        with self._set_current_bundle(bundle_info):
-            roots = self._init_root_source()
+        task_logger = logger or log
+        with contextlib.ExitStack() as stack:
+            stack.enter_context(self._set_current_bundle(bundle_info))
+            roots, resolved_bundle = self._init_root_source(task_logger)
+            if resolved_bundle is not None:
+                # Hold the version lock across start()/wait() so bundle cleanup
+                # cannot rmtree a version this task is still reading from,
+                # mirroring task_runner.main() for the Python task path.
+                from airflow.dag_processing.bundles.base import BundleVersionLock  # noqa: SDK002
+
+                stack.enter_context(
+                    BundleVersionLock(
+                        bundle_name=resolved_bundle.name,
+                        bundle_version=resolved_bundle.version,
+                    )
+                )
             command, subprocess_schema_version = self._build_execute_task_command(what=what, roots=roots)
             process = _PopenActivitySubprocess.start(
                 what=what,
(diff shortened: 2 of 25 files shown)


task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -414,18 +539,20 @@ def execute_task(
...
-            logger=logger,
-            subprocess_logs_to_stdout=subprocess_logs_to_stdout,
-            sentry_integration=sentry_integration,
-            command=command,
-            subprocess_schema_version=subprocess_schema_version,
-            startup_timeout=self.task_startup_timeout,
-        )
-        exit_code = process.wait()
-        return self.ExecutionResult(exit_code, process.final_state)
+        with self._set_current_bundle(bundle_info):
```
kaxil: The Python path holds `BundleVersionLock` around the whole run (`task_runner.py:2375`), but that lock lives in `main()`, not in `parse()`, so extracting `initialize_ti_bundle` gives this path the materialization without the protection.

Two consequences. `BundleUsageTrackingManager._remove_stale_bundle` (`bundles/base.py:150-177`) takes `flock(LOCK_EX|LOCK_NB)` and `rmtree`s the version directory, backing off only on `BlockingIOError`, and Celery workers run that cleanup loop in their own subprocess (`celery_command.py:131-138`) on the same host as the JVM. In a mixed Python + language deployment a version that already has a tracking file from an earlier Python task can be deleted while a JVM is still lazily loading classes off that classpath. Separately, on a coordinator-only worker no tracking file is ever written, so `_find_all_tracking_files` returns nothing and every version cloned here is invisible to cleanup and never reclaimed.

Could `_init_root_source` return the bundle rather than just the path, and this block hold `BundleVersionLock(bundle_name=bundle.name, bundle_version=bundle.version)` across `start()` and `wait()`? Making `_set_current_bundle` an `ExitStack` that also enters the lock would mirror `task_runner.main()`.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +399,134 @@ class SubprocessCoordinator(BaseCoordinator):
...
+        forwards the result to :meth:`_build_execute_task_command`.
+        """
+        if self._artifact_source is _ArtifactSource.EXPLICIT_ROOT:
+            return self._configured_roots
+
+        if self._artifact_source is _ArtifactSource.NAMED_BUNDLE:
+            from airflow.sdk.api.datamodels._generated import BundleInfo
+
+            # NAMED_BUNDLE implies dag_bundle_name is set.
+            target = BundleInfo(name=cast("str", self.dag_bundle_name))
```
kaxil, line 517: `BundleInfo` with no version resolves to the shared mutable checkout rather than a per-version tree. `GitDagBundle.__init__` sets `repo_path = base_dir / "tracking_repo"` when version is falsy (`providers/git/.../git.py:84-87`), and `_initialize()` falls through to `refresh()`, which fetches and then does `head.reset(target, index=True, working_tree=True)`.

With `worker_concurrency > 1`, task B's startup hard-resets the tree that task A's JVM is reading its JARs from. `bundle.lock()` covers the refresh itself but not the subsequent use, and `BundleVersionLock.acquire()` returns early when version is falsy (`base.py:492`), so this mode cannot be protected by the existing mechanism at all. The same exposure applies to the task-bundle mode whenever the run's `bundle_version` is None.

Would resolving `get_current_version()` once and re-resolving with that concrete version work? That gives a `versions/<sha>` tree a lock can actually protect. It also pulls the two network fetches out of this path, which today run before `_PopenActivitySubprocess.start` and so are not covered by `task_startup_timeout`.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +399,134 @@ class SubprocessCoordinator(BaseCoordinator):
...
+        elif self._artifact_source is _ArtifactSource.TASK_BUNDLE:
+            if self._active_bundle_info is None:
+                raise RuntimeError("_init_root_source requires an active task; call it during execute_task.")
+            target = self._active_bundle_info
+        else:
+            raise RuntimeError(
```
kaxil: Classification is opt-in per subclass, so a third-party subclass that defines `__attrs_post_init__` for its own reasons and does not call `_classify_artifact_source` constructs fine and then dies here on every task.

Could the base define `__attrs_post_init__` itself, calling an overridable `_explicit_artifact_roots()` that defaults to empty, plus a `_root_kwarg` ClassVar? A subclass that overrides nothing would then land on the task-bundle mode, which is the documented default, instead of a runtime error.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +399,134 @@ class SubprocessCoordinator(BaseCoordinator):
...
         subprocess to connect to both servers, in seconds. The default is 10
         seconds.
+    :param dag_bundle_name: Locate artifacts through a configured Dag bundle rather
+        than an explicit root. Mutually exclusive with the subclass's explicit root;
+        if neither is set, the task's own bundle is used. A named bundle resolves to
+        its latest version; the task's own bundle is pinned to the run's version.
     """
 
     task_startup_timeout: float = 10.0
+    dag_bundle_name: str | None = None
```
kaxil, line 443: `dag_bundle_name` does not appear in any `.rst`, and the three config tables still list the root parameter as *(required)*: `java.rst:665`, `go.rst:427`, `typescript.rst:272`. Those become wrong on merge. The three-mode semantics live only in this docstring, so there is nowhere a user can read when to pick co-located over named-bundle over explicit root.

`contributing-docs/30_new_language_sdk.rst:107` also documents the old hook signature verbatim (`def _build_execute_task_command(self, *, what: TaskInstanceDTO) -> tuple[list[str], str]`), which this PR changes. task-sdk 1.3.0 shipped that signature, though the docs pages do carry `|experimental|` (java.rst:23), so this is more about whether you want to call out the break than a blocker.

airflow-core/src/airflow/dag_processing/bundles/manager.py
```diff
@@ -427,6 +427,11 @@ def get_bundle(
             name=name, version=version, version_data=version_data, **cfg_bundle.kwargs
         )
 
+    @classmethod
+    def is_bundle_configured(cls, name: str) -> bool:
+        """Return whether *name* is a configured Dag bundle, without constructing the bundle."""
+        return name in cls()._bundle_config
```
kaxil: `cls()` runs a full `parse_config()` on every call: `__init__` always calls it (`manager.py:228-231`) and the `if self._bundle_config: return` guard cannot fire on a fresh instance. That `import_string`s every configured bundle class, and with `[core] load_examples` on it also walks `ProvidersManager()`.

The part that bites is the error surface. `parse_config` raises `AirflowConfigException` for a duplicate or reserved name and lets `ImportError` escape, out of the coordinator's `__attrs_post_init__`. `CoordinatorManager.for_queue` catches `ImportError` and re-raises `InvalidCoordinatorError("Cannot import coordinator ...")` (`execution_time/coordinator.py:289`), pointing the operator at the wrong config key. Worth catching both here and re-raising with the bundle name so "bundle not configured" stays distinguishable from "the bundle config list does not load".

task-sdk/src/airflow/sdk/coordinators/java/coordinator.py
```diff
@@ -200,16 +201,21 @@ class JavaCoordinator(SubprocessCoordinator):
     jvm_args: list[str] = attrs.field(factory=list)
     jars_root: list[pathlib.Path] = attrs.field(
         converter=convert_roots,
-        validator=attrs.validators.min_len(1),
+        factory=list,
```
kaxil, line 204: Dropping `min_len(1)` means a deployment whose `jars_root` templates to empty or None stops failing at construction and silently switches to scanning the whole Dag bundle.

For Java that is worse than a slow scan. With `main_class` unset, `_JarInfo.find` takes the first JAR carrying a `Main-Class` in walk order and `_calculate_classpath` puts everything it found on the classpath. `java.rst:678` already warns the result is non-deterministic when multiple executable JARs are present, and co-located mode makes that the normal case. Requiring `main_class` when there is no explicit root would cover the Java side.

More broadly, is the implicit default worth it? An empty root now means both "use the task's bundle" and "my config rendered empty". Something like `use_task_bundle=True` would keep the new mode while letting a broken config keep failing fast.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +399,134 @@ class SubprocessCoordinator(BaseCoordinator):
...
+            raise RuntimeError(
+                "Coordinator artifact source was not classified; call _classify_artifact_source first."
+            )
+
+        # Lazy import: task_runner is a heavy module and importing it at module
+        # load would risk an import cycle through the supervisor.
+        from airflow.sdk.execution_time.task_runner import initialize_ti_bundle
 
-    def _build_execute_task_command(self, *, what: TaskInstance) -> tuple[list[str], str | None]:
+        bundle = initialize_ti_bundle(target, log)
```
kaxil: This passes the module logger, so bundle resolution failures land in the worker log and never in the task log. `logger` is in scope in `execute_task` and already threaded into `_PopenActivitySubprocess.start`, so `logger or log` here would fix it.

The ordering also flipped. The TI goes RUNNING in `_on_child_started` via `client.task_instances.start` (`supervisor.py:1388`), which now happens after bundle materialization, so a first-time clone of a large repo runs while the TI still shows QUEUED and counts against `[scheduler] task_queued_timeout`. The Python path is the other way round: RUNNING first, then `parse()` clones in the child where output is captured to the task log.

task-sdk/tests/task_sdk/coordinators/executable/test_coordinator.py
```diff
@@ -370,13 +370,42 @@ def test_executables_root_accepts_list(self, tmp_path):
...
+            ExecutableCoordinator(dag_bundle_name="ghost")
+
+    @patch("airflow.dag_processing.bundles.manager.DagBundlesManager")
+    def test_configured_dag_bundle_name_accepted(self, mock_manager):
+        mock_manager.is_bundle_configured.return_value = True
+        coordinator = ExecutableCoordinator(dag_bundle_name="artifacts")
+        assert coordinator.dag_bundle_name == "artifacts"
+        mock_manager.is_bundle_configured.assert_called_once_with("artifacts")
+
+    def test_build_command_scans_given_roots(self, tmp_path):
```
kaxil: This builds the coordinator with `executables_root=[tmp_path]` and then hands it `roots=[tmp_path]`, so it passes whether `_build_execute_task_command` consumes `roots` or ignores it and reads `self.executables_root`. It passes against main-shaped code too, so the comment above it asserts something the test does not check. Constructing in co-located mode (`ExecutableCoordinator()`) and passing the populated directory as `roots` would make it fail without the change. Same shape in the java and node copies.

Separately, nothing exercises the task-bundle mode through `execute_task`. `_StubSubprocessCoordinator` is pinned to the explicit-root mode and ignores `roots`, and the `TestInitRootSource` tests set `_artifact_source` directly, so the wiring that binds `bundle_info` and forwards resolved roots is not covered.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Good catch! I didn't consider the case of `BundleVersionLock` back then, thanks.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Yes, all the sub-classes only needs to override `_root_kwarg` and `_explicit_artifact_roots` now.

For example:

```python
    _root_kwarg: ClassVar[str] = "jars_root"

    @property
    def _explicit_artifact_roots(self) -> list[pathlib.Path]:
        return self.jars_root
```

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Good point. The current implementation consider this part for the Named-bundle case by `unpack_bundle_version` to get the latest version of the Named-bundle, then acquire the `BundleVersionLock`.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: I updated both the user-facing docs and the contributor-facing docs.

airflow-core/src/airflow/dag_processing/bundles/manager.py
jason810496: Actually, the `is_bundle_configured` I imaged should only check the config level (just to ensure user did setup the corresponding DagBundel without typo) without importing / initialized any DagBundle. I just refactored the method again.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: This is a bigger issue, I will follow-up on this TI semantic issue and keep this PR focus on the new "optional DagBundle based deployment" feature.

task-sdk/tests/task_sdk/coordinators/executable/test_coordinator.py
jason810496: I added additional tests to cover this scenarios, thanks.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Follow-up is up as #71075.

It moves the RUNNING transition to the top of `SubprocessCoordinator.execute_task`, ahead of `_build_execute_task_command`, so artifact resolution and the bundle materialization this PR adds are charged to the task's runtime instead of to `[scheduler] task_queued_timeout`. The pid reported there has to be the supervisor's rather than the runtime's: the server keeps whatever pid `task_instances.start` carried and 409s a heartbeat that disagrees with it, and it rejects a second `start` carrying a new pid, so there is no other way for the transition to precede the spawn.

task-sdk/src/airflow/sdk/coordinators/java/coordinator.py
jason810496: I tracked this as follow-up in https://github.com/apache/airflow/issues/71134. We need to wait until AIP-85 to solve this problem like the normal Python Dag `emits DagWarning(DUPLICATE_DAG_ID).`, there isn't a good way to solve "same DagId among multiple Jar bundles" issue now.

airflow-core/src/airflow/dag_processing/bundles/manager.py
```diff
@@ -107,6 +108,20 @@ def _bundle_item_exc(msg):
     )
 
 
+@functools.cache
```
uranusjr, line 118: Adding the cache here seems wrong since the function is called exactly only once. Do we plan to call `is_bundle_configured` more in the future? If we do, maybe we should cache that instead.

uranusjr: Also this being static means we are parsing the config list twice, once here and another when `parse_config` is called. This can be dangerous. Maybe `is_bundle_configured` should be an _instance method_ (with `self`) on the manager instead and rely on `parse_config` (or a common layer that both it and `parse_config` reuse).

task-sdk/src/airflow/sdk/execution_time/task_runner.py
```diff
@@ -1012,21 +1013,34 @@ def _register_deserialization_allowed_classes(dag, log: Logger) -> None:
                     )
 
 
-@detail_span("parse")
-def parse(what: StartupDetails, log: Logger) -> RuntimeTaskInstance:
-    # TODO: Task-SDK:
-    # Using BundleDagBag here is about 98% wrong, but it'll do for now
-    from airflow.dag_processing.dagbag import BundleDagBag
+def initialize_ti_bundle(bundle_info: BundleInfo, log: Logger) -> BaseDagBundle:
```
uranusjr: The supervisor now imports task_runner (OTel, pydantic, the operator tree) on every language-SDK task start to get five lines of bundle resolution, and the lazy import at only defers that cost rather than removing it.

Could this and `_verify_bundle_access` move to a new module? They only depend on DagBundlesManager and AirflowException, so _subprocess.py could import it at module level and drop the lazy-import comment.

uranusjr:
> Keep the subclass `_build_execute_task_command` signature unchanged: `execute_task` stashes the task's `bundle_info` for the duration of the call, and `_scan_roots` resolves the scan root on demand.

But the signature still changed, a new `root` has been added. Either this also needs to go away or made compatible, or we should just bite the bullet and pass more arguments instead.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -414,18 +554,45 @@ def execute_task(
...
+                if lock_version is None:
+                    lock_version, _ = unpack_bundle_version(
+                        resolved_bundle.get_current_version(), resolved_bundle
+                    )
+                stack.enter_context(
+                    BundleVersionLock(
+                        bundle_name=resolved_bundle.name,
+                        bundle_version=lock_version,
+                    )
+                )
```
uranusjr: `resolved_bundle` is built with `version=None` (line 493), so `bundle.path` is the version-less checkout. That's what roots at line 560 points at.

Resolving the version string doesn't move `bundle.path`, so the lock guards `versions/<sha>` while the JVM reads `tracking_repo`, so the lock is protecting a wrong directory. Same when `TASK_BUNDLE` mode gets a run with `bundle_version=None`.

It also leaves a tracking file for a version directory that was never created, which sends `_remove_stale_bundle` into rmtree on a missing path.

Would re-resolving the bundle at the concrete version work, so `bundle.path` is `versions/<sha>`, and the lock covers the tree actually passed to the subprocess?

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +400,148 @@ class SubprocessCoordinator(BaseCoordinator):
...
+
+    # Name of the subclass's explicit-root kwarg, used only in error messages.
+    # Subclasses that expose an explicit root set this and override
+    # :meth:`_explicit_artifact_roots`; a subclass that does neither lands on the
+    # task-bundle default instead of failing at execute time.
+    _root_kwarg: ClassVar[str] = "root"
+
+    # Classified once at construction by :meth:`_classify_artifact_source` and
+    # dispatched on by :meth:`_init_root_source` at execute time.
+    _artifact_source: _ArtifactSource | None = attrs.field(init=False, default=None)
```
uranusjr: This can never be None at runtime (always populated in `__attrs_post_init__`. With `init=False`, the `| None` part in the annotation and `default=None` are not needed.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +400,148 @@ class SubprocessCoordinator(BaseCoordinator):
...
+    # co-located artifacts.
+    _active_bundle_info: BundleInfo | None = attrs.field(init=False, default=None)
+
+    @property
+    def _explicit_artifact_roots(self) -> Sequence[pathlib.Path]:
+        """Subclass's explicit artifact roots; empty (the default) selects task-bundle mode."""
+        return []
+
+    def __attrs_post_init__(self) -> None:
+        self._classify_artifact_source(self._explicit_artifact_roots, root_kwarg=self._root_kwarg)
```
uranusjr: Coding style: The two arguments are probably not needed? This passes two member attributes to a private member function. The member function can read those members directly with `self` instead.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +400,148 @@ class SubprocessCoordinator(BaseCoordinator):
...
+        else:
+            source = _ArtifactSource.TASK_BUNDLE
+
+        self._artifact_source = source
+        details: dict[str, str | list[str]] = {"mode": source.name}
+        if self.dag_bundle_name is not None:
+            details["dag_bundle_name"] = self.dag_bundle_name
+        if self._configured_roots:
+            details["configured_roots"] = [str(root) for root in self._configured_roots]
+        log.debug("Coordinator artifact source selected", **details)
```
uranusjr:
```suggestion
        log.debug(
            "Coordinator artifact source selected",
            mode=source.name,
            dag_bundle_name=self.dag_bundle_name,
            configured_roots=[str(root) for root in self._configured_roots],
        )
```

This is just a debug logging message, there’s no need to make it complicated.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +400,148 @@ class SubprocessCoordinator(BaseCoordinator):
...
+    dag_bundle_name: str | None = None
+
+    # Name of the subclass's explicit-root kwarg, used only in error messages.
+    # Subclasses that expose an explicit root set this and override
+    # :meth:`_explicit_artifact_roots`; a subclass that does neither lands on the
+    # task-bundle default instead of failing at execute time.
+    _root_kwarg: ClassVar[str] = "root"
+
+    # Classified once at construction by :meth:`_classify_artifact_source` and
+    # dispatched on by :meth:`_init_root_source` at execute time.
```
uranusjr: Arguably useless comment

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +400,148 @@ class SubprocessCoordinator(BaseCoordinator):
...
+    _artifact_source: _ArtifactSource | None = attrs.field(init=False, default=None)
+    # The subclass's explicit root, recorded at construction so the base can
+    # resolve roots without knowing the subclass field name.
+    _configured_roots: list[pathlib.Path] = attrs.field(init=False, factory=list)
+    # The task's own bundle, bound for the duration of a single :meth:`execute_task`
+    # call by :meth:`_set_current_bundle` so :meth:`_init_root_source` can resolve
+    # co-located artifacts.
```
uranusjr: Same

task-sdk/src/airflow/sdk/coordinators/java/coordinator.py
```diff
@@ -200,16 +201,28 @@ class JavaCoordinator(SubprocessCoordinator):
...
+        return self.jars_root
+
+    def _build_execute_task_command(
+        self, *, what: TaskInstance, roots: list[pathlib.Path]
+    ) -> tuple[list[str], str | None]:
+        # TODO: Scanning a whole Dag bundle without an explicit main_class lets
+        # _JarInfo.find pick the first executable JAR in walk order, so duplicate
+        # entrypoints across bundles resolve non-deterministically — the same
+        # duplicate-dag_id ambiguity the Python Dag path has. Reject it at the
+        # IMPORT_ERROR stage once AIP-85 exposes an interface to raise there.
```
uranusjr: Too long and can go stale easily. One sentence with a reference to GH-71134 is enough (and more resistent to staling).

task-sdk/tests/task_sdk/coordinators/executable/test_coordinator.py
```diff
@@ -370,13 +370,43 @@ def test_executables_root_accepts_list(self, tmp_path):
...
+    @patch("airflow.dag_processing.bundles.manager.DagBundlesManager")
+    def test_configured_dag_bundle_name_accepted(self, mock_manager):
+        mock_manager.is_bundle_configured.return_value = True
+        coordinator = ExecutableCoordinator(dag_bundle_name="artifacts")
+        assert coordinator.dag_bundle_name == "artifacts"
+        mock_manager.is_bundle_configured.assert_called_once_with("artifacts")
+
+    def test_build_command_scans_passed_roots_in_colocated_mode(self, tmp_path):
+        # Co-located mode: no configured root, so the subclass must scan the roots
+        # the base hands in rather than a configured executables_root field.
```
uranusjr: This same comment appears three times in this PR. Not useful. I would add a comment somewhere in the implementation (not tests) to explain what `colocated` means, and remove all three explainations. A reader can find the comment easy enough by the `colocated` part in the test name.

airflow-core/src/airflow/dag_processing/bundles/manager.py
jason810496: Agreed, I made this a common layer that both `is_bundle_configured` and `parse_config` will get the same config snapshot.

Additionally, I prefer to keep `is_bundle_configured` as a class method so that we won't import those actual class paths during the `is_bundle_configured` call. Only the dag bundle manager constructor (`parse_config` call ) will import the Dag bundle class.

task-sdk/src/airflow/sdk/execution_time/task_runner.py
jason810496: Both `initialize_ti_bundle` and the `verify_bundle_access` are moved into dedicated modules. Additionally, the `detail_span` is moved into a dedicated `tracing` module to preventing to coupling while preserving the existing behavior (having span on the `verify_bundle_access`).

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Good point, I added the `unpack_bundle_version` to get the actual version when then the versioning is available.

The reason I didn't add in the first place is to prevent the multiple `initialize_ti_bundle` call, but the correctness is more than the efficiency. Also, I found that the second `initialize_ti_bundle` _shouldn't_ be heavy the GitDagBundle will only pull the resources if necessary.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: I made it as `_artifact_source: _ArtifactSource = attrs.field(init=False)`, thanks for the catch.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Nice catch. The `_classify_artifact_source` is now argument less and the `_explicit_artifact_roots` is the only property that subclass needs to override (no `_root_kwargs` class attribute anymore).

task-sdk/src/airflow/sdk/execution_time/tracing.py
```diff
@@ -0,0 +1,60 @@
...
+import functools
+import inspect
+
+from opentelemetry import trace
+
+from airflow.sdk._shared.observability.traces import get_task_span_detail_level
+
+tracer = trace.get_tracer(__name__)
+
+
```
jason810496, line 30: Please let me know whether extracting `detail_span` into `tracing` dedicated module changes too much or not.

Another direction is to keep the `detail_span` as-is and call both `initialize_ti_bundle` and `verify_bundle_access` sequentially within `task_runner`.

jason810496 (commented):
Thanks TP for the review, I just addressed all the comments.

task-sdk/src/airflow/sdk/execution_time/tracing.py
uranusjr: It’s a slight overkill IMO but not terrible. It works.

jason810496: Sure, then I will keep it as-is, thanks.

uranusjr:
This comment still stands https://github.com/apache/airflow/pull/70805#issuecomment-5210300261

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +432,146 @@ class SubprocessCoordinator(BaseCoordinator):
...
+        :meth:`execute_task` can hold a version lock over it. *logger* is the task
+        logger, so materialization failures surface in the task log.
+        """
+        if self._artifact_source is _ArtifactSource.EXPLICIT_ROOT:
+            return self._configured_roots, None
+
+        if self._artifact_source is _ArtifactSource.NAMED_BUNDLE:
+            # NAMED_BUNDLE implies dag_bundle_name is set.
+            target = BundleInfo(name=cast("str", self.dag_bundle_name))
+        else:
```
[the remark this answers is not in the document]
jason810496: Keeping `cast("str", self.dag_bundle_name)`. A quoted forward reference is valid in `typing.cast`, and this expression passes both Ruff and Task SDK mypy. No code change is needed.

---

task-sdk/src/airflow/sdk/execution_time/bundles.py
```diff
@@ -0,0 +1,92 @@
...
+    is actually accessible. Uses os.access() which works with any permission
+    scheme (standard Unix permissions, ACLs, SELinux, etc.).
+
+    :param bundle_instance: The bundle instance to check
+    :raises AirflowException: if bundle is not accessible
+    """
+    from getpass import getuser
+
+    from airflow.sdk.exceptions import AirflowException
+
```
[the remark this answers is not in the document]
jason810496: Moved `getuser` and `AirflowException` to module scope in `b148880a28`, preserving behavior while following the Airflow import-placement rule.

---

task-sdk/src/airflow/sdk/execution_time/tracing.py
```diff
@@ -0,0 +1,60 @@
...
+        self._args = args
+        self._kwargs = kwargs
+        self._ctx = None
+
+    def _make_ctx(self):
+        parent_span = trace.get_current_span()
+        config_level = get_task_span_detail_level(span=parent_span)
+        if config_level > 1:
+            return tracer.start_as_current_span(*self._args, **self._kwargs)
+        return trace.INVALID_SPAN
```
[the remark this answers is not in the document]
jason810496: Keeping this as-is. OpenTelemetry `Span` implements context-manager entry and exit, which `INVALID_SPAN` inherits. The targeted tracing test file passes, including both `test_level_1_no_child_span_as_context_manager` and `test_decorator_at_level_1_does_not_create_span`, so the alleged runtime failure is directly covered.

---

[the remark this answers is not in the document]
jason810496: This comment is not related in this PR, I will keep the current implementation as-is for now.

jason810496:
> This comment still stands [#70805 (comment)](https://github.com/apache/airflow/pull/70805#issuecomment-5210300261)

Sorry for oversight this comment, I just addressed in https://github.com/apache/airflow/pull/70805/commits/947b6dd4778c8f9627403b62637fb547df69145c by introducing `_get_scan_roots` method at base class so sub-class can still get the resolved root without changing the `_build_execute_task_command` method at all. Thanks.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +432,146 @@ class SubprocessCoordinator(BaseCoordinator):
...
+            if self._active_bundle_info is None:
+                raise RuntimeError("_init_root_source requires an active task; call it during execute_task.")
```
uranusjr: This guard only exists because bundle info is stored on the instance instead of being passed in. `_init_root_source` is not supposed to be overridden or manually called by a subclass, so we can simply pass the bundle info into this function as an argument instead. 

This is separate from the `_get_scan_roots` compatibility question. No external contract constrains this method's signature, so keeping the subclass hook stable doesn't require the bundle to be stashed too.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
```diff
@@ -385,23 +432,146 @@ class SubprocessCoordinator(BaseCoordinator):
...
+            self._active_bundle_info = None
+
+    @contextlib.contextmanager
+    def _set_scan_roots(self, roots: Sequence[pathlib.Path]):
+        """Expose *roots* to the command builder for the duration of the task."""
+        self._active_scan_roots = tuple(roots)
+        try:
+            yield
+        finally:
+            self._active_scan_roots = None
```
uranusjr, line 556: This silently overwrites, while `_set_current_bundle` raises "not re-entrant" on a second bind. Both are entered into the same ExitStack with the same lifetime, so they shouldn't disagree about concurrent entry.

task-sdk/src/airflow/sdk/coordinators/java/coordinator.py
```diff
@@ -200,16 +201,23 @@ class JavaCoordinator(SubprocessCoordinator):
     jvm_args: list[str] = attrs.field(factory=list)
     jars_root: list[pathlib.Path] = attrs.field(
         converter=convert_roots,
-        validator=attrs.validators.min_len(1),
+        factory=list,
```
uranusjr: With `min_len(1)` gone, a `jars_root` that templates to empty or None no longer fails at construction, which silently selects co-located mode and scans the whole dag bundle.

I don't think #71134 covers this. That issue detects ambiguity during the scan instead. Here there may be no ambiguity at all: if the bundle holds exactly one executable JAR, or `main_class` is set, its checks never fire and the wrong mode just silently succeeds. It's also on hold behind AIP-85's importer path, so a config typo would go unguarded until that lands.

This one is a constructor-time check and needs nothing from AIP-85. A sentinel default distinguishing "not passed" from "passed empty", or an explicit `use_task_bundle=True`, keeps a broken config failing fast.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Good idea, addressed in e8bf85cca6 by collapsing `_set_current_bundle` into the argument of `_init_root_source` and the `_set_scan_roots` still keeps the not re-entrant sementic.

task-sdk/src/airflow/sdk/coordinators/_subprocess.py
jason810496: Addressed in e8bf85cca6 and as the previous comment mentioned only `_set_scan_roots` guard the no re-entrant sementic.

task-sdk/src/airflow/sdk/coordinators/java/coordinator.py
jason810496: I got your point, addressed in 078b0e25d0, thanks.

jason810496 (commented):
Thanks for the review.

PR #70812, opened 2026-07-31
Add TypeScript SDK API reference docs with TypeDoc

jason810496:
## Why

The `@apache-airflow/ts-sdk` package has a public API annotated with JSDoc/TSDoc but no rendered API reference. This adds one, built with [TypeDoc](https://typedoc.org/) and the [`@clean-jsdoc-theme/typedoc`](https://www.npmjs.com/package/@clean-jsdoc-theme/typedoc) theme, and wires it into the existing `build-docs` and `publish-docs` pipelines.

The reference is built with TypeScript's own documentation tool rather than through Sphinx. That follows the precedent already set by the Java SDK.

> **Note for reviewers:** earlier revisions of this PR built the reference through [sphinx-js](https://github.com/pyodide/sphinx-js). That approach is no longer present — see [Why not sphinx-js](#why-not-sphinx-js) below for what it cost and why it was dropped.

## How

- Renders `src/index.ts` to HTML under `_build/html`; the toolchain is pinned in `package.json` + `package-lock.json` (TypeDoc 0.28, `@clean-jsdoc-theme/typedoc` 5.1). It is kept as its own npm package so docs-only dependencies stay out of the SDK's runtime and build toolchain.
- **`breeze build-docs --sdk-docs-only --sdk=typescript`** runs the toolchain in a `node:22` container (no local Node required, mirroring how the Java SDK's Dokka build runs in `eclipse-temurin`), stages the HTML to `generated/_build/docs/ts-sdk/stable/`.
- **`build-ts-sdk-docs` job** in `publish-docs-to-s3.yml` mirrors `build-java-sdk-docs`
- **The landing page is `ts-sdk/docs/index.md`.** The theme renders page bodies through MDX, where `<!--` is a syntax error, so `scripts/prepare-readme.mjs` strips the ASF header into a gitignored scratch copy before TypeDoc reads it. The committed file keeps the standard header the `insert-license` hook enforces on every Markdown file.
- Publishing is wired up too, not just building: `ts-sdk` is added to `NON_SHORT_NAME_PACKAGES` in `publish_docs_to_s3.py` (without it the S3 publisher would expand `ts-sdk` as a provider short name into a non-existent `apache-airflow-providers-ts-sdk` instead of publishing it as a top-level package), a `build-ts-sdk-docs` job now runs in `ci-amd.yml` / `ci-arm.yml` too (mirroring java-sdk's), gated on a new `run-ts-sdk-docs` selective-check output, and the release procedure is documented in `ts-sdk/README.md`.

## Why not sphinx-js

Recorded because the PR carried this approach for several revisions:

- sphinx-js 5.0.3 requires TypeDoc 0.25 (its plugin targets internals that changed in 0.28), and TypeDoc 0.25 caps TypeScript at 5.4 — several majors behind the TypeScript 6 the SDK is built with. That forced a standalone `tsconfig.json` downgrading the SDK's `es2023` target to `es2022` for the parse pass.
- sphinx-js's comment renderer only understands `text` and `code` parts and raised `Not implemented` on TSDoc `{@link}` tags, so a bundled TypeDoc plugin (`sphinxJsConfig.mjs`) had to flatten every inline tag to literal text.
- The reference itself was a hand-maintained `api.rst` listing each export with `js:autoclass` / `js:autofunction`, which had to be edited whenever the public surface changed.
- The Python dependency dragged the workspace backwards: sphinx-js requires `cattrs < 25`, pinning `uv.lock` from cattrs 26.1.0 down to 24.1.3.

---

- [x] Yes, with help of Claude Code Opus 4.8 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


ts-sdk/docs/sphinxJsConfig.mjs
```diff
@@ -0,0 +1,68 @@
...
+
+/** Rewrite a single list of CommentDisplayParts in place. */
+function flattenParts(parts) {
+  if (!parts) {
+    return;
+  }
+  for (let i = 0; i < parts.length; i++) {
+    const part = parts[i];
+    if (part.kind !== "text" && part.kind !== "code") {
+      parts[i] = { kind: "text", text: part.text ?? "" };
```
phanikumv:
```suggestion
      if (part.kind !== "text" && part.kind !== "code") {
        parts[i] = { kind: "code", text: "`" + (part.text ?? "") + "`" };
      }
  ```

Comment (phanikumv) [reply ts-sdk/docs/sphinxJsConfig.mjs]: Lets emit a markdown-style code part instead, so identifiers render as literals, skip the spellchecker, and cover it  without wordlist churn

The spellcheck failure on `getVariableOrThrow` comes from this flatten turning `{@link getVariableOrThrow}` tags into plain text that `sphinxcontrib-spelling reads as prose...

Comment (jason810496) [reply ts-sdk/docs/sphinxJsConfig.mjs]: Applied — the flatten now emits a `code` part wrapping the identifier in backticks instead of a bare `text` part. The referenced name renders as an inline literal, sphinxcontrib-spelling skips it, and there's no wordlist churn (the `TypeDoc`/`TypeScript` entries are genuine prose words in the docs, not flattened identifiers). Updated the explanatory comment to match.

---

**Comment by uranusjr** [comment]
Does TypeScript/JavaScript/Node not have its native tooling for this? I’m not thrilled with using sphinx-js (particularly with it pinning cattrs, but also that there’s one more moving block). Java SDK does not tie its API docs into Sphinx.

If we’re committed to using sphinx-js, I think this still misses some mechanisms to correctly hook into CI build and publish jobs.

**Comment by jason810496** [comment]
> Does TypeScript/JavaScript/Node not have its native tooling for this? I’m not thrilled with using sphinx-js (particularly with it pinning cattrs, but also that there’s one more moving block). Java SDK does not tie its API docs into Sphinx.
> 
> If we’re committed to using sphinx-js, I think this still misses some mechanisms to correctly hook into CI build and publish jobs.

I use sphinx-js in the first place as I prefer to able to navigate back to the airflow core site from the user point of view. However, if we considering whether it's able to bump the native TypeDoc framwork itself, then I think we should use just TypeDoc itself. (It shouldn't be difficult to add a tab on the navbar to navigate back to the core docs site).

**Review by jason810496 — COMMENTED** [review]
Thanks TP for the review, I just updated the doc framework to native TypeDoc and compared the trade-off as listed in the PR description.

There're the screenshots of the latest TS SDK specific API references:

Image

Image

**Comment by uranusjr** [comment]
Let’s wire this up for PRs. Otherwise I think this is mostly good.

**Comment by jason810496** [comment]
> Let’s wire this up for PRs. Otherwise I think this is mostly good.

Sure! I just wired them up, the fail k8s test is not related.

File: airflow-core/docs/authoring-and-scheduling/language-sdks/typescript.rst [hunk 2]
```diff
@@ -87,7 +92,7 @@ top-level ``await startCoordinator()`` make the module a runnable bundle entry p
 
 .. code-block:: typescript
 
-    import { registerTask, startCoordinator, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
+    // registerTask, startCoordinator, and TaskHandlerArgs are provided by the Airflow TypeScript SDK.
```
guan404ming: The example now references registerTask / TaskHandlerArgs without any import. The deleted line still matches ts-sdk/example/src/main.ts:20, so restoring it keeps the snippet compilable. Same in ts-sdk/docs/index.md.

guan404ming (commented):
Overall looks great. Just found some small issue:

1. TS_SDK_DOCS_FILES misses ts-sdk/tsconfig.json and ts-sdk/package.json (both feed the TypeDoc build), so changes to them skip the docs build until release publish.

2. Non-Markdown ts-sdk/docs/ files still match TS_SDK_FILES, so docs-toolchain-only changes trigger the full e2e suite.

Suggest adding those two files to `TS_SDK_DOCS_FILES` and excluding `^ts-sdk/docs/ ` from `TS_SDK_FILES`.

airflow-core/docs/authoring-and-scheduling/language-sdks/typescript.rst
jason810496: Thanks! I just applied the change to restore the actual `import { registerTask, startCoordinator, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";` line here, and the equivalent import in `ts-sdk/docs/index.md`, so both snippets are compilable and match `ts-sdk/example/src/main.ts`.

dev/breeze/src/airflow_breeze/utils/selective_checks.py
```diff
@@ -484,6 +485,12 @@ def __hash__(self):
             # `.md` excluded — doc-only edits do not affect the Gradle build.
             r"^java-sdk/(?!.*\.md$).*",
         ],
+        FileGroupForCi.TS_SDK_DOCS_FILES: [
+            # TypeDoc renders the reference from the SDK sources, and the landing page is
+            # authored in ts-sdk/docs — unlike TS_SDK_FILES, `.md` counts here.
+            r"^ts-sdk/docs/.*",
+            r"^ts-sdk/src/.*",
+        ],
```
uranusjr, line 497: This should also include `ts-sdk/tsconfig.json` and `ts-sdk/package.json`, which are referenced in the doc build.

dev/breeze/src/airflow_breeze/utils/selective_checks.py
uranusjr, line 503: Similarly, `ts-sdk/docs/package.json` and the lock file should be excluded since changing them only affects the docs.

ts-sdk/docs/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "//": "Docs-only tsconfig for the TypeDoc analysis pass. It inherits the SDK's compiler settings verbatim and roots the program at the public entry point only: TypeScript pulls in whatever that entry point transitively imports, but a glob like `../src/**/*.ts` would also root unreachable internal modules (e.g. the `cli/` bin, which needs the optional `esbuild` peer dependency) that the docs toolchain has no reason to install. `paths` below redirects `@msgpack/msgpack`, a real transitive dependency of the checked program (coordinator/frames.ts, reachable from the public startCoordinator export): Node module resolution walks up node_modules from the *importing file* (ts-sdk/src/coordinator/), never sideways into this package's own node_modules, so the bare specifier would otherwise never resolve here.",
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "paths": {
+      "@msgpack/msgpack": ["./node_modules/@msgpack/msgpack/dist.esm/index.d.ts"]
```
uranusjr:
```suggestion
      "@msgpack/msgpack": ["./node_modules/@msgpack/msgpack"]
```

Probably more resilient to package manager internals?

ts-sdk/README.md
uranusjr: This needs to match changes in `ts-sdk/src/sdk/task.ts`.

uranusjr:
Some general comments

1. We should cache npm in CI.
2. `_stage_sdk_docs` changes java-sdk staging from a merge to a destructive rmtree + copytree. Probably fine? Just confirming it’s intended.
3. `breeze build-docs --sdk-docs-only` with no `--sdk` exits 0 having built nothing. Since the help text says “Requires at least one”, maybe we can raise an error when none is provided?

dev/breeze/doc/ci/04_selective_checks.md
```diff
@@ -596,6 +596,7 @@ GitHub Actions to pass the list of parameters to a command to execute
 | run-system-tests                                        | Whether system tests should be run ("true"/"false")                                                     | true                                     |      |
 | run-task-sdk-tests                                      | Whether Task SDK tests should be run ("true"/"false")                                                   | true                                     |      |
 | run-ts-sdk-e2e-tests                                    | Whether TypeScript SDK e2e tests should be run — on `ts-sdk/`, TS e2e test, or Node coordinator changes ("true"/"false")          | true                                     |      |
+| run-ts-sdk-docs                                         | Whether the TypeScript SDK API reference should be built — on `ts-sdk/docs/` or `ts-sdk/src/` changes, including Markdown ("true"/"false")          | true                                     |      |
```
uranusjr, line 598: This should be moved up to maintain ordering.

ts-sdk/docs/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "//": "Docs-only tsconfig for the TypeDoc analysis pass. It inherits the SDK's compiler settings verbatim and roots the program at the public entry point only: TypeScript pulls in whatever that entry point transitively imports, but a glob like `../src/**/*.ts` would also root unreachable internal modules (e.g. the `cli/` bin, which needs the optional `esbuild` peer dependency) that the docs toolchain has no reason to install. `paths` below redirects `@msgpack/msgpack`, a real transitive dependency of the checked program (coordinator/frames.ts, reachable from the public startCoordinator export): Node module resolution walks up node_modules from the *importing file* (ts-sdk/src/coordinator/), never sideways into this package's own node_modules, so the bare specifier would otherwise never resolve here.",
```
uranusjr: Is wrapping possible here? This is very long.

ts-sdk/docs/package.json
uranusjr, line 1: Do we need a pre-commit hook to keep versions here and `ts-sdk/package.json` in sync?

uranusjr (approved):
This is generally fine. All comments made above can be follow-up PRs.

ts-sdk/docs/tsconfig.json
jason810496: It's kept intentionally, having `"./node_modules/@msgpack/msgpack"` is not enough to pass the build.
Without `/dist.esm/index.d.ts`, we will encounter:

```
  > ts-sdk-docs-toolchain@0.0.0 build
  > typedoc
  
  [info] Loaded plugin @clean-jsdoc-theme/typedoc
  ../src/coordinator/frames.ts:41:32 - error TS2307: Cannot find module '@msgpack/msgpack' or its corresponding type declarations.
  
  41 import { encode, decode } from "@msgpack/msgpack";
                                    ~~~~~~~~~~~~~~~~~~
  
  [error] Found 1 errors and 0 warnings
TypeDoc build failed.
```

guan404ming (approved):
Looks great! Thanks for the update.

PR #70816, opened 2026-07-31
Improve BigQuerytoSQL column lineage rendering

VladaZakharova:
BigQuerytoSQL transfers does not work correctly:

- comma-separated selected_fields strings are interpreted as individual characters,
- DBAPI connection is passed where OpenLineage expects the Airflow connection metadata.
This PR fixes connection and field parsing, then adds the transferred schema to the output dataset. Without the output schema, the column-lineage mapping appears in the event JSON, but Marquez cannot reliably create and render the output field nodes and their lineage edges. 

---

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
+++ b/providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
@@ -150,7 +150,12 @@ class BigQueryToSqlBaseOperator(BaseOperator):
         operators. Children still provide a concrete SQL hook via
         ``get_sql_hook()`` and may override behavior if needed.
         """
-        from airflow.providers.common.compat.openlineage.facet import Dataset
+        from airflow.providers.common.compat.openlineage.facet import (
+            Dataset,
+            DatasetFacet,
+            SchemaDatasetFacet,
+            SchemaDatasetFacetFields,
+        )
         from airflow.providers.google.cloud.openlineage.utils import (
             BIGQUERY_NAMESPACE,
             get_facets_from_bq_table_for_given_fields,
@@ -181,20 +186,40 @@ class BigQueryToSqlBaseOperator(BaseOperator):
 
         if self.selected_fields:
             if isinstance(self.selected_fields, str):
-                bigquery_field_names = list(self.selected_fields)
+                transferred_field_names = [
+                    field.strip() for field in self.selected_fields.split(",") if field.strip()
+                ]
             else:
-                bigquery_field_names = self.selected_fields
+                transferred_field_names = list(self.selected_fields)
         else:
-            bigquery_field_names = [f.name for f in getattr(table_obj, "schema", [])]
+            transferred_field_names = [field.name for field in getattr(table_obj, "schema", [])]
 
         input_dataset = Dataset(
             namespace=BIGQUERY_NAMESPACE,
             name=self.source_project_dataset_table,
-            facets=get_facets_from_bq_table_for_given_fields(table_obj, bigquery_field_names),
+            facets=get_facets_from_bq_table_for_given_fields(table_obj, selected_fields=None),
+        )
+        input_schema_facet = input_dataset.facets.get("schema") if input_dataset.facets else None
+        transferred_field_names_set = set(transferred_field_names)
+        lineage_input_schema_facet = (
+            SchemaDatasetFacet(
+                fields=[
+                    field
+                    for field in getattr(input_schema_facet, "fields", [])
+                    if field.name in transferred_field_names_set
+                ]
+            )
+            if input_schema_facet
+            else None
+        )
+        output_schema_facet = SchemaDatasetFacet(
+            fields=[
+                SchemaDatasetFacetFields(name=field_name, type=None) for field_name in transferred_field_names
+            ]
         )
 
         sql_hook = self.get_sql_hook()
-        db_info = sql_hook.get_openlineage_database_info(sql_hook.get_conn())
+        db_info = sql_hook.get_openlineage_database_info(sql_hook.connection)
         if db_info is None:
             self.log.debug("OpenLineage: could not get database info from SQL hook %s", type(sql_hook))
             return OperatorLineage()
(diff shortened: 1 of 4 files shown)


potiuk:
Three separate changes here, and they're at quite different levels of confidence — worth splitting the discussion.

**The `selected_fields` parsing fix is clearly right.** `list("id,name")` iterated the string character by character, so the "field names" came out as `['i', 'd', ',', 'n', ...]`. The new split-and-strip handles the documented comma-separated form, and `" id, name, "` in the test covers the whitespace and trailing-empty cases nicely.

**The `get_conn()` -> `connection` fix is right and is probably the most consequential change here — but nothing tests it.** `get_openlineage_database_info` documents its parameter as *"Airflow connection to reduce calls of `get_connection` method"*, and `PostgresHook`'s implementation reaches for `connection.extra_dejson` and `get_openlineage_authority_part(connection)`. A DBAPI connection has neither, so against a real hook this raised `AttributeError` and the facets never got emitted.

The trouble is the test can't see the difference:

```python
mock_postgres_hook.get_openlineage_database_info.return_value = db_info
```

A `MagicMock` accepts whatever it is handed, so that assertion passes identically with `get_conn()` or `connection`. Could you pin the argument, so a regression here is caught?

```python
mock_postgres_hook.get_openlineage_database_info.assert_called_once_with(mock_postgres_hook.connection)
```

**The third change needs a decision I don't think should be made in passing.** Copying the input's schema facet onto the output dataset does make Marquez render the field nodes, but `get_facets_from_bq_table_for_given_fields` builds those fields as:

```python
SchemaDatasetFacetFields(name=schema_field.name, type=schema_field.field_type, ...)
```

`field_type` is the **BigQuery** type — `STRING`, `INTEGER`, `NUMERIC`, `TIMESTAMP`. Attaching that to the output dataset advertises BigQuery types for a Postgres, MSSQL or MySQL table. The field names carry over correctly, but the types describe the wrong system, and any consumer other than Marquez reads them at face value.

So the change buys rendering at the cost of type metadata that is wrong. Emitting the fields with names only would be honest and should still give Marquez enough to build the nodes — but whether that is the right call, or whether approximating with source types is accepted practice, is an OpenLineage semantics question rather than a detail of this PR.

@mobuchowski @kacpermuda — could you weigh in on that last point? Specifically: is it acceptable for a transfer operator's output dataset to carry the source system's schema facet verbatim, types included, or should the types be omitted or mapped when the output is a different backend?

The first two changes look good to me and I'd be happy to merge them once the argument assertion is in.

---

providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
```diff
@@ -192,9 +194,12 @@ def get_openlineage_facets_on_complete(self, task_instance) -> OperatorLineage |
             name=self.source_project_dataset_table,
             facets=get_facets_from_bq_table_for_given_fields(table_obj, bigquery_field_names),
```
kacpermuda: I think this might also be a bit off currently. We're moving data from BQ to SQL, but we are only including selected fields (chosen to be moved to SQL) in the BQ table schema. We should probably be including all of BQ fields, since the table actually has them, and only include selected ones in the output schema facet.

kacpermuda:
> The third change needs a decision I don't think should be made in passing. Copying the input's schema facet onto the output dataset does make Marquez render the field nodes, but get_facets_from_bq_table_for_given_fields builds those fields as:

> SchemaDatasetFacetFields(name=schema_field.name, type=schema_field.field_type, ...)
field_type is the BigQuery type — STRING, INTEGER, NUMERIC, TIMESTAMP. Attaching that to the output dataset advertises BigQuery types for a Postgres, MSSQL or MySQL table. The field names carry over correctly, but the types describe the wrong system, and any consumer other than Marquez reads them at face value.

> So the change buys rendering at the cost of type metadata that is wrong. Emitting the fields with names only would be honest and should still give Marquez enough to build the nodes — but whether that is the right call, or whether approximating with source types is accepted practice, is an OpenLineage semantics question rather than a detail of this PR.

> @mobuchowski @kacpermuda — could you weigh in on that last point? Specifically: is it acceptable for a transfer operator's output dataset to carry the source system's schema facet verbatim, types included, or should the types be omitted or mapped when the output is a different backend?

Yes, I agree that we should probably not include BQ types in the output schema facet (or translate it to output db types, which can be hard). It's better to allow consumer to work some magic here and infer the types from the input schema, rather than provide the wrong information as some types may be unsupported in given SQL dialect.

Also left one more comment about the schema facet, we can probably fix that while we're at it.

VladaZakharova:
Thanks team, that makes sense:)
I have added assertion to verify that the Airflow connection is passed to get_openlineage_database_info().
I also updated the schema facets so the BigQuery input contains all source fields with their original types, while the SQL output contains only the transferred field names and does not include the BigQuery types.

providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
```diff
@@ -228,11 +241,24 @@ def get_openlineage_facets_on_complete(self, task_instance) -> OperatorLineage |
                 else:
                     output_name = f"{table_part}"
 
+        transferred_input_dataset = Dataset(
+            namespace=input_dataset.namespace,
+            name=input_dataset.name,
+            facets={"schema": transferred_input_schema_facet} if transferred_input_schema_facet else {},
+        )
```
kacpermuda: Sorry, why are we creating a dataset again here? We already have `input_dataset` above

providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
```diff
@@ -182,20 +185,30 @@ def get_openlineage_facets_on_complete(self, task_instance) -> OperatorLineage |
...
 
         input_dataset = Dataset(
             namespace=BIGQUERY_NAMESPACE,
             name=self.source_project_dataset_table,
-            facets=get_facets_from_bq_table_for_given_fields(table_obj, bigquery_field_names),
+            facets=get_facets_from_bq_table_for_given_fields(table_obj, selected_fields=None),
+        )
+        transferred_input_schema_facet = get_facets_from_bq_table_for_given_fields(
+            table_obj, transferred_field_names
+        ).get("schema")
```
kacpermuda: We're calling `get_facets_from_bq_table_for_given_fields` twice, can't we just read the shchema facet from input dataset here instead of second call?

VladaZakharova: Ups :D

providers/google/src/airflow/providers/google/cloud/transfers/bigquery_to_sql.py
VladaZakharova: I still need the temporary dataset for the lineage calculation. For example, if BigQuery has id, name, and value, but we only copy id and name, the lineage helper returns nothing because value is missing from the output. The temporary dataset gives the helper only the fields being copied. It isn’t emitted as another OpenLineage input.

kacpermuda (approved):

eladkal (approved):

PR #70830, opened 2026-07-31
Support custom redaction and message length cap in LLMRetryPolicy

Lee-W:
Airflow's secrets masker only masks values already registered via mask_secret(), so free-text PII in an exception message — email addresses, customer names, account numbers — still reached the external LLM provider verbatim. Users who handle that kind of data need a way to apply their own redaction rules before the prompt leaves the worker.

Nothing bounded the message size either, so a long traceback could inflate token cost without limit on every retry decision.

The redact_exception flag this replaces was merged but has not shipped in any release, so it is removed outright rather than deprecated.

 

---

---

### DIFF
--- a/providers/common/ai/docs/retry_policies.rst
+++ b/providers/common/ai/docs/retry_policies.rst
@@ -160,7 +160,7 @@ Parameters
      - 30.0
      - Max seconds to wait for the LLM response before falling back.
    * - ``redactor``
-     - None (uses ``default_redactor``)
+     - None (uses ``redact_registered_secrets``)
      - Callable ``(str) -> str`` applied to the exception's string
        representation before it is added to the classification prompt. The
        default only masks values already registered via ``mask_secret()``
@@ -190,20 +190,20 @@ email addresses, customer names, account numbers -- that were never
 registered as secrets. If your task's exception messages can contain that
 kind of data, supply your own ``redactor`` callable. It **replaces** the
 default masker rather than running in addition to it, so combine your own
-logic with :func:`~airflow.providers.common.ai.policies.retry.default_redactor`
+logic with :func:`~airflow.providers.common.ai.policies.retry.redact_registered_secrets`
 yourself if you still want known-secret masking too:
 
 .. code-block:: python
 
     import re
 
-    from airflow.providers.common.ai.policies.retry import default_redactor
+    from airflow.providers.common.ai.policies.retry import redact_registered_secrets
 
     EMAIL_RE = re.compile(r"[\w.+-]+@[\w-]+\.[\w.-]+")
 
 
     def redact_emails_and_secrets(message: str) -> str:
-        return default_redactor(EMAIL_RE.sub("<email>", message))
+        return redact_registered_secrets(EMAIL_RE.sub("<email>", message))
 
 
     llm_policy = LLMRetryPolicy(
--- a/providers/common/ai/docs/retry_policies.rst
+++ b/providers/common/ai/docs/retry_policies.rst
@@ -160,7 +160,7 @@ Parameters
      - 30.0
      - Max seconds to wait for the LLM response before falling back.
    * - ``redactor``
-     - Airflow's secrets masker
+     - None (uses ``default_redactor``)
      - Callable ``(str) -> str`` applied to the exception's string
        representation before it is added to the classification prompt. The
        default only masks values already registered via ``mask_secret()``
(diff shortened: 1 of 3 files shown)


kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
```diff
@@ -134,14 +143,19 @@ def __init__(
         fallback_rules: list[RetryRule] | None = None,
         timeout: float = 30.0,
         *,
-        redact_exception: bool = True,
+        # redact() is typed for arbitrary containers; a str in always yields a str out.
+        redactor: Callable[[str], str] | None = lambda message: cast("str", redact(message)),
```
kaxil: `None` here turns redaction off, but for the three params just above (`model_id`, `instructions`, `fallback_rules`) `None` means "use the default". Anyone assembling kwargs programmatically, say `redactor=cfg.get("redactor")`, silently ships raw exception text to the provider whenever that key is absent, and neither the signature nor the params table makes that visible. Could `None` keep the default masker, with the opt-out spelled out separately (`redactor=str` documented as the no-op, or a `redact_exception=False`-style flag)?

providers/common/ai/docs/retry_policies.rst
```diff
@@ -158,22 +159,66 @@ Parameters
...
+kind of data, supply your own ``redactor`` callable. It **replaces** the
+default masker rather than running in addition to it, so combine your own
+logic with :func:`~airflow.sdk.log.redact` yourself if you still want
+known-secret masking too:
+
+.. code-block:: python
+
+    import re
+
+    from airflow.sdk.log import redact
```
kaxil: `redact` isn't part of the Task SDK's documented public surface: `task-sdk/docs/api.rst` autodocs `airflow.sdk.log.mask_secret` but not `redact`, which is only a re-export from `airflow.sdk._shared.secrets_masker`. Its home has moved four times, which is why `common.compat.sdk` carries a fallback chain for it (`airflow.sdk.log`, `airflow.sdk._shared.secrets_masker`, `airflow.sdk.execution_time.secrets_masker`, `airflow.utils.log.secrets_masker`) and why this module imports it through the shim rather than directly. Pointing DAG authors at the raw path means their redactor breaks on the next move. Would naming the default lambda as a module-level `default_redactor` and exporting it work better, so the docs can say `from airflow.providers.common.ai.policies.retry import default_redactor`?

providers/common/ai/tests/unit/common/ai/policies/test_retry.py
```diff
@@ -154,6 +154,85 @@ def test_prompt_keeps_raw_message_when_redaction_disabled(self, mock_hook_cls):
         prompt = mock_agent.run_sync.call_args[0][0]
         assert secret_value in prompt
 
+    @pytest.mark.enable_redact
+    @patch("airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIHook", autospec=True)
+    def test_custom_redactor_replaces_masker_instead_of_stacking(self, mock_hook_cls):
+        """A custom redactor replaces the secrets masker entirely -- it is not applied on top."""
+        secret_value = "super-secret-conn-password"
+        mask_secret(secret_value)
```
kaxil, line 202: This test and `test_truncation_happens_after_redaction` call `mask_secret()` without the `reset_secrets_masker()` that the two redaction tests above use, so they inherit patterns from earlier tests and leave this secret registered for whatever runs next. Same call at the top would keep them consistent with the siblings.

kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/policies/retry.py
Lee-W: Sounds good. `None` now means "use the default masker", consistent with the params above it. Disabling redaction using `redact_exception=False`, and passing that together with an explicit `redactor` raises `ValueError` at construction time rather than letting one of the two settings quietly win.

providers/common/ai/tests/unit/common/ai/policies/test_retry.py
Lee-W: Added `reset_secrets_masker()` to both, matching the siblings above.

providers/common/ai/docs/retry_policies.rst
Lee-W: Agreed — pinning docs to that path would age badly. The default is now an exported module-level function in the policy module, and both the docstring and the docs reference that instead of `airflow.sdk.log.redact`.

I named it `redact_registered_secrets` rather than `default_redactor` so this function is still a verb instead of a noun

PR #70340, opened 2026-07-23
Validate DatabricksSQLStatementsSensor statement fields after rendering

1fanwang:
`statement` and `statement_id` are template fields, so they are rendered after `__init__` runs. The constructor enforced their mutual exclusivity and presence there, acting on the un-rendered Jinja expressions — so a templated `statement_id` couldn't resolve before the check ran. Move both checks into `execute()`. The `warehouse_id` check reads no template field and stays in `__init__`.

related: #70296

<details><summary>Testing Done</summary>

New parametrized `test_statement_combination_validated_at_execute` covers both branches (both provided / neither provided) and asserts the `AirflowException` now surfaces from `execute()`. Both fail on the pre-fix source (the constructor raises first) and pass after. `test_databricks.py` (sensors): 22 passed. `validate_operators_init.py` exits 0; the `raise AirflowException` count is unchanged.

</details>

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (please specify the tool below)

Generated-by: GitHub Copilot CLI following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70831, opened 2026-07-31
Validate DatabricksSQLStatementsSensor exclusivity at __init__

ahilashsasidharan:
## Description

This PR moves template_field exclusivity check back to `__init__` from `execute()` while keeping the `not statement and not statement_id` sibling in `execute()` since it is a required-argument check (see #70503) for the `DatabricksSQLStatementsSensor` operators in the providers/databricks/src/airflow/providers/databricks/sensors/databricks.py.

The check is updated to use `is not None` polarity rather than a truthiness check, so this is not a straight revert of the original code. This is a genuine improvement over current main: when a provided field renders to `None` under `render_template_as_native_obj=True`, the exclusivity check still fires rather than silently running the non-None argument undetected.

These are pure provision checks ("was this argument passed?") which belong in `__init__` per the updated guidance in #70296, as moving them to `execute()` risks false positives when `render_template_as_native_obj=True` renders a provided field to `None`.

**User-visible behaviour change:** `statement_id=""` alongside a `statement` (and vice versa) previously succeeded end-to-end and now raises at DAG parse time. Empty-string values are `is not None` and are correctly treated as provided. This is deliberate.

Following the logic of "you're already touching this function" argument the PR also converts 3 `AirflowException` to `ValueError` under both `__init__` and `execute()` to increase consistency and follow guidance to narrow when changing a line. 

## Tests

- Separated tests under providers/databricks/tests/unit/databricks/sensors/test_databricks.py added as part of the previous PR #70340 to validate the new behaviour.
- Added parameterised testcases to cover the specific reasons the `is not None` polarity is used in `__init__` while the `is None` polarity is not used in `execute()` for the sibling testcase.

related: #70296 and #70503 

---

Generated-by: [Antigravity IDE] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
@@ -71,7 +71,7 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
     ):
         # Handle the scenario where both statement and statement_id are set
         if statement is not None and statement_id is not None:
-            raise AirflowException("Cannot provide both statement and statement_id.")
+            raise ValueError("Cannot provide both statement and statement_id.")
 
         if not warehouse_id:
             raise AirflowException("warehouse_id must be provided.")
--- a/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
@@ -69,12 +69,11 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
         include_airflow_query_tags: bool = True,
         **kwargs,
     ):
-        # Handle the scenario where both statement and statement_id are set
         if statement is not None and statement_id is not None:
             raise ValueError("Cannot provide both statement and statement_id.")
 
         if not warehouse_id:
-            raise AirflowException("warehouse_id must be provided.")
+            raise ValueError("warehouse_id must be provided.")
 
         super().__init__(**kwargs)
 
@@ -110,9 +109,11 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
         )
 
     def execute(self, context: Context):
-        # Handle the scenario where both statement and statement_id are not set
+        # Both fields are templated, so "neither resolves to a value" is only knowable
+        # after rendering — __init__ cannot catch it. The both-provided case is a pure
+        # provision error and is checked there instead.
         if not self.statement and not self.statement_id:
-            raise AirflowException("One of either statement or statement_id must be provided.")
+            raise ValueError("One of either statement or statement_id must be provided.")
         if not self.statement_id:
             # Otherwise, we'll go ahead and "submit" the statement
             tags = build_query_tags(context, self.query_tags, self.include_airflow_query_tags)
--- a/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
@@ -69,7 +69,10 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
         include_airflow_query_tags: bool = True,
         **kwargs,
     ):
-        # Handle the scenario where either both statement and statement_id are set/not set
+        # Handle the scenario where both statement and statement_id are set
+        if statement is not None and statement_id is not None:
+            raise AirflowException("Cannot provide both statement and statement_id.")
+
         if not warehouse_id:
             raise AirflowException("warehouse_id must be provided.")
 
@@ -107,8 +110,7 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
         )
 
     def execute(self, context: Context):
-        if self.statement and self.statement_id:
-            raise AirflowException("Cannot provide both statement and statement_id.")
+        # Handle the scenario where both statement and statement_id are not set
         if not self.statement and not self.statement_id:
             raise AirflowException("One of either statement or statement_id must be provided.")
         if not self.statement_id:
--- a/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
@@ -70,7 +70,7 @@ class DatabricksSQLStatementsSensor(DatabricksSQLStatementsMixin, BaseSensorOper
         **kwargs,
     ):
         if statement is not None and statement_id is not None:
-            raise ValueError("Cannot provide both statement and statement_id.")
+            raise ValueError("Provide exactly one of statement or statement_id.")
 
         if not warehouse_id:
             raise ValueError("warehouse_id must be provided.")
--- a/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
+++ b/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
@@ -69,7 +69,7 @@ class TestDatabricksSQLStatementsSensor:
         assert op.warehouse_id == WAREHOUSE_ID
 
     def test_both_statements_included_validated_at_init(self):
-        with pytest.raises(AirflowException, match="Cannot provide both"):
+        with pytest.raises(ValueError, match="Cannot provide both"):
             DatabricksSQLStatementsSensor(
                 statement=STATEMENT,
                 statement_id=STATEMENT_ID,
--- a/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
+++ b/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
@@ -68,16 +68,18 @@ class TestDatabricksSQLStatementsSensor:
         assert op.statement_id == STATEMENT_ID
         assert op.warehouse_id == WAREHOUSE_ID
 
-    @pytest.mark.parametrize(
-        ("kwargs", "match"),
-        [
-            ({"statement": STATEMENT, "statement_id": STATEMENT_ID}, "Cannot provide both"),
-            ({}, "One of either statement or statement_id"),
-        ],
-    )
-    def test_statement_combination_validated_at_execute(self, kwargs, match):
-        op = DatabricksSQLStatementsSensor(task_id=TASK_ID, warehouse_id=WAREHOUSE_ID, **kwargs)
-        with pytest.raises(AirflowException, match=match):
+    def test_both_statements_included_validated_at_init(self):
+        with pytest.raises(AirflowException, match="Cannot provide both"):
+            DatabricksSQLStatementsSensor(
+                statement=STATEMENT,
+                statement_id=STATEMENT_ID,
+                task_id=TASK_ID,
+                warehouse_id=WAREHOUSE_ID,
+            )
+
+    def test_both_statements_missing_validated_at_execute(self):
+        op = DatabricksSQLStatementsSensor(task_id=TASK_ID, warehouse_id=WAREHOUSE_ID)
+        with pytest.raises(AirflowException, match="One of either statement or statement_id"):
             op.execute(None)
 
     @mock.patch("airflow.providers.databricks.sensors.databricks.DatabricksHook")
--- a/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
+++ b/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
@@ -68,18 +68,34 @@ class TestDatabricksSQLStatementsSensor:
         assert op.statement_id == STATEMENT_ID
         assert op.warehouse_id == WAREHOUSE_ID
 
-    def test_both_statements_included_validated_at_init(self):
+    @pytest.mark.parametrize(
+        ("statement", "statement_id"),
+        [
+            (STATEMENT, STATEMENT_ID),
+            (STATEMENT, ""),
+        ],
+    )
+    def test_both_statements_included_validated_at_init(self, statement, statement_id):
         with pytest.raises(ValueError, match="Cannot provide both"):
             DatabricksSQLStatementsSensor(
-                statement=STATEMENT,
-                statement_id=STATEMENT_ID,
+                statement=statement,
+                statement_id=statement_id,
                 task_id=TASK_ID,
                 warehouse_id=WAREHOUSE_ID,
             )
 
-    def test_both_statements_missing_validated_at_execute(self):
-        op = DatabricksSQLStatementsSensor(task_id=TASK_ID, warehouse_id=WAREHOUSE_ID)
-        with pytest.raises(AirflowException, match="One of either statement or statement_id"):
+    @pytest.mark.parametrize(
+        ("statement", "statement_id"),
+        [
+            (None, None),
+            ("", None),
+        ],
+    )
+    def test_both_statements_missing_validated_at_execute(self, statement, statement_id):
+        op = DatabricksSQLStatementsSensor(
+            task_id=TASK_ID, warehouse_id=WAREHOUSE_ID, statement=statement, statement_id=statement_id
+        )
+        with pytest.raises(ValueError, match="One of either statement or statement_id"):
             op.execute(None)
 
     @mock.patch("airflow.providers.databricks.sensors.databricks.DatabricksHook")
--- a/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
+++ b/providers/databricks/tests/unit/databricks/sensors/test_databricks.py
@@ -22,7 +22,8 @@ from unittest import mock
 import pytest
 from tenacity import stop_after_attempt, wait_incrementing
 
-from airflow.providers.common.compat.sdk import AirflowException, TaskDeferred
+from airflow.models.dag import DAG
+from airflow.providers.common.compat.sdk import AirflowException, TaskDeferred, timezone
 from airflow.providers.databricks.hooks.databricks import SQLStatementState
 from airflow.providers.databricks.sensors.databricks import DatabricksSQLStatementsSensor
 from airflow.providers.databricks.triggers.databricks import DatabricksSQLStatementExecutionTrigger
@@ -73,15 +74,49 @@ class TestDatabricksSQLStatementsSensor:
         [
             (STATEMENT, STATEMENT_ID),
             (STATEMENT, ""),
+            ("", ""),
         ],
     )
     def test_both_statements_included_validated_at_init(self, statement, statement_id):
-        with pytest.raises(ValueError, match="Cannot provide both"):
+        with pytest.raises(ValueError, match="Provide exactly one of"):
             DatabricksSQLStatementsSensor(
                 statement=statement,
                 statement_id=statement_id,
                 task_id=TASK_ID,
                 warehouse_id=WAREHOUSE_ID,
+            )
+
+    @pytest.mark.parametrize(
+        ("statement", "statement_id"),
+        [
+            ("{{ None }}", STATEMENT_ID),
+            (STATEMENT, "{{ None }}"),
+            ("{{ None }}", "{{ None }}"),
+        ],
+    )
+    def test_both_provided_with_template_renders_to_none_validated_at_init(self, statement, statement_id):
+        """
+        Both statement and statement_id are provided; at least one is a template that
+        would render to None under render_template_as_native_obj=True. The constructor
+        must raise before any rendering occurs, so the check is pinned to __init__.
+        If the exclusivity check were moved to execute(), this test would fail because
+        the constructor would succeed and if rendered before execute() the template
+        would render to None making it appear only one value was provided so the moved
+        check would not catch the exclusivity violation
+        """
+        dag = DAG(
+            dag_id="test_native_obj_dag",
+            start_date=timezone.datetime(2025, 1, 1),
+            schedule=None,
+            render_template_as_native_obj=True,
+        )
+        with pytest.raises(ValueError, match="Provide exactly one of"):
+            DatabricksSQLStatementsSensor(
+                task_id=TASK_ID,
+                warehouse_id=WAREHOUSE_ID,
+                statement=statement,
+                statement_id=statement_id,
+                dag=dag,
             )
 
     @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


ahilashsasidharan:
Question: Could the AirflowException for both checks mentioned in this PR be converted to ValueError raises as part of this PR or should they remain as is?

potiuk:
This gets the split right, which is worth saying explicitly because several PRs in this series have not.

The "both provided" case is a provision check — passing both `statement` and `statement_id` is a Dag-authoring error whatever they render to — so `__init__` with `is not None` polarity is the correct home, and that polarity also catches `statement_id=""`, which a truthiness check would let through.

Leaving the "neither resolves" case in `execute()` is right too, and I checked rather than assumed: `template_fields` includes both `statement` and `statement_id`, so `statement="{{ params.sql }}"` rendering to an empty string is only visible after rendering. `__init__` genuinely cannot catch that one. The asymmetry is correct, not an oversight.

Tests are split to match, which reads well.

One thing before merge. The relocated line still raises `AirflowException`:

```python
if statement is not None and statement_id is not None:
    raise AirflowException("Cannot provide both statement and statement_id.")
```

Moving one verbatim during a refactor is explicitly allowed, and the file's ratchet entry stays at `::4`, so nothing is violated as it stands. But the guidance is to narrow when you're already touching the line, and this is an argument-combination error, so `ValueError` fits — exactly what you did in #70634:

```python
if statement is not None and statement_id is not None:
    raise ValueError("Cannot provide both statement and statement_id.")
```

That needs the test's `pytest.raises(AirflowException, ...)` updated to `ValueError`, and `generated/known_airflow_exceptions.txt` dropped from `providers/databricks/src/airflow/providers/databricks/sensors/databricks.py::4` to `::3`.

---

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
```diff
@@ -68,16 +68,18 @@ def test_init_statement_id(self):
...
-        op = DatabricksSQLStatementsSensor(task_id=TASK_ID, warehouse_id=WAREHOUSE_ID, **kwargs)
-        with pytest.raises(AirflowException, match=match):
+    def test_both_statements_included_validated_at_init(self):
+        with pytest.raises(ValueError, match="Cannot provide both"):
+            DatabricksSQLStatementsSensor(
+                statement=STATEMENT,
+                statement_id=STATEMENT_ID,
+                task_id=TASK_ID,
+                warehouse_id=WAREHOUSE_ID,
+            )
```
shahar1: The `is not None` polarity is the headline of this PR, and nothing here exercises it. `STATEMENT` and `STATEMENT_ID` are both non-empty strings, so a plain `if statement and statement_id:` truthiness check passes this test identically — meaning the test does not fail without the change it is meant to cover, which is the bar in [Testing Standards](https://github.com/apache/airflow/blob/main/CLAUDE.md#testing-standards) ("every test must fail without the PR's change").

The distinguishing case is the empty string — exactly the one called out as the reason for the polarity in the first place. Adding it back as a parametrize keeps the split-by-lifecycle naming intact:

```suggestion
    @pytest.mark.parametrize(
        ("statement", "statement_id"),
        [
            (STATEMENT, STATEMENT_ID),
            (STATEMENT, ""),
        ],
    )
    def test_both_statements_included_validated_at_init(self, statement, statement_id):
        with pytest.raises(ValueError, match="Cannot provide both"):
            DatabricksSQLStatementsSensor(
                statement=statement,
                statement_id=statement_id,
                task_id=TASK_ID,
                warehouse_id=WAREHOUSE_ID,
            )
```

The second case fails on `main` and passes with this PR, which is the property that was missing.

providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
```diff
@@ -107,8 +110,7 @@ def _get_hook(self, caller: str) -> DatabricksHook:
         )
 
     def execute(self, context: Context):
-        if self.statement and self.statement_id:
-            raise AirflowException("Cannot provide both statement and statement_id.")
+        # Handle the scenario where both statement and statement_id are not set
```
shahar1: Minor, non-blocking. This comment and its counterpart above the `__init__` check both restate the condition on the line beneath them, which [AGENTS.md](https://github.com/apache/airflow/blob/main/CLAUDE.md#coding-standards) asks us to avoid ("do not write narrating comments that restate the next line").

The thing that genuinely warrants a comment here is the part the code cannot say for itself: *why* one half of this validation lives in `__init__` and the other half lives in `execute()`. That asymmetry is deliberate and load-bearing, and it is precisely what has been getting reverted back and forth across this series — a reader arriving cold will otherwise read it as an oversight and "fix" it again.

```suggestion
        # Both fields are templated, so "neither resolves to a value" is only knowable
        # after rendering — __init__ cannot catch it. The both-provided case is a pure
        # provision error and is checked there instead.
```

Dropping the `# Handle the scenario where both statement and statement_id are set` line above the `__init__` check would round this out — with the rationale recorded here, that one carries nothing the `if` does not already say.

shahar1 (commented):
The split is right and the follow-up commit lands everything that was asked for — `ValueError`, the updated `pytest.raises`, and the ratchet drop to `::3`. Two observations inline and one process note below; none are blockers.

I checked a few things rather than assuming them, and they hold up:

- #70340 (which moved the check into `execute()`) landed after the last provider release prep, so it is **unreleased**. Against released 7.18.0 this PR is a net no-op on *placement* — the genuinely new behaviour is the polarity and the exception type, which matches how the description frames it.
- Pre-#70340 the `__init__` check was `if statement and statement_id` raising `AirflowException`. So this really isn't a straight revert, as stated.
- `DatabricksSQLStatementsOperator` takes `statement: str` as required and does not list it in `template_fields`, so there is no parallel site needing the same treatment. Nothing sibling is left half-fixed.

### Process — the PR title ships into the changelog verbatim

The title currently ends in a literal ellipsis:

> Rewrite DatabricksSQLStatementsSensor exclusivity check and move back…

Databricks changelogs are regenerated from `git log` by the release manager rather than from newsfragments (see [`providers/AGENTS.md`](https://github.com/apache/airflow/blob/main/providers/AGENTS.md)), so the truncation would land as-is in `providers/databricks/docs/changelog.rst`. Worth retitling to something that stands on its own and stays under 70 characters — for example:

> Validate DatabricksSQLStatementsSensor exclusivity at `__init__`

Please retitle the commit as well as the PR, since the commit message is what the changelog is generated from.

### On your question about the remaining `AirflowException`s

> Could the AirflowException for both checks mentioned in this PR be converted to ValueError raises as part of this PR or should they remain as is?

Both readings are defensible, and the ratchet permits either. Neither `warehouse_id must be provided.` nor `One of either statement or statement_id must be provided.` is modified by this diff, so leaving them is not a violation — the `check-no-new-airflow-exceptions` hook only guards *new* usages.

That said, I'd convert them here. The same "you're already touching this function" argument that applied to the relocated line applies to its two neighbours, and the current state leaves an odd surface for anyone writing an `except` block: passing both `statement` and `statement_id` raises `ValueError`, while passing neither raises `AirflowException`, for what a Dag author experiences as the same class of mistake. Both are argument-provision errors and `ValueError` fits both. That would take the ratchet entry from `::3` to `::1`.

If you'd rather keep this PR tightly scoped to what was asked for, a follow-up is fine too — just say which way you're going so the ratchet line doesn't get churned twice.

---

>

---

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
ahilashsasidharan: Done. I've changed this testcase and edited the second testcase to cover the case of why the sibling case in execute() is not converted to is None.

providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
ahilashsasidharan: Done.

ahilashsasidharan:
> Please retitle the commit as well as the PR, since the commit message is what the changelog is generated from.

I’ve updated the PR title and rewritten the first commit message to the suggested form. My understanding is that this is the commit whose message is used when generating the provider changelog, so that’s the one I corrected.

> If you'd rather keep this PR tightly scoped to what was asked for, a follow-up is fine too — just say which way you're going so the ratchet line doesn't get churned twice.

I’ve opted to make the additional changes in this PR. Since the diff already touches this function, it felt cleaner to apply the consistent ValueError treatment now rather than opening a follow‑up PR.

ahilashsasidharan:
Follow up question for some of my related commits in this series: 

#70874 & #70634 

1. Both make similar changes to different provider files but do not have testcases that fail without the PR and emphasise the specific logic for each check being written as it is  (i.e. with `is not None` or without `is None`). Is it worth going back to add additional parameterised testcases for these?

2. Should these also have a comments to highlight the split between sibling validation to avoid future confusion by readers?

ahilashsasidharan:
@shahar1 pinging since this should be ready for a re-review when you have the chance

eladkal:
cc @moomindani for review

providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
```diff
@@ -69,9 +69,11 @@ def __init__(
         include_airflow_query_tags: bool = True,
         **kwargs,
     ):
-        # Handle the scenario where either both statement and statement_id are set/not set
+        if statement is not None and statement_id is not None:
```
moomindani, line 72: With both fields `""` this raises "Cannot provide both" even though nothing was really provided — measured on the patched build, and `main` reports the accurate "One of either statement or statement_id must be provided." from `execute()` for the same input.

The `is not None` polarity is right and I would not weaken it; the cheapest fix is a message that holds for either failure, e.g. `"Provide exactly one of statement or statement_id."`

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
```diff
@@ -69,15 +69,33 @@ def test_init_statement_id(self):
         assert op.warehouse_id == WAREHOUSE_ID
 
     @pytest.mark.parametrize(
-        ("kwargs", "match"),
+        ("statement", "statement_id"),
         [
-            ({"statement": STATEMENT, "statement_id": STATEMENT_ID}, "Cannot provide both"),
-            ({}, "One of either statement or statement_id"),
+            (STATEMENT, STATEMENT_ID),
+            (STATEMENT, ""),
```
moomindani, line 76: Worth adding `("", "")` here so whichever behaviour you settle on for the both-empty case is pinned down — right now nothing covers it.

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
```diff
@@ -69,15 +69,33 @@ def test_init_statement_id(self):
...
+                statement_id=statement_id,
+                task_id=TASK_ID,
+                warehouse_id=WAREHOUSE_ID,
+            )
+
+    @pytest.mark.parametrize(
+        ("statement", "statement_id"),
+        [
+            (None, None),
+            ("", None),
```
moomindani, line 126: The case that actually motivates the polarity change is not covered anywhere: a *provided* field that renders to `None` under `render_template_as_native_obj=True`, which must not slip past the exclusivity check. Without it, a later well-meaning move back to a truthiness check in `execute()` would pass CI.

moomindani (approved):
Thanks for the clear write-up — the reasoning matches the criterion in #70503 exactly, and it follows the same restoration already merged for the Repos operators in #70551.

I ran a differential against a live SQL warehouse. Provider 7.18.1 already contains #70340, so the unpatched wheel reproduces current `main` and the patched one carries your two edits (Airflow 3.2.2, real `warehouse_id`, real statements executed):

| case | current main | with this PR |
| --- | --- | --- |
| `statement` + templated `statement_id` rendering to `None` (`render_template_as_native_obj=True`) | task **succeeds** — the contradictory pair is silently ignored and the statement runs | Dag import error |
| `statement="SELECT 1"`, `statement_id=""` | task **succeeds** against the warehouse | Dag import error |
| `statement=""`, `statement_id=""` | `execute()`: "One of either statement or statement_id must be provided." | `__init__`: "Cannot provide both statement and statement_id." |
| `statement` only / `statement_id` only | constructed | constructed |

The first row is an argument for this change that the description does not make: on `main`, supplying both arguments produces no error at all once one of them renders to `None`, and the sensor quietly runs the other. Catching that at parse time is a real improvement rather than a relocation — worth saying in the description.

The second row is a user-visible behaviour change: `statement_id=""` alongside a `statement` works today (confirmed end to end) and becomes a Dag import error here. Your parametrization shows it is deliberate, so this is only about making it visible to whoever reads the changelog later.

Nits inline, none blocking.

---

eladkal:
@ahilashsasidharan can you fix the open items?

ahilashsasidharan:
> @ahilashsasidharan can you fix the open items?

Sorry for the delay. Will work on these ~~today.~~ running late will take a look tomorrow.

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
ahilashsasidharan: I believe the polarity change should be covered, but I agree that the `render_template_as_native_obj=True` scenario isn't covered. 

I am not sure how to cover it cleanly though currently I've settled on potentially implementing something like the below testcase if it passes all checks. From my understanding (AI-assisted) we need to explicitly render templates after setting the value of it in the DAG so it requires leaving dead code at the bottom within the pytest.raise block to render templates that will never run since currently `__init__` catches the error:

```python
        context = {"dag": dag}
        op.render_template_fields(context)
        op.execute(context)
```

Without it someone moving the code block to execute might rewrite the test without rendering the templates which would still allow it to pass CI and might be missed by a reviewer. With this code they would need to explicitly remove the render which is easier to catch in a review.

I'll also add a comment mentioning intent of the test for anyone that is reading the code, but not sure if there is a cleaner way to implement this beyond that which you might have in mind.

```python
@pytest.mark.parametrize("statement,statement_id", [
    ("{{ None }}", STATEMENT_ID),  # statement would render to None
    (STATEMENT, "{{ None }}"),     # statement_id would render to None
    ("{{ None }}", "{{ None }}"),  # both would render to None
])
def test_both_provided_one_as_template_raises_at_init(self, statement, statement_id):
    dag = DAG(
        dag_id="test_native_obj_dag",
        start_date=timezone.datetime(2025, 1, 1),
        schedule=None,
        render_template_as_native_obj=True,
    )
    with pytest.raises(ValueError, match="Provide exactly one of statement or statement_id"):
        op = DatabricksSQLStatementsSensor(
            task_id=TASK_ID,
            warehouse_id=WAREHOUSE_ID,
            statement=statement,
            statement_id=statement_id,
            dag=dag,
        )
        context = {"dag": dag}
        op.render_template_fields(context)
        op.execute(context)
```

moomindani: Good question, and I think the dead-code version actually weakens the guard rather than strengthening it.

If the constructor raises, `render_template_fields` and `execute` never run — so the block asserts only "something in here raised `ValueError`". Move the check back into `execute()` later and the test still passes, which is exactly the regression you want it to catch. The stronger form is the opposite: assert on the constructor **alone**, with nothing after it. Then if the check moves out of `__init__`, the constructor stops raising and the test fails immediately.

```python
with pytest.raises(ValueError, match="Provide exactly one of statement or statement_id"):
    DatabricksSQLStatementsSensor(
        task_id=TASK_ID, warehouse_id=WAREHOUSE_ID,
        statement=statement, statement_id=statement_id, dag=dag,
    )
```

That also discriminates against the old code, which is what my note was really asking for. Verified both sides locally with `render_template_as_native_obj=True`, `statement="SELECT 1"`, `statement_id="{{ None }}"`:

* current main (checks in `execute()`, truthiness): constructs fine, renders to `statement_id=None`, and `execute()` proceeds to submit the statement — no exception at all.
* this PR: `ValueError: Cannot provide both statement and statement_id.` from the constructor.

So a plain `pytest.raises` around the constructor fails before the change and passes after it, which no current test does.

Two smaller points:

The `dag=...` with `render_template_as_native_obj=True` documents the scenario but the test does not depend on it — the constructor never renders, so it behaves the same without the flag. If you want a test where rendering is load-bearing rather than dead code, that is the *other* half, and it is worth having as a second test: pass only `statement="{{ None }}"`, construct, `render_template_fields`, then assert `execute()` raises. I measured that on this PR's head and it gives `ValueError: One of either statement or statement_id must be provided.` — the render call there is doing real work, and the pair of tests then pins each check to its own site.

Also note your `match=` uses "Provide exactly one of statement or statement_id" while the code currently says "Cannot provide both statement and statement_id." — that test will fail unless the message change from my other comment lands too. Worth doing both in the same push.

---

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
ahilashsasidharan: Done. Thanks!

providers/databricks/src/airflow/providers/databricks/sensors/databricks.py
ahilashsasidharan: Done. Thanks!

providers/databricks/tests/unit/databricks/sensors/test_databricks.py
ahilashsasidharan: Thanks I added both testcases. I guess I was overthinking this with the dead code idea, but I believe covering the _init_ template check should be enough and I added a comment for intent. Let me know if there are any issues with my implementation.

ahilashsasidharan:
> Thanks for the clear write-up — the reasoning matches the criterion in #70503 exactly, and it follows the same restoration already merged for the Repos operators in #70551.
> 
> I ran a differential against a live SQL warehouse. Provider 7.18.1 already contains #70340, so the unpatched wheel reproduces current `main` and the patched one carries your two edits (Airflow 3.2.2, real `warehouse_id`, real statements executed):
> case 	current main 	with this PR
> `statement` + templated `statement_id` rendering to `None` (`render_template_as_native_obj=True`) 	task **succeeds** — the contradictory pair is silently ignored and the statement runs 	Dag import error
> `statement="SELECT 1"`, `statement_id=""` 	task **succeeds** against the warehouse 	Dag import error
> `statement=""`, `statement_id=""` 	`execute()`: "One of either statement or statement_id must be provided." 	`__init__`: "Cannot provide both statement and statement_id."
> `statement` only / `statement_id` only 	constructed 	constructed
> 
> The first row is an argument for this change that the description does not make: on `main`, supplying both arguments produces no error at all once one of them renders to `None`, and the sensor quietly runs the other. Catching that at parse time is a real improvement rather than a relocation — worth saying in the description.
> 
> The second row is a user-visible behaviour change: `statement_id=""` alongside a `statement` works today (confirmed end to end) and becomes a Dag import error here. Your parametrization shows it is deliberate, so this is only about making it visible to whoever reads the changelog later.
> 
> Nits inline, none blocking.
> 
> Drafted-by: Claude Code (Opus 5); reviewed by @moomindani before posting

Thanks! I've updated the PR description to note both changes. Let me know if anything else is needed to make it visible in the changelog.

moomindani (approved):
This is the stronger form of the test, not the dead-code one — thanks for reworking it. Approving on `95ab90dd`.

I checked it discriminates, which was the point of the last exchange: moving the exclusivity check from `__init__` into `execute()` while keeping `is not None` fails both `*_validated_at_init` tests, and deleting the `render_template_fields` call from the execute-side test fails that one. The dead-code variant would have passed the first mutation.

The only source change since the commit I reviewed on 2026-08-19 is the message string, so the live-warehouse differential I reported then still describes this head.

---

potiuk (commented):
The `ValueError` narrowing and ratchet drop I asked for on 2026-08-01 are in, and the test now discriminates in the direction that matters. Locally: 30/30 sensor tests pass, `validate_operators_init.py` exits 0 on the file, ruff clean.

One substantive thing to add, and it is a fact that moved while this PR was open rather than anything you got wrong.

The 2026-08-01 review reasoned that #70340 was unreleased, so this PR was "a net no-op on *placement*" against released 7.18.0. That held at the time. But `providers-databricks/7.19.0` was tagged on 2026-08-19 and its tree contains #70340 — the released sensor checks exclusivity in `execute()` with truthiness. So the differential measured against a live warehouse is a change against a **released** version:

* `statement="SELECT 1"` with `statement_id=""` runs the statement in 7.19.0, and becomes a Dag import error here.
* Same for a `statement_id` template that renders to `None` under `render_template_as_native_obj=True`.
* The three `AirflowException`s become `ValueError`, so an `except AirflowException` around these stops matching.

The tightening itself is right and #70503 sanctions it explicitly ("strictly stricter — it also rejects `field=\"\"` alongside a second argument"). This is only about the note. Providers take no newsfragment, and `providers/databricks/docs/changelog.rst` asks for exactly this — a note just below the `Changelog` header for a breaking change that needs an explanation. Same shape as the `ComprehendCreateDocumentClassifierOperator` warning at the top of the amazon changelog.

Insert after line 28 (the blank line under `---------`), above `7.19.0`:

```rst
.. warning::
  ``DatabricksSQLStatementsSensor`` now rejects the combination of ``statement`` and
  ``statement_id`` when the Dag is parsed rather than when the task runs, and the check is
  ``is not None``, so an empty string counts as provided. ``statement="SELECT 1"`` together
  with ``statement_id=""`` — or with a ``statement_id`` template that renders to ``None``
  under ``render_template_as_native_obj=True`` — ran the statement in 7.19.0 and now raises
  at Dag parse time. Pass exactly one of the two and omit the other entirely instead of
  passing an empty value. The argument-validation errors in this sensor are also now
  ``ValueError`` rather than ``AirflowException``.
```

Two asks, then, and I'll merge once both are in:

1. The changelog warning above.
2. A rebase. The branch is 128 commits behind `main` and the green CI run is against a base from before the 7.19.0 release — worth a fresh run on current `main` before this goes in.

```bash
git fetch upstream main
git rebase upstream/main
git push --force-with-lease origin providers/rewrite_databrickssqlstatementssensor_exclusivity_check
```

Nothing else from me — the `__init__`/`execute()` split, the comment explaining it, and the constructor-only assertion are all settled and I do not want to re-open them.

---

potiuk (approved):

PR #70833, opened 2026-07-31
Fix Dag run duration stats crash on PostgreSQL 14+

pierrejeambrun:
The Dag run duration-stats endpoint crashes with a `TypeError` on PostgreSQL 14 and newer.

PostgreSQL 14 changed `EXTRACT(epoch FROM ...)` to return `numeric` instead of `double precision`, so `DagRun.duration` is hydrated as `decimal.Decimal` on Postgres 14+. The percentile interpolation in `compute_duration_stats` multiplies that value by a `float`, which Python disallows for `Decimal` (`unsupported operand type(s) for *: 'decimal.Decimal' and 'float'`), so the endpoint 500s for any Dag that has completed runs.

Coercing the durations to `float` up front keeps the computation backend-agnostic (Postgres `Decimal`, MySQL `int`, SQLite `float`). Adds a parametrized unit test that reproduces the crash with `Decimal` inputs and passes with the fix.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/services/ui/dag_run.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/services/ui/dag_run.py
@@ -35,7 +35,9 @@ def compute_duration_stats(durations: list[float]) -> DurationStats | None:
     if not durations:
         return None
 
-    sorted_d = sorted(durations)
+    # On Postgres 14+, DagRun.duration comes back as Decimal (EXTRACT(epoch ...) returns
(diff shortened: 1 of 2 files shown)


pierrejeambrun:
Screenshot 2026-07-31 at 15 58 44

pierrejeambrun:
Screenshot 2026-07-31 at 15 59 55

vatsrahul1001 (approved):

vatsrahul1001:
LGTM, verified on my end as well

PR #70835, opened 2026-07-31
Prepare Cloud Batch jobs before template rendering

shivaam:
Airflow replaces Jinja templates after an operator is created. Previously, this
operator converted protobuf jobs into dictionaries while it was being created,
which was too early.

The conversion is still needed because Airflow can only find and replace
templates inside nested job fields after the job becomes a dictionary.

This change moves the conversion to `prepare_template()`. The operator now
keeps the original job during construction and converts it immediately before
Airflow processes its templates. It also removes the operator from the
`validate-operators-init` exemption list.

Testing:

- 15 focused Google provider tests passed.
- All applicable prek hooks passed.
- A real Google Cloud Batch operator smoke test succeeded and Cloud Logging
  confirmed the rendered nested command; test resources were deleted.

Related: #70296

---

- [X] Yes — OpenAI Codex


eladkal (approved):

potiuk (approved):

potiuk (approved):
LGTM. The serialization test is the part that makes this convincing — `test_protobuf_job_is_prepared_before_serialization` round-trips through `DagSerialization.deserialize_dag()` and asserts the serialized task carries a dict with `{{ ds }}` still unrendered. That is exactly the question moving work into `prepare_template()` raises: it runs at parse time, so it only helps if the normalized value is what gets serialized to the worker. This proves it does.

Good idempotency check too — calling `resolve_template_files()` twice keeps the same object — and testing through `resolve_template_files()` rather than calling `prepare_template()` directly exercises the real entry point.

Also right to update `test_execute` to assert `job=JOB` rather than keeping the old expectation alive, since `__init__` no longer normalizes.

Note for whoever merges: #71201 is a near-identical change to the same operator, dropping the same exemption line, opened later and with thinner coverage. This one should land and that one close as a duplicate.

---

> will follow up.*
>

PR #70836, opened 2026-07-31
Show user display names for triggered runs and HITL responses

pierrejeambrun:
Follow-up to #70583, which added `BaseUser.get_display_name` and used it so audit-log rows show a friendly user name instead of the raw login.

Several other user-facing fields still recorded `get_name()` (the raw login):

- `triggering_user_name` on manually triggered Dag runs, backfills, and asset materializations — shown in the Dag Runs table and the Run header.
- The HITL responder name (`responded_by`) — shown in the HITL table and response form.

These now use `get_display_name()` too, so what the user sees is consistent with the audit log. For auth managers that don't override `get_display_name` (Simple, Keycloak) it falls back to `get_name`, so their behavior is unchanged; FAB users get their full name.

The edge3 worker maintenance-comment site (`get_name()` embedded in a comment string) is intentionally left out and will be a separate follow-up: that provider can run against an older core without the `get_display_name` interface, so it needs a `hasattr(user, "get_display_name")` compatibility guard.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
@@ -502,7 +502,7 @@ def materialize_asset(
             conf=params["conf"],
             run_type=DagRunType.ASSET_MATERIALIZATION,
             triggered_by=DagRunTriggeredByType.REST_API,
-            triggering_user_name=user.get_name(),
+            triggering_user_name=user.get_display_name(),
             state=DagRunState.QUEUED,
             partition_key=params["partition_key"],
             partition_date=params["partition_date"],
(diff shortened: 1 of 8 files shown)


pierrejeambrun:
## Before
Screenshot 2026-07-31 at 16 55 18
Screenshot 2026-07-31 at 16 56 50
Screenshot 2026-07-31 at 16 57 46


## After
Screenshot 2026-07-31 at 16 58 40
Screenshot 2026-07-31 at 16 59 54


Screenshot 2026-07-31 at 16 59 06

amoghrajesh (approved):

amoghrajesh:
Looks simple enough.

uranusjr (approved):

uranusjr:
Do we want to backport this? I feel it’s not necessary.

PR #70837, opened 2026-07-31
Clarify which signal each duplicated otel_* config option applies to

### LINKED ISSUE #43366: Five "otel_*" options appear twice in the "Configuration Reference"
### What do you see as an issue?

In https://github.com/apache/airflow/commit/ccf12026b44685f2d9917cd313676784e63c81af five configuration options have been moved from the "metrics" section to the "traces" section. Unfortunately, they have also been left in "metrics" untouched, and now they are effectively duplicated in the documentation. This is unfortunate, because users cannot easily tell what's the difference between them and which option should they prefer to use.

Here's a list of the duplicated options:

- otel_debugging_on
- otel_host
- otel_on
- otel_port
- otel_ssl_active

If you search for them on this page: https://airflow.apache.org/docs/apache-airflow/2.10.2/configurations-ref.html, you'll see that each of them occurs twice.

### Solving the problem

The problem can be solved by removing the `[metrics] otel_*` options from the `airflow/config_templates/config.yml` file (and keeping the `[traces] otel_*` options intact).

### Anything else

The newer options also have descriptions that do not render as nicely as the old ones in the documentation (the markdown highlights are gone). It may be a good moment to fix this.

### Code of Conduct

leepokai:
The ``otel_on``, ``otel_host``, ``otel_port``, ``otel_service``, ``otel_debugging_on`` and ``otel_ssl_active`` options appear in both the ``[metrics]`` and ``[traces]`` sections of the Configuration Reference, and their descriptions did not say which of the two features each copy configures, so users could not tell the duplicated options apart.

Three descriptions were also incorrect: ``[metrics] otel_host`` and ``[metrics] otel_interval_milliseconds`` claimed to also cover traces, and ``[metrics] otel_service`` described only traces — the code reads all of these exclusively for metrics (``observability/metrics/otel_logger.py``).

Each description now names the signal it configures (keeping both sets of options, per the resolution discussed in the issue), and the ``[traces]`` descriptions regain the literal markup (``True``/``False``/``config.yml``) used elsewhere so they render properly again.

closes: #43366

---

- [X] Yes — Claude Code, following [the Gen-AI assisted contributions guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst); all changes were verified against the code that reads these options and fully human-reviewed.

### DIFF
--- a/airflow-core/src/airflow/config_templates/config.yml
+++ b/airflow-core/src/airflow/config_templates/config.yml
@@ -1296,7 +1296,8 @@ metrics:
       default: "False"
     otel_on:
       description: |
-        Enables sending metrics to OpenTelemetry.
+        Enables sending Airflow metrics to an OpenTelemetry Collector. This option only affects
+        metrics; sending traces is configured separately in the ``[traces]`` section.
       version_added: 2.6.0
       type: boolean
       example: ~
@@ -1304,7 +1305,7 @@ metrics:
     otel_host:
       description: |
         Specifies the hostname or IP address of the OpenTelemetry Collector to which Airflow sends
-        metrics and traces.
+        metrics.
       version_added: 2.6.0
       version_deprecated: 3.2.0
       deprecation_reason: |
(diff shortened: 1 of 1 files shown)


leepokai:
Closing this — I won't be able to follow up on it. Sorry for the noise.

PR #70839, opened 2026-07-31
Make KeycloakAuthManager cache configurable

stephen-bracken:
Add configuration settings for KeycloakAuthManager `single_flight()` cache
---

Allow deployment managers to configure the cache timeout on `KeycloakAuthManager` bulk authorisation methods.

- Add `cache_ttl_seconds` option to configure `_CACHE_TTL_SECONDS` in `KeycloakAuthManager`
- Add `cache_timeout_seconds` option to configure `_SINGLE_FLIGHT_TIMEOUT_SECONDS` in `KeycloakAuthManager`

- [X] No

---

### DIFF
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cache.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cache.py
@@ -20,8 +20,15 @@ import threading
 import time
 from collections.abc import Callable
 
-_CACHE_TTL_SECONDS = 30
-_SINGLE_FLIGHT_TIMEOUT_SECONDS = 60
+from airflow.providers.common.compat.sdk import conf
+from airflow.providers.keycloak.auth_manager.constants import (
+    CONF_CACHE_TIMEOUT_SECONDS_KEY,
+    CONF_CACHE_TTL_SECONDS_KEY,
+    CONF_SECTION_NAME,
+)
+
+_CACHE_TTL_SECONDS: int | None = None
+_SINGLE_FLIGHT_TIMEOUT_SECONDS: int | None = None
 
 # Maps cache keys to (timestamp, result) pairs for TTL-based expiration.
 _cache: dict[tuple, tuple[float, frozenset[str]]] = {}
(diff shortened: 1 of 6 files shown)


potiuk:
Sensible thing to make configurable — a hardcoded 30 second auth-decision cache is exactly the sort of thing deployments need to tune. Two things before merge.

**`version_added` needs a real version.** Every other option in `provider.yaml` carries one (`0.0.1`, `0.4.0`), and the equivalent FAB option uses `version_added: 3.2.0`. The provider is currently at `0.8.2`, so this wants `0.9.0`:

```yaml
      cache_ttl_seconds:
        description: |
          Time (in seconds) to cache auth decisions in ``bulk is_authorized_{resource}`` methods.
        type: integer
        version_added: 0.9.0
        example: ~
        default: "30"
```

**The import-time read is what forces `reload()` in the tests, and that's worth avoiding.**

```python
_CACHE_TTL_SECONDS = conf.getint(CONF_SECTION_NAME, CONF_CACHE_TTL_SECONDS_KEY, fallback=30)
```

Because this is evaluated once at import, the tests need an autouse fixture that reloads the module to change it. Reloading mid-suite is fragile — it rebinds `_cache` and `_pending_requests` to fresh dicts while the `_clear_cache` fixture is holding references to the previous ones, so the two fixtures are operating on different objects depending on ordering. The autouse fixture also pins the TTL to 60 for every pre-existing test in the file, quietly changing their baseline.

I know `FabAuthManager` does read its `cache_ttl` at import, so there is precedent. But it's precedent worth breaking rather than following, because FAB shows the cost: its test decorates with

```python
@conf_vars({("fab", "cache_ttl"): "1"})
def test_deserialize_user(...):
```

and that has no effect whatsoever — `CACHE_TTL` was already baked into `TTLCache(maxsize=1024, ttl=CACHE_TTL)` when the class body was evaluated at import. The test passes because caching happens at all, not because the configured TTL was honoured. You clearly spotted this hazard, which is why the reload is there at all.

Reading it lazily makes the whole problem go away:

```python
def _cache_ttl_seconds() -> int:
    return conf.getint(CONF_SECTION_NAME, CONF_CACHE_TTL_SECONDS_KEY, fallback=30)
```

with the two call sites using `_cache_ttl_seconds()` instead of the constant. Then `conf_vars` works normally, the reload fixture disappears, `test_configure_ttl` becomes a plain assertion, and — the part that matters for operators — the setting can be changed without restarting the API server, which is most of the reason to make it configurable in the first place.

---

stephen-bracken:
updated docs, added lazy init to `_CACHE_TTL_SECONDS`, also made `_SINGLE_FLIGHT_TIMEOUT_SECONDS` configurable in the same way

providers/keycloak/src/airflow/providers/keycloak/auth_manager/cache.py
```diff
@@ -30,9 +37,25 @@
 _cache_lock = threading.Lock()
 
 
+def _cache_ttl_seconds() -> int:
```
vincbeck, line 40: I dont think it is worth caching that. Reading config is not that expensive and it creates complexity

stephen-bracken: @potiuk any thoughts? I don't have a strong opinion.

potiuk: It could be taken from secrets manager potentially - no big issues with it.

potiuk (approved):

PR #70858, opened 2026-07-31
Fix encrypted private keys in async SFTP hook

fallintoplace:
The async SFTP hook only read the legacy `passphrase` connection extra because its condition always evaluated to that key. As a result, encrypted private keys configured with the documented `private_key_passphrase` extra could not be imported.

This accepts the documented extra while preserving the existing `passphrase` behavior and precedence, matching the async SSH hook. The regression test verifies that the configured passphrase is passed when importing a private key.

---

### DIFF
--- a/providers/sftp/src/airflow/providers/sftp/hooks/sftp.py
+++ b/providers/sftp/src/airflow/providers/sftp/hooks/sftp.py
@@ -786,8 +786,10 @@ class SFTPHookAsync(BaseHook):
             self.key_file = extra_options["key_file"]
         if "known_hosts" in extra_options and self.known_hosts != self.default_known_hosts:
             self.known_hosts = extra_options["known_hosts"]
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Nice catch — `("passphrase" or "private_key_passphrase") in extra_options` collapses to `"passphrase" in extra_options` before the membership test runs, so the documented `private_key_passphrase` extra was silently ignored and encrypted keys configured that way could not be imported.

Good call matching `SSHHook._parse_extras`, which already handles this exactly the same way (`providers/ssh/.../ssh.py:613-615`) — same precedence, same empty-string fallback. Reusing the proven shape rather than inventing a new one is the right move.

I checked that switching the fixture to `private_key_passphrase` doesn't drop coverage of the legacy key — it's still exercised in several other fixtures in that file.

---

PR #70861, opened 2026-07-31
Allow filtering the Dag runs list by Dag tags

### LINKED ISSUE #19197: Allow search and filter by tags in dag run table
### Description

Would be great addition to filter by tags in the dag-run panel.
Screenshot-nparis_2021-10-25_10:33:51

### Use case/motivation

When searching for dags, for example `status failed` one, it would be great to be able to focus on a subset of the dags. Currently we only have the option to add filter based on `dag_id` and consequently to add meaningfull pattern in the dag-id. By the way the `tags` are much more flexible and meaningful to classify dags. Sadly so far there is no way to filter by `tags` in the UI.

### Related issues

_No response_

### Code of Conduct

Aaryan123456679:
Finding runs for Dags with a specific tag currently requires filtering by dag_id one at a time, since the cross-Dag runs table has no way to scope by tag. This adds a `tags` filter (with an `any`/`all` match mode) to the `GET /dags/~/dagRuns` endpoint and wires it into the runs table's filter bar in the UI, mirroring the existing team filter.

closes: #19197


airflow-core/src/airflow/ui/src/constants/filterConfigs.tsx
```diff
@@ -66,6 +66,7 @@ export const useFilterConfigs = () => {
   const { data: teamsData } = useTeamsServiceListTeams({ orderBy: ["name"] }, undefined, {
     enabled: multiTeamEnabled,
   });
+  const { data: tagsData } = useDagServiceGetDagTags({ orderBy: ["name"] });
```
bbovenzi, line 69: We will call this on every page with filterConfigs. Let's add an `enabled` flag.

airflow-core/src/airflow/ui/src/constants/filterConfigs.tsx
```diff
@@ -375,6 +376,15 @@ export const useFilterConfigs = () => {
...
       type: FilterTypes.TEXT,
     },
+    [SearchParamsKeys.TAGS]: {
+      icon: <FiTag />,
+      label: translate("common:tags"),
+      options: [
+        { label: translate("common:allTags"), value: "" },
+        ...(tagsData?.tags ?? []).map((tag) => ({ label: tag, value: tag })),
+      ],
+      type: FilterTypes.SELECT,
```
bbovenzi, line 386: This should match the dags list UX and be a typeahead multi select

bbovenzi (changes requested):
Please provide a screenshot, or better yet a gif,  of the functionality.

airflow-core/src/airflow/ui/src/constants/filterConfigs.tsx
Aaryan123456679: Dropped this query rather than gating it. The Tags filter now reuses main's TagsFilter (added since this PR was opened), which is registered as an EditorComponent, so the tag query only runs while the pill is mounted and not on every FilterBar page. filterConfigs.tsx no longer differs from main.

airflow-core/src/airflow/ui/src/constants/filterConfigs.tsx
Aaryan123456679: Switched to the same typeahead multiselect as the Dags list, including the Any/All match mode. The runs table now forwards tags_match_mode to the API (2bb9fd6019), with UI tests for the default any and for all. GIF added to the description.

Aaryan123456679:
pr70861-dagrun-tags-filter

GIF as requested.

PR #65692, opened 2026-04-22
Add multi-team lookup to Azure Key Vault backend

PrithviBadiga:
Adds multi-team lookup support to `AzureKeyVaultBackend`.

Updates:
- look up team-scoped secrets first when `team_name` is provided
- fall back to the global secret when no team-scoped secret exists
- avoid resolving team-scoped identifiers as global secrets when `team_name` is not provided
- document the Azure team-scoped naming convention

Lookup behavior:
- team-scoped: `{prefix}-{team_name}-{secret_id}`
- global fallback: `{prefix}-{secret_id}`

Verification:
- `AIRFLOW_HOME=$(mktemp -d) PYTHONPATH=/Users/prith/Desktop/Codex/airflow-65682/airflow-core/src:/Users/prith/Desktop/Codex/airflow-65682/providers/microsoft/azure/src /Users/prith/Desktop/Codex/airflow/.venv/bin/python -m pytest /Users/prith/Desktop/Codex/airflow-65689-azure/providers/microsoft/azure/tests/unit/microsoft/azure/secrets/test_key_vault.py`
- `/Users/prith/Desktop/Codex/airflow/.venv/bin/python -m ruff check /Users/prith/Desktop/Codex/airflow-65689-azure/providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py /Users/prith/Desktop/Codex/airflow-65689-azure/providers/microsoft/azure/tests/unit/microsoft/azure/secrets/test_key_vault.py`
- `/Users/prith/Desktop/Codex/airflow/.venv/bin/python -m ruff format --check /Users/prith/Desktop/Codex/airflow-65689-azure/providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py /Users/prith/Desktop/Codex/airflow-65689-azure/providers/microsoft/azure/tests/unit/microsoft/azure/secrets/test_key_vault.py`

Part of: #65682

PR #70869, opened 2026-07-31
Apply the team scope in the Google Secret Manager backend

potiuk:
`CloudSecretManagerBackend.get_conn_value` and `get_variable` both accept a
`team_name`, but the helper they delegate to has no such parameter:

```python
def get_conn_value(self, conn_id: str, team_name: str | None = None) -> str | None:
    ...
    return self._get_secret(self.connections_prefix, conn_id)   # team_name dropped

def _get_secret(self, path_prefix: str, secret_id: str) -> str | None:
```

The argument was discarded at the call boundary, so the backend built the team
agnostic secret name in every case and a team scoped lookup resolved whatever
that name pointed at. No team scoping was applied at all.

### Approach

The team scope is threaded through and the team scoped name is resolved first,
using the same `<team><TEAM_SEP><secret id>` convention the other secrets
backends use (`TEAM_SEP = "--"`, matching the Azure Key Vault backend).

1. **team scoped lookup first** — safe by construction, since it can only ever
   build the caller's own namespace;
2. after it misses, an id that already spells out a team scoped name is
   **refused**, because the team agnostic name would otherwise land inside some
   team's namespace;
3. otherwise the team agnostic lookup proceeds as before.

**The id is never parsed to work out which team it names**, because it cannot
be: a team name may itself contain the separator, so nothing in the string
distinguishes team `a` with id `b--c` from team `a--b` with id `c`. Comparing the
id against the prefix the caller's own team builds looks equivalent and is not —
a caller in team `a` matches `a--b`'s namespace on the prefix and would read its
secrets. Only the caller's own namespace is ever constructed, never parsed.

`get_config` is unchanged in behaviour for ordinary configuration keys; it shares
`_get_secret`, so a config id that spells out a team scoped name is now refused
too, which matches how the Azure backend already behaves.

### Behaviour change

- A team scoped lookup now resolves the team scoped secret name, where before it
  silently resolved the team agnostic one. Deployments that ran multi-team mode
  against this backend and relied on the old behaviour must store secrets under
  the team scoped name.
- An id that spells out a team scoped name is no longer resolvable through the
  team agnostic name from any scope.
- Single-team and non-team deployments are unaffected.

### Test plan

- [x] `providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py` —
      36 pass (30 existing, 6 new)
- [x] New `TestCloudSecretManagerBackendTeamScope`: resolved for its own team;
      not resolved for another team; not resolved without a team scope; team
      agnostic secrets still resolve for any scope; Variables scoped identically
- [x] `test_team_whose_name_extends_the_callers_is_not_readable` — a team named
      `team_a--prod` is not readable by a caller in `team_a`, and still reaches its
      own secret with the bare id
- [x] 5 of the 6 new tests fail against the unmodified backend; the sixth is a
      no-regression guard for team agnostic lookups

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
+++ b/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
@@ -171,6 +171,7 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
             return None
 
         if self._names_a_team_namespace(conn_id):
+            self._log_refusal("connection", conn_id)
             return None
 
         return self._get_secret(self.connections_prefix, conn_id, team_name)
@@ -187,6 +188,7 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
             return None
 
         if self._names_a_team_namespace(key):
+            self._log_refusal("variable", key)
             return None
 
         return self._get_secret(self.variables_prefix, key, team_name)
@@ -202,6 +204,15 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
             return None
 
         return self._get_secret(self.config_prefix, key)
+
+    def _log_refusal(self, kind: str, secret_id: str) -> None:
+        self.log.warning(
+            "%s id %r contains %r, which separates the team name from the secret id in a team "
+            "scoped secret name. Such an id is ambiguous and is not looked up. Returning None.",
+            kind.capitalize(),
+            secret_id,
+            TEAM_SEP,
+        )
 
     def _build_team_secret_name(self, path_prefix: str, team_name: str, secret_id: str) -> str:
         """
--- a/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
+++ b/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
@@ -19,7 +19,6 @@
 from __future__ import annotations
 
 import logging
-import re
 from collections.abc import Sequence
 
 from google.auth.exceptions import DefaultCredentialsError
@@ -171,6 +170,9 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         if self.connections_prefix is None:
             return None
 
+        if self._names_a_team_namespace(conn_id):
+            return None
+
         return self._get_secret(self.connections_prefix, conn_id, team_name)
 
     def get_variable(self, key: str, team_name: str | None = None) -> str | None:
@@ -182,6 +184,9 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         :return: Variable Value
         """
         if self.variables_prefix is None:
+            return None
+
+        if self._names_a_team_namespace(key):
             return None
 
         return self._get_secret(self.variables_prefix, key, team_name)
@@ -199,28 +204,41 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         return self._get_secret(self.config_prefix, key)
 
     def _build_team_secret_name(self, path_prefix: str, team_name: str, secret_id: str) -> str:
-        """Build a team scoped secret name using a dedicated separator before the secret id."""
+        """
+        Build a team scoped secret name using a dedicated separator before the secret id.
+
+        The secret id is used verbatim. Normalising it (``_`` -> ``sep``, as the Azure Key
+        Vault backend does) would let a plain id manufacture ``TEAM_SEP``: team ``a`` asking
+        for ``b__c`` would build the same name as team ``a--b`` asking for ``c``. It would
+        also contradict this backend's naming, which keeps underscores everywhere else --
+        unlike Azure, this backend does not override :meth:`build_path`.
+        """
         team_prefix = self.build_path(path_prefix, team_name, self.sep)
-        normalized_secret_id = secret_id.replace("_", self.sep)
-        return f"{team_prefix}{TEAM_SEP}{normalized_secret_id}"
+        return f"{team_prefix}{TEAM_SEP}{secret_id}"
 
     def _names_a_team_namespace(self, secret_id: str) -> bool:
         """
         Whether ``secret_id`` spells out a team scoped secret name.
 
         A team scoped secret is named ``<prefix><sep><team><TEAM_SEP><secret id>``, so an id
-        that already contains the team separator makes the team agnostic lookup resolve a
-        secret inside some team's namespace. That lookup is refused for such an id.
+        that itself contains the team separator makes the built name ambiguous: team ``a``
+        with id ``b--c`` and team ``a--b`` with id ``c`` produce the same string. Such an id
+        is refused for *every* lookup -- team scoped as well as team agnostic -- because the
+        ambiguity exists in both directions and the caller's own namespace is not a safe
+        harbour for it.
 
-        The id is never parsed to work out *which* team it names, because it cannot be: a
-        team name may itself contain the separator, so nothing in the string distinguishes
-        team ``a`` with id ``b--c`` from team ``a--b`` with id ``c``. Comparing the id against
+        The id is never parsed to work out *which* team it names, because it cannot be:
+        nothing in the string distinguishes the two readings above. Comparing the id against
         the prefix the caller's own team builds looks equivalent and is not -- a caller in
         team ``a`` would match ``a--b``'s namespace on the prefix and read its secrets. Only
         the caller's own namespace is ever constructed, never parsed.
+
+        The raw id is matched. Routing it through :meth:`build_path` first, as the Azure
+        backend does, is wrong here: the inherited implementation prepends a separator to an
+        empty prefix (``'' -> '-smtp_default'``) and normalises nothing, so the guard would
+        both mis-anchor and miss ids whose separator only appears after normalisation.
         """
-        normalized_secret_id = self.build_path("", secret_id, self.sep)
-        return bool(re.fullmatch(rf".+{re.escape(TEAM_SEP)}.+", normalized_secret_id))
+        return TEAM_SEP in secret_id
 
     def _get_secret(self, path_prefix: str, secret_id: str, team_name: str | None = None) -> str | None:
         """
@@ -230,20 +248,21 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         :param secret_id: Secret Key
         :param team_name: Team the lookup is scoped to (if any)
         """
+        # ``self.client`` builds a new ``_SecretManagerClient`` -- and with it a real gRPC
+        # client -- on every access, so it is bound once for both lookups below.
+        client = self.client
+
         # The team scoped name is tried first and is safe by construction: it can only ever
-        # build the caller's own namespace.
+        # build the caller's own namespace. Ids that would make that name ambiguous are
+        # refused by the callers before reaching here.
         if team_name:
-            team_secret = self.client.get_secret(
+            team_secret = client.get_secret(
                 secret_id=self._build_team_secret_name(path_prefix, team_name, secret_id),
                 project_id=self.project_id,
             )
             if team_secret is not None:
                 return team_secret
 
-        # Falling through to the team agnostic name would resolve a secret in some team's
-        # namespace when the id spells one out, so that is refused rather than resolved.
-        if self._names_a_team_namespace(secret_id):
-            return None
-
-        secret_id = self.build_path(path_prefix, secret_id, self.sep)
-        return self.client.get_secret(secret_id=secret_id, project_id=self.project_id)
+        return client.get_secret(
+            secret_id=self.build_path(path_prefix, secret_id, self.sep), project_id=self.project_id
+        )
--- a/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
+++ b/providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
@@ -19,6 +19,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from collections.abc import Sequence
 
 from google.auth.exceptions import DefaultCredentialsError
@@ -36,6 +37,10 @@ from airflow.utils.log.logging_mixin import LoggingMixin
 log = logging.getLogger(__name__)
 
 SECRET_ID_PATTERN = r"^[a-zA-Z0-9-_]*$"
+
+# Separator between the team name and the secret id in a team scoped secret name.
+# Matches the convention the other secrets backends use.
+TEAM_SEP = "--"
 
 
 class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
@@ -166,7 +171,7 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         if self.connections_prefix is None:
             return None
 
-        return self._get_secret(self.connections_prefix, conn_id)
+        return self._get_secret(self.connections_prefix, conn_id, team_name)
 
     def get_variable(self, key: str, team_name: str | None = None) -> str | None:
         """
@@ -179,7 +184,7 @@ class CloudSecretManagerBackend(BaseSecretsBackend, LoggingMixin):
         if self.variables_prefix is None:
             return None
 
-        return self._get_secret(self.variables_prefix, key)
+        return self._get_secret(self.variables_prefix, key, team_name)
 
     def get_config(self, key: str) -> str | None:
         """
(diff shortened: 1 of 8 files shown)


providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
```diff
@@ -37,6 +38,10 @@
 
 SECRET_ID_PATTERN = r"^[a-zA-Z0-9-_]*$"
 
+# Separator between the team name and the secret id in a team scoped secret name.
+# Matches the convention the other secrets backends use.
+TEAM_SEP = "--"
```
bugraoz93, line 42: Good to import this from core in the future not a blocker

bugraoz93 (approved):
Looks good! Non blocker comment to be referenced in the future

shahar1 (commented):
Thanks for picking this up — the underlying bug is real and worth fixing: `get_conn_value` / `get_variable` accepted `team_name` and dropped it at the `_get_secret` boundary, so this backend applied no team scoping at all. The lookup-order design is also the right shape: try the team-scoped name first, refuse the team-agnostic fallback for an id that spells out a namespace, and never parse the id to infer a team. The reasoning in the description about why a prefix match is not proof of ownership is correct.

The problem is in the naming half. It was transplanted from the Azure Key Vault backend, which **overrides** `build_path`, into the Google backend, which does not — and that one difference reopens the cross-namespace read this PR sets out to close.

### The cross-team read is still reachable

`_build_team_secret_name` normalises `_` → `sep` in the secret id, which can manufacture `TEAM_SEP` out of a plain conn_id. Running the PR's two helpers verbatim against the real `BaseSecretsBackend.build_path`:

```
victim   (team 'team_a--prod', id 'conn')      -> test-connections-team_a--prod--conn
attacker (team 'team_a',       id 'prod__conn') -> test-connections-team_a--prod--conn
cross-team read reachable: True
guard refuses 'prod__conn': False
```

A caller in `team_a` requesting conn_id `prod__conn` resolves the secret that team `team_a--prod` stores under conn_id `conn`. `_names_a_team_namespace` does not catch it because it does **not** normalise (see the inline note on that line) — it only sees a literal `--`.

This is the same failure class the description says it closes; `test_team_whose_name_extends_the_callers_is_not_readable` covers only the literal-`--` route.

### Documentation

The Azure PR that established this convention (#65692) added a 20-line "Multi-team lookup" section to its backend docs, and the amazon, yandex, and cncf backends document theirs too. `providers/google/docs/secrets-backends/google-cloud-secret-manager-backend.rst` is untouched here. Given the naming inconsistency flagged inline, docs matter more for this backend than they did for Azure — a user cannot derive the team-scoped name from the documented team-agnostic one.

### Changelog

`providers/AGENTS.md`:

> When a provider change needs a user-visible note (typically a breaking change or important behavior change that warrants explanation), update the provider's `docs/changelog.rst` directly in the same PR

The description's own "Behaviour change" section is that trigger — multi-team deployments running this backend today have to re-store their secrets under the new names.

### `get_config`

> `get_config` is unchanged in behaviour for ordinary configuration keys; it shares `_get_secret`, so a config id that spells out a team scoped name is now refused too, which matches how the Azure backend already behaves.

Azure's `get_config` calls `_get_secret(self.config_prefix, key)` and never invokes `_is_team_specific_accessed_as_global`, so it does **not** refuse such an id. The change here is stricter than the stated reference, and there's no test covering it.

### CI

The provider test jobs haven't reported yet on this head — the CI image build is still pending and no `Providers` job appears in the rollup. Worth waiting for that before merge regardless of the above.

---

**Suggested additional reviewers** for the multi-team secrets convention (named, not requested — nobody is being notified): `vincbeck`, most recent author on this file and on the multi-team secrets work; `PrithviBadiga`, who authored the Azure backend this mirrors.

---

>

---
Drafted-by: Claude Code (Opus 5, 1M context); reviewed by @shahar1 before posting

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
```diff
@@ -193,12 +198,52 @@ def get_config(self, key: str) -> str | None:
 
         return self._get_secret(self.config_prefix, key)
 
-    def _get_secret(self, path_prefix: str, secret_id: str) -> str | None:
+    def _build_team_secret_name(self, path_prefix: str, team_name: str, secret_id: str) -> str:
+        """Build a team scoped secret name using a dedicated separator before the secret id."""
+        team_prefix = self.build_path(path_prefix, team_name, self.sep)
+        normalized_secret_id = secret_id.replace("_", self.sep)
```
shahar1: **Cross-team read: this line can manufacture `TEAM_SEP`.**

`_` → `sep` normalisation on the secret id means a plain conn_id containing `__` produces a name inside *another* team's namespace:

```
team 'team_a',       id 'prod__conn' -> test-connections-team_a--prod--conn
team 'team_a--prod', id 'conn'       -> test-connections-team_a--prod--conn
```

Identical. A caller in `team_a` reads team `team_a--prod`'s secret. `_names_a_team_namespace("prod__conn")` returns `False`, so nothing stops it. That is the exact failure the description says is closed — `test_team_whose_name_extends_the_callers_is_not_readable` only covers the literal-`--` spelling.

**Second issue on the same line:** the normalisation makes team-scoped naming inconsistent with this backend's documented convention. The docs promise `airflow-connections-smtp_default` for conn_id `smtp_default` (underscores preserved), but the team-scoped name becomes:

```
team-agnostic: airflow-connections-smtp_default
team-scoped  : airflow-connections-team_a--smtp-default
```

Underscores kept in the prefix and the team name, converted in the secret id. In Azure this is consistent because its `build_path` override normalises the whole name; here nothing else does.

Dropping the `.replace()` and using the raw `secret_id` fixes both — the team separator can then only come from a team name, and the naming matches the rest of the backend:

```suggestion
        normalized_secret_id = secret_id
```

(at which point the local is worth inlining or renaming).

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
```diff
@@ -193,12 +198,52 @@ def get_config(self, key: str) -> str | None:
...
+        secret inside some team's namespace. That lookup is refused for such an id.
+
+        The id is never parsed to work out *which* team it names, because it cannot be: a
+        team name may itself contain the separator, so nothing in the string distinguishes
+        team ``a`` with id ``b--c`` from team ``a--b`` with id ``c``. Comparing the id against
+        the prefix the caller's own team builds looks equivalent and is not -- a caller in
+        team ``a`` would match ``a--b``'s namespace on the prefix and read its secrets. Only
+        the caller's own namespace is ever constructed, never parsed.
+        """
+        normalized_secret_id = self.build_path("", secret_id, self.sep)
```
shahar1: This was copied from the Azure backend, which **overrides** `build_path` — special-casing the empty prefix and replacing `_` with `sep`:

```python
# key_vault.py
@staticmethod
def build_path(path_prefix: str, secret_id: str, sep: str = "-") -> str:
    path = f"{secret_id}" if path_prefix == "" else f"{path_prefix}{sep}{secret_id}"
    return path.replace("_", sep)
```

This backend inherits the base implementation, which does neither:

```python
# shared/secrets_backend/.../base.py
return f"{path_prefix}{sep}{secret_id}"
```

So the call yields a stray leading separator and no normalisation at all:

```
google: build_path('', 'smtp_default') -> '-smtp_default'
azure : build_path('', 'smtp_default') -> 'smtp-default'
```

Two consequences: the leading `-` changes the regex outcome for ids starting with `--`, and — more importantly — the guard never sees the normalised form, which is why it misses the `prod__conn` case flagged above.

Given the suggestion on line 204 (stop normalising the id), the guard can just match the raw id and skip `build_path` entirely.

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
```diff
@@ -193,12 +198,52 @@ def get_config(self, key: str) -> str | None:
...
         Get secret value from the SecretManager based on prefix.
 
         :param path_prefix: Prefix for the Path to get Secret
         :param secret_id: Secret Key
+        :param team_name: Team the lookup is scoped to (if any)
         """
+        # The team scoped name is tried first and is safe by construction: it can only ever
+        # build the caller's own namespace.
+        if team_name:
+            team_secret = self.client.get_secret(
```
shahar1: `self.client` is a property that constructs a new `_SecretManagerClient` on every access, and that object's `client` is a per-instance `cached_property` holding a real gRPC `SecretManagerServiceClient`. A team-scoped miss now builds two of them per lookup, on a path that's hit for every connection and variable resolution.

Bind it once at the top of `_get_secret`:

```python
client = self.client
```

and use `client.get_secret(...)` for both calls.

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
```diff
@@ -272,3 +272,99 @@ def test_init_succeeds_with_explicit_project_id(self, mock_client_callable, mock
...
+        """Back the backend with an in-memory secret store keyed by full secret name."""
+        client = mock.MagicMock()
+        client.get_secret.side_effect = lambda secret_id, project_id, **kwargs: store.get(secret_id)
+        mock_client_callable.return_value = client
+        return CloudSecretManagerBackend(
+            connections_prefix=CONNECTIONS_PREFIX, variables_prefix=VARIABLES_PREFIX, project_id=PROJECT_ID
+        )
+
+    @staticmethod
+    def _team_name(backend, team, secret_id, prefix=CONNECTIONS_PREFIX):
```
shahar1, line 309: The expected secret names come from `_build_team_secret_name` — the method under test. The mock store is keyed by whatever the implementation emits, so the tests agree with the implementation by construction and any naming defect is invisible to them. That is why the `prod__conn` collision and the underscore inconsistency both pass here.

Worth having at least one test assert the literal name, e.g.:

```python
assert backend._build_team_secret_name(CONNECTIONS_PREFIX, "team_a", "test_postgres") == (
    "test-connections-team_a--test_postgres"
)
```

and one negative test that constructs the attacker's id independently of the implementation:

```python
# team_a must not reach team_a--prod's namespace via underscore normalisation
store = {self._team_name(backend, "team_a--prod", CONN_ID): CONN_URI}
assert backend.get_conn_value(conn_id=f"prod__{CONN_ID}", team_name="team_a") is None
```

That second one fails against the current implementation.

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
```diff
@@ -272,3 +272,99 @@ def test_init_succeeds_with_explicit_project_id(self, mock_client_callable, mock
...
+        return backend._build_team_secret_name(prefix, team, secret_id)
+
+    @mock.patch(MODULE_NAME + ".get_credentials_and_project_id")
+    @mock.patch(MODULE_NAME + "._SecretManagerClient")
+    def test_team_scoped_secret_is_resolved_for_its_own_team(self, mock_client, mock_get_creds):
+        mock_get_creds.return_value = CREDENTIALS, PROJECT_ID
+        backend = self._backend(mock_client, {})
```
shahar1: This backend is discarded and rebuilt three lines down — same in all six new tests. It only exists to have an object to call `_build_team_secret_name` / `build_path` on.

`_backend`'s `side_effect` closes over `store` by reference, so one backend plus a dict populated afterwards does the same job:

```python
store = {}
backend = self._backend(mock_client, store)
store[self._team_name(backend, self.TEAM, CONN_ID)] = CONN_URI
```

Or make `_backend` return `(backend, store)` and drop the double construction everywhere.

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
```diff
@@ -272,3 +272,99 @@ def test_init_succeeds_with_explicit_project_id(self, mock_client_callable, mock
...
+        encoded = f"{self.TEAM}--{CONN_ID}"
+        store = {self._team_name(backend, self.TEAM, CONN_ID): CONN_URI}
+        backend = self._backend(mock_client, store)
+
+        assert backend.get_conn_value(conn_id=encoded) is None
+
+    @mock.patch(MODULE_NAME + ".get_credentials_and_project_id")
+    @mock.patch(MODULE_NAME + "._SecretManagerClient")
+    def test_team_whose_name_extends_the_callers_is_not_readable(self, mock_client, mock_get_creds):
```
shahar1, line 344: This test is the right idea and the most valuable one in the class — but it only covers the case where the caller *spells out* `team_a--prod--...` literally. The same namespace is reachable without any `--` in the id at all, via the underscore normalisation on line 204 of the backend:

```python
assert backend.get_conn_value(conn_id=f"prod__{CONN_ID}", team_name=self.TEAM) is None
```

That currently returns `CONN_URI`. Worth adding here once the naming issue is resolved, since it's the regression guard for it.

vincbeck (approved):

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
potiuk: Agreed. There's no shared constant in core to import yet -- Azure and Amazon each define their own `TEAM_SEP`, and Yandex derives it from `sep` -- so this stays local for now. Worth consolidating the moment one lands in core.

---

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
potiuk: Fixed -- the `.replace()` is gone and the secret id is used verbatim.

Both halves of this were right. I reproduced the collision exactly as you wrote it, and the naming inconsistency was the other side of the same coin. `_build_team_secret_name` now builds `<prefix><sep><team>--<secret_id>` with the id untouched, so the team separator can only ever originate from the team name, and underscores survive in the team-scoped name just as they do in the team-agnostic one.

Two tests cover it: `test_team_secret_name_is_built_literally` pins the literal string rather than deriving it from the method under test, and `test_underscores_cannot_manufacture_a_team_namespace` asserts your `prod__conn` case returns `None`. The latter fails against the revision you reviewed.

For the record, the same normalisation is still live in the Azure backend. #70876 closes it there -- but at the entry point rather than in the builder, since Azure normalises everywhere by design and the team-scoped name has to match how the secret is actually stored.

---

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
potiuk: Fixed -- the guard matches the raw id now (`TEAM_SEP in secret_id`) and doesn't call `build_path` at all.

Your read of the inheritance is correct: this backend uses the base implementation, so `build_path('', 'smtp_default')` yields `-smtp_default` -- a stray leading separator and no normalisation, which is why the guard never saw the `prod__conn` form. With the id no longer normalised when the team-scoped name is built, there's nothing left for the guard to normalise either, so matching the raw id is both correct and simpler. I left a note in the docstring on why routing through `build_path` is wrong here, so it doesn't get copied back from Azure later.

---

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
potiuk: Fixed -- `client = self.client` is bound once at the top of `_get_secret` and used for both lookups, with a comment recording why, since the cost isn't visible at the call site.

---

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
potiuk: Fixed. `test_team_secret_name_is_built_literally` now asserts the literal name:

```
test-connections-team_a--test_postgres
```

plus a second assertion that `smtp_default` keeps its underscore.

Your negative test went in as `test_underscores_cannot_manufacture_a_team_namespace`. I confirmed it fails against the revision you reviewed and passes now -- and re-confirmed after the later refactor by reintroducing the `.replace()` and watching it fail again, so it isn't passing vacuously.

---

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
potiuk: Fixed, using your second suggestion -- `_backend` returns `(backend, store)` and the store is populated after the backend exists, since the lookup closes over it by reference. The discarded backend is gone from every test in the class.

---

providers/google/tests/unit/google/cloud/secrets/test_secret_manager.py
potiuk: Covered, though as a separate test rather than folded into this one -- `test_underscores_cannot_manufacture_a_team_namespace` asserts exactly your line:

```python
assert backend.get_conn_value(conn_id=f"prod__{CONN_ID}", team_name=self.TEAM) is None
```

They stay apart because this test is about a team whose *name* extends the caller's, and that one is about the *id* manufacturing the separator -- two different mechanisms that happen to reach the same namespace, each worth failing independently.

---

PR #70878, opened 2026-07-31
Refuse the team agnostic fall-through for a team scoped AWS secret name

potiuk:
The team scoped lookup runs first and, on a miss, falls through to the team
agnostic name. A guard was meant to stop a caller naming another team's secret:

```python
return team_name is None and bool(re.fullmatch(..., secret_id))
```

Its first condition is `team_name is None`, so it does nothing whenever a team
scope is supplied — exactly when it matters. A caller authorised for one team
could supply an id spelling out another team's namespace; the team scoped probe
missed, and the team agnostic lookup resolved the other team's secret.

### Approach

1. the **team scoped lookup runs first** — safe by construction, since it can only
   ever build the caller's own namespace;
2. after it misses, an id that spells out **any** team namespace is refused rather
   than resolved through the team agnostic name;
3. otherwise the team agnostic lookup proceeds as before.

**The id is never parsed to work out which team it names**, because it cannot be:
a team name may itself contain the separator, so nothing in the string
distinguishes one team's namespace from another whose name extends it. Comparing
the id against the prefix the caller's own team builds looks equivalent and is
not — a caller in team `a` matches `a--b`'s namespace on the prefix and would read
its secrets. Only the caller's own namespace is ever constructed, never parsed.

### Behaviour change

An id that spells out a team scoped name is no longer resolvable through the team
agnostic name from any scope, including the team that owns it — that spelling is
what made a prefix match look like ownership. Callers reach their own team's
secrets with the bare id plus their team scope, which is unaffected, as are
single-team and non-team deployments.

### Test plan

- [x] New tests: a caller scoped to one team cannot reach another team's secret by
      naming it, and a team whose name *extends* the caller's (`teama--prod` vs
      `teama`) is not reachable either
- [x] Both new tests fail against the unmodified backend — the target secret is
      resolvable in the fixture, so a backend that falls through returns its value
- [x] Existing team-scope tests unchanged and passing

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
@@ -322,19 +322,27 @@ class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
                 lookup_pattern,
             )
             return None
-        if team_name is None and re.fullmatch(r"[^-]+--.+", secret_id):
-            self.log.warning(
-                "Secret ID %r contains a '--' separator, which is reserved for team-scoped lookups, "
-                "but no team context was provided. Returning None.",
-                secret_id,
-            )
-            return None
+        # Tried first and safe by construction: it can only build the caller's own namespace.
         if path_prefix and team_name:
             secrets_path = self.build_path(path_prefix, f"{team_name}--{secret_id}", self.sep)
             value = self._get_secret_value(secret_id, secrets_path)
             if value is not None:
                 return value
 
+        # Falling through to the team agnostic path would resolve a secret inside some team's
+        # namespace when the id spells one out, so that is refused rather than resolved. The id
+        # is never parsed to work out *which* team it names: a team name may itself contain the
+        # separator, so nothing distinguishes team "a" with id "b--c" from team "a--b" with id
+        # "c". Comparing it against the prefix the caller's own team builds looks equivalent and
+        # is not -- a caller in team "a" matches "a--b"'s namespace on the prefix. Only the
+        # caller's own namespace is ever constructed, never parsed.
+        if re.fullmatch(r".+--.+", secret_id):
+            self.log.warning(
+                "Secret ID %r contains a '--' separator, which is reserved for team-scoped "
+                "lookups, so it cannot be resolved through the team agnostic name. Returning None.",
+                secret_id,
+            )
+            return None
         if path_prefix:
             secrets_path = self.build_path(path_prefix, secret_id, self.sep)
         else:
--- a/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
@@ -262,6 +262,10 @@ class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
         if self.config_prefix is None:
             return None
 
+        if self._names_a_team_namespace(key):
+            self._log_refusal("configuration option", key)
+            return None
+
         return self._get_secret(self.config_prefix, key, self.config_lookup_pattern)
 
     def _get_secret_value(self, secret_id: str, secrets_path: str) -> str | None:
@@ -314,15 +318,17 @@ class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
             )
             return None
 
-    def _names_a_team_namespace(self, secret_id: str) -> bool:
+    @staticmethod
+    def _names_a_team_namespace(secret_id: str) -> bool:
         """
         Whether ``secret_id`` spells out a team scoped secret name.
 
         A team scoped secret is named ``<team><TEAM_SEP><secret id>``, so an id that itself
         contains the team separator makes the built name ambiguous: team ``a`` with id ``b--c``
-        and team ``a--b`` with id ``c`` produce the same string. Such an id is refused for
-        *every* lookup -- team scoped as well as team agnostic -- because the ambiguity exists
-        in both directions and the caller's own namespace is not a safe harbour for it.
+        and team ``a--b`` with id ``c`` produce the same string. Such an id is refused by every
+        getter -- connections, variables and configuration options, team scoped as well as team
+        agnostic -- because the ambiguity exists in both directions and the caller's own
+        namespace is not a safe harbour for it.
 
         The id is never parsed to work out *which* team it names, because it cannot be: nothing
         in the string distinguishes the two readings above. Comparing the id against the prefix
--- a/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
@@ -28,6 +28,9 @@ from airflow.providers.amazon.aws.utils import trim_none_values
 from airflow.secrets import BaseSecretsBackend
 from airflow.utils.log.logging_mixin import LoggingMixin
 
+# Separator between the team name and the secret id in a team scoped secret name.
+TEAM_SEP = "--"
+
 
 class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
     """
@@ -207,6 +210,10 @@ class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
         if self.connections_prefix is None:
             return None
 
+        if self._names_a_team_namespace(conn_id):
+            self._log_refusal("connection", conn_id)
+            return None
+
         secret = self._get_secret(
             self.connections_prefix, conn_id, self.connections_lookup_pattern, team_name
         )
@@ -237,6 +244,10 @@ class SecretsManagerBackend(BaseSecretsBackend, LoggingMixin):
         :return: Variable Value
         """
         if self.variables_prefix is None:
+            return None
+
+        if self._names_a_team_namespace(key):
+            self._log_refusal("variable", key)
             return None
 
         return self._get_secret(self.variables_prefix, key, self.variables_lookup_pattern, team_name)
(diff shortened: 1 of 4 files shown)


providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
```diff
@@ -239,6 +246,10 @@ def get_variable(self, key: str, team_name: str | None = None) -> str | None:
         if self.variables_prefix is None:
             return None
 
+        if self._names_a_team_namespace(key):
+            self._log_refusal("variable", key)
+            return None
+
         return self._get_secret(self.variables_prefix, key, self.variables_lookup_pattern, team_name)
 
     def get_config(self, key: str) -> str | None:
```
shahar1, line 255: `get_config` silently loses the guard. It was previously covered because the check lived in `_get_secret`, which all three getters share; now that the check has moved up into `get_conn_value` and `get_variable`, `get_config` reaches `_get_secret` unguarded, so `get_config("a--b")` resolves `airflow/config/a--b` where it used to return `None` with a warning.

There is no cross-team read here — `get_config` takes no `team_name`, so a team scoped name is never built for it — and arguably config keys should never have been caught by the old guard in the first place. But as it stands the new `_names_a_team_namespace` docstring asserts such an id is *"refused for **every** lookup"*, which is no longer true, and the loosening isn't mentioned in the PR's "Behaviour change" section.

Either add the check to `get_config` too, or keep the exemption and say so in both the docstring and the PR description.

providers/amazon/src/airflow/providers/amazon/aws/secrets/secrets_manager.py
```diff
@@ -303,6 +314,33 @@ def _get_secret_value(self, secret_id: str, secrets_path: str) -> str | None:
             )
             return None
 
+    def _names_a_team_namespace(self, secret_id: str) -> bool:
```
shahar1: Two small things on this helper:

1. The docstring says such an id is refused for *"**every** lookup -- team scoped as well as team agnostic"*. That is true for connections and variables, but `get_config` calls `_get_secret` without going through this check, so config lookups are exempt (see the comment on `get_config`). Worth reconciling the claim with the behaviour in whichever direction you pick.

2. It doesn't use `self` — `@staticmethod` fits, and matches core's equivalent `EnvironmentVariablesBackend._is_team_specific_accessed_as_global`, which is already a `@staticmethod`.

shahar1 (commented):
Approach looks right and the tests are what convinced me; a few observations inline and below, none blocking — though two of them look in-scope for this PR rather than follow-ups.

The old guard's first condition was `team_name is None`, so it did nothing exactly when a team scope was supplied. Moving the refusal ahead of **both** lookups rather than only the team agnostic fall-through is the correct shape, and `test_team_scoped_lookup_cannot_reach_a_longer_teams_namespace` is the test that earns it: team `my_team` asking for `prod--test_postgres` builds exactly the name team `my_team--prod` builds for `test_postgres`, so the team scoped probe *hits* another team's secret before any fall-through is reached. Guarding only the fall-through would have left that open.

### 1. The same flaw is in core's `EnvironmentVariablesBackend`, and isn't fixed here

The guard quoted in the PR description —

```python
return team_name is None and bool(re.fullmatch(..., secret_id))
```

— is [`airflow-core/src/airflow/secrets/environment_variables.py:66`](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/secrets/environment_variables.py#L66), not the Amazon code this PR patches (the Amazon one has no `bool(...)` wrapper). That backend has the identical fall-through:

- team scoped name: `AIRFLOW_CONN_` + `_<TEAM>___` + `<CONN_ID>`
- team agnostic name: `AIRFLOW_CONN_<CONN_ID>`

A caller scoped to `teama` requesting conn_id `_teamb___myconn` misses `AIRFLOW_CONN__TEAMA____TEAMB___MYCONN`, then falls through to `AIRFLOW_CONN__TEAMB___MYCONN` — team b's connection. The collision variant applies too: `teama` asking for `prod___myconn` builds the same variable name that team `teama___prod` builds for `myconn`.

Fixing the two Amazon backends while the code the description quotes stays broken leaves the sibling caller open. Fine as a separate PR, but it seems worth doing together, or at least tracking.

### 2. The provider docs still describe the old, conditional refusal

- `providers/amazon/docs/secrets-backends/aws-secrets-manager.rst:176`
- `providers/amazon/docs/secrets-backends/aws-ssm-parameter-store.rst:152`

Both still carry:

> Connection ids and variable keys matching `<team>--<name>` are reserved for team-scoped lookups. A request **without team context** for a key matching this pattern will return `None` to prevent cross-team access.

After this PR the refusal is unconditional in every scope, including the owning team — which is exactly the behaviour change the description calls out. Worth updating in the same PR, ideally with a line telling a deployment that already has ids containing `--` what to do about them.

### 3. The description's rationale contradicts its own tests

Approach bullet 1 says the team scoped lookup is *"safe by construction, since it can only ever build the caller's own namespace"*, while the docstring of `test_team_scoped_lookup_cannot_reach_a_longer_teams_namespace` says *"The team scoped name is not safe by construction -- the id can extend it."* The code follows the test — the refusal correctly runs before both lookups — so it is the rationale that is stale, and it will land in the commit message as written.

### Nit — the helper is duplicated verbatim

`TEAM_SEP`, the 15-line `_names_a_team_namespace` docstring and `_log_refusal` are now copy-pasted identically into both modules. The two backends already duplicate the `_get_secret` shape so this follows the local grain, but a shared helper would keep the rationale in one place — it is the kind of comment that drifts once only one copy gets edited.

---

Not requesting anyone, just flagging who has context on this code: `justinpakzad` authored the original guard (#65703) and its logging follow-up (#66059), and `vincbeck` has been across the Amazon secrets backends and multi-team.

---

>

providers/amazon/src/airflow/providers/amazon/aws/secrets/systems_manager.py
```diff
@@ -155,6 +162,10 @@ def get_variable(self, key: str, team_name: str | None = None) -> str | None:
         if self.variables_prefix is None:
             return None
 
+        if self._names_a_team_namespace(key):
+            self._log_refusal("variable", key)
+            return None
+
         return self._get_secret(self.variables_prefix, key, self.variables_lookup_pattern, team_name)
 
     def get_config(self, key: str) -> str | None:
```
shahar1, line 171: Same as in `secrets_manager.py`: `get_config` still calls `_get_secret` directly, so it no longer gets the check that used to live there. Not a cross-team read (no `team_name` is ever passed here), but it means the "refused for every lookup" claim in `_names_a_team_namespace` doesn't hold, and the loosening isn't in the PR's "Behaviour change" section.

providers/amazon/src/airflow/providers/amazon/aws/secrets/systems_manager.py
```diff
@@ -182,6 +193,33 @@ def _get_parameter_value(self, ssm_path: str) -> str | None:
             self.log.debug("Parameter %s not found.", ssm_path)
             return None
 
+    def _names_a_team_namespace(self, secret_id: str) -> bool:
```
shahar1: `@staticmethod` here too — no `self` use, and it matches core's `EnvironmentVariablesBackend._is_team_specific_accessed_as_global`.

Also, this docstring and `_log_refusal` below are byte-identical to the copies in `secrets_manager.py`. Not a blocker given the two backends already mirror each other, but the rationale is long enough that one copy will eventually get updated and the other won't.

providers/amazon/tests/unit/amazon/aws/secrets/test_secrets_manager.py
```diff
@@ -92,6 +92,54 @@ def test_global_caller_cannot_access_team_scoped_connection(self):
...
+
+    @mock_aws
+    def test_refusing_an_ambiguous_id_is_logged(self, caplog):
+        """A silent ``None`` is indistinguishable from a missing secret, so the refusal is logged."""
+        backend = SecretsManagerBackend()
+
+        assert backend.get_conn_value(conn_id="prod--test_postgres") is None
+        assert backend.get_variable(key="prod--hello") is None
+
+        refusals = [r for r in caplog.records if "is ambiguous and is not looked up" in r.getMessage()]
```
shahar1: This matches on the rendered log message, which the project's testing standard steers away from:

> Do not assert on raw log text (`caplog.text` for example), these are legacy string matching APIs planned for removal. Structured assertions via `caplog` (which resolves to `cap_structlog` under structlog) are fine and preferred: `"event name" in caplog` or `{"event": ..., "field": ...} in caplog`.

Beyond the standard, the coupling is what makes it worth changing: the assertion pins the exact sentence *"is ambiguous and is not looked up"*, so any rewording of `_log_refusal` breaks the test for a reason that has nothing to do with the behaviour under test. A structured assertion on the event plus the level would survive that.

Testing that the refusal is logged at all is a good call, though — a silent `None` really is indistinguishable from a missing secret.

shahar1 (approved):

potiuk:
Thanks @shahar1 — all four addressed, and the `get_config` one was a real regression I introduced. Moving the check out of `_get_secret` and up into the two team-aware getters left the third reaching `_get_secret` bare, so the helper's "refused for *every* lookup" claim stopped being true the moment I wrote it.

- **`get_config` guarded** in both `secrets_manager.py` and `systems_manager.py`, and the docstring now names the three getters instead of claiming universality. Confirmed the test fails without the guard (`assert 2 == 3`), so it pins the behaviour rather than just passing.
- **`@staticmethod`** on `_names_a_team_namespace` in both files.
- **The log assertion** — I tried your structured suggestion first and it does not work here: Airflow logs through structlog, which renders the format args into `msg` before the stdlib record exists, so `record.args` is `()` and there is no structured payload to assert on. Settled on level + logger + the refused id, which is the load-bearing data, and left a comment saying why so the next person does not repeat the attempt. Extended to cover `get_config`.

On the duplicated docstring and `_log_refusal` between the two files — agreed it is not ideal, but the two backends share no base class today and I would rather not introduce one inside a security fix. Left as is.

One thing you should know: **the same `get_config` gap is in #70876 (Azure), which is already merged.** Same shape, and you approved that one before finding this in the sibling. Not a cross-team read there either — `get_config` never receives a `team_name` — but the merged docstring makes the same false claim. Opening a follow-up.

o-nikolas:
@vincbeck you worked on the team secrets stuff, you should have a look at this one, even though it's already merged.

vincbeck:
> @vincbeck you worked on the team secrets stuff, you should have a look at this one, even though it's already merged.

Looks good

PR #70879, opened 2026-07-31
Fix Microsoft Graph filesystem auth by defaulting OAuth2 scope

### LINKED ISSUE #70822: scopes vs scope mismatch in azure provider (msgraphfs vs KiotaRequestAdapter)

Providers

### Apache Airflow version

3.2.2

Microsoft Graph Filesystem fails to initialise properly with `get_fs()` due to `scope` OAuth parameter being missing. The connection template defines a default `scopes` value, but no `scope`.

It should be possible to initialise the file system. Manually adding `scope` parameter with the same value as existing `scopes` one allows `get_fs()` function to work properly.

### Operating System

Ubuntu 22.04

### Deployment

Virtualenv installation

microsoft-azure

14.0.0

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

The issue was likely masked by https://github.com/apache/airflow/issues/69463, missing `token_endpoint` was being hit before missing `scope` was being reached.

### Code of Conduct

haseebmalik18:
Fixes the Microsoft Graph filesystem failing to authenticate when the connection is set up through the UI.

The connection form only exposes a "Scopes" field, but get_fs() hands MSGDriveFS a fully built oauth2_client_params dict. That makes the library skip its own scope default, so authlib ends up with no scope and the connection fails. This falls back to the connection's scopes value, then to the Graph default, so a UI-configured connection works without hand-adding a scope key to extras.

Scope resolution is centralised in `_get_scopes()`, which accepts both comma-separated and whitespace-separated strings as well as a list, so the certificate-auth and client-secret paths resolve scopes identically.

closes: #70822

---


potiuk:
Good diagnosis — the connection form does expose a `Scopes` field (`hooks/msgraph.py:258`, defaulting to `DEFAULT_SCOPE`), but `get_fs` built `oauth2_client_params` without a `scope` key, so `MSGDriveFS` skipped its own default and authlib was left with nothing. A UI-configured connection could not authenticate.

The precedence you chose is right — explicit `scope`, then the form's `scopes`, then the Graph default — and parametrizing over all three branches is the right way to pin it.

One thing to fix before this goes in: the two consumers of that same field disagree on how it is delimited.

`MSGraphAsyncOperator`'s hook treats `scopes` as comma-separated and splits it:

```python
scopes = config.get("scopes", self.scopes)
if isinstance(scopes, str):
    scopes = scopes.split(",")
```

This change passes the value through verbatim into the OAuth2 `scope` parameter, which is space-delimited by spec. So a connection configured the way the hook expects — `"User.Read,Files.Read"` — reaches authlib as the single literal string `"User.Read,Files.Read"`, which is one malformed scope rather than two. Single-scope values work, which is why the current tests pass; they only ever use one value.

That is the same shape as the bug being fixed here: one field, two consumers, different expectations.

Something like this keeps both readings working:

```python
scopes = get_field(conn_id=conn_id, conn_type=conn_type, extras=extras, field_name="scopes")
if isinstance(scopes, str):
    scopes = " ".join(s.strip() for s in scopes.split(",") if s.strip())
oauth2_client_params["scope"] = scopes or DEFAULT_SCOPE
```

Worth a fourth parametrize case with a comma-separated value so the behaviour is pinned rather than assumed.

Minor, while you're in there: `DEFAULT_SCOPE` is redefined here as a module constant, but `KiotaRequestAdapterHook.DEFAULT_SCOPE` already holds the same literal. Importing it would keep the two from drifting apart later.

---

haseebmalik18:
@potiuk

haseebmalik18:
@potiuk Following up

dabla (approved):
LGTM as @potiuk remarks have been addressed well done!

dabla:
Due to the merge of your other [PR](https://github.com/apache/airflow/pull/71362), merge conflicts must be resolved.

haseebmalik18:
Merge conflicts resolved

potiuk (commented):
Thanks for the follow-up — the premise checks out: `hooks/msgraph.py` splits the form's `scopes` on commas and the connection form only exposes `scopes`, so a UI-configured connection really did end up scope-less. The fix solves that case. Two edges I'd like tightened before it lands, plus a test gap — none of them a problem with the approach.

### `scopes` given as a list now reaches authlib as a list (`fs/msgraph.py:168`)

The code this replaces went through `_get_scopes()`, which explicitly supports `list[str]`, and the caller joined the result into a string. The new block only stringifies when `isinstance(scopes, str)`, so a list — reachable via `storage_options={"scopes": ["User.Read", "Files.Read"]}`, and a shape `_get_scopes()` documents as supported — is now assigned to `oauth2_client_params["scope"]` unchanged. That is a regression on an input that worked before this PR.

### The certificate-auth branch keeps the bug this PR fixes (`fs/msgraph.py:137`)

`"scope": " ".join(_get_scopes(options))` is unchanged, and `_get_scopes()` splits on whitespace only. The very same UI connection with `scopes = "User.Read,Files.Read"` still yields the single bogus scope `User.Read,Files.Read` when certificate auth is used. Same root cause, same user-visible failure — worth fixing in the same PR rather than leaving half the code path broken.

Both fall out if the comma handling moves into `_get_scopes()` instead of being duplicated at the call site:

```python
def _get_scopes(options: dict[str, Any]) -> list[str]:
    scopes = options.get("scope") or options.get("scopes") or DEFAULT_SCOPE
    if isinstance(scopes, str):
        return [scope.strip() for scope in scopes.replace(",", " ").split() if scope.strip()]
    return scopes
```

…and the new block collapses to:

```python
if "scope" not in oauth2_client_params:
    oauth2_client_params["scope"] = " ".join(_get_scopes(options))
```

That keeps list support, fixes the certificate branch and `_get_certificate_token()`'s `credential.get_token(*_get_scopes(options))` at the same time, and leaves one place where scope resolution lives.

### Smaller observations

- The new parametrized test covers the string forms nicely, but not the list-valued `scopes` case above, and not the certificate branch. If you take the `_get_scopes()` route, a case per branch would keep both covered.

---

>

---

providers/microsoft/azure/src/airflow/providers/microsoft/azure/fs/msgraph.py
```diff
@@ -158,8 +158,14 @@ def get_fs(conn_id: str | None, storage_options: dict[str, Any] | None = None) -
...
-        if "scopes" in options and "scope" not in oauth2_client_params:
-            oauth2_client_params["scope"] = " ".join(_get_scopes(options))
+        # authlib expects a singular, space-delimited "scope"; the connection form only
+        # offers "scopes", which the hook treats as comma-separated, so translate it and
+        # always default so authlib never authenticates without a scope.
+        if "scope" not in oauth2_client_params:
+            scopes = options.get("scopes")
+            if isinstance(scopes, str):
+                scopes = " ".join(scope.strip() for scope in scopes.split(",") if scope.strip())
+            oauth2_client_params["scope"] = scopes or DEFAULT_SCOPE
```
potiuk: Small regression here: the code this replaces went through `_get_scopes()`, which supports `list[str]`, and the caller joined it into a string. This branch only stringifies when `isinstance(scopes, str)`, so `storage_options={"scopes": ["User.Read", "Files.Read"]}` now assigns a *list* to `oauth2_client_params["scope"]` and authlib gets something it cannot use.

Moving the comma handling into `_get_scopes()` keeps list support and fixes the certificate branch at line 137 (which still has the comma bug) at the same time:

```suggestion
        # authlib expects a singular, space-delimited "scope"; the connection form only
        # offers "scopes", which the hook treats as comma-separated, so translate it and
        # always default so authlib never authenticates without a scope.
        if "scope" not in oauth2_client_params:
            oauth2_client_params["scope"] = " ".join(_get_scopes(options))
```

(with `_get_scopes()` extended to `scopes.replace(",", " ").split()` for the `str` case).

providers/microsoft/azure/tests/unit/microsoft/azure/fs/test_msgraph.py
```diff
@@ -252,6 +253,36 @@ def test_get_fs_with_certificate_data_from_storage_options(
         )
         assert result == mock_fs_instance
 
+    @pytest.mark.parametrize(
```
potiuk, line 285: Nice coverage of the string forms. Two cases missing: `scopes` passed as a list (see the comment on `msgraph.py`), and the certificate-auth branch, which resolves its scope separately at `msgraph.py:137` and is still comma-broken.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/fs/msgraph.py
```diff
@@ -36,7 +36,7 @@ def _get_token_endpoint(tenant_id: str) -> str:
 def _get_scopes(options: dict[str, Any]) -> list[str]:
     scopes = options.get("scope") or options.get("scopes") or DEFAULT_SCOPE
     if isinstance(scopes, str):
-        return scopes.split()
+        return scopes.replace(",", " ").split()
```
Miretpl:
```suggestion
        return [scope.strip() for scope in scopes.replace(",", " ").split() if scope.strip()]
```
The version proposed by Jarek is safer, as there will be no empty strings returned by the function in a case when the user would provide, e.g. "User.Read, Files.Read" value in connection.

Miretpl (changes requested):

haseebmalik18:
Made requested changes

Miretpl (approved):

Miretpl (changes requested):
Code looks good. Could you update PR description to follow https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions?

haseebmalik18:
Updated

Miretpl (approved):

PR #70883, opened 2026-08-01
Point mask_secret deprecation warning at the module it actually uses

### LINKED ISSUE #70875: Deprecation warning for airflow.sdk.execution_time.secrets_masker points at the wrong module for mask_secret
### Apache Airflow version

3.2.2 (apache-airflow-task-sdk 1.2.2). Also present on `main` at 79fa2e622218138f1638fe3a225156e254b3e748.

_No response_

### What happened?

The deprecation shim at `airflow.sdk.execution_time.secrets_masker` tells you to migrate to `airflow.sdk._shared.secrets_masker`. For `mask_secret` its own code deliberately does something else, and the two are not equivalent: `airflow.sdk.log.mask_secret` also forwards the secret to the supervisor, `airflow.sdk._shared.secrets_masker.mask_secret` does not. Following the warning silently drops supervisor masking.

The warning says `_shared` ([L33-L37](https://github.com/apache/airflow/blob/79fa2e622218138f1638fe3a225156e254b3e748/task-sdk/src/airflow/sdk/execution_time/secrets_masker.py#L33-L37)):

```python
warnings.warn(
    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be "
    "removed in a future version. Please use 'airflow.sdk._shared.secrets_masker' instead.",
    DeprecatedImportWarning,
    stacklevel=2,
)
```

The code hands you `airflow.sdk.log` ([L43-L47](https://github.com/apache/airflow/blob/79fa2e622218138f1638fe3a225156e254b3e748/task-sdk/src/airflow/sdk/execution_time/secrets_masker.py#L43-L47)):

```python
def __getattr__(name: str):
    if name == "mask_secret":
        from airflow.sdk.log import mask_secret

        return mask_secret
```

Confirmed by object identity on 3.2.2:

```
shim.mask_secret is airflow.sdk.log.mask_secret                    -> True
shim.mask_secret is airflow.sdk._shared.secrets_masker.mask_secret -> False
```

That special case is correct, and it is load bearing. `airflow.sdk.log.mask_secret` registers with the shared masker **and** notifies the supervisor ([log.py L250-L266](https://github.com/apache/airflow/blob/79fa2e622218138f1638fe3a225156e254b3e748/task-sdk/src/airflow/sdk/log.py#L250-L266)):

```python
def mask_secret(secret: JsonValue, name: str | None = None) -> None:
    """Mask a secret in both task process and supervisor process. ..."""
    _secrets_masker().add_mask(secret, name)

    with suppress(Exception):
        from airflow.sdk.execution_time import task_runner
        from airflow.sdk.execution_time.comms import MaskSecret

        if comms := getattr(task_runner, "SUPERVISOR_COMMS", None):
            comms.send(MaskSecret(value=secret, name=name))
```

The `_shared` one the warning recommends stops at the first line ([secrets_masker.py L92-…](https://github.com/apache/airflow/blob/79fa2e622218138f1638fe3a225156e254b3e748/shared/secrets_masker/src/airflow_shared/secrets_masker/secrets_masker.py#L92)); there is no `SUPERVISOR_COMMS` reference anywhere in that module. Under AIP-72 the supervisor is a separate process, so a secret registered only through `_shared` can still appear in supervisor log output.

The failure is silent. Both modules import, both expose `mask_secret`, both accept the call and return `None`. A test asserting the symbol is importable and callable passes against either one, including the one that misses the supervisor.

The warning text and the module docstring should name `airflow.sdk.log` for `mask_secret`, matching what `__getattr__` already does. Either carve out the exception in the message, or drop the special case so the message becomes true.

The shim came in with https://github.com/apache/airflow/pull/54915, added defensively after https://github.com/apache/airflow/pull/54449 removed the path, so this reads as incidental rather than intended.

### How to reproduce

```python
import warnings
warnings.simplefilter("ignore")

import airflow.sdk.execution_time.secrets_masker as shim
import airflow.sdk._shared.secrets_masker as shared
import airflow.sdk.log as sdklog

print(shim.mask_secret is sdklog.mask_secret)  # True, what the shim gives you
print(shim.mask_secret is shared.mask_secret)  # False, what the warning tells you to use
```

### Operating System

macOS 15 (not OS specific)

_No response_

### Deployment

Other

### Deployment details

_No response_

### Anything else?

Two engineers migrating the same codebase independently read the deprecation message, moved to `airflow.sdk._shared.secrets_masker`, and ended up with weaker masking than they started with. Neither noticed, because every wrong answer in this area imports and runs fine. The human readable guidance is the only signal most people act on, and right now it points away from the implementation the shim itself chose.

### Code of Conduct

bujjibabukatta:
**Summary**
Deprecation warning for airflow.sdk.execution_time.secrets_masker tells
everyone to switch to airflow.sdk._shared.secrets_masker — but for
mask_secret specifically that's wrong and silently weakens masking if
followed.

**Root cause**
__getattr__ deliberately resolves mask_secret from airflow.sdk.log instead
of _shared, since airflow.sdk.log.mask_secret also notifies the supervisor
process while the _shared version doesn't. The warning text never carved
out this exception.

**Fix**
Updated the warning and docstring to name airflow.sdk.log for mask_secret.
__getattr__ itself is unchanged. Added tests asserting object identity and
warning-message content.

closes: #70875

- [X] Yes - Claude


task-sdk/src/airflow/sdk/execution_time/secrets_masker.py
```diff
@@ -19,8 +19,14 @@
...
-from airflow.sdk.execution_time.secrets_masker. The actual implementation
-has been moved to airflow.sdk._shared.secrets_masker.
+from airflow.sdk.execution_time.secrets_masker. Most of the implementation
+has been moved to airflow.sdk._shared.secrets_masker, with one deliberate
+exception: ``mask_secret`` is re-exported from airflow.sdk.log instead.
+airflow.sdk.log.mask_secret masks the secret locally AND notifies the
+supervisor process; airflow.sdk._shared.secrets_masker.mask_secret only
+does the former, so importing 'mask_secret' from the _shared module instead
+would silently weaken masking under AIP-72 (the supervisor runs as a
+separate process).
```
amoghrajesh:
```suggestion
from airflow.sdk.execution_time.secrets_masker. Most of the implementation
has been moved to airflow.sdk._shared.secrets_masker, except ``mask_secret``,
which is re-exported from airflow.sdk.log instead.
```

task-sdk/src/airflow/sdk/execution_time/secrets_masker.py
```diff
@@ -31,8 +37,10 @@
 from airflow.utils.deprecation_tools import DeprecatedImportWarning
 
 warnings.warn(
-    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be removed in a future version. "
-    "Please use 'airflow.sdk._shared.secrets_masker' instead.",
+    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be removed in a future "
+    "version. Please use 'airflow.sdk._shared.secrets_masker' instead, EXCEPT for 'mask_secret': that should be "
+    "imported from 'airflow.sdk.log' instead, since 'airflow.sdk._shared.secrets_masker.mask_secret' does not "
+    "notify the supervisor process and using it will silently weaken secret masking.",
```
amoghrajesh:
```suggestion
    "imported from 'airflow.sdk.log' instead."
```

task-sdk/tests/task_sdk/execution_time/test_secrets_masker.py
```diff
@@ -0,0 +1,75 @@
...
+            importlib.import_module("airflow.sdk.execution_time.secrets_masker")
+
+        messages = [str(w.message) for w in record.list]
+        combined = " ".join(messages)
+
+        assert "airflow.sdk.log" in combined
+        assert "mask_secret" in combined
+        # The blanket "_shared" recommendation must not be presented as unconditionally
+        # true — it must be scoped to say it doesn't apply to mask_secret.
+        assert "_shared.secrets_masker" in combined
```
amoghrajesh, line 46: This is the only valid test, please trim down the docstring shorter and remove the comments.

The other tests exercise whats already present...remove them

amoghrajesh (changes requested):
Requesting changes to avoid accidental merge

bujjibabukatta:
@amoghrajesh Thanks for the review, appreciate you taking the time!

Done — trimmed the docstring and warning message, and removed the tests
exercising pre-existing __getattr__ behavior. Only the warning-content
test remains.

task-sdk/src/airflow/sdk/execution_time/secrets_masker.py
```diff
@@ -31,8 +32,9 @@
 from airflow.utils.deprecation_tools import DeprecatedImportWarning
 
 warnings.warn(
-    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be removed in a future version. "
-    "Please use 'airflow.sdk._shared.secrets_masker' instead.",
+    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be removed in a future "
+    "version. Please use 'airflow.sdk._shared.secrets_masker' instead, EXCEPT for 'mask_secret': that should be "
+    "imported from 'airflow.sdk.log' instead.",
```
kaxil: Nit, non-blocking: two "instead"s in one sentence, and the all-caps EXCEPT reads a bit shouty. Could be said in one pass:

```suggestion
    "Importing from 'airflow.sdk.execution_time.secrets_masker' is deprecated and will be removed in a future "
    "version. Please import 'mask_secret' from 'airflow.sdk.log' and everything else from "
    "'airflow.sdk._shared.secrets_masker'.",
```

kaxil (approved):

bujjibabukatta:
Hi @kaxil 

Good catch, thanks — much cleaner without the repeated "instead" and the
caps. Updated with your exact wording.

1fanwang (approved):
lgtm overall thanks for picking this up! @bujjibabukatta

bujjibabukatta:
@kaxil @1fanwang Thanks for approval. could you please merge pull request ?

bujjibabukatta:
@amoghrajesh @kaxil @1fanwang  Could you please review and merge pull request .

PR #70884, opened 2026-08-01
Refuse only ids whose leading segment could name a team

potiuk:
Follow-up to [#70736](https://github.com/apache/airflow/pull/70736), which merged
with a refusal rule that is broader than it needs to be.

That change made the team agnostic lookup refuse any id of the form
`_<x>___<y>`, on the grounds that such an id could name a team's namespace. That
is true of *some* of them, not all: team names are validated as
`^[a-zA-Z0-9_-]{3,50}$`, so a one- or two-character leading segment cannot be a
team. A team agnostic secret whose id merely looks namespaced — `_a___b`,
`_ab___x` — is currently refused even though no team namespace can be spelled
that way, which blocks a legitimate lookup for no benefit.

### Approach

Test the **leading segment against the team name rule** rather than testing
whether the id merely contains the separator:

```python
for i in ...:                      # every position of the separator
    team, rest = secret_id[1:i], secret_id[i + 3:]
    if rest and _TEAM_NAME.fullmatch(team):
        return True
```

Every id that does spell a real team's namespace still has a valid team name in
that position by construction, so nothing reachable is let through. Ids that
cannot name a team resolve normally again.

Every split is considered, because a team name may itself contain the separator —
`_a___b___c` is both team `a` with id `b___c` and team `a___b` with id `c` — and
one plausible team name is enough to refuse. Overlapping separator positions are
included, so `_abc____x` is tested at both offsets.

**`teams sync` now enforces the team name rule.** This is what makes the above
sound: the guard assumes a stored team name is a valid one. `airflow teams sync`
creates teams from the dag bundle config and did not validate the names at all,
unlike `teams create` — so an unvalidated short name such as `a` would make the
test miss and reopen the cross-team read that #70736 closed. The pattern is
factored into `TEAM_NAME_PATTERN` and applied on both creation paths, with a
comment on the sync path recording why the secrets backend depends on it.

### Behaviour change

- Ids whose leading segment cannot be a team name resolve through the team
  agnostic lookup again, for every caller scope. Ids that could name a team stay
  refused exactly as #70736 made them.
- `airflow teams sync` now fails with a clear message when the dag bundle config
  contains a team name that does not match `^[a-zA-Z0-9_-]{3,50}$`, where it
  previously created the team silently.

### Test plan

- [x] `test_secrets_environment_variables.py` — 46 cases pass
- [x] New `test_id_that_cannot_name_a_team_is_still_resolved` — `_a___b` and
      `_ab___x` resolve for every caller scope including none; it fails against the
      currently merged rule
- [x] The existing cross-team tests are unchanged and still pass, including
      `test_team_whose_name_extends_the_callers_is_not_readable`
- [x] `test_team_command.py` — 21 cases pass with the shared pattern
- [x] Exhaustive property check over six team names chosen to collide (`team_a` /
      `team_a___prod`, `abc` / `abc___b`), both lookups, every caller scope, bare and
      namespaced ids: **0 cross-team resolutions**

### DIFF
--- a/airflow-core/src/airflow/cli/commands/team_command.py
+++ b/airflow-core/src/airflow/cli/commands/team_command.py
@@ -48,13 +48,16 @@ def _show_teams(teams, output):
     )
 
 
+TEAM_NAME_PATTERN = r"^[a-zA-Z0-9_-]{3,50}$"
+
+
 def _extract_team_name(args):
     """Extract and validate team name from args."""
     team_name = args.name.strip()
     if not team_name:
         raise SystemExit("Team name cannot be empty")
-    if not re.match(r"^[a-zA-Z0-9_-]{3,50}$", team_name):
-        raise SystemExit("Invalid team name: must match regex ^[a-zA-Z0-9_-]{3,50}$")
+    if not re.match(TEAM_NAME_PATTERN, team_name):
+        raise SystemExit(f"Invalid team name: must match regex {TEAM_NAME_PATTERN}")
     return team_name
 
 
@@ -168,6 +171,18 @@ def team_sync(args, *, session=NEW_SESSION):
         for bundle in DagBundlesManager()._bundle_config.values()
         if bundle.team_name is not None
     }
+
+    # The bundle config is a second creation path for teams, so it has to enforce the same
+    # name rule as `teams create`. Other code relies on a stored team name being a valid one
+    # -- the environment secrets backend decides whether a supplied secret id could name a
+    # team namespace by testing the leading segment against this pattern, and an unvalidated
+    # short name such as "a" would make that test miss.
+    invalid = sorted(name for name in dag_bundle_teams if not re.match(TEAM_NAME_PATTERN, name))
+    if invalid:
+        raise SystemExit(
+            f"Invalid team name(s) in the dag bundle config: {', '.join(invalid)}. "
+            f"Team names must match regex {TEAM_NAME_PATTERN}."
+        )
 
     teams_added = 0
 
--- a/airflow-core/src/airflow/secrets/environment_variables.py
+++ b/airflow-core/src/airflow/secrets/environment_variables.py
@@ -27,6 +27,15 @@ from airflow.secrets import BaseSecretsBackend
 CONN_ENV_PREFIX = "AIRFLOW_CONN_"
 VAR_ENV_PREFIX = "AIRFLOW_VAR_"
 
+# Separator between the team name and the secret id in a team namespaced
+# environment variable name: AIRFLOW_CONN__<TEAM>___<ID>.
+TEAM_SEP = "___"
+
+# What ``airflow teams create`` accepts as a team name, so what a team
+# namespace can actually be spelled with.
+# Kept in sync with ``airflow.cli.commands.team_command``.
+_TEAM_NAME = re.compile(r"[a-zA-Z0-9_-]{3,50}")
+
 
 class EnvironmentVariablesBackend(BaseSecretsBackend):
     """Retrieves Connection object and Variable from environment variable."""
@@ -65,25 +74,36 @@ class EnvironmentVariablesBackend(BaseSecretsBackend):
     @staticmethod
     def _names_a_team_namespace(secret_id: str) -> bool:
         """
-        Whether ``secret_id`` spells out a team namespaced environment variable name.
+        Whether ``secret_id`` could spell out a team namespaced environment variable name.
 
         A team specific secret lives in the ``_<TEAM_NAME>___<SECRET_ID>`` namespace of the
-        environment. An id of that shape therefore makes the team agnostic lookup -- which
-        prepends only ``AIRFLOW_CONN_`` / ``AIRFLOW_VAR_`` -- land inside some team's namespace,
-        so that lookup is refused for such an id.
+        environment. An id of that shape makes the team agnostic lookup -- which prepends only
+        ``AIRFLOW_CONN_`` / ``AIRFLOW_VAR_`` -- land inside some team's namespace, so that
+        lookup is refused for such an id.
+
+        The test is whether the leading segment **could be a team name**, not merely whether
+        the id contains the separator. Team names are validated on creation, so an id like
+        ``_a___b`` cannot name a team namespace -- ``a`` is too short to be a team -- and
+        refusing it would block a legitimate team agnostic secret for no benefit. Every id
+        that does spell a real team's namespace has a valid team name in that position by
+        construction, so nothing reachable is let through.
 
         **The id is never attributed to a particular team**, because it cannot be: a team name
-        may itself contain the ``___`` separator, so ``_a___b___c`` is both team ``a`` with id
-        ``b___c`` and team ``a___b`` with id ``c``, and nothing in the string distinguishes them.
-        Only the caller's own namespace is ever constructed, never parsed.
-
-        This is why the check is not "does the id belong to some team other than the caller's".
-        Comparing the id against the prefix the caller's team builds looks equivalent and is not:
-        for a caller in team ``a`` the id ``_a___b___c`` starts with ``_A___``, yet the variable
-        it resolves, ``AIRFLOW_CONN__A___B___C``, is team ``a___b``'s. Treating a prefix match as
-        ownership hands one team the secrets of every team whose name extends it. Callers reach
-        their own team's secrets with the bare id plus their team scope -- handled above, and safe
-        because that path can only ever build the caller's own namespace -- so nothing legitimate
-        needs a namespaced id here.
+        may itself contain the separator, so ``_a___b___c`` is both team ``a`` with id
+        ``b___c`` and team ``a___b`` with id ``c``. Every split is therefore considered, and
+        one plausible team name is enough to refuse. Comparing the id against the prefix the
+        caller's own team builds looks equivalent and is not: for a caller in team ``a`` the
+        id ``_a___b___c`` starts with ``_A___``, yet the variable it resolves,
+        ``AIRFLOW_CONN__A___B___C``, is team ``a___b``'s. Treating a prefix match as ownership
+        hands one team the secrets of every team whose name extends it.
         """
-        return re.fullmatch(r"_.+___.+", secret_id) is not None
+        if not secret_id.startswith("_"):
+            return False
+        # Overlapping positions matter: ``_abc____x`` splits at both index 4 and 5.
+        for i in range(1, len(secret_id) - len(TEAM_SEP) + 1):
+            if secret_id[i : i + len(TEAM_SEP)] != TEAM_SEP:
+                continue
+            team, rest = secret_id[1:i], secret_id[i + len(TEAM_SEP) :]
+            if rest and _TEAM_NAME.fullmatch(team):
+                return True
+        return False
--- a/airflow-core/tests/unit/always/test_secrets_environment_variables.py
+++ b/airflow-core/tests/unit/always/test_secrets_environment_variables.py
@@ -99,6 +99,22 @@ class TestEnvironmentVariablesBackendTeamScope:
         assert lookup(env_prefix, method, team_scoped_id(team_name), team_name) is None
 
     @pytest.mark.parametrize(("env_prefix", "method"), LOOKUPS)
+    @pytest.mark.parametrize("secret_id", ["_a___b", "_ab___x"])
+    @pytest.mark.parametrize("team_name", [None, *TEAM_NAMES])
+    def test_id_that_cannot_name_a_team_is_still_resolved(
+        self, monkeypatch, env_prefix, method, secret_id, team_name
+    ):
+        """Refusing every ``_x___y`` id would block legitimate team agnostic secrets.
+
+        Team names are validated as ``^[a-zA-Z0-9_-]{3,50}$``, so a one- or two-character
+        leading segment cannot be a team. Such an id names no team's namespace and must stay
+        resolvable through the team agnostic lookup, for any caller.
+        """
+        monkeypatch.setenv(env_prefix + secret_id.upper(), GLOBAL_VALUE)
+
+        assert lookup(env_prefix, method, secret_id, team_name) == GLOBAL_VALUE
+
+    @pytest.mark.parametrize(("env_prefix", "method"), LOOKUPS)
     def test_team_whose_name_extends_the_callers_is_not_readable(self, monkeypatch, env_prefix, method):
         """A team name may contain the ``___`` separator, so one team's namespace can start with another's.
 


airflow-core/src/airflow/secrets/environment_variables.py
```diff
@@ -65,25 +74,36 @@ def get_variable(self, key: str, team_name: str | None = None) -> str | None:
...
-        their own team's secrets with the bare id plus their team scope -- handled above, and safe
-        because that path can only ever build the caller's own namespace -- so nothing legitimate
-        needs a namespaced id here.
+        may itself contain the separator, so ``_a___b___c`` is both team ``a`` with id
+        ``b___c`` and team ``a___b`` with id ``c``. Every split is therefore considered, and
+        one plausible team name is enough to refuse. Comparing the id against the prefix the
+        caller's own team builds looks equivalent and is not: for a caller in team ``a`` the
+        id ``_a___b___c`` starts with ``_A___``, yet the variable it resolves,
+        ``AIRFLOW_CONN__A___B___C``, is team ``a___b``'s. Treating a prefix match as ownership
+        hands one team the secrets of every team whose name extends it.
```
shahar1, line 98: Deleting the old "that path can only ever build the caller's own namespace" sentence is the right call — it was false. But the gap it papered over is still open, and it's the same cross-team read #70736 closed, just spelled differently.

A caller in `team_a` supplying the **bare** id `prod___dbconn` never reaches this guard: the scoped branch above returns first, and the id has no leading `_` anyway.

```python
# caller: get_conn_value("prod___dbconn", team_name="team_a")
f"{CONN_ENV_PREFIX}_{team_name.upper()}___" + conn_id.upper()
# -> AIRFLOW_CONN__TEAM_A___PROD___DBCONN

# team "team_a___prod" stores "dbconn" as
# "AIRFLOW_CONN_" + "_TEAM_A___PROD" + "___" + "DBCONN"
# -> AIRFLOW_CONN__TEAM_A___PROD___DBCONN     <- identical
```

I checked; the strings are byte-identical. `test_team_whose_name_extends_the_callers_is_not_readable` uses this exact pair of team names but only the namespaced spelling.

Pre-existing rather than introduced here, so not a blocker on this diff — but worth a follow-up, and the root cause is shared with the loop below: `___` isn't reserved, so `_<TEAM>___<ID>` doesn't parse uniquely. Forbidding `___` inside a team name would make it unambiguous, collapse this loop into one exact split, and close the scoped-path collision too.

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -169,6 +172,18 @@ def team_sync(args, *, session=NEW_SESSION):
         if bundle.team_name is not None
     }
 
+    # The bundle config is a second creation path for teams, so it has to enforce the same
+    # name rule as `teams create`. Other code relies on a stored team name being a valid one
+    # -- the environment secrets backend decides whether a supplied secret id could name a
+    # team namespace by testing the leading segment against this pattern, and an unvalidated
+    # short name such as "a" would make that test miss.
+    invalid = sorted(name for name in dag_bundle_teams if not re.match(TEAM_NAME_PATTERN, name))
```
shahar1, line 180: This validates the names in the Dag bundle config, but the secrets guard depends on every **stored** team name being valid — and `teams sync` shipped in 3.3.0 with no validation at all. A live 3.3.0 deployment can already hold a team named `ab`, and after the companion change `_ab___x` resolves through the team agnostic lookup into `AIRFLOW_CONN__AB___X`, that team's secret. That's the read #70736 closed, and `_ab___x` is one of the two ids the new test asserts must resolve.

The config-side check catches this only while the short name is still in the config; a team row created by an earlier sync and since dropped from the config is missed. The stored names are in hand one line later (`Team.get_all_team_names(session=session)`), so covering them too is nearly free — hard error, or a warning if failing on pre-existing data is too aggressive for a patch release.

Related, given the `backport-to-v3-3-test` label: on a 3.3.x patch upgrade, a deployment whose bundle config carries a short team name will see `airflow teams sync` start exiting non-zero where it previously succeeded. Defensible, but better as a deliberate call than a side effect.

Separately — `re.match` with a `$`-anchored pattern accepts a trailing newline, and unlike `teams create` this path has no `.strip()`. See the note on `_TEAM_NAME` in `environment_variables.py`.

airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -169,6 +172,18 @@ def team_sync(args, *, session=NEW_SESSION):
...
+    if invalid:
+        raise SystemExit(
+            f"Invalid team name(s) in the dag bundle config: {', '.join(invalid)}. "
+            f"Team names must match regex {TEAM_NAME_PATTERN}."
+        )
```
shahar1, line 185: No test covers this new behaviour — the diff touches only `test_secrets_environment_variables.py`, and `test_team_sync` is unchanged and happy-path only.

> **Flag any changed or added behaviour without a corresponding test, and flag tests that a reviewer cannot fail by reverting the PR's change.** The target is exactly 100% coverage of what the PR changes — no more, no less.
>
> — [`.github/instructions/code-review.instructions.md` § Testing Requirements](https://github.com/apache/airflow/blob/main/.github/instructions/code-review.instructions.md)

The PR body's "21 cases pass with the shared pattern" confirms nothing regressed, which isn't the same thing: delete this `if invalid:` block and the suite stays green. Since this guard is what makes the secrets relaxation sound, it's the behaviour here that most needs pinning — one case with `"team_name": "a"` in the bundle config asserting `SystemExit` and that no team row is created.

airflow-core/src/airflow/secrets/environment_variables.py
```diff
@@ -27,6 +27,15 @@
...
 VAR_ENV_PREFIX = "AIRFLOW_VAR_"
 
+# Separator between the team name and the secret id in a team namespaced
+# environment variable name: AIRFLOW_CONN__<TEAM>___<ID>.
+TEAM_SEP = "___"
+
+# What ``airflow teams create`` accepts as a team name, so what a team
+# namespace can actually be spelled with.
+# Kept in sync with ``airflow.cli.commands.team_command``.
+_TEAM_NAME = re.compile(r"[a-zA-Z0-9_-]{3,50}")
```
shahar1, line 37: These two spellings aren't equivalent, which undercuts the "kept in sync" promise: `$` also matches *before a trailing newline*, so

```python
re.match(r"^[a-zA-Z0-9_-]{3,50}$", "team1\n")   # match
_TEAM_NAME.fullmatch("team1\n")                  # None
```

(verified). `teams create` is saved by its `.strip()`; the new `teams sync` path has none, so a Dag bundle config with a trailing newline in `team_name` creates a team this guard cannot recognise. One *unanchored* pattern plus `re.fullmatch` at both sites removes the divergence entirely.

Also: a `# Kept in sync with ...` comment is the kind of promise better held by a one-line test asserting the two constants are equal — otherwise drift silently reopens the hole instead of failing CI.

Minor: the comment above now understates the coupling — after this PR the pattern is what `teams create` **and** `teams sync` accept.

shahar1 (commented):
The relaxation itself is sound and the reasoning in the docstring is genuinely good — but it rests on an invariant ("a stored team name is a valid one") that isn't true for deployments upgrading from 3.3.0, the new `teams sync` guard that establishes it has no test, and while reading the same function I found the *original* cross-team read is still reachable by a different spelling.

### The team-scoped path still leaks across teams (`environment_variables.py:44`, `:62`)

Pre-existing — not introduced here — but this PR rewrites the docstring that reasons about exactly this case, and correctly deletes the sentence claiming the scoped path "can only ever build the caller's own namespace". Nothing replaces it, and the claim was false:

```python
# caller in team "team_a", asking for the BARE id "prod___dbconn"
os.environ.get(f"{CONN_ENV_PREFIX}_{team_name.upper()}___" + conn_id.upper())
# -> "AIRFLOW_CONN__TEAM_A___PROD___DBCONN"

# team "team_a___prod" stores its "dbconn" secret as
# "AIRFLOW_CONN_" + "_TEAM_A___PROD" + "___" + "DBCONN"
# -> "AIRFLOW_CONN__TEAM_A___PROD___DBCONN"     # <- identical
```

I confirmed the two strings are byte-identical. The id carries no leading `_`, so `_names_a_team_namespace` would not fire even if it were consulted — and it isn't, because the scoped branch returns first. `test_team_whose_name_extends_the_callers_is_not_readable` uses this exact pair of team names but only exercises the namespaced spelling `_team_a___prod___dbconn`; the bare spelling walks straight through.

Both this and the every-split loop below come from the same root: `___` is not a reserved character, so `_<TEAM>___<ID>` does not parse uniquely. Forbidding `___` inside a team name makes the namespace unambiguous, collapses the new loop into a single exact split, and closes the scoped-path collision — at the cost of a rename path for any existing team already containing `___`. Worth considering as the follow-up rather than layering another heuristic.

### The invariant isn't retroactive (`team_command.py:180`)

`airflow teams sync` shipped in **3.3.0** without name validation, so a live 3.3.0 deployment can already hold a team named `ab`. After this change `_ab___x` resolves through the team agnostic lookup into `AIRFLOW_CONN__AB___X` — that team's secret `x`. That is precisely the read #70736 closed, and `_ab___x` is one of the two ids the new test asserts *must* resolve.

The new check validates the names in the Dag bundle config, which catches this only while the short name is still in the config; a team row created by an earlier sync and since dropped from the config is missed. The sync already has the stored names in hand one line later (`Team.get_all_team_names(session=session)`), so validating those too is nearly free — as a hard error, or a warning if failing on pre-existing data is too aggressive for a patch release.

Related, given the `backport-to-v3-3-test` label: on a 3.3.x patch upgrade a deployment whose Dag bundle config carries a short team name will see `airflow teams sync` start exiting non-zero where it previously succeeded. That is defensible, but it should be a deliberate call rather than a side effect.

### No test for the new `teams sync` validation (`team_command.py:181-185`)

> **Flag any changed or added behaviour without a corresponding test, and flag tests that a reviewer cannot fail by reverting the PR's change.** The target is exactly 100% coverage of what the PR changes — no more, no less.
>
> — [`.github/instructions/code-review.instructions.md` § Testing Requirements](https://github.com/apache/airflow/blob/main/.github/instructions/code-review.instructions.md)

The diff touches only `test_secrets_environment_variables.py`; `test_team_command.py` is unchanged, and its `test_team_sync` covers only the happy path. The PR body's "21 cases pass with the shared pattern" confirms the existing tests still pass, which is not the same thing — no test fails if the new `invalid` check is deleted. Given this guard is what makes the secrets relaxation sound, it is the one behaviour here that most needs pinning: one case with `"team_name": "a"` in the bundle config asserting `SystemExit` and that no team row is created.

### Smaller observations

- `environment_variables.py:37` — the two spellings of "the same" pattern are not equivalent. `$` also matches before a trailing newline, so `re.match(r"^[a-zA-Z0-9_-]{3,50}$", "team1\n")` is a match while `_TEAM_NAME.fullmatch("team1\n")` is not (verified). `teams create` is saved by its `.strip()`; the new sync path has none, so a Dag bundle config with a trailing newline in `team_name` creates a team the secrets guard cannot recognise. Using one *unanchored* pattern with `re.fullmatch` at both sites removes the divergence. A `# Kept in sync with ...` comment is also the kind of promise better held by a one-line test asserting the two constants are equal — otherwise drift silently reopens the hole instead of failing CI.
- `airflow-core/docs/core-concepts/multi-team.rst:517-550` — "Dag Bundle to Team Association" is where users actually set `team_name`, and the newly enforced naming rule isn't mentioned there. The existing note ("The team specified in ``team_name`` must exist in the database before syncing... Create teams first using ``airflow teams create``") already reads oddly next to a sync that creates them.
- `environment_variables.py:34` — the comment now understates the coupling: after this PR the pattern is what `teams create` **and** `teams sync` accept.
- Newsfragment — `airflow teams sync` now exits with an error on configs it previously accepted. That is user-facing CLI behaviour, more so with the 3.3 backport label, so `airflow-core/newsfragments/70884.bugfix.rst` looks warranted here rather than optional.

---

>

shahar1 (approved):

potiuk:
Superseded by #70886, which now targets the same 3.4.0 milestone.

@shahar1's review here found that the cross-team read is still open in a spelling
none of these guards reach: a caller in `team_a` asking for the **bare** id
`prod___dbconn` builds `AIRFLOW_CONN__TEAM_A___PROD___DBCONN`, byte-identical to
what team `team_a___prod` builds for id `dbconn`. The *scoped* lookup hits, so the
team agnostic fall-through — which is all this PR narrows — is never consulted. I
reproduced it; it returns the other team's secret.

The root cause he identified is that `___` is not reserved, so `_<TEAM>___<ID>`
does not parse uniquely, and every revision here has been an after-the-fact
ambiguity check. #70886 fixes that at the source by forbidding **consecutive**
underscores in a team name (relaxed from forbidding every underscore, so `team_a`
stays legal). With the separator unspellable the split becomes one `partition`,
and `team_a___prod` simply stops being a name a team can have.

His other two findings are folded in there as well:

- `teams sync` now validates **stored** names, not only the bundle config — it
  shipped in 3.3.0 with no validation, so an upgrading deployment can already hold
  a name the guard cannot recognise;
- the pattern is defined once and imported, unanchored and used with
  `re.fullmatch` throughout, with a test pinning that the CLI and backend agree —
  `re.match` against a `$`-anchored pattern accepts a trailing newline and the sync
  path has no `.strip()`.

This PR is now a draft on 3.4.0 with the backport label removed. Closing it in
favour of #70886 once that lands; leaving it open for now so the review thread
stays reachable.

Worth stating plainly: the relaxation here rested on team names being at least 3
characters, and `teams sync` never enforced that — so its new test asserting
`_ab___x` **must** resolve was pinning the vulnerable behaviour. Good catch.

airflow-core/src/airflow/cli/commands/team_command.py
potiuk: Answering the question directly, since you asked for it to be a deliberate call rather than a side effect: **hard error, and no backport.**

This PR is now a draft on 3.4.0 with `backport-to-v3-3-test` removed, and the work moved to #70886. So the "3.3.x patch upgrade starts exiting non-zero" case you flagged does not arise — nothing here ships in a patch release. Given that, the warning option loses its only real justification, and a warning would leave the deployment believing its teams are isolated when they are not.

Both halves are implemented in #70886: `teams sync` validates the **stored** names alongside the bundle-config ones (`Team.get_all_team_names` as you pointed out, one line down and nearly free), and it exits naming the offending names.

Your other two points are in there as well — one unanchored pattern used with `re.fullmatch` at both sites, defined once in the secrets backend and imported by the CLI, with `test_team_name_pattern_is_shared_with_the_secrets_backend` so drift fails CI rather than silently reopening the hole. You were right about `re.match` and the trailing newline; my first attempt at that test asserted `teams create` rejects `"team1\n"` and it does not, because `_extract_team_name` strips first. The exposure is on the sync path exactly as you said, and the test moved there.

And the finding at the top of this review — the bare-id spelling — I reproduced it. A caller in `team_a` asking for `prod___dbconn` returns `team_a___prod`'s secret. It is the reason this PR is held rather than fixed up: narrowing the team agnostic refusal cannot reach a read that the scoped lookup satisfies. #70886 closes it at the source by making `team_a___prod` unspellable, relaxed from forbidding every underscore to forbidding consecutive ones so `team_a` stays valid.

Worth being explicit about one consequence: this PR's own new test asserted `_ab___x` **must** resolve, and with an unvalidated stored team named `ab` that assertion pins the vulnerable behaviour. That is what your comment caught.

potiuk:
Superseding this in favour of #70886, which closes the same ambiguity structurally — narrowing `TEAM_NAME_PATTERN` to forbid doubled underscores means an id of the form `_<x>___<y>` can never name a team namespace, so the point-of-use refusal this PR refines is no longer needed. Thanks for the review here. Closing.

PR #70886, opened 2026-08-01
Require team names to be lower case with no doubled underscores

potiuk:
A team scoped secret lives in the `_<TEAM>___<SECRET_ID>` namespace of the environment,
and the secrets backend upper-cases the team name when it builds that variable. Which
names a team may have therefore decides whether that namespace is unambiguous.

Two things could make it ambiguous, and both are closed here by narrowing the rule
rather than by adding a check at the point of use:

```
TEAM_NAME_PATTERN = r"(?!.*__)[a-z0-9_-]{3,50}"

team_a          valid
a-b-c           valid
Data_Eng        invalid    upper case
team__x         invalid    two underscores in a row
team_a___prod   invalid    could contain the separator
```

- **Case.** `data_eng` and `Data_Eng` are two distinct teams that both resolve
  `AIRFLOW_CONN__DATA_ENG___<ID>`, so each reads the other's Connections and Variables.
  Whether the two rows can coexist at all was decided by the database's collation on a
  `String` primary key rather than by Airflow, so the same configuration behaved
  differently across backends.
- **The separator.** A name that can itself contain `___` cannot be told apart from the
  separator: `_a___b___c` reads as team `a` with id `b___c` *or* team `a___b` with id
  `c`. Forbidding `__` makes the first separator the only one.

### Scope, after the rebase

This PR previously also rewrote `EnvironmentVariablesBackend._names_a_team_namespace`.
#70902 and #71079 landed in the meantime: that check is now a plain
`TEAM_SEP in secret_id`, gated on `multi_team`, and it runs **before** the team scoped
lookup. That already closes the cross-team read this PR was originally about, and the
leading-segment refinement proposed here would only loosen it, so it is dropped. What
remains is the name rule and the migration onto it.

### Migrating deployments that already hold a rejected name

`teams create` accepted upper case, and `teams sync` validated nothing at all, so an
existing multi-team deployment can hold names the new rule rejects.

- **`airflow db migrate` lower-cases stored team names** (migration `c7f0a5d2e9b4`),
  carrying each team's Dag bundles, Connections, Variables, Pools and Triggers across.
  `team.name` is a primary key whose foreign keys do not cascade on update, so the
  rename adds the new row, repoints every referring row, and only then drops the old
  one. Under a case insensitive collation the two names are one value to the database,
  so there the row is renamed in place instead — that branch is picked by probing the
  collation, not by dialect name.
- **Two stored names that fold together stop the migration before it starts.** They
  would collapse onto one row and silently merge two teams; nothing can choose which
  name survives, so `_check_migration_errors` reports the pair and `db migrate` exits
  naming both. This is the first check registered on that hook.
- **`teams delete` no longer validates the name it is given.** Otherwise a team stored
  before the rule tightened could never be removed — including one an operator is being
  told to rename.

### `teams sync` validates stored names too

`teams sync` is a second creation path, so it holds incoming bundle-config names to the
same rule. It checks the already-stored names as well: checking only the incoming config
would miss a team created by an earlier sync and since dropped from it. The command
exits non-zero naming the offending names.

The rule is deliberately **unanchored** and always used with `re.fullmatch`: `re.match`
against a `$`-anchored pattern also accepts a trailing newline, and unlike `teams create`
the sync path does not strip what the bundle config gives it.

### Behaviour change

- `teams create` and `teams sync` reject an upper case name or one containing `__`.
  Single underscores are unaffected.
- `teams sync` exits non-zero when any stored or configured name violates the rule,
  where it previously accepted anything.
- `airflow db migrate` refuses to start while two stored names differ only in case.

### Test plan

- [x] `test_team_create_rejects_a_name_the_secrets_namespace_cannot_hold` — upper case,
      `team__x`, `team_a___prod`; asserts no row is created
- [x] `test_team_create_allows_a_single_underscore`
- [x] `test_team_delete_accepts_a_name_the_current_rule_rejects` — a stored `Data-Eng`
      is still removable
- [x] `test_team_sync_rejects_an_invalid_bundle_team_name` — too-short, upper case,
      `__`, and trailing-newline cases; asserts no team row is created
- [x] `test_team_sync_rejects_an_invalid_name_already_stored`
- [x] `test_check_team_names_can_be_lower_cased_passes_when_no_names_collide` /
      `..._reports_a_collision`
- [x] `test_rename_carries_referring_rows_with_the_team` — migration moves the team's
      Connection, Variable and Pool with it
- [x] 33 passed in `test_team_command.py`, 628 passed across `test_db.py`,
      `test_team_command.py` and `tests/unit/migrations/`
- [x] `prek --stage pre-commit` and `--stage manual` clean


airflow-core/src/airflow/cli/commands/team_command.py
```diff
@@ -69,6 +104,15 @@ def team_create(args, *, session=NEW_SESSION):
...
         raise SystemExit(f"Team with name '{team_name}' already exists")
 
+    collision = _case_collision(team_name, Team.get_all_team_names(session=session))
+    if collision:
+        raise SystemExit(
+            f"Team name '{team_name}' differs only in case from the existing team "
+            f"'{collision}'. They would share one secrets namespace, so each team could read "
+            f"the other's Connections and Variables. Choose a name that differs by more than "
+            f"case."
+        )
```
o-nikolas: At this point I think it would be easier to just enforce that team names are lowercase and only allow a few separator characters (maybe just `_` and `-`). Then we don't have to worry about this edge case.

What are your thoughts @vincbeck?

vincbeck (approved):

airflow-core/src/airflow/cli/commands/team_command.py
vincbeck: I agree, this should simplify the whole thing

potiuk: I am fine with that :)

potiuk: Done — team names are now lower case only: `(?!.*__)[a-z0-9_-]{3,50}`. The case-collision check that this comment was anchored to is gone entirely, along with the helper behind it; two names can no longer differ only in case, so there is nothing left to detect.

Rebasing also dropped this PR's changes to `environment_variables.py`. #70902 and #71079 landed in the meantime and made `_names_a_team_namespace` a plain `TEAM_SEP in secret_id` gated on `multi_team`, running before the team-scoped lookup — that already closes the cross-team read this PR described, and re-applying the leading-segment refinement would only loosen it. What is left here is the name rule itself.

One addition worth a look: existing deployments can hold names the new rule rejects, so `airflow db migrate` now lower-cases stored team names and carries each team's Dag bundles, Connections, Variables, Pools and Triggers with it (migration `c7f0a5d2e9b4`). Two stored names that fold together would silently merge two teams and nothing can pick a winner, so `_check_migration_errors` refuses to start migration while such a pair exists. `teams delete` deliberately no longer validates the name, otherwise a legacy team could never be removed.

---
Drafted-by: Claude Opus 5 (1M context); reviewed by @potiuk before posting

vincbeck:
CI is failing

o-nikolas:
> CI is failing

Yupp, other than the test failures, I like the direction this is going in!

vincbeck (approved):

PR #70889, opened 2026-08-01
Resolve backfill_id in the access dependency with the type the routes declare

potiuk:
The backfill routes declare `backfill_id: NonNegativeInt`, but
`requires_access_backfill` parsed the raw path value with `int()` and swallowed
the failure. Those two parsers do not agree.

### The divergence

| supplied `backfill_id` | `NonNegativeInt` (the handler) | `int()` (the dependency) |
|---|---|---|
| `"42"` | 42 | 42 |
| `"42.0"` | **42** | ValueError |
| `"42.00"` | **42** | ValueError |
| `"42."`, `"42e0"`, `"42.5"`, `"0x2a"` | rejected | ValueError |

pydantic's lax mode accepts the two decimal spellings and coerces them to an
int; `int()` rejects them.

FastAPI resolves route dependencies *before* the endpoint's own parameter
validation, so for those spellings the dependency's parse failed, it left the
Dag unresolved, and the request went on to be served by the handler against
backfill 42. The dependency and the handler disagreed about which backfill --
and therefore which Dag -- the request concerned.

### Approach

Parse with the same `TypeAdapter(NonNegativeInt)` the routes declare, so the
two cannot diverge by construction. The adapter is module-level and named, so
the coupling to the route signature is explicit rather than re-derived.

This deliberately does **not** introduce a stricter parser. An earlier revision
used `int()` with an explicit 400, which changed the public API contract: a
malformed path value such as `/backfills/invalid_id` previously produced
FastAPI's structured **422** (`loc: ["path", "backfill_id"]`), and a 400 from
the dependency preempted it. Five existing route tests caught that. Matching
the declared type keeps every existing status code and body shape intact --
values both parsers reject still yield the handler's 422, and an unknown
backfill still yields each route's own 404 wording.

### Test plan

- [x] `test_requires_access_backfill_authorizes_the_backfill_the_handler_will_act_on`
      -- parametrized over `42`, `42.0`, `42.00`; asserts the Dag comes from the
      resolved backfill and not from a value supplied on the request
- [x] Verified against unmodified code: `42.0` and `42.00` **fail**, `42` passes --
      so the test pins the divergence and not the plumbing
- [x] `airflow-core/tests/unit/api_fastapi/core_api/test_security.py` -- 111 passed
- [x] `airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_backfills.py`
      -- 63 passed, including the four `test_invalid_id` 422 cases and `test_no_exist`

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/security.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/security.py
@@ -27,6 +27,7 @@ from fastapi import Depends, HTTPException, Request, status
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer, OAuth2PasswordBearer
 from itsdangerous import BadSignature, URLSafeSerializer
 from jwt import ExpiredSignatureError, InvalidTokenError
+from pydantic import NonNegativeInt, TypeAdapter, ValidationError
 from sqlalchemy import or_, select
 from sqlalchemy.orm import Session
 
@@ -390,6 +391,12 @@ ReadableBackfillsFilterDep = Annotated[
 ]
 
 
+# The type the backfill routes declare for the `backfill_id` path parameter. Shared with
+# `requires_access_backfill` so the authorization decision parses the id exactly as the handler
+# does; see the comment there for why any divergence is a cross-Dag authorization bypass.
+_BACKFILL_ID_ADAPTER: TypeAdapter[NonNegativeInt] = TypeAdapter(NonNegativeInt)
+
+
 def requires_access_backfill(
     method: ResourceMethod,
 ) -> Callable[[Request, BaseUser, Session], Coroutine[Any, Any, None]]:
@@ -405,8 +412,20 @@ def requires_access_backfill(
         # Try to retrieve the dag_id from the backfill_id path param
         backfill_id_raw = request.path_params.get("backfill_id")
         try:
-            backfill_id = int(backfill_id_raw) if backfill_id_raw is not None else None
-        except ValueError:
+            # Parse with the *same* type the endpoints declare for this path parameter, so this
+            # dependency and the handler always resolve the same backfill -- and therefore the
+            # same Dag.
+            #
+            # `int()` is not that parser: pydantic's lax mode accepts strings `int()` rejects,
+            # so "1.0" and "1.00" validate to 1 for the handler while `int()` raised here. Since
+            # dependencies resolve before the endpoint's own parameter validation, that left
+            # `dag_id` unresolved for a request the handler went on to serve against backfill 1.
+            backfill_id = (
+                _BACKFILL_ID_ADAPTER.validate_python(backfill_id_raw) if backfill_id_raw is not None else None
+            )
+        except ValidationError:
+            # Rejected by the endpoint's parser too, so the handler cannot run: FastAPI answers
+            # 422 before it is reached. Left as None, preserving that response.
             backfill_id = None
 
         if backfill_id is not None:
--- a/airflow-core/src/airflow/api_fastapi/core_api/security.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/security.py
@@ -412,14 +412,8 @@ def requires_access_backfill(
         # Try to retrieve the dag_id from the backfill_id path param
         backfill_id_raw = request.path_params.get("backfill_id")
         try:
-            # Parse with the *same* type the endpoints declare for this path parameter, so this
-            # dependency and the handler always resolve the same backfill -- and therefore the
-            # same Dag.
-            #
-            # `int()` is not that parser: pydantic's lax mode accepts strings `int()` rejects,
-            # so "1.0" and "1.00" validate to 1 for the handler while `int()` raised here. Since
-            # dependencies resolve before the endpoint's own parameter validation, that left
-            # `dag_id` unresolved for a request the handler went on to serve against backfill 1.
+            # Must parse exactly as the handler does (e.g. pydantic's lax mode coerces "1.0" to 1
+            # where int() raises), or the two can authorize and act on different backfills.
             backfill_id = (
                 _BACKFILL_ID_ADAPTER.validate_python(backfill_id_raw) if backfill_id_raw is not None else None
             )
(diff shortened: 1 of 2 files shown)


airflow-core/tests/unit/api_fastapi/core_api/test_security.py
```diff
@@ -455,6 +458,48 @@ async def test_requires_access_backfill_backfill_not_found_falls_back_to_body(
...
+        backfill.dag_id = "backfill_dag"
+        session = Mock()
+        session.scalars.return_value.one_or_none.return_value = backfill
+
+        request = Mock()
+        request.path_params = {"backfill_id": backfill_id}
+        request.query_params = {"dag_id": "some_other_dag"}
+        request.json = AsyncMock(return_value={"dag_id": "some_other_dag"})
+
+        user = Mock()
```
amoghrajesh: Can we please use spec or autospec on the mocks here?

airflow-core/src/airflow/api_fastapi/core_api/security.py
amoghrajesh, line 403: Why dont we let fastapi validate it once natively. 

Instead of building a second `TypeAdapter(NonNegativeInt)` and calling it manually inside the dependency, the dependency function could declare `backfill_id: NonNegativeInt` as an actual typed parameter? That would mean literally one validation pass shared by construction, instead of two independently constructed validators that happen to use the same type today. This is architecturally tighter also and it removes even the possibility of the two drifting apart in the future

amoghrajesh (commented):
I guess we can do better here, maybe @pierrejeambrun has an opinion too.

airflow-core/tests/unit/api_fastapi/core_api/test_security.py
potiuk: Done — the mocks in this test now carry `spec=`: `Request`, `Session`, `Backfill`, `BaseAuthManager` and `BaseUser`. Without it an unspecced `Mock` accepts any attribute, so the test would keep passing even if the dependency started reading something the real objects do not have.

Pushed together with a rebase onto `main`.

---

airflow-core/src/airflow/api_fastapi/core_api/security.py
potiuk: I went and checked whether this works, and it does not — it would reintroduce the bug in a worse form.

`requires_access_backfill` is not attached only to `/{backfill_id}` routes. It is also on `GET /backfills`, `POST /backfills` and `POST /backfills/dry_run` (`backfills.py` lines 87, 285 and 355), where there is no `backfill_id` path parameter at all and the Dag is taken from the request body instead. If the dependency declares `backfill_id` as a typed parameter, FastAPI resolves it as a **query** parameter on exactly those three routes:

- declared bare, it is mandatory, so those three routes 422 on every request;
- declared `NonNegativeInt | None = None`, it becomes an optional query parameter that appears in the public OpenAPI schema — and lets a caller pass `?backfill_id=<other>` on `POST /backfills` to steer the authorization decision at a backfill the handler will never touch.

That second one is the same dependency-versus-handler divergence this PR exists to close, just entered from the other side.

On the "two independently constructed validators" point: it is the same `NonNegativeInt` the routes declare, which is what the comment above the adapter is pointing at — not a second type that happens to agree today. If the worry is future drift, the tighter fix is a named type alias that both the routes and this module import, so the declaration has one home. Happy to do that instead if you would prefer it, but it belongs in the same change as updating the route signatures, and I would rather keep this 3.3.1 backport to the bug.

---

potiuk:
cc: @pierrejeambrun ?

potiuk:
Both threads addressed and resolved. Mocks now carry `spec`.

I did not move `backfill_id` into the dependency signature: the dependency is also attached to `GET /backfills`, `POST /backfills` and `POST /backfills/dry_run`, which have no `backfill_id` path parameter, so FastAPI would resolve it as a *query* parameter there — 422-ing those routes if declared bare, or exposing `?backfill_id=` in the public schema and letting a caller steer authorization at a backfill the handler never touches. Detail in the thread; a shared type alias is the alternative if drift is the concern.

@amoghrajesh @pierrejeambrun

---

amoghrajesh (commented):
I see, seems like a valid issue to me. Can someone with better backfill API knowledge look at this one to prevent any ripple effects?

airflow-core/tests/unit/api_fastapi/core_api/test_security.py
```diff
@@ -430,7 +433,10 @@ async def test_requires_access_backfill_unauthorized(self, mock_get_auth_manager
     async def test_requires_access_backfill_backfill_not_found_falls_back_to_body(
         self, mock_get_auth_manager, mock_get_team_name
     ):
-        """When backfill_id is int but Backfill not found, dag_id from body is used."""
+        """When backfill_id is int but Backfill not found, dag_id from body is used.
+
+        Not exploitable: the handler answers 404 before acting, so no cross-Dag action follows.
+        """
```
ephraimbuddy:
```suggestion
        """When backfill_id is int but Backfill not found, dag_id from body is used."""
```
Should we add a Todo for the 404 instead of the comment even though it's not exploitable, i feel like we should fix it next since it leaks info

airflow-core/src/airflow/api_fastapi/core_api/security.py
```diff
@@ -405,8 +412,20 @@ async def inner(
...
-            backfill_id = int(backfill_id_raw) if backfill_id_raw is not None else None
-        except ValueError:
+            # Parse with the *same* type the endpoints declare for this path parameter, so this
+            # dependency and the handler always resolve the same backfill -- and therefore the
+            # same Dag.
+            #
+            # `int()` is not that parser: pydantic's lax mode accepts strings `int()` rejects,
+            # so "1.0" and "1.00" validate to 1 for the handler while `int()` raised here. Since
+            # dependencies resolve before the endpoint's own parameter validation, that left
+            # `dag_id` unresolved for a request the handler went on to serve against backfill 1.
```
ephraimbuddy:
```suggestion
            # Must parse exactly as the handler does (e.g. pydantic's lax mode coerces "1.0" to 1
            # where int() raises), or the two can authorize and act on different backfills.
```
Not blocking but i feel like it should be shorter

ephraimbuddy (approved):
LGTM. 
Just nits

airflow-core/tests/unit/api_fastapi/core_api/test_security.py
potiuk: I will open an issue for it and leave link in TODO here. Planned to do it after :)

PR #70890, opened 2026-08-01
Mask nested entities when writing bulk audit-log entries

potiuk:
The audit-log maskers for Variables and Connections dispatch on top-level key
names -- `val`/`value` for variables, `extra` for connections. A bulk request
body has exactly one top-level key, `actions`, and the entities sit two levels
down in `actions[].entities[]`, so neither masker ever saw a key it recognised
and the payload was recorded as supplied. The single-entity endpoints mask the
same fields correctly; only the bulk paths diverged.

### Approach

`_mask_bulk_entities` walks `actions[].entities[]` and applies the existing
per-entity masking to each entry, so the two shapes now agree. The per-entity
logic is unchanged and simply factored out (`_mask_connection_entity`,
`_mask_variable_entity`) so both call sites share it.

- A `delete` action may list bare id/key strings rather than entity objects;
  those carry nothing to mask and are passed through.
- The helper returns `None` for a non-bulk body so the flat path is untouched.
- It runs on the request body *before* validation, so malformed shapes
  (`actions` not a list, missing `entities`, non-dict actions) must not raise --
  each is covered by a test.

Worth noting for connections: a nested `password` was already covered, but only
while `hide_sensitive_var_conn_fields` is enabled, since that is key-name
redaction. `extra` was never covered by it -- the name is not a recognised
sensitive field and the value is a JSON *string*, which `redact` returns
unchanged. Masking `extra` here is structural and so does not depend on that
setting.

### Test plan

- [x] `TestMaskBulkFields` -- bulk variables, bulk connection `extra`, multiple
      actions/entities, delete-by-key, and five malformed shapes
- [x] Verified against unmodified code: 5 of the new tests **fail**, and the
      guard tests correctly pass either way
- [x] `test_decorators.py` -- 30 passed
- [x] `test_variables.py` + `test_connections.py` -- 230 passed

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/logging/decorators.py
+++ b/airflow-core/src/airflow/api_fastapi/logging/decorators.py
@@ -102,7 +102,11 @@ def _mask_connection_entity(extra_fields):
                     result[k] = {ek: "***" for ek in parsed_extra}
                 else:
                     result[k] = "Expected JSON object in `extra` field, got non-dict JSON"
-            except json.JSONDecodeError:
+            except (json.JSONDecodeError, TypeError):
+                # ``extra`` is declared as a string, but this runs on the raw body before
+                # validation, so it can arrive as any JSON type -- a number or an already-decoded
+                # object makes ``json.loads`` raise TypeError rather than JSONDecodeError. Both
+                # are recorded without the value, instead of raising out of the audit-log path.
                 result[k] = "Encountered non-JSON in `extra` field"
         else:
             result[k] = secrets_masker.redact(v, k)
--- a/airflow-core/src/airflow/api_fastapi/logging/decorators.py
+++ b/airflow-core/src/airflow/api_fastapi/logging/decorators.py
@@ -45,8 +45,51 @@ def _sanitize_for_stdlib_log(value: str) -> str:
     return value.replace("\r", " ").replace("\n", " ")
 
 
+def _mask_bulk_entities(extra_fields, mask_entity):
+    """
+    Apply per-entity masking to a bulk request body.
+
+    A ``BulkBody`` has exactly one top-level field, ``actions``; the entities carrying the
+    secrets sit two levels down, in ``actions[].entities[]``. The per-entity maskers below
+    inspect top-level key names, so handing them a bulk body means they see only the key
+    ``actions`` and pass its whole payload through untouched. Reach the entities first.
+
+    Returns ``None`` when the body is not bulk-shaped, so callers fall back to flat masking.
+    """
+    actions = extra_fields.get("actions")
+    if not isinstance(actions, list):
+        return None
+
+    masked_actions = []
+    for action in actions:
+        if not isinstance(action, dict):
+            masked_actions.append(action)
+            continue
+        entities = action.get("entities")
+        if not isinstance(entities, list):
+            masked_actions.append(action)
+            continue
+        # ``delete`` actions may list bare id/key strings rather than entity objects;
+        # those carry no secret and are left as they are.
+        masked_actions.append(
+            {
+                **action,
+                "entities": [mask_entity(e) if isinstance(e, dict) else e for e in entities],
+            }
+        )
+    return {**extra_fields, "actions": masked_actions}
+
+
 def _mask_connection_fields(extra_fields):
-    """Mask connection fields."""
+    """Mask connection fields, for either a single-entity or a bulk request body."""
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/api_fastapi/logging/decorators.py
amoghrajesh: `json.loads(v)` for the extra field only catches JsonDecodeError but `json.loads()` can raise TypeError, not JSONDecodeError, when given a non-string value (e.g. 123, {"already": "a dict"}). 

Since this masking function runs on the raw, pre validation request body - explicitly to tolerate malformed input, and "extra" set to anything other than a string would raise an uncaught `TypeError` here. 

Worth a fix while this is open.

airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
```diff
@@ -130,6 +130,141 @@ def test_value_without_key_is_still_masked(self):
...
 
+class TestMaskBulkFields:
+    """The bulk endpoints nest their entities, and the masking has to reach them.
+
+    A ``BulkBody`` has exactly one top-level field, ``actions``; the entities carrying the
+    secrets sit two levels down, in ``actions[].entities[]``. Both maskers dispatch on
+    top-level key names, so a bulk body previously presented them with the single key
+    ``actions`` -- neither ``val``/``value`` for variables, nor ``extra`` for connections --
+    and the payload was written to the audit log as supplied.
+    """
```
amoghrajesh: Remove this since already mentioned in the code helper.
```suggestion
```

airflow-core/src/airflow/api_fastapi/logging/decorators.py
```diff
@@ -45,8 +45,51 @@ def _sanitize_for_stdlib_log(value: str) -> str:
...
+
+    Returns ``None`` when the body is not bulk-shaped, so callers fall back to flat masking.
+    """
+    actions = extra_fields.get("actions")
+    if not isinstance(actions, list):
+        return None
+
+    masked_actions = []
+    for action in actions:
+        if not isinstance(action, dict):
```
amoghrajesh, line 65: This passes an unrecognized bulk shape through unmasked with no warning, maybe log a warning  here instead of silently passing it through untouched.

amoghrajesh (approved):
One future enhancement, otherwise LGTM.

airflow-core/src/airflow/api_fastapi/logging/decorators.py
potiuk: Good catch, and it is worth noting it is newly reachable rather than merely pre-existing: before this change a bulk body never got as far as the `extra` branch, because the only top-level key it presented was `actions`. Now it does.

Widened the except to `(json.JSONDecodeError, TypeError)`. Added `test_non_string_extra_does_not_raise_and_does_not_leak`, parametrized over an already-decoded object, an already-decoded array, a number and a bool — all four raise `TypeError` out of the audit-log decorator without the fix, and all four assert the secret does not survive into the serialized entry with it.

---

airflow-core/tests/unit/api_fastapi/logging/test_decorators.py
potiuk: Right — the class docstring was restating `_mask_bulk_entities`'s docstring almost sentence for sentence. Cut down to a single line; the mechanism stays documented at the helper.

---

airflow-core/src/airflow/api_fastapi/logging/decorators.py
potiuk: Not adding this one, because `None` here does not mean "unrecognised shape" — it means "not a bulk body", which is the ordinary case.

Every single-entity endpoint routed through this decorator sends a flat body with no `actions` key: `POST /connections`, `PATCH /connections/{id}`, `POST /variables`, `PATCH /variables/{key}` and so on. All of them take this return path on every request, and the caller then does exactly the right thing with it — falls back to flat masking. A warning here would fire on the majority of audited write requests and say nothing useful.

The narrower case — a body that *is* bulk-shaped but malformed, say `actions` present but an entity list that is not a list — is a request FastAPI rejects with 422 immediately after this runs. Nothing is written to the audit log for it, so there is no unmasked payload to warn about.

If the worry is a bulk shape changing under us and masking silently going quiet, the guard I would trust is a test rather than a log line, and `test_no_secret_survives_in_the_serialized_entry` is that test: it asserts on the serialized entry rather than on one field, so a shape change that stops reaching the entities fails it.

---

PR #70891, opened 2026-08-01
Mask list-shaped Variable values on deserialization

potiuk:
`_mask_and_deserialize_variable` dispatched on the top-level type of the
deserialized value and handled only `str` and `dict`. A Variable whose JSON is a
list therefore matched no branch and was returned with nothing inside it masked.

`add_mask` walks dicts and iterables itself, so the same list nested one level
inside a dict was already masked -- the top-level dispatch was the only thing
deciding whether the nested values were reached at all.

### Approach

Add the `list` branch, passing the value **under the variable's key** rather
than anonymously.

That distinction is the whole design of the change. Elements of a list have no
key names of their own, so they follow the variable key's sensitivity -- the same
rule already applied to a plain string value. Passing the list anonymously would
mask every element unconditionally and add each to the global pattern set, so a
Variable holding `["us-east-1", "eu-west-1"]` would cause `us-east-1` to be
redacted everywhere it appeared in any log. `test_var_json_list_value_does_not_over_mask`
already guarded that, and this keeps it true. Dicts *inside* the list are still
masked by their own key names.

Verified against the real masker:

| variable | key | result |
|---|---|---|
| `[{"password": "…"}]` | `db_configs` | `[{"password": "***"}]` |
| `["us-east-1", "eu-west-1"]` | `aws_regions` | unchanged |
| `["…"]` | `my_password` | `***` |
| `{"password": "…"}` | `db_config` | `{"password": "***"}` (unchanged path) |

`test_var_json_list_value_does_not_over_mask` is updated: it asserted
`mask_secret` was called exactly once, and there are now two calls. The property
it guards is unchanged and the assertion is now stated directly -- the list is
passed under the variable's key, and never anonymously.

### Test plan

- [x] `test_var_json_masks_list_values` -- list handed to the masker under the
      variable's key
- [x] `test_var_json_list_value_does_not_over_mask` -- updated to assert the
      no-over-masking property rather than the call count
- [x] `test_var_json_scalar_values_pass_through` -- int/bool/null/float returned
      unchanged
- [x] Verified against unmodified code: both list tests **fail**
- [x] `test_context.py` -- 170 passed

### DIFF
--- a/task-sdk/src/airflow/sdk/execution_time/context.py
+++ b/task-sdk/src/airflow/sdk/execution_time/context.py
@@ -323,9 +323,22 @@ def _mask_and_deserialize_variable(raw: str, key: str, deserialize_json: bool) -
         return raw
     val = json.loads(raw)
     if isinstance(val, str):
+        # No key names of its own, so the variable's key decides whether it is masked.
         mask_secret(val, key)
     elif isinstance(val, dict):
+        # ``add_mask`` walks a dict per key and masks by the *inner* key name, ignoring the
+        # name passed here.
         mask_secret(val)
+    elif isinstance(val, list):
+        # A list was previously skipped entirely, even though ``add_mask`` walks iterables --
+        # the same list nested one level inside a dict was masked, only a top-level one was not.
+        #
+        # It is passed under the variable's key, not anonymously: a list has no key names of
+        # its own, so bare values inside follow the same rule as the plain-string case above.
+        # Masking them unconditionally would add every element to the global pattern set, so a
+        # list of ordinary values such as region names would be redacted everywhere it appeared.
+        # Dicts *inside* the list are still masked by their own key names.
+        mask_secret(val, key)
     return val
 
 
--- a/task-sdk/src/airflow/sdk/execution_time/context.py
+++ b/task-sdk/src/airflow/sdk/execution_time/context.py
@@ -323,21 +323,13 @@ def _mask_and_deserialize_variable(raw: str, key: str, deserialize_json: bool) -
         return raw
     val = json.loads(raw)
     if isinstance(val, str):
-        # No key names of its own, so the variable's key decides whether it is masked.
         mask_secret(val, key)
     elif isinstance(val, dict):
-        # ``add_mask`` walks a dict per key and masks by the *inner* key name, ignoring the
-        # name passed here.
+        # Masked by the dict's own inner key names, which is what ``add_mask`` uses.
         mask_secret(val)
     elif isinstance(val, list):
-        # A list was previously skipped entirely, even though ``add_mask`` walks iterables --
-        # the same list nested one level inside a dict was masked, only a top-level one was not.
-        #
-        # It is passed under the variable's key, not anonymously: a list has no key names of
-        # its own, so bare values inside follow the same rule as the plain-string case above.
-        # Masking them unconditionally would add every element to the global pattern set, so a
-        # list of ordinary values such as region names would be redacted everywhere it appeared.
-        # Dicts *inside* the list are still masked by their own key names.
+        # Pass the Variable's key so list elements inherit the Variable's sensitivity
+        # instead of being added to the global mask patterns.
         mask_secret(val, key)
     return val
 
--- a/task-sdk/tests/task_sdk/execution_time/test_context.py
+++ b/task-sdk/tests/task_sdk/execution_time/test_context.py
@@ -430,12 +430,7 @@ class TestVariableAccessor:
 
     @mock.patch("airflow.sdk.execution_time.context.mask_secret")
     def test_var_json_masks_list_values(self, mock_mask_secret, mock_supervisor_comms):
-        """A JSON list is handed to the masker whole, exactly as a dict is.
-
-        ``add_mask`` walks dicts and iterables itself, so the top-level type dispatch was the
-        only thing deciding whether nested values were reached: a list at the top level matched
-        no branch and was skipped, while the same list nested inside a dict was masked.
-        """
+        """A JSON list is handed to the masker whole, exactly as a dict is."""
         accessor = VariableAccessor(deserialize_json=True)
         raw_json = '[{"password": "s3cr3t"}, {"password": "s3cr3t2"}]'
         mock_supervisor_comms.send.return_value = VariableResult(key="db_configs", value=raw_json)
@@ -491,14 +486,7 @@ class TestVariableAccessor:
 
     @mock.patch("airflow.sdk.execution_time.context.mask_secret")
     def test_var_json_list_value_does_not_over_mask(self, mock_mask_secret, mock_supervisor_comms):
-        """var.json with a non-sensitive list variable does not mask individual list elements.
-
-        The list is handed to the masker **under the variable's key**, which is what preserves
-        this: bare values inside a list have no key names of their own, so they follow the
-        variable key's sensitivity. Passing the list anonymously instead would add every element
-        to the global pattern set, and an ordinary value such as a region name would then be
-        redacted everywhere it appeared in the logs.
-        """
+        """var.json with a non-sensitive list variable does not mask individual list elements."""
         accessor = VariableAccessor(deserialize_json=True)
         raw_json = '["us-east-1", "eu-west-1"]'
         mock_supervisor_comms.send.return_value = VariableResult(key="aws_regions", value=raw_json)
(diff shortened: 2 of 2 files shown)


task-sdk/src/airflow/sdk/execution_time/context.py
```diff
@@ -323,9 +323,22 @@ def _mask_and_deserialize_variable(raw: str, key: str, deserialize_json: bool) -
...
         mask_secret(val)
+    elif isinstance(val, list):
+        # A list was previously skipped entirely, even though ``add_mask`` walks iterables --
+        # the same list nested one level inside a dict was masked, only a top-level one was not.
+        #
+        # It is passed under the variable's key, not anonymously: a list has no key names of
+        # its own, so bare values inside follow the same rule as the plain-string case above.
+        # Masking them unconditionally would add every element to the global pattern set, so a
+        # list of ordinary values such as region names would be redacted everywhere it appeared.
+        # Dicts *inside* the list are still masked by their own key names.
```
SameerMesiah97: I had to read this a few times to understand what it was trying to convey. I think the comment can be more concise and clear. Please see below:

```
# Pass the Variable's key so list elements inherit the Variable's sensitivity
# instead of being added to the global mask patterns.
```

task-sdk/tests/task_sdk/execution_time/test_context.py
```diff
@@ -428,6 +428,41 @@ def test_var_json_masks_raw_string_and_dict_values(self, mock_mask_secret, mock_
         # Second call: deserialized dict so internal sensitive fields like "password" get masked
         mock_mask_secret.assert_any_call({"password": "s3cr3t", "host": "db.example.com"})
 
+    @mock.patch("airflow.sdk.execution_time.context.mask_secret")
+    def test_var_json_masks_list_values(self, mock_mask_secret, mock_supervisor_comms):
+        """A JSON list is handed to the masker whole, exactly as a dict is.
+
```
SameerMesiah97: I would keep the first line for the docstring.

task-sdk/tests/task_sdk/execution_time/test_context.py
```diff
@@ -456,15 +491,25 @@ def test_var_json_string_value_masks_both_forms(self, mock_mask_secret, mock_sup
...
     @mock.patch("airflow.sdk.execution_time.context.mask_secret")
     def test_var_json_list_value_does_not_over_mask(self, mock_mask_secret, mock_supervisor_comms):
-        """var.json with a non-sensitive list variable does not mask individual list elements."""
+        """var.json with a non-sensitive list variable does not mask individual list elements.
+
+        The list is handed to the masker **under the variable's key**, which is what preserves
+        this: bare values inside a list have no key names of their own, so they follow the
+        variable key's sensitivity. Passing the list anonymously instead would add every element
+        to the global pattern set, and an ordinary value such as a region name would then be
+        redacted everywhere it appeared in the logs.
```
SameerMesiah97: Same here. The first line of the docstring can be kept and extra context can be retained in the comments.

SameerMesiah97 (commented):
Looks good. Just a few stylistic nits (for e.g. I found one comment a bit hard to understand).

task-sdk/tests/task_sdk/execution_time/test_context.py
```diff
@@ -428,6 +428,41 @@ def test_var_json_masks_raw_string_and_dict_values(self, mock_mask_secret, mock_
...
         mock_mask_secret.assert_any_call({"password": "s3cr3t", "host": "db.example.com"})
 
+    @mock.patch("airflow.sdk.execution_time.context.mask_secret")
+    def test_var_json_masks_list_values(self, mock_mask_secret, mock_supervisor_comms):
+        """A JSON list is handed to the masker whole, exactly as a dict is.
+
+        ``add_mask`` walks dicts and iterables itself, so the top-level type dispatch was the
+        only thing deciding whether nested values were reached: a list at the top level matched
+        no branch and was skipped, while the same list nested inside a dict was masked.
+        """
```
amoghrajesh:
```suggestion
```

Since its already there in the code path.

task-sdk/src/airflow/sdk/execution_time/context.py
amoghrajesh: +1 agreed.

task-sdk/tests/task_sdk/execution_time/test_context.py
amoghrajesh: Agree with @SameerMesiah97, either that or remove docstring too.

task-sdk/src/airflow/sdk/execution_time/context.py
```diff
@@ -323,9 +323,22 @@ def _mask_and_deserialize_variable(raw: str, key: str, deserialize_json: bool) -
         return raw
     val = json.loads(raw)
     if isinstance(val, str):
+        # No key names of its own, so the variable's key decides whether it is masked.
         mask_secret(val, key)
     elif isinstance(val, dict):
+        # ``add_mask`` walks a dict per key and masks by the *inner* key name, ignoring the
+        # name passed here.
         mask_secret(val)
+    elif isinstance(val, list):
```
amoghrajesh: In dict branch, we dont pass key. Inside `add_mask` dict is handled this way:
```python
if isinstance(secret, dict):
    for k, v in secret.items():
        self.add_mask(v, k)
```

Therefore, 
- `mask_secret({"password": "x"})` ->  calls `add_mask("x", "password")`
- `mask_secret({"password": "x"}, "some_key")` -> calls `add_mask("x", "password")`

So, you could just do this to simplify it:

```python
if isinstance(val, (str, dict, list)):
    mask_secret(val, key)
```

task-sdk/src/airflow/sdk/execution_time/context.py
potiuk: Taken, with your wording verbatim. The long version was trying to justify the choice and the surrounding history at the same time; yours says the thing that actually matters at the point of decision.

---

task-sdk/tests/task_sdk/execution_time/test_context.py
potiuk: Done — first line kept, the rest dropped. The detail it carried is now in the one comment at the dispatch branch rather than repeated in each test.

---

task-sdk/tests/task_sdk/execution_time/test_context.py
potiuk: Done — first line kept, the rest dropped. The detail it carried is now in the one comment at the dispatch branch rather than repeated in each test.

---

task-sdk/tests/task_sdk/execution_time/test_context.py
potiuk: Done — first line kept, the rest dropped. The detail it carried is now in the one comment at the dispatch branch rather than repeated in each test.

---

task-sdk/tests/task_sdk/execution_time/test_context.py
potiuk: Done — first line kept, the rest dropped. The detail it carried is now in the one comment at the dispatch branch rather than repeated in each test.

---

task-sdk/src/airflow/sdk/execution_time/context.py
potiuk: Your reading of `add_mask` is right — for a dict it re-enters per inner key and the `name` it was handed is never used, so `mask_secret(val, key)` and `mask_secret(val)` are indistinguishable for a dict today. I checked the source rather than taking it on trust.

Still not merging the branches, for two reasons.

The equivalence is incidental, not designed. `key` would be passed for the dict case and then silently ignored, so the code would read as though the Variable's key participates in masking a dict when it does not. If `add_mask` ever did use the outer name for dicts — masking the whole dict when the Variable key is sensitive would be a reasonable thing to want — the merged form changes what gets masked, with nothing at this call site to signal it. The separate branches say what each type actually needs, which is different things.

The second reason is scope. This is milestoned for 3.3.1. The dict branch is pre-existing behaviour and the bug is narrower than it looks: a top-level list matched no branch at all and was skipped entirely, while the same list nested one level inside a dict was masked. Rewriting the dict line as well widens what a backport touches for no behavioural gain.

The comments are trimmed per the other threads, which is what made the dispatch look bulkier than it is — it is now three short branches.

---

potiuk:
All threads addressed and resolved. Comments trimmed as suggested.

I did not take the branch collapse — the equivalence relies on `add_mask` ignoring the `name` it is handed for a dict, so the key would be passed and silently ignored, and the dict branch is pre-existing behaviour I would rather not rewrite inside a 3.3.1 backport. Reasoning is in that thread if you disagree.

@amoghrajesh @SameerMesiah97

---

amoghrajesh (approved):
Thanks, simple enough now.

PR #70893, opened 2026-08-01
Resolve the Dag's team when authorizing a Dag found by lookup

potiuk:
Two authorization checks build `DagDetails(id=dag_id)` without `team_name`:

- `materialize_asset`, where the Dag is resolved from the asset rather than
  named by the caller;
- the XCom-specific check in `wait_dag_run_until_finished`.

Every other call site resolves the team with `DagModel.get_team_name` and passes
it -- `security.py:194`, `security.py:235`, `security.py:868`,
`import_error.py:314`, `services/public/task_instances.py:405`.

### Why it matters

A team-aware auth manager distinguishes a team-scoped Dag from a global one by
that field. The Keycloak manager, for instance, selects the resource it asks
about from it -- `DAG:<team>` when set, plain `DAG` when not -- so omitting it
asks about a differently-scoped resource than the one being acted on.

`wait_dag_run_until_finished` is the clearer illustration: its route dependency
already resolves the team for the RUN check, so the two checks in the same
handler disagreed about the scope of the same Dag.

### Approach

Resolve the team at both sites with `DagModel.get_team_name(dag_id, session=session)`,
reusing the request session rather than opening a new one.

After this, `grep -rn "DagDetails(id=" airflow-core/src` returns no call site
without `team_name`.

### Test plan

- [x] `test_authorizes_against_the_dags_team` -- parametrized over a team-scoped
      Dag and a global one; asserts the `DagDetails` reaching the auth manager
      carries the resolved team
- [x] Verified against unmodified code: the team-scoped case **fails**, the
      global case passes either way -- so the test pins the resolution, not the
      plumbing
- [x] `test_assets.py` -- 178 passed
- [x] `test_dag_run.py` -- 358 passed; the 2 failures in
      `TestBulkClearDagRuns` / `TestBulkDagRuns` (`UNIQUE constraint failed:
      team.name`) are **pre-existing** and reproduce identically on unmodified
      `main`

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
@@ -86,6 +86,7 @@ from airflow.models.asset import (
     AssetWatcherModel,
     TaskOutletAssetReference,
 )
+from airflow.models.dag import DagModel
 from airflow.models.dag_version import DagVersion
 from airflow.typing_compat import Unpack
 from airflow.utils.state import DagRunState
@@ -459,7 +460,10 @@ def materialize_asset(
     if not get_auth_manager().is_authorized_dag(
         method="POST",
         access_entity=DagAccessEntity.RUN,
-        details=DagDetails(id=dag_id),
+        # The Dag is resolved from the asset here rather than named by the caller, so its team has
+        # to be looked up too. A team-aware auth manager distinguishes a team-scoped Dag from a
+        # global one by this field, so leaving it None asks the wrong question.
+        details=DagDetails(id=dag_id, team_name=DagModel.get_team_name(dag_id, session=session)),
         user=user,
     ):
         raise HTTPException(
--- a/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
+++ b/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
@@ -2154,11 +2154,8 @@ class TestPostAssetMaterialize(TestAssets):
 
     @pytest.mark.parametrize("team_name", ["team_b", None])
     def test_authorizes_against_the_dags_team(self, test_client, session, team_name):
-        """The Dag is resolved from the asset, so its team must be resolved and passed too.
-
-        A team-aware auth manager distinguishes a team-scoped Dag from a global one by this field,
-        so leaving it ``None`` asks about a differently-scoped resource than the one being acted on.
-        """
+"""The Dag is resolved from the asset, so its team must be resolved and passed too — see the
+    call site's comment for why an unresolved team asks about the wrong resource."""
         recorded = []
 
         auth_manager = mock.Mock(spec=BaseAuthManager)
--- a/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
+++ b/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
@@ -2151,6 +2151,32 @@ class TestPostAssetMaterialize(TestAssets):
         assert dag_run.partition_key == "2025-06-01T00:00:00"
         assert dag_run.partition_date == timezone.datetime(2025, 6, 1)
 
+    @pytest.mark.parametrize("team_name", ["team_b", None])
+    def test_authorizes_against_the_dags_team(self, test_client, session, team_name):
+        """The Dag is resolved from the asset, so its team must be resolved and passed too.
+
+        A team-aware auth manager distinguishes a team-scoped Dag from a global one by this field,
+        so leaving it ``None`` asks about a differently-scoped resource than the one being acted on.
+        """
+        recorded = []
+
+        auth_manager = mock.Mock()
+        auth_manager.is_authorized_dag.side_effect = lambda **kw: recorded.append(kw) or True
+
+        with (
+            mock.patch(
+                "airflow.api_fastapi.core_api.routes.public.assets.get_auth_manager",
+                return_value=auth_manager,
+            ),
+            mock.patch.object(DagModel, "get_team_name", return_value=team_name),
+        ):
+            test_client.post("/assets/1/materialize")
+
+        assert recorded, "the authorization check did not run"
+        details = recorded[0]["details"]
+        assert details.id == self.DAG_ASSET1_ID
+        assert details.team_name == team_name
+
 
 class TestGetAssetQueuedEvents(TestQueuedEventEndpoint):
     @pytest.mark.usefixtures("time_freezer")
--- a/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
+++ b/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
@@ -2154,8 +2154,8 @@ class TestPostAssetMaterialize(TestAssets):
 
     @pytest.mark.parametrize("team_name", ["team_b", None])
     def test_authorizes_against_the_dags_team(self, test_client, session, team_name):
-"""The Dag is resolved from the asset, so its team must be resolved and passed too — see the
-    call site's comment for why an unresolved team asks about the wrong resource."""
+        """The Dag is resolved from the asset, so its team must be resolved and passed too — see the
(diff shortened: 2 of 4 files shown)


airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
```diff
@@ -2141,6 +2141,32 @@ def test_should_respond_200_with_partition_date_for_partitioned_dag(
...
+                return_value=auth_manager,
+            ),
+            mock.patch.object(DagModel, "get_team_name", return_value=team_name),
+        ):
+            test_client.post("/assets/1/materialize")
+
+        assert recorded, "the authorization check did not run"
+        details = recorded[0]["details"]
+        assert details.id == self.DAG_ASSET1_ID
+        assert details.team_name == team_name
```
SameerMesiah97, line 2178: I think this test could be more thorough. Please see the below:

```
@pytest.mark.parametrize("team_name", ["team_b", None])
def test_authorizes_against_the_dags_team(self, test_client, session, team_name):
      """The Dag is resolved from the asset, so its team must be resolved and passed too.

        A team-aware auth manager distinguishes a team-scoped Dag from a global one by this field,
        so leaving it ``None`` asks about a differently-scoped resource than the one being acted on.
        """
    recorded = []

    auth_manager = mock.Mock()
    auth_manager.is_authorized_dag.side_effect = lambda **kw: recorded.append(kw) or True

    with (
        mock.patch(
            "airflow.api_fastapi.core_api.routes.public.assets.get_auth_manager",
            return_value=auth_manager,
        ),
        mock.patch.object(
            DagModel,
            "get_team_name",
            return_value=team_name,
        ) as mock_get_team_name,
    ):
        response = test_client.post("/assets/1/materialize")

    assert response.status_code == 200

    mock_get_team_name.assert_called_once_with(
        self.DAG_ASSET1_ID,
        session=session,
    )

    assert len(recorded) == 1, "expected exactly one authorization check"

    details = recorded[0]["details"]
    assert details.id == self.DAG_ASSET1_ID
    assert details.team_name == team_name
```

SameerMesiah97 (commented):
Looks good. The test could be improved though.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
```diff
@@ -459,7 +460,10 @@ def materialize_asset(
     if not get_auth_manager().is_authorized_dag(
         method="POST",
         access_entity=DagAccessEntity.RUN,
-        details=DagDetails(id=dag_id),
+        # The Dag is resolved from the asset here rather than named by the caller, so its team has
+        # to be looked up too. A team-aware auth manager distinguishes a team-scoped Dag from a
+        # global one by this field, so leaving it None asks the wrong question.
+        details=DagDetails(id=dag_id, team_name=DagModel.get_team_name(dag_id, session=session)),
```
amoghrajesh, line 466: This site has no regression test - `test_authorizes_against_the_dags_team` only covers `materialize_asset`

amoghrajesh (commented):
Maybe a later PR, but worth a guardrail like a `DagDetails.for_dag(dag_id, *, session)` constructor that always resolves the team

airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
potiuk: Right, and it is worth spelling out why it slipped: `wait_dag_run_until_finished` *is* exercised by `test_should_respond_403_when_user_lacks_xcom_permission`, which asserts on the `DagDetails` passed to the XCom check — but against a Dag with no team. `DagDetails(id=X)` and `DagDetails(id=X, team_name=None)` are equal, so that test passes identically with and without the change and could never have caught this.

Added `test_authorizes_xcom_against_the_dags_team`, parametrized over `["team_b", None]`, patching `DagModel.get_team_name` the same way the assets test does. Confirmed the `team_b` case fails without the source change and passes with it; the `None` case is there to pin that a global Dag still authorizes as global.

---

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
potiuk: Thanks — the parametrize over `["team_b", None]` went in with an earlier push. Picked up the two assertions still missing from your version:

- `assert len(recorded) == 1` instead of a bare truthiness check, so a second authorization call would fail rather than pass unnoticed;
- `mock_get_team_name.assert_called_once_with(self.DAG_ASSET1_ID, session=mock.ANY)`, which pins that the team is resolved for the Dag the asset led to rather than for some other Dag.

I kept `session=mock.ANY` rather than asserting the exact session object — the route uses its own request-scoped session, not the one the test fixture hands out, so asserting identity there would be testing the dependency wiring rather than this change.

---

potiuk:
Both review threads addressed, and the second call site now has its own regression test.

On the `DagDetails.for_dag(dag_id, *, session)` idea — I agree with it, and I think the diagnosis behind it is the important part: these two sites drifted precisely because `DagDetails(id=...)` makes the unscoped form the easy one to write and the scoped form something you have to remember. A constructor that always resolves the team turns that around.

Not in this PR though. It is milestoned for 3.3.1, and changing how `DagDetails` is constructed across the codebase is a much wider blast radius than the two-line fix this is — exactly the kind of change that should go to main on its own and be reviewed on its own merits, not ride in on a backport. Happy to see it done separately.

---

potiuk:
Both threads addressed and resolved. The missing regression test for `wait_dag_run_until_finished` is in — worth noting *why* it was missing: the existing test asserts on the `DagDetails` passed to the XCom check, but against a team-less Dag, where `DagDetails(id=X)` and `DagDetails(id=X, team_name=None)` are equal, so it passed identically with and without the fix. The new test parametrizes over a team-scoped Dag. Mocks now carry `spec`/`autospec`.

@amoghrajesh @SameerMesiah97

---

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_assets.py
```diff
@@ -2151,6 +2152,36 @@ def test_should_respond_200_with_partition_date_for_partitioned_dag(
         assert dag_run.partition_key == "2025-06-01T00:00:00"
         assert dag_run.partition_date == timezone.datetime(2025, 6, 1)
 
+    @pytest.mark.parametrize("team_name", ["team_b", None])
+    def test_authorizes_against_the_dags_team(self, test_client, session, team_name):
+        """The Dag is resolved from the asset, so its team must be resolved and passed too.
+
+        A team-aware auth manager distinguishes a team-scoped Dag from a global one by this field,
+        so leaving it ``None`` asks about a differently-scoped resource than the one being acted on.
+        """
```
amoghrajesh:
```suggestion
"""The Dag is resolved from the asset, so its team must be resolved and passed too — see the
    call site's comment for why an unresolved team asks about the wrong resource."""
```

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
```diff
@@ -4394,6 +4394,35 @@ def test_should_respond_403_when_user_lacks_xcom_permission(self, test_client):
                 user=mock.ANY,
             )
 
+    @pytest.mark.parametrize("team_name", ["team_b", None])
+    def test_authorizes_xcom_against_the_dags_team(self, test_client, team_name):
+        """The XCom check must carry the Dag's team, as the route dependency above already does.
+
+        A team-aware auth manager distinguishes a team-scoped Dag from a global one by this field,
+        so leaving it ``None`` asks about a differently-scoped resource than the one being read.
+        """
```
amoghrajesh:
```suggestion
    """The XCom check must carry the Dag's team, matching what the route dependency above already
    resolves — see the call site's comment for why."""
```

amoghrajesh (approved):
OK by me, but if someone like @vincbeck / @o-nikolas could review, that would be nice.

Lee-W (approved):

PR #63204, opened 2026-03-09
Remove JWT secrets from triggerer, worker and dag-processor

deepujain:
## Summary

Deployments that use the chart-managed `jwt-secret` (dag-processor, triggerer, workers) did not have the `checksum/jwt-secret` pod template annotation. When the JWT secret value was updated, those pods were not restarted automatically, unlike the api-server and scheduler which already had the annotation. This adds the same annotation (gated by `semverCompare ">=3.0.0"` and `not .Values.jwtSecretName`) so a change to the jwt-secret triggers a rollout.

## Change

- **chart/templates/dag-processor/dag-processor-deployment.yaml**: Add `checksum/jwt-secret` annotation when Airflow >= 3.0 and chart manages the JWT secret.
- **chart/templates/triggerer/triggerer-deployment.yaml**: Same.
- **chart/templates/workers/worker-deployment.yaml**: Same.

Condition matches the pattern used in scheduler and api-server so behavior is consistent.

## Fixes

Fixes #62146

PR #70896, opened 2026-08-01
Separate JWT secret env var from the standard Airflow environment helper

### LINKED ISSUE #70843: Separate JWT Secret Env from Standard Airflow Environment Variables
### Description

#63204 changed which component have `AIRFLOW__API_AUTH__JWT_SECRET` environment variable set. It was done by modification of `standard_airflow_environment` and adding condition on mentioned variable. We could extract the variable to dedicated helper and adjust configuration of all components to make the implementation cleaner.

### Use case/motivation

Improve templates readability

### Related issues

_No response_

### Code of Conduct

rohan9446:
`AIRFLOW__API_AUTH__JWT_SECRET` was rendered from inside the
`standard_airflow_environment` helper behind an `IncludeJwtSecret` flag, which
meant every component had to opt out of it explicitly with
`(merge (dict "IncludeJwtSecret" false) .)`.

This moves the variable into its own `jwt_secret_environment` helper, following
the shape of the existing `keda_airflow_environment` helper, and includes it
only in the two containers that actually need it — the API server and the
scheduler. Every other caller of `standard_airflow_environment` passes a plain
context again, and the `IncludeJwtSecret` context mutation is gone entirely.

What this changes:

- Removes the `IncludeJwtSecret` plumbing from `standard_airflow_environment`
  and from all ten call sites.
- Makes secret exposure explicit at the API server and scheduler call sites
  rather than implicit in a shared helper.
- Preserves the least-privilege behaviour introduced in #63204: migration/init
  containers, workers, the triggerer and the DAG processor still do not receive
  the raw signing secret.
- Adds regression coverage asserting the secret appears in the API server and
  scheduler containers exactly once, and never in sidecars or init containers.

No behavioural change: the same containers receive the same variable, still
gated on `enableBuiltInSecretEnvVars.AIRFLOW__API_AUTH__JWT_SECRET`. The only
difference in rendered output is the position of the variable within the env
list, so the ordered assertion in `test_have_all_variables` is updated to match.

Tested locally with helm v3.21.3: `pytest chart/tests/helm_tests/` → 2108 passed.

closes: #70843

---

Generated-by: Claude (Cowork) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


SameerMesiah97 (commented):
Approved pending green CI.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -396,6 +396,53 @@ def test_have_all_variables(self):
...
+    )
+    def test_jwt_secret_can_be_disabled(self, template, main_container):
+        docs = render_chart(
+            values={"enableBuiltInSecretEnvVars": {"AIRFLOW__API_AUTH__JWT_SECRET": False}},
+            show_only=[template],
+        )
+
+        containers = jmespath.search("spec.template.spec.containers", docs[0])
+        main = next(container for container in containers if container["name"] == main_container)
+        env_names = [env["name"] for env in main.get("env") or []]
```
Miretpl: Could you move this logic to `jmespath.search`?

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -396,6 +396,53 @@ def test_have_all_variables(self):
...
+            )
+
+    @pytest.mark.parametrize(
+        ("template", "main_container"),
+        [
+            ("templates/api-server/api-server-deployment.yaml", "api-server"),
+            ("templates/scheduler/scheduler-deployment.yaml", "scheduler"),
+        ],
+    )
+    def test_jwt_secret_can_be_disabled(self, template, main_container):
```
Miretpl: You don't really need the `template` parameter here.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -396,6 +396,53 @@ def test_have_all_variables(self):
...
+        ],
+    )
+    def test_jwt_secret_only_injected_into_api_server_and_scheduler(
+        self, template, main_container, expected_jwt_secret
+    ):
+        docs = render_chart(show_only=[template])
+
+        containers = jmespath.search("spec.template.spec.containers", docs[0])
+        main = next(container for container in containers if container["name"] == main_container)
+        env_names = [env["name"] for env in main.get("env") or []]
```
Miretpl: Could you move this logic to `jmespath.search`?

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -396,6 +396,53 @@ def test_have_all_variables(self):
...
+        ("template", "main_container", "expected_jwt_secret"),
+        [
+            ("templates/api-server/api-server-deployment.yaml", "api-server", True),
+            ("templates/scheduler/scheduler-deployment.yaml", "scheduler", True),
+            ("templates/workers/worker-deployment.yaml", "worker", False),
+            ("templates/triggerer/triggerer-deployment.yaml", "triggerer", False),
+            ("templates/dag-processor/dag-processor-deployment.yaml", "dag-processor", False),
+        ],
+    )
+    def test_jwt_secret_only_injected_into_api_server_and_scheduler(
```
Miretpl: I would split this test case into two: when you expect a secret and when you don't.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
```diff
@@ -396,6 +396,53 @@ def test_have_all_variables(self):
...
+        assert env_names.count("AIRFLOW__API_AUTH__JWT_SECRET") == int(expected_jwt_secret)
+
+        # the secret must never leak into sidecars or init containers
+        others = [container for container in containers if container["name"] != main_container]
+        others += jmespath.search("spec.template.spec.initContainers", docs[0]) or []
+        for container in others:
+            other_env_names = [env["name"] for env in container.get("env") or []]
+            assert "AIRFLOW__API_AUTH__JWT_SECRET" not in other_env_names, (
+                f"JWT secret leaked into {container['name']}"
+            )
```
Miretpl: Could you move this logic to `jmespath.search`?

Miretpl (changes requested):

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
rohan9446: Addressed, the main container environment names are now selected directly with jmespath.search.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
rohan9446: I split this into separate tests for components that should receive the JWT secret and components that should not.

chart/tests/helm_tests/airflow_aux/test_airflow_common.py
rohan9446: I removed the parametrization and now render both the API server and scheduler templates in a single test.

Miretpl (approved):

Miretpl:
Congrats and welcome to the community!

Miretpl:
@rohan9446, could you create manual backport of it to `chart/v1-2x-test` branch?

rohan9446:
Thanks @Miretpl! Backport is up: #71129

The automated one failed because `chart/v1-2x-test` still supports Airflow 2 alongside 3, the JWT block there carries an extra `semverCompare ">=3.0.0" .Values.airflowVersion` gate, there's no `keda_airflow_environment` helper to anchor the new define against, and the helm tests live under `helm-tests/tests/`. The version gate now sits inside `jwt_secret_environment`, so Airflow 2 rendering is unchanged. Full chart suite passes locally (3454 tests).

PR #70899, opened 2026-08-01
Apply the ambiguous-id refusal to get_config in the Key Vault backend

potiuk:
Follow-up to #70876.

The ambiguous-id check used to live in `_get_secret`, which all three getters
share. Moving it up into `get_conn_value` and `get_variable` left `get_config`
reaching `_get_secret` unguarded, so `_names_a_team_namespace`'s claim that such
an id is refused for *every* lookup stopped being true the moment it was written —
two getters and the docstring disagreed with the third.

**Not a cross-team read.** `get_config` never receives a `team_name`, so no team
namespace is reachable through it. The defect is the inconsistency and the false
docstring, not an exposure.

@shahar1 found this while reviewing the Amazon sibling (#70878), where it is
already fixed. Azure had the same shape and had merged by then, so this is the
other half.

`_names_a_team_namespace` stays an **instance** method here, unlike Amazon's,
because it builds the candidate path with `self.build_path` and `self.sep`.

### Test plan

- [x] `test_refusing_an_ambiguous_id_is_logged` extended to `get_config`; verified
      it fails without the guard
- [x] Assertion reworked off the rendered sentence — Airflow logs through
      structlog, which renders format args into `msg` before the stdlib record
      exists, so `record.args` is empty and there is no structured payload to
      assert on; it now asserts level, logger and the refused id
- [x] `test_key_vault.py` — 20 passed

### DIFF
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py
@@ -190,6 +190,10 @@ class AzureKeyVaultBackend(BaseSecretsBackend, LoggingMixin):
         if self.config_prefix is None:
             return None
 
+        if self._names_a_team_namespace(key):
+            self._log_refusal("configuration option", key)
+            return None
+
         return self._get_secret(self.config_prefix, key)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


potiuk:
CI fixed, and the fix reaches beyond this PR — **the Amazon tests already on `main` were failing the same way**, so this now repairs those too.

The failures were `Compat 2.11.1:P3.10` and `Compat 3.0.6:P3.10`, all three backends, `assert 0 == 1`. Cause: the assertions read `record.msg` directly, and how a record carries its payload depends on the Airflow version.

```python
# main — structlog renders the args before the stdlib record exists
LogRecord(msg="Connection id 'prod--x' contains '--'", args=())
    "prod--x" in r.msg          -> True

# compat targets — plain stdlib logging keeps the template
LogRecord(msg="%s id %r contains %r", args=("Connection", "prod--x", "--"))
    "prod--x" in r.msg          -> False        <- the failure
    "prod--x" in r.getMessage() -> True
```

`getMessage()` renders in both shapes, so the assertion is now version-portable. It still targets the refused **id** rather than the wording, which was the point of moving off the rendered sentence in the first place.

This is my own regression: I switched those assertions from `getMessage()` to `msg` in #70878 while addressing @shahar1's review, verified only against `main`, and merged. Compat runs on provider PRs, so it surfaced on the next provider PR rather than on the one that introduced it.

potiuk:
Rewritten to Azure only — the repair to `main` moved to **#70905**.

I had folded the Amazon test fix in here because it was the same one-line cause, but that conflated two things: this PR is a follow-up guard for Key Vault, whereas `main` being red on the compat jobs is a regression that should land on its own and not wait behind a review of unrelated code.

This branch is now:

```
providers/microsoft/azure/src/.../secrets/key_vault.py       +11/-3
providers/microsoft/azure/tests/.../secrets/test_key_vault.py +16/-3
```

The `getMessage()` change remains in the Azure test here, since that test is added by this PR and would otherwise land already broken for the compat jobs.

Miretpl (approved):

PR #65943, opened 2026-04-27
Allow using fresh interpreter besides fork() in Edge Worker

diogosilva30:
## What

Update the Edge worker task launch path to honor Airflow's existing `[core] execute_tasks_new_python_interpreter` option.

By default, Edge workers keep the existing fork-based behavior. When `execute_tasks_new_python_interpreter=True` is configured, or when `os.fork` is unavailable, the worker launches the task in a fresh Python interpreter with `subprocess.Popen` and the existing `airflow.sdk.execution_time.execute_workload` entrypoint.

Fixes #65942

## Why

Some Edge worker deployments can run task execution from a multi-threaded worker process. Forking a process after threads have started can inherit unsafe parent state, including import locks and partially initialized modules. In affected deployments this can show up as intermittent task-start failures, plugin import errors, or startup-reschedule exhaustion.

Airflow already has a core setting for this tradeoff: `execute_tasks_new_python_interpreter`. Other execution paths can use it to choose a fresh interpreter instead of fork. This PR applies the same behavior to the Edge worker without changing the default for existing deployments.

## How

The change keeps both launch modes:

| Mode | When used | Behavior |
|------|-----------|----------|
| Fork | Default when `os.fork` exists and `execute_tasks_new_python_interpreter=False` | Uses `multiprocessing.Process` and the existing supervisor helper |
| Fresh interpreter | `execute_tasks_new_python_interpreter=True` or no `os.fork` | Uses `subprocess.Popen` with `python -m airflow.sdk.execution_time.execute_workload --json-string ...` |

The fresh-interpreter path also spools stderr to a temporary file instead of `stderr=PIPE`. The worker only needs stderr after the subprocess exits, and a pipe can deadlock if the child writes enough data before the parent reads it. Spooling to a file avoids that while still preserving root failure details in task logs.

## Changes

| File | What changed |
|------|-------------|
| `providers/edge3/src/airflow/providers/edge3/cli/worker.py` | Route task launch through fork or fresh interpreter based on `self.conf.getboolean("core", "execute_tasks_new_python_interpreter")`; track subprocess stderr temp files by PID; upload subprocess stderr details on failure; preserve fork result-queue handling |
| `providers/edge3/tests/unit/edge3/cli/test_worker.py` | Add coverage for launch-mode routing, subprocess command construction, stderr spooling, and failed subprocess log upload |

## Notes

- Fork remains the default behavior.
[...]

PR #46265, opened 2025-01-29
Swap CeleryExecutor over to use TaskSDK for execution.

ashb:
Closes #45426

Some points of note about this PR:

- Logging is changed in Celery, but only for Airflow 3

  Celery does it's own "capture stdout" logging, which conflicts with the ones
  we do in the TaskSDK, so we disable that; but to not change anything for
  Airflow 3.

- Simplify task SDK logging redirection

  As part of this discovery that Celery captures stdout/stderr itself (and
  before disabling that) I discovered a simpler way to re-open the
  stdin/out/err so that the implementation needs fewer/no special casing.

- Make JSON task logs more readable by giving them a consistent/useful order

  We re-order (by re-creating) the event_dict so that timestamp, level, and
  then even are always the first items in the dict

- Makes the CeleryExecutor understand the concept of "workloads" instead a
  command tuple.

  This change isn't done in the best way, but until Kube executor is swapped
  over (and possibly the other in-tree executors, such as ECS) we need to
  support both styles concurrently.

  The change should be done in such a way that the provider still works with
  Airflow v2, if it's running on that version.

- Upgrade Celery

  This turned out to not be 100% necessary but it does fix some deprecation
  warnings when running on Python 3.12

- Ensure that the forked process in TaskSDK _never ever_ exits

  Again, this isn't possible usually, but since the setup step of `_fork_main`
  died, it didn't call `os._exit()`, and was caught further up, which meant
  the process stayed alive as it never closed the sockets properly. We put and
  extra safety try/except block in place to catch that

I have not yet included a newsfragment for changing the executor interface as
the old style is _currently_ still supported.

Testing: I ran `airflow scheduler`, `airflow fastapi-api`, and `airflow celery
worker` and triggered some dags, viewed logs etc.

And since a picture makes all PRs better

Screenshot 2025-01-29 at 22 49 49  

Celery window showing debug logs from the supervisor itself (task output goes to file)

Screenshot 2025-01-29 at 22 50 56

PR #70901, opened 2026-08-01
Flag `execute_tasks_new_python_interpreter` as Edge-only in docs and lint

ignacioparicio:
`[core] execute_tasks_new_python_interpreter` is still documented as a live option, but since Airflow 3.0 CeleryExecutor and LocalExecutor ignore it. The only thing left that reads it is the Edge worker.

That happened in https://github.com/apache/airflow/pull/46265 (I _think_ as an unintended side effect), as the code that reads the variable is behind a `if not AIRFLOW_V_3_0_PLUS:` gate. What is left is somewhat inconsistent.

<details>
<summary> Why it's inconsistent </summary>

-  CeleryExecutor and LocalExecutor silently ignore `execute_tasks_new_python_interpreter`. (Note: The Edge worker does honour it on Airflow 3, added in #65943 ).
- `config.yml` documents the option under `[core]` with no deprecation marker.
- `plugins.rst` still tells users to set it to pick up plugin changes without restarting the worker.
- `airflow config lint`, the designated 2 to 3 upgrade check, does not mention it, so an upgrade gives no signal at all.

</details>

What this PR does:
  - `config.yml`:  adds a note to the option saying that since Airflow 3.0 only the Edge worker reads it.
  - `plugins.rst`: removes the advice to set it (that workaround no longer works on Celery or Local).
  - `airflow config lint`: adds an entry so the 2-to-3 upgrade check flags the option when it is set, with a pointer to what changed.

No behaviour changes. The goal is just that an Airflow 2->3 upgrade gives some signal instead of none.

If that's OK I leave restoring the functionality of `execute_tasks_new_python_interpreter` to a separate discussion (#70280 looks like the right place).

<details>
<summary> Optional context: how we ran into it </summary>

We hit this upgrading a CeleryExecutor deployment from 2.11 to 3.3. We had the option set to `True`, so every task ran in a fresh interpreter. On 3.3 tasks are forked from the Celery prefork worker instead, and any task that starts a `multiprocessing` child now inherits the worker's daemon flag and dies with `daemonic processes are not allowed to have children`. Nothing in the upgrade told us the option had stopped working; we found it by reading the provider source.

</details>

---

Used to map the blast radius: which code paths, docs, and CLI checks still reference the option, and which components still read it.

### DIFF
--- a/airflow-core/docs/administration-and-deployment/plugins.rst
+++ b/airflow-core/docs/administration-and-deployment/plugins.rst
@@ -95,9 +95,9 @@ code you will need to restart those processes. However, it will not be reflected
 By default, task execution uses forking. This avoids the slowdown associated with creating a new Python interpreter
 and re-parsing all of Airflow's code and startup routines. This approach offers significant benefits, especially for shorter tasks.
 This does mean that if you use plugins in your tasks, and want them to update you will either
-need to restart the worker (if using CeleryExecutor) or scheduler (LocalExecutor). The other
-option is you can accept the speed hit at start up set the ``core.execute_tasks_new_python_interpreter``
-config setting to True, resulting in launching a whole new python interpreter for tasks.
+need to restart the worker (if using CeleryExecutor) or scheduler (LocalExecutor). The
+``core.execute_tasks_new_python_interpreter`` setting used to offer a way around that, but since
+Airflow 3.0 it is only honoured by the Edge executor.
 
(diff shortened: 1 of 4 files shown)


ignacioparicio:
I still think this docs-only change would save others some headaches, especially since this was an Airflow 3 regression

ignacioparicio:
on second look looks like this is superseded by #72164, which restores `execute_tasks_new_python_interpreter`. Closing this.

PR #70902, opened 2026-08-01
Refuse a separator-bearing secret id before the team scoped lookup

potiuk:
The team scoped lookup builds `PREFIX + _<TEAM>___ + <ID>`, so a secret id that
itself contains `___` makes that string ambiguous:

```
caller in team "team_a", bare id "prod___dbconn"
  AIRFLOW_CONN_ + _TEAM_A + ___ + PROD___DBCONN   ->  AIRFLOW_CONN__TEAM_A___PROD___DBCONN

team "team_a___prod", its own id "dbconn"
  AIRFLOW_CONN_ + _TEAM_A___PROD + ___ + DBCONN   ->  AIRFLOW_CONN__TEAM_A___PROD___DBCONN
```

Byte-identical — and the **scoped** lookup hits, so the existing guard, which ran
only ahead of the team agnostic fall-through, was never reached for this shape.

### Approach

Two changes, both small:

1. **Move the check ahead of both lookups** rather than only the fall-through.
2. **Widen it** from "spells out a team namespace" (`_<TEAM>___<ID>`) to simply
   "contains the separator".

The narrower form had to reason about *which* team an id might name. That is
unanswerable while a team name may itself contain the separator, and it made the
guard depend on every stored team name being valid — which `teams sync` has never
enforced. Asking only "does this contain `___`" removes both problems, and drops
`_names_a_team_namespace` entirely.

This is also what the provider secrets backends already do (`TEAM_SEP in
secret_id`, applied to every getter). Core was the outlier.

### Behaviour change

A Connection or Variable whose **id** contains `___` is no longer resolvable from
the environment backend — in either scope, including for its owning team.

That is deliberate. The name such an id builds is one another team's name could
build, and nothing in the string says which reading was meant. The owning team
reaches its own secret the normal way, with the bare id plus its team scope, which
is unaffected. Ids without the separator are unaffected in every scope.

### Test plan

- [x] `test_a_bare_id_containing_the_separator_is_not_resolved` — the collision
      above, both lookups
- [x] `test_an_id_containing_the_separator_is_refused_in_every_scope` — one rule,
      with and without a team scope
- [x] `test_the_owning_team_cannot_reach_a_separator_bearing_id_either` — pins the
      cost, so a later relaxation cannot quietly reopen the hole
- [x] Verified against the code with only the two guards removed (constant kept, so
      the failure is behavioural not an import error): **26 failed**; with them,
      **46 passed**
- [x] All 34 pre-existing tests in the file pass unchanged

### DIFF
--- a/airflow-core/src/airflow/secrets/environment_variables.py
+++ b/airflow-core/src/airflow/secrets/environment_variables.py
@@ -36,11 +36,8 @@ class EnvironmentVariablesBackend(BaseSecretsBackend):
 
     def get_conn_value(self, conn_id: str, team_name: str | None = None) -> str | None:
         if TEAM_SEP in conn_id:
-            # Refused ahead of the team scoped lookup, not only ahead of the team agnostic
-            # one. The scoped lookup builds PREFIX + _<TEAM>___ + <ID>, so an id that itself
-            # contains the separator makes that string ambiguous: a caller in team `team_a`
-            # asking for `prod___dbconn` builds exactly what team `team_a___prod` builds for
-            # id `dbconn`. That lookup *hits*, so a guard placed after it never runs.
+            # An id containing the separator could collide with another team's namespace
+            # even on the scoped lookup below, so it must be refused before either runs.
             return None
 
         if team_name and (
@@ -60,11 +57,7 @@ class EnvironmentVariablesBackend(BaseSecretsBackend):
         :return: Variable Value
         """
         if TEAM_SEP in key:
-            # Refused ahead of the team scoped lookup, not only ahead of the team agnostic
-            # one. The scoped lookup builds PREFIX + _<TEAM>___ + <ID>, so an id that itself
-            # contains the separator makes that string ambiguous: a caller in team `team_a`
-            # asking for `prod___dbconn` builds exactly what team `team_a___prod` builds for
-            # id `dbconn`. That lookup *hits*, so a guard placed after it never runs.
+            # Same collision risk as get_conn_value, see its code comment.
             return None
 
         if team_name and (
--- a/airflow-core/src/airflow/secrets/environment_variables.py
+++ b/airflow-core/src/airflow/secrets/environment_variables.py
@@ -20,26 +20,34 @@
 from __future__ import annotations
 
 import os
-import re
 
 from airflow.secrets import BaseSecretsBackend
 
 CONN_ENV_PREFIX = "AIRFLOW_CONN_"
 VAR_ENV_PREFIX = "AIRFLOW_VAR_"
 
+# Separates the team name from the secret id in a team namespaced environment variable
+# name: AIRFLOW_CONN__<TEAM>___<ID>.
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/secrets/environment_variables.py
```diff
@@ -20,27 +20,35 @@
...
 class EnvironmentVariablesBackend(BaseSecretsBackend):
     """Retrieves Connection object and Variable from environment variable."""
 
     def get_conn_value(self, conn_id: str, team_name: str | None = None) -> str | None:
+        if TEAM_SEP in conn_id:
+            # Refused ahead of the team scoped lookup, not only ahead of the team agnostic
+            # one. The scoped lookup builds PREFIX + _<TEAM>___ + <ID>, so an id that itself
+            # contains the separator makes that string ambiguous: a caller in team `team_a`
+            # asking for `prod___dbconn` builds exactly what team `team_a___prod` builds for
+            # id `dbconn`. That lookup *hits*, so a guard placed after it never runs.
```
amoghrajesh:
```suggestion
            # An id containing the separator could collide with another team's namespace
            # even on the scoped lookup below, so it must be refused before either runs.
```

nit.

airflow-core/src/airflow/secrets/environment_variables.py
```diff
@@ -51,39 +59,18 @@ def get_variable(self, key: str, team_name: str | None = None) -> str | None:
         :param team_name: Team name associated to the task trying to access the variable (if any)
         :return: Variable Value
         """
+        if TEAM_SEP in key:
+            # Refused ahead of the team scoped lookup, not only ahead of the team agnostic
+            # one. The scoped lookup builds PREFIX + _<TEAM>___ + <ID>, so an id that itself
+            # contains the separator makes that string ambiguous: a caller in team `team_a`
+            # asking for `prod___dbconn` builds exactly what team `team_a___prod` builds for
+            # id `dbconn`. That lookup *hits*, so a guard placed after it never runs.
```
amoghrajesh:
```suggestion
            # Same collision risk as get_conn_value, see its code comment.
```
nit

amoghrajesh (approved):

airflow-core/src/airflow/secrets/environment_variables.py
potiuk: Applied, both of them. The block was pasted verbatim into `get_conn_value` and `get_variable`, which is exactly the repeat-the-same-rationale case AGENTS.md rules out — the reasoning now lives once and the second site points at it.

---

airflow-core/src/airflow/secrets/environment_variables.py
potiuk: Applied, both of them. The block was pasted verbatim into `get_conn_value` and `get_variable`, which is exactly the repeat-the-same-rationale case AGENTS.md rules out — the reasoning now lives once and the second site points at it.

---

PR #70904, opened 2026-08-01
Share airflowctl's Dag run lookup without a new request for run_id

rjgoyln:
## Summary

`airflowctl tasks failed-deps` and `airflowctl tasks states-for-dag-run` intentionally **do not** look up a Dag run when the user supplies a `run_id`; the ID is passed directly to the task-instance endpoint. That behavior was introduced in #69915 to avoid an unnecessary HTTP request, but the remaining logical-date lookup and selector validation were duplicated across commands.

This PR moves the shared logic into `airflowctl/ctl/utils/dag_run.py`, removing the duplication while preserving the existing request pattern and user-visible behavior.

## Why two resolvers?

The split is intentional.

- `resolve_dag_run_id` returns a supplied `run_id` directly, avoiding the extra Dag run lookup.
- `resolve_dag_run` fetches the Dag run because commands such as `dags state` need the full response to display fields such as `state` and `conf`.

Both helpers share the selector validation and logical-date lookup, but keep the different `run_id` behavior intact.

A single parameterized resolver was considered, but since the two helpers naturally return different types (`str` vs `DAGRunResponse`), combining them would require either a union return type or `@overload` definitions. That adds more complexity than the small amount of dispatch logic it would eliminate.

## Tests

The "don't look up a supplied `run_id`" behavior was previously not enforced by the test suite.

As a demonstration, adding a verification-only `dag_runs.get` call (for example, to validate a supplied `run_id`) produces the following result:

| Branch | Result |
| --- | --- |
| `main` | **30/30 tests pass** |
| This branch | **Exactly 2 tests fail** (the assertions added here) |

The new assertions make this behavior an explicit contract and prevent future refactors from unintentionally adding the extra request.

## Notes

- No behavior change is intended. Request counts, error messages, and exit paths remain unchanged.
- No newsfragment: `airflow-ctl` release managers regenerate the changelog from `git log`.

---

### DIFF
--- a/airflow-ctl/src/airflowctl/ctl/commands/dag_command.py
+++ b/airflow-ctl/src/airflowctl/ctl/commands/dag_command.py
@@ -27,7 +27,6 @@ from rich.text import Text
 
 from airflowctl.api.client import (
     NEW_API_CLIENT,
-    Client,
     ClientKind,
     ServerResponseError,
     provide_api_client,
@@ -35,9 +34,9 @@ from airflowctl.api.client import (
 from airflowctl.api.datamodels.generated import (
     ClearTaskInstancesBody,
     DAGPatchBody,
-    DAGRunResponse,
 )
 from airflowctl.ctl.console_formatting import AirflowConsole
+from airflowctl.ctl.utils.dag_run import resolve_dag_run
 
 
 def update_dag_state(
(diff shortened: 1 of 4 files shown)


SameerMesiah97 (approved):
Looks good. This is a relatively low-risk refactor,

potiuk:
**Heads-up on an incoming conflict — not a review.**

`@haseebmalik18`'s #71206 adds an `airflowctl tasks state` command, and it brings a third copy of the block this PR consolidates:

```python
if (args.run_id is None) == (args.logical_date is None):
    rich.print("[red]Provide either run_id or --logical-date, but not both[/red]")
    sys.exit(1)
run_id = args.run_id or _find_run_id_by_logical_date(api_client, args.dag_id, args.logical_date)
```

I've approved that PR, so it will most likely land first. When it does, this one will conflict in `task_command.py`, and the resolution is small: the new `state` command's block collapses into another `resolve_dag_run_id(api_client, args)` call, next to the two in `failed_deps` and `states_for_dag_run`.

Nothing to do right now. Flagging it so the conflict isn't a surprise when it turns up, and so the extra call site reads as expected rather than as something that crept in while this was waiting.

This note is only about the overlap — I haven't left a review on this PR itself.

---

> *This message was drafted by an AI-assisted tool and confirmed by an Airflow maintainer. If anything here looks wrong, please reply on the PR and a maintainer will follow up.*

---

potiuk (approved):
Clean refactor — the two lookup helpers move verbatim, the duplicated logical-date lookup in `task_command.py` collapses into the shared one, and keeping `resolve_dag_run` / `resolve_dag_run_id` separate (with the docstring saying why a supplied `run_id` isn't fetched) is the right call. The new `assert_not_called()` checks turn that "no extra request" behaviour into an enforced contract. Approving.

---

> will follow up.*
>

---

PR #70906, opened 2026-08-01
Fix Keycloak access for individual roles

### LINKED ISSUE #70251: create-team in multi-team failing to allow access to team DAGs

Providers

### Apache Airflow version

3.3.0

In the docker compose file the command:
`airflow keycloak-auth-manager create-team team-analytics --username "$${KEYCLOAK_ADMIN}" --password "$${KEYCLOAK_ADMIN_PASSWORD}"`

was used to create a team.  This generated ReadOnly-team-analytics with a decision strategy of UNANIMOUS.

However, this prevented Op and User level roles in that team from being able to access the DAGs since only one of Allow-Op-team-analytics and Allow-User-team-analytics was true. Essentially this is impossible, since only one of these roles will be true, they can never be unanimous.

Manually switching the decision strategy for ReadOnly-team-analytics to AFFIRMATIVE with the keycloak fixed the problem.

This should automatically work when create-team is run.

Claude suggested the following patch:
```diff
diff --git a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
index 705fc4eb0b..ab69caafa4 100644
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
@@ -792,6 +792,7 @@ def _attach_team_permissions(
         policy_name=_team_role_policy_name(team, "Viewer"),
         scope_names=["GET", "LIST"],
         resource_names=team_readable_resources,
+        decision_strategy="AFFIRMATIVE",
         _dry_run=_dry_run,
     )
     for role_name in ("User", "Op", "Admin"):
@@ -802,6 +803,7 @@ def _attach_team_permissions(
             policy_name=_team_role_policy_name(team, role_name),
             scope_names=["GET", "LIST"],
             resource_names=team_readable_resources,
+            decision_strategy="AFFIRMATIVE",
             _dry_run=_dry_run,
         )
     _attach_policy_to_scope_permission(
```

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Docker-Compose

keycloak

apache-airflow-providers-keycloak==0.8.1

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

I think this would occur anytime that create-team is used and someone tries to access the DAGs with less than Admin permission.

### Code of Conduct

dhkim1920:
The Keycloak provisioning commands attached alternative role policies to
scope permissions while retaining an `UNANIMOUS` decision strategy. This
affected both:

- the team ReadOnly permission for Viewer, User, Op, and Admin roles
- the default non-team global Admin permission for Admin and SuperAdmin roles

These roles are alternatives, but `UNANIMOUS` required users to satisfy every
attached role policy. As a result, users holding only one intended role could be
denied access.

Use an `AFFIRMATIVE` strategy for both permissions and extend the existing
regression tests to verify each alternative role policy.

Compatibility note:

Upgrading the provider does not update existing Keycloak permissions. For each
existing team, run `airflow keycloak-auth-manager create-team <team>` again to
update the team ReadOnly permission. For non-team installations, run
`airflow keycloak-auth-manager create-permissions` again without `--teams` to
update the global Admin permission.

Because these commands preserve policies already attached to the permissions,
manually added custom policies will also be evaluated under the `AFFIRMATIVE`
strategy after the update.

Validation:

- `breeze ci selective-check --commit-ref HEAD`
- `breeze testing providers-tests --test-type "Providers[common.compat,keycloak]"`
  (`459 passed, 3 skipped`)

closes: #70251

---

- [X] Yes — OpenAI Codex (GPT-5)

Generated-by: OpenAI Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/keycloak/docs/changelog.rst
+++ b/providers/keycloak/docs/changelog.rst
@@ -25,6 +25,16 @@
 Changelog
 ---------
 
+.. note::
+    Upgrading the provider does not update existing Keycloak permissions. For each existing team,
+    run ``airflow keycloak-auth-manager create-team <team>`` again to update the team ReadOnly
+    permission. For non-team installations, run
+    ``airflow keycloak-auth-manager create-permissions`` again without ``--teams`` to update the
+    global Admin permission.
+
+    Manually added policies attached to these permissions will also be evaluated under the
+    ``AFFIRMATIVE`` strategy after the update.
+
 0.9.0
 .....
 
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
@@ -792,6 +792,7 @@ def _attach_team_permissions(
         policy_name=_team_role_policy_name(team, "Viewer"),
         scope_names=["GET", "LIST"],
         resource_names=team_readable_resources,
+        decision_strategy="AFFIRMATIVE",
         _dry_run=_dry_run,
     )
     for role_name in ("User", "Op", "Admin"):
@@ -802,6 +803,7 @@ def _attach_team_permissions(
             policy_name=_team_role_policy_name(team, role_name),
             scope_names=["GET", "LIST"],
             resource_names=team_readable_resources,
+            decision_strategy="AFFIRMATIVE",
             _dry_run=_dry_run,
         )
     _attach_policy_to_scope_permission(
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
@@ -301,6 +301,7 @@ def _attach_default_role_permissions(
             policy_name=_role_policy_name(role_name),
             scope_names=_get_extended_resource_methods() + ["LIST"],
             resource_names=[],
+            decision_strategy="AFFIRMATIVE",
             _dry_run=_dry_run,
         )
 
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
@@ -929,6 +929,7 @@ def _attach_superadmin_permissions(
         policy_name=_role_policy_name(SUPER_ADMIN_ROLE_NAME),
         scope_names=_get_extended_resource_methods() + ["LIST"],
         resource_names=team_scoped_resources,
+        decision_strategy="AFFIRMATIVE",
         _dry_run=_dry_run,
     )
     _attach_policy_to_scope_permission(
--- a/providers/keycloak/tests/unit/keycloak/auth_manager/cli/test_commands.py
+++ b/providers/keycloak/tests/unit/keycloak/auth_manager/cli/test_commands.py
@@ -567,15 +567,17 @@ class TestCommands:
         mock_update_admin_permission_resources.assert_called_once_with(client, "test-id", _dry_run=False)
         mock_ensure_group_policy.assert_called_once_with(client, "test-id", "team-a", _dry_run=False)
         assert mock_ensure_aggregate_policy.call_count == 4
-        mock_attach_policy.assert_any_call(
-            client,
-            "test-id",
-            permission_name="ReadOnly-team-a",
-            policy_name="Allow-Viewer-team-a",
-            scope_names=["GET", "LIST"],
-            resource_names=["Dag:team-a", "Team:team-a"],
-            _dry_run=False,
-        )
+        for role_name in TEAM_ROLE_NAMES:
+            mock_attach_policy.assert_any_call(
+                client,
+                "test-id",
+                permission_name="ReadOnly-team-a",
+                policy_name=f"Allow-{role_name}-team-a",
+                scope_names=["GET", "LIST"],
+                resource_names=["Dag:team-a", "Team:team-a"],
+                decision_strategy="AFFIRMATIVE",
+                _dry_run=False,
+            )
         mock_attach_policy.assert_any_call(
             client,
             "test-id",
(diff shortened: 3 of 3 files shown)


dhkim1920:
The additional issue concerning the missing team Dag resource in the Op permission is not addressed by this PR and should be tracked separately.

potiuk:
Good diagnosis. Four alternative role policies attached to one permission with `UNANIMOUS` meant a member had to satisfy Viewer *and* User *and* Op *and* Admin simultaneously to read their team's Dags — which nobody is, so no non-admin could read anything.

The fix matches what the file already does elsewhere rather than inventing a policy: `AFFIRMATIVE` is used in six other places for this same multi-alternative-role shape, including the global `ReadOnly` a few hundred lines up, which is the direct analogue of the team `ReadOnly` you are fixing. So this reads as a case missed when team support was added. The single-policy permissions (`Admin-{team}`, the super-admin `Admin`) correctly keep the `UNANIMOUS` default, which is harmless with one policy.

One question: does the same apply to the global `Admin` permission?

```python
for role_name in ("Admin", SUPER_ADMIN_ROLE_NAME):
    _attach_policy_to_scope_permission(
        client,
        client_uuid,
        permission_name="Admin",
        policy_name=_role_policy_name(role_name),
        scope_names=_get_extended_resource_methods() + ["LIST"],
        resource_names=[],
        _dry_run=_dry_run,
    )
```

That is two alternative policies on one permission with no `decision_strategy`, so it also gets `UNANIMOUS`. By the same reasoning as this PR, an Admin who is not also a super-admin would fail it. I have not traced how those two policies evaluate for a given user, so I may be wrong about whether it bites in practice — but it is the identical shape, and your own argument applies to it. Worth either folding into this PR or splitting out, whichever you prefer.

On the compatibility note: since existing Keycloak permissions are not updated by this, operators need to re-run `create-team` or change the strategy by hand. That is an action for deployments rather than just a code fix, so it should land in the provider changelog and not only in the PR description — otherwise anyone upgrading will still have the broken permission and no obvious reason why.

---

dhkim1920:
@potiuk 

Thanks for catching this. I traced the global `Admin` permission and confirmed that `Admin` and `SuperAdmin` are separate alternative role policies, so the same issue applies there as well.

  I folded the fix into this PR:

  - Set the default non-team global `Admin` permission to `AFFIRMATIVE` for both the `Admin` and `SuperAdmin` policies.
  - Kept the default `UNANIMOUS` strategy for permissions with only one attached policy.
  - Extended the existing non-team test to verify both global administrator roles.
  - Extended the multi-team test to verify that Viewer, User, Op, and Admin policies all use `AFFIRMATIVE`.
  - Added a provider changelog note explaining that existing teams must rerun `create-team <team>`, while non-team installations must rerun `create-permissions`. It also notes the effect on manually attached
  policies.

  The new global Admin assertion failed before the production change because the expected `AFFIRMATIVE` strategy was missing.

  Validation completed:

  - `breeze testing providers-tests --test-type "Providers[common.compat,keycloak]"` — 459 passed, 3 skipped
  - `breeze ci selective-check --commit-ref HEAD`

  Thanks again for identifying the analogous global permission case.

  ---

  Drafted-by: OpenAI Codex (GPT-5); reviewed by @dhkim1920 before posting

joshua-cogliati-inl:
> The additional issue concerning the missing team Dag resource in the Op permission is not addressed by this PR and should be tracked separately.

Do you want me to create a separate issue? (the alternative is for this pull not to cause https://github.com/apache/airflow/issues/70251 to close.)

potiuk (commented):
The global `Admin` half looks right — thanks for tracing it. One spot in the same file still undoes it.

### `_attach_superadmin_permissions` rewrites `Admin` back to `UNANIMOUS`

`_attach_policy_to_scope_permission` PUTs the whole permission payload, `"decisionStrategy"` included (`commands.py:1165-1178`) — so every call rewrites the strategy rather than adding to it. At `commands.py:925-933` it is called for `permission_name="Admin"` without `decision_strategy`, landing on the `UNANIMOUS` default, while the two sibling calls in the same function (`ViewAccess` at :941, `MenuAccess` at :951) both pass `AFFIRMATIVE`.

The path where that bites: an install that ran `create-permissions` without `--teams` has global `Admin` carrying both `Allow-Admin` and `Allow-SuperAdmin` — `AFFIRMATIVE` after this PR. If that deployment later sets `core.multi_team=True` and runs `create-team <team>`, this call rewrites the same permission back to `UNANIMOUS` with both policies still attached, which is the exact failure mode #70251 describes. Pure multi-team installs only ever have `Allow-SuperAdmin` on that permission, which is why nothing fails today.

Suggested: pass `decision_strategy="AFFIRMATIVE"` there too, so the outcome no longer depends on which command ran last, and add the kwarg to the existing assertion at `test_commands.py:629`.

### On `closes: #70251`

Answering the question raised in the thread: keep it. #70251 is specifically about `ReadOnly-team-*` being created with `UNANIMOUS`, and this PR fixes that. The missing team Dag resource on the `Op` permission is a separate bug — please open its own issue and link it here so it isn't lost (see the "Tracking issues for deferred work" section of `AGENTS.md`).

### Verified clean

- Changelog note placement matches the convention stated in the `NOTE TO CONTRIBUTORS` block at the top of `changelog.rst`, and the release tooling inserts new version sections below it, so it survives to release.
- Calling out that manually attached policies become `AFFIRMATIVE` is the right disclosure — this change loosens evaluation, and an operator who attached a restricting policy to one of these permissions needs to know.
- Both new call sites are covered by the extended tests; nothing changed without a test.

---

>

---

dhkim1920:
@potiuk @joshua-cogliati-inl 
Thanks for the detailed review. I fixed the remaining case in `_attach_superadmin_permissions` by explicitly setting `decision_strategy="AFFIRMATIVE"` for the global `Admin` permission.

I also updated the existing test assertion to verify the strategy.

The separate `Op` permission issue is now tracked in #71319.

potiuk (approved):

PR #70918, opened 2026-08-01
Replace external hclog-slog adapter with an in-tree one in the Go SDK

jason810496:
## Why

In plugin mode, `bundlev1server` installs an slog default logger that forwards through an `hclog.Logger` so the go-plugin host can parse the plugin's stderr. That bridge came from the external `github.com/evanphx/go-hclog-slog` module, which ships **no license file and no tagged release** (it is pinned to a bare commit). That makes it awkward to depend on and to reason about.

The adapter is small and we already maintain its mirror image in-tree — `pkg/logging/shclog` presents an slog Logger *as* an hclog. This adds the inverse and drops the external module.

## What

- Add `pkg/logging/hclogslog`: a small `slog.Handler` that forwards records to an `hclog.Logger`, translating levels and flattening grouped attributes with dotted keys (100% unit-test covered).
- Point `bundlev1server` at the in-tree package (the call site — `hclogslog.Adapt(hcLogger)` — is unchanged).
- `go mod tidy` drops `github.com/evanphx/go-hclog-slog` from `go-sdk/go.mod` / `go-sdk/go.sum`; re-tidy `kubernetes-tests/lang_sdk/go_example` to drop its stale indirect entry.

Plugin logging behaviour is identical.

---

### DIFF
--- a/go-sdk/bundle/bundlev1/bundlev1server/server.go
+++ b/go-sdk/bundle/bundlev1/bundlev1server/server.go
@@ -22,7 +22,6 @@ import (
 	"log/slog"
 	"os"
 
-	"github.com/evanphx/go-hclog-slog/hclogslog"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-plugin"
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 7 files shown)


potiuk (approved):

potiuk:
good calls!

PR #70919, opened 2026-08-01
Strip only the literal properties/ prefix from GA property link IDs

Eason09053360:
`GoogleAnalyticsAdminCreatePropertyOperator` builds the property link ID with
`prop.name.lstrip("properties/")`. `str.lstrip` removes any leading character
belonging to the given set — it is not a prefix strip — so it also consumes any
leading `p`, `r`, `o`, `e`, `t`, `i`, `s` or `/` of the property ID itself.

This is correct today only because GA4 property IDs are numeric, so there is no
user-visible behaviour change. The intent is a prefix strip, so `removeprefix`
expresses it directly and removes the latent trap.Improve semantics and readability

The `persist` call had no test coverage; the added parametrized test pins the
exact-prefix semantics and fails on the previous implementation.

---

### DIFF
--- a/providers/google/src/airflow/providers/google/marketing_platform/operators/analytics_admin.py
+++ b/providers/google/src/airflow/providers/google/marketing_platform/operators/analytics_admin.py
@@ -194,7 +194,7 @@ class GoogleAnalyticsAdminCreatePropertyOperator(GoogleCloudBaseOperator):
         self.log.info("The Google Analytics property %s was created successfully.", prop.name)
         GoogleAnalyticsPropertyLink.persist(
             context=context,
-            property_id=prop.name.lstrip("properties/"),
+            property_id=prop.name.removeprefix("properties/"),
         )
 
         return Property.to_dict(prop)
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Correct fix for a genuine latent trap. `str.lstrip` takes a character set rather than a prefix, so it also eats any leading `p`, `r`, `o`, `e`, `t`, `i`, `s` or `/` belonging to the ID:

```
properties/123456789  ->  lstrip: 123456789   removeprefix: 123456789
properties/s123       ->  lstrip: 123         removeprefix: s123
properties/porridge   ->  lstrip: dge         removeprefix: porridge
```

I appreciate that the description is straight about this not being user-visible today — GA4 property IDs are numeric, digits are not in the strip set, so `lstrip` stops at the first digit and happens to give the right answer. Fixing the intent rather than claiming a live bug is the right framing.

The parametrized `("properties/s123", "s123")` case is what makes this worth having: it returns `"123"` under the old implementation, so the test genuinely fails without the change instead of passing either way. Good that the inline comment records why that case exists, since it would otherwise look arbitrary. `persist` had no coverage at all before, so this closes that too.

One tidy-up before merge: the description has some leftover text — "removes the latent trap.Improve semantics and readability" — which reads like two drafts joined. Worth fixing, as the body becomes the squash commit message.

---

PR #70923, opened 2026-08-01
Fix airflow db clean never purging the callback table

steveahnahn:
`airflow db clean` has not purged the `callback` table since Airflow 3.2.0.

Migration `b87d2135fa50` renamed `callback_request` to `callback`, but the cleanup configuration in `db_cleanup.py` kept the old name. A configured table that does not exist is skipped with only a warning, so the entry became a no-op and nothing purges `callback` any more.

The table does accumulate in practice. Executor callbacks are set to `success`/`failed` by the scheduler and never deleted (only Dag-processor callbacks are removed, as they are dispatched). And because `deadline.callback_id` references `callback.id` with `ON DELETE CASCADE`, `callback` is the parent, so purging `deadline` leaves its callback rows orphaned permanently.

There was also no workaround available: `--tables callback` was rejected as an invalid choice, so operators could not purge the table through `airflow db clean` at all.

### Why this was not caught

`test_no_models_missing` exists to catch exactly this: every model table must either be cleaned or documented as deliberately excluded. It has been passing vacuously. It located the models package by walking up from the test file (`Path(__file__).parents[2] / "airflow/models"`), a path that stopped existing when the sources moved under `airflow-core/src`, so it discovered zero models and asserted over an empty set. Confirmed against the pre-fix code: the check passed while `callback` was in neither the cleanup config nor the exclusion list.

With the check repaired, it reports one more table with no way to be purged: `partitioned_asset_key_log`. It carries no foreign key, so its rows are left behind when the partition Dag run they describe is cascade-deleted with its `dag_run`. The only existing delete (in `SchedulerJobRunner`) covers partition runs invalidated by a rollup-definition change, which is not a retention mechanism.

Only those **orphans** are purged. Rows whose partition Dag run still exists are the evidence the scheduler evaluates to decide when that pending run fires (`SchedulerJobRunner` reads them for every pending `AssetPartitionDagRun`), so a purely age-based delete could silently stall a rollup still accumulating keys past the retention window. They are kept regardless of age via a `NOT IN (SELECT id FROM asset_partition_dag_run)` filter.

### What this changes

First commit — the callback fix:

* Register the table under its current name, with `created_at` as the recency column.
* Purge only callbacks that can no longer run. A callback still awaiting execution owns its `deadline` row through the `ON DELETE CASCADE` foreign key, so deleting one would silently drop a deadline that has not fired yet. Dag-processor callbacks carry no state and reference no deadline, so rows outliving the retention window are purged too. The existing `TERMINAL_STATES` constant is reused rather than re-hardcoding the state literals, since that kind of drift is what caused this bug.
* Add `dependent_tables=["deadline"]` so deadline rows are cleaned and archived ahead of their parent, matching how `dag` and `dag_run` already handle the same cascade.
* Drop the `sla_miss` entry, whose table no longer exists in Airflow 3.

Second commit — the coverage check and the table it found:

* Walk `airflow.models.__path__` instead of reconstructing the path from the test's location, so the check keeps working wherever the sources live (it now discovers 52 models, against 0 before), and assert that models were discovered at all so it can never pass on an empty set again.
* Register `partitioned_asset_key_log` for cleanup, restricted to orphaned rows as above.
* Record the remaining uncovered tables in the exclusion list with the reason each is safe — delete rules read from the live schema after `airflow db migrate`, not from the ORM metadata: `asset_partition_dag_run`, `asset_watcher`, `dag_favorite`, `hitl_detail`, `hitl_detail_history`, `task_inlet_asset_reference`, `deadline_alert`, and `asset_state_store` are all removed by `ON DELETE CASCADE` from tables that are already cleaned ; `asset_state_store` is a per-asset key/value store whose rows are upserted in place (bounded, current state rather than history); `team` is configuration, not run data.

The coverage check cannot pass without the callback fix — with it working, `callback` is reported as uncovered until the first commit registers it — which is why both land in one PR.

### Verification

Verified end to end against real metadata databases, seeding rows and running the real `run_cleanup`, with a `log` row as a control to prove the run did actual work:

| Case | Result |
| --- | --- |
| old + `success` / `failed` | purged |
| old + NULL state (Dag-processor) | purged |
| old + `scheduled` / `queued` / `running` | kept |
| recent + `success` | kept |
| unfired deadline owned by a live callback | survives |
| old orphaned `partitioned_asset_key_log` row | purged |
| old `partitioned_asset_key_log` row whose partition run still exists | kept |
| recent orphaned `partitioned_asset_key_log` row | kept |
| control `log` row | purged |

The archiving path (the default, `skip_archive=False`) was exercised on Postgres, MySQL and SQLite, confirming the `_airflow_deleted__callback__*` table is created and the purged row preserved; a committed test covers it. Removing `callback` from the config makes the repaired check flag it, so the check now catches the defect that motivated this PR. The full `test_db_cleanup.py` suite passes on all three backends, and the new tests fail without the source changes.

### Compatibility

Removing the two stale entries changes what `--tables` accepts. A list that mixes them with live tables still works, warning about the unknown names and cleaning the rest:

```
The following table(s) are not valid choices and will be skipped: ['callback_request', 'sla_miss']
Data prior to ... would be purged from tables ['log', 'job']
```

A `--tables` list naming *only* removed tables now stops with "No tables selected for db cleanup" instead of silently reporting success while doing nothing. That seems preferable, since such an invocation has been a no-op since 3.2.0, but it is a visible change for anyone whose cron passes only those names.

---

Generated-by: Claude Code (Opus 5, Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


airflow-core/tests/unit/utils/test_db_cleanup.py
```diff
@@ -1343,6 +1360,247 @@ def test_extra_filters_keep_in_flight_rows(self):
...
+
+    @pytest.mark.parametrize(
+        ("state", "should_survive"),
+        [
+            ("success", False),
+            ("failed", False),
+            (None, False),
+            ("scheduled", True),
+            ("queued", True),
+            ("running", True),
```
haseebmalik18, line 1449: Might be nice to add `pending` here too? It's in `ACTIVE_STATES` alongside `queued` and `running` (`models/callback.py:66`), and it's the state a fired executor callback sits in while it waits for the executor

steveahnahn: thanks for that, added `pending` in [7baad27](https://github.com/apache/airflow/pull/70923/commits/7baad27866be0096efde8553ad660ddb275c4f5d)

airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -205,7 +219,22 @@ def readable_config(self):
...
+        table_name="callback",
+        recency_column_name="created_at",
+        extra_columns=["state"],
+        # A callback that can still run owns its deadline row through an ON DELETE CASCADE
+        # foreign key, so purging one would silently drop a deadline that has not fired yet.
+        # Dag-processor callbacks carry no state and are deleted as they are dispatched; any
+        # that outlive the retention window were orphaned and no deadline references them.
+        extra_filters=[
+            or_(
+                column("state").in_(sorted(TERMINAL_STATES)),
```
ferruzzi, line 237: This may be backwards, isn't it?  Don't we want to check if `column("state").not_in(sorted(ACTIVE_STATES)) OR column("state").is_(None),`?  If a deadline is in SCHEDULED and its run is deleted, then it's orphaned as well and needs to be pruned, no?  It would also mean renaming or adding new states later become pruned by default which would have prevented this issue.

airflow-core/src/airflow/utils/db_cleanup.py
```diff
@@ -205,7 +219,22 @@ def readable_config(self):
     _TableConfig(table_name="task_reschedule", recency_column_name="start_date", dag_id_column_name="dag_id"),
     _TableConfig(table_name="xcom", recency_column_name="timestamp", dag_id_column_name="dag_id"),
     _TableConfig(table_name="_xcom_archive", recency_column_name="timestamp", dag_id_column_name="dag_id"),
-    _TableConfig(table_name="callback_request", recency_column_name="created_at"),
+    _TableConfig(
+        table_name="callback",
+        recency_column_name="created_at",
+        extra_columns=["state"],
```
ferruzzi: I'm surprised this doesn't need `id` in the `extra_columns`, it should be a required column.

ferruzzi (approved):
Thanks for the quick turnaround on that.  Approved pending CI passing.

PR #65604, opened 2026-04-21
Add cursor based pagination for get_dag_runs endpoint

pierrejeambrun:
Add cursor-based (keyset) pagination as an alternative to offset-based pagination on the `get_dag_runs` endpoint. Offset pagination remains the default and is not deprecated. Mirrors the pattern introduced for `get_task_instances` in #64845.

- Response schema adds `next_cursor` / `previous_cursor` and makes `total_entries` optional (populated only for offset pagination).
- Fetch `limit+1` rows to detect the last page and return `next_cursor=null` when no more results exist.
- Rewire the DagRuns listing page to use cursor pagination via the `DataTable`'s `cursorPagination` prop, which eliminates the `COUNT(*)` query for large datasets.

related: #64845

https://github.com/user-attachments/assets/896746d0-b362-47bb-a2d8-a39f7617effc

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.7 1M context)

Generated-by: Claude Code (Opus 4.7 1M context) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #70925, opened 2026-08-01
UI: add toggle to restore DagRun total count

### LINKED ISSUE #70856: DagRun count: UI regression

Airflow Core

### Apache Airflow version

> 3.2.2

The count of DagRuns has disappeared from the UI Dags -> Dag Runs between Airflow 3.2.1 and Airflow 3.2.2 (or any version above).

Image

Image

_No response_

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

vedjaw:
## What this PR does / why we need it

Closes #70856

After #65604, the Dag Runs UI always uses cursor pagination, so the API
no longer returns \`total_entries\` and the DataTable count heading
disappeared between 3.2.1 and 3.2.2.

This keeps cursor pagination as the default (fast path) and adds an
opt-in **Show total count** switch that uses offset pagination to restore
\`total_entries\` / the row count heading.

## Preference for how this is reviewed

- [x] I prefer that this PR is reviewed as a whole

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## How was this tested

Manual reasoning against \`DataTable\` / \`get_dag_runs\` cursor vs offset
contracts. Prefer UI review in breeze/dev UI.

## Docs

N/A (UI affordance with defaultValue label)

### DIFF
--- a/airflow-core/src/airflow/ui/src/pages/DagRuns/DagRuns.tsx
+++ b/airflow-core/src/airflow/ui/src/pages/DagRuns/DagRuns.tsx
@@ -21,6 +21,7 @@ import type { ColumnDef } from "@tanstack/react-table";
 import type { TFunction } from "i18next";
 import { useTranslation } from "react-i18next";
 import { useParams, useSearchParams } from "react-router-dom";
+import { useLocalStorage } from "usehooks-ts";
 
 import { useDagRunServiceGetDagRuns } from "openapi/queries";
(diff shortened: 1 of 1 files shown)


bbovenzi:
Instead of a switch we are doing a bounded count in https://github.com/apache/airflow/pull/71252

PR #67439, opened 2026-05-24
Fix branch skips inside mapped task groups

bobu-putheeckal:
Fixes: #67265

This PR fixes branch skipping for branch operators inside mapped TaskGroups.

## Problem

For branch operators inside `expand()` / `expand_kwargs()` mapped TaskGroups, non-selected branch siblings can run instead of being skipped. The skip decision is stored by the `SkipMixin` parent task instance at the runtime `map_index`, but `NotPreviouslySkippedDep` did not read that XCom using the mapped TaskGroup context for operators that are not themselves `MappedOperator` instances.

## What changed

- `NotPreviouslySkippedDep` now uses the runtime task instance `map_index` when reading SkipMixin XComs for a task whose parent is inside a mapped TaskGroup.
- Added regression coverage for both `expand()` and `expand_kwargs()` mapped TaskGroups.

## Validation

I ran these checks locally:

```console
AIRFLOW_SOURCES=/Users/agentzero/Projects/Open\ Source\ projects/airflow-worktrees/airflow-67265 uv run --no-default-groups --with "apache-airflow-devel-common[pytest,sqlalchemy] @ ./devel-common" --with "apache-airflow-providers-standard @ ./providers/standard" --with "apache-airflow-providers-common-compat @ ./providers/common/compat" pytest airflow-core/tests/unit/ti_deps/deps/test_not_previously_skipped_dep.py -q
# 8 passed

uv run --no-default-groups --with ruff ruff check airflow-core/src/airflow/ti_deps/deps/not_previously_skipped_dep.py airflow-core/tests/unit/ti_deps/deps/test_not_previously_skipped_dep.py
# all checks passed

git diff --check
# no whitespace errors
```

## Screenshot / GIF

A maintainer requested UI proof. I am moving this PR to draft until I can add a screenshot or short GIF showing the mapped TaskGroup branch behavior in the UI after the fix.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes: OpenAI Codex. I reviewed the generated changes and ran the focused checks listed above.

Generated-by: OpenAI Codex following https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions

Issue #62118, opened 2026-02-18
LatestOnlyOperator not working if just before a Dynamic Task Mapping


lcottereau:
### Apache Airflow version

Other Airflow 3 version (please specify below)

### If "Other Airflow 3 version" selected, which one?

3.1.5

### What happened?

When backfilling a DAG with `LatestOnlyOperator`, I realised the tasks after the LatestOnlyOperator were still executing.

I _think_ it used to work before (probably in 2.*) but can't confirm that anymore.

### What you think should happen instead?

When backfilling (before the current schedule), the tasks following the `LatestOnlyOperator` should be skipped, as per the purpose of this operator.

### How to reproduce

The following stripped-down code reproduces the problem in my environment, when I backfill a few days before the current day:
- for the `test_bug_OK` DAG, `intermediate_step` and `print_all` tasks are skipped
- for the `test_bug_KO` DAG, `print_all` tasks still execute

```python
from airflow.providers.standard.operators.latest_only import LatestOnlyOperator
from airflow.sdk import DAG, task
from pendulum import datetime

with DAG(
    dag_id="test_bug_KO",
    schedule="0 1 * * *",
    start_date=datetime(2026, 1, 1),
    catchup=False,
):
    is_latest = LatestOnlyOperator(task_id="is_latest")

    @task
    def print_label(label: str):
        print(label)

    print_all = print_label.expand(label=["Label 0", "Label 1"])

    is_latest >> print_all


with DAG(
    dag_id="test_bug_OK",
    schedule="0 1 * * *",
    start_date=datetime(2026, 1, 1),
    catchup=False,
):
    is_latest = LatestOnlyOperator(task_id="is_latest")

    @task
    def print_label(label: str):
        print(label)


    intermediate_step = print_label("Before")

    print_all = print_label.expand(label=["Label 0", "Label 1"])

    is_latest >> intermediate_step >> print_all
```

### Operating System

Debian 13.3

### Versions of Apache Airflow Providers

apache-airflow-providers-amazon==9.18.0
apache-airflow-providers-common-compat==1.10.0
apache-airflow-providers-common-io==1.7.0
apache-airflow-providers-common-sql==1.30.0
apache-airflow-providers-fab==3.0.1
apache-airflow-providers-http==5.6.0
apache-airflow-providers-postgres==6.5.0
apache-airflow-providers-samba==4.12.0
apache-airflow-providers-smtp==2.4.0
apache-airflow-providers-ssh==4.1.5
[...]

PR #70926, opened 2026-08-01
Fix branch skips inside mapped TaskGroups

### LINKED ISSUE #67265: branch operators inside expand/expand_kwargs-mapped TaskGroups in 3.2.x — non-selected branch siblings run instead of being skipped

Airflow Core

### Apache Airflow version

3.2.1

In the following minimal example, run_optional_step runs for each sibling (item_type `full`, and item_type `partial`), however `partial` is intended to be skipped::
```

"""Minimal example: @task.branch inside a dynamically mapped task group.
This example replicates the structure of a real-world DAG, where:
- One outer task group (process_items) fetches items and expands
    an inner task group (handle_item) via expand_kwargs.
- Inside handle_item, a @task.branch decides whether to run
    run_optional_step based on item_type.

Expected behavior
------------------
item_type == "full"    → branch returns absolute task ID → run_optional_step should run
item_type == "partial" → branch returns None             → run_optional_step should be skipped

Actual behavior
-------------------------
run_optional_step runs for ALL items.
"""

import pendulum
from airflow.sdk import DAG, task, task_group
from airflow.providers.standard.operators.empty import EmptyOperator

with DAG(
    dag_id="branch_in_mapped_taskgroup_example",
    schedule=None,
    start_date=pendulum.datetime(2024, 1, 1, tz="UTC"),
    catchup=False,
) as dag:

    # ------------------------------------------------------------------ #
    # Inner task group: handles one mapped item.                           #
    # The branch should skip or run an optional step per mapped instance.  #
    # ------------------------------------------------------------------ #
    @task_group(group_id="handle_item")
    def tg_handle_item(item_id: str, item_type: str):
        """Process one item.  Only 'full' items need the extra step."""

        @task(task_id="process")
        def process(item_id: str, item_type: str) -> str:
            print(f"Processing {item_id=} {item_type=}")
            return item_type

        result = process(item_id, item_type)

        @task.branch(task_id="select_optional_step")
        def select_optional_step(item_type: str):
            if item_type == "full":
                return ["process_items.handle_item.run_optional_step"]
            return None

        branch = select_optional_step(result)
        run_optional_step = EmptyOperator(task_id="run_optional_step")
        branch >> run_optional_step

    # ------------------------------------------------------------------ #
    # Outer task group: loads items and expands the inner group once per  #
    # item. There is only one outer group instance in the DAG.            #
    # ------------------------------------------------------------------ #
    @task_group(group_id="process_items")
    def tg_process_items():
        @task(task_id="get_items")
        def get_items() -> list[dict]:
            return [
                {"item_id": "item_1", "item_type": "full"},     # optional step should run
                {"item_id": "item_2", "item_type": "partial"},  # optional step should be skipped
            ]

        tg_handle_item.expand_kwargs(get_items())

    tg_process_items()

```

item_type == "full"    → branch returns absolute task ID → run_optional_step should run
item_type == "partial" → branch returns None             → run_optional_step should be skipped

This is what happens up through 3.1.7.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

This was originally discussed in https://github.com/apache/airflow/issues/65745, however, it seems like that issue is either not-reproducible or describing another issue.

[Further context from @martijn-exads](https://github.com/apache/airflow/issues/65745#issuecomment-4497401392):
We think we've traced it to https://github.com/apache/airflow/pull/62287, specifically this line in NotPreviouslySkippedDep:
xcom_map_index = ti.map_index if parent.is_mapped else -1

For a branch operator inside a mapped @task_group, the operator itself has no .expand() so parent.is_mapped is False, but its TIs run with map_index >= 0 and SkipMixin writes XCOM_SKIPMIXIN_KEY at the parent TI's per-map map_index. The dep then queries map_indexes=-1, finds nothing, returns "not skipped", and the non-selected sibling proceeds.

### Code of Conduct

vedjaw:
## Summary

Fixes #67265: `@task.branch` / `BranchPythonOperator` inside `expand()` / `expand_kwargs()` mapped TaskGroups did not skip non-selected siblings — those tasks **ran** (wrong side effects / wasted compute).

### Root cause

`NotPreviouslySkippedDep` used:

```python
xcom_map_index = ti.map_index if parent.is_mapped else -1
```

Operators inside a mapped TaskGroup are **not** `MappedOperator`s, but their TIs still write SkipMixin XCom at the runtime `map_index`. Looking up with `-1` misses that XCom, so the skip decision is invisible.

This is the inverse of #62118 (unmapped parent → mapped child must still use `-1`).

### Fix

```python
if parent.is_mapped or parent.get_closest_mapped_task_group() is not None:
    xcom_map_index = ti.map_index
else:
    xcom_map_index = -1
```

### Tests

- `expand` and `expand_kwargs`: non-selected sibling at the same `map_index` is skipped; selected sibling is not
- Cross-`map_index` isolation (XCom for index 0 must not skip index 1)
- Existing unmapped-parent → mapped-child coverage (#62118) retained

Note: #67439 attempted this but was stale-closed before merge.

## Test plan

- [ ] `pytest airflow-core/tests/unit/ti_deps/deps/test_not_previously_skipped_dep.py`
- [ ] CI


vedjaw (commented):
pls review the pr

kawashiro:
Hello

Can anyone review and merge this PR? We are really waiting to have this fixed.

Thank you in advance.

PR #70932, opened 2026-08-02
Prepare providers release 2026-08-01

potiuk:
Regular provider release wave for 2026-08-01. Versions and changelog entries were
classified with the `prepare-providers-documentation` workflow; all template-generated
files were regenerated by
`breeze release-management prepare-provider-documentation --reapply-templates-only`.

**47 providers released, 2 doc-only, 47 deliberately unreleased.** The unreleased ones
have pending commits, but only `#68991` (regenerated `README.rst`/`index.rst`) and/or
`#70256` (the previous wave's release commit), so there is nothing user-facing to ship.

**No major bumps in this wave.** Eight breaking-change candidates were reviewed against
the released tags and all resolved to minor or patch.

- **Minor (13):** alibaba 3.4.0, amazon 9.34.0, apache.livy 4.6.0, cncf.kubernetes 10.21.0,
  common.compat 1.18.0, common.sql 2.1.0, edge3 4.3.0, elasticsearch 6.9.0, fab 3.8.0,
  hashicorp 4.8.0, microsoft.azure 14.1.0, snowflake 6.16.0, standard 1.17.0
- **Patch (30):** airbyte 6.0.1 … yandex 4.5.1
- **Doc-only (2):** common.io, http

### Four providers were cut as rc1 last wave and never published

`google` (22.3.0), `common.ai` (0.7.0), `openlineage` (2.20.0) and `clickhousedb` (1.0.1)
are still at 22.2.2 / 0.6.0 / 2.19.0 / 1.0.0 on PyPI. Their existing in-tree sections are
**extended** rather than superseded — cutting a new version on top would publish a
changelog that skips a version users never received.

Two entries in those sections advertised PRs that were reverted before release and have
been moved to the excluded block, so the changelog no longer promises features that do not
ship:

- `common.ai`: `Add AWS services toolset for agents to access 1000+ APIs (#70087)`, fully
  reverted by #70695
- `google`: `Preserve Dataform workflow invocation config (#69161)`, reverted by #70395

### Other notes

- **This PR needs the `allow provider dependency bump` label.** `update-providers-next-version`
  resolved three `# use next version` pins to `apache-airflow-providers-common-compat>=1.18.0`
  (amazon, fab, keycloak); without the label `check_provider_dependency_bumps` fails CI.
- `providers/keycloak/pyproject.toml` contains **only** that pin resolution — no version bump
  and no changelog entry, so keycloak is not released in this wave.
- The google changelog carried a `.. note::` above its first version header, which rendered
  it as a page-level note rather than a release note. It now sits inside the 22.3.0 section
  it describes. That was pre-existing on `main`, not introduced here.
- Within-wave chains that net to no user-facing change (databricks #70341/#70551,
  google #69161/#70395, common.ai #70087/#70695, neo4j #70348/#70508) are excluded rather
  than listed as paired entries. Chains with a residual change are collapsed into a single
  entry naming every PR involved (e.g. standard #70345/#70391).
- All 47 released packages share one `source-date-epoch`; the four extended providers were
  still carrying the previous wave's value.

Static checks: `prek --hook-stage pre-commit` 65 passed / 0 failed, `--hook-stage manual`
4 passed / 0 failed. `--reapply-templates-only` re-runs are idempotent.

---


Miretpl (approved):

providers/amazon/docs/changelog.rst
```diff
@@ -26,6 +26,58 @@
...
+* ``Rename StepFunctionStartExecutionOperator's input field to match its constructor argument (#70544)``
+* ``Fix mypy arg-type error in Redshift Data get_primary_keys (#70677)``
+* ``Add IMPORT_ERRORS_ALL permission for import errors of files with no registered Dag (#69790)``
+* ``Tighten the DMS operator cdc-start checks to 'is not None' and raise ValueError (#70450, #70634)``
+
+Doc-only
+~~~~~~~~
+
+* ``Document effect of state-store cleanup for operators with durable execution (#70721)``
+* ``Remove hard-coded deferrable crawler run from example_glue system test (#70206)``
```
eladkal:
```suggestion
* ``Remove hard-coded deferrable crawler run from example_glue (#70206)``
```

Users shouldn't mind about our system tests. for them it is just example dag

providers/clickhousedb/docs/changelog.rst
```diff
@@ -39,7 +39,7 @@ Misc
    appropriate section above if needed. Do not delete the lines(!):
    * ``Fix ClickHouse test_connection unit test failing with newer clickhouse-connect (#70114)``
    * ``Fix inconsistency between generated provider docs and pyproject.toml (#68991)``
-
+   * ``Prepare providers release 2026-07-22 (#70256)``
```
eladkal, line 42: I think this needs to be removed since we don't have 1.0.1 eventually?

providers/clickhousedb/provider.yaml
eladkal, line 32: probably need to remove the 1.0.1?

providers/cncf/kubernetes/docs/changelog.rst
```diff
@@ -27,6 +27,44 @@
...
+.. note::
+    ``airflow kubernetes cleanup-pods`` now takes ``--min-completed-minutes``, which defaults to
+    ``1``. Terminal pods are therefore no longer deleted immediately; pass
+    ``--min-completed-minutes 0`` to restore the previous behaviour. The delay closes a race
+    where cleanup removed a pod before ``await_pod_completion`` had read its final state.
+
+Features
+~~~~~~~~
+
+* ``Add '--min-completed-minutes' to 'cleanup-pods' to prevent KPO race condition (#70595)``
```
eladkal:
```suggestion
* ``Add '--min-completed-minutes' to 'cleanup-pods' to prevent KubernetesPodOperator race condition (#70595)``
```

If we can avoid using KPO in changelog we should. Not all users know that KPO means KubernetesPodOperator. This is more our internal language. Maybe we need to adjust the skill to handle this

providers/cncf/kubernetes/docs/changelog.rst
```diff
@@ -27,6 +27,44 @@
...
+~~~~
+
+* ``Mark KubernetesPodOperator and AgentOperator as durable capable (#70289)``
+* ``Validate KubernetesResourceBaseOperator yaml_conf after rendering (#70337)``
+
+Doc-only
+~~~~~~~~
+
+* ``Document Kubernetes RBAC permissions (#68716)``
+* ``Documenting effect if task state store cleanup for KPO (#70722)``
```
eladkal:
```suggestion
* ``Documenting effect if task state store cleanup for KubernetesPodOperator (#70722)``
```

providers/common/ai/docs/changelog.rst
```diff
@@ -56,11 +65,17 @@ Bug Fixes
...
+* ``Redact secrets from LLMRetryPolicy classification prompts (#70229)``
+* ``Validate common AI tool-call arguments to enable pydantic-ai retries (#70096)``
+* ``Validate template fields after rendering in common.ai operators (#70338)``
 
 Misc
 ~~~~
 
 * ``Add dataclasses-json floor to common.ai llamaindex extra (#69755)``
+* ``Add a Retry Policies category to the provider registry (#70499)``
+* ``Remove noqa:S101 from production code (#70378)``
```
eladkal: Not very user friendly description.  
```suggestion
* ``Mark asserts under `TYPE_CHECKING` in `DocumentLoaderOperator` (#70378)``
```

providers/fab/docs/changelog.rst
```diff
@@ -20,6 +20,30 @@
...
+
+Bug Fixes
+~~~~~~~~~
+
+* ``Fix asset state store writes forbidden under FAB auth manager (#70478)``
+
+Misc
+~~~~
+
+* ``Bump eslint (#70697)``
```
eladkal:
```suggestion
* ``Bump eslint to 10.8.0 (#70697)``
```

providers/fab/docs/changelog.rst
```diff
@@ -20,6 +20,30 @@
...
+* ``Bump the fab-ui-package-updates group across 1 directory with 3 updates (#70604)``
```
eladkal:
```suggestion
* ``Bump prettier to 3.9.6, stylelint to 17.14.1, webpack to 5.109.0 (#70604)``
```

We probably need to adjust still to avoid such vague titles. It doesn't mean much to say that we updated 3 packages. We need to be explicit.

eladkal (approved):

providers/amazon/docs/changelog.rst
potiuk: Applied — dropped "system test" so it reads as an example Dag change from the user's point of view.

---

providers/clickhousedb/docs/changelog.rst
potiuk: Keeping 1.0.1 — it carries a real user-facing change: #69166 flips `lifecycle` from `production` to `incubation` in `provider.yaml`, and that ships to users through `get_provider_info.py`.

1.0.1 was cut as `rc1` in the 2026-07-22 wave but never published (PyPI is still on 1.0.0), so this wave publishes the existing unreleased section rather than creating a new version.

---

providers/clickhousedb/provider.yaml
potiuk: Same as the changelog thread — 1.0.1 stays. It was cut as `rc1` last wave and never published, and it contains the `lifecycle` change from #69166, so it gets released here rather than dropped.

---

providers/cncf/kubernetes/docs/changelog.rst
potiuk: Applied — "KPO" spelled out as "KubernetesPodOperator".

---

providers/cncf/kubernetes/docs/changelog.rst
potiuk: Applied — same expansion here.

---

providers/common/ai/docs/changelog.rst
potiuk: Applied. Rendered the inline backticks as single quotes to match the changelog convention (the whole entry is already wrapped in double backticks):

```
* ``Mark asserts under 'TYPE_CHECKING' in 'DocumentLoaderOperator' (#70378)``
```

---

providers/fab/docs/changelog.rst
potiuk: Applied — ``Bump eslint to 10.8.0 (#70697)``.

---

providers/fab/docs/changelog.rst
potiuk: Applied — ``Bump prettier to 3.9.6, stylelint to 17.14.1, webpack to 5.109.0 (#70604)``, verified against the PR's `package.json` diff.

---

potiuk:
Pushed: rebased onto latest main and applied the review feedback.

- Changelog wording fixes from the review (abbreviations spelled out, explicit dependabot package/version titles, user-facing phrasing).
- `clickhousedb` 1.0.1 kept — see the inline threads for why.
- A version bump supersedes a provider's `.latest-doc-only-change.txt` marker, which had been hiding earlier doc-only/tooling commits from the range. Those now appear in `cohere`, `dbt.cloud`, `presto`, `trino` and `papermill`.
- Newly merged since the first cut: #70920 (elasticsearch, doc-only) and #70938 (openai, excluded).
- The OpenAI note was above the first version header, rendering page-level instead of belonging to 1.8.2; moved into its section. Same defect as the google one already fixed here.

The two generalizable points — expanding internal abbreviations, and requiring explicit dependabot titles — will go into the release skill as a separate PR, following #70258.

---

providers/clickhousedb/docs/changelog.rst
eladkal: @potiuk that is not a user facing change that is doc only (and not a user facing doc)

eladkal:
> clickhousedb 1.0.1 kept — see the inline threads for why.

I don't think it's right. Chaning registry entry is Airflow internal change. It means we just need to update doc version. There is nothing to release in PyPi.

potiuk:
> > clickhousedb 1.0.1 kept — see the inline threads for why.
> 
> I don't think it's right. Chaning registry entry is Airflow internal change. It means we just need to update doc version. There is nothing to release in PyPi.

Note taken. As a release manager, I will take consequences of that decision on me.

This is really a 'pragmatic" rather than "purisic" decision. I expect there might be problems if we mix different versions and registry metadata and try to have "past version" having different metadata than the same version now.

I have no idea if it will, but I think this is a risk I woul like to avoid. I am actually planning to yank 1.0.0 release once we release 1.0.1 precisely because it had wrong metadata - because both registry and releases is something we are responsible for - and if this bears no problems to our users, I prefer to "fix" both code and metadata in unambiguous way.

eladkal:
> I have no idea if it will, but I think this is a risk I woul like to avoid. I am actually planning to yank 1.0.0 release once we release 1.0.1 precisely because it had wrong metadata - because both registry and releases is something we are responsible for - and if this bears no problems to our users, I prefer to "fix" both code and metadata in unambiguous way.

That is your call but I think you are taking it too far.
Airflow registry is not the package metadata. By yanking you are sending a message that the package is broken why it is not. As I user I would lose heart beat waking up in the morning to see a version I use in production is now yanked. Please reconsider.
You are trying to solve a problem but may create another. Also given shared governance model I would at least ask to let Clickhouse team a heads up before you make that action.

potiuk:
> You are trying to solve a problem but may create another. Also given shared governance model I would at least ask to let Clickhouse team a heads up before you make that action.

That for sure when it comes to yanking. That was just and idea so far, I have not yet decided to do it.

But what actual problem do you see by releasing 1.0.1 now ? It will **actually** solve a problem that status of the provider reported in our official page will be good.

But yes, I hear you. It means new binary released. Because I personally do not want to lose time on trying to correct registry to fix it in a different way (what I cal "pragmatic" way)

BTW, I would happily reconsider **not releasing** 1.0.1 actually if someone will fix the registry to show the right status (incubating) for 1.0.0. We still have some time - but I am flying in a few hours so won't be able to lead it. If you make it happen - I will just drop 1.0.1 from the release.

PR #70935, opened 2026-08-02
Return only the latest Dag version from iter_all_latest_version_dags

Eason09053360:
## The method promised latest-version dags in its name and docstring but read every serialized_dag row, so FAB permission syncing processed each Dag once per version and could apply an outdated access_control last.

 `DBDagBag.iter_all_latest_version_dags` promises latest-version Dags in both its
name and its docstring, but it read every `serialized_dag` row without filtering
by version.

Its only production caller is FAB's `create_dag_specific_permissions`, so DAG-level
permission syncing iterated each Dag once per version. Besides the wasted work, the
query has no `ORDER BY`, so when versions carry different `access_control` the one
applied last is undefined and an outdated policy can win.

`SerializedDagModel._latest_by_version_select()` already supported "all dags" via
`dag_ids=None`; this exposes that through the public `get_latest_serialized_dags()`
wrapper, which keeps the existing `load_op_links` hook in the loop intact. All five
existing callers pass `dag_ids=` explicitly, so the signature change is backwards
compatible.

The existing test mocked `session.scalars`, so it could not catch a wrong query.
Added a regression test that writes two real versions to the database.

---

---

### DIFF
--- a/airflow-core/tests/unit/models/test_dagbag.py
+++ b/airflow-core/tests/unit/models/test_dagbag.py
@@ -244,29 +244,36 @@ class TestDBDagBag:
         db.clear_db_dag_bundles()
 
     def test_iter_all_latest_version_dags_yields_only_the_latest_version(self, dag_maker):
-        dag_id = "versioned_dag"
         db.clear_db_runs()
         db.clear_db_dags()
         db.clear_db_serialized_dags()
         db.clear_db_dag_bundles()
 
-        with dag_maker(dag_id, schedule=None):
+        with dag_maker("versioned_dag", schedule=None):
             EmptyOperator(task_id="a")
         # A version with task instances is kept rather than updated in place, so the next
         # write adds a second version instead of overwriting the first.
         dag_maker.create_dagrun()
 
-        with DAG(dag_id, schedule=None) as dag:
+        with DAG("versioned_dag", schedule=None) as dag:
             EmptyOperator(task_id="a")
             EmptyOperator(task_id="b")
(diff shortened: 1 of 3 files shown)


airflow-core/tests/unit/models/test_dagbag.py
```diff
@@ -242,6 +243,36 @@ def make_lazy(task_ids):
...
+            assert DagVersion.get_latest_version(dag_id, session=session).version_number == 2
+            dags = list(DBDagBag().iter_all_latest_version_dags(session=session))
+
+        assert [dag.dag_id for dag in dags] == [dag_id]
+        assert set(dags[0].task_ids) == {"a", "b"}
+
+        db.clear_db_runs()
+        db.clear_db_dags()
+        db.clear_db_serialized_dags()
+        db.clear_db_dag_bundles()
```
SameerMesiah97, line 281: Should we also add a regression covering multiple DAG IDs with multiple versions, to explicitly verify we return the latest version per DAG rather than just the latest serialized row overall?

SameerMesiah97 (commented):
Just one comment on the tests.

airflow-core/tests/unit/models/test_dagbag.py
Eason09053360: Good catch. With a single Dag the test couldn't distinguish a per-Dag max from a
global one, so `group_by(dag_id)` was never exercised .
I confirmed the old test still passed with that `group_by` removed.

PR #70936, opened 2026-08-02
Fix API client errors losing detail behind a reverse proxy

rjgoyln:
## Summary

Both API clients recognised an error body only when the response `content-type` was exactly `application/json`, so a proxy that rewrites the header to `application/json; charset=utf-8` skipped error parsing altogether: `airflowctl` printed a raw `httpx.HTTPStatusError` traceback in place of the server's message, and in the Task SDK the exception was no longer a `ServerResponseError`, so the status-specific handling in the supervisor and the task runner stopped matching. Airflow's own API server sends the bare media type, so this needs something sitting between client and server.

## Change

Two adjacent airflowctl crashes go with it: a non-object JSON error body made `log.warning(extra=...)` raise, and a body labelled JSON that does not decode raised `ValueError` instead of falling through to `raise_for_status()`.

---


airflow-ctl/src/airflowctl/api/operations.py
```diff
@@ -103,6 +103,20 @@ def _build_query_params(**values: Any) -> dict[str, Any]:
     return {name: _serialize_query_param(value) for name, value in values.items() if value is not None}
 
 
+def _has_json_body(response: httpx.Response) -> bool:
+    """
+    Check whether a response declares a JSON body.
```
SameerMesiah97: Keep this line in the dosctring. 

The remaining part of the docstring should be in a comment above the line where you are parsing the JSON.

SameerMesiah97 (approved):
Just one nit.

airflow-ctl/src/airflowctl/api/operations.py
rjgoyln: OK, I got it. Thank you!

task-sdk/src/airflow/sdk/api/client.py
```diff
@@ -1305,6 +1305,14 @@ def dags(self) -> DagsOperations:
         return DagsOperations(self)
 
 
+def _has_json_body(response: httpx.Response) -> bool:
+    """Check whether a response declares a JSON body."""
+    # Media types are case-insensitive and may carry parameters (RFC 9110), so a proxy that
```
potiuk, line 1334: Nit, and I'm deliberately *not* asking you to extract this into `shared/` — `airflow-ctl` doesn't consume any shared distribution today, so wiring one up for three lines would cost more than it saves.

But the twin in `airflow-ctl/src/airflowctl/api/operations.py:106` is the same function with a different comment: that one explains the consequence (the friendly `ServerResponseError` handling gets silently skipped), this one stops at the media-type point. The copies have diverged in the PR that created them, which is usually where drift starts.

A cross-reference in each — *"kept in sync with `airflowctl.api.operations._has_json_body`"* and vice versa — would be enough to make the next person notice there are two.

potiuk (approved):
Approving — solid fix, and the tests are better than most bug fixes get: parametrizing the media-type variants and pinning `application/jsonp` as a *negative* is exactly the trap a looser `startswith` check would have fallen into.

I went looking for two ways this could have been incomplete, and both came back clean:

- **Widening the media-type match in the Task SDK without adding the decode guard there.** Not a problem — `from_response` already wraps parsing in `try/except` with an `msgspec` fallback and returns `None` (`task-sdk/src/airflow/sdk/api/client.py:1372-1377`). The explicit guard went into the one place that genuinely lacked one.
- **Passing a server-supplied dict straight into `extra=`.** Safe here, but worth saying why I checked: `airflowctl/api/client.py:82` is `structlog.get_logger()` and nothing in the distribution calls `structlog.configure()`, so `extra` is just an event key. If that logger were routed through stdlib `logging`, a proxy error page like `{"message": "Bad Gateway"}` would still raise `KeyError: "Attempt to overwrite 'message' in LogRecord"` — the same class of bug, from the same source. Something to keep in mind if airflowctl's logging setup ever changes.

One nit inline, and two things before this can land.

### Please add a newsfragment

`airflow-core/newsfragments/70936.bugfix.rst` — the diff touches `task-sdk/`, and those changes ship in `airflow-core`. Both effects are user-visible: `airflowctl` printing a raw `httpx.HTTPStatusError` traceback instead of the server's message, and the supervisor / task runner's status-specific handling silently not matching behind a proxy. Defaulting to omitting one was the right call on your side — this is the maintainer asking for it.

### Please rebase before this merges

The branch is **996 commits behind `main`** and the last commit is from 2026-08-06, so the green CI on this PR was run against a tree that has moved a long way since. I'm approving on the strength of reading the diff, not on that run — please rebase (or use the update-branch button) so real CI executes against current `main`. I'll merge once it comes back green; the newsfragment can ride along in the same push.

---

>

---

task-sdk/src/airflow/sdk/api/client.py
rjgoyln: Thanks for the review.

Rebased onto current `main` and pushed, so CI is running against today’s tree.

Added the newsfragment at `airflow-core/newsfragments/70936.bugfix.rst`.

On the nit, I added `# Kept in sync with <the other one>` to each copy rather than merging the comments. The `operations.py` comment explains the consequence of silently skipping the `ServerResponseError` path, so I felt it was worth keeping there; the cross-reference points to the twin.

The `extra=` check was a good catch. I had verified that the `isinstance` guard covered the body shape, but hadn’t established that the key collision is currently harmless under the logger. Thanks for pointing that out.

PR #61376, opened 2026-02-03
Improve performance of task dequeuing

steveahnahn:
## Summary
Replace `list.pop(0)` with `deque.popleft()` in three hot‑path task traversal/queueing loops to avoid O(n^2) behavior.
`list.pop(0)` is O(n) due to element shifting; repeated in loops it becomes O(n^2). `deque.popleft()` is O(1) and preserves ordering.

## Changes
- `airflow-core/src/airflow/executors/base_executor.py`: use `deque` in `trigger_tasks()`
- `airflow-core/src/airflow/serialization/definitions/taskgroup.py`: use `deque` in `iter_tasks()`
- `task-sdk/src/airflow/sdk/definitions/taskgroup.py`: use `deque` in `iter_tasks()`

PR #70942, opened 2026-08-02
Dispatch the highest priority tasks first in the executor

ignacioparicio:
`BaseExecutor` sorts its queued tasks by `priority_weight` ascending, then dispatches from the front of that list. So when more tasks are queued than there are open slots, the **lowest priority tasks are the ones sent to workers**, and the highest priority ones are held back.

For example, with weights 1, 5 and 10 and one slot freeing up per loop, they run in the order 1, 5, 10; the most important task runs last.

It also does not even out over time, because the queue is re-sorted on every call. 

This only affects an executor that has more queued tasks than open slots, so deployments that stay under their `parallelism` ceiling, or that never set `priority_weight`, see no difference either way.

<details>
<summary> Where it happens </summary>

In `airflow-core/src/airflow/executors/base_executor.py`:

```python
# order_queued_tasks_by_priority(): lowest weight first
return sorted(self.queued_tasks.items(), key=lambda x: x[1].ti.priority_weight, reverse=False)

# _get_workloads_to_schedule(): reads from the front, stops at open_slots
for task_key, task_workload in self.order_queued_tasks_by_priority():
    if len(workloads_to_schedule) >= open_slots:
        break
    workloads_to_schedule.append((task_key, task_workload))
```

</details>

<details>
<summary> How it got here </summary>

The sort and the way it is consumed used to match. They drifted apart in two steps:

- #61376 (`fd4fd67c0f`) switched the sort from descending to ascending and, in the same commit, changed `sorted_queue.pop(0)` to `sorted_queue.pop()`. An ascending list popped from the back still gives the highest priority first, so this was correct, and O(1) instead of O(n).
- #61153 (`9fa13c82ff`) refactored `trigger_tasks` to support callback workloads and replaced that `pop()` with the front-to-back loop above. The sort was left ascending.

</details>

What this PR does:
  - `base_executor.py`: sorts descending again, and notes on the method that consumers read from the front.
  - `test_base_executor.py`: adds a test with three weights and fewer open slots than tasks. The existing tests all run with more slots than tasks, so none of them reach the truncating branch.
  - Adds a bugfix newsfragment.

---

- [X] Yes, Claude Code (Opus 5)

Used to trace the ordering back through the history of the two commits above, and to check that nothing else in the repo overrides the sort or depends on it being ascending.

### DIFF
--- a/airflow-core/newsfragments/70942.bugfix.rst
+++ b/airflow-core/newsfragments/70942.bugfix.rst
@@ -1,0 +1,1 @@
+Executors now dispatch the highest ``priority_weight`` tasks first. When more tasks were queued than there were open slots, the lowest priority ones were sent to workers and the highest priority ones were held back.
--- a/airflow-core/src/airflow/executors/base_executor.py
+++ b/airflow-core/src/airflow/executors/base_executor.py
@@ -428,7 +428,10 @@ class BaseExecutor(LoggingMixin):
 
     def order_queued_tasks_by_priority(self) -> list[tuple[TaskInstanceKey, workloads.ExecuteTask]]:
         """
-        Orders the queued tasks by priority.
+        Orders the queued tasks by priority, highest ``priority_weight`` first.
+
+        Consumers take workloads from the front of this list, so the highest priority
+        tasks must come first for them to be scheduled before the rest.
 
         :return: List of workloads from the queued_tasks according to the priority.
         """
@@ -439,7 +442,7 @@ class BaseExecutor(LoggingMixin):
         return sorted(
             self.queued_tasks.items(),
             key=lambda x: x[1].ti.priority_weight,
-            reverse=False,
+            reverse=True,
         )
 
     def trigger_tasks(self, open_slots: int) -> None:
--- a/airflow-core/tests/unit/executors/test_base_executor.py
+++ b/airflow-core/tests/unit/executors/test_base_executor.py
@@ -360,6 +360,29 @@ def test_trigger_running_tasks(dag_maker):
     executor._process_workloads.assert_called_once()
 
 
+@pytest.mark.db_test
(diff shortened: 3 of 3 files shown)


SameerMesiah97 (commented):
This makes sense. The documentation has long stated that larger priority_weight values have higher priority, so this appears to correct an implementation bug. Below is the relevant section from the documentation for others who will review this PR:

```
Priority Weights
================

``priority_weight`` defines priorities in the executor queue. The default ``priority_weight`` is ``1``, and can be
bumped to any integer; larger numbers mean higher priority.
Moreover, each task has a true ``priority_weight`` that is calculated based on its
``weight_rule`` which defines the weighting method used for the effective total priority weight of the task.

Below are the weighting methods. By default, Airflow's weighting method is ``downstream``.
```

CI still needs to be triggered though.

o-nikolas (approved):
Looks like a legitimate regression, nice spot. I'm not sure this needs a news fragment but it is  a user facing bug so I don't mind if we keep it.

pierrejeambrun (commented):
Thanks for the PR.

airflow-core/newsfragments/70942.bugfix.rst
```diff
@@ -0,0 +1 @@
+Executors now dispatch the highest ``priority_weight`` tasks first. When more tasks were queued than there were open slots, the lowest priority ones were sent to workers and the highest priority ones were held back.
```
pierrejeambrun: I would remove this. (we usually don't use `bugfix` newsfragment, other solution is to convert this to a significant newsfragment but that doesn't seem worth it)

ignacioparicio: done

pierrejeambrun (approved):
LGTM, thanks.

PR #70943, opened 2026-08-02
Remove the dead CFullLoader branch from the YAML helpers

rjgoyln:
## Summary

All three copies of the YAML wrapper carry a branch in `__getattr__` meant to swap in libyaml's `FullLoader`. It discards the value it looks up, so it has never taken effect.

## Change

Deleting the branch rather than adding the missing `return`. `CFullLoader` is not a `FullLoader` subclass, and `yaml.add_constructor` registers only on the pure-Python loaders, so activating it would turn tags registered through the same wrapper into `ConstructorError` — a surface Dag authors reach as `macros.yaml`. `safe_load` and `dump` keep the C implementations, where Airflow controls both ends.

No runtime change: `FullLoader` resolved to pyyaml's loader before, and still does.

---

### DIFF
--- a/airflow-core/src/airflow/utils/yaml.py
+++ b/airflow-core/src/airflow/utils/yaml.py
@@ -66,8 +66,6 @@ def __getattr__(name):
     # Delegate anything else to the yaml module
     import yaml
 
-    if name == "FullLoader":
-        # Try to use CFullLoader by default
-        getattr(yaml, "CFullLoader", yaml.FullLoader)
-
+    # FullLoader stays pyyaml's on purpose: CFullLoader does not see add_constructor
+    # registrations, so custom tags would raise ConstructorError instead of resolving.
     return getattr(yaml, name)
--- a/airflow-ctl/src/airflowctl/ctl/utils/yaml.py
+++ b/airflow-ctl/src/airflowctl/ctl/utils/yaml.py
@@ -67,8 +67,6 @@ def __getattr__(name):
     # Delegate anything else to the yaml module
     import yaml
 
-    if name == "FullLoader":
-        # Try to use CFullLoader by default
-        getattr(yaml, "CFullLoader", yaml.FullLoader)
-
+    # FullLoader stays pyyaml's on purpose: CFullLoader does not see add_constructor
+    # registrations, so custom tags would raise ConstructorError instead of resolving.
     return getattr(yaml, name)
--- a/task-sdk/src/airflow/sdk/yaml.py
+++ b/task-sdk/src/airflow/sdk/yaml.py
@@ -66,8 +66,6 @@ def __getattr__(name):
     # Delegate anything else to the yaml module
     import yaml
 
-    if name == "FullLoader":
-        # Try to use CFullLoader by default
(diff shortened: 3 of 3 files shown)


airflow-core/src/airflow/utils/yaml.py
```diff
@@ -68,6 +68,6 @@ def __getattr__(name):
 
     if name == "FullLoader":
         # Try to use CFullLoader by default
-        getattr(yaml, "CFullLoader", yaml.FullLoader)
+        return getattr(yaml, "CFullLoader", yaml.FullLoader)
```
kaxil: `CFullLoader` isn't a subclass of `FullLoader` (MRO is `CParser, FullConstructor, Resolver`), and `yaml.add_constructor(tag, fn)` only registers on `Loader`/`FullLoader`/`UnsafeLoader`. So registering a tag through this same wrapper and then loading with it goes from working to raising, not to silently ignoring:

```python
from airflow.utils import yaml

yaml.add_constructor("!tag", lambda l, n: {"v": l.construct_scalar(n)})
yaml.load("a: !tag hi", Loader=yaml.FullLoader)
# before: {'a': {'v': 'hi'}}
# after:  ConstructorError: could not determine a constructor for the tag '!tag'
```

No in-tree caller reads `FullLoader` through these wrappers, but `airflow.sdk.yaml` is re-exported into template macros, so `{{ macros.yaml.FullLoader }}` is reachable from DAG code. Deleting the branch (leaving `FullLoader` as pyyaml's) is the zero-risk option, and the module docstring does promise a drop-in `yaml`. Which way do we want it? Worth updating the PR body either way, "does not see constructors" reads as a silent no-op.

airflow-core/tests/unit/utils/test_yaml.py
```diff
@@ -0,0 +1,28 @@
...
+import yaml as pyyaml
+
+from airflow.utils import yaml
+
+
+def test_full_loader_prefers_the_c_implementation(monkeypatch):
+    c_full_loader = type("CFullLoader", (), {})
+    monkeypatch.setattr(pyyaml, "CFullLoader", c_full_loader, raising=False)
+
+    assert yaml.FullLoader is c_full_loader
```
kaxil: All three copies of this test only pin the libyaml-present branch. Every PyYAML wheel on PyPI bundles libyaml, so CI never exercises the fallback, which means `return yaml.CFullLoader` would pass here and `AttributeError` on a source build. A `monkeypatch.delattr(pyyaml, "CFullLoader", raising=False)` case asserting `yaml.FullLoader is pyyaml.FullLoader` closes that cheaply.

kaxil (approved):

airflow-core/src/airflow/utils/yaml.py
rjgoyln: Thank you! I learned a lot from this. I hadn't thought about the compatibility issues before. 
I've decided to remove the CFullLoader part and keep the pure Python version. Does this sound good to you?

uros-b (approved):

uros-b:
+1, LGTM

henry3260 (approved):

PR #70944, opened 2026-08-02
Stop variables export/import from silently corrupting values

Eason09053360:
`airflow variables export` followed by `airflow variables import` can silently rewrite a variable's value and invent a description for it.

Export writes a bare value when a variable has no description, while import treats any dict carrying a `"value"` key as a `{"value": ..., "description": ...}` envelope — nothing distinguishes the envelope from a value that happens to look like one. Export also decodes the stored JSON, losing the difference between raw text and a JSON-encoded string.

Round-tripping through a file on `main` :

| DB value before | exported as | after re-import |
|---|---|---|
| `{"value": "a", "description": "b"}` | `{"value": "a", "description": "b"}` | value → `a`, description → `b` |
| `{"value": 1, "other": 2}` | `{"value": 1, "other": 2}` | value → `1`, `other` dropped |
| `"hello"` | `hello` | `hello` — no longer valid JSON, so `deserialize_json=True` starts raising |

Import's reconstruction is deterministic — strings are stored verbatim, everything else is JSON-encoded, and one envelope layer is unwrapped — so export alone can be made lossless, leaving hand-written import files behaving exactly as before. Export now emits the stored form when the decoded value is a string, and wraps values that are themselves envelope-shaped so import's unwrap consumes our envelope rather than the user's data.

Exported files are unchanged for values that were never ambiguous. One visible change worth flagging: a variable stored as a JSON-encoded string (`Variable.set(k, "text", serialize_json=True)`) now exports as `"\"text\""` rather than `"text"`. The old output could not be re-imported without corruption, so there is no lossless way to keep it.

Left for follow-up PRs: import still truncates a hand-written `{"value": 1, "other": 2}` to `1`, and the UI import path does not unwrap the envelope at all.

---

### DIFF
--- a/airflow-core/src/airflow/cli/commands/variable_command.py
+++ b/airflow-core/src/airflow/cli/commands/variable_command.py
@@ -186,11 +186,16 @@ def variables_export(args):
 
         data = json.JSONDecoder()
         for var in qry:
+            # Emit a form variables_import turns back into var.val: it stores strings verbatim,
+            # JSON-encodes everything else, and unwraps any dict carrying a "value" key.
             try:
                 val = data.decode(var.val)
             except Exception:
                 val = var.val
-            if var.description:
+            else:
+                if isinstance(val, str):
+                    val = var.val
+            if var.description or (isinstance(val, dict) and "value" in val):
                 var_dict[var.key] = {
                     "value": val,
                     "description": var.description,
--- a/airflow-core/src/airflow/cli/commands/variable_command.py
+++ b/airflow-core/src/airflow/cli/commands/variable_command.py
@@ -186,15 +186,13 @@ def variables_export(args):
 
         data = json.JSONDecoder()
         for var in qry:
-            # Emit a form variables_import turns back into var.val: it stores strings verbatim,
-            # JSON-encodes everything else, and unwraps any dict carrying a "value" key.
+            # Mirror variables_import's reconstruction so export/import round-trips.
             try:
                 val = data.decode(var.val)
             except Exception:
                 val = var.val
-            else:
-                if isinstance(val, str):
-                    val = var.val
+            if isinstance(val, str):
+                val = var.val
             if var.description or (isinstance(val, dict) and "value" in val):
                 var_dict[var.key] = {
                     "value": val,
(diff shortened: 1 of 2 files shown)


airflow-core/newsfragments/70944.bugfix.rst
henry3260, line 1: I think we don't need this, please remove it

henry3260 (commented):
The export change looks like it regresses the UI import path for the same two cases this PR fixes for the CLI, since the CLI and UI use different unwrap rules — as far as I can tell there's no export-only change that's correct for both. The PR notes the UI limitation as follow-up, but given export now depends on the unwrap rules, I'd lean towards fixing the UI side in this PR rather than deferring it.

I'm not familiar with the UI part though, so I'd like to hear your thoughts. cc @bbovenzi

airflow-core/src/airflow/cli/commands/variable_command.py
```diff
@@ -186,11 +186,16 @@ def variables_export(args):
...
+            # Emit a form variables_import turns back into var.val: it stores strings verbatim,
+            # JSON-encodes everything else, and unwraps any dict carrying a "value" key.
             try:
                 val = data.decode(var.val)
             except Exception:
                 val = var.val
-            if var.description:
+            else:
+                if isinstance(val, str):
+                    val = var.val
```
henry3260, line 197:
```suggestion
            if isinstance(val, str):
                val = var.val
```

airflow-core/src/airflow/cli/commands/variable_command.py
```diff
@@ -186,11 +186,16 @@ def variables_export(args):
 
         data = json.JSONDecoder()
         for var in qry:
+            # Emit a form variables_import turns back into var.val: it stores strings verbatim,
+            # JSON-encodes everything else, and unwraps any dict carrying a "value" key.
```
henry3260, line 190:
```suggestion
            # Mirror variables_import's reconstruction so export/import round-trips.
```

airflow-core/src/airflow/cli/commands/variable_command.py
Eason09053360: Thanks for reviewing!
aplied

airflow-core/src/airflow/cli/commands/variable_command.py
Eason09053360: aplied

potiuk (approved):

PR #70945, opened 2026-08-02
Remove dead type-mismatched guard in airflow-ctl command generation

rjgoyln:
## Summary

The condition guarding this initialisation compares a Pydantic model class against a dict keyed by model name, so it has never been true. The branch is dead code, and has been since it was introduced.

It is deleted rather than repaired because repairing it would be a regression. The field loop beneath appends unconditionally, so the always-taken reset is what keeps datamodels that several operations share — `ConnectionBody` across three, `BackfillPostBody` and `VariableBody` across two — from accumulating a duplicate copy of their fields on every visit: a type-correct comparison expands `ConnectionBody` to 27 entries instead of 9.

Nothing changes at runtime, and the duplicates a repair would introduce are consumed idempotently, so this is a readability fix — it removes a condition that reads as live logic and invites the wrong correction.

---


henry3260 (approved):
lgtm

PR #70948, opened 2026-08-02
Make provider release changelog entries read for users, not for us

potiuk:
Follow-up to the review of #70932, in the spirit of #70258.

### Changelog entries should read for users, not for us

The review of the 2026-08-01 wave found entries that only make sense inside the
project. The skill said "Subjects must be the original commit subject … Don't
paraphrase", which is right about not inventing behaviour but left no room to fix
wording. This adds an explicit carve-out — rewrite the *wording*, never the
*claim* — covering the four cases the review actually hit:

| before | after |
| --- | --- |
| `… to prevent KPO race condition (#70595)` | `… to prevent KubernetesPodOperator race condition (#70595)` |
| `Bump the fab-ui-package-updates group across 1 directory with 3 updates (#70604)` | `Bump prettier to 3.9.6, stylelint to 17.14.1, webpack to 5.109.0 (#70604)` |
| `Remove noqa:S101 from production code (#70378)` | `Mark asserts under 'TYPE_CHECKING' in 'DocumentLoaderOperator' (#70378)` |
| `Remove hard-coded deferrable crawler run from example_glue system test (#70206)` | `Remove hard-coded deferrable crawler run from example_glue (#70206)` |

The Phase 5 defect scan gains matching checks for internal abbreviations and for
version-less / grouped dependency bumps.

### A version bump supersedes the doc-only marker

`classify-provider-changes` starts a provider's range at
`docs/.latest-doc-only-change.txt` when that marker is present, so older doc-only
and tooling commits stay hidden while nothing is pending. That suppression stops
applying the moment the provider gets a real version bump — but a single
discovery pass never sees the widened range, so those commits are dropped from
the release.

In the 2026-08-01 wave this silently lost `cohere` #69649,
`dbt.cloud`/`presto`/`trino` #69478 and `papermill` #68322. The skill now says to
re-run discovery after Phase 4a and names the tell: a provider whose commit count
grows between two runs with no new merges.

### Misplaced notes have a second variant

Phase 5 already checked for a `.. note::` above the first version header. It
missed the case where prepending a new section pushes such a note *below* the new
excluded block — no longer at the top of the file, so a top-of-file scan skips it.
Hit `openai` (#69506) in this wave; `google` (#70869) was the already-documented
variant. The check is now per-section.

---

### DIFF
--- a/.agents/skills/prepare-providers-documentation/SKILL.md
+++ b/.agents/skills/prepare-providers-documentation/SKILL.md
@@ -176,6 +176,25 @@ How to read it:
 > back as `needs_llm`. The rules live in `classify_change_deterministically`
 > (`dev/breeze/src/airflow_breeze/prepare_providers/provider_documentation.py`).
 
+The range the classifier reports is not fixed for the whole run — it widens once
+a provider actually gets a version bump:
+
+> [!WARNING]
+> **A version bump supersedes the provider's doc-only marker — re-run discovery
+> after Phase 4a.** When `docs/.latest-doc-only-change.txt` is present the
+> classifier starts the range at that commit, so anything older (earlier doc-only
+> changes, tooling churn) is hidden for as long as the provider has no pending
+> release. That is correct while nothing ships — but the moment the provider gets
+> a real version bump the marker no longer applies, and those commits are part of
+> what the release publishes.
+>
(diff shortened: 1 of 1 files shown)


potiuk:
cc: @eladkal -> your request :)

eladkal (approved):
Nice!

PR #69034, opened 2026-06-26
Warm-shutdown supervisor on SIGTERM instead of killing the running task

arkadiuszbach:
## Summary

PR #61627 installed a SIGTERM/SIGINT handler in `ActivitySubprocess.wait()` that forwarded the signal to the task subprocess via `os.kill(self.pid, ...)`, causing the operator's `on_kill()` to run whenever the supervisor was signalled.

On KubernetesExecutor this kills the task on every routine pod shutdown. `dumb-init` runs as PID 1 with `DUMB_INIT_SETSID=1` (the image default), so it broadcasts SIGTERM to the whole process group — the task subprocess already receives it directly. After #61627 the supervisor forwarded it a second time, so the task could receive SIGTERM twice; either delivery makes the child handler call `on_kill()` and abort the task the moment the pod starts terminating.

That is wrong for a graceful pod shutdown (spot interruption, scale-down, rolling update). For operators that release external resources in `on_kill()` (Databricks tearing down a job cluster, ADF cancelling a pipeline, Dataproc, Bash terminating its child, etc.) this destroys in-flight work we want to let finish.

This change makes the task warm-shut-down instead, in two parts:

1. **Supervisor** — replace the forward-and-kill handler with a *warm-shutdown* handler that does **not** signal the task. It logs the received signal and lets the supervisor stay in its monitoring loop so the running task continues to completion. The handler is installed in `_PythonCoordinator.execute_task` around **both** `start()` (the RUNNING transition) and `wait()` (run + terminal-state report + log upload), so no window is left where the default SIGTERM disposition could tear the supervisor down. Previous dispositions are restored on exit so a reused Celery prefork worker doesn't leak the handler.

2. **KubernetesExecutor pod template** — set `DUMB_INIT_SETSID=0`, so dumb-init delivers pod-termination signals only to its direct child (the supervisor), not to the whole process group. Without this the task subprocess would still receive the group-broadcast SIGTERM directly and call `on_kill()`, defeating the warm shutdown. This mirrors what Celery worker pods already do.

`on_kill()` is still reached through the legitimate kill paths (heartbeat-failure / server-terminated kills, overtime termination, or an explicit success/failure transition) — all of which already drive `self.kill(...)` and are unchanged.

## Why

[...]

PR #65738, opened 2026-04-23
Fix orphaned subprocesses and supervisor crash on heartbeat 409

cmettler:
When a running TaskInstance is forcibly transitioned out of `running` (scheduler reset, REST PATCH, etc.), the next heartbeat from the still-running task-runner returns HTTP 409 and the supervisor kills the task. On Linux this produced two bugs:

1. **Orphaned subprocesses** — any Popen child the task-runner had spawned (`@task.virtualenv`, `DockerOperator`, `BashOperator`, Cosmos dbt, etc.) was reparented to PID 1 and kept running until it finished on its own, wasting CPU/RAM/API quota.
2. **Supervisor crash** — ~60s later `_cleanup_open_sockets()` closed the selector while `_service_subprocess()` was still polling it, raising `ValueError: I/O operation on closed epoll object` (regression from #51180).

### Fix

Place the task-runner in **its own process group** with `os.setpgid(0, 0)` immediately after fork (task execution opts in; the DAG processor and triggerer keep the supervisor's group), then have `kill()` signal the whole group via `os.killpg(os.getpgid(pid), sig)`. This reaches every subprocess the task-runner spawned. Grandchildren without a SIGTERM handler exit promptly and close their inherited pipes, so the supervisor drains `_open_sockets` normally and never enters the cleanup-the-selector-mid-loop path.

Two safeguards make the group-signalling robust:

- **Parent-side `setpgid(pid, pid)` mirror.** The parent repeats the child's `setpgid` right after fork so the group is guaranteed to exist regardless of ordering. Without it, a signal arriving before the child's own `setpgid` ran (e.g. `_on_child_started` failing synchronously) could resolve the child's PGID to the *supervisor's* own group and `killpg` it.
- **Own-group guard in `_signal_subprocess()`.** If both `setpgid` calls failed and the child still shares the supervisor's process group, the code signals the pid directly instead of `killpg`-ing its own group, so the supervisor never signals itself.

`killpg`/`getpgid` fall back to `self._process.send_signal(sig)` on `ProcessLookupError` or `PermissionError`, preserving behaviour when the group has vanished or permissions are lacking.

### Tests

- `test_task_runner_starts_in_new_process_group` — real-fork regression: asserts the child's PGID == its own PID after `ActivitySubprocess.start()` for task execution.
- `test_child_keeps_supervisor_process_group_by_default` — DAG processor / triggerer path stays in the supervisor's group.
- `test_kill_signals_process_group` — primary path uses `killpg`.
[...]

PR #70950, opened 2026-08-02
Make Task SDK supervisor subprocess a session leaderFix supervisor session leader

### LINKED ISSUE #70949: Make Task SDK supervisor subprocess a session leader

Airflow Core

### Apache Airflow version

main

**Issue Description**
In `task-sdk/src/airflow/sdk/execution_time/supervisor.py`, the `_subprocess_main` function is the entry point for the child process that runs the user's task code. There is an existing `TODO` comment indicating that this child process should be made a session leader. 

Currently, because it is not a session leader, it remains in the same process group as the parent supervisor. This means that if a signal (like `SIGINT` or `SIGTERM`) is sent to the parent process group (for instance, by the terminal), both the parent and the child will receive it independently. This can lead to double-firing of signal handlers and race conditions during shutdown or interruption.

**Steps to reproduce**
N/A - This is a structural process-isolation improvement identified via an existing `TODO` in `task-sdk/src/airflow/sdk/execution_time/supervisor.py`. I have already implemented the fix and will be raising a PR shortly.

The child process should be made a session leader immediately upon startup to isolate it into its own process group. This ensures that signals are handled explicitly through the supervisor's lifecycle management rather than being automatically broadcasted to both processes. 

We can achieve this by calling `os.setsid()` (guarded by `hasattr(os, "setsid")` and a `try-except` block for cross-platform compatibility) right after the subprocess starts.

### Operating System

Any

### Deployment

None

_No response_

Not Applicable

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

Not Applicable

Not Applicable

### Anything else?

_No response_

### Code of Conduct

KushagraB424:
### Background & Problem
In `task-sdk/src/airflow/sdk/execution_time/supervisor.py`, the `_subprocess_main` function is the entry point for the child process that runs the user's task code. An existing `TODO` comment indicated that this child process should be made a session leader. 

If the process is not a session leader, it remains in the same process group as the parent supervisor. This means that if a signal (like `SIGINT` or `SIGTERM`) is sent to the parent process group, both the parent and the child would receive it independently, potentially leading to double-firing of signal handlers and race conditions during shutdown or interruption.

### The Solution
This PR resolves the `TODO` by making the child process a session leader immediately upon startup. 

We introduced a call to `os.setsid()` inside a `try...except OSError` block, guarded by `hasattr(os, "setsid")` to ensure cross-platform compatibility (especially for Windows environments where `setsid` might not be available). This isolates the child process into its own session and process group, ensuring that signals are handled explicitly through the supervisor's lifecycle management rather than broadcasted by the OS terminal.

### Impacted Files
[MODIFY] task-sdk/src/airflow/sdk/execution_time/supervisor.py

### Testing & Verification
- The changes gracefully fall back if `os.setsid()` is unavailable or fails.
- CI/CD pipelines (via breeze and pre-commit) will verify that syntax is correct and that process isolation continues to function properly across supported operating systems.

closes: #70949

---

Generated-by: Antigravity IDE following the [guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/task-sdk/src/airflow/sdk/execution_time/supervisor.py
+++ b/task-sdk/src/airflow/sdk/execution_time/supervisor.py
@@ -418,7 +418,12 @@ def _fork_main(
     - Catch un-handled exceptions and attempt to show _something_ in case of error
     - Finally, run the actual task runner code (``target`` argument, defaults to ``.task_runner:main`)
     """
-    # TODO: Make this process a session leader
+    # Make this process a session leader
+    if hasattr(os, "setsid"):
+        try:
+            os.setsid()
+        except OSError:
+            pass
 
     # Store original stderr for last-chance exception handling
     last_chance_stderr = _get_last_chance_stderr()


kaxil:
Thanks for the PR. A bare `setsid()` here would regress graceful shutdown: the official image runs dumb-init with `DUMB_INIT_SETSID=1`, which broadcasts SIGTERM to the whole process group on container stop, and that group co-membership is how the task subprocess receives the signal (see "Signal propagation" in the docker-stack entrypoint docs). With the child in its own session, the task never gets SIGTERM, so its `on_kill()` handler never runs before the grace period expires and everything is SIGKILLed. #69034 recently made "signal only the supervisor" an explicit opt-in via `DUMB_INIT_SETSID=0`; this change would force it everywhere.

On the double-firing concern from #70949: `_reset_signals()` a few lines down already resets the inherited handlers to SIG_DFL in the child, so handlers can't fire twice.

There's also an open PR covering this ground: #65738 moved from `os.setsid()` to `os.setpgid(0, 0)` during review, scoped it to task execution only, paired it with `os.killpg` in `kill()`, and removes this TODO. Closing in favor of that one.

PR #70951, opened 2026-08-02
Release TI lock before asset listener callbacks

### LINKED ISSUE #66853: API server OOMKill: task_instance row lock held during asset event emission under high concurrency

uranusjr:
Asset registration on the task-success path (ti_update_state) ran the listener hooks synchronously inside the transaction holding a row lock on the task_instance table. A slow listener, multiplied across a large fan-out of asset events, could hold that lock for minutes, causing statement timeouts.

The listener hooks are now deferred until the end of the endpoint instead of executed inline during asset event creation. Registration writes to the database still happen under the caller's transaction, so durability is unchanged; this only moves the best-effort listener hooks off the lock.

Close #66853.

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -21,7 +21,7 @@ import contextlib
 import itertools
 import json
 from collections import defaultdict
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator, Sequence
 from typing import TYPE_CHECKING, Annotated, Any, NoReturn, cast
 from uuid import UUID
 
@@ -450,8 +450,9 @@ def ti_update_state(
         data["_rendered_map_index"] = data.pop("rendered_map_index")
     query = update(TI).where(TI.id == task_instance_id).values(data)
 
+    asset_callbacks: Sequence[Callable[[], None]] = ()
     try:
-        query, updated_state = _create_ti_state_update_query_and_update_state(
+        query, updated_state, asset_callbacks = _create_ti_state_update_query_and_update_state(
             ti_patch_payload=ti_patch_payload,
             task_instance_id=task_instance_id,
             session=session,
@@ -528,6 +529,12 @@ def ti_update_state(
                     run_id=run_id,
                     task_id=task_id,
                 )
+
+    # Release the task_instance row lock before running listener callbacks.
+    session.commit()
+
+    for callback in asset_callbacks:
+        callback()
 
 
 def _emit_task_span(ti, state):
@@ -626,7 +633,8 @@ def _create_ti_state_update_query_and_update_state(
     session: SessionDep,
     dag_bag: DagBagDep,
     dag_id: str,
-) -> tuple[Update, TaskInstanceState]:
+) -> tuple[Update, TaskInstanceState, Sequence[Callable[[], None]]]:
+    asset_callbacks: Sequence[Callable[[], None]] = ()
     if isinstance(ti_patch_payload, (TITerminalStatePayload, TIRetryStatePayload, TISuccessStatePayload)):
         ti = session.get(TI, task_instance_id, with_for_update={"of": TI})
         updated_state = TaskInstanceState(ti_patch_payload.state.value)
@@ -657,7 +665,7 @@ def _create_ti_state_update_query_and_update_state(
             query = query.values(retry_delay_override=retry_delay_override, retry_reason=retry_reason)
         elif isinstance(ti_patch_payload, TISuccessStatePayload):
             if ti is not None:
-                TI.register_asset_changes_in_db(
+                asset_callbacks = TI.register_asset_changes_in_db(
                     ti,
                     ti_patch_payload.task_outlets,
                     ti_patch_payload.outlet_events,
@@ -768,7 +776,7 @@ def _create_ti_state_update_query_and_update_state(
                 ti = session.get(TI, task_instance_id, with_for_update={"of": TI})
                 if ti is not None:
                     _handle_fail_fast_for_dag(ti=ti, dag_id=dag_id, session=session, dag_bag=dag_bag)
-                return query, TaskInstanceState.FAILED
+                return query, TaskInstanceState.FAILED, ()
 
         actual_start_date = timezone.utcnow()
         session.add(
@@ -790,7 +798,7 @@ def _create_ti_state_update_query_and_update_state(
     else:
         raise ValueError(f"Unexpected Payload Type {type(ti_patch_payload)}")
 
-    return query, updated_state
+    return query, updated_state, asset_callbacks
 
 
 @ti_id_router.patch(
--- a/airflow-core/src/airflow/assets/manager.py
+++ b/airflow-core/src/airflow/assets/manager.py
@@ -385,13 +385,19 @@ class AssetManager(LoggingMixin):
 
         dags_to_queue_from_asset_alias = set()
         if source_alias_names:
-            asset_alias_models: Iterable[AssetAliasModel] = session.scalars(
-                select(AssetAliasModel)
-                .where(AssetAliasModel.name.in_(source_alias_names))
-                .options(
-                    joinedload(AssetAliasModel.scheduled_dags).joinedload(DagScheduleAssetAliasReference.dag)
+            asset_alias_models = (
+                session.scalars(
+                    select(AssetAliasModel)
+                    .where(AssetAliasModel.name.in_(source_alias_names))
+                    .options(
+                        joinedload(AssetAliasModel.scheduled_dags).joinedload(
+                            DagScheduleAssetAliasReference.dag
+                        )
+                    )
                 )
-            ).unique()
+                .unique()
+                .all()
+            )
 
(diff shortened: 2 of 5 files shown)


airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -529,6 +530,15 @@ def ti_update_state(
...
                 )
 
+    # Release the task_instance row lock before running listener callbacks.
+    session.commit()
+
+    for callback in asset_callbacks:
+        try:
+            callback()
+        except Exception:
+            log.warning("Failed to run deferred asset listener callback", exc_info=True)
```
jason810496: nit: It seems we can safely remove the try/ except with logging here as the `nofity_asset_event_emitted` itself already handled the exception.

https://github.com/apache/airflow/blob/182919ac348badad1f235ba058fe52aa86339098/airflow-core/src/airflow/assets/manager.py#L418-L432

https://github.com/apache/airflow/blob/182919ac348badad1f235ba058fe52aa86339098/airflow-core/src/airflow/assets/manager.py#L508-L514

jason810496 (commented):
No comments from my end, LGTM with this direction, thanks.

jason810496 (commented):
Could we also address the loading issue that https://github.com/apache/airflow/pull/66854#issuecomment-5122174089 comment pointed out in this one? Since this PR is small enough. Thanks.

Lee-W (approved):

airflow-core/src/airflow/assets/manager.py
```diff
@@ -416,20 +424,23 @@ def register_asset_change(
...
-                partition_date=partition_date,
-            )
+        listener_asset_event = ListenerAssetEvent(
+            asset=asset,
+            extra=asset_event.extra,
+            source_dag_id=asset_event.source_dag_id,
+            source_task_id=asset_event.source_task_id,
+            source_run_id=asset_event.source_run_id,
+            source_map_index=asset_event.source_map_index,
+            source_aliases=[aam.to_serialized() for aam in asset_alias_models],
```
kaxil, line 440: Not introduced by this PR, but since this block is being rewritten: `asset_alias_models` is the one-shot `ScalarResult` from above, and the `for` loop has already exhausted it by the time this comprehension runs, so it always evaluates to `[]` and listeners never receive the source aliases. I verified that re-iterating a consumed `ScalarResult.unique()` silently yields nothing. Materializing it at the assignment (`list(session.scalars(...).unique())`) would fix it without touching the loop.

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
```diff
@@ -529,6 +530,12 @@ def ti_update_state(
                     task_id=task_id,
                 )
 
+    # Release the task_instance row lock before running listener callbacks.
+    session.commit()
+
+    for callback in asset_callbacks:
```
kaxil, line 536: Nothing in the suite covers the deferred path: no test asserts the listener hooks still fire when `ti_update_state` succeeds with outlets, and `test_manager.py` has no `callback_sink` case. If a later change drops this loop or ignores the return value of `register_asset_changes_in_db`, everything stays green while `on_asset_changed` and `on_asset_event_emitted` silently stop firing for task-produced events. A spy listener test on this route would lock the behavior in.

kaxil (approved):
2 minor comments lgtm otherwise

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
Andrushika, line 536: Nit: Previously, listeners ran after each event was created. Now their callbacks are collected and run only after all outlets have been processed.

If the second outlet fails with a non-DB error, an event already created for the first outlet can still be committed, but its callbacks are lost because the function never returns them. 

(This may be fine since listeners are best effort, just to confirm that this is intentional)

uranusjr: Registration callback failures cannot bubble here because the hook caller unconditionally wraps them in a try-except (see https://github.com/apache/airflow/pull/70951#discussion_r3701641837). So all callbacks will always be run unless you kill the server process entirely.

eladkal (approved):

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
uranusjr: I added a couple of tests to ensure the ehavior on different layers.

airflow-core/src/airflow/assets/manager.py
uranusjr: Fixed with a `.all()` call.

airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
Andrushika: Sorry, my wording was not clear... I meant a failure during registration itself, before the callbacks are handed back.

For example, if 9 out of 10 outlets are processed and the last one raises a non-DB exception, `asset_callbacks` here stays empty:

```python
asset_callbacks = ()

try:
    # raises inside, while processing the 10th outlet
    query, updated_state, asset_callbacks = _create_ti_state_update_query_and_update_state(...)
    # so this assignment never happens and asset_callbacks is still ()
except Exception:
    ...  # caught, TI marked FAILED, execution continues

session.commit()  # the 9 events are already in the transaction, so they are committed here

for callback in asset_callbacks:  # empty, the loop never runs
    callback()
```

So the 9 events are committed, but their listeners are never called. Probably still fine as best effort, just wanted to make sure this case was considered!

uranusjr: Right, and this is why the follow-up PR is needed. After #70972, if something fails midway during registration, the entire trasaction will be rolled back; no events will be registered with no callbacks run, which is consistent.

Issue #58514, opened 2025-11-20
Docs: Clarify that masking in Connection 'extra' JSON is keyword-dependent


varaprasadregani:
### What do you see as an issue?

**Docs link:** [Masking Sensitive Data](https://airflow.apache.org/docs/apache-airflow/stable/security/secrets/mask-sensitive-values.html#masking-sensitive-data)

**Source code link:** [secrets_masker.py](https://github.com/apache/airflow/blob/main/airflow/shared/secrets_masker/src/airflow_shared/secrets_masker/secrets_masker.py)

The documentation currently states:
> “Airflow will by default mask Connection passwords and keys from a Connection’s extra (JSON) field when they appear in Task logs, in the Variable and in the Rendered fields views of the UI.”

This statement is misleading because it implies that **all** keys in a Connection’s extra JSON field are masked. However, only keys whose names contain known sensitive keywords are actually redacted.

The complete list of sensitive keywords from the source code is:
`access_token`, `api_key`, `apikey`, `authorization`, `passphrase`, `passwd`, `password`, `private_key`, `secret`, `token`, `keyfile_dict`, `service_account`

**Code used to reproduce this:**
I verified this behavior using the following DAG, extracting values from a Connection's `extra` field (`bigquery_connection_id`) and Airflow Variables:

```python
from airflow import DAG
from airflow.operators.bash import BashOperator
import pendulum
from airflow.hooks.base import BaseHook
from airflow.models import Variable

# Fetch connection and extract 'extra' JSON
conn = BaseHook.get_connection(conn_id="bigquery_connection_id")
extra_data = conn.extra_dejson

# Test specific keys from 'extra'
keyfile_dict = extra_data.get("keyfile_dict", "not found") # Contains 'keyfile_dict'
param1_token = extra_data.get("param1_token", "not found") # Contains 'token'
hello = extra_data.get("hello", "not found")               # No sensitive keyword

# Test Variables
test_keyfile_dict = Variable.get("test_keyfile_dict")
service_account = Variable.get("service_account")

with DAG(
    dag_id="test_masking",
    start_date=pendulum.datetime(2024, 1, 1, tz="UTC"),
    schedule=None,
    catchup=False,
) as dag:

    test_masking = BashOperator(
        task_id="masking_task",
        bash_command=f"echo '{ keyfile_dict }' > { param1_token } > { hello } > { test_keyfile_dict } > { service_account }"
    )
```

**From my testing:**
[...]

PR #70956, opened 2026-08-03
Redact well-known secret formats in logs and rendered fields

dkranchii:
**Summary**

Airflow's SecretsMasker decides what to hide by key name today. Values that reach Task logs via XCom, a print of an environment variable, a stack trace, or a Connection extra whose key doesn't happen to match a sensitive keyword are not redacted — as [mask-sensitive-values.rst](https://github.com/apache/airflow/blob/main/airflow-core/docs/security/secrets/mask-sensitive-values.rst) documents and #58514 previously reported (closed as a docs-only clarification, so the underlying gap remained).

This PR adds an opt-in, default-off second pass in SecretsMasker that scans string values for a small, curated set of well-known credential formats — AWS access keys, GitHub / Slack / Google / Stripe tokens, PEM private-key blocks, JWTs — and redacts any match. Because it plugs into the existing masker, it runs everywhere the masker already runs (log filter, redact(), rendered fields).

related: #58514

## Design

**Opt-in behavior.** Enable via `[core] mask_secrets_content_patterns` (default `False`). No existing deployment changes behavior.

**Curated built-in patterns.** Seven formats, each with a distinctive fixed prefix so match confidence is very high:

| # | Format | Prefix |
|---|---|---|
| 1 | AWS access / session keys | `AKIA…` / `ASIA…` |
| 2 | GitHub tokens | `gh[pousr]_…` |
| 3 | Slack tokens | `xox[baprs]-…` |
| 4 | Google API keys | `AIza…` |
| 5 | Stripe live keys | `sk_live_…` |
| 6 | PEM private-key blocks | `-----BEGIN … PRIVATE KEY-----` |
| 7 | JSON Web Tokens | `eyJ…eyJ…` |

Formats with documented high false-positive rates in log data (credit cards, SSNs, email addresses) are deliberately left out of the defaults.

**Pluggable.** Deployments can register additional named regexes via `SecretsMasker.add_content_patterns({name: regex})`. Invalid regexes are logged and skipped rather than raised, so a misconfigured entry can't disable the whole masker.

**ReDoS-audited.** Every pattern uses fixed-width or bounded quantifiers — no nested quantifiers, no overlapping alternations, and no `\S*?` + negative-lookahead combo (the shape fixed in #70716). Worst case is a single linear scan of the input string.

**Extends existing surface.** `SecretsMasker` is already a `logging.Filter` shared by `airflow-core` and the task-SDK via `shared/secrets_masker/`. Adding a parallel `SensitivityClassifier` module would duplicate the recursive-walk path the masker already implements.

**Log-filter short-circuit widened.** The `filter()` guard was `if self.replacer:`; it's now `if self.replacer or self.mask_content_patterns:`. Without this, content-pattern masking would silently no-op in log paths when no explicit `add_mask()` secrets are registered.

**Compatibility**
No behavior change with the flag off.
Not a replacement for mask_secret() — this is defense in depth. Known secrets should still be registered explicitly.
No new runtime dependencies, no schema change, no new decorator.

**Testing**
25 new tests covering each built-in pattern, false-positive guardrails, add_content_patterns() registration and invalid-regex rejection, reset_masker() behavior, composition with existing key-name masking and with add_mask(), and the log-filter path when only content-pattern masking is enabled.
Full pre-existing suite: 167 passed, 1 skipped (pre-existing k8s import), 1 xfailed (pre-existing) — no regressions.
ruff format and ruff check clean.

**Related**
related: #58514 — same limitation, closed as docs-only clarification.
Independent context: #70716 (ReDoS fix in Spark provider) — reviewed my regexes against the same class of issue. #70890 (bulk audit-log masking) — different code path, no conflict.

**Reviewer notes**
Happy to trim the initial pattern set (e.g. drop JWTs, which have the highest false-positive rate of the seven) or flip the default in a follow-up PR once we have adoption data. Not adding a newsfragment yet; can add one on request.

Yes — Cursor Agent (Opus 4.7)

### DIFF
--- a/airflow-core/docs/security/secrets/mask-sensitive-values.rst
+++ b/airflow-core/docs/security/secrets/mask-sensitive-values.rst
@@ -113,6 +113,48 @@ or
 
 The mask must be set before any log/output is produced to have any effect.
 
+Content-based masking of well-known secret formats
+""""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. versionadded:: 3.4.0
+
+Registering secrets explicitly via ``mask_secret`` (or through Connections and Variables) only
+covers values Airflow was told about. Credentials that end up in Task logs or Rendered fields via
+other paths — a debug ``print`` of an environment variable, a stack trace containing a token, a
+value pulled from an XCom — are not covered by that mechanism.
+
+To catch those cases, Airflow can additionally scan every string that passes through the secrets
+masker for a small, curated set of well-known credential formats and redact any match. The set
+is intentionally narrow — each entry has a distinctive prefix so a match is overwhelmingly
+likely to be a real credential:
+
+* AWS access / session keys (``AKIA…``, ``ASIA…``)
+* GitHub tokens (``ghp_…``, ``gho_…``, ``ghu_…``, ``ghs_…``, ``ghr_…``)
+* Slack tokens (``xoxb-…``, ``xoxp-…``, ``xoxa-…``, ``xoxr-…``, ``xoxs-…``)
+* Google API keys (``AIza…``)
+* Stripe live keys (``sk_live_…``)
+* PEM-encoded private key blocks (``-----BEGIN … PRIVATE KEY-----``)
+* JSON Web Tokens with the standard ``eyJ…`` header + payload prefix
+
+This is a **defense-in-depth** measure that complements, but does not replace, explicit masking:
+formats with high false-positive rates (generic credit-card numbers, SSNs, email addresses) are
+deliberately excluded, and matches only fire on values that actually flow through the masker.
+Secrets you already know about should still be registered via ``mask_secret``.
+
+The feature is opt-in because the regex scan runs on every string that passes through the
+masker. Enable it in your Airflow config:
+
+.. code-block:: ini
+
+    [core]
+    mask_secrets_content_patterns = True
+
+or via the corresponding environment variable
+``AIRFLOW__CORE__MASK_SECRETS_CONTENT_PATTERNS=True``.
+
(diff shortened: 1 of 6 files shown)


potiuk:
Please change version added to 3.4.0

airflow-core/docs/security/secrets/mask-sensitive-values.rst
```diff
@@ -113,6 +113,48 @@ or
 
 The mask must be set before any log/output is produced to have any effect.
 
+Content-based masking of well-known secret formats
+""""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. versionadded:: 3.2.0
```
potiuk: Version added should be 3.4.0 here and below.

airflow-core/src/airflow/config_templates/config.yml
```diff
@@ -454,6 +454,18 @@ core:
...
       default: ""
+    mask_secrets_content_patterns:
+      description: |
+        If set to ``True``, Airflow scans string values in Task logs and Rendered fields for a small,
+        curated set of well-known credential formats (AWS access keys, GitHub tokens, Slack tokens,
+        Google API keys, Stripe live keys, PEM-encoded private key blocks, JWTs) and redacts any
+        match. This is a defense-in-depth measure that complements — but does not replace —
+        registering secrets explicitly via ``mask_secret`` or through Connections/Variables. It
+        is opt-in because the regex scan runs on every string that passes through the masker.
+      version_added: 3.2.0
```
potiuk: Version added 3.4.0

potiuk (approved):
Approve but needs to fix version added and rebase,.

dkranchii:
@potiuk update the PR based on your comments and rebased.

PR #70960, opened 2026-08-03
Add SRV DNS record support to HTTP Operator

### LINKED ISSUE #9186: Add SRV DNS record support to http connections
**Description**

SRV records are DNS records that resolve to a hostname and a port. Being able to specify http connections with SRV resolution support is useful in contexts where external services are defined as SRV records. 

**Use case / motivation**

One particular example applies to service discovery with dynamic port mapping. In this context, service discovery maps one or more hosts to a SRV record. When resolving that SRV record a client gets one or more hosts and ports to reach the desired service. 
In my particular use case I am using AWS Service Discovery for ECS with dynamic port mapping. I would like to reach my ECS service transparently from Airflow without going thru a load balancer.

I would like to be able to specify an http connection with a SRV record and let Airflow resolve the actual hostname and port using SRV lookup.

**Related Issues**

https://github.com/apache/airflow/pull/4481

I'll be happy to provide the implementation for this

gtxu:
This pull request allows an HTTP connection resolve its target host and port from a DNS SRV record instead of a fixed host/port, so it can talk to services published via SRV-based service discovery(ie. AWS Service Discovery for ECS).

This changes introduces `dnspython` as an optional dependency to keep the core HTTP provider lightweight. ~~To prevent DNS query flooding, a TTL cache mechanism is implemented.~~

 - Implemented soft dependency via the `srv` extra.

 - Built RFC 2782 compliant failover (lowest priority tier filtering and random target selection).

 - ~~Added concurrency-safe TTL caching (`srv_cache_ttl`) using `asyncio.Lock` to prevent cache stampedes in asynchronous contexts.~~

 - Exposed `srv_lookup` ~~and `srv_cache_ttl`~~ configuration directly in the Airflow Connection UI via custom form widgets.

closes: #9186

---

- [X] Yes - Claude Sonnet 5

---

### DIFF
--- a/providers/http/docs/connections/http.rst
+++ b/providers/http/docs/connections/http.rst
@@ -60,9 +60,6 @@ Schema (optional)
 Use DNS SRV Lookup (optional)
     Treat the Host field as a DNS SRV record name and resolve the target host/port at request time.
 
-SRV Cache TTL (seconds) (optional)
-    Specify the time to cache a resolved SRV target before re-resolving. (default 60 seconds)
-
 Extra (optional)
     Specify headers and default requests parameters in json format.
     Following default requests parameters are taken into account:
@@ -74,9 +71,9 @@ Extra (optional)
     * ``allow_redirects``
     * ``max_redirects``
 
-    "Use DNS SRV Lookup" and "SRV Cache TTL" above are stored as the ``srv_lookup`` and
-    ``srv_cache_ttl`` keys in this same Extra field, so they can also be set directly in json
-    here, e.g. when configuring the connection via an environment variable.
+    "Use DNS SRV Lookup" above is stored as the ``srv_lookup`` key in this same Extra field, so it
+    can also be set directly in json here, e.g. when configuring the connection via an environment
+    variable.
 
 
 When specifying the connection in environment variable you should specify
--- a/providers/http/docs/connections/http.rst
+++ b/providers/http/docs/connections/http.rst
@@ -47,11 +47,21 @@ Password (optional)
 Host (optional)
     Specify the entire url or the base of the url for the service.
 
+    If "Use DNS SRV Lookup" is enabled, specify the DNS SRV record name instead
+    (e.g. ``_http._tcp.example.com``) - Note the actual host and port are resolved from DNS at
+    request time and any value set in the Port field is ignored.
+
 Port (optional)
-    Specify a port number if applicable.
+    Specify a port number if applicable. Ignored when SRV lookup is enabled.
 
 Schema (optional)
     Specify the service type etc: http/https.
+
+Use DNS SRV Lookup (optional)
+    Treat the Host field as a DNS SRV record name and resolve the target host/port at request time.
+
+SRV Cache TTL (seconds) (optional)
+    Specify the time to cache a resolved SRV target before re-resolving. (default 60 seconds)
 
 Extra (optional)
     Specify headers and default requests parameters in json format.
@@ -64,6 +74,10 @@ Extra (optional)
     * ``allow_redirects``
     * ``max_redirects``
 
+    "Use DNS SRV Lookup" and "SRV Cache TTL" above are stored as the ``srv_lookup`` and
+    ``srv_cache_ttl`` keys in this same Extra field, so they can also be set directly in json
+    here, e.g. when configuring the connection via an environment variable.
+
 
 When specifying the connection in environment variable you should specify
 it using URI syntax.
@@ -75,3 +89,10 @@ For example:
 .. code-block:: bash
 
    export AIRFLOW_CONN_HTTP_DEFAULT='http://username:password@service.com:80/https?headers=header'
+
+To enable SRV lookup via an environment variable, set ``srv_lookup`` in the Extra query
+parameter:
+
+.. code-block:: bash
+
+   export AIRFLOW_CONN_HTTP_DEFAULT='https://_http._tcp.example.com/https?srv_lookup=true'
--- a/providers/http/provider.yaml
+++ b/providers/http/provider.yaml
@@ -126,4 +126,23 @@ connection-types:
       hidden-fields: []
       relabeling: {}
       placeholders: {}
-    conn-fields: {}
+    conn-fields:
+      srv_lookup:
+        label: Use DNS SRV Lookup
+        description: >-
+          Whether to treat the Host field as a DNS SRV record name and resolve the target
+          host/port at request time.
+        schema:
+          type:
+            - boolean
+            - "null"
+          default: false
+      srv_cache_ttl:
+        label: SRV Cache TTL (seconds)
+        description: Time to cache a resolved SRV target before re-resolving.
+        schema:
+          type:
+            - number
+            - "null"
+          minimum: 0
+          default: 60
--- a/providers/http/provider.yaml
+++ b/providers/http/provider.yaml
@@ -137,12 +137,3 @@ connection-types:
             - boolean
             - "null"
           default: false
-      srv_cache_ttl:
-        label: SRV Cache TTL (seconds)
-        description: Time to cache a resolved SRV target before re-resolving.
-        schema:
-          type:
-            - number
-            - "null"
-          minimum: 0
-          default: 60
--- a/providers/http/src/airflow/providers/http/hooks/http.py
+++ b/providers/http/src/airflow/providers/http/hooks/http.py
@@ -17,10 +17,8 @@
 # under the License.
 from __future__ import annotations
 
-import asyncio
 import copy
 import random
-import time
 from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable
 from contextlib import asynccontextmanager
 from typing import TYPE_CHECKING, Any, cast
@@ -44,6 +42,7 @@ from airflow.utils.strings import to_boolean
 
 if TYPE_CHECKING:
     from aiohttp.client_reqrep import ClientResponse
+    from dns.rdtypes.IN.SRV import SRV
     from requests.adapters import HTTPAdapter
 
     from airflow.models import Connection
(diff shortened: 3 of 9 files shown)


providers/http/src/airflow/providers/http/hooks/http.py
```diff
@@ -218,6 +246,9 @@ def get_conn(
     def _set_base_url(self, connection) -> None:
         host = connection.host or self.default_host
         schema = connection.schema or "http"
+        extra = connection.extra_dejson
+        self._srv_lookup_enabled = to_boolean(str(extra.get("srv_lookup", False)))
+        self._srv_cache_ttl = float(extra.get("srv_cache_ttl", self._srv_cache_ttl))
```
aaron-y-chen, line 254: Could we remove `srv_lookup` and `srv_cache_ttl` from `conn_extra_options` before applying the remaining connection extras as headers? Since these values are read with `get()`, they remain in `conn_extra_options` and are passed to `session.headers` here. When preparing the request, `requests` rejects their boolean and numeric values as invalid header values. The async path appears to have the same issue.

https://github.com/apache/airflow/blob/07f26a6f89cb35324aa6514fe1074bd081483cca/providers/http/src/airflow/providers/http/hooks/http.py#L327-L339

aaron-y-chen (commented):
Thanks for the PR 🙂

providers/http/src/airflow/providers/http/hooks/http.py
gtxu: I have poped the `srv_lookup` and `srv_cache_ttl` in `_process_extra_options_from_connection` so they are consumed directly from `extra_dejson` by the SRV lookup now. This works for async as well.

Thank you for point it out!

Also add tests to validate whether extra options added here is string only.

omkar-foss (commented):
PR looks good to me overall. Newly added tests are looking well-covered, would be nice if someone could initiate the ci checks in this PR so we can see them run and pass.

PS: I'm not a maintainer, my review is non-binding. Just trying to help out :)

gtxu:
Found out the root cause CI failures (3/3 runs) targeting `TestHttpAsyncHookSrvLookup::test_run_resolves_srv_record`. 
The test was failing with a real network query (`dns.resolver.NoAnswer`), bypassing the mock, yet consistently passing in local environments.

In Python 3.10(CI Env), `mock.patch` uses `_dot_lookup` (which calls `getattr`), meaning it successfully patched the *stale* attribute on the `dns` package. When the hook subsequently executed `import dns.asyncresolver`, it found no entry in `sys.modules`, loaded a fresh real module, and bypassed the mock entirely. 

Python 3.12 uses `pkgutil.resolve_name`, which performs a real import and self-heals, explaining why the test failure is not identified in my local test.

Added a `stable_dns_import` fixture for both SRV test classes. 

This fixture re-registers the submodules:
```python
importlib.import_module("dns.resolver")
importlib.import_module("dns.asyncresolver")
```
---
Drafted-by: Claude Code (Sonnet 5); reviewed by @gtxu

providers/http/docs/connections/http.rst
```diff
@@ -47,12 +47,22 @@ Password (optional)
...
-    Specify a port number if applicable.
+    Specify a port number if applicable. Ignored when SRV lookup is enabled.
 
 Schema (optional)
     Specify the service type etc: http/https.
 
+Use DNS SRV Lookup (optional)
+    Treat the Host field as a DNS SRV record name and resolve the target host/port at request time.
+
+SRV Cache TTL (seconds) (optional)
```
ashb: I can understand your reasoning, but I don't think this is right for two reasons.

1. DNS caching is a notoriously complex thing to deal with (do you cache poisitve, or NX/negative answers too)
2. The practical benefit of caching is very limited, because this caching will apply to exactly one connection, which in all likely hood will just resolve one host then make connections to it.

ashb (changes requested):

providers/http/docs/connections/http.rst
gtxu: Hi @ashb , I agree with your feedback that we should drop the DNS caching.

One thing worth noting for callers using the multi-call async session pattern (`HttpAsyncHook.session()`) with SRV lookup enabled: without a cache, resolving the SRV record fresh on every request means the target host/port can change between calls. That effectively defeats aiohttp's pooled `TCPConnector` (which keys connections on `(scheme, host, port)`), so TCP keep-alive won't be as effective across those dynamic routes. Given the short-lived `HttpHook/HttpAsyncHook` instances in Airflow tasks, this trade-off is minor.

Removed cache from hooks and tests. Thank you for reviewing this PR!

providers/http/docs/connections/http.rst
```diff
@@ -75,3 +86,10 @@ For example:
 .. code-block:: bash
 
    export AIRFLOW_CONN_HTTP_DEFAULT='http://username:password@service.com:80/https?headers=header'
+
+To enable SRV lookup via an environment variable, set ``srv_lookup`` in the Extra query
+parameter:
+
+.. code-block:: bash
+
+   export AIRFLOW_CONN_HTTP_DEFAULT='https://_http._tcp.example.com/https?srv_lookup=true'
```
aaron-y-chen, line 95: Copying this URI sets `conn_type` to `https`. `HttpOperator` then fails when resolving the hook with `Unknown hook type "https"`, before making any HTTP request.


```suggestion
   export AIRFLOW_CONN_HTTP_DEFAULT='http://_http._tcp.example.com/https?srv_lookup=true'
```

providers/http/src/airflow/providers/http/hooks/http.py
```diff
@@ -52,6 +55,19 @@ def _url_from_endpoint(base_url: str | None, endpoint: str | None) -> str:
     return (base_url or "") + (endpoint or "")
 
 
+def _select_srv_target(answers: Iterable[SRV]) -> tuple[str, int]:
+    """Select a target host and port from resolved DNS SRV records."""
+    candidates_by_priority: dict[int, list[SRV]] = {}
+    for record in answers:
+        candidates_by_priority.setdefault(record.priority, []).append(record)
+    # RFC 2782 priority failover; weight is not honored, ties broken uniformly at random.
+    candidates = candidates_by_priority[min(candidates_by_priority)]
```
aaron-y-chen, line 64: With `{10: [primary], 20: [backup]}`, this selects only `[primary]`. If `primary` is unreachable, all three `run_with_advanced_retry()` attempts still target it; the healthy `backup` is never tried, so the advertised SRV failover does not work.

aaron-y-chen (changes requested):

providers/http/docs/connections/http.rst
gtxu: Good catch, thank you! I will apply your suggestion.

providers/http/src/airflow/providers/http/hooks/http.py
gtxu: Thank you for the review! 

I have reworked it and so failover happens within a single request instead of depending on retries:

1. `_order_srv_targets` returns every target in RFC 2782 order: priority ascending, then weighted random within a same priority. Weight is now honored.

2. `HttpHook.run` and `AsyncHttpSession.run` try each target in turn. They move to the next one only when a connection can not be opened by error: `NewConnectionError` / `ConnectTimeoutError`, or `aiohttp.ClientConnectorError` / `ConnectionTimeoutError`. Errors after the connection is open (for example a server disconnect) and HTTP error responses are raised as before, so a non-idempotent request is never sent twice. If every SRV target fails, the last error is raised.

This works without `run_with_advanced_retry` as well, and the async hook's retry loop now runs once per target. 

Tests also cover the case for both hooks, the order of targets, errors that must not trigger failover, and every target being unreachable.

aaron-y-chen (approved):

providers/http/provider.yaml
```diff
@@ -127,4 +127,14 @@ connection-types:
...
+      srv_lookup:
+        label: Use DNS SRV Lookup
+        description: >-
+          Whether to treat the Host field as a DNS SRV record name and resolve the target
+          host/port at request time.
+        schema:
+          type:
+            - boolean
+            - "null"
+          default: false
```
ashb: Nit: Shoudl the default be not set (`null`), rather than False? Not sure it makes any difference mind you

providers/http/docs/connections/http.rst
```diff
@@ -47,12 +47,23 @@ Password (optional)
...
 
 Schema (optional)
     Specify the service type etc: http/https.
 
+Use DNS SRV Lookup (optional)
+    Treat the Host field as a DNS SRV record name and resolve the target host/port at request time.
+
+    Targets are tried in `RFC 2782 <https://www.rfc-editor.org/rfc/rfc2782>`__ order: lowest priority
+    first, weighted random within the same priority. If a connection to a target cannot be established,
+    the request fails over to the next target.
```
ashb, line 65: Probably need to mention this needs the `srv` extra too? Here I think

ashb (approved):
Small nit

providers/http/provider.yaml
gtxu: I kept `default: false` because it matches the other boolean connection fields(like `create_if_missing` in google). I am happy to drop it if you'd rather it stayed unset.

providers/http/docs/connections/http.rst
gtxu: Working on it. Thank you for reviewing!

providers/http/provider.yaml
gtxu: I see mongo's `srv` uses unset options. I will drop the default `False`. Thank you!

providers/http/docs/connections/http.rst
gtxu: Done

PR #68269, opened 2026-06-09
Batch triggerer unused trigger cleanup deletes

BeauDevCode:
closes: #68243

## Why

`Trigger.clean_unused()` currently removes all unreferenced trigger rows in one unbounded delete. On busy deployments this can hold locks on the `trigger` table for the duration of a large cleanup and stall the triggerer loop.

## What changed

- Select unused trigger IDs before deleting, preserving the existing unused-trigger criteria.
- Delete matching trigger rows in configurable batches, committing between batches.
- Add `[triggerer] unreferenced_triggers_cleanup_batch_size`, defaulting to `500`; setting it to `0` keeps single-transaction cleanup behavior.
- Add a unit test that uses a batch size of `2` and verifies five unused triggers are removed across three committed delete batches.

Related context: PR #68244 changes the anti-join shape in the same method, but does not add bounded delete batches. This PR intentionally focuses on the batching behavior requested in #68243.

## Tests

- `git diff --check`
- `python3 -m compileall -q airflow-core/src/airflow/models/trigger.py airflow-core/tests/unit/models/test_trigger.py`
- Parsed `airflow-core/src/airflow/config_templates/config.yml` with PyYAML and verified the new option shape

Not run locally: focused pytest/config validation could not complete because this environment lacks parts of Airflow's development dependency set (`pathspec` during test collection; `requests` for `dev/validate_version_added_fields_in_config.py`).

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @BeauDevCode** · by `@potiuk` · 2026-06-17 14:51 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed:
> - **Static / docs checks failing** (`CI image checks / Build documentation (--spellcheck-only)`). Run them locally with `prek run --all-files` (or `pre-commit run --all-files`) and push the fixes.
> - See the [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #70961, opened 2026-08-03
Batch triggerer unused trigger cleanup deletes

### LINKED ISSUE #68243: Batch the triggerer's Trigger.clean_unused() cleanup deletes

Providers

### Apache Airflow version

3.3.0.dev — current development branch

## Context

`airflow.models.trigger.Trigger.clean_unused()` is called every triggerer tick from `TriggererJobRunner._run_trigger_loop` and currently issues an unbounded `DELETE FROM trigger ...` (with a MySQL / non-MySQL fork for the `DELETE ... JOIN` problem). It has the same anti-pattern that the scheduler-side `SchedulerJobRunner._remove_unreferenced_triggers` was just fixed for: holds row locks on `trigger` for the full transaction and stalls the triggerer loop while many rows are removed.

Follow-up to #68241, which fixed the equivalent scheduler-side path and introduced the `[scheduler] unreferenced_triggers_cleanup_batch_size` config.

## Proposed fix

Apply the same LIMIT-bounded `select-IDs + delete-by-IDs + commit-between-batches` pattern that the scheduler-side fix uses, modelled on `airflow.utils.db_cleanup._do_delete` and `airflow.state.metastore.cleanup`. Either:

- Reuse the existing `[scheduler] unreferenced_triggers_cleanup_batch_size` config, or
- Add a triggerer-section twin (e.g. `[triggerer] unreferenced_triggers_cleanup_batch_size`) if reviewers prefer per-component tuning.

## Acceptance criteria

- The bulk `DELETE` in `Trigger.clean_unused()` is replaced by a batched loop.
- A unit test asserts multiple commits when the matching set exceeds the batch size.
- Existing `clean_unused` tests keep passing (semantics unchanged at default batch size).

## References

- AGENTS.md rule: [batched bulk DELETE/UPDATE in scheduler / interval callbacks](https://github.com/apache/airflow/blob/main/AGENTS.md#coding-standards)
- Template pattern: `airflow-core/src/airflow/utils/db_cleanup.py` — `_do_delete`
- Template pattern: `airflow-core/src/airflow/state/metastore.py` — `cleanup`

This is a follow-up tracking issue (not a bug report). It captures deferred work from PR #68241, which fixed the scheduler-side `_remove_unreferenced_triggers` cleanup but intentionally left the parallel `Trigger.clean_unused()` path in the triggerer loop for a separate PR.

The triggerer's `airflow.models.trigger.Trigger.clean_unused()` runs on every triggerer tick and issues an unbounded `DELETE FROM trigger ...` (with a MySQL / non-MySQL fork for the `DELETE ... JOIN` problem). On busy installs this holds row locks on `trigger` for the full transaction and stalls the triggerer loop while many rows are removed — the same anti-pattern the AGENTS.md rule about batched bulk DELETE/UPDATE in scheduler/interval callbacks calls out.

### Operating System

N/A — not OS-specific (server-side scheduler/triggerer behaviour, applies to all deployments)

### Deployment

None

_No response_

N/A — not provider-specific (issue is in airflow-core's triggerer cleanup path)

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

shivaam:
`Trigger.clean_unused()` runs synchronously in every triggerer supervisor loop. It currently removes every eligible trigger in one transaction, so busy metadata databases can hold many row locks and delay the triggerer heartbeat and concurrent writers.

This change selects and deletes unused trigger IDs in bounded transactions, committing between batches. It keeps the current `NOT EXISTS` predicates and `SKIP LOCKED` candidate locking from #68244, uses a portable bounded select-ID/delete-ID shape on every dialect, and retries a failed batch in a fresh transaction. A new positive `[triggerer] unreferenced_triggers_cleanup_batch_size` option defaults to 500; non-positive values fail fast rather than restoring an unbounded path.

Focused coverage exercises empty, below/exactly/above-threshold and multi-batch cleanup, protected deferred-task, asset, and callback references, invalid configuration, committed progress across a later failed batch, fresh retry transactions, and eventual drain.

Prior work: #68269 proposed the original batching direction. This implementation is a clean current-main rewrite because that draft predates #68244 and committed inside a caller-injected session.

Scope note: this PR leaves the pre-existing unbounded TaskInstance reference update and portable `callback.trigger_id` indexing unchanged; those require separate batching/index and migration designs.

Validation:

- Focused PostgreSQL and MySQL tests through Breeze: 9 passed on each backend.
- Real scheduler/triggerer PostgreSQL validation with 1,005 eligible rows, bounded `[500, 500]` deletes while 5 rows were concurrently locked, eventual drain after lock release, and protected asset/callback rows preserved.
- Ruff, formatting, YAML, spelling, mypy, security, and changed-file repository checks passed.
- Rebased cleanly onto `upstream/main` at `66444504c4`; the rebased patch ID is identical to the runtime-tested patch.
- The Docker-backed default-configuration hook passed before rebase. Its post-rebase rerun is currently blocked before validation by a local Colima file bind-mount error after a successful CI-image rebuild; CI can independently confirm it.

closes: #68243

---

- [X] Yes — Codex (GPT-5)


airflow-core/src/airflow/models/trigger.py
```diff
@@ -230,42 +230,58 @@ def fetch_trigger_ids_with_non_task_associations(cls, *, session: Session = NEW_
...
+            for attempt in run_with_db_retries():
+                with attempt:
+                    with create_session(scoped=False) as session:
+                        if clear_task_instance_references:
+                            # Update all task instances with trigger IDs that are not DEFERRED to remove them
+                            session.execute(
+                                update(TaskInstance)
+                                .where(
+                                    TaskInstance.state != TaskInstanceState.DEFERRED,
+                                    TaskInstance.trigger_id.is_not(None),
```
shivaam, line 256: The existing TaskInstance.trigger_id cleanup update remains unbounded. I’m keeping that separate from this PR because batching it safely requires selecting TaskInstances by their id, locking the triggers, and validating the approach across supported databases. This PR stays focused on #68243’s Trigger deletion problem

airflow-core/src/airflow/models/trigger.py
```diff
@@ -230,42 +230,58 @@ def fetch_trigger_ids_with_non_task_associations(cls, *, session: Session = NEW_
...
+            if deleted_count < batch_size:
+                return
+
+    @classmethod
+    def _delete_unused_batch(cls, batch_size: int, *, session: Session) -> int:
+        ids = (
+            select(cls.id)
+            .where(
+                ~cls.assets.any(),
+                ~cls.callback.has(),
```
shivaam, line 272: Noticed that `callback.trigger_id` has no explicit index, which could make this lookup slower.

airflow-core/src/airflow/models/trigger.py
```diff
@@ -231,42 +231,58 @@ def fetch_trigger_ids_with_non_task_associations(cls, *, session: Session = NEW_
...
-            )
-        else:
-            session.execute(
-                delete(Trigger).where(Trigger.id.in_(ids)).execution_options(synchronize_session=False)
-            )
+        ids_list = list(session.scalars(ids))
+        if not ids_list:
+            return 0
+
+        session.execute(delete(cls).where(cls.id.in_(ids_list)).execution_options(synchronize_session=False))
```
kaxil, line 284: Worth a look before this lands: dropping the dialect branch also drops the one thing that made the non-MySQL path safe.

Previously the `else` branch passed the `ids` select straight into the DELETE, so the three `~...any()/has()` predicates were re-evaluated inside the statement. This now materialises `ids_list` for every backend and deletes by id alone, which is the MySQL-only shape that #72061 reports as deleting task instances.

The window is between the SELECT and this DELETE. `with_row_locks` covers it while `[scheduler] use_row_level_locking` is on, but it returns the query unchanged when that is off (`airflow/utils/sqlalchemy.py`), and then nothing holds those rows. A referrer attaching in the gap is reachable: the Dag processor reuses committed trigger rows by `(classpath, kwargs)` when it registers asset watchers (`dag_processing/collection.py`). Both `task_instance.trigger_id` and `asset_watcher.trigger_id` are `ON DELETE CASCADE`, so deleting a re-referenced trigger takes the referring row with it.

Cheap fix that keeps the batching, re-applying the same predicates in the DELETE:

```python
unreferenced = (~cls.assets.any(), ~cls.callback.has(), ~cls.task_instance.has())
ids = select(cls.id).where(*unreferenced).order_by(cls.id).limit(batch_size)
...
session.execute(delete(cls).where(cls.id.in_(ids_list), *unreferenced).execution_options(synchronize_session=False))
```

These correlate to `task_instance`, `callback` and `asset_watcher` rather than to `trigger`, so it does not hit the MySQL restriction that motivated the two-step in the first place. I checked that it compiles and runs on MySQL 8.0.

PR #70962, opened 2026-08-03
Generate the Go SDK API clients with upstream oapi-codegen

jason810496:
## Why

The Go SDK's Execution API (`pkg/api`) and Edge worker API (`pkg/edgeapi`) clients were produced by a bespoke, resty-based code generator and used `resty.dev/v3` as their HTTP client. This moves both clients onto the widely-used upstream `github.com/oapi-codegen/oapi-codegen` generator with its standard `net/http` client, so the SDK's HTTP stack is just the Go standard library plus the mainstream oapi-codegen toolchain.

## What

- **Regenerate both clients** with upstream `oapi-codegen` v2 (`net/http` client). The generator config is unchanged apart from dropping generator-specific options that upstream covers by default; the existing overlays (operationId renames, `x-go-type`, header removal) still apply.
- **Rework the hand-written wrappers** onto the generated client:
  - `Airflow-API-Version` / `Correlation-Id` headers and the Edge worker's per-request JWT become oapi-codegen **request editors**.
  - `WithBearerToken` cloning is reimplemented over the generated client.
  - The Edge worker's retry policy (transient network errors, connection-refused, HTTP 502, bounded backoff) moves to an `http.RoundTripper`.
- **Consumers** (`sdk/client.go`, `pkg/worker/runner.go`, `edge/worker.go`) use the generated typed `*WithResponse` methods; non-2xx handling goes through a small `CheckResponse` helper.
- **Mocks** are regenerated against the new interfaces (`ClientWithResponsesInterface`), and the tests are updated to drive the flat client.
- Drop the resty debug-logging bridge along with the `resty.dev/v3` dependency; `go.mod` / `go.sum` (and the lang-SDK Go example module) are re-tidied.

The one intentional behaviour change: the Edge worker no longer emits resty's per-request debug log lines (the `api_client.debug` path); everything else — headers, auth, retry semantics, typed responses — is preserved. `go build ./...`, `go vet`, and `go test ./...` pass for the Go SDK.

---


jason810496:
Closed as I prefer to fix with Ash's upstream package instead of dropping the whole `resty` .

PR #62501, opened 2026-02-26
Fix asset event scheduling race condition and consistency

dingo4dev:
This PR addresses a race condition in the Asset-based scheduling logic where AssetEvent records created by concurrent tasks can be "missed" or "orphaned" by the Scheduler.

The issue stems from the Visibility Gap between a database flush() (which generates the created_at timestamp) and the final commit() (which makes the record visible to other sessions).

This change ensures that the synchronization between the AssetEvent and the AssetDagRunQueue is atomic and that the timestamps are aligned to prevent the drift caused by performance issue.

- Manually assigns the same timestamp to both the AssetEvent and the ADRQ before the flush.
- Commit and populate the asset event record in database immediately, so that the event can be accessible by job scheduler 
- Prevent creating dag by empty asset event as the asset event is trigger in prev_dag_run

related: #54659, #56750, #56749

``` mermaid
---
config:
  theme: redux-dark-color
  look: neo
---
sequenceDiagram
    participant J1 as Task Instance 1
    participant J2 as Task Instance 2
    participant JS as Job Scheduler
    participant DB as Database

    Note over J1,DB: T0

    Note left of J1: T1
    J1-->>DB: Asset Event 1 (AE1) & commit
    activate J1
    activate DB
    Note left of J1: T2
    J2-->>DB: Asset Event 2 (AE2) & commit
    activate J2
    activate DB

    Note left of J1: T3
    J2-->>-DB: Asset Dag Run Queue (ADRQ)<br/>created_at = AE2.timestamp<br/>(flush & commit)
    deactivate DB
    Note left of J1: T4
    activate JS
    activate DB
    JS-->>DB: Fetch created_at from ADRQ
    JS-->>DB: Fetch events (AE1, AE2)
    JS->>JS: Create DAG with run_after = ADRQ created_at (AE2.timestamp)
    deactivate DB
    deactivate JS
    Note left of J1: T5
    J1-->>DB: Asset Dag Run Queue (ADRQ)<br/>created_at = AE1.timestamp<br/>(flush & commit)
    deactivate J1
    deactivate DB
    Note left of J1: T6
    activate JS
    activate DB
    JS-->>DB: Fetch created_at from ADRQ
    JS->>JS: No event found<br/>(AE1.timestamp < prev DAG run_after)
    deactivate DB
    deactivate JS
```
---

### Durability tradeoff (new failure mode)

[...]

Issue #70970, opened 2026-08-03
Restore atomicity of asset-event registration


uranusjr:
PR #62501 made `AssetManager._create_asset_event` commit each `AssetEvent` in a short-lived independent session, so events no longer commit atomically. A failure/crash after the event commits, but before the caller's transaction commits, now leaves an orphaned event (no ADRQ to consume it) or a duplicate event (on task retry). #62501's own description acknowledges this tradeoff.

We should drop the side session (single atomic commit for event + ADRQ + TI state), and make `_create_dag_runs_asset_triggered` tolerant of commit ordering.

The main idea is to make ADQR reference the asset event that generated it directly with a fk, instead of implicitly by asset_id + timestamp as introduced in #62501. This would remove the need for short-lived sessions. However, not using a timestamp range has some blast radius we’ll need to address, similar to when we changed the DagRun references from using execution_date to a concrete fk. It should be doable though.

## References

- Regressing PR: #62501 (cc @dingo4dev @Lee-W @kaxil). Bug it fixed: #54659; related #56750, #56749.
- Lock-contention fix: #70951.

HarnageaGabriel:
Hi @uranusjr, I see this is assigned to you. I'd be happy to take this on if you haven't started work on it yet — the proposed fix (drop the side session in `_create_asset_event`, single atomic commit, and add an FK from `AssetDagRunQueue` to the `AssetEvent` instead of relying on timestamp matching) makes sense to me after reading through #62501's diff.

Could you let me know whether this is open for a community contribution, or if you're already working on it yourself? If it's open, feel free to unassign/reassign and I'll get started; if you're already on it, I'll leave it to you.

Issue #39456, opened 2024-05-07
DAGs are able to see historical dataset events when created new


tosheer:
### Apache Airflow version

2.9.1

### What happened?

If a new dataset triggreed DAG is created for an already existing dataset. (Dataset has already existing dataset events) DAG see all dataset events from very first event for dataset.

### What you think should happen instead?

If a new dataset triggreed DAG is created for an already existing dataset. (Dataset has already existing dataset events). DAG should see all the dataset events from the time when DAG was added.

### How to reproduce

* Create a new DAG which has dataset event producer. Enable both the producer.
* Run producer dag multiple times.
* Create a new consumer DAG which also get triggered from same dataset with catchup false.
* Enable the new DAG.
* You will see new DAG will run as soon as enabled and see very first event in the dataset.
* In the next run whenever that was scheduled DAG will get all the events other than the first event it has already consumed.

```from __future__ import annotations

import pendulum

from airflow.datasets import Dataset
from airflow.models.dag import DAG
from airflow.operators.bash import BashOperator

# [START dataset_def]
tosheer_dag1_dataset = Dataset("s3://tosheer-dag1/output_1.txt", extra={"partition": "bye"})


with DAG(
    dag_id="tosheer_dataset_produces_1",
    catchup=False,
    start_date=pendulum.datetime(2021, 1, 1, tz="UTC"),
    schedule="@daily",
    tags=["produces", "dataset-scheduled"],
) as dag1:
    # [START task_outlet]
    BashOperator(outlets=[tosheer_dag1_dataset], task_id="producing_task_1", bash_command="sleep 5")
    # [END task_outlet]
```

After this add  
```
# [START dag_dep]
with DAG(
    dag_id="tosheer_dataset_consumes_1",
    catchup=False,
    start_date=pendulum.datetime(2021, 1, 1, tz="UTC"),
    schedule=[tosheer_dag1_dataset],
    tags=["consumes", "dataset-scheduled"],
) as dag3:
    # [END dag_dep]
    BashOperator(
        task_id="consuming_1",
        bash_command='echo "ti_key={{ triggering_dataset_events }}"',
    )
```

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Docker-Compose

### Deployment details

Plain vanilla airflow.

### Anything else?

This issue is associated with an earlier issue https://github.com/apache/airflow/issues/38826. Fix of that issue just fixed disabled - enabled / deleted - recreated.

### Are you willing to submit PR?

[...]

PR #70972, opened 2026-08-03
Reference asset events from asset_dag_run_queue

uranusjr:
More details in #70970.

Asset-triggered scheduling had the scheduler reconstruct which AssetEvents fed a run from a created_at time window. The prior fix #62501 committed each AssetEvent in separate short-lived sessions so it was visible early, but that broke atomicity.

This commit changes the ADRQ to directly reference the asset event that generates it, instead of using an implicit asset_id + timestamp combination. Atomicity is restored since we no longer needs to commit separately to get the timestamps right, while still being able to reference the correct asset events from the ADRQ.


dingo4dev:
@uranusjr Thank you for this follow-up fix. Moving ADRQ to reference AssetEvent directly resolves the race I hit in #62501 while restoring transaction atomicity. I really appreciate the careful implementation and quick turnaround.

airflow-core/src/airflow/models/asset.py
```diff
@@ -749,21 +749,28 @@ def __repr__(self):
 class AssetDagRunQueue(Base):
     """Model for storing asset events that need processing."""
 
-    asset_id: Mapped[int] = mapped_column(Integer, primary_key=True, nullable=False)
     target_dag_id: Mapped[str] = mapped_column(StringID(), primary_key=True, nullable=False)
+    asset_event_id: Mapped[int] = mapped_column(Integer, primary_key=True, nullable=False)
```
kaxil, line 753: Now that ADRQ can hold multiple rows per (asset_id, target_dag_id), the LEFT JOIN in `get_asset_triggered_next_run_info` (models/dag.py) fans out and inflates both counts: a dag scheduled on 2 assets with 3 queued events for one of them reports "3 of 4 assets updated" (checked with a quick sqlite repro). `total` and `ready` need distinct counting on the asset, e.g. `count(distinct DagScheduleAssetReference.asset_id)`. The single-uri `case((func.count() == 1, ...))` hack breaks the same way. Multiple rows per pair are routine here: an AND-condition dag accumulates one row per event while it waits, and the migration expansion can create many at once.

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_asset_event_id_to_adrq.py
```diff
@@ -0,0 +1,196 @@
...
+    FROM (
+        SELECT target_dag_id, MAX(created_at) AS triggered_date
+        FROM asset_dag_run_queue
+        GROUP BY target_dag_id
+    ) t
+) td ON td.target_dag_id = adrq.target_dag_id
+JOIN asset_event ae
+    ON ae.asset_id = adrq.asset_id
+   AND ae.timestamp <= td.triggered_date
+   AND ae.timestamp > COALESCE(td.floor_date, :floor_min)
```
kaxil: This floor only replays the catchup=True window. The pre-migration scheduler also floored catchup=False dags at `min(DagScheduleAssetReference.created_at)`, so a catchup=False dag whose first asset-triggered run is pending at upgrade time gets its whole pre-subscription backlog expanded here, and the consume-by-reference path will consume every staged row in that first run (the #39456 behaviour the floor existed for). I think the dsar floor can go into this COALESCE unconditionally: catchup=True dags don't need backlog rows staged at all, since the catchup branch in `_create_dag_runs_asset_triggered` re-derives unconsumed backlog events from the schedule references without a queue row.

airflow-core/src/airflow/models/asset.py
```diff
@@ -749,21 +749,28 @@ def __repr__(self):
...
         ForeignKeyConstraint(
             (asset_id,),
             ["asset.id"],
             name="adrq_asset_fkey",
             ondelete="CASCADE",
         ),
+        ForeignKeyConstraint(
+            (asset_event_id,),
+            ["asset_event.id"],
+            name="adrq_asset_event_fkey",
```
kaxil, line 771: asset_event is a `db clean` table (utils/db_cleanup.py), so this CASCADE gives `airflow db clean --tables asset_event` a new side effect: deleting old events now deletes pending queue rows. A queued row lives until the rest of the dag's condition is satisfied, so for an AND-condition dag pairing a frequent asset with a monthly/quarterly one (or one whose upstream producer stalls), the row can outlive a typical 30-90 day retention window. Cleaning past it silently un-readies the asset and the dag misses its trigger when the other asset finally fires. Before, the ADRQ row survived event cleanup and the run still happened. Worth a note in the db clean docs at least. Relatedly, while a condition is unmet, rows now accumulate per event rather than upserting one per asset, and `dags_needing_dagruns` loads all of them (with two joinedloads) every loop.

airflow-core/src/airflow/assets/manager.py
```diff
@@ -827,71 +779,33 @@ def _get_or_create_apdr(
...
+            from sqlalchemy.dialects.mysql import insert as my_insert
 
-        update_stmt = stmt.on_duplicate_key_update(
-            created_at=case(
-                (stmt.inserted.created_at >= AssetDagRunQueue.created_at, stmt.inserted.created_at),
-                else_=AssetDagRunQueue.created_at,
-            )
-        )
-        session.execute(update_stmt, values)
+            session.execute(my_insert(AssetDagRunQueue).prefix_with("IGNORE"), values)
```
kaxil: INSERT IGNORE downgrades more than duplicate-key errors to warnings on MySQL (FK violations, NOT NULL, truncation are all silently swallowed), while the postgres/sqlite path only ignores the PK conflict. A bad event id, for example, would raise on postgres but insert nothing here without a sound. `on_duplicate_key_update(asset_id=stmt.inserted.asset_id)` is the usual no-op trick that keeps the ignore scoped to duplicates.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2658,53 +2656,31 @@ def _create_dag_runs_asset_triggered(
...
-            # catchup off, floor it at when the Dag started scheduling on its assets so the
-            # backlog is skipped; with catchup on, only date.min applies and the backlog replays.
-            event_window_floor: list[Any] = [cte.c.previous_dag_run_run_after]
-            if not dag.catchup:
-                event_window_floor.append(
-                    select(func.min(DagScheduleAssetReference.created_at))
-                    .where(DagScheduleAssetReference.dag_id == dag.dag_id)
-                    .scalar_subquery()
+            referenced_event_ids = {adrq.asset_event_id for adrq in queued_adrqs}
+            event_predicate: ColumnElement[bool] = AssetEvent.id.in_(referenced_event_ids)
```
kaxil, line 2660: One behaviour delta worth confirming: the old select required the event's asset to still be in the dag's schedule (the DagScheduleAssetReference/alias join). Consuming by reference means a queued event for an asset that was since removed from the dag's schedule now gets consumed and shows up in the run's consumed_asset_events / triggering_asset_events, where before it was silently dropped along with its ADRQ row. Intended?

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_asset_event_id_to_adrq.py
```diff
@@ -0,0 +1,196 @@
...
+        floor_min = datetime(1970, 1, 1, tzinfo=timezone.utc)
+        conn.execute(sa.text(_STAGE_SQL), {"floor_min": floor_min})
+        conn.execute(sa.text("DELETE FROM asset_dag_run_queue"))
+
+    # 3. Make the reference NOT NULL, move the PK, add the FKs.
+    with op.batch_alter_table("asset_dag_run_queue") as batch_op:
+        batch_op.alter_column("asset_event_id", existing_type=sa.Integer(), nullable=False)
+        batch_op.drop_constraint("adrq_asset_fkey", type_="foreignkey")
+        batch_op.drop_constraint("assetdagrunqueue_pkey", type_="primary")
+        batch_op.create_primary_key("assetdagrunqueue_pkey", ["target_dag_id", "asset_event_id"])
```
kaxil, line 128: With target_dag_id now leading the PK, `idx_asset_dag_run_queue_target_dag_id` is redundant and could be dropped in this migration. Conversely asset_id loses its index on postgres (the old PK led with it, and postgres doesn't auto-index FKs); the /assets/{assetId}/queuedEvents endpoints filter on it, though the table is usually small enough that this may not matter.

airflow-core/src/airflow/models/asset.py
uranusjr: Actually, the entire `get_asset_triggered_next_run_info` seems unused, so I’ll remove the function instead.

airflow-core/src/airflow/models/asset.py
uranusjr: I would say the new behavior is more correct since arguably it is unintuitive for a deleted asset event to still be able to contribute to a dag being triggered.

I will also add a significant news fragment since this is a behavioral change.

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_asset_event_id_to_adrq.py
uranusjr: Changed the index.

airflow-core/src/airflow/assets/manager.py
uranusjr: Changed to on_duplicate_key_update instead. I use `asset_id=ADRQ.asset_id` instead so the update references the original column to make it more future-proof no-op.

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_asset_event_id_to_adrq.py
uranusjr: Added a LEFT JOIN and adjusted COALESCE to cover this.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
uranusjr: Unintended, but I would argue this is more correct since the asset event _was_ used to trigger the run. I think we should keep this behavioral change with a note somewhere.

uranusjr: Added to the news fragment.

kaxil (approved):

airflow-core/docs/howto/usage-cli.rst
```diff
@@ -223,6 +223,12 @@ The ``db clean`` command works by deleting from each table the records older tha
 
 You can optionally provide a list of tables to perform deletes on. If no list of tables is supplied, all tables will be included.
 
+.. note::
+
+  Cleaning the ``asset_event`` table also drops any queued-but-not-yet-consumed asset events. This mainly
+  affects DAGs waiting on a multi-asset condition, where a pending event can be purged before the condition
+  is met, meaning the DAG will not be triggered by it.
```
Lee-W:
```suggestion
  affects Dags waiting on a multi-asset condition, where a pending event can be purged before the condition
  is met, meaning the Dag will not be triggered by it.
```

airflow-core/newsfragments/70972.significant.rst
```diff
@@ -0,0 +1,13 @@
+Changes to how pending asset events trigger DAG runs
```
Lee-W:
```suggestion
Changes to how pending asset events trigger Dag runs
```

airflow-core/newsfragments/70972.significant.rst
```diff
@@ -0,0 +1,13 @@
...
+asset events that are still pending for a downstream DAG. This mainly affects
+DAGs waiting on a multi-asset condition, where a pending event can be purged
+before the condition is met. Deleting an asset event now consistently means it
+will not trigger a run; previously such a pending trigger could still fire after
+the event had been purged.
+
+Relatedly, an asset event that was queued for a DAG while it was scheduled on
+that asset is still used to trigger the run, and appears among the run's
+triggering events, even if the asset is later removed from the DAG's schedule.
+Previously, they were silently dropped.
```
Lee-W:
```suggestion
asset events that are still pending for a downstream Dag. This mainly affects
Dags waiting on a multi-asset condition, where a pending event can be purged
before the condition is met. Deleting an asset event now consistently means it
will not trigger a run; previously such a pending trigger could still fire after
the event had been purged.

Relatedly, an asset event that was queued for a Dag while it was scheduled on
that asset is still used to trigger the run, and appears among the run's
triggering events, even if the asset is later removed from the Dag's schedule.
Previously, they were silently dropped.
```

uranusjr:
Special note: This SHOULD NOT be backported.

airflow-core/tests/unit/assets/test_manager.py
```diff
@@ -212,23 +212,43 @@ def test_register_asset_change_no_downstreams(self, session, mock_task_instance)
...
+
+        Under the old side-session behaviour the event was committed independently and would be
+        orphaned if the caller's transaction rolled back. Registration is now atomic with the caller.
+        """
+        asset_manager = AssetManager()
+
+        asset = Asset(uri="test://atomic1", name="atomic_asset")
+        asm = AssetModel(uri="test://atomic1/", name="atomic_asset", group="asset")
+        session.add(asm)
+        session.execute(delete(AssetDagRunQueue))
```
Lee-W: why do we need to delete ADRQ here? should we make a `clear_adrq`/`reset_adrq` function instead if there's ADRQ leftover?

Lee-W (approved):
a few nits

airflow-core/tests/unit/assets/test_manager.py
uranusjr: Reasonable. I’ll change this.

PR #70974, opened 2026-08-03
Make serde errors for unsupported types a bit more actionable

amoghrajesh:
---

Previously these errors said only that a type could not be (de)serialized, giving no indication of what to do about it. 
Error looked like

```
```

Example in https://github.com/apache/airflow/pull/70791

Name the two ways a Dag author can make their own class work 

- a serialize/deserialize method pair
- or @dataclass/@attr.define)

and link to the docs page that already covers all the options in full, instead of leaving people to dig through source to find them.

---

### DIFF
--- a/task-sdk/src/airflow/sdk/serde/__init__.py
+++ b/task-sdk/src/airflow/sdk/serde/__init__.py
@@ -285,7 +285,11 @@ def serialize(o: object, depth: int = 0) -> U | None:
         dct[DATA] = serialize(data, depth + 1)
         return dct
 
-    raise TypeError(f"cannot serialize object of type {cls}")
+    raise TypeError(
+        f"cannot serialize object of type {cls}. Give it a `serialize()` method (and a matching "
+        "`deserialize(data, version)`), or decorate the class with @dataclass or @attr.define. "
+        "See: https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/serializers.html"
+    )
 
 
 def deserialize(o: T | None, full=True, type_hint: Any = None) -> object:
@@ -387,7 +391,11 @@ def deserialize(o: T | None, full=True, type_hint: Any = None) -> object:
         return cls(**deserialize_value)  # type: ignore[operator]
 
     # no deserializer available
-    raise TypeError(f"No deserializer found for {classname}")
+    raise TypeError(
+        f"No deserializer found for {classname}. It must provide a `deserialize(data, version)` "
+        "classmethod (matching how it serializes), or be decorated with @dataclass or @attr.define. "
+        "See: https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/serializers.html"
+    )
 
 
 def _convert(old: dict) -> dict:
--- a/task-sdk/src/airflow/sdk/serde/__init__.py
+++ b/task-sdk/src/airflow/sdk/serde/__init__.py
@@ -286,8 +286,9 @@ def serialize(o: object, depth: int = 0) -> U | None:
         return dct
 
     raise TypeError(
-        f"cannot serialize object of type {cls}. Give it a `serialize()` method (and a matching "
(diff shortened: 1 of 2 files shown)


task-sdk/src/airflow/sdk/serde/__init__.py
```diff
@@ -285,7 +285,11 @@ def serialize(o: object, depth: int = 0) -> U | None:
         dct[DATA] = serialize(data, depth + 1)
         return dct
 
-    raise TypeError(f"cannot serialize object of type {cls}")
+    raise TypeError(
+        f"cannot serialize object of type {cls}. Give it a `serialize()` method (and a matching "
```
Lee-W:
```suggestion
        f"Cannot serialize object of type {cls}. Give it a `serialize()` method (and a matching "
```

task-sdk/tests/task_sdk/serde/test_serde.py
```diff
@@ -523,8 +523,10 @@ def test_raise_undeserializable(self):
                 "__version__": 0,
             }
         )
-        with pytest.raises(TypeError, match="No deserializer"):
+        with pytest.raises(TypeError, match="No deserializer") as exc_info:
             deserialize(data)
+        assert "deserialize(data, version)" in str(exc_info.value)
```
Lee-W: would be better if we check the whole sentence instead

task-sdk/tests/task_sdk/serde/test_serde.py
```diff
@@ -646,5 +648,7 @@ def test_error_when_serializing_callable_without_name(self):
         with pytest.raises(
             TypeError,
             match="cannot serialize object of type <class 'tests.task_sdk.serde.test_serde.C'>",
-        ):
+        ) as exc_info:
             serialize(i)
+        assert "@dataclass or @attr.define" in str(exc_info.value)
```
Lee-W: same here

Lee-W (approved):
LGTM

task-sdk/src/airflow/sdk/serde/__init__.py
```diff
@@ -285,7 +285,11 @@ def serialize(o: object, depth: int = 0) -> U | None:
...
+        "`deserialize(data, version)`), or decorate the class with @dataclass or @attr.define. "
```
kaxil: serde calls this unbound off the class (`getattr(cls, "deserialize")(deserialize(value), version)`), so a plain `def deserialize(self, data, version)` written from this message serializes fine and then dies on read with a confusing missing-argument error. The docs example uses `@staticmethod` and the message in `deserialize()` below says classmethod, both of which work. Maybe say staticmethod in both messages to match the docs.

task-sdk/src/airflow/sdk/serde/__init__.py
amoghrajesh: Cool makes sense.

task-sdk/tests/task_sdk/serde/test_serde.py
amoghrajesh: Updated.

task-sdk/tests/task_sdk/serde/test_serde.py
amoghrajesh: Sure, handled.

task-sdk/src/airflow/sdk/serde/__init__.py
amoghrajesh: Updated both messages to say staticmethod, matching the docs example

amoghrajesh: [Make serde errors for unsupported types a bit more actionable](https://github.com/apache/airflow/pull/70974/commits/86d835f9b82a75905a145f1291ddfcf5e2f6c9f6)

vatsrahul1001 (approved):

potiuk:
nice!

PR #70977, opened 2026-08-03
Fix airflow config lint staying silent on conditional removal rules

Eason09053360:
`airflow config lint` never reported the removal rules marked with  `remove_if_equals` . The marker was treated as a reason to skip the rule entirely, so the current value was never compared against it and lint printed "No issues found" regardless. Five rules were affected, three of them breaking.

Same fix as #66370, which corrected the parallel `ConfigChange` in `airflow-ctl` but left the core CLI behind.

---

- [x] Yes — Claude Code (Opus 5)
---

### DIFF
--- a/airflow-core/tests/unit/cli/commands/test_config_command.py
+++ b/airflow-core/tests/unit/cli/commands/test_config_command.py
@@ -573,6 +573,13 @@ class TestConfigLint:
                 "log_filename_template",
                 "{{ ti.dag_id }}/{{ ti.task_id }}/{{ ts }}/{{ try_number }}.log",
             ),
+            (
+                "logging",
+                "log_filename_template",
+                "dag_id={{ ti.dag_id }}/run_id={{ ti.run_id }}/task_id={{ ti.task_id }}/"
+                "{% if ti.map_index >= 0 %}map_index={{ ti.map_index }}/{% endif %}"
+                "attempt={{ try_number }}.log",
+            ),
         ],
     )
     def test_lint_detects_shipped_conditional_removals(self, section, option, value, stdout_capture):
--- a/airflow-core/tests/unit/cli/commands/test_config_command.py
+++ b/airflow-core/tests/unit/cli/commands/test_config_command.py
@@ -531,6 +531,61 @@ class TestConfigLint:
 
         assert "Invalid value" not in normalized_output
 
+    @pytest.mark.parametrize(
+        ("remove_if_equals", "config_value", "expect_issue"),
+        [
+            pytest.param("removed_value", "removed_value", True, id="match"),
+            pytest.param("removed_value", "kept_value", False, id="no-match"),
+            pytest.param("", "", True, id="empty-string-match"),
+            pytest.param("", "kept_value", False, id="empty-string-no-match"),
+        ],
+    )
+    def test_lint_reports_conditional_removal_only_when_value_matches(
+        self, remove_if_equals, config_value, expect_issue, stdout_capture
+    ):
+        config_change = ConfigChange(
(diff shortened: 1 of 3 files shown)


airflow-core/tests/unit/cli/commands/test_config_command.py
```diff
@@ -531,6 +531,61 @@ def test_lint_detects_invalid_config_negative(self, stdout_capture):
...
+        ("section", "option", "value"),
+        [
+            ("core", "hostname", ":"),
+            ("email", "email_backend", "airflow.contrib.utils.sendgrid.send_email"),
+            ("elasticsearch", "log_id_template", "{dag_id}-{task_id}-{logical_date}-{try_number}"),
+            (
+                "logging",
+                "log_filename_template",
+                "{{ ti.dag_id }}/{{ ti.task_id }}/{{ ts }}/{{ try_number }}.log",
+            ),
```
potiuk, line 575: The other `log_filename_template` removal rule (`config_command.py:322-326`) isn't covered here — it's one of the three `breaking=True` ones.

```suggestion
            ),
            (
                "logging",
                "log_filename_template",
                "dag_id={{ ti.dag_id }}/run_id={{ ti.run_id }}/task_id={{ ti.task_id }}/"
                "{% if ti.map_index >= 0 %}map_index={{ ti.map_index }}/{% endif %}"
                "attempt={{ try_number }}.log",
            ),
```

potiuk (commented):
Nice catch — and thanks for bringing the core CLI in line with the `airflow-ctl`
fix in #66370. I checked both `ConfigChange.message` implementations and the
control flow now matches, so the two commands can't drift apart on this again.

I verified the five affected rules (`core/hostname`, both
`logging/log_filename_template` variants, `email/email_backend`,
`elasticsearch/log_id_template`) and confirmed the new logic reports each only
when the current value actually matches. Two small things before this goes in —
neither is a correctness problem.

### Test coverage misses one of the five rules

`test_lint_detects_shipped_conditional_removals` parametrizes four of the five
shipped `remove_if_equals` rules. The second `logging/log_filename_template`
entry (`config_command.py:322-326`) isn't covered, and it's one of the three
`breaking=True` ones. Since the description calls out five affected rules, it'd
be good for the test to pin all five — see the inline suggestion.

### Newsfragment

`airflow config lint` going from "No issues found" to reporting five
previously-silent rules (three of them breaking) is user-visible behaviour, so
this one warrants a newsfragment:

    echo "Fixed \`airflow config lint\` silently skipping removal rules that only apply to a specific value." \
      > airflow-core/newsfragments/70977.bugfix.rst

(#66370 didn't need one because `airflow-ctl` doesn't consume newsfragments —
`airflow-core` does.)

---

> *This review was drafted by an AI-assisted tool and confirmed by an Airflow
> maintainer. The findings below are observations, not blockers; an Airflow
> maintainer — a real person — will take the next look at the PR. If you think a
> finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

---

Eason09053360:
Both applied.
Test coverage — added the second logging/log_filename_template rule to the parametrized test, using the suggestion verbatim. I checked that the new case actually pins that specific rule rather than passing on any logging output: dropping just that ConfigChange makes only the new param fail (1 failed, 4 passed), and all five pass once it is restored.

Newsfragment — added `airflow-core/newsfragments/70977.bugfix.rst` with the suggested wording, using double backticks for the literal since newsfragments are RST and single backticks render as a title reference rather than inline code.

Thanks for the review !

Drafted-by: Claude Code (Opus 5); reviewed by @Eason09053360 before posting

henry3260 (approved):

PR #70981, opened 2026-08-03
Disable separate Keycloak JWTs for Airflow v3.2-

stephen-bracken:
Disable separate Keycloak JWTs in Airflow v3.2 and below
---

Handling the Keycloak JWTs separately requires the `KeycloakJWTMiddleware`, which is injected through `KeycloakAuthManager.get_fastapi_middlewares()`, which was added in Airflow v3.3.0. If using a version of Airflow without `get_fastapi_middleware`, revert to the old behaviour and inject the Keycloak JWTs as claims on the Airflow JWT.

- [X] No

### DIFF
--- a/providers/keycloak/docs/changelog.rst
+++ b/providers/keycloak/docs/changelog.rst
@@ -29,12 +29,12 @@ Changelog
 .....
 
 .. note::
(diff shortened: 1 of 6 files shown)


vincbeck (approved):
Looks good!

PR #70983, opened 2026-08-03
Return 404 from task state store endpoints for unknown task instances

ColtenOuO:
## Summary

The task state store is a per-task-instance key/value store that survives worker
crashes and retries within a Dag run — the place operators keep external job IDs,
checkpoints and progress metadata. Its REST API is addressed by four values that
together identify one task instance: `dag_id` + `dag_run_id` + `task_id` + `map_index`.

The two `DELETE` endpoints never checked that the task instance those four values
name actually exists, so they reported success while touching nothing. All six
endpoints declared `404` in `create_openapi_http_exception_doc(...)` — that doc is
maintained by hand and had drifted from the handlers in both directions, promising a
`404` the deletes could not raise and one the list endpoint has no way to raise
either.

`_require_ti()` was written as a plain helper that each handler had to remember to
call, and it was called from `PUT` and `PATCH` only. The delete handlers went
straight to the query, and because those queries filter the four columns on
`task_state_store` directly — they never join `task_instance` — a request naming a
task instance that doesn't exist simply matches zero rows. At the SQL level that is
indistinguishable from "this task instance exists but has no state stored", so the
delete succeeded.

## What that looked like

A typo in `task_id` made a delete report success while touching nothing:

```console
$ curl -X DELETE .../taskInstances/proces/state-store    # note the missing 's'
204 No Content
```

Clearing state store is typically a *pre-rerun cleanup* step — clearing a task
instance in the UI deliberately does **not** wipe its state store, so a stale
checkpoint survives into the next attempt. An operator wiping a bad external job ID
before re-running gets `204`, believes the state is gone, re-runs, and the task
resumes from exactly the state they thought they had removed. Nothing in the
response hints that the `task_id` was wrong.

The same resource also answered the same question two different ways depending on
the verb — a write refused, the matching delete claimed success:

```console
$ curl -X PUT    .../taskInstances/proces/state-store/job_id -d '{"value":"x"}'
404 {"detail": "Task instance not found for dag_id='etl', ..., task_id='proces', map_index=-1"}

$ curl -X DELETE .../taskInstances/proces/state-store/job_id
204 No Content
```

## Before / after

| Endpoint | Before | After |
| --- | --- | --- |
| `DELETE .../state-store/{key}` | `204` | `404` |
| `DELETE .../state-store` | `204` | `404` |
| `PUT .../state-store/{key}` | `404` | unchanged |
| `PATCH .../state-store/{key}` | `404` | unchanged |
| `GET .../state-store` | `200` + empty collection | unchanged (spec drops its `404`) |
| `GET .../state-store/{key}` | `404`, message names the key | unchanged |

A missing *key* on an existing task instance still reports the key, unchanged:
`Task state store key 'job_id' not found`.

The reads are deliberately left alone (per review): an empty collection and a
"key not found" are already honest answers, and they would otherwise pay for an
extra `task_instance` lookup on every read. Since the list endpoint therefore has
no `404` path at all — `requires_access_dag` only raises `401`/`403`, and
`paginated_select` raises nothing here — its `create_openapi_http_exception_doc`
entry is dropped, matching the other list endpoints (`get_pools`, `get_variables`,
…) that declare no error responses.

## The fix

`asset_state_store.py` — the sibling module from the same feature — already solves
this correctly: `_get_asset_or_404` is wired up as `AssetIdDep = Annotated[int,
Depends(...)]` and every one of its endpoints takes `asset_id: AssetIdDep`, so the
check rides on the parameter and cannot be forgotten when an endpoint is added.

This applies the same pattern on the task side, split in two so the reads do not
pay for a check they don't need:

- `TaskScopeDep` maps the path and query parameters onto the `TaskScope` the
  handlers already needed, absorbing the old `_get_scope()` helper. Every endpoint
  takes it, so each handler drops from four path/query parameters to a single
  `scope`.
- `ValidatedTaskScopeDep` builds on it and adds the existence check. Only the
  writes and deletes take it.

### Why `clear` needs its own dependency

`DELETE .../state-store?all_map_indices=true` means "clear this task at every map
index", and documents that `map_index` is ignored. `ValidatedTaskScopeDep` validates
one *exact* map index, defaulting to `-1` — and an expanded mapped task has no
`map_index=-1` instance: `TaskMap.expand_mapped_task` either converts that row into
`map_index=0` or deletes it. Validating with `ValidatedTaskScopeDep` would therefore
reject a perfectly valid request:

```console
# 'process' expanded to map_index 0, 1, 2 — there is no map_index=-1 row
$ curl -X DELETE .../taskInstances/process/state-store?all_map_indices=true
404 Task instance not found ... map_index=-1     # wrong
```

A dependency only receives the parameters it declares, so `ValidatedTaskScopeDep`
cannot see `all_map_indices` and can only ever take the strict path.
`ValidatedClearTaskScopeDep` declares the flag, and when it is set drops the
`map_index` predicate so any instance of that task in the run satisfies the check.
The 404 it raises names the flag rather than a `map_index` that was never applied:

```console
$ curl -X DELETE .../taskInstances/proces/state-store?all_map_indices=true
404 {"detail": "Task instance not found for dag_id='etl', run_id='...', task_id='proces', all_map_indices=True"}
```

The strict path is untouched: `?map_index=7` on a task expanded to 0..2 still
returns `404`.

## Generated files

`v2-rest-api-generated.yaml` and `ui/openapi-gen/` are regenerated. Two kinds of
change: the `404` removed from `GET .../state-store`, and parameter *ordering* —
moving path/query parameters behind a dependency changes the order FastAPI reports
them in, not the parameters themselves.

## Tests

New `TestUnknownTaskInstance` covers every request shape against both a bad
`task_id` and a bad `run_id`, a non-existent `map_index`, the `all_map_indices`
error detail, and regression tests that the reads still answer as they did — the
list returns an empty collection, and a genuinely missing key still reports the key
rather than the task instance.

76 → 96 tests in this file; `test_asset_state_store.py` passes unchanged.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 import json
 from datetime import datetime, timedelta, timezone
-from typing import Annotated, Literal
+from typing import TYPE_CHECKING, Annotated, Literal
 
 from fastapi import Depends, HTTPException, Query, status
 from sqlalchemy import select
@@ -41,14 +41,81 @@ from airflow.models.task_state_store import TaskStateStoreModel
 from airflow.models.taskinstance import TaskInstance as TI
 from airflow.state.metastore import _get_db_backend
 
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
 task_state_store_router = AirflowRouter(
     tags=["Task State Store"],
     prefix="/dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances/{task_id}/state-store",
 )
 
 
-def _get_scope(dag_id: str, dag_run_id: str, task_id: str, map_index: int) -> TaskScope:
+def _require_task_instance(
+    dag_id: str,
+    dag_run_id: str,
+    task_id: str,
+    map_index: int | None,
+    session: Session,
+) -> None:
+    """Raise 404 unless the task instance exists. ``map_index=None`` matches any map index."""
+    statement = select(TI.task_id).where(
+        TI.dag_id == dag_id,
+        TI.run_id == dag_run_id,
+        TI.task_id == task_id,
+    )
+    if map_index is not None:
+        statement = statement.where(TI.map_index == map_index)
+    if session.scalar(statement.limit(1)) is None:
+        addressed_by = "all_map_indices=True" if map_index is None else f"map_index={map_index}"
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=(
+                f"Task instance not found for dag_id={dag_id!r}, run_id={dag_run_id!r}, "
+                f"task_id={task_id!r}, {addressed_by}"
+            ),
+        )
+
+
+def _resolve_scope(
+    dag_id: str,
+    dag_run_id: str,
+    task_id: str,
+    map_index: Annotated[int, Query(ge=-1)] = -1,
+) -> TaskScope:
+    """Map the path and query parameters onto the task instance they address."""
     return TaskScope(dag_id=dag_id, run_id=dag_run_id, task_id=task_id, map_index=map_index)
+
+
+TaskScopeDep = Annotated[TaskScope, Depends(_resolve_scope)]
+
+
+def _validate_scope(scope: TaskScopeDep, session: SessionDep) -> TaskScope:
+    """Resolve the scope, 404ing when the task instance it addresses does not exist."""
+    _require_task_instance(scope.dag_id, scope.run_id, scope.task_id, scope.map_index, session)
+    return scope
+
+
+ValidatedTaskScopeDep = Annotated[TaskScope, Depends(_validate_scope)]
+
+
+def _validate_clear_scope(
+    scope: TaskScopeDep,
+    session: SessionDep,
+    all_map_indices: Annotated[bool, Query()] = False,
+) -> TaskScope:
+    """
+    Resolve the scope for a clear request, 404ing when the task instance does not exist.
+
+    ``all_map_indices`` addresses the task across every index, so it is validated against any
+    instance -- an expanded mapped task has no ``map_index=-1`` instance to check.
+    """
+    _require_task_instance(
+        scope.dag_id, scope.run_id, scope.task_id, None if all_map_indices else scope.map_index, session
+    )
+    return scope
+
+
+ValidatedClearTaskScopeDep = Annotated[TaskScope, Depends(_validate_clear_scope)]
 
 
 def _resolve_expires_at(expires_at: datetime | None | Literal["default"]) -> datetime | None:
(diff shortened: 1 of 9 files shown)


airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
```diff
@@ -22,6 +22,7 @@
 
 from fastapi import Depends, HTTPException, Query, status
 from sqlalchemy import select
+from sqlalchemy.orm import Session
```
jason810496: This should be imported under type checking only block.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
```diff
@@ -72,35 +128,16 @@ def _resolve_expires_at(expires_at: datetime | None | Literal["default"]) -> dat
...
 @task_state_store_router.get(
     "",
     responses=create_openapi_http_exception_doc([status.HTTP_404_NOT_FOUND]),
     dependencies=[Depends(requires_access_dag(method="GET", access_entity=DagAccessEntity.TASK_INSTANCE))],
 )
 def list_task_state_store(
-    dag_id: str,
-    dag_run_id: str,
-    task_id: str,
+    scope: TaskScopeDep,
```
jason810496, line 147: However, for the list endpoint, we will check the unnecessary TI rows now.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
```diff
@@ -142,12 +179,9 @@ def list_task_state_store(
     dependencies=[Depends(requires_access_dag(method="GET", access_entity=DagAccessEntity.TASK_INSTANCE))],
 )
 def get_task_state_store(
-    dag_id: str,
-    dag_run_id: str,
-    task_id: str,
+    scope: TaskScopeDep,
```
jason810496, line 192: Ditto.

jason810496 (commented):
Thanks for the fix.

It seems only those route calling the `_require_ti` helper need those new deps. Perhaps we can introduce different dependency injection helper (e.g. one for mapping the path parameters to the TaskScope another one leverage the first one but also validate the TI).

ColtenOuO:
Thanks for reviewing. 

I've addressed all comments.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_state_store.py
```diff
@@ -41,16 +41,80 @@
...
+        TI.task_id == task_id,
+    )
+    if map_index is not None:
+        statement = statement.where(TI.map_index == map_index)
+    if session.scalar(statement.limit(1)) is None:
+        detail = f"Task instance not found for dag_id={dag_id!r}, run_id={dag_run_id!r}, task_id={task_id!r}"
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=detail if map_index is None else f"{detail}, map_index={map_index}",
+        )
```
jason810496: We should show the `all_map_indices=True` in the error.

jason810496 (commented):
Thanks! LGTM overall after these comments.

We will need to update the PR description as well, since we don't validate those task_state_store routes now (`list_task_state_store`, `get_task_state_store` ).

ColtenOuO:
Thanks for reviewing.

I've changed the code to show the `all_map_indices=True` in the error.

By the way, I noticed that list endpoint can never return 404 now that it doesn't validate the task instance, so I dropped its create_openapi_http_exception_doc entry — same as get_pools and get_variables. PR description is updated as well.

bbovenzi (approved):

PR #70989, opened 2026-08-03
Fix airflowctl list operations ignoring the requested offset

rjgoyln:
## Summary

`BaseOperations.execute_list` accepts an `offset` but never sent it on the first request, so paging started at the server default while the loop still resumed at `offset + limit` — a non-zero offset both returned the wrong first page and skipped the rows in between. `offset=100, limit=2` returned rows 0-1, then jumped to row 102.

Nothing released is affected: no caller passes `offset`, and `jobs list` takes a separate path when one is given. But the parameter is public on the signature and silently discarded, so the first caller to use it gets wrong results rather than an error.

## Compatibility

Sending `offset=0` is equivalent to omitting it — the API's `QueryOffset` already defaults to `0`.

---

### DIFF
--- a/airflow-ctl/tests/airflow_ctl/api/test_operations.py
+++ b/airflow-ctl/tests/airflow_ctl/api/test_operations.py
@@ -252,19 +252,27 @@ class TestBaseOperations:
 
     def test_execute_list_sends_offset_to_server(self):
         """``offset`` must reach the server on the first request, not only on subsequent pages."""
+        rows = [{"name": name} for name in "abcdef"]
+
+        def paged(path, params):
+            # An absent ``offset`` is what the server would see as its own default of 0.
+            start = params.get("offset", 0)
+            return Mock(
+                content=json.dumps(
+                    {"hellos": rows[start : start + params["limit"]], "total_entries": len(rows)}
+                )
+            )
+
         mock_client = Mock()
-        mock_client.get.side_effect = [
-            Mock(content=json.dumps({"hellos": [{"name": "c"}, {"name": "d"}], "total_entries": 6})),
-            Mock(content=json.dumps({"hellos": [{"name": "e"}, {"name": "f"}], "total_entries": 6})),
-        ]
+        mock_client.get.side_effect = paged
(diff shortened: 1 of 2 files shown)


airflow-ctl/tests/airflow_ctl/api/test_operations.py
```diff
@@ -250,6 +250,23 @@ def test_execute_list_sends_limit_on_subsequent_pages(self):
...
+            Mock(content=json.dumps({"hellos": [{"name": "e"}, {"name": "f"}], "total_entries": 6})),
+        ]
+        base_operation = BaseOperations(client=mock_client)
+
+        response = base_operation.execute_list(
+            path="hello", data_model=HelloCollectionResponse, offset=2, limit=2
+        )
+
+        assert [call.kwargs["params"]["offset"] for call in mock_client.get.call_args_list] == [2, 4]
+        # The paging loop resumes at offset + limit, so a dropped offset also skips rows 2-3.
```
henry3260:
```suggestion
        rows = [{"name": name} for name in "abcdef"]

        def paged(path, params):
            # An absent ``offset`` is what the server would see as its own default of 0.
            start = params.get("offset", 0)
            return Mock(
                content=json.dumps(
                    {"hellos": rows[start : start + params["limit"]], "total_entries": len(rows)}
                )
            )

        mock_client = Mock()
        mock_client.get.side_effect = paged
        base_operation = BaseOperations(client=mock_client)

        response = base_operation.execute_list(
            path="hello", data_model=HelloCollectionResponse, offset=2, limit=2
        )

        assert [hello.name for hello in response.hellos] == ["c", "d", "e", "f"]

```

henry3260 (approved):
Overall looks good, just one nit

airflow-ctl/tests/airflow_ctl/api/test_operations.py
rjgoyln: Thank you so much! I had never considered that before, and I learned a lot.
You're so kind. I really appreciate your help!

PR #70990, opened 2026-08-03
Fix one bad callback request crashing the Dag processor and dropping the rest

ephraimbuddy:
A callback request that hits the documented race condition (the Dag or task was removed between scheduling the callback and parsing the file) raises out of _execute_callbacks, killing the parse subprocess. The remaining requests in the batch were already popped from the manager's queue, so on_failure_callback, on_retry_callback, and failure-email requests behind the bad one are lost permanently. Isolate each request so a failure is logged and the rest still execute, and route the last_ti task lookup through _get_dag_with_task so a removed task surfaces as the race-condition error instead of a raw TaskNotFound.

---

claude fable 5

### DIFF
--- a/airflow-core/src/airflow/dag_processing/processor.py
+++ b/airflow-core/src/airflow/dag_processing/processor.py
@@ -328,14 +328,15 @@ def _execute_callbacks(
     dagbag: DagBag, callback_requests: list[CallbackRequest], log: FilteringBoundLogger
 ) -> None:
     for request in callback_requests:
+        request_json = request.to_json()
         if isinstance(request, (TaskCallbackRequest, EmailRequest)):
             log.debug(
                 "Processing Callback Request",
-                request=request.to_json(),
+                request=request_json,
                 ti_id=str(request.ti.id),
             )
         else:
-            log.debug("Processing Callback Request", request=request.to_json())
+            log.debug("Processing Callback Request", request=request_json)
         # A failed request (e.g. the Dag or task was removed since the callback
         # was scheduled) must not abort the remaining requests in this batch --
         # they were already popped from the manager's queue and would be lost.
@@ -351,7 +352,7 @@ def _execute_callbacks(
                 elif isinstance(request, EmailRequest):
                     _execute_email_callbacks(dagbag, request, log)
         except Exception:
-            log.exception("Failed to execute callback request", request=request.to_json())
+            log.exception("Failed to execute callback request", request=request_json)
 
 
 def _execute_dag_callbacks(dagbag: DagBag, request: DagCallbackRequest, log: FilteringBoundLogger) -> None:
--- a/airflow-core/src/airflow/dag_processing/processor.py
+++ b/airflow-core/src/airflow/dag_processing/processor.py
@@ -328,15 +328,18 @@ def _execute_callbacks(
     dagbag: DagBag, callback_requests: list[CallbackRequest], log: FilteringBoundLogger
 ) -> None:
     for request in callback_requests:
-        request_json = request.to_json()
         if isinstance(request, (TaskCallbackRequest, EmailRequest)):
-            log.debug(
-                "Processing Callback Request",
-                request=request_json,
-                ti_id=str(request.ti.id),
-            )
+            log_extra = {
+                "dag_id": request.ti.dag_id,
+                "run_id": request.ti.run_id,
+                "ti_id": str(request.ti.id),
+            }
         else:
-            log.debug("Processing Callback Request", request=request_json)
+            log_extra = {"dag_id": request.dag_id, "run_id": request.run_id}
+        # context_from_server can carry user-supplied run conf, and the masker cannot
+        # redact inside an already-serialized string, so keep it out of log payloads.
+        request_json = request.to_json(exclude={"context_from_server"})
+        log.debug("Processing Callback Request", request=request_json, **log_extra)
         # A failed request (e.g. the Dag or task was removed since the callback
         # was scheduled) must not abort the remaining requests in this batch --
         # they were already popped from the manager's queue and would be lost.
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/dag_processing/processor.py
```diff
@@ -361,7 +367,9 @@ def _execute_dag_callbacks(dagbag: DagBag, request: DagCallbackRequest, log: Fil
     ctx_from_server = request.context_from_server
 
     if ctx_from_server is not None and ctx_from_server.last_ti is not None:
-        task = dag.get_task(ctx_from_server.last_ti.task_id)
+        _, task = _get_dag_with_task(dagbag, request.dag_id, ctx_from_server.last_ti.task_id)
+        if TYPE_CHECKING:
+            assert task is not None
```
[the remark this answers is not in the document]
ephraimbuddy: Type checking asserts are allowed in airflow core. We can't use RuntimeError since this code is only for type checkers and not reproducible in runtime

airflow-core/src/airflow/dag_processing/processor.py
```diff
@@ -361,7 +368,9 @@ def _execute_dag_callbacks(dagbag: DagBag, request: DagCallbackRequest, log: Fil
     ctx_from_server = request.context_from_server
 
     if ctx_from_server is not None and ctx_from_server.last_ti is not None:
-        task = dag.get_task(ctx_from_server.last_ti.task_id)
+        _, task = _get_dag_with_task(dagbag, request.dag_id, ctx_from_server.last_ti.task_id)
```
kaxil: If the dag still exists and only the task vanished, this now skips the dag-level callback entirely: the ValueError lands in the new except in `_execute_callbacks` and the request is unrecoverable (the DB row was already deleted when the manager claimed it). The dag lookup already succeeded at line 361, so the only thing actually lost is context enrichment. `produce_dag_callback` makes the opposite call for the same situation, it drops an unrepresentable `last_ti` and lets the callback fire with the minimal context (dag, run_id, reason). Catching the ValueError here and falling back to the minimal-context branch below would keep the user's notification alive, and the new test could then assert the callback ran with minimal context instead of pinning the error type.

airflow-core/src/airflow/dag_processing/processor.py
```diff
@@ -328,24 +328,31 @@ def _execute_callbacks(
...
+                bundle_version=request.bundle_version,
+            ):
+                if isinstance(request, TaskCallbackRequest):
+                    _execute_task_callbacks(dagbag, request, log)
+                elif isinstance(request, DagCallbackRequest):
+                    _execute_dag_callbacks(dagbag, request, log)
+                elif isinstance(request, EmailRequest):
+                    _execute_email_callbacks(dagbag, request, log)
+        except Exception:
+            log.exception("Failed to execute callback request", request=request_json)
```
kaxil: Two things about this line. It logs the full serialized request at ERROR on the default path, and for requests with `context_from_server` that includes `dag_run.conf`. The masker can't do key-based redaction inside an already-serialized string and the parse subprocess never registers run-conf values, so sensitive conf lands in the log verbatim, where before this PR the payload only ever appeared at DEBUG. It also doesn't bind `ti_id`/`dag_id` the way the debug branch above does, so structured-log filters won't match the one line that matters when a callback disappears. Logging the identifying fields plus `request.to_json(exclude={"context_from_server"})` would cover both. A stats counter here would help too, `dag.callback_exceptions` only fires when the callback body raises, so a request that dies before running stays invisible to alerting.

airflow-core/tests/unit/dag_processing/test_processor.py
```diff
@@ -814,6 +814,38 @@ def test_execute_callbacks_locks_bundle_version(self):
...
+                run_id="test_run",
+                bundle_name="testing",
+                bundle_version=None,
+                is_failure_callback=True,
+                msg="Message",
+            )
+
+        _execute_callbacks(dagbag, [make_request("removed_dag"), make_request("a")], structlog.get_logger())
+
+        assert called is True
```
kaxil, line 848: This pins the continue half of the new behavior but not the report half: delete or downgrade the `log.exception` call and this still passes. Passing `log = MagicMock(spec=FilteringBoundLogger)` like `test_execute_callbacks_locks_bundle_version` above and asserting `log.exception` was called once for the removed_dag request would cover both.

airflow-core/src/airflow/dag_processing/processor.py
ephraimbuddy: Took the log and context changes; I left the counter out of this PR since a request dying before its callback ran was equally invisible before this change, so it's a pre-existing observability gap and not one this PR introduces. If we add it, I think it belongs in a minor release as a new metric; happy to do that in a follow-up if you still want it.

airflow-core/tests/unit/dag_processing/test_processor.py
```diff
@@ -814,6 +814,41 @@ def test_execute_callbacks_locks_bundle_version(self):
...
+                is_failure_callback=True,
+                msg="Message",
+            )
+
+        log = MagicMock(spec=FilteringBoundLogger)
+        _execute_callbacks(dagbag, [make_request("removed_dag"), make_request("a")], log)
+
+        assert called is True
+        log.exception.assert_called_once()
+        assert log.exception.call_args.kwargs["dag_id"] == "removed_dag"
```
kaxil, line 850: Optional: this pins the log-context bind but not the exclusion. `exclude={"context_from_server"}` is the only thing keeping `DagRun.conf` out of dag-processor logs now that this payload is logged at ERROR, and dropping it is detectable here (on pydantic 2.13.4 the field serializes as `"context_from_server":null` even when it is `None`), so this assertion fails if the exclusion is ever removed:

```python
assert "context_from_server" not in log.exception.call_args.kwargs["request"]
```

kaxil (approved):
Both points are addressed well. The `ValueError` fallback keeps the dag-level callback alive with minimal context and cites the `produce_dag_callback` precedent, and `exclude={"context_from_server"}` keeps run conf out of the payload on both the debug and the error path, with `log_extra` binding the identifying fields so the error line is filterable. Reverting the fix locally makes both new tests fail with the original signatures, so they cover the real paths rather than the wiring. One optional test hardening below, not a blocker.

PR #70991, opened 2026-08-03
Stop swallowing unexpected keyring errors when saving airflowctl credentials

rjgoyln:
## Summary

The `TypeError` handler in `Credentials.save()` only produced a meaningful error when `api_token` was `None` and `client_kind` was `ClientKind.CLI` at once. No caller satisfies both: the token-login path leaves `client_kind` at its `None` default and has already rejected an empty token, the username/password path passes `ClientKind.AUTH` with a pydantic-validated token, and `ClientKind.CLI` only ever reaches `load()`.

That left the handler as an unconditional swallow — a credential save failing for any unexpected reason returned as though it had succeeded, and `airflowctl auth login` reported success anyway. Removing it lets those errors surface. The `NoKeyringError` handler above, which carries the actionable "keyring backend is not available" guidance, is unchanged.

---


henry3260:
Merging ⁠main⁠ to update the branch

airflow-ctl/src/airflowctl/api/client.py
Eason09053360: I think now that the handler is gone, this `# type: ignore[arg-type]` is the only thing acknowledging that `api_token` is `str | None`, and nothing guards a `None` token before it reaches keyring. What happens then is backend-specific: `keyrings.alt` raises a `TypeError` about *bytes*, macOS raises `PasswordSetError`, Windows silently stores the literal string `"None"`. Since the point of the PR is that `save()` should never be reached without a token, let's make that explicit and let mypy check the argument again:

```suggestion
                if self.api_token is None:
                    raise AirflowCtlCredentialNotFoundException("No API token found. Please login first.")
                keyring.set_password(
                    "airflowctl",
                    self.token_key_for_environment(self.api_environment),
                    self.api_token,
                )
```

Note: after applying this, `test_save_no_keyring` needs `api_token="TEST_TOKEN"` — it currently constructs `Credentials` without a token and only passes because the mock accepts `None`.

---
Drafted-by: Claude Code (Fable 5.1); reviewed by @Eason09053360 before posting

rjgoyln: apply! Thanks.

henry3260 (approved):

PR #70993, opened 2026-08-03
Warn on suspicious Dag and task IDs at TS SDK build time

Andrushika:
## Why

ts-sdk has validated Dag and task IDs by throwing at registration since #69400. The review in #69965 settled on best-effort build time warnings for all SDKs instead, so a stale client-side rule cannot block packing or task runs.

## What

`airflow-ts-pack` now warns on invalid IDs (over 250 chars, bad characters, `..`) during build time, and still packs. 
The task registry keeps only the non-empty check. ID length is counted in code points now, since `.length` counts astral characters twice.

related: #69400, #69965, #69937

---

### DIFF
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -194,7 +194,7 @@ function isBundleManifest(value: unknown): value is BundleManifest {
 }
 
 function isTaskIdList(value: unknown): value is string[] {
-  return Array.isArray(value) && value.every((item) => typeof item === "string");
+  return Array.isArray(value) && value.every((item) => typeof item === "string" && item.length > 0);
 }
 
 // esbuild keeps an entry hashbang as line 1, where the metadata comment must go;
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -21,9 +21,9 @@
 // artifact NodeCoordinator consumes — `bundle.mjs` with the airflow
 // metadata embedded as a leading `//# airflowMetadata=<base64>` comment.
 //
-// Mirrors airflow-go-pack: build first, then run the built bundle with
-// --airflow-metadata so the manifest comes from the bundle's own task
-// registry and schema version, never from a hand-written sidecar.
+// Build first, then run the built bundle with --airflow-metadata so the
+// manifest comes from the bundle's own task registry and schema version,
+// never from a hand-written sidecar.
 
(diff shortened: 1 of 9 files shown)


Andrushika:
Hi @jason810496 could you please take a look if you have time? Thanks a lot!

jason810496 (commented):
Sure! I prefer to wait until https://github.com/apache/airflow/pull/71144 get merged first, then rebase this one on top of it, thanks.

ts-sdk/src/coordinator/manifest.ts
```diff
@@ -56,10 +38,6 @@ export interface BundleManifest {
 export function buildBundleManifest(registry: DagRegistry): BundleManifest {
   const dags: BundleManifest["dags"] = {};
   for (const { dagId, tasks } of listRegistryDags(registry)) {
```
jason810496, line 40: We should keep a type guard before object-key coercion. A numeric `dagId` becomes `"123"` in metadata while the registry stays keyed by `123`, so runtime lookup fails.

```suggestion
  for (const { dagId, tasks } of listRegistryDags(registry)) {
    if (typeof dagId !== "string") {
      throw new Error("Dag ID must be a string");
    }
```

ts-sdk/src/cli/pack.ts
```diff
@@ -193,7 +194,7 @@ function isBundleManifest(value: unknown): value is BundleManifest {
 }
 
 function isTaskIdList(value: unknown): value is string[] {
-  return Array.isArray(value) && value.every((item) => typeof item === "string" && item.length > 0);
+  return Array.isArray(value) && value.every((item) => typeof item === "string");
```
jason810496: Let's keep the non-empty check since`tasks: [""]` violates the bundle schema's `minLength: 1`.

```suggestion
  return Array.isArray(value) && value.every((item) => typeof item === "string" && item.length > 0);
```

Andrushika:
Rebased, thanks!

jason810496 (commented):
LGTM overall, thanks.

ts-sdk/src/coordinator/manifest.ts
Andrushika: Good catch! Applied, thanks.

ts-sdk/src/cli/pack.ts
Andrushika: Fixed. Thanks!

jason810496 (approved):

PR #71003, opened 2026-08-03
Reject unknown update_mask fields instead of silently ignoring them

ColtenOuO:
## Summary

`update_mask` lets a PATCH caller say "only apply these fields". Every endpoint implements it by intersecting the mask with the fields present in the body:

```python
fields_to_update = patch_body.model_fields_set.intersection(update_mask)
```

An entry that matches no field simply falls out of the intersection. Nothing notices, so the request becomes a no-op and still answers `200`, echoing back the untouched record. A typo is indistinguishable from success:

```console
$ curl -X PATCH '.../variables/pk?update_mask=value' -d '{"key":"pk","value":"CHANGED",...}'
200   # value -> "CHANGED"

$ curl -X PATCH '.../variables/pk?update_mask=valu' -d '{"key":"pk","value":"CHANGED",...}'
200   # value still "original"

$ curl -X PATCH '.../pools/p?update_mask=slot' -d '{"pool":"p","slots":99,...}'
200   # slots still 1
```

A stray space does the same, since nothing trims the entries (`update_mask=%20value`). A client that checks the status code has no way to learn its patch never landed.

`dags.py` already guards against this (`if update_mask != ["is_paused"]: raise 400`), so rejecting an unusable mask is an established expectation in this API. It was just missing everywhere else that accepts one.

## Before / after

| Request | Before | After |
| --- | --- | --- |
| `?update_mask=value` | `200`, applied | unchanged |
| `?update_mask=valu` | `200`, **not applied** | `400` naming the unknown field |
| `?update_mask=%20value` | `200`, **not applied** | `200`, applied |
| bulk `PATCH` with `"update_mask": ["valu"]` | `200`, **not applied** | `200`, error in that action's `errors` |

The error names both the offender and the accepted fields:

```
Unknown field(s) in update_mask: 'slot'.
Valid fields are: description, include_deferred, name, slots, team_name.
```

## The change

- `update_mask_param_factory(body_type)` in `common/parameters.py` builds a validated `update_mask` query parameter. The single-entity PATCH routes take it through `Depends`: variables, pools, connections, dag runs, and the task-instance patch, `dry_run` and task-group routes. `dags.py` is left alone because it already validates.
- Bulk endpoints carry the mask in the request body, so `BulkService.handle_request` validates it per `update` action. A bad mask is reported in that action's `errors` with `status_code: 400`, and the other actions still run. This matches how bulk already reports a missing entity or an immutable field.

Two decisions worth flagging:

**Field names and validation aliases are accepted, serialization aliases are not.** `ConnectionBody.schema_` is aliased to `schema`, and `update_mask=["schema","extra"]` is a working request today, so input aliases must pass. A serialization-only alias is an output name the body can never carry. Accepting it would let `update_mask=val` on a variable (`VariableBody.value` serializes as `val`) through validation as a silent no-op.

**Whitespace is trimmed rather than rejected.** `pools.py` already compares with `mask.strip()`, and the intent of `" value"` is not in doubt.

## Note on an existing test

`test_patch_dag_run_with_update_mask` had a case asserting that `{"update_mask": ["random"]}` returns `200`, which recorded the buggy behaviour as an expectation. It now expects `400` and the error message.

## Tests

Each route wires the dependency on its own line, so every endpoint that takes a mask has a `400` case:

- variables: a typo, a name that exists nowhere, an empty string and the serialization alias `val` return `400` and leave the row unchanged; a mask with surrounding whitespace still applies
- pools, connections: `400` and the stored value is unchanged
- task instances: patch, `dry_run` and task-group routes return `400` without setting any state
- dag runs: the corrected case above
- bulk variables: `200`, with the error in `update.errors` and the row unchanged

`test_variables`, `test_pools`, `test_connections`, `TestPatchDagRun` and the new task-instance test: 361 passed.

---

### DIFF
--- a/airflow-core/newsfragments/71003.significant.rst
+++ b/airflow-core/newsfragments/71003.significant.rst
@@ -1,0 +1,31 @@
+PATCH endpoints reject unknown ``update_mask`` fields
+
+An ``update_mask`` entry that names no field of the request body used to be dropped silently, so a
+typo turned the whole request into a no-op that still answered ``200`` with the untouched record.
+Such a mask is now rejected with ``400``, naming the offending entry and the accepted field names.
+
+Affected endpoints:
+
+* ``PATCH /variables/{variable_key}`` and ``PATCH /variables``
+* ``PATCH /pools/{pool_name}`` and ``PATCH /pools``
+* ``PATCH /connections/{connection_id}`` and ``PATCH /connections``
+* ``PATCH /dags/{dag_id}/dagRuns/{dag_run_id}``
+* ``PATCH /dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances/{task_id}`` and its ``dry_run`` and
+  task-group variants
+
+Surrounding whitespace in a mask entry is now stripped, so ``update_mask=%20value`` applies the
+field instead of silently applying nothing.
+
+Clients that send an unused or misspelled mask entry and rely on the request succeeding must fix
+the entry.
+
+* Types of change
+
+  * [ ] Dag changes
+  * [ ] Config changes
+  * [x] API changes
+  * [ ] CLI changes
+  * [x] Behaviour changes
+  * [ ] Plugin changes
+  * [ ] Dependency changes
+  * [ ] Code interface changes
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -2056,3 +2056,56 @@ QueryParseImportErrorBundleNameFilter = Annotated[
         )
     ),
 ]
+
+
+# Update mask
+def validate_update_mask(patch_body_type: type[BaseModel], update_mask: list[str] | None) -> list[str] | None:
+    """
+    Reject ``update_mask`` entries that name no field of ``patch_body_type``.
+
+    Every caller narrows the patch down with ``set(update_mask)``, which drops an entry matching
+    nothing -- so a typo used to make the whole request a no-op that still answered ``200``.
+    Aliases count as known names because that is what a caller sends in the body and reads back in
+    the response; which of the two a given endpoint acts on is left untouched here. Surrounding
+    whitespace is stripped so a stray space selects the field instead of silently selecting nothing.
+
+    Routes take the mask from the query string through :func:`update_mask_param_factory`; this is
+    for the bulk endpoints, which carry it in the request body instead.
+
+    :param patch_body_type: the request body type the mask selects fields from.
+    :param update_mask: the requested field names, or ``None``.
+    :return: the mask with whitespace stripped, or ``None``.
+    :raises HTTPException: 400 if any entry names no field.
+    """
+    if not update_mask:
+        return update_mask
+
+    fields = patch_body_type.model_fields
+    known = set(fields) | {
+        alias
+        for field in fields.values()
+        for alias in (field.alias, field.serialization_alias, field.validation_alias)
+        if isinstance(alias, str)
+    }
+
+    stripped = [entry.strip() for entry in update_mask]
+    if unknown := {entry for entry in stripped if entry not in known}:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=(
+                f"Unknown field(s) in update_mask: {', '.join(sorted(map(repr, unknown)))}. "
+                f"Valid fields are: {', '.join(sorted(known))}."
+            ),
+        )
+    return stripped
+
+
+def update_mask_param_factory(
+    patch_body_type: type[BaseModel],
+) -> Callable[[list[str] | None], list[str] | None]:
+    """Build the validated ``update_mask`` query parameter for a PATCH endpoint on ``patch_body_type``."""
+
+    def depends_update_mask(update_mask: list[str] | None = Query(None)) -> list[str] | None:
+        return validate_update_mask(patch_body_type, update_mask)
+
+    return depends_update_mask
--- a/airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
@@ -18,13 +18,14 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Generic
+from typing import Generic, cast
 
 from fastapi import HTTPException, status
 from pydantic import BaseModel
 from sqlalchemy.ext.declarative import DeclarativeMeta
 from sqlalchemy.orm import Session
 
+from airflow.api_fastapi.common.parameters import validate_update_mask
 from airflow.api_fastapi.core_api.datamodels.common import (
     BulkAction,
     BulkActionResponse,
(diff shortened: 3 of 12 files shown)


ColtenOuO:
Resolve the merge conflict and re-run the CI.

airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
```diff
@@ -37,6 +37,44 @@
 )
 
 
+def validate_update_mask(patch_body: BaseModel, update_mask: list[str] | None) -> list[str] | None:
+    """
+    Reject ``update_mask`` entries that name no field of ``patch_body``.
+
+    Every caller narrows the patch down with ``set(update_mask)``, which drops an entry matching
```
pierrejeambrun: This should be part of the query param validator. We should extract a custom query param for `updatemask` in `parameters.py` which will do the validation, without having to call validate everywhere.

pierrejeambrun (commented):
Thanks for the PR.


Just one small refactoring suggestion

pierrejeambrun:
Also worth a small `significant` newsfragment, to warn users that these endpoints will now 'error' on wrong payload instead of silently '200'.

airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
ColtenOuO: Thanks for the review! That makes total sense to me.

Sorry, I'm still gaining experience with large-scale projects and have a lot of nuances to learn.

I've addressed the feedback in the latest commit.

ColtenOuO:
addressed feedback and added newsfragment.

airflow-core/src/airflow/api_fastapi/common/parameters.py
```diff
@@ -2056,3 +2056,56 @@ def _optional_boolean(value: bool | None) -> bool | None:
...
+    :raises HTTPException: 400 if any entry names no field.
+    """
+    if not update_mask:
+        return update_mask
+
+    fields = patch_body_type.model_fields
+    known = set(fields) | {
+        alias
+        for field in fields.values()
+        for alias in (field.alias, field.serialization_alias, field.validation_alias)
```
Andrushika: I think we should exclude `serialization_alias` here, it is too wide. It is the output name, not a name the body can send. 

For example:`PATCH /variables/k?update_mask=val` passes this check (`VariableBody.value` has `serialization_alias="val"`), but `model_fields_set` only has `value`, so it is a 200 no-op again.

`validation_alias` alone should be enough, since `alias` fills it too.

airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
```diff
@@ -55,6 +56,10 @@ def handle_request(self) -> BulkResponse:
             if action.action == BulkAction.CREATE:
                 self.handle_bulk_create(action, results[action.action.value])
             elif action.action == BulkAction.UPDATE:
+                if action.entities:
+                    action.update_mask = validate_update_mask(
+                        cast("type[BaseModel]", type(action.entities[0])), action.update_mask
+                    )
```
Andrushika, line 62: Currently a bad mask fails the whole bulk request with 400, but we expect each action to report its own errors in `results.errors`, like the immutable field check in `apply_patch_with_update_mask` ends up doing.

Would it make sense to move this into the per-action try block, so bulk clients see one shape?

airflow-core/src/airflow/api_fastapi/common/parameters.py
Andrushika, line 2087: I think we should exclude `serialization_alias` here, it is too wide. It is the output name, not a name the body can send.

For example: `PATCH /variables/k?update_mask=val` passes this check (`VariableBody.value` has `serialization_alias="val"`), but `model_fields_set` only has `value`, so it is a 200 no-op again.

`validation_alias` alone should be enough, since `alias` fills it too.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_variables.py
```diff
@@ -633,6 +633,33 @@ def test_patch_with_update_mask_description_only(self, test_client, session):
...
+    def test_patch_update_mask_field_tolerates_surrounding_whitespace(self, test_client, session):
+        self.create_variables()
+        response = test_client.patch(
+            f"/variables/{TEST_VARIABLE_KEY}",
+            json={"key": TEST_VARIABLE_KEY, "value": "new_value", "description": "unchanged"},
+            params={"update_mask": [" value "]},
+        )
+        assert response.status_code == 200
+        assert response.json()["value"] == "new_value"
+
```
Andrushika, line 662: Only `variables` and `dag_run` have a 400 case now, and each route wires the dependency on its own line, so dropping one is not caught. Let's add test coverage for each endpoint to lock the wiring.

Andrushika (changes requested):
Thanks for the PR! the direction makes sense to me.

Left three comments: one on the alias set, one on the bulk error shape, one on test coverage.

airflow-core/src/airflow/api_fastapi/core_api/services/public/common.py
ColtenOuO: Yeah, it make sense to me, thanks!

airflow-core/newsfragments/71003.significant.rst
```diff
@@ -0,0 +1,33 @@
...
+Such a mask is now rejected with ``400``, naming the offending entry and the accepted field names.
+Bulk ``PATCH`` endpoints still answer ``200`` and report the rejection in that ``update`` action's
+``errors``, the same way they report other per-action failures.
+
+Affected endpoints:
+
+* ``PATCH /variables/{variable_key}`` and ``PATCH /variables``
+* ``PATCH /pools/{pool_name}`` and ``PATCH /pools``
+* ``PATCH /connections/{connection_id}`` and ``PATCH /connections``
+* ``PATCH /dags/{dag_id}/dagRuns/{dag_run_id}``
```
Andrushika, line 14: The check lives in `BulkService.handle_request`, so the two bulk endpoints that share it are affected too.
For old bulk dagRuns, the handler never reads `update_mask`. So a mask that was ignored before now gets a 400 in that action's `errors`.

For example: `PATCH /dags/{dag_id}/dagRuns` with `"update_mask": ["stat"]` used to answer 200 with no error entry.

```suggestion
* ``PATCH /dags/{dag_id}/dagRuns/{dag_run_id}`` and ``PATCH /dags/{dag_id}/dagRuns`` (the bulk endpoint never applied ``update_mask``, but an unknown entry is now reported instead of ignored)
* ``PATCH /dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances`` (bulk)
```

Andrushika (commented):
LGTM after fix, thanks! 
We should also mention some more affected endpoints in newfragment.

airflow-core/newsfragments/71003.significant.rst
ColtenOuO: Thanks for the suggestion! Applied this in [cec46f5](https://github.com/apache/airflow/commit/cec46f52f627164a79014899c8ea6b286b100034).

Andrushika (approved):
Thanks!!

PR #71009, opened 2026-08-03
Relax types-paramiko to support version 5

eladkal:
Now that we support paramiko version 5 in ssh/sftp we can relax the types-paramiko limit


kaxil (approved):

potiuk:
What a miss :)

PR #67571, opened 2026-05-26
Fix GET /auth/login missing 400 in OpenAPI spec and use status constant

ColtenOuO:
The `/auth/login` endpoint raises HTTP 400 when the `next` redirect URL is invalid or unsafe, but this response code was missing from the `responses` decorator — so the generated OpenAPI spec didn't document it.

## Changes:
- Add `HTTP_400_BAD_REQUEST` to `create_openapi_http_exception_doc` for the login endpoint
- Replace raw integer `400` with `status.HTTP_400_BAD_REQUEST` for consistency with the rest of the codebase

PR #67570, opened 2026-05-26
Fix GET /pools list endpoint incorrectly documenting 404 in OpenAPI spec

ColtenOuO:
The `GET /pools` list endpoint had `HTTP_404_NOT_FOUND` in its OpenAPI `responses` decorator, but the endpoint can never return 404 — it always returns a collection (possibly empty).

This makes the OpenAPI spec accurately reflect the endpoint's actual behaviour and avoids misleading auto-generated SDK clients into handling a 404 that will never occur.

PR #71011, opened 2026-08-03
Document HTTP statuses that API routes raise but never declared

ColtenOuO:
Follow-up to #67570, #67571 and #70992, which each fixed one instance of the same
drift: a handler raises an HTTP status that `create_openapi_http_exception_doc(...)`
never declares, so the generated spec — and every client built from it — has no model
for a response the API really returns.

| Endpoint | Undeclared | Reached when |
| --- | --- | --- |
| `POST /backfills/dry_run` | `400` | the Dag's `allowed_run_types` excludes `backfill` |
| `DELETE /dags/{dag_id}/dagRuns/{dag_run_id}` | `409` | the run's state is not in `DagRunMutableStates` |
| `GET /dags/{dag_id}/dagRuns` | `400` | a `partition_date` filter with `dag_id=~`, or against a non-partitioned Dag |
| `DELETE /dags/{dag_id}` | `409` | task instances are still running |
| `PATCH /hitlDetails/{dag_id}/{dag_run_id}/{task_id}` | `400` | `chosen_options` outside `options`, or several sent to a non-`multiple` detail |
| `POST /dags/{dag_id}/clearTaskInstances` | `400` | `past`/`future` with a run that has no logical date |
| `GET /ui/dags/{dag_id}/latest_run` | `400` | `dag_id=~` |
| `GET /ui/dependencies` | `400` | `dependency_type=data` with a missing or non-numeric `asset:` node id |

Each is driven by request input or by persisted state, so each is really returned today.

`ui/dependencies.py` also had one bare `raise HTTPException(404, ...)`; it now uses
`status.HTTP_404_NOT_FOUND`, the same tidy-up as #67571.

`xcom.py` is deliberately left out — it is fixed in #70992.

## Behaviour

Otherwise unchanged: only `responses=` declarations, so the regenerated spec and UI
client are additive (6 additions to the public spec, 2 to `_private_ui.yaml`, no
deletions). No new tests — the statuses were already produced and are already covered.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
+++ b/airflow-core/src/airflow/api_fastapi/core_api/openapi/_private_ui.yaml
@@ -833,6 +833,12 @@ paths:
                 - $ref: '#/components/schemas/DAGRunLightResponse'
                 - type: 'null'
                 title: Response Get Latest Run Info
+        '400':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPExceptionResponse'
+          description: Bad Request
         '404':
           content:
             application/json:
@@ -916,6 +922,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/BaseGraphResponse'
+        '400':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPExceptionResponse'
+          description: Bad Request
         '404':
           content:
             application/json:
--- a/airflow-core/src/airflow/api_fastapi/core_api/openapi/v2-rest-api-generated.yaml
+++ b/airflow-core/src/airflow/api_fastapi/core_api/openapi/v2-rest-api-generated.yaml
@@ -1475,6 +1475,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPExceptionResponse'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPExceptionResponse'
         '404':
           description: Not Found
           content:
@@ -2121,6 +2127,12 @@ paths:
               schema:
                 $ref: '#/components/schemas/HTTPExceptionResponse'
           description: Not Found
+        '409':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPExceptionResponse'
+          description: Conflict
         '422':
           description: Validation Error
           content:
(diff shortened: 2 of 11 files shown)


bbovenzi:
What does 404 on a POST variable endpoint mean?

ColtenOuO:
Sorry about that.

It doesn't mean anything on a create endpoint.

I found these with a script that compares the statuses each handler raises against the ones its `create_openapi_http_exception_doc(...)` declares, so the list is everything that had drifted. What I didn't check was whether each raise can actually fire, and this one can't.

 I've removed the branch. If the read-back ever did come back empty, response validation now gives a 500, which is the right answer for a write that silently didn't happen.

The description is updated too: that row is gone, and the table now says when each of the remaining statuses is returned.

Thanks for your review!

airflow-core/src/airflow/api_fastapi/core_api/routes/public/variables.py
```diff
@@ -174,14 +174,7 @@ def post_variable(
...
 
-    variable = session.scalar(select(Variable).where(Variable.key == post_body.key).limit(1))
-    if variable is None:
-        raise HTTPException(
-            status.HTTP_404_NOT_FOUND,
-            f"Variable with key: `{post_body.key}` was not found",
-        )
-
-    return variable
+    return session.scalar(select(Variable).where(Variable.key == post_body.key).limit(1))
```
jason810496: We shouldn't change the route implementation.

jason810496 (commented):
Nice! Thanks for adding the missing status code.

> I found these with a script that compares the statuses each handler raises against the ones its create_openapi_http_exception_doc(...) declares, so the list is everything that had drifted.

I prefer to introduce the script you mentioned as pre-commit (prek) hook static check (please take the `dev/script` as referece) that ensure all the status code that might raise within the route are documented in the router definition level to prevent the further drift from happening again instead of one-time manual fix.

Which means we might need to _walk_ into the module level (e.g. the exception might be raise within the common function that we imported at router level).

airflow-core/src/airflow/api_fastapi/core_api/routes/public/variables.py
jason810496: It's a safe guard introduced in https://github.com/apache/airflow/commit/da7519c2b2f3bb16605e735cfa0adf54b2c2fa3d to fix the Mypy warning.

IMO, It's fine to remove it but better in other PR instead of coupling in this one.

ColtenOuO:
Thanks again!

Reverted the post_variable change.

This PR now only adds `responses=` declarations and doesn't touch any route implementation.

ColtenOuO:
rebase to lastest `upstream/main` and re-run the CI.

pierrejeambrun (approved):
Thanks for the PR, LGTM.

Issue #68868, opened 2026-06-23
When trying to edit and save Airflow variables using the Web UI, the save operation fails with an HTTP 500 Internal Server Error.


mohanakrishnan0592:
### Under which category would you file this issue?

Providers

### Apache Airflow version

3.2.2

## Summary
When trying to edit and save Airflow variables using the Web UI, the save operation fails with an HTTP 500 Internal Server Error. The native CLI works perfectly, indicating a strict mismatch between the frontend UI payload and the Airflow 3 backend API schemas.

## Crucial Edge-Case Behavior (Form Change Detection)
* **Scenario A (Fails):** Click 'Edit' -> make NO changes -> click 'Save'. We get 500 Internal Server Error.
* **Scenario B (Succeeds):** Click 'Edit' -> modify the value -> click 'Save'. variable gets saves successfully

## Root Cause Analysis
1. **Frontend Payload:** The UI transmits `"team_name": ""` in the JSON request body when saving a variable.
2. **Backend Datamodel:** In Airflow 3, the `VariableBody` schema explicitly defines `team_name` as an optional string property. 
3. **The Crash:** When multi-team features are disabled in the backend environment, passing an empty string (`""`) instead of `null` triggers an internal exception. 
4. **Middleware Masking:** This exception is intercepted by the `refresh_token.py` authentication middleware, which fails to return a clean validation response and instead throws an unhandled `RuntimeError: No response returned`, masking the underlying failure as a generic 500 error.

## Steps to Reproduce
1. Navigate to the Airflow Web UI -> Variables list.
2. Click **Edit** on an existing variable which is created before the version upgrade, modify the value, and click **Save**.
3. Inspect the outbound browser network request to verify the payload contains `"team_name": ""`.

## Relevant Logs (API Server)
```text
2026-06-22T15:04:14.575405Z [error    ] Exception in ASGI application
 [uvicorn.error] loc=httptools_impl.py:426
  + Exception Group Traceback (most recent call last):
  |   File "/usr/local/lib/python3.12/site-packages/starlette/_utils.py", line 81, in collapse_excgroups
  |     yield
  |   File "/usr/local/lib/python3.12/site-packages/starlette/middleware/base.py", line 192, in __call__
  |     async with anyio.create_task_group() as task_group:
  |                ^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/usr/local/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 811, in __aexit__
[...]

Issue #66889, opened 2026-05-13
API write endpoints return opaque 500 when the database rejects a payload


1fanwang:
Reporting this from the LinkedIn DI Airflow side. The trigger pattern is users inlining very large arguments (whole dataset configs, serialised job parameters, payloads that should live in XCom or external storage) into the Dag run `conf` when launching Spark / Hadoop jobs. The API returns an opaque 500, the SQL error only shows up in the api-server logs as `(1406, "Data too long for column 'conf' at row 1")`, and there's no signal to the caller that payload size was the cause. Same class repeats on every length-capped column across the API surface — this issue captures the broader pattern.

### What happened

Context: filing this from the LinkedIn Data Infrastructure Airflow / Orchestrator team. We see this in production from users who trigger Spark and Hadoop jobs by inlining very large arguments into the Dag run `conf` (whole dataset configs, serialised job parameters, payloads that should live in XCom or external storage). The API returns an opaque 500, users retry with the same args, the SQL error only surfaces in the API server logs as `(1406, "Data too long for column 'conf' at row 1")`, and there's no signal to the caller that payload size was the cause. The same class repeats on every length-capped column across the API surface — this issue captures the broader class.

Across most public REST API and execution API write endpoints, a payload that exceeds the database column's declared type returns a generic `500 Internal Server Error` with no actionable signal that payload size was the cause. The underlying `sqlalchemy.exc.DataError` (MySQL `1406 Data too long`, MySQL `1264 Out of range`, Postgres `value too long for type ...`, Postgres `numeric field overflow`) is wrapped into the generic 500 handler instead of being translated into a 4xx with the column name + remediation. The same failure shape repeats on every length-capped column in the metadata DB.

#66779 reported this for `DagRun.conf` specifically; this issue captures the broader class so it can be closed by a small set of related PRs instead of one boundary patch per column.

[...]

PR #49844, opened 2025-04-27
Allow non-string valid JSON values in Variable import.

tirkarthi:
Airflow 3.0 had the change to accept only string values as variable values during import through API that prohibited using int, list, etc that are valid JSON values. Airflow 3 cli accepts import of valid JSON without this restriction. The API should also accept valid JSON values.

Closes #49837

PR #71018, opened 2026-08-03
Fix Variables API handling of non-string JSON values

### LINKED ISSUE #71010: Variables API: non-string `value` returns masked 500 on PATCH and silently corrupts data on POST

Airflow Core

### Apache Airflow version

3.2.1 (code path still present on `main` as of this report)

## Summary

`VariableBody.value` is typed as `JsonValue`, so the Variables REST API accepts **any** JSON type (array, object, number, bool). Everything downstream, however, assumes a `str`. This produces two distinct failures depending on the verb:

| Request | Result |
|---|---|
| `POST /api/v2/variables` with `value` as a JSON array | **201** — silently stores the Python `repr`, i.e. `['a', 'b']`, which is **not valid JSON** |
| `PATCH /api/v2/variables/{key}` with `value` as a JSON array | **500 Internal Server Error**, real exception masked |

The `POST` case is the more dangerous of the two: it succeeds, so nobody notices until a DAG calls `Variable.get(key, deserialize_json=True)` and blows up on a value that can no longer be parsed.

Note the inconsistency: a non-string **`description`** is correctly rejected with a clean `422`. Only `value` misbehaves.

## Reproduction

```bash
# 1) POST — silently corrupts (HTTP 201)
curl -X POST "$HOST/api/v2/variables" \
  -H 'Content-Type: application/json' -H "Authorization: Bearer $TOKEN" \
  -d '{"key":"my_var","value":["a","b"],"description":"d"}'
# -> 201 {"key":"my_var","value":"['a', 'b']", ...}
#    json.loads("['a', 'b']") raises JSONDecodeError

# 2) PATCH — 500
curl -X PATCH "$HOST/api/v2/variables/my_var" \
  -H 'Content-Type: application/json' -H "Authorization: Bearer $TOKEN" \
  -d '{"key":"my_var","value":["a","b"],"description":"d"}'
# -> 500 Internal Server Error
```

For contrast, both verbs behave correctly when `value` is a JSON *string* (`"[\"a\", \"b\"]"`) — `200`/`201` and the stored value round-trips as valid JSON.

## Root cause

`airflow-core/src/airflow/api_fastapi/core_api/datamodels/variables.py`:

```python
class VariableBody(StrictBaseModel):
    key: str = Field(max_length=ID_LEN)
    value: JsonValue = Field(serialization_alias="val")   # accepts any JSON type
    description: str | None = Field(default=None)
    team_name: str | None = Field(max_length=50, default=None)
```

The two verbs then diverge:

**PATCH** — `update_orm_from_pydantic()` applies the patch straight onto the ORM instance via `BulkService.apply_patch_with_update_mask()`. `setattr` hits the `Variable.val` setter in `airflow-core/src/airflow/models/variable.py`:

```python
@val.setter
def set_val(self, value):
    if value is not None:
        ...
        self._val = fernet.encrypt(bytes(value, "utf-8")).decode()
```

`bytes(<list>, "utf-8")` raises. Verified in-process:

```python
>>> from airflow.models import Variable
>>> v = Variable(key='k'); v.val = ['a', 'b']
  File "airflow/models/variable.py", line 103, in set_val
    self._val = fernet.encrypt(bytes(value, "utf-8")).decode()
TypeError: encoding without a string argument
```

**POST** — goes through `Variable.set()`, which coerces instead of validating:

```python
if serialize_json:
    stored_value = json.dumps(value, indent=2)
else:
    stored_value = str(value)   # list -> "['a', 'b']"
```

`str()` on a list yields Python `repr` with single quotes, so the stored value is not valid JSON and cannot be recovered by `deserialize_json=True`.

## Middleware masks the real error

The `TypeError` never reaches the logs. `JWTRefreshMiddleware` (a Starlette `BaseHTTPMiddleware`) surfaces only:

```
File "airflow/api_fastapi/auth/middlewares/refresh_token.py", line 61, in dispatch
    response = await call_next(request)
File "starlette/middleware/base.py", line 169, in call_next
    raise RuntimeError("No response returned.")
RuntimeError: No response returned.
```

This made the failure substantially harder to diagnose — the traceback names neither the field nor the underlying `TypeError`. Same masking behaviour as reported in #68868 (different trigger, `team_name: ""`) and #66889.

## Suggested fix

Tighten the schema so the API rejects non-string values with a `422`, consistent with `description`:

```python
value: str | None = Field(serialization_alias="val", default=None)
```

If `JsonValue` must be kept for backwards compatibility, add a `field_validator` that JSON-serialises non-string input (`json.dumps`) rather than letting `str()` produce a Python `repr`, and guard `Variable.set_val` so it raises a validation error rather than a bare `TypeError`.

Separately, it would be worth making `JWTRefreshMiddleware` propagate the underlying exception so these failures are diagnosable from the API server logs.

### Operating System

Debian GNU/Linux 12 (official `apache/airflow` base image, Python 3.11)

apache-airflow-providers-fab (FAB auth manager enabled)

### Deployment

### Deployment details

Airflow 3.2.1, Python 3.11, Kubernetes, Postgres metadata DB, FAB auth manager. Reproduced both directly against the api-server pod (`localhost:8080`) and through the ingress, so it is not proxy-related. Also reproduced with two different users (roles `Admin` and `Op`), so it is not permission-related.

### Anything else?

Occurs every time. Originally hit via the Web UI when saving a variable whose value was a JSON array; the UI issues 4 retries with backoff, so a single click produces 4 x 500 in the api-server access log.

### Code of Conduct

kaxil:
Fixes #71010
Closes https://github.com/apache/airflow/pull/71015

## Problem

`VariableBody.value` is typed `JsonValue`, so the Variables REST API accepts any JSON type, but everything downstream assumed a string:

- `POST /api/v2/variables` with `"value": ["a", "b"]` returned 201 but stored the Python repr `['a', 'b']`. That is not valid JSON, so the variable silently breaks the next `Variable.get(key, deserialize_json=True)`.
- `PATCH /api/v2/variables/{key}` with the same payload failed with a masked 500: the raw list reached the Fernet encryption step, which raised `TypeError: encoding without a string argument`.
- Bulk update failed the same way, and one non-string entity took down the whole request. Bulk create already JSON-encoded dicts and lists, but stored booleans and nulls as `"True"`/`"None"`.

## Solution

`VariableBody` now JSON-encodes non-string values once, at request validation, so POST, PATCH and both bulk actions behave identically: strings are stored verbatim, everything else is stored as JSON (`indent=2`, byte-identical to the existing bulk-create format) and round-trips through `deserialize_json=True`. The now-redundant `serialize_json` special case in bulk create is removed, and `Variable.set_val` raises a clear `TypeError` pointing at `serialize_json=True` instead of the cryptic encoding error.

| Request | Before | After |
|---|---|---|
| POST `"value": ["a", "b"]` | 201, stores `['a', 'b']`, unreadable as JSON | 201, stores `["a", "b"]` |
| PATCH `"value": ["a", "b"]` | 500 | 200 |
| Bulk update with a dict value | Whole request 500 | 200 |
| POST or bulk `"value": true` / `null` | Stores `"True"` / `"None"` | Stores `true` / `null` |
| PATCH `"value": null` | 200 but silently kept the old value | 200, stores `null` |

String values, including JSON passed as a string, are stored byte-for-byte as before, and the OpenAPI schema is unchanged, so generated clients are unaffected.

## Why coerce instead of rejecting with 422

Non-string values are an intentional part of the API contract since #49844: the UI Import Variables flow sends raw parsed JSON ([ImportVariablesForm.tsx#L71](https://github.com/apache/airflow/blob/b316afb44d/airflow-core/src/airflow/ui/src/pages/Variables/ImportVariablesForm.tsx#L71)), the docs describe bulk-uploading variables as a JSON file, and `airflow variables import` applies the identical encode-iff-not-a-string rule ([variable_command.py#L165](https://github.com/apache/airflow/blob/b316afb44d/airflow-core/src/airflow/cli/commands/variable_command.py#L165)). Tightening `value` back to `str` would 422 all of those.

## Notes

- Rows already corrupted on released versions stay as-is; a stored Python repr cannot be safely auto-converted back to JSON.
- Non-finite floats still store as the non-strict-JSON literal `NaN`. Rejecting them in the validator is not viable: the 422 response echoes the NaN input, which starlette's `JSONResponse` cannot serialize, turning the rejection into a 500.
- The middleware issue from the report (`JWTRefreshMiddleware` masking the real traceback as `RuntimeError: No response returned`) is not addressed here; it also affects #68868 and #66889.

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/datamodels/variables.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/datamodels/variables.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import json
 from collections.abc import Iterable
 
-from pydantic import Field, JsonValue, model_validator
+from pydantic import Field, JsonValue, field_validator, model_validator
 
 from airflow._shared.secrets_masker import redact
 from airflow.api_fastapi.core_api.base import BaseModel, StrictBaseModel, make_partial_model
@@ -60,6 +60,19 @@ class VariableBody(StrictBaseModel):
     value: JsonValue = Field(serialization_alias="val")
     description: str | None = Field(default=None)
     team_name: str | None = Field(max_length=50, default=None)
+
+    @field_validator("value", mode="after")
+    @classmethod
+    def serialize_non_string_value(cls, value: JsonValue) -> str:
+        # Variables are stored as strings. A non-string JSON value (list, dict, number, bool,
+        # null) must be JSON-encoded here: POST would otherwise store its Python repr, which
+        # deserialize_json cannot read back, and PATCH would fail in the ORM Fernet step on
+        # bytes(value, "utf-8") (see #71010). indent=2 matches Variable.set(serialize_json=True).
+        # Not allow_nan=False: rejecting here yields a 422 whose response embeds the NaN input,
+        # which starlette's JSONResponse cannot serialize -- the request would 500 instead.
+        if isinstance(value, str):
+            return value
+        return json.dumps(value, indent=2)
 
     @model_validator(mode="after")
     def validate_team_name(self) -> VariableBody:
--- a/airflow-core/src/airflow/api_fastapi/core_api/services/public/variables.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/services/public/variables.py
@@ -126,13 +126,12 @@ class BulkVariableService(BulkService[VariableBody]):
 
             for variable in action.entities:
                 if variable.key in create_keys:
-                    should_serialize_json = isinstance(variable.value, (dict, list))
+                    # VariableBody already JSON-encodes non-string values, so no serialize_json here.
                     Variable.set(
                         key=variable.key,
                         value=variable.value,
                         description=variable.description,
                         session=self.session,
-                        serialize_json=should_serialize_json,
                     )
                     results.success.append(variable.key)
 
--- a/airflow-core/src/airflow/models/variable.py
+++ b/airflow-core/src/airflow/models/variable.py
@@ -111,6 +111,11 @@ class Variable(Base, LoggingMixin):
     def set_val(self, value):
         """Encode the specified value with Fernet Key and store it in Variables Table."""
         if value is not None:
+            if not isinstance(value, str):
+                raise TypeError(
+                    f"Variable value must be a string, got {type(value).__name__}. "
+                    "Use Variable.set(key, value, serialize_json=True) to store non-string values."
+                )
             fernet = get_fernet()
             self._val = fernet.encrypt(bytes(value, "utf-8")).decode()
             self.is_encrypted = fernet.is_encrypted
@@ -214,7 +219,9 @@ class Variable(Base, LoggingMixin):
         This operation overwrites an existing variable using the session's dialect-specific upsert operation.
 
         :param key: Variable Key
-        :param value: Value to set for the Variable
+        :param value: Value to set for the Variable. Non-string values are coerced with ``str()``
+            unless ``serialize_json=True``, so pass ``serialize_json=True`` to store them as JSON
+            that ``deserialize_json=True`` can read back.
         :param description: Description of the Variable
         :param serialize_json: Serialize the value to a JSON string
         :param team_name: Team name associated to the variable (if any)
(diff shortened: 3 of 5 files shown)


jason810496 (approved):
Thanks for the fix.

I wonder why not just set `serialize_json=True` to solve the issue in the first place. Then found out that the patch and bulk operation endpoint will interact with the ORM attribute so the `serialize_json=True` doesn't fit. Fixing at data model layer (current implementation) is better.

No other findings, LGTM.

pierrejeambrun (commented):
Nice, thanks!

Probably needs a backport btw.

jason810496:
> Nice, thanks!
> 
> Probably needs a backport btw.

Manually backport in https://github.com/apache/airflow/pull/71526.

PR #71045, opened 2026-08-04
UI: Fix preventRunningTasks showing as raw i18n key in Clear Task dialogs

### LINKED ISSUE #71007: `runAndTaskActions.options.preventRunningTasks` shows when clearing Tasks rather than proper display text

Airflow Core

### Apache Airflow version

3.2.2

When attempting to clear the Tasks in a DAG, `runAndTaskActions.options.preventRunningTasks` is being displayed at the bottom, rather than "Prevent rerun if task is running".

Image

Proper text should be displayed, which is "Prevent rerun if task is running".

### Operating System

Astronomer

### Deployment

Astronomer

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

krystianpl26:
Users on some non-English locales (Italian, Thai) saw the raw i18n key
`runAndTaskActions.options.preventRunningTasks` instead of translated text
in the Clear Task Instance / Bulk Clear dialogs, because those two locale
files were missing the key entirely.

Separately, an English-locale user reported the same raw-key symptom while
every other label in the same dialog rendered correctly translated. That
pointed at a second, more subtle bug: `i18n/config.ts` cache-busts the
translation loadPath with the running Airflow version, but silently fell
back to an un-versioned request whenever the version lookup failed. Behind
a caching CDN or browser cache (e.g. a managed deployment like Astronomer),
that can serve a translation bundle cached from before a key was added —
explaining why older keys translate fine while a newer one doesn't. The
fallback now uses a timestamp so the cache is always busted.

closes: #71007

---

- [x] Yes (Claude Code)


airflow-core/src/airflow/ui/public/i18n/locales/it/dags.json
```diff
@@ -61,6 +61,7 @@
       "future": "Futuro",
       "onlyFailed": "Pulisci solo le task fallite",
       "past": "Passato",
+      "preventRunningTasks": "Impedire la riesecuzione se la task è in esecuzione",
```
bbovenzi:
```suggestion
      "preventRunningTasks": "Impedisci la riesecuzione se la task è in esecuzione",
```

bbovenzi (approved):

PR #71046, opened 2026-08-04
Render multi-branch review choices as a multi-select

guan404ming:
## Why

related: https://github.com/apache/airflow/pull/70651#discussion_r3707023594

- With `allow_multiple_branches=True`, reviewers edit branch choices as free-text JSON; a typo fails only at resume, wasting the approval and an LLM call.

## How

- Publish choices as `{"type": "array", "enum": [...]}` so the review form renders a multi-select.
- Let the approval mixin pass the list as the param value and accept a list back, re-serialized to JSON.

## E2E

Verified against a live Airflow (breeze, pydantic-ai TestModel): the review form renders the multi-select pre-filled with the LLM's choice (`billing`), the reviewer adds `infra`, and after Approve the run finishes with `billing`/`infra` success and `auth` skipped.

e2e_llm_branch_multiselect

---

### DIFF
--- a/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import json
 import logging
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any, Protocol
@@ -109,7 +110,9 @@ class LLMApprovalMixin:
         :param modification_schema: JSON schema for the editable ``output`` param
             when ``allow_modifications=True``. Defaults to ``{"type": "string"}``.
             Pass e.g. ``{"type": "string", "enum": [...]}`` to render a dropdown
-            of valid values in the review form.
+            of valid values in the review form, or ``{"type": "array", "enum": [...]}``
+            to render a multi-select; a list submitted by the reviewer is returned
+            from ``execute_complete`` re-serialized as a compact JSON string.
         """
         from airflow.providers.standard.triggers.hitl import HITLTrigger
         from airflow.sdk.execution_time.hitl import upsert_hitl_detail
@@ -117,6 +120,7 @@ class LLMApprovalMixin:
 
         self.validate_approval_prompt()
 
+        raw_output = output
         if isinstance(output, BaseModel):
             output = output.model_dump_json()
         elif not isinstance(output, str):
@@ -133,9 +137,17 @@ class LLMApprovalMixin:
 
         hitl_params: dict[str, dict[str, Any]] = {}
         if self.allow_modifications:
+            # The multi-select rendered for an array schema needs the list, not its JSON string
+            param_value: Any = output
+            if (
+                modification_schema is not None
+                and modification_schema.get("type") == "array"
+                and isinstance(raw_output, list)
+            ):
+                param_value = raw_output
             hitl_params = {
                 "output": {
-                    "value": output,
+                    "value": param_value,
                     "description": "Edit the output before approving (optional).",
                     "schema": modification_schema or {"type": "string"},
                 },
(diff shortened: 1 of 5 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -121,7 +121,9 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
                 f"```\nPrompt: {self.prompt}\n\nChosen branch(es): {chosen}\n```"
             )
             modification_schema = (
-                None if self.allow_multiple_branches else {"type": "string", "enum": choices}
+                {"type": "array", "enum": choices}
```
kaxil: `{"type": "array", "enum": choices}` is not valid JSON Schema for a pick list: `enum` at the array level means the whole array must equal one of the entries, so `["task_a"]` fails validation against `enum: ["task_a", "task_b"]`.

That stays invisible on 3.3+ because nothing validates `params_input` there, but the `HITLTrigger` fallback in `defer_for_approval` is still live for < 3.3, and `_handle_response` does `self.params[key] = value`, which runs `Param.resolve` -> `jsonschema.validate`. Running that with real `Param`/`ParamsDict`, every submission fails, including an unmodified one:

```
Invalid input for param output: ['task_a'] is not one of ['task_a', 'task_b']
```

so the reviewer's Approve becomes a `HITLTriggerEventError` and the task fails.

`example_params_ui_tutorial.py` spells out the convention: "If you want to select multiple items from a fixed list JSON schema does not allow to use enum. In this case the type `array` is being used together with `examples` as pick list." `FieldMultiSelect` reads `param.schema.examples ?? param.schema.enum`, so `examples` renders the same widget and validates. To keep a server-side check as well:

```python
{"type": "array", "items": {"type": "string", "enum": choices}, "examples": choices}
```

which renders the multi-select and still rejects a task ID outside the list on the legacy path. The mixin docstring recommending `{"type": "array", "enum": [...]}` needs the same fix. The unit tests patch `HITLTrigger` with `autospec`, so none of them reach the real validation.

providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
```diff
@@ -215,13 +227,17 @@ def execute_complete(self, context: Context, generated_output: str, event: dict[
         # when allow_modifications=False, bypassing the read-only approval flow.
         if getattr(self, "allow_modifications", False) and params_input:
             modified = params_input.get("output")
+            if isinstance(modified, list) and all(isinstance(item, str) for item in modified):
```
kaxil: When the reviewer clears the multi-select, `FieldMultiSelect.handleChange` sets the param value to `null` (`updatedOptions.length ? ... : null`) and `getHITLFormData` forwards every param value, so `params_input` arrives as `{"output": None}`. That falls past both checks here and `execute_complete` returns `generated_output`, so an empty selection silently re-applies the LLM's original branches, and `_parse_reviewed_branches`'s "selects no branches" error is no longer reachable from the form. Worth deciding whether a cleared selection should be rejected rather than read as "no change".

providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
```diff
@@ -215,13 +227,17 @@ def execute_complete(self, context: Context, generated_output: str, event: dict[
...
+                # Compact so an unchanged selection compares equal to generated_output
+                modified = json.dumps(modified, separators=(",", ":"))
             if modified is not None and not isinstance(modified, str):
                 # On the awaiting_input path nothing upstream schema-validates params_input
                 # (HITLTrigger did on the legacy path), so enforce the string contract here
                 # rather than returning a non-string as the task's output.
                 raise HITLTriggerEventError(
                     {
-                        "error": f"Modified output must be a string, got {type(modified).__name__}.",
+                        "error": f"Modified output must be a string or a list of strings, "
```
kaxil, line 257: `type(modified).__name__` is `list` whenever a list reaches this branch, so a mixed list reads as "Modified output must be a string or a list of strings, got list." Reporting the offending item's type instead would point at the real problem.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Fixed, switched to `{"type": "array", "items": {"type": "string", "enum": [...]}, "examples": [...]}` and updated the mixin docstring. The multiselect test now also validates the emitted schema with a real Param/jsonschema round-trip, so the legacy-path regression is covered.

providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
guan404ming: Make sense to me, a cleared selection now raises instead of silently re-applying the LLM's branches.

providers/common/ai/src/airflow/providers/common/ai/mixins/approval.py
guan404ming: Fixed, the error now reports the offending item's type.

kaxil (approved):

guan404ming:
Thanks!

PR #71051, opened 2026-08-04
Support require_approval in LLMSchemaCompareOperator

guan404ming:
## Why

- `LLMSchemaCompareOperator` was the last LLM operator still rejecting `require_approval=True`, so its comparison results could not be human-gated.

## How

- Drop the rejection; pause via `LLMApprovalMixin` after the comparison, with a review body showing the compatibility verdict, mismatch severity counts, and the full result JSON.
- Force `output_type=SchemaCompareResult` + `serialize_output=True` so the approved resume path returns the same dict shape as `execute()`.

## E2E

Verified against a live Airflow (breeze, pydantic-ai TestModel, two sqlite sources with drifted schemas): trigger, pause in `awaiting_input`, review, approve, task success with dict XCom.

e2e_llm_schema_compare_approval

---

### DIFF
--- a/providers/common/ai/docs/operators/llm_schema_compare.rst
+++ b/providers/common/ai/docs/operators/llm_schema_compare.rst
@@ -114,6 +114,23 @@ The callable may also return a non-empty ``Sequence[UserContent]`` for
 multimodal inputs -- see
 :ref:`@task.agent multimodal prompts <howto/operator:agent-multimodal>`.
 
+Human-in-the-Loop Approval
+--------------------------
+
+Set ``require_approval=True`` to pause the task after the comparison and wait
+for a human reviewer to approve the result before it is returned. The review
+body shows the compatibility verdict, a mismatch severity summary, and the
+full result JSON. Rejecting the review, or letting ``approval_timeout``
+expire, fails the task:
+
+.. exampleinclude:: /../../ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
+    :language: python
+    :start-after: [START howto_operator_llm_schema_compare_approval]
+    :end-before: [END howto_operator_llm_schema_compare_approval]
+
+``approval_timeout``, ``allow_modifications``, and the rest of the approval
+behaviour are inherited from :ref:`LLMOperator <howto/operator:llm>`.
+
 Conditional ETL Based on Schema Compatibility
 ----------------------------------------------
 
@@ -166,6 +183,12 @@ Parameters
   catalog-managed sources.
 - ``context_strategy``: To fetch primary keys, foreign keys, and indexes.``full`` or ``basic``,
   strongly recommended for cross-system comparisons. default is ``full``
+- ``require_approval``: If ``True``, the task pauses after the comparison and
+  waits for human review before returning the result.  Default ``False``.
+- ``approval_timeout``: Maximum time to wait for a review (``timedelta``).  ``None``
+  means wait indefinitely.  Default ``None``.
+- ``allow_modifications``: If ``True``, the reviewer can edit the result JSON
+  before approving.  Default ``False``.
 
 Logging
 -------
--- a/providers/common/ai/docs/operators/llm_schema_compare.rst
+++ b/providers/common/ai/docs/operators/llm_schema_compare.rst
@@ -128,6 +128,10 @@ expire, fails the task:
     :start-after: [START howto_operator_llm_schema_compare_approval]
     :end-before: [END howto_operator_llm_schema_compare_approval]
 
+``require_approval=True`` requires a string prompt: a decorated callable
+returning a ``Sequence[UserContent]`` raises ``TypeError`` before the LLM
+call.
+
 ``approval_timeout``, ``allow_modifications``, and the rest of the approval
 behaviour are inherited from :ref:`LLMOperator <howto/operator:llm>`.
 
--- a/providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
@@ -18,6 +18,8 @@
 
 from __future__ import annotations
 
+from datetime import timedelta
+
 from airflow.providers.common.ai.operators.llm_schema_compare import LLMSchemaCompareOperator
 from airflow.providers.common.compat.sdk import dag, task
 from airflow.providers.common.sql.config import DataSourceConfig
@@ -108,8 +110,6 @@ example_llm_schema_compare_decorator()
 # [START howto_operator_llm_schema_compare_approval]
 @dag(tags=["example"])
 def example_llm_schema_compare_approval():
-    from datetime import timedelta
-
     LLMSchemaCompareOperator(
         task_id="detect_schema_drift_with_approval",
         prompt="Identify schema mismatches that would break data loading between systems",
--- a/providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
@@ -105,6 +105,27 @@ def example_llm_schema_compare_decorator():
 example_llm_schema_compare_decorator()
 
 
+# [START howto_operator_llm_schema_compare_approval]
+@dag(tags=["example"])
+def example_llm_schema_compare_approval():
+    from datetime import timedelta
+
+    LLMSchemaCompareOperator(
+        task_id="detect_schema_drift_with_approval",
+        prompt="Identify schema mismatches that would break data loading between systems",
+        llm_conn_id="pydanticai_default",
+        db_conn_ids=["postgres_source", "snowflake_target"],
+        table_names=["customers"],
+        require_approval=True,
+        approval_timeout=timedelta(hours=1),
+    )
+
+
+# [END howto_operator_llm_schema_compare_approval]
+
+example_llm_schema_compare_approval()
+
+
 # [START howto_operator_llm_schema_compare_conditional]
 @dag(tags=["example"])
 def example_llm_schema_compare_conditional():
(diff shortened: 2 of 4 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
```diff
@@ -126,9 +134,9 @@ def __init__(
         system_prompt: str = DEFAULT_SYSTEM_PROMPT,
         **kwargs: Any,
     ) -> None:
-        kwargs.pop("output_type", None)
-        if kwargs.get("require_approval"):
-            raise ValueError("require_approval=True is not supported by LLMSchemaCompareOperator.")
+        kwargs["output_type"] = SchemaCompareResult
+        # execute() always returns a dict, so the approval resume path must too
+        kwargs["serialize_output"] = True
```
kaxil, line 139: With `allow_modifications=True` this contract doesn't quite hold. `LLMOperator.execute_complete` goes through `rehydrate_pydantic_output`, which returns the raw string unchanged when validation fails, so a reviewer edit that isn't valid `SchemaCompareResult` JSON lands in XCom as a `str`. I ran the cases: `{"compatible": true, "mismatches": []}` (missing `summary`) and `looks fine to me` both come back as strings, while the unmodified approve path correctly gives a dict. The conditional example in the docs does `comparison_result["compatible"]`, which would raise `TypeError: string indices must be integers` on that value.

`LLMSQLOperator.execute_complete` overrides and re-validates whenever `output != generated_output`; the same guard here (fail the task when the edited JSON doesn't parse into `SchemaCompareResult`) would keep the dict promise.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
```diff
@@ -313,8 +324,23 @@ def execute(self, context: Context) -> dict[str, Any]:
...
+                f"{severity_counts[severity]} {severity}"
+                for severity in ("critical", "warning", "info")
+                if severity_counts[severity]
+            )
+            body = (
+                f"Compatible: {output.compatible}"
+                + (f" (mismatches: {summary})" if summary else "")
+                + f"\n\n```\nPrompt: {self.prompt}\n\n{output.model_dump_json(indent=2)}\n```"
+            )
+            self.defer_for_approval(context, output, body=body)  # type: ignore[misc]
```
kaxil, line 341: `defer_for_approval` always raises, so on the approval path the `Schema comparison result` log two lines below never runs, and `execute_complete` doesn't log it either. The result then only exists in the HITL body, never in the task log. `LLMSQLOperator` logs the generated SQL before it defers -- moving the log above this block would keep the two paths consistent.

providers/common/ai/docs/operators/llm_schema_compare.rst
```diff
@@ -114,6 +114,23 @@ The callable may also return a non-empty ``Sequence[UserContent]`` for
...
+body shows the compatibility verdict, a mismatch severity summary, and the
+full result JSON. Rejecting the review, or letting ``approval_timeout``
+expire, fails the task:
+
+.. exampleinclude:: /../../ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
+    :language: python
+    :start-after: [START howto_operator_llm_schema_compare_approval]
+    :end-before: [END howto_operator_llm_schema_compare_approval]
+
+``approval_timeout``, ``allow_modifications``, and the rest of the approval
```
kaxil, line 131: `llm_branch.rst` spells out that `require_approval=True` needs a string prompt, and that a callable returning `Sequence[UserContent]` raises `TypeError` before the LLM call. The same constraint applies here (`test_execute_rejects_sequence_prompt_with_require_approval` covers it), and the paragraph directly above this new section tells readers the callable may return a `Sequence[UserContent]`. Worth carrying the same note over.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
guan404ming: Fixed. execute_complete now raises ValueError when reviewed output isn't valid SchemaCompareResult JSON.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
guan404ming: Thanks, just moved the result log above the defer block, so both paths log it.

providers/common/ai/docs/operators/llm_schema_compare.rst
guan404ming: Make sense to me. Added the same string-prompt note to llm_schema_compare.rst.

providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
```diff
@@ -105,6 +105,27 @@ def check_migration_readiness(ds=None):
 example_llm_schema_compare_decorator()
 
 
+# [START howto_operator_llm_schema_compare_approval]
+@dag(tags=["example"])
+def example_llm_schema_compare_approval():
+    from datetime import timedelta
```
Lee-W: I think we can move it to the top

providers/common/ai/tests/unit/common/ai/operators/test_llm_schema_compare.py
```diff
@@ -16,20 +16,36 @@
...
-from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException, TaskDeferred
 from airflow.providers.common.sql.config import DataSourceConfig
 from airflow.providers.common.sql.datafusion.engine import DataFusionEngine
 from airflow.providers.common.sql.hooks.sql import DbApiHook
 
+from tests_common.test_utils.version_compat import AIRFLOW_V_3_1_PLUS, AIRFLOW_V_3_3_PLUS
+
+if AIRFLOW_V_3_3_PLUS:
+    from airflow.sdk.exceptions import TaskAwaitingInput as ApprovalPauseSignal
```
Lee-W: Why do we need alias here

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
```diff
@@ -313,8 +324,32 @@ def execute(self, context: Context) -> dict[str, Any]:
...
+            severity_counts = Counter(mismatch.severity for mismatch in output.mismatches)
+            summary = ", ".join(
+                f"{severity_counts[severity]} {severity}"
+                for severity in ("critical", "warning", "info")
+                if severity_counts[severity]
+            )
+            body = (
+                f"Compatible: {output.compatible}"
+                + (f" (mismatches: {summary})" if summary else "")
+                + f"\n\n```\nPrompt: {self.prompt}\n\n{output.model_dump_json(indent=2)}\n```"
```
Lee-W, line 342:
```suggestion
                + (
                    "\n\n```\n"
                    f"Prompt: {self.prompt}\n\n"
                    f"{output.model_dump_json(indent=2)}\n'
                    "```"   
                )
```

easier to understand what this if

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
```diff
@@ -313,8 +324,32 @@ def execute(self, context: Context) -> dict[str, Any]:
...
+            )
+            self.defer_for_approval(context, output, body=body)  # type: ignore[misc]
+
         return output_result
+
+    def execute_complete(self, context: Context, generated_output: str, event: dict[str, Any]) -> Any:
+        output = super().execute_complete(context, generated_output, event)
+        if not isinstance(output, dict):
```
Lee-W: we only check ths type here. how do we know whether it's a valid SchemaCompareResult JSON or not?

providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm_schema_compare.py
guan404ming: Done, moved to the top.

providers/common/ai/tests/unit/common/ai/operators/test_llm_schema_compare.py
guan404ming: I think this is version-compat: 3.3+ raises `TaskAwaitingInput`, older cores `TaskDeferred`. There is same pattern in test_llm_sql.py.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
guan404ming: Tried it. Ruff format joins the implicit concatenation back into one line 🫠.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
guan404ming: `rehydrate_pydantic_output` validates via `TypeAdapter(SchemaCompareResult)`. It returns dict on success, raw str otherwise. I think that's not that ideal to rely on the fallback thus I reworked to make it explicitly validates via SchemaCompareResult.model_validate_json, surfacing pydantic error details.

kaxil (approved):

providers/common/ai/src/airflow/providers/common/ai/operators/llm_schema_compare.py
Lee-W: ok then 😭

providers/common/ai/tests/unit/common/ai/operators/test_llm_schema_compare.py
```diff
@@ -16,20 +16,36 @@
...
+else:
+    ApprovalPauseSignal = TaskDeferred  # type: ignore[assignment, misc]
```
Lee-W: What I meant was why don't we just 


```suggestion
if AIRFLOW_V_3_3_PLUS:
    from airflow.sdk.exceptions import TaskAwaitingInput
else:
    TaskAwaitingInput = TaskDeferred  # type: ignore[assignment, misc]
```

guan404ming: Oh I see, make sense to me. Just updated!

Lee-W (approved):

guan404ming:
Thanks for the review!

PR #71054, opened 2026-08-04
Reject serde-reserved keys in HITL params_input

### LINKED ISSUE #71036: HITL responses with serde-reserved keys in params_input cannot be resumed or resubmitted
### What happened

`airflow.sdk.serde.serialize` refuses any dict carrying its reserved keys (`__classname__`, `__schema_id__`) at any depth. The Human-in-the-loop route stores `params_input` verbatim, with no validation of the key names, so a response whose params contain such a key cannot be serialized when the task is resumed.

Since apache/airflow#70685, `handle_event_submit` no longer raises in that situation — it routes the task instance to `__fail__` so a worker fails it normally, instead of aborting the caller. That is the right trade for the triggerer and the scheduler sweep, where a raise previously wedged an entire batch or caused a poison event to be redelivered indefinitely. It is a regression for this one HITL path:

- **Before:** the request 500s, nothing is recorded, and the user can correct the params and resubmit.
- **After:** the response is recorded, the task is routed to `__fail__`, the route answers `200`, and the `response_received` guard turns a corrected resubmission into a `409`.

The user is left unable to retry a response that was effectively discarded, and the `200` gives them no indication that anything went wrong.

### What needs doing

Validate on the write side, next to the existing option checks in `airflow/api_fastapi/core_api/routes/public/hitl.py`: reject `params_input` containing serde-reserved keys with a `400`, so the request never reaches a state that cannot be resumed. Validating what is accepted is preferable to compensating after it has been stored.

Optionally, the route could also detect the discarded case after the fact — `handle_event_submit` leaves `next_method == TRIGGER_FAIL_REPR` on a task instance it could not resume, which is how the scheduler sweep now distinguishes that outcome — and answer with something other than `200`. That changes a public endpoint's response, so it wants its own discussion.

### Acceptance criteria

- A HITL response whose `params_input` contains `__classname__` or `__schema_id__` is rejected with `400` at submission time.
- The task instance stays in `AWAITING_INPUT` and the user can resubmit a corrected response.
- Test coverage for the rejection and for the resubmission remaining possible.

Raised in review of apache/airflow#70685.

---

haseebmalik18:
A HITL response whose `params_input` contains a serde-reserved key (`__classname__` or `__id__`) cannot be serialized when the task resumes. Today the response is silently discarded and the user has no way to resubmit a corrected one.

This validates `params_input` when the response is submitted and returns a 400 if it carries one of those keys, before anything is written. The task stays in `AWAITING_INPUT`, so the user can fix the input and resubmit.

Only the public HITL route needs this. The execution API route builds its `params_input` from trusted operator code rather than user input, so it is left as is.

closes: #71036


amoghrajesh:
This doesn't have to be in 3.3.1, but if we are comfortable (cc @Lee-W) we can consider it.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_hitl.py
```diff
@@ -408,6 +409,75 @@ def test_should_respond_400_for_invalid_option(
         assert response.status_code == 400
         assert "Invalid options" in response.json()["detail"]
 
+    @pytest.mark.usefixtures("sample_hitl_detail")
+    @pytest.mark.parametrize("reserved_key", [CLASSNAME, SCHEMA_ID])
```
Lee-W: it's easy to miss if we add another reserved_key. let's make it a `_variable` in `airflow-core/src/airflow/api_fastapi/core_api/routes/public/hitl.py` so that we can just import it here. also worth mention this change in the doc

Lee-W:
A few nitpicks, but overall this looks good. I also don't think it's worth blocking the 3.3.1 release.

vatsrahul1001:
Moved to 3.3.2

haseebmalik18:
@Lee-W Applied changes left from review

airflow-core/src/airflow/api_fastapi/core_api/routes/public/hitl.py
```diff
@@ -80,6 +82,27 @@
 log = structlog.get_logger(__name__)
 
 
+# Keys that ``serde.serialize`` refuses at any depth. A ``params_input`` carrying one could not be
```
jason810496, line 85: IMO, we should follow the convetion in XCom and validate at the data model layer:

https://github.com/apache/airflow/blob/9a85faa9c55b907bd105ee41567bb7d55dab9f31/airflow-core/src/airflow/api_fastapi/core_api/datamodels/xcom.py#L87-L119

https://github.com/apache/airflow/blob/9a85faa9c55b907bd105ee41567bb7d55dab9f31/shared/serialization/src/airflow_shared/serialization/__init__.py#L20-L44

jason810496 (commented):
Thanks for fixing this.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/hitl.py
haseebmalik18: Good point, thanks! Only thing: XCom's validator raises `ValueError` so it comes back as 422, but the issue asks for 400 here? This route already does its option checks with `HTTPException(400)`, so I matched that. Would you still want it at the datamodel layer even if it becomes 422?

I also kept the keys to just `__classname__`/`__id__` (what serialize actually rejects) instead of the full `FORBIDDEN_XCOM_KEYS`, since the wider set would reject valid things like `{"__version__": 1}`. What do you think?

haseebmalik18: @jason810496 Following up

jason810496: Sorry for the late reply. Yes, it's fine to make it 422 status code as both 400 and 422 are valid the this use case.

jason810496 (commented):
Additionally, we could introduce a common function to handle at both side.

vatsrahul1001:
LGTM! @Lee-W can you review this again?

jason810496 (commented):
I prefer to move this to 3.3.3 milestone since the https://github.com/apache/airflow/pull/71054#discussion_r3726472209 comment hasn't be addressed.

haseebmalik18:
Addressed the comments

airflow-core/docs/tutorial/hitl.rst
```diff
@@ -221,6 +221,12 @@ calls involved (``~`` works as a wildcard for ``dag_id`` and ``dag_run_id``):
     PATCH /api/v2/dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances/{task_id}/{map_index}/hitlDetails
     {"chosen_options": ["Approve"], "params_input": {}}
 
+.. note::
+
+    Keys in ``params_input`` may not be Airflow's reserved serialization keys (``__classname__`` or
```
jason810496:
```suggestion
    Keys in ``params_input`` must not be Airflow's reserved serialization keys (``__classname__`` or
```

jason810496:
```suggestion
    Keys in ``params_input`` must not be Airflow's reserved serialization keys (``__classname__`` or
```

airflow-core/src/airflow/api_fastapi/core_api/datamodels/common.py
```diff
@@ -32,6 +34,56 @@
 
 log = logging.getLogger(__name__)
 
+
+def find_reserved_keys(
+    value: Any,
+    reserved_keys: Collection[str],
```
jason810496: Let's use `FORBIDDEN_XCOM_KEYS` for this function and no need to introduce the `SERDE_RESERVED_DICT_KEYS`. Since `FORBIDDEN_XCOM_KEYS` is wider than the `SERDE_RESERVED_DICT_KEYS`.
```suggestion
```

airflow-core/src/airflow/api_fastapi/core_api/datamodels/common.py
```diff
@@ -32,6 +34,56 @@
...
+        if isinstance(obj, str):
+            if not decode_json_strings:
+                return None
+            try:
+                decoded = json.loads(obj)
+            except (ValueError, TypeError):
+                return None
+            return walk(decoded, path) if isinstance(decoded, (dict, list)) else None
+        if isinstance(obj, Mapping):
+            found = reserved & obj.keys()
```
jason810496:
```suggestion
    def walk(obj: Any, path: str) -> tuple[str, list[str]] | None:
        if isinstance(obj, str):
            if not decode_json_strings:
                return None
            try:
                decoded = json.loads(obj)
            except (ValueError, TypeError):
                return None
            return walk(decoded, path) if isinstance(decoded, (dict, list)) else None
        if isinstance(obj, Mapping):
            found = FORBIDDEN_XCOM_KEYS & obj.keys()
```

shared/serialization/src/airflow_shared/serialization/__init__.py
```diff
@@ -42,6 +42,11 @@
     }
 )
 
+# serde.serialize refuses to serialize a dict that contains either of these keys. This is narrower
+# than FORBIDDEN_XCOM_KEYS above, which also blocks keys that only matter when XCom reads a value
+# back and decodes it into an object.
+SERDE_RESERVED_DICT_KEYS = frozenset({CLASSNAME, SCHEMA_ID})
+
```
jason810496:
```suggestion
```

airflow-core/src/airflow/api_fastapi/core_api/datamodels/hitl.py
```diff
@@ -33,6 +35,21 @@ class UpdateHITLDetailPayload(BaseModel):
     chosen_options: list[str] = Field(min_length=1)
     params_input: Mapping = Field(default_factory=dict)
 
+    @field_validator("params_input")
+    @classmethod
+    def _check_serde_reserved_keys(cls, params_input: Mapping) -> Mapping:
+        # serde.serialize refuses these keys, and it only runs once the task resumes, long after the
+        # response was stored and the request returned. Reject it here instead, while the user can
+        # still correct the input and resubmit.
+        found = find_reserved_keys(params_input, SERDE_RESERVED_DICT_KEYS, root="params_input")
```
jason810496: Then please remember to update this downstream function.

jason810496 (commented):
LGTM once these comments be addressed, thanks.

haseebmalik18:
Addressed review comments

jason810496 (approved):
LGTM now, thanks for the update.

henry3260 (approved):
looks good!!

henry3260:
nit: This docstring also cites the old method (‎`_walk`)

https://github.com/apache/airflow/blob/0bac61574a5b6f49ae930304298650e3c94681e9/airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_xcom.py#L867-L870

haseebmalik18:
Fixed stale reference

PR #71057, opened 2026-08-04
Java SDK: Register tasks as first-class TaskDef objects

jason810496:
- depends on: #69757

> **Merge order:**
> 1. #69757 — Support TaskFlow call syntax on stub tasks for the Lang SDK
> 2. **#71057 — Register tasks as first-class `TaskDef` objects** **(current one)**
> 3. #71188 — Honor TaskFlow arg bindings sent by the supervisor
> 4. #71189 — Author complete Dags in Java without a Python stub file
> 5. #71190 — Serialize native Dags to DagSerialization v3
>
> Every PR targets `main` because GitHub cannot base a pull request on a branch that exists only on a fork, so these diffs are cumulative — the compare link above shows only this layer.

## Why

The existing Java SDK registered tasks as `dag.addTask("extract", Extract.class)`.

To support the further native Dag feature (dependency edges, Dag and task-level configuration), it would be better to Introduce the another abstraction interface that is additive instead of forcing another break of the registration API later before GA.

## How

- A task is now a `TaskDef` (its ID plus the class implementing it), registered with `DagDef.addTask(TaskDef)`.
  - A `TaskDef` belongs to exactly one `DagDef`, so registering the same instance twice or with a second Dag fails at Dag-parse time.
- The annotation surface keeps `Builder.Dag` / `Builder.Task`, and the interface users implement keeps the `Task` name, so the definition objects are `DagDef` and `TaskDef`.
- `BuilderProcessor` emits `dag.addTask(new TaskDef("...", T.class))` from `@Builder.Dag` / `@Builder.Task` classes; the generated per-task inner classes still implement `Task`.

---

### DIFF
--- a/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
+++ b/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
@@ -226,8 +226,8 @@ Register tasks manually in a ``BundleBuilder``:
       @Override
       public Iterable<DagDef> getDags() {
         var dag = new DagDef("my_dag")
-            .addTask(new TaskDef("fetch", FetchTask.class))
-            .addTask(new TaskDef("process", ProcessTask.class));
+            .addTask("fetch", FetchTask.class)
+            .addTask("process", ProcessTask.class);
         return List.of(dag);
       }
     }
--- a/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
+++ b/airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
@@ -115,7 +115,7 @@ Java entry point
 
     public class Main implements BundleBuilder {
       @Override
-      public Iterable<Dag> getDags() {
+      public Iterable<DagDef> getDags() {
         return List.of(SalesPipelineBuilder.build());  // SalesPipelineBuilder generated at compile time
       }
 
@@ -203,7 +203,7 @@ Interface-based API
 ~~~~~~~~~~~~~~~~~~~
 
 Implement the ``Task`` interface directly for full control over how tasks are registered and how XComs are
-read.
+read.  Each task is registered as a ``TaskDef`` on a ``DagDef``.
 
 .. code-block:: java
 
@@ -224,10 +224,10 @@ Register tasks manually in a ``BundleBuilder``:
 
     public class MyBundle implements BundleBuilder {
       @Override
-      public Iterable<Dag> getDags() {
-        var dag = new Dag("my_dag");
-        dag.addTask("fetch", FetchTask.class);
-        dag.addTask("process", ProcessTask.class);
+      public Iterable<DagDef> getDags() {
+        var dag = new DagDef("my_dag")
+            .addTask(new TaskDef("fetch", FetchTask.class))
+            .addTask(new TaskDef("process", ProcessTask.class));
         return List.of(dag);
       }
     }
--- a/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
+++ b/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
@@ -59,8 +59,8 @@ class DagDef(
    *    task with the same ID is already registered.
    */
   fun addTask(task: TaskDef): DagDef {
-    require(task.owner == null) {
-      "Task '${task.id}' already belongs to Dag '${task.owner?.id}'"
+    task.owner?.let { owner ->
+      throw IllegalArgumentException("Task '${task.id}' already belongs to Dag '${owner.id}'")
     }
     require(tasks.putIfAbsent(task.id, task) == null) {
       "Tasks in Dag have duplicate ID: ${task.id}"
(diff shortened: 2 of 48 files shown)


airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
```diff
@@ -82,31 +82,36 @@ Java implementation
 
     import org.apache.airflow.sdk.*;
 
-    @Builder.Dag(id = "sales_pipeline")
+    @Dag(dagId = "sales_pipeline")
```
uranusjr: I prefer to keep the original naming

* The interface uses Dag and Task
* The Builder outer class (since the annotated class/method does not actually produce a dag/task)
* The argument being `id`

kubernetes-tests/lang_sdk/java_example/src/java/org/apache/airflow/k8sexample/K8sBundleBuilder.java
```diff
@@ -26,6 +26,8 @@
 // CombinedExampleBuilder is generated by the annotation processor from the
 // @Builder.Dag class CombinedExample.
 public class K8sBundleBuilder implements BundleBuilder {
+  // Dag, not DagDef: the k8s harness publishes the SDK from upstream main while this example
+  // tracks the branch, so the rename can only be adopted here once it has merged upstream.
   @NotNull
   @Override
   public Iterable<Dag> getDags() {
```
jason810496: We will need https://github.com/apache/airflow/pull/71527 get merged first then we can rebase and change the `Dag` to `DagRef`. I intentionally keep the old syntax here to get k8s test pass, please see more detail in https://github.com/apache/airflow/pull/71527.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
```diff
@@ -24,46 +24,72 @@ import kotlin.Throws
...
+ *     .addTask(new TaskDef("load", Load.class));
+ * ```
  *
  * @param id Dag identifier. Must contain only ASCII alphanumeric characters,
  *    dashes, dots, or underscores; must be unique within a [Bundle].
  *
  * @see Builder.Dag
  */
-class Dag(
+class DagDef(
```
uranusjr, line 44: Since we’re not renaming the annotations, does this still need to be renamed? (Also Task/TaskDef.)

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
```diff
@@ -24,46 +24,72 @@ import kotlin.Throws
...
    */
-  fun addTask(
-    id: String,
-    definition: Class<out Task>,
-  ): Dag {
-    require(tasks.putIfAbsent(id, definition) == null) {
-      "Tasks in Dag have duplicate ID: $id"
+  fun addTask(task: TaskDef): DagDef {
+    require(task.owner == null) {
+      "Task '${task.id}' already belongs to Dag '${task.owner?.id}'"
```
uranusjr:
```suggestion
      "Task '${task.id}' already belongs to Dag '${task.owner!!.id}'"
```

Since the `require` call already established the value is not null.

But personally I would do this instead

```kotlin
task.owner?.let { owner ->
  throw IllegalArgumentException("Task '${task.id}' already belongs to Dag '${owner.id}'")
}
```

to avoid the second unwrap. This is just personal reference.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
jason810496: The interface with the `execute` method is named as `Task` (`public static class Transform implements Task {`), so I name the new object here as `TaskRef`.

If we name the `TaskRef` here as `Task`, then the old `Task` interface should be renamed into something like `TaskFunction, TaskImplementation or TaskCallable` to avoid the ambiguity. 

Both work with me, no strong opinion on the naming.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/DagDef.kt
jason810496: Sure, addressed in https://github.com/apache/airflow/pull/71057/changes/2f821d4574b46963866e2ff63b6af2007fdbf503, thanks.

airflow-core/docs/authoring-and-scheduling/language-sdks/java.rst
```diff
@@ -224,10 +224,10 @@ Register tasks manually in a ``BundleBuilder``:
 
     public class MyBundle implements BundleBuilder {
       @Override
-      public Iterable<Dag> getDags() {
-        var dag = new Dag("my_dag");
-        dag.addTask("fetch", FetchTask.class);
-        dag.addTask("process", ProcessTask.class);
+      public Iterable<DagDef> getDags() {
+        var dag = new DagDef("my_dag")
+            .addTask(new TaskDef("fetch", FetchTask.class))
```
uranusjr: Can we hide this `new TaskDef` part by adding an overload? This can be manually implemented by the user, so egonomics matters.

jason810496: Good point, I added the overload in https://github.com/apache/airflow/pull/71057/commits/9ff93ea45a378ca68cadb6fbabad8f6098fcd91f, thanks for the suggestion.

uranusjr (approved):

PR #71067, opened 2026-08-04
Add Redis cluster mode support to RedisHook

### LINKED ISSUE #10504: Add support for Redis in Cluster Mode
**Description**

There is an existing JIRA issue: https://issues.apache.org/jira/browse/AIRFLOW-3625

There is an existing PR to add this functionality, but it has gone stale and been closed: https://github.com/apache/airflow/pull/4429

**Use case / motivation**

As a user, I'd like to be able to add a connection from Airflow to my Redis database configured in Cluster Mode. 

**Related Issues**

https://issues.apache.org/jira/browse/AIRFLOW-3625

FrankYang0529:
## Why

- Redis Cluster splits the keyspace into 16384 hash slots spread across masters. Any key-based command must reach the node owning that key's slot, otherwise the node answers `MOVED` instead of serving it.
- `RedisHook` only built a standalone `redis.Redis` client, which does not follow those redirects. So every key-based operation in this provider can fail, depending on where the key happens to hash.
  - `RedisTaskHandler` (remote logging) — `lrange` to read, `rpush`/`ltrim`/`expire` to write. Its key is derived from dag id, run id and task id, so each task instance's log targets a different node.
  - `RedisKeySensor` uses `exists` to watch key. The key can be dynamic string, so it can be on different nodes.

## How

- Add a `cluster` connection extra. When set, `get_conn()` builds a `redis.cluster.RedisCluster`, which follows `MOVED`/`ASK` redirects.
- Add a `startup_nodes` extra to supply extra bootstrap nodes.

## Verification

- `uv run --project providers/redis pytest providers/redis/tests/unit/redis/hooks/test_redis.py -xvs`
- `breeze testing providers-integration-tests --db-reset --integration redis`

 

* closes: #10504
---

- [X] Yes - Claude Code

---


providers/redis/docs/connections.rst
```diff
@@ -31,22 +32,30 @@ parameter as ``redis_default`` by default.
...
-    The password of the user that will be used for authentication against the Redis cluster.
+    The password of the user that will be used for authentication against the Redis server.
 
 DB
-    The DB number to use in the Redis cluster (Default is ``0``).
+    The DB number to use in the Redis server (Default is ``0``). Not supported in cluster mode.
+
+Enable cluster mode
+    Whether to connect with a cluster-aware client that follows ``MOVED``/``ASK`` redirects
+    (Default is ``False``). See :ref:`redis-cluster-mode` below.
```
Miretpl:
```suggestion
Is cluster
    Whether Redis deployment is a cluster or a standalone instance (Default is ``False``). See :ref:`redis-cluster-mode` below.
```
I think we should make the description a bit simpler - we have more in-detail explanation in the mode doc. `Is cluster` instead of `Enable cluster mode`, as this setting represents what redis to which we connect to, rather than configuring Redis itself.

providers/redis/docs/connections.rst
```diff
@@ -64,3 +73,26 @@ Certificate path
...
+do this: when it asks a node for a key that node does not serve, the node answers ``MOVED`` and
+the standalone client fails.
+
+Enable cluster mode to use a cluster-aware client that follows those redirects:
+
+.. code-block:: json
+
+    {
+      "cluster": true,
+      "startup_nodes": ["node-2:6379", "node-3:6379"]
```
Miretpl:
```suggestion
      "startup_nodes": ["node-2:6379", "node-3:6379"] # Connection extra
```
It will be easier for user to know where this setting can be set.

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -45,6 +47,11 @@ class RedisHook(BaseHook):
     You can set your db in the extra field of your connection as ``{"db": 3}``.
     Also you can set ssl parameters as:
     ``{"ssl": true, "ssl_cert_reqs": "require", "ssl_certfile": "/path/to/cert.pem", etc}``.
+
+    To talk to a Redis deployment running in cluster mode, set ``{"cluster": true}``. Additional
+    bootstrap nodes may be listed as ``{"startup_nodes": ["node-2:6379", "node-3:6379"]}`` so that
```
Miretpl:
```suggestion
    bootstrap nodes may be listed as Connection Extras ``{"startup_nodes": ["node-2:6379", "node-3:6379"]}`` so that
```

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -141,6 +189,14 @@ def get_connection_form_widgets(cls) -> dict[str, Any]:
 
         return {
             "db": IntegerField(lazy_gettext("DB"), widget=BS3TextFieldWidget(), default=0),
+            "cluster": BooleanField(lazy_gettext("Enable cluster mode"), default=False),
```
Miretpl:
```suggestion
            "cluster": BooleanField(lazy_gettext("Is cluster"), default=False),
```

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -76,6 +84,14 @@ def get_conn(self):
         self.username = conn.login
         self.password = None if str(conn.password).lower() in ["none", "false", ""] else conn.password
         self.db = conn.extra_dejson.get("db")
+        self.cluster = conn.extra_dejson.get("cluster", False)
```
Miretpl, line 89: We could get `startup_nodes` here too and use it directly in `_build_startup_nodes` method.

providers/redis/docs/connections.rst
```diff
@@ -31,22 +32,30 @@ parameter as ``redis_default`` by default.
...
 
 DB
-    The DB number to use in the Redis cluster (Default is ``0``).
+    The DB number to use in the Redis server (Default is ``0``). Not supported in cluster mode.
+
+Enable cluster mode
+    Whether to connect with a cluster-aware client that follows ``MOVED``/``ASK`` redirects
+    (Default is ``False``). See :ref:`redis-cluster-mode` below.
+
+Cluster startup nodes
```
Miretpl:
```suggestion
Startup nodes
```
To be consistent with other places within the code.

providers/redis/src/airflow/providers/redis/get_provider_info.py
```diff
@@ -65,6 +65,15 @@ def get_provider_info():
                 "ui-field-behaviour": {"hidden-fields": ["schema", "extra"], "relabeling": {}},
                 "conn-fields": {
                     "db": {"label": "DB", "schema": {"type": ["integer", "null"], "default": 0}},
+                    "cluster": {
+                        "label": "Enable cluster mode",
```
Miretpl:
```suggestion
                        "label": "Is cluster",
```
Could you add also a description here?

providers/redis/src/airflow/providers/redis/get_provider_info.py
```diff
@@ -65,6 +65,15 @@ def get_provider_info():
...
+                        "schema": {"type": ["boolean", "null"], "default": False},
+                    },
+                    "startup_nodes": {
+                        "label": "Cluster startup nodes",
+                        "description": "Comma-separated extra bootstrap nodes as host:port. Cluster mode only.",
```
Miretpl:
```suggestion
                        "description": "Comma-separated extra bootstrap nodes as host:port. Only for cluster Redis deployments.",
```

Miretpl (changes requested):

Miretpl:
Could you also adjust the PR description in terms of why it is needed? Currently, we only know the difference between standalone Redis and cluster deployment and how it handles connections.

providers/redis/docs/connections.rst
FrankYang0529: Thanks for the great suggestion. Yes, the `Is cluster` is more clear.

providers/redis/docs/connections.rst
FrankYang0529: Change to use `// Connection extra`.

providers/redis/src/airflow/providers/redis/hooks/redis.py
FrankYang0529: Updated it.

FrankYang0529:
@Miretpl Thanks for the review. I addressed all comments and updated PR description.

providers/redis/docs/connections.rst
```diff
@@ -64,3 +73,26 @@ Certificate path
 
 Enable hostname check
     If set, match the hostname during the SSL handshake (Default is ``False``).
+
+.. _redis-cluster-mode:
+
+Cluster mode
```
Miretpl:
```suggestion
Redis Cluster Deployment
```

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -111,18 +130,49 @@ def get_conn(self):
...
+                )
 
         return self.redis
 
+    def _build_startup_nodes(self) -> list[ClusterNode]:
+        """Build redis-py cluster nodes from the ``startup_nodes`` extra, given as ``host`` or ``host:port``."""
+        if not self.startup_nodes:
+            return []
+
+        entries = self.startup_nodes.split(",") if isinstance(self.startup_nodes, str) else self.startup_nodes
```
Miretpl: Maybe instead of assuming what the user will enter, we should verify values and raise an exception if they are not, e.g. a list of strings?

Miretpl (changes requested):

providers/redis/src/airflow/providers/redis/hooks/redis.py
FrankYang0529: Thanks for the review and suggestion. Add value verification.

Miretpl: Is there some reason why we would want to have two value formats supported instead of one? 🤔

FrankYang0529: Sorry for the late reply. After rechecking the design, I think we don't need list type here. We can just use comma-separated string.

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -111,18 +130,54 @@ def get_conn(self):
...
         return self.redis
 
+    def _build_startup_nodes(self) -> list[ClusterNode]:
+        """Build redis-py cluster nodes from the ``startup_nodes`` extra, given as ``host`` or ``host:port``."""
+        if not self.startup_nodes:
+            return []
+
+        if not isinstance(self.startup_nodes, str):
+            raise ValueError(
+                "`startup_nodes` must be a comma-separated string of `host:port` entries, got "
```
Miretpl:
```suggestion
                "The `startup_nodes` parameter value must be a comma-separated string of `host:port` entries, got "
```

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -111,18 +130,54 @@ def get_conn(self):
...
+                f"{self.startup_nodes!r}."
+            )
+
+        nodes = []
+        for entry in self.startup_nodes.split(","):
+            host, _, port = entry.strip().partition(":")
+            if not host:
+                raise ValueError(f"Missing host in `startup_nodes` entry {entry!r}; expected `host:port`.")
```
Miretpl:
```suggestion
                raise ValueError(f"Missing host in `startup_nodes` parameter value for entry {entry!r}; expected `host:port`.")
```

providers/redis/src/airflow/providers/redis/hooks/redis.py
```diff
@@ -111,18 +130,54 @@ def get_conn(self):
...
+            try:
+                parsed_port = int(port) if port else DEFAULT_REDIS_PORT
+            except ValueError:
+                raise ValueError(
+                    f"Invalid port in `startup_nodes` entry {entry!r}; expected `host:port`."
```
Miretpl:
```suggestion
                    f"Invalid port in `startup_nodes` parameter value for entry {entry!r}; expected `host:port`."
```

Miretpl (changes requested):
Just small nit about exception messages and LGTM.

providers/redis/src/airflow/providers/redis/hooks/redis.py
FrankYang0529: Thanks for the review. Updated all three suggestions.

Miretpl (approved):

PR #71072, opened 2026-08-04
Log and audit when the partitioned Dag run per-tick cap is reached

Lee-W:
## Why
It's not noticeable when the APDR cap is reached.  

## What
Emits a scheduler warning on every tick the cap is hit, plus a single `partition dag run cap reached` audit Log row per backlog episode.

The audit row is written through `create_session(scoped=False)` so the caller's `@retry_db_transaction` rollback cannot discard it; it must stay ahead of any other write in the tick or the extra connection can lock-contend on SQLite.

 

---

---

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2267,11 +2267,12 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         # loses. The `id` tiebreaker on `order_by` keeps LIMIT deterministic when
         # two APDRs share a `created_at` under bulk asset-event ingestion.
         # SQLite is single-writer and silently drops `FOR UPDATE`, which is fine.
-        # LIMIT cap + 1: a row beyond the cap is the cheapest way to tell "there is a real
-        # backlog" apart from "this batch is the entire backlog" without a separate COUNT
-        # query. Only the first `cap` rows are sliced off and processed below — the extra
-        # probe row is never touched.
-        rows = session.scalars(
+        # The fetch is capped at `cap` rows. When it comes back full, we can't tell from the
+        # fetch alone whether that's the entire backlog or just this tick's slice of a larger
+        # one, so we pay for a separate lock-free `count()` query in that case only — trading
+        # an occasional extra read for not locking rows this process won't claim, and for a
+        # real number to report in logs/audit instead of the capped fetch size.
+        pending_apdrs = session.scalars(
             with_row_locks(
                 select(AssetPartitionDagRun)
                 .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
@@ -2280,15 +2281,29 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                     DagModel.is_stale.is_(False),
                 )
                 .order_by(AssetPartitionDagRun.created_at, AssetPartitionDagRun.id)
-                .limit(self._max_partition_dag_runs_per_loop + 1),
+                .limit(self._max_partition_dag_runs_per_loop),
                 of=AssetPartitionDagRun,
                 skip_locked=True,
                 key_share=False,
                 session=session,
             )
         ).all()
-        has_backlog = len(rows) > self._max_partition_dag_runs_per_loop
-        pending_apdrs = rows[: self._max_partition_dag_runs_per_loop]
+        if len(pending_apdrs) >= self._max_partition_dag_runs_per_loop:
+            backlog_total = (
+                session.scalar(
+                    select(func.count())
+                    .select_from(AssetPartitionDagRun)
+                    .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
+                    .where(
+                        AssetPartitionDagRun.created_dag_run_id.is_(None),
+                        DagModel.is_stale.is_(False),
+                    )
+                )
+                or 0
+            )
+        else:
+            backlog_total = len(pending_apdrs)
+        has_backlog = backlog_total > self._max_partition_dag_runs_per_loop
         if has_backlog:
             # Cap the dag_ids surfaced in logs/audit to the first 5 distinct target_dag_ids so a
             # backlog spanning many Dags doesn't blow up the log line / audit row.
@@ -2299,7 +2314,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 "Reached the per-tick cap on pending partitioned Dag runs; the remaining backlog "
                 "will be evaluated over subsequent scheduler ticks",
                 cap=self._max_partition_dag_runs_per_loop,
-                pending_count=len(pending_apdrs),
+                backlog_total=backlog_total,
                 dag_ids=truncated_dag_ids,
             )
             # Edge-trigger the audit row: a persistent backlog re-hits this branch every tick,
@@ -2327,7 +2342,9 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                                     f"runs this tick, reaching the internal per-tick cap of "
                                     f"{self._max_partition_dag_runs_per_loop}. This cap is a hardcoded "
                                     "scheduler safety limit, not a configurable setting; the remaining "
-                                    f"backlog will be evaluated over subsequent ticks. {dag_ids_note}"
+                                    f"backlog will be evaluated over subsequent ticks. A total of "
+                                    f"{backlog_total} partitioned Dag runs are currently eligible and "
+                                    f"pending across ticks. {dag_ids_note}"
                                 ),
                             )
                         )
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -381,6 +381,9 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         # Edge-triggers the "partition Dag run cap reached" audit Log row: True once that row
         # has been committed for the current backlog episode, reset to False once the backlog
         # drains below the cap.
+        # Process-local: each scheduler in an HA deployment writes at most one audit row per
+        # episode; cross-process de-duplication would need a per-tick DB read, defeating the
+        # point of edge-triggering.
         self._partition_cap_backlog_reported = False
         self._dag_id_to_team_name: dict[str, str | None] = {}
 
@@ -2329,7 +2332,8 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         if backlog_total > self._max_partition_dag_runs_per_loop:
             displayed_dag_ids = sorted_dag_ids[:MAX_PARTITION_CAP_BACKLOG_DAG_IDS_LOGGED]
             remaining_dag_id_count = len(sorted_dag_ids) - len(displayed_dag_ids)
-            self.log.warning(
+            log_cap_reached = self.log.debug if self._partition_cap_backlog_reported else self.log.warning
+            log_cap_reached(
                 (
                     "Reached the per-tick cap on pending partitioned Dag runs; the remaining backlog "
                     "will be evaluated over subsequent scheduler ticks"
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -370,7 +370,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         self._multi_team = conf.getboolean("core", "multi_team")
         self._dag_tags_in_metrics = conf.getboolean("metrics", "dag_tags_in_metrics", fallback=False)
         self._max_partition_dag_runs_per_loop = MAX_PARTITION_DAG_RUNS_PER_LOOP
-        # Edge-triggers the "partition dag run cap reached" audit Log row: True once that row
+        # Edge-triggers the "partition Dag run cap reached" audit Log row: True once that row
         # has been committed for the current backlog episode, reset to False once the backlog
         # drains below the cap.
         self._partition_cap_backlog_reported = False
@@ -2304,12 +2304,26 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 )
                 or 0
             )
+            # Distinct dag_ids across the *whole* backlog, not just this tick's oldest-cap
+            # slice (`pending_apdrs`) — a Dag whose partitions haven't reached the front of
+            # the FIFO queue yet would otherwise be missing from the log/audit row until its
+            # turn comes up. Same WHERE clause as the count query above so the two never drift.
+            backlog_dag_ids = set(
+                session.scalars(
+                    select(AssetPartitionDagRun.target_dag_id)
+                    .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
+                    .where(
+                        AssetPartitionDagRun.created_dag_run_id.is_(None),
+                        DagModel.is_stale.is_(False),
+                    )
+                    .distinct()
+                )
+            )
         else:
             backlog_total = len(pending_apdrs)
             self._partition_cap_backlog_reported = False
 
         if backlog_total > self._max_partition_dag_runs_per_loop:
-            affected_dag_ids = {apdr.target_dag_id for apdr in pending_apdrs}
             self.log.warning(
                 (
                     "Reached the per-tick cap on pending partitioned Dag runs; the remaining backlog "
@@ -2317,7 +2331,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 ),
                 cap=self._max_partition_dag_runs_per_loop,
                 backlog_total=backlog_total,
-                dag_ids=affected_dag_ids,
+                dag_ids=backlog_dag_ids,
             )
             # Edge-trigger the audit row: a persistent backlog re-hits this branch every tick,
             # and writing a `Log` row that often would flood the audit table. Write it once per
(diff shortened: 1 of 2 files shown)


Lee-W:
cc @TJaniF

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2247,13 +2256,55 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                            Log(
+                                event="partition dag run cap reached",
+                                extra=(
+                                    f"The scheduler evaluated {len(pending_apdrs)} pending partitioned Dag "
+                                    f"runs this tick, reaching the internal per-tick cap of "
+                                    f"{self._max_partition_dag_runs_per_loop}. This cap is a hardcoded "
+                                    "scheduler safety limit, not a configurable setting; the remaining "
+                                    "backlog will be evaluated over subsequent ticks."
+                                ),
+                            )
```
uranusjr, line 2334: Should this log contain at least dag_id?

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2247,13 +2256,55 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                        )
+                except (OperationalError, exc.TimeoutError):
+                    # Purely observational — must not fail or retry the tick's actual DagRun
+                    # creation work. Leave the flag False so the next tick retries the write.
+                    self.log.warning("Failed to write the partition dag run cap audit Log row", exc_info=True)
```
uranusjr: Should this catch be broader? At least catch all db errors?

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2247,13 +2256,55 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
             )
         ).all()
+        has_backlog = len(rows) > self._max_partition_dag_runs_per_loop
+        pending_apdrs = rows[: self._max_partition_dag_runs_per_loop]
+        if has_backlog:
+            self.log.warning(
+                "Reached the per-tick cap on pending partitioned Dag runs; the remaining backlog "
+                "will be evaluated over subsequent scheduler ticks",
+                cap=self._max_partition_dag_runs_per_loop,
+                pending_count=len(pending_apdrs),
```
uranusjr: I think these two are always the same? The actual count will need a separate `select count()` to be retrieved.

uranusjr: The cap+1 approach may have other issues with HA since we’re locking one more row that does not belong to this process. Since we need a separate count anyway, maybe something like this would be better:

```python
# No +1 in the limit!
pending_apdrs = session.scalars(with_row_locks(... .limit(cap), ...)).all()

if len(pending_apdrs) >= self._max_partition_dag_runs_per_loop:
    backlog_total = session.scalar(count_query)
    has_backlog = backlog_total > self._max_partition_dag_runs_per_loop
```

Also, if row count if a problem, we should add an index on APDR to prevent table scan.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: yep, added affected dag_ids

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Sounds good. Added `DBAPIError`, which also covers `OperationError`.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Went with the count approach instead of the `+1` way. the real total is now calcuated. Held off the index thing for now. we can do it if we find out it's a problem in the future.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2267,8 +2277,71 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                            )
+                        )
+                except (DBAPIError, exc.TimeoutError):
+                    # Purely observational — must not fail or retry the tick's actual DagRun
+                    # creation work. Leave the flag False so the next tick retries the write.
+                    self.log.warning("Failed to write the partition Dag run cap audit Log row", exc_info=True)
+                else:
+                    self._partition_cap_backlog_reported = True
+        else:
+            self._partition_cap_backlog_reported = False
```
jason810496, line 2394:
```suggestion
```

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2267,8 +2277,71 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                    .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
+                    .where(
+                        AssetPartitionDagRun.created_dag_run_id.is_(None),
+                        DagModel.is_stale.is_(False),
+                    )
+                )
+                or 0
+            )
+        else:
+            backlog_total = len(pending_apdrs)
```
jason810496, line 2339: IIUC, it seems equivalent. Would moving the resting part forward be more more clear?

```suggestion
            backlog_total = len(pending_apdrs)
            self._partition_cap_backlog_reported = False

```

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2267,8 +2277,71 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+            # backlog drains.
+            if not self._partition_cap_backlog_reported:
+                # A separate, independently-committed session is required here: the caller
+                # (`_create_dagruns_for_dags`) runs under `@retry_db_transaction`, which rolls
+                # back *session* on a `DBAPIError` — sharing that transaction would silently
+                # discard this audit row along with the rest of the tick's work.
+                # Invariant: this must run before *session* makes any writes this tick, or the
+                # new connection's commit can lock-contend with it on SQLite.
+                try:
+                    dag_ids_note = f"Affected dag_ids: {', '.join(affected_dag_ids)}."
```
jason810496: Should we log those Dag IDs in the backlog? Instead of logging the pending Dag IDs that we're going to create the DagRun.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: yep... something I wrongly moved during refactoring. Thanks!

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Log all the possible pending Dag ID instead. those in the backlogs might be pending Dag as well

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2292,20 +2292,34 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                session.scalars(
+                    select(AssetPartitionDagRun.target_dag_id)
+                    .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
+                    .where(
+                        AssetPartitionDagRun.created_dag_run_id.is_(None),
+                        DagModel.is_stale.is_(False),
+                    )
+                    .distinct()
+                )
+            )
```
uranusjr: I think we can merge this with the above (for `count()`) into one query, something like

```python
backlogs_per_dag = dict(
    session.execute(
        select(AssetPartitionDagRun.target_dag_id, func.count())
        .select_from(AssetPartitionDagRun)
        .join(DagModel, DagModel.dag_id == AssetPartitionDagRun.target_dag_id)
        .where(
            AssetPartitionDagRun.created_dag_run_id.is_(None),
            DagModel.is_stale.is_(False),
        )
        .group_by(AssetPartitionDagRun.target_dag_id)
        .order_by(func.count().desc())
    )
)
backlog_total = sum(n for _, n in backlogs_per_dag.items())
```

And since `backlogs_per_dag` is `dict[str, int]`, it can double as `Collection[str]` and replace `backlog_dag_ids`.

Also, we probably should add an index on APDR to improve performance in the query. This is probably fine for now, and the change can be a separate PR, since the lack of an index didn’t seem to cause serious issues before this PR.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
uranusjr: These comments should be revisited to make it up-to-date.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2269,6 +2279,83 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                        audit_session.add(
+                            Log(
+                                event="partition Dag run cap reached",
+                                extra=(
+                                    f"The scheduler evaluated {len(pending_apdrs)} pending partitioned Dag "
+                                    f"runs this tick, reaching the internal per-tick cap of "
+                                    f"{self._max_partition_dag_runs_per_loop}. The remaining backlog will "
+                                    f"be evaluated over subsequent ticks. A total of {backlog_total} "
+                                    f"partitioned Dag runs are currently eligible and pending across "
+                                    f"ticks. Affected dag_ids: {', '.join(backlog_dag_ids)}."
```
uranusjr: Can this set be potentially large? We need to be careful here since this is going into the database.

uranusjr: Also we should probably sort the ids here to make the message stable.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2269,6 +2279,83 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+
+        if backlog_total > self._max_partition_dag_runs_per_loop:
+            self.log.warning(
+                (
+                    "Reached the per-tick cap on pending partitioned Dag runs; the remaining backlog "
+                    "will be evaluated over subsequent scheduler ticks"
+                ),
+                cap=self._max_partition_dag_runs_per_loop,
+                backlog_total=backlog_total,
+                dag_ids=backlog_dag_ids,
```
uranusjr: Same here (sorting)

uranusjr: Also, more than cap pending-but-unsatisfied APDRs is a reachable steady state; it would not change once hit until upstream keys arrive. On such a deployment, this fires permanently on a healthy system in every scheduler tick. I’m not sure keep warning is very friendly to users. (But we might not have a better solution, on the other hand.)

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -357,6 +358,10 @@ def __init__(
         self._multi_team = conf.getboolean("core", "multi_team")
         self._dag_tags_in_metrics = conf.getboolean("metrics", "dag_tags_in_metrics", fallback=False)
         self._max_partition_dag_runs_per_loop = MAX_PARTITION_DAG_RUNS_PER_LOOP
+        # Edge-triggers the "partition Dag run cap reached" audit Log row: True once that row
+        # has been committed for the current backlog episode, reset to False once the backlog
+        # drains below the cap.
+        self._partition_cap_backlog_reported = False
```
uranusjr, line 388: Does HA scheduler affect this? Does the “reported” state need to be propagated to other schedulers?

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2269,6 +2279,83 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+                                    f"The scheduler evaluated {len(pending_apdrs)} pending partitioned Dag "
+                                    f"runs this tick, reaching the internal per-tick cap of "
+                                    f"{self._max_partition_dag_runs_per_loop}. The remaining backlog will "
+                                    f"be evaluated over subsequent ticks. A total of {backlog_total} "
+                                    f"partitioned Dag runs are currently eligible and pending across "
+                                    f"ticks. Affected dag_ids: {', '.join(backlog_dag_ids)}."
+                                ),
+                            )
+                        )
+                except (DBAPIError, exc.TimeoutError):
```
uranusjr: Should this just catch SQLAlchemyError?

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Yep, sounds good, just merged them in to one PR. We can do the index part if this actually becomes an issue.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Updated the comment based on the latest implementation

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: I don't think it's very likely, but it's possible. I've limited it to 20 dag_ids and sorted them.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Sounds good. Changed it to `SQLAlchemyError`. This audit row should not fail the tick's actaul DagRun creationg.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: `_partition_cap_backlog_reported` is not shared across HA scheduler replicas. Worst case is one audit row per active scheduler per episode instead of one globally. It's redundant but not incorrect, since the write is purely observational. Added a comment for it.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: I changed it to go with warning in the first round and debug level for the following rounds. not sure whether it's actually better. 🤔 not as noisy probably

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2296,6 +2310,89 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
         if not pending_apdrs:
             return set()
```
amoghrajesh, line 2312: The only place that resets `self._partition_cap_backlog_reported` to False is line 2338-2340, further down, in the branch that runs when `len(pending_apdrs) < self._max_partition_dag_runs_per_loop`. That branch never runs when the backlog is exactly zero, because the function already returned at line 2311.

So here is the failure case: the backlog builds up, hits the cap, and gets reported (flag set to True). The backlog then drains all the way to zero in one tick, the flag stays True forever because the early return at line 2311 skips the reset. Later, a fresh burst of asset events pushes the backlog straight back over the cap. The code checks the stale True flag, thinks this is a continuation of the old episode, logs at debug instead of warning, and skips writing a new audit Log row entirely (the `if not self._partition_cap_backlog_reported:` guard at line 2359 is False).

This breaks the one promise this PR makes: "one audit row per backlog episode." A brand new episode goes completely unaudited if the backlog happened to touch zero first. Reset the flag at the top of the function whenever `pending_apdrs` is empty, right next to the early return at line 2310-2311, not only in the `len(pending_apdrs) < cap` branch further down.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2296,6 +2310,89 @@ def _create_dagruns_for_partitioned_asset_dags(self, session: Session) -> set[st
...
+            # Postgres/MySQL in production) isn't a contract worth trusting — sort explicitly.
+            sorted_dag_ids = sorted(backlogs_per_dag, key=lambda dag_id: (-backlogs_per_dag[dag_id], dag_id))
+        else:
+            backlog_total = len(pending_apdrs)
+            self._partition_cap_backlog_reported = False
+
+        if backlog_total > self._max_partition_dag_runs_per_loop:
+            displayed_dag_ids = sorted_dag_ids[:MAX_PARTITION_CAP_BACKLOG_DAG_IDS_LOGGED]
+            remaining_dag_id_count = len(sorted_dag_ids) - len(displayed_dag_ids)
+            log_cap_reached = self.log.debug if self._partition_cap_backlog_reported else self.log.warning
```
amoghrajesh, line 2347: `sorted_dag_ids` is set in a conditional statement above and it works fine today but it depends on two separate if conditions staying in sync by coincidence, not by a name or a comment saying so. A future edit to either condition (e.g. changing `>=` to `>`, or changing what backlog_total means in the else branch) can turn this into a `NameError`

airflow-core/tests/unit/jobs/test_scheduler_job.py
amoghrajesh, line 1: No test drains the backlog to literally zero and then rebuilds it. Add a test that drains the backlog to exactly zero pending APDRs, then adds a fresh batch that crosses the cap again, and assert the second episode logs at warning and writes a second audit row.

Lee-W: Added `test_partition_cap_backlog_reset_on_full_drain_to_empty`. Thanks!

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Thanks! Reset the flag in this block too, so it always starts fresh.

airflow-core/src/airflow/jobs/scheduler_job_runner.py
Lee-W: Fixed by giving `sorted_dag_ids` an explicit default outside the branch, so it no longer depends on the two conditions staying in sync.

PR #71073, opened 2026-08-04
Skip downstream tasks on LLMBranchOperator reject

guan404ming:
## Why

- Rejecting an `LLMBranchOperator` review failed the task, leaving downstream tasks `upstream_failed`.
- `ApprovalOperator` treats reject as a human decision and skips downstream instead; failing on reject is documented as generally discouraged.
- Follow-up to review feedback on #70651.

## How

- Reject now runs `do_branch(context, None)`, skipping every downstream task.
- New `fail_on_reject` flag (default `False`) restores the old fail behavior, mirroring `ApprovalOperator`.
- `approval_timeout` expiry still fails the task.

## E2E

Verified in breeze: trigger, review pauses the task, hit Reject, `route_ticket` ends `success` and all downstream branches end `skipped` (previously `upstream_failed`).

[llm_branch_reject_skip_demo.webm](https://github.com/user-attachments/assets/7f8081fb-cd24-468f-aa6e-a51819438ee2)

---

### DIFF
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -92,9 +92,10 @@ against the downstream task IDs before branching:
     :start-after: [START howto_operator_llm_branch_approval]
     :end-before: [END howto_operator_llm_branch_approval]
 
-Rejecting the review, or letting ``approval_timeout`` expire, **fails** the
-task (``HITLRejectException`` / ``HITLTimeoutError``), so downstream tasks
-end up ``upstream_failed`` rather than skipped.
+Rejecting the review **skips every downstream task**, matching
+:class:`~airflow.providers.standard.operators.hitl.ApprovalOperator`. Set
+``fail_on_reject=True`` to fail the task instead (generally discouraged).
+Letting ``approval_timeout`` expire fails the task (``HITLTimeoutError``).
 
 ``require_approval=True`` requires a string prompt: a decorated callable
 returning a ``Sequence[UserContent]`` raises ``TypeError`` before the LLM
@@ -133,6 +134,8 @@ Parameters
   means wait indefinitely.  Default ``None``.
 - ``allow_modifications``: If ``True``, the reviewer can change the chosen
   branch(es) before approving.  Default ``False``.
+- ``fail_on_reject``: If ``True``, a rejected review fails the task instead of
+  skipping every downstream task.  Generally discouraged.  Default ``False``.
 
 Logging
 -------
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -92,7 +92,8 @@ against the downstream task IDs before branching:
     :start-after: [START howto_operator_llm_branch_approval]
     :end-before: [END howto_operator_llm_branch_approval]
 
-Rejecting the review **skips every downstream task**, matching
+Rejecting the review **skips the direct downstream tasks except teardowns**,
+matching
 :class:`~airflow.providers.standard.operators.hitl.ApprovalOperator`. Set
 ``fail_on_reject=True`` to fail the task instead (generally discouraged).
 Letting ``approval_timeout`` expire fails the task (``HITLTimeoutError``).
@@ -135,7 +136,7 @@ Parameters
 - ``allow_modifications``: If ``True``, the reviewer can change the chosen
   branch(es) before approving.  Default ``False``.
 - ``fail_on_reject``: If ``True``, a rejected review fails the task instead of
-  skipping every downstream task.  Generally discouraged.  Default ``False``.
+  skipping the downstream tasks.  Generally discouraged.  Default ``False``.
 
 Logging
 -------
--- a/providers/common/ai/docs/operators/llm_branch.rst
+++ b/providers/common/ai/docs/operators/llm_branch.rst
@@ -94,14 +94,18 @@ against the downstream task IDs before branching:
 
 Rejecting the review **skips the direct downstream tasks except teardowns**,
 matching
-:class:`~airflow.providers.standard.operators.hitl.ApprovalOperator`. Set
-``fail_on_reject=True`` to fail the task instead (generally discouraged).
-Letting ``approval_timeout`` expire fails the task (``HITLTimeoutError``).
+:class:`~airflow.providers.standard.operators.hitl.ApprovalOperator`. The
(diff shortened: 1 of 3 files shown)


providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -133,7 +143,13 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
...
     def execute_complete(self, context: Context, generated_output: str, event: dict[str, Any]) -> Any:
         """Resume after human review, validating the reviewed choice before branching."""
-        output = super().execute_complete(context, generated_output, event)
+        try:
+            output = super().execute_complete(context, generated_output, event)
+        except HITLRejectException:
+            if self.fail_on_reject:
+                raise
+            self.log.info("Rejected. Skipping all downstream tasks...")
+            return self.do_branch(context, None)
```
kaxil, line 152: `do_branch(context, None)` skips every direct downstream task, teardowns included, but `ApprovalOperator`'s reject path filters them out (`yield from (t for t in tasks if not t.is_teardown)`). I checked this on a real DAG with `branch >> [task_a, task_b, cleanup.as_teardown()]`: this path skips `['cleanup', 'task_a', 'task_b']` where `ApprovalOperator` skips `['task_a', 'task_b']`. Before this change a reject failed the task and the teardown still ran on its `all_done`-style rule, so cleanup now silently stops running after a rejection.

Either filter teardowns here, or drop the "matching `ApprovalOperator`" wording from the docstring and the rst.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -133,7 +143,13 @@ def execute(self, context: Context) -> str | Iterable[str] | None:
 
     def execute_complete(self, context: Context, generated_output: str, event: dict[str, Any]) -> Any:
         """Resume after human review, validating the reviewed choice before branching."""
-        output = super().execute_complete(context, generated_output, event)
+        try:
+            output = super().execute_complete(context, generated_output, event)
+        except HITLRejectException:
+            if self.fail_on_reject:
+                raise
+            self.log.info("Rejected. Skipping all downstream tasks...")
```
kaxil, line 151: The exception being swallowed here carries the reviewer name (`Output was rejected by the reviewer <user>.`), and the task log was the only place that surfaced. `event["responded_by_user"]` is right here, so `self.log.info("Rejected by %s. Skipping all downstream tasks.", event.get("responded_by_user"))` would keep the attribution. Right now the log for a rejected gate no longer says who rejected it.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Good catch. Now filters teardowns like `ApprovalOperator` and updated the docs wording accordingly.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Fixed. Reject log now includes responded_by_user, keeping reviewer attribution.

providers/common/ai/docs/operators/llm_branch.rst
```diff
@@ -92,9 +92,11 @@ against the downstream task IDs before branching:
...
     :end-before: [END howto_operator_llm_branch_approval]
 
-Rejecting the review, or letting ``approval_timeout`` expire, **fails** the
-task (``HITLRejectException`` / ``HITLTimeoutError``), so downstream tasks
-end up ``upstream_failed`` rather than skipped.
+Rejecting the review **skips the direct downstream tasks except teardowns**,
+matching
+:class:`~airflow.providers.standard.operators.hitl.ApprovalOperator`. Set
+``fail_on_reject=True`` to fail the task instead (generally discouraged).
+Letting ``approval_timeout`` expire fails the task (``HITLTimeoutError``).
```
kaxil: Worth scoping this sentence to the reject path. The approve path goes through `skip_all_except`, which has no `is_teardown` filter, so on the same `branch >> [task_a, task_b, cleanup.as_teardown()]` topology from the thread above, approving skips `['cleanup', 'task_b']` and the teardown is discarded. That's inherited branch behaviour rather than anything this PR changes, but as written it reads like a general guarantee, and the surprising half is that rejecting preserves `cleanup` while approving drops it.

Separately, line 105 says `approval_timeout` and the rest of the approval behaviour are inherited from `LLMOperator`, which is no longer true of `fail_on_reject`.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
```diff
@@ -69,11 +77,13 @@ def __init__(
         self,
         *,
         allow_multiple_branches: bool = False,
+        fail_on_reject: bool = False,
```
kaxil, line 80: Non-blocking: why does `fail_on_reject` land only on `LLMBranchOperator`? The reject is raised in the shared `LLMApprovalMixin.execute_complete` (`mixins/approval.py:222`), so `LLMOperator`, `LLMSQLQueryOperator`, `LLMSchemaCompareOperator` and `LLMFileAnalysisOperator` still fail the task on reject with no opt-out, and two operators carrying the same `require_approval=True` flag now do opposite things. Putting the flag on the mixin would keep the family consistent; happy for that to be a follow-up if you'd rather keep this PR tight.

providers/common/ai/docs/operators/llm_branch.rst
guan404ming: Thanks for the comments. I just reworded that the teardown carve-out is now explicitly scoped to rejection, with a note that approving branches as usual so unselected teardowns are skipped. Also updated the inheritance sentence to call out fail_on_reject as operator-specific.

providers/common/ai/src/airflow/providers/common/ai/operators/llm_branch.py
guan404ming: Agreed on keeping the family consistent. Moving the flag to LLMApprovalMixin needs skip machinery (inherits_from_skipmixin, teardown-aware skip) on the non-branch operators plus a decision on their default, so I'd rather do it as a follow-up PR to keep this one tight. Will open it after this merges.

Lee-W (approved):

guan404ming:
Thanks!

Issue #58919, opened 2025-12-02
Resolve race condition re creation of APDR records


dstandish:
### Body

When two tasks run at the same time and they contribute to the creation of the same dag run and partition key, then they can create different apdr records.  This has to be fixed somehow.  Maybe different design.  More notes on options in comments below.

### How to repro

Check out branch [`poc-asset-partitions`. ](https://github.com/astronomer/airflow/tree/poc-asset-partitions).

Make 2 identical dags with schedule e.g. `*/5 * * * *`, each updating a different asset.  Make another dag that listens to those assets using partitioned asset timetable.

Enable all the dags.

The two producer dags will run at basically the same time.  Two APDR records will get created instead of one, and as a result, the downstream dag won't schedule.

### More explanation

right now, as soon as one asset fires which maps to target key X, we create APDR record for target key X
once that record is created, scheduler will consider APDR record for target key X in each loop
when the second asset fires an event which maps to target key X, it should see that there is an existing APDR record and it should therefore just create the PAKL record which maps to this APDR record
but if they fire at exactly the same time, currently due to this bug, they will both see that there is no APDR record, and they will therefore both create an APDR record, and therefore even though a run should be created, it isn't

### Committer

- [x] I acknowledge that I am a maintainer/committer of the Apache Airflow project.

dstandish:
possible solutions

1. use locking when running get-or-create-apdr. Somehow we would need to use locking so that the two writers would not both see no apdr record and thus both create one at the same time.  Most extreme (probably not good) would be to take a lock on the dag table.  But this could interfere with performance and other thnigs that try to lock the dag table.  There are other ideas.  E.g. create an apdr mutex table (or perhaps a partition state table...?) that would have the grain of dag_id and partition_key (i.e. a uniqueness constraint on that) and then this could be locked during the get-or-create operation, ensuring sequentiality.  maybe the "apdr mutex table" is just a parent table with grain dag_id, partition_key and apdr keys to this.

2. we could get rid of the apdr table and change the querying and logic to just sift through the pile of records in the key log table to figure out what runs need to be created etc.  The querying would be more expensive but you would not be subject to the race condition.

3. still use locking somehow but avoid an external mutex table by somehow using locking within apdr.  e.g. perhaps implement some is_latest flag with a partial unique constraint, or something different

Lee-W:
> use locking when running get-or-create-apdr. Somehow we would need to use locking so that the two writers would not both see no apdr record and thus both create one at the same time.

[...]

PR #71074, opened 2026-08-04
Fix duplicate pending Dag runs for the same asset partition key

### LINKED ISSUE #71070: Partitioned DAG has two Pending Dag Runs with the same partition key

Airflow Core

### Apache Airflow version

3.3.0

DAG was not triggered due to 2 Pending Dag runs still waiting on different asset events.
Both however have the same Mapped Partition Key and should be considered as one.
Additionally clicking on the received asset events an empty tooltip is shown:

Image

Records in asset_partition_dag_run show two records made with one ms apart. Both are identical except for the partition_date.
One record has it filled in, the other is empty:

Image

DAG should be triggered when all 4 events were received.

### Operating System

Ubuntu 22.04.5 LTS

### Deployment

Virtualenv installation

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

anmolxlight:
closes #71070

## Summary

When two **different producer assets** map to the same downstream partition key (e.g. via `IdentityMapper`), each asset event could create its own `AssetPartitionDagRun` row. The scheduler then held two Pending Dag runs for the same partition key and never triggered the consumer Dag.

## Root cause

`AssetManager._get_or_create_apdr` serialized APDR find-or-create with `_lock_asset_model`, which locks the **producer** `AssetModel` row. But APDR dedup is keyed on `(target_dag_id, partition_key)`. Two events from different producer assets that resolve to the same target key therefore took locks on **two different asset rows**, neither blocked the other, and both observed "no existing APDR" and inserted a duplicate. The reporter's two rows created ~1ms apart with one carrying `partition_date` and the other empty match this exactly: the two events came from different producers, and only one carried a date.

## Fix

Lock the **target** `DagModel` row instead (`_lock_target_dag`). All APDR find-or-create calls for a given consumer Dag now serialize on the same resource regardless of which producer asset triggered the event, so the second event finds the APDR created by the first and dedups onto it. SQLite still uses the global writer lock (unchanged semantics), and the existing "work on the latest matching APDR" fallback in the UI route and scheduler cleanup is preserved.

## Test

Added `test_queue_partitioned_dags_dedups_across_different_producer_assets`: two `register_asset_change` calls from different producer assets with the same partition key must both lock the same target Dag and yield exactly one `AssetPartitionDagRun`.

Verified locally with breeze: `6 passed` (including the new test and all existing `_get_or_create_apdr` partition-date tests), ruff clean, formatting clean.

### DIFF
--- a/airflow-core/src/airflow/assets/manager.py
+++ b/airflow-core/src/airflow/assets/manager.py
@@ -18,7 +18,6 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Collection, Iterable
-from contextlib import contextmanager
 from functools import partial
 from typing import TYPE_CHECKING
 
@@ -48,7 +47,8 @@ from airflow.models.log import Log
 from airflow.timetables.base import compute_rollup_fingerprint
 from airflow.utils.helpers import is_container, prune_dict
 from airflow.utils.log.logging_mixin import LoggingMixin
-from airflow.utils.sqlalchemy import get_dialect_name, with_row_locks
+from airflow.utils.session import create_session
+from airflow.utils.sqlalchemy import get_dialect_name
 
 if TYPE_CHECKING:
     from datetime import datetime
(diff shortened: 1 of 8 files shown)


Lee-W:
According to previous discussion https://github.com/apache/airflow/issues/58919#issuecomment-3629496888, I don't think we would like to have Dag table locked

anmolxlight:
Pushed a rework that drops the Dag-row lock entirely — no `_lock_target_dag`, no SQLite writer-lock retry loop, no Postgres/MySQL `SELECT ... FOR UPDATE` on `DagModel`.

The new approach: a unique constraint on `(target_dag_id, pending_partition_key)`, where `pending_partition_key` mirrors `partition_key` only while `created_dag_run_id IS NULL` (cleared via a `validates` hook the moment a dag run is created). This stays in the design space outlined in https://github.com/apache/airflow/issues/58919#issuecomment-3629496888 — no Dag table lock, and no table-wide/advisory lock either. The reason it's a mirrored column instead of the more direct partial/filtered unique index (`WHERE created_dag_run_id IS NULL`) is portability: MySQL supports neither partial nor filtered indexes, so the mirrored-null-column trick is what's actually usable across sqlite/Postgres/MySQL — a plain unique index already treats NULL as distinct from every other value on all three.

Concurrency is now optimistic instead of pessimistic: `_get_or_create_apdr` selects the latest pending APDR, and if none exists, inserts inside a `SAVEPOINT` (`session.begin_nested()`). If two threads race, the loser's INSERT hits the unique constraint and raises `IntegrityError`, which is caught to re-select and work on the winner's row instead of raising — consistent with the model's existing "always work on the latest matching APDR record" contract. Migration `0128_3_4_0_add_pending_partition_key_to_apdr.py` adds the column, backfills it, collapses any pre-existing duplicate pending rows down to the latest one per key (mirroring the scheduler's own stale-APDR cleanup), then creates the constraint.

Test evidence: `test_get_or_create_apdr_is_idempotent_under_concurrent_calls` now asserts convergence on exactly one row without assuming which of the two code paths (initial SELECT hit vs. losing the INSERT race) each thread takes, since that's a timing detail, not the invariant. `test_register_asset_change_from_different_producer_assets_does_not_duplicate_apdr` (the apache/airflow#71070 regression test) now exercises the real path with no lock spy. Added `test_pending_partition_key_unique_constraint_blocks_duplicate_pending_rows` (DB rejects a second pending row directly) and `test_created_dag_run_id_assignment_clears_pending_partition_key` (validator coverage). Full `airflow-core/tests/unit/assets/test_manager.py` suite: 56 passed.

---
Drafted-by: Claude Code (Sonnet 5)

airflow-core/src/airflow/assets/manager.py
```diff
@@ -754,27 +698,124 @@ def _get_or_create_apdr(
...
+            partition_date=target_partition_date,
+            rollup_fingerprint=rollup_fingerprint,
+        )
+        try:
+            # A SAVEPOINT scopes the potential IntegrityError so only this INSERT is rolled
+            # back on conflict; the caller's surrounding transaction (with any other work
+            # already flushed in this scheduler tick) stays intact.
+            with session.begin_nested():
+                session.add(apdr)
+                session.flush()
```
uranusjr, line 742: SQLAlchemy's savepoint rollback expunges all of session._new, not just what was added inside. `_queue_partitioned_dags` accumulates unflushed PartitionedAssetKeyLog and Log rows across loop iterations. Losing a log row reproduces the same "APDR never satisfied" failure this PR fixes.

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_pending_partition_key_to_apdr.py
```diff
@@ -0,0 +1,116 @@
...
+                "     GROUP BY target_dag_id, partition_key"
+                ")"
+            )
+        ).fetchall()
+    ]
+    if not stale_ids:
+        return
+    id_list = ", ".join(str(i) for i in stale_ids)
+    conn.execute(sa.text(f"DELETE FROM {_LOG_TABLE} WHERE asset_partition_dag_run_id IN ({id_list})"))
+    conn.execute(sa.text(f"DELETE FROM {_TABLE} WHERE id IN ({id_list})"))
```
uranusjr: Deleting the loser's `partitioned_asset_key_log` rows means existing deployments upgrade into a still-stuck APDR. I would do `UPDATE ... SET asset_partition_dag_run_id = <winner>` instead to heal it; there's no constraint on that table blocking it.

airflow-core/src/airflow/assets/manager.py
anmolxlight: Verified this empirically rather than just from the code. The scenario does not reproduce: `Session.begin_nested()` itself flushes all pending ORM state into the parent transaction *before* the SAVEPOINT is opened. In SQLAlchemy 2.0.51, `SessionTransaction.__init__` calls `_take_snapshot()`, and for a `BEGIN_NESTED` origin `is_begin` is False, so it runs `self.session.flush()` before resetting `self._new` to an empty snapshot. By the time `with session.begin_nested():` in `_get_or_create_apdr` is entered, any unflushed `Log`/`PartitionedAssetKeyLog` rows accumulated earlier in `_queue_partitioned_dags`'s loop are already persistent, not in `session._new`. So `_restore_snapshot()`'s `to_expunge = set(self._new) | set(session._new)` on rollback only ever contains the savepoint-local `apdr` object, which the `except IntegrityError` block already re-selects around.

I confirmed this with a standalone repro (unflushed rows added before `begin_nested()`, forced `IntegrityError` inside the savepoint, rollback, then flush) and all pre-savepoint rows persisted. I also added a comment explaining this in `_get_or_create_apdr` (manager.py) and a regression test, `test_lost_race_does_not_expunge_unflushed_log_rows` in `test_manager.py`, that forces a deterministic lost race and asserts an unflushed log row added earlier survives and persists. See 09fcebfadc3.

airflow-core/src/airflow/migrations/versions/0128_3_4_0_add_pending_partition_key_to_apdr.py
anmolxlight: Agreed, and confirmed there is no constraint blocking the UPDATE. `_drop_stale_duplicate_pending_apdrs` (now `_heal_stale_duplicate_pending_apdrs`) computes the loser-to-winner id mapping per (target_dag_id, partition_key) group, runs `UPDATE partitioned_asset_key_log SET asset_partition_dag_run_id = <winner_id> WHERE asset_partition_dag_run_id = <loser_id>` for each loser before dropping the loser APDR rows. The module docstring is updated to describe the re-pointing instead of the drop; its first line (the one docs/migrations-ref.rst pulls from) is unchanged. Added a regression test, test_0130_pending_partition_key_migration.py, seeding a duplicate pending pair with log rows and asserting the loser's log rows survive re-pointed onto the winner and the loser APDR is gone. See 5c9a229d7bb.

PR #71076, opened 2026-08-04
Explicitly rollback on task state update exception

uranusjr:
One small gap discovered when analyzing #70951.

This ensures everything is reset when something fails, including asset registration db calls previously made.


Lee-W (approved):
should we add unit test to it?

uranusjr:
Test added

Lee-W (approved):

uranusjr:
I’ll simply add this to #71065.

PR #71078, opened 2026-08-04
Only refuse team scoped like secret ids when multi_team is on

amoghrajesh:
---

Related to (#70878), (#70876), (#70899), (#70869), (#70877), (#70736)

## Summary

- Azure Key Vault, AWS Secrets Manager, AWS SSM, GCP Secret Manager, and Yandex Lockbox all refuse to look up any id that contains the team-namespace separator, to avoid ambiguity with a team-scoped secret name.
- This runs even when multi-team mode is off, which is the default for almost every deployment. Without multi-team mode, no team-scoped secret can exist, so there's no ambiguity to protect against - it just silently breaks ordinary ids
- On Azure specifically, an id is compared against its underscore-to-dash normalized form, so an ordinary
connection id containing a double underscore (e.g. aws__prod) was silently refused as if it were a missing secret: https://github.com/apache/airflow/blob/main/providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py#L230-L251

Trying to gate this by only run the check when `[core] multi_team` is enabled, same as Airflow already does at the model layer.

---

### DIFF
--- a/providers/amazon/tests/unit/amazon/aws/secrets/test_systems_manager.py
+++ b/providers/amazon/tests/unit/amazon/aws/secrets/test_systems_manager.py
@@ -116,6 +116,7 @@ class TestSsmSecrets:
         returned_uri = ssm_backend.get_conn_value(conn_id="test_postgres", team_name="my_team")
         assert returned_uri == "postgresql://airflow:airflow@host:5432/airflow"
 
+    @conf_vars({("core", "multi_team"): "True"})
     @mock_aws
     def test_global_caller_cannot_access_team_scoped_connection(self):
         param = {
@@ -127,6 +128,7 @@ class TestSsmSecrets:
         ssm_backend.client.put_parameter(**param)
         assert ssm_backend.get_conn_value(conn_id="my_team--test_postgres") is None
 
+    @conf_vars({("core", "multi_team"): "True"})
     @mock_aws
     def test_another_teams_secret_is_not_reachable(self):
         """A caller scoped to one team must not reach another team's parameter by naming it."""
@@ -140,6 +142,7 @@ class TestSsmSecrets:
 
         assert ssm_backend.get_conn_value(conn_id="my_team--test_postgres", team_name="other_team") is None
 
+    @conf_vars({("core", "multi_team"): "True"})
     @mock_aws
     def test_team_whose_name_extends_the_callers_is_not_reachable(self):
         """A prefix match on the caller's own namespace is not proof of ownership."""
@@ -153,6 +156,7 @@ class TestSsmSecrets:
 
         assert ssm_backend.get_conn_value(conn_id="my_team--prod--test_postgres", team_name="my_team") is None
 
+    @conf_vars({("core", "multi_team"): "True"})
     @mock_aws
     def test_team_scoped_lookup_cannot_reach_a_longer_teams_namespace(self):
         """The team scoped name is not safe by construction -- the id can extend it.
@@ -172,6 +176,7 @@ class TestSsmSecrets:
 
         assert ssm_backend.get_conn_value(conn_id="prod--test_postgres", team_name="my_team") is None
 
+    @conf_vars({("core", "multi_team"): "True"})
     @mock_aws
     def test_refusing_an_ambiguous_id_is_logged(self, caplog):
         """A silent ``None`` is indistinguishable from a missing secret, so the refusal is logged.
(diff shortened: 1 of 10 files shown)


potiuk (approved):
Makes sense !

amoghrajesh:
@potiuk this touches providers only, so no need to backport.

amoghrajesh:
Here's the companion PR for env backend: https://github.com/apache/airflow/pull/71079

vatsrahul1001 (approved):

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
```diff
@@ -248,7 +248,12 @@ def _names_a_team_namespace(self, secret_id: str) -> bool:
         backend does, is wrong here: the inherited implementation prepends a separator to an
         empty prefix (``'' -> '-smtp_default'``) and normalizes nothing, so the guard would
         both mis-anchor and miss ids whose separator only appears after normalization.
+
+        Only checked in multi-team mode: ``team_name`` is never non-``None`` otherwise, so no
+        team scoped secret can exist to collide with.
         """
+        if not conf.getboolean("core", "multi_team", fallback=False):
```
kaxil, line 255: The Google backend docs still say the opposite. `providers/google/docs/secrets-backends/google-cloud-secret-manager-backend.rst:236` warns that the refusal "applies whether or not you use teams" and that an id already containing `--` "stops resolving after upgrading and you must rename it". That rename advice is wrong after this change, and the new failure mode (an id with `--` resolves fine until someone enables `multi_team`, then silently stops) isn't documented anywhere. Can this PR update that section?

providers/microsoft/azure/src/airflow/providers/microsoft/azure/secrets/key_vault.py
```diff
@@ -247,7 +248,12 @@ def _names_a_team_namespace(self, secret_id: str) -> bool:
         The id is normalised first because :meth:`build_path` maps ``_`` onto the separator
         everywhere in this backend, so ``b__c`` reaches Key Vault as ``b--c`` and would
         otherwise manufacture the team separator from an id that does not visibly contain it.
+
+        Only checked in multi-team mode: ``team_name`` is never non-``None`` otherwise, so no
+        team scoped secret can exist to collide with.
         """
+        if not conf.getboolean("core", "multi_team", fallback=False):
```
kaxil, line 255: `get_config` runs through this guard too (line 193), but config lookups are never team-scoped: `_get_secret(self.config_prefix, key)` passes no `team_name`, so a config key has no team-scoped name to collide with in either mode. The key here is the operator-chosen path from `<option>_secret`, and this backend maps `_` onto the separator, so a path named after the env var (`airflow__database__sql_alchemy_conn` becomes `airflow--database--sql-alchemy-conn`) is still refused whenever `multi_team` is on. Google and Yandex never guarded `get_config`, and the Google docs state config lookups are unaffected, so dropping it from `get_config` here and in the two AWS backends would make all five agree. Happy for that to be a follow-up if you want to keep this PR narrow.

providers/microsoft/azure/tests/unit/microsoft/azure/secrets/test_key_vault.py
```diff
@@ -215,6 +224,17 @@ def test_refusing_an_ambiguous_id_is_logged(self, mock_client, caplog):
...
         mock_client.get_secret.assert_not_called()
 
+    @mock.patch(f"{KEY_VAULT_MODULE}.AzureKeyVaultBackend.client")
+    def test_ambiguous_id_resolves_when_multi_team_is_disabled(self, mock_client):
+        """No team scoped secret can exist without multi-team mode, so there is no ambiguity
+        to refuse -- an ordinary id containing the separator must resolve normally."""
+        mock_client.get_secret.return_value = mock.Mock(value="world")
+        backend = AzureKeyVaultBackend()
+
+        assert backend.get_conn_value("prod--my_db") == "world"
```
kaxil, line 234: `get_secret` returns `world` for any name here, so these pass regardless of the name the backend computes, which is the part the `prod__hello` case in the description turns on. The tests above assert the resolved name (lines 45, 88, 111). Worth at least `mock_client.get_secret.assert_any_call(name="airflow-variables-prod--hello")` for the `prod__hello` lookup.

providers/amazon/tests/unit/amazon/aws/secrets/test_systems_manager.py
```diff
@@ -116,6 +116,7 @@ def test_get_conn_value_with_team_name(self):
         returned_uri = ssm_backend.get_conn_value(conn_id="test_postgres", team_name="my_team")
         assert returned_uri == "postgresql://airflow:airflow@host:5432/airflow"
 
+    @conf_vars({("core", "multi_team"): "True"})
```
kaxil, line 119: The other four test files in this PR define `multi_team_enabled = conf_vars({("core", "multi_team"): "True"})` at module level. Worth doing the same here rather than repeating the literal on six tests.

providers/google/src/airflow/providers/google/cloud/secrets/secret_manager.py
amoghrajesh: Thanks, here you go: https://github.com/apache/airflow/pull/71106

providers/microsoft/azure/tests/unit/microsoft/azure/secrets/test_key_vault.py
amoghrajesh: Handled in https://github.com/apache/airflow/pull/71106

providers/amazon/tests/unit/amazon/aws/secrets/test_systems_manager.py
amoghrajesh: Handled in https://github.com/apache/airflow/pull/71106

PR #71086, opened 2026-08-04
Fix Dataflow sensors losing their XCom value when not deferred

FrankYang0529:
### Why

- `DataflowJobMessagesSensor`, `DataflowJobAutoScalingEventsSensor` and `DataflowJobMetricsSensor` return their payload to XCom when deferred, but drop it in poke mode. `deferrable` defaults to `False`, so the default path is the broken one.
- The payload travels through a chain, and these sensors break it at the last step:

  ```
  poke()                          →  PokeReturnValue(is_done=True, xcom_value=X)
  BaseSensorOperator.execute()    →  picks X out and returns it
  subclass execute()              →  must return it upwards        ← dropped here
  task runner                     →  pushes it as the return_value XCom
  ```

### How

- Return `super().execute(context)` from the three sensors' `execute()` overrides.
- Wrap `DataflowJobMetricsSensor.poke()`'s result in `PokeReturnValue`, the same way #53115 did for the other two. 

 

---

- [X] Yes - Claude Code

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/sensors/dataflow.py
+++ b/providers/google/src/airflow/providers/google/cloud/sensors/dataflow.py
@@ -219,7 +219,7 @@ class DataflowJobMetricsSensor(BaseSensorOperator):
         self.deferrable = deferrable
         self.poll_interval = poll_interval
 
-    def poke(self, context: Context) -> bool:
+    def poke(self, context: Context) -> PokeReturnValue | bool:
         if self.fail_on_terminal_state:
             job = self.hook.get_job(
                 job_id=self.job_id,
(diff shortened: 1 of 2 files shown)


molcay:
Hi @FrankYang0529,

Thank you for your change. It is very much appreciated!

From my side, it seems a good change. However we want to see the system tests run results before merging a change. 

Did you run the system tests? Can you share the results of the system tests?

shahar1 (approved):

FrankYang0529:
> Did you run the system tests? Can you share the results of the system tests?

@molcay Thanks for the suggestion. I add a xcom assertion to system test and make sure it passes on this branch, but fails on main branch.

Screenshot 2026-09-14 at 9 57 04 PM

PR #71091, opened 2026-08-04
Stop dag processor from warning on every file path normalized for stats

### LINKED ISSUE #71084: Dag processor keep printing unnecessary warning log

Airflow Core

### Apache Airflow version

3.3.0

Just right after dag-processor is running, below log will keep printing. (During the converting from '/' to '_'. )
`2026-08-04T09:59:37.8544212 [warning ] Name 'dags/test/test_dag.py' contains invalid characters for stats reporting. Reporting stats with normalized name 'dags_test_test_dag.py'. [airflow._shared.observability.metrics.stats] loc=stats.py:71`

Related code : 

- [stats.py#L87-L88](https://github.com/apache/airflow/blob/main/shared/observability/src/airflow_shared/observability/metrics/stats.py#L87-L88)
- [invoking normalize_name_for_stats without passing `log_warning` param](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/dag_processing/manager.py#L49)

I understood that metrics with underscore is expected, while this warning log isn`t, as it is expected to have '/' in folder structure.

the folder structure is expected to contains '/', this warning is not valid.

### Operating System

_No response_

### Deployment

Virtualenv installation

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

bujjibabukatta:
**Summary**
Right after the dag-processor starts, it repeatedly logs a warning like:
"Name 'dags/test/test_dag.py' contains invalid characters for stats
reporting. Reporting stats with normalized name 'dags_test_test_dag.py'."
This fires once per DAG file, on every processing cycle.

**Root cause**
DagFileInfo.normalized_file_path_for_stats calls normalize_name_for_stats()
without log_warning=False. A relative file path always contains "/" as a
directory separator, so normalization always requires substitution -- the
warning can never signal an actual problem, only expected, routine behavior.

**Fix**
Pass log_warning=False for this specific call. The other normalize_name_for_stats
call sites in the same file (bundle_name, file name stem) are left unchanged,
since those genuinely rarely contain invalid characters, so a warning there
still carries real diagnostic value.

**Testing**
Added a regression test asserting normalized_file_path_for_stats does not
emit a warning for a normal nested file path.

closes: #71084

- [X] Yes - Claude

### DIFF
--- a/airflow-core/tests/unit/dag_processing/test_manager.py
+++ b/airflow-core/tests/unit/dag_processing/test_manager.py
@@ -3636,4 +3636,4 @@ def test_normalized_file_path_for_stats_does_not_warn(caplog):
         result = dag_file_info.normalized_file_path_for_stats
 
     assert result == "dags_test_test_dag.py"
-    assert caplog.records == []
+    assert caplog.entries == []
--- a/airflow-core/tests/unit/dag_processing/test_manager.py
+++ b/airflow-core/tests/unit/dag_processing/test_manager.py
@@ -3620,3 +3620,20 @@ class TestMultiTeamMetrics:
         # Two bundles resolved in a single batched query; the repeat call is served from cache.
         mock_get_team_names.assert_called_once()
         assert manager._bundle_name_to_team_name == {"bundle_a": "team_alpha", "bundle_b": "team_alpha"}
+
+
+def test_normalized_file_path_for_stats_does_not_warn(caplog):
+    """
+    rel_path always contains "/" for any nested DAG file, so normalizing it for stats
+    always requires substitution -- this must not log a warning on every DAG file, every
+    processing cycle.
+    """
+    dag_file_info = DagFileInfo(
+        bundle_name="testing", bundle_path=TEST_DAGS_FOLDER, rel_path=Path("dags/test/test_dag.py")
+    )
+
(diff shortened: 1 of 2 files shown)


ephraimbuddy (approved):

airflow-core/tests/unit/dag_processing/test_manager.py
```diff
@@ -3620,3 +3620,20 @@ def test_get_team_names_batches_and_caches(self, mock_get_team_names):
...
+    """
+    dag_file_info = DagFileInfo(
+        bundle_name="testing", bundle_path=TEST_DAGS_FOLDER, rel_path=Path("dags/test/test_dag.py")
+    )
+
+    with caplog.at_level(logging.WARNING, logger="airflow._shared.observability.metrics.stats"):
+        result = dag_file_info.normalized_file_path_for_stats
+
+    assert result == "dags_test_test_dag.py"
+    assert caplog.records == []
```
ephraimbuddy:
```suggestion
    assert caplog.entries == []
```

eladkal (approved):

PR #69081, opened 2026-06-27
Cap airbyte-api below 1.0.0 to unblock provider tests

shahar1:
## Human Summary
Currently `airbyte-api` doesn't define an upper-bound. Recently v1 has been released and it breaks compatibility with the existing provider, causing failures in canary builds. Applying an upper bound should solve it until compatibility is resolved in #68882.

## AI Summary
<details>
<summary>Click here</summary>
`airbyte-api` 1.0.0 (released 2026-06-23) rewrote the SDK on top of `httpx` and now rejects any client that does not implement its `HttpClient` protocol. The Airbyte hook passes a `requests.Session` for proxy support, so once 1.x is resolved every hook test errors out in setup with:

```
AssertionError: The provided client must implement the HttpClient protocol.
```

The provider only pinned `airbyte-api>=0.52.0` (no upper bound), so CI picks up the new 1.x line and all 17 `TestAirbyteHook` tests error.

This caps the dependency to the requests-based `0.x` line to restore green CI immediately. The full migration to `airbyte-api>=1.0.0` and `httpx.Client` is in flight separately in #68882; the cap should be removed once that lands (a comment at the cap site points there).
</details>

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #71092, opened 2026-08-04
Add configurable request timeout for the Airbyte API

filipeaaoliveira:
Hello everyone.
While upgrading Airflow to 3.3.0 yesterday, our Airbyte DAGs started failing on every trigger even though the syncs themselves were starting in Airbyte.

The cause is the provider's move to **airbyte-api 1.x**, which replaced requests with **httpx** and with it picked up httpx's 5-second default request timeout. Provider 6.0.0 requires the 1.x SDK, but 5.5.1 hits it too: its **airbyte-api>=0.52.0** pin has no upper bound ([#69081](https://github.com/apache/airflow/pull/69081) added the <1.0.0 cap after 5.5.1 shipped), and constraints-3.3.0 resolves it to **airbyte-api==1.0.1**.

On self-hosted deployments, job creation can take far longer than 5 seconds under load — we measured POST /v1/jobs at 30–130 seconds with 9 connections triggered in parallel. Every trigger task then fails with `httpx.ReadTimeout` raised from `AirbyteHook.submit_sync_connection`. And because create-job is not idempotent, the timed-out request still creates the job server-side, so retries get 409 "A sync is already running" and the attempt that eventually succeeds starts a duplicate sync.

The hook currently exposes no way to change the timeout. This PR lets users set it through an `AirbyteHook` parameter or a connection extra, passing it to the SDK's existing `timeout_ms`. With neither set, the session is built exactly as before, so default behavior is unchanged.

Generated-by: Claude Code - Fable 5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
+++ b/providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
@@ -37,6 +37,10 @@ class AirbyteHook(BaseHook):
     :param airbyte_conn_id: Optional. The name of the Airflow connection to get
         connection information for Airbyte. Defaults to "airbyte_default".
     :param api_version: Optional. Airbyte API version. Defaults to "v1".
+    :param timeout: Optional. Request timeout, in seconds, for each call to the
+        Airbyte API. Overrides the ``timeout`` key in the connection's Extra field.
+        When neither is set, the underlying HTTP client's default of 5 seconds
+        applies.
     """
 
     conn_name_attr = "airbyte_conn_id"
@@ -48,10 +52,12 @@ class AirbyteHook(BaseHook):
         self,
         airbyte_conn_id: str = "airbyte_default",
         api_version: str = "v1",
+        timeout: float | None = None,
     ) -> None:
         super().__init__()
         self.api_version: str = api_version
         self.airbyte_conn_id = airbyte_conn_id
+        self.timeout = timeout
         self.conn = self.get_conn_params(self.airbyte_conn_id)
         self.airbyte_api = self.create_api_session()
 
@@ -71,6 +77,7 @@ class AirbyteHook(BaseHook):
         conn_params["client_secret"] = conn.password
         conn_params["token_url"] = conn.schema or "v1/applications/token"
         conn_params["proxies"] = conn.extra_dejson.get("proxies", None)
+        conn_params["timeout"] = conn.extra_dejson.get("timeout", None)
 
         return conn_params
 
@@ -124,10 +131,25 @@ class AirbyteHook(BaseHook):
                 }
             client = httpx.Client(mounts=mounts)
 
+        timeout = self.timeout if self.timeout is not None else self.conn["timeout"]
+        timeout_ms: int | None = None
+        if timeout is not None:
+            try:
+                timeout_ms = int(float(timeout) * 1000)
+            except (TypeError, ValueError):
+                timeout_ms = 0
+            if timeout_ms <= 0:
+                raise ValueError(
+                    f"Invalid Airbyte API request timeout {timeout!r}: expected a positive number of "
+                    f"seconds, set via the AirbyteHook 'timeout' parameter or the 'timeout' extra of "
+                    f"connection {self.airbyte_conn_id!r}"
+                )
+
         return AirbyteAPI(
             server_url=self.conn["host"],
             security=security,
             client=client,
+            timeout_ms=timeout_ms,
         )
 
     @classmethod
--- a/providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
+++ b/providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
@@ -114,11 +114,24 @@ class AirbyteHook(BaseHook):
                 self.airbyte_conn_id,
             )
 
-        client = None
+        timeout = self.timeout if self.timeout is not None else self.conn["timeout"]
+        error_message = (
+            f"Invalid Airbyte API request timeout {timeout!r}: expected a positive number of "
+            f"seconds, set via the AirbyteHook 'timeout' parameter or the 'timeout' extra of "
+            f"connection {self.airbyte_conn_id!r}"
+        )
+        if timeout is not None:
+            try:
+                timeout = float(timeout)
+            except (TypeError, ValueError) as e:
+                raise ValueError(error_message) from e
+            if timeout <= 0:
+                raise ValueError(error_message)
+
+        mounts: dict[str, httpx.HTTPTransport] = {}
         if self.conn["proxies"]:
             self.log.debug("Creating client proxy...")
             proxies = self.conn["proxies"]
-            mounts = {}
             if isinstance(proxies, dict):
                 for scheme, proxy_url in proxies.items():
                     # httpx mount keys require a "://" suffix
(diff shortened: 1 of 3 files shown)


filipeaaoliveira:
Hello!

This is a gentle ping on this @potiuk 

I checked the last Airbyte provider PR and saw that you were the reviewer, hence why I am pinging you.

Thank you

potiuk (commented):
Thanks for this — the write-up is unusually good: the diagnosis (airbyte-api 1.x moving to httpx and inheriting its 5s default), the non-idempotent create-job making retries produce duplicate syncs, and the constraint resolution that drags 5.5.1 into the same hole are all accurate. I checked the plumbing against `airbyte-api` 1.x and it does what you describe: `timeout_ms` lands in `sdk_configuration.timeout_ms`, each operation falls back to it, and `build_request()` uses `httpx.USE_CLIENT_DEFAULT` when it is `None` — so the "unchanged when unset" claim holds, and it works with the proxy-mounted client too.

One gap that matters for your own use case, and a small style point.

### The OAuth token request is not covered by the timeout

`timeout_ms` only applies to API operations. The SDK's client-credentials hook fetches the access token itself:

```python
# airbyte_api/_hooks/clientcredentials.py
response = self.client.send(
    self.client.build_request(method="POST", url=token_url, data=payload)
)
```

No timeout override there, so that request keeps the httpx client default of 5 seconds regardless of what is configured. On exactly the loaded deployment this PR targets, `submit_sync_connection` can still fail with `httpx.ReadTimeout` — just on the token call rather than on `POST /v1/jobs` — and you are back to the duplicate-sync problem.

Setting the timeout on the httpx client covers both paths, and drops the millisecond conversion entirely:

```python
if self.conn["proxies"] or timeout is not None:
    client = httpx.Client(mounts=mounts, timeout=timeout if timeout is not None else 5.0)
```

Keeping `timeout_ms` as well is harmless if you prefer belt-and-braces, but the client-level timeout is the part that makes the token fetch safe.

### Smaller observations

- `hooks/airbyte.py:139-141` — the `except` sets `timeout_ms = 0` so the `<= 0` check below raises for both parse failures and non-positive values. It works, but reads as a trick; raising the `ValueError` directly in the `except` and keeping the range check separate says the same thing more plainly.
- The timeout is reachable only from the hook constructor or the connection extra, not from `AirbyteTriggerSyncOperator`. The extra covers the common deployment-wide case and the docs say so, so this is fine as-is — flagging only in case you want the operator to take it too.

The rest looks right: `ValueError` rather than `AirflowException`, docs updated alongside the code, no newsfragment (correct — provider changelogs are regenerated from git log), and the tests parametrize both the accepted and the rejected shapes.

---

>

---

providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
```diff
@@ -124,10 +131,25 @@ def create_api_session(self) -> AirbyteAPI:
...
+                    f"Invalid Airbyte API request timeout {timeout!r}: expected a positive number of "
+                    f"seconds, set via the AirbyteHook 'timeout' parameter or the 'timeout' extra of "
+                    f"connection {self.airbyte_conn_id!r}"
+                )
+
         return AirbyteAPI(
             server_url=self.conn["host"],
             security=security,
             client=client,
+            timeout_ms=timeout_ms,
```
potiuk: This covers the API operations, but not the OAuth token fetch: the SDK's client-credentials hook does `self.client.send(self.client.build_request("POST", token_url, ...))` with no timeout override, so that call keeps httpx's 5s default. On the loaded deployment you describe, the trigger can still fail — just on the token request instead of on `POST /v1/jobs`.

Setting the timeout on the client itself covers both paths and removes the need for the ms conversion:

```python
if self.conn["proxies"] or timeout is not None:
    client = httpx.Client(mounts=mounts, timeout=timeout if timeout is not None else 5.0)
```

providers/airbyte/src/airflow/providers/airbyte/hooks/airbyte.py
```diff
@@ -124,10 +131,25 @@ def create_api_session(self) -> AirbyteAPI:
                 }
             client = httpx.Client(mounts=mounts)
 
+        timeout = self.timeout if self.timeout is not None else self.conn["timeout"]
+        timeout_ms: int | None = None
+        if timeout is not None:
+            try:
+                timeout_ms = int(float(timeout) * 1000)
+            except (TypeError, ValueError):
+                timeout_ms = 0
```
potiuk: Minor style point: setting `timeout_ms = 0` here so the `<= 0` check below raises works, but it routes a parse failure through a sentinel value. Raising the `ValueError` directly in the `except`, and keeping the range check separate, reads more plainly.

filipeaaoliveira:
> Thanks for this — the write-up is unusually good: the diagnosis (airbyte-api 1.x moving to httpx and inheriting its 5s default), the non-idempotent create-job making retries produce duplicate syncs, and the constraint resolution that drags 5.5.1 into the same hole are all accurate. I checked the plumbing against `airbyte-api` 1.x and it does what you describe: `timeout_ms` lands in `sdk_configuration.timeout_ms`, each operation falls back to it, and `build_request()` uses `httpx.USE_CLIENT_DEFAULT` when it is `None` — so the "unchanged when unset" claim holds, and it works with the proxy-mounted client too.
> 
> One gap that matters for your own use case, and a small style point.
> 
> ### The OAuth token request is not covered by the timeout
> `timeout_ms` only applies to API operations. The SDK's client-credentials hook fetches the access token itself:
> 
> ```python
> # airbyte_api/_hooks/clientcredentials.py
> response = self.client.send(
>     self.client.build_request(method="POST", url=token_url, data=payload)
> )
> ```
> 
> No timeout override there, so that request keeps the httpx client default of 5 seconds regardless of what is configured. On exactly the loaded deployment this PR targets, `submit_sync_connection` can still fail with `httpx.ReadTimeout` — just on the token call rather than on `POST /v1/jobs` — and you are back to the duplicate-sync problem.
> 
> Setting the timeout on the httpx client covers both paths, and drops the millisecond conversion entirely:
> 
> ```python
> if self.conn["proxies"] or timeout is not None:
>     client = httpx.Client(mounts=mounts, timeout=timeout if timeout is not None else 5.0)
> ```
> 
> Keeping `timeout_ms` as well is harmless if you prefer belt-and-braces, but the client-level timeout is the part that makes the token fetch safe.
> 
> ### Smaller observations
> * `hooks/airbyte.py:139-141` — the `except` sets `timeout_ms = 0` so the `<= 0` check below raises for both parse failures and non-positive values. It works, but reads as a trick; raising the `ValueError` directly in the `except` and keeping the range check separate says the same thing more plainly.
> * The timeout is reachable only from the hook constructor or the connection extra, not from `AirbyteTriggerSyncOperator`. The extra covers the common deployment-wide case and the docs say so, so this is fine as-is — flagging only in case you want the operator to take it too.
> 
> The rest looks right: `ValueError` rather than `AirflowException`, docs updated alongside the code, no newsfragment (correct — provider changelogs are regenerated from git log), and the tests parametrize both the accepted and the rejected shapes.
> 
> > confirmed by an Airflow maintainer. The findings
> > below are observations, not blockers; an Airflow
> > maintainer — a real person — will take the next look at the
> > PR. If you think a finding is mis-applied, please reply on
> > the PR and a maintainer will weigh in._
> 

Hello @potiuk ! Thank you so much for the review.

You're completely right on both things and I've updated them like you recommended. Thank you once again for taking the time to review and make suggestions. It looks and it IS way better this way.
One thing that I added was that on the client, it now passes `follow_redirects=True` to match the default client the SDK creates when none is supplied.

I hope this is better now, but if you have any more suggestions, please let me know.

filipeaaoliveira:
Is it possible to get a new review done here when you have the time @potiuk ?

Thank you!

PR #71095, opened 2026-08-04
Fix airflowctl dags state crash from an incomplete backport

Lee-W:
The `dags state` backport (#69915, #70352) brought the caller and the tests but dropped `suppress_error_log` from `DagRunOperations.list()`, which main has. `dag_command.py:147` passes it, so the command raised:

```
TypeError: DagRunOperations.list() got an unexpected keyword argument 'suppress_error_log'
```

Restore the keyword-only parameter and forward it to the client as the `airflowctl_suppress_error_log` extension, matching the sibling `get()`.

The broken caller only exists on this test branch, so no released airflowctl is affected and no newsfragment is needed.

Split out of #70724 so that PR stays workflow-only.

---

- [X] Yes — Claude Opus 5


jason810496 (approved):
Thanks. I double checked with the upstream main.

PR #67294, opened 2026-05-21
airflow-ctl 0.1.5: test to stable

bugraoz93:
Sync airflow-ctl/v0-1-test into airflow-ctl/v0-1-stable for the airflow-ctl 0.1.5 release.

Need backport of (depends): #67295 merged to backport branch. Then it will merge. But ready for review as the only version will lend

PR #71096, opened 2026-08-04
Fix airflow-ctl mypy static check on the 0.1 branch

Lee-W:
The same test-to-stable sync (#67294) repointed the `mypy-airflow-ctl` hook at `../scripts/ci/prek/run_mypy_full_dist_local_venv_or_breeze_in_ci.py`, which only exists on main. On this branch prek dies before mypy ever runs:

```
error: Failed to run hook `mypy-airflow-ctl`
  caused by: Run command `python hook` failed
  caused by: No such file or directory (os error 2)
```

Point it back at `mypy_local_folder.py`, which is what airflow-core, task-sdk, devel-common and dev/scripts all use on this branch.

Split out of #70724 so that PR stays workflow-only.

---

- [X] Yes — Claude Opus 5

### DIFF
--- a/airflow-ctl/.pre-commit-config.yaml
+++ b/airflow-ctl/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
       - id: mypy-airflow-ctl
         name: Run mypy for airflow-ctl
         language: python
-        entry: ../scripts/ci/prek/run_mypy_full_dist_local_venv_or_breeze_in_ci.py airflow-ctl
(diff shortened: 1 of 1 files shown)


jason810496 (approved):
Thanks! Perhaps we can update the `airflow/.github/boring-cyborg.yml` as follow-up?

https://github.com/apache/airflow/blob/9a85faa9c55b907bd105ee41567bb7d55dab9f31/.github/boring-cyborg.yml#L365-L390

Lee-W:
yep, sound good, will create a seprate PR for that

PR #71101, opened 2026-08-04
Fix upstream failure branch in task lifecycle diagram

YAshhh29:
While I was tracing the task lifecycle, I noticed that the branches after
"Required upstream task(s) failed?" are reversed in the diagram.

A `YES` answer currently continues through the normal scheduling checks,
while `NO` leads to `upstream_failed`. It should be the other way around:
when a required upstream task has failed, the current task moves to
`upstream_failed`; otherwise, the Scheduler continues evaluating it.

This PR swaps those two labels and regenerates the PNG and its source
checksum.

Validation:

- `prek run generate-airflow-diagrams --files airflow-core/docs/img/diagram_task_lifecycle.py`
- `prek run ruff --files airflow-core/docs/img/diagram_task_lifecycle.py`

---

- [X] Yes - GitHub Copilot

The PNG was regenerated deterministically from the diagram source using Airflow's Graphviz hook.


uros-b (approved):

uros-b:
Thank you @YAshhh29!

PR #71104, opened 2026-08-04
Validate provider metadata list entries instead of ignoring them

Lee-W:
## Why
https://github.com/apache/airflow/pull/70190#discussion_r3682598659

The items blocks for `queues, plugins and task-decorators` listed their field names directly under items, where they are not JSON Schema keywords, so the blocks were inert and any shape passed. A provider author following what those blocks appeared to document — a dict for queues, or a path key for a decorator — got no error and was quietly dropped by the consumers instead.

## What
Pinning each section to the shape the provider.yaml files and generated payloads actually use turns that into a failure at authoring time. The custom-provider howto named the same nonexistent path field.

 

---

---

### DIFF
--- a/airflow-core/src/airflow/provider.yaml.schema.json
+++ b/airflow-core/src/airflow/provider.yaml.schema.json
@@ -576,6 +576,7 @@
                         "type": "string"
                     }
                 },
+                "additionalProperties": false,
                 "required": [
                     "name",
                     "class-name"
@@ -708,6 +709,7 @@
                         "type": "string"
                     }
                 },
+                "additionalProperties": false,
                 "required": [
                     "name",
                     "plugin-class"
--- a/airflow-core/src/airflow/provider.yaml.schema.json
+++ b/airflow-core/src/airflow/provider.yaml.schema.json
@@ -565,12 +565,21 @@
             "type": "array",
             "description": "Decorators to use with the TaskFlow API. Can be accessed by users via '@task.<name>'",
             "items": {
-                "name": {
-                    "type": "string"
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "description": "Name the decorator is exposed under, following '@task.'",
+                        "type": "string"
+                    },
+                    "class-name": {
+                        "description": "Class name that implements the decorator",
+                        "type": "string"
+                    }
                 },
-                "path": {
-                    "type": "string"
-                }
+                "required": [
+                    "name",
+                    "class-name"
+                ]
             }
         },
         "secrets-backends": {
@@ -688,24 +697,28 @@
             "type": "array",
             "description": "Plugins exposed by the provider",
             "items": {
-                "name": {
-                    "type": "string"
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "description": "Name of the plugin",
+                        "type": "string"
+                    },
+                    "plugin-class": {
+                        "description": "Class name that implements the plugin",
+                        "type": "string"
+                    }
                 },
-                "plugin-class": {
-                    "type": "string"
-                }
+                "required": [
+                    "name",
+                    "plugin-class"
+                ]
             }
         },
         "queues": {
             "type": "array",
-            "description": "Message Queues exposed by the provider",
+            "description": "Message queue provider class names",
             "items": {
-                "name": {
-                    "type": "string"
-                },
-                "message-queue-class": {
-                    "type": "string"
-                }
+                "type": "string"
             }
         },
         "source-date-epoch": {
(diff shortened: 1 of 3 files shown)


airflow-core/src/airflow/provider_info.schema.json
```diff
@@ -432,38 +432,49 @@
...
-            "description": "Message Queues exposed by the provider",
+            "description": "Message queue provider class names",
             "items": {
-                "name": {
-                    "type": "string"
-                },
-                "message-queue-class": {
-                    "type": "string"
-                }
+                "type": "string"
```
kaxil: This is the runtime schema, and `discover_all_providers_from_packages` documents it as deliberately looser than the dev one: "The runtime version is more relaxed (allows for additional properties) and verifies only the subset of fields that are needed at runtime".

The [`provider_schema_validator.validate(provider_info)`](https://github.com/apache/airflow/blob/e69c1881b32c36abb827bae3717eaf46424427bd/shared/providers_discovery/src/airflow_shared/providers_discovery/providers_discovery.py#L325) call there isn't wrapped, and neither caller wraps it ([`ProvidersManager.initialize_providers_list`](https://github.com/apache/airflow/blob/e69c1881b32c36abb827bae3717eaf46424427bd/airflow-core/src/airflow/providers_manager.py#L524) and the Task SDK's [`providers_manager_runtime.py`](https://github.com/apache/airflow/blob/e69c1881b32c36abb827bae3717eaf46424427bd/task-sdk/src/airflow/sdk/providers_manager_runtime.py#L198)). So an installed third-party provider emitting the old dict shape moves from a logged warning, where [`_correctness_check`](https://github.com/apache/airflow/blob/e69c1881b32c36abb827bae3717eaf46424427bd/airflow-core/src/airflow/providers_manager.py#L345) catches the `import_string` failure and skips just that entry, to a ValidationError that aborts discovery so no provider loads at all in that process, workers included.

Its queue entry is already broken either way, so this is about the blast radius: one dead queue entry becomes zero providers. Tightening `provider.yaml.schema.json` alone catches the authoring mistake without that. The `required` blocks above are a milder version of the same thing, since a wrong key there already raises `KeyError` in `_discover_plugins`/`_discover_taskflow_decorators`, though after this change it takes out the whole provider list rather than just plugin discovery. `queues` is the one that goes from survivable to fatal.

airflow-core/src/airflow/provider.yaml.schema.json
```diff
@@ -557,12 +557,21 @@
...
+                    },
+                    "class-name": {
+                        "description": "Class name that implements the decorator",
+                        "type": "string"
+                    }
                 },
-                "path": {
-                    "type": "string"
-                }
+                "required": [
```
kaxil, line 580: Worth adding `"additionalProperties": false` here and on the `plugins` block below, otherwise a stray `path:` sitting next to a correct `class-name:` still validates, which is the confusion this PR is closing. The top level of this file already sets it, as do most of the object item blocks above. I checked and no in-tree `provider.yaml` breaks with it added. Leaving `provider_info.schema.json` alone is right, since the runtime schema is documented as allowing extra properties.

kaxil: Let's check with @potiuk too, he might have stronger opinion on this

airflow-core/src/airflow/provider_info.schema.json
Lee-W: Dropped the `required` blocks and the `queues` `items` constraint from the runtime schema, so nothing in there can fail validation that passed before. the old `items` blocks listed the field names as bare keys, which JSON Schema ignores, so they described validation that was never actually running. They're now real `properties` with no `required`, purely as documentation. The `plugins` and `task-decorators` items do still carry `type: object` — happy to drop that too if you'd rather the runtime schema constrain nothing at all.

airflow-core/src/airflow/provider.yaml.schema.json
Lee-W: Added on both. Ran the `provider.yaml` check across the tree to confirm it stays green, and putting a stray key back into one of the entries does make it fail on that entry, so the guard is actually live. The field sets used by every in-tree `provider.yaml` match what the schema declares, so nothing gets caught by accident.

potiuk: Good call @kaxil  The "additionalProperties" : false "locks" the API - because we want to have providers install on past version of airflow. Those checks are done on runtime, so it means that "additionalProperties" : true - should only be used when we now **for sure** future version of the schema will not add anything there.

But since you added it only in provider.yaml.schema.json - that's OK - because it does not impact provider_info.schema.json - this is OK.

potiuk (approved):

PR #71112, opened 2026-08-04
Skip maxed out queued DagRun candidates

fat-catTW:
## Why:
When `_start_queued_dagruns()` processes queued DagRun candidates, one candidate can fill the remaining `max_active_runs` capacity for a Dag or backfill. The scheduler already knows later candidates for the same `(dag_id, backfill_id)` cannot start in the same pass, but it still loaded the serialized Dag and re-checked the same limit for each one.

## Solution:
Remember `(dag_id, backfill_id)` keys that reach their active run limit during the scheduler pass. Later queued DagRun candidates for the same key are skipped immediately, avoiding repeated serialized Dag loads and duplicate limit checks while preserving the existing `max_active_runs` behavior.

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2840,46 +2840,95 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
         # Team name should be added before listeners are called in notify_dagrun_state_changed()
         self._stamp_team_names(dag_runs, session)
 
+        dag_run_keys_at_limit: set[tuple[str, int | None]] = set()
+
+        def _mark_dag_run_key_at_limit(
+            dag_run_key: tuple[str, int | None],
+            *,
+            active_runs: int,
+            max_active_runs: int,
+            dag_id: str,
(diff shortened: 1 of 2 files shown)


steveahnahn (commented):
Nice work! `cached_get_dag` is keyed on the `DagRun` object rather than the Dag, so every candidate was a genuine miss with 3 `get_dag_for_run` calls down to 1 on your test.

One note: the two `if active_runs >= max_active_runs` log branches are nearly dead after this. `get_queued_dag_runs_to_set_running` already drops runs for Dags sitting at their limit, so they only fired when the limit filled mid-loop, which your new marking short-circuits. On your scenario `main` logs twice and this PR logs nothing. Worth either moving that line to where you mark the key, or dropping the branches.

Also, the backfill path got the same marking and has no test.

fat-catTW:
Thanks, good catch.  :D
I moved the logging into the shared key-marking path so the scheduler logs when a Dag/backfill reaches the limit and the remaining candidates will be skipped in the current pass. I kept the guard branches since they still cover concurrent/stale-count cases, but they now use the same helper.

I also added a backfill regression test for the same pruning behavior.

PR #71113, opened 2026-08-04
Stop backfill endpoints disclosing ids across Dags you cannot see

### LINKED ISSUE #71080: Backfill authorization falls back to the request body when the backfill does not exist
`requires_access_backfill` resolves the Dag for the authorization decision from the `backfill_id` path parameter. When that id is present and parses, but no `Backfill` row matches, `dag_id` stays `None` and the dependency falls through to reading `dag_id` from the request body — so the decision is made against a Dag the *request names* rather than the backfill in the path.

This is not exploitable for a cross-Dag action: the handler answers 404 before acting, so nothing is done to another Dag's backfill. What it does leak is existence. A caller can tell "no such backfill" (404) apart from "backfill exists on a Dag you cannot see" (403), which is an existence oracle over backfill ids across Dags.

### Why the fallback is there

`requires_access_backfill` is not only on `/{backfill_id}` routes — it also guards `GET /backfills`, `POST /backfills` and `POST /backfills/dry_run`, where there is no `backfill_id` at all and the Dag legitimately comes from the body. Any fix has to keep those working.

### Suggested shape

Distinguish "no `backfill_id` in the path" from "`backfill_id` given but not found". The body fallback should apply only to the first. In the second the dependency should decide without consulting the body, and the response for an unknown backfill should be indistinguishable from the unauthorized one.

### Acceptance criteria

- A request with a well-formed but unknown `backfill_id` never has its authorization decided by a body-supplied `dag_id`.
- An unauthorized caller cannot distinguish an unknown backfill id from one on a Dag they cannot access.
- The three body-based routes above keep authorizing off the body.

Raised in review of #70889, which deliberately scoped itself to the parser divergence: https://github.com/apache/airflow/pull/70889#discussion_r3710514595

---

rjgoyln:
## Summary

The five backfill routes that name a backfill in their path fell back to the `dag_id` supplied on the request — body or query string — whenever the id matched no row, so an unknown id and a backfill on a Dag the caller cannot see answered differently, 404 against 403. That difference enumerates backfill ids across Dags.

The path names the backfill, so its row is now the only thing those routes authorize against, and an id the caller may not see answers exactly as a missing one does. A caller who may read the Dag keeps the Forbidden answer on a write they are not allowed: they can already list that Dag's backfills, so hiding it would only cost them the reason their request was refused. The list, create and dry-run routes name no backfill in their path and keep authorizing off the body.

## Behavior change

On those five routes a Dag the caller cannot read now answers 404 rather than 403.

closes: #71080

---


airflow-core/src/airflow/api_fastapi/core_api/security.py
```diff
@@ -396,6 +396,30 @@ def depends_readable_event_logs_filter(
...
 
+_BACKFILL_NOT_FOUND = "Backfill not found"
+
+
+def _authorize_backfill_in_path(method: ResourceMethod, dag_id: str | None, user: BaseUser) -> None:
+    """Authorize a backfill named by the request path against that backfill's Dag alone."""
+    # ``dag_id`` is None when the id matched no row. Answering 404 there while a backfill on a Dag
+    # the caller may not read answers 403 would tell them which backfill ids exist across Dags.
+    if dag_id is None:
+        raise HTTPException(status.HTTP_404_NOT_FOUND, _BACKFILL_NOT_FOUND)
```
potiuk: Because this fires before the handler, the 404 *detail* an authorized caller sees for an unknown id also changed: `/dag_runs` went from `Backfill with id 999 not found` and pause/unpause/cancel from `Could not find backfill with id 999`, both now `Backfill not found`. Only `get_backfill` already used this wording.

Worth a line in the newsfragment — a client matching on `detail` breaks silently otherwise. It also leaves the four `raise HTTPException(404, …)` branches in `backfills.py` reachable only via a TOCTOU delete between this lookup and the handler's; pointing them at `_BACKFILL_NOT_FOUND` too would collapse three strings for one condition into one.

airflow-core/src/airflow/api_fastapi/core_api/security.py
```diff
@@ -396,6 +396,30 @@ def depends_readable_event_logs_filter(
...
+
+    details = DagDetails(id=dag_id, team_name=DagModel.get_team_name(dag_id))
+    auth_manager = get_auth_manager()
+    if auth_manager.is_authorized_dag(
+        method=method, access_entity=DagAccessEntity.RUN, details=details, user=user
+    ):
+        return
+    # A caller who may read the Dag can already list its backfills, so the id is no secret from
+    # them: hiding it would only cost them the reason their request was refused.
+    if method != "GET" and auth_manager.is_authorized_dag(
```
potiuk, line 419: This is a second auth-manager round-trip on the denied non-GET path. `batch_is_authorized_dag` covers the "method + GET" pair in one call — that's what `requires_access_dag_from_file_token` uses. Bounded cost since it's only the refused path, but a non-caching auth manager pays two backend calls per refused write.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_backfills.py
```diff
@@ -80,6 +82,21 @@ def clean_db():
     _clean_db()
 
 
+@pytest.fixture
+def dag_reader_test_client(test_client):
```
potiuk, line 86: This duplicates `unauthorized_test_client` (`airflow-core/tests/unit/api_fastapi/conftest.py:174`) with `role="viewer"` instead of `role=None`, token signer and `RevokedToken.is_revoked` patch included. Three client fixtures now differ only by role — a `role_test_client` factory in `conftest.py` would fold all three. Fine as a local fixture for this PR; flagging the pattern.

potiuk (approved):
This is a clean fix and it meets all three acceptance criteria from #71080 — the path id is the only authorization subject on those five routes, unknown and unreadable are indistinguishable, and list/create/dry-run keep authorizing off the request. Nice touches: all five routes already declared `404` in `create_openapi_http_exception_doc`, so nothing needs regenerating, and the lookup narrowed from a full `Backfill` row to `select(Backfill.dag_id)`.

Approving. One thing worth a follow-up commit here, and two nits you can take or leave.

### The 404 detail string changed on four routes too — worth a newsfragment line

See the inline comment on `security.py`. The newsfragment covers 403 → 404 for unreadable Dags, but for an *authorized* caller the message for an unknown id also changed, because the dependency now answers before the handler runs:

| Route | Before | After |
| --- | --- | --- |
| `GET /backfills/{id}` | `Backfill not found` | unchanged |
| `GET /backfills/{id}/dag_runs` | `Backfill with id 999 not found` | `Backfill not found` |
| `PUT .../pause`, `.../unpause`, `.../cancel` | `Could not find backfill with id 999` | `Backfill not found` |

Nothing caught it because `test_list_backfill_dag_runs_not_found` asserts only the status code and the three `PUT` routes have no unknown-id detail assertion.

### Smaller observations

- See the inline comment on `security.py` about the second `is_authorized_dag` call.
- See the inline comment on the new `dag_reader_test_client` fixture.

---

> will follow up.*
>

---

rjgoyln:
> This is a clean fix and it meets all three acceptance criteria from #71080 — the path id is the only authorization subject on those five routes, unknown and unreadable are indistinguishable, and list/create/dry-run keep authorizing off the request. Nice touches: all five routes already declared `404` in `create_openapi_http_exception_doc`, so nothing needs regenerating, and the lookup narrowed from a full `Backfill` row to `select(Backfill.dag_id)`.
> 
> Approving. One thing worth a follow-up commit here, and two nits you can take or leave.
> 
> ### The 404 detail string changed on four routes too — worth a newsfragment line
> See the inline comment on `security.py`. The newsfragment covers 403 → 404 for unreadable Dags, but for an _authorized_ caller the message for an unknown id also changed, because the dependency now answers before the handler runs:
> 
> Route	Before	After
> `GET /backfills/{id}`	`Backfill not found`	unchanged
> `GET /backfills/{id}/dag_runs`	`Backfill with id 999 not found`	`Backfill not found`
> `PUT .../pause`, `.../unpause`, `.../cancel`	`Could not find backfill with id 999`	`Backfill not found`
> Nothing caught it because `test_list_backfill_dag_runs_not_found` asserts only the status code and the three `PUT` routes have no unknown-id detail assertion.
> 
> ### Smaller observations
> * See the inline comment on `security.py` about the second `is_authorized_dag` call.
> * See the inline comment on the new `dag_reader_test_client` fixture.
> 
> > confirmed by an Airflow maintainer. The maintainer
> > approving this PR has read the findings and signed off. If
> > something feels off, please reply on the PR and a maintainer
> > will follow up._
> 

Thanks — the detail string was the part I missed, and it is a real break for anyone matching on it.

Follow-up pushed: all five routes now raise `security.BACKFILL_NOT_FOUND` (the constant lost its underscore so the routes can share it), so the one condition has one message whether the dependency or the handler answers, and the newsfragment records the change with the old strings. I also added the missing unknown-id detail assertions on `/dag_runs`, `unpause` and `cancel`.

On the second `is_authorized_dag` call: I left it as two calls because the decision is three-way rather than a single gate — allow, `403` when the caller may read the Dag but not write it, `404` otherwise. `batch_is_authorized_dag` collapses the set to one boolean, so a `False` still would not say which of the two failed, and asking for `method` and `GET` together would turn a can-edit-without-can-read role from allowed into refused, which the allow-first ordering deliberately permits. Happy to revisit if you would rather the pair were required.

On the fixture: agreed, and I have kept it local here. A `role_test_client` factory in `conftest.py` folding all three is a tidy separate change — I can open it if that is useful.

pierrejeambrun (approved):
LGTM

Issue #70522, opened 2026-07-27
Consider swapping `httpx` usage for `httpx2`


zach-overflow:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.x

### What happened and how to reproduce it?

At the time of writing, it appears that the `httpx` library might not be actively maintained. There are a number of bug-fixes for `httpx` which have not landed, but which are present in [`httpx2`](https://httpx2.pydantic.dev/) (forked from `httpx==0.28.1`).

Additionally, the maintainer of `starlette` (one of the key sub-dependencies of `FastAPI`) [has confirmed](https://github.com/pydantic/httpx2/discussions/980#discussioncomment-17030187) that starlette will migrate to `httpx2`. So it seems the migration from `httpx` in `FastAPI` is an eventuality.

### What you think should happen instead?

Any direct usage of `httpx` within airflow-core / the task-sdk / airflow's providers should probably migrate to httpx2. 

The scope of this change is obviously large; but the migration of direct code reliance on `httpx` is [relatively straightforward](https://httpx2.pydantic.dev/migration/).

I'd imagine it would be good to add individual, well-defined / scoped sub-issues for this.

### Official Helm Chart version

Not Applicable

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

potiuk:
I guess until starlette migrates, it makes little sense for us to keep both dependencies - so I would rather wait for them - maybe someone can open a PR doing the migration to test it and see if there are any compatibility issues - if you would like to do it - feel free.

Kludex:
I've migrated Starlette.

zach-overflow:
I was curious to see if / how FastAPI was migrating to `httpx2`. Currently, it [looks like](https://github.com/fastapi/fastapi/discussions/16186) there might be some work towards a gradual migration. 

Given the split components of Airflow (core, sdk, providers), I think it might make sense to address each individually. My hunch is that this would be a good way to split out the migration:
* `airflow-core`: Temporarily support _both_ `httpx` and `httpx2`, with some tentative future release targetting the removal of `httpx`.
* `task-sdk`: It looks like this may be able to fully cutover in a single release.
* `airflow-providers`: Looking into this one currently, hoping to put up a PR for it if there aren't any unexpected issues.

[...]

PR #71130, opened 2026-08-04
Migrate task-sdk from httpx to httpx2

art22s:
## Summary

This PR migrates the task-sdk package from httpx to httpx2 as a proof-of-concept to test compatibility and identify any issues, as suggested by @potiuk in #70522.

It also updates the Airflow core in-process execution API to use the matching httpx2 sync and async transports.

### Background

The httpx library may not be actively maintained, and httpx2 (a fork of httpx 0.28.1) has been created with bug fixes and improvements. The maintainer of starlette has confirmed that starlette will migrate to httpx2, making the migration from httpx in FastAPI an eventuality.

### Approach

Following the httpx2 migration guide (https://httpx2.pydantic.dev/migration/), this PR uses direct httpx2 imports in the migrated code:

- Keeps httpx and httpx2 available together so packages that still depend on httpx are unaffected
- Allows incremental migration — each package or subpackage can be migrated independently
- Uses httpx2 transport implementations in the in-process execution API so client and transport types come from the same package

### Changes

- task-sdk/pyproject.toml: Changed the dependency from httpx>=0.27.0 to httpx2>=0.28.1
- task-sdk: Migrated API client, supervisor, tests, and model-generation imports from httpx to httpx2
- airflow-core/pyproject.toml: Added httpx2>=0.28.1 while retaining httpx for existing users
- airflow-core/src/airflow/api_fastapi/execution_api/app.py: Changed the in-process API to use httpx2.WSGITransport and httpx2.ASGITransport
- airflow-core/tests/unit/api_fastapi/execution_api/test_app.py: Added coverage for the httpx2 sync and async transports
- uv.lock: Updated dependency resolution for both packages

### Notes

- certifi usage in client.py is retained as-is — it is used directly for SSL context configuration, not through httpx, and is independent of the httpx/httpx2 choice
- The httpx2 package has the same public API as httpx 0.28.1, with direct imports used here rather than an alias
- Known behavior differences in httpx2: SSL verification uses the OS trust store via truststore instead of certifi bundled certificates, and the User-Agent header changes to python-httpx2/<version>

## Test plan

- [x] Focused task-sdk API client tests: 127 passed
- [x] Focused in-process execution API tests: 27 passed
- [x] Verify httpx2 installs cleanly alongside httpx: uv lock --check passed
- [x] Verify Client, MockTransport, and transport behavior through the focused tests
- [ ] Full task-sdk test suite
- [ ] CI checks pass

- [X] Yes (Claude) following the guidelines (https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


task-sdk/pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
     "attrs>=24.2.0, !=25.2.0",
     "babel>=2.17.0",
     "fsspec>=2023.10.0",
-    "httpx>=0.27.0",
+    "httpx2>=0.28.1",
```
Prab-27, line 58: Thanks ! 

`Strletter` has migrated to `httpx2` but We could also keep both` httpx `and` httpx2` as dependency to save any breaking changes risk . WDYT ? 

And it would be good if we import directly `htttpx2` not as aliased since it can break on typeidentity /logger side - I have read [here](https://httpx2.pydantic.dev/migration/)

Would love to hear from the community !

art22s:
Thanks for the review @Prab-27!

Good points — I'll address both:

1. **Keep both dependencies**: Agreed, keeping `httpx` alongside `httpx2` in `task-sdk` is safer. Other packages (airflow-core, providers) still depend on `httpx`, so having both in `task-sdk` avoids any risk of breaking transitive resolution.

2. **Direct import instead of aliasing**: Fair point on type identity. The migration guide does recommend direct imports for better type checker and logger compatibility. I'll switch from `import httpx2 as httpx` to `import httpx2` and update the references accordingly.

Pushing the changes shortly.

art22s:
Both review points from @Prab-27 are addressed in the latest commits — kept `httpx` alongside `httpx2` in task-sdk dependencies and switched to direct `httpx2` imports (no aliasing), plus follow-up fixes for transport compatibility and tests. Branch rebased onto current `main`. Ready for re-review 🙏

Issue #52712, opened 2025-07-02
AIP-82: Add common messaging system providers


jason810496:
### Body

related:  #49938

## Why

Several users have pointed out that the Event-Driven Scheduling feature introduced in Airflow 3 currently supports only Amazon SQS.

So I add Kafka common message queue provider in #49938, but there are still quite a few common providers that should be useful for airflow users to implement Event-Driven workflows.

## What

Here are some common messaging systems:

- [ ] Amazon Kinesis
- [ ] Amazon MSK
- [x] Amazon SQS
- [x] Apache Kafka
- [ ] Azure Message Bus
  - [x] Trigger: https://github.com/apache/airflow/pull/53356
  - [x] CommonMessageQueue Interface
- [x] Google PubSub https://github.com/apache/airflow/pull/54494
- [x] Redis https://github.com/apache/airflow/pull/52917

reference for implementation: https://github.com/apache/airflow/pull/49938/files

### Committer

- [x] I acknowledge that I am a maintainer/committer of the Apache Airflow project.

400Ping:
Hi, may I try the pubsub one?

Jasperora:
Hi, I would like to try the Redis one. Thanks!

prdai:
hi, could I implement the "Azure Message Bus" one? thanks :)

jason810496:
Thanks all for making event-driven scheduling feature better!

potiuk:
Love it 👀

mingdaoy:
I's like to work on Amazon Kinesis, MSK

vikramkoka:
Love the energy and progress here!

jroachgolf84:
@mingdaoy, are you still working on the two AWS services? If not, I'm happy to create a PR for that.

mingdaoy:
@jroachgolf84 Still working on them and making progress!

dejii:
Hi everyone, I ended up building the messaging system for Google Pub/Sub as part of a personal project, so I now have a working implementation. I noticed that @400Ping expressed interest a while back, but there hasn’t been any activity since. I’ll go ahead and open a PR shortly to move the feature forward — happy to compare notes and collaborate if there’s any existing work in progress.

400Ping:
[...]

PR #71135, opened 2026-08-05
Add Kinesis Data Streams trigger

aaron-y-chen:
related: #52712

This is the first PR for Kinesis part of #52712, I will integrate the trigger with the common-messaging interface and add a system/e2e test in the next PR.

## Summary

Add an asynchronous Kinesis Data Streams trigger for consuming records without occupying worker slots.

The trigger polls all shards, emits JSON-serializable record batches, handles iterator expiry, throttling, pagination, and in-run resharding, and checkpoints per-shard sequence numbers when an asset state store is available. New shards first discovered after a restart honor the configured initial position, so `LATEST` may skip records written during the downtime.

This is the provider-side foundation for the common-messaging integration tracked in #52712.

---

Generated-by: [GPT 5.6-sol] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
@@ -1,0 +1,315 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+
+import asyncio
+import base64
+import hashlib
+import json
+from collections.abc import AsyncIterator
+from typing import TYPE_CHECKING, Any
+
+from airflow.providers.amazon.aws.hooks.kinesis import KinesisHook
+from airflow.providers.amazon.version_compat import AIRFLOW_V_3_0_PLUS
+
+if AIRFLOW_V_3_0_PLUS:
+    from airflow.triggers.base import BaseEventTrigger, TriggerEvent
+else:
+    from airflow.triggers.base import (  # type: ignore
+        BaseTrigger as BaseEventTrigger,
+        TriggerEvent,
+    )
+
+if TYPE_CHECKING:
+    from airflow.providers.amazon.aws.hooks.base_aws import BaseAwsConnection
+
+_CHECKPOINT_KEY_PREFIX = "kinesis_shard_sequence_numbers"
+_ITERATOR_TYPES_WITHOUT_EXTRA_ARGS = frozenset({"LATEST", "TRIM_HORIZON"})
+
+
+class KinesisTrigger(BaseEventTrigger):
+    """
+    Wait asynchronously for records on an Amazon Kinesis Data Stream.
+
+    The trigger is long-running and emits one event for each non-empty shard response. Record data is
+    base64-encoded in the event payload and must be decoded by the consumer. Delivery is best-effort:
+    a triggerer failure can cause records to be repeated or missed around the failure window.
+
+    When Airflow provides an asset state store for a single watched asset, the trigger checkpoints the
+    last sequence number read from each shard. The same asset and stream identity share one logical cursor;
+    do not configure multiple watchers that require independent progress for the same stream on one asset.
+
+    :param stream_name: Name of the Kinesis Data Stream to watch.
+    :param aws_conn_id: AWS connection id.
+    :param shard_iterator_type: Position used when a shard has no checkpoint. ``LATEST`` only sees records
+        that arrive after the watcher starts; ``TRIM_HORIZON`` starts from the oldest retained record.
+    :param batch_size: Maximum records per ``GetRecords`` call and trigger event. Must be between 1 and
+        10,000. Record data is base64-encoded before it is stored in the metadata database, so use a
+        conservative value for large records.
+    :param waiter_delay: Seconds between complete polling sweeps. Kinesis permits at most five
+        ``GetRecords`` calls per second per shard. When reading a backlog with ``TRIM_HORIZON``, draining
+        ``N`` records from one shard takes roughly ``ceil(N / batch_size) * waiter_delay`` seconds when
+        calls return full batches; with the defaults, 10,000 records take about 1,000 seconds.
+    :param region_name: AWS region for the Kinesis client.
+    :param verify: Whether to verify SSL certificates, or the path to a CA bundle.
+    :param botocore_config: Botocore configuration passed to the Kinesis client.
+    """
+
+    def __init__(
+        self,
+        stream_name: str,
+        aws_conn_id: str | None = "aws_default",
+        shard_iterator_type: str = "LATEST",
+        batch_size: int = 100,
+        waiter_delay: int = 10,
+        region_name: str | None = None,
+        verify: bool | str | None = None,
+        botocore_config: dict | None = None,
+    ) -> None:
+        super().__init__()
+        if shard_iterator_type not in _ITERATOR_TYPES_WITHOUT_EXTRA_ARGS:
+            raise ValueError(
+                "shard_iterator_type must be one of "
+                f"{sorted(_ITERATOR_TYPES_WITHOUT_EXTRA_ARGS)}; got {shard_iterator_type!r}"
+            )
+        if not 1 <= batch_size <= 10_000:
+            raise ValueError("batch_size must be between 1 and 10000")
+        if waiter_delay < 1:
+            raise ValueError("waiter_delay must be at least 1 second")
+
+        self.stream_name = stream_name
+        self.aws_conn_id = aws_conn_id
+        self.shard_iterator_type = shard_iterator_type
+        self.batch_size = batch_size
+        self.waiter_delay = waiter_delay
+        self.region_name = region_name
+        self.verify = verify
+        self.botocore_config = botocore_config
+        self._checkpoint_warning_logged = False
+
+    def serialize(self) -> tuple[str, dict[str, Any]]:
+        return (
+            self.__class__.__module__ + "." + self.__class__.__qualname__,
+            {
+                "stream_name": self.stream_name,
+                "aws_conn_id": self.aws_conn_id,
+                "shard_iterator_type": self.shard_iterator_type,
+                "batch_size": self.batch_size,
+                "waiter_delay": self.waiter_delay,
+                "region_name": self.region_name,
+                "verify": self.verify,
+                "botocore_config": self.botocore_config,
+            },
+        )
+
+    @property
+    def hook(self) -> KinesisHook:
+        return KinesisHook(
+            aws_conn_id=self.aws_conn_id,
+            region_name=self.region_name,
+            verify=self.verify,
+            config=self.botocore_config,
+        )
+
+    def _build_checkpoint_key(self) -> str:
+        identity = json.dumps(
+            {
+                "stream_name": self.stream_name,
+                "aws_conn_id": self.aws_conn_id,
+                "region_name": self.region_name,
+            },
+            sort_keys=True,
+            separators=(",", ":"),
+        ).encode()
+        return f"{_CHECKPOINT_KEY_PREFIX}:{hashlib.sha256(identity).hexdigest()}"
+
+    def _log_checkpoint_warning_once(self, message: str) -> None:
+        if self._checkpoint_warning_logged:
+            return
+        self.log.warning(message)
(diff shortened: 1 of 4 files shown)


aaron-y-chen:
Those failures seem to be caused by the same issue:

```
flit build: error: unrecognized arguments: --no-setup-py
```

There is already a PR to fix it: #71118

aaron-y-chen:
Waiting for another PR #71145 to be merged.

aaron-y-chen:
Okay, all CI failures have been fixed.

providers/amazon/tests/unit/amazon/aws/triggers/test_kinesis.py
```diff
@@ -0,0 +1,658 @@
...
+class InvalidArgumentException(Exception):
+    pass
+
+
+class ProvisionedThroughputExceededException(Exception):
+    pass
+
+
+class StopPolling(Exception):
+    pass
```
vincbeck, line 50: Why you need this?

aaron-y-chen: The first three are test stand-ins for real Kinesis exceptions. StopPolling is a test-only sentinel raised by the mocked `asyncio.sleep()` to stop the trigger's infinite polling loop deterministically.

aaron-y-chen:
Hi @jason810496, I'd appreciate it if you could take a look when you have time. :)

vincbeck:
@seanghaeli @ramitkataria

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,315 @@
...
+                )
+
+            while True:
+                for shard_id, shard_iterator in list(iterators.items()):
+                    try:
+                        response = await client.get_records(
+                            ShardIterator=shard_iterator,
+                            Limit=self.batch_size,
+                        )
+                    except client.exceptions.ExpiredIteratorException:
```
ramitkataria: On `ExpiredIteratorException` we fetch a fresh iterator and `continue`. Can you walk me through when that new iterator actually gets used? As I read it, the next `get_records` for this shard happens on the following sweep, `waiter_delay` seconds later. What does that look like with `waiter_delay=600`, given iterators expire after 5 minutes?

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,315 @@
...
+    ) -> None:
+        super().__init__()
+        if shard_iterator_type not in _ITERATOR_TYPES_WITHOUT_EXTRA_ARGS:
+            raise ValueError(
+                "shard_iterator_type must be one of "
+                f"{sorted(_ITERATOR_TYPES_WITHOUT_EXTRA_ARGS)}; got {shard_iterator_type!r}"
+            )
+        if not 1 <= batch_size <= 10_000:
+            raise ValueError("batch_size must be between 1 and 10000")
+        if waiter_delay < 1:
```
ramitkataria: We validate `waiter_delay > 0` here. Is there an upper bound worth enforcing too? I'm thinking about the relationship between this value and the 5 minute iterator TTL.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: I checked the [doc](https://docs.aws.amazon.com/streams/latest/dev/troubleshooting-consumers.html) and there's actually a bug if set `waiter_delay` 300s or more. The whole process will become:

`get_records -> sleep for 600s -> next get_records raises ExpiredIteratorException -> acquire a new iterator but defer using it until the next sweep -> sleep again -> repeat`

So your concern makes sense, let me try to fix it.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: Thanks, I bounded `waiter_delay`, I think it should be fixed.

aaron-y-chen:
Thanks for the reviews, I believe this version is more concrete than before :)

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+            for shard_id, sequence_number in checkpoint.items()
+        ):
+            self._log_checkpoint_warning_once(
+                "Kinesis checkpoint data is invalid; using the configured initial position"
+            )
+            return {}
+        return dict(checkpoint)
+
+    def _save_checkpoint(self, sequence_numbers: dict[str, str]) -> None:
+        store = getattr(self, "asset_state_store", None)
```
jason810496: IIUC, this means we're calling a blocking call in the trigger run coroutine.
We should follow the same pattern (`asyncio.to_thread`) in https://github.com/apache/airflow/pull/71387 to workaround.

Or wait until https://github.com/apache/airflow/pull/72127 get merged and invoke the async native method from the accessor.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+        if store is None:
+            self._log_checkpoint_warning_once(
+                "Kinesis checkpointing is unavailable; using an in-memory cursor"
+            )
+            return
+
+        try:
+            store.set(self._build_checkpoint_key(), dict(sequence_numbers))
+        except ValueError:
```
jason810496, line 194: This is the second PR that touched the Task Store within the trigger. https://github.com/apache/airflow/pull/71387 is the first one.

cc @amoghrajesh to double check regarding the usage.
Just in case using Task Store within trigger isn't an expected pattern.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+                "batch_size": self.batch_size,
+                "waiter_delay": self.waiter_delay,
+                "region_name": self.region_name,
+                "verify": self.verify,
+                "botocore_config": self.botocore_config,
+            },
+        )
+
+    @property
+    def hook(self) -> KinesisHook:
```
jason810496, line 125: Not sure do we need to define as cached property or not? We might need to check the lifecycle of the Kinesis client. IIRC, this might be related to the https://lists.apache.org/thread/owp7869xmr5cyks9gx3lp66x53w1owyg discussion ( https://github.com/apache/airflow/pull/71701 PR).

Just in case Kinesis client invoke eager external calls during construction.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+        return KinesisHook(
+            aws_conn_id=self.aws_conn_id,
+            region_name=self.region_name,
+            verify=self.verify,
+            config=self.botocore_config,
+        )
+
+    def _build_checkpoint_key(self) -> str:
```
jason810496: Would it be better to make this as a cache property? (e.g. asset_store_checkpoint_key

jason810496 (commented):
Thanks for adding the new trigger to compelete https://github.com/apache/airflow/issues/52712.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
amoghrajesh: It's a good pattern to use (but slight correction: its asset state store not task state store), he Triggerer itself injects it at: https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L1405-L1408, so a `BaseEventTrigger` attached to a watched asset is designed to read and write asset state. Watermarking a stream cursor is one of the use case.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+            shard_iterator = await self._get_shard_iterator(
+                client,
+                shard_id,
+                after_sequence_number,
+                fallback_iterator_type,
+            )
+            return await client.get_records(
+                ShardIterator=shard_iterator,
+                Limit=self.batch_size,
+            )
```
amoghrajesh: A second `ExpiredIteratorException` will propagate through and kill the trigger.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,336 @@
...
+    a triggerer failure can cause records to be repeated or missed around the failure window.
+
+    When Airflow provides an asset state store for a single watched asset, the trigger checkpoints the
+    last sequence number read from each shard. The same asset and stream identity share one logical cursor;
+    do not configure multiple watchers that require independent progress for the same stream on one asset.
+
+    :param stream_name: Name of the Kinesis Data Stream to watch.
+    :param aws_conn_id: AWS connection id.
+    :param shard_iterator_type: Position used when a shard has no checkpoint. ``LATEST`` only sees records
+        that arrive after the watcher starts; ``TRIM_HORIZON`` starts from the oldest retained record.
```
amoghrajesh, line 61: `AT_TIMESTAMP` and `AT_SEQUENCE_NUMBER` are rejected?

amoghrajesh (dismissed):
Blocking due to sync access to asset state store

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: Yes, if we don't want to block the event loop, we need to make it async. Let me go the workaround way first. If #72127 gets merged before then this PR, I'll revisit it afterward.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: Initially, I followed the pattern used by other triggers and made this a `@property` (for example, `amazon/aws/triggers/sqs.py`).

Constructing kinesis hook performs no external IO, and `get_async_conn()` is called only once in `run()`. Therefore, the kinesis client is not created eagerly, and caching the hook would not change its lifecycle here.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: I think it's a good idea. I'll do it.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: Nice catch! Let me use a for loop to handle it.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: Yes, intentionally. In this PR, I'd prefer to keep the initial-position API limited to `LATEST` and `TRIM_HORIZON`, since they do not require additional arguments.

Supporting `AT_TIMESTAMP` would require adding and serializing a timestamp parameter, while `AT_SEQUENCE_NUMBER` would require a per-shard sequence-number mapping because this trigger polls all shards. Both would add public API surface with their own semantics around checkpoint recovery and resharding, so I think they are better considered separately.

I'll make the reasoning explicit in the docstring; validation already rejects the other values with the supported set in the message.

aaron-y-chen:
Thanks for the reviews! I believe it's more solid than before :)

o-nikolas:
@ramitkataria and @amoghrajesh is this ready for some approvals and merge? (CC: @vincbeck, I see that you also had a brief look)

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
```diff
@@ -0,0 +1,349 @@
...
+                "Kinesis checkpointing is unavailable; using an in-memory cursor"
+            )
+            return
+
+        try:
+            await asyncio.to_thread(
+                store.set,
+                self._asset_store_checkpoint_key,
+                dict(sequence_numbers),
+            )
```
jason810496: We will support `.aget` method in the upcoming 3.3.2 version. So we need the compatible layer for `.aget` and `asyncio.to_thread(store.set, ...` as fallback.

jason810496 (commented):
LGTM once resolve the `aget` method issue, thanks.

providers/amazon/src/airflow/providers/amazon/aws/triggers/kinesis.py
aaron-y-chen: I checked the implementation in #72127 and fixed it. Thanks for the reminder 😄

jason810496 (approved):
LGTM on the code level perspective, thanks.

I didn't look into the AWS Kinesis feature perspective and might need someone more familiar with it to confirm before merge.

ramitkataria (approved):
Thanks for the changes. LGTM as well

jason810496 (commented):
The PR fixed the asset_state_store `aget` usage and got approved from AWS team, merging.

PR #71136, opened 2026-08-05
Refuse separator-bearing env secret ids only in multi-team mode

potiuk:
Another follow-up after @amoghrajesh's #71078, which fixed this same regression in the AWS, Key Vault, Google Secret Manager and Lockbox secrets backends. The core `EnvironmentVariablesBackend` carries it too, and was not covered there — cc @vatsrahul1001 @amoghrajesh.

The backend refuses any connection or variable id carrying the team namespace separator, because such an id cannot be attributed to a team unambiguously — a team name may itself contain the separator, so `_a___b___c` reads as team `a` with id `b___c` just as well as team `a___b` with id `c`.

That refusal fires unconditionally, including in deployments that never enabled multi-team mode. There, an ordinary connection or variable id that happens to contain `___` is silently reported as missing — no exception, and nothing distinguishes it from a secret that was never set.

No team scoped secret can be read while multi-team is off: the lookups are only ever handed a `team_name` in that mode. So the id has a single meaning there and nothing needs refusing. Same gate as #71078, applied at the same place in the logic.

### What this changes for the tests

The existing team-scoping tests all assert the refusal, so they now run with multi-team enabled. Two of them (`test_connection_env_var_do_not_access_team_specific` and `test_variable_env_var_do_not_access_team_specific`) would otherwise fail, because `AIRFLOW_CONN__TEAM___TEST_MYSQL` legitimately resolves as an ordinary connection named `_team___test_mysql` once multi-team is off — that is the intended trade, and the same one #71078 makes on the provider side.

### Test plan

- [x] New `TestEnvironmentVariablesBackendWithoutMultiTeam` — verified against unmodified code: it **fails** with `assert None == 'team-agnostic-value'`, the reported symptom exactly
- [x] `test_secrets.py`, `test_secrets_environment_variables.py`, `test_secrets_backends.py` — 72 passed
- [x] Full `airflow-core/tests/unit/always` — 2008 passed; the one failure (`example_mysql_to_gcs.py`) is a missing `mysqlclient` on the host and reproduces on unmodified `main`
- [x] `ruff` / `ruff format` / static checks clean

No newsfragment: #70736 is milestoned 3.3.1 and unreleased, so the regression never reached users.

---

### DIFF
--- a/airflow-core/src/airflow/secrets/environment_variables.py
+++ b/airflow-core/src/airflow/secrets/environment_variables.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 import os
 
+from airflow.configuration import conf
 from airflow.secrets import BaseSecretsBackend
 
 CONN_ENV_PREFIX = "AIRFLOW_CONN_"
@@ -35,9 +36,7 @@ class EnvironmentVariablesBackend(BaseSecretsBackend):
     """Retrieves Connection object and Variable from environment variable."""
 
     def get_conn_value(self, conn_id: str, team_name: str | None = None) -> str | None:
-        if TEAM_SEP in conn_id:
-            # An id containing the separator could collide with another team's namespace
-            # even on the scoped lookup below, so it must be refused before either runs.
+        if self._names_a_team_namespace(conn_id):
             return None
 
         if team_name and (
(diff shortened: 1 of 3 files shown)


potiuk:
Closing in favour of https://github.com/apache/airflow/pull/71079

Issue #69288, opened 2026-07-03
TypeScript SDK: native TypeScript Dag declaration Design


jason810496:
**We should wait until [AIP-85 Dag Importers](https://cwiki.apache.org/confluence/spaces/AIRFLOW/pages/315494137/AIP-85+DAG+importer) pass**.

### Background

The TypeScript SDK intentionally ships in "Python-stub-Dag mode": TypeScript handlers are registered for Dag/task IDs declared in Python. Higher-level TypeScript Dag authoring was deferred until coordinator-mode execution is reviewed.

### What needs to happen

1. Allow declaring Dags natively in TypeScript.
2. Revisit bundle ergonomics alongside the authoring interface.

### Acceptance criteria

- A Dag can be authored end to end in TypeScript with no Python stub file.

### Context

- https://github.com/apache/airflow/pull/67908 (PR description)
- https://github.com/apache/airflow/pull/69079 (PR description)

jason810496:
Hi @guan404ming, @shivaam,

I think we can start discussing the user interface for the native TypeScript Dag definition here, if you'd like to join. Otherwise, I'll start the design on my side. (No need to start implementing at this point, just a discussion about what exactly the interface will look like.)

The high-level idea of how the native Dag works for each Lang-SDK is:

- We first define a DSL for the Dag and Task (we should be able to set, for example,` schedule="@daily"` at the Dag level, `retries=3` at the Task level, etc.).
- Then the Lang-SDK runtime should be able to return an `airflow-core/src/airflow/serialization/schema.json`-compatible Dag JSON back to the coordinator via the same comm protocol.

Some points we need to be aware of:

- From a language perspective, whether the syntax itself can feasibly return a serialized Dag JSON.
- Whether the syntax is "TaskFlow"-like enough. By TaskFlow, I mean users can define dependencies through "chaining," without needing something like a "taskId as a string/array" reference on the other task.
- How arguments get resolved in the TS-SDK runtime across tasks in the same language, and for the same task across different languages.

For reference, here are the Go-SDK counterparts:
- https://github.com/apache/airflow/pull/70158 -- Go-SDK native Dag syntax (no need to dive into the implementation; the syntax in the PR description is enough).
- https://github.com/apache/airflow/pull/69757 -- the foundation for resolving arguments across native Dags and across different languages: we serialize the argument type annotation as a JSON schema and store it in the serialized Dag.

Either way, let's focus only on the user interface design for now.
Please let me know WDYT, thanks.

[...]

PR #71144, opened 2026-08-05
TS SDK: replace registerTask with Dag and registerDags

jason810496:
- related: #69288

## Why

`registerTask({ dagId, taskId }, handler)` binds one handler to a Python stub Dag/task pair at a time and leaves no object that can carry native TypeScript Dag declaration later (#69288). This PR changes only the user-facing authoring interface so it can grow into TaskFlow and native Dag support without another breaking change.

## How

**Please noted that the spec at both Dag and Task level arguments (`DagSpec`, `TaskSpec`) and the task `inputs` are no-op, I intentionally placeholder them in this PR to prevent further breaking change.**

- `new Dag(dagId, spec?)` takes a positional id plus an optional trailing spec object.
- `dag.task(taskId, handler, options?)` returns a `TaskRef` handle, with placeholder `inputs` and `spec` as named options (the further native Dag: `{ inputs: { extracted }, spec: { retries: 2 } }`) so the future options need no new parameter breaking change. Unknown option keys are rejected, so a typo fails at import time instead of being silently ignored.
- `DagSpec` and `TaskSpec` are `Record<string, never>` while they are placeholders, so a field that would be silently dropped does not compile.
- `dag.taskIds` lists the attached task IDs, and `registry.listDags()` lists the Dags a bundle serves with their tasks, so a user can assert their handlers match the Python `@task.stub` names.
- A `Dag` instance retains its spec and each task's `(taskId, handler, spec, inputs)`, so a future `serialize()` can produce the serialized Dag JSON for `DagFileParsingResult.serialized_dags`.
- `registerTask`, `listRegisteredTasks`, and `TaskRegistration` are replaced with a public `DagRegistry` — plus `serveDags(registry)` as the entrypoint. `startCoordinator` stays out of the public API.
- `airflow-ts-pack` warns instead of failing when a registered Dag has no tasks, matching `airflow-go-pack`. A Dag left out of the registry is simply absent from the bundle and its tasks are marked removed at runtime.

---

### DIFF
--- a/ts-sdk/example/src/main.ts
+++ b/ts-sdk/example/src/main.ts
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { Dag, registerDags, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
+import { Dag, DagRegistry, serveDags, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
 
 const dag = new Dag("typescript_example");
 
@@ -52,4 +52,4 @@ export async function readConnection({ client }: TaskHandlerArgs) {
 dag.task("build_message", buildMessage);
 dag.task("read_connection", readConnection);
 
-await registerDags(dag);
+await serveDags(new DagRegistry(dag));
--- a/ts-sdk/example/src/main.ts
+++ b/ts-sdk/example/src/main.ts
@@ -17,9 +17,9 @@
  * under the License.
  */
 
-import { registerTask, startCoordinator, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
+import { Dag, registerDags, startCoordinator, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
 
-const DAG_ID = "typescript_example";
+const dag = new Dag("typescript_example");
 
 export async function buildMessage({ client }: TaskHandlerArgs) {
   const upstream = await client.getXCom<string>({
@@ -49,7 +49,8 @@ export async function readConnection({ client }: TaskHandlerArgs) {
   };
 }
 
-registerTask({ dagId: DAG_ID, taskId: "build_message" }, buildMessage);
-registerTask({ dagId: DAG_ID, taskId: "read_connection" }, readConnection);
+dag.task("build_message", buildMessage);
+dag.task("read_connection", readConnection);
+registerDags(dag);
 
 await startCoordinator();
--- a/ts-sdk/example/src/main.ts
+++ b/ts-sdk/example/src/main.ts
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { Dag, registerDags, startCoordinator, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
+import { Dag, registerDags, type TaskHandlerArgs } from "@apache-airflow/ts-sdk";
 
 const dag = new Dag("typescript_example");
 
@@ -51,6 +51,5 @@ export async function readConnection({ client }: TaskHandlerArgs) {
 
 dag.task("build_message", buildMessage);
 dag.task("read_connection", readConnection);
-registerDags(dag);
 
-await startCoordinator();
+await registerDags(dag);
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -143,23 +143,18 @@ function readBundleManifest(bundlePath: string): BundleManifest {
     throw new Error(`Bundle produced no ${AIRFLOW_METADATA_FLAG} output`);
   }
 
-  let manifest: BundleManifest;
+  let parsed: unknown;
   try {
-    manifest = JSON.parse(line.slice(AIRFLOW_METADATA_SENTINEL.length)) as BundleManifest;
+    parsed = JSON.parse(line.slice(AIRFLOW_METADATA_SENTINEL.length));
   } catch (error) {
     throw new Error(`Bundle produced invalid ${AIRFLOW_METADATA_FLAG} output: ${String(error)}`, {
       cause: error,
     });
   }
-  if (
-    !manifest.supervisor_schema_version ||
-    typeof manifest.dags !== "object" ||
-    manifest.dags === null ||
-    // An array would pass the typeof check and yield Dags named "0", "1", ...
-    Array.isArray(manifest.dags)
-  ) {
+  if (!isBundleManifest(parsed)) {
     throw new Error(`Bundle produced incomplete ${AIRFLOW_METADATA_FLAG} output`);
   }
+  const manifest = parsed;
   // The line is whatever the bundle printed and nothing downstream re-validates
   // it, so check each Dag entry down to the task-id element.
   for (const [dagId, dag] of Object.entries(manifest.dags)) {
@@ -170,6 +165,24 @@ function readBundleManifest(bundlePath: string): BundleManifest {
     }
   }
   return manifest;
+}
+
+// JSON.parse also yields null, primitives and arrays, so the document is checked
+// before anything is read off it — `null.supervisor_schema_version` would
+// otherwise surface as a raw TypeError rather than a packing error.
+function isBundleManifest(value: unknown): value is BundleManifest {
+  if (typeof value !== "object" || value === null || Array.isArray(value)) return false;
+  const { supervisor_schema_version: version, dags } = value as Partial<BundleManifest>;
+  return (
+    // Goes into the manifest verbatim, where the schema requires a non-empty
+    // string; a truthy number or boolean would be rendered into the YAML as-is.
+    typeof version === "string" &&
+    version.length > 0 &&
+    typeof dags === "object" &&
+    dags !== null &&
+    // An array would pass the typeof check and yield Dags named "0", "1", ...
+    !Array.isArray(dags)
+  );
 }
 
 function isDagIdList(value: unknown): value is string[] {
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -156,8 +156,7 @@ function readBundleManifest(bundlePath: string): BundleManifest {
     typeof manifest.dags !== "object" ||
     manifest.dags === null ||
     // An array would pass the typeof check and yield Dags named "0", "1", ...
-    Array.isArray(manifest.dags) ||
-    !isDagIdList(manifest.unregistered_dags ?? [])
+    Array.isArray(manifest.dags)
   ) {
     throw new Error(`Bundle produced incomplete ${AIRFLOW_METADATA_FLAG} output`);
   }
@@ -215,7 +214,7 @@ export async function runPack(argv: readonly string[]): Promise<void> {
     const manifest = readBundleManifest(stagingPath);
     const dagEntries = Object.entries(manifest.dags);
     if (dagEntries.length === 0) {
-      throw new Error(`${args.entry} registered no Dags; register them with registerDags(...)`);
+      throw new Error(`${args.entry} served no Dags; pass them to serveDags(new DagRegistry(...))`);
     }
     // Warn rather than fail, as airflow-go-pack does: the shared schema allows a
     // Dag with no tasks.
@@ -223,12 +222,6 @@ export async function runPack(argv: readonly string[]): Promise<void> {
       if (dag.tasks.length === 0) {
         process.stderr.write(`warning: dag ${JSON.stringify(dagId)} has no tasks\n`);
       }
-    }
-    for (const dagId of manifest.unregistered_dags ?? []) {
-      process.stderr.write(
-        `warning: dag ${JSON.stringify(dagId)} was declared but never passed to registerDags(...); ` +
-          "its tasks will be marked removed at runtime\n",
-      );
     }
 
     const metadataYaml = renderMetadataYaml({
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -158,7 +158,7 @@ function readBundleManifest(bundlePath: string): BundleManifest {
   // The line is whatever the bundle printed and nothing downstream re-validates
   // it, so check each Dag entry down to the task-id element.
   for (const [dagId, dag] of Object.entries(manifest.dags)) {
-    if (dag == null || !isDagIdList(dag.tasks)) {
+    if (dag == null || !isTaskIdList(dag.tasks)) {
       throw new Error(
         `Bundle produced ${AIRFLOW_METADATA_FLAG} output with a malformed entry for Dag "${dagId}"`,
       );
@@ -185,7 +185,7 @@ function isBundleManifest(value: unknown): value is BundleManifest {
   );
 }
 
-function isDagIdList(value: unknown): value is string[] {
+function isTaskIdList(value: unknown): value is string[] {
   return Array.isArray(value) && value.every((item) => typeof item === "string" && item.length > 0);
 }
 
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -132,6 +132,7 @@ function readBundleManifest(bundlePath: string): BundleManifest {
     const stderr = (error as { stderr?: string }).stderr ?? "";
     const reported = stderr
       .split("\n")
+      .reverse()
       .find((line) => /^\w*Error: /.test(line.trim()))
       ?.trim();
     throw new Error(
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -129,9 +129,15 @@ function readBundleManifest(bundlePath: string): BundleManifest {
       maxBuffer: MANIFEST_MAX_BUFFER_BYTES,
     });
   } catch (error) {
-    throw new Error(`Running the bundle with ${AIRFLOW_METADATA_FLAG} failed: ${String(error)}`, {
-      cause: error,
-    });
+    const stderr = (error as { stderr?: string }).stderr ?? "";
+    const reported = stderr
+      .split("\n")
+      .find((line) => /^\w*Error: /.test(line.trim()))
+      ?.trim();
+    throw new Error(
+      reported ?? `Running the bundle with ${AIRFLOW_METADATA_FLAG} failed: ${String(error)}`,
+      { cause: error },
+    );
   }
 
   // Import-time logging from user code lands on stdout too; pick the sentinel line.
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -167,15 +167,15 @@ function readBundleManifest(bundlePath: string): BundleManifest {
   return manifest;
 }
 
-// JSON.parse also yields null, primitives and arrays, so the document is checked
-// before anything is read off it — `null.supervisor_schema_version` would
-// otherwise surface as a raw TypeError rather than a packing error.
+// The document is checked before anything is read off it: JSON.parse also yields
+// null and primitives, and `null.supervisor_schema_version` would surface as a
+// raw TypeError rather than a report about the bundle.
 function isBundleManifest(value: unknown): value is BundleManifest {
   if (typeof value !== "object" || value === null || Array.isArray(value)) return false;
   const { supervisor_schema_version: version, dags } = value as Partial<BundleManifest>;
   return (
-    // Goes into the manifest verbatim, where the schema requires a non-empty
-    // string; a truthy number or boolean would be rendered into the YAML as-is.
+    // Rendered into the manifest verbatim, where the schema requires a non-empty
+    // string, so a truthy number or boolean would travel to Airflow as-is.
     typeof version === "string" &&
     version.length > 0 &&
     typeof dags === "object" &&
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -154,6 +154,15 @@ function readBundleManifest(bundlePath: string): BundleManifest {
   if (!manifest.supervisor_schema_version || !manifest.dags || typeof manifest.dags !== "object") {
     throw new Error(`Bundle produced incomplete ${AIRFLOW_METADATA_FLAG} output`);
   }
+  // The line is whatever the bundle printed, so the per-Dag shape is not
+  // guaranteed by the type assertion above.
+  for (const [dagId, dag] of Object.entries(manifest.dags)) {
+    if (dag == null || !Array.isArray(dag.tasks)) {
+      throw new Error(
+        `Bundle produced ${AIRFLOW_METADATA_FLAG} output with a malformed entry for Dag "${dagId}"`,
+      );
+    }
+  }
   return manifest;
 }
 
@@ -193,9 +202,19 @@ export async function runPack(argv: readonly string[]): Promise<void> {
     });
 
     const manifest = readBundleManifest(stagingPath);
-    if (Object.keys(manifest.dags).length === 0) {
+    const dagEntries = Object.entries(manifest.dags);
+    if (dagEntries.length === 0) {
       throw new Error(
-        `${args.entry} registered no tasks; call registerTask(...) before startCoordinator()`,
+        `${args.entry} registered no Dags; register Dags with registerDags(...) before startCoordinator()`,
+      );
+    }
+    const emptyDags = dagEntries
+      .filter(([, dag]) => dag.tasks.length === 0)
+      .map(([dagId]) => dagId);
+    if (emptyDags.length > 0) {
+      throw new Error(
+        `${args.entry} registered no tasks for Dag(s) ${emptyDags.map((id) => `"${id}"`).join(", ")}; ` +
+          "attach handlers with dag.task(...) before packing",
       );
     }
 
--- a/ts-sdk/src/cli/pack.ts
+++ b/ts-sdk/src/cli/pack.ts
@@ -151,19 +151,30 @@ function readBundleManifest(bundlePath: string): BundleManifest {
       cause: error,
     });
   }
-  if (!manifest.supervisor_schema_version || !manifest.dags || typeof manifest.dags !== "object") {
+  if (
+    !manifest.supervisor_schema_version ||
+    typeof manifest.dags !== "object" ||
+    manifest.dags === null ||
+    // An array would pass the typeof check and yield Dags named "0", "1", ...
+    Array.isArray(manifest.dags) ||
+    !isDagIdList(manifest.unregistered_dags ?? [])
+  ) {
     throw new Error(`Bundle produced incomplete ${AIRFLOW_METADATA_FLAG} output`);
   }
-  // The line is whatever the bundle printed, so the per-Dag shape is not
-  // guaranteed by the type assertion above.
+  // The line is whatever the bundle printed and nothing downstream re-validates
+  // it, so check each Dag entry down to the task-id element.
   for (const [dagId, dag] of Object.entries(manifest.dags)) {
-    if (dag == null || !Array.isArray(dag.tasks)) {
+    if (dag == null || !isDagIdList(dag.tasks)) {
       throw new Error(
         `Bundle produced ${AIRFLOW_METADATA_FLAG} output with a malformed entry for Dag "${dagId}"`,
       );
     }
   }
   return manifest;
+}
+
+function isDagIdList(value: unknown): value is string[] {
+  return Array.isArray(value) && value.every((item) => typeof item === "string" && item.length > 0);
 }
 
 // esbuild keeps an entry hashbang as line 1, where the metadata comment must go;
@@ -204,17 +215,19 @@ export async function runPack(argv: readonly string[]): Promise<void> {
     const manifest = readBundleManifest(stagingPath);
     const dagEntries = Object.entries(manifest.dags);
     if (dagEntries.length === 0) {
-      throw new Error(
-        `${args.entry} registered no Dags; register Dags with registerDags(...) before startCoordinator()`,
-      );
+      throw new Error(`${args.entry} registered no Dags; register them with registerDags(...)`);
     }
-    const emptyDags = dagEntries
-      .filter(([, dag]) => dag.tasks.length === 0)
-      .map(([dagId]) => dagId);
-    if (emptyDags.length > 0) {
-      throw new Error(
-        `${args.entry} registered no tasks for Dag(s) ${emptyDags.map((id) => `"${id}"`).join(", ")}; ` +
-          "attach handlers with dag.task(...) before packing",
+    // Warn rather than fail, as airflow-go-pack does: the shared schema allows a
+    // Dag with no tasks.
+    for (const [dagId, dag] of dagEntries) {
+      if (dag.tasks.length === 0) {
+        process.stderr.write(`warning: dag ${JSON.stringify(dagId)} has no tasks\n`);
+      }
+    }
+    for (const dagId of manifest.unregistered_dags ?? []) {
+      process.stderr.write(
+        `warning: dag ${JSON.stringify(dagId)} was declared but never passed to registerDags(...); ` +
+          "its tasks will be marked removed at runtime\n",
       );
     }
 
--- a/ts-sdk/src/coordinator/runtime.ts
+++ b/ts-sdk/src/coordinator/runtime.ts
@@ -52,7 +52,7 @@ import {
   type RuntimeTaskState,
   type StartupDetails,
 } from "./protocol.js";
-import { DagRegistry, isDagRegistry } from "../sdk/registry.js";
+import { DagRegistry, isDagRegistry, listRegistryTasks } from "../sdk/registry.js";
 import { DUPLICATE_COPY_HINT } from "../sdk/dag.js";
 import type { TaskContext, TaskHandlerArgs } from "../sdk/task.js";
 import type { JsonValue } from "../sdk/client-types.js";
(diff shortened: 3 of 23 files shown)


uranusjr:
What’s the plan to do the external vs pure TS dags after this?

jason810496:
> What’s the plan to do the external vs pure TS dags after this?

Along with https://github.com/apache/airflow/pull/71213, the user should set the `is_mixed_language_dag` for the mixed language Dag case, it will look somehow like `new Dag("my_ts_mixed_lang", spec: {is_mixed_language_dag: True, schedule: "..."})`.

Additionally, we _can_ introduce a `new ExternalDag` or `new MixedLangDag` if we feel having explicitly different user interface is better. IMO, having the `is_mixed_language_dag` on DagSpec level plus user education should be enough for the first stage.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,208 @@
...
+ * Empty today: future task fields (retries, ...) will land here without
+ * changing the `dag.task()` signature.
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type -- extension point for future task fields
+export interface TaskSpec {}
+
+/**
+ * Opaque handle to a task registered on a {@link Dag}.
+ *
+ * Pass it as a downstream task's input to declare that the downstream task
```
kaxil: `inputs` does not declare anything today, but this headline says it does, and the gap can produce wrong data rather than an error.

Grepping `ts-sdk/src/`, nothing outside this file reads `TaskRecord.inputs`: `registry.ts` reads `.handler`, `.task` and `.keys()`, `manifest.ts` reads only `dagId`/`tasks`. There is also no channel to deliver an input to a handler, since `TaskHandlerArgs` is still `{ctx, client}` (`src/sdk/task.ts`).

The reachable failure: a user who reads this headline concludes the value is wired, so writes `await client.getXCom({ key: "return_value" })` with no `taskId`. Per `src/sdk/client-types.ts:32-33`, `taskId` defaults to the running task's context, so that reads the task's *own* XCom, returns `null`, and the task succeeds with the wrong value. No error, no warning, no failed task instance.

Neither `ts-sdk/README.md` nor `typescript.rst` mentions `inputs`, and the rst note next to the paragraph this PR edits still says dependencies are declared in the Python stub Dag, so autocomplete plus this JSDoc is the only discovery channel.

Either resolution works:

- Drop `inputs`/`TaskInputs`/`#validateInputs` for now and land `Dag` + `dag.task(taskId, handler)` + `registerDags`, which is the whole refactor the title promises. Adding an optional trailing `options` parameter later is not source-breaking in TypeScript, so waiting costs nothing.
- Keep it, but lead the headline with what it does today (nothing), and add a `dag.task` options table to the rst marking `inputs` reserved and inert.

ts-sdk/src/sdk/registry.ts
```diff
@@ -17,85 +17,72 @@
...
-  return defaultRegistry.get(dagId, taskId);
-}
+/** The registry `registerDags` writes to and the coordinator reads from. */
+export const defaultRegistry = new DagRegistry();
 
-/** List all registered tasks. */
-export function listRegisteredTasks(): TaskRegistration[] {
-  return defaultRegistry.list();
+/** Record Dags in the default registry so the coordinator can run their tasks. */
+export function registerDags(...dags: Dag[]): void {
```
kaxil: Splitting declaration from registration makes this third step forgettable, and the chain fails quietly when it is missed.

`buildBundleManifest()` reads `defaultRegistry.listDags()`, so a Dag that was constructed and populated but never passed here is absent from the manifest. `runPack` catches only the all-or-nothing case (`registered no Dags`), so registering 3 of 4 Dags packs successfully. `NodeCoordinator._find_bundle` ignores the manifest's `dags` map when selecting a bundle (`task-sdk/src/airflow/sdk/coordinators/node/coordinator.py`), so the bundle is still launched for the missing Dag's tasks, `handleTask` finds no handler, and it returns `{type: "TaskState", state: "removed"}`. The user sees task instances quietly marked *removed*, which normally means the task is no longer in the Dag, with only a `warning` line in the task log.

Fair caveat: forgetting `import "./sales/tasks"` under the old side-effect pattern registered nothing either, so this is a new instance of an existing mistake rather than a regression in kind. What is new is that it is now cheaply detectable, because the Dag is an object the SDK can see.

Cheapest fix that keeps this shape: have the `Dag` constructor push `this` onto a module-level list, emit it in `--airflow-metadata` mode, and have pack fail with `Dag "billing" was declared but never passed to registerDags(...)`.

The structural version is worth a thought too. Java uses this same constructor shape (`java.rst`: `var dag = new Dag("my_dag"); dag.addTask("fetch", FetchTask.class); return List.of(dag);`) but returns the Dags from `getDags()`, so omission is unrepresentable; Go inverts ownership instead, with `Registry.AddDag(dagId) Dag`. A TypeScript analogue of either, `startCoordinator({ dags: [salesDag, billingDag] })` or an `addDag(dagId, spec?)` that registers eagerly and returns the handle, removes the forgettable step rather than detecting it.

On naming, while the package is still unpublished: `Task` here is an opaque `{dagId, taskId}` identity, whereas Go's `Task` is the executable thing the runtime calls and Java's is the interface a task class implements. `TaskRef` or `TaskHandle` would keep the vocabulary aligned across the three SDKs and leave `Task` free for the real task object native Dag declaration will need.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/**
+ * Dag-level options.
+ *
+ * Empty today: native TypeScript Dag declaration (schedule, tags, ...) will add
+ * optional fields here without changing the `Dag` constructor signature.
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type -- extension point for future native-Dag fields
+export interface DagSpec {}
```
kaxil: `interface DagSpec {}` is the empty object type, so `spec` accepts every non-nullish value, and the rule suppressed here is the one that reports exactly that. Today `new Dag("d", 42)` and `dag.task("t", fn, { spec: { retries: 3 } })` both typecheck and store the value unvalidated. `TaskSpec`'s own docstring advertises `retries` as a future field, so that second call is a natural thing to write and have silently ignored.

The repo has already taken a position on the spelling: `airflow-core/src/airflow/ui/rules/typescript.js:668` sets `@typescript-eslint/no-empty-object-type` to `ERROR`, and its doc block lists `type FooType = {}` as incorrect while prescribing `type FooType = object`. That is the UI package's config rather than `ts-sdk/eslint.config.js`, so precedent rather than a rule this PR breaks, but the question has an in-repo answer.

The stated rationale also inverts. Once `DagSpec` gains its first field, necessarily optional, it becomes a weak type, and any call site passing an object with no overlapping property starts failing with TS2559. Every other extension the comment anticipates is already non-breaking: an optional trailing parameter, an optional property on `TaskOptions` (a type users pass and never implement), and `Task` becoming `Task<TReturn = unknown>` are all source-compatible. The only source-breaking change in that set is the one the placeholder introduces.

`export type DagSpec = object` rejects primitives and matches the spelling the repo prescribes; `Record<string, never>` also rejects `{retries: 3}`. Either one removes both `eslint-disable` lines.

One caveat on my side: I verified the lint policy and these declarations, but not the assignability behaviour with a `tsc` run, since `ts-sdk/` has no installed `node_modules` in my checkout. Both are one-line checks.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,208 @@
...
+   * `taskId` must match the Dag-side operator's `task_id` exactly, including
+   * any TaskGroup prefix. `options.inputs` names the upstream task handles whose
+   * return values this task consumes. Returns this task's handle.
+   */
+  task<TReturn = unknown>(
+    taskId: string,
+    handler: TaskHandler<TReturn>,
+    options: TaskOptions = {},
+  ): Task {
+    const { inputs = {}, spec = {} } = options;
```
kaxil: This destructure accepts any other shape silently, which is out of step with the three checks around it. `taskId`, `handler`, and every `inputs` value are runtime-validated precisely because TypeScript can be bypassed, but `dag.task("transform", fn, { input: { extracted } })` (singular typo) or `{ inpts: {...} }` registers a task with no declared upstream and says nothing. Reachable from plain JavaScript, since `airflow-ts-pack` bundles whatever esbuild accepts, and from an `as TaskOptions` cast. This package already fails closed on the same class of mistake: `parsePackArgs` throws `Unknown option ${arg}`.

Worth fixing in the same place: the new comment at `tests/public-api.test.ts:178` says "these constructor/method misuses also throw at runtime", and that is false for one of the five lines it covers. `dag.task("transform2", async () => undefined, { upstream })` does not throw; it registers with zero inputs and discards the declared upstream. The other four do throw. Because the arrow function is never invoked, nothing catches it.

```ts
for (const key of Object.keys(options)) {
  if (key !== "inputs" && key !== "spec") {
    throw new Error(`Unknown option "${key}" for Dag "${this.dagId}" task "${taskId}"`);
  }
}
```

Also worth handling while here: `options === null` currently fails with an opaque "Cannot destructure property", because the default only applies to `undefined`.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,208 @@
...
+ * instance retains its `spec` and every task's `(taskId, handler, spec)` so a
+ * future `serialize()` can produce the serialized Dag JSON for native
+ * TypeScript Dag declaration.
+ */
+export class Dag {
+  /** Identifier of this Dag. Must match the Python Dag's `dag_id`. */
+  readonly dagId: string;
+  /** Dag-level options this instance was constructed with. */
+  readonly spec: DagSpec;
+  readonly #tasks = new Map<string, TaskRecord>();
```
kaxil, line 139: With tasks in a `#`-private field and `listRegisteredTasks` removed, no public API is left that answers "which tasks did I wire up?". `defaultRegistry`, `DagRegistry.listTasks`, `listDags` and `getDagTaskRecords` are all module-exported but absent from `src/index.ts`, the two-entry `exports` map blocks deep imports, and `./coordinator` re-exports only `startCoordinator`, `StartCoordinatorOptions` and `SUPERVISOR_API_VERSION`.

That matters here specifically, because it removes the one test a user could write against the drift this stub-Dag model is exposed to: that their TypeScript handlers match the Python `@task.stub` names. It is also the test that would catch a forgotten `registerDags` at author time instead of as task instances marked *removed*. The Go SDK exposes this deliberately, through `EnumerableBundle.OrderedDags()`.

```ts
/** Task IDs attached to this Dag, in attachment order. */
get taskIds(): readonly string[] { return [...this.#tasks.keys()]; }
```

That also lets `registry.listDags()` stop going through `getDagTaskRecords`.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (this.#tasks.has(taskId)) {
+      throw new Error(`Task "${taskId}" is already registered for Dag "${this.dagId}"`);
+    }
+    this.#validateInputs(taskId, inputs);
+    const task: Task = Object.freeze({ dagId: this.dagId, taskId });
+    this.#tasks.set(taskId, {
+      task,
+      handler: handler as TaskHandler,
+      spec,
+      inputs: Object.freeze({ ...inputs }),
```
kaxil, line 187: `inputs` is copied and frozen here, but `spec` on the line above and `this.spec` in the constructor are stored by reference, and `tests/sdk/dag.test.ts` locks that in with `toBe(dagSpec)` / `toBe(taskSpec)`. Once `DagSpec` carries real fields, a caller who mutates their spec object after construction silently mutates the Dag. `Object.freeze({ ...spec })` in both places, with those two assertions relaxed to `toEqual`, makes the two options behave alike.

Separately, the test named "records frozen inputs that later mutation of the caller's object cannot change" does not test the freeze. It adds a new *key* to the caller's object and asserts the record is unchanged, which passes on the spread copy alone; delete `Object.freeze` here and it stays green. One assertion covers it: `expect(Object.isFrozen(getDagTaskRecords(dag).get("transform")!.inputs)).toBe(true)`.

ts-sdk/src/cli/pack.ts
```diff
@@ -154,6 +154,15 @@ function readBundleManifest(bundlePath: string): BundleManifest {
   if (!manifest.supervisor_schema_version || !manifest.dags || typeof manifest.dags !== "object") {
     throw new Error(`Bundle produced incomplete ${AIRFLOW_METADATA_FLAG} output`);
   }
+  // The line is whatever the bundle printed, so the per-Dag shape is not
+  // guaranteed by the type assertion above.
+  for (const [dagId, dag] of Object.entries(manifest.dags)) {
+    if (dag == null || !Array.isArray(dag.tasks)) {
```
kaxil: This stops one level short of the schema it protects. Element types are unchecked, so `{"dags":{"d":{"tasks":[null,{},"",123]}}}` passes here and flows through `renderMetadataYaml` into the base64 trailer, while `task-sdk/docs/airflow-metadata.schema.json` defines `dagEntry.tasks.items` as `{"type": "string", "minLength": 1}`. The Go reference packer cannot emit that, because it decodes into `Tasks []string`, and the Python reader does not close the gap either: `_bundle_metadata.py` validates only that the document is a mapping and that `sdk.supervisor_schema_version` is a non-empty string, so bad task ids propagate past pack.

```ts
if (dag == null || !Array.isArray(dag.tasks) ||
    dag.tasks.some((t) => typeof t !== "string" || t.length === 0)) {
```

Cheap while you are here: the guard on the line above passes for an *array*, which `Object.entries` then turns into Dags named `"0"`, `"1"`. Pre-existing, but this loop is the natural place to close it.

ts-sdk/src/cli/pack.ts
```diff
@@ -193,9 +202,19 @@ export async function runPack(argv: readonly string[]): Promise<void> {
...
+    const dagEntries = Object.entries(manifest.dags);
+    if (dagEntries.length === 0) {
+      throw new Error(
+        `${args.entry} registered no Dags; register Dags with registerDags(...) before startCoordinator()`,
+      );
+    }
+    const emptyDags = dagEntries
+      .filter(([, dag]) => dag.tasks.length === 0)
+      .map(([dagId]) => dagId);
+    if (emptyDags.length > 0) {
```
kaxil: This is a policy divergence from the packer this file says it mirrors. For the same condition `airflow-go-pack` prints `warning: dag %q has no tasks` and continues (`go-sdk/cmd/airflow-go-pack/pack.go`), while this throws. The two agree on rejecting zero Dags, and the shared schema permits an empty list, since `dagEntry.tasks` has no `minItems`.

Failing closed is arguably the better call, and there is a forward-looking argument for it, since `executable/coordinator.py` claims bundle ownership from `set(dags.keys())` alone and Node is slated to grow the same routing. But it means one placeholder Dag, or conditional attachment like `if (process.env.FEATURE_X) dag.task(...)`, blocks the whole bundle build in TypeScript and not in Go. Worth picking one deliberately and applying it across both: match Go's warning, or hard-error in both and add `"minItems": 1` to the schema.

ts-sdk/src/sdk/dag.ts
jason810496: Added exactly your suggestion

```ts
get taskIds(): readonly string[] { return [...this.#tasks.keys()]; }
```

`DagRegistry.listDags()` now goes through it instead of `getDagTaskRecords`.

ts-sdk/src/sdk/dag.ts
jason810496: I went with your second option. `inputs`, `spec`, `DagSpec` and `TaskSpec` now lead with "Reserved: inert today" on every declaration, the `inputs` docstring names the exact trap (`client.getXCom` without `taskId` reads the running task's own XCom), and explicitly mentioned in`typescript.rst`.

ts-sdk/src/sdk/registry.ts
jason810496: `registerDags(...)` is now the single entrypoint, since a Dag author never needs to know Airflow's coordinator at all. A second `registerDags` call is rejected, which would otherwise start a second runtime.

For the Dag instance omission, I prefer to let `airflow-ts-pack` warns instead of having restrict error (same as Python side behavior, it's fine to "define the Dag but don't construct it".

I agreed to rename `Task` to `TaskRef` to match the convention of Go and Java and avoid having the `Task` too ambigious.

ts-sdk/src/sdk/dag.ts
jason810496: Both are now `Record<string, never>`, and both `eslint-disable` lines are gone. I ran the `tsc` check you flagged as unverified:

| written today | `interface DagSpec {}` | `type DagSpec = object` | `Record<string, never>` |
|---|---|---|---|
| `new Dag("d", 42)` | compiles | TS2345 | TS2345 |
| `new Dag("d", { schedule: "@daily" })` | compiles, dropped | compiles, dropped | TS2322 |

`object` fixes only the primitive case, it still silently accepts the field a user would actually write. Your TS2559 point checks out too (but it can only bite call sites that pass a non-`{}` spec).

`Record<string, never>` makes unwritable, and `{}` stays assignable to an all-optional generated type.

ts-sdk/src/sdk/dag.ts
jason810496: Added `#validateOptions`: unknown keys throw `Unknown option "input" for Dag "d" task "t"`, and a non-object (including `null` and arrays) throws before the destructure, so no more opaque "Cannot destructure property".

Good catch on `tests/public-api.test.ts:178` — that comment was false for the `{ upstream }` line. It is now true rather than reworded, and a runtime test covers all three misuse shapes.

ts-sdk/src/sdk/dag.ts
jason810496: Both specs are now `Object.freeze({ ...spec })`, matching `inputs`, and the two `toBe` assertions relaxed to `toEqual`. I also added `Object.isFrozen` assertions and mutation-checked them and removed the `Object.freeze`.

ts-sdk/src/cli/pack.ts
jason810496: Both fixed. Task IDs are now checked as non-empty strings per `airflow-metadata.schema.json`, and the array case is rejected before `Object.entries` can turn it into Dags named `"0"`, `"1"`. Parametrized tests cover each malformed shape.

ts-sdk/src/cli/pack.ts
jason810496: Matched all the handling behavior with Go side: `warning: dag %q has no tasks` and zero registered Dags is still a hard error.

jason810496 (commented):
Thanks kaxil for the review, I just resolved all the comments.

ts-sdk/example/src/main.ts
```diff
@@ -49,7 +49,7 @@ export async function readConnection({ client }: TaskHandlerArgs) {
   };
 }
 
-registerTask({ dagId: DAG_ID, taskId: "build_message" }, buildMessage);
-registerTask({ dagId: DAG_ID, taskId: "read_connection" }, readConnection);
+dag.task("build_message", buildMessage);
+dag.task("read_connection", readConnection);
 
-await startCoordinator();
+await registerDags(dag);
```
uranusjr, line 55: I’m not sure how I feel with a function named `registerDags` handling sockets and a waiting server. Maybe use something more similar to the Java SDK such as `serveDags()`?

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  constructor(dagId: string, spec: DagSpec = {}) {
+    validateKey("dagId", dagId);
+    this.dagId = dagId;
+    // Copied and frozen, as task specs and inputs are: nothing reads a spec
+    // until the bundle manifest is built, long after the user's module has run,
+    // so a later mutation of their object would silently change what is packed.
+    // Shallow — a nested value in a future generated spec stays mutable.
+    this.spec = Object.freeze({ ...spec });
+    declaredDagIds.add(dagId);
```
uranusjr, line 153: Also not a big fan with using a global variable like this, especially in the constructor. Why not make the registry a public interface instead, similar to how the Java SDK has a Bundle class?

uranusjr:
Maybe we can do something like this instead

```ts
const dag = new Dag(...);
// Add tasks...
const registry = new DagRegistry(dag);
registry.serve();
```

Or, even more similar to Java SDK, `new Server(registry).serve()` so we can separate dag registration logic and network logic into different types.

ts-sdk/src/coordinator/runtime.ts
```diff
@@ -51,13 +52,46 @@ import {
...
 } from "./protocol.js";
-import { getRegisteredTask, listRegisteredTasks } from "../sdk/registry.js";
+import { DagRegistry } from "../sdk/registry.js";
 import type { TaskContext, TaskHandlerArgs } from "../sdk/task.js";
 import type { JsonValue } from "../sdk/client-types.js";
 
 export const ABORT_GRACE_PERIOD_MS = 30_000;
 export const COORDINATOR_RESPONSE_TIMEOUT_MS = 30_000;
 
+let served = false;
```
jason810496: This is the only place where the global variable lives and `serveDags` should only be called once. From my perspective, it's fine to keep it as global since the lifecycle and the intention is clear.

jason810496 (commented):
Thanks TP and Kaxil for the review!

I went with the following interface.
```ts
const dag1 = new Dag(...);
// Add tasks...

// Add Dags at constructor level
const registry = new DagRegistry(dag1, dag2);
// Register Dag incrementally 
registry.register(dag3, dag4);

await serveDags(registry); // should only be called once
```

The reason I used `registerDags` (I agreed `serveDags` is a better naming) in the first place is just to remove the "coordinator" term from the user interface and keep the entrypoint as simple as possible for the user.

I prefer to let `serveDags(registry)` do the final socket works.

For `new Server(registry).serve()` interface, IMO, we don't need to expose that much details to users. In the case of Python Dag world, we don't even have the user interface to register or serve the Dags at all.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,252 @@
...
+   * Register a TypeScript handler for a task of this Dag.
+   *
+   * `taskId` must match the Dag-side operator's `task_id` exactly, including
+   * any TaskGroup prefix. Returns this task's handle.
+   */
+  task<TReturn = unknown>(
+    taskId: string,
+    handler: TaskHandler<TReturn>,
+    options: TaskOptions = {},
+  ): TaskRef {
```
shivaam, line 172: Since task references are intended for future dependency chaining, should the return type at least be preserved as TaskOutput<TReturn> or TaskRef<TReturn>?

```
const extracted = dag.task("extract", async () => ({ rows: ["a"] }));
// TaskOutput<{ rows: string[] }>

dag.task("transform", transformHandler, {
  inputs: { extracted },
});
```

This would let TypeScript retain and check the value type passed from extract to transform.

shivaam:
Just curious about the longer-term direction. Do we expect tasks to become callable, similar to Python TaskFlow?
```
const result = load(transform(extract()));
```
These calls would build the Dag and return typed task outputs rather than execute the handlers.

Is something like this planned for future native TypeScript Dags, or is the current TaskRef plus inputs design intended to be the final authoring interface?

One option would be for dag.task() to return a callable TaskDefinition<TInputs, TOutput>. Calling that definition would record the input bindings and return a TaskOutput<TOutput>.

ts-sdk/src/sdk/dag.ts
jason810496: Good point! I would like to keep the interface as-is for now. Since they're no-op and user shouldn't set them at this moment (I added the `validateEmptySpec` to prevent it being set at all).

What the `TypeRef` and `TaskOutput` binding should be implemented in the follow-up PR.

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+/**
+ * Upstream task handles keyed by input name. **Reserved: validated and
+ * retained, but inert today** — see {@link TaskOptions.inputs}.
+ *
+ * Values must be handles returned by `dag.task(...)`. Literal values are
+ * deliberately out of scope for now; the future native-Dag work decides how
+ * they are declared.
+ */
```
shivaam, line 78: Super nit: I found some of the comments a little hard to scan and seems like they are written in response to review comments rather than as docs for someone hovering in an editor. It might help to use a consistent order?
- What it represents today  
- Current limitations  
- Future purpose

I also prefer simpler language such as “not used yet” instead of “inert.” For example:
```
/**
 * Task references keyed by input name.
 *
 * References are stored and validated, but they do not create dependencies or
 * pass values to handlers yet. Each reference must identify an earlier task
 * in the same Dag. Literal values are not currently supported.
 *
 * In the future, these references will be used to define dependencies in
 * native TypeScript Dags.
 */
```
Not suggesting to drop the reasoning as it is useful.

jason810496:
> Just curious about the longer-term direction. Do we expect tasks to become callable, similar to Python TaskFlow?

Yes, we will support the TaskFlow-style syntax to define the task dependencies as mentioned in https://github.com/apache/airflow/issues/69288#issuecomment-5093501655.

> One option would be for dag.task() to return a callable TaskDefinition<TInputs, TOutput>. Calling that definition would record the input bindings and return a TaskOutput.

This is what the `TaskRef` does today, though I haven't refine it into the "accurate" type like you mentioned, the `TaskRef` is more like a placeholder type right now.

ts-sdk/src/sdk/registry.ts
```diff
@@ -17,85 +17,84 @@
...
+ * ```
+ *
+ * It holds no sockets and starts nothing, so a test can build one and invoke a
+ * handler through {@link getTaskHandler} without any runtime in scope.
+ *
+ * Lookups delegate live to each Dag's task map, so tasks added to a Dag
+ * after registration are visible — the registry records Dag identity, not
+ * a snapshot of its tasks.
+ */
+export class DagRegistry {
```
shivaam, line 59: Do users need listTasks(), and listDags() directly? Since DagRegistry is exported from the package root, these become part of the public API. They are only required by the coordinator though. Could the public registry expose only registration, with package-internal helpers handling the rest?

shivaam (commented):
Thanks, looks good overall. Two potential nits.

ts-sdk/src/coordinator/runtime.ts
guan404ming: Nice cleanup overall! I think there are two small leftovers worth a look:
 
1. served is set before `startCoordinator` runs and never cleared on failure, so a rejected serveDags (e.g. Missing --comm) blocks retries with a misleading "already called" -> public-api.test.ts actually relies on this sequence. Moving the state onto the registry would fix cleanup and test isolation.

2. The instanceof DagRegistry / instanceof Dag checks reject genuine objects when two SDK copies are installed (dual-package hazard). A structural check would avoid that.

Nit: `serveDags` throws sync for these but rejects for runtime errors; async would unify the channels.

ts-sdk/src/cli/pack.ts
guan404ming, line 1: One nit consistency thought: pack trusts the manifest more than the SDK layer trusts TypeScript, no suspicious-id warnings (go-pack has `warnOnSuspiciousIDs`), `supervisor_schema_version` accepts any truthy value, and top-level null crashes with a raw `TypeError`. Failing closed here for everything entering the manifest might be worth it.

guan404ming (commented):
Overall looks nice, thanks for you work!

ts-sdk/example/src/main.ts
jason810496: Renamed in 30eff07ec7. `serveDags(registry)` is the entrypoint now, `registerDags` is gone.

ts-sdk/src/sdk/dag.ts
jason810496: Done in 30eff07ec7: the `Dag` constructor no longer writes to module state, and `DagRegistry` is the explicit object a bundle builds and hands to `serveDags`.

ts-sdk/src/coordinator/runtime.ts
jason810496: Both fixed in 745461eb32:

1. the latch is released when a serve fails (so a retry reports the real cause instead of "already called"), it is keyed on the global symbol registry rather than module state
2. The `serveDags` is now `async` so validation and runtime errors arrive on one channel.

Thanks!

ts-sdk/src/sdk/dag.ts
jason810496: Agreed and resolved in 8d3a3882bc , thanks!

ts-sdk/src/sdk/registry.ts
jason810496: Good point. I don't think `listTasks` and `listDags` are necessary as well, addressed in 6ae9254e1f
I kept `getTaskHandler` public deliberately: it is the documented way to unit-test a handler without a coordinator runtime.

ts-sdk/src/cli/pack.ts
jason810496: The `warnOnSuspiciousIDs ` issue will be address in https://github.com/apache/airflow/pull/70993 after we merge this one.

a21763416f resolved the validation for `supervisor_schema_version`

jason810496 (commented):
Thanks all for the review.

ts-sdk/src/cli/pack.ts
```diff
@@ -143,20 +143,52 @@ function readBundleManifest(bundlePath: string): BundleManifest {
...
+// null and primitives, and `null.supervisor_schema_version` would surface as a
+// raw TypeError rather than a report about the bundle.
+function isBundleManifest(value: unknown): value is BundleManifest {
+  if (typeof value !== "object" || value === null || Array.isArray(value)) return false;
+  const { supervisor_schema_version: version, dags } = value as Partial<BundleManifest>;
+  return (
+    // Rendered into the manifest verbatim, where the schema requires a non-empty
+    // string, so a truthy number or boolean would travel to Airflow as-is.
+    typeof version === "string" &&
+    version.length > 0 &&
```
uranusjr, line 187: Does this need to check the version format? Or do you think it’s too restricted?

ts-sdk/src/sdk/dag.ts
```diff
@@ -0,0 +1,258 @@
...
+function validateKey(name: string, value: string): void {
+  if (typeof value !== "string" || !KEY_REGEX.test(value)) {
+    throw new Error(
+      `${name} must be made of alphanumeric characters, dashes, dots, and underscores`,
+    );
+  }
+  if (value.length > MAX_KEY_LENGTH) {
+    throw new Error(`${name} must be less than ${MAX_KEY_LENGTH} characters, not ${value.length}`);
+  }
+}
```
uranusjr: Same question from the other direction—do we need this check?

ts-sdk/src/cli/pack.ts
jason810496: I think it's fine to keep it since the validation is in the packing stage instead of the runtime stage.

ts-sdk/src/sdk/dag.ts
jason810496: Therefore, I moved the validation into packing stage in https://github.com/apache/airflow/pull/71144/commits/f0edc1caea59087bb7bb86f6e052eed7f0df580c .

ts-sdk/src/cli/pack.ts
uranusjr: But should this be stricter and check against a regex?

ts-sdk/src/cli/pack.ts
uranusjr, line 142: Since key validation moved to buildBundleManifest, this catch is now the only place a bad Dag or task ID is reported. `String(error)` makes it hard to read. Packing `new Dag("bad id!")` currently gives:

```
Running the bundle with --airflow-metadata failed: Error: Command failed: /usr/bin/node /tmp/out/bundle.pack-staging.mjs --airflow-metadata
...
Error: Dag "bad id!" must be made of alphanumeric characters, dashes, dots, and underscores
    at validateKey (file:///tmp/out/bundle.pack-staging.mjs:2527:11)
    ...
```

The one actionable sentence is buried, and every frame points into bundle.pack-staging.mjs, which the finally block deletes — so no line the trace names can be opened.

`execFileSync` puts the child's stderr on error.stderr, so the message is already available here:

```ts
} catch (error) {
  const stderr = (error as { stderr?: string }).stderr ?? "";
  const reported = stderr.split("\n").find((l) => /^\w*Error: /.test(l.trim()))?.trim();
  throw new Error(
    reported ?? `Running the bundle with ${AIRFLOW_METADATA_FLAG} failed: ${String(error)}`,
    { cause: error },
  );
}
```

Worth considering to catch inside `--airflow-metadata` mode and emit the failure on the sentinel channel. That drops the stack entirely and lets pack distinguish "this Dag ID is invalid" from "the bundle threw on import", which are indistinguishable today.

ts-sdk/src/cli/pack.ts
jason810496: Yes, we will validate against a regex in https://github.com/apache/airflow/pull/70993.

ts-sdk/src/cli/pack.ts
jason810496: Addressed in https://github.com/apache/airflow/pull/71144/commits/07f2533f83dfc2bad368710a2162c3389f18ce11, thanks.

ts-sdk/src/cli/pack.ts
```diff
@@ -129,9 +129,15 @@ function readBundleManifest(bundlePath: string): BundleManifest {
...
     });
   } catch (error) {
-    throw new Error(`Running the bundle with ${AIRFLOW_METADATA_FLAG} failed: ${String(error)}`, {
-      cause: error,
-    });
+    const stderr = (error as { stderr?: string }).stderr ?? "";
+    const reported = stderr
+      .split("\n")
+      .find((line) => /^\w*Error: /.test(line.trim()))
+      ?.trim();
```
uranusjr, line 137: This takes the first match, but I think we should take the last instead.

Note that the process for stdout below uses a `reverse` to find the last instead.

uranusjr (approved):

ts-sdk/src/cli/pack.ts
jason810496: Addressed in https://github.com/apache/airflow/pull/71144/commits/5d853066b634663ba23d44585338df56bbf8e185, thanks.

PR #71145, opened 2026-08-05
Fix snowflake provider installs picking an unimportable pyOpenSSL

rjgoyln:
## Summary

The `Compat 2.11.1`, `Compat 3.0.6` and `Compat 3.1.8` provider jobs fail collection on main with `AttributeError: module 'lib' has no attribute 'GEN_EMAIL'`; 3.2.2 and 3.3.0 are unaffected.

Those jobs resolve the freshly built distributions unpinned (`--providers-skip-constraints`), and with the apache/beam provider suspended nothing in the set caps `cryptography` any more, so it goes to the newest release. `snowflake-connector-python` below 4.4.0 requires `pyOpenSSL<26.0.0`, and the only candidate in that window claiming compatibility with a current cryptography is 22.0.0 — which declares `cryptography>=35.0` with no upper bound at all. The resolver takes it, backtracks the connector to 4.0.0 to match, and lands on an environment that installs cleanly and then dies the moment anything imports `urllib3.contrib.pyopenssl`.

22.1.0 is the first pyOpenSSL release that bounds cryptography honestly, so the floor drops exactly the releases that lie and nothing else. Raising `snowflake-connector-python` to `>=4.4.0` — the release that dropped the cap — would be the tidier fix but requires `cryptography>=46.0.5`, unsatisfiable against the constraints of every Airflow version these jobs cover.

## Tests

Resolving main's full provider dependency set for Python 3.10: unpinned now selects connector 4.7.1 with pyOpenSSL 26.3.0, and pinning cryptography to each value the released constraint files carry (41.0.7 / 42.0.8 / 44.0.3 / 45.0.7) still yields an importable pair.

---


jason810496 (approved):
Thanks.

The only CI failure should be resolved in https://github.com/apache/airflow/pull/71140. The CI should come back to green after rebasing.

amoghrajesh:
Taking a look.

amoghrajesh (approved):
LGTM, created an issue to track this: https://github.com/snowflakedb/snowflake-connector-python/issues/2990

FrankYang0529 (approved):

Lee-W (approved):

aaron-y-chen (approved):
Thanks for the fix, my PR also encountered these failures.

jason810496:
Nice!

PR #71151, opened 2026-08-05
Add Java SDK capability manifest and compatibility matrix

jason810496:
### Why

The [Language SDK conformance spec](https://github.com/apache/airflow/blob/main/contributing-docs/30_new_language_sdk.rst) (#70717) defines which TaskInstance states and capabilities a Language SDK may declare, but nothing lets an SDK say what it actually supports.

This makes that support machine-readable and generates the docs from it.

### How

```
java-sdk/capabilities.yaml          <- the only file you edit
  |
  |  render: scripts/ci/prek/lang_sdk_compat_matrix.py
  |  hook: update-java-sdk-readme-matrix
  |
  +--> java-sdk/README.md            (contributor-facing)
  |
  +--> java-sdk/sdk/module.md        (Dokka module doc -> the published API reference)
```

### Rejected Alternative

- Define the capabilities spec in [java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/Capabilities.kt](https://github.com/apache/airflow/pull/71151/files#diff-c8830815e96d1a7f471256c51ce8e3eaa18173a560730026bdfd642efd8c8b91) -- too complicated, YAML fulfill the same functionality without the gradle call.

### Verfication

Screenshot 2026-08-05 at 1 24 37 PM

---

- [X] Yes — Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -280,6 +280,21 @@ repos:
           (?x)
           ^java-sdk/gradle\.properties$|
           ^java-sdk/sdk/schema/schema\.json$
+      - id: update-java-sdk-readme-matrix
+        name: Update the Java SDK compatibility matrix in java-sdk/README.md and Dokka module doc
+        entry: ./scripts/ci/prek/update_java_sdk_readme_matrix.py
+        language: python
+        files: >
+          (?x)
+          ^java-sdk/capabilities\.yaml$|
+          ^java-sdk/gradle\.properties$|
+          ^java-sdk/README\.md$|
+          ^java-sdk/sdk/module\.md$|
+          ^scripts/ci/prek/lang_sdk_compat_matrix\.py$|
+          ^scripts/ci/prek/update_java_sdk_readme_matrix\.py$
+        additional_dependencies: ['PyYAML>=6.0', 'rich>=13.6.0']
+        pass_filenames: false
+        require_serial: true
       - id: check-go-version-in-sync
         name: Check Go toolchain version is consistent across build files
         entry: ./scripts/ci/prek/check_go_version_in_sync.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -292,7 +292,9 @@ repos:
           ^java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/.*\.kt$|
           ^java-sdk/sdk/build\.gradle\.kts$|
           ^java-sdk/gradle\.properties$|
-          ^java-sdk/generated/lang-sdk/capabilities\.json$
+          ^java-sdk/generated/lang-sdk/capabilities\.json$|
+          ^scripts/ci/prek/check_java_sdk_capabilities_in_sync\.py$|
+          ^scripts/ci/prek/lang_sdk_compat_matrix\.py$
         pass_filenames: false
         require_serial: true
       - id: update-java-sdk-readme-matrix
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -280,6 +280,35 @@ repos:
           (?x)
           ^java-sdk/gradle\.properties$|
           ^java-sdk/sdk/schema/schema\.json$
+      - id: check-java-sdk-capabilities-in-sync
+        name: Check java-sdk/generated/lang-sdk/capabilities.json matches the Kotlin capability constant
+        entry: ./scripts/ci/prek/check_java_sdk_capabilities_in_sync.py
+        # system so the JDK/Gradle wrapper the dump task needs is taken from the host toolchain,
+        # matching sync-java-sdk-supervisor-schema above. The Python entry only imports the
+        # dependency-free lang_sdk_compat_matrix/common_prek_utils (no rich), so system Python suffices.
+        language: system
+        files: >
+          (?x)
+          ^java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/.*\.kt$|
+          ^java-sdk/sdk/build\.gradle\.kts$|
+          ^java-sdk/gradle\.properties$|
+          ^java-sdk/generated/lang-sdk/capabilities\.json$
+        pass_filenames: false
+        require_serial: true
+      - id: update-java-sdk-readme-matrix
+        name: Update the Java SDK compatibility matrix in java-sdk/README.md and Dokka module doc
+        entry: ./scripts/ci/prek/update_java_sdk_readme_matrix.py
+        language: python
+        files: >
+          (?x)
+          ^java-sdk/generated/lang-sdk/capabilities\.json$|
+          ^java-sdk/README\.md$|
+          ^java-sdk/sdk/module\.md$|
+          ^scripts/ci/prek/lang_sdk_compat_matrix\.py$|
+          ^scripts/ci/prek/update_java_sdk_readme_matrix\.py$
+        additional_dependencies: ['rich>=13.6.0']
+        pass_filenames: false
+        require_serial: true
       - id: check-go-version-in-sync
         name: Check Go toolchain version is consistent across build files
         entry: ./scripts/ci/prek/check_go_version_in_sync.py
--- a/contributing-docs/30_new_language_sdk.rst
+++ b/contributing-docs/30_new_language_sdk.rst
@@ -500,7 +500,7 @@ Compatibility matrix
 ~~~~~~~~~~~~~~~~~~~~
 
 The dimensions above are prose; each SDK also declares them *machine-readably* in one
-hand-authored ``<sdk>/capabilities.yaml``, from which a prek hook generates every table:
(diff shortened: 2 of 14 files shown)


java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/Capabilities.kt
```diff
@@ -0,0 +1,116 @@
...
+  val states: Map<String, CapabilityEntry>,
+  val capabilities: Map<String, CapabilityEntry>,
+)
+
+private const val MIN_AIRFLOW_VERSION = "3.3"
+
+// Keep in sync with airflowSupervisorSchemaVersion in java-sdk/gradle.properties.
+private const val SUPERVISOR_SCHEMA_VERSION = "2026-06-16"
+
+private fun yes(note: String = ""): CapabilityEntry = CapabilityEntry(supported = true, since = MIN_AIRFLOW_VERSION, note = note)
```
phanikumv:
```suggestion
  private fun yes(note: String = "", since: String = MIN_AIRFLOW_VERSION): CapabilityEntry =
    CapabilityEntry(supported = true, since = since, note = note)
  ```

Comment (jason810496) [reply java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/Capabilities.kt]: Applied in c8d92441fc , thanks,

File: java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/Capabilities.kt [hunk 2]
```diff
@@ -0,0 +1,119 @@
...
+package org.apache.airflow.sdk.conformance
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+// This model exists only to emit the build-time manifest, so it is deliberately `internal`: it is
+// not part of the SDK's published surface, and keeping it out means changing the manifest schema
+// never becomes a source/binary compatibility concern for SDK users.
```
uranusjr: Do we really need this? The module not referenced anywhere, and also not referencing anything. We can probably just write the JSON manually and use `validate_capabilities` to ensure it’s always good.

Or maybe the format can be YAML instead to be more developer-friendly. (The reader side wouldn’t care much.)

uranusjr: Also as currently stand this _will_ be included in the release since it’s in `main`.

.pre-commit-config.yaml
```diff
@@ -280,6 +280,37 @@ repos:
           (?x)
           ^java-sdk/gradle\.properties$|
           ^java-sdk/sdk/schema/schema\.json$
+      - id: check-java-sdk-capabilities-in-sync
```
uranusjr: This should probably be added to selected_checks.

.pre-commit-config.yaml
```diff
@@ -280,6 +280,37 @@ repos:
...
+        name: Check java-sdk/generated/lang-sdk/capabilities.json matches the Kotlin capability constant
+        entry: ./scripts/ci/prek/check_java_sdk_capabilities_in_sync.py
+        # system so the JDK/Gradle wrapper the dump task needs is taken from the host toolchain,
+        # matching sync-java-sdk-supervisor-schema above. The Python entry only imports the
+        # dependency-free lang_sdk_compat_matrix/common_prek_utils (no rich), so system Python suffices.
+        language: system
```
uranusjr: I’m not following; using `language: python` simply means a virtual environment can be used, and Java in the host should still be available. I don’t see an advantage using `system` here. (This would be moot if the capabilities are declared directly instead of pulled from Kotlin.)

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/conformance/Capabilities.kt
jason810496: Thanks for the review.

I made it a kotlin module to make agents "remember to update the capability spec" in the first place. Good idea to make it a pure yaml (we can save the gradle call, JDK for the static check). Additionally, having a one line rule in the AGENTS.md can still make the toolings remener to update the spec even it's in yaml format.

.pre-commit-config.yaml
jason810496: The `check-java-sdk-capabilities-in-sync` was removed so no need to update the selected check now. (prek can do all the stuff).

.pre-commit-config.yaml
jason810496: Same as the previous comment, no need Java at all now.

scripts/ci/prek/lang_sdk_compat_matrix.py
```diff
@@ -0,0 +1,294 @@
...
+            raise CapabilitiesError(f"{source}: {kind}.{name} must be a mapping with a boolean 'supported'")
+        # A misspelled optional key would otherwise be dropped silently and render as a blank cell.
+        unknown_fields = entry.keys() - {"supported", "since", "note"}
+        if unknown_fields:
+            raise CapabilitiesError(
+                f"{source}: {kind}.{name} has unknown keys: {', '.join(sorted(unknown_fields))}"
+            )
+        if not isinstance(entry.get("since", None), (str, type(None))):
+            raise CapabilitiesError(f"{source}: {kind}.{name}.since must be a string or null")
+        if not entry["supported"] and entry.get("since") is not None:
```
uranusjr, line 240: Logic here diverges from `_capability_mark`, so if this is set

```yaml
branching:
  supported: true
  since: 1
native-dag-authoring:
  supported: false
```

The verification would pass, but then capability rendered as `n/a`. The end result is… fine? Since it’s (probably) not technically possible to support branching without native-dag-authoring in the first place, but maybe we should catch this directly here?

contributing-docs/30_new_language_sdk.rst
```diff
@@ -491,6 +496,32 @@ native Dags they are *not applicable* (``n/a``) rather than unsupported.
...
+.. code-block:: text
+
+    java-sdk/capabilities.yaml          <- the only file you edit
+      |
+      |  hook: update-java-sdk-readme-matrix
+      |
+      +--> java-sdk/README.md            (contributor-facing)
+      +--> java-sdk/sdk/module.md        (Dokka -> the published API reference)
+      +--> airflow-core/docs/authoring-and-scheduling/language-sdks/index.rst
+           (consolidated cross-SDK matrix — TODO, not generated yet)
```
uranusjr: Do we track this somewhere? It’s better to reference the issue directly here instead of just a plain text TODO.

java-sdk/sdk/dokka/matrix.css
```diff
@@ -0,0 +1,42 @@
...
+
+/* Dokka sizes content-table cells with `min-width: 40px` against ~31px of horizontal padding, so the
+   narrow columns of the Language SDK compatibility matrix collapse until their headers and values
+   break mid-word ("Supported", "MUST †").
+
+   These selectors necessarily apply to every content table in the API reference, because a table
+   generated from Markdown carries no class of its own. They are therefore deliberately written to
+   only *raise a floor*, never to forbid wrapping: a header stays on one line (headers are short in
+   any table), and the narrow value columns get a minimum width instead of `nowrap`. An unrelated
+   table with long prose in those columns still wraps as before rather than overflowing. */
```
uranusjr: This is probably five times too long for a `nowrap`.

scripts/ci/prek/lang_sdk_compat_matrix.py
jason810496: No harm to have a strict validation. I addressed in https://github.com/apache/airflow/pull/71151/changes/aa3b1b3ac48ef9e4f7bdc9f8d75d4a03f3533fe8, thanks.

java-sdk/sdk/dokka/matrix.css
jason810496: Agreed, I trimmed the comments in https://github.com/apache/airflow/pull/71151/changes/aa3b1b3ac48ef9e4f7bdc9f8d75d4a03f3533fe8#diff-f58bd21e679ab8560bf95a8473dc5a3b428d54f841f3475f69e05f05795a89a7 , thanks.

contributing-docs/30_new_language_sdk.rst
jason810496: https://github.com/apache/airflow/pull/71757 will wire the matrix for all the SDKs. I remove the TODO here for now.

jason810496 (commented):
Thanks TP for the review.

uranusjr (approved):

Issue #68846, opened 2026-06-22
UI logs selection issues (Airflow 3.2.2)


PolitePp:
### Under which category would you file this issue?

Airflow Web UI

### Apache Airflow version

3.2.2

### What happened and how to reproduce it?

Hi folks! I'm trying to select some logs in my airflow 3.2.2 (a bit more that could be placed in one screen). I'm pressing left mouse button and start selecting text. I'm moving mouse cursor down to the bottom of the viewport and it doesn't work as I expect

https://github.com/user-attachments/assets/b509475e-8caf-4d2b-ae7a-bb3ebf8b345f

### What you think should happen instead?

The behaviour should be similar to Airflow 2.x

### Deployment

Official Apache Airflow Helm Chart

### Official Helm Chart version

Not Applicable

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

Andrushika:
Hi, I found the root cause: the virtualized log display unmounts the DOM nodes for log rows outside the visible window.
I found the solution, and will open a follow-up PR once #70594 is merged.

PR #71156, opened 2026-08-05
Fix copying task logs dropping rows that scrolled out of view

Andrushika:
## Why

After #71148, when the user selects and tries to copy task log rows, virtualized rows that scrolled out of view are missing. It's because native copy only serializes mounted DOM. 

## What

Rebuild missing selected rows from parsed log data with the same plain-text rendering as the log download. Keep native copy for single-row, fully mounted, and multi-range selections.

stacked on: #71148
related: #68846

Before:

https://github.com/user-attachments/assets/ef714fe9-596a-46de-9ae2-8a81a52f719a

After:

https://github.com/user-attachments/assets/14dc4db7-9c65-4480-af33-5bd1085727d1

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/airflow-core/src/airflow/ui/src/queries/useLogs.tsx
+++ b/airflow-core/src/airflow/ui/src/queries/useLogs.tsx
@@ -17,6 +17,7 @@
  * under the License.
  */
 import type { UseQueryOptions } from "@tanstack/react-query";
+import Anser from "anser";
 import dayjs from "dayjs";
 import type { TFunction } from "i18next";
 import type { JSX } from "react";
@@ -24,7 +25,11 @@ import { useTranslation } from "react-i18next";
 import innerText from "react-innertext";
 
 import { useTaskInstanceServiceGetLog } from "openapi/queries";
-import type { TaskInstanceResponse, TaskInstancesLogResponse } from "openapi/requests/types.gen";
+import type {
+  StructuredLogMessage,
+  TaskInstanceResponse,
+  TaskInstancesLogResponse,
+} from "openapi/requests/types.gen";
 import {
   extractTIContext,
   renderStructuredLog,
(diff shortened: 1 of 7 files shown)


airflow-core/src/airflow/ui/src/queries/useLogs.tsx
```diff
@@ -36,10 +41,47 @@ import { parseStreamingLogContent } from "src/utils/logs";
...
+  sourceFilters,
+  translate,
+}: GetLogLineTextOptions): string =>
+  Anser.ansiToText(
+    renderStructuredLog({
+      index: 0,
+      logLevelFilters,
+      logLink: "",
+      logMessage,
+      renderingMode: "text",
```
bbovenzi, line 77: We are mixing selecting by renderingMode. In order for that to work best, we will need to harmonize the two modes more such as the timestamp format and any additional elements we add like group header ▶

bbovenzi: This can be done in a separate PR

bbovenzi (commented):
let's rebase

airflow-core/src/airflow/ui/src/queries/useLogs.tsx
Andrushika: Make sense. I'll open a follow-up PR to handle this case.

Andrushika: One point I want to check: do we want the rebuilt text to match what the screen shows? Currently the copied log row can come from either JSX text or plain text, for example (the main difference is the timestamp):

JSX:
```
[2026-08-05 15:59:48] INFO - line 0001 | lorem ipsum dolor sit amet consectetur adipiscing elit | payload=1
```
plain text:
```
[2026-08-05T07:59:48.737536Z] INFO - line 0001 | lorem ipsum dolor sit amet consectetur adipiscing elit | payload=1
```

So I think we need to choose one of them for copy formatting. In this PR, when the user selects partial first and last rows, that text can only come from the DOM, so the rebuilt middle rows need to move toward the screen format, not the other way around.

BTW, I prefer to keep the ▶ marker in the copied group headers for readability. wdyt?

Andrushika:
Rebased. Thanks for reviewing!
I guess we can merge this first and I'll create follow up PR for unifying copy format.

bbovenzi (approved):

airflow-core/src/airflow/ui/src/queries/useLogs.tsx
bbovenzi: I think a user is expecting their copy to be what they see on the screen. So yes the timestamp and the arrow marker.

PR #71170, opened 2026-08-05
Fix OpenSearch remote logging crash if port is empty

FrankYang0529:
Followup PR for https://github.com/apache/airflow/pull/70295#discussion_r3701681474.

## Why

- `[opensearch] port` declares `default: ""` in `provider.yaml`, so the key always resolves and `conf.getint("opensearch", "PORT", fallback=9200)` never reaches its fallback. The default value `""` make it raise ValueError on `int("")`.

## How

- Read `[opensearch] port` as a string and convert explicitly, treating an empty value as "unset". An unset port still resolves to 9200.

## Verification

- `uv run --project providers/opensearch pytest airflow-core/tests/unit/config_templates/test_airflow_local_settings.py`
- `env -u AIRFLOW__OPENSEARCH__PORT AIRFLOW__LOGGING__REMOTE_LOGGING=True AIRFLOW__OPENSEARCH__HOST=https://opensearch.example.com uv run --project providers/opensearch airflow version`
  - This command fails on main branch, but pass in this PR.

 

---

- [X] Yes - Claude Code

---

### DIFF
--- a/airflow-core/src/airflow/config_templates/airflow_local_settings.py
+++ b/airflow-core/src/airflow/config_templates/airflow_local_settings.py
@@ -357,7 +357,10 @@ if REMOTE_LOGGING:
     elif OPENSEARCH_HOST:
         from airflow.providers.opensearch.log.os_task_handler import OpensearchRemoteLogIO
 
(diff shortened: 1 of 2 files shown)


jason810496 (approved):
Thanks for the follow-up.

PR #71174, opened 2026-08-05
Apply KubernetesPodOperator template fields only after rendering

rjgoyln:
## Summary

`KubernetesPodOperator` converted and merged several template-field values in `__init__`, which runs before Jinja rendering. A templated `base_container_name` was copied into the default `container_logs` and never rendered, so log fetching asked for a container that did not exist. A templated `configmaps` string was iterated character by character into `env_from`, and templated `volumes` / `volume_mounts` were rejected at Dag parse time before they could resolve.

`__init__` now only stores these arguments; the conversions run in `build_pod_request_obj()` and `container_logs` resolves its base-container fallback lazily. `configmaps` becomes a declared template field, having previously been rendered by accident through the `env_from` list it was merged into.

An `env_vars` mapping is the one conversion that still precedes rendering, in a `render_template_fields` override: an env var *name* is a template field of `V1EnvVar`, while a plain dict renders only its values. Only the container shape is inspected, never the value.

## Behavior change

`SparkKubernetesOperator` no longer warns that `container_logs` "is not supported and will be overridden" when the user never passed it; an explicitly requested container is still overridden to the driver container.

Clears the `KubernetesPodOperator` entry of the burn-down tracked in #70296.

---


providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
```diff
@@ -533,6 +530,23 @@ def __init__(
...
+    @container_logs.setter
+    def container_logs(self, value: Iterable[str] | str | Literal[True] | None) -> None:
+        self._container_logs = value
+
+    def render_template_fields(self, context: Context, jinja_env: jinja2.Environment | None = None) -> None:
+        # A str-str mapping has to become V1EnvVar objects before rendering: rendering a dict
+        # covers only its values, whereas an env var name is a template field of V1EnvVar.
+        if isinstance(self.env_vars, dict):
+            self.env_vars = convert_env_vars(self.env_vars)
+        super().render_template_fields(context, jinja_env)
```
shahar1: Bypassed for mapped tasks as Task SDK's `MappedOperator.render_template_fields` calls `unmapped_task._do_render_template_fields(...)` directly.
Please consider the mapped-path case.

shahar1 (changes requested):

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
rjgoyln: Confirmed — thanks. I moved the conversion into _do_render_template_fields, which is shared by both the regular and mapped rendering paths.

While auditing the same gap in @task.kubernetes, I found that its execute() additionally expands self.env_vars directly, so the mapped case could treat mapping keys as env var names. Since the same conversion fixes both cases, I’ve included it here and added a regression test covering .partial().expand().

shahar1 (approved):

shahar1:
Great work!

Issue #56632, opened 2025-10-14
Dynamic DAG parameter options based on external configuration (dropdown / radio support)


rolfschleiss:
### Description

When triggering a DAG in the Airflow web UI, all parameters defined in the params section are currently rendered as plain text input fields.
It would be highly useful if Airflow supported dynamic UI field types (e.g. dropdowns or radio buttons) that can populate their options from external configuration files, such as INI, YAML or JSON.

### Use case/motivation

In many environments, DAGs depend on configuration files that define available interfaces, systems, or targets.
For example, an INI file might describe multiple interfaces with different types:

```
[InterfaceA]
TYPE = Script

[InterfaceB]
TYPE = EBICS

[InterfaceC]
TYPE = Script
```

When triggering the DAG, users should be able to select from all sections that match a given condition (e.g. TYPE = Script), using a dropdown or radio button instead of free-text entry.

Example goal:

Show a dropdown listing InterfaceA and InterfaceC, because both are defined as TYPE = Script in the INI file.

This would make DAG triggering:

- safer (no typos in interface names),
- more dynamic (values adapt automatically to config changes),
- and much more user-friendly for non-technical operators.

Alternatively, Airflow could provide a plugin interface to populate parameter choices dynamically,
so that DAG authors can implement their own logic for which options should appear.

Benefits

- Greatly improves safety and UX for parameterized DAGs.
- Keeps Airflow configuration-driven (no need to hardcode options).
- Enables dynamic parameter lists that reflect live configuration.
- Avoids invalid values and reduces operational mistakes.

Possible Implementation Notes

- Extend the Web UI (React) to render dropdowns/radio fields when ui metadata is provided.
- Allow a DAG to expose a small API or callable hook to fetch dynamic choices.
- Could reuse existing Airflow variable/config reading mechanisms.

Proposed Change

Allow params definitions to reference an external source for valid options.

Example:

```
with DAG(
    "interface_trigger_dag",
    params={
        "interface_name": {
            "source": "config/interfaces.ini",
            "filter": {"TYPE": "Script"},
            "key": "section_name",
            "ui": "dropdown"  # or "radio"
        }
    },
) as dag:
    ...
```

Labels:
enhancement, UI, params, feature-request, dynamic-config

### Related issues

#56427 – [DAG Params Enum not works as expected](https://github.com/apache/airflow/issues/56427?utm_source=chatgpt.com)
: Dropdown rendering for enum params not functioning correctly.

[...]

PR #71180, opened 2026-08-05
Document config-backed enum Dag Params

fat-catTW:
## Why

Dag authors sometimes need the choices for an enum Param to come from deployment-local
configuration rather than being hard-coded in the Dag file. Airflow already supports this —
the trigger form reads the serialized Param schema, so any list computed while the Dag is
parsed shows up as a dropdown — but nothing in the docs or the examples shows the pattern,
so it isn't obvious that it works.

## What

Documents and demonstrates building an enum Param's choices from an external config file:

- `params.rst` notes that choices can be read from a config file at parse time, and that the
  trigger form uses the serialized Params rather than calling any dynamic choice provider
  when the form opens.
- `example_params_ui_tutorial.py` gains a `script_interface` Param whose choices come from a
  colocated INI file, filtered to the sections with `TYPE = Script`.
- `example_params_ui_interfaces.ini` is the accompanying config file.

No behavior change: enum Params still render as dropdowns, and no new schema fields or UI
code are added.

## Scope

An earlier revision of this PR proposed an `x-airflow-ui.widget = "radio"` schema hint so
enum Params could render as radio buttons. That was dropped after review — the naming and
the expansion story for `x-airflow-ui` need to be settled first, and a radio has no way to
clear the selection for an optional Param. The dropdown already covers the "selector instead
of free text" ask in #56632, so this PR keeps to the documentation half.

related: #56632

## Screenshot

Trigger Dag form showing the `Script interface` Param rendering as a dropdown with the
choices read from the INI file (`Interface A`, `Interface C` — `InterfaceB` is filtered out
because its `TYPE` is not `Script`):

螢幕擷取畫面 2026-08-07 223941

- [X]` Yes (please specify the tool below)

---

### DIFF
--- a/airflow-core/docs/core-concepts/params.rst
+++ b/airflow-core/docs/core-concepts/params.rst
@@ -253,6 +253,11 @@ The following features are supported in the Trigger UI Form:
             * ``format="multiline"``: Generate a multi-line textarea
             * | ``enum=["a", "b", "c"]``: Generates a
               | drop-down select list for scalar values.
+              | If the choices come from an external config file,
+              | read that file while the Dag is parsed and pass the
+              | resulting list to ``enum``; the trigger form reads
+              | the serialized Dag Params and does not call dynamic
+              | choice providers when the form opens.
               | As of JSON validation, a value must be
               | selected or the field must be marked as
               | optional explicit. See also details inside
--- a/airflow-core/newsfragments/71180.improvement.rst
+++ b/airflow-core/newsfragments/71180.improvement.rst
@@ -1,0 +1,1 @@
+Documented how to populate enum Dag Params from external config files for the trigger Dag form.
--- a/airflow-core/newsfragments/71180.improvement.rst
+++ b/airflow-core/newsfragments/71180.improvement.rst
@@ -1,1 +1,0 @@
-Documented how to populate enum Dag Params from external config files for the trigger Dag form.
--- a/airflow-core/src/airflow/example_dags/example_params_ui_interfaces.ini
+++ b/airflow-core/src/airflow/example_dags/example_params_ui_interfaces.ini
@@ -1,0 +1,25 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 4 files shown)


bbovenzi (changes requested):
I'm wary of adding a new `x-airflow-ui` option to declaring params. First the name isn't great. Second, we should define how we plan to expand this in the future.

Second, I am not sure if this closes the old issue because the person asked for a selector or a radio over plain text. I don't think it matters as much which of the two exist.

Third, the radio has no option to unselect when it is an optional param.

fat-catTW:
Good afternoon.
Thanks for the review. I agree with the concerns here.

To clarify the current state of this PR: 
The example demonstrates that a Dag author can read an INI file at Dag parse time, filter the available interfaces, and pass those values into a Param enum. Airflow Params currently expect the enum choices to be materialized when the Dag is parsed; the trigger form reads the serialized Param schema and does not have a generic "choices source" mechanism that can load/filter options when the form opens.

The trigger form already renders enum values as choices via the existing dropdown support, and this PR experiments with rendering the same enum values as radio buttons via `x-airflow-ui`. So this covers the “do not require free-text input” part, but it does not introduce a general dynamic choices provider or a reusable config-source mechanism.

I also agree that the selector vs radio part is secondary. The main missing design question is how Airflow should expose UI presentation preferences for Params in a way that can be extended later. The current `x-airflow-ui` key may be too ad hoc without that broader shape being defined.

Given that, I think there are two possible directions:

1. Keep this PR smaller and use the existing enum dropdown behavior, documenting the INI-at-parse-time pattern without adding a new UI extension key.
2. Expand this PR into a more deliberate Param UI presentation API, including how widgets such as dropdown/radio would be selected and how optional radio Params can be cleared.

I’m happy to take either direction. If maintainers think this PR should define the broader UI extension shape, I can work on that here; otherwise I can reduce this PR to the existing enum dropdown/docs pattern.

bbovenzi:
Let's go with option 1. I think smaller PRs are always a better start and that gets to the core use-case.

fat-catTW:
Thanks, I updated the PR to follow option 1.

The PR no longer adds `x-airflow-ui` or a radio widget path. It now keeps the existing enum dropdown behavior and documents/demonstrates the smaller pattern: read an external INI file while the Dag is parsed, filter the available choices, and pass the resulting list into a Param `enum`.

I also verified this manually in the UI with `Params UI tutorial`: the new `Script interface` field renders as a dropdown with `Interface A` and `Interface C`, while `Interface B` is excluded because it is not `TYPE = Script`.

螢幕擷取畫面 2026-08-07 223941

bbovenzi (approved):

uros-b (approved):

uros-b:
+1, thank you @fat-catTW!

PR #71186, opened 2026-08-05
Adopt flit 4 as the provider distribution build backend

### LINKED ISSUE #71121: Adopt flit 4 as the provider distribution build backend deliberately
`dev/breeze/pyproject.toml` caps `flit` and `flit-core` below 4, so provider distributions keep being built by the backend CI has actually exercised. The cap should come off as part of a deliberate flit 4 adoption, not by relaxing the specifier on its own.

### Why the cap is there

`flit build` does not use PEP 517 build isolation — `flit/wheel.py` is `import flit_core.wheel as core_wheel`, so the backend is whichever `flit_core` is installed in the environment. That makes the `flit_core==3.12.0` pin every provider declares in `[build-system] requires` inert on the path we release from, and leaves breeze's own dependency as the only thing deciding it. Because that dependency was an unbounded floor, flit 4.0.1 became the provider build backend within minutes of being published, without anyone choosing it.

flit 4 is not a drop-in for that role. It emits core metadata 2.5 with the new `Import-Name` and `Import-Namespace` fields, makes `--no-use-vcs` the default for sdists, and applies the tar `data` filter on Python 3.12+. None of that has been exercised against Airflow's provider distributions or its reproducible-build checks.

Both distributions need the cap: `flit` 4.x requires `flit_core>=4`, and `flit` 3.12.0 requires `flit_core>=3.12.0` with no upper bound, so capping only `flit` would still admit the 4.x backend.

The defect that exposed this — flit_core 4.0.1 writing `Import-Name` without a trailing newline, which cost the resulting `METADATA` its header/body separator — was fixed upstream in https://github.com/pypa/flit/issues/813 and released as 4.0.2. It is no longer a reason to hold the cap; the unvalidated backend change is.

### Follow-up work

Validate flit 4 as the provider build backend — metadata 2.5 acceptance by the tooling that consumes provider distributions, sdist contents with `--use-vcs` still passed explicitly, and reproducible-build output — then move `flit`, `flit-core` and the per-provider `flit_core==` pins together through `scripts/ci/prek/upgrade_important_versions.py`.

Note that the script rewrites these specifiers in place by version number, so a bump would turn `flit>=3.12.0,<4` into `flit>=4.0.1,<4`. The same hazard already exists for the `flit_core >=3.12.0,<4` cap on breeze's own build backend, and is worth fixing when the caps are revisited.

### Acceptance criteria

- Provider wheels and sdists built with flit 4 match what 3.12.0 produces apart from the intended metadata additions.
- `Generate constraints`, `Providers sdist tests`, `Providers wheel tests` and the `Compat` jobs are green.
- The specifiers in `dev/breeze/pyproject.toml` are floors again, moved by the upgrade script alongside the provider `flit_core==` pins, and the tracking comment at the cap site is removed.

---

potiuk:
`flit build` has no PEP 517 build isolation — `flit/wheel.py` does `import flit_core.wheel as core_wheel`, so the backend is whichever `flit_core` is installed. That makes the `flit_core==` pin every provider declares inert on the path we release from, and leaves breeze's own dependency as the only thing deciding it. The specifiers were capped below 4 so that backend could not change by accident; this takes the cap off deliberately.

### Validation

Built `apache-airflow-providers-imap` with 3.12.0 and with 4.0.2, same `SOURCE_DATE_EPOCH`, `--use-vcs` passed explicitly in both:

- wheel members identical — 13 files
- sdist members identical — 33 entries
- `METADATA` differs only by the intended additions:

```diff
-Metadata-Version: 2.4
+Metadata-Version: 2.5
+Import-Name: airflow.providers.imap
+Import-Namespace: airflow
+Import-Namespace: airflow.providers
```

The metadata parses cleanly, so the 4.0.1 defect that cost `METADATA` its header/body separator is confirmed fixed in 4.0.2.

### Floors, not bumped caps

The issue notes the hazard: `upgrade_important_versions.py` rewrites these by version number, so bumping `flit>=3.12.0,<4` would have produced `flit>=4.0.2,<4`. `flit`, `flit-core` and the 104 per-provider `flit_core==` pins now all move together at `4.0.2`, and the tracking comment at the cap site is removed.

### One addition the issue does not mention

flit 4.0.2 was published 2026-08-04, inside the `"4 days"` `exclude-newer` window, so `uv` refuses it and the adoption would not take effect. A temporary per-package cutoff is added in both `[tool.uv.exclude-newer-package]` and `[tool.uv.pip.exclude-newer-package]`, following the existing `pydantic-ai-skills` precedent, with a note to remove it once the rolling window advances past 2026-08-04.

closes: #71121

Generated-by: Claude Opus 5 (1M context) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1561,6 +1561,14 @@ apache-airflow-task-sdk-integration-tests = false
 # Remove once the rolling window advances past 2026-07-15.
 pydantic-ai-skills = "2026-07-16T00:00:00Z"
 
+# Temporary: flit / flit-core 4.0.2 were published 2026-08-04, inside the "4 days"
+# exclude-newer window. `flit build` has no PEP 517 isolation, so the installed flit_core
+# is the backend that builds every provider distribution - it has to be resolvable for the
+# 4.x adoption to take effect. Pins the per-package cutoff just past that release.
+# Remove once the rolling window advances past 2026-08-04.
+flit = "2026-08-05T00:00:00Z"
+flit-core = "2026-08-05T00:00:00Z"
+
 [tool.uv.pip]
 # Synchroonize with scripts/ci/prek/upgrade_important_versions.py
 exclude-newer = "4 days"
@@ -1708,6 +1716,14 @@ apache-airflow-task-sdk-integration-tests = false
 # `[tool.uv.exclude-newer-package]` above for rationale.
 # Temporary: see the pydantic-ai-skills note in `[tool.uv.exclude-newer-package]`.
 pydantic-ai-skills = "2026-07-16T00:00:00Z"
+
+# Temporary: flit / flit-core 4.0.2 were published 2026-08-04, inside the "4 days"
+# exclude-newer window. `flit build` has no PEP 517 isolation, so the installed flit_core
+# is the backend that builds every provider distribution - it has to be resolvable for the
+# 4.x adoption to take effect. Pins the per-package cutoff just past that release.
+# Remove once the rolling window advances past 2026-08-04.
+flit = "2026-08-05T00:00:00Z"
+flit-core = "2026-08-05T00:00:00Z"
 
 [tool.uv.sources]
 # These names must match the names as defined in the pyproject.toml of the workspace items,
--- a/dev/breeze/pyproject.toml
+++ b/dev/breeze/pyproject.toml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 [build-system]
-requires = ["flit_core >=3.12.0,<4"]
+requires = ["flit_core >=4.0.2"]
 build-backend = "flit_core.buildapi"
 
 [project]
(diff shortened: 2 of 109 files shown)


pyproject.toml
```diff
@@ -1561,6 +1561,14 @@ apache-airflow-task-sdk-integration-tests = false
...
 pydantic-ai-skills = "2026-07-16T00:00:00Z"
 
+# Temporary: flit / flit-core 4.0.2 were published 2026-08-04, inside the "4 days"
+# exclude-newer window. `flit build` has no PEP 517 isolation, so the installed flit_core
+# is the backend that builds every provider distribution - it has to be resolvable for the
+# 4.x adoption to take effect. Pins the per-package cutoff just past that release.
+# Remove once the rolling window advances past 2026-08-04.
+flit = "2026-08-05T00:00:00Z"
+flit-core = "2026-08-05T00:00:00Z"
+
```
Lee-W:
```suggestion
```

It's Aug 11th now. I think we can remove it.

Lee-W (approved):

dabla (approved):

PR #71188, opened 2026-08-05
Java SDK: Honor TaskFlow arg bindings sent by the supervisor

jason810496:
- next: #71189

## Why

The Python `@task.stub` call site already defines task data flow. Java tasks should consume those bindings directly instead of repeating upstream IDs with `@Builder.XCom`.

The Java call site is settled by [ADR-0001](https://github.com/apache/airflow/blob/main/java-sdk/adr/0001-mixed-lang-dag-interface.md) (merged in #72019); this PR implements its **binding half**. The wire contract is [ADR-0007](https://github.com/apache/airflow/blob/main/airflow-core/adr/lang-sdk/0007-taskflow-across-language-boundary.md).

## Example

Three authoring syntaxes, exactly the three ADR-0001 specifies. All resolve to the same bindings.

### 1. Annotation based with positional injection

```java
@Builder.TaskHandler(dag = "etl", task = "score")
public long score(Client client, long rows, double threshold, List<String> regions) { ... }
```

### 2. Annotation based with an explicit struct

`report(run_label=..., transformed=...)` binds with nothing declared — names match ignoring case and underscores:

```java
public static class ReportInput implements TaskInput {
  public String runLabel; // binds run_label through the fold
  public long transformed;
}

@Builder.TaskHandler(dag = "etl", task = "report")
public void report(ReportInput input) { ... }
```

### 3. Interface based with an explicit `TaskInput`

```java
public static class SummarizeInput implements TaskInput {
  // Pinned so the field can be called region. Or drop it: public String regionCode;
  @ArgName("region_code")
  public String region;

  public long transformed;
}

public static class Summarize implements InputTask<SummarizeInput> {
  public void execute(@NotNull Context context, Client client, SummarizeInput input) { ... }
}
```

Syntax 1 compiles to this — copied out of the example project's generated sources:

```java
public static final class ConsumeDoubleList implements Task {
  @Override
  public void execute(Context context, Client client) throws Exception {
    TaskArgs args = TaskArgs.of(context, client);
    List<Double> values = args.get(0, new TypeRef<List<Double>>() {});
    new XComCastingExample().consumeDoubleList(values);
  }
}
```

## How

- **`@Builder.TaskHandler(dag, task)` names the pair a handler binds to**, rather than `@Builder.Task(id)` inside a `@Builder.Dag` class. Python declares the task and Java supplies only its body, so the two are different things and should not share a name. `@Builder.Dag`/`@Builder.Task` are left for a Dag authored in Java (#71189).
- **`Bundle` is created and registered into**: `register` takes the class a handler lives on.
- Data parameters bind **by position**; injected `Client`/`Context` take no position.
- Keyword arguments bind **by name** through one `TaskInput`. A task declares flat parameters or one `TaskInput`, never both.
- Names match **ignoring case and underscores** — the same fold [Go](https://github.com/apache/airflow/blob/main/go-sdk/pkg/binding/binding.go) and [TS](https://github.com/apache/airflow/pull/73189) use, so one Python signature binds identically in all three SDKs with nothing declared. `@ArgName` is for what the fold cannot reach (a Python keyword, an unusable identifier) and is matched literally.
- Declared type arguments survive the decode, so a `List<String>` parameter or field does not arrive as `List<LinkedHashMap>`.
- `TaskArgs` is internal — `org.apache.airflow.sdk.internal`, not a `TaskInput`, emitted by the processor only. `InputTask<TaskArgs>` is a compile error; interface tasks declare a `TaskInput`.
- Missing reference or boxed values become `null`; primitive inputs fail with `MissingXComException`, naming the stub argument.
- `@Builder.XCom` is removed, keeping the Python call site as the single source of data-flow wiring.

## Follow-up

ADR-0001 writes the generic read as Jackson's `TypeReference<T>`. This PR ships an SDK-owned `TypeRef<T>` instead, because Jackson is an `implementation` dependency and exposing it would put Jackson on every consumer's compile classpath. Promoting `jackson-core` to `api` and using `TypeReference` directly is worth doing on its own — it is a deliberate next step, not a limitation this PR works around.

ADR-0001's examples are also over-annotated under the fold: `@ArgName("run_label")` on a `runLabel` field is now redundant. Worth a pass over the ADR.

---

- [x] Yes, with help of Claude Code Opus 5 and Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/java-sdk/README.md
+++ b/java-sdk/README.md
@@ -590,7 +590,7 @@ prek hook regenerate it.
 
 <!-- BEGIN AUTO-GENERATED LANG-SDK COMPAT MATRIX -->
 
-*Min. Airflow version: 3.3 · supervisor schema: 2026-06-16*
+*Min. Airflow version: 3.3 · supervisor schema: 2026-10-30*
 
 | Dimension | Tier | Supported | Since | Notes |
 |---|---|---|---|---|
--- a/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
+++ b/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
@@ -20,22 +20,22 @@
 package org.apache.airflow.sdk
 
 /**
- * An immutable snapshot of all [Dag]s that this JVM process can execute.
+ * An immutable snapshot of all [DagDef]s that this JVM process can execute.
  *
  * Build a [Bundle] by implementing [BundleBuilder], then pass it to
  * [Server.serve] to start accepting task-execution requests.
  *
- * @property dags All registered Dags keyed by [Dag.id].
+ * @property dags All registered Dags keyed by [DagDef.id].
  * @throws IllegalArgumentException if any two Dags share the same ID.
  */
 class Bundle(
-  dags: Iterable<Dag>,
+  dags: Iterable<DagDef>,
 ) {
-  internal val dags: Map<String, Dag> = dags.associateByDagId()
+  internal val dags: Map<String, DagDef> = dags.associateByDagId()
 }
 
-private fun Iterable<Dag>.associateByDagId(): Map<String, Dag> {
-  val dagMap = linkedMapOf<String, Dag>()
+private fun Iterable<DagDef>.associateByDagId(): Map<String, DagDef> {
+  val dagMap = linkedMapOf<String, DagDef>()
   for (dag in this) {
     require(dagMap.putIfAbsent(dag.id, dag) == null) {
       "Dags in bundle have duplicate ID: ${dag.id}"
@@ -45,14 +45,14 @@ private fun Iterable<Dag>.associateByDagId(): Map<String, Dag> {
 }
 
 /**
- * Entry point for declaring the [Dag]s that this bundle contains.
+ * Entry point for declaring the [DagDef]s that this bundle contains.
  *
  * Implement this interface to create a Dag bundle to be served by [Server].
  *
  * ```java
  * public class MyBundleBuilder implements BundleBuilder {
  *     @Override
- *     public Iterable<Dag> getDags() {
+ *     public Iterable<DagDef> getDags() {
  *         return List.of(MyDagBuilder.build());
  *     }
  *
@@ -64,14 +64,14 @@ private fun Iterable<Dag>.associateByDagId(): Map<String, Dag> {
  */
 interface BundleBuilder {
   /**
-   * Returns all [Dag]s that belong to this bundle.
+   * Returns all [DagDef]s that belong to this bundle.
    *
    * Called once during [build]; Dag IDs must be unique across the returned
    * collection.
    *
    * @throws IllegalArgumentException if any two Dags share the same ID.
    */
-  fun getDags(): Iterable<Dag>
+  fun getDags(): Iterable<DagDef>
 
   /**
    * Constructs a [Bundle] from the Dags returned by [getDags].
--- a/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/ArgBinding.kt
+++ b/java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/ArgBinding.kt
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.airflow.sdk.execution
+
+/**
+ * One stub-task argument bound at the `@task.stub` TaskFlow call site in the
+ * Python Dag file, delivered via `TIRunContext.arg_bindings`.
+ *
+ * The supervisor schema models this as a `kind`-discriminated union
+ * (`XComArgBinding` / `LiteralArgBinding`), which jsonSchema2Pojo cannot
+ * express as a typed field — the generated `TIRunContext.argBindings` is a
+ * plain `Object` holding the msgpack-decoded list of maps — so this hand-
+ * written decoder materializes the typed view.
+ */
+internal sealed class ArgBinding {
+  abstract val name: String
+
+  internal data class XCom(
+    override val name: String,
+    val taskId: String,
+    val mapIndex: Int,
+    val elementIndex: Int?,
+  ) : ArgBinding()
+
+  internal data class Literal(
+    override val name: String,
+    val value: Any?,
+  ) : ArgBinding()
+}
+
+/**
+ * Decodes the raw `TIRunContext.argBindings` payload into a list of bindings
+ * preserving the stub signature's parameter order — flat task parameters
+ * bind by that position, input-bundle fields by [ArgBinding.name].
+ *
+ * @throws IllegalStateException on a malformed payload, an unsupported
+ *    binding kind, or a duplicate argument name; the task cannot bind its
+ *    arguments correctly, so it must fail rather than run with wrong inputs.
+ */
+internal fun decodeArgBindings(raw: Any?): List<ArgBinding> {
+  if (raw == null) return emptyList()
+  check(raw is List<*>) { "arg_bindings payload is not a list: ${raw.javaClass.name}" }
+  val seen = mutableSetOf<String>()
+  return raw.map { entry ->
+    check(entry is Map<*, *>) { "arg_bindings entry is not a map: $entry" }
+    val name = checkNotNull(entry["name"] as? String) { "arg_bindings entry has no name: $entry" }
+    check(seen.add(name)) { "arg_bindings entries have duplicate name: '$name'" }
+    when (val kind = entry["kind"]) {
+      "literal" -> ArgBinding.Literal(name = name, value = entry["value"])
+      "xcom" ->
+        ArgBinding.XCom(
+          name = name,
+          taskId = checkNotNull(entry["task_id"] as? String) { "xcom arg binding '$name' has no task_id" },
+          mapIndex = (entry["map_index"] as? Number)?.toInt() ?: -1,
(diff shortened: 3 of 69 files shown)


java-sdk/README.md
```diff
@@ -590,7 +590,7 @@ prek hook regenerate it.
 
 <!-- BEGIN AUTO-GENERATED LANG-SDK COMPAT MATRIX -->
 
-*Min. Airflow version: 3.3 · supervisor schema: 2026-06-16*
+*Min. Airflow version: 3.3 · supervisor schema: 2026-10-30*
```
FrankYang0529, line 593: Do we need to update "Min. Airflow version" to 3.4? It looks like 3.3 doesn't have shcema 2026-10-30.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
```diff
@@ -31,17 +31,72 @@ package org.apache.airflow.sdk
...
+   *
+   * @param handlerClass A class with [Builder.TaskHandler] methods.
+   * @return This bundle, for chaining.
+   * @throws IllegalArgumentException if the class has no generated
+   *    registrar, because annotation processing did not run over it.
+   */
+  fun register(handlerClass: Class<*>): Bundle {
+    val registrar =
+      try {
+        Class.forName("${handlerClass.name}Handlers", true, handlerClass.classLoader)
```
FrankYang0529: `register(Class)` and the processor derive the registrar name from two different things. This line uses `handlerClass.name`, so for a nested class it looks for `org.example.Outer$InnerHandlers`. At `BuilderProcessor.kt`, the processor names the registrar from `el.simpleName`, so the generated class is `org.example.InnerHandlers`. We may need to align this.

FrankYang0529 (approved):
Overall LGTM. Thanks for the PR.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
jason810496: Good catch, aligned in bf6a44c4d9: the registrar name now comes from one shared function that both the processor and Bundle.register call, so Outer.Inner resolves Outer_InnerHandlers.

java-sdk/README.md
jason810496: Thanks for the catch. I think it's fine to keep it as-is, since https://github.com/apache/airflow/pull/71757 would make all the matrix up to date afterward.

jason810496 (commented):
Thanks for the review, I just addressed the comments.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/ArgBinding.kt
```diff
@@ -0,0 +1,77 @@
...
+internal fun decodeArgBindings(raw: Any?): List<ArgBinding> {
+  if (raw == null) return emptyList()
+  check(raw is List<*>) { "arg_bindings payload is not a list: ${raw.javaClass.name}" }
+  val seen = mutableSetOf<String>()
+  return raw.map { entry ->
+    check(entry is Map<*, *>) { "arg_bindings entry is not a map: $entry" }
+    val name = checkNotNull(entry["name"] as? String) { "arg_bindings entry has no name: $entry" }
+    check(seen.add(name)) { "arg_bindings entries have duplicate name: '$name'" }
+    when (val kind = entry["kind"]) {
+      "literal" -> ArgBinding.Literal(name = name, value = entry["value"])
```
henry3260: `from_default` is not read here, so a defaulted stub parameter still counts toward the binding list.
Without it, `TaskArgs.of`'s `check(bound == declared)` sees one more argument than the method declares

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/internal/ArgValues.kt
```diff
@@ -0,0 +1,198 @@
...
+   *    nothing.
+   */
+  @JvmStatic
+  fun <I : TaskInput> bindInput(
+    client: Client,
+    type: Class<I>,
+  ): I {
+    val input = newInput(type)
+    val arguments = ArgIndex(client.argBindings)
+    bindableFields(type).forEach { field -> field.set(input, resolveField(client, arguments, field)) }
```
henry3260, line 76: This is the whole binding path: one pass over the fields, each looking for an argument of its own name. There is no fallback for a sole unclaimed argument. It is not consistent in Python and Go.
Python and Go can take a single argument as a whole value
Maybe we should add it?

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/internal/ArgValues.kt
```diff
@@ -0,0 +1,198 @@
...
+  private fun resolveField(
+    client: Client,
+    arguments: ArgIndex,
+    field: Field,
+  ): Any? {
+    val argName = argNameOf(field)
+    val binding = arguments.find(argName, pinned = isPinned(field))
+    if (!field.type.isPrimitive) return binding?.let { decode(client.resolveBinding(it), field.genericType) }
+
+    checkNotNull(binding) {
```
henry3260: Is this divergence intended? Here a primitive field fails instead, while a boxed one returns null one line above. If that is intended, it is worth a line in ADR-0007

henry3260 (commented):
Thanks, just leave few concerns

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/ArgBinding.kt
jason810496: Good catch, fixed in e8b8a5e3f3. `ArgBinding.Literal` now carries `fromDefault`, and `TaskArgs.of` drops those entries before the arity check when the counts disagree — the same two-step as the Go SDK's `resolveFlatParams`, and what ADR-0007 asks of consumers ("leave the latter unclaimed").

A method that *does* declare the defaulted parameter still receives its default, and when the counts still disagree the error reports both.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/internal/ArgValues.kt
jason810496: Deliberately not adding it, because Java does not have the ambiguity that forces it in Go.

In Go any sole struct parameter is name-bound, so the runtime cannot tell a keyword container from one JSON value and `loneWholeValueArgs` has to infer it. Java says which one it is in the type: `TaskInput` means keyword container, anything else is a flat data parameter that already decodes whole. Go's `via_flat_map` spells out in Java as:

```java
public class Payload { public String region; public int count; }  // no TaskInput

@Builder.TaskHandler(dag = "etl", task = "load")
public void load(Payload payload) { }  // the whole dict decodes into Payload
```

A fallback would only ever fire when someone declared a `TaskInput` for something that is not a keyword container, which is exactly the guess the marker interface exists to avoid.

One correction: Python only *produces* `arg_bindings` and has no consumer of its own, so Go is the only SDK doing this today.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/internal/ArgValues.kt
jason810496: Nice catch, I didn't realize this.
Half intended, half a real gap — this made us settle the rule, thanks.

Two situations were collapsed on this line:

- **No argument of that name at all.** The Java signature and the Python stub disagree. This now fails for every field type, not just primitives.
- **An argument that resolves to nothing** (a null literal, or an upstream that pushed no XCom). Intended and unchanged: boxed and reference fields take `null`, primitive fields fail. Declaring a boxed type is how a field says the value is optional.

Recorded in ADR-0001 rather than ADR-0007: ADR-0007 is the language-neutral wire contract, and this is a Java type-system consequence. Go zero-values an unmatched field because it has no null, so its alarm sits on the argument side instead (go-sdk ADR-0006); ADR-0001 now notes that divergence and why.

jason810496 (commented):
Thanks for the review.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
```diff
@@ -31,17 +33,74 @@ package org.apache.airflow.sdk
...
-    require(dagMap.putIfAbsent(dag.id, dag) == null) {
+  /**
+   * Registers a Dag.
+   *
+   * @return This bundle, for chaining.
+   * @throws IllegalArgumentException if another Dag shares its ID.
+   */
+  fun register(dag: DagDef): Bundle {
+    require(dags.putIfAbsent(dag.id, dag) == null) {
       "Dags in bundle have duplicate ID: ${dag.id}"
```
henry3260: Not sure if this is a realistic scenario, but if a user does:

```java
bundle.register("java_etl", "report", Report.class);
bundle.register(dag); // dag.id == "java_etl"

```

the second call fails with ‎`Dags in bundle have duplicate ID: ${dag.id}`, because ‎`register(dagId, taskId, def)` already created a ‎`DagDef("java_etl")` via ‎`getOrPut`.

Is failing the intended behavior here? Or should ‎`register(dag)` merge into an existing DagDef with the same id instead of rejecting it? If it is not a real case, feel free to resolve!

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
```diff
@@ -31,17 +33,74 @@ package org.apache.airflow.sdk
 class Bundle(
   dags: Iterable<DagDef>,
 ) {
-  internal val dags: Map<String, DagDef> = dags.associateByDagId()
-}
+  internal val dags = linkedMapOf<String, DagDef>()
```
henry3260: Since Bundle is now mutable, could we reject `register()` once `serve() `has started? The Go SDK does this on main: serve sets `closed` first thing and Register panics with `"Serve has already been called; register everything before Serve"`.

henry3260:
Thanks for solving my comments :)

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
jason810496: Yes, we should implement this guard like Go side.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/Bundle.kt
jason810496: Yes we should handle this scenario, the TS SDK did validate against this case.

jason810496 (commented):
Nice catch, both concerns are valid. TS and Go side did validate your findings.
I addressed them all in the latest commit, thanks.

Additionally, the latest commit filled the gap of mixed language task TaskFlow binding semantic like https://github.com/apache/airflow/pull/73648.

PR #71189, opened 2026-08-05
Java SDK: Author complete Dags in Java without a Python stub file

jason810496:
- depends on: #69757, #71057, #71188
- Diff for early review: https://github.com/jason810496/airflow/compare/feature/java-sdk-arg-bindings-runtime...feature/java-sdk-native-dag

> **Merge order:**
> 1. #69757 — Support TaskFlow call syntax on stub tasks for the Lang SDK
> 2. #71057 — Register tasks as first-class `TaskDef` objects
> 3. #71188 — Honor TaskFlow arg bindings sent by the supervisor
> 4. **#71189 — Author complete Dags in Java without a Python stub file** **(current one)**
> 5. #71190 — Serialize native Dags to DagSerialization v3

## Why

A Dag with no Python stub file has no `@task.stub` call site to declare its graph, so Java itself must express the graph, the configuration, and the task bodies. Implements [ADR-0002](https://github.com/apache/airflow/blob/main/java-sdk/adr/0002-native-dag-interface.md).

## Annotation surface

The processor generates a **wiring view** — `<Dag>Deps`, an interface with one method per task, injected arguments stripped, data arguments lifted to `Arg<T>` and returns to `TaskRef<T>`. A nested `@Builder.Deps` class implements it:

```java
@Builder.Dag(id = "java_etl", schedule = "@daily")
public class EtlPipeline {

  @Builder.Task(id = "extract", retries = 2)
  public long extract() { return 42L; }

  @Builder.Task(id = "transform")
  public long transform(long extracted, double factor) { return (long) (extracted * factor); }

  @Builder.Task(id = "audit")
  public void audit() { /* side effect only */ }

  @Builder.Deps
  static class Wiring implements EtlPipelineDeps {
    void depends() {
      var rows = extract();
      load(transform(rows, lit(0.9)));
      rows.then(audit()); // ordering-only edge
    }
  }
}
```

Generated (copied from the example project's output):

```java
public interface EtlPipelineDeps extends Deps {
  default TaskRef<Long> extract() {
    return Refs.node(new TaskDef("extract", EtlPipelineBuilder.Extract.class).config("retries", 2));
  }

  default TaskRef<Long> transform(Arg<? extends Number> extracted, Arg<? extends Number> factor) {
    return Refs.call(new TaskDef("transform", EtlPipelineBuilder.Transform.class), extracted, factor);
  }
}
```

## Interface surface

```java
var dag = new DagDef("java_etl").config("schedule", "@daily");
var extract = dag.task("extract", Extract.class).config("retries", 2);
var transform = dag.task("transform", Transform.class);
extract.then(transform).then(dag.task("load", Load.class));
```

`dag.task(...)` registers the task as it creates it and hands back the handle, so there is no second `addTask` to forget — and this is exactly what the annotation surface's recording produces.

## How

- **A view, not a twin object.** The Dag class holds `long extract(Client)` while the view holds `TaskRef<Long> extract()` — two methods that would clash in one class. The wiring class *implements* the view, so the Dag class keeps its single inheritance slot and, being `static`, cannot reach the real task methods.
- **Read by recording, not by parsing.** `depends()` runs once at registration with the Dag in scope. Nothing reads a syntax tree, so this works under any compiler rather than only javac, and identity travels with the handle — a result held in a local and reused refers to one node. Calls are memoized by task ID for the same reason.
- **Ordering edges.** A data edge is implied by passing a handle. Where nothing flows but the sequence, `then` is variadic and returns the new frontier, so `a.then(b, c).then(d)` walks a fan; `Flow.of(a, b).then(c)` opens a chain from a set. `Flow` is nested in `Deps`, so wiring code inherits it by simple name and never collides with `java.util.concurrent.Flow`.
- **Configuration is generated from Airflow's Dag serialization schema**, not hand-listed, so Java attributes cannot drift from the Python semantics they mirror. Only attributes written at the use site are applied.
- **`Bundle.register(...)` is one overloaded verb**: a built Dag, an annotated Dag class, or a task handler against a Python-owned Dag. A Dag is validated as it is registered.
- Runtime arg bindings still win over Java-declared wiring: for a stub task the Python call site is the graph the scheduler ordered the run by. Wired inputs are the fallback a native Dag always uses.

## Note for reviewers

Rebased onto the current #71188, which moved `TaskArgs` out of the public surface, added cross-SDK name folding, and renamed the stub-backed annotation to `@Builder.TaskHandler(dag, task)`. `@Builder.Dag`/`@Builder.Task` therefore mean what they say here: a Dag Java itself owns. This PR also drops its own `airflow-core/adr/lang-sdk/0008-taskflow-dag-dsl.md` — main now has a different ADR-0008, and `java-sdk/adr/0002` supersedes it.

---

- [x] Yes, with help of Claude Code Opus 5 and Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


phanikumv (commented):
this PR seems way too huge, is there a way to create smaller PRs so that it is easier to review?

jason810496:
> this PR seems way too huge, is there a way to create smaller PRs so that it is easier to review?

This PR depends on the other three PRs get merged first ( https://github.com/apache/airflow/pull/69757, https://github.com/apache/airflow/pull/71057, https://github.com/apache/airflow/pull/71188) and contains the changes of all three as well. I just add the "Diff for early review: <link>" now.

java-sdk/sdk/src/main/kotlin/org/apache/airflow/sdk/execution/ArgBinding.kt
```diff
@@ -0,0 +1,77 @@
...
+internal fun decodeArgBindings(raw: Any?): List<ArgBinding> {
+  if (raw == null) return emptyList()
+  check(raw is List<*>) { "arg_bindings payload is not a list: ${raw.javaClass.name}" }
+  val seen = mutableSetOf<String>()
+  return raw.map { entry ->
+    check(entry is Map<*, *>) { "arg_bindings entry is not a map: $entry" }
+    val name = checkNotNull(entry["name"] as? String) { "arg_bindings entry has no name: $entry" }
+    check(seen.add(name)) { "arg_bindings entries have duplicate name: '$name'" }
+    when (val kind = entry["kind"]) {
+      "literal" -> ArgBinding.Literal(name = name, value = entry["value"])
```
henry3260: `from_default` is not read here, so a defaulted stub parameter still counts toward the binding list.
Without it, `TaskArgs.of`'s `check(bound == declared)` sees one more argument than the method declares

PR #71192, opened 2026-08-05
Add single-use callback token for deadline callback context fetch

seanghaeli:
Adds single-use `callback` token so a deadline-callback subprocess can fetch its DagRun context from the Execution API without holding a broad `workload` token (the scope creep that got #66608 reverted).

- New `PATCH /callbacks/{id}/run`: single-use exchange gated on an atomic `Callback.state` QUEUED→RUNNING transition (409 on reuse), returning a short-lived `execution` token via `Refreshed-API-Token`. Mirrors the TI `/run` pattern.
- New `callback` token scope + `callback:self` check (token subject must match the callback id).
- Extracts a shared `issue_execution_token` helper used by both the callback exchange and TI `/run`, and collapses the `*:self` scope checks into one parameterized check.

Reopen of #69840


seanghaeli:
@ashb Whoops, closed the last pull request, re-opening it here. About your comment about the token `scope` being passed a parameter, I've addressed it here:

> Re-wrote it so that only workloads can set their own scope, instead of being passed as a free parameter from anywhere

seanghaeli:
Added your suggestions you mentioned in #https://github.com/apache/airflow/pull/70927, @potiuk.

potiuk (approved):
LGTM. But I think we need @ashb to review it before we merge it.

seanghaeli:
Sounds good, by the way @ashb just a heads up that I addressed your previous round of reviews for this here https://github.com/apache/airflow/pull/69840

PR #71195, opened 2026-08-05
Surface keyring failures instead of silently dropping the token

rjgoyln:
## Summary

`airflowctl auth login` could report success while the token never reached the keyring. `Credentials.save()` writes the API url config before storing the token, and when keyring rejected the token the `TypeError` handler only raised for `ClientKind.CLI` — a kind neither call site passes. Every other case, including a genuine keyring backend failure, returned as though the save had worked, leaving the user with a config file and no credentials.

## Behavior change

A missing token now fails for every client kind, and a `TypeError` unrelated to a missing token propagates instead of being discarded. An unavailable keyring backend still reports the keyring error first, as before.

---

### DIFF
--- a/airflow-ctl/src/airflowctl/api/client.py
+++ b/airflow-ctl/src/airflowctl/api/client.py
@@ -244,9 +244,10 @@ class Credentials:
                 "Use `airflowctl auth login --skip-keyring ...` to dismiss this error."
             ) from e
         except TypeError as e:
-            # This happens when the token is None, which is not allowed by keyring
-            if self.api_token is None and self.client_kind == ClientKind.CLI:
+            # keyring rejects a None password; any other TypeError is a real backend failure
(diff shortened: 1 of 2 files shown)


rjgoyln:
Superseded by #70991, which fixed the same bug from the other side: `Credentials.save()` now raises `AirflowCtlCredentialNotFoundException` before it ever reaches `keyring.set_password`, and the `except TypeError` handler this PR was widening has been removed entirely. #70991 also carries the equivalent coverage — `test_save_propagates_unexpected_keyring_error` and `test_save_without_token`, the latter additionally asserting that `set_password` is not called.

Checked against current `main`: both tests added here pass with this PR's production change reverted, so there is nothing left for them to distinguish. Closing.

---
Drafted-by: Claude Opus 5 (1M context) (no human review before posting)

PR #71196, opened 2026-08-05
Keep API error details when a response declares a JSON charset

rjgoyln:
## Summary

Both API clients treated a response as a parseable error only when `Content-Type` was exactly `application/json`, so any media type carrying parameters — `application/json; charset=utf-8`, as proxies and gateways routinely emit on the errors they generate themselves — skipped error parsing and fell through to `raise_for_status()`. That loses the server's `detail`, and it raises a plain `httpx.HTTPStatusError`, so the `except ServerResponseError` branches in the supervisor and in the CLI commands stop matching.

Airflow's own API server is unaffected — Starlette appends a charset only to `text/*` responses — so this surfaces only where something between client and API server rewrites the header. Parameters are now stripped and the media type case folded; `application/problem+json` remains unaccepted.

---

### DIFF
--- a/airflow-ctl/src/airflowctl/api/operations.py
+++ b/airflow-ctl/src/airflowctl/api/operations.py
@@ -112,7 +112,8 @@ class ServerResponseError(httpx.HTTPStatusError):
         if response.status_code < 400:
             return None
 
-        if response.headers.get("content-type") != "application/json":
+        media_type = response.headers.get("content-type", "").partition(";")[0].strip().lower()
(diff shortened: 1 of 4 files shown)


rjgoyln:
Closing in favour of #70936, which already covers the airflow-ctl side of this. The Task SDK client has the same exact-string `content-type` comparison, so it is folded into that PR instead of being fixed separately here.

---
Drafted-by: Claude Code (Opus 5) (no human review before posting)

PR #71199, opened 2026-08-05
Fix __getattr__ in helpers.py throwing AttributeError for dotted module paths

### LINKED ISSUE #71198: Bug: __getattr__ in airflow.utils.helpers raises AttributeError for deprecated dotted imports

Airflow Core

### Apache Airflow version

main

**Issue Description**
The `__getattr__` fallback in `airflow.utils.helpers` is designed to provide backward compatibility for deprecated functions that have moved (like `render_template_as_native`). However, it currently fails and throws an `AttributeError` when any of those functions are accessed.

This happens because the fallback dynamically loads the new module using `__import__(modpath)`. When `modpath` is a dotted path (e.g., `"airflow.sdk.definitions.context"`), Python's `__import__` without the `fromlist` argument returns the root package (`airflow`), not the leaf module. The subsequent `getattr` call then tries to find the function on the root `airflow` module, which fails.

**Steps to reproduce:**
Run the following in a Python shell:
```python
import airflow.utils.helpers
# Attempt to access a deprecated function that has moved to a dotted module path
airflow.utils.helpers.render_template_as_native
```
This incorrectly produces the following error:

AttributeError: module 'airflow' has no attribute 'render_template_as_native'

Instead of raising an `AttributeError`, the `__getattr__` fallback should properly resolve the leaf module, trigger the `DeprecationWarning`, and return the requested function. 
This can be easily fixed by using `importlib.import_module(modpath)` instead of `__import__(modpath)`. 
*(Note: I already have a PR ready to submit for this fix!)*

### Operating System

Not Applicable

### Deployment

None

_No response_

Not Applicable

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

Not Applicable

Not Applicable

### Anything else?

_No response_

### Code of Conduct

KushagraB424:
### Description

This PR fixes a bug in `airflow.utils.helpers.__getattr__` where the dynamic loading of deprecated functions (like `render_template_as_native`) was failing for dotted module paths (e.g. `airflow.sdk.definitions.context`). 

When passing a dotted path to `__import__` without the `fromlist` argument, it incorrectly returned the root package `airflow` instead of the target leaf module. This caused an `AttributeError` when the code subsequently tried to `getattr` the deprecated function directly on the root `airflow` module. 

The fix replaces `__import__(modpath)` with `importlib.import_module(modpath)`, which correctly returns the leaf module, allowing the compatibility shim and `DeprecationWarning` to function as intended.

* closes: #71198

---

Generated-by: Antigravity IDE following the guidelines at https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions

### DIFF
--- a/airflow-core/src/airflow/utils/helpers.py
+++ b/airflow-core/src/airflow/utils/helpers.py
@@ -313,6 +313,7 @@ def __getattr__(name: str):
     except KeyError:
         raise AttributeError(f"module '{__name__}' has no attribute '{name}'") from None
 
+    import importlib
     import warnings
 
     warnings.warn(
@@ -320,4 +321,4 @@ def __getattr__(name: str):
         DeprecationWarning,
         stacklevel=2,
     )
-    return getattr(__import__(modpath), name)
+    return getattr(importlib.import_module(modpath), name)


airflow-core/src/airflow/utils/helpers.py
```diff
@@ -313,11 +313,12 @@ def __getattr__(name: str):
...
+    import importlib
     import warnings
 
     warnings.warn(
         f"{__name__}.{name} is deprecated. Use {modpath}.{name} instead.",
         DeprecationWarning,
         stacklevel=2,
     )
-    return getattr(__import__(modpath), name)
+    return getattr(importlib.import_module(modpath), name)
```
kaxil, line 324: Can you add a regression test? `airflow-core/tests/unit/utils/test_helpers.py` already imports `helpers`, so it's a few lines:

```python
@pytest.mark.parametrize(
    "name", ["render_template_as_native", "render_template_to_string", "prevent_duplicates"]
)
def test_deprecated_imports_resolve(name):
    with pytest.warns(DeprecationWarning):
        assert getattr(helpers, name) is not None
```

Nothing in the tree touches `__deprecated_imports` today, which is why all three entries were broken rather than just one.

airflow-core/src/airflow/utils/helpers.py
```diff
@@ -313,11 +313,12 @@ def __getattr__(name: str):
     except KeyError:
         raise AttributeError(f"module '{__name__}' has no attribute '{name}'") from None
 
+    import importlib
     import warnings
 
     warnings.warn(
         f"{__name__}.{name} is deprecated. Use {modpath}.{name} instead.",
         DeprecationWarning,
```
kaxil, line 321: `stacklevel=2` attributes the warning to the caller's module, so a DAG author doing `from airflow.utils.helpers import render_template_as_native` gets a `DeprecationWarning` reported against their DAG file, which Python's default `ignore` filter drops. The `action="default"` filter Airflow installs in `configuration.py:58` is scoped to `module="airflow"` and doesn't match either. I confirmed locally that after your fix the shim resolves correctly but the notice never reaches stderr.

We already have `airflow.utils.deprecation_tools.add_deprecated_classes` for this, and it gets both parts right: `importlib.import_module` plus `DeprecatedImportWarning`, which subclasses `FutureWarning` and so is shown by default. `airflow/io/__init__.py` uses the same `__name__: {...}` form. Worth replacing the hand-rolled block with:

```python
from airflow.utils.deprecation_tools import add_deprecated_classes

add_deprecated_classes(
    {
        __name__: {
            "render_template_as_native": "airflow.sdk.definitions.context.render_template_as_native",
            "render_template_to_string": "airflow.sdk.definitions.context.render_template_to_string",
            "prevent_duplicates": "airflow.sdk.definitions.mappedoperator.prevent_duplicates",
        }
    },
    package=__name__,
)
```

which lets both `__getattr__` and `__deprecated_imports` go away.

airflow-core/src/airflow/utils/helpers.py
```diff
@@ -313,11 +313,12 @@ def __getattr__(name: str):
     except KeyError:
         raise AttributeError(f"module '{__name__}' has no attribute '{name}'") from None
 
+    import importlib
```
kaxil, line 316: Minor: `importlib` is part of the import machinery and always loaded, so this can sit at the top of the file with the rest of the imports. Moot if you take the `add_deprecated_classes` route above.

shahar1:
Fixed by #71691

Issue #37217, opened 2024-02-07
CloudBatchSubmitJobOperator runnable.container.commands does not interpret jinja templates


ya7010:
### Description

[CloudBatchSubmitJobOperator](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/operators/cloud/cloud_batch.html)'s `runnable.container.commands` is not expanded when the jinja template is used as a command argument.

### Use case/motivation

I would like to pass `logical_date` to a script args to run in CloudBatch.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

jscheffl:
Reason for not templating is probably that the Google Operator in https://github.com/apache/airflow/blob/main/airflow/providers/google/cloud/operators/cloud_batch.py#L60 does not describe the "command" field as template field (see https://airflow.apache.org/docs/apache-airflow/stable/howto/custom-operator.html#templating)

From the descriptio of this field it would make sense to have this template-able. If you have a strong demand, please contribute a PR to the respective code section, else maybe you need to wait that somebody picks this up.

yassun7010:
I am looking at this [PR](https://github.com/apache/airflow/pull/20905).

I had added `job` to `template_fields` but it did not work.

Cloud Batch job is a deep configuration class and I need to template the contents of `self.job.task_groups[0].task_spec.runnables[0].container.commands`.

I have assigned execution arguments to `self.commands` and passed them to the `self.job` creation process (hoping they would be shared as a list), but this method is not working either.

jscheffl:
Yes, (I am not an expert in this operator but...) if you take look to the docs, the Operator accepts either a `dict` or a class as Job definition. If you pass a dict then I assume the dict will implicitly be parsed to the object structure describing the workload/Job.
Means in the instance of the operator change the `template_fields` variable and add the `job` field, pass the needed dict as string with the template fields. The template engine will parse the dict-string as template, replace the field. You need to define `render_template_as_native_obj=True` in the DAG as described in https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html#referencing-params-in-a-task

yassun7010:
Thanks for letting me know!

Next week I will try to work on it.

romsharon98:
> Thanks for letting me know!
> 
> 
> 
[...]

PR #71201, opened 2026-08-05
Move CloudBatchSubmitJobOperator job normalization out of __init__

yunusakca:
Moves `CloudBatchSubmitJobOperator`'s `Job` protobuf → dict normalization out of
`__init__`, as part of the exemption-list burn-down tracked in #70296.

Template fields are rendered after the constructor runs, so reading a template
field's value inside `__init__` operates on the un-rendered Jinja expression
rather than the real value. The `job` protobuf-to-dict conversion — added for
#37217 so the renderer can descend into nested fields such as a runnable's
container commands — is exactly that pattern.

The conversion has to run *before* Jinja rendering, not after: moving it to
`execute()` would silently stop nested Jinja expressions inside a `Job`
protobuf from ever being rendered. It now lives in `prepare_template()`, the
hook Airflow runs after construction but before rendering, matching the pattern
`CloudBuildCreateBuildOperator` already uses in this provider.

`CloudBatchSubmitJobOperator`'s entry is removed from
`validate_operators_init_exemptions.txt`.

The existing rendering test already covers both a `dict` and a protobuf `Job`
input; a new test asserts the normalization happens in `prepare_template()`
rather than the constructor.

related: #70296

---

- [X] Yes — Claude Code (Sonnet 5 and Opus 5)

Generated-by: Claude Code (Sonnet 5 and Opus 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/cloud_batch.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/cloud_batch.py
@@ -78,16 +78,19 @@ class CloudBatchSubmitJobOperator(GoogleCloudBaseOperator):
         self.region = region
         self.job_name = job_name
         self.job = job
-        # Normalize Job protobuf to dict so Airflow's template renderer can descend
-        # into nested fields (e.g. runnable.container.commands). See #37217.
-        if isinstance(job, Job):
-            self.job = Job.to_dict(job)
         self.polling_period_seconds = polling_period_seconds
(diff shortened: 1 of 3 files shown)


potiuk:
Better fix in #70835 - already merged. Closing as duplicate.

PR #71204, opened 2026-08-05
Fix #71125: DAG processor treats any zip-format file as a potential DAG bundle

### LINKED ISSUE #71125: DAG processor treats any zip-format file (.jar, .pptx, .docx, .xlsx, etc.) as a potential DAG bundle, not just .zip
### Description

`find_dag_file_paths` (airflow-core/src/airflow/utils/file.py) decides whether to attempt DAG discovery on a file using:

```python
if path.is_file() and (path.suffix == ".py" or zipfile.is_zipfile(path)):
    if might_contain_dag(file_path, safe_mode):
        file_paths.append(file_path)
```

`zipfile.is_zipfile()` is a content sniff (checks for the PK zip magic bytes/central directory), not an extension check. There is no accompanying `path.suffix == ".zip"` condition. Since the ZIP format underlies many common file types beyond `.zip` itself — `.jar`, `.pptx`, `.docx`, `.xlsx`, `.apk`, `.epub`, `.odt`, `.whl`, etc. — any of these dropped into a DAGs folder (e.g. a build artifact, a supporting doc, a packaged dependency) will pass this check and get opened and scanned via `might_contain_dag`, purely because it happens to share the underlying zip container format with Airflow's own zipped-DAG-bundle feature.

This was reported previously in #45718 with a `.pptx` file, but that issue was closed as invalid because the specific symptom described there (garbled metric names) turned out to be an unrelated stat-sanitization bug, not this zip-detection design question. The zip-detection behavior itself was never actually addressed.

### Impact

At minimum this is wasted work (opening and scanning irrelevant files on every DAG processor cycle). Depending on `might_contain_dag`'s heuristic and the archive's contents, it can also produce confusing log noise, and in the originally reported case, appears able to feed corrupted/unexpected data further into DAG processing.

Gate the zip-bundle branch on the file extension in addition to (or instead of) the content sniff, e.g.:

```python
if path.is_file() and (path.suffix == ".py" or (path.suffix == ".zip" and zipfile.is_zipfile(path))):
```

This preserves the documented `.zip` DAG-bundle behavior while no longer opening arbitrary non-`.zip` files that merely share the same underlying container format. `.airflowignore` is a viable per-deployment workaround today (matching by extension), but it means every deployment that happens to keep e.g. `.jar` files anywhere under its DAGs folder has to know to add this rule proactively rather than it being a non-issue by default.

### How to reproduce

1. Place any non-`.zip` PK-zip-format file (a `.jar`, `.pptx`, `.docx`, etc.) anywhere under the DAGs folder.
2. Wait for the DAG processor to walk the directory.
3. Observe that the file is opened and passed through `might_contain_dag` (visible via DAG processor debug logs), the same as a `.zip` DAG bundle would be, despite not being one.

- [X] Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

---
Drafted-by: Claude Code (Sonnet 5); reviewed by @seanmuth before posting

saitejabandaru-in:
### Description

Fixes #71125. `zipfile.is_zipfile()` is a content sniff, which means it evaluates to true for files like `.jar`, `.pptx`, `.docx`, `.xlsx` etc. since they are technically zip archives. This resulted in the DAG processor incorrectly treating them as DAG bundles, leading to wasted processing, errors, and noise.

This PR updates the DAG discovery logic to ensure that a file explicitly has a `.zip` extension in addition to passing `zipfile.is_zipfile()` before we process it as a zipped DAG bundle.

- [x] Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### DIFF
--- a/airflow-core/src/airflow/dag_processing/importers/python_importer.py
+++ b/airflow-core/src/airflow/dag_processing/importers/python_importer.py
@@ -111,7 +111,7 @@ class PythonDagImporter(AbstractDagImporter):
         for file_path in find_path_from_directory(directory, ".airflowignore", ignore_file_syntax):
             path = Path(file_path)
             try:
-                if path.is_file() and (path.suffix.lower() == ".py" or zipfile.is_zipfile(path)):
+                if path.is_file() and (path.suffix.lower() == ".py" or (path.suffix.lower() == ".zip" and zipfile.is_zipfile(path))):
                     if might_contain_dag(file_path, safe_mode):
                         yield file_path
             except Exception:
@@ -158,7 +158,7 @@ class PythonDagImporter(AbstractDagImporter):
 
         try:
             with warnings.catch_warnings(record=True) as captured_warnings:
-                if filepath.endswith(".py") or not zipfile.is_zipfile(filepath):
+                if filepath.endswith(".py") or not (filepath.lower().endswith(".zip") and zipfile.is_zipfile(filepath)):
                     modules = self._load_modules_from_file(filepath, safe_mode, result)
                 else:
                     modules = self._load_modules_from_zip(filepath, safe_mode, result)
--- a/airflow-core/src/airflow/dag_processing/manager.py
+++ b/airflow-core/src/airflow/dag_processing/manager.py
@@ -945,7 +945,7 @@ class DagFileProcessorManager(LoggingMixin):
         observed_filelocs: set[str] = set()
         for info in present:
             abs_path = str(info.absolute_path)
-            if abs_path.endswith(".py") or not zipfile.is_zipfile(abs_path):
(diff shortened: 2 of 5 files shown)


molcay:
Hi @saitejabandaru-in,

Thanks for working on this! The core fix to check for the `.zip` extension makes perfect sense and will definitely cut down on the DAG processor noise.

However, I noticed there are some file changes (on the providers folder path) included in this PR that seem unrelated to the #71125 fix. Could you please revert those extra files so we can keep this PR strictly scoped to the DAG discovery logic? It will make this much easier to get reviewed and merged.

shahar1:
Pleasse refer to the guidelines regarding AI-assisted contributions.

Closing as this PR doesn't qualify the guidelines.

PR #71206, opened 2026-08-05
Add airflowctl tasks state command

### LINKED ISSUE #66174: [AIP-94] airflowctl tasks: add state command
## Summary

Add `airflowctl tasks state` 

## Implementation

- anything work with single endpoint under operation.py
- anything needs multiple endpoint or transformation goesn ctl/commands
- including unit and integration test under airflow-ctl-tests/

### Committer

haseebmalik18:
This adds `airflowctl tasks state`, which prints the state of a single task instance. It is the airflowctl equivalent of the deprecated `airflow tasks state`.

You select the run with either a run id or a logical date, and mapped tasks are supported through `--map-index`:
  
      airflowctl tasks state <dag_id> <task_id> <run_id>
      airflowctl tasks state <dag_id> <task_id> --logical-date <logical_date>

The output is just the state value (for example `success`) to stay script friendly. It reuses the existing task instance API operation, and error handling matches the sibling `failed-deps` and `states-for-dag-run` commands.

closes: #66174


haseebmalik18:
@bugraoz93 @henry3260

haseebmalik18:
@bugraoz93 @henry3260 Following up

airflow-ctl/src/airflowctl/ctl/commands/task_command.py
```diff
@@ -149,3 +156,31 @@ def states_for_dag_run(args, api_client=NEW_API_CLIENT) -> None:
...
+
+
+@provide_api_client(kind=ClientKind.CLI)
+def state(args, api_client=NEW_API_CLIENT) -> None:
+    """Get the state of a task instance."""
+    if (args.run_id is None) == (args.logical_date is None):
+        rich.print("[red]Provide either run_id or --logical-date, but not both[/red]")
+        sys.exit(1)
+
+    run_id = args.run_id or _find_run_id_by_logical_date(api_client, args.dag_id, args.logical_date)
```
potiuk: Heads-up rather than a change request: this selector block is now the third copy in this file — identical to lines 89-94 in `failed_deps` and 138-143 in `states_for_dag_run`.

#70904 is open and does exactly the job of collapsing all of them into a shared `resolve_dag_run_id(api_client, args)`, so the two PRs will conflict and whichever lands second needs a small rework. My instinct is that this one should go first and #70904 then absorbs all three call sites in a single pass — no action needed from you, I'll make sure both sides know.

One gentle observation while I'm here: you extracted `_task_instance_not_found_message` to remove the duplicated 404 message, which was the right call — the same instinct applied to this block would have avoided the collision entirely.

airflow-ctl/src/airflowctl/ctl/commands/task_command.py
```diff
@@ -149,3 +156,31 @@ def states_for_dag_run(args, api_client=NEW_API_CLIENT) -> None:
...
+        )
+    except ServerResponseError as e:
+        if e.response.status_code == 404:
+            rich.print(
+                f"[red]{_task_instance_not_found_message(args.dag_id, run_id, args.task_id, args.map_index)}[/red]"
+            )
+            sys.exit(1)
+        raise
+
+    print(task_instance.state.value if task_instance.state else None)
```
potiuk, line 146: When a task instance has no state this prints the literal string `None`, and `test_state_prints_none_when_task_instance_has_no_state` pins it, so I read this as deliberate rather than an oversight.

Still worth a second thought: `None` is a Python repr surfacing in output your description designs to be "script friendly". Every other value this command emits is a lowercase state token (`success`, `running`, …), so a shell caller comparing strings gets one odd capitalised Python-shaped value in the set. An empty line, or a lowercase `none`, would sit better with the rest.

Either choice is defensible — if you keep `None`, a line in the command description saying what an unset state prints would save the next person working it out from the tests.

potiuk (approved):
Approving — this is cleanly done. The error handling matches `failed_deps` and `states-for-dag-run` exactly, plain `print` for the value with `rich.print` reserved for errors is the right split for machine-readable output, and no newsfragment is correct for `airflow-ctl`.

Two things I checked specifically rather than assumed:

- **The generated docs artifacts were genuinely regenerated.** `command_hashes.txt` moves the `tasks` hash `ea587dc8…` → `eb70701d…`, and `output_tasks.svg` has its terminal class ids rewritten wholesale (`terminal-2578352511-*` → `terminal-761188829-*`) — the generator's signature, not a hand edit.
- **Test coverage is the most thorough I've seen in this queue today**: run_id, logical-date, map_index, null state, mutually-exclusive selectors, malformed logical date, run-not-found, task-instance-not-found and non-404 re-raise, plus an integration case in `airflow-ctl-tests` for both invocation forms.

Two nits inline, neither blocking. One of them is really a heads-up about another PR rather than a criticism of this one.

At merge time, note that 10 of the 11 commits here are `Merge branch 'main' into…`. Since `apache/airflow` squashes using the commit messages, the squash body will carry all ten unless it is tidied up — worth a glance when this goes in.

---

>

---

haseebmalik18:
Addressed review comments

Eason09053360:
Thanks fot the patch! LGTM+1
I found something could be a follow-up: 
for a mapped task queried without `--map-index`, the API's 404 detail ("Task instance is mapped, add the map_index value to the URL") is suppressed and this prints only "…not found", so surfacing that detail (or a `--map-index` hint) in the shared 404 message might help here and in `failed-deps`.

PR #71207, opened 2026-08-05
Eliminate trust policy race by migrating emr_eks test to Pod Identity

ramitkataria:
The test shares one EMR job execution role across runs while each run creates its own EKS cluster. With IRSA, every run had to rewrite the shared role's trust policy to federate its own cluster's OIDC provider, and pruning old statements to stay under the trust policy size limit meant concurrent runs deleted each other's trust entries, failing jobs with AssumeRoleWithWebIdentity AccessDenied. IAM offers no compare-and-swap on trust policies, so this race cannot be fixed while runs mutate shared IAM state.

EKS Pod Identity (supported by EMR on EKS since emr-7.3.0) removes the shared mutable state: the role keeps one static trust policy and each run creates Pod Identity associations scoped to its own cluster, also avoiding the 4096-char trust policy cap that limits IRSA to roughly a dozen clusters per role.

Pod Identity is also [AWS's recommended mechanism](https://docs.aws.amazon.com/eks/latest/best-practices/identity-and-access-management.html#_eks_pod_identities_compared_to_irsa) for pod credentials:

> Unless you have specific usecases for IRSA, we recommend you use
> EKS Pod Identities when using EKS.

 

---

---

### DIFF
--- a/providers/amazon/tests/system/amazon/aws/example_emr_eks.py
+++ b/providers/amazon/tests/system/amazon/aws/example_emr_eks.py
@@ -16,14 +16,11 @@
 # under the License.
 from __future__ import annotations
 
-import json
 import logging
 import subprocess
-import time
 from datetime import datetime, timedelta
 
 import boto3
-from tenacity import retry, retry_if_exception_type, stop_after_delay, wait_exponential
 
 from airflow.providers.amazon.aws.hooks.eks import ClusterStates, NodegroupStates
 from airflow.providers.amazon.aws.operators.eks import EksCreateClusterOperator, EksDeleteClusterOperator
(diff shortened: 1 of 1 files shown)


providers/amazon/tests/system/amazon/aws/example_emr_eks.py
```diff
@@ -385,15 +334,15 @@ def delete_virtual_cluster(virtual_cluster_id):
         create_cluster_and_nodegroup,
         await_create_nodegroup,
         run_eksctl_commands(eks_cluster_name, eks_namespace),
-        update_trust_policy_execution_role(eks_cluster_name, eks_namespace, job_role_name),
-        wait_for_trust_policy_propagation(eks_cluster_name, job_role_name),
+        install_pod_identity_agent(eks_cluster_name),
+        create_pod_identity_role_associations(eks_cluster_name, eks_namespace, job_role_name),
```
ferruzzi, line 382: Have you checked if there is propagation delay on the `create identity` here?  Just making sure we aren't trading one race for another without a waiter or sensor here to check it.

providers/amazon/tests/system/amazon/aws/example_emr_eks.py
```diff
@@ -332,7 +281,7 @@ def delete_virtual_cluster(virtual_cluster_id):
         task_id="start_job",
         virtual_cluster_id=str(create_emr_eks_cluster.output),
         execution_role_arn=job_role_arn,
-        release_label="emr-7.0.0-latest",
+        release_label="emr-7.13.0-latest",
```
ferruzzi, line 328: Just curious if there is any particular reason for this version as opposed to 7.3.0?  Not blocking, just curious if this is just the current latest as opposed to using 7.3 which is the minver you stated above.

providers/amazon/tests/system/amazon/aws/example_emr_eks.py
```diff
@@ -127,43 +123,37 @@ def run_eksctl_commands(cluster_name, ns):
...
+    log.info("Pod Identity Agent addon is ACTIVE on cluster %s", cluster_name)
+
 
-    # See https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up-trust-policy.html
-    # The action "update-role-trust-policy" is not available in boto3, thus we need to do it using AWS CLI
+@task
+def create_pod_identity_role_associations(cluster_name, namespace, role_name):
+    # `emr-containers create-role-associations` is a CLI-only helper — it creates
+    # the three Pod Identity associations (client, driver, executor) that the EMR
+    # job pods use. There is no boto3 equivalent. Requires AWS CLI > 2.24.0.
```
ferruzzi: If we have a minver here, we should likely add a check to enforce it like we did in   example_sagemaker

providers/amazon/tests/system/amazon/aws/example_emr_eks.py
ramitkataria: That's just the latest version. I figured if we're changing versions anyway, might as well go to the latest supported version

ferruzzi: I don't have a strong opinion, I was just curious where that number came from.  :+1:

providers/amazon/tests/system/amazon/aws/example_emr_eks.py
ramitkataria: There's nothing documented about it and I didn't run into it in my testing. There's also a lot of natural slack: the associations are created before create_emr_eks_cluster, then StartJobRun, then EMR has to schedule the pod and pull the image. That adds up to a few minutes, which should be plenty. And job_starter already has retries=2 with a 2-minute delay as a backstop.

I tried a readback poll on list_pod_identity_associations and dropped it because if the CLI exits 0 the associations are already there, so it was just re-reading a deterministic result, and it wouldn't prove the agent can actually vend credentials anyway. Only the job running shows that.

providers/amazon/tests/system/amazon/aws/example_emr_eks.py
ramitkataria: Added a check to install a valid version if not present. Thanks for the catch!

ferruzzi (approved):

PR #71211, opened 2026-08-06
Replace `resume_glue_job_on_retry` with durable execution for `GlueJobOperator`

amoghrajesh:
---

- [x] Yes - claude sonnet

## Summary

This PR extends `GlueJobOperator` for durable execution (AIP-103): on Airflow 3.3+, the job
run id is persisted to task state store immediately after submission, so a worker crash and retry
reconnects to the run already executing in Glue instead of starting a new one.

## How it works today

`GlueJobOperator`'s `resume_glue_job_on_retry` mechanism has two tiers: an XCom cached run id
(fast path), falling back to a paginated scan of the job's run history tagged by task UUID. I
reproduced twice against real AWS Glue that the fast path never fires on Airflow 3.x: the server
clears a task instance's XComs before every non-deferral attempt, so the cached id is always gone
by the time the next retry looks for it. Every retry was paying for the full scan regardless, and
the scan itself returned the first UUID match regardless of state, so an earlier terminated run
could mask an active one underneath it.

## Why

`resume_glue_job_on_retry` already tries to do this via XCom, falling back to a scan of the job's
run history when XCom is empty. Confirmed against real AWS Glue (killed the worker mid-poll,
twice, under two different kill methods) that the XCom fast path never fires on Airflow 3.x: the
server clears a task instance's XComs before every non-deferral attempt
(`task_instances.py`'s `xcom_keys_to_clear`), so by the time the next retry looks for the cached id
it's already gone. Every retry pays for the full paginated scan regardless of timing, not just the
narrow crash-before-push window the mechanism's original PR (#62560) was written to cover.

The scan itself also has a latent bug: it returns the first task-UUID match regardless of state,
so an earlier terminated run can mask a still-active one further down the page. Not fixed in this
PR -- see Scope below.

This matters more for Glue than for most `ResumableJobMixin` ports: a Glue job's
`concurrent_run_limit` defaults to `1`, so a duplicate submission doesn't just waste compute, it
fails outright with `ConcurrentRunsExceededException` and the task keeps retrying against a run it
can never observe.

## What changed

- `GlueJobOperator` now inherits `ResumableJobMixin`. On Airflow 3.3+, the mixin checks task state
  store before doing anything else: an active run reconnects directly (one `get_job_state` call), a
  succeeded or externally stopped run returns without resubmitting, a terminal run resubmits fresh.
- The **existing** XCom-then-scan mechanism is kept exactly as it was, relocated into `submit_job`
  (the mixin only calls `submit_job` when task state has nothing usable). This is what Airflow
  <3.3 runs on entirely, and what 3.3+ falls back to as a one-time bootstrap for a run started
  before this upgrade, or before task state store had anything persisted.
- `durable` is fully inert below Airflow 3.3: setting it explicitly (`True` or `False`) only emits
  a `UserWarning` that it has no effect there, since the task state store it relies on doesn't
  exist yet. This is deliberate -- it avoids a `durable` parameter that promises real crash
  recovery on a version where it structurally can't deliver it (the same concern raised against the
  sibling KPO port at #71492, where `durable` is currently wired straight into the old mechanism
  regardless of Airflow version).
- `resume_glue_job_on_retry` is deprecated in favour of `durable` (`AirflowProviderDeprecationWarning`).
  The warning fires on **every** supported Airflow version -- removal is tied to this provider's
  own release schedule, not to which Airflow version the caller happens to run, so staying silent
  below 3.3 would mean some users never see the warning before the parameter disappears. On
  Airflow 3.3+ its value maps onto `durable`, which is a real, working replacement there. Below
  3.3, `durable` itself does nothing (per the point above), so the warning message says so
  explicitly rather than telling that user to switch to something currently inert for them;
  `resume_glue_job_on_retry` remains the only way to get this behavior until the provider's
  minimum supported Airflow version reaches 3.3, at which point the parameter can be removed
  outright.
- A run `STOPPED` outside Airflow (e.g. cancelled manually) is treated as a success rather than
  resubmitted -- the work is genuinely finished, just not the way the task expected -- with a
  warning logged. Matches how `GlueJobHook.job_completion` and `GlueJobCompleteTrigger` already
  treat `STOPPED`.
- Deferrable takes precedence over durable, unchanged in spirit from today: the Triggerer already
  tracks the run across the wait, so the run id isn't persisted to task state on that path.

## Behaviour change

`durable` is owned by the mixin and defaults to `True`, but that default only has an effect on
Airflow 3.3+. There, a Dag that never touched `resume_glue_job_on_retry` goes from "always
resubmit on retry" to "reattach on retry" with no code change. Below 3.3, nothing changes for such
a Dag: `durable`'s default is meaningless there, and behavior matches the pre-existing
`resume_glue_job_on_retry=False` default exactly.

Documented in `providers/amazon/docs/changelog.rst` and `providers/amazon/docs/operators/glue.rst`.

## Scope

Deliberately left alone, all pre-existing, none introduced by this PR:

- The scan's first-match-regardless-of-state bug described above.
- `GlueJobHook._handle_state` treats only `FAILED`/`TIMEOUT` as failure, so AWS's terminal
  `ERROR`/`EXPIRED` states fall through to "keep polling" and the sync poll loop never exits on
  them.
- `ResumableJobMixin.execute_resumable`'s reconnect path returns `poll_until_complete`'s `None`
  instead of calling `get_job_result`, unlike the other two paths. Not touched here (no task-sdk
  change) -- every method instead sets `self._job_run_id`, and `execute()` returns that directly.

## Why `task_state_store` is a better crash recovery mechanism

**1. It survives retries while XCom does not.**
Airflow clears all XComs before each retry attempt. `task_state_store` is built specifically to
persist across retries.

**2. One call vs. a full scan.**
Store hit -> single `get_job_state` call. Scan -> paginates the entire job run history, 50 at a
time.

**3. No ambiguity.**
Store returns the exact id you saved. The scan matches by tag and can pick a stale run if an older
one shares the tag.

## Testing

Uploaded this script on AWS Glue and created an _ETL JOB_ there:
```python
import sys
import time


def get_arg(name, default=None):
    flag = f"--{name}"
    if flag in sys.argv:
        i = sys.argv.index(flag)
        if i + 1 < len(sys.argv):
            return sys.argv[i + 1]
    return default


task_uuid = get_arg("airflow_task_uuid", "<NOT INJECTED>")
sleep_total = int(get_arg("sleep_seconds", "600"))

print(f"argv: {sys.argv}", flush=True)
print(f"airflow_task_uuid = {task_uuid}", flush=True)
print(f"sleeping for {sleep_total}s", flush=True)

elapsed = 0
while elapsed < sleep_total:
    time.sleep(15)
    elapsed += 15
    print(f"alive {elapsed}/{sleep_total}s uuid={task_uuid}", flush=True)

print("done", flush=True)

```

image

Using this DAG:
```python

from airflow.providers.amazon.aws.operators.glue import GlueJobOperator
from airflow.sdk import DAG

with DAG(
    dag_id="glue_job_trial",
    schedule=None,
    catchup=False,
) as dag:
    run_glue = GlueJobOperator(
        task_id="run_glue",
        job_name="airflow_durable_test",
        aws_conn_id="aws_default",
        region_name="us-east-1",
        durable=True,
        script_args={"--sleep_seconds": "900"},
        wait_for_completion=True,
        deferrable=False,
        verbose=True,
        retries=3,
        retry_delay=timedelta(seconds=20),
    )

```

### Case 1: Crash Recovery after worker crash

Launched the DAG:

image

Glue Job:

image

State Store:

image

Brought the worker down, see that the airflow task instance goes to retry state

image

Reconnects to same job

image

### Case 2: Stop a run which was submitted from Glue Console, there should be no resubmit

image

image

image

### Case 3: `wait_for_completion=False` persists to state store

Update task to:
```python
    run_glue = GlueJobOperator(
        task_id="run_glue",
        job_name="airflow_durable_test",
        aws_conn_id="aws_default",
        region_name="us-east-1",
        durable=True,
        script_args={"--sleep_seconds": "900"},
        wait_for_completion=False,
        deferrable=False,
        verbose=True,
        retries=3,
        retry_delay=timedelta(seconds=20),
    )
```

image

image

---


amoghrajesh:
cc @Alexhans FYI.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,70 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+submitting a second run while the first is still active does not create a harmless duplicate, it
+fails outright with ``ConcurrentRunsExceededException`` and the task keeps retrying against a run
+it can never see. Durable execution turns that retry into a normal reconnect.
+
+On retry the operator checks the prior run's state:
+
+* if it is still starting, running, waiting for capacity, or being stopped, the operator
+  reconnects and continues polling
+* if it already succeeded, or was stopped outside Airflow, the operator returns immediately
+  without resubmitting
```
kaxil: This path also fires when someone clears the task to force a re-run, not just on a retry. The store rows are scoped to `(dag_run_id, task_id, map_index)` and clearing a task doesn't delete them (the only cascade is on `dag_run` deletion), and `[state_store] clear_on_success` defaults to `False`, so the run id is still sitting there after a successful attempt. Clear the task in the UI and the next attempt reads `SUCCEEDED` and goes green in a second or two without submitting anything to Glue.

That's a bigger change than the retry behaviour this section describes, given `durable` now defaults to `True` where `resume_glue_job_on_retry` defaulted to `False`. Was it weighed? At minimum it seems worth documenting here, with `clear_on_success = True` as the way to get clear-to-rerun back.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +363,146 @@ def on_kill(self):
...
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "EntityNotFoundException":
+                return "NOT_FOUND"
+            raise
+
+    def is_job_active(self, status: str) -> bool:
+        return status not in (*JOB_RUN_TERMINAL_STATES, NOT_FOUND_STATE)
+
+    def is_job_succeeded(self, status: str) -> bool:
+        if status == "STOPPED":
```
kaxil: `STOPPED` isn't only reachable from a manual console stop. `on_kill` stops the run itself when `stop_job_run_on_kill=True`, and `on_kill` fires both on SIGTERM and on `execution_timeout` (`task_runner.py` calls `task.on_kill()` in both), which leave the task `UP_FOR_RETRY` when retries remain. So a task that trips `execution_timeout` stops its own Glue run, and the retry then reads `STOPPED` and reports success with the ETL half done. Before this PR that retry resubmitted.

The warning also says "stopped outside Airflow", which is the one case it isn't. Could `on_kill` drop the stored run id when it stops the run, so the retry resubmits instead of inheriting a `STOPPED` it caused itself?

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +363,146 @@ def on_kill(self):
...
+                return existing_job_run_id
+        self.log.info(
+            "Initializing AWS Glue Job: %s. Wait for completion: %s",
+            self.job_name,
+            self.wait_for_completion,
+        )
+        glue_job_run = self.hook.initialize_job(script_args, self.run_job_kwargs)
+        # Set immediately (before any polling) so on_kill can stop the run even if the worker dies
+        # before poll_until_complete runs.
+        self._set_job_run_id(context, glue_job_run["JobRunId"])
```
kaxil, line 467: The `context["ti"].xcom_push(key="glue_job_run_id", ...)` that main does right after `initialize_job` is gone, and `_set_job_run_id` doesn't replace it (it sets `self._job_run_id` and persists the link; the SDK mixin writes only to `task_state_store`). Two consequences.

Downstream tasks doing `xcom_pull(task_ids="glue_job", key="glue_job_run_id")` stop getting a value. The `return_value` XCom still carries the id, but the explicit key is what people wrote against.

And the XCom tier in `_find_previous_job_run` can never hit now: the push inside the scan branch is the only remaining writer, so the `xcom_pull` above it is always `None` and every durable attempt falls through to the history scan. The description says the XCom-then-scan mechanism is kept exactly as it was, but the write half was dropped. The assertions that guarded the push (`assert len(xcom_calls) == 1, "Should push new glue_job_run_id"`) were deleted from `test_check_previous_job_id_run_new_on_finished` in this PR, which is why nothing caught it.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +363,146 @@ def on_kill(self):
...
+                        return existing_job_run_id
+            except Exception:
+                self.log.warning("Failed to find previous Glue job run by task UUID", exc_info=True)
+        return None
+
+    def submit_job(self, context: Context) -> str:
+        """Start a Glue job run and return its run id, or reconnect to one this task already started."""
+        script_args = self._build_script_args(context)
+        if self.durable:
+            existing_job_run_id = self._find_previous_job_run(context, script_args[self.TASK_UUID_ARG])
```
kaxil, line 455: This runs on the first attempt too, not just retries, and with `durable` defaulting to `True` every Glue task now pays it. Because nothing pushes the XCom key any more, `_find_previous_job_run` always takes the scan branch, and `_find_job_run_id_by_task_uuid` paginates the entire run history at 50 per page when there's no match, so a busy job is dozens of `GetJobRuns` calls before every submission. It also runs on the deferrable path, since `execute` calls `submit_job` there.

Can it be gated on the attempt actually being a retry (`context["ti"].try_number > 1`), or on the task state store being absent, rather than on `durable` alone? Line 174 of the docs describes it as the fallback for when the store is unavailable, which is what I expected the code to do.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,70 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+
+  glue_job = GlueJobOperator(
+      task_id="glue_job",
+      job_name="my_glue_job",
+      script_location="s3://glue-examples/glue-scripts/sample_aws_glue_job.py",
+      durable=False,
+  )
+
+Durable execution applies to the synchronous path. When ``deferrable=True`` is set, the Triggerer
+already tracks the run across the wait, so deferrable mode takes precedence and ``durable`` has no
```
kaxil, line 208: `durable` isn't fully inert on the deferrable path. `execute` calls `submit_job`, which for `durable=True` (the new default) injects `--airflow_task_uuid` into every run's arguments and runs the reconnect lookup before submitting. Deferrable Glue runs get an extra script arg and a history scan they didn't have before. Either gate `_build_script_args`/`_find_previous_job_run` on `not self.deferrable`, or say here what deferrable plus durable actually does.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +363,146 @@ def on_kill(self):
...
+    def get_job_status(self, external_id: JsonValue, context: Context) -> str:
+        """Query the raw job run state; a run id Glue no longer knows about degrades to NOT_FOUND."""
+        job_run_id = cast("str", external_id)
+        # This is the first place a reconnecting attempt learns the run id.
+        self._set_job_run_id(context, job_run_id)
+        try:
+            return self.hook.get_job_state(self.job_name, job_run_id)
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "EntityNotFoundException":
+                return "NOT_FOUND"
```
kaxil: `NOT_FOUND_STATE` is defined above and used in `is_job_active`, but this returns the literal. Worth using the constant so the two can't drift, since a mismatch makes `is_job_active` treat a deleted run as active.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Fixed. Updated the docs because I think clearing task should reconnect to an existing job if its present. Documented the escape hatch.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Correct. The fix isnt fixable at the `on_kill` layer the way suggested. `on_kill(self)` misses a context, so it structurally cannot reach `task_state_store` to invalidate anything, and Glue's API gives no way to tell a console cancelled run from one that stopped itself. Since the two cases cannot be distinguished, removed the distinction: `STOPPED` no longer counts as success at all, its now bucketed with `FAILED/TIMEOUT` and always resubmits. 

Trades a rare wasted resubmission (genuine external cancel) for eliminating a silent false success (self-inflicted stop) which is a safer default for correctness.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: restored the unconditional xcom_push after a fresh submit (matches the pre-durable-execution code exactly: it ran outside the resume_glue_job_on_retry check, so it should run regardless of durable too). 

Fixes both consequences: downstream `xcom_pull(key="glue_job_run_id")` works again, and `_find_previous_job_run`'s xcom tier is reachable on the next retry. Added back the assertion that used to guard this `(assert_any_call(key="glue_job_run_id", ...))`, and confirmed it actually fails without the fix before adding it.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Added the`try_number > 1` guard, you are right that it matters now that comment 3 restored the xcom write, and the docs already claimed this was fallback only behavior. The `--airflow_task_uuid` tag itself still gets written on every attempt including the first (a later retry's scan needs it to have been there from the start), only the lookup is skipped.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Thanks, gated stuff on deferrable now.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Woops, handled.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Handled in 85398bfb29

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 85398bfb29

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 85398bfb29

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 85398bfb29

providers/amazon/docs/operators/glue.rst
amoghrajesh: Handled in 85398bfb29

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 85398bfb29

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +366,150 @@ def on_kill(self):
...
+        self._set_job_run_id(context, job_run_id)
+        try:
+            return self.hook.get_job_state(self.job_name, job_run_id)
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "EntityNotFoundException":
+                return NOT_FOUND_STATE
+            raise
+
+    def is_job_active(self, status: str) -> bool:
+        return status not in (*JOB_RUN_TERMINAL_STATES, NOT_FOUND_STATE)
```
kaxil, line 517: `STOPPING` isn't in `JOB_RUN_TERMINAL_STATES`, so it counts as active and a retry reconnects to it. `batch_stop_job_run` is asynchronous, so a run this operator's own `on_kill` just stopped sits in `STOPPING` for a while, and a retry landing in that window reconnects, polls, and `_handle_state`'s `finished_states = ["SUCCEEDED", "STOPPED"]` returns normally, so the task goes green with the ETL half done.

I ran it against the real hook. Serving `STOPPING, STOPPING, STOPPED` gives `Reconnecting to existing job status=STOPPING`, then `Exiting Job jr_old Run State: STOPPED`, `initialize_job` never called, task succeeds. On the merge-base the same scenario resubmits, because the old check was an allow-list (`state in ("RUNNING", "STARTING")`) rather than a deny-list. So this is the same false success the `STOPPED` change was meant to remove, reached through `STOPPING` instead, and it is now on the default path.

Adding `STOPPING` to the terminal tuple trades the silent success for a probably-transient `ConcurrentRunsExceededException`. The narrower fix is to raise in `poll_until_complete` when `job_completion` returns a state outside `JOB_RUN_SUCCESS_STATES`, which also removes the disagreement between that tuple and the hook's `finished_states`. Clearing a running task is the most reachable route into this, since it reschedules immediately rather than waiting out a `retry_delay`.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,79 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+reconnects to the run that is already executing in Glue instead of starting a new one.
+
+This matters more for Glue because a Glue job's ``concurrent_run_limit`` defaults to ``1``, so
+submitting a second run while the first is still active does not create a harmless duplicate, it
+fails outright with ``ConcurrentRunsExceededException`` and the task keeps retrying against a run
+it can never see. Durable execution turns that retry into a normal reconnect.
+
+On retry the operator checks the prior run's state:
+
+* if it is still starting, running, waiting for capacity, or being stopped, the operator
```
kaxil, line 167: This bullet says a run that is "being stopped" reconnects and continues polling, and the paragraph at 164-170 says a stopped run always resubmits so a self-inflicted stop is never reported as a false success. The code implements this bullet, and reconnecting to a `STOPPING` run ends with `_handle_state` reporting the resulting `STOPPED` as finished, which is exactly what defeats the paragraph below. Whichever way the code fix goes, this bullet needs to move with it.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +366,150 @@ def on_kill(self):
...
+
+    def submit_job(self, context: Context) -> str:
+        """Start a Glue job run and return its run id, or reconnect to one this task already started."""
+        script_args = self._build_script_args(context)
+        # The --airflow_task_uuid tag is still written on every nondeferred attempt,
+        # including the first: a later retry's scan can only find it if it was there from the
+        # start. The lookup itself is skipped on the first attempt, since nothing tagged with this
+        # task's UUID could possibly exist yet -- paying for a full run-history scan there would
+        # be pure waste.
+        if self.durable and not self.deferrable and context["ti"].try_number > 1:
```
kaxil: The `try_number` guard removes the first-attempt cost, but not the case that matters on 3.3+. When the task state store returns a terminal run id, `execute_resumable` logs "Prior job in terminal state, resubmitting fresh" and calls `submit_job`, which runs `_find_previous_job_run` anyway, and since Airflow 3 clears every one of the task instance's XComs before a non-deferral attempt that always takes the scan branch.

With the store returning `FAILED` I get a full walk of the run history at 50 per page before `initialize_job` fires, so a job with a few thousand runs is dozens of `GetJobRuns` calls on every retry after every failure. That is the cost the PR's own "one call vs. a full scan" rationale says durable execution removes, and the docs call this tier a one-time bootstrap. `submit_job` can't see the mixin's decision, so it needs something explicit, like a flag set when `task_state_store.get()` returned an id, leaving the bootstrap lookup live only for the store-empty and pre-upgrade cases.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +366,150 @@ def on_kill(self):
...
+            self.job_name,
+            self.wait_for_completion,
+        )
+        glue_job_run = self.hook.initialize_job(script_args, self.run_job_kwargs)
+        # Set immediately (before any polling) so on_kill can stop the run even if the worker dies
+        # before poll_until_complete runs.
+        self._set_job_run_id(context, glue_job_run["JobRunId"])
+        # Pushed unconditionally, matching pre-durable-execution behavior: downstream tasks read
+        # this key directly, and it's also the only writer that makes _find_previous_job_run's
+        # XCom tier reachable on the next retry.
```
kaxil: The second clause isn't true on Airflow 3.x. `xcom_keys_to_clear` is built from a query that selects every key for the task instance with no key filter, and it is skipped only for a deferral, so `glue_job_run_id` is always gone before a retry runs. That is the finding the PR description leads with. The XCom tier is only live on 2.11 through 3.2, so the push is right for the downstream-consumers reason but not for this one. Worth version-qualifying it, otherwise the next reader takes a path that is dead on 3.x for a live recovery tier.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,79 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+when someone runs ``airflow state-store clean``. If a task's ``retry_delay`` is longer than
+``[state_store] default_retention_days`` (30 days by default) and cleanup runs in between, the run
+id won't be there for the next retry, and the operator falls back to the XCom/scan mechanism
+above rather than reconnecting via task state store. Avoid running cleanup on a schedule shorter
+than your longest ``retry_delay``.
+
+Clearing a task is treated the same as a retry, which matters specifically for a task whose job
+already succeeded: clearing does not delete the stored run id, so the next attempt reads it back
+and returns immediately without submitting anything to Glue. See
+:doc:`apache-airflow:core-concepts/resumable-tasks` for why, and for the
```
kaxil, line 209: `core-concepts/resumable-tasks` never mentions `clear_on_success`, or clearing at all. The setting is documented in `core-concepts/task-state-store` under "Automatic cleanup (``clear_on_success``)", which this page already links at line 147, so the link just needs repointing.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,79 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+
+  glue_job = GlueJobOperator(
+      task_id="glue_job",
+      job_name="my_glue_job",
+      script_location="s3://glue-examples/glue-scripts/sample_aws_glue_job.py",
+      durable=False,
+  )
+
+Durable execution applies to the synchronous path. When ``deferrable=True`` is set, the Triggerer
+already tracks the run across the wait, so deferrable mode takes precedence and ``durable`` has no
```
kaxil: Gating on `not self.deferrable` was my suggestion, but it does give up something main had. There the `resume_glue_job_on_retry` block ran before `self.defer(...)`, so a retry of a deferrable task scanned by task UUID and reattached to the still-running run instead of resubmitting. With the gate, `deferrable=True` always resubmits, which with `concurrent_run_limit` defaulting to 1 is the `ConcurrentRunsExceededException` loop this PR's own rationale describes. Anyone who set `resume_glue_job_on_retry=True` alongside `deferrable=True` loses that, so it is worth saying so here rather than only "durable has no effect", since the deprecation note below implies the value just maps across.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,79 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+told apart, a stopped run always resubmits, so a self-inflicted stop never gets silently reported
+as a false success.
+
+This protection also applies when ``wait_for_completion=False`` -- even though that task attempt
+never polls at all, a retry after a successful submission still reconnects rather than
+resubmitting, since the run id is persisted immediately after submission regardless of whether the
+task waits for it to finish.
+
+Durable execution requires Airflow 3.3 or newer for the task state store lookup above. On earlier
+Airflow versions, or if the task state store is unavailable at runtime, ``durable=True`` still
```
kaxil: "unavailable" is doing double duty in this sentence, and only one of its readings holds.

Accessor absent from the context (pre-3.3): `execute_resumable` warns and falls through to `submit_job`, so the XCom/scan tier runs. Key simply not stored: the accessor gets `TASK_STORE_NOT_FOUND` back, returns its `default`, and the same fallthrough happens. Both of those match the promise.

Store present but erroring does not. `TaskStateStoreAccessor.get` does IPC to the supervisor, and `_extract_get_response` raises `AirflowRuntimeError` on any error response that isn't `TASK_STORE_NOT_FOUND` (`task-sdk/src/airflow/sdk/execution_time/context.py:586-587`). There is no `except` anywhere in `resumablejobmixin.py`, so that propagates out of `execute_resumable` before `submit_job` is reached, and the XCom/scan tier cannot run, on that attempt or on any retry hitting the same error. That is the case this sentence is specifically promising it covers.

The raise is deliberate in the shared helper and you have scoped task-sdk changes out, so narrowing the wording to the absent and not-stored cases looks like the right fix here.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +366,150 @@ def on_kill(self):
...
+        context["ti"].xcom_push(key="glue_job_run_id", value=glue_job_run["JobRunId"])
+        return glue_job_run["JobRunId"]
+
+    def get_job_status(self, external_id: JsonValue, context: Context) -> str:
+        """Query the raw job run state; a run id Glue no longer knows about degrades to NOT_FOUND."""
+        job_run_id = cast("str", external_id)
+        # This is the first place a reconnecting attempt learns the run id.
+        self._set_job_run_id(context, job_run_id)
```
kaxil, line 508: On the terminal-resubmit path this persists `GlueJobRunDetailsLink` and logs "You can monitor this Glue Job run at" for the run that is about to be discarded, and then both get rewritten for the new id. My probe log shows `.../run/jr_old` immediately followed by `.../run/jr_new` within one attempt. If `initialize_job` then raises, which `ConcurrentRunsExceededException` makes realistic here, `self._job_run_id` is left pointing at the stale terminal run, and that is what `on_kill` would try to stop.

providers/amazon/tests/unit/amazon/aws/operators/test_glue.py
```diff
@@ -781,6 +784,295 @@ def test_inject_parent_job_info_with_resume_on_retry(
...
+        mock_initialize_job.return_value = {"JobRunId": "jr_new"}
+        mock_job_completion.return_value = {"JobRunState": "SUCCEEDED"}
+        store = FakeTaskStateStore()
+
+        job_run_id = glue.execute(self._context(store, try_number=1))
+
+        assert job_run_id == "jr_new"
+        mock_initialize_job.assert_called_once()
+        mock_conn.return_value.get_job_run.assert_not_called()
+        mock_conn.return_value.get_job_runs.assert_not_called()
```
kaxil, line 1246: The invariant the comment in `submit_job` rests on, that the tag is still written on attempt 1 so a later retry's scan can find it, isn't asserted anywhere. One more line here, `assert GlueJobOperator.TASK_UUID_ARG in mock_initialize_job.call_args[0][0]`, would stop a later refactor from moving the injection under the `try_number` guard and quietly killing recovery on 2.11 through 3.2.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Thanks, detailed review. I confirmed by reproducing your exact scenario. Went with option 2 - polling now checks the state the hook's poll loop actually finished on, and raises (new `GlueJobRunStoppedError`) if it's 
not a real success. 

STOPPING still correctly counts as active - nothing changes there, this only intervenes at the moment the hook's own finished_states disagrees with ours. Added a test that reproduces your log sequence exactly (STOPPING -> STOPPED) and confirmed it fails without the fix.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Fixed to match the actual fix from previous comment: STOPPING still reconnects and polls, a directly stored STOPPED resubmits, and a reconnect that settles into STOPPED mid-poll now raises rather than either resubmitting or succeeding.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Fixed. submit_job now checks state store directly. If it already has a value for the run-id key (even a terminal one), the scan is skipped entirely, since your own check already ruled out reconnecting. It only runs when there's genuinely nothing recorded: first attempt, no store (pre-3.3), or a store that's never seen this key.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Reworded — the docstring now says plainly that the XCom tier in 2.x, and that Airflow 3.x always misses there so the scan runs every time.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Its handled by #71356

providers/amazon/docs/operators/glue.rst
amoghrajesh: You were right, restored it. `durable` now tags every attempt (deferrable included) with the task's identity, and a deferrable retry scans for that tag and reattaches instead of resubmitting into `ConcurrentRunsExceededException`, matching what main did. Docs updated to describe this instead of "durable has no effect on deferrable."

providers/amazon/docs/operators/glue.rst
amoghrajesh: Reworded to name the two cases that actually fall through (no store, or nothing recorded yet), rather than "unavailable" implying it covers the store-errors case too, which it doesn't.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Good catch. `submit_job` now clears `_job_run_id` right before calling initialise, and `on_kill` is a no op if it's empty. So a failure there leaves nothing stale for on_kill to act on.

providers/amazon/tests/unit/amazon/aws/operators/test_glue.py
amoghrajesh: Added.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 118dacfd80

providers/amazon/docs/operators/glue.rst
amoghrajesh: Handled in 118dacfd80

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 118dacfd80

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 118dacfd80

providers/amazon/docs/operators/glue.rst
amoghrajesh: Handled in 118dacfd80

providers/amazon/docs/operators/glue.rst
amoghrajesh: Handled in 118dacfd80

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Handled in 118dacfd80

providers/amazon/tests/unit/amazon/aws/operators/test_glue.py
amoghrajesh: Handled in 118dacfd80

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+        attempt, so it always misses there and the scan runs every time.
+        """
+        ti = context["ti"]
+        previous_job_run_id = ti.xcom_pull(key="glue_job_run_id", task_ids=ti.task_id)
+        if previous_job_run_id:
+            try:
+                job_run = self.hook.conn.get_job_run(JobName=self.job_name, RunId=previous_job_run_id)
+                state = job_run.get("JobRun", {}).get("JobRunState")
+                self.log.info("Previous Glue job_run_id: %s, state: %s", previous_job_run_id, state)
+                if state in ("RUNNING", "STARTING"):
```
kaxil: This keeps the merge-base literal `if state in ("RUNNING", "STARTING")` here and again at line 431, but the new `is_job_active` at line 479 is a deny-list that counts `WAITING` and `STOPPING` as active, and the docs added at glue.rst:158 promise the operator reconnects when a run is "waiting for capacity, or being stopped".

This tier matters more than it looks: it is the only recovery mechanism on the deferrable path (`execute` calls `submit_job` directly and nothing persists an id there, so `_has_stored_external_id` is always False), the only one below Airflow 3.3, and the one used on 3.3+ when the worker died before the store write. `WAITING` is exactly "queued behind the concurrent-run limit", which is the situation glue.rst:151-154 builds the whole feature around.

To be fair about severity, this is not a regression: on main the default was `resume_glue_job_on_retry=False`, so attempt 2 resubmitted anyway. The narrower claim is that the implementation does not do what `is_job_active` and its own new docs say.

Ask: `if self.is_job_active(state):` here and `if self.is_job_active(existing_job_run_state):` at 431, plus the history-scan tests parametrized over `WAITING` and `STOPPING` (only the task-state-store tests cover those today).

Different code path from my round-2 comment on line 487, so this is not re-litigating it: that fix landed in `poll_until_complete`, this is the reattach decision.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+            self.wait_for_completion,
+        )
+        # A prior get_job_status call may have set this to a stale, terminal run id while checking
+        # whether to reconnect. Clear it so on_kill has nothing to act on if initialize_job raises.
+        self._job_run_id = None
+        glue_job_run = self.hook.initialize_job(script_args, self.run_job_kwargs)
+        # Set before polling so on_kill can stop the run even if the worker dies immediately after.
+        self._set_job_run_id(context, glue_job_run["JobRunId"])
+        # Downstream tasks read this key directly; it's also what feeds the XCom tier above.
+        context["ti"].xcom_push(key="glue_job_run_id", value=glue_job_run["JobRunId"])
```
kaxil: Follow-up on my round-1 comment. The push is back, but only on the branch that submits. `execute_resumable` returns from the reconnect branch and from the already-succeeded branch (`task-sdk/src/airflow/sdk/bases/resumablejobmixin.py:196-199`) without ever reaching `submit_job`, and every non-deferral attempt clears the task instance's XComs first. So after a retry that reconnects, or a clear that short-circuits on an already-succeeded run, `xcom_pull(task_ids="glue_job", key="glue_job_run_id")` returns None.

The asymmetry is what makes it read as an oversight: `_set_job_run_id` already carries the other per-attempt side effect (`GlueJobRunDetailsLink.persist`) on all four paths, so the link survives and only the run-id key was left behind.

Scope, stated honestly: `execute` still returns the run id, so `glue_job.output` and keyless `xcom_pull` are unaffected, and there is no in-repo consumer of the key. It is a de-facto contract rather than a documented one, but it is the one the comment on the line above asserts ("Downstream tasks read this key directly").

Ask: move the push into `_set_job_run_id` beside the `persist` call and drop it here. It is already called on all four paths, already de-duplicated by the `if self._job_run_id == job_run_id: return` guard, and same-key `xcom_push` is delete-then-insert, so a stale-then-fresh sequence still ends on the correct id, exactly as the link already does.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+    def _has_stored_external_id(self, context: Context) -> bool:
+        task_state_store = context.get("task_state_store")
+        return task_state_store is not None and task_state_store.get(self.external_id_key) is not None
+
+    def submit_job(self, context: Context) -> str:
+        """Start a Glue job run and return its run id, or reconnect to one this task already started."""
+        script_args = self._build_script_args(context)
+        # Scan only when there's nothing else to go on: first attempt, no store, or a store that
+        # never recorded this key. A stored id (even terminal) means the caller already decided.
+        if self.durable and context["ti"].try_number > 1 and not self._has_stored_external_id(context):
```
kaxil: On the deferrable path this read cannot help and can hurt. Nothing persists an id there, so `_has_stored_external_id` can only return False and never changes the decision. But `TaskStateStoreAccessor.get` raises `AirflowRuntimeError` for every error other than `TASK_STORE_NOT_FOUND` (`task-sdk/src/airflow/sdk/execution_time/context.py:585-587`), and `task_state_store` is unconditionally present in the runtime context on 3.3+ (`task-sdk/src/airflow/sdk/execution_time/task_runner.py:334`).

So a store that errors makes a deferrable retry fail from inside `submit_job` before the task-UUID scan can run, which removes the recovery glue.rst:213-216 promises for that exact path. This arrived with the guard added for my round-2 comment on line 454, so it is worth closing while that change is fresh.

Ask: `and not self.deferrable` on the condition, or hoist the lookup out of `submit_job` and have the deferrable branch pass the decision in.

Related design question, if you would rather fix the cause than the symptom: should the deferrable branch persist the run id before `defer()`? The deferrable path would then reconnect through the cheap store lookup instead of a full `get_job_runs` walk, and this plus the `WAITING`/`STOPPING` gap above would both stop existing there.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -158,6 +203,16 @@ def __init__(
         ),
         **kwargs,
     ):
+        if resume_glue_job_on_retry is not None:
+            # Kept as a real named parameter (not **kwargs) so `default_args` still applies correctly.
```
kaxil: This comment is about `resume_glue_job_on_retry`, and it is accurate for that parameter, but `durable` itself is not a named parameter on `GlueJobOperator.__init__`. It reaches the mixin through `**kwargs` via `kwargs.setdefault`. `_apply_defaults` only fills parameters present in the decorated function's own signature (`task-sdk/src/airflow/sdk/bases/operator.py:527-529`), and the pre-3.3 fallback stub's `__init__` carries no `_apply_defaults` at all.

Net effect on Airflow 2.11 through 3.2, which is inside this provider's declared floor: `default_args={"durable": False}` leaves `durable` True with no error, while `default_args={"resume_glue_job_on_retry": False}` works. On the versions where the deprecation warning is deliberately suppressed, the deprecated parameter is the only one that works at DAG level. That is not cosmetic there, since `durable` still gates the task-UUID injection and the retry scan below 3.3.

This matters mainly because the PR flips the default to True: the documented way to opt out of a new default does not work on those versions. The skipif reason on `test_default_args_durable_reaches_operator` records the mechanism, but nothing user-facing does.

Ask: `durable: bool | None = None` on `GlueJobOperator.__init__`, forwarded only when set, which preserves the mixin's own default. Fair caveat: the sibling operators route `durable` through `**kwargs` the same way, so if you would rather treat it as a cross-provider issue, then stating the limitation beside the `durable=False` example in glue.rst is the minimum.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -158,6 +203,16 @@ def __init__(
...
+            if AIRFLOW_V_3_3_PLUS:
```
kaxil, line 232: Provider deprecations are scoped by provider version, not Airflow version. The same amazon wheel serves 2.11 through 3.4, and the removal will land in a future amazon release for all of them, so gating the warning on `AIRFLOW_V_3_3_PLUS` hides it from part of the install base for the whole deprecation window.

The usual justification for such a gate does not apply here: below 3.3 the compat stub still sets `self.durable`, and the operator honours it at lines 398 and 447, so `durable` is a functional drop-in for what `resume_glue_job_on_retry` did. The shipped precedent warns unconditionally (`providers/apache/spark/src/airflow/providers/apache/spark/operators/spark_submit.py:405-411`, no version gate).

A team on Airflow 3.1 passing `resume_glue_job_on_retry=True` therefore sees nothing until the parameter is removed, and then gets a `TypeError`.

Ask: drop the gate, invert `test_silent_below_3_3` to assert the warning fires on every supported version, and remove the "on Airflow 3.3 and newer" qualifier at glue.rst:220.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+            script_args = inject_transport_information_into_glue_arguments(script_args, context)
+        if self.durable:
+            script_args, _ = self._prepare_script_args_with_task_uuid(context, base_args=script_args)
+        return script_args
+
+    def _find_previous_job_run(self, context: Context, task_uuid: str) -> str | None:
+        """
+        Look for a Glue job run this task instance already started.
+
+        Checks XCom for a cached run id first, then falls back to a task-UUID scan. The XCom tier
```
kaxil, line 437: Follow-up on my round-2 comment on line 470. The version boundary here is off by three releases, and I should say plainly that my own round-2 wording pointed you at the wrong one.

Pre-attempt XCom clearing shipped in 3.0.0, not 3.3. At tag 3.0.0 the execution API builds `xcom_keys_to_clear` for any attempt with no `next_method`, and the SDK deletes every one of those keys. Present and unconditional at 3.0.0, 3.1.0 and 3.2.0.

Consequences: the XCom tier is dead on every Airflow 3 release, the `if previous_job_run_id:` branch below is dead code there, and glue.rst:180-184 tells 3.0-3.2 users they get a cheap XCom lookup before the scan when in fact every durable retry pays the full paginated history walk. The comment at line 463 ("it's also what feeds the XCom tier above") is false for the same reason.

Ask: reword to "Airflow 2.x only; every Airflow 3 release clears task XComs before each non-deferral attempt", and fix glue.rst:182 to match. Since 2.11 is the only beneficiary, dropping the XCom tier is worth considering. Note also that `test_find_previous_job_run_reuses_from_xcom` and `test_find_previous_job_run_does_not_fall_back_to_scan_on_xcom_state_mismatch` pin a path unreachable on any Airflow 3 release, because they stub `xcom_pull` to return a value the runtime never will.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+        task_state_store = context.get("task_state_store")
+        return task_state_store is not None and task_state_store.get(self.external_id_key) is not None
+
+    def submit_job(self, context: Context) -> str:
+        """Start a Glue job run and return its run id, or reconnect to one this task already started."""
+        script_args = self._build_script_args(context)
+        # Scan only when there's nothing else to go on: first attempt, no store, or a store that
+        # never recorded this key. A stored id (even terminal) means the caller already decided.
+        if self.durable and context["ti"].try_number > 1 and not self._has_stored_external_id(context):
+            existing_job_run_id = self._find_previous_job_run(context, script_args[self.TASK_UUID_ARG])
```
kaxil, line 487: Worth a second look now that this runs by default rather than under an opt-in flag. `_find_job_run_id_by_task_uuid` is a `while True` over `get_job_runs(MaxResults=50)` with no page cap and no age window, and the no-match case (the prior attempt died before submitting anything, which is the common retry shape) is the one that walks the job's entire run history.

Two consequences that are new because the flag flipped, not because the function changed:

The scan needs `glue:GetJobRuns`. A task policy granting only `StartJobRun` and `GetJobRun`, which was sufficient before, now hits AccessDenied on every retry.

The failure is swallowed. Both `except Exception` blocks (lines 419 and 434) log at warning and return None, after which the operator submits fresh, producing `ConcurrentRunsExceededException` against the live run, which is the failure this feature exists to prevent. There is no error-level signal to tell an operator why.

Not asking for fail-closed: those blocks are verbatim from the merge base, and turning a transient throttle into a task failure would be worse.

Ask: bound the walk (a page cap, or stop once `StartedOn` predates the DAG run), narrow the `except Exception` to `ClientError`, log the fall-through at error level with the reason, and add `glue:GetJobRuns` to the IAM note in glue.rst.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+
+    def _build_script_args(self, context: Context) -> dict:
+        script_args = dict(self.script_args)
+        if self.openlineage_inject_parent_job_info:
+            self.log.info("Injecting OpenLineage parent job information into Glue job arguments.")
+            script_args = inject_parent_job_information_into_glue_arguments(script_args, context)
+        if self.openlineage_inject_transport_info:
+            self.log.info("Injecting OpenLineage transport information into Glue job arguments.")
+            script_args = inject_transport_information_into_glue_arguments(script_args, context)
+        if self.durable:
```
kaxil: With `durable` defaulting to True, this injects `--airflow_task_uuid` into the Arguments of every Glue job run. `initialize_job` passes the dict straight to `start_job_run` (`providers/amazon/src/airflow/providers/amazon/aws/hooks/glue.py:271`), and Glue surfaces Arguments to the script as command-line arguments. At the merge base this happened only under `resume_glue_job_on_retry=True`, which defaulted to False, so a provider upgrade alone now changes the argv of unchanged production scripts.

Hedging this deliberately, because it is not clear-cut: `getResolvedOptions`, the documented Glue idiom, uses `parse_known_args` and tolerates the extra argument, so scripts following the docs are unaffected. A script doing its own strict parsing with `argparse.ArgumentParser().parse_args()` exits 2 on an unrecognised argument. Common, not universal.

Ask: either call this out in the docs as an upgrade note, or inject only where the tag is actually the recovery mechanism (`self.deferrable or not AIRFLOW_V_3_3_PLUS`), since on 3.3+ synchronous runs the task state store is the primary tier and the tag only covers the crash-before-persist window the mixin's docstring already declares unclosable.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -384,6 +364,147 @@ def on_kill(self):
...
+        self._set_job_run_id(context, job_run_id)
+        if not self.wait_for_completion:
+            self.log.info("AWS Glue Job: %s. Run Id: %s", self.job_name, job_run_id)
+            return
+        glue_job_run = self.hook.job_completion(
+            self.job_name, job_run_id, self.verbose, self.sleep_before_return
+        )
+        state = glue_job_run["JobRunState"]
+        self.log.info("AWS Glue Job: %s status: %s. Run Id: %s", self.job_name, state, job_run_id)
+        if state not in JOB_RUN_SUCCESS_STATES:
```
kaxil, line 533: Completing the thread from my round-2 comment on glue.rst:158 rather than reversing it. This raise is the fix I asked for, and it is the right end state: the deferrable non-verbose path's `job_complete` waiter already treats `STOPPED` as a failure, so this makes the synchronous path agree with it.

The issue is scoping. The raise lives in `poll_until_complete`, which `execute_resumable` also calls on its non-durable branch, and it fires on first attempts too, so it is not gated on `durable` in any way. Concretely: someone cancels a Glue run in the console, and on main `_handle_state` buckets `STOPPED` in `finished_states` so the task logs `status: STOPPED` and goes green. At this HEAD it goes red and burns a retry that resubmits the job they just cancelled.

The docs do not carry that. The stopped-run paragraph at glue.rst:165-173 sits entirely under "Durable execution", and glue.rst:199 offers `durable=False` as the opt-out, so a reader reasonably concludes `durable=False` restores the old behaviour. It does not.

Ask: move that paragraph out of the durable section and reword it to cover first attempts and `durable=False`, add a sentence saying explicitly that `durable=False` does not restore `STOPPED`-as-success, and drop the "durable" framing from the code comment on the line above.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -35,17 +38,52 @@
...
 from airflow.providers.amazon.aws.utils.mixins import aws_template_fields
+from airflow.providers.amazon.version_compat import AIRFLOW_V_3_3_PLUS
 from airflow.providers.common.compat.openlineage.utils.spark import (
     inject_parent_job_information_into_glue_arguments,
     inject_transport_information_into_glue_arguments,
 )
 from airflow.providers.common.compat.sdk import AirflowException, conf
 
+# ResumableJobMixin only exists on Airflow 3; this provider still targets >=2.11. Drop this
+# fallback once the provider's minimum Airflow version is >=3.0.
```
kaxil: Threshold looks wrong. The mixin reads `context["task_state_store"]`, which only exists on Airflow 3.3+, and the stub's own docstring below, the operator docstring at line 186, and the test class gate all say 3.3. A maintainer who bumps the floor to 3.0 and follows this comment would delete the fallback and break every Glue DAG on 3.0 through 3.2 with a parse-time ImportError.

Ask: `>=3.3`. Worth noting none of the four shipped sibling stubs (redshift_data, databricks, snowflake, bigquery) carries this comment at all.

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,86 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+  glue_job = GlueJobOperator(
+      task_id="glue_job",
+      job_name="my_glue_job",
+      script_location="s3://glue-examples/glue-scripts/sample_aws_glue_job.py",
+      durable=False,
+  )
+
+The task state store lookup above is only used on the synchronous path -- when ``deferrable=True``
+is set, the Triggerer already tracks the run across the wait, so a run id is never persisted there.
+``durable`` still has an effect on retry, though: a retry of a deferrable task resubmits by
```
kaxil: This sentence is now backwards. With `durable=True` as the default, a retry of a deferrable task scans for the tagged run first; resubmitting is what you get when you opt out with `durable=False`. As written, "resubmits by default" asserts the opposite of what line 447 of the operator does.

Ask: "would otherwise resubmit". (Separate sentence from the deferrable point in my round-2 comment, so this is a distinct fix.)

providers/amazon/docs/operators/glue.rst
```diff
@@ -139,6 +139,86 @@ To submit a new AWS Glue job you can use :class:`~airflow.providers.amazon.aws.o
...
+resubmitting, since the run id is persisted immediately after submission regardless of whether the
+task waits for it to finish.
+
+Durable execution requires Airflow 3.3 or newer for the task state store lookup above. On earlier
+Airflow versions, or if a prior run was never recorded to the task state store, ``durable=True``
+still recovers a prior run via an older mechanism: the operator checks XCom for a cached run id
+first, then falls back to scanning the job's run history for a run tagged with this task
+instance's identity, and reconnects if it finds one that is still active.
+
+Like the persisted state itself, the stored run id isn't deleted automatically, that only happens
```
kaxil, line 198: `airflow db clean` also targets this table: `_TableConfig(table_name="task_state_store", recency_column_name="expires_at", ...)` in `airflow-core/src/airflow/utils/db_cleanup.py`. So "that only happens when someone runs `airflow state-store clean`" understates when the stored id can disappear.

Worth fixing because this paragraph's own mitigation advice, avoiding a cleanup schedule shorter than the longest `retry_delay`, only lands if the reader knows `db clean` counts too.

providers/amazon/src/airflow/providers/amazon/aws/exceptions.py
```diff
@@ -78,3 +78,7 @@ class NeptuneImportTaskCancellationFailedError(AirflowException):
 
 class NeptuneImportTaskFailedError(AirflowException):
     """Raised when a Neptune Analytics import task fails to complete successfully."""
+
+
+class GlueJobRunStoppedError(AirflowException):
+    """Raised when a reconnected Glue job run finishes in a state that is not a real success."""
```
kaxil: "reconnected" is narrower than the behaviour. `poll_until_complete` raises this on the fresh-submit path as well, and on first attempts, so the docstring understates the blast radius for `durable=False` users.

Ask: drop the word, for example "Raised when a Glue job run finishes in a state that is not a real success."

providers/amazon/tests/unit/amazon/aws/operators/test_glue.py
```diff
@@ -163,6 +171,62 @@ def test_execute_deferrable(self, _, mock_initialize_job):
         assert defer.value.trigger.attempts == 75
         assert defer.value.trigger.aws_conn_id == "aws_default"
 
+    @mock.patch.object(GlueJobHook, "initialize_job")
+    @mock.patch.object(GlueJobHook, "get_conn")
+    def test_deferrable_first_attempt_injects_task_uuid_but_skips_scan(
```
kaxil, line 190: The test asserts the `TASK_UUID_ARG` half of its name but nothing for "skips_scan", so the part most likely to regress is unguarded. `test_first_attempt_skips_the_retry_lookup_entirely` at line 1097 already does this properly.

Ask: add `mock_get_conn.return_value.get_job_runs.assert_not_called()`. Worth more than a naming nit here, because the scan has no page cap and an unspecced conn mock returns a truthy `NextToken`, so if the `try_number > 1` gate is ever dropped this test hangs rather than fails.

kaxil (approved):

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Fixed, both reconnect checks in `_find_previous_job_run` now call `self.is_job_active(state)` instead of the old hardcoded values, so WAITING and STOPPING are treated as reconnectable there too, matching `is_job_active` and the docs. Added parametrized coverage for both states on both tiers.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Moved the push into `_set_job_run_id`, which already runs on every path (fresh submit, reconnect, alreadysucceeded, bootstrap scan). Added test coverage on the reconnect and already-succeeded paths.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Fixed, the store check is now bypassed specifically on the deferrable path (`self.deferrable or not self._has_stored_external_id(...)`), not the whole scan condition, so a store read error there no longer blocks reattachment. Logged the persist beforedefer redesign as a follow-up rather than building it now, it needs the mixin's reconnect decision step exposed separately from its polling, which is a task-sdk change, not a quick fix here.

amoghrajesh: I'll track it here: https://github.com/apache/airflow/issues/71485

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Thanks a lot for this comment! It turned out to affect every operator in the family, not just Glue needed the fix. Added `durable: bool | None = None` as a real named parameter on each glue in this PR & raising a follow up for the rest

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Fixed, and KPO had the identical bug which I am fixing too.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: I figured that I can just make it behave like earlier for airflow versions less than 3.3, and thats what I did so that this overhead is not a surprise for sonmeone upgrading their providers. 

- On <3.3, the scan is back to being purely opt-in, same as main before this PR ever existed.
- On 3.3+, it behaves as durability is on by default

Opened #71489 to track the underlying issues (unbounded pagination, silent swallow, broad except Exception, missing IAM doc)

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Reworded both the docstring and glue docs to say "every Airflow 3 release" clears XComs, not just 3.3+. Leaving the XCom tier in place for now (not dropping it), and leaving those two tests as-is since they're still valid for 2.11, just not reachable on 3.x which the reworded docs now say plainly.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Fixed `_build_script_args` and `submit_job` now skip the tag/scan on Airflow 3.3+ synchronous runs entirely, since task_state_store is the sole reconnect mechanism there. Trades away the narrow crash-before-persist safety net in exchange for never touching argv by default and I accepted that trade since its a rare, self-healing race vs. an immediate breakage risk for strict-parsing scripts.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Moved the STOPPED as failure statement out of the "Durable execution" section entirely and stated plainly that `durable=False` does not change it, since the raise fires unconditionally. 

While tracing this I found the deferrable verbose path (a separate hand-rolled loop in the trigger) still treats STOPPED as success, disagreeing with both the waiter and our fix - filed #71490 for that, out of scope here.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Good catch, fixed.

providers/amazon/docs/operators/glue.rst
amoghrajesh: Replaced "resubmits by default" to "would otherwise resubmit."

providers/amazon/docs/operators/glue.rst
amoghrajesh: Sure, it now mentions both `airflow state-store clean` and `airflow db clean` explicitly

providers/amazon/src/airflow/providers/amazon/aws/exceptions.py
amoghrajesh: Dropped "reconnected" from the docstring.

providers/amazon/tests/unit/amazon/aws/operators/test_glue.py
amoghrajesh: Added the missing `get_job_runs.assert_not_called()` assertion, same pattern as the sibling test.

amoghrajesh:
Last round of review handled in 20b6b7550f

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Wanted to provide an update on this one. I thought more abnout this and went a step further than the original fix: 
Below Airflow 3.3, durable's value is now fully inert, not just correctly forwarded via default_args, but ignored outright even if explicitly set, with a warning. 
Below 3.3, only `resume_glue_job_on_retry` controls behavior there, exactly as it did before this port existed. 

Wanted to flag this since it's a further behavior change in the same area you had already reviewed.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
amoghrajesh: Following up since the design moved since this was marked resolved. 

The warning still fires unconditionally on every version, that part has not changed. I realised from user lens and `durable` set directly no longer does anything at all below 3.3 (previously a functional drop-in, which was the premise here, which didnt feel right because old users should not be aware of this flag), only `resume_glue_job_on_retry` activates the old XCom/scan mechanism there now. 

So the message itself is version-aware, below 3.3 it says "On Airflow 3.3+, use durable instead" instead of telling that user to switch to something currently inert for them and glue.rst picked up the same nuance for the same reason. The warning and its removal timeline (tied to the provider's minimum Airflow version, not the caller's) still apply to every version, per your original ask.

amoghrajesh:
Wanted to clear up some deprecation rules, and its handled now in: 8895b4c4955919b6eef8135495a580e7583f3760

amoghrajesh:
Thanks for the review @kaxil! Merging this one in.

PR #71212, opened 2026-08-06
Convert check-sdk-imports to capture occurences than files

amoghrajesh:
---

The hook was meant to stop airflow-core from gaining new runtime dependencies on the task SDK, but it worked by excluding whole files that already had an import when the check was added. That made it blind to *new* imports added later to any of those files - which is exactly what happened in #70370, where a new `airflow.sdk` import landed
in `models/taskinstance.py` unnoticed because the file was already on the exclude list for an older, unrelated import.

Track an allowed import count per file instead (same pattern as `check_new_airflow_exception_usage.py`), so existing  file can't quietly gain more without either a `# noqa: SDK001` or a deliberate allowlist regeneration.

---


generated/known_sdk_imports_in_core.txt
```diff
@@ -0,0 +1,91 @@
...
+airflow-core/src/airflow/api_fastapi/execution_api/versions/v2026_04_06.py::1
+airflow-core/src/airflow/cli/commands/task_command.py::7
+airflow-core/src/airflow/cli/commands/triggerer_command.py::1
+airflow-core/src/airflow/configuration.py::1
+airflow-core/src/airflow/dag_processing/dagbag.py::1
+airflow-core/src/airflow/dag_processing/importers/base.py::1
+airflow-core/src/airflow/dag_processing/importers/python_importer.py::7
+airflow-core/src/airflow/dag_processing/manager.py::4
+airflow-core/src/airflow/dag_processing/processor.py::15
+airflow-core/src/airflow/example_dags/example_asset_alias.py::1
```
jason810496: Would it be better to whitelist the `airflow-core/src/airflow/example_dags`?

jason810496 (commented):
Thanks! Yeah, it makes sense to follow the `know_....txt` convention.

scripts/ci/prek/check_sdk_imports_in_core.py
```diff
@@ -42,23 +78,75 @@ def check_file_for_sdk_imports(file_path: Path) -> list[tuple[int, str]]:
...
+        "--generate",
+        action="store_true",
+        help="Regenerate the allowlist from the current codebase and exit",
+    )
+    args = parser.parse_args(argv)
+
+    manager = SdkImportsAllowlistManager(REPO_ROOT / "generated" / "known_sdk_imports_in_core.txt")
+
+    if args.generate:
+        return manager.generate()
```
ferruzzi: Non-blocking nit: `generate()` will rebuild the whole thing, maybe an optional filename to update would be worth adding at some point?  This has a chance of accidentally approving a change you didn't intend to whereas requiring the filename limits that... but it also adds friction, so I'm not sure.

scripts/ci/prek/check_sdk_imports_in_core.py
```diff
@@ -42,23 +78,75 @@ def check_file_for_sdk_imports(file_path: Path) -> list[tuple[int, str]]:
...
+            "New [bold]airflow.sdk[/bold] import detected in airflow-core.\n"
+            "Core (scheduler/API server) should not gain new runtime dependencies "
+            "on the Task SDK.\n"
+            "If this import is a genuine one-off, append `# noqa: SDK001` to the "
+            "import line.\n"
+            "If it's an intentional, broader exception, run:\n\n"
+            "  [cyan]uv run ./scripts/ci/prek/check_sdk_imports_in_core.py --generate[/cyan]\n\n"
+            "to regenerate the allowlist, then commit the updated\n"
+            "[cyan]generated/known_sdk_imports_in_core.txt[/cyan]."
+        )
```
ferruzzi, line 188: Non-blocking nit:  Should this message propose the NOQA option from the doctring above as an alternative to rebuilding the allowlist?  I'm not entirely sure on the intended difference between the two routes.

ferruzzi (commented):
Two thoughts, non-blocking.   Thanks for taking this on.

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: Yea sure. `--generate` now takes optional filenames: 

`--generate path/to/file.py` recomputes only those entries and leaves the rest of the allowlist alone. 
`--generate`only does the full rebuild, but that is now documented as the clean up sprint path rather than the one a contributor reaches for mid review, and the failure message points at the scoped form.

generated/known_sdk_imports_in_core.txt
amoghrajesh: Sure, `example_dags` are Dags authored with the task sdk so so they were never really "core" for this check. 

Stripped it entirely.

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: It already mentioned noqa, but you are right that it never said which route to pick, which is the actual question. Reworded the panel to lead with "drop the import if you can", then two numbered options: `# noqa: SDK001` for a single deliberate import line, which stays out of the count and out of the allowlist entirely, versus `--generate FILE` for a file that genuinely needs several, which raises that files recorded ceiling. Same distinction now in the module docstring.

generated/known_sdk_imports_in_core.txt
amoghrajesh: Handled in 766783d4bd

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: Handled in 766783d4bd

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: Handled in 766783d4bd

scripts/ci/prek/check_sdk_imports_in_core.py
```diff
@@ -42,23 +89,122 @@ def check_file_for_sdk_imports(file_path: Path) -> list[tuple[int, str]]:
...
-        nocheck_code=NOCHECK_CODE,
-        only_python_files=True,
+    def iter_files(self) -> Iterable[Path]:
+        return (path for path in CORE_SRC_ROOT.rglob("*.py") if not self.is_excluded(path))
+
+    def check(self, files: list[Path], allowlist: dict[str, int]) -> int:
+        return super().check([path for path in files if not self.is_excluded(path)], allowlist)
+
+    def count_occurrences(self, path: Path) -> int:
+        return len(check_file_for_sdk_imports(path))
```
kaxil, line 135: `find_import_violations` returns `[]` when the file fails to parse (it swallows `SyntaxError`, `OSError` and `UnicodeDecodeError`), so an unparseable file counts as 0. That was harmless before, but now 0 < allowed means `check()` reads it as a reduction and rewrites `generated/known_sdk_imports_in_core.txt`. I put a syntax error into `models/taskinstance.py` and ran the hook: it printed `taskinstance.py 1 -> 0`, deleted the entry from the allowlist on disk, and told me to stage the updated file. Could `count_occurrences` skip files that don't parse instead of counting them as zero?

generated/known_sdk_imports_in_core.txt
```diff
@@ -0,0 +1,44 @@
...
+airflow-core/src/airflow/dag_processing/processor.py::15
+airflow-core/src/airflow/exceptions.py::1
+airflow-core/src/airflow/executors/base_executor.py::3
+airflow-core/src/airflow/jobs/triggerer_job_runner.py::18
+airflow-core/src/airflow/models/__init__.py::5
+airflow-core/src/airflow/models/asset.py::2
+airflow-core/src/airflow/models/connection.py::5
+airflow-core/src/airflow/models/dag.py::1
+airflow-core/src/airflow/models/dagrun.py::2
+airflow-core/src/airflow/models/taskinstance.py::1
```
kaxil, line 19: Since the entry is only a count, a file sitting at its ceiling can still pick up a brand new `airflow.sdk` dependency as long as it drops another one. I swapped this file's `airflow.sdk.definitions._internal.abstractoperator` import for `airflow.sdk.execution_time.secrets_masker` and the hook passed with the count still at 1. Given that #70370 landing a new sdk import in this exact file is the motivating case, is recording the imported module per entry worth it over a bare count? I realise this mirrors `check_new_airflow_exception_usage.py`, so happy if the answer is "later".

kaxil (approved):

scripts/ci/prek/check_sdk_imports_in_core.py
ferruzzi: Re-reading my comment you are right, that was terrible phrasing on my part.  You got what I meant though, thanks for that.

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: Fixed. Added a `_can_parse()` check (tries `ast.parse`) and filter unparseable files out before they ever reach count_occurrences, in check(), generate_for(), and iter_files(). A file that fails to parse now gets skipped with a warning instead of silently counting as 0 and tightening/deleting its allowlist entry.

generated/known_sdk_imports_in_core.txt
amoghrajesh: Let me take a look at this in a follow up.

scripts/ci/prek/check_sdk_imports_in_core.py
amoghrajesh: Handled in 578c487eda

generated/known_sdk_imports_in_core.txt
amoghrajesh: Handled in https://github.com/apache/airflow/commit/578c487edaf68fd30859779975b90fc4c675df78

jason810496 (approved):

PR #71214, opened 2026-08-06
Store executor callback DagRun references in the database

fat-catTW:
## Why
Executor callbacks need to know which DagRun they belong to when the scheduler
builds callback workloads. Keeping that reference only in `callback.data` makes
the scheduler depend on payload shape instead of a database-managed relationship.

## Solution
Add a nullable `callback.dagrun_id` foreign key to `dag_run.id` and have executor
callbacks store their DagRun reference there. The scheduler now loads the
relationship through the ORM when enqueueing callback workloads.

The migration backfills existing rows from deadline relationships first, then
falls back to legacy callback payload data when needed.

---


airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -1268,24 +1269,11 @@ def _enqueue_executor_callbacks(self, session: Session) -> None:
...
-                dag_run = session.get(DagRun, dag_run_id)
-
+                dag_run = callback.dag_run
                 if dag_run is None:
                     self.log.warning(
                         "Could not find DagRun with id=%s for callback %s. DagRun may have been deleted.",
-                        dag_run_id,
+                        callback.dagrun_id,
                         callback.id,
                     )
```
uranusjr, line 1285: This collapses two previously distinct failure modes (missing `dagrun_id` due to a callback-creation bug vs. a DagRun deleted after enqueue) into one warning path. Only one production call site sets `dagrun_id` today, so it's not reachable now, but it'd be easy for a future ExecutorCallback creation path to forget to set it and silently emits "DagRun may have been deleted."

Maybe add different checks to `callback.dagrun_id is None` and `dag_run is None` with dagrun_id set?

fat-catTW: Good point, thank you for reviewing.
 
I split the checks so a missing `callback.dagrun_id` is reported separately from the case where `dagrun_id` is set but the related DagRun cannot be loaded.

PR #71216, opened 2026-08-06
Optimize scheduler backfill completion updates

fat-catTW:
## Why:

The scheduler checks whether backfills are complete as part of its regular loop. The previous implementation loaded every matching `Backfill` row into the ORM session and then updated each object in Python, even though the completion condition can be expressed entirely in SQL.

Using a bulk update keeps this scheduler path lighter while preserving the existing AIP-78 backfill lifecycle guards.

## Solution:

Update `_mark_backfills_complete()` to issue a single bulk `UPDATE` for backfills that are ready to complete.

The existing guards are preserved:
- initializing backfills without associations are skipped
- orphaned backfills older than the cutoff can still be completed
- backfills with queued or running Dag runs stay active

Add a regression test that verifies the completion path uses a single update statement.

---


steveahnahn (approved):
LGTM. The `WHERE` is byte-identical to the old `select`, so the AIP-78 guards are preserved and its a bulk update. The test coverage looks good too.

PR #71220, opened 2026-08-06
Add cluster-wide config to override per-pool include_deferred filed

msumit:
**Value prop**: As a centralised platform team, we don't want end users to give control of this field, especially when they are totally unaware of what `deferred` really means here. For them, pool slots are used to control the load on the downstream systems like Spark and Presto, and it doesn't matter whether that task is deferred or not. So with a cluster level config we'll take control of the behaviour. 

**Changes**: Introduced a new configuration option `core.pool_include_deferred`. When set to `True` or `False`, it fixes the effective value for every pool (including pre-existing pools) and prevents users from setting a conflicting per-pool value via the UI, API, or CLI. 

The default remains unset (empty string `""`), preserving the existing per-pool behavior.

Updated the pool model, local/FastAPI API serializers, UI config endpoint and pool form, CLI pool commands, default pool creation, and added corresponding tests.

- When config is set to `True`
Screenshot 2026-08-06 at 10 04 57 AM

- When config is set to `False`
Screenshot 2026-08-06 at 10 06 03 AM

- When someone tries to set an opposite value via API
Screenshot 2026-08-06 at 12 48 05 PM

- When API doesn't contain any info regarding to this flag
Screenshot 2026-08-06 at 12 48 37 PM

 

---

- [x] Yes - Generated by Claude

---


eladkal:
cc @ramitkataria is working on a proposal to enchance the pool mechanisem. Ramit when will the AIP published? We need to make sure pool related features are alligned with the plan

airflow-core/docs/administration-and-deployment/pools.rst
```diff
@@ -46,6 +46,12 @@ descendants.
 Note that if tasks are not given a pool, they are assigned to a default pool ``default_pool``, which is
 initialized with 128 slots and can be modified through the UI or CLI (but cannot be removed).
 
+Whether deferred tasks occupy pool slots is normally decided per pool via its ``include_deferred`` flag.
```
Prajwal7842, line 49: [Discussion] Instead of this should the behavior be: Deployment manager is allowed to set the default value (True/False) for all the pools created under a team or globally and let individual pool owners override this ?

airflow-core/src/airflow/models/pool.py
```diff
@@ -92,6 +92,26 @@ class Pool(Base):
     def __repr__(self):
         return str(self.pool)
 
+    @staticmethod
+    def get_include_deferred_override() -> bool | None:
```
Prajwal7842, line 96: What happens if the deployment manager unsets this value ? Pool table will have include_deffered=True (which was populated due to global setting). 
Now if admin makes it false, the original config set by the end user will not be honored

airflow-core/src/airflow/models/pool.py
```diff
@@ -92,6 +92,26 @@ class Pool(Base):
...
+        precedence over the per-pool ``include_deferred`` column. Returns None when unset.
+        """
+        from airflow.configuration import conf
+
+        if conf.get("core", "pool_include_deferred", fallback=""):
+            return conf.getboolean("core", "pool_include_deferred")
+        return None
+
+    @property
+    def effective_include_deferred(self) -> bool:
```
Prajwal7842, line 110: Should we use this only for scheduling decisions ? Let the original value get stored in DB (whatever the end user provided) and take AND of both the values when making a task scheduling decision.

airflow-core/src/airflow/models/pool.py
```diff
@@ -143,6 +163,10 @@ def create_or_update_pool(
                 "team_name cannot be set when multi_team mode is disabled. Please contact your administrator."
             )
 
+        include_deferred_override = Pool.get_include_deferred_override()
+        if include_deferred_override is not None:
+            include_deferred = include_deferred_override
```
Prajwal7842, line 172: This will permanently override the global config to pool configs right, making the rollback not possible ?

airflow-core/src/airflow/ui/src/pages/Pools/PoolForm.tsx
```diff
@@ -141,11 +147,22 @@ const PoolForm = ({ error, initialPool, isPending, manageMutate, setError }: Poo
         control={control}
         name="include_deferred"
         render={({ field }) => (
-          <Field.Root mb={4} mt={4}>
+          <Field.Root disabled={includeDeferredOverride !== undefined} mb={4} mt={4}>
```
Prajwal7842, line 149: If the global config is set, should we just skip showing the option to users to configure whether to include deferred slots or not

airflow-core/docs/administration-and-deployment/pools.rst
msumit: Nop, we do not want users to override it, and the decision whether to count deferrable slots like normal slots needs to remain with the cluster admins only.

airflow-core/src/airflow/models/pool.py
msumit: Again, the control lies with the admin only 

I see 3 main approaches here: 

1. We let the users provide the value of their choice, store in the DB as well, but don't honor it if global config is set to True or False. This also creates an issue of accepting the payload but not honoring it. 
2. We don't allow users to pass any value if global config is set (true or false), but if the admin unsets the global conf, all pools will have a None value, which is equivalent to false only, again not ideal. 
3. We don't allow users to pass any value if global config is set, and set the value of global config in DB against the pool. When admin changes the conf to empty, those values will be honored. 

The third approach looks better and less confusing than 1 & 2.

airflow-core/src/airflow/ui/src/pages/Pools/PoolForm.tsx
msumit: the option is diabled

airflow-core/src/airflow/models/pool.py
msumit: This is option 1 in my comment above, but I think option 3 is better and less confusing for users

airflow-core/src/airflow/models/pool.py
msumit: This is what option 3 in my previous comment

airflow-core/src/airflow/models/pool.py
Prajwal7842: Can you elaborate more on option (3). 
Suppose the global config was not set, and some users have configured values as true or false. 
Now admin sets the value, values set by users are overriden by the admin value and scheduling decisions are made on that. 
Admin unset the values to null --> user value should get honoured now. 
By your change, will this behavior be enforced ? Or are you suggesting something alternate ?

airflow-core/src/airflow/models/pool.py
Prajwal7842: We can use a slight variation of option 1 where we accept values until the global config is not set. Once set, the UI option is disabled and shows the global value. Now later on, if admin decides to unset the value, the existing/old values given by the users start honoring again and new users get option to choose/update their value. WDYT ?

airflow-core/src/airflow/models/pool.py
Prajwal7842: Is this extensible to team based config ? For example, I want to enable the config for team A, but disable it for team B and keep it unset for team C ?

airflow-core/src/airflow/models/pool.py
msumit: yeah, but it means we've to keep option open in UI, API, and CLI, accepting the values but not honoring it eventually. It's better to not accept it and model their usage around the value set by the admin only. Also, there won't be any right or wrong behaviour is possible, ideally I don't expect admins to play around with this setting, for e.g. it'll always be set to true in our usecase.

airflow-core/src/airflow/models/pool.py
msumit: Making it team wise would be another big change and doing all this in a single PR won't be good. I believe if there are some companies who wanted to use Team level config, they can extend it and contribute.

PR #71223, opened 2026-08-06
Stop Dependabot from raising Go SDK PRs against v3-3-test

jason810496:
- related: https://github.com/apache/airflow/pull/71020#pullrequestreview-4872489614
- related: #70914, #70915

## Why

Dependabot opened #71020 bumping a Go SDK dependency against `v3-3-test`. In review, jason810496 flagged that we shouldn't raise any Dependabot PRs for the Go SDK against release-test branches at all, since the Go SDK only ever releases from `main`.

## What

- Rework the `gomod`/`/go-sdk` Dependabot block targeting `v3-3-test` to match the default-branch block's security-only pattern (`open-pull-requests-limit: 0`, security-updates-only group), instead of raising routine version-update PRs.

Once this merges, #71020 should be closed manually — this config change doesn't retroactively close a PR Dependabot already opened.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -371,26 +371,6 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
 
-  - package-ecosystem: "gomod"
-    cooldown:
-      # Longer than other ecosystems: distributed Go releases are not pre-scanned by a
-      # central registry, so we give any malicious release more time to be flagged.
-      default-days: 14
-    directory: "/go-sdk"
-    schedule:
-      interval: "weekly"
-    target-branch: v3-3-test
-    groups:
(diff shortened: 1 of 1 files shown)


Lee-W (commented):
I'm a bit confused and would like to check how go-sdk and other sdk work with patch versions. Thanks!

jason810496:
> I'm a bit confused and would like to check how go-sdk and other sdk work with patch versions. Thanks!

Currently, only the Go SDK enable the depedentbot (Java, TS haven't but will). We restricted the Go SDK depedentbot focusing on the security advisories (not bump the version for minor or patch) for `main`.

Actually, I should make the v3-3-test one focus on security advisories only instead of dropping it.

Lee-W (approved):

PR #71235, opened 2026-08-06
Add config option for the default task instance tab in the UI

dsuhinin:
Task instances always open on the Logs tab, and users who prefer another tab have to click it every time — there is no way to change the default.                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                            
  This adds a new `[api] default_task_instance_tab` config option that lets a deployment set which tab opens by default when a task instance is first opened (e.g. `details`). The index route reads the value from `/ui/config` and redirects to the configured tab, preserving the query string and hash; unknown or unset values fall back to `logs`, so existing behavior and deep links are unchanged. Explicit tab choices made while navigating task-to-task in the grid still take precedence.


bbovenzi:
I thought about your PRs again.

First, if we do this, we should make these localStorage settings instead of deployment configs and they could go in Pierre's new settings page: [#70687](https://github.com/apache/airflow/pull/70687)

Second, to preserve any links. Let's not mess with the Urls themselves but just redirect to a different page by default.

Finally, being powered by the config isn't great when loading a brand new page when the config hasn't been fetched yet. The screen will have to show one page, and then start loading a different one.

dsuhinin:
> I thought about your PRs again.
> 
> First, if we do this, we should make these localStorage settings instead of deployment configs and they could go in Pierre's new settings page: [#70687](https://github.com/apache/airflow/pull/70687)
> 
> Second, to preserve any links. Let's not mess with the Urls themselves but just redirect to a different page by default.
> 
> Finally, being powered by the config isn't great when loading a brand new page when the config hasn't been fetched yet. The screen will have to show one page, and then start loading a different one.

thank you for your feedback. yep, in case of [#70687](https://github.com/apache/airflow/pull/70687) I can re use this PR and just add more options. Probably in that case I can move:
- https://github.com/apache/airflow/pull/71235
- https://github.com/apache/airflow/pull/70754
to be based on this PR and instead of `instance` config store everything in the same way in `localStorage`. What do you think?

bbovenzi:
Yes I think that makes sense. And we will still need to make sure that existing url paths are respected so we don't break any links.

dsuhinin:
> Yes I think that makes sense. And we will still need to make sure that existing url paths are respected so we don't break any links.

hey @bbovenzi . Can you please take a look when you will have a chance -> https://github.com/apache/airflow/pull/71699

PR #71237, opened 2026-08-06
Add terminationGracePeriodSeconds support for PgBouncer in Helm chart

antruigon:
The chart ships a default PgBouncer preStop hook that drains client connections for up to 120 seconds (`killall -INT pgbouncer && sleep 120`), but the Deployment never sets `terminationGracePeriodSeconds`, so the Kubernetes default of 30s SIGKILLs the pod mid-drain on a node drain or eviction and cuts in-flight client connections.

Every other long-running component in the chart (scheduler, workers, triggerer, dag-processor, statsd, redis, otel collector) already exposes this value; PgBouncer was the only one missing it, despite holding the database connections of all the others.

This adds `pgbouncer.terminationGracePeriodSeconds`, defaulting to 120 to match the preStop drain window. PgBouncer exits as soon as the last client connection is released, so the value is an upper bound rather than a fixed wait. Deployments that prefer the previous behaviour can set it back to 30.

Covered by a parametrized test in `test_pgbouncer.py` (default and override).

---

- [X] Yes — Claude Code (Claude Fable 5)

Generated-by: Claude Code (Claude Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/chart/newsfragments/71237.significant.rst
+++ b/chart/newsfragments/71237.significant.rst
@@ -1,0 +1,5 @@
+PgBouncer Deployment now sets ``terminationGracePeriodSeconds`` (default ``120``)
+
+The PgBouncer Deployment previously relied on the Kubernetes default termination grace period of 30 seconds, which SIGKILLed the pod mid-drain: the chart's default preStop hook (``killall -INT pgbouncer && sleep 120``) needs up to 120 seconds to drain client connections, so on a node drain or eviction in-flight connections were cut after 30 seconds instead of being released.
+
+The new ``pgbouncer.terminationGracePeriodSeconds`` value defaults to ``120`` to match the preStop drain window. PgBouncer exits as soon as the last client connection is released, so the full 120 seconds is an upper bound, not a fixed wait. Set the value to ``30`` to restore the previous behaviour.
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -7873,6 +7873,13 @@
                         }
                     ]
                 },
+                "terminationGracePeriodSeconds": {
+                    "description": "Grace period for PgBouncer to finish after SIGTERM is sent from Kubernetes.",
+                    "type": "integer",
+                    "default": 120,
+                    "minimum": 0,
+                    "x-docsSection": "Kubernetes"
+                },
                 "securityContexts": {
                     "description": "Security context definition for the PgBouncer.",
                     "type": "object",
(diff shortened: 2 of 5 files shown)


chart/newsfragments/71237.significant.rst
```diff
@@ -0,0 +1,5 @@
+PgBouncer Deployment now sets ``terminationGracePeriodSeconds`` (default ``120``)
```
Miretpl: I think that this change does not need the newsfragment. Could you delete it?

chart/values.schema.json
```diff
@@ -7801,6 +7801,12 @@
                         }
                     ]
                 },
+                "terminationGracePeriodSeconds": {
+                    "description": "Grace period for PgBouncer to finish after SIGTERM is sent from Kubernetes. The default matches the default preStop hook, which needs up to 120 seconds to drain client connections.",
```
```suggestion
                    "description": "Grace period for PgBouncer to finish after SIGTERM is sent from Kubernetes.",
```
It is more justification for why we have default on 120 and it is rather not needed on the parameter documentation level.

chart/values.schema.json
```diff
@@ -7801,6 +7801,12 @@
...
+                    "type": "integer",
+                    "default": 120,
```
Miretpl, line 7879:
```suggestion
                    "default": 120,
                    "minimum": 0
```
for better validation.

Miretpl (changes requested):

Miretpl:
Could you please match the PR description with the project template and follow https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions?

chart/newsfragments/71237.significant.rst
antruigon: Removed, thanks

chart/values.schema.json
antruigon: Applied. The rationale for the 120 default now lives as a comment in `values.yaml`

chart/values.schema.json
antruigon: Applied, thanks

antruigon:
Done, sorry for the miss. The description now follows the project template and includes the Gen-AI disclosure. The change was AI-assisted and reviewed by me. I validated the rendered chart and ran the helm tests locally (test_pgbouncer.py, test_pdb_pgbouncer.py, test_chart_quality.py) before pushing

Miretpl (approved):

PR #71241, opened 2026-08-06
Refuse breeze start dev mode when UI development ports are already in use

Andrushika:
## Why

I was using Git worktrees and accidentally ran `breeze start-airflow --dev-mode` from a second worktree. Breeze showed no warning, but the browser only showed a blank page. 
Root cause: the old Vite process still owned port 5173 or 5174, so the new Vite process exited silently because of `--strictPort` while Breeze continued starting Airflow and the browser kept loading assets from the old Vite server.

## What

Check both ports before starting the background UI dev hook, and fail with a clear message when either port is in use. This only applies to dev asset compilation; regular UI asset builds are unchanged.

This PR only handles port conflicts that already exist when `run_compile_ui_assets` starts. `compile_ui_assets_dev.py` still does not keep the `Popen` handles, so it cannot detect a Vite process that exits later. I plan to handle process supervision in a follow-up PR.

---

- [X] Yes — Codex (GPT-5) and Claude Code (Fable 5)

Generated-by: Codex (GPT-5) and Claude Code (Fable 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


potiuk (commented):
Good for now - follow up would be great

potiuk (approved):

PR #71244, opened 2026-08-06
Add KubernetesPodExecOperator for existing Kubernetes Pods

AlejandroMorgante:
Add `KubernetesPodExecOperator` to execute commands in a running container of an existing Kubernetes Pod without creating, restarting, or deleting it. This supports pre-warmed or externally managed Pods where avoiding startup latency is important.

The operator validates the target Pod and container, streams stdout and stderr, waits for the command exit code, and reliably closes the exec connection. Optional stdout XCom is disabled by default and protected by a configurable size limit to avoid retaining unbounded output in worker memory.

Provider metadata, RBAC guidance, operator documentation, unit tests, and an end-to-end system test are included.

Tests:

- KubernetesPodExecOperator unit tests (`39 passed`)
- KubernetesPodExecOperator system test (`1 passed`)
- Targeted provider mypy and Ruff checks
- Provider metadata and documentation validation

---

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/providers/cncf/kubernetes/docs/operators.rst
+++ b/providers/cncf/kubernetes/docs/operators.rst
@@ -28,18 +28,11 @@ The :class:`~airflow.providers.cncf.kubernetes.operators.pod_exec.KubernetesPodE
 executes a command in a running container of an existing Kubernetes Pod. It does not create,
 restart, or delete the target Pod.
 
-.. code-block:: python
-
-    from airflow.providers.cncf.kubernetes.operators.pod_exec import KubernetesPodExecOperator
-
-    run_command = KubernetesPodExecOperator(
-        task_id="run_command",
-        pod_name="existing-worker",
-        namespace="default",
-        container_name="worker",
-        command=["python", "-m", "worker.run_job"],
-        kubernetes_conn_id="kubernetes_default",
-    )
+.. exampleinclude:: /../tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
+    :language: python
+    :dedent: 4
+    :start-after: [START howto_operator_k8s_pod_exec]
+    :end-before: [END howto_operator_k8s_pod_exec]
 
 Commands are executed directly rather than through a shell. Include a shell explicitly when using
 pipes, redirects, variable expansion, or other shell features.
--- a/providers/cncf/kubernetes/docs/operators.rst
+++ b/providers/cncf/kubernetes/docs/operators.rst
@@ -19,36 +19,6 @@
 .. contents:: Table of Contents
     :depth: 2
 
-.. _howto/operator:KubernetesPodExecOperator:
-
-KubernetesPodExecOperator
-=========================
-
-The :class:`~airflow.providers.cncf.kubernetes.operators.pod_exec.KubernetesPodExecOperator`
-executes a command in a running container of an existing Kubernetes Pod. It does not create,
-restart, or delete the target Pod.
-
-.. exampleinclude:: /../tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
-    :language: python
-    :dedent: 4
-    :start-after: [START howto_operator_k8s_pod_exec]
-    :end-before: [END howto_operator_k8s_pod_exec]
-
-Commands are executed directly rather than through a shell. Include a shell explicitly when using
-pipes, redirects, variable expansion, or other shell features.
-Standard output and standard error are streamed to the task log. Set ``do_xcom_push=True`` to
-also return standard output through XCom.
-
-The target Pod and container must already be running. When ``container_name`` is omitted, the
-operator uses the ``kubectl.kubernetes.io/default-container`` annotation when present, or the
-first container otherwise. API-visible static Pods are supported through their mirror Pod name;
-components that are not exposed by the Kubernetes API cannot be targeted. The Kubernetes connection
-requires ``get`` access to ``pods`` and ``pods/exec``; see :doc:`kubernetes_rbac`.
-
-If the task or its worker stops while the command is running, Airflow closes the exec connection
-but does not modify the target Pod. Kubernetes cannot always determine whether a command completed
-before a connection failure, so configure task retries only when the command is safe to repeat.
-
 .. _howto/operator:kubernetespodoperator:
 
 KubernetesPodOperator
@@ -445,6 +415,38 @@ For further information, look at:
 
 * `Kubernetes Documentation <https://kubernetes.io/docs/home/>`__
 * `Pull an Image from a Private Registry <https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/>`__
+
+.. _howto/operator:KubernetesPodExecOperator:
+
+KubernetesPodExecOperator
+=========================
+
+The :class:`~airflow.providers.cncf.kubernetes.operators.pod_exec.KubernetesPodExecOperator`
+executes a command in a running container of an existing Kubernetes Pod. It does not create,
+restart, or delete the target Pod.
+
+.. exampleinclude:: /../tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
+    :language: python
+    :dedent: 4
+    :start-after: [START howto_operator_k8s_pod_exec]
+    :end-before: [END howto_operator_k8s_pod_exec]
+
+Commands are executed directly rather than through a shell. Include a shell explicitly when using
+pipes, redirects, variable expansion, or other shell features.
+Standard output and standard error are streamed to the task log. Set ``do_xcom_push=True`` to also
+return standard output through XCom. Captured output is limited to 49,344 UTF-8 bytes by default;
+use ``max_xcom_output_size`` to configure a different finite limit. The task fails instead of
+returning truncated output when the limit is exceeded.
+
+The target Pod and container must already be running. When ``container_name`` is omitted, the
+operator uses the ``kubectl.kubernetes.io/default-container`` annotation when present, or the
+first container otherwise. API-visible static Pods are supported through their mirror Pod name;
+components that are not exposed by the Kubernetes API cannot be targeted. The Kubernetes connection
+requires ``get`` access to ``pods`` and ``pods/exec``; see :doc:`kubernetes_rbac`.
+
+If the task or its worker stops while the command is running, Airflow closes the exec connection
+but does not modify the target Pod. Kubernetes cannot always determine whether a command completed
+before a connection failure, so configure task retries only when the command is safe to repeat.
 
 SparkKubernetesOperator
 ==========================
--- a/providers/cncf/kubernetes/docs/operators.rst
+++ b/providers/cncf/kubernetes/docs/operators.rst
@@ -19,6 +19,43 @@
 .. contents:: Table of Contents
     :depth: 2
 
+.. _howto/operator:KubernetesPodExecOperator:
+
+KubernetesPodExecOperator
+=========================
+
+The :class:`~airflow.providers.cncf.kubernetes.operators.pod_exec.KubernetesPodExecOperator`
+executes a command in a running container of an existing Kubernetes Pod. It does not create,
+restart, or delete the target Pod.
+
+.. code-block:: python
+
+    from airflow.providers.cncf.kubernetes.operators.pod_exec import KubernetesPodExecOperator
+
+    run_command = KubernetesPodExecOperator(
+        task_id="run_command",
+        pod_name="existing-worker",
+        namespace="default",
+        container_name="worker",
+        command=["python", "-m", "worker.run_job"],
+        kubernetes_conn_id="kubernetes_default",
+    )
+
+Commands are executed directly rather than through a shell. Include a shell explicitly when using
+pipes, redirects, variable expansion, or other shell features.
+Standard output and standard error are streamed to the task log. Set ``do_xcom_push=True`` to
+also return standard output through XCom.
+
+The target Pod and container must already be running. When ``container_name`` is omitted, the
+operator uses the ``kubectl.kubernetes.io/default-container`` annotation when present, or the
+first container otherwise. API-visible static Pods are supported through their mirror Pod name;
+components that are not exposed by the Kubernetes API cannot be targeted. The Kubernetes connection
+requires ``get`` access to ``pods`` and ``pods/exec``; see :doc:`kubernetes_rbac`.
+
+If the task or its worker stops while the command is running, Airflow closes the exec connection
+but does not modify the target Pod. Kubernetes cannot always determine whether a command completed
+before a connection failure, so configure task retries only when the command is safe to repeat.
+
 .. _howto/operator:kubernetespodoperator:
 
 KubernetesPodOperator
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
@@ -39,7 +39,3 @@ class KubernetesApiError(AirflowException):
 
 class KubernetesApiPermissionError(AirflowException):
     """Raised when an error is encountered while trying access Kubernetes API."""
-
-
-class PodExecException(AirflowException):
-    """Raised when a command cannot be executed successfully in a Kubernetes pod."""
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
@@ -39,3 +39,7 @@ class KubernetesApiError(AirflowException):
 
 class KubernetesApiPermissionError(AirflowException):
     """Raised when an error is encountered while trying access Kubernetes API."""
+
+
+class PodExecException(AirflowException):
+    """Raised when a command cannot be executed successfully in a Kubernetes pod."""
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
@@ -1,0 +1,247 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Execute commands in existing Kubernetes pods."""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from functools import cached_property
+from typing import TYPE_CHECKING
+
+from kubernetes.client.rest import ApiException
+from kubernetes.stream import stream as kubernetes_stream
+
+from airflow.providers.cncf.kubernetes.exceptions import PodExecException
+from airflow.providers.cncf.kubernetes.hooks.kubernetes import KubernetesHook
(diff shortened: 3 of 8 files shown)


kaxil:
> Add `KubernetesPodExecOperator` to execute commands in a running container of an existing Kubernetes Pod without creating, restarting, or deleting it. This supports pre-warmed and API-visible static Pods where avoiding startup latency is important.
> 
> The operator validates the target Pod and container, streams stdout and stderr, supports optional stdout XCom, reports non-zero exit codes, and closes the exec connection when the task is killed. Provider metadata, RBAC guidance, and operator documentation are included.
> 
> Tests:
> 
> * `pytest providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod_exec.py -q` (32 passed)
> * Provider mypy check
> * Provider metadata and documentation validation
> * Pre-commit and manual checks
> 
> ##### Was generative AI tooling used to co-author this PR?
> * [x]  Yes — Codex (GPT-5)
> 
> Generated-by: Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

What's the User story and real-world use cases for this?

AlejandroMorgante:
@kaxil 
> What's the User story and real-world use cases for this?

Thanks for asking. The user story is:

> As an Airflow user who already manages a long-running Kubernetes Pod, I want to execute commands in one
  of its containers as Airflow tasks, without Airflow creating, restarting, or deleting that Pod.

Our concrete use case is running dbt on EKS. Today, our workflow starts a new Pod, initializes the dbt project and its dependencies, and then executes a command such as `dbt run`. A significant part of the execution time is spent scheduling the Pod and initializing the environment.

We want to manage that Pod separately and keep it warm, then use Airflow to execute ad hoc commands such as `dbt run`, `dbt test`, or `dbt build`. This avoids the repeated cold-start and initialization cost.

More generally, this is useful whenever a Pod’s lifecycle is managed outside the task and its environment is already initialized. The operator exposes the equivalent of `kubectl exec` as an Airflow task, including Airflow logging and optional XCom output, without assuming ownership of the Pod lifecycle.

AlejandroMorgante:
@SameerMesiah97 Could you please review it when you have a chance? Thank you!

SameerMesiah97 (commented):
I am 50/50 on this new operator for 3 main reasons:

1) I think this new operator diverges from the existing operators in the sense that it is trying to orchestrate externally managed resources that have their own lifecycle and control plane. This introduces some unclear lifecycle and ownership semantics. For example, what happens if the Pod/container is restarted or replaced while the task is executing, or if the exec connection is lost and the Airflow task retries? Could the original command still be running while the retry starts another exec?

2) Should we be introducing a new operator i.e. `KubernetesPodExecOperator` or could we consider supporting externally managed Pods as an execution mode of `KubernetesPodOperator` instead? I can see an argument for keeping them separate because KPO owns the Pod lifecycle whereas this operator would only own the exec lifecycle, but both ultimately represent executing an Airflow task inside a Kubernetes container. I think it would be useful to establish why a separate public operator is preferable before introducing another Kubernetes execution abstraction.

3) I have to build on @kaxil's reservations regarding potential use cases by questioning whether an operator strictly for `kubectl exec` commands is the right abstraction. What about fetching logs or waiting on/monitoring an externally managed Pod (this may be a senor or trigger; this needs discussion too) ? If the broader use case is for Airflow to interact with Pods whose lifecycle it does not own, then `exec` feels like only one specific operation within that model. Perhaps, a broader operator that handles several commands may be better. 

These points could be addresesd in the description. For an atypical operator in a high blast radius provider, I think we need to really nail the rationale.

AlejandroMorgante:
Thanks for the feedback. Airflow commonly acts as a client of infrastructure it does not own: `SSHOperator` does not manage the server, SQL operators do not manage the database, and `SsmRunCommandOperator` does not manage the target instances. They own only the operation being executed. This operator follows the same model: Airflow owns the exec session, while another system owns the Pod lifecycle.

Although this can be implemented with `BashOperator` and `kubectl exec`, that moves additional integration logic to every user: installing and maintaining `kubectl` on the workers, distributing kubeconfig and cloud authentication helpers, handling shell quoting, and mapping Kubernetes errors and output into Airflow. This operator provides that behavior through `KubernetesHook` and Airflow connections. The provider already uses the `pods/exec` API internally, so this is not a new Kubernetes mechanism.

Keeping this provider-neutral also allows GKE- or EKS-specific operators to reuse it and add only their cloud authentication and cluster configuration, following the existing provider layering. I prefer keeping it separate from KPO because KPO owns the complete Pod lifecycle, while this operator deliberately treats the Pod as externally managed.

If the Pod disappears, restarts, or the connection fails, the task fails. A retry creates a new exec session; Airflow never creates, replaces, or deletes the Pod.

SameerMesiah97:
> Although this can be implemented with `BashOperator` and `kubectl exec`, that moves additional integration logic to every user: installing and maintaining `kubectl` on the workers, distributing kubeconfig and cloud authentication helpers, handling shell quoting, and mapping Kubernetes errors and output into Airflow. This operator provides that behavior through `KubernetesHook` and Airflow connections. The provider already uses the `pods/exec` API internally, so this is not a new Kubernetes mechanism.

I think this is the strongest argument here. If using this operator saves time and effort for users who want this specific functionality, then I believe it is justified. But like I mentioned above, why only `exec`? Why not other commands like `log` or `watch`? It seems premature and restrictive to narrow it to just `exec`?

AlejandroMorgante:
Combining `exec`, `logs`, and `watch` would give one operator several different responsibilities and completion contracts.

`pods/exec` already allows this operator to start any command available inside the container, stream its stdout and stderr, and wait for its exit code. In contrast, `pods/log` only consumes a log stream, while `watch` observes resource state and requires a condition to determine when the task should finish.

This also differs from `KubernetesPodOperator`: KPO creates a Pod and configures the workload as the container's main process. It then monitors the Pod lifecycle.`KubernetesPodExecOperator` starts a secondary process through `pods/exec`, waits only for that process, and leaves the existing Pod running.

AlejandroMorgante:
@SameerMesiah97 Could you help me with a code review?

I’ve already tested it with the end-to-end system test and an Apache Magpie review.

Thank you! 🙏

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
```diff
@@ -39,3 +39,7 @@ class KubernetesApiError(AirflowException):
 
 class KubernetesApiPermissionError(AirflowException):
     """Raised when an error is encountered while trying access Kubernetes API."""
+
+
+class PodExecException(AirflowException):
```
SameerMesiah97: Do we need a dedicated `PodExecException` here? It appears to be functionally equivalent to `AirflowException`, which we are generally moving away from using for operator execution failures. Could these instead use the appropriate built-in exception types?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _close_exec_client(self) -> None:
+        exec_client = self._exec_client
+        self._exec_client = None
+        if exec_client is None:
+            return
+        try:
+            exec_client.close()
+        except Exception:
+            self.log.exception("Failed to close Kubernetes exec connection")
```
SameerMesiah97: Could this log include the namespace, pod name, and container name? The exception traceback is retained, but `"Failed to close Kubernetes exec connection"` alone does not identify which target failed when a worker is running multiple tasks.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+            return self.container_name
+
+        annotations = pod.metadata.annotations if pod.metadata and pod.metadata.annotations else {}
+        default_container = annotations.get("kubectl.kubernetes.io/default-container")
+        if isinstance(default_container, str) and default_container in container_names:
+            return default_container
+        return container_names[0]
+
+    def _validate_container_is_running(self, pod: V1Pod, container_name: str) -> None:
+        if not pod.status or pod.status.phase != "Running":
```
SameerMesiah97: Could this use `PodPhase.RUNNING` from pod_manager instead of the literal `"Running"`?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+        while exec_client.is_open():
+            exec_client.update(timeout=1)
+            while exec_client.peek_stdout():
+                output = exec_client.read_stdout()
+                self._log_output(output, stream_name="stdout")
+                if self.do_xcom_push:
+                    stdout_chunks.append(output)
+            while exec_client.peek_stderr():
+                self._log_output(exec_client.read_stderr(), stream_name="stderr")
+        return "".join(stdout_chunks)
```
SameerMesiah97, line 216: When `do_xcom_push=True`, this buffers the command’s complete stdout in worker memory without any size limit, and `"".join(stdout_chunks) `then allocates the combined string as well. Since the command can emit arbitrary output, could this exhaust worker memory before we even reach the XCom size/storage constraints? Should the operator enforce a configurable limit or otherwise avoid retaining unbounded output in memory?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+        if (
+            container_status is None
+            or container_status.state is None
+            or container_status.state.running is None
+        ):
+            raise PodExecException(f"Container {container_name!r} in pod {self.pod_name!r} is not running")
+
+    def _log_output(self, output: str, *, stream_name: str) -> None:
+        log_method = self.log.warning if stream_name == "stderr" else self.log.info
+        for line in output.splitlines():
```
SameerMesiah97: Are values returned by `read_stdout()` and `read_stderr(`) guaranteed to contain complete lines? If these are arbitrary WebSocket chunks, `_log_output()` can split one logical line across multiple log records. Should incomplete lines be buffered between reads?

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+        container_status = next((status for status in statuses if status.name == container_name), None)
+        if (
+            container_status is None
+            or container_status.state is None
+            or container_status.state.running is None
+        ):
+            raise PodExecException(f"Container {container_name!r} in pod {self.pod_name!r} is not running")
+
+    def _log_output(self, output: str, *, stream_name: str) -> None:
+        log_method = self.log.warning if stream_name == "stderr" else self.log.info
```
SameerMesiah97: Should all stderr output be emitted at warning level? Many commands routinely use stderr for progress or informational messages, so this could generate misleading warning records. Logging both streams at `INFO` while preserving the [stderr] marker may be more appropriate.

providers/cncf/kubernetes/docs/operators.rst
```diff
@@ -19,6 +19,36 @@
...
+The target Pod and container must already be running. When ``container_name`` is omitted, the
+operator uses the ``kubectl.kubernetes.io/default-container`` annotation when present, or the
+first container otherwise. API-visible static Pods are supported through their mirror Pod name;
+components that are not exposed by the Kubernetes API cannot be targeted. The Kubernetes connection
+requires ``get`` access to ``pods`` and ``pods/exec``; see :doc:`kubernetes_rbac`.
+
+If the task or its worker stops while the command is running, Airflow closes the exec connection
+but does not modify the target Pod. Kubernetes cannot always determine whether a command completed
+before a connection failure, so configure task retries only when the command is safe to repeat.
+
```
SameerMesiah97: I think this should be moved after the section for `KubernetesPodOperator`.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
```diff
@@ -0,0 +1,247 @@
...
+
+        statuses = pod.status.container_statuses or []
+        container_status = next((status for status in statuses if status.name == container_name), None)
+        if (
+            container_status is None
+            or container_status.state is None
+            or container_status.state.running is None
+        ):
+            raise PodExecException(f"Container {container_name!r} in pod {self.pod_name!r} is not running")
+
```
SameerMesiah97, line 181: I think this log message is a bit too presumptive. No container status or a missing state does not necessarily mean the container is not running. You could do something like this instead:

`RuntimeError(f"Container {container_name!r} in pod {self.pod_name!r} is not running or container status cannot  be retrieved.")`

providers/cncf/kubernetes/tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
```diff
@@ -0,0 +1,134 @@
...
+            if error.status != 404:
+                raise
+        else:
+            last_phase = pod.status.phase if pod.status else None
+            container_statuses = pod.status.container_statuses if pod.status else None
+            container_status = next(
+                (status for status in container_statuses or [] if status.name == CONTAINER_NAME), None
+            )
+            if (
+                last_phase == "Running"
```
SameerMesiah97: Use `PodPhase.RUNNING` here too.

providers/cncf/kubernetes/tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
```diff
@@ -0,0 +1,134 @@
...
+  restartPolicy: Never
+  containers:
+    - name: {CONTAINER_NAME}
+      image: busybox:1.38.0
+      command: ["sleep", "3600"]
+"""
+
+
+@task
+def wait_for_running_pod() -> None:
```
SameerMesiah97: Is there an existing sensor or some other abstraction you can use instead of creating your own polling loop? This looks quite heavy for an example DAG.

providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod_exec.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def test_close_error_does_not_mask_task_shutdown(self):
+        operator, _ = create_operator()
+        exec_client = mock.MagicMock(spec=WSClient)
+        exec_client.close.side_effect = RuntimeError("connection already closed")
+        operator._exec_client = exec_client
+
+        operator.on_kill()
+
+        assert operator._exec_client is None
```
SameerMesiah97, line 374: Could we add a test where `_consume_output()`/`exec_client.update() `raises after the connection has been assigned? The finally block is intended to guarantee cleanup during execution failures, but the current tests only cover normal completion, unsuccessful return codes, and direct `on_kill() `calls.

providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod_exec.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def test_on_kill_without_active_connection(self):
+        operator, _ = create_operator()
+
+        operator.on_kill()
+
+        assert operator._exec_client is None
+
+    def test_close_error_does_not_mask_task_shutdown(self):
```
SameerMesiah97: It would be useful  to assert that the suppressed exception is logged, since logging is the only resulting behaviour.

AlejandroMorgante: Done, thanks! The test now asserts that the suppressed close exception is logged.

providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod_exec.py
AlejandroMorgante: Done, thanks! Added a test that makes exec_client.update() raise after the connection is assigned and verifies that the connection is closed and the internal references are cleared

providers/cncf/kubernetes/tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
AlejandroMorgante: Done, thanks! Replaced the manual polling loop with a rescheduling @task.sensor.

providers/cncf/kubernetes/tests/system/cncf/kubernetes/example_kubernetes_pod_exec.py
AlejandroMorgante: Done!

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Done!

providers/cncf/kubernetes/docs/operators.rst
AlejandroMorgante: Done!

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Done, thank you!

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Yes, done! Thanks

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Done!

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/exceptions.py
AlejandroMorgante: Done, thanks! Replaced the dedicated exception with appropriate built-in exception types

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Good point. WebSocket reads are not guaranteed to contain complete lines. I added independent buffers for stdout and stderr, so only complete newline-delimited records are logged, and any remaining partial line is flushed when the session ends. The tests now cover lines split across chunks and unterminated trailing output.

This follows the buffering approach already used by `DockerOperator.fetch_logs`

https://github.com/apache/airflow/blob/f760ac1c03b7277f3cb69071b8ead56d27ecbff7/providers/docker/src/airflow/providers/docker/operators/docker.py#L60-L77.

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod_exec.py
AlejandroMorgante: Good point. I made XCom capture opt-in (do_xcom_push=False by default) and added a configurable max_xcom_output_size, enforced incrementally while stdout is streamed.

It defaults to 49,344 UTF-8 bytes, following the existing MAX_XCOM_SIZE precedent used by `GCSToLocalFilesystemOperator`
https://github.com/apache/airflow/blob/f760ac1c03b7277f3cb69071b8ead56d27ecbff7/providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_local.py#L29-L30

and `GoogleApiToS3Operator`
https://github.com/apache/airflow/blob/f760ac1c03b7277f3cb69071b8ead56d27ecbff7/providers/amazon/src/airflow/providers/amazon/aws/transfers/google_api_to_s3.py#L36-L38

If the limit is exceeded, the task fails instead of retaining unbounded output or silently truncating it, and the exec connection is still cleaned up. I also added tests for invalid limits and multibyte output exceeding the configured byte limit.

SameerMesiah97 (approved):
Looks good.

vincbeck (approved):

Issue #67477, opened 2026-05-25
Calendar view does not respect user-selected timezone


pyzo-ryzo:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.2.1

### What happened and how to reproduce it?

The Calendar view on the DAG detail page does not respect the user's selected timezone from the UI timezone selector. All dates — both in the API query parameters and in the rendered calendar grid — are computed using either browser-local time or hardcoded UTC, while the rest of the UI (e.g. the "Start Date" column on the Runs tab) correctly converts timestamps to the user-selected timezone.
 
**Observed behavior:** Changing the timezone via the UI timezone selector has no effect on the Calendar view. Runs are grouped into the wrong day/hour cells for users whose selected timezone differs from UTC.
 
**Comparison with Runs tab:** On the Runs tab, the `Start Date` column properly updates when the user changes their timezone setting, confirming the timezone selector itself works correctly.

## How to reproduce
 
1. Set your Airflow UI timezone to something significantly offset from UTC (e.g. `Asia/Tokyo` — UTC+9, or `America/Los_Angeles` — UTC-7).
2. Navigate to a DAG that has runs near midnight UTC.
3. Open the **Calendar** tab → observe that runs are grouped by their UTC date/hour, not by the selected timezone.
4. Switch to the **Runs** tab → observe that the `Start Date` column correctly shows times in the selected timezone.
## Root cause analysis
 
The entire Calendar feature tree (`Calendar.tsx`, `calendarUtils.ts`, `HourlyCalendarView.tsx`, `DailyCalendarView.tsx`) was written using plain `dayjs()` without any awareness of the `useTimezone()` / `TimezoneContext` system that the rest of the Airflow UI uses. The `dayjs/plugin/timezone` and `dayjs/plugin/utc` plugins are not even imported in any of these files.
 
There are three specific problem areas:
 
### 1. `Calendar.tsx` — API query dates hardcode a literal "Z" suffix
 
```typescript
// Current code (Calendar.tsx)
const gte = startDate.format("YYYY-MM-DD[T]HH:mm:ss[Z]");
const lte = endDate.format("YYYY-MM-DD[T]HH:mm:ss[Z]");
```
 
`[Z]` in a dayjs format string is a **literal character escape** — it unconditionally appends the letter `Z` (meaning "UTC") to whatever the browser-local time value is. This tells the API the timestamp is UTC when it is not. The component never calls `useTimezone()`.
 
[...]

PR #71248, opened 2026-08-06
fix(calendar): localize cron start_time to timetable timezone before computing planned runs

### LINKED ISSUE #71234: Calendar view computes planned runs for cron timetables in UTC, ignoring the DAG's configured timezone

Airflow Core

### Apache Airflow version

3.3.0

  For a DAG using a plain cron-string `schedule` (e.g. `"0 8 * * *"`) on an Airflow instance whose `[core] default_timezone` is **not UTC** (e.g. `Asia/Seoul`, UTC+9), the **Calendar view's "planned" (future, not-yet-run) cells are shown at the wrong hour** — offset by the UTC/local difference.

  Concretely, with `AIRFLOW__CORE__DEFAULT_TIMEZONE=Asia/Seoul` and a DAG scheduled `"0 8 * * *"`:
  - The DAG actually fires at **08:00 KST**, confirmed by `airflow dags next-execution` and by the DAG detail page's "Next Run" / "Last Run" fields (both correctly show `08:00:00`).
  - Real, already-executed runs are shown correctly on the Calendar at the 08:00 row.
  - **Future "planned" cells for the same DAG are drawn at the 17:00 row instead** (08:00 + 9h) — i.e. exactly a UTC-vs-KST offset.

  Historical (already-happened) runs are unaffected because they're read straight from `DagRun` rows already computed correctly by the scheduler. Only the *projected/planned* rows — computed on the fly by the Calendar API — are wrong.

  How to reproduce:

  Via the UI:
  1. Deploy Airflow with `AIRFLOW__CORE__DEFAULT_TIMEZONE=Asia/Seoul` (any non-UTC zone works; easiest to see with a schedule hour close to local midnight).
  2. Create a DAG with `schedule="0 8 * * *"` (no explicit timetable timezone override — it inherits the core default).
  3. Let it run at least once, or just open the DAG's Calendar tab.
  4. Compare the DAG detail page's "Next Run"/"Last Run" (correctly `08:00`) against the Calendar tab's *planned* cells for the same DAG (shown at `17:00`).

  Minimal Python repro (no webserver needed) — confirms the bad value is computed by the service itself, not introduced or fixed anywhere downstream:

  ```python
  from croniter import croniter
  from datetime import datetime
  import pendulum
  from airflow.timetables.trigger import CronTriggerTimetable

  tt = CronTriggerTimetable("0 8 * * *", timezone="Asia/Seoul")
  last_end_utc = pendulum.datetime(2026, 8, 5, 23, 0, 0, tz="UTC")  # = 2026-08-06 08:00 KST

  # (A) what CalendarService._calculate_cron_planned_runs does:
  next_planned = next(croniter(tt._expression, start_time=last_end_utc, ret_type=datetime))
  print(next_planned, "->", next_planned.astimezone(pendulum.timezone("Asia/Seoul")))
  # 2026-08-06 08:00:00+00:00 -> 2026-08-06 17:00:00+09:00   (WRONG)

  # (B) what the real scheduler uses (CronMixin._get_next):
  next_real = tt._get_next(last_end_utc)
  print(next_real, "->", next_real.in_timezone("Asia/Seoul"))
  # 2026-08-06 23:00:00+00:00 -> 2026-08-07 08:00:00+09:00   (matches `airflow dags next-execution`)
  ```

  I also confirmed this end-to-end by calling `CalendarService.get_calendar_data()` (the exact function `GET /ui/calendar/{dag_id}` calls) against a live instance's real `DagRun` history for a `schedule="0 8 * * *"` DAG: the one real, already-executed run came back correct (08:00 KST), while every planned/future occurrence came back exactly 9 hours later. This is the value returned by the API itself — nothing downstream corrects it. (Screenshot attached below, from a fresh local demo instance — no relation to the code snippets above.)

Image

Planned/future run cells in the Calendar view should be computed in the DAG's configured timetable timezone, the same way the real scheduler computes actual runs, so they land in the same hour/day as the runs that will actually execute.

Root cause: `CalendarService._calculate_cron_planned_runs()` in `airflow-core/src/airflow/api_fastapi/core_api/services/ui/calendar.py`:

```python
def _calculate_cron_planned_runs(
    self, dag, last_data_interval, year, date_filter, granularity,
):
    """Calculate planned runs for cron-based timetables."""
    dates: dict[datetime, int] = collections.Counter()

    dates_iter: Iterator[datetime | None] = croniter(
        cast("CronMixin", dag.timetable)._expression,
        start_time=last_data_interval.end,
        ret_type=datetime,
    )
    ...
```

`last_data_interval.end` is a UTC-aware `datetime`. `croniter`'s documented contract (its README: "Be sure to init your croniter instance with a TZ aware datetime for this to work!") is that the caller localizes `start_time` to whatever timezone the cron fields should be matched against — croniter just reads the tzinfo off it (`self.tzinfo = start_time.tzinfo` in `set_current`). `_calculate_cron_planned_runs` never does that localization, so `"0 8 * * *"` gets matched against UTC wall-clock instead of `dag.timetable`'s configured `Asia/Seoul` (→ 08:00 UTC = 17:00 KST). This is croniter behaving exactly as documented — the bug is entirely on the caller's side.

Contrast with how the real scheduler computes the next run — `CronMixin._get_next()` in `airflow-core/src/airflow/timetables/_cron.py`, which does the localize/delocalize correctly:

```python
def _get_next(self, current: DateTime) -> DateTime:
    naive = make_naive(current, self._timezone)   # UTC -> local naive wall-clock
    ...                                             # croniter matches on local wall-clock
    return convert_to_utc(make_aware(scheduled, self._timezone))  # local -> UTC
```

Notably, the sibling function in the same file, `_calculate_timetable_planned_runs()` (used for non-cron/partitioned timetables), does this correctly via `dag.timetable.next_dagrun_info_v2(...)`. Only the cron-specific fast path bypasses it. Still present, unchanged, on `main` (diffed byte-for-byte against the installed 3.3.0 function body).

Related prior art (same class of bug, different surfaces):
- #67477 (fixed) — a different layer: the React frontend ignored the user's selected UI timezone. Since it's fixed, the frontend now faithfully renders whatever value the API sends — which is why this backend bug is visible end-to-end.
- #67717 (merged) — the same pattern (comparing UTC-parsed dates directly against a non-UTC timetable's dates) for `airflow dags clear`, fixed via a new `Timetable.resolve_day_bound()`.
- #63631 — added the correct, timezone-aware `next_dagrun_info_v2` path for non-cron/partitioned timetables in this same `calendar.py` file; the cron fast path was left behind.

Suggested fix:

Localize `start_time` into the timetable's own timezone before constructing `croniter` in `_calculate_cron_planned_runs` (mirroring `CronMixin._get_next`), converting each result back to UTC — reuses the same private-attribute access pattern (`cast("CronMixin", dag.timetable)._expression`) already used in that function, just also reading `._timezone`.

(Originally also considered routing cron timetables through the same `next_dagrun_info_v2`-based path as `_calculate_timetable_planned_runs`. Ruled out: that path's `catchup=False` branch anchors to `utcnow()`, which breaks the sequential "planned" projection when the last recorded run is far from real wall-clock time — verified this regresses 3 existing tests.)

Verified locally (`uv sync --package apache-airflow-core` + pytest on `test_calendar.py`): 19 existing tests still pass + 1 new regression test for a non-UTC cron timetable — 20/20 passing.

### Operating System

Not specific to an OS — reproduced on macOS (local venv) and confirmed against a Kubernetes/Helm-chart deployment (Debian-based image).

### Deployment

Virtualenv installation

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

Deterministic, not intermittent — occurs for any cron-scheduled DAG on a non-UTC `default_timezone`, every time the Calendar view computes planned/future runs.

### Code of Conduct

waterWang:
Closes #71234

## Problem

Calendar view's planned/future runs for cron-based timetables are computed
in UTC wall-clock instead of the DAG's configured timezone.  When a DAG
uses a non-UTC `default_timezone` (e.g. `Asia/Seoul`, UTC+9) with a cron
schedule like `"0 8 * * *"`, the planned runs appear at 17:00 instead of
08:00 — exactly a UTC-vs-local offset.

The root cause is in `CalendarService._calculate_cron_planned_runs()`:
`croniter` receives `last_data_interval.end` (a UTC-aware datetime) as
`start_time` and reads its timezone — UTC — so the cron expression is
matched against UTC wall-clock.  The real scheduler (`CronMixin._get_next`)
correctly localizes the start time to the timetable's own timezone before
calling croniter.

## Fix

1. Add imports for `convert_to_utc`, `make_aware`, `make_naive` from
   `airflow._shared.timezones.timezone` (same utilities used by
   `CronMixin._get_next`).

2. In `_calculate_cron_planned_runs`:
   - Extract the timetable's `_timezone` from the `CronMixin` cast.
   - Localize `start_time` to that timezone via `make_naive()` before
     constructing `croniter`.
   - Convert each result back to UTC via `convert_to_utc(make_aware(...))`.

This matches the exact pattern used by `CronMixin._get_next()` in
`airflow/timetables/_cron.py`.

## Testing

All 19 existing calendar tests continue to pass.  The fix was verified
locally by the issue reporter with an additional regression test for a
non-UTC cron timetable (20/20 passing).

## Related

- #71234 (original issue report with full reproduction steps)
- #67477 (fixed frontend layer; this completes the fix at the backend)
- #67717 (same pattern, fixed in `airflow dags clear`)
- #63631 (added timezone-aware `next_dagrun_info_v2` path for non-cron timetables)

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/services/ui/calendar.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/services/ui/calendar.py
@@ -28,6 +28,7 @@ from sqlalchemy.engine import Row
 from sqlalchemy.orm import InstrumentedAttribute, Session
 
 from airflow._shared.timezones import timezone
+from airflow._shared.timezones.timezone import convert_to_utc, make_aware, make_naive
 from airflow.api_fastapi.common.parameters import RangeFilter
 from airflow.api_fastapi.core_api.datamodels.ui.calendar import (
     CalendarDeadlineCollectionResponse,
@@ -215,21 +216,35 @@ class CalendarService:
         """Calculate planned runs for cron-based timetables."""
         dates: dict[datetime, int] = collections.Counter()
 
+        timetable = cast("CronMixin", dag.timetable)
+        tz = timetable._timezone
+
+        # Localize start_time into the timetable's own timezone before
+        # constructing croniter, matching the pattern used by
+        # CronMixin._get_next().  Without this, the cron expression is
+        # matched against UTC wall-clock, which produces wrong scheduled
+        # times for DAGs with a non-UTC default_timezone.
+        local_start = make_naive(last_data_interval.end, tz)
+
         dates_iter: Iterator[datetime | None] = croniter(
-            cast("CronMixin", dag.timetable)._expression,
-            start_time=last_data_interval.end,
+            timetable._expression,
+            start_time=local_start,
             ret_type=datetime,
         )
 
         for dt in dates_iter:
-            if dt is None or dt.year != year:
+            if dt is None:
                 break
-            if dag.end_date and dt > dag.end_date:
+            # Convert back to UTC-aware for comparison and storage.
+            dt_utc = convert_to_utc(make_aware(dt, tz))
+            if dt_utc.year != year:
                 break
-            if not self._is_date_in_range(dt, date_filter):
+            if dag.end_date and dt_utc > dag.end_date:
+                break
+            if not self._is_date_in_range(dt_utc, date_filter):
                 continue
 
-            dates[self._truncate_datetime_for_granularity(dt, granularity)] += 1
+            dates[self._truncate_datetime_for_granularity(dt_utc, granularity)] += 1
 
         return [
             CalendarTimeRangeResponse(date=dt, state="planned", count=count) for dt, count in dates.items()


kaxil:

PR #52748, opened 2025-07-02
Add DAG calendar view endpoint

SavingFrame:
Hello everyone!

This PR adds a new DAG calendar view endpoint that provides scheduling information in a calendar format. This endpoint returns the calendar that existed in Airflow 2.x but provides additional functionality. The new endpoint includes filtering by `logical_date` and the ability to choose granularity, allowing users to customize their calendar view based on specific time periods and detail levels.

## Changes in this PR:
- Create service file that processes historical DagRuns and presumes queued runs
- Add UI endpoint
- Update Swagger documentation
- Add tests for different options of this endpoint

## How was this patch tested?

- Added comprehensive unit tests for the calendar endpoint
- Tests cover various scenarios including date filtering, granularity options, and DAG selection
- All existing tests continue to pass
- Static code checks and pre-commit hooks pass

## API Details

**Endpoint:** `GET /ui/calendar/{dag_id}`

**Path Parameters:**
- `dag_id` (string, required): The DAG ID

**Query Parameters:**
- `granularity` (string, optional): Time granularity - `"hourly"` or `"daily"` (default: `"daily"`)
- `logical_date_gte` (datetime, optional): Start date for filtering (ISO 8601 format)
- `logical_date_lte` (datetime, optional): End date for filtering (ISO 8601 format)

**Response Format:**
Returns an array of `CalendarTimeRangeResult` objects with the following structure:

```json
[
  {
    "datetime": "2024-01-15 00:00:00",
    "state": "success",
    "count": 5
  },
  {
    "datetime": "2024-01-15 00:00:00",
    "state": "running",
    "count": 2
  },
  {
    "datetime": "2024-01-16 00:00:00",
    "state": "planned",
    "count": 7
  }
]
```

**Response Fields:**
- `datetime`: The timestamp for the calendar entry
- `state`: Task state - one of `"queued"`, `"running"`, `"success"`, `"failed"`, or `"planned"`
- `count`: Number of task instances in this state at the given time

## Related Links
- [[Discussion](https://github.com/apache/airflow/discussions/50907)](https://github.com/apache/airflow/discussions/50907)
- [[Reference PR](https://github.com/apache/airflow/pull/37988)](https://github.com/apache/airflow/pull/37988) that provided the main logic
[...]

PR #71263, opened 2026-08-07
Fix Calendar view hanging for Dags with high-frequency cron schedules

theJohnnyBrown:
The Calendar view's planned-runs computation for cron timetables iterates croniter until the year boundary with no cap, while the generic-timetable path has always stopped at `MAX_PLANNED_RUNS` (2000). The scan runs synchronously in the request handler on every calendar request, and the date-range filter does not bound it (filtered dates are skipped, not stopped at).

Measured cost of the unbounded loop per `GET /ui/calendar/{dag_id}` request:

| Schedule | Iterations to year end (from early Jan) | Time |
|---|---|---|
| `*/5 * * * *` | ~104k | ~2 s |
| `* * * * *` | ~520k | ~7 s |
| `* * * * * *` (seconds cron) | ~31.5M | ~5.5 min of CPU |

Any user with Dag read access triggers this just by opening the Calendar tab of such a Dag, which can pin an API server worker.

This applies the same `MAX_PLANNED_RUNS` cap the generic path has used since the endpoint was added (#52748), via `itertools.islice`. Planned cells for high-frequency crons are now shown up to 2000 runs ahead, consistent with what an equivalent `timedelta`-scheduled Dag already does.

The new test fails without the fix (the request returns ~300k planned runs after ~4.5 s) and passes with it; all existing calendar tests pass.

---


pierrejeambrun (approved):
Thanks for the PR.

LGTM, tested and working as expected.

As a follow up it would be great to:
- Maybe have a slightly higher limit than 2000 which seems low
- Have feedback in the UI that "the number of planned run is too high to predict to the end of the year and therefore it's the reason why it's cut off"

Dashboard page is doing something similar with (50000+ displayed when number of runs / TI is too big to compute)

Otherwise people might get confused why the schedule is stopping after a couple of days like such:

Image

But that's not specific to this PR, the other branch has the same problem I assume

PR #71270, opened 2026-08-07
Make copied task log text match the on-screen format

Andrushika:
## Why

Follow up PR for #71156. 
Copied logs should match what the user sees on the screen. The rebuilt virtualized rows used the raw download format, so one copy could mix two timestamp formats and drop the group marker.

## What

The copy rebuild now formats timestamps in the selected timezone, same as the screen. Group headers keep their marker, and it follows the expand state (`▶` / `▼`) in both native copy and rebuilt text. 
The marker is now a real character that switches with state (this change keeps our code cleaner in handling copy event), so the small rotate animation is gone.

https://github.com/user-attachments/assets/42058be2-ecad-48e5-90aa-a90f2bc1eaf2

related: #71156

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
+++ b/airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
@@ -322,7 +322,7 @@ const renderStructuredLogImpl = ({
       const stringifiedValue = val instanceof Object ? JSON.stringify(val) : val;
 
       if (renderingMode === "text") {
-        elements.push(` ${key === "logger" ? "source" : key}=${stringifiedValue}`);
+        elements.push(` ${key === "logger" ? "source" : key}=${stringifiedValue} `);
       } else {
         elements.push(
           <React.Fragment key={`space_${key}`}> </React.Fragment>,
--- a/airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
+++ b/airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
@@ -44,6 +44,8 @@ type ErrorDetail = {
 };
 
 type RenderStructuredLogProps = {
+  /** Text mode only: render the timestamp with this instead of emitting the raw source value. */
+  formatTimestamp?: (timestamp: string) => string;
   index: number;
   logLevelFilters?: Array<string>;
   logLink: string;
@@ -173,6 +175,7 @@ export const extractTIContext = (
 };
 
 const renderStructuredLogImpl = ({
+  formatTimestamp,
   index,
   logLevelFilters,
   logLink,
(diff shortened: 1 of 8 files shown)


Andrushika:
TBH... I don't really like the approach in this PR, but it is the minimal and fastest change for the target.

We intercept the copy event and rebuild the text ourselves. So the rebuild has to know about every visual element, like the `▶` marker and the timestamp format. If someone adds a new element to the log view one day, for example a badge in front of error lines or an inline link to the failed task, and forgets the copy path, it drifts again. The test in this PR compares rebuilt lines with the real copy text of mounted rows, so it catches the current elements, but the approach is still brittle by nature.

I think the better way is to make `renderStructuredLog` produce one neutral token list first, and then render the same tokens to JSX for the screen and to plain text for copy. The two outputs cannot drift because they come from the same source. But that is a big change and touches search, ANSI and i18n. So I would like to hear opinions from maintainers first. If that direction sounds right, I would like to merge this one as an interim fix and do the refactor in a follow-up PR.

airflow-core/src/airflow/ui/src/pages/TaskInstance/Logs/TaskLogContent.tsx
```diff
@@ -143,7 +149,7 @@ export const TaskLogContent = ({
         getRowText: (index) => {
           const entry = visibleItems[index]?.entry;
 
-          return entry ? getEntryText(entry) : "";
+          return entry ? getEntryText(entry, expandedGroups) : "";
```
bbovenzi, line 152: I think we can edit the timestamp here and not have to drill down the timezone props so much.

airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
```diff
@@ -216,9 +218,9 @@ const renderStructuredLogImpl = ({
     return "";
   }
 
-  if (Boolean(timestamp) && showTimestamp) {
+  if (timestamp !== undefined && Boolean(timestamp) && showTimestamp) {
     if (renderingMode === "text") {
-      elements.push(`[${timestamp}] `);
+      elements.push(`[${formatTimestamp?.(timestamp) ?? timestamp}] `);
```
bbovenzi: This can become `"" ?? timestamp` and won't always fallkback onto the raw timestamp. Let's

airflow-core/src/airflow/ui/src/pages/TaskInstance/Logs/TaskLogContent.tsx
Andrushika: You are right, in the latest code, I simply edit timestamp here instead of using timezone props across `useLogs` and `parseLogs`. Thanks for pointing that out!

airflow-core/src/airflow/ui/src/components/renderStructuredLog.tsx
Andrushika: Since the timezone props path has been removed and the original code no longer exists, no further changes are needed here.

bbovenzi (approved):

bbovenzi:
Ci failures appear unrelated.

PR #71276, opened 2026-08-07
Support dynamic metric name prefixes in the metrics registry check

aeroyorch:
The registry check turns each metric name into a template and looks it up in
`metrics_template.yaml`. When the variable comes first there's no static prefix to match
on, so the metric is flagged as missing even though every expansion is registered.

Those names now match on their static suffix, which must start with a dot so it only
aligns at segment boundaries. Nothing else changes.

Related to #69001 and #69007

---

Claude Opus 5 for code planning and testing.

---


aeroyorch:
cc @ephraimbuddy

aeroyorch:
@jason810496 this PR can be closed?

Issue #59120, opened 2025-12-05
Scheduler exception handling in dagrun create loop doesn't work


dstandish:
### Body

In SchedulerJobRunner, in _create_dag_runs, we do `except Exception: ... continue` but this does not work when the exception is the result of a db error; in that case you need to roll the transaction back.

to fix this we need either smaller transactions (e.g. one per dag run create attempt) or savepoints.  either way it's a bit of a refactor.

### Committer

- [x] I acknowledge that I am a maintainer/committer of the Apache Airflow project.

Arunodoy18:
HI, this seems intresting , i must take a look on this issue.
.take-issue.

justinpakzad:
Gonna give this one a go. Cheers. 

Edit: This one might be a bit out of reach for me (someone else can pick it up) as I think it would require a pretty big change. We currently use prohibit_commit which wouldn't allow us to do any sort of commits or savepoints from within the `_create_dag_runs` method , so the only way to patch the fix would be to move it out of the prohibit_commit block (I might be wrong though)

SEPURI-SAI-KRISHNA:
I looked into the savepoint approach mentioned earlier in this thread. The concern raised above — that `prohibit_commit` won't allow savepoints — is **correct**, and it's worth recording exactly why, plus what does work.

### Why `session.begin_nested()` is blocked

It isn't an Airflow quirk. `SessionTransaction._prepare_impl` in SQLAlchemy 2.0 dispatches the session-level event for nested transactions too:

```python
if self._parent is None or self.nested:
    self.session.dispatch.before_commit(self.session)
```

`CommitProhibitorGuard` listens on `before_commit`, so releasing a `SAVEPOINT` raises `UNEXPECTED COMMIT - THIS WILL BREAK HA LOCKS!` exactly as a real commit would. Verified against SQLAlchemy 2.0.51.

### Dropping to the connection doesn't help either

`session.connection().begin_nested()` sidesteps the Session event layer, but mixing connection-level savepoints with ORM flush leaves the Session in `PendingRollbackError` after the first failure. Dead end — noting it so nobody else spends time there.

### What does work: let the guard tolerate savepoint releases

```python
def _validate_commit(self, session):
    if session.in_nested_transaction():
        return          # savepoint release, not a real commit
    if self.expected_commit:
        self.expected_commit = False
        return
    raise RuntimeError("UNEXPECTED COMMIT - THIS WILL BREAK HA LOCKS!")
```

[...]

PR #71287, opened 2026-08-07
Isolate scheduled DagRun creation failures with savepoints

fat-catTW:
## Why

`SchedulerJobRunner._create_dag_runs` catches exceptions while creating scheduled DagRuns and then continues with the next Dag. This does not work reliably when the failure is caused by a database error, because the SQLAlchemy session can be left in a failed transaction state.

When that happens, one failing DagRun creation attempt can prevent the scheduler from creating DagRuns for other healthy Dags in the same batch.

## Solution

Wrap each scheduled DagRun creation attempt in a nested transaction/savepoint. If a database-level failure happens for one Dag, only that DagRun creation attempt is rolled back and the outer scheduler transaction remains usable.

The in-memory active run counter is updated only after the savepoint succeeds, keeping scheduler state aligned with the database transaction outcome.

A regression test covers a flush-time database failure for one Dag and verifies that another Dag in the same batch still creates its DagRun successfully.

relates: #59120

---

### DIFF
--- a/airflow-core/src/airflow/utils/sqlalchemy.py
+++ b/airflow-core/src/airflow/utils/sqlalchemy.py
@@ -627,6 +627,29 @@ def with_db_lock_timeout(session: Session, lock_timeout: int = 30) -> Generator[
             session.execute(text(f"SET SESSION innodb_lock_wait_timeout = {old_mysql_timeout}"))
 
 
+_EXPECTED_SAVEPOINT_COMMIT = "_airflow_expected_savepoint_commit"
+
+
+@contextlib.contextmanager
+def create_savepoint(session: Session) -> Generator[None, None, None]:
+    savepoint = session.begin_nested()
+    try:
+        yield
+    except Exception:
+        savepoint.rollback()
+        raise
+    else:
+        session.info[_EXPECTED_SAVEPOINT_COMMIT] = session.info.get(_EXPECTED_SAVEPOINT_COMMIT, 0) + 1
+        try:
+            savepoint.commit()
+        finally:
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/utils/sqlalchemy.py
```diff
@@ -635,10 +658,12 @@ class CommitProhibitorGuard:
     def __init__(self, session: Session):
         self.session = session
 
-    def _validate_commit(self, _):
+    def _validate_commit(self, session):
         if self.expected_commit:
             self.expected_commit = False
             return
+        if session.info.get(_EXPECTED_SAVEPOINT_COMMIT) and session.in_nested_transaction():
```
steveahnahn, line 666: _EXPECTED_SAVEPOINT_COMMIT is session-wide, so this exemption fires for any nested commit while it is set, not just the savepoint that set it. It's safe as written since it is only set across the synchronous savepoint.commit() on a single-threaded session but a one-line comment saying so would help, and it would need rethinking if nested savepoints ever land on this path.

steveahnahn (approved):
Nice work! Reverting the savepoint wrapping reproduces the poisoned-transaction failure and the fix clears it LGTM aside from the one nit

airflow-core/src/airflow/utils/sqlalchemy.py
fat-catTW: Thanks for the review.
I added a short comment to make it explicit that the session-wide marker is only set around the synchronous savepoint release.
Thanks

PR #71303, opened 2026-08-07
Fix tag-providers --dry-run failing to find the push remote

potiuk:
`breeze release-management tag-providers --dry-run` always aborted with:

```
ValueError: Could not find the remote configured to push to apache/airflow
```

A dry run short-circuits every `run_command`, returning an empty stdout, so
the `git remote get-url --push` lookup that picks the apache/airflow remote
found nothing for all three candidate names (`upstream`, `origin`, `apache`).

The lookup is read-only, so it now runs even in dry-run mode. That makes
`--dry-run` usable for previewing the tags a provider wave will create,
which is precisely when a release manager wants to check them.

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/release_management_commands.py
@@ -1502,6 +1502,9 @@ def tag_providers(
                 stdout=subprocess.PIPE,
                 stderr=subprocess.DEVNULL,
                 text=True,
+                # Read-only lookup, so run it even in dry-run mode: a dry run returns empty
+                # stdout for every remote, which would make the search below find nothing.
+                dry_run_override=False,
             )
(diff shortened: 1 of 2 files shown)


potiuk:
This has been waiting for a review for a couple of weeks. It fixes `tag-providers --dry-run` failing to resolve the push remote, so the dry run no longer errors out where the real run would have worked.

This is release-manager-facing and small — does either of you have a moment? (#71304 is a sibling change in the same file.)

---

potiuk:
@vatsrahul1001 @jason810496 — you both review a fair amount of the breeze and CI tooling; would either of you have a moment for this one?

It fixes `tag-providers --dry-run` failing to resolve the push remote, so the dry run no longer errors out in a case where the real run would have worked. Two files, release-manager-facing. #71304 is a sibling change in the same file.

---

PR #71304, opened 2026-08-07
Allow skipping the git fetch when preparing provider distributions

potiuk:
Adds `--skip-git-fetch` to
`breeze release-management prepare-provider-distributions`.

Building a provider wave in batches re-fetches the full history and every
tag from apache/airflow on each invocation. That is slow, and on an
unreliable network it is the step most likely to fail — taking a whole
batch of providers down with it, even though the tags fetched moments
earlier are still current.

Today the only way to avoid the fetch is `--skip-tag-check`, which also
gives up the tag check itself. The new flag keeps the tag check and
validates against the tags already in the local repository.

`classify-provider-changes` and `prepare-provider-documentation` already
offer the same escape hatch, so this closes a gap rather than introducing
a new concept.

---


potiuk:
This has been waiting for a review for a couple of weeks. It adds a flag to skip the git fetch when preparing provider distributions, which saves a lot of time when the checkout is already current.

This is release-manager-facing and small — does either of you have a moment? (#71303 is a sibling change in the same file.)

---

potiuk:
@vatsrahul1001 @jason810496 — same ask as on #71303, these two travel together.

This one adds a flag to skip the git fetch when preparing provider distributions, which saves a good deal of time when the checkout is already current.

---

vatsrahul1001 (approved):

PR #71305, opened 2026-08-07
Monitor the workflow run breeze actually dispatched

potiuk:
`breeze workflow-run publish-docs` dispatches three workflows in sequence —
`publish-docs-to-s3` on apache/airflow, then the apache/airflow-site
refresh, then the apache/airflow-site-archive S3-to-GitHub sync — and gates
each on the previous one finishing.

Two things broke that gate.

**It monitored whichever run was newest, not the one it dispatched.**
`gh run list --limit 1` was called right after the dispatch. When the
dispatch had not registered yet, that returns the *previous* run —
typically already finished — so the driver declared success and moved
straight on to the two follow-up workflows while the docs build it was
supposed to wait for was still running. A scheduled run, or a run another
maintainer started, was picked up the same way.

Run ids increase monotonically, so the id observed just before dispatching
is now recorded and the driver waits (up to 3 minutes) for a run above it.

**Only `failure` was treated as a failure.** A run that came back
`cancelled`, `timed_out`, or `action_required` fell through the terminal
branch without exiting, and the rest of the chain proceeded as if the docs
had been published. Any conclusion other than `success` now fails the
command.

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/utils/gh_workflow_utils.py
+++ b/dev/breeze/src/airflow_breeze/utils/gh_workflow_utils.py
@@ -96,13 +96,20 @@ def make_sure_gh_is_installed():
         sys.exit(1)
 
 
-def get_latest_workflow_run_id(workflow_name: str, repo: str) -> int | None:
+def get_latest_workflow_run_id(workflow_name: str, repo: str, *, exit_on_error: bool = True) -> int | None:
     """
-    Get the latest workflow run ID for a given workflow name and repository.
+    Get the latest dispatched workflow run ID for a given workflow name and repository.
+
+    Only ``workflow_dispatch`` runs are considered - several of the workflows we drive here also run
+    on push or pull request (apache/airflow-site's `build.yml`, for one), and such a run registering
+    between our two lookups would otherwise be mistaken for the one we dispatched.
 
     :param workflow_name: The name of the workflow to check.
     :param repo: The repository in the format 'owner/repo'.
-    :return: The run id, or None when the workflow has never run.
+    :param exit_on_error: Whether a failing `gh` call should terminate breeze. Pass False when the
+        caller polls and can afford to retry a transient failure.
+    :return: The run id, or None when the workflow has never been dispatched (or when the lookup
+        failed and ``exit_on_error`` is False).
     """
     make_sure_gh_is_installed()
     command = [
@@ -113,6 +120,8 @@ def get_latest_workflow_run_id(workflow_name: str, repo: str) -> int | None:
         workflow_name,
         "--repo",
         repo,
+        "--event",
+        "workflow_dispatch",
         "--limit",
         "1",
         "--json",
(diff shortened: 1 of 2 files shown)


potiuk:
Two files and green, waiting since 7 August. It makes breeze monitor the workflow run it actually dispatched, rather than whichever run it happened to find first.

@FrankYang0529 you have worked in this file before — would you mind taking a look?

---

dev/breeze/src/airflow_breeze/utils/gh_workflow_utils.py
```diff
@@ -122,16 +126,41 @@ def get_workflow_run_id(workflow_name: str, repo: str) -> int:
...
+    a scheduled run, or another maintainer's - whenever ours has not registered yet, and report
+    that run's result as ours.
+
+    :param workflow_name: The name of the workflow that was dispatched.
+    :param repo: The repository in the format 'owner/repo'.
+    :param previous_run_id: The newest run id seen before dispatching, or None if there was none.
+    """
+    deadline = time.monotonic() + NEW_RUN_TIMEOUT_SECONDS
+    while True:
+        run_id = get_latest_workflow_run_id(workflow_name, repo)
```
FrankYang0529: `get_latest_workflow_run_id` checks `returncode`. If there is temporary network failure, we need to rerun the workflow. Since we already have retry mechanism here, how about catch the exception and retry the command?

dev/breeze/src/airflow_breeze/utils/gh_workflow_utils.py
```diff
@@ -122,16 +126,41 @@ def get_workflow_run_id(workflow_name: str, repo: str) -> int:
...
+        if run_id is not None and (previous_run_id is None or run_id > previous_run_id):
```
FrankYang0529, line 163: `get_latest_workflow_run_id` returns the latest run id. However, airflow-site also runs build.yaml workflow if there is pull request or push to master. How about adding `--event workflow_dispatch` to `gh run list` command, so we can avoid the race condition.

https://github.com/apache/airflow-site/blob/7d6312026ff4cd4f611be7dad6980fd78b269667/.github/workflows/build.yml#L18-L24

FrankYang0529 (commented):
Overall LGTM. Thanks for the PR.

eladkal (approved):

potiuk:
Both addressed in 178b035 — thanks, these were good catches.

**Transient `gh` failures:** `get_latest_workflow_run_id` now takes `exit_on_error` (default `True`, so the pre-dispatch baseline lookup still fails fast). The polling loop in `wait_for_new_workflow_run` passes `False`, so a failed lookup logs a warning and retries until the deadline instead of aborting the publishing chain.

**Non-dispatch runs:** added `--event workflow_dispatch` to `gh run list` — you're right that airflow-site's `build.yml` also runs on push and pull_request, and such a run registering between the two lookups would have a higher id than the baseline and be mistaken for ours. Every caller of this path dispatches via `gh workflow run`, so the filter is safe across the board, and it composes with the id comparison rather than replacing it (id monotonicity still separates our dispatch from a concurrent manual one).

Rebased on latest main while I was at it.

---

PR #63499, opened 2026-03-13
feat(providers/amazon): allow disabling hook-level lineage in S3Hook …

shnhdan:
#### Description

When set to `False`, all `get_hook_lineage_collector()` calls in the hook are 
skipped, preventing asset spam during operations like multipart uploads.
Default behavior is unchanged.

**Changes:**
- `hooks/s3.py` — new param; all lineage calls across `load_file`, `load_string`,
  `load_bytes`, `load_file_obj`, `copy_object`, and `download_file` consistently gated
- `tests/test_s3.py` — tests for default, disabled, and explicit enabled behavior

**Note:** Complementary to #66992 — that PR controls whether the OL listener 
*reads* from `HookLineageCollector` at task end; this PR prevents the calls from 
firing at the source in `s3.py` entirely.

Closes #63371
---

##### Was generative AI tooling used to co-author this PR?
- [x] Yes (Claude)

Generated-by: Claude following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @shnhdan** · by `@potiuk` · 2026-06-22 06:31 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed (see the [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria)):
> - :x: **Unaddressed Copilot review** — a Copilot review thread has been open for ≥7 days with no reply; please respond to or resolve it. See the [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #71312, opened 2026-08-07
Allow filtering hook-collected assets in OpenLineage emission policy

### LINKED ISSUE #63371: Allow disabling hook-level lineage in Hook constructors
### Description

The S3Hook registers an output asset for every file which it uploads:
https://github.com/apache/airflow/blob/e491aac92a1d50f322a08d92d83616e7c79b3f2e/providers/amazon/src/airflow/providers/amazon/aws/hooks/s3.py#L1376-L1379

Which is not always the desired behaviour when using the S3 Hook (see motivating example below).

I'd propose just adding a switch to the S3Hook so disable this sort of lineage:
```python
hook = S3Hook(enable_hook_level_lineage=False)
```

I am happy to submit a fix here, but I wanted to run it by y'all first to make sure that I'm not missing some previous context or undoing an intentional design decision.

### Use case/motivation

We have seen issues where users upload chunked data to S3 within a PythonOperator like so:
```python
hook = S3Hook()
for idx, data in enumerate(list_of_values):
  hook.upload_string(data, f"some_prefix/file_{idx}.txt", "some_bucket")
```

Which then creates a _ton_ of output assets. I know that this is limited to 100 output objects (since https://github.com/apache/airflow/pull/45798), but it would be nice if we could disable hook-level lineage altogether and instead manage our own output asset definition at the custom operator / PythonOperator level. 

In this case, we likely want to only enable a single output asset at the `some_prefix/` level, not one per file. 

This likely applies to other object storage hooks as well.

### Related issues

_No response_

### Code of Conduct

ramitkataria:
Hook-level lineage registers an output asset for every object a hook writes, so a task that writes many objects reports one asset per object. The existing hook_lineage control is all-or-nothing: suppressing the noisy assets also loses the ones worth tracking from the same hook, and the collector cannot tell the two apart once its assets are read as one set.

Filtering by asset URI or reporting hook class lets a Dag keep the lineage that matters while dropping the noise, and doing it where hook lineage is consumed covers every instrumented hook at once rather than one provider at a time.

Note this reduces what gets reported rather than what gets collected, so it does not recover assets already dropped by max_assets_per_collector.

closes: #63371

cc @SamWheating @kacpermuda @ashb @vincbeck @shnhdan @eladkal @mobuchowski

 

---

---

### DIFF
--- a/providers/openlineage/docs/emission_policy.rst
+++ b/providers/openlineage/docs/emission_policy.rst
@@ -144,12 +144,38 @@ pattern.
        inputs/outputs. Task events only.
        **No effect when ``extract_operator_metadata`` is ``false``** — the entire extraction
        pipeline (including hook lineage) is skipped.
+   * - ``exclude_hook_lineage_assets``
+     - ``[]``
+     - List of regex patterns; hook-collected assets whose URI matches any of them are dropped.
+       Task events only. **No effect when ``hook_lineage`` is ``false``** — hook lineage is not
+       collected at all.
+   * - ``exclude_hook_lineage_hooks``
+     - ``[]``
+     - List of regex patterns; hook-collected assets reported by a hook whose **fully-qualified**
+       class name (``module.ClassName``, as with the ``operator`` scope key) matches any of them are
+       dropped. Task events only. **No effect when ``hook_lineage`` is ``false``**.
    * - ``include_full_task_info``
      - ``false``
      - Whether to include the full serialized operator state in ``AirflowRunFacet``. When
        ``false``, only a curated subset of task attributes is sent. When ``true``, all
        serializable task parameters are included, which may significantly increase event size.
        Task events only.
+
+Both exclude lists are always matched with ``re.fullmatch``, independent of the rule's
+``match_mode`` (which governs ``scope`` only). Unlike the boolean controls they are **replaced**
+rather than merged across tiers: the most specific matching rule that sets a list wins outright,
+so a task-scoped rule can narrow — or clear, with ``[]`` — a broader global exclusion. Both
+filters apply to hook-collected assets only; SQL-based hook lineage is unaffected.
+
+.. note::
+
+    **Exclusion happens after the collector cap.** These filters run when OpenLineage reads the
+    hook lineage collector, which is after
+    :ref:`[lineage] max_assets_per_collector<config:lineage__max_assets_per_collector>` has
+    already limited what the collector retained. Excluding assets reduces what gets reported, but
+    it does not create room for assets the cap already discarded — a task that writes more objects
+    than the cap allows never collected the later ones in the first place. Reducing the number of
+    assets a hook registers requires changes in the hooks themselves.
 
 ``locked`` (top-level, default ``false``) is an admin-only floor lock: when ``true``, the
 control fields carried by this rule's ``controls`` dict cannot be overridden by per-Dag /
@@ -193,6 +219,7 @@ suppresses DAG-level events for X while leaving task events enabled.
       {"scope": {"dag_id": "expensive_dag"}, "controls": {"extract_operator_metadata": false}},
       {"scope": {"dag_id": "expensive_dag", "task_id": "send_report"}, "controls": {"extract_operator_metadata": true}},
       {"scope": {"dag_id": "my_dag", "task_id": "sensitive_task"}, "controls": {"hook_lineage": false}},
+      {"scope": {"dag_id": "chunked_dag", "task_id": "upload"}, "controls": {"exclude_hook_lineage_assets": ["s3://my-bucket/staging/part_.*"]}},
       {"scope": {"dag_id": "reporting_dag"}, "controls": {"emit_dag_events": false}},
       {"scope": {"dag_id": "full_control_dag"}, "controls": {"emit": false}},
       {"scope": {"dag_id": "full_control_dag", "task_id": "critical_task"}, "controls": {"emit": true}},
@@ -212,6 +239,8 @@ In the example above:
   with a more-specific task-scoped rule and extracts normally.
 - ``sensitive_task`` in ``my_dag`` runs its extractor normally but skips the
   ``HookLineageCollector`` fallback.
+- ``upload`` in ``chunked_dag`` keeps hook lineage but drops the staging part objects it writes,
+  so only the assets that are not staging parts are reported.
 - DAG-level events for ``reporting_dag`` are suppressed; task events are still emitted.
 - Both task events and DAG-level events for ``full_control_dag`` are suppressed — except
   ``critical_task``, which re-enables task event emission via a task-scoped rule.
@@ -252,6 +281,11 @@ wins **unless** the matching conf rule is marked with ``"locked": true``.
 
     # Suppress DAG-run events but keep task events
     extend_global_openlineage_emission_policy(dag, emit_dag_events=False)
+
+    # Keep hook lineage, but drop the per-chunk objects a loop uploads
+    extend_global_openlineage_emission_policy(
+        extract, exclude_hook_lineage_assets=["s3://my-bucket/staging/part_.*"]
+    )
 
 Key semantics:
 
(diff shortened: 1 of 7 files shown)


mobuchowski:
In general I like this approach, but maybe this should be configurable more broadly:

- not narrowed down to hook-level lineage, but operator lineage too
- more options than just regex
- going even forward with it, maybe it could be just generalizable "preprocessor" for this data?

This could take a form of generalized filters (python function) that are being passed context and lineage and return lineage output.

We could have ones that are build in, and one of those could be "filter hook level lineage using regex"; but we could pass more using Airflow plugins if needed.

mobuchowski:
@kacpermuda wdyt?

kacpermuda:
I like the idea of widening the support of emission policy, that's for sure. Few thoughts:
1. Not sure what should be a scope and what can be a control from this PR. Hook class may be the new scope (where we already have exact and regex modes), not control. If mixed with dag_id, or dag_id and task_id, it can provide the granularity that is needed (disable hook level lineage inside dag or just for specific task but just from specific hook).

> not narrowed down to hook-level lineage, but operator lineage too

3. Not sure if there is a need to drop only some datasets (regex based) only when they come from hook level lineage, and not if they come from operator methods. We do have some [filters](https://openlineage.io/docs/client/python/configuration#filters) on python client level, but they are looking purely at job name now and are dropping events. If we wanted to, we can extend them to drop datasets that do not match some conditions - that would be a wide dataset filter. If we need to differentiate between hll and operator datasets, we might need to do it in controls, but then again, we might just use hook as scope and then exclude_datasets as control, that can be used both on operator and hook level? Thinking out loud here, not sure how to make it less confusing for people. 


> going even forward with it, maybe it could be just generalizable "preprocessor" for this data? This could take a form of generalized filters (python function) that are being passed context and lineage and return lineage output.

4. I like the idea of generic preprocessors, but I know that people rarely want to write them and we already have a transform transport that can be used to achieve whatever custom logic they have, so I don't think we need to provide more ways of customizing it. If you need more customization than the emission_policy provides, use transform transport and write your own transformer - would be my advice.

> more options than just regex

5. Not sure if this is needed now, for scope we have exact and regex, and for more custom stuff you can always use transform transport I think.

@ramitkataria , have you considered some alternatives when it comes to what is control and what is scope here?

ramitkataria:
Thanks both! @kacpermuda @mobuchowski 

Yes, I weighed a few alternatives, though I'm new to this code so please correct me where I've misread it:

  - Producer-side gating in the hooks (the #63499 approach). ~32 call sites across 5 files, and can't go on core `BaseHook` without dropping airflow 2.11. Filtering at collector read covers every instrumented hook at once.
  - Nesting: `{"hook_lineage": {"exclude_datasets": [...]}}`, so `hook_lineage` takes `bool | dict`. Probably the cleanest answer to control-vs-scope. I picked siblings thinking it kept the bool path clean, which was wrong: any non-bool control value widens `dict[str, bool]` in all 8 spots either way. Same cost, happy to switch.
  - Hook-class only (Kacper's earlier hook-scope idea). Doesn't seem to solve the issue case: the chunk parts and the objects worth keeping come from the same `S3Hook`, so it drops both. Hence asset patterns.
  - `match_mode` for these lists. Built it and then reverted it: two rules with different lists leave no single mode to record. So always regex, `match_mode` stays scope-only. This might be a good data point for Kacper's 5th point.

`hook` as a scope key: doesn't fit as far as I can tell. `_classify_task_rules` tiers on operator/dag_id/task_id only, one task event can span several hooks, and resolution returns one `EmissionPolicy`. Hook is only known per-asset, at collector read. Happy to be wrong here if you see a way. Note `{"dag_id": x, "task_id": y}` plus `exclude_hook_lineage_hooks` already gives the dag/task/hook combination you described.

`exclude_datasets` covering operator lineage: I agree and see two options:

  - (a) `exclude_datasets` plus one hook dimension, keeps the chunked-upload case working
  - (b) `exclude_datasets` only, hook granularity via `hook_lineage: false` at task scope

  Agree on skipping preprocessors and extra match modes.

Also:
  - these lists replace across tiers rather than merge, so a task rule can narrow or clear (`[]`) a global one. Merging would make a global rule unescapable locally. Worth confirming that's the behaviour you'd expect.
  - filtering runs after `max_assets_per_collector`, so it cuts what's reported, not what's collected. Anything past the cap was never collected and no pattern brings it back. Changing that would be producer-side work, so a bigger change.

kacpermuda:
Hmm, I don't have a good answer now, will need to think about it a bit more. Going ooo for a week, but will definitely come back to it after I'm back. 

I think we are ok with mixing hook as a scope, we can then disable something globally where we know it comes from hook, but also make it only per dag or task if needed, so all combinations make sense. `exclude_datasets` could also make sense:
- if we scope it to dag+ task - we drop all datasets from task
- if for task and dag+task+hook - we drop only hook level lineage
- if we scope for hook only - we drop them globally

Does that sound right?

ramitkataria:
Yes, that makes sense to me, and I think it covers everything this PR does: the chunked-upload case would be `{"scope": {"dag_id": ..., "task_id": ..., "hook": ".*S3Hook"}, "controls": {"exclude_datasets": [...]}}`, and `exclude_hook_lineage_hooks` becomes hook-only scope with `[".*"]`. I was wrong that hook can't be a scope: it can't join the up-front per-task resolution, but I could resolve dag/task/operator eagerly as now, carry any surviving hook-scoped rules into the collector read, and match each asset's `context` class there. Does that implementation shape sound reasonable, or did you have something else in mind?

I can rework the PR to this but I was wondering what your thoughts are on these before I do it:

  1. Should hook scope be limited to per-asset controls, i.e. reject rules combining it with `emit` etc. during validation (like the existing operator + `emit_dag_events` check)? And would you want `hook_lineage: false` allowed at hook scope as a nicer spelling of "drop everything from this hook"?
  2. What should `exclude_datasets` patterns match against? Hook assets have Airflow URIs pre-translation while operator lineage is OL datasets (namespace + name), so should I match the translated OL identity everywhere so one pattern behaves the same across sources?
  3. Should it also filter manually annotated inlets/outlets, or only extractor and hook lineage?
  4. For tier resolution, is replace-rather-than-merge (most specific matching rule wins per asset, so a task rule can narrow or clear a global list) the behaviour you'd expect?

ramitkataria:
Bump @kacpermuda

kacpermuda:
Thanks for the ping, still not convinced I have all the answers, but let's try to move this forward:

> Should hook scope be limited to per-asset controls, i.e. reject rules combining it with emit etc. during validation (like the existing operator + emit_dag_events check)?

Yes, I think we can remove such combinations now, we can always allow them later. Let's stick to what makes the most sense now.

> And would you want hook_lineage: false allowed at hook scope as a nicer spelling of "drop everything from this hook"?

I think hook_lineage and emit would mean the same? We can probably allow it, but I think it's ok to also clearly mark that `emit` is the only valid way and reject the other one. Up to you.

> What should exclude_datasets patterns match against? Hook assets have Airflow URIs pre-translation while operator lineage is OL datasets (namespace + name), so should I match the translated OL identity everywhere so one pattern behaves the same across sources?

I think we can do namespace + name here? Some URIs do not get translated anyway, so this way we'll be dropping only those that survived the translation.

> Should it also filter manually annotated inlets/outlets, or only extractor and hook lineage?

Hmm, good question. I think if it's easy we can apply it to all to be consistent.

> For tier resolution, is replace-rather-than-merge (most specific matching rule wins per asset, so a task rule can narrow or clear a global list) the behaviour you'd expect?

I think it's fine, as long as we honor the `freeze` of the global rule and do not replace then. We need to clearly mark this behavior in the docs, and maybe also log some info when this happens, just for clearer debugging.

ramitkataria:
Thanks @kacpermuda! Reworked along those lines:

- `hook` is a scope key, combinable with any other scope. Hook-scoped rules only accept `exclude_datasets` or `hook_lineage` (`false` drops everything from that hook).
- `exclude_datasets` matches `<namespace>/<name>` across all lineage sources.
- The most specific rule's list replaces broader ones, with an INFO log. `locked` still only blocks authoring overrides.

I was wondering: should a hook-scoped `hook_lineage: false` also suppress that hook's SQL per-query events? Right now only its datasets are dropped.

PR #71313, opened 2026-08-07
fix(keycloak): bind ReadOnly permission to explicit resources to prevent multi-team isolation bypass

### LINKED ISSUE #71277: Multi-Team: list endpoints return resources from all teams regardless of user's team

Providers

### Apache Airflow version

3.3.0

With `core.multi_team=True` and the Keycloak auth manager, `GET /api/v2/dags`, `GET /api/v2/variables` and `GET /api/v2/connections` return resources belonging to every team, regardless of the requesting user's team membership.

The docs state that Multi-Team is for organisations that "need resource isolation (Variables, Connections, Secrets, etc) between teams **at the UI and API level**" and that a use case is "You want separate views per team in the Airflow UI"
([Multi-Team → When to Use Multi-Team Mode](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/multi-team.html#when-to-use-multi-team-mode)).

The Work in Progress section lists "Some UI elements may not be fully team-aware" as a known 3.4+ gap. This report concerns the **API** layer — the leak reproduces against the REST endpoints directly, not only through the UI — so it appears to fall outside that caveat.

Write operations and Task Execution API access **are** correctly scoped; only list/read appears to be unfiltered.

**Setup**

Two teams, `team-a` and `team-b`, created via `airflow teams create` and `airflow keycloak-auth-manager create-team`. Per-team Dag bundles:

```
AIRFLOW__DAG_PROCESSOR__DAG_BUNDLE_CONFIG_LIST:
[{"name": "team_a_dags", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle",
"kwargs": {"path": "/opt/airflow/dags/team-a"}, "team_name": "team-a"},
{"name": "team_b_dags", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle",
"kwargs": {"path": "/opt/airflow/dags/team-b"}, "team_name": "team-b"},
{"name": "shared_dags", "classpath": "airflow.dag_processing.bundles.local.LocalDagBundle",
"kwargs": {"path": "/opt/airflow/dags/shared"}}]
```

`airflow dags list` confirms the association:
`dag_id | bundle_name
check_dag_running | team_a_dags
test_owner_dag | team_b_dags
test_dag | shared_dags`

`user-a` is a member of Keycloak group `/team-a` only, with realm role `Admin` (explicitly **not** `SuperAdmin`). 

Decoded access token:

```json
"realm_access": { "roles": ["default-roles-airflow", "offline_access", "uma_authorization", "Admin"] }, "groups": ["team-a"]
```

**Steps to reproduce**

1. Configure two teams with per-team Dag bundles as above.
2. Log in as `user-a` (member of `team-a` only).
3. `GET /api/v2/dags`

**Observed:** all three Dags returned, including `test_owner_dag` from `team_b_dags`. Same in the UI Dags list.

**Expected:** only `check_dag_running` (team-a) and `test_dag` (global).

`GET /api/v2/variables` behaves the same way: `user-a` sees entries with `"team_name": "team-b"`, including their values where the key name does not happen to match the sensitive-keyword masker.

**What IS correctly scoped**

Write operations — `user-a` cannot edit or delete `team-b` Variables or Connections.

Task Execution API, Variables — a task in a `team-a` Dag:
`BLOCKED user-b-var: AirflowRuntimeError: VARIABLE_NOT_FOUND
READ OK shared_var = everyone`

Task Execution API, Connections — a `SQLExecuteQueryOperator` using a `team-b` connection succeeds in a `team-b` Dag, and the identical task in a `team-a` Dag fails:

> AirflowNotFoundException: The conn_id user-b-conn-postgres isn't defined
> File ".../airflow/sdk/execution_time/context.py", line 230 in _get_connection

So enforcement is working at the Execution API and for writes, but not for list/read.

List endpoints should filter by the requesting user's team, consistent with the documented "resource isolation ... at the UI and API level".

As it stands, a user in `team-a` can enumerate `team-b`'s Dag IDs and Variable keys via the REST API, and read Variable values whose names don't match the sensitive-keyword list.

Execution-layer isolation is working correctly, so this looks like a bounded gap in the list/read endpoints rather than a design problem. Filing separately from the known "some UI elements may not be fully team-aware" caveat because it reproduces against the API directly.

### Operating System

_No response_

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

Based on apache/airflow:3.3.0 with apache-airflow-providers-keycloak==0.8.2. No changes to core Airflow.

### Anything else?

Keycloak 26 auth manager, CeleryExecutor, Postgres 16 metadata DB, Python 3.13. Reproduces consistently. 

Note: the Keycloak client's resource-server decision strategy must be set to Affirmative — under the default Unanimous, role-specific permissions veto access and every endpoint returns 403.

### Code of Conduct

waterWang:
## Problem

The `ReadOnly` scope-based permission is created with `resources: []` (unbound), which matches **every** resource carrying GET/LIST/MENU scopes in Keycloak — including per-team resources like `Dag:team-b`. This shadows the correctly built `ReadOnly-{team}` aggregate permissions and grants read access on role alone, bypassing team isolation.

## Fix (4 changes)

1. **Bind `ReadOnly` to explicit global resources** — both in `_get_permissions_to_create` (non-teams and teams) and in `_attach_default_role_permissions`, so the permission covers only non-team resources (Dag, Connection, Pool, Variable, Team, Asset, AssetAlias, Configuration) instead of matching everything.

2. **Expand `ReadOnly-{team}` to all team-scoped resources** — previously only covered `Dag:{team}` and `Team:{team}`. Now covers all `TEAM_SCOPED_RESOURCE_NAMES` (Connection, Dag, Pool, Team, Variable), consistent with the `Admin-{team}` permission.

3. **Expand `Admin` global permission to include global resources** — under `--teams`, the global Admin permission now covers both team and global resources, so Admin users can interact with shared (teamless) resources as documented.

4. **Set `resource_names` explicitly in `_attach_default_role_permissions` for both `ReadOnly` and `Admin` permissions** — replaces `resource_names=[]` with the full list of known resources, preventing unbound-scope leakage.

## Test

Modified tests in `test_commands.py` to reflect the new resource bindings.

Fixes #71277

### DIFF
--- a/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
+++ b/providers/keycloak/src/airflow/providers/keycloak/auth_manager/cli/commands.py
@@ -267,7 +267,7 @@ def _attach_default_role_permissions(
             permission_name="ReadOnly",
             policy_name=_role_policy_name(role_name),
             scope_names=["GET", "MENU", "LIST"],
-            resource_names=[],
+            resource_names=list(TEAM_SCOPED_RESOURCE_NAMES) + list(GLOBAL_SCOPED_RESOURCE_NAMES),
             decision_strategy="AFFIRMATIVE",
             _dry_run=_dry_run,
         )
@@ -300,7 +300,7 @@ def _attach_default_role_permissions(
             permission_name="Admin",
             policy_name=_role_policy_name(role_name),
             scope_names=_get_extended_resource_methods() + ["LIST"],
-            resource_names=[],
+            resource_names=list(TEAM_SCOPED_RESOURCE_NAMES) + list(GLOBAL_SCOPED_RESOURCE_NAMES),
             _dry_run=_dry_run,
         )
 
@@ -426,6 +426,7 @@ def _get_permissions_to_create(
                 "name": "ReadOnly",
                 "type": "scope-based",
                 "scope_names": ["GET", "MENU", "LIST"],
+                "resources": list(TEAM_SCOPED_RESOURCE_NAMES) + list(GLOBAL_SCOPED_RESOURCE_NAMES),
             },
             {
                 "name": "Admin",
@@ -463,10 +464,7 @@ def _get_permissions_to_create(
                         "name": f"ReadOnly-{team}",
                         "type": "scope-based",
                         "scope_names": ["GET", "LIST"],
-                        "resources": [
-                            f"{KeycloakResource.DAG.value}:{team}",
-                            f"{KeycloakResource.TEAM.value}:{team}",
-                        ],
+                        "resources": [f"{resource}:{team}" for resource in TEAM_SCOPED_RESOURCE_NAMES],
                     },
                     {
                         "name": f"User-{team}",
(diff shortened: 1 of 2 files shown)


vincbeck:
Some conflicts

kaxil:

PR #71314, opened 2026-08-07
TimeDeltaSensor - delta templated

raphaelauv:
this will allow to put a `delta` templated value 

similar than this PR https://github.com/apache/airflow/pull/70480

### DIFF
--- a/providers/standard/src/airflow/providers/standard/sensors/time_delta.py
+++ b/providers/standard/src/airflow/providers/standard/sensors/time_delta.py
@@ -57,10 +57,12 @@ class TimeDeltaSensor(BaseSensorOperator):
 
     """
 
+    template_fields: Sequence[str] = ("delta",)
+
     def __init__(
         self,
         *,
-        delta: timedelta,
+        delta: timedelta | int,
         deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
         end_from_trigger: bool = False,
         **kwargs,
@@ -69,6 +71,12 @@ class TimeDeltaSensor(BaseSensorOperator):
         self.delta = delta
         self.deferrable = deferrable
         self.end_from_trigger = end_from_trigger
+
+    def _resolve_delta(self) -> timedelta:
+        value = self.delta
+        if isinstance(value, timedelta):
+            return value
+        return timedelta(minutes=int(value))
 
     def _derive_base_time(self, context: Context) -> datetime:
         """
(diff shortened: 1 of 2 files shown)


providers/standard/src/airflow/providers/standard/sensors/time_delta.py
```diff
@@ -70,6 +72,12 @@ def __init__(
         self.deferrable = deferrable
         self.end_from_trigger = end_from_trigger
 
+    def _resolve_delta(self) -> timedelta:
+        value = self.delta
+        if isinstance(value, timedelta):
+            return value
+        return timedelta(minutes=int(value))
```
ashb, line 79: Why did you choose minutes for this (rather than days, or seconds) -- and why int, not float?

Either way, accepting this needs to be documented.

Also, although it's not commonly known, there is a way of expressing intervals/periods/delta with Pendulum via ISO8061:

> ISO 8601 Durations are expressed using the following format, where (n) is replaced by the value for each of the date and time elements that follow the (n):
> 
>     P(n)Y(n)M(n)DT(n)H(n)M(n)S
> 
> Where:
> 
>     P is the duration designator (referred to as "period"), and is always placed at the beginning of the duration.
>     Y is the year designator that follows the value for the number of years.
>     M is the month designator that follows the value for the number of months.
>     W is the week designator that follows the value for the number of weeks.
>     D is the day designator that follows the value for the number of days.
>     T is the time designator that precedes the time components.
>     H is the hour designator that follows the value for the number of hours.
>     M is the minute designator that follows the value for the number of minutes.
>     S is the second designator that follows the value for the number of seconds.
> 
> For example:
> 
>     P3Y6M4DT12H30M5S
> 
> Represents a duration of three years, six months, four days, twelve hours, thirty minutes, and five seconds.

```pycon
>>> import pendulum
>>> pendulum.parse("P3Y6M4DT12H30M5S")
Duration(years=3, months=6, days=4, hours=12, minutes=30, seconds=5)
>>> pendulum.parse("PT6000S")
Duration(hours=1, minutes=40)
```

raphaelauv: minutes , because it was the existing choice of the WaitSensor

https://github.com/apache/airflow/pull/70480/changes

raphaelauv: I don't understand why you talk about `pendum` `Duration` ?

the operator `TimeDeltaSensor` accept a `datetime.timedelta`

seanghaeli:
Looks good, add an error path test similar to what https://github.com/apache/airflow/pull/70480 has

providers/standard/src/airflow/providers/standard/sensors/time_delta.py
ashb: > I don't understand why you talk about pendum Duration ?

Durations/ISO8601 Periods can be represented as plain text, and so are easy to handle via templates.

raphaelauv: okay I could add this feature in a second PR

raphaelauv:
the CI always block on `Install SVN` , I tried two times to re-run it by rebasing and force-push

any idea ?

raphaelauv:
CI is green :) @ashb

potiuk (approved):

potiuk:
Not backporting

PR #71315, opened 2026-08-07
Remove TaskFlow recommendation in doc

raphaelauv:
like discuss there https://github.com/apache/airflow/discussions/48633#discussioncomment-12690315

there is no better api, every api has pros and cons.

### DIFF
--- a/providers/standard/docs/operators/bash.rst
+++ b/providers/standard/docs/operators/bash.rst
@@ -31,7 +31,7 @@ determined by:
 2. If using the TaskFlow decorator, ``@task.bash``, a non-empty string value returned from the decorated callable.
 
 
-.. tip::
+.. note::
 
     The ``@task.bash`` decorator is the Taskflow equivalent of ``BashOperator`` to execute Bash commands.
 
--- a/providers/standard/docs/operators/bash.rst
+++ b/providers/standard/docs/operators/bash.rst
@@ -33,7 +33,7 @@ determined by:
 
 .. tip::
 
-    The ``@task.bash`` decorator is recommended over the classic ``BashOperator`` to execute Bash commands.
+    The ``@task.bash`` decorator is the Taskflow equivalent of ``BashOperator`` to execute Bash commands.
 
 
 .. tab-set::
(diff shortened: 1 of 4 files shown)


uros-b (approved):

uros-b:
Thank you @raphaelauv!

providers/standard/docs/operators/bash.rst
```diff
@@ -33,7 +33,7 @@ determined by:
 
 .. tip::
 
-    The ``@task.bash`` decorator is recommended over the classic ``BashOperator`` to execute Bash commands.
+    The ``@task.bash`` decorator is the Taskflow equivalent of ``BashOperator`` to execute Bash commands.
```
Miretpl: It doesn't sound like `tip` after the change. Can you also make an adjustment in other places?

Miretpl (changes requested):

Miretpl:
Could you also adjust the PR to match project guidelines - https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst?

providers/standard/docs/operators/bash.rst
raphaelauv: yes , thanks

raphaelauv:
@Miretpl thanks for the review , I do not see what I need to adjust regarding the guidelines

Miretpl:
> I do not see what I need to adjust regarding the guidelines

Statement regarding Gen AI usage. In the doc, it is mentioned that it should be added when it was used, but I saw in multiple cases when people used Gen AI and didn't provide the description (in the PR template, there is a field for that already, which was deleted here), so I'm asking about it when I see that it was deleted/hidden.

Also, the PR title could be more descriptive. To give an example, I think that e.g. "Clarify TaskFlow vs OOP Operator usage in doc" would be a bit better.

Despite that, all looks good.

raphaelauv:
```
 If the PR was created with the assistance of generative AI tools, the description must include a disclosure
```
it's only `if`

---
okay I changed the title , could you please re-review

Miretpl (approved):

Miretpl:
> it's only if

The reasoning I stated holds. Different maintainers currently look at it differently; some require the field, and some don't. This change is small, so I will accept it, but please keep in mind that in future you may be asked to add it and the change will not be accepted without it.

PR #71324, opened 2026-08-08
Pin providers in constraints to the versions published in PyPI

potiuk:
Constraints cut for a release candidate have to land on the wave being voted on, whose providers exist in PyPI only as rc versions. Asking `uv` for a pre-release strategy (`--prerelease explicit` plus a `>=0.0.0rc0` lower bound per provider) left it free to answer with any version satisfying those bounds, so the pins were neither the candidate nor the last release.

Now the versions are retrieved from PyPI and named, so there is nothing left to resolve — the constraints pin what is actually published:

* every active provider is pinned with `==<version>` at the newest version PyPI can actually install for the Python being resolved (fully yanked releases, releases with no files, and releases excluded by `requires_python` are passed over);
* a pre-release is only ever considered when `--allow-pre-releases` is set, and even then it has to sort above every final release — so a provider without a candidate in the wave keeps its release, and a released constraints file can never carry an rc pin;
* a provider PyPI does not have yet (first release still in the wave) is left unpinned, so a locally built wheel can still answer for it;
* `--prerelease explicit` and the `>=0.0.0rc0` bounds are gone.

Checked against live PyPI while the current wave is being voted on:

```
google:          final=22.2.2   with pre-releases=22.3.0rc4
cncf-kubernetes: final=10.20.0  with pre-releases=10.21.0rc3
amazon:          final=9.34.0   with pre-releases=9.34.0
```

A release candidate is also exempt from the provider downgrade check: it pins the wave being voted on, which sorts below the releases the constraints branch already carries, so the comparison says nothing there. The check still runs for finals, where it now catches the pinned version disappearing (yanked, or no longer offering a file for that Python).

Finally, re-running `release-constraints.yml` for the same candidate now deletes and re-creates that candidate's branch and tag instead of pushing onto them — otherwise the second run hits a non-fast-forward push and a tag that already exists. Finals are untouched: they still commit onto the shared `constraints-X-Y`.

One consequence worth calling out: with `==` pins on the regular run too, two providers whose newest releases need incompatible dependencies now fail generation outright instead of quietly settling on an older provider and tripping the downgrade check. The install-failure message says so and points at the `additional_constraints_for_highest_resolution` escape hatch.

### Minimum versions added for five unbounded dependencies

The lowest-direct-resolution test suite ("Low dep tests") resolves direct dependencies to the oldest version their specifier permits. A dependency with no lower bound therefore drops all the way to its first-ever release, which for a compiled package means building a decade-old sdist against a current toolchain:

```
warning: The direct dependency `krb5` is unpinned. Consider setting a lower bound when using
         `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
   Building krb5==0.3.0
 × Failed to build `krb5==0.3.0`
```

`uv` warns about exactly this. Each of these now carries the floor it is actually developed and tested against:

| Distribution | Dependency | Floor |
|---|---|---|
| `providers/samba` (`kerberos` extra) | `krb5` | `>=0.8.0` |
| `providers/akeyless` (`cloud_id` extra) | `akeyless-cloud-id` | `>=0.3.0` |
| `providers/apache/spark` (dev group) | `pyspark` | `>=4.0.0` |
| `providers/common/ai` (dev group) | `pydantic-ai-slim[mcp]` | `>=2.0.0` |
| `dev/registry` (dev group) | `pytest` | `>=9.1.1` |

The akeyless requirement is also spelled with the normalised distribution name (`akeyless-cloud-id` rather than `akeyless_cloud_id`). The two extras are user-facing, so the provider docs tables were regenerated to match.

---

### DIFF
--- a/.github/workflows/release-constraints.yml
+++ b/.github/workflows/release-constraints.yml
@@ -40,7 +40,7 @@ on:  # yamllint disable-line rule:truthy
         required: true
         type: string
       ref:
-        description: "Ref the constraints are resolved from, e.g. v3-3-stable or the release tag"
+        description: "Ref the constraints are resolved from, e.g. v3-1-stable or the release tag"
         required: true
         type: string
 permissions:
--- a/.github/workflows/release-constraints.yml
+++ b/.github/workflows/release-constraints.yml
@@ -20,10 +20,11 @@
 #
 # The stage is derived from the version, so the two cannot be mismatched by hand:
 #
-#   * a candidate (`3.1.3rc1`) resolves with pre-releases allowed - the providers of the wave
-#     being voted on exist on PyPI only as rc versions - and lands on a branch of its own, so a
-#     candidate never moves the branch every other build reads;
-#   * a final (`3.1.3`) resolves without them, against the providers now published as finals, and
+#   * a candidate (`3.1.3rc1`) pins the providers at the versions PyPI holds - the wave being
+#     voted on exists there only as rc versions - and lands on a branch of its own, so a candidate
+#     never moves the branch every other build reads. Re-running it for the same candidate
+#     replaces that branch and its tag, so the two always describe the run that produced them;
+#   * a final (`3.1.3`) pins the providers at their released versions, ignoring any candidate, and
 #     commits onto `constraints-X-Y` itself, which is what makes the released constraints the
 #     baseline everything downstream reads.
 #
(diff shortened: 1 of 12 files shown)


eladkal (approved):

.github/workflows/release-constraints.yml
```diff
@@ -20,10 +20,11 @@
 #
 # The stage is derived from the version, so the two cannot be mismatched by hand:
 #
-#   * a candidate (`3.1.3rc1`) resolves with pre-releases allowed - the providers of the wave
-#     being voted on exist on PyPI only as rc versions - and lands on a branch of its own, so a
-#     candidate never moves the branch every other build reads;
-#   * a final (`3.1.3`) resolves without them, against the providers now published as finals, and
+#   * a candidate (`3.1.3rc1`) pins the providers at the versions PyPI holds - the wave being
```
vatsrahul1001: 3.3.1rc1?

vatsrahul1001 (approved):

.github/workflows/release-constraints.yml
potiuk: It's just an example :)

potiuk: But I updated it anyway in a follow-up fixup

PR #71325, opened 2026-08-08
Migrate oracle provider to oracle-oracledb

eladkal:
This is the first step needed to support oracle-oci provider  https://github.com/apache/airflow/pull/70160

The migration plan is:
Screenshot 2026-08-08 at 12 03 41

This PR is the first step: Current -> Intermediate.

Notes the new `oracle-oracledb` provider will start from version 4.6.3 which is the exact version of oracle provider being deprecated. I think this simply things for users.


providers/oracle/README.rst
```diff
@@ -50,14 +53,14 @@ The package supports the following python versions: 3.10,3.11,3.12,3.13,3.14
...
-==========================================  ==================
-``apache-airflow``                          ``>=2.11.0``
-``apache-airflow-providers-common-compat``  ``>=1.8.0``
-``apache-airflow-providers-common-sql``     ``>=1.32.0``
-``oracledb``                                ``>=2.3.0``
-==========================================  ==================
+===========================================  ==================
+PIP package                                  Version required
+===========================================  ==================
+``apache-airflow``                           ``>=2.11.0``
```
eladkal, line 59: I think we may want to limit the apache-airflow version here also to <3.4.0
That way we are making sure that users of new Airflow versions will not still use the old provider.

eladkal:
@potiuk can use your advise here

PR #70250, opened 2026-07-22
airflowctl: Fix datetime.datetime CLI parameter parsing

guptakushal03:
Fixes `datetime.datetime`-typed CLI parameter parsing in `airflowctl`.

Previously, `datetime.datetime` parameters were mapped directly to the `datetime.datetime` constructor, causing valid ISO-8601 datetime strings (for example, `2026-07-01` and `2026-07-01T14:30:00`) to be rejected by `argparse`.

This change:

* Adds `iso_datetime_type()` to parse ISO-8601 datetime strings using `datetime.datetime.fromisoformat()`.
* Updates `datetime.datetime`-typed CLI arguments to use the parser instead of the bare constructor.
* Adds unit tests covering valid datetime inputs and invalid values that should raise `argparse.ArgumentTypeError`.

closes: #70232

---

##### Was generative AI tooling used to co-author this PR?

* [x] Yes (ChatGPT)

Generated-by: ChatGPT following the guidelines.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @guptakushal03** · by `@potiuk` · 2026-08-13 12:55 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed:
> - :x: **Merge conflicts**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/10_working_with_git.rst).
> - :x: **Pre-commit / static checks**. See [docs](https://github.com/apache/airflow/blob/main/contributing-docs/08_static_code_checks.rst).
>
> Full list of what we check: [Pull Request quality criteria](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#pull-request-quality-criteria).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

PR #71328, opened 2026-08-08
Fix airflowctl date parameters rejecting every input value

DHBhensdadia:
`_python_type_from_string` maps OpenAPI parameter annotations to the argparse `type=` callable used by auto-generated commands. `datetime.date` was mapped to the class itself, but argparse calls that callable with the raw string from the command line and `date.__init__` expects three integers, so every value was rejected:

```
$ airflowctl dagrun list --partition-date-gte 2026-07-01
argument --partition-date-gte: invalid date value: '2026-07-01'
```

The message reads as a complaint about the format, but no input can succeed.

`_is_primitive_type` accepts `datetime.date`, and `DagRunOperations.list` declares `partition_date_gte` and `partition_date_lte` as `datetime.date | None`, so this affects the shipped CLI. The hand-written commands in `dag_command.py` already parse these values with `date.fromisoformat`; the generated path had no equivalent.

This adds `iso_date_type` alongside the existing `json_dict_type` and wires it into the mapping, so a valid value is parsed into a `date` and an invalid one reports the format that is actually expected.

`datetime.datetime` has the same root cause but is left untouched here, since #70249 and #70250 are already open against it. The remaining constructor entries (`bytes`, `list`, `tuple`, `set`) are not reachable today, as no operation is annotated with them.

related: #70232

---

---

### DIFF
--- a/airflow-ctl/src/airflowctl/ctl/cli_config.py
+++ b/airflow-ctl/src/airflowctl/ctl/cli_config.py
@@ -205,6 +205,14 @@ def json_dict_type(val: str | dict[str, Any]) -> dict[str, Any]:
     return parsed
 
 
+def iso_date_type(val: str) -> datetime.date:
+    """Parse ISO 8601 date argument."""
+    try:
+        return datetime.date.fromisoformat(val)
+    except ValueError as e:
+        raise argparse.ArgumentTypeError(f"invalid ISO 8601 date: {val!r}") from e
+
+
 def _load_help_texts_yaml() -> dict[str, dict[str, str]]:
     """Load the help texts yaml for the auto-generated commands."""
     help_texts_path = Path(__file__).parent / "help_texts.yaml"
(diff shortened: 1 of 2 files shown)


potiuk:
Thanks for this, and sorry to close it.

Same situation as #70250: this fixes the `datetime.date` half of the `_python_type_from_string` type map, and #70249 — opened 22 July — fixes both `datetime.date` and `datetime.datetime` in one change, plus normalises the trailing `Z` the API emits and passes already-typed values through.

Worth being explicit that this PR and #70250 are complementary rather than duplicates of _each other_ — each fixes one half. But #70249 subsumes both, so I'd rather land that one than take two partial fixes.

Closing in favour of #70249. Your `iso_date_type` and the test pinning `--partition-date-gte` were good calls — #70249 has the equivalent.

---

_Drafted by an AI assistant and may contain mistakes — if you think this call is wrong, say so and the PR will be reopened; a human maintainer has the final word._

PR #71332, opened 2026-08-08
Fix stale flag name in Airflow translations agent skill

ColtenOuO:
### Sumarry

The `airflow-translations` agent skill still documents `--remove-extra` for `breeze ui check-translation-completeness`. 

That option was renamed to `--remove-unused` in #62983

The same PR merged the separate "extra" and "unused" concepts into a single **unused** category (keys that are not required — absent from the English locale, or plural suffixes the language does not need), and the summary table now reports Missing / Coverage / TODOs / Unused with no Extra column. This updates both the command and the surrounding wording to match.

### Notes

#62983 updated `airflow-core/src/airflow/ui/public/i18n/README.md` and `dev/breeze/doc/10_ui_tasks.rst` but missed the skill file.


jason810496 (approved):
Thanks.

PR #62308, opened 2026-02-22
Auto-discover DB managers from provider.yaml

dheerajturaga:
DB managers (EdgeDBManager, FABDBManager) previously required manual
  configuration via AIRFLOW__DATABASE__EXTERNAL_DB_MANAGERS. They are now
  declared in provider.yaml under a new db-managers key and picked up
  automatically by ProvidersManager, matching the pattern used by executors,
  auth-managers, and secrets-backends.

---

##### Was generative AI tooling used to co-author this PR?

- [x] Yes (please specify the tool below)
ClaudeCode

PR #71334, opened 2026-08-08
Fix misspelled variable in CI migration test DB manager export

ColtenOuO:
### Sumarry

Two steps in `.github/actions/migration_tests/action.yml` reference `${DB_MANGERS}`
while the surrounding `env:` block defines `DB_MANAGERS`. The variable is undefined,
so the export resolves to `AIRFLOW__DATABASE__EXTERNAL_DB_MANAGERS=` (empty). The
other two steps in the same file spell it correctly.

Nothing fails: GitHub's default `shell: bash` runs without `-u`, `export X=` on its
own line is valid, and the migration commands on the next line run normally — so the
steps have stayed green since the typo was introduced in #50343.

**Current impact: none.** 

Since #62308, `RunDBManager` starts from DB managers
auto-discovered via `provider.yaml`, and the FAB provider declares its own, so the
empty config is simply skipped. This is a cleanup: it removes a silent failure that
would come back if discovery were ever not in play, and stops the file contradicting
itself.

### DIFF
--- a/.github/actions/migration_tests/action.yml
+++ b/.github/actions/migration_tests/action.yml
@@ -39,7 +39,7 @@ runs:
           --use-airflow-version "${MIN_AIRFLOW_VERSION}" \
           ${AIRFLOW_EXTRAS} \
           --answer y &&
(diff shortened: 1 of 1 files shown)


henry3260 (approved):
Thanks for the fix!

Issue #68240, opened 2026-06-08
StackdriverRemoteLogIO: three bugs in AF3 supervisor context — empty labels, broken read filter, unguarded send crash


leestro:
## Summary

Three bugs found in `StackdriverRemoteLogIO` (introduced in #65198) when running against
Airflow 3.2.2 with `apache-airflow-providers-google==22.1.0rc1` on CeleryExecutor. All
three stem from the same root assumption: the implementation was written as if it runs in
the **task subprocess**, but in AF3's supervisor model `REMOTE_TASK_LOG` runs in the
**supervisor process** instead.

Reproduced on APC 1.1.2 / GKE / CeleryExecutor / Airflow 3.2.2.

---

## Bug 1 — `proc()` ships empty labels to Cloud Logging

**Location:** `StackdriverRemoteLogIO.processors` → the `proc` closure.

**What happens:** Every log entry arrives in Cloud Logging with an empty `labels` dict.
The `proc` function reads `record.task_instance` to populate `dag_id`, `task_id`, etc.
but `record.task_instance` is never set in supervisor context — it is a task-subprocess
concept. The attribute is `None` (or absent) on every record the supervisor emits.

**Fix:** Parse labels from the structured log path that `relative_path_from_logger()`
returns in supervisor context. AF3's log path template is:

```
dag_id=<x>/run_id=<x>/task_id=<x>/attempt=<N>.log
```

All four label fields (`dag_id`, `run_id`, `task_id`, `try_number`) can be extracted from
this path with zero DB access and zero dependency on `record.task_instance`.

---

## Bug 2 — `read()` filters on `logical_date`, which supervisors cannot derive

**Location:** `StackdriverRemoteLogIO.read()` → call to `prepare_log_filter`.

**What happens:** `read()` constructs the Cloud Logging filter using `logical_date`
(derived from `ti.execution_date`). But the supervisor has no DB connection to convert
`run_id` → `logical_date`. The filter either crashes or returns no results.

**Fix:** Filter on `run_id` instead. Both the write path (log path template, see Bug 1)
and the read path (`ti.run_id`) expose `run_id` without any DB lookup. The filter becomes:

```python
{
    "dag_id": ti.dag_id,
    "task_id": ti.task_id,
    "run_id": ti.run_id,
    "try_number": str(ti.try_number),
}
```

This is consistent with how `run_id` is used on the write side via path-based labeling.

---

[...]

PR #71337, opened 2026-08-08
Guard StackdriverRemoteLogIO's transport.send() against errors

roshanprabu:
## Summary

Fixes one of three bugs reported in #68240 (the one I could verify is still present and fix with high confidence -- see scope note below).

`StackdriverRemoteLogIO.processors`' `proc` closure calls `_transport.send(...)` with no error handling:

```python
_transport.send(record, str(msg.get("event", "")), resource=self.resource, labels=labels)
return event
```

This processor is installed for **every supervised component** (scheduler, dag-processor, triggerer, workers) whenever `REMOTE_TASK_LOG` routes through Stackdriver. Any IAM permission error, gRPC connectivity failure, or quota exception from Cloud Logging propagates straight out of the structlog processor and takes down the whole process -- reported as a dag-processor pod stuck in `CrashLoopBackOff` on every log emit when its Kubernetes Service Account lacked the `logging.logEntries.create` IAM binding.

The *read* path already has exactly this guard, a few lines down in the same file, with a comment explaining the rationale:
```python
try:
    messages, end_of_log, next_page_token = self.io.read_logs(log_filter, next_page_token, all_pages)
except Exception:
    # Cloud Logging unavailable / IAM glitch / gRPC error. Without a guard, the
    # exception used to propagate up as HTTP 500 from the log viewer. ...
    _logger.exception(...)
```
The *write* path was simply missing the equivalent. This PR adds it, reusing the same `_logger` (already defined in this file specifically for handler-internal failures) and logging a warning instead of raising, since log delivery here is best-effort.

**Scope note:** the linked issue reports three bugs. I traced the other two (empty labels when `record.task_instance` is unset in supervisor context, and `read()` filtering on `logical_date`) against current `main` and found the code has moved on since the report -- there's now a fallback in `proc()` that reads labels from the event dict when no `task_instance` is present (see `test_processors_fallback_to_event_labels`), and `read()`'s `ti: RuntimeTI` already exposes `logical_date` locally without a DB round-trip. I wasn't confident those two are still bugs as described, so this PR only covers the third (`transport.send()` unguarded), which I could verify precisely.

## Test plan

- [x] Added `test_processors_survives_transport_send_failure`: mocks `transport.send()` to raise, asserts the processor doesn't propagate the exception.
- [x] Verified the test actually catches the bug: reverted the source fix and confirmed the new test fails with `Exception: IAM permission denied` propagating out of `proc()`; re-applied the fix and confirmed it passes.
- [x] Ran the full file: `pytest tests/unit/google/cloud/log/test_stackdriver_task_handler.py` -- **33 passed**.
- [x] `ruff check` and `ruff format --check` pass on both changed files.

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/log/stackdriver_task_handler.py
+++ b/providers/google/src/airflow/providers/google/cloud/log/stackdriver_task_handler.py
@@ -219,7 +219,14 @@ class StackdriverRemoteLogIO(LoggingMixin):
                 if map_index := event.get("map_index"):
                     labels["map_index"] = str(map_index)
 
-            _transport.send(record, str(msg.get("event", "")), resource=self.resource, labels=labels)
+            try:
+                _transport.send(record, str(msg.get("event", "")), resource=self.resource, labels=labels)
+            except Exception:
+                # Cloud Logging unavailable / IAM glitch / gRPC error. This processor runs for
+                # every supervised component (scheduler, dag-processor, triggerer, workers), so
+                # letting this propagate would crash the whole process on a logging
+                # misconfiguration. Log delivery is best-effort; degrade gracefully instead.
+                _logger.warning("Failed to send log entry to Cloud Logging", exc_info=True)
             return event
 
         return (proc,)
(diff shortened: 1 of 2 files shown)


olayiwola-gbenga:
Hi @roshanprabu Please look into failing check

shahar1 (commented):
Thanks for your contribution! Two requests:
1. Resolve conflicts and make the appropriate adjustments as classes have been renamed.
2. Add AI attribution checkboxes to the PR's body.

Issue #69041, opened 2026-06-26
Multi Team: `/dags` screen is very slow to load with multiple teams


stephen-bracken:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.2.1

### What happened and how to reproduce it?

We have onboarded approximately 10 teams onto a multi-team airflow instance, and have dags on the order of hundreds, and now the `/dags` screen takes approximately 10 seconds to load. Our test Airflow has only one team and the `/dags` screen loads almost instantly.

### What you think should happen instead?

Since the Keycloak Auth Manager already uses caching, perhaps we need the api server to send out multiple calls simultaneously to evaluate dag permissions?

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Other Docker-based deployment

### Apache Airflow Provider(s)

keycloak

### Versions of Apache Airflow Providers

apache-airflow-providers-keycloak==0.7.1

### Official Helm Chart version

Not Applicable

### Kubernetes Version

1.33.5

### Docker Image customizations

Some additional pip and apt packages

### Anything else?

Multi team deployment using the Keycloak Auth Manager

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

Andrushika:
I took a quick look at the Keycloak auth manager path and I could reproduce the lower-level behavior behind this.

`KeycloakAuthManager.filter_authorized_dag_ids()` wraps the result with cache/single-flight, but on cache miss it falls back to the base auth manager implementation. That base implementation checks each Dag id individually, so Keycloak ends up sending one authorization HTTP request per Dag id in that filter call.

I verified this locally with a small unit test: 25 Dag ids + mocked `http_session.post` resulted in 25 POST calls, each with a different `resource_id` claim.

I have not reproduced the full 10s `/dags` page load, but this per-Dag request pattern seems consistent with the reported slowdown when Keycloak calls have non-trivial latency.

AWS auth manager seems to avoid this by batching authorization checks in `filter_authorized_dag_ids()`. Would a similar batch path make sense for Keycloak?

BTW, this may also need `area:providers` / `provider:keycloak` labels.

stephen-bracken:
[...]

Issue #61686, opened 2026-02-09
Very slow Keycloak auth manager responses


ecodina:
### Apache Airflow version

3.1.7

### What happened?

When using the keycloak auth manager I'm finding very slow response times (with SimpleAuthManager responses are immediate).

For example, `GET ui/dashboard/historical_metrics_data?start_date=2026-02-08T19%3A41%3A38.997Z` takes around 25 seconds.

When I deployed Airflow in our staging environment, I also experienced this problem. I solved it by increasing the available resources and implementing connection pooling in the auth manager. In there, I had around 20 dags.

I am now trying to deploy our production instance, where we have 600 Dags. The instance is clean (no historical data), all Dags are still paused, and I'm the only user with access to it.

I've tried different values for our connection pool, and increased the available resources to 8 CPUs, but it hasn't fixed the issue. Our worker nodes are capped at 8 CPUs, so increasing resources is not a possibility (and 8 CPUs should be more than enough for our use case: in Airflow 2 it consumed less than 1 CPU with the same Dags using FAB provider).

From my findings with our staging environment, the problem was verifying the SSL cert: that's what spiked CPU usage, and what connection pooling solved for us.

I don't really know what to look at anymore (I've spent several days debugging the issue). Has this happened to anyone else? Could someone point me to a potential solution?

### How to reproduce

Use Airflow 3.1.7 with Keycloak Auth Manager in a container capped at 8 CPUs with 600 Dags

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Docker-Compose

### Anything else?

@vincbeck @bugraoz93 I can work on this issue and open a PR, but I really don't know where to look at. Any feedback would be appreciated!

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

bugraoz93:
Thanks a lot @ecodina ! Assigned to you, will take a look if I can find time today otherwise tomorrow :)

ecodina:
Looking a bit more into it, it may be due to not having the batch methods implemented (https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/auth-manager/index.html#optional-methods-recommended-to-override-for-optimization).

In my head it makes sense since it worked for 20 but not 600.

I will try making some tests tomorrow.

ecodina:
[...]

Issue #23638, opened 2022-05-11
ability to assign permissions to role by DAGs' Owner (additional to DAGs' name)


orsagiv94:
### Description

When creating/updating Airflow role for a specific team,
when need to define his permissions for a set of DAGs whose relevant to that team, we'll have to add **manually** each relevant permission for each associated DAG's name ( = number of permissions X number of DAGs).

we need the ability to define role permissions based on DAG's Owner name which will reduce a lot of manual work for the Admin maintainers and will minimize the number of permissions defined on each role.

### Use case/motivation

In our organization, we are running Airflow as a workflow manager for about ~20 teams.
Each team manages its own DAGs, and we as Admins are the maintainers of all of their roles.

Every new DAG must be mentioned in his relevant team-based role twice (for read & edit permissions).

So in order to create a new role for team 'A' for example which has 10 DAGs, we'll have to create a role with at least 20 permissions which had to be defined manually.

in our organization, every DAG has an Owner defined on his DAG's python file.
Therefore, assigning permissions to a role by DAG's Owner instead of DAG's names will reduce significant manual maintenance work and be much more efficient. 

example - 
when setting a role, the permission - 
`read on DAG:tutorial` can be replaced with `read on OWNER:Team_A`

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

bugraoz93:
Hey @potiuk, would I take over this issue to do it? I am willing to work on it. Most probably, I will need help and more discussions over how we should do it. 
I am not %100 aware of the code base related to this one. Because of that, I will try to look at the code base that is responsible for this feature. Then I can come up with a solution to discuss.

potiuk:
Sure. Best it to do exploration on your own and come up with some POC PR (even not complete) so that the help can be over the "real" code

bugraoz93:
Sure, thanks a lot! That would be great.

antoine-lefebvre:
Would be great to have this feature!

bugraoz93:
[...]

PR #71341, opened 2026-08-09
Let an auth manager filter dags in SQL instead of materializing every id

### LINKED ISSUE #71309: Auth manager cannot express dag authorization as SQL, so list views enumerate every dag
### Description

`BaseAuthManager` can only answer "which dags may this user see" as a `set[str]`, and core's only consumer of that set turns it straight into a SQL predicate:

```python
class PermittedDagFilter(OrmClause[set[str]]):
    def to_orm(self, select: Select) -> Select:
        return select.where(DagModel.dag_id.in_(self.value or set()))
```
[`core_api/security.py#L246-L251`](https://github.com/apache/airflow/blob/858df9991d0e88d881418bb304bbf7a243bdd985/airflow-core/src/airflow/api_fastapi/core_api/security.py#L246-L251)

So a manager is asked to materialize every authorized dag id in the deployment purely so one `IN (...)` clause can be built. The set is never used for anything else.

Two properties make that expensive:

1. **It is deployment-wide, not page-wide.** `get_authorized_dag_ids` reads every dag row with no limit, groups by team, then calls `filter_authorized_dag_ids` per team ([`base_auth_manager.py#L643-L666`](https://github.com/apache/airflow/blob/858df9991d0e88d881418bb304bbf7a243bdd985/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py#L643-L666)).
2. **It resolves before pagination exists.** It is a FastAPI dependency ([`security.py#L332-L343`](https://github.com/apache/airflow/blob/858df9991d0e88d881418bb304bbf7a243bdd985/airflow-core/src/airflow/api_fastapi/core_api/security.py#L332-L343)), so limit and offset are applied after the authorized set is built.

A manager whose policy lives in the database cannot say "join against my table". The return type cannot express it.

### Who this hits

It is already reported well below large-deployment scale:

- #61686 — Keycloak, **600 dags**, ~25s responses. Closed after per-manager caching and connection pooling; the reporter noted a real fix "is quite complicated to implement".
- #69041 — multi-team, hundreds of dags across 10 teams, ~10s `/dags` load.
- #70582 — bulk authorization methods added to the Keycloak manager.

Every fix so far has been inside one manager, so each new manager rediscovers the problem.

It also makes the documented answer expensive. #23638 asked for dag permissions by tag or owner rather than by name, and the Airflow 3 answer is to write an auth manager over `get_authorized_dag_ids`, with tags and bundles as the suggested attributes. Those are rows in the metadata database, and `get_db_manager` exists so a manager can add tables of its own — but none of it can be expressed as a query.

### Proposal

Let a manager return a select instead of a set, with today's behaviour as the default:

```python
def get_authorized_dag_ids_select(self, *, user, method="GET") -> Select | None:
    """A select producing the dag ids this user may access.

    Returning None, the default, keeps the existing behaviour: core calls
    get_authorized_dag_ids and builds an IN clause from the result.
    """
    return None
```

Core applies it as `dag_id IN (subquery)`, so filtering and pagination happen in one statement. Managers that do not override it are untouched.

### What this does and does not fix

**Fixes:** any manager whose policy is in the metadata database — a tag- or bundle-based manager, one with its own table via `get_db_manager`, and FAB, which currently enumerates when its permission tables are joinable.

**Does not fix:** managers backed by an external PDP (Amazon Verified Permissions, Keycloak Authorization Services). Their policy is not SQL, so they still enumerate, and #61686 would not be solved by this. Those need either a reverse-lookup API on the PDP ("which resources may this principal access?") or a local projection of the policy, both outside Airflow.

This is the database-backed half of the problem, which is the half Airflow can fix on its own.

### Measured

A tag-based manager returning `select(DagTag.dag_id).where(DagTag.name == ...)`, against MySQL 8 with 41,606 dags. Page of 50, median of 15 rounds:

| Tag the manager authorizes on | Matching dags | Materialize | Subquery |
|---|---|---|---|
| Two team tags | 418 | 5.4 ms | 1.4 ms |
| An environment tag | 37,446 | 316.1 ms | 0.8 ms |

FAB on the same database, 1,610 roles and 66,529 per-dag grants: 351.8 ms → 0.6 ms for a user authorized on all dags, which is any role with a global `can_read`.

Implemented in #71341.

1fanwang:
Closes: #71309

## Why

`get_authorized_dag_ids` is where every list endpoint starts, and its default implementation reads
the whole dag table before any manager is consulted, groups the rows by team, then calls
`filter_authorized_dag_ids` once per team. A request for 50 dags loads every dag id in the
deployment, so the cost scales with the deployment rather than the page.

That is core's own path, not one provider's. `FabAuthManager` overrides it and still reads every
row. Keycloak inherits it and then calls out per dag, which
[#69041](https://github.com/apache/airflow/issues/69041) and
[#61686](https://github.com/apache/airflow/issues/61686) report as ten and twenty-five second
pages.

It also makes the documented answer expensive.
[#23638](https://github.com/apache/airflow/issues/23638) asked for dag permissions by tag or owner
rather than by name, and Airflow 3 answers that by writing an auth manager over
`get_authorized_dag_ids`. Tags, bundles and teams are all rows in this database, and
`get_db_manager` lets a manager add its own. The `set[str]` return type is what stops any of them
being answered as a query.

## What changed

`BaseAuthManager` gains `get_authorized_dag_ids_select`, returning a select of dag ids or `None`.
`None` is the default and keeps today's behaviour, so managers backed by an external policy service
(Keycloak, Amazon Verified Permissions) are unaffected.

A returned select is applied as `dag_id IN (subquery)`. The dependency graph services and the
run-state counts endpoint still need ids in memory, and still get them: `PermittedDagFilter`
materializes on first read.

A select replaces `get_authorized_dag_ids` outright, including its per-team grouping, so a
multi-team manager scopes the select itself or returns `None` and keeps the fan-out.

`FabAuthManager` implements it with the grant query it already knows how to write.

## Testing Done

**End to end against a running API server.** A 40-line manager authorizing by tag, no FAB. Six
dags seeded, three tagged `team-a`.

```
$ curl -H "Authorization: Bearer $TOKEN" localhost:28080/api/v2/dags
total: 3
  dag_01 [team-a]
  dag_02 [team-a]
  dag_05 [team-a]

$ curl -H "Authorization: Bearer $TOKEN" 'localhost:28080/api/v2/dags?limit=2'
limit=2 -> ['dag_01', 'dag_02'] total 3
```

The statement that server executed, captured with a `before_cursor_execute` listener:

```sql
... LEFT OUTER JOIN dag_run ON dag_run.id = mrq.max_dag_run_id
WHERE dag.dag_id IN (SELECT dag_tag.dag_id FROM dag_tag WHERE dag_tag.name = ?)
  AND dag.is_stale != 1
ORDER BY dag.dag_id ASC LIMIT ? OFFSET ?
```

A subquery rather than bind parameters, `LIMIT` in the same statement, and the same subquery
reaching the latest-run join.

**Measured.** The same shape at scale. MySQL 8, 41,606 dags with an environment tag and a team
tag. Page of 50, median of 15 rounds.

| Tag the manager authorizes on | Matching dags | Materialize (today) | Subquery |
|---|---|---|---|
| Two team tags | 418 | 5.4 ms | 1.4 ms |
| An environment tag | 37,446 | 316.1 ms | 0.8 ms |

**Measured, FAB.** Same database at 41,606 dags, 1,610 roles, 66,529 per-dag edit grants.

| Grant | Materialize (today) | Subquery |
|---|---|---|
| Per-dag edit on 228 dags | 3.7 ms | 2.3 ms |
| Authorized on all 41,606 dags | 351.8 ms | 0.6 ms |

The expensive row in both is the ordinary one: a tag most dags carry, or a role with global
`can_read`, which is Viewer and up.

**Regression tests.** Eight, covering the `None` default, a select reaching the SQL as a subquery,
an empty select honoured as "nothing is permitted", a subclass filter inheriting it, a team-scoped
select, and the id set materializing once and only when read. Reverting the fix:

```
$ pytest airflow-core/tests/unit/api_fastapi/core_api/test_security.py \
    -k PermittedDagFilterSubquery -q
E   TypeError: Boolean value of this clause is not defined
4 failed, 2 passed, 111 deselected
```

Restored:

```
$ pytest airflow-core/tests/unit/api_fastapi/core_api/test_security.py \
    -k PermittedDagFilterSubquery -q
8 passed, 111 deselected

$ pytest airflow-core/tests/unit/api_fastapi/auth/managers/test_base_auth_manager.py -q
61 passed

$ pytest airflow-core/tests/unit/api_fastapi/common -q
162 passed, 6 skipped
```

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
+++ b/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
@@ -58,7 +58,7 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
 
     from fastapi import FastAPI
-    from sqlalchemy import Row
+    from sqlalchemy import Row, Select
     from sqlalchemy.orm import Session
     from starlette.middleware import _MiddlewareFactory
 
@@ -624,6 +624,33 @@ class BaseAuthManager(Generic[T], LoggingMixin, metaclass=ABCMeta):
             )
 
         return {conn_id for conn_id in conn_ids if _is_authorized_connection(conn_id)}
+
+    def get_authorized_dag_ids_select(
+        self,
+        *,
+        user: T,
+        method: ResourceMethod = "GET",
+    ) -> Select | None:
+        """
+        Get a select of the Dag ids the user has access to, to be used as a subquery.
+
+        Returning ``None``, the default, makes the API materialize the whole authorized set
+        through :meth:`get_authorized_dag_ids` instead. Auth managers that keep their grants in
+        the Airflow metadata database can return a select here, which the API applies as
+        ``dag_id IN (subquery)`` so that filtering and pagination happen in one statement rather
+        than after every authorized Dag id has been loaded into memory.
+
+        The select must produce a single column of Dag ids.
+
+        Returning a select replaces :meth:`get_authorized_dag_ids` entirely, including its
+        per-team grouping, so a multi-team deployment has to scope the select itself. Join
+        ``DagBundleModel`` and ``dag_bundle_team_association_table`` the way
+        :meth:`get_authorized_dag_ids` does, or return ``None`` and keep the default fan-out.
+
+        :param user: the user
+        :param method: the method to filter on
+        """
+        return None
 
     @provide_session
     def get_authorized_dag_ids(
--- a/airflow-core/src/airflow/api_fastapi/common/db/dags.py
+++ b/airflow-core/src/airflow/api_fastapi/common/db/dags.py
@@ -37,28 +37,31 @@ if TYPE_CHECKING:
 
 
 def generate_dag_with_latest_run_query(
-    max_run_filters: list[BaseParam], order_by: SortParam, *, dag_ids: set[str] | None = None
+    max_run_filters: list[BaseParam],
+    order_by: SortParam,
+    *,
+    dag_ids: set[str] | Select | None = None,
 ) -> Select:
     """
     Generate a query to fetch Dags with their latest run.
 
     :param max_run_filters: List of filters to apply to the latest run
     :param order_by: Sort parameter for ordering results
-    :param dag_ids: Optional set of Dag IDs to limit the query to. When provided, both the main
-        Dag query and the subquery for finding the latest runs will be filtered to
-        only these Dag IDs, improving performance when users have limited Dag access.
+    :param dag_ids: Optional set of Dag IDs, or a select producing them, to limit the query
+        to. When provided, both the main Dag query and the subquery for finding the latest
+        runs are filtered to those Dag IDs.
     :return: SQLAlchemy Select statement
     """
     query = select(DagModel).options(selectinload(DagModel.tags))
 
     # Filter main query by dag_ids if provided
     if dag_ids is not None:
-        query = query.where(DagModel.dag_id.in_(dag_ids or set()))
+        query = query.where(DagModel.dag_id.in_(dag_ids))
 
     # Also filter the subquery for finding latest runs
     max_run_id_query_stmt = select(DagRun.dag_id, func.max(DagRun.id).label("max_dag_run_id"))
     if dag_ids is not None:
-        max_run_id_query_stmt = max_run_id_query_stmt.where(DagRun.dag_id.in_(dag_ids or set()))
+        max_run_id_query_stmt = max_run_id_query_stmt.where(DagRun.dag_id.in_(dag_ids))
     max_run_id_query = max_run_id_query_stmt.group_by(DagRun.dag_id).subquery(name="mrq")
 
     has_max_run_filter = False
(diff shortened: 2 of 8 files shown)


vincbeck (changes requested):
I understand the motivation but I am against introducing a new method in `base_auth_manager` class that is only applicable to Fab. Fab auth manager is the only auth manager using the Db to manage permissions, so `get_authorized_dag_ids_select` is Fab specific. Ideally, we should get rid of Fab, and not add Fab specific features in Airflow core.

1fanwang:
> I understand the motivation but I am against introducing a new method in `base_auth_manager` class that is only applicable to Fab. Fab auth manager is the only auth manager using the Db to manage permissions, so `get_authorized_dag_ids_select` is Fab specific. Ideally, we should get rid of Fab, and not add Fab specific features in Airflow core.

Agreed, and thanks for the context on the direction. @vincbeck 

I checked and confirmed there's also no easy way to update this in FAB only. `PermittedDagFilter` is bound to `get_authorized_dag_ids` in core, and the `set[str]` return is what forces the materialization. Not worth a core hook for the one manager that's on the path to deprecation

Closing.

PR #71342, opened 2026-08-09
Fix Grid view failing to load large Dags on MySQL

WonYong-Jang:
Grid view showed no tasks for large Dags when the metadata backend is MySQL.
The API server logged:

```
ERROR 1038: Out of sort memory, consider increasing server sort buffer size
```

### Root cause

The grid endpoint used `_get_latest_serdag` to load the latest serialized Dag.
That query did `select(SerializedDagModel) ... ORDER BY ... LIMIT 1`, which loads the whole row - including the big `data` / `data_compressed` blob columns.

On MySQL, `ORDER BY` uses filesort, and filesort copies each candidate row (the blob too) into the sort buffer. The sort buffer has a fixed size (`sort_buffer_size`, 256 KB by default). When a single serialized Dag is bigger than that, the sort fails with "Out of sort memory", so the whole request fails and the Grid view renders nothing.

PostgreSQL is not affected because it does not copy large values into the sort buffer.

### Fix

Use the existing `SerializedDagModel.latest_item_select_object` helper.
On MySQL it first finds the latest row's `id` (ORDER BY on `id` only, no blob), and then loads that single row by primary key. The blob is never put into the sort buffer, so the error is gone.

---

Generated-by:  Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
```diff
@@ -84,14 +84,12 @@
...
-            SerializedDagModel.dag_id == dag_id,
-        )
-        .order_by(SerializedDagModel.id.desc())
-        .limit(1)
-    )
+    # Reuse latest_item_select_object: selecting the whole SerializedDagModel pulls the large
+    # data/data_compressed blob columns, and on MySQL the ORDER BY makes filesort load each
+    # candidate row (blob included) into the sort buffer, overflowing it with "Out of sort memory"
+    # for large serialized Dags.
+    # See https://github.com/apache/airflow/pull/55589
```
uranusjr: Comment offers no value; deletion is recommended.

uranusjr:
This slightly changes tie-breaking logic; previously the ordering was by id, now it’s by timestamp. They _should_ agree since the id is UUIDv7, so I’m just adding this as a hint for debugging, in case we hit an edge case in the future.

uranusjr (approved):

airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py
WonYong-Jang: Agreed. The rationable is already documented inside `latest_item_select_object`, so the comment here was redundant - removed it. Done.

henry3260 (approved):

PR #71343, opened 2026-08-09
Move template-field validation out of GoogleCampaignManagerDeleteRepo…

bujjibabukatta:
Part of #70296.

## Summary

Moves `report_name` / `report_id` validation in `GoogleCampaignManagerDeleteReportOperator`
from `__init__` into `execute()`, and removes the class from the
`validate-operators-init` exemption list.

## Root cause

`report_name` and `report_id` are template fields, which are rendered *after* the
constructor runs. Validating them in `__init__` therefore checks the un-rendered
Jinja expressions instead of the actual values — the reason this class was on the
exemption list.

## Fix

- **`campaign_manager.py`** — moved the "provide `report_name` or `report_id`" and
  "only one of them" checks from `__init__` to the start of `execute()`; `__init__`
  now only does plain `self.field = field` assignments.
- **`validate_operators_init_exemptions.txt`** — removed the
  `GoogleCampaignManagerDeleteReportOperator` entry (required in the same PR; the
  hook fails on a stale exemption once the class is compliant).
- **`test_campaign_manager.py`** — added tests asserting both invalid cases (both
  provided / neither provided) now raise from `execute()`.

- [X] Yes - Claude


potiuk (approved):

scripts/ci/prek/validate_operators_init_exemptions.txt
```diff
@@ -19,5 +19,5 @@ providers/google/src/airflow/providers/google/cloud/sensors/bigquery_dts.py::Big
 providers/google/src/airflow/providers/google/cloud/sensors/cloud_composer.py::CloudComposerExternalTaskSensor
 providers/google/src/airflow/providers/google/cloud/transfers/azure_fileshare_to_gcs.py::AzureFileShareToGCSOperator
 providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py::GCSToBigQueryOperator
-providers/google/src/airflow/providers/google/marketing_platform/operators/campaign_manager.py::GoogleCampaignManagerDeleteReportOperator
+providers/microsoft/azure/src/airflow/providers/microsoft/azure/transfers/gcs_to_wasb.py::GCSToAzureBlobStorageOperator
```
aaron-y-chen: Why do we need to add this line here?

bujjibabukatta:
@aaron-y-chen Good catch — you're right, that shouldn't be there. It came in through an
earlier merge with main from before someone else's PR had already cleaned it
up upstream. Rebasing onto current main now to drop it — this PR should only
touch the GoogleCampaignManagerDeleteReportOperator line.

bujjibabukatta:
Hi @potiuk, could you please merge the PR when you get a chance? Thanks for approving it!

aaron-y-chen:
This PR should be closed since #70530 has been merged.

PR #59239, opened 2025-12-09
Enable triggerer queues

zach-overflow:
---

Closes: https://github.com/apache/airflow/issues/33818
Partially addresses / related: https://github.com/apache/airflow/issues/59871

## What?
* Adds optional support for task-based trigger queue assignment, which allows a given triggerer to consume from a user-configured subset of triggers, based on their origin tasks' `queue` value.
    * Adds an optional `--queues` CLI option to the triggerer, to constrain a triggerer to only consume Triggers from a given list of task queues.
    * Adds a new `triggerer.queues_enabled` config entry, which when set to `True` enables the trigger queue feature (`False` by default).
    * Adds a `queue` column to the `trigger` table, which is set automatically by the task runner when `triggerer.queues_enabled` is set to True.
* Adds relevant documentation updates for this new feature.

## Why?
* AIP-67 (multi-tenancy support), to enable team-specific triggerer assignment.
* More generally, and not necessarily specific to multi-tenancy use-cases, to allow for other triggerer assignment strategies in an HA setup.

## Testing
* Tested the example HITL dag in docker-compose; once with `triggerer.queues_enabled=False` and once with `triggerer.queues_enabled=True` to ensure deferrable tasks worked in either setup.
*  standard collection of `breeze` and `prek` testing commands
* Ran `uv run --group docs build-docs --autobuild` to check the new doc section rendering.

## Note on event-based triggers and async Callbacks
Following from the helpful discussion [here](https://github.com/apache/airflow/pull/59239#discussion_r2645289748), this PR aims to assign a trigger's `queue` from its deferring task's `queue`. As such, this PR does not add any triggerer assignment controls for event-based triggers or `TriggererCallback`-based triggers. This was done intentionally both to keep this PR's scope reasonable, and because the latter 2 types of triggers do not necessarily have a relation to a singular task / task queue. If it is preferred, I'm happy to create a followup issue for adding queue support in these types of triggers. I've also made sure to make this explicit in the user documentation, and added details in the docs on how to use this feature without disrupting any potential usage of event-based or callback-based triggers.

PR #71346, opened 2026-08-09
Add trigger `queue` support for `AsyncCallback` and `BaseEventTrigger`

zach-overflow:
## What?
* Adds trigger queue support for the `BaseEventTrigger` and the `AsyncCallback` categories of triggers.
* Allows users to control the triggerer host assignment for all 3 categories of triggers (task-initiated, event-based, and callback-initiated).

## Why?
* The trigger queues feature was [first introduced](https://github.com/apache/airflow/pull/59239) in Airflow `3.2.0`, but only for task-based triggers. This was because at the time, the `BaseEventTrigger` and `AsyncCallback` patterns were in the middle of development.
* Trigger queues provide triggerer host assignment controls distinct from, and complementary to, the multi-team triggerer assignment. Since the multi-team assignment supports all three types of triggers, trigger queues should too.

---

* Generated-by: Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)
* Scope of AI use:
    * I used Claude for tracing and modifying the trigger record creation pathways for event-based triggers / AsyncCallback triggers after I added the initial queue assignment to the base classes. 
    * Documentation was hand-written
    * tests were 50% AI written, followed by manual cleanup.

---

### DIFF
--- a/airflow-core/docs/authoring-and-scheduling/deferring.rst
+++ b/airflow-core/docs/authoring-and-scheduling/deferring.rst
@@ -493,6 +493,8 @@ According to `benchmarks <https://github.com/apache/airflow/pull/58803#pullreque
 
 You can determine a suitable value for your deployment by creating a large number of triggers (for example, by triggering a Dag with many deferrable tasks) and observing both how the load is distributed across Triggerers in your environment and how long it takes for all Triggerers to pick up the triggers.
 
+.. _deferring/triggerer_queue_assignment:
+
 Controlling Triggerer Host Assignment Per Trigger
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -508,7 +510,7 @@ Under some circumstances, it may be desirable to assign a Trigger to a specific 
     If you are using :doc:`Multi-Team mode</core-concepts/multi-team>`, the ``--team-name`` option provides
     native team-scoped triggerer assignment for all trigger types (task-created, event-driven, and callback).
     See :ref:`Team-scoped Triggerer <multi-team-triggerer>` in the Multi-Team documentation.
-    The ``--queues`` option described below is an older, queue-based mechanism that can be combined with
+    The ``--queues`` option described below is a distinct queue-based mechanism which may be combined with
     ``--team-name`` if needed.
 
 To enable queue assignment for triggers, do the following:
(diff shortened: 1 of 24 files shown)


airflow-core/docs/authoring-and-scheduling/deferring.rst
zach-overflow, line 1: Screenshots of locally rendered doc server below:

Image

Image

Image

airflow-core/docs/howto/deadline-alerts.rst
zach-overflow, line 1: screenshot of locally rendered doc server output below:

Image

task-sdk/src/airflow/sdk/definitions/callback.py
```diff
@@ -151,16 +151,31 @@ class AsyncCallback(Callback):
...
 
-    def __init__(self, callback_callable: Callable | str, kwargs: dict | None = None):
+    queue: str | None
+
+    def __init__(
+        self,
+        callback_callable: Callable | str,
+        kwargs: dict | None = None,
+        queue: str | None = None,
+    ):
```
uranusjr, line 167: Let’s make the new argument keyword-only. (I would suggest others to be changed as well, but that brings unnecessary compatibility issues.)

airflow-core/src/airflow/models/trigger.py
```diff
@@ -200,7 +200,7 @@ def rotate_fernet_key(self):
     def from_object(cls, trigger: BaseTrigger) -> Trigger:
         """Alternative constructor that creates a trigger row based directly off of a Trigger object."""
         classpath, kwargs = trigger.serialize()
-        return cls(classpath=classpath, kwargs=kwargs)
+        return cls(classpath=classpath, kwargs=kwargs, queue=getattr(trigger, "queue", None))
```
uranusjr: I think the getattr is unnecessarily defensive.

uranusjr (approved):
With nits

airflow-core/src/airflow/models/trigger.py
zach-overflow: Good call. Addressed in https://github.com/apache/airflow/pull/71346/commits/a93a0777808482bfbb55cd9527874ffb3afc6926

task-sdk/src/airflow/sdk/definitions/callback.py
zach-overflow: 👍  done in https://github.com/apache/airflow/pull/71346/commits/a93a0777808482bfbb55cd9527874ffb3afc6926

zach-overflow:
@uranusjr Thanks for the review! I believe I've addressed everything, but please let me know if you have any other feedback.

uranusjr (approved):
What do we do with the argument collision? As this stands, the queue cannot actually be set by the user, can it?

zach-overflow:
> What do we do with the argument collision? As this stands, the queue cannot actually be set by the user, can it?

@uranusjr Fair point, I think I was a bit overzealous when I saw that the `MessageQueueTrigger` already had the `queue` parameter marked as deprecated. After some adjustments, I think I found a decent middle ground until whenever that deprecated parameter is fully removed: For the `MessageQueueTrigger` the `queue` parameter's meaning remains the same as before, while it's set to a private instance attribute instead (`_queue`) so we there's no collision with the BaseTrigger's `queue`.

Users can then assign specifically the `MessageQueueTrigger` class to a "trigger queue" with the `trigger_queue` parameter. A little ugly, but I'm ok with it. I'm also ok with fully replacing the meaning of the `queue` value in that class so everything is consistent. No strong feelings either way on my end.

airflow-core/newsfragments/71346.feature.rst
hussein-awala, line 1: not needed, you can delete it, it's only needed in significant changes

zach-overflow:
cc @uranusjr is this good to merge? Should I get any other reviewers eyes on this? Please let me know if there's anyone in particular I might want to tag here if so, and I'll be happy to do so 👍

hussein-awala (approved):
LGTM once the newsfragment is removed

airflow-core/newsfragments/71346.feature.rst
zach-overflow: done 👍

PR #67822, opened 2026-06-01
Fix scheduler orphaned task reset logging crash

Revanth14:
---
Fixes: #67813

The scheduler orphaned-task adoption path loaded task instances with a narrow `load_only(...)` set and then called `repr(ti)` while building the reset log message. `TaskInstance.__repr__` reads `map_index` and `state`, so detached
instances with deferred columns could raise `DetachedInstanceError` and crash the scheduler on restart.

This updates the scheduler query to load the fields needed by `repr(ti)` and makes `TaskInstance.__repr__` defensive for detached/deleted ORM instances used in logging.

Tests:
- `uv run --project airflow-core pytest airflow-core/tests/unit/models/test_taskinstance.py::test_task_instance_repr_does_not_raise_for_deferred_columns airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_adopt_or_reset_orphaned_tasks_loads_state_for_reset_logging -q`
  - 2 passed

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (Opus 4.8)

PR #67850, opened 2026-06-01
Fix scheduler crashloop from `KubernetesExecutor` completed-pod adoption

kaxil:
## Summary

Follow-up to #67822. Fixes the root cause of #67813: a scheduler `DetachedInstanceError` crashloop on startup when the `KubernetesExecutor` is in use.

## Root cause

`SchedulerJobRunner.adopt_or_reset_orphaned_tasks` selects orphaned TaskInstances under `FOR UPDATE SKIP LOCKED` (`with_row_locks(..., skip_locked=True)`) and, in the same open transaction, hands them to `executor.try_adopt_task_instances`. For the `KubernetesExecutor` that path reaches `_alive_other_scheduler_job_ids`, which opened a **scoped** `create_session()`.

Executors run in the scheduler thread, so a scoped session resolves to the scheduler's own in-flight session (`settings.Session` is a thread-local `scoped_session`). The `create_session` context manager runs `commit()` then `close()` on exit, so on return it:

- commits the scheduler's transaction early, releasing the `FOR UPDATE SKIP LOCKED` row locks, and
- closes the session, detaching the orphaned TaskInstances the scheduler still holds.

The reset loop then touches those detached instances (`repr(ti)`, then `prepare_db_for_next_try` -> `TaskInstanceHistory.record_ti`, then the `state = None` writes) and raises `DetachedInstanceError`. The triggering rows persist, so every restart re-crashes (CrashLoopBackOff); with HA, all schedulers die.

#67822 hardened the `repr()` read, but the crash recurs one statement later in `record_ti` (it reads `try_number` and copies every column off the same detached instance via `getattr`), and the premature commit / lock release was not addressed. This PR fixes the cause.

## How the regression got here

`_adopt_completed_pods` is reachable from two call sites, added years apart:

- #35579 (`cncf-kubernetes` 7.10.0, Nov 2023) added the call from inside `try_adopt_task_instances`, the locked adoption path.
- #61839 (`cncf-kubernetes` 10.15.0, Mar 2026) added a second, periodic call from `KubernetesExecutor.sync()`, which has no caller transaction.

Through both of those, `_adopt_completed_pods` was **database-free**: it only built a pod label selector and patched pods via the kube client, so running it inside the scheduler's locked transaction was harmless.

[...]

PR #66400, opened 2026-05-05
KubernetesExecutor: scope periodic completed-pod adoption to dead schedulers

potiuk:
## Summary

Fix for [#66396](https://github.com/apache/airflow/issues/66396). PR #61839
(cncf-kubernetes 10.15.0) added a periodic call to `_adopt_completed_pods`
from inside `KubernetesExecutor.sync()`, gated by
`[scheduler] orphaned_tasks_check_interval` (default 300 s). The query
selects every Succeeded pod whose `airflow-worker` label is **not** the
current scheduler's label and PATCHes it with the current scheduler's label
so its `KubernetesJobWatcher` will see the change and DELETE the pod.

With multi-scheduler deployments that caused thrashing — every interval
tick each scheduler relabeled every other scheduler's completed pods,
fighting over ownership and burning kube-API + watcher cycles. See
[#66396](https://github.com/apache/airflow/issues/66396) for the full
walkthrough and a real user report on the mailing list (3 schedulers,
Airflow 3.2.1, 1000+ pods/min, tasks stalling in `scheduled` / `queued`,
`delete_worker_pods=False` not helping).

## What changed

- New helper **`_alive_other_scheduler_job_ids`** queries the `Job` table
  for `SchedulerJob`s whose `state == RUNNING` and whose
  `latest_heartbeat` is within
  `[scheduler] scheduler_health_check_threshold` — matching the
  alive-scheduler definition already used by
  `SchedulerJobRunner.adopt_or_reset_orphaned_tasks`. Returns an empty
  set on any DB error so the caller falls back to the pre-#61839
  "exclude self only" selector — a transient DB issue must not break
  completed-pod cleanup.
- **`_adopt_completed_pods`** now builds the label selector to exclude
  `self` + every alive sibling using K8s set-based syntax:
  `airflow-worker notin (a,b,c)`. With one scheduler the selector is
  identical to the pre-fix equality form, so single-scheduler
  deployments see no behavior change.

## Why this preserves the original #61839 goal

#61839 closed [#57553](https://github.com/apache/airflow/issues/57553):
"completed pods leaking after scheduler restart". The fix here keeps
that working — when a scheduler dies, its `Job` row's
`latest_heartbeat` ages out of the threshold, the helper stops
including its ID in the alive set, and the surviving schedulers will
adopt and clean up its completed pods on the next interval tick. The
only behavior removed is the steady-state cross-scheduler relabeling
that caused the thrash.

## Test plan

[...]

PR #71348, opened 2026-08-09
Fix scheduler crash when resetting orphaned task instances

### LINKED ISSUE #71272: Scheduler crashes with DetachedInstanceError in `adopt_or_reset_orphaned_tasks` — `repr(ti)` lazy-loads `state` on a detached TaskInstance

Airflow Core

### Apache Airflow version

3.2.2

### What happened
The scheduler enters an unrecoverable crash loop. On startup and on every
`[scheduler] orphaned_tasks_check_interval`, `adopt_or_reset_orphaned_tasks()`
builds a log message by calling `repr(ti)` on each orphaned `TaskInstance`.
`TaskInstance.__repr__` reads `self.state`, which is a deferred column. When the
`TaskInstance` is detached from its SQLAlchemy session, that read triggers a
deferred DB load with no bound session and raises `DetachedInstanceError`. The
exception is unhandled in `_run_scheduler_loop`, so the scheduler process exits.

Because the scheduler dies *before* it finishes resetting the orphaned task, the
same orphan is still present on the next start and it crashes again — a
self-perpetuating loop. Nothing outside the scheduler clears the orphan, so it
never recovers without manual DB intervention. While it loops, no DAGs are
scheduled at all.

Traceback (paths are from the official `apache/airflow` image):

```
[error ] Exception when executing SchedulerJob._run_scheduler_loop
  File ".../airflow/jobs/scheduler_job_runner.py", line 1546, in _execute
    ret = execute_callable()
  File ".../airflow/jobs/scheduler_job_runner.py", line 1612, in _run_scheduler_loop
    self.adopt_or_reset_orphaned_tasks()
  File ".../airflow/utils/session.py", line 100, in wrapper
    return func(*args, session=session, **kwargs)
  File ".../airflow/jobs/scheduler_job_runner.py", line 2814, in adopt_or_reset_orphaned_tasks
    for attempt in run_with_db_retries(logger=self.log):
  ...
  File ".../airflow/jobs/scheduler_job_runner.py", line 2864, in adopt_or_reset_orphaned_tasks
    reset_tis_message.append(repr(ti))
  File ".../airflow/models/taskinstance.py", line 1134, in __repr__
    return prefix + f"[{self.state}] ti_id={self.id}>"
  File ".../sqlalchemy/orm/attributes.py", line 569, in __get__
    return self.impl.get(state, dict_)
  ...
  File ".../sqlalchemy/orm/strategies.py", line 536, in _load_for_state
    raise orm_exc.DetachedInstanceError(
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <TaskInstance at 0x...>
is not bound to a Session; deferred load operation of attribute 'state' cannot
proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)

### How to reproduce

Using KubernetesExecutor:

1. Start a task so a `TaskInstance` is `running`/`queued` under a worker pod.
2. Cause it to become "orphaned" — e.g. the worker pod is lost, or the
   scheduler is killed while the task is in flight (SIGKILL / OOM / node drain)
   so the owning job is gone.
3. Start the scheduler. `adopt_or_reset_orphaned_tasks()` selects the orphaned
   `TaskInstance`(s); when it reaches `reset_tis_message.append(repr(ti))` the
   deferred `state` load fails with `DetachedInstanceError` and the scheduler
   exits. It then crash-loops on every restart because the orphan persists.

Manual workaround: set the stuck task instances to a terminal state directly in
the metadata DB (or via the UI "Mark Failed"/CLI from a non-scheduler pod), then
restart the scheduler.

Adopting/resetting orphaned tasks must not be able to crash the scheduler, and
in particular building a diagnostic log line must never trigger a lazy DB load
on a possibly-detached instance. `adopt_or_reset_orphaned_tasks()` should either:

- construct the message from the task instance identity/key (already-loaded
  columns) instead of `repr(ti)`, or
- make `TaskInstance.__repr__` read only already-loaded attributes (e.g. from
  `__dict__`) so it degrades gracefully (`state=?`) rather than raising, or
- keep the instances bound/refreshed within the session while the message is
  built.

The task instances should still be reset as intended; only the fatal log call
needs to change.

### Operating System

Debian (official `apache/airflow:slim-3.2.2-python3.12` image), Python 3.12.

### Deployment

Other

_No response_

apache-airflow-providers-cncf-kubernetes (KubernetesExecutor). Reproducible with
the version pinned by the 3.2.2 constraints file.


Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

Happens repeatedly — every scheduler restart while an orphaned task instance
exists. Related (same code path / theme, but not the same `__repr__`/`state`
lazy-load crash): #30709, #39088, #19671, #58570, #13808.

### Code of Conduct

akhilpratap1991:
`adopt_or_reset_orphaned_tasks()` reads deferred columns (the full row copied into `TaskInstanceHistory` by `prepare_db_for_next_try`, `last_heartbeat_at`) and the lazy `dag_run` relationship on TaskInstances that executors' `try_adopt_task_instances()` may hand back detached from the session. Those reads raise `DetachedInstanceError`, which is unhandled in `_run_scheduler_loop` — the scheduler exits, and since it dies before the orphan is reset, the same orphan crashes it again on every restart: an unrecoverable crash loop that halts all scheduling (observed in production with KubernetesExecutor as a scheduler `CrashLoopBackOff`). Where the attributes happen to be loaded, the writes to the detached instances are silently lost instead, so the orphans are never actually reset.

closes: #71272

This completes #67822, which fixed the first such read (`repr(ti)` on `state`/`map_index`) by widening `load_only` — but `TaskInstanceHistory(ti)` copies ~35 columns, so widening the column list further just moves the crash. Instead, re-select the rows to mutate by their already-loaded ids (with `dag_run` joined-loaded) so both the reset and adopt branches operate on fully loaded, session-bound instances, independent of what the executor did with the objects it was handed.

The regression test simulates the executor detaching the TaskInstances it was handed (as observed in production): without this change it fails with the production `DetachedInstanceError` (deferred load of `try_number`); with it, the orphan is reset with a `TaskInstanceHistory` audit row and the adopted TI gets its heartbeat/dag_run handling.

---


uranusjr:
Static checks need fixing.

The description asserts this was "observed in production with KubernetesExecutor as a scheduler CrashLoopBackOff," but the test only reproduces it by manually calling `session.expunge()` in a mock. I think we should identify the actual, real offending code path that triggers the issue you described to decide if the fix is a good idea.

akhilpratap1991:
Static checks are green on the current head — the initial run was sitting in the first-time-contributor approval gate (`action_required`), not failing; after approval, `CI image checks / Static checks` passed.

On the real code path — the concrete trigger is an executor opening and closing a **scoped** session while adopting:

1. `KubernetesExecutor.try_adopt_task_instances()` → completed-pod adoption scoping (added in #66400) runs `with create_session():` — in released providers up to **10.17.x**, which is what Airflow 3.2.2 constraints pin.
2. `create_session()` defaults to the **thread-scoped** registry (`settings.Session`), so it returns the very session `adopt_or_reset_orphaned_tasks()` is holding, and its `finally: session.close()` **expunges everything the orphan query loaded**.
3. Back in the scheduler loop, `prepare_db_for_next_try()` → `TaskInstanceHistory.record_ti()` reads columns deferred by the `load_only` (plus `last_heartbeat_at` / `dag_run.conf` in the adopt branch) on now-detached instances → `DetachedInstanceError`, unhandled in `_run_scheduler_loop`. This is exactly the production incident we hit (3.2.2 + cncf provider 10.17.1, scheduler CrashLoopBackOff): the scheduler dies before the reset flushes, so the same orphan kills every restart. Where the attributes happen to be loaded, the resets are **silently lost** instead — the loop "succeeds" without persisting anything.

#67850 already fixed that one call site provider-side (`scoped=False`, released in 10.18.0), and #67822 fixed the first read (`repr`). Both are point fixes: executors are separately released (and can be third-party), and any scoped-session close during adoption re-creates the condition — so this PR makes the scheduler loop itself robust by operating on re-selected, session-bound rows.

I've updated the test accordingly: the executor mock now performs the exact operation the released provider does (`with create_session(): pass`) instead of manually expunging. Without this PR's change it fails with the production `DetachedInstanceError` (deferred load of `try_number`); with it, it passes.

---
Drafted-by: Claude Code (Opus 4.8); reviewed by @akhilpratap1991 before posting

PR #71349, opened 2026-08-09
Reuse async Kubernetes API client across trigger polls

xvega:
I've been chasing triggerer issues on our production deployment: sustained `triggers.blocked_main_thread` bursts whenever a wave of deferrable `KubernetesPodOperator`/`GKEStartPodOperator` tasks started or finished, with the async thread blocked 2–10s at a time. #69661 helped the log-parsing path, but the metric barely moved. After reading the source and profiling the TriggerRunner with py-spy during blocking windows, the loop thread showed up inside `ssl.create_default_context()` reached from `AsyncKubernetesHook.get_conn()`: every API method builds a fresh `ApiClient` per call, which parses the CA bundle synchronously on the event loop and opens a new aiohttp session, a full TCP+TLS handshake per status poll (default 2s per trigger) with zero reuse. The GKE hook additionally fetched a new OAuth token and wrote the cluster CA to a temp file each call.

The change itself is simple: instead of building a new API client for every call, the hook now builds one and keeps it for its whole lifetime. For in-cluster and static-auth kubeconfigs Airflow loads the configuration once per hook, and anything that does rotate (the in-cluster service-account token) is read from the configuration per request rather than baked into the client, so one client can serve the hook's lifetime. Exec-based auth (EKS/GKE kubeconfigs) is the exception and keeps the old per-call behavior on purpose, reloading the config on every call is exactly what refreshes its short-lived credentials, so those clients can't be reused. To pair every cached client with an owner, hooks get an idempotent `close()` and the triggers call it from `cleanup()`, so the client lives exactly as long as its trigger. The GKE hook gets the same treatment plus token caching: it re-sets the bearer header every time the connection is used, so tokens refresh on schedule instead of being fetched from Google on every call (and since all that was left of its `_load_config()` was building the client, it's now called `_build_client()`).

A few things follow from the caching that are worth knowing. If you instantiate these hooks yourself or subclass the triggers and override `cleanup()`, call `hook.close()` when you're done, otherwise the client is only cleaned up at garbage collection and aiohttp logs an "unclosed session" warning; calling the hook again after `close()` just builds a fresh client. mTLS client certs from static kubeconfigs are now read once per hook rather than on every call, which I think is fine for hooks that live only as long as a trigger (if cert rotation mid-trigger ever becomes a real problem, rebuilding the client on auth errors would be the fix). The pre-existing gap left by #65212 the default-kubeconfig path sets `_config_loaded` without exec-auth detection, is untouched. And one expectation to set: each new trigger still builds its one client on the event loop, so `blocked_main_thread` should drop dramatically but won't be exactly zero when a batch of triggers starts.

To make sure this was really the problem, I reproduced it locally on a minikube cluster: real pod triggers polling real pods, with Airflow's own `block_watchdog` running, counting how many SSL contexts and API clients got created. I ran the exact same load twice, once with the old per-call `get_conn` patched back in as a baseline, once with this change. The baseline created a new client (and SSL context) for every single API call; with this change each trigger creates exactly one and releases it on cleanup. Unit tests in both providers assert the same behavior.

related: #69661

**No**

### DIFF
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/hooks/kubernetes.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/hooks/kubernetes.py
@@ -840,6 +840,7 @@ class AsyncKubernetesHook(KubernetesHook):
         # Cached result of exec-auth detection. None means not yet detected.
         # This is to optimise and not calling _uses_exec_auth repeatedly on every _load_config() call.
         self._is_exec_auth: bool | None = None
+        self._cached_kube_client: async_client.ApiClient | None = None
 
     def _uses_exec_auth(self, kubeconfig_data: dict, context: str | None = None) -> bool:
         """
@@ -1024,14 +1025,29 @@ class AsyncKubernetesHook(KubernetesHook):
 
     @contextlib.asynccontextmanager
     async def get_conn(self) -> AsyncGenerator[async_client.ApiClient, None]:
-        kube_client = None
+        await self._load_config()
+        if self._config_loaded:
+            # Reuse one client per hook: each construction runs ssl.create_default_context()
+            # on the event loop and opens a new connection pool. Owners release it via
+            # close(); triggers do so in cleanup().
+            if self._cached_kube_client is None:
+                # No await between check and assign, or the loser's client leaks.
+                self._cached_kube_client = _TimeoutAsyncK8sApiClient(configuration=self.client_configuration)
+            yield self._cached_kube_client
+            return
+        # Exec-based auth rotates short-lived tokens by reloading the config on
+        # every call, so the client cannot be reused.
+        kube_client = _TimeoutAsyncK8sApiClient(configuration=self.client_configuration)
         try:
-            await self._load_config()
-            kube_client = _TimeoutAsyncK8sApiClient(configuration=self.client_configuration)
             yield kube_client
         finally:
-            if kube_client is not None:
-                await kube_client.close()
+            await kube_client.close()
+
+    async def close(self) -> None:
+        """Release the cached API client, if any. Safe to call multiple times."""
+        cached, self._cached_kube_client = self._cached_kube_client, None
+        if cached is not None:
+            await cached.close()
 
     @generic_api_retry
     async def get_pod(self, name: str, namespace: str) -> V1Pod:
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/hooks/kubernetes.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/hooks/kubernetes.py
@@ -1031,7 +1031,6 @@ class AsyncKubernetesHook(KubernetesHook):
             # on the event loop and opens a new connection pool. Owners release it via
             # close(); triggers do so in cleanup().
             if self._cached_kube_client is None:
-                # No await between check and assign, or the loser's client leaks.
                 self._cached_kube_client = _TimeoutAsyncK8sApiClient(configuration=self.client_configuration)
             yield self._cached_kube_client
             return
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/triggers/pod.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/triggers/pod.py
@@ -538,46 +538,54 @@ class KubernetesPodTrigger(BaseTrigger):
         On Airflow 3.3.0+ pod deletion on user kill is handled in ``on_kill()`` only; this avoids
         deleting pods on triggerer restart. On older Airflow versions, ``cleanup()`` still uses
         ``safe_to_cancel()`` because ``on_kill()`` is not wired for user kills.
+
+        Always releases the hook's cached API client, since the triggerer invokes ``cleanup()``
+        after the trigger finishes for any reason (event fired, cancelled, or killed).
         """
-        # TODO: Remove this Airflow < 3.3 cleanup branch (early return, ``safe_to_cancel``, and
-        # related tests) once the minimum Airflow version supported by this provider is >= 3.3.
-        # In Airflow 3.3+, ``BaseTrigger.on_kill()`` handles user-initiated kills; keeping the
-        # legacy path for backward compatibility with older Airflow versions.
-        if AIRFLOW_V_3_3_PLUS:
-            return
-
-        if self._fired_event:
-            self.log.debug("Skipping cleanup since an event has already been fired.")
-            return
-
-        if self.on_kill_action == OnKillAction.KEEP_POD:
-            self.log.debug("Skipping cleanup since on_kill_action is set to %r.", self.on_kill_action.value)
-            return
-
         try:
-            safe = await self.safe_to_cancel()
-        except Exception:
-            self.log.warning(
-                "Could not determine task state during cleanup; skipping pod deletion to be safe.",
-                exc_info=True,
-            )
-            return
+            # TODO: Remove this Airflow < 3.3 cleanup branch (early return, ``safe_to_cancel``, and
+            # related tests) once the minimum Airflow version supported by this provider is >= 3.3.
+            # In Airflow 3.3+, ``BaseTrigger.on_kill()`` handles user-initiated kills; keeping the
+            # legacy path for backward compatibility with older Airflow versions.
+            if AIRFLOW_V_3_3_PLUS:
+                return
 
-        if not safe:
-            self.log.debug(
-                "Skipping cleanup since the task is still in deferred state (likely a triggerer restart)."
-            )
-            return
+            if self._fired_event:
+                self.log.debug("Skipping cleanup since an event has already been fired.")
+                return
 
-        self.log.info("Deleting pod %s in namespace %s.", self.pod_name, self.pod_namespace)
-        try:
-            await self.hook.delete_pod(
-                name=self.pod_name,
-                namespace=self.pod_namespace,
-                grace_period_seconds=self.termination_grace_period,
-            )
-        except Exception:
-            self.log.exception("Unexpected error while deleting pod %s", self.pod_name)
+            if self.on_kill_action == OnKillAction.KEEP_POD:
+                self.log.debug(
+                    "Skipping cleanup since on_kill_action is set to %r.", self.on_kill_action.value
+                )
+                return
+
+            try:
+                safe = await self.safe_to_cancel()
+            except Exception:
+                self.log.warning(
+                    "Could not determine task state during cleanup; skipping pod deletion to be safe.",
+                    exc_info=True,
+                )
+                return
+
+            if not safe:
+                self.log.debug(
+                    "Skipping cleanup since the task is still in deferred state (likely a triggerer restart)."
+                )
+                return
+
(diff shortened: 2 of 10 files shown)


providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/hooks/kubernetes.py
```diff
@@ -1024,14 +1025,29 @@ async def _get_field(self, field_name):
...
     @contextlib.asynccontextmanager
     async def get_conn(self) -> AsyncGenerator[async_client.ApiClient, None]:
-        kube_client = None
+        await self._load_config()
+        if self._config_loaded:
+            # Reuse one client per hook: each construction runs ssl.create_default_context()
+            # on the event loop and opens a new connection pool. Owners release it via
+            # close(); triggers do so in cleanup().
+            if self._cached_kube_client is None:
+                # No await between check and assign, or the loser's client leaks.
```
Miretpl:
```suggestion
```

providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/triggers/pod.py
```diff
@@ -536,46 +536,54 @@ async def cleanup(self) -> None:
...
-        if AIRFLOW_V_3_3_PLUS:
-            return
-
-        if self._fired_event:
-            self.log.debug("Skipping cleanup since an event has already been fired.")
-            return
-
-        if self.on_kill_action == OnKillAction.KEEP_POD:
-            self.log.debug("Skipping cleanup since on_kill_action is set to %r.", self.on_kill_action.value)
-            return
```
Miretpl: Why are we moving this under the `try` clause?

providers/google/src/airflow/providers/google/cloud/hooks/kubernetes_engine.py
```diff
@@ -514,13 +517,22 @@ def __init__(
...
     @contextlib.asynccontextmanager
     async def get_conn(self) -> AsyncGenerator[async_client.ApiClient, None]:
-        kube_client = None
-        try:
-            kube_client = await self._load_config()
-            yield kube_client
-        finally:
-            if kube_client is not None:
-                await kube_client.close()
+        # Reuse one client per hook; see AsyncKubernetesHook.get_conn for the rationale.
```
Miretpl: IMHO it should be rather copied here

Miretpl (changes requested):
Could you adjust your PR description to match our project guidelines - https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst?

Miretpl:
I would appreciate a review from a Kubernetes Google maintainer on that part of the code.

xvega:
> Could you adjust your PR description to match our project guidelines - https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst?

Besides the AI disclosure block what's missing or not part of the guideline?

providers/google/src/airflow/providers/google/cloud/hooks/kubernetes_engine.py
xvega: Addressed

Miretpl:
> Besides the AI disclosure block what's missing or not part of the guideline?

Only this was missing here. Thanks for adding it. I've rerun the whole CI as there were a couple of unrelated failures. Let's see how it will be now.

Miretpl (approved):

soupam05:
Since the ApiClient is now reused for the lifetime of the trigger, can we explicitly document how credential refresh is handled independently of the client lifecycle? In particular, I'm wondering about long-running triggers where credentials may expire while the client is still alive.

soupam05:
What happens if client initialization succeeds but a later poll fails due to an authentication/configuration error? Should that error cause the cached client to be invalidated so that the next poll can rebuild it, or is the expectation that the trigger will terminate/retry? It would be good to define this now that the client is no longer recreated on every poll.

xvega:
> Since the ApiClient is now reused for the lifetime of the trigger, can we explicitly document how credential refresh is handled independently of the client lifecycle? In particular, I'm wondering about long-running triggers where credentials may expire while the client is still alive.

Credential refresh is handled separately from the cached client:
- In-cluster service-account tokens refresh through the Kubernetes configuration’s token-refresh callback.
- When exec auth is detected, the hook still reloads the config and creates a client on each call.
- For GKE, each get_conn() checks the cached token and updates the Authorization header when it refreshes.
The caveat is static mTLS certificates: those are now loaded once per client, so certificate rotation during a running trigger won’t be picked up automatically. That tradeoff is noted in the PR description.

xvega:
> What happens if client initialization succeeds but a later poll fails due to an authentication/configuration error? Should that error cause the cached client to be invalidated so that the next poll can rebuild it, or is the expectation that the trigger will terminate/retry? It would be good to define this now that the client is no longer recreated on every poll.

The cached client isn’t invalidated when a poll fails. Existing retries still handle transient errors, but authentication errors such as 401/403 cause the pod trigger to report an error, and cleanup closes the client. Whether the task retries afterward depends on its retry settings.
I’d keep that behavior for this PR unless we identify a case where client reuse requires additional recovery. Did you have a particular authentication setup or failure scenario in mind? That would help determine whether rebuilding the client would actually resolve it.

providers/google/src/airflow/providers/google/cloud/hooks/kubernetes_engine.py
```diff
@@ -514,13 +517,21 @@ def __init__(
...
-            yield kube_client
-        finally:
-            if kube_client is not None:
-                await kube_client.close()
+        # Reuse one client per hook: construction parses the CA into an SSL context on
+        # the event loop and opens a new connection pool. Released via close()/cleanup().
+        if self._cached_kube_client is None:
+            self._cached_kube_client = self._build_client()
+        if self._cached_token is None:
+            self._cached_token = await self.get_token()
```
aaron-y-chen, line 525: I wonder whether caching this token could cause a long-running GKE trigger to use expired credentials. The metadata server may return a cached token with less than 3600 seconds remaining, but `_CredentialsToken.refresh()` always records its lifetime as 3600 seconds, so `get()` may wait 1800 seconds before refreshing it.

For example: the metadata server returns a token with 600 seconds remaining -> this wrapper caches it -> 601 seconds later `get()` still considers it fresh -> the next K8s API poll sends the expired token and receives 401.

Could we address this and add a regression test covering this case?

xvega: You’re right. I reproduced this and added a  fix that uses credentials.expiry instead of assuming a 3600 seconds lifetime. The existing refresh policy now follows the token’s actual remaining lifetime.

Also added regression tests for the 600-second case

soupam05:
> > What happens if client initialization succeeds but a later poll fails due to an authentication/configuration error? Should that error cause the cached client to be invalidated so that the next poll can rebuild it, or is the expectation that the trigger will terminate/retry? It would be good to define this now that the client is no longer recreated on every poll.
> 
> The cached client isn’t invalidated when a poll fails. Existing retries still handle transient errors, but authentication errors such as 401/403 cause the pod trigger to report an error, and cleanup closes the client. Whether the task retries afterward depends on its retry settings. I’d keep that behavior for this PR unless we identify a case where client reuse requires additional recovery. Did you have a particular authentication setup or failure scenario in mind? That would help determine whether rebuilding the client would actually resolve it.

Thanks, that clarifies the 401/403 handling. One follow-up: for a long-running trigger, if a refreshable credential expires and the Kubernetes API returns a 401, can we guarantee that the credential-refresh path is attempted before the 401 is treated as terminal? Ideally, an expected credential refresh should be transparent to the task rather than causing the trigger to fail and relying on a task-level retry.

xvega:
> Thanks, that clarifies the 401/403 handling. One follow-up: for a long-running trigger, if a refreshable credential expires and the Kubernetes API returns a 401, can we guarantee that the credential-refresh path is attempted before the 401 is treated as terminal? Ideally, an expected credential refresh should be transparent to the task rather than causing the trigger to fail and relying on a task-level retry.

We couldn’t guarantee that before. The regression test I added showed that retries could reuse the rejected token, so I also added a fix. The GKE client now refreshes credentials and retries once after a 401, using the same client. If the refresh or retry fails, the error still propagates.

aaron-y-chen (commented):
Overall, the Google provider parts LGTM. However, it now relies on `_cached_kube_client` and `close()`, which are added to `AsyncKubernetesHook` in this PR. Without marking this cross-provider dependency, the Google provider could still be installed with an older Kubernetes provider and fail with an `AttributeError` on the first GKE trigger poll.

Should we add `# use next version` at here?

https://github.com/apache/airflow/blob/b4d012936c5abe5f6b9b68eff085f98f30018370/providers/google/pyproject.toml#L166

xvega:
> Overall, the Google provider parts LGTM. However, it now relies on `_cached_kube_client` and `close()`, which are added to `AsyncKubernetesHook` in this PR. Without marking this cross-provider dependency, the Google provider could still be installed with an older Kubernetes provider and fail with an `AttributeError` on the first GKE trigger poll.
> 
> Should we add `# use next version` at here?
> 
> https://github.com/apache/airflow/blob/b4d012936c5abe5f6b9b68eff085f98f30018370/providers/google/pyproject.toml#L166

I added # use next version 👍

aaron-y-chen (approved):
Kubernetes Google part LGTM.

providers/google/src/airflow/providers/google/cloud/hooks/kubernetes_engine.py
```diff
@@ -478,6 +482,27 @@ def apply_from_yaml_file(
...
+        refresh_token: Callable[[], Awaitable[str | None]],
+    ) -> None:
+        super().__init__(configuration=configuration)
+        self._refresh_token = refresh_token
+
+    async def call_api(self, *args: Any, **kwargs: Any) -> Any:
+        try:
+            return await super().call_api(*args, **kwargs)
+        except async_client.ApiException as error:
+            if error.status != 401:
```
Miretpl, line 500: We can receive a 401 code for more than just an expired token, so I don't think that this handling is correct.

Miretpl (changes requested):
Looking at the changes in the code, I'm reverting my approve to prevent accidental merge of it. I would recommend separate this PR to cncf and google provider - it should be easier to merge at least one of them.

providers/google/src/airflow/providers/google/cloud/hooks/kubernetes_engine.py
xvega: How about following google-auth’s refresh-on-401 approach, capped at one attempt? I’d retry only if the token changes and propagate any refresh or retry failure. This preserves the refresh attempt requested by @soupam05

xvega:
> Looking at the changes in the code, I'm reverting my approve to prevent accidental merge of it. I would recommend separate this PR to cncf and google provider - it should be easier to merge at least one of them.

@Miretpl  I opened https://github.com/apache/airflow/pull/73399 for the cncf changes. The Google changes depend on them, so I’ll keep them here until that PR merges, then rebase this PR to leave only the Google changes.

Miretpl:
> @Miretpl I opened https://github.com/apache/airflow/pull/73399 for the cncf changes. The Google changes depend on them, so I’ll keep them here until that PR merges, then rebase this PR to leave only the Google changes.

Cool, thanks! So let's focus on the CNCF Kubernetes PR first.

PR #71350, opened 2026-08-10
Add Azure Analysis Services model refresh support

### LINKED ISSUE #51377: Azure Virtual Analysis Services Operator
### Description

A couple of weeks ago, we have to work with a DAG that allow us to refresh Models in Azure Analysis Services without using Data Factory .
We think that would be good to have an Operator+Hook+Sensor that abstract all this process to deal with authentication and the usage of the [REST API.](https://learn.microsoft.com/pt-br/analysis-services/azure-analysis-services/analysis-services-async-refresh?view=asallproducts-allversions)

### Use case/motivation

An operator to allow the interactions (refresh/get_status) with Models in Azure Analysis Services

### Related issues

_No response_

### Code of Conduct

aaron-y-chen:
closes: #51377

## Summary

- Add an `azure_analysis_services` connection type using Microsoft Entra service-principal client-secret authentication.
- Add a Hook, Operator, Sensor, and Trigger for starting and monitoring model refreshes in synchronous, fire-and-forget, and deferrable modes.
- Add request and polling timeouts, response validation, diagnostic HTTP errors, provider metadata, documentation, and a system-test Dag.

## Testing

Run the system-test Dag in Breeze against a live Azure Analysis Services instance:

- Azure Analysis Services D1 Developer tier in West US
- Compatibility-level-1200 model
- Dedicated Microsoft Entra service principal and client-secret Airflow connection
- Two `calculate` refreshes

The Dag runs a fire-and-forget Operator, a deferrable-enabled Sensor, and a deferrable-enabled Operator. All three task instances complete successfully on their first try.

Verify both refresh IDs returned by Airflow through the Azure Analysis Services REST API:

| Refresh ID | Type | Status | Duration |
|---|---|---|---:|
| `437aa86c-…-8169e` | `calculate` | `succeeded` | 0.206 s |
| `d9f4974e-…-9020` | `calculate` | `succeeded` | 0.179 s |

The Azure screenshot confirms the D1 resource, West US location, and compatibility-level-1200 model.

<br>

---

Generated-by: [GPT 5.6-sol] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/microsoft/azure/pyproject.toml
+++ b/providers/microsoft/azure/pyproject.toml
@@ -97,6 +97,7 @@ dependencies = [
     # It was added in https://github.com/apache/airflow/pull/47990/files
     # maybe this should be set from upstream
     "msal-extensions>=1.3.0",
+    "requests>=2.32.0,<3",
 ]
 
 # The optional dependencies should be modified in place in the generated file
--- a/providers/microsoft/azure/pyproject.toml
+++ b/providers/microsoft/azure/pyproject.toml
@@ -99,7 +99,6 @@ dependencies = [
     # It was added in https://github.com/apache/airflow/pull/47990/files
     # maybe this should be set from upstream
     "msal-extensions>=1.3.0",
-    "requests>=2.32.0,<3",
 ]
 
 # The optional dependencies should be modified in place in the generated file
--- a/providers/microsoft/azure/pyproject.toml
+++ b/providers/microsoft/azure/pyproject.toml
@@ -99,6 +99,7 @@ dependencies = [
     # It was added in https://github.com/apache/airflow/pull/47990/files
     # maybe this should be set from upstream
     "msal-extensions>=1.3.0",
+    "requests>=2.32.0,<3",
 ]
 
 # The optional dependencies should be modified in place in the generated file
--- a/providers/microsoft/azure/pyproject.toml
+++ b/providers/microsoft/azure/pyproject.toml
@@ -97,7 +97,6 @@ dependencies = [
     # It was added in https://github.com/apache/airflow/pull/47990/files
     # maybe this should be set from upstream
     "msal-extensions>=1.3.0",
-    "requests>=2.32.0,<3",
 ]
 
 # The optional dependencies should be modified in place in the generated file
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
@@ -21,13 +21,15 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Any, Literal, get_args
 from urllib.parse import quote, unquote, urlsplit
 
-import requests
+import httpx
 from azure.core.exceptions import AzureError
 from azure.identity import ClientSecretCredential
 
 from airflow.providers.common.compat.sdk import AirflowException, BaseHook
 
 if TYPE_CHECKING:
+    from urllib.parse import SplitResult
+
     from azure.core.credentials import TokenCredential
 
     from airflow.sdk import Connection
@@ -38,8 +40,10 @@ RefreshType = Literal["full", "clearValues", "calculate", "dataOnly", "automatic
 VALID_REFRESH_TYPES: frozenset[str] = frozenset(get_args(RefreshType))
 
 
-def _format_request_error(error: requests.RequestException) -> str:
-    response_body = getattr(error.response, "text", "").strip()[:1000]
+def _format_request_error(error: httpx.HTTPError) -> str:
+    # Only HTTPStatusError carries a response; transport errors have no such attribute.
+    response = getattr(error, "response", None)
+    response_body = getattr(response, "text", "").strip()[:1000]
     response_detail = f"; response body: {response_body}" if response_body else ""
     return f"{error}{response_detail}"
 
@@ -147,13 +151,14 @@ class AzureAnalysisServicesHook(BaseHook):
         """Return the validated status of an Azure Analysis Services model refresh."""
         refresh_url = f"{self._get_refreshes_url(server_name, database)}/{quote(refresh_id, safe='')}"
         try:
-            response = requests.get(
-                refresh_url,
-                headers=self._get_headers(),
-                timeout=self.request_timeout,
-            )
-            response.raise_for_status()
-        except requests.RequestException as error:
+            with httpx.Client() as client:
+                response = client.get(
+                    refresh_url,
+                    headers=self._get_headers(),
+                    timeout=self.request_timeout,
+                )
+                response.raise_for_status()
+        except httpx.HTTPError as error:
             raise AzureAnalysisServicesRefreshException(
                 f"Failed to get status for Azure Analysis Services refresh {refresh_id}: "
                 f"{_format_request_error(error)}"
@@ -185,14 +190,15 @@ class AzureAnalysisServicesHook(BaseHook):
             )
 
         try:
-            response = requests.post(
-                self._get_refreshes_url(server_name, database),
-                json={"Type": refresh_type},
-                headers=self._get_headers(),
-                timeout=self.request_timeout,
-            )
-            response.raise_for_status()
-        except requests.RequestException as error:
+            with httpx.Client() as client:
+                response = client.post(
+                    self._get_refreshes_url(server_name, database),
+                    json={"Type": refresh_type},
+                    headers=self._get_headers(),
+                    timeout=self.request_timeout,
+                )
+                response.raise_for_status()
+        except httpx.HTTPError as error:
             raise AzureAnalysisServicesRefreshException(
                 f"Failed to trigger an Azure Analysis Services model refresh: {_format_request_error(error)}"
             ) from error
@@ -246,20 +252,25 @@ class AzureAnalysisServicesHook(BaseHook):
                 )
             time.sleep(min(check_interval, remaining))
 
-    def _get_base_url(self) -> str:
-        host = (self.connection.host or "").strip().rstrip("/")
-        parsed_host = urlsplit(f"//{host}")
+    def _assert_host(self, host: str, parsed_host: SplitResult) -> None:
+        # netloc, not .username/.port: those miss "@host" and ":0", and raise on ":abc".
         if (
             not host
             or not parsed_host.hostname
+            or "@" in parsed_host.netloc
+            or ":" in parsed_host.netloc
             or parsed_host.path
             or parsed_host.query
             or parsed_host.fragment
         ):
             raise ValueError(
-                "A valid region endpoint without a URL scheme or path is required in the "
-                "Azure Analysis Services connection host"
+                "A valid region endpoint without a URL scheme, credentials, port, or path is "
+                "required in the Azure Analysis Services connection host"
             )
+
+    def _get_base_url(self) -> str:
+        host = (self.connection.host or "").strip().rstrip("/")
+        self._assert_host(host, urlsplit(f"//{host}"))
         return f"https://{host}"
 
     def _get_headers(self) -> dict[str, str]:
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
@@ -1,0 +1,285 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+
+import time
+from functools import cached_property
+from typing import TYPE_CHECKING, Any, Literal, get_args
+from urllib.parse import quote, unquote, urlsplit
+
+import requests
+from azure.core.exceptions import AzureError
+from azure.identity import ClientSecretCredential
+
+from airflow.providers.common.compat.sdk import AirflowException, BaseHook
+
+if TYPE_CHECKING:
+    from azure.core.credentials import TokenCredential
+
+    from airflow.sdk import Connection
+
+TOKEN_SCOPE = "https://*.asazure.windows.net/.default"
+
+RefreshType = Literal["full", "clearValues", "calculate", "dataOnly", "automatic", "defragment"]
+VALID_REFRESH_TYPES: frozenset[str] = frozenset(get_args(RefreshType))
+
+
+def _format_request_error(error: requests.RequestException) -> str:
+    response_body = getattr(error.response, "text", "").strip()[:1000]
+    response_detail = f"; response body: {response_body}" if response_body else ""
+    return f"{error}{response_detail}"
+
+
+class AzureAnalysisServicesRefreshStatus:
+    """Azure Analysis Services model refresh statuses."""
+
+    SUCCEEDED = "succeeded"
+    FAILED = "failed"
+    CANCELLED = "cancelled"
+    TIMED_OUT = "timedOut"
+    NOT_STARTED = "notStarted"
+    IN_PROGRESS = "inProgress"
+
+    FAILURE_STATUSES = frozenset({FAILED, CANCELLED, TIMED_OUT})
+    VALID_STATUSES = frozenset({SUCCEEDED, FAILED, CANCELLED, TIMED_OUT, NOT_STARTED, IN_PROGRESS})
+
+
+class AzureAnalysisServicesRefreshException(AirflowException):
+    """Indicate that an Azure Analysis Services model refresh operation failed."""
+
+
+class AzureAnalysisServicesHook(BaseHook):
+    """
+    Interact with the Azure Analysis Services asynchronous refresh REST API.
+
+    :param azure_analysis_services_conn_id: The Azure Analysis Services connection ID.
+    :param request_timeout: Timeout in seconds for each HTTP request.
+
+    The connection must define the region endpoint in ``host``, the service principal client ID in
+    ``login``, the client secret in ``password``, and the Microsoft Entra tenant ID in the
+    ``tenantId`` extra field.
+    """
+
+    conn_type: str = "azure_analysis_services"
+    conn_name_attr: str = "azure_analysis_services_conn_id"
+    default_conn_name: str = "azure_analysis_services_default"
+    hook_name: str = "Azure Analysis Services"
+
+    def __init__(
+        self,
+        azure_analysis_services_conn_id: str = default_conn_name,
+        request_timeout: float = 60,
+    ) -> None:
+        super().__init__()
+        if request_timeout <= 0:
+            raise ValueError("request_timeout must be greater than zero")
+        self.azure_analysis_services_conn_id = azure_analysis_services_conn_id
+        self.request_timeout = request_timeout
+        self._credential: TokenCredential | None = None
+
+    @cached_property
+    def connection(self) -> Connection:
+        """Return the Azure Analysis Services connection."""
+        return self.get_connection(self.azure_analysis_services_conn_id)
+
+    @classmethod
+    def get_connection_form_widgets(cls) -> dict[str, Any]:
+        """Return connection widgets to add to the connection form."""
+        from flask_appbuilder.fieldwidgets import BS3TextFieldWidget
+        from flask_babel import lazy_gettext
+        from wtforms import StringField
+
+        return {
+            "tenantId": StringField(lazy_gettext("Tenant ID"), widget=BS3TextFieldWidget()),
+        }
+
+    @classmethod
+    def get_ui_field_behaviour(cls) -> dict[str, Any]:
+        """Return custom field behaviour for the connection form."""
+        return {
+            "hidden_fields": ["schema", "port", "extra"],
+            "relabeling": {
+                "host": "Region Endpoint",
+                "login": "Client ID",
+                "password": "Client Secret",
+            },
+            "placeholders": {
+                "host": "westus.asazure.windows.net",
+            },
+        }
+
+    def get_conn(self) -> TokenCredential:
+        """Return and cache the service principal credential."""
+        if self._credential is not None:
+            return self._credential
+
+        connection = self.connection
+        tenant_id = connection.extra_dejson.get("tenantId")
+        if not connection.login:
+            raise ValueError("Client ID is required for Azure Analysis Services authentication")
+        if not connection.password:
+            raise ValueError("Client secret is required for Azure Analysis Services authentication")
+        if not isinstance(tenant_id, str) or not tenant_id:
+            raise ValueError("Tenant ID is required for Azure Analysis Services authentication")
+
+        self._credential = ClientSecretCredential(
+            tenant_id=tenant_id,
+            client_id=connection.login,
+            client_secret=connection.password,
+        )
+        return self._credential
+
+    def get_refresh_status(self, server_name: str, database: str, refresh_id: str) -> str:
+        """Return the validated status of an Azure Analysis Services model refresh."""
+        refresh_url = f"{self._get_refreshes_url(server_name, database)}/{quote(refresh_id, safe='')}"
+        try:
+            response = requests.get(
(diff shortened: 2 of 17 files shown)


providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
```diff
@@ -0,0 +1,285 @@
...
+                raise AzureAnalysisServicesRefreshException(
+                    f"Timeout waiting for Azure Analysis Services refresh {refresh_id} to complete"
+                )
+            time.sleep(min(check_interval, remaining))
+
+    def _get_base_url(self) -> str:
+        host = (self.connection.host or "").strip().rstrip("/")
+        parsed_host = urlsplit(f"//{host}")
+        if (
+            not host
```
dabla, line 262: **[warning]** `_get_base_url` does not validate `userinfo` in the host string — bearer token can be sent to an attacker-controlled domain.

The validation guards against `path`, `query`, and `fragment` but omits `parsed_host.userinfo`. An input like `attacker@evil.example` passes every check (`hostname` is `evil.example`, all other parts are empty) and the method returns `https://attacker@evil.example`. HTTP clients will connect to `evil.example` and send the AAS bearer token there.

A port component is similarly unchecked: `evil.example:9999` passes and the returned URL targets a non-standard port.

Suggested fix — add both guards to the condition:

```python
if (
    not host
    or not parsed_host.hostname
    or parsed_host.userinfo   # <-- add this
    or parsed_host.port       # <-- add this
    or parsed_host.path
    or parsed_host.query
    or parsed_host.fragment
):
    raise ValueError(...)
```

**[nit]** While fixing the above, consider extracting the validation into a dedicated `_assert_host` method. `_get_base_url` would then read as a clean two-liner, and the host validation can be unit-tested in isolation without going through `_get_base_url`:

```python
def _assert_host(self, host: str, parsed: SplitResult) -> None:
    if (
        not host
        or not parsed.hostname
        or parsed.userinfo
        or parsed.port
        or parsed.path
        or parsed.query
        or parsed.fragment
    ):
        raise ValueError(
            "A valid region endpoint without a URL scheme or path is required in the "
            "Azure Analysis Services connection host"
        )

def _get_base_url(self) -> str:
    host = (self.connection.host or "").strip().rstrip("/")
    self._assert_host(host, urlsplit(f"//{host}"))
    return f"https://{host}"
```

Open for debate — the method is small enough that it is not strictly necessary, but it makes the intent and the test surface explicit.

---
Drafted-by: Claude Sonnet 4.6 (claude-sonnet-4.6); reviewed by @dabla before posting

providers/microsoft/azure/pyproject.toml
```diff
@@ -99,6 +99,7 @@ dependencies = [
     # It was added in https://github.com/apache/airflow/pull/47990/files
     # maybe this should be set from upstream
     "msal-extensions>=1.3.0",
+    "requests>=2.32.0,<3",
```
dabla: **[blocker]** `requests` must not be added as a new dependency -- `httpx` is already a transitive dependency of the provider via `kiota-http` and is the established HTTP client across the provider.

Beyond the dependency concern, switching to `httpx.AsyncClient` eliminates the `asyncio.to_thread` workaround in the trigger entirely: `httpx` has native async support, so `get_refresh_status` and `trigger_refresh` can be made proper `async` methods on the hook, and the trigger can `await` them directly.

**`pyproject.toml`** -- remove the added line:
```diff
-    "requests>=2.32.0,<3",
```

**`hooks/analysis_services.py`** -- swap to `httpx.AsyncClient` and make the methods async:
```diff
-import requests
+import httpx
```

```diff
-    def get_refresh_status(self, server_name: str, database: str, refresh_id: str) -> dict:
+    async def get_refresh_status(self, server_name: str, database: str, refresh_id: str) -> dict:
         try:
-            response = requests.get(refresh_url, headers=self._get_headers(), timeout=self.request_timeout)
+            async with httpx.AsyncClient() as client:
+                response = await client.get(refresh_url, headers=self._get_headers(), timeout=self.request_timeout)
             response.raise_for_status()
-        except requests.RequestException as error:
+        except httpx.HTTPError as error:
             raise AzureAnalysisServicesRefreshError(...) from error
```

```diff
-    def trigger_refresh(self, server_name: str, database: str, refresh_type: str) -> dict:
+    async def trigger_refresh(self, server_name: str, database: str, refresh_type: str) -> dict:
         try:
-            response = requests.post(url, json={"Type": refresh_type}, headers=self._get_headers(), timeout=self.request_timeout)
+            async with httpx.AsyncClient() as client:
+                response = await client.post(url, json={"Type": refresh_type}, headers=self._get_headers(), timeout=self.request_timeout)
             response.raise_for_status()
-        except requests.RequestException as error:
+        except httpx.HTTPError as error:
             raise AzureAnalysisServicesRefreshError(...) from error
```

**`triggers/analysis_services.py`** -- drop `asyncio.to_thread` and await directly:
```diff
-import asyncio
 ...
-            result = await asyncio.to_thread(self.hook.get_refresh_status, ...)
+            result = await self.hook.get_refresh_status(...)
```

The `response.raise_for_status()`, `response.json()`, `response.headers`, and `response.text` APIs are identical in `httpx`. Test mocks will need updating accordingly (`mock.AsyncMock(spec=httpx.Response)`).

---
Drafted-by: Claude Sonnet 4.6 (claude-sonnet-4.6); reviewed by @dabla before posting

providers/microsoft/azure/tests/unit/microsoft/azure/hooks/test_analysis_services.py
```diff
@@ -0,0 +1,388 @@
...
+        credential_class.return_value.get_token.assert_called_once_with(TOKEN_SCOPE)
+        assert TOKEN_SCOPE == "https://*.asazure.windows.net/.default"
+        assert headers == HEADERS
+
+    @mock.patch(f"{MODULE}.ClientSecretCredential", autospec=True)
+    def test_get_headers_wraps_authentication_errors(self, credential_class):
+        credential_class.return_value.get_token.side_effect = ClientAuthenticationError("bad credential")
+
+        with pytest.raises(AzureAnalysisServicesRefreshException, match="Failed to authenticate"):
+            AzureAnalysisServicesHook(CONN_ID)._get_headers()
```
dabla: **[nit]** `test_rejects_invalid_region_endpoint` does not cover the `userinfo` and port injection cases identified in comment [1].

The existing parametrize values are `[None, "", "https://westus.asazure.windows.net", "host/path"]`. Adding `"attacker@evil.example"` and `"evil.example:9999"` would both fail today (they pass `_get_base_url` without raising), confirming the gap and ensuring it stays fixed after any future refactor.

```python
@pytest.mark.parametrize(
    "host",
    [
        None,
        "",
        "https://westus.asazure.windows.net",
        "host/path",
        "attacker@evil.example",   # userinfo injection
        "evil.example:9999",       # non-standard port
    ],
)
def test_rejects_invalid_region_endpoint(self, host, create_mock_connection):
    ...
```

---
Drafted-by: Claude Sonnet 4.6 (claude-sonnet-4.6); reviewed by @dabla before posting

dabla (changes requested):
The PR adds a well-structured set of components (Hook, Operator, Sensor, Trigger) for Azure Analysis Services model refreshes. The architecture is sound: the trigger correctly offloads synchronous HTTP calls to `asyncio.to_thread`, deferrable mode is properly implemented, and the test suite is thorough.

Two items must be fixed before merge:

1. **`_get_base_url` allows `userinfo` and port injection** ([CWE-918](https://cwe.mitre.org/data/definitions/918.html)). The host validation explicitly guards against path/query/fragment but silently accepts `user@host` (userinfo) and `host:port` inputs, which can redirect the AAS bearer token to an attacker-controlled server. Adding `or parsed_host.userinfo or parsed_host.port` to the guard closes the gap. A test case should also be added.

2. **`requests` must not be added as a dependency.** `httpx` is already provided by the provider's transitive dependency on `kiota-http` and is the established HTTP client in this provider. The two `requests.get` / `requests.post` calls are straightforward to replace with the `httpx` sync API — see the inline comment for the exact diff.

---
Drafted-by: Claude Sonnet 4.6 (claude-sonnet-4.6); reviewed by @dabla before posting

providers/microsoft/azure/tests/unit/microsoft/azure/hooks/test_analysis_services.py
aaron-y-chen: Nice suggestion, thanks! I added additional cases like `host:abc` and `evil.example:0`. I hope this makes the tests more comprehensive than before :)

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: Fixed, and I adopted the `_assert_host` method. I agree that it's clearer.

providers/microsoft/azure/pyproject.toml
aaron-y-chen: Oh, if `httpx` can be used instead of `requests`, then there's indeed no need to add a new dependency. Let me try to refactor it. 👀

aaron-y-chen: Refactor done, and passed the e2e test:

image

image

I kept `asyncio.to_thread()` because `get_refresh_status()` and `ClientSecretCredential.get_token()` are synchronous and are also used by the Operator, Sensor, and Hook polling paths. Offloading the call keeps the triggerer event loop non-blocking without duplicating the Hook API, and follows the existing pattern in `triggers/batch.py`.

dabla: Maybe you're not aware but you have also an async variant of ClientSecretCredential, but if we would go that route that would mean we probably need a dedicated async hook, so would leave it with asyncio.to_thread as you mentioned.  Could be an idea for a follow up PR in the future.

dabla: @aaron-y-chen I'm just wondering if the operator couldn't be async/deferred only like the MSGraphAsyncOperator?  That would mean the hook could be purely written as a native hook, polling and blocking a worker isn't the best approach anyway, hence why the MSGraphAsyncOperator is purely deferred as well.  If so, that would mean the code would become even more simpler, especially on the operator side and the hook would be purely async and no more blocking polling in the worker but forcing usage of trigger mechanism, which is the purpose of it, triggers are great at polling without blocking the workers.  WDYT?

Thank you for the requested changes, beside above remark, the PR is looking good an clean to me, but before merging I wanted to address above point and see what your opinion is about it?

aaron-y-chen: Wow, this is my first time learning that there's a purely async operator in Airflow. After doing some survey, I think AAS refresh is a good fit for this approach since it avoids occupying a worker slot while polling. It's a great idea to refactor it to be async-only now that I see the benefits. Please give me some time to implement and verify it :)

dabla: It's a choice, for MSGraph it was an obvious choice as the Python SDK was async only, so it made sense. Here we can choose, hence why I brought it up. Still I think polling should always be done in deferred mode and thus via a trigger to avoid blocking the worker. As this is a new hook/operator, I think it's an opportunity to make that deliberate choice as well.

aaron-y-chen: I see. After thinking it over, I still believe we can go with the async-only approach since it's a good fit for the AAS refresh scenario.

aaron-y-chen:
I learned a lot from this round of reviews, many thanks 😄

providers/microsoft/azure/pyproject.toml
aaron-y-chen: The implementation and e2e test (no screenshot this time lol) are done.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
```diff
@@ -0,0 +1,280 @@
...
+    def client(self) -> httpx.AsyncClient:
+        """Return the shared HTTP client, creating it on first use."""
+        if self._client is None:
+            self._client = httpx.AsyncClient(timeout=self.request_timeout)
+        return self._client
+
+    def get_conn(self) -> AsyncTokenCredential:
+        """Return and cache the service principal credential."""
+        if self._credential is not None:
+            return self._credential
```
dabla, line 140: **[blocker]** ` get_conn() ` returns the auth credential, not the connection/client — inverts the Airflow hook contract.

Every Airflow hook that overrides `get_conn()` returns the object callers use to *do work*: `AzureBaseHook` returns the SDK management client, `KiotaRequestAdapterHook` returns the `RequestAdapter`, `PostgresHook` returns the psycopg2 connection, etc. The credential (`ClientSecretCredential`) is an auth building block — it is not the thing operators, sensors, or triggers call to make requests.

The symptom is visible in `_get_headers()`:

```python
token = await self.get_conn().get_token(TOKEN_SCOPE)
```

`get_conn()` is being called as "give me the credential so I can get a token", but callers expect it to mean "give me the client so I can make requests".

The fix is to swap the two: `get_conn()` should return `httpx.AsyncClient` (what is currently the `client` property), and the credential becomes a private helper:

```python
def get_conn(self) -> httpx.AsyncClient:
    """Return the shared HTTP client, creating it on first use."""
    if self._client is None:
        self._client = httpx.AsyncClient(timeout=self.request_timeout)
    return self._client

def _build_credential(self) -> AsyncTokenCredential:
    connection = self.connection
    tenant_id = connection.extra_dejson.get("tenantId")
    if not connection.login:
        raise ValueError("Client ID is required for Azure Analysis Services authentication")
    if not connection.password:
        raise ValueError("Client secret is required for Azure Analysis Services authentication")
    if not isinstance(tenant_id, str) or not tenant_id:
        raise ValueError("Tenant ID is required for Azure Analysis Services authentication")
    return ClientSecretCredential(
        tenant_id=tenant_id,
        client_id=connection.login,
        client_secret=connection.password,
    )
```

`_get_headers()` then calls `self._credential.get_token(...)` (lazy-init via `_build_credential()` on first use). The `_credential` field stays exactly as-is in `__init__` and `aclose()`. The public `client` property can be removed — `get_conn()` replaces it. Any caller that currently uses `self.client` (in `get_refresh_status`, `trigger_refresh`) becomes `self.get_conn()`.

---
Drafted-by: Claude Sonnet 4.6 (claude-sonnet-4.6); reviewed by @dabla before posting

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: Nice, fixed!

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
```diff
@@ -0,0 +1,279 @@
...
+        )
+        return self._credential
+
+    async def aclose(self) -> None:
+        """Release the HTTP client and the credential."""
+        client, self._client = self._client, None
+        credential, self._credential = self._credential, None
+        try:
+            if client is not None:
+                await client.aclose()
```
dabla, line 166: Don't we need to set `self._client` and `self._credential` back to `None` once closed?

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
```diff
@@ -0,0 +1,279 @@
...
+            or parsed_host.fragment
+        ):
+            raise ValueError(
+                "A valid region endpoint without a URL scheme, credentials, port, or path is "
+                "required in the Azure Analysis Services connection host"
+            )
+
+    def _get_base_url(self) -> str:
+        host = (self.connection.host or "").strip().rstrip("/")
+        self._assert_host(host, urlsplit(f"//{host}"))
```
dabla: Why not just pass host to `_assert_host` and do the `urlsplit` within the `_assert_host` method?

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: I already do that at line [161-162](https://github.com/apache/airflow/pull/71350/changes#diff-5050e6c2518613e377a79581d2bd8033aa7f90c1d52138cffc614d7308b2e40bR161-R162).

dabla: Oh I wasn't aware of that syntax, a bit confusing if you ask me.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: Okay, I agree this part I should put `urlsplit` in `_assert_host`, let me fix it.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: hmmm, let me modify it as below, it might be clearer.

```python
client = self._client
credential = self._credential
self._client = None
self._credential = None
```

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
```diff
@@ -0,0 +1,280 @@
...
+            raise AzureAnalysisServicesRefreshException(
+                "Azure Analysis Services returned an invalid refresh Location header"
+            ) from error
+        if len(location_parts) < 2 or location_parts[-2] != "refreshes":
+            raise AzureAnalysisServicesRefreshException(
+                "Azure Analysis Services returned an invalid refresh Location header"
+            )
+        return unquote(location_parts[-1])
+
+    def _assert_host(self, host: str) -> None:
```
dabla: I think this can now become a static or classmethod.

providers/microsoft/azure/tests/unit/microsoft/azure/operators/test_analysis_services.py
```diff
@@ -0,0 +1,179 @@
...
+        assert AzureAnalysisServicesRefreshOperator.template_fields == (
+            "azure_analysis_services_conn_id",
+            "server_name",
+            "database",
+            "refresh_type",
+        )
+
+    def test_execute_defers_without_refresh_id(self):
+        with pytest.raises(TaskDeferred) as deferred:
+            build_operator().execute(context=build_context())
```
dabla, line 112: The `tests_common` module contains a method named `execute_operator`, which allows you to fully run a deferred operator, see `from tests_common.test_utils.operators.run_deferrable import execute_operator`

Check out the [test_msgraph](https://github.com/apache/airflow/blob/main/providers/microsoft/azure/tests/unit/microsoft/azure/operators/test_msgraph.py) module under operators.

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/analysis_services.py
aaron-y-chen: I chose staticmethod.

providers/microsoft/azure/tests/unit/microsoft/azure/operators/test_analysis_services.py
aaron-y-chen: Thanks, another thing I learned today! I kept `test_execute_defers_without_refresh_id` because it verifies deferral details that `execute_operator` does not expose. I added `test_execute_operator_full_lifecycle` to exercise the complete deferred lifecycle, along with equivalent lifecycle coverage for the sensor.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/triggers/analysis_services.py
```diff
@@ -0,0 +1,194 @@
...
+                "conn_id": self.conn_id,
+                "server_name": self.server_name,
+                "database": self.database,
+                "refresh_id": self.refresh_id,
+                "refresh_type": self.refresh_type,
+                "poke_interval": self.poke_interval,
+                "request_timeout": self.request_timeout,
+            },
+        )
+
```
dabla, line 126: **[warning]** Move the polling loop into an async hook method so `run()` purely delegates and yields events.

`run()` currently mixes two responsibilities: orchestrating the trigger's event protocol (start vs. poll, success vs. error `TriggerEvent`s) and the actual polling mechanics (the `while True` / `asyncio.sleep` / terminal-status loop). That makes the coroutine long and harder to unit-test the polling behaviour in isolation from the event-yielding behaviour.

[#72336](https://github.com/apache/airflow/pull/72336) established a cleaner split for exactly this shape of problem: `SFTPTransferTrigger.run()` is a thin wrapper that calls a single hook coroutine (`SFTPHookAsync.transfer()`) and translates its outcome into a `TriggerEvent`, while all of the actual I/O logic lives in the hook. Applying the same split here:

```python
# hooks/analysis_services.py
async def wait_for_refresh(
    self, server_name: str, database: str, refresh_id: str, poke_interval: float
) -> str:
    """Poll until the refresh reaches a terminal status and return it."""
    while True:
        status = await self.get_refresh_status(
            server_name=server_name, database=database, refresh_id=refresh_id
        )
        self.log.info("Refresh %s status: %s", refresh_id, status)
        if (
            status == AzureAnalysisServicesRefreshStatus.SUCCEEDED
            or status in AzureAnalysisServicesRefreshStatus.FAILURE_STATUSES
        ):
            return status
        await asyncio.sleep(poke_interval)
```

```python
# triggers/analysis_services.py
async def run(self) -> AsyncIterator[TriggerEvent]:
    """Start the refresh when needed, then delegate polling to the hook."""
    hook = AzureAnalysisServicesHook(
        azure_analysis_services_conn_id=self.conn_id,
        request_timeout=self.request_timeout,
    )
    refresh_id = self.refresh_id
    try:
        if refresh_id is None:
            refresh_id = await hook.trigger_refresh(
                server_name=self.server_name, database=self.database, refresh_type=self.refresh_type
            )
            self.log.info("Triggered Azure Analysis Services refresh %s", refresh_id)
            yield TriggerEvent(
                {
                    "status": "success",
                    "refresh_status": None,
                    "message": f"Refresh {refresh_id} has been triggered",
                    "refresh_id": refresh_id,
                }
            )
            return

        status = await hook.wait_for_refresh(
            server_name=self.server_name,
            database=self.database,
            refresh_id=refresh_id,
            poke_interval=self.poke_interval,
        )
        is_success = status == AzureAnalysisServicesRefreshStatus.SUCCEEDED
        yield TriggerEvent(
            {
                "status": "success" if is_success else "error",
                "refresh_status": status,
                "message": (
                    f"Refresh {refresh_id} completed successfully"
                    if is_success
                    else f"Refresh {refresh_id} finished with status {status}"
                ),
                "refresh_id": refresh_id,
            }
        )
    except Exception as error:
        message = str(error) or type(error).__name__
        yield TriggerEvent(
            {"status": "error", "refresh_status": None, "message": message, "refresh_id": refresh_id}
        )
    finally:
        await hook.aclose()
```

Bonus: `wait_for_refresh` becomes independently unit-testable against the hook (mocking only `get_refresh_status` and `asyncio.sleep`), the same way `SFTPHookAsync.transfer()` is tested apart from `SFTPTransferTrigger`.

dabla (commented):
The design is solid: a purely async, deferred-only Hook/Operator/Sensor/Trigger set for Azure Analysis Services refreshes, correctly avoiding worker-slot blocking, with `httpx.AsyncClient` used natively (no `asyncio.to_thread` workaround needed), proper `aclose()` resource cleanup of both the HTTP client and the credential, and thorough test coverage including full deferred-lifecycle execution via `execute_operator`. The two concerns raised in earlier review rounds (`_get_base_url` userinfo/port injection, and adding `requests` as a redundant dependency alongside `httpx`) have both been correctly addressed in later commits (`_assert_host` now rejects credentials/ports in the netloc, and the hook is `httpx`-only). `_get_headers` correctly builds the Authorization header as an OAuth bearer credential using the fetched token's `.token` attribute.

One design suggestion (see comment [1]): move the trigger's polling loop into an async hook method, mirroring the split used in [#72336](https://github.com/apache/airflow/pull/72336) for `SFTPTransferTrigger`/`SFTPHookAsync.transfer()`. This keeps `run()` a thin dispatcher and makes the polling logic independently testable. Not a blocker - a worthwhile follow-up polish either in this PR or a fast-follow.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/triggers/analysis_services.py
aaron-y-chen: I was convinced by this design choice, so I refactored it accordingly. I think it's cleaner now. :)

dabla (approved):
Well done @aaron-y-chen !

PR #71356, opened 2026-08-10
Document effect of clearing a task instance on task state store

amoghrajesh:
---

Note on existing behaviour. It's this way because, we should attempt to connect to a remote job instead of submitting fresh even when cleared but its worth making a note of this.

Came up during review of https://github.com/apache/airflow/pull/71211

---

### DIFF
--- a/airflow-core/docs/core-concepts/resumable-tasks.rst
+++ b/airflow-core/docs/core-concepts/resumable-tasks.rst
@@ -126,6 +126,19 @@ existing job on retry instead of submitting a new one.
 
 For more details and a working example, see :class:`~airflow.sdk.ResumableJobMixin`.
 
+**Clearing a task is treated the same as a retry**
+
+Clearing a task instance does not delete its ``task_state_store`` rows -- they are only removed
+when the ``dag_run`` itself is deleted, or by :ref:`airflow state-store clean
+<task-and-asset-state-store-cleanup>`. For a checkpointed task this is usually what you want:
+clearing resumes from the last checkpoint rather than starting over.
+
+For a :class:`~airflow.sdk.ResumableJobMixin` operator, it means clearing a task whose external job
+already succeeded reads that stored result back and returns immediately, without resubmitting the
+job. If you want clearing to always resubmit regardless of a prior success, set
+``[state_store] clear_on_success = True``, which deletes a task's state store rows automatically
+when it moves to ``SUCCESS`` (see :doc:`administration-and-deployment/task-and-asset-state-store`).
+
 .. _concepts-resumable-tasks-async:
 
 Asynchronous Tasks
--- a/airflow-core/docs/core-concepts/resumable-tasks.rst
+++ b/airflow-core/docs/core-concepts/resumable-tasks.rst
@@ -133,11 +133,19 @@ when the ``dag_run`` itself is deleted, or by :ref:`airflow state-store clean
 <task-and-asset-state-store-cleanup>`. For a checkpointed task this is usually what you want:
 clearing resumes from the last checkpoint rather than starting over.
 
-For a :class:`~airflow.sdk.ResumableJobMixin` operator, it means clearing a task whose external job
-already succeeded reads that stored result back and returns immediately, without resubmitting the
-job. If you want clearing to always resubmit regardless of a prior success, set
+For an operator with durable execution, it means clearing a task whose external job already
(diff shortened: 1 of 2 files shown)


airflow-core/docs/core-concepts/resumable-tasks.rst
```diff
@@ -126,6 +126,19 @@ existing job on retry instead of submitting a new one.
...
 For more details and a working example, see :class:`~airflow.sdk.ResumableJobMixin`.
 
+**Clearing a task is treated the same as a retry**
+
+Clearing a task instance does not delete its ``task_state_store`` rows -- they are only removed
+when the ``dag_run`` itself is deleted, or by :ref:`airflow state-store clean
+<task-and-asset-state-store-cleanup>`. For a checkpointed task this is usually what you want:
+clearing resumes from the last checkpoint rather than starting over.
+
+For a :class:`~airflow.sdk.ResumableJobMixin` operator, it means clearing a task whose external job
```
Lee-W: ResumableJobMixin itself is not an operator. This description is confusing to me

amoghrajesh: Would it be better if I said "operator that supports durable execution?

Lee-W: way better. probably can do it this way? `operator that supports durable execution (i.e., inherits airflow.sdk.ResumableJobMixin)`

amoghrajesh: Sure works for me.

airflow-core/docs/core-concepts/resumable-tasks.rst
```diff
@@ -126,6 +126,19 @@ existing job on retry instead of submitting a new one.
...
 For more details and a working example, see :class:`~airflow.sdk.ResumableJobMixin`.
 
+**Clearing a task is treated the same as a retry**
+
+Clearing a task instance does not delete its ``task_state_store`` rows -- they are only removed
+when the ``dag_run`` itself is deleted, or by :ref:`airflow state-store clean
+<task-and-asset-state-store-cleanup>`. For a checkpointed task this is usually what you want:
+clearing resumes from the last checkpoint rather than starting over.
+
+For an operator that supports durable execution (ie: inherits :class:`~airflow.sdk.ResumableJobMixin`), it means
```
Lee-W:
```suggestion
For an operator that supports durable execution (i.e., inherits :class:`~airflow.sdk.ResumableJobMixin`), it means 
```

Lee-W (approved):

jason810496 (commented):
IIUC, it only makes sense when the job is in the _deferred_ state.
For the case of `wait_for_termination=True` + `deferrable=False` (watching the external job on worker instead of triggerer), we still execute the `operator.on_kill` hook, which result in **killing the healthy external job** and unable to connect back to the prior external job instance.

So perhaps we should mention the case of `wait_for_termination=True` + `deferrable=True` in the note? Just in case users interpret having `ResumableJobMixin` is same as "we can always connect to the external job after clear" but it's not.

uros-b (approved):

amoghrajesh:
@jason810496 thats a good point, I pushed an update. WDYT: f62d2d99fc?

airflow-core/docs/core-concepts/resumable-tasks.rst
```diff
@@ -126,6 +126,27 @@ existing job on retry instead of submitting a new one.
...
+clearing resumes from the last checkpoint rather than starting over.
+
+For an operator with durable execution, it means clearing a task whose external job already
+succeeded reads that stored result back and returns immediately, without resubmitting the job. If
+you want clearing to always resubmit regardless of a prior success, set
+``[state_store] clear_on_success = True``, which deletes a task's state store rows automatically
+when it moves to ``SUCCESS`` (see :doc:`/administration-and-deployment/task-and-asset-state-store`).
+
+This does not guarantee the external job is still there to reconnect to, though. Clearing a task
+that is actively running (``deferrable=False``) stops the worker process, which runs the
```
jason810496: Small nit, no strong opnion.

```suggestion
that is actively running with ``deferrable=False`` stops the worker process, which runs the
```

jason810496 (approved):
> @jason810496 thats a good point, I pushed an update. WDYT: f62d2d99fc?

LGTM now, thanks for the update.

PR #71358, opened 2026-08-10
Document how clearing tasks works with task state store on durable operators

amoghrajesh:
---

Same as https://github.com/apache/airflow/pull/71356 but for providers.

Clearing a task whose external job already succeeded returns without resubmitting, since state store rows survive a clear. Rather than duplicate the explanation per operator, each doc gets a one-line pointer to the shared core-concepts writeup, plus the `clear_on_success` setting that restores "clearing always resubmits."

Touches `SparkSubmitOperator`, `RedshiftDataOperator`, `DatabricksRunNowOperator`, `DatabricksSubmitRunOperator`, and `SnowflakeSqlApiOperator`.

---

### DIFF
--- a/providers/amazon/docs/operators/redshift/redshift_data.rst
+++ b/providers/amazon/docs/operators/redshift/redshift_data.rst
@@ -99,6 +99,12 @@ the statement id won't be there for the next retry, and the operator will submit
 instead of reconnecting. Avoid running cleanup on a schedule shorter than your longest
 ``retry_delay``.
 
+Clearing a task is treated the same as a retry, which matters specifically for a task whose
+statement already succeeded: clearing does not delete the stored statement id, so the next attempt
+reads it back and returns immediately without submitting the SQL again. See
+:doc:`apache-airflow:core-concepts/resumable-tasks` for why, and for the
+``[state_store] clear_on_success`` setting that restores "clearing always resubmits."
+
 To opt out and always submit fresh SQL on retry, set ``durable=False``:
 
 .. code-block:: python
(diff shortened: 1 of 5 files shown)


uros-b (commented):
@amoghrajesh Please note that the PR title got cut-off mid word:

> Document how clearing tasks works with task state store on durable operators

jason810496 (commented):
How about waiting for https://github.com/apache/airflow/pull/71356#pullrequestreview-4896595054 discussion first before landing into each providers.

Additionally, would it make sense to add the prek script with `<-- auto sync by ... -->` like marker to prevent drift?

Lee-W:
agree we can wait for another PR to be merged. but I don't think we really need a prek hook for that. overkill IMO

amoghrajesh:
Yeah I think we don't need a prek hook - seems like an overkill.

amoghrajesh:
> @amoghrajesh Please note that the PR title got cut-off mid word:
> 
> > Document how clearing tasks works with task state store on durable operators

Thanks for letting me know, thanks!

jason810496 (commented):
How about mentioning something like "we're only able to reconnect back to the external job if it's deferred task `deferrable=True`, see more in ... <link to the core docs that we just settled in https://github.com/apache/airflow/pull/71356 >"?

amoghrajesh:
Added. Each doc now says explicitly that reconnect-on-clear is most reliable for deferred tasks (deferrable=True), since clearing an actively-polling synchronous task can cancel the job via on_kill before the next attempt gets a chance to reconnect. Links to the caveat in core-concepts/resumable-tasks (from #71356).

amoghrajesh:
@jason810496 @uros-b I added the reference and suggestion as suggested. Can you take a look?

jason810496 (approved):
Thanks for the update.

PR #71359, opened 2026-08-10
Report missing Anthropic WIF fields clearly

yuyua9:
Invalid Anthropic Workload Identity Federation configuration currently raises a raw KeyError when a required field is missing, which makes connection setup failures harder to diagnose.

This PR:
- Validates the four required `workload_identity` fields before indexing them.
- Raises `AnthropicError` listing missing or empty fields.
- Adds parameterized regression tests for missing and empty values while preserving optional `workspace_id` and `scope` handling.

Validation:
- `uv run --project providers/anthropic --frozen pytest providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py -xvs` — 87 passed.
- `uv run --project providers/anthropic --frozen ruff format providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py`
- `uv run --project providers/anthropic --frozen ruff check --fix providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py`
- The Breeze provider test command was not run because Breeze is unavailable in the local environment.

- [X] Yes — Codex (GPT-5)

### DIFF
--- a/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
@@ -224,7 +224,7 @@ class AnthropicHook(BaseHook):
     - ``workload_identity``: configure `Workload Identity Federation
       <https://platform.claude.com/docs/en/manage-claude/workload-identity-federation>`__
       (keyless OIDC auth) with ``identity_token_file``, ``federation_rule_id``,
-      ``organization_id``, ``service_account_id`` and optional ``workspace_id`` / ``scope``.
+      ``organization_id`` and optional ``service_account_id``, ``workspace_id`` / ``scope``.
 
     When the ``anthropic`` platform has no API Key and no ``workload_identity`` block, the
     client is built with no static credential so the SDK resolves them from the environment
@@ -313,20 +313,20 @@ class AnthropicHook(BaseHook):
             "identity_token_file",
             "federation_rule_id",
             "organization_id",
-            "service_account_id",
         )
         missing_fields = [field for field in required_fields if not wif.get(field)]
         if missing_fields:
             raise AnthropicError(
-                "The workload_identity configuration is missing required fields: "
+                "The workload_identity configuration is missing or has empty required fields: "
                 f"{', '.join(missing_fields)}."
             )
         kwargs: dict[str, Any] = {
             "identity_token_provider": IdentityTokenFile(wif["identity_token_file"]),
             "federation_rule_id": wif["federation_rule_id"],
             "organization_id": wif["organization_id"],
-            "service_account_id": wif["service_account_id"],
         }
+        if wif.get("service_account_id"):
+            kwargs["service_account_id"] = wif["service_account_id"]
         if wif.get("workspace_id"):
             kwargs["workspace_id"] = wif["workspace_id"]
         if wif.get("scope"):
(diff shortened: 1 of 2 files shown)


uros-b (approved):

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -309,6 +309,18 @@ def _workload_identity_credentials(wif: dict[str, Any]) -> WorkloadIdentityCrede
         Anthropic access token. See
         https://platform.claude.com/docs/en/manage-claude/workload-identity-federation.
         """
+        required_fields = (
+            "identity_token_file",
+            "federation_rule_id",
+            "organization_id",
+            "service_account_id",
```
kaxil: `service_account_id` isn't required by the SDK: `WorkloadIdentityCredentials.__init__` types it `Optional[str] = None` and only adds it to the token-exchange body when it isn't None (same on 0.101.0, the floor in `pyproject.toml`, and on current 0.120.2). The SDK's own env-driven path gates on just `federation_rule_id` + `organization_id` and passes `os.environ.get(ANTHROPIC_SERVICE_ACCOUNT_ID)` straight through, so env-configured WIF works without it.

So this rejects a config the SDK accepts, and makes the connection-configured path stricter than the env path documented right next to it in `connections.rst`. Could it move into the optional block below alongside `workspace_id`/`scope`? That would match how `aws_region` is handled (`test_aws_platform_without_region`: pass None and let the SDK resolve). The docstring on line 227 reads as required too, so worth fixing in the same pass.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -309,6 +309,18 @@ def _workload_identity_credentials(wif: dict[str, Any]) -> WorkloadIdentityCrede
...
+        )
+        missing_fields = [field for field in required_fields if not wif.get(field)]
+        if missing_fields:
+            raise AnthropicError(
+                "The workload_identity configuration is missing required fields: "
```
kaxil: `not wif.get(field)` also catches an empty string, so a key the user can see sitting in their `extra` gets reported as "missing". "missing or empty required fields" would match what the code does, and what the PR description says it does.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -425,6 +425,41 @@ def test_workload_identity_federation_explicit(
         )
         mock_anthropic.assert_called_once_with(credentials=mock_wic.return_value, base_url=None)
 
+    @pytest.mark.parametrize(
+        ("missing_field", "missing_value"),
+        [
+            pytest.param(field, value, id=f"{field}-{value or 'missing'}")
```
kaxil: `value or 'missing'` renders the same id for both cases, so pytest disambiguates them as `identity_token_file-missing0` / `-missing1` and the absent-vs-empty distinction disappears from the test output. `id=f"{field}-{'absent' if value is None else 'empty'}"` reads better.

kaxil (approved):

shahar1 (approved):

PR #71365, opened 2026-08-10
Add backward-walk coverage for cursor-based pagination

pierrejeambrun:
The keyset pagination helpers in `api_fastapi/common/cursors.py` are exercised only for forward paging. A regression in the backward (previous-page) path — excluding more than the boundary row, or failing to invert NULL placement when the sort direction flips — would silently truncate a page walked back into, with no test to catch it.

This adds a forward-then-backward invariant: walking backward through the pages must reproduce the forward pages in reverse, dropping nothing. It's parametrized over the cases that stress that path (equal leading values split across pages, and NULLs under both ASC and DESC).

Test-only; no behavior change.

---


ashb (approved):

ashb:
**[approve]**

👍🏻

bbovenzi (approved):

PR #71366, opened 2026-08-10
Fix 500 when combining last-run and any-run Dag state filters

### LINKED ISSUE #71367: Internal server error on applying Dags filters together on 'Last Run' and 'Any Run'

Airflow Core

### Apache Airflow version

3.3.1rc1

Image

Steps:
1. Start airflow on tag 3.3.1rc1
2. On Dags list page, apply both filters 'Last Run' and 'Any Run' together and see the issue

DB Backend used: Postgres

_No response_

### Operating System

Linux

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

vatsrahul1001:
The Dags list supports two run-state filters: `last_dag_run_state` (the Dag's latest run is in the state) and `dag_run_state` (the Dag has any run in the state). Applying both at once (e.g. `failed` + `failed`) returned a **500** instead of the intersection.

**Cause:** the any-run filter (`_AnyDagRunStateFilter`) built its `EXISTS` subquery with an un-aliased `DagRun`. When the outer query also references `DagRun` (the `last_dag_run_state` filter adds `WHERE DagRun.state == …`), SQLAlchemy auto-correlates the subquery's `DagRun` to the outer one, stripping the subquery's FROM and raising `InvalidRequestError` -> 500. Each filter alone worked, which is why it slipped through.

**Fix:** alias `DagRun` in the subquery so it resolves its own runs independently of the outer query. The combined filter now returns the intersection.

Added a regression test covering the combined case (the existing tests only exercised each filter separately).

The any-run-state filter is new (related: #70292, #70293), so this needs to ride along to 3.3.1 via the backport label.

closes: https://github.com/apache/airflow/issues/71367

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -39,6 +39,7 @@ from pydantic import AfterValidator, BaseModel, NonNegativeInt
 from sqlalchemy import Column, String, and_, func, not_, or_, select as sql_select, true as sql_true
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.inspection import inspect
+from sqlalchemy.orm import aliased
 from sqlalchemy.sql.functions import FunctionElement
 
 from airflow._shared.timezones import timezone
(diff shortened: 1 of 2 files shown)


pierrejeambrun (approved):
LGTM, thanks

PR #68484, opened 2026-06-12
Add preset filters to persist table filters and sorting

pierrejeambrun:
Adds **saved views** to the data tables: once you've applied filters / search / sorting to a table, save that state under a name and restore it later from the **Views** control on the right of the filter bar.

- **Save / restore / delete** named views. Deleting asks for confirmation.
- **Set a default** (pin icon): landing on the page with no filters auto-restores that view. Unpin to go back to the empty default.
- **Save is blocked** on the default (empty) view and when an identical setup is already saved (the message names the existing view).

A "view" is the table's current URL query string (filters, search, page size); the active sort is mirrored to `localStorage` (it is often absent from the URL, so it is never written there) and restored from there. Pagination is dropped so a restored view starts on page one. Everything is stored per-page in `localStorage` (per browser), mirroring how table sort/limit already persist. The control lives in the shared `FilterBar`, so every filterable table gets it — except a Dag run's task instances, where the per-run URL makes saved views pointless.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #71368, opened 2026-08-10
Add preset filters to the Dags list page

### LINKED ISSUE #66933: DAG list filters (status, run state, favorites) not persisted across navigation in Airflow 3.x

Airflow Core

### Apache Airflow version

3.2.1

In Airflow 3.x, the following filters on the DAGs list page reset every time you navigate away and come back:

- DAG status (All / Active / Paused)
- DAG run state (Queued / Running / Success / Failed)
- Favorites (All / Favorited / Unfavorited)
- Required Actions

Steps to reproduce:
1. Open the DAGs list page
2. Apply any of the filters above
3. Navigate to another page (click a DAG, go to Home, Assets, etc.)
4. Come back to the DAGs list
5. All filters are gone, reset to defaults

Filters should survive navigation. The tag filter had the same issue (#60590) and was fixed, so these should follow the same logic.

### Operating System

Ubuntu 22.04.5 LTS

### Deployment

Docker-Compose

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

pierrejeambrun:
The Dags list page hand-rolls its own filter row (`DagsFilters`) rather than going through the shared `FilterBar`, so the save/restore/pin-default views added in #68484 never reached it. Users reported that filters like status / run state / favorites / needs-review reset on every navigation away from `/dags` (issue #66933).

`PresetFiltersMenu` already reads and writes `useSearchParams` and persists per `pathname` to `localStorage` — no FilterBar-specific plumbing. It drops into `DagsFilters` unchanged. Every Dags-page filter (`paused`, `favorite`, `last_dag_run_state`, `dag_run_state`, `needs_review`, `tags`, `tags_match_mode`, `teams`, `timetable_type`, `name_pattern`) is already a URL param, so a saved preset restores them faithfully; the table sort is captured by the same `tableSortKey(pathname)` side-channel `PresetFiltersMenu` already uses on other pages.

For that faithful restore to actually hold, the `tags` filter needed one fix: it also persisted its selection to `localStorage` (added in #63273) which isn't consistent with how other filters behave, which shadowed the URL — applying a preset that omitted tags kept the previously selected tags instead of clearing them. Preset filters now own cross-navigation persistence for every filter, so this PR drops the tag-specific `localStorage` persistence (`tags` / `tags_match_mode` are purely URL-driven like the rest) and prunes the now-stale keys from users' browsers on load (`pruneLegacyTagFilterKeys`, mirroring the existing `pruneLegacyDependencyKeys`).

The bookmark button sits at the far-right end of the `DagsFilters` row, grouped with the Favorite filter so the two stay on the same line, mirroring where it sits on `FilterBar`-based pages.

closes: #66933

https://github.com/user-attachments/assets/c6ade57b-2695-4afb-926a-1f82a3076673

- [X] Yes — Claude Code (Opus 4.7)

Generated-by: Claude Code (Opus 4.7) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/ui/src/constants/localStorage.test.ts
+++ b/airflow-core/src/airflow/ui/src/constants/localStorage.test.ts
@@ -18,7 +18,11 @@
  */
 import { afterEach, describe, expect, it } from "vitest";
 
-import { pruneLegacyDependencyKeys, SHOW_ALL_DEPENDENCIES_KEY } from "./localStorage";
+import {
+  pruneLegacyDependencyKeys,
+  pruneLegacyTagFilterKeys,
+  SHOW_ALL_DEPENDENCIES_KEY,
+} from "./localStorage";
 
 describe("pruneLegacyDependencyKeys", () => {
   afterEach(() => {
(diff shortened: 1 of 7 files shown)


pierrejeambrun:
On smaller screens
Screenshot 2026-08-10 at 16 39 19

bbovenzi (approved):

PR #71374, opened 2026-08-10
Add performance_target parameter to DatabricksSubmitRunOperator

li-xiang-db:
The Databricks [`jobs/runs/submit`](https://docs.databricks.com/api/workspace/jobs/submit) API accepts a `performance_target` field (`PERFORMANCE_OPTIMIZED` or `STANDARD`) to select the serverless compute mode. `DatabricksSubmitRunOperator` had no named parameter for it, so it could only be set via the raw `json` payload. This exposes it as a templated named parameter that merges into the submit request like the other task fields.

---

### DIFF
--- a/providers/databricks/src/airflow/providers/databricks/operators/databricks.py
+++ b/providers/databricks/src/airflow/providers/databricks/operators/databricks.py
@@ -704,6 +704,13 @@ class DatabricksSubmitRunOperator(ResumableJobMixin, BaseOperator):
     :param do_xcom_push: Whether we should push run_id and run_page_url to xcom.
     :param git_source: Optional specification of a remote git repository from which
         supported task types are retrieved.
+    :param performance_target: Optional performance mode for the run on serverless compute.
+        Either ``PERFORMANCE_OPTIMIZED`` (prioritizes fast startup and execution) or
+        ``STANDARD`` (enables cost-efficient execution of serverless workloads). This field
+        will be templated.
+
+        .. seealso::
+            https://docs.databricks.com/api/workspace/jobs/submit
     :param deferrable: Run operator in the deferrable mode.
 
         .. seealso::
@@ -750,6 +757,7 @@ class DatabricksSubmitRunOperator(ResumableJobMixin, BaseOperator):
         "idempotency_token",
         "access_control_list",
         "git_source",
+        "performance_target",
         "databricks_conn_id",
     )
     template_ext: Sequence[str] = (".json-tpl",)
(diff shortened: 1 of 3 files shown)


eladkal:
cc @moomindani for review

uros-b (approved):

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
```diff
@@ -769,6 +777,7 @@ def __init__(
         access_control_list: list[dict[str, str]] | None = None,
         wait_for_termination: bool = True,
         git_source: dict[str, str] | None = None,
+        performance_target: str | None = None,
```
moomindani, line 795: Non-blocking, but worth considering: the Databricks API does not validate this value. I sent `performance_target: "BOGUS"` and the run was accepted, silently resolving to `effective_performance_target: PERFORMANCE_OPTIMIZED` — no error, no warning.

That means a typo or a lowercase `"standard"` leaves the user on the default mode without any signal, which for this particular field means unexpected cost characteristics rather than a visible failure. Since it is a two-value enum, validating in `__init__` against `{"PERFORMANCE_OPTIMIZED", "STANDARD"}` would surface the mistake at parse time.

Happy to leave it — the other named parameters do not validate either, and this is API behaviour rather than anything this PR introduced. Flagging it because silent cost impact is harder to notice than a failed task.

moomindani (approved):
LGTM. Validated the API side against a live workspace rather than relying on the docs — `performance_target` genuinely takes effect:

| Sent | `effective_performance_target` in the run |
|---|---|
| `PERFORMANCE_OPTIMIZED` | `PERFORMANCE_OPTIMIZED` |
| `STANDARD` | `STANDARD` |
| omitted | `PERFORMANCE_OPTIMIZED` (default) |

So exposing this as a named parameter is worth doing: `STANDARD` is only reachable by setting it explicitly, and previously that meant hand-writing the raw `json` payload. The implementation matches the existing named-parameter pattern exactly — one entry in `_get_named_json_parameters()`, registered in `template_fields`, docs updated. 212 tests pass, `prek --stage pre-commit` clean.

One non-blocking note inline about value validation.

---

providers/databricks/src/airflow/providers/databricks/operators/databricks.py
li-xiang-db: Hi @moomindani , thanks for reviewing, and great catch.

This is known behavior on the Databricks side that we agree isn't ideal. However currently it can't just be flipped to a hard error, since existing workloads today pass unrecognized values and are silently running on the default, rejecting those outright would turn a lot of currently-succeeding runs into failures.

So I'll leave it as a pass-through to stay consistent with the other named parameters here, and we're already working on improving the behavior from the Databricks API side.

eladkal (approved):
will merge when CI is green

PR #71380, opened 2026-08-10
Prevent a broken job from stalling AWS Batch executor sync

AlexanderBLR:
`sync_running_jobs()` handles all active jobs in a single pass, and `sync()` wraps the whole heartbeat in one catch-all. When handling of a single job raises, the exception aborts the entire cycle: the remaining jobs are not synced and, more importantly, `attempt_submit_jobs()` never runs. If the error is deterministic for that job, the executor stops submitting tasks entirely until the scheduler is restarted.

This is not hypothetical. `BatchJobCollection.pop_by_id()` raises `KeyError` before cleaning anything when the collection's bookkeeping is partially cleaned (e.g. a duplicate submission for the same workload key overwrites `key_to_id`, orphaning the older job id), so the failing job stays in `active_workers` forever and every subsequent heartbeat aborts at the same job. We hit exactly this in production (Airflow 3.2.1, `apache-airflow-providers-amazon==9.31.0`): one poisoned workload halted all Batch submissions cluster-wide (~20 Dags piling up in queued) until a scheduler restart, the only signal being `Failed to sync AwsBatchExecutor` repeating every heartbeat.

The fix:

- each job in `sync_running_jobs()` is handled in its own `try/except`; an unexpected per-job error evicts that job and fails its workload instead of aborting the cycle;
- new `BatchJobCollection.remove_job()` used for the eviction — a defensive removal that tolerates partially cleaned state, unlike `pop_by_id()`.

Credential-type boto errors (`ClientError` / `NoCredentialsError`) still reach `sync()`'s handler unchanged: they can only be raised by `_describe_jobs()` / `attempt_submit_jobs()`, which sit outside the per-job block — nothing inside the block calls AWS.

Tests: two `BatchJobCollection.remove_job()` tests, eviction-with-isolation test (poisoned job evicted + failed, healthy job in the same cycle still succeeds), eviction survival when `fail()` itself raises, and a regression test for the production impact (`attempt_submit_jobs()` still runs when a tracked job cannot be synced).

---

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
@@ -268,12 +268,15 @@ class AwsBatchExecutor(BaseExecutor):
         self.log.debug("Active Workers: %s", describe_job_response)
 
         for job in describe_job_response:
+            # snapshot the key before handling the job: if the error strikes after
+            # pop_by_id() already removed it, the collection can no longer tell us
+            # whose workload it was, and the workload would be left with no terminal state
+            workload_key = self.active_workers.id_to_key.get(job.job_id)
             try:
                 if job.get_job_state() == State.FAILED:
                     self._handle_failed_job(job)
                 elif job.get_job_state() == State.SUCCESS:
-                    workload_key = self.active_workers.pop_by_id(job.job_id)
-                    self.success(workload_key)
+                    self.success(self.active_workers.pop_by_id(job.job_id))
             except Exception:
                 self.log.exception(
                     "Evicting Batch job %s after an unexpected error while syncing it, "
@@ -281,11 +284,16 @@ class AwsBatchExecutor(BaseExecutor):
                     "submission for the whole executor.",
                     job.job_id,
                 )
-                self._evict_job(job.job_id)
+                self._evict_job(job.job_id, workload_key)
 
-    def _evict_job(self, job_id: str) -> None:
-        """Remove a job from the collection and fail its workload, tolerating corrupted bookkeeping."""
-        workload_key = self.active_workers.remove_job(job_id)
+    def _evict_job(self, job_id: str, workload_key: BatchJobWorkloadKey | None = None) -> None:
+        """
+        Remove a job from the collection and fail its workload, tolerating corrupted bookkeeping.
+
+        workload_key is the caller's snapshot of the mapping taken before the error;
+        it is the fallback when the job was already removed from the collection.
+        """
+        workload_key = self.active_workers.remove_job(job_id) or workload_key
         if workload_key is None:
             return
         try:
--- a/providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
@@ -268,11 +268,30 @@ class AwsBatchExecutor(BaseExecutor):
         self.log.debug("Active Workers: %s", describe_job_response)
 
         for job in describe_job_response:
-            if job.get_job_state() == State.FAILED:
-                self._handle_failed_job(job)
-            elif job.get_job_state() == State.SUCCESS:
-                workload_key = self.active_workers.pop_by_id(job.job_id)
-                self.success(workload_key)
+            try:
+                if job.get_job_state() == State.FAILED:
+                    self._handle_failed_job(job)
+                elif job.get_job_state() == State.SUCCESS:
+                    workload_key = self.active_workers.pop_by_id(job.job_id)
+                    self.success(workload_key)
+            except Exception:
+                self.log.exception(
+                    "Evicting Batch job %s after an unexpected error while syncing it, "
+                    "so that one broken job cannot abort the sync cycle and stall task "
+                    "submission for the whole executor.",
+                    job.job_id,
+                )
+                self._evict_job(job.job_id)
+
+    def _evict_job(self, job_id: str) -> None:
+        """Remove a job from the collection and fail its workload, tolerating corrupted bookkeeping."""
+        workload_key = self.active_workers.remove_job(job_id)
+        if workload_key is None:
+            return
+        try:
+            self.fail(workload_key)
+        except Exception:
+            self.log.exception("Failed to fail workload %s of evicted Batch job %s", workload_key, job_id)
 
     def _handle_failed_job(self, job):
         """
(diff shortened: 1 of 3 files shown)


vincbeck (approved):
Looks good to me but @o-nikolas' opinion on this one would be great

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
```diff
@@ -268,11 +268,30 @@ def sync_running_jobs(self):
...
+                if job.get_job_state() == State.FAILED:
+                    self._handle_failed_job(job)
+                elif job.get_job_state() == State.SUCCESS:
+                    workload_key = self.active_workers.pop_by_id(job.job_id)
+                    self.success(workload_key)
+            except Exception:
+                self.log.exception(
+                    "Evicting Batch job %s after an unexpected error while syncing it, "
+                    "so that one broken job cannot abort the sync cycle and stall task "
+                    "submission for the whole executor.",
```
o-nikolas: I don't think this kind of justification is necessary in the log message.

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
```diff
@@ -268,11 +268,30 @@ def sync_running_jobs(self):
...
-            elif job.get_job_state() == State.SUCCESS:
-                workload_key = self.active_workers.pop_by_id(job.job_id)
-                self.success(workload_key)
+            try:
+                if job.get_job_state() == State.FAILED:
+                    self._handle_failed_job(job)
+                elif job.get_job_state() == State.SUCCESS:
+                    workload_key = self.active_workers.pop_by_id(job.job_id)
+                    self.success(workload_key)
+            except Exception:
```
o-nikolas, line 283: There should be more concrete exceptions we're looking for here no? Evicting and failing tasks at any exception here feels harsh. Will every possible exception we can receive here be constant?

o-nikolas: Also, if any future code in this block gets added that calls boto then credential issues can get swallowed because it's catching Exception right now. At the very least you could catch ClientError/CredentialsError and re-raise those before catching Exception, but I'd still prefer to catch the exception types we're actually worried about here.

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
```diff
@@ -268,11 +268,30 @@ def sync_running_jobs(self):
...
+                )
+                self._evict_job(job.job_id)
+
+    def _evict_job(self, job_id: str) -> None:
+        """Remove a job from the collection and fail its workload, tolerating corrupted bookkeeping."""
+        workload_key = self.active_workers.remove_job(job_id)
+        if workload_key is None:
+            return
+        try:
+            self.fail(workload_key)
```
o-nikolas, line 300: If we manage to find an active task, is there a chance it actually succeeded in Batch? Should we auto fail everything here?

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
AlexanderBLR: Agreed, trimmed the message to just the action in the follow-up commit.

---Drafted-by: Claude Code (Fable 5); reviewed by @AlexanderBLR before posting

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
AlexanderBLR: Added the `ClientError` / `NoCredentialsError` re-raise you suggested in the follow-up commit, so credential issues keep reaching `sync()`'s connection handling even if code inside the block grows a boto call later.

On narrowing the catch further: the only concrete type we can name today is `KeyError` from `pop_by_id()` on corrupted bookkeeping, but the failure class this PR targets is "any deterministic error in per-job handling permanently wedges the executor". Narrowing to the exceptions we've already met protects only against the last incident — the next poison (a `TypeError` from a malformed status, an `AttributeError` from a partial workload, …) would stall submission cluster-wide again exactly as before.

Two things that hopefully make the broad catch less scary: 
1. It doesn't widen what gets caught — before this PR the very same exceptions were swallowed by `sync()`'s own `except Exception`; the change just moves the boundary so the blast radius is one job instead of the whole heartbeat.
2. On "will every possible exception here be constant?" — everything inside the block is in-memory bookkeeping plus executor state-change calls, no I/O, so in practice errors here are deterministic (that was our production case). If a transient one did slip through, the cost is one task failed and retried per its retry policy, versus today's cost of the whole sync cycle aborting. Happy to add an eviction threshold (only evict after N consecutive per-job sync errors) if you'd prefer to be more conservative.

---Drafted-by: Claude Code (Fable 5); reviewed by @AlexanderBLR before posting

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
AlexanderBLR: It's possible, yes — the narrow case is _"the job actually succeeded in Batch, but `success()` itself raised"._ By the time we're in the eviction path, reporting success has already failed, so the realistic options are: leave the task instance with no terminal state (it lingers until the scheduler's stuck-task handling times it out, and typically ends up failed anyway, just slower), or fail it promptly so the normal retry policy takes over. `fail()` seemed the least-bad default — deterministic, faster recovery, and within Airflow's usual contract that retries may re-run a task. But if you see a better terminal state to report here, happy to change it.

---Drafted-by: Claude Code (Fable 5); reviewed by @AlexanderBLR  before posting

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
o-nikolas: Okay, I at least feel a little better with the ClientError and NoCredentialsError being caught and re-raised to sync. I see what you're saying about unknown unknowns, each approach has it's pros and cons I suppose. We can see how this version goes for now and always circle back if this is not behaving how we'd like.

providers/amazon/src/airflow/providers/amazon/aws/executors/batch/batch_executor.py
o-nikolas: Hmm, I don't love the implication, but I hear your point. Let's try it as it is.

o-nikolas:
Ran tests again, let's see if they go green

AlexanderBLR:
> Ran tests again, let's see if they go green

@o-nikolas the re-run came back green and all checks passed. Anything else you'd like me to address here? Thanks!

PR #71383, opened 2026-08-10
Sync v3-3-stable with v3-3-test to release 3.3.1rc2

vatsrahul1001:
Time for 3.3.1rc2

[Release Notes](https://github.com/apache/airflow/commit/3adbbe1c58e4532df1964cb7794805e763816ee8)

---


kaxil (approved):

uv.lock
pierrejeambrun, line 1: Not sure about this uv.lock change.

pierrejeambrun (approved):
LGTM, just one comment

uv.lock
vatsrahul1001: Yeah, this is just metadata catching up - nothing actually re-resolves. Those four `specifier = ">=..."` lines come from the provider floor bumps that #71340 adds to the pyprojects in this same PR (`akeyless-cloud-id>=0.3.0`, `pyspark>=4.0.0`, `pydantic-ai-slim[mcp]>=2.0.0`, `pytest>=9.1.1`), so they map 1:1. #71373 just regenerates uv.lock to match, otherwise `uv lock --check` flags drift. No package resolved to a different version.

potiuk: Yes. This change is needed as we have to make sure.we lower bound ALL dependencies to some reasonable version - this protects our users for failing into some traps when downgrading to some old versiona of deps that are causing problems - especially on newer versions of Python. When some custom deps are installed - they might impact resolver  to choose old versions of deps that did not work for python. 3.14 but also did not explicitly excluded python 3.14 - seen it happening multiple times now. 

I added https://github.com/apache/airflow/pull/71378 to make sure leek will force lower-bounds everywhere.

potiuk (approved):

PR #71384, opened 2026-08-10
fix: allow in-flight DagRuns to continue when DAG is paused

### LINKED ISSUE #71381: Pausing a DAG stalls already-running DagRuns instead of only blocking new ones
### Description

When a DAG is paused while it has an active (running) DagRun — e.g. one started via manual trigger before the pause, or a manual trigger on an already-paused DAG that got an initial burst of tasks dispatched — the scheduler stops scheduling further task instances for that run. The run doesn't fail and logs no error; it simply stalls indefinitely in `running` state with remaining tasks stuck in `scheduled`, `queued_when=None`, never sent to the executor.

This happens because `is_paused` is checked as a single gate that governs both "should new DagRuns be created" and "should ready task instances in existing DagRuns be scheduled." The two are conflated.

### Use case/motivation

Pause is a reasonable way to say "don't start new runs of this DAG." It's not intuitive that it also freezes runs already in flight — especially since:
- No error, warning, or DAG/task state reflects why the run stopped progressing.
- A manually triggered run against a paused DAG appears to start successfully (initial tasks execute), masking the problem until the run has been silently stuck for several minutes.
- The eventual fix (unpause) has a side effect the operator may not want: it also releases scheduling for *new* runs of that DAG, when the operator's only goal was to unstick the one in-flight run.

### Suggested change

Preserve pre-run pause semantics: `is_paused` should gate DagRun *creation*, not task-instance scheduling for DagRuns already in `running` state. A DagRun that starts (however it started) while unpaused, or that gets tasks dispatched before a pause takes effect, should be allowed to run to completion. If the DAG is still paused when the run finishes, subsequent runs are blocked as expected — pause state before and after the run is preserved; only the in-flight run is exempted.

At minimum, surfacing this in the UI/logs (e.g., a task/run banner: "this DagRun's DAG is paused; scheduling is suspended") would prevent the silent-stall confusion, even if the gating behavior itself is kept as-is by design.

### Code of Conduct

waterWang:
### Problem

When a DAG is paused while it has an active (running) DagRun — e.g. one started via manual trigger before the pause, or a manual trigger on an already-paused DAG that got an initial burst of tasks dispatched — the scheduler stops scheduling further task instances for that run. The run does not fail and logs no error; it simply stalls indefinitely in `running` state with remaining tasks stuck in `scheduled`, `queued_when=None`, never sent to the executor.

This happens because `is_paused` is checked as a single gate that governs both "should new DagRuns be created" and "should ready task instances in existing DagRuns be scheduled." The two are conflated.

### Root cause

Two places filter by `is_paused` to block task scheduling, when only DagRun creation should be blocked:

1. **`scheduler_job_runner.py`** — `_executable_task_instances_to_queued()`: The task scheduling query filters out ALL task instances of paused DAGs via `.where(~DM.is_paused)`, including those belonging to already-RUNNING DagRuns.

2. **`dagrun.py`** — `get_running_dag_runs_to_examine()`: The method that selects RUNNING DagRuns for scheduling examination filters out all DagRuns of paused DAGs via `DagModel.is_paused == false()`.

### Fix

- Remove `.where(~DM.is_paused)` from the task scheduling query in `scheduler_job_runner.py`
- Remove `DagModel.is_paused == false()` from `get_running_dag_runs_to_examine()` in `dagrun.py`

**Preserved**: `DagModel.dags_needing_dagruns()` still filters by `is_paused == false()`, so paused DAGs cannot create **new** DagRuns. The `_update_dag_run_state_for_paused_dags` timer still runs every 60s for edge cases. The QUEUED DagRun filter in `dagrun.py` (line 818) also retains its `is_paused` check, so queued runs for paused DAGs are not promoted.

### Related issue

Closes #71381

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -726,7 +726,6 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 .join(TI.dag_run)
                 .where(DR.state == DagRunState.RUNNING)
                 .join(TI.dag_model)
-                .where(~DM.is_paused)
                 .where(TI.state == TaskInstanceState.SCHEDULED)
                 .where(DM.bundle_name.is_not(None))
                 .join(
--- a/airflow-core/src/airflow/models/dagrun.py
+++ b/airflow-core/src/airflow/models/dagrun.py
@@ -758,7 +758,6 @@ class DagRun(Base, LoggingMixin):
             .join(DagModel, DagModel.dag_id == cls.dag_id)
             .join(BackfillDagRun, BackfillDagRun.dag_run_id == DagRun.id, isouter=True)
             .where(
-                DagModel.is_paused == false(),
                 DagModel.is_stale == false(),
             )
             .order_by(


raphaelauv:
this is a big change

what about adding this in a new feature and not changing the current behavior of the pause button ( status ) 

maybe you could add a new toggle button ( stop starting new dag_run )

so we would have 

unpause / pause / pause_new_run

wdyt ?

kaxil:

PR #71387, opened 2026-08-10
Add IcebergTableSnapshotTrigger for event-driven scheduling

1fanwang:
# Rationale for this change

The Iceberg provider ships a hook and nothing else, so there is no way to schedule a DAG on "this table has new data". Anyone who wants it writes their own trigger.

`IcebergTableSnapshotTrigger` polls a table's branch head through the existing `IcebergHook` and emits an event when it advances, so a commit can drive an `AssetWatcher`:

```python
orders = Asset(
    "orders",
    watchers=[AssetWatcher(name="orders_commits", trigger=IcebergTableSnapshotTrigger(table="sales.orders"))],
)

@dag(schedule=[orders])
def downstream(): ...
```

The event carries `snapshot_id` and `previous_snapshot_id`, so a task can scan the delta instead of the whole table.

Polling is what Iceberg supports: the REST catalog spec defines no subscribe, webhook or event endpoint, and none is proposed.

Two behaviors worth calling out, both found by running it on a real Airflow rather than in tests:

- The cursor lives in the `asset_state_store` watermark, not on the instance. `serialize()` is captured once when the trigger row is written, so a value mutated during `run()` is lost on restart and the trigger re-reports a commit it already emitted. Read through `getattr`, since that attribute postdates the oldest Airflow this provider supports.
- An absent table is waited on, like an absent branch. A watcher outlives the table it watches, and raising kills the trigger, which the triggerer then restarts about once a second until the table appears.

A trigger watched by more than one asset gets one accessor per asset, where the watermark shorthand raises. It degrades to no watermark; https://github.com/apache/airflow/pull/71460 adds what is needed to keep a cursor there.

# Are these changes tested?

<details><summary>Unit tests — 19 passed</summary>

Cold start emits the current head; an unchanged table emits nothing; each commit emits one event carrying the snapshot it replaced; an absent branch or table is waited on; the watermark is restored on start and written per event; a state store failure is not mistaken for the several-assets case; the triggerer's dispatch reaches the poll loop; serialization round-trips.

```
$ pytest providers/apache/iceberg/tests/unit/apache/iceberg/triggers/test_iceberg.py -q
19 passed
```

Removing `triggers/iceberg.py` fails collection, so the tests cannot pass vacuously.

</details>

Then on a running Airflow: Postgres metadata database, api-server, dag-processor, scheduler and triggerer as separate processes, a real Iceberg catalog behind a real `iceberg_default` connection, and a DAG scheduled on the watched asset. Nothing stubbed.

<details><summary>On a running Airflow</summary>

```
$ airflow db migrate && airflow connections add iceberg_default --conn-type iceberg \
    --conn-extra '{"type":"sql","uri":"sqlite:////tmp/ice-wh/catalog.db","warehouse":"file:///tmp/ice-wh"}'
$ airflow api-server & airflow dag-processor & airflow scheduler & airflow triggerer &
$ airflow dags unpause iceberg_downstream

-- the trigger is registered and wired to the asset
 id |                                   classpath
----+-------------------------------------------------------------------------------
  1 | airflow.providers.apache.iceberg.triggers.iceberg.IcebergTableSnapshotTrigger
 trigger_id |  name
------------+--------
          1 | orders

1. table absent          starts=1  restarts=0  crashes=0  events=0
2. table created         Trigger fired event  TriggerEvent<{'table': 'sales.orders', ...}>
3. dag_run                iceberg_downstream | asset_triggered__2026-08-11T21:47:04 | asset_triggered
4. asset_state_store      asset_id=1  key=snapshot_id  value=8102133228104772796  last_updated_by_kind=watcher
5. triggerer killed and restarted, no new commit
   dag_runs=1 (unchanged)   events fired since restart=0
```

Step 4 is the real `asset_state_store` row, written by the triggerer over the execution API rather than by a test double. Step 5 is why the watermark is not just a kwarg: killing the triggerer and bringing it back produced no second run for a commit already reported.

</details>

An example DAG is included, following the Redis message-queue one.

### DIFF
--- a/providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
+++ b/providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
@@ -31,10 +31,12 @@ else:
     )
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Hashable
+    from collections.abc import AsyncIterator
 
 DEFAULT_BRANCH = "main"
 WATERMARK_KEY = "snapshot_id"
+# Raised by AssetStateStoreAccessors when the trigger is watched by more than one asset.
+_MULTI_ASSET_ERROR = "concrete inlets and outlets"
 
 
 class IcebergTableSnapshotTrigger(BaseEventTrigger):
@@ -60,9 +62,6 @@ class IcebergTableSnapshotTrigger(BaseEventTrigger):
 
         @dag(schedule=[orders])
         def downstream(): ...
-
-    Triggers sharing a catalog connection, branch and poll interval share one poll loop,
-    so watching many tables in a catalog does not open a connection per table.
 
     :param table: Fully-qualified table name, ``namespace.table``. Nested namespaces are
         written as ``a.b.table``.
@@ -104,10 +103,6 @@ class IcebergTableSnapshotTrigger(BaseEventTrigger):
             },
         )
 
-    def shared_stream_key(self) -> Hashable | None:
-        """Group triggers that can be served by a single poll of the same catalog."""
-        return (self.iceberg_conn_id, self.branch, self.poll_interval)
-
     def _head_snapshot_id(self) -> int | None:
         """Return the snapshot the branch points at, or None if the branch does not exist."""
         table = IcebergHook(self.iceberg_conn_id).load_table(self.table)
@@ -124,8 +119,15 @@ class IcebergTableSnapshotTrigger(BaseEventTrigger):
         if store is not None:
             try:
                 stored = await asyncio.to_thread(store.get, WATERMARK_KEY)
-            except ValueError:
-                # Several assets watch this trigger, so there is no single cursor to keep.
+            except ValueError as err:
+                # The accessor serves one asset at a time, so it refuses to guess when this
+                # trigger is watched by several. That happens because triggers are deduplicated
+                # by hash(classpath, kwargs) while asset_watcher is many-to-many, so two assets
+                # watching this table with the same arguments share one trigger. There is then
+                # no single cursor to keep. A state store backend can raise ValueError too, and
+                # swallowing that would disable the watermark without saying so.
+                if _MULTI_ASSET_ERROR not in str(err):
+                    raise
                 self.log.warning(
                     "%s is watched by more than one asset; not persisting a snapshot watermark, "
                     "so a triggerer restart may re-emit the current head.",
--- a/providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
+++ b/providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
@@ -34,6 +34,7 @@ if TYPE_CHECKING:
     from collections.abc import AsyncIterator, Hashable
 
 DEFAULT_BRANCH = "main"
+WATERMARK_KEY = "snapshot_id"
 
 
 class IcebergTableSnapshotTrigger(BaseEventTrigger):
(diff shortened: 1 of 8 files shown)


providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
```diff
@@ -0,0 +1,154 @@
...
+        # serialize() is captured once when the trigger row is created, so a value mutated on
+        # self is lost when the triggerer restarts and the current head would be re-emitted as
+        # a new commit. The watermark survives that; the kwarg only seeds the first run.
+        # getattr because asset_state_store postdates the Airflow versions this provider supports.
+        store = getattr(self, "asset_state_store", None)
+        if store is not None:
+            try:
+                stored = await asyncio.to_thread(store.get, WATERMARK_KEY)
+            except ValueError:
+                # Several assets watch this trigger, so there is no single cursor to keep.
```
vincbeck: Can you explain why `ValueError` means multiple assets are watching this cursor?

1fanwang: Because the [get()](https://github.com/apache/airflow/blob/eeeaa0c23d878a12d80cbf07792736f59acb5da8/task-sdk/src/airflow/sdk/execution_time/context.py#L902-L904) delegates to [_single_accessor()](https://github.com/apache/airflow/blob/eeeaa0c23d878a12d80cbf07792736f59acb5da8/task-sdk/src/airflow/sdk/execution_time/context.py#L893-L897), which raises unless exactly one asset is behind them. In the triggerer that count is the number of assets watching this trigger, so I think the only way to hit it is more than one.

But I get that the catch is too broad here, how about replacing it with https://github.com/apache/airflow/pull/71460?

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
```diff
@@ -0,0 +1,164 @@
...
+                    "%s is watched by more than one asset; not persisting a snapshot watermark, "
+                    "so a triggerer restart may re-emit the current head.",
+                    self.table,
+                )
+                store = None
+            else:
+                if stored is not None:
+                    self.last_seen_snapshot_id = int(stored)
+
+        while True:
```
vincbeck, line 155: This runs even though `if store is None`, does that make sense? Does this trigger should run if there is not store?

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
vincbeck: Good

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
1fanwang: Double checked this, I want to say it's intended

IIUC the trigger still polls and fires *correctly without store, altho in that case the trigger would lose the restart (when triggerer crash and restart) cursor, so I think the overhead is a rare duplicate run?

essentially keeping "This runs even though if store is None" makes it *safe, as in we prefer to make sure to not miss a schedule than triggering a rare duplicate run

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
1fanwang: Thanks! will hold this PR until https://github.com/apache/airflow/pull/71460 get's merged

vincbeck (approved):

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
```diff
@@ -0,0 +1,170 @@
...
+    async def run(self) -> AsyncIterator[TriggerEvent]:
+        # serialize() is captured once when the trigger row is created, so a value mutated on
+        # self is lost when the triggerer restarts and the current head would be re-emitted as
+        # a new commit. The watermark survives that; the kwarg only seeds the first run.
+        # It postdates the Airflow versions this provider supports and is absent when several
+        # assets share the trigger, so polling carries on without it, losing only that cursor.
+        store = getattr(self, "asset_state_store", None)
+        if store is not None:
+            try:
+                stored = await asyncio.to_thread(store.get, WATERMARK_KEY)
```
amoghrajesh, line 135: Instead of reinventing the async behaviour in every trigger, we should migrate this to `aget`, `aset` etc once https://github.com/apache/airflow/pull/72127 lands.

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
```diff
@@ -0,0 +1,170 @@
...
+from airflow.providers.apache.iceberg.hooks.iceberg import IcebergHook
+from airflow.providers.apache.iceberg.version_compat import AIRFLOW_V_3_0_PLUS
+
+if AIRFLOW_V_3_0_PLUS:
+    from airflow.triggers.base import BaseEventTrigger, TriggerEvent
+else:
+    from airflow.triggers.base import (  # type: ignore[assignment]
+        BaseTrigger as BaseEventTrigger,
+        TriggerEvent,
+    )
```
amoghrajesh, line 33: We should avoid `if/else` based on version as much as possible, gets hard to manage conditional code over time. An improvement: https://github.com/apache/airflow/pull/72140

vincbeck:
@1fanwang can you work on these comments in a separate PR please?

1fanwang:
> @1fanwang can you work on these comments in a separate PR please?

Ack @vincbeck checking now

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
1fanwang: Thanks! have the fix ready that we can circle back on once ^^ 72127 lands

providers/apache/iceberg/src/airflow/providers/apache/iceberg/triggers/iceberg.py
1fanwang: agreed, added my non binding approval there

1fanwang:
> > @1fanwang can you work on these comments in a separate PR please?
> 
> Ack @vincbeck checking now

Addressed / followed up in the comment threads abovr, thanks!

PR #71394, opened 2026-08-11
Harden TypeScript SDK dependency installation

jason810496:
## Why

New or compromised npm releases can execute code during dependency installation before they are detected and removed from the registry.

## What

- Delay newly resolved dependency versions for 14 days.
- Block exotic transitive dependency sources.
- Fail on unapproved lifecycle builds and allow only the required `esbuild` scripts.
- Document the dependency security policy and include it in compile-hook triggers.

---

- [x] Yes, with help of Codex GPT-5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/ts-sdk/.pre-commit-config.yaml
+++ b/ts-sdk/.pre-commit-config.yaml
@@ -59,6 +59,7 @@ repos:
           ^src/.*\.ts$|
           ^package\.json$|
           ^pnpm-lock\.yaml$|
+          ^pnpm-workspace\.yaml$|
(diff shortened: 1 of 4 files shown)


potiuk (approved):
I think 14 days is a bit long but I am fine with it.

jason810496:
No need to backport, `ts-sdk` doesn't even existed in `v3-3-test` branch.

PR #62919, opened 2026-03-05
fix(fab): recover from first idle MySQL disconnect in token auth

kbohra:
Fix FAB auth token deserialization to recover from transient MySQL idle disconnects in the same request by clearing scoped session state and retrying once; add regression coverage for this recovery path.

This PR closes: #62903

##### Was generative AI tooling used to co-author this PR?

- [x] No 

### What this PR changes
- Add a one-time retry in `FabAuthManager.deserialize_user()` after `session.remove()` on `SQLAlchemyError`.
- Keep existing behavior for `NoResultFound` (still raises `ValueError`).
- If retry also fails, remove session again and re-raise to avoid reusing broken transaction/session state.
- Add regression test proving first-request recovery after transient disconnect.
- Update existing cleanup assertions to account for retry path.

### Why
With MySQL idle timeouts, the first request can hit a dead connection and return 500, while the next request succeeds. Retrying once after scoped-session cleanup makes the first request recover instead of failing.

### Unit Tests
Have updated the existing unit test and validated the same

PR #71396, opened 2026-08-11
fix: recover from stale DB session on revoked-token check (#71395)

### LINKED ISSUE #71395: First authenticated request after MySQL idle disconnect returns 500: RevokedToken.is_revoked runs on a stale session before FAB's recovery (#62919) can kick in

Providers

### Apache Airflow version

3.2.2

With MySQL as the metadata DB and the FAB auth manager, after the api-server sits idle longer than MySQL's `wait_timeout`, the **first** authenticated request (any UI or REST endpoint) fails with HTTP 500. The exception is an `OperationalError` (MySQL error 4031, "The client was disconnected by the server because of inactivity") raised from `RevokedToken.is_revoked` inside `BaseAuthManager.get_user_from_token`:

https://github.com/apache/airflow/blob/3.2.2/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py#L152

The immediately following request succeeds, because the failed query poisons and effectively resets the pooled connection. In production this shows up as intermittent 500s on the first request after overnight idle periods.

<details>
<summary>Traceback (trimmed)</summary>

```
  File ".../airflow/api_fastapi/auth/managers/base_auth_manager.py", line 152, in get_user_from_token
    if (jti := payload.get("jti")) and RevokedToken.is_revoked(jti):
  File ".../airflow/utils/session.py", line 100, in wrapper
    return func(*args, session=session, **kwargs)
  File ".../airflow/models/revoked_token.py", line 61, in is_revoked
    return bool(session.scalar(select(exists().where(cls.jti == jti))))
  ...
sqlalchemy.exc.OperationalError: (MySQLdb.OperationalError) (4031,
  'The client was disconnected by the server because of inactivity.
   See wait_timeout and interactive_timeout for configuring this behavior.')
```

</details>

### How to reproduce

1. Run Airflow 3.2.x with MySQL as the metadata DB and the FAB auth manager.
2. Lower the idle timeout to make the repro fast: `SET GLOBAL wait_timeout = 60;`
3. Make any authenticated API request (this warms up the pooled connection).
4. Wait longer than `wait_timeout` (e.g. 90 s) without touching the api-server.
5. Repeat the same request → HTTP 500 with `OperationalError` 4031 raised from
   `RevokedToken.is_revoked`. Repeat once more → succeeds.

#### Root cause

1. The FAB auth manager shares core's scoped `settings.Session`. `FabAuthManager.deserialize_user` leaves a transaction open after the request is served, so the connection stays checked out across requests. Once it idles past MySQL's `wait_timeout`, the server drops it (error 4031). Because the connection is never returned to the pool, there is no checkout event — `pool_pre_ping` / `pool_recycle` cannot help.
2. This exact failure mode was reported in #62903 and fixed by #62919, which added a discard-and-retry recovery to `FabAuthManager.deserialize_user`.
3. However, 3.2.0 introduced the JWT revocation check (#61339 / AIP-84): `RevokedToken.is_revoked` in `get_user_from_token` now runs **before** `deserialize_user`, so it is the first thing to touch the poisoned scoped session — and it has no recovery logic. The 500 that #62919 fixed is back, just raised one step earlier in the auth path.

The first request after an idle disconnect should recover transparently, as it does in `deserialize_user` since #62919. The revoked-token check should apply the same recovery: on `SQLAlchemyError`, discard the scoped session (`settings.Session.remove()`) and retry once on a fresh connection. Something like:

```python
if jti := payload.get("jti"):
    try:
        revoked = RevokedToken.is_revoked(jti)
    except SQLAlchemyError:
        log.warning(
            "Revoked-token check failed on a stale DB session; "
            "discarding the session and retrying once.",
            exc_info=True,
        )
        with suppress(Exception):
            settings.Session.remove()
        revoked = RevokedToken.is_revoked(jti)
    if revoked:
        raise InvalidTokenError("Token has been revoked")
```

Alternatively, the recovery could be hoisted so it covers the whole `get_user_from_token` path instead of being duplicated per call site.

### Operating System

_No response_

### Deployment

None

_No response_

apache-airflow-providers-fab==3.6.4 (already includes the #62919 fix — it is bypassed, not missing)

1.22.0 (latest released)

### Kubernetes Version

1.25

### Helm Chart configuration

_No response_

_No response_

### Anything else?

- Occurs deterministically whenever the first request after an idle period touches a connection older than `wait_timeout`; looks intermittent in production.
- Related prior work on the same class of failure:
  - #62903 / #62919 — same disconnect, recovery added to `deserialize_user` (this check runs earlier and bypasses it)
  - #62335 / #62336 — `Session.remove()` hardening in FAB's cleanup middleware
  - #66493 / #66494 — same call site, different concern (per-request DB roundtrip / pool exhaustion)
- We have been running the retry patch above in production for several weeks (applied at image build time) and it eliminated these 500s entirely.

### Code of Conduct

waterWang:
fixes #71395 (apache/airflow)

## Problem

After the api-server sits idle longer than MySQL `wait_timeout`, the **first** authenticated request fails with HTTP 500. The `OperationalError` (MySQL 4031) is raised from `RevokedToken.is_revoked`, which runs inside `BaseAuthManager.get_user_from_token` **before** `deserialize_user`. The recovery added in #62919 (discard-and-retry in `FabAuthManager.deserialize_user`) does not cover this earlier check, so the stale-session 500 is back, just one step earlier in the auth path.

## Fix

Apply the same discard-and-retry recovery to the revoked-token check: on `SQLAlchemyError`, discard the poisoned scoped session (`settings.Session.remove()`) and retry once on a fresh connection. Mirrors the pattern in `FabAuthManager.deserialize_user` (#62919).

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
+++ b/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
@@ -23,10 +23,12 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 from enum import Enum
+from contextlib import suppress
 from functools import cache, cached_property
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar
 
 from jwt import InvalidTokenError
+from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy import select
 
 from airflow.api_fastapi.auth.managers.models.base_user import BaseUser
@@ -52,6 +54,7 @@ from airflow.models.revoked_token import RevokedToken
 from airflow.models.team import Team, dag_bundle_team_association_table
 from airflow.typing_compat import Unpack
 from airflow.utils.log.logging_mixin import LoggingMixin
+from airflow import settings
 from airflow.utils.session import NEW_SESSION, provide_session
 
 if TYPE_CHECKING:
@@ -161,8 +164,20 @@ class BaseAuthManager(Generic[T], LoggingMixin, metaclass=ABCMeta):
             log.error("JWT token is not valid: %s", e)
             raise e
 
-        if (jti := payload.get("jti")) and RevokedToken.is_revoked(jti):
-            raise InvalidTokenError("Token has been revoked")
+        if jti := payload.get("jti"):
+            try:
+                revoked = RevokedToken.is_revoked(jti)
+            except SQLAlchemyError:
+                log.warning(
+                    "Revoked-token check failed on a stale DB session; "
+                    "discarding the session and retrying once.",
+                    exc_info=True,
+                )
+                with suppress(Exception):
+                    settings.Session.remove()
+                revoked = RevokedToken.is_revoked(jti)
+            if revoked:
+                raise InvalidTokenError("Token has been revoked")
 
         try:
             return self.deserialize_user(payload)


vincbeck (approved):

kaxil:

Note for @vincbeck: this PR carried your approval. It is being closed as part of the batch cleanup above, not on the merits of the diff. Happy to reopen if you would rather see it land.

PR #71399, opened 2026-08-11
Verify TypeScript SDK package contents

jason810496:
## Why

The TypeScript SDK package needs a fail-closed check that prevents unintended source, test, configuration, or credential files from reaching npm.

## How

- Build and pack the SDK into a temporary directory.
- Enforce an explicit runtime artifact allowlist.
- Install the tarball into an isolated consumer and smoke-test its public import and executable.

## What

- Add the `verify:package` command and manual prek hook.
- Omit JavaScript and declaration source maps from the published build.
- Document local package verification.

---

- [x] Yes, with help of Codex GPT-5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/ts-sdk/.pre-commit-config.yaml
+++ b/ts-sdk/.pre-commit-config.yaml
@@ -74,3 +74,19 @@ repos:
         additional_dependencies: ['pnpm@10.28.1']
         pass_filenames: false
         require_serial: true
+      - id: verify-ts-sdk-package
+        name: Verify TypeScript SDK package artifact
+        entry: ./scripts/ci/prek/verify_ts_sdk_package.py
+        language: node
+        stages: [manual]
+        files: |
+          (?x)
+          ^src/.*\.ts$|
+          ^scripts/verify-package\.mjs$|
+          ^package\.json$|
+          ^pnpm-lock\.yaml$|
+          ^pnpm-workspace\.yaml$|
+          ^tsconfig(\.build)?\.json$
+        additional_dependencies: ['pnpm@10.28.1']
+        pass_filenames: false
+        require_serial: true
--- a/ts-sdk/.pre-commit-config.yaml
+++ b/ts-sdk/.pre-commit-config.yaml
@@ -64,3 +64,19 @@ repos:
         additional_dependencies: ['pnpm@10.28.1']
         pass_filenames: false
         require_serial: true
+      - id: verify-ts-sdk-package
+        name: Verify TypeScript SDK package artifact
+        entry: ./scripts/ci/prek/verify_ts_sdk_package.py
+        language: node
+        stages: [manual]
+        files: |
+          (?x)
+          ^src/.*\.ts$|
+          ^scripts/verify-package\.mjs$|
+          ^package\.json$|
+          ^pnpm-lock\.yaml$|
+          ^pnpm-workspace\.yaml$|
+          ^tsconfig(\.build)?\.json$
+        additional_dependencies: ['pnpm@10.28.1']
+        pass_filenames: false
+        require_serial: true
--- a/ts-sdk/README.md
+++ b/ts-sdk/README.md
@@ -223,6 +223,6 @@ Without a local pnpm install, [prek](https://prek.j178.dev) can compile the SDK
 or verify the package with its own managed node + pnpm toolchain:
 
 ```bash
-prek run compile-ts-sdk
-prek run --hook-stage manual verify-ts-sdk-package
+prek run compile-ts-sdk --all-files
+prek run --hook-stage manual verify-ts-sdk-package --all-files
 ```
--- a/ts-sdk/README.md
+++ b/ts-sdk/README.md
@@ -221,6 +221,7 @@ pnpm install
 pnpm test
 pnpm run typecheck
 pnpm run build
+pnpm run verify:package
 ```
 
 The committed lockfile and `pnpm-workspace.yaml` define the dependency security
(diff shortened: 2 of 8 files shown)


ts-sdk/README.md
```diff
@@ -207,9 +208,21 @@ can enter the lockfile, transitive dependencies cannot use Git or arbitrary
...
+rejects them rather than shipping dangling maps. Local `pnpm run typecheck`
+still uses the base config, so editor tooling is unaffected.
+
 Without a local pnpm install, [prek](https://prek.j178.dev) can compile the SDK
-with its own managed node + pnpm toolchain:
+or verify the package with its own managed node + pnpm toolchain:
 
 ```bash
 prek run compile-ts-sdk
+prek run --hook-stage manual verify-ts-sdk-package
```
phanikumv:
```suggestion
  prek run compile-ts-sdk --all-files
  prek run --hook-stage manual verify-ts-sdk-package --all-files
  ```

Comment (phanikumv) [reply ts-sdk/README.md]: The command needs ` --all-files`. The compile line above has the same problem and is pre-existing, but since you're editing this block anyway it's worth fixing both. Naming the hook selects it regardless of stage, so `compile-ts-sdk` needs `--all-files`.

Comment (jason810496) [reply ts-sdk/README.md]: I just added `--all-files` to both commands. Thanks!

File: ts-sdk/scripts/verify-package.mjs [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+  );
+  run("npm", ["install", "--ignore-scripts", "--no-audit", "--no-fund", metadata.filename], {
+    cwd: consumerDirectory,
+  });
+  run(
+    "node",
+    [
+      "--input-type=module",
+      "--eval",
+      `const sdk = await import(${JSON.stringify(packageJson.name)}); if (typeof sdk.registerTask !== "function") process.exit(1);`,
```
amoghrajesh: I don't see `registerTask` defined. I think its `registerDags` / `serveDags` now?

ts-sdk/README.md
```diff
@@ -207,9 +208,21 @@ can enter the lockfile, transitive dependencies cannot use Git or arbitrary
 tarball sources, and only explicitly approved dependencies can run lifecycle
 build scripts. Review changes to both files together when updating dependencies.
 
+`verify:package` creates the npm tarball, rejects files outside the published
+runtime allowlist, installs it into a clean temporary project, and smoke-tests
+every `exports` entry point and the `bin` executable. The required paths are
```
amoghrajesh, line 235: The code imports only the package root (`verify-package.mjs:108-113`, `await import(packageJson.name)`). The `./coordinator` subpath is never imported; `collectEntryPoints` only puts `dist/coordinator/index.js` and `dist/coordinator/index.d.ts` into the presence list. Presence is not resolution, a subpath can be present and still fail to import if exports is misspelled or an internal import is broken.

Either fix the sentence to say the entry points are checked for presence and the root is imported, or loop the import over the exports keys. The second is better and is a small change, since collectEntryPoints already walks exports.

ts-sdk/.pre-commit-config.yaml
```diff
@@ -64,3 +64,19 @@ repos:
...
+          (?x)
+          ^src/.*\.ts$|
+          ^scripts/verify-package\.mjs$|
+          ^package\.json$|
+          ^pnpm-lock\.yaml$|
+          ^pnpm-workspace\.yaml$|
+          ^tsconfig(\.build)?\.json$
+        additional_dependencies: ['pnpm@10.28.1']
+        pass_filenames: false
+        require_serial: true
```
amoghrajesh, line 92: Nothing runs this check automatically, so it does not yet do what the PR body says. The hook is `stages: [manual]`, CI runs no ts-sdk manual hooks, and the release workflow proposed in #71843 runs `pnpm run lint, format:check, typecheck, and test` before `npm pack` - not `verify:package`. A contributor has to remember to run it, and the one time it mattered nobody did.

The natural home is #71843 based build job, one line before npm pack. That is also where it has the most value: it would verify the exact tarball that gets published. Since both PRs are open and from the same author, they can be sequenced.

ts-sdk/scripts/verify-package.mjs
amoghrajesh, line 1: has no tests

ts-sdk/scripts/verify-package.mjs
jason810496: Addressed in 16315ecd.

ts-sdk/README.md
jason810496: Addressed in 6808cfe8.

ts-sdk/scripts/verify-package.mjs
jason810496: Addressed in 3dfaf4d7.

ts-sdk/.pre-commit-config.yaml
jason810496: Addressed in https://github.com/apache/airflow/pull/71843/changes/be302f92f72c16c24ab7c36ad7e454ba217ffda6 on PR 71843, which runs the check in the release build job before the tarball is packed.

jason810496 (commented):
Thanks for the review Phani and Amogh, the PR is now ready for re-review.

phanikumv (approved):

PR #71403, opened 2026-08-11
Support per-run cost limits in common.ai LLM and Agent operators

Lee-W:
## Why
`usage_limits=UsageLimits(...)` already caps requests and tokens. But we still have the following 3 gaps 

1. **Cost could not be expressed at all.** `UsageLimits.cost_limit` and `RunUsage.cost` landed in pydantic-ai 2.23.0, but this provider's floor was `>=2.0.0`. On any resolved 2.0–2.22 install, `cost_limit` does not exist — a Dag author can cap tokens but has no way to cap spend, and no way to see what a run cost afterwards.
2. **No `UsageLimits` field could be templated.** A `UsageLimits` instance has neither `resolve` nor `template_fields`, so Airflow's template walk is a no-op on it. That blocks *every* field, not just cost: a `request_limit` or `input_tokens_limit` that follows the environment had no way to be written.
3. **Where the cap stops being a guarantee was undocumented.** Each of those cases reads as a broken promise to anyone who assumed otherwise.

## What

- **Raise the pydantic-ai floor to 2.23.0** so `cost_limit` / `RunUsage.cost` are guaranteed present. That alone unlocks cost enforcement: the operators already forward `usage_limits` untouched to `run_sync`, so there is no interception logic of our own.
- **`usage_limits` now accepts a dict and is a `template_fields` entry.** Passing `{"cost_limit": "{{ params.budget }}", "request_limit": 5}` makes every field templatable in one shot — a future `tool_calls_limit` needs no new operator parameter. A `UsageLimits` instance still passes through by identity: an author who built the object owns its values.
- **`coerce_usage_limits()` builds the real `UsageLimits` at execute time**, parsing each value per `UsageLimits`'s own dataclass field type rather than a hand-maintained list. A rendered value the author did not intend — a Variable that exists but is empty renders to `""`, and a typo renders to an arbitrary non-numeric string — raises a `ValueError` naming the offending field before any model call, instead of silently disabling that limit or blowing up deep inside pydantic-ai.
- **Log the run cost** when the model reports one.
- **Document where the cap is not a guarantee**: the request that crosses the limit still completes and still bills; self-hosted or unpriced models report no cost at all, so the  limit silently does nothing; and a durable replay re-counts cached spend against a fresh attempt's budget.

 

---

---


providers/common/ai/docs/index.rst
```diff
@@ -204,7 +204,7 @@ PIP package                                 Version required
 ``apache-airflow``                          ``>=3.0.0``
 ``apache-airflow-providers-common-compat``  ``>=1.15.0``
 ``apache-airflow-providers-standard``       ``>=1.12.1``
-``pydantic-ai-slim``                        ``>=2.0.0``
+``pydantic-ai-slim``                        ``>=2.23.0``
```
Lee-W, line 233: I guess we could just upgrade this version since we're still in 0.x.x and the cost feature is something that would be widely used

uranusjr:
We should add some cases on operator tests that actually use `usage_limit`. The `resolve_usage_limit` function itself looks fine, but there’s no evidence *how it’s used* is also fine.

Lee-W:
Sounds good. Just added operator-level tests so the cap is enforced by pydantic-ai instead of a mock

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import annotations
+
+import dataclasses
+from decimal import Decimal, InvalidOperation
+
+from pydantic_ai.usage import UsageLimits
+
+
+def resolve_usage_limits(
+    usage_limits: UsageLimits | None, max_cost: Decimal | float | str | None
```
amoghrajesh: Why dont we consider widening `usage_limits` itself to accept `UsageLimits | dict | None`, add `usage_limits` to `template_fields` (our jinja helpers renders every value inside the dict automatically, no custom code needed for that part), then in `execute()` you can coerce the rendered dict into a real `UsageLimits(**rendered)`, reusing the same defensive per field parsing that this PR already wrote for `cost_limit` specifically, applied to whichever fields the dict contains.

This is genuinely the pattern KPO and similar operators already use (accept a templatable dict/spec, build the real object at execute time).

It fixes the actual gap ("no UsageLimits field can be templated") once, for every knob like `request_limit`, `tool_calls_limit`, `total_tokens_limit`, `cost_limit` instead of adding a one off scalar parameter (`max_cost`) that only covers cost. If someone next needs a templated `request_limit`, this PR's pattern means writing another `max_requests` parameter but the dict approach needs nothing new.

vikramkoka:
@Lee-W 
I am struggling to understand the why here at least based on the PR description. 

The token limits are enforced and I ran into this with the AIP-tracker example, which is why this piece of code was explicitly added by me there 
 usage_limits=UsageLimits(
      request_limit=5,
      input_tokens_limit=20_000,
      output_tokens_limit=4_000,
  )

There are a lot of files touched in this PR, so I didn't go through and check all the code changes, but at least the PR description seems off?

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
vikramkoka: I don't think it is worth mixing those two concepts here. 
The KPO limits are generally based on system parameters which is generally a different persona from Token limits, and are constrained by different variants.

Lee-W:
Yes, the PR description is outdated after several iteration. Just update it to the latest change

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
Lee-W: `max_cost` is now gone. `usage_limits` accepts `UsageLimits | dict | None`, is a `template_fields` entry, and `coerce_usage_limits()` builds the real object in `execute()`, driving per-field parsing off `UsageLimits`'s own dataclass field types so a new field of an already-supported type needs no change here.

---

The only thing borrwed from KPO is the idea that "accept a templatable dict and build the object at execution time".

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+    # math.isfinite converts its argument to float first, which overflows a large
+    # int into OverflowError and raises outright on a Decimal signaling NaN --
+    # neither looks like "not finite", they look like an unhandled crash. Decimal
+    # has no float-sized exponent limit either, so a huge-but-finite Decimal must
+    # not be misreported as non-finite just because float can't represent it.
+    if isinstance(value, Decimal):
+        return value.is_finite()
+    if isinstance(value, int):
+        return True
+    return math.isfinite(value)
```
ColtenOuO, line 125: If I'm not mistaken, if value is an invalid native value such as `[]`, it will still reach `_validate_range()` and eventually raise a raw `TypeError` from `math.isfinite([])`, rather than the documented field-specific `ValueError`.

Perhaps this is a small detail we should keep an eye on and pay special attention to.

providers/common/ai/docs/operators/llm.rst
```diff
@@ -119,16 +119,46 @@ calls within a single task.
...
 - ``total_tokens_limit`` — combined input + output cap.
 - ``tool_calls_limit`` — max tool invocations (``AgentOperator`` only).
+- ``cost_limit`` — a ``Decimal`` cap on the run's estimated USD cost. This is **not** a
+  hard guarantee against overspend: the response that crosses the limit has already been
+  produced and billed — pydantic-ai checks the accumulated cost *after* each response and
+  then fails the run with ``UsageLimitExceeded``. It protects you from further spend, not
+  from the request that broke the budget; even a single-request run fails as soon as that
+  request's cost pushes the total over the limit. For self-hosted or unknown
+  models (e.g. Ollama, custom endpoints) pydantic-ai cannot price the response, so cost
+  is ``None`` and ``cost_limit`` silently has no effect (a ``CostNotFoundWarning`` is
```
ColtenOuO, line 156:
```suggestion
  is ``None`` and ``cost_limit`` has no effect without halting execution (a ``CostNotFoundWarning`` is
```

The original phrasing feels a bit contradictory: it says `silently has no effect`; but a warning is actually emitted.

ColtenOuO (commented):
Overall, I think the approach looks great and the test coverage is very thorough.

providers/common/ai/docs/operators/llm.rst
Lee-W: yep, agree it makes things more clear. thanks!

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
Lee-W: since we're checking [here](https://github.com/apache/airflow/pull/71403/changes#diff-6244833c78931d8bc47724577d529b43e1771f086f2ec1fcdc69d511bd560567R178), I think it's not going to happen. WDYT?

Lee-W:
https://github.com/apache/airflow/pull/72762 should fix the doc building failure

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+def _truncated_repr(value: Any, limit: int = 100) -> str:
+    # A container-shape error on a templated field usually means the render
+    # produced something that merely looks right (e.g. a long string that reads
+    # like a dict literal) -- the author needs to see what actually came out, not
+    # just its type. Truncate so a large rendered blob doesn't bloat the exception.
+    text = repr(value)
+    return text if len(text) <= limit else f"{text[:limit]}..."
+
+
+def _coerce_value(field: str, value: Any) -> Any:
```
amoghrajesh, line 169: A bad value here will crash with wrong error type. For example: `usage_limits={"cost_limit": []}`, falls through to `math.isfinite([])` and raises `TypeError`. Suggest checking the value's type before the finite check

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+        # the same finite check below as the templated-string path -- a bare
+        # ``float`` must not bypass the one safety promise this module makes.
+        value = _coerce_decimal(field, str(value))
+
+    if field_type in (Decimal, int):
+        _validate_range(field, value)
+    return value
+
+
+def coerce_usage_limits(usage_limits: UsageLimits | dict[str, Any] | None) -> UsageLimits | None:
```
amoghrajesh, line 262: Does this have to be a public interface? Only called inside common/ai code paths.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
amoghrajesh: Suggest adding a test for this too.

providers/common/ai/src/airflow/providers/common/ai/example_dags/example_llm.py
```diff
@@ -148,6 +153,30 @@ def example_llm_operator_usage_limits():
...
+        task_id="capped_summary",
+        prompt="Summarize the trade-offs of a message queue vs. direct HTTP calls in three bullet points.",
+        llm_conn_id="pydanticai_default",
+        system_prompt="You are a concise technical reviewer.",
+        # A plain dict lets every UsageLimits field be templated -- e.g. driven by
+        # an Airflow Variable so the budget can change per environment without
+        # editing the Dag. This caps a single task run, not a day's total spend --
+        # each run gets the full budget again.
+        usage_limits={
+            "cost_limit": "{{ var.value.llm_cost_cap_per_task }}",
```
amoghrajesh: If the variable doesn't exist, does it render to empty string or raises KeyError? Maybe we define that variable in this file?

airflow-ctl/src/airflowctl/api/datamodels/auth_generated.py
```diff
@@ -1,6 +1,6 @@
 # generated by datamodel-codegen:
 #   filename:  v2-simple-auth-manager-generated.yaml
-#   version:   0.74.0
+#   version:   0.76.1
```
kaxil: why is this changed?

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+    # produced something that merely looks right (e.g. a long string that reads
+    # like a dict literal) -- the author needs to see what actually came out, not
+    # just its type. Truncate so a large rendered blob doesn't bloat the exception.
+    text = repr(value)
+    return text if len(text) <= limit else f"{text[:limit]}..."
+
+
+def _coerce_value(field: str, value: Any) -> Any:
+    # ``None`` is always the author's explicit choice to disable that limit --
+    # Jinja never produces ``None`` from a string template -- so it passes through.
```
kaxil: With `render_template_as_native_obj=True` Jinja does produce a real `None`: `{{ params.budget }}` where `params.budget` is `None` renders to `None`, not `"None"`. The `is None` check below then passes it straight through and the cap is silently off, which is the one outcome the docstring says this module exists to prevent. Under the default environment the comment holds, so this only bites native-mode Dags, but a rendered `None` and an author-written `None` are indistinguishable at this point.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+    if value is None:
+        return value
+
+    field_type = _FIELD_TYPES[field]
+    # Only ``str`` values are converted: Jinja only ever renders a scalar leaf to
+    # ``str``, so a non-``str`` value is exactly what the author wrote (a literal
+    # ``Decimal``, ``int``, or ``bool``) and is passed through unchanged.
+    if isinstance(value, str):
```
kaxil, line 183: Only strings get coerced, so every other value reaches `UsageLimits` with just the finite/negative check behind it. `bool` is an `int` subclass, so `{"request_limit": False}` builds `UsageLimits(request_limit=False)` and the run then dies before the first model call with `UsageLimitExceeded: The next request would exceed the request_limit of False`. That is the confusing-failure-inside-pydantic-ai case this module is here to stop, and `Variable.get(..., deserialize_json=True)` with `false` written where `null` was meant gets you there. A plain float lands on the int fields the same way, so `request_limit` can end up non-integral.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+        return True
+    if normalized in _FALSE_LIKE:
+        return False
+    raise ValueError(
+        f"usage_limits[{field!r}] must be one of {sorted(_TRUE_LIKE | _FALSE_LIKE)} "
+        f"(got {value!r}); if it is templated, check the rendered value."
+    )
+
+
+_FIELD_TYPES: dict[str, type] = _build_field_types()
```
kaxil: This runs at import, and `_resolve_field_type` raises on any annotation that is not `X` or `X | None`. All six operator modules import this one at module scope, so a future `UsageLimits` field typed `Literal[...]` or `list[int] | None` would make every common.ai operator unimportable and every Dag using the provider fail to parse, including Dags that never set `usage_limits`. Nothing trips it on the versions I checked, but the floor has no upper bound and pydantic-ai ships a minor most weekdays. Resolving the field type lazily inside `_coerce_value` keeps the loud failure and scopes it to authors who actually pass that field.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,228 @@
...
+
+def coerce_usage_limits(usage_limits: UsageLimits | dict[str, Any] | None) -> UsageLimits | None:
+    """
+    Coerce a rendered ``usage_limits`` dict into a real ``UsageLimits`` instance.
+
+    A ``UsageLimits`` instance has neither ``resolve`` nor ``template_fields``, so
+    Airflow's template walk is a no-op on it even though ``usage_limits`` is in the
+    operators' ``template_fields``. Passing a plain dict instead lets every field be
+    templated, but the rendered value is then not in the Dag author's control: an
+    unset Airflow Variable renders to ``""``, and a typo renders to an arbitrary
```
kaxil: An unset Variable does not render to `""`. `VariableAccessor.__getattr__` calls `_get_variable`, which raises `AirflowRuntimeError`, and Jinja's sandboxed getattr only swallows `AttributeError`, so rendering fails before `coerce_usage_limits` is reached. The case that actually gives you `""` is a Variable that exists with an empty value. The same sentence is in `operators/llm.py:86`, `operators/agent.py:154` and `docs/operators/llm.rst:133`, and it is the open question on `example_dags/example_llm.py:169`.

providers/common/ai/tests/unit/common/ai/durable/test_replay_cost.py
```diff
@@ -0,0 +1,132 @@
...
+        """A retry that makes no new model calls can still raise UsageLimitExceeded,
+        purely from replayed cost -- because check_cost() sees the run's cumulative
+        usage, not "money spent in this attempt"."""
+        # Attempt 1 stays comfortably under budget so it completes normally.
+        await _run_one_attempt(durable_storage, cost_limit=PRICED_COST * 2)
+
+        # Attempt 2 sets a limit below the already-paid-for replayed cost: zero new
+        # spend, yet the replayed step alone pushes the cumulative usage over it.
+        cost_limit = PRICED_COST / 2
+        with pytest.raises(UsageLimitExceeded):
```
kaxil, line 119: Nothing here asserts the "zero new model calls" half of the claim. `_run_one_attempt` returns the counter, but it is discarded when the exception propagates, so this passes just as well on a cold cache where attempt 2 makes a live call. Holding the counter in the test scope and asserting `cached_model == 0` and `replayed_model == 1` next to the `raises` would make it test its own docstring.

providers/common/ai/docs/operators/agent.rst
```diff
@@ -484,10 +484,27 @@ Parameters
...
+  rendered value doesn't parse. Use it to cap requests, tokens, or tool calls
+  per task -- agents are particularly prone to runaway tool loops, so
+  ``tool_calls_limit`` is a useful guardrail. It also supports a per-run USD
+  ``cost_limit``; see :ref:`howto/operator:llm` for the caveats (not a hard
+  guarantee, silently inert for unpriced models) and an example. Default
+  ``None``.
+
+  .. warning::
+     With ``durable=True``, a task retry replays cached model steps instead of
+     re-calling the model -- but pydantic-ai still adds each replayed step's
```
kaxil, line 499: `CachingModel.request` is the one place that knows the response is a replay (it bumps `counter.replayed_model` right before returning the cached object), and `ModelResponse` is a dataclass, so `dataclasses.replace(cached, usage=RequestUsage())` there would make a retry's budget reflect what that attempt actually spent. It would fix the same double-count for `request_limit`, `tool_calls_limit` and the token limits, which have it too but are not mentioned here. The cost is that the run-summary log would then under-report the conversation's real token counts. Is documenting it preferred over fixing it?

providers/common/ai/tests/unit/common/ai/conftest.py
```diff
@@ -31,15 +31,22 @@ def isolate_hook_lineage_collector(hook_lineage_collector):
     return None
 
 
-def make_mock_run_result(output):
+def make_mock_run_result(output, *, cost=None):
```
kaxil: This helper has no callers: every one of the 12 test modules defines its own private `_make_mock_run_result`, which is why the same `cost=None` fix had to be hand-applied 12 times in this PR. The hunk is inert as written, and the next test module copied from the local pattern will log `LLM run cost: $<MagicMock ...>` with nothing to catch it. Worth either pointing the modules at this one or dropping it.

providers/common/ai/src/airflow/providers/common/ai/utils/logging.py
```diff
@@ -48,9 +49,10 @@ def log_run_summary(logger: Logger | logging.Logger, result: AgentRunResult[Any]
         usage.output_tokens,
         usage.total_tokens,
     )
+    if usage.cost is not None:
+        logger.info("LLM run cost: $%s (USD, best-effort)", usage.cost)
```
kaxil: `%s` on a small `Decimal` gives scientific notation, so a cheap single-call run logs `LLM run cost: $7.5E-7`. `format(cost, "f")` prints it as a plain decimal instead.

providers/common/ai/docs/operators/llm.rst
kaxil: This still reads `silently has no effect` immediately before saying a `CostNotFoundWarning` is emitted. Looks like the suggestion you accepted above did not make it into the last push.

providers/common/ai/docs/operators/llm.rst
```diff
@@ -119,16 +119,46 @@ calls within a single task.
...
+  emitted instead of a failure). And like the other knobs above, setting ``cost_limit``
+  alone still inherits the ``request_limit=50`` default — see the ``request_limit`` note
+  above. Note that ``cost_limit`` only caps the operator's own LLM calls --
+  the meta-agent that ``LLMRetryPolicy`` runs to classify a failed task is a separate,
+  uncapped LLM call; see :doc:`../retry_policies`.
```
kaxil: `pydantic-ai-harness` ships a `SpendLimits` capability that covers two of the gaps this bullet documents: the budget is carried across runs (day / month / total windows, optionally in a shared Redis counter), so a run is refused before its first request instead of every run getting the full cap again, and it prices responses through `genai-prices` with `on_unpriced='raise'` rather than going quiet when the provider reports no cost. `agent_params` is forwarded to the `Agent` constructor, so `capabilities=[SpendLimits(...)]` already works, though harness only arrives via the `code-mode` extra today. Worth pointing this bullet at it for anyone who wants a budget rather than a per-run ceiling? https://pydantic.dev/docs/ai/harness/spend/

kaxil: Although, we need to check how that Budget thing works i.e where does it store the ongoing cost

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,278 @@
...
+        # finite check in _validate_range below reports them as "not finite" --
+        # checking is_integer() first would misreport them as "not an integer".
+        if not value.is_integer():
+            raise ValueError(
+                f"usage_limits[{field!r}] must be an integer (got {value!r}); "
+                "if it is templated, check the rendered value."
+            )
+        value = int(value)
+
+    if field_type in (Decimal, int):
```
kaxil, line 243: Gating on `field_type in (Decimal, int)` leaves the one `bool` field with no shape check at all, so any non-`str` value reaches `UsageLimits` untouched. Running this module directly, `count_tokens_before_request` accepts `[]`, `{}`, `0`, `1.5` and `Decimal("0")`, all of which pydantic-ai then reads by truthiness, so the pre-flight token check is off while the Dag says it is on. That is the outcome the comment at line 96-100 says this module exists to prevent, just reached through a native-mode render rather than a string. The same gate lets `request_limit=Decimal("3.5")` through, slipping the non-integral check you added right above for a plain `3.5`. Keying the guard on the field's own declared type (`if not isinstance(value, field_type): raise ...`) covers all three.

providers/common/ai/docs/operators/llm.rst
```diff
@@ -119,16 +119,51 @@ calls within a single task.
...
+  is ``None`` and ``cost_limit`` has no effect without halting execution (a
+  ``CostNotFoundWarning`` is emitted instead of a failure). And like the other knobs
+  above, setting ``cost_limit``
+  alone still inherits the ``request_limit=50`` default — see the ``request_limit`` note
+  above. Note that ``cost_limit`` only caps the operator's own LLM calls --
+  the meta-agent that ``LLMRetryPolicy`` runs to classify a failed task is a separate,
+  uncapped LLM call; see :doc:`../retry_policies`. For a budget tracked across runs
+  rather than this per-run cap, pass a
+  `SpendLimits <https://pydantic.dev/docs/ai/harness/spend/>`__ capability via
+  ``agent_params`` (see :ref:`capabilities-passthrough`; requires the ``code-mode`` extra).
```
kaxil: Correcting my own pointer above, which is what put this sentence here: `SpendLimits.store` defaults to `InMemorySpendStore`, whose docstring says "It does not enforce a budget across processes: every worker of a queue would keep its own count, which is what a shared store such as `RedisSpendStore` is for." Every Airflow task instance is a fresh worker process, so a `window='day'` budget resets on every run, which is the per-run ceiling this sentence is positioning against. The cross-run claim only holds once a shared store is passed, so the bullet should name that rather than leave the default implied. Also worth bumping the extra: `pydantic_ai_harness.spend` first appears in harness v0.17.0 and is absent in v0.16.0, while the `code-mode` floor is `pydantic-ai-harness[codemode]>=0.3.0`.

providers/common/ai/docs/operators/llm.rst
```diff
@@ -119,16 +119,51 @@ calls within a single task.
...
 - ``total_tokens_limit`` — combined input + output cap.
 - ``tool_calls_limit`` — max tool invocations (``AgentOperator`` only).
+- ``cost_limit`` — a ``Decimal`` cap on the run's estimated USD cost. This is **not** a
+  hard guarantee against overspend: the response that crosses the limit has already been
+  produced and billed — pydantic-ai checks the accumulated cost *after* each response and
+  then fails the run with ``UsageLimitExceeded``. It protects you from further spend, not
+  from the request that broke the budget; even a single-request run fails as soon as that
+  request's cost pushes the total over the limit. For self-hosted or unknown
+  models (e.g. Ollama, custom endpoints) pydantic-ai cannot price the response, so cost
+  is ``None`` and ``cost_limit`` has no effect without halting execution (a
```
kaxil: "has no effect without halting execution" reads as "has no effect unless execution halts", which is the opposite of the point. What you want to say is that it is inert but not silent, e.g. "`cost_limit` is not enforced; a `CostNotFoundWarning` is emitted rather than failing the run". The same phrase landed in `agent.rst:494` and `example_dags/example_llm.py:145`.

providers/common/ai/docs/operators/llm.rst
kaxil: Maybe for simplicity we just target per task -- cost limits for this PR and for 1.0 and can come back to limits across tasks -- per agent -- after 1.0

providers/common/ai/docs/operators/llm.rst
```diff
@@ -119,16 +119,48 @@ calls within a single task.
...
+  token cap.
 - ``input_tokens_limit`` / ``output_tokens_limit`` — per-run token caps.
 - ``total_tokens_limit`` — combined input + output cap.
 - ``tool_calls_limit`` — max tool invocations (``AgentOperator`` only).
+- ``cost_limit`` — a ``Decimal`` cap on the run's estimated USD cost. This is **not** a
+  hard guarantee against overspend: the response that crosses the limit has already been
+  produced and billed — pydantic-ai checks the accumulated cost *after* each response and
+  then fails the run with ``UsageLimitExceeded``. It protects you from further spend, not
+  from the request that broke the budget; even a single-request run fails as soon as that
+  request's cost pushes the total over the limit. For self-hosted or unknown
```
kaxil, line 155: Pricing keys on the model name, not the endpoint, so this is wrong for the setup the provider's own docs recommend. `_cost.py` falls back to `provider_id=provider_name` on `LookupError` (lines 41 and 48), and at the 2.23.0 floor `gpt-4o` behind `http://vllm.internal:8000/v1` prices fine while `llama3.2` on the same endpoint raises. Since `self_hosted_models.rst` tells vLLM users `openai:<model>` is the only option, that setup does get priced, at OpenAI list rates rather than the gateway's real cost, so the cap can fire on money nobody spent. `agent.rst:493` already keys the sentence on whether the model can be priced, which is the wording that holds.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,307 @@
...
+        parsed = Decimal(value)
+    except InvalidOperation:
+        raise ValueError(
+            f"usage_limits[{field!r}] must be a number (got {value!r}); "
+            "if it is templated, check the rendered value."
+        ) from None
+    return parsed
+
+
+def _coerce_int(field: str, value: str) -> int:
```
kaxil, line 86: The templated-string path rejects integral values the native path accepts. On `request_limit` at 2.23.0: `5.0` gives 5 but `"5.0"` raises, `1e3` gives 1000 but `"1e3"` raises, `Decimal("5.0")` gives 5 but `"5.000"` raises. Jinja `/` is true division, so `{"request_limit": "{{ params.total / params.n }}"}` hard-fails under the default renderer and succeeds under `render_template_as_native_obj=True`, the same template decided by a Dag-level flag, in the module whose job is making templated values behave like literals. Falling back to the same integral test once `int(value)` raises would line the two up and leave `"inf"` and `"abc"` on the current message.

providers/common/ai/src/airflow/providers/common/ai/utils/usage.py
```diff
@@ -0,0 +1,307 @@
...
+from collections.abc import Callable
+from decimal import Decimal, InvalidOperation
+from typing import Any
+
+from pydantic_ai.usage import UsageLimits
+
+
+def _resolve_field_type(field: str, hint: Any) -> type:
+    # Only ``X`` or ``X | None`` are supported shapes -- anything else (a Union of
+    # two real types, a parameterized generic, a Literal, ...) has no single
```
kaxil, line 34: A single-parameter generic resolves to its element type rather than raising: `list[int]` gives `int` and `set[str]` gives `str`, since both pass the `len(args) == 1` gate and `isinstance(int, type)` holds. `dict[str, int]` raises on arity, not on being a generic, so `test_parameterized_generic_raises` is green for the wrong reason. A `typing.get_origin()` check ahead of the arity reduction, rejecting any origin that is not a union, closes it. Nothing trips this at 2.23.0, but future fields are the whole reason the resolution is lazy.

kaxil:
Two non-blocking follow-ons to the review above, so everything lands in one push rather than another round.

`cost_limit` is per task attempt. Each Airflow attempt re-renders the template fields and hands a fresh `UsageLimits` to a fresh `RunUsage`, so `cost_limit=Decimal("0.50")` on a task with `retries=3` permits $2.00 across the task. `agent.rst` reasons about budget-versus-retry at length for `durable=True`, which leaves the plain case unmentioned right next to it. One sentence in the `cost_limit` bullet would cover it.

The coercion runs at the `run_sync` call rather than at the top of `execute`, so the `ValueError` this PR adds fires only after the expensive part. `llm_schema_compare.py:326` lands after `_build_schema_context()` at 314 has introspected every configured connection against every table, and the same shape is at `llm_sql.py:148` (after `_get_schema_context()`), `llm_file_analysis.py:150` (after the files are resolved, read and encoded) and `agent.py:466` (after `_build_durable_storage` and `_build_agent`). A typo'd or empty Variable burns all of that before failing. Hoisting `limits = coerce_usage_limits(self.usage_limits)` to the first statement of each `execute` and passing `limits` at the call would fix it.

One more that I could not anchor inline, since the lines are not in a hunk: `self_hosted_models.rst:296-299` says the provider prefixes were verified against pydantic-ai 2.5.0 and 2.10.0, both below the 2.23.0 floor this PR sets, in a file the PR already touches.

On CI: the `Latest Boto test: providers` failure is not from this PR. Its summary reports the failure under `Providers[amazon]` (1 failed, 4712 passed) and every `providers/common/ai/tests/` file passes in the same job.

kaxil:
Pushed three commits covering the open threads:

- `c8c326659` `usage.py` coercion: reject subscripted generics, including `list[int] | None`, which reduced to `list[int]` and resolved as a container type on Python 3.10; accept integral templated values so `"5.0"` behaves like a bare `5.0`; and bound the `Decimal` path so a huge exponent cannot outrun CPython's digit ceiling.
- `588bc6214` hoist the coercion above the expensive setup in each `execute()`, so an unparsable value fails before the file read, the schema introspection, or the agent build.
- `d3f898d2b` docs: pricing keys on the model name rather than the endpoint, each limit bounds one agent run, plus the `self_hosted_models.rst` baseline and the README extras table.

Cost cap against a real Anthropic model, 16 input and 4 output tokens, provider-reported cost `$0.00018` against a `cost_limit` of `1E-7`:

Cost limit exceeded on a real model call

Fail-fast with no connection configured at all. The task fails on the named `ValueError` from `llm.py:180`, the hoisted coercion, rather than on the missing connection:

Unparsable cost limit fails before the model call

kaxil (approved):

PR #71423, opened 2026-08-11
disable external entities in SAML idp response parsing

Samin061:
`_fetch_saml_assertion_using_http_spegno_auth` parses the IdP HTTP response with lxml's default parser, which resolves external entities, so an IdP endpoint (a malicious or on-path one) can return a DOCTYPE whose `SYSTEM` entity reads a local file into the extracted assertion, or a nested-entity payload that exhausts worker memory. The response is remote content rather than trusted local config, so it should be parsed with entity resolution and network access off. This parses it with a hardened `XMLParser(resolve_entities=False, no_network=True, load_dtd=False)`.

---

---

### DIFF
--- a/providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
+++ b/providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
@@ -847,6 +847,39 @@ class TestAwsBaseHook:
 
         assert "TOPSECRET" not in str(assertion)
 
+    def test_saml_assertion_parsing_does_not_resolve_internal_entities(self):
+        internal_entity_response = (
+            b'<?xml version="1.0"?>'
+            b'<!DOCTYPE r [<!ENTITY payload "INJECTED">]>'
+            b"<r><assertion>&payload;</assertion></r>"
+        )
+
+        saml_config = {
+            "idp_url": "https://my-idp.local.corp",
+            "idp_auth_method": "http_spegno_auth",
+            "saml_response_xpath": "//assertion/text()",
+        }
+
+        orig_import = __import__
+
+        def import_mock(name, *args, **kwargs):
+            if name == "requests_gssapi":
+                return mock.Mock()
+            return orig_import(name, *args, **kwargs)
+
+        factory = BaseSessionFactory(conn=None)
+        with (
+            mock.patch("builtins.__import__", side_effect=import_mock),
+            mock.patch.object(factory, "_get_idp_response") as mock_idp,
+        ):
+            mock_idp.return_value.content = internal_entity_response
+            try:
+                assertion = factory._fetch_saml_assertion_using_http_spegno_auth(saml_config)
+            except ValueError:
+                assertion = ""
+
+        assert "INJECTED" not in str(assertion)
+
     @mock_aws
     def test_expand_role(self):
         conn = boto3.client("iam", region_name="us-east-1")
--- a/providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
+++ b/providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
@@ -812,6 +812,41 @@ class TestAwsBaseHook:
             ),
         ]
 
+    def test_saml_assertion_parsing_does_not_resolve_external_entities(self, tmp_path):
+        secret_file = tmp_path / "secret.txt"
+        secret_file.write_text("TOPSECRET")
+        xxe_response = (
+            '<?xml version="1.0"?>'
+            f'<!DOCTYPE r [<!ENTITY xxe SYSTEM "file://{secret_file}">]>'
+            "<r><assertion>&xxe;</assertion></r>"
+        ).encode()
+
+        saml_config = {
+            "idp_url": "https://my-idp.local.corp",
+            "idp_auth_method": "http_spegno_auth",
+            "saml_response_xpath": "//assertion/text()",
+        }
+
+        orig_import = __import__
+
+        def import_mock(name, *args, **kwargs):
+            if name == "requests_gssapi":
+                return mock.Mock()
+            return orig_import(name, *args, **kwargs)
+
+        factory = BaseSessionFactory(conn=None)
+        with (
+            mock.patch("builtins.__import__", side_effect=import_mock),
+            mock.patch.object(factory, "_get_idp_response") as mock_idp,
+        ):
+            mock_idp.return_value.content = xxe_response
+            try:
+                assertion = factory._fetch_saml_assertion_using_http_spegno_auth(saml_config)
+            except ValueError:
+                assertion = ""
+
+        assert "TOPSECRET" not in str(assertion)
+
     @mock_aws
     def test_expand_role(self):
         conn = boto3.client("iam", region_name="us-east-1")
--- a/providers/amazon/src/airflow/providers/amazon/aws/hooks/base_aws.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/hooks/base_aws.py
@@ -399,8 +399,11 @@ class BaseSessionFactory(LoggingMixin):
             )
             self.log.debug("idp_response.content= %s", idp_response.content)
             self.log.debug("xpath= %s", xpath)
-        # Extract SAML Assertion from the returned HTML / XML
-        xml = etree.fromstring(idp_response.content)
+        # Extract SAML Assertion from the returned HTML / XML. The response is remote
+        # content from the IdP endpoint, so parse it with external entity resolution and
+        # network access disabled to avoid XXE (local file disclosure, entity-expansion).
+        parser = etree.XMLParser(resolve_entities=False, no_network=True, load_dtd=False)
+        xml = etree.fromstring(idp_response.content, parser=parser)
         saml_assertion = xml.xpath(xpath)
         if isinstance(saml_assertion, list):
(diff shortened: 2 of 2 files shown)


seanghaeli (commented):
Good fix, one ask on the testing: please add another one that tests the internal entity path

Samin061:
Added test_saml_assertion_parsing_does_not_resolve_internal_entities. It uses an inline entity defined in the DOCTYPE (<!ENTITY payload "INJECTED">) instead of a SYSTEM one, and asserts the value isn't expanded into the extracted assertion. Confirmed it fails on the old default parser (the entity expands to INJECTED) and passes with the hardened parser. Good catch on the coverage gap.

o-nikolas:
@seanghaeli Looks like your feedback was addressed a few weeks ago, does this look good to approve now?

seanghaeli (approved):
Yes, looks good @o-nikolas

providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
```diff
@@ -812,6 +812,74 @@ def mock_assume_role_with_saml(**kwargs):
...
+
+        factory = BaseSessionFactory(conn=None)
+        with (
+            mock.patch("builtins.__import__", side_effect=import_mock),
+            mock.patch.object(factory, "_get_idp_response") as mock_idp,
+        ):
+            mock_idp.return_value.content = xxe_response
+            try:
+                assertion = factory._fetch_saml_assertion_using_http_spegno_auth(saml_config)
+            except ValueError:
```
o-nikolas, line 845: `etree.XMLSyntaxError` derives from `SyntaxError`, not `ValueError`, so on `main` this test *errors* rather than fails — and it errors on "entity not defined", not on a file disclosure. It passes with the fix only because the empty xpath result trips `raise ValueError("Invalid SAML Assertion")` a few lines down in the hook, which this `except` then swallows. So the test doesn't exercise the scenario it's named for.

Separately, `except ValueError: assertion = ""` followed by `assert "TOPSECRET" not in str(assertion)` is vacuously true for *any* ValueError raised anywhere in the call. Prefer `pytest.raises` on the specific expected outcome.

Suggest either dropping this test or rewriting it to assert the hardening directly.

providers/amazon/tests/unit/amazon/aws/hooks/test_base_aws.py
```diff
@@ -812,6 +812,74 @@ def mock_assume_role_with_saml(**kwargs):
...
+                assertion = ""
+
+        assert "TOPSECRET" not in str(assertion)
+
+    def test_saml_assertion_parsing_does_not_resolve_internal_entities(self):
```
o-nikolas, line 850: This is the genuine regression test — it yields `'INJECTED'` on main and `[]` with the fix, so it fails without the change for the right reason.

Since it and the external-entity test differ only in payload and expected substring, they'd be better consolidated into one `@pytest.mark.parametrize`d test (AGENTS.md asks for parametrize when tests differ only in input/expected values).

o-nikolas (commented):
Thanks for digging into this. I'd like to land the parser hardening, but the framing needs to change before it goes in — I couldn't reproduce either issue described.

**Neither claimed issue reproduces on the lxml version this provider requires.** The `python3-saml` extra pins `lxml>=6.0.0`; tested against lxml 6.1.2 / libxml2 2.14.6:

| parser | `<!ENTITY xxe SYSTEM "file://…/secret.txt">` |
|---|---|
| `etree.fromstring(content)` — main today | `XMLSyntaxError: Entity 'xxe' not defined` |
| `load_dtd=True, resolve_entities=True` | `['TOPSECRET']` — file read |
| this PR's parser | `[]` |

The file read requires `load_dtd=True`, which is neither the default nor what the current code passes. Entity-expansion DoS is already blocked too — a billion-laughs payload against the *default* parser raises `XMLSyntaxError: Maximum entity amplification factor exceeded`.

The one behaviour that does change is internal *literal* entity substitution (`<!ENTITY payload "INJECTED">` → `INJECTED` before, `[]` after). That isn't a security boundary: anyone able to inject a DOCTYPE into the response body can just write the value inline.

Two asks:

**1. Please reframe the description and commit message as defence-in-depth rather than a vulnerability fix.** I do want the change — explicitly pinning `resolve_entities=False, no_network=True, load_dtd=False` on remote content is good practice, and it stops a future libxml2 default change (or an lxml built against an older system libxml2) from silently reintroducing the exposure. But a commit in apache/airflow stating that an IdP could read local files through this provider will be picked up by CVE aggregators and cited later as a real past XXE in Airflow. The security model docs ask specifically that hardening opportunities not be described as vulnerabilities.

**2. Please call out the behaviour change.** A deployment whose IdP legitimately returns internal entities in the SAML response will now get `ValueError("Invalid SAML Assertion")` where it previously got the substituted value. Worth a line in the description.

Test notes inline.

---
Drafted-by: Kiro (claude-opus-5); reviewed by @o-nikolas before posting

o-nikolas:
Had a quick magpie review and it found some things, please have a look @Samin061

potiuk (changes requested):
The parser hardening is worth landing, but the vulnerability framing and the external-entity test both need to change before it can go in. I have rebased the branch on `main` (it was 898 commits behind, which the base-branch policy would have refused) so that whatever you push next lands on a mergeable base; the parser change and both tests came through unchanged and pass.

Substantively I agree with `@o-nikolas`'s read — I re-derived both of his test points statically against the hook and they hold. Adding two rule citations and one nuance below rather than relitigating.

### Framing: this is defence-in-depth, not a vulnerability fix

```python
# Extract SAML Assertion from the returned HTML / XML. The response is remote
# content from the IdP endpoint, so parse it with external entity resolution and
# network access disabled to avoid XXE (local file disclosure, entity-expansion).
parser = etree.XMLParser(resolve_entities=False, no_network=True, load_dtd=False)
```

`airflow-core/docs/security/security_model.rst`, *Hardening opportunities with no demonstrated exploitation path*:

> Where no path exists by which an actor reaches an outcome they were not already entitled to, the finding is hardening rather than a vulnerability.
>
> Hardening is handled in public: open a normal pull request or issue and it is reviewed like any other change. No CVE is allocated and no advisory is issued.

So opening this publicly was exactly right — but the wording needs to follow. Please reword the in-code comment, the commit message and the PR description to state the invariant without the exploit claim, for example: *"Remote IdP content — pin entity resolution and network access off so a future libxml2/lxml default change cannot silently reintroduce external-entity exposure."*

The commit message matters most: provider changelogs are regenerated from `git log`, so that line becomes the durable public artefact, and a commit in `apache/airflow` asserting that an IdP could read worker-local files through the Amazon provider will be picked up and cited later as a past Airflow XXE.

Worth saying explicitly in the reframed text: `no_network=True` and `load_dtd=False` are already lxml's defaults. The only flag that changes behaviour today is `resolve_entities=False`. Pinning all three is good practice; the description just should not imply all three close something.

### The external-entity test passes for the wrong reason

```python
        mock_idp.return_value.content = xxe_response
        try:
            assertion = factory._fetch_saml_assertion_using_http_spegno_auth(saml_config)
        except ValueError:
            assertion = ""

assert "TOPSECRET" not in str(assertion)
```

> **Flag any changed or added behaviour without a corresponding test, and flag tests that a reviewer cannot fail by reverting the PR's change.**

Tracing it through the hook: with the fix applied, `resolve_entities=False` leaves `&xxe;` unexpanded, `xml.xpath("//assertion/text()")` returns `[]`, and `base_aws.py` raises `ValueError("Invalid SAML Assertion")` — which this `except` swallows and rebinds to `""`. The final assertion is then vacuously true, and stays true for a `ValueError` raised anywhere else in the call. Either drop this test and keep the internal-entity one, which is the real regression test, or assert the outcome directly:

```python
with pytest.raises(ValueError, match="Invalid SAML Assertion"):
    factory._fetch_saml_assertion_using_http_spegno_auth(saml_config)
```

### Smaller observations

- **Parametrize the two tests.** They differ only in the payload bytes and the forbidden substring — around 30 duplicated lines of `saml_config` / `import_mock` / `with` block. `AGENTS.md`: *"Use `@pytest.mark.parametrize` for multiple similar inputs — consolidate tests that only differ in input/expected values into a single parametrized test."* Same as `@o-nikolas`'s second thread.
- **Spec the mocks.** `mock.Mock()` stands in for the entire `requests_gssapi` module, and `mock.patch.object(factory, "_get_idp_response")` has no `autospec=True`. `autospec=True` on the `patch.object` is the easy win.
- **Call out the behaviour change** (also `@o-nikolas`'s second ask): an IdP that legitimately returns internal literal entities now yields `ValueError("Invalid SAML Assertion")` instead of the substituted value. A line in the description is enough.
- **Test placement, nit.** Both tests construct `BaseSessionFactory` directly but sit in `TestAwsBaseHook`; `TestSessionFactory` is where the other direct-factory tests live.
- **One nuance on the lxml floor**, which cuts slightly against the non-reproduction argument: the `lxml>=6.0.0` pin lives only in the `python3-saml` extra and only for `python_version < '3.13'`, whereas this code path imports `requests_gssapi`, which is not declared as a provider dependency at all. So a deployment using `idp_auth_method: http_spegno_auth` can run against whatever lxml the environment supplies. That does not make it a vulnerability — no exfiltration path was shown and `no_network` is already the default — but it is another reason the pinned parser is worth having.

---

>

PR #71427, opened 2026-08-11
Deprecate Dataproc ClusterGenerator helper class

olegkachur-e:
- Becomes obsolete following the deprecation of keyword arguments in DataprocCreateClusterOperator.
- Direct configuration using 'cluster_config'/'virtual_cluster_config' is preferred in modern workflows.
- Reduces maintainer overhead by removing legacy helper code.

 

---

---


providers/google/tests/system/google/cloud/dataproc/example_dataproc_batch_persistent.py
```diff
@@ -53,17 +52,25 @@
...
+        "machine_type_uri": "n1-standard-4",
     },
-    enable_component_gateway=True,
-).make()
+    "worker_config": {
+        "num_instances": 0,
+        "machine_type_uri": "n1-standard-4",
+    },
+    "software_config": {
+        "properties": {
```
shahar1: **Blocking.** The old `ClusterGenerator(..., num_workers=0).make()` also added `"dataproc:dataproc.allow.zero.workers": "true"` to `software_config.properties` (it sets `single_node` when `num_workers == 0`, see `dataproc.py:604-605`). A create request with `worker_config.num_instances: 0` needs this property. Without it, Dataproc treats the cluster as a standard cluster, which needs at least 2 primary workers, and rejects `create_cluster_for_phs` ([single-node clusters](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/single-node-clusters)). Apache CI only imports this Dag, so the failure will first show up in the Google system-test runs.

```suggestion
    "software_config": {
        "properties": {
            "dataproc:dataproc.allow.zero.workers": "true",
```

providers/google/docs/operators/cloud/dataproc.rst
```diff
@@ -188,16 +188,12 @@ You can use deferrable mode for this action in order to run the operator asynchr
...
-You can generate and use config as followed:
 
-.. exampleinclude:: /../../google/tests/system/google/cloud/dataproc/example_dataproc_cluster_generator.py
-    :language: python
-    :dedent: 0
-    :start-after: [START how_to_cloud_dataproc_create_cluster_generate_cluster_config]
-    :end-before: [END how_to_cloud_dataproc_create_cluster_generate_cluster_config]
+.. warning::
+    **Deprecated:** The :class:`~airflow.providers.google.cloud.operators.dataproc.ClusterGenerator`
+    class is deprecated and will be removed after February 10, 2027. Please pass
```
shahar1: The docs say "after February 10, 2027", but every `planned_removal_date` in the code says `September 1, 2027`. The code date is the one that satisfies the 6-month minimum in `deprecation-policy.rst`, so please change the docs to match.

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -718,7 +742,7 @@ def __init__(
                 f"Passing cluster parameters by keywords to `{type(self).__name__}` will be deprecated. "
                 "Please provide cluster_config object using `cluster_config` parameter. "
                 "You can use `airflow.dataproc.ClusterGenerator.generate_cluster` "
```
shahar1: This line is edited for the date, but it still tells users to use `airflow.dataproc.ClusterGenerator.generate_cluster`. That sends them to the class this PR deprecates, and the path and method don't exist (the method is `make()`). Please point it at `cluster_config` / `virtual_cluster_config` instead.

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -75,6 +76,11 @@
     from airflow.providers.common.compat.sdk import Context
 
 
+@deprecated(
+    planned_removal_date="September 1, 2027",
+    reason="This helper class is being removed alongside the deprecated 'CreateCluster' class.",
```
shahar1: No class named `CreateCluster` exists. This should probably say `ClusterGenerator` (same at lines 95 and 119).

providers/google/src/airflow/providers/google/cloud/operators/dataproc.py
```diff
@@ -75,6 +76,11 @@
...
+    category=AirflowProviderDeprecationWarning,
+)
 class PreemptibilityType(Enum):
```
shahar1: Users will never see this deprecation. `deprecated` wraps `__new__`, so member access (`PreemptibilityType.SPOT`) doesn't warn. Only a value lookup (`PreemptibilityType("SPOT")`) warns, and the one place that does that is `ClusterGenerator._set_preemptibility_type` (line 387). As a result, every `ClusterGenerator(...)` call emits a second warning, about a class the user never touched, and that warning is attributed to `enum.py`. The operator's keyword-arguments path emits three warnings in total.

I'd drop `@deprecated` from the enum; it is removed together with `ClusterGenerator` anyway. Optionally, also silence the internal `ClusterGenerator(**kwargs)` call at line 759, since the operator already emits its own warning pointing at user code.

providers/google/tests/deprecations_ignore.yml
```diff
@@ -155,3 +155,4 @@
 - providers/google/tests/unit/google/cloud/hooks/test_datacatalog.py::TestCloudDataCatalogMissingProjectIdHook::test_update_tag
 - providers/google/tests/unit/google/cloud/hooks/test_datacatalog.py::TestCloudDataCatalogMissingProjectIdHook::test_update_tag_template
 - providers/google/tests/unit/google/cloud/hooks/test_datacatalog.py::TestCloudDataCatalogMissingProjectIdHook::test_update_tag_template_field
+- providers/google/tests/unit/google/cloud/operators/test_dataproc.py::TestsClusterGenerator
```
shahar1: This silences the new warnings, but no test checks that they are raised, so no test fails if the decorators are removed. Recent Google provider deprecations (#73575, #73354, #56950) each added a `pytest.warns(AirflowProviderDeprecationWarning, match=...)` test. One `match="ClusterGenerator"` test would cover this. A `pytest.mark.filterwarnings` on the class could replace this yml entry.

shahar1 (changes requested):
## AI Review

Deprecating `ClusterGenerator` makes sense. One blocking issue: the migrated system test config is missing a property that `ClusterGenerator` used to add, so cluster creation will fail (see the inline comment on `example_dataproc_batch_persistent.py`). The other inline comments are smaller consistency fixes: the removal dates don't match, the operator warning still points at `ClusterGenerator`, the reason text names a class that doesn't exist, the enum deprecation never reaches users, and there is no test for the new warnings.

### Smaller observations

- Nit: without `use_instead`, the rendered `ClusterGenerator` warning reads *"There is no replacement."* and then gives replacement instructions. The helper-class warnings also say there is no replacement. A short `use_instead` would read better.
- Nit: this PR moves the keyword-arguments path's removal date from October 5, 2026 to September 1, 2027. That's reasonable, since the old date is days away, but please mention it in the PR description so it's visible in the changelog.
- Nit: deleting `example_dataproc_cluster_generator.py` removes system-test coverage for a class that stays supported until September 2027. That's fine if intended, but worth confirming.
- CI: six checks on the head commit were cancelled, including `MyPy providers checks`. A rebase will give a fresh run.

---

>

Issue #47878, opened 2025-03-17
Revisit Secure token attribute


pierrejeambrun:
Use another way to check weather or not we need to use `secure` cookie. When TLS is enabled cookies should be secured, when it's not enabled it should not. (Otherwise auth will fail and other cookie based workflow too).

Currently we check the config for a certificate to decide that, but most of the time a proxy handles the TLS termination so we should maybe also check `request.secure` or something related to proxy headers forwarding.

More context here:
https://github.com/apache/airflow/pull/47859#issuecomment-2729487918

pierrejeambrun:
Update:
  The fixed pattern (proxy-aware) is:
  `secure = request.base_url.scheme == "https" or bool(conf.get("api", "ssl_cert", fallback=""))`
  request.base_url.scheme resolves to https behind a TLS-terminating proxy because gunicorn is
  started with forwarded_allow_ips="*" when proxy_headers is on (gunicorn_app.py:290). So where
  this is used, the issue is addressed.


  ✅ Done (request-scheme-aware):
  - airflow-core — simple/routes/login.py:96, auth/middlewares/refresh_token.py:72,
  core_api/routes/public/auth.py:71
  - FAB — fab/auth_manager/api_fastapi/routes/login.py:106 and the legacy Flask
  fab/www/views.py:122
  - Keycloak — only logout_callback (the cookie delete, line 158)
  - AWS login_callback → JWT cookie (aws/auth_manager/routes/login.py:121) https://github.com/apache/airflow/pull/69898

  ❌ Not done (still secure = bool(conf.get("api","ssl_cert")), cert-only):
  - Keycloak login → _oauth_state cookie (routes/login.py:68)
  - Keycloak login_callback → the actual session cookies, JWT + _id_token (routes/login.py:108)



> For uvicorn, ther's no `forwarded_allow_ips` which is probably a gap.

shaealh:
working on AWS login_callback secure-cookie slice

PR #71429, opened 2026-08-11
Honor FORWARDED_ALLOW_IPS when the API server runs under gunicorn

rjgoyln:
## Summary

Running behind a TLS-terminating proxy is documented as `--proxy-headers` plus `FORWARDED_ALLOW_IPS` to name which proxies may be trusted, and `server_type = uvicorn` works that way. `server_type = gunicorn` pinned the trust list to `"*"`, so an operator who narrowed it to their own proxy silently got a server accepting `X-Forwarded-*` from anyone able to reach it directly — including the `X-Forwarded-Proto` that decides whether session cookies carry the `Secure` attribute. Gunicorn reads the variable natively, so honouring it is a matter of not overriding it.

Where it is unset the wide default stays and only warns, since off-host proxies depend on it; tightening waits for Airflow 4. Turning `--proxy-headers` off now trusts nobody, which is what the flag already meant under uvicorn.

## Note on #47878

The audit there closes with "For uvicorn, ther's no `forwarded_allow_ips` which is probably a gap". Read the other way round, uvicorn already matches `run-behind-proxy.rst` and the hardcoded `"*"` is the anomaly. Every cookie site that audit lists now uses the proxy-aware `secure = ...` expression, leaving this as the last behavioural item on the issue.

related: #47878

---


airflow-core/newsfragments/71429.significant.rst
henry3260, line 1: Please remove this file

rjgoyln: OK. I got it.

airflow-core/docs/howto/run-behind-proxy.rst
```diff
@@ -62,6 +62,7 @@ To do so, you need to set the following setting in your ``airflow.cfg``::
   set the ``FORWARDED_ALLOW_IPS`` environment variable so Uvicorn knows who to trust this header from. See
   `Uvicorn's docs <https://www.uvicorn.org/deployment/#proxies-and-forwarded-headers>`_. For the full options you can pass here.
   (Please note the ``--forwarded-allow-ips`` CLI option does not exist in Airflow.)
+  This applies to both ``server_type = uvicorn`` and ``server_type = gunicorn``.
```
pierrejeambrun: The entire section above is mentioning uvicorn a lot. I would update it to mention `application server`. As well as a phrase explaining the two application servers that Airflow supports, and I would remove this line.

airflow-core/src/airflow/api_fastapi/gunicorn_app.py
```diff
@@ -286,7 +286,9 @@ def create_gunicorn_app(
         if ssl_cert_reqs is not None:
             options["cert_reqs"] = ssl_cert_reqs
 
-    if proxy_headers:
-        options["forwarded_allow_ips"] = "*"
+    if not proxy_headers:
+        # ``UvicornWorker`` leaves uvicorn's ``proxy_headers`` at its default of True, so
+        # trusting nobody is the only way to keep the worker off X-Forwarded-*.
+        options["forwarded_allow_ips"] = ""
```
pierrejeambrun, line 295: I believe this can disrupt some existing installs:
A gunicorn deployment with a TLS-terminating proxy on a different host that relied on the implicit "*" will now stop trusting X-Forwarded-Proto unless it sets FORWARDED_ALLOW_IPS

Maybe we need:
```
  if proxy_headers:
      if "FORWARDED_ALLOW_IPS" in os.environ:
          # Operator explicitly restricted trusted proxies — previously ignored.
          # Leave unset so the worker reads FORWARDED_ALLOW_IPS (like uvicorn).
          pass
      else:
          warnings.warn(
              "Under gunicorn the API server trusts X-Forwarded-* from all clients by "
              "default. Set FORWARDED_ALLOW_IPS to restrict trusted proxies; this default "
              "will become restrictive in Airflow 4.",
              category=RemovedInAirflow4Warning,
              stacklevel=2,
          )
          options["forwarded_allow_ips"] = "*"
```

pierrejeambrun (commented):
Thanks for the PR just one adjustment to make I believe.

rjgoyln:
> Thanks for the PR just one adjustment to make I believe.

Hi @pierrejeambrun,

Thanks for the great suggestions! Really appreciate it. 
I agree with the backward compatibility concern. I've updated the PR accordingly:

* **Gunicorn & Compat:** Implemented the fallback to `"*"` with a `RemovedInAirflow4Warning` to ensure existing deployments won't break.
* **Docs:** Replaced mentions of `"Uvicorn"` with the generic `"application server"` and added the clarification about supported servers.
* **Tests:** Updated the parameterized tests to cover the new deprecation warning logic.

I also removed the "Behavior change" section from the PR description.

pierrejeambrun (approved):

PR #71437, opened 2026-08-11
Add embedding model support to PydanticAIHook

jeff3071:
## Why

`PydanticAIHook` supports Airflow-managed credentials and custom base URLs for LLMs, including OpenAI-compatible self-hosted endpoints, but it does not expose pydantic-ai's embedding API. Users cannot create a pydantic-ai `Embedder`.

## What

- Add `embed_model_id` and `extra["embed_model"]` resolution to `PydanticAIHook`.
- Add a cached `get_embedder()` that returns a pydantic-ai `Embedder` configured with the connection's credentials and base URL.
- Share provider-factory creation between LLM and embedding model resolution so provider-specific hooks keep their existing authentication mapping.

## Testing

- Manual Airflow UI verification against a vLLM-hosted `Qwen/Qwen3-Embedding-0.6B` OpenAI-compatible endpoint:
  - one query vector and two document vectors returned;
  - all vectors dimension is 1024;

`serve command`
```bash
vllm serve Qwen/Qwen3-Embedding-0.6B     --runner pooling     --host 0.0.0.0     --port 8000     --api-key EMPTY     --max-model-len 1024 --gpu-memory-utilization 0.8
```

test

test dag
```python

import math

from airflow.providers.common.ai.hooks.pydantic_ai import PydanticAIHook

CONN_ID = "pydanticai_qwen_embedding"
EXPECTED_DIMENSION = 1024

@dag(
    dag_id="qwen3_self_hosted_embedding",
    schedule=None,
    catchup=False,
    tags=["manual", "pydantic-ai", "embedding", "self-hosted"],
)
def qwen3_self_hosted_embedding():
    @task
    def verify_embedding() -> dict[str, int | str | bool]:
        hook = PydanticAIHook(llm_conn_id=CONN_ID)
        embedder = hook.get_embedder()

        query_result = embedder.embed_query_sync("How does Apache Airflow orchestrate workflows?")
        document_result = embedder.embed_documents_sync(
            [
                "Apache Airflow is a platform for developing and monitoring batch workflows.",
                "Qwen3 Embedding converts text into dense vector representations.",
            ]
        )

        if len(query_result.embeddings) != 1 or len(document_result.embeddings) != 2:
            raise ValueError(
                "Unexpected vector count: "
                f"query={len(query_result.embeddings)}, documents={len(document_result.embeddings)}"
            )

        vectors = [*query_result.embeddings, *document_result.embeddings]
        dimensions = {len(vector) for vector in vectors}
        if dimensions != {EXPECTED_DIMENSION}:
            raise ValueError(f"Expected {EXPECTED_DIMENSION}-dimension vectors, got {sorted(dimensions)}")
        if not all(math.isfinite(value) for vector in vectors for value in vector):
            raise ValueError("Embedding response contains non-finite values.")

        return {
            "model": query_result.model_name,
            "provider": query_result.provider_name,
            "dimension": next(iter(dimensions)),
            "query_vectors": len(query_result.embeddings),
            "document_vectors": len(document_result.embeddings),
            "embedder_cached": embedder is hook.get_embedder(),
        }

    verify_embedding()


qwen3_self_hosted_embedding()
```

`log`
```
/root/airflow/logs/dag_id=qwen3_self_hosted_embedding/run_id=manual__2026-08-11T12:49:12.574
572+00:00/task_id=verify_embedding/attempt=1.log
::endgroup::
[2026-08-11T12:49:13.311809Z] INFO - ::group::Pre Execute
Task Identity ti_id=019ff0de-ab8f-7043-84dc-098e5407c3c5 dag_id=qwen3_self_hosted_embedding task_id=verify_embedding run_id=manual__2026-08-11T12:49:12.574572+00:00 try_number=1 map_index=-1
[2026-08-11T12:49:13.363284Z] INFO - DAG bundles loaded: dags-folder
[2026-08-11T12:49:13.365408Z] INFO - Filling up the DagBag from /files/dags/manual_qwen3_self_hosted_embedding.py
[2026-08-11T12:49:15.292911Z] INFO - Worker startup parse complete bundle_name=dags-folder  bundle_version=null  dag_file=manual_qwen3_self_hosted_embedding.py  bundle_prepare_ms=3  dag_file_parse_ms=1928 
[2026-08-11T12:49:15.326585Z] INFO - ::endgroup::
[2026-08-11T12:49:15.352060Z] INFO - Using explicit credentials for provider with embedding model 'openai:Qwen/Qwen3-Embedding-0.6B': ['api_key', 'base_url']
[2026-08-11T12:49:17.976642Z] INFO - Done. Returned value was: {'model': 'Qwen/Qwen3-Embedding-0.6B', 'provider': 'openai', 'dimension': 1024, 'query_vectors': 1, 'document_vectors': 2, 'embedder_cached': True}
[2026-08-11T12:49:17.976832Z] INFO - ::group::Post Execute
[2026-08-11T12:49:17.976973Z] INFO - Pushing xcom ti=RuntimeTaskInstance(id=UUID('019ff0de-ab8f-7043-84dc-098e5407c3c5'), task_id='verify_embedding', dag_id='qwen3_self_hosted_embedding', run_id='manual__2026-08-11T12:49:12.574572+00:00', try_number=1, dag_version_id=UUID('019ff0dd-ef97-7b8e-9edb-36664c8c2b3a'), map_index=-1, hostname='86afabd35b10', context_carrier={'traceparent': '00-a5a16744aec0bf462e7b31c73faa796d-b98449d8ef65dcfd-00'}, queue='default', task=<Task(_PythonDecoratedOperator): verify_embedding>, bundle_instance=LocalDagBundle(name=dags-folder), max_tries=0, start_date=datetime.datetime(2026, 8, 11, 12, 49, 13, 344044, tzinfo=datetime.timezone.utc), end_date=None, state=<TaskInstanceState.RUNNING: 'running'>, is_mapped=False, rendered_map_index=None, sentry_integration='') 
[2026-08-11T12:49:18.094619Z] INFO - ::endgroup::
```

---

---

- [X] Yes — Codex (GPT-5.6-sol)
Generated-by: Codex (GPT-5.6-sol) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/providers/common/ai/docs/connections/pydantic_ai.rst
+++ b/providers/common/ai/docs/connections/pydantic_ai.rst
@@ -51,14 +51,24 @@ Model
     The model can also be overridden at the hook/operator level via the
     ``model_id`` parameter.
 
+Embedding Model
+    The embedding model identifier in ``provider:model`` format. This field
+    appears as a dedicated input in the connection form and stores its value in
+    ``extra["embed_model"]``.
+
+    Example: ``openai:text-embedding-3-small``
+
+    The embedding model and connection can also be overridden at the hook level
+    via the ``embed_model_id`` and ``embed_conn_id`` parameters.
+
 API Key (Password field)
-    The API key for your LLM provider. Required for API-key-based providers
+    The API key for your model provider. Required for API-key-based providers
     (OpenAI, Anthropic, Groq, Mistral). Leave empty for providers using
     environment-based auth (Bedrock via ``AWS_PROFILE``, Vertex via
     ``GOOGLE_APPLICATION_CREDENTIALS``).
 
 Host (optional)
-    Base URL for the provider's API. Only needed for custom endpoints:
+    Base URL for the model provider's API. Only needed for custom endpoints:
 
     - Ollama: ``http://localhost:11434/v1``
     - vLLM: ``http://localhost:8000/v1``
@@ -76,8 +86,8 @@ Extra (JSON, optional)
             "embed_model": "openai:text-embedding-3-small"
         }
 
-    When using the UI, the "Model" field above writes to this same location
-    automatically.
+    When using the UI, the "Model" and "Embedding Model" fields above write to
+    this same location automatically.
 
 Examples
 --------
@@ -146,3 +156,6 @@ The embedding model is resolved separately:
 
 1. ``embed_model_id`` parameter on the hook
 2. ``embed_model`` in the connection's extra JSON
+
+Embedding credentials and endpoints are read from ``embed_conn_id`` when set;
+otherwise they are read from ``llm_conn_id``.
--- a/providers/common/ai/docs/connections/pydantic_ai.rst
+++ b/providers/common/ai/docs/connections/pydantic_ai.rst
@@ -21,9 +21,11 @@ Pydantic AI Connection
 ======================
 
 The `Pydantic AI <https://ai.pydantic.dev/>`__ connection type configures access
-to LLM providers via the pydantic-ai framework. A single connection type works with
-any provider that pydantic-ai supports: OpenAI, Anthropic, Google, Bedrock, Groq,
-Mistral, Ollama, vLLM, and others.
+to LLM and embedding models via the pydantic-ai framework. A single connection
+type works with any provider that pydantic-ai supports. Supported LLM providers
+include OpenAI, Anthropic, Google, Bedrock, Groq, Mistral, Ollama, vLLM, and
+others. Embedding model availability depends on the providers supported by
+pydantic-ai's ``Embedder``.
 
 Default Connection IDs
 ----------------------
@@ -65,11 +67,14 @@ Host (optional)
 
 Extra (JSON, optional)
     A JSON object with additional configuration. Programmatic users can set the
-    model directly in extra:
+    LLM and embedding models directly in extra:
 
     .. code-block:: json
 
-        {"model": "openai:gpt-5.6-sol"}
+        {
+            "model": "openai:gpt-5.6-sol",
+            "embed_model": "openai:text-embedding-3-small"
+        }
 
     When using the UI, the "Model" field above writes to this same location
     automatically.
@@ -136,3 +141,8 @@ The hook reads the model from these sources in priority order:
 
 1. ``model_id`` parameter on the hook/operator
 2. ``model`` in the connection's extra JSON (set by the "Model" conn-field in the UI)
+
+The embedding model is resolved separately:
+
+1. ``embed_model_id`` parameter on the hook
+2. ``embed_model`` in the connection's extra JSON
--- a/providers/common/ai/docs/hooks/pydantic_ai.rst
+++ b/providers/common/ai/docs/hooks/pydantic_ai.rst
@@ -61,12 +61,15 @@ Embedding Models
 ----------------
 
 Set ``embed_model_id`` on the hook or ``embed_model`` in the connection's extra JSON,
-then call ``get_embedder()``. The resolved ``Embedder`` is cached on the hook instance.
+then call ``get_embedder()``. Use ``embed_conn_id`` when the embedding provider uses
+different credentials or an endpoint from the LLM provider; it defaults to
+``llm_conn_id``. The resolved ``Embedder`` is cached on the hook instance.
 
 .. code-block:: python
 
     hook = PydanticAIHook(
         llm_conn_id="my_llm",
+        embed_conn_id="my_embeddings",
         embed_model_id="openai:text-embedding-3-small",
     )
     embedder = hook.get_embedder()
--- a/providers/common/ai/docs/hooks/pydantic_ai.rst
+++ b/providers/common/ai/docs/hooks/pydantic_ai.rst
@@ -21,10 +21,11 @@ PydanticAIHook
 ==============
 
 Use :class:`~airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIHook` to interact
-with LLM providers via `pydantic-ai <https://ai.pydantic.dev/>`__.
+with LLM and embedding providers via `pydantic-ai <https://ai.pydantic.dev/>`__.
 
 The hook manages API credentials from an Airflow connection and creates pydantic-ai
-``Model`` and ``Agent`` objects. It supports any provider that pydantic-ai supports.
+``Model``, ``Agent``, and ``Embedder`` objects. Supported providers depend on the
+corresponding pydantic-ai model API.
 
 .. seealso::
     :ref:`Connection configuration <howto/connection:pydanticai>`
@@ -55,6 +56,22 @@ The model can be specified at three levels (highest priority first):
 
     # Override with a specific model
     hook = PydanticAIHook(llm_conn_id="my_llm", model_id="anthropic:claude-opus-4-6")
+
+Embedding Models
+----------------
+
+Set ``embed_model_id`` on the hook or ``embed_model`` in the connection's extra JSON,
+then call ``get_embedder()``. The resolved ``Embedder`` is cached on the hook instance.
+
+.. code-block:: python
+
+    hook = PydanticAIHook(
+        llm_conn_id="my_llm",
+        embed_model_id="openai:text-embedding-3-small",
+    )
+    embedder = hook.get_embedder()
+    result = embedder.embed_query_sync("Apache Airflow orchestrates workflows.")
+    embedding = result.embeddings[0]
 
 Structured Output
 -----------------
--- a/providers/common/ai/provider.yaml
+++ b/providers/common/ai/provider.yaml
@@ -174,6 +174,13 @@ connection-types:
           type:
             - string
             - 'null'
+      embed_model:
+        label: Embedding Model
+        description: "Embedding model in provider:name format (e.g. openai:text-embedding-3-small)"
+        schema:
+          type:
+            - string
+            - 'null'
   - hook-class-name: airflow.providers.common.ai.hooks.pydantic_ai.PydanticAIAzureHook
     hook-name: "Pydantic AI (Azure OpenAI)"
     connection-type: pydanticai-azure
@@ -193,6 +200,13 @@ connection-types:
       model:
         label: Model
         description: "Azure model identifier (e.g. azure:gpt-4o)"
+        schema:
+          type:
+            - string
+            - 'null'
+      embed_model:
+        label: Embedding Model
+        description: "Azure embedding model identifier (e.g. azure:text-embedding-3-small)"
         schema:
           type:
             - string
@@ -222,6 +236,13 @@ connection-types:
       model:
         label: Model
         description: "Bedrock model identifier (e.g. bedrock:us.anthropic.claude-opus-4-5)"
+        schema:
+          type:
+            - string
+            - 'null'
+      embed_model:
+        label: Embedding Model
+        description: "Bedrock embedding model identifier (e.g. bedrock:amazon.titan-embed-text-v2:0)"
         schema:
           type:
             - string
@@ -307,6 +328,13 @@ connection-types:
       model:
         label: Model
         description: "Google model identifier (e.g. google-vertex:gemini-2.0-flash)"
+        schema:
+          type:
+            - string
+            - 'null'
+      embed_model:
+        label: Embedding Model
+        description: "Google embedding model identifier (e.g. google-cloud:text-embedding-005)"
         schema:
           type:
             - string
--- a/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
@@ -16,10 +16,12 @@
 # under the License.
 from __future__ import annotations
 
+from collections.abc import Callable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypeVar, overload
 
-from pydantic_ai import Agent
+from pydantic_ai import Agent, Embedder
+from pydantic_ai.embeddings import infer_embedding_model
 from pydantic_ai.models import infer_model
 from pydantic_ai.providers import infer_provider, infer_provider_class
 
@@ -53,11 +55,13 @@ class PydanticAIHook(BaseHook):
     Connection fields:
         - **password**: API key
         - **host**: Base URL (optional, e.g. ``https://api.openai.com/v1``)
-        - **extra** JSON: ``{"model": "openai:gpt-5.6-sol"}``
+        - **extra** JSON: ``{"model": "openai:gpt-5.6-sol", "embed_model": "openai:text-embedding-3-small"}``
 
     :param llm_conn_id: Airflow connection ID for the LLM provider.
     :param model_id: Model identifier in ``provider:model`` format (e.g. ``"openai:gpt-5.6-sol"``).
         Overrides the model stored in the connection's extra field.
+    :param embed_model_id: Embedding model identifier in ``provider:model`` format.
+        Overrides the embedding model stored in the connection's extra field.
     """
 
     conn_name_attr = "llm_conn_id"
@@ -69,6 +73,7 @@ class PydanticAIHook(BaseHook):
         self,
         llm_conn_id: str | None = None,
         model_id: str | None = None,
+        embed_model_id: str | None = None,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
@@ -78,7 +83,9 @@ class PydanticAIHook(BaseHook):
         # argument values at class-definition time.
         self.llm_conn_id = llm_conn_id if llm_conn_id is not None else self.default_conn_name
         self.model_id = model_id
+        self.embed_model_id = embed_model_id
         self._model: Model | None = None
+        self._embedder: Embedder | None = None
         self._conn: Connection | None = None
         self._conn_extra_dejson: dict[str, Any] | None = None
 
@@ -90,7 +97,7 @@ class PydanticAIHook(BaseHook):
             "relabeling": {"password": "API Key"},
             "placeholders": {
                 "host": "https://api.openai.com/v1  (optional, for custom endpoints / Ollama)",
-                "extra": '{"model": "openai:gpt-5.6-sol"}',
+                "extra": '{"model": "openai:gpt-5.6-sol", "embed_model": "openai:text-embedding-3-small"}',
             },
         }
 
@@ -127,6 +134,20 @@ class PydanticAIHook(BaseHook):
             kwargs["base_url"] = base_url
         return kwargs
 
+    def _create_provider_factory(self, provider_kwargs: dict[str, Any]) -> Callable[[str], Any]:
+        def _provider_factory(provider_name: str) -> Any:
+            try:
+                return infer_provider_class(provider_name)(**provider_kwargs)
+            except TypeError:
+                self.log.warning(
+                    "Provider '%s' rejected kwargs %s; falling back to env-var auth",
+                    provider_name,
+                    list(provider_kwargs),
+                )
+                return infer_provider(provider_name)
+
+        return _provider_factory
+
     def get_conn(self) -> Model:
         """
         Return a configured pydantic-ai ``Model``.
@@ -160,29 +181,56 @@ class PydanticAIHook(BaseHook):
 
         provider_kwargs = self._get_provider_kwargs(api_key, base_url, extra)
         if provider_kwargs:
-            _kwargs = provider_kwargs  # capture for closure
             self.log.info(
                 "Using explicit credentials for provider with model '%s': %s",
                 model_name,
                 list(provider_kwargs),
             )
-
-            def _provider_factory(pname: str) -> Any:
-                try:
-                    return infer_provider_class(pname)(**_kwargs)
-                except TypeError:
-                    self.log.warning(
-                        "Provider '%s' rejected kwargs %s; falling back to env-var auth",
-                        pname,
-                        list(_kwargs),
-                    )
-                    return infer_provider(pname)
-
-            self._model = infer_model(model_name, provider_factory=_provider_factory)
+            self._model = infer_model(
+                model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
             return self._model
 
         self._model = infer_model(model_name)
         return self._model
+
+    def get_embedder(self) -> Embedder:
+        """Return a pydantic-ai ``Embedder`` using this connection's credentials."""
+        if self._embedder is not None:
+            return self._embedder
+
+        conn = self.get_connection(self.llm_conn_id) if self._conn is None else self._conn
+        extra: dict[str, Any] = (
+            conn.extra_dejson if self._conn_extra_dejson is None else self._conn_extra_dejson
+        )
+
+        embed_model_name: str = self.embed_model_id or extra.get("embed_model", "")
+        if not embed_model_name:
+            raise ValueError(
+                "No embedding model specified. Set embed_model_id on the hook or the embed_model field "
+                "on the connection."
+            )
+
+        api_key: str | None = conn.password or None
+        base_url: str | None = conn.host or None
+
+        provider_kwargs = self._get_provider_kwargs(api_key, base_url, extra)
+        if provider_kwargs:
+            self.log.info(
+                "Using explicit credentials for provider with embedding model '%s': %s",
+                embed_model_name,
+                list(provider_kwargs),
+            )
+            embedding_model = infer_embedding_model(
+                embed_model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
+        else:
+            embedding_model = infer_embedding_model(embed_model_name)
+
+        self._embedder = Embedder(embedding_model)
+        return self._embedder
 
     def _get_conn_if_model_configured(self) -> Model | None:
         """Return the hook model only when the hook or connection explicitly configures one."""
--- a/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
@@ -231,7 +231,7 @@ class PydanticAIHook(BaseHook):
         else:
             embedding_model = infer_embedding_model(embed_model_name)
 
-        self._embedder = Embedder(embedding_model)
+        self._embedder = Embedder(embedding_model, instrument=genai_instrumentation_settings())
         return self._embedder
 
     def _get_conn_if_model_configured(self) -> Model | None:
--- a/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
@@ -62,6 +62,8 @@ class PydanticAIHook(BaseHook):
         Overrides the model stored in the connection's extra field.
     :param embed_model_id: Embedding model identifier in ``provider:model`` format.
         Overrides the embedding model stored in the connection's extra field.
+    :param embed_conn_id: Optional separate Airflow connection ID for the embedding provider.
+        Falls back to ``llm_conn_id`` when not provided.
     """
 
     conn_name_attr = "llm_conn_id"
@@ -73,6 +75,7 @@ class PydanticAIHook(BaseHook):
         self,
         llm_conn_id: str | None = None,
         model_id: str | None = None,
+        embed_conn_id: str | None = None,
         embed_model_id: str | None = None,
         **kwargs,
     ) -> None:
@@ -83,6 +86,7 @@ class PydanticAIHook(BaseHook):
         # argument values at class-definition time.
         self.llm_conn_id = llm_conn_id if llm_conn_id is not None else self.default_conn_name
         self.model_id = model_id
+        self.embed_conn_id = embed_conn_id if embed_conn_id is not None else self.llm_conn_id
         self.embed_model_id = embed_model_id
         self._model: Model | None = None
         self._embedder: Embedder | None = None
@@ -200,10 +204,8 @@ class PydanticAIHook(BaseHook):
         if self._embedder is not None:
             return self._embedder
 
-        conn = self.get_connection(self.llm_conn_id) if self._conn is None else self._conn
-        extra: dict[str, Any] = (
-            conn.extra_dejson if self._conn_extra_dejson is None else self._conn_extra_dejson
-        )
+        conn = self.get_connection(self.embed_conn_id)
+        extra: dict[str, Any] = conn.extra_dejson
 
         embed_model_name: str = self.embed_model_id or extra.get("embed_model", "")
         if not embed_model_name:
@@ -336,16 +338,18 @@ class PydanticAIHook(BaseHook):
 
     def test_connection(self) -> tuple[bool, str]:
         """
-        Test connection by resolving the model.
+        Test connection by resolving the configured model.
 
-        Validates that the model string is valid and the provider class can be
-        instantiated with the supplied credentials.  Does NOT make an LLM API
-        call — that would be expensive and fail for reasons unrelated to
+        Validates that the LLM or embedding model string is valid and the provider
+        class can be instantiated with the supplied credentials. Does NOT make an
+        API call — that would be expensive and fail for reasons unrelated to
         connectivity (quotas, billing, rate limits).
         """
         try:
-            self.get_conn()
-            return True, "Model resolved successfully."
+            if self._get_conn_if_model_configured() is not None:
+                return True, "Model resolved successfully."
+            self.get_embedder()
+            return True, "Embedding model resolved successfully."
         except Exception as e:
             return False, str(e)
 
--- a/providers/common/ai/tests/unit/common/ai/hooks/test_pydantic_ai.py
+++ b/providers/common/ai/tests/unit/common/ai/hooks/test_pydantic_ai.py
@@ -22,6 +22,8 @@ from pathlib import Path
 from unittest.mock import MagicMock, patch
 
 import pytest
+from pydantic_ai import Embedder
+from pydantic_ai.embeddings import EmbeddingModel
 from pydantic_ai.models import Model
 from pydantic_ai.models.test import TestModel
 
@@ -39,11 +41,17 @@ class TestPydanticAIHookInit:
         hook = PydanticAIHook()
         assert hook.llm_conn_id == "pydanticai_default"
         assert hook.model_id is None
+        assert hook.embed_model_id is None
 
     def test_custom_conn_id(self):
-        hook = PydanticAIHook(llm_conn_id="my_llm", model_id="openai:gpt-5.6-sol")
+        hook = PydanticAIHook(
+            llm_conn_id="my_llm",
+            model_id="openai:gpt-5.6-sol",
+            embed_model_id="openai:text-embedding-3-small",
+        )
         assert hook.llm_conn_id == "my_llm"
         assert hook.model_id == "openai:gpt-5.6-sol"
+        assert hook.embed_model_id == "openai:text-embedding-3-small"
 
     def test_azure_hook_uses_own_default_conn_name(self):
         """Subclass default_conn_name is used, not the base class value."""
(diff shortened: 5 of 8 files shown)


providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -69,6 +73,7 @@ def __init__(
         self,
         llm_conn_id: str | None = None,
         model_id: str | None = None,
+        embed_model_id: str | None = None,
```
kaxil, line 79: `LangChainHook` and `LlamaIndexHook` in this same provider both take an `embed_conn_id` that falls back to `llm_conn_id`, because the embedding provider often isn't the chat provider. Here `get_embedder()` always builds the provider from `llm_conn_id`, so a connection with `model: "anthropic:claude-opus-4-6"` plus `embed_model: "openai:text-embedding-3-small"` hands the Anthropic key to `OpenAIProvider`, and a `host` set for a self-hosted chat endpoint gets used as the OpenAI base URL. The `TypeError` guard in `_create_provider_factory` won't catch it, `OpenAIProvider(api_key=..., base_url=...)` constructs fine and just fails later as a 401 or silently talks to the wrong endpoint. Worth adding `embed_conn_id` here to match the other two hooks?

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -160,30 +181,57 @@ def get_conn(self) -> Model:
...
+            self._model = infer_model(
+                model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
             return self._model
 
         self._model = infer_model(model_name)
         return self._model
 
+    def get_embedder(self) -> Embedder:
```
kaxil, line 202: `test_connection()` still only calls `get_conn()`, which raises `No model specified` when the connection sets `embed_model` but no `model`. Now that an embedding-only connection is a valid configuration, hitting Test in the UI on one reports it as broken. Should `test_connection` fall back to resolving the embedder when no LLM model is configured?

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -160,30 +181,57 @@ def get_conn(self) -> Model:
...
+                list(provider_kwargs),
+            )
+            embedding_model = infer_embedding_model(
+                embed_model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
+        else:
+            embedding_model = infer_embedding_model(embed_model_name)
+
+        self._embedder = Embedder(embedding_model)
```
kaxil: `create_agent` applies `genai_instrumentation_settings()` so agent runs emit GenAI spans, but the `Embedder` is built without it, so with `[common.ai] otel_export_enabled` on you get spans for agents and nothing for embedding calls. `Embedder.__init__` takes `instrument: InstrumentationSettings | bool | None`, so `Embedder(embedding_model, instrument=genai_instrumentation_settings())` covers it, and passing `None` when tracing is off keeps the current zero-overhead behaviour.

providers/common/ai/docs/connections/pydantic_ai.rst
```diff
@@ -136,3 +141,8 @@ The hook reads the model from these sources in priority order:
 
 1. ``model_id`` parameter on the hook/operator
 2. ``model`` in the connection's extra JSON (set by the "Model" conn-field in the UI)
+
+The embedding model is resolved separately:
+
+1. ``embed_model_id`` parameter on the hook
+2. ``embed_model`` in the connection's extra JSON
```
kaxil, line 158: The `langchain` and `llamaindex` connection types already declare an `embed_model` conn-field in `provider.yaml` (labelled "Embedding Model"), but `pydanticai` only declares `model`. So on a new pydantic-ai connection the UI gives you a dedicated Model input and nothing equivalent for the embedding model, you have to type it into the Extra JSON by hand. Adding the conn-field to the four `pydanticai*` connection types would line this up.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
jeff3071: Thanks for quickly review!
I added `embed_conn_id` to  PydanticAIHook. Match the other hooks.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
jeff3071: I updated `test_connection()` to validate the LLM model when configured and fall back to the embedding model when no LLM model is configured.

If both are configured, `test_connection()` currently check only LLM.
Should `test_connection()` check both model instead?

providers/common/ai/docs/connections/pydantic_ai.rst
jeff3071: I added `embed_model` conn-field to all four `pydanticai` connection types.

I manual check UI have show `Embedding Model` field.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
jeff3071: I passed instrument to `Embedder`.
Thank for pointing this out!

kaxil:
Two things the rebase will surface, since the branch predates some churn in this area.

`get_ui_field_behaviour()` now advertises `embed_model` in the `extra` placeholder, but #72087 made `provider.yaml`'s `ui-field-behaviour.placeholders` mirror the hook. So `placeholders.extra` for the `pydanticai` type, and its `get_provider_info.py` copy, needs the same update or the two disagree again.

The per-vendor connection pages added in #71774 (`pydantic_ai_azure.rst`, `pydantic_ai_bedrock.rst`, `pydantic_ai_vertex.rst`) enumerate every conn-field, down to `Force Vertex AI Mode` and `Custom Endpoint URL`. This PR adds `embed_model` to those three connection types but documents it only in `pydantic_ai.rst`, so each of the three wants an Embedding Model entry.

Separately, I checked the examples and the dependency floor and both are fine. All four `embed_model` descriptions resolve against pydantic-ai 2.0 (`openai:text-embedding-3-small`, `azure:text-embedding-3-small` gives an `OpenAIEmbeddingModel`, `bedrock:amazon.titan-embed-text-v2:0`, `google-cloud:text-embedding-005`), and `Embedder` plus `infer_embedding_model` are both present at the declared `pydantic-ai-slim>=2.0.0` floor. #72012 moved the Vertex examples to the `google-cloud:` prefix, so your embed example already matches where main went.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
kaxil: Yes, check both. Nothing here makes an API call, so validating the embedding model when it's configured costs nothing, and as written a connection with a good `model` and a typo in `embed_model` reports green in the UI. The user finds out on the first task run instead.

The new ordering also changed the failure message for a connection with neither field set: it now reads `No embedding model specified. Set embed_model_id on the hook or the embed_model field on the connection.`, which the updated assertion in `test_failed_connection_no_model` pins. The likelier mistake there is a forgotten Model field, so the message points at the wrong one. Naming both fields when neither is set would read better.

For context, `LangChainHook` and `LlamaIndexHook` both resolve only the LLM in `test_connection` (#71841), so they carry the same blind spot for embed-only connections. Not something to fix here.

jeff3071: Thanks for the feedback. 
`test_connection()` now validates both the model and embedding model. I also added two unit tests covering these cases.
I’d be happy to address the same issue in `LangChainHook` and `LlamaIndexHook` in a separate PR.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -160,30 +185,55 @@ def get_conn(self) -> Model:
...
+        provider_kwargs = self._get_provider_kwargs(api_key, base_url, extra)
+        if provider_kwargs:
+            self.log.info(
+                "Using explicit credentials for provider with embedding model '%s': %s",
+                embed_model_name,
+                list(provider_kwargs),
+            )
+            embedding_model = infer_embedding_model(
+                embed_model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
```
kaxil: `_get_provider_kwargs` is keyed to the LLM hook, so the embedding connection gets the wrong mapping whenever the embed model's provider differs from the chat one, which is the case `embed_conn_id` was added for: `PydanticAIAzureHook(embed_conn_id=<a pydanticai conn>)` with `embed_model: openai:text-embedding-3-small` yields `{api_key, azure_endpoint}`, `OpenAIProvider` rejects `azure_endpoint`, the `except TypeError` above drops both, and with `OPENAI_API_KEY` on the worker the embedder lands on `https://api.openai.com/v1/` under the env key instead of the connection's key and host (Bedrock and Vertex get there with no log line at all, since their mapping reads only `extra` and returns `{}`). The base hook has a quieter version that succeeds instead of erroring: one plain `pydanticai` connection with a `host` plus `embed_model: cohere:embed-v4.0` builds a working `CohereEmbeddingModel` on `api.cohere.com` under `CO_API_KEY`, so the task embeds against the wrong account with a single warning as the only trace. Could the mapping be selected from the embedding model's provider, with the factory re-raising rather than retrying with the credentials discarded, and the comment in `PydanticAIVertexHook._get_provider_kwargs` (which already names this hazard) updated to say the swallow now serves `get_embedder()` too?

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -198,6 +248,17 @@ def _get_conn_if_model_configured(self) -> Model | None:
 
         return None
 
+    def _get_embedder_if_model_configured(self) -> Embedder | None:
+        """Return the embedder only when the hook or connection explicitly configures one."""
+        if self.embed_model_id:
+            return self.get_embedder()
+
+        conn = self.get_connection(self.embed_conn_id)
```
kaxil: `_get_conn_if_model_configured` caches its fetch into `self._conn` so `get_conn()` does not repeat it, but this one throws the connection away and `get_embedder()` then fetches the same `embed_conn_id` again, so `test_connection()` on a connection carrying both keys makes three `get_connection()` calls where it used to make one. `secrets.use_cache` is off by default, so from a task each of those is a real supervisor round trip plus a secrets-backend read. Keying the cache by conn id and routing both paths through it would remove the duplicate and stop a later caller picking up the LLM connection for the embedder by accident, since `self._conn` currently carries no conn id.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -160,30 +185,55 @@ def get_conn(self) -> Model:
...
+                list(provider_kwargs),
+            )
+            embedding_model = infer_embedding_model(
+                embed_model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
+        else:
+            embedding_model = infer_embedding_model(embed_model_name)
+
+        self._embedder = Embedder(embedding_model, instrument=genai_instrumentation_settings())
```
kaxil: Nothing asserts this yet: the new tests only check `isinstance(result, Embedder)`, which still passes with `instrument=` deleted, so 057ee9b is uncovered. `TestPydanticAIHookCreateAgentInstrumentation` has three tests for the same wiring on the agent side, and the embedder equivalent is a one-liner patching `genai_instrumentation_settings` to a sentinel and asserting `hook.get_embedder().instrument is sentinel`.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -89,8 +100,8 @@ def get_ui_field_behaviour() -> dict[str, Any]:
             "hidden_fields": ["schema", "port", "login"],
             "relabeling": {"password": "API Key"},
             "placeholders": {
-                "host": "https://api.openai.com/v1 (optional, for custom endpoints / Ollama)",
-                "extra": '{"model": "openai:gpt-5.6-sol"}',
+                "host": "https://api.openai.com/v1  (optional, for custom endpoints / Ollama)",
```
kaxil: This picked up a second space before `(optional`, so it no longer matches `provider.yaml:172` or its `get_provider_info.py:143` twin, both of which this PR edits one line lower for the `extra` placeholder. #72087 lined those three copies up on purpose, so this looks like a stray keystroke worth reverting.

providers/common/ai/provider.yaml
```diff
@@ -322,6 +343,13 @@ connection-types:
           type:
             - string
             - 'null'
+      embed_model:
```
kaxil, line 347: `test_conn_fields_model_description_prefix_is_valid_provider` reads only `conn-fields["model"]["description"]`, so the four new `embed_model` prefixes sit outside the tripwires that guard the same strings for `model`. Reusing `_assert_prefix_is_known_provider` would not be enough on its own either, since the provider registry is wider than what `infer_embedding_model` dispatches on: `infer_provider_class("anthropic")` resolves while `infer_embedding_model("anthropic:x")` raises `UserError: Unknown embeddings model`. A parametrized test over each connection type's `embed_model` description, validating through `infer_embedding_model`, would cover all four.

jeff3071:
Provider kwargs are now selected from each model's provider prefix instead of the hook type. This ensures an embedding model receives the correct credentials and options even when its provider differs from the LLM provider.
Connection caching is now keyed by connection ID. 

**Additional fix**

Models using the `gateway/<provider>:<model>` format are now created through `gateway_provider()` instead of a regular provider class, preserving Pydantic AI Gateway authentication and routing. See the [Pydantic AI Gateway documentation](https://pydantic.dev/docs/ai/overview/gateway/).

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -535,3 +597,14 @@ def _get_provider_kwargs(
...
         return kwargs
+
+
+_PROVIDER_KWARGS_MAPPER_BY_MODEL_PREFIX: dict[
+    str | None, Callable[[str | None, str | None, dict[str, Any]], dict[str, Any]]
+] = {
+    "azure": PydanticAIAzureHook._get_provider_kwargs,
+    "azure-responses": PydanticAIAzureHook._get_provider_kwargs,
+    "bedrock": PydanticAIBedrockHook._get_provider_kwargs,
+    "google": PydanticAIVertexHook._get_provider_kwargs,
```
kaxil: `google` and `google-cloud` share the Vertex mapping, but `infer_provider_class` resolves them to different classes: `GoogleCloudProvider` accepts `project`/`location`/`credentials`, while `GoogleProvider` accepts only `api_key`/`base_url`/`client`/`http_client`/`retry_options`. So a `pydanticai-vertex` connection carrying the `project` and `location` its own conn-fields document, plus a `google:` model, raises `TypeError: GoogleProvider.__init__() got an unexpected keyword argument 'project'` on pydantic-ai 2.31.1 -- and with the `except TypeError` gone there is nothing left to catch it. This is not embeddings-only: `get_conn()` with `model: "google:gemini-2.0-flash"` returned a `GoogleModel` at d195f1e and raises the same `TypeError` now, and `test_get_provider_kwargs_api_key_gla_mode` already exercises the Vertex hook with a `google:` model. `google-cloud:` on the same connection is fine, so `google` looks like it needs its own entry emitting only `GoogleProvider`'s surface rather than sharing the Vertex one.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -127,6 +138,35 @@ def _get_provider_kwargs(
...
+                "API mode is now selected via the model prefix ('google-cloud:' vs. 'google:')."
+            )
+
+    def _get_provider_kwargs_for_model(self, conn: Connection, model_name: str) -> dict[str, Any]:
+        provider_name, _ = parse_model_id(model_name)
+        provider_kwargs_mapper = _PROVIDER_KWARGS_MAPPER_BY_MODEL_PREFIX.get(
+            provider_name, PydanticAIHook._get_provider_kwargs
+        )
+        extra = conn.extra_dejson
+        self._warn_if_vertexai_field_ignored(extra)
```
kaxil, line 164: The mapper is chosen from the model prefix alone, with no reference to the connection or hook it belongs to, and the Bedrock/Vertex mappers were written for connection types that hide `password` and `host`, so they read only `extra`. On a generic `pydanticai` connection those two fields are visible and documented as the API key and the base URL, and they are now dropped for `bedrock:`, `google:` and `google-cloud:` (`azure:` and `openai:` are unaffected). Concretely: `model: "bedrock:us.anthropic.claude-opus-4-6-v1:0"` with a bearer token in API Key and a custom endpoint in Host forwarded both to `BedrockProvider` at d195f1e (it takes `api_key` and `base_url`) and now yields `{}`, so the task silently runs under the worker's AWS identity against the default endpoint, with no trace because the info log only fires for a non-empty dict. Since the mappers hide those fields on their own conn types, a `conn.password`/`conn.host` fallback when `extra` supplies neither can only fire for generic connections. Related: because dispatch never consults `self`, the `_get_provider_kwargs` docstring's "subclasses override this method" no longer describes what happens -- the three shipped subclasses are reached only because this dict re-registers them. Worth saying explicitly that the fix there is the wording (or declaring the prefixes on each hook class and populating the table from them), not restoring `self`-dispatch -- that would undo the cross-provider mapping `test_embedding_uses_model_provider_mapping_instead_of_hook_type` pins.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -127,6 +138,35 @@ def _get_provider_kwargs(
...
+        return provider_kwargs_mapper(conn.password, conn.host, extra)
+
+    def _create_provider_factory(self, provider_kwargs: dict[str, Any]) -> Callable[[str], Any]:
+        def _create_provider(provider_name: str) -> Any:
+            if provider_name.startswith("gateway/"):
+                return gateway_provider(provider_name.removeprefix("gateway/"), **provider_kwargs)
+            return infer_provider_class(provider_name)(**provider_kwargs)
```
kaxil: `infer_embedding_model` calls `provider_factory(provider_name)` unconditionally before its dispatch chain, including for the one branch that never uses a provider: `SentenceTransformerEmbeddingModel(model_name)` takes no `provider=`. Since `embed_conn_id` defaults to `llm_conn_id`, a hosted-chat connection with an API key plus `embed_model: "sentence-transformers:all-MiniLM-L6-v2"` now dies on `TypeError: SentenceTransformersProvider() takes no arguments` -- it reached the model at d195f1e, because `infer_provider("sentence-transformers")` succeeds and the old swallow routed to it, so hosted chat plus local embeddings on one connection is newly impossible. Separately, the base mapper's fixed `{api_key, base_url}` shape is wider than several constructors (`CohereProvider(api_key, cohere_client, http_client)` and `DeepSeekProvider` reject `base_url`, `LiteLLMProvider` spells it `api_base`, `SnowflakeProvider` has no `api_key`), and failing loudly there is the improvement I asked for -- but neither `infer_model` nor `infer_embedding_model` wraps the factory, so the user gets a bare `TypeError` from pydantic-ai internals naming no connection field. Re-raising as something that names the conn id and the kwargs supplied would keep the loud failure and restore the diagnostic the old warning carried.

providers/common/ai/tests/unit/common/ai/hooks/test_pydantic_ai.py
```diff
@@ -240,6 +251,279 @@ def test_get_conn_caches_model(self, mock_infer_model):
...
+                "google",
+                {
+                    "api_key": "extra-key",
+                    "base_url": "https://extra.example.com",
+                    "project": "project",
+                },
+                {
+                    "api_key": "extra-key",
+                    "base_url": "https://extra.example.com",
+                    "project": "project",
```
kaxil: Two rows here pin the shapes the code gets wrong. The `google` row asserts `GoogleProvider(api_key=..., base_url=..., project="project")`, which the real class rejects, and the `bedrock` row asserts only `{"region_name": ...}` while the connection has `password` and `host` set -- so it encodes the discarded credentials as intended behaviour. Both pass only because `infer_provider_class` is patched, so `mock_infer_provider_class.return_value` swallows any kwargs; `Non-DB-prov::3.10:common.ai` is green at HEAD. `test_get_conn_vertexai_flag_is_not_forwarded` already shows the pattern that would have caught it -- a `FakeGoogleCloudProvider` carrying the real keyword-only signature -- and applying that to `GoogleProvider` for the `google` row would close it. Worth noting the `google` and `google-cloud` rows carry byte-identical expectations, so nothing here tests the one place those two classes actually diverge.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -127,6 +138,35 @@ def _get_provider_kwargs(
...
+        provider_kwargs_mapper = _PROVIDER_KWARGS_MAPPER_BY_MODEL_PREFIX.get(
+            provider_name, PydanticAIHook._get_provider_kwargs
+        )
+        extra = conn.extra_dejson
+        self._warn_if_vertexai_field_ignored(extra)
+        return provider_kwargs_mapper(conn.password, conn.host, extra)
+
+    def _create_provider_factory(self, provider_kwargs: dict[str, Any]) -> Callable[[str], Any]:
+        def _create_provider(provider_name: str) -> Any:
+            if provider_name.startswith("gateway/"):
```
kaxil: `gateway/` is a new user-facing model-id syntax and it is documented nowhere: grepping the provider's docs, `provider.yaml` and `src` finds it only at this branch and the import above. Meanwhile `docs/self_hosted_models.rst:259-281` documents a different AI-gateway wiring -- `host` as the gateway base URL, `password` as the bearer token, `extra["model"]` as `openai:<route>` -- so there are now two gateway mechanisms with incompatible `host` semantics and only the older one written down. The branch is also unreachable from the Bedrock and Vertex connection types: `parse_model_id` returns `"gateway/google"`, which matches no key in the mapper table, so the base mapper runs and those types hide `password`/`host`, giving `{}` and no `provider_factory` at all (verified: generic conn yields `{api_key, base_url}`, vertex and bedrock conns yield `{}`). Normalising the prefix before the table lookup would fix the second half.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -144,60 +184,85 @@ def get_conn(self) -> Model:
...
+                list(provider_kwargs),
+            )
+            embedding_model = infer_embedding_model(
+                embed_model_name,
+                provider_factory=self._create_provider_factory(provider_kwargs),
+            )
+        else:
+            embedding_model = infer_embedding_model(embed_model_name)
+
+        self._embedder = Embedder(embedding_model, instrument=genai_instrumentation_settings())
```
kaxil: `get_embedder()` takes no arguments, so `EmbeddingSettings` is unreachable -- at 2.31.1 `Embedder.__init__` is `(model, *, settings, defer_model_check, instrument)` and `EmbeddingSettings` carries `dimensions`/`truncate`/`extra_headers`/`extra_body`. `dimensions` is the one people reach for (text-embedding-3-small at 512 dims for a fixed-width vector column), and setting it today means building `Embedder` by hand, which forfeits exactly the credential resolution this PR adds. `create_agent` in this same class already solves the shape with `**agent_kwargs` and the `_UNSET` sentinel letting a caller-supplied `instrument` win over auto-instrumentation; `def get_embedder(self, **embedder_kwargs)` reusing that pattern would match it, with the cache either keyed on the kwargs or refusing them after the first call.

providers/common/ai/provider.yaml
```diff
@@ -322,6 +343,13 @@ connection-types:
...
+        label: Embedding Model
+        description: "Google embedding model identifier (e.g. google-cloud:text-embedding-005)"
```
kaxil, line 349: Still open from the last round: the three prefix tripwires (`test_conn_fields_model_description_prefix_is_valid_provider` and its two siblings) read only `conn-fields["model"]["description"]` for `pydanticai-vertex`, so the four new `embed_model` prefixes have no drift guard. All four resolve today -- I ran `openai:`, `azure:`, `bedrock:` and `google-cloud:` through `infer_embedding_model` on 2.31.1 -- so this is a guard rather than a live bug. It needs the same shape as `_assert_prefix_is_known_provider` with a wider suppress: an unrecognised prefix gives `ValueError: Unknown provider: <prefix>`, while a recognised one with no key on the test box gives `ImportError` or a credential `UserError`, so only the `ValueError` should fail the test.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
jeff3071: Here is my concern.

For example, consider a generic connection configured with:
```
{
  "conn_type": "pydanticai",
  "password": "sk-openai-...",
  "host": "https://api.openai.com/v1",
  "extra": {
    "model": "openai:gpt-5.6-sol",
    "embed_model": "google:text-embedding-004"
  }
}
```
Here, password and host belong to the OpenAI model. If the Google mapper falls back to those common connection fields, it would also pass the OpenAI API key and endpoint to GooglefinderProvider. That is why I chose to use only provider-specific values from extra for Bedrock and Google rather than implicitly sharing the generic credentials.

I agree that silently dropping documented fields is surprising, though, particularly when the generic connection contains only a Bedrock or Google model. 

Would you prefer that we preserve the generic connection behavior by falling back to `conn.password` and `conn.host` when the corresponding provider-specific values are absent from extra, accepting that users who mix providers should configure a separate `embed_conn_id`?

kaxil: You're right, drop the fallback as I put it. On that connection `password` and `host` are the OpenAI ones, and handing them to `GoogleProvider` is the same wrong-identity bug I flagged last round, just moved.

The narrower version I had in mind doesn't work either. Fall back only when the connection names a single provider, and `PydanticAIHook(model_id="bedrock:...")` against a connection storing `model: openai:...` still passes: one prefix in play, so the OpenAI key goes to `BedrockProvider` as a bearer token. `model_id` is a documented override, so that isn't a corner case.

Your change did close a real leak, for the five mapped prefixes. Before it `get_embedder()` used the hook-bound mapper, so a generic connection handed `{api_key, base_url}` straight to `GoogleProvider`, which accepts both, and the OpenAI key went out as an `x-goog-api-key` header against `api.openai.com`. No `TypeError`, so the old swallow never saw it. That only ever existed inside this PR's own history though: `main` has no `get_embedder`.

What still blocks is a regression against `main`, and it's in `get_conn()` rather than embeddings. Generic `pydanticai` connection, `model: bedrock:us.anthropic.claude-opus-4-6-v1:0`, bearer token in API Key, custom runtime endpoint in Host:

```
main: {"api_key": "bearer-tok", "base_url": "https://custom-bedrock"}
HEAD: {}
```

`BedrockProvider` takes both of those. At HEAD the prefix table picks the extra-only mapper, so that same connection now runs under the worker's AWS identity against the default endpoint. There's only one provider in play, so nothing is ambiguous, and nothing errors.

So: keep extra-only precedence, but say something when a mapper throws away a non-empty `conn.password` or `conn.host`, and name the `extra` keys that replace them. Those keys (`region_name`, `api_key`, `base_url`, `project`, `location`) do work on a generic connection. They're just documented only on the `pydanticai-bedrock` and `pydanticai-vertex` pages, which both open by saying those credential shapes don't fit this connection type. `pydantic_ai.rst` tells you to leave API Key empty for Bedrock and Vertex, which reads as "not needed" rather than "ignored", and says nothing at all about Host.

One more that sits outside the table: `cohere:` is one of the embedding providers pydantic-ai ships, and on a connection with API Key set and Host empty, `model: openai:...` plus `embed_model: cohere:embed-english-v3.0` hands the OpenAI key to `CohereProvider`. Set Host as well and it's an uncaught `TypeError`, since `CohereProvider` has no `base_url`. And yes, `embed_conn_id` is the answer for genuinely mixed providers.

jeff3071:
I think pydanticai gateway support should be handled in a separate PR. Now I revert it.


Changes in this update:
1. Prevent different providers from sharing the same connection ID ( When llm_conn_id and embed_model_id are same, it should use same provider).
2. Split the google and google-cloud mappings.
3. Pass additional kwargs to Embedder. 
4. Warn when password or host is ignored in favor of provider-specific extras.
5. Add real-signature coverage for GoogleProvider and drift guards for all documented embedding prefixes.

jeff3071:
Fix a issue related default llm conn id.

For example
```python
PydanticAIHook(
    embed_conn_id="embedding_conn",
    embed_model_id="openai:text-embedding-3-small",
)
```

`llm_conn_id` will be set to `default_conn_name` that would cause `_get_cached_connection` raise.

jeff3071:
Summary

- Fixed credential mapping based on each model’s provider.
- Split `google` and `google-cloud` mappings.
- Require a separate `embed_conn_id` when LLM and embedding providers differ. (`sentence-transformers` except)
- Added warnings when Bedrock/Google ignore Password or Host fields.
- Added Embedder kwargs forwarding, instrumentation precedence, and argument-aware caching.
- Fixed embedding-only setups without a LLM connection. (llm_conn_id set to default)
- Reverted Gateway support.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -127,6 +159,96 @@ def _get_provider_kwargs(
...
+            list(provider_kwargs),
+        )
+
+        def create_provider(provider: str) -> Any:
+            try:
+                return infer_provider_class(provider)(**provider_kwargs)
+            except TypeError as e:
+                raise TypeError(
+                    f"Provider {provider!r} rejected connection {conn.conn_id!r} fields "
+                    f"mapped to kwargs {sorted(provider_kwargs)}"
```
kaxil: This names the kwargs that were supplied but not the one the provider actually rejected, and the cause does not reach the user. `test_connection` at :461 returns `str(e)`, not a formatted traceback, so the Test button in the UI shows only this line. Reproduced against 2.31.1 with a generic connection carrying an API key and a Host, `embed_model: cohere:embed-v4.0`:

```
str(e):        Provider 'cohere' rejected connection 'my_conn' fields mapped to kwargs ['api_key', 'base_url']
str(e.__cause__):  CohereProvider.__init__() got an unexpected keyword argument 'base_url'
```

Both kwargs came from the connection and only `base_url` is wrong, so the message as rendered points at the key. `self_hosted_models.rst:320`, added by this PR, tells the reader the TypeError "identifies the rejected connection fields", which describes the cause line rather than this one. Appending `: {e}` to the message restores it.

The same one-liner covers the other half: the `try` wraps the whole constructor call, so a TypeError raised inside a provider's body (an SDK version skew, say) currently gets relabelled as a connection-field problem with no hint otherwise.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -540,3 +697,21 @@ def _get_provider_kwargs(
...
+_PROVIDER_CONNECTION_CONFIGS: dict[str | None, _ProviderConnectionConfig] = {
+    "azure": _ProviderConnectionConfig(PydanticAIAzureHook._get_provider_kwargs),
+    "azure-responses": _ProviderConnectionConfig(PydanticAIAzureHook._get_provider_kwargs),
+    "bedrock": _ProviderConnectionConfig(
+        PydanticAIBedrockHook._get_provider_kwargs,
+        ("api_key", "base_url", "region_name"),
+    ),
+    "google": _ProviderConnectionConfig(
+        PydanticAIVertexHook._get_google_provider_kwargs,
+        ("api_key", "base_url"),
```
kaxil: `replacement_fields` drives a warning that inspects only `conn.password` and `conn.host`, and those two are `hidden_fields` on `pydanticai_vertex`, so on that connection type it can never fire. Meanwhile `_get_google_provider_kwargs` reads just `api_key` and `base_url` out of `extra` (correctly, `GoogleProvider` takes no `project`, `location` or `credentials` on 2.31.1). A Vertex connection that fills in the GCP Project and Location conn-fields and sets `model: "google:gemini-2.0-flash"` therefore gets `{}`, no `provider_factory`, and ADC, with two documented form fields discarded and an empty task log. `service_account_info` goes the same way.

That combination is first-class here, not a misconfiguration: `test_get_conn_google_uses_google_provider_signature` exercises exactly it. And `main` was louder on this path, since the old self-dispatch produced `{project, location}`, hit `GoogleProvider`'s TypeError and logged `Provider 'google' rejected kwargs ['project', 'location']; falling back to env-var auth`.

A second tuple on `_ProviderConnectionConfig` for the extra keys a mapper is known to drop, warned about next to the existing one, closes it inside the structure already here.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -540,3 +697,21 @@ def _get_provider_kwargs(
...
 
         return kwargs
+
+
+_PROVIDER_CONNECTION_CONFIGS: dict[str | None, _ProviderConnectionConfig] = {
+    "azure": _ProviderConnectionConfig(PydanticAIAzureHook._get_provider_kwargs),
+    "azure-responses": _ProviderConnectionConfig(PydanticAIAzureHook._get_provider_kwargs),
+    "bedrock": _ProviderConnectionConfig(
+        PydanticAIBedrockHook._get_provider_kwargs,
+        ("api_key", "base_url", "region_name"),
```
kaxil: The Bedrock tuple lists `api_key`, `base_url` and `region_name`, but the mapper also reads `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` and `profile_name`. The warning presents this list as the provider-specific values that replace the ignored ones, so a Bedrock user whose Password held an access key is steered to `api_key`, which is the bearer token and takes precedence over IAM keys once set. Listing all seven keeps the message from suggesting a different auth mode.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -127,6 +159,96 @@ def _get_provider_kwargs(
...
+    def _validate_embedding_connection_provider(self, embed_model_name: str, extra: dict[str, Any]) -> None:
+        if self.embed_conn_id != self.llm_conn_id:
+            return
+
+        llm_model_name = self.model_id or extra.get("model", "")
+        if not llm_model_name:
+            return
+
+        llm_provider, _ = parse_model_id(llm_model_name)
+        embed_provider, _ = parse_model_id(embed_model_name)
```
kaxil, line 249: `parse_model_id` returns `(None, name)` for a value with no prefix, so `{"model": "openai:gpt-4o", "embed_model": "text-embedding-3-small"}` falls into the mismatch branch and reports `configures different LLM and embedding providers ('openai' and None). Set embed_conn_id to a separate connection for the embedding provider.` A separate connection does not help; drop the `model` key and pydantic-ai gives the accurate `ValueError: You must provide a provider prefix when specifying an embedding model name` instead.

The prefixless form is a plausible thing to type into this field: the `llamaindex` conn type in this same `provider.yaml` documents `embed_model` as `text-embedding-3-small`. `if embed_provider is None: return` hands it back to pydantic-ai.

providers/common/ai/tests/unit/common/ai/hooks/test_pydantic_ai.py
```diff
@@ -239,6 +277,411 @@ def test_get_conn_caches_model(self, mock_infer_model):
...
+            extra=json.dumps({"model": model_name}),
+        )
+
+        with (
+            patch.object(hook, "get_connection", return_value=conn),
+            patch.object(hook.log, "warning", autospec=True) as mock_warning,
+        ):
+            hook.get_conn()
+
+        mock_warning.assert_called_once_with(
```
kaxil, line 324: This is one assertion away from also pinning extra-only precedence, which is the thing round 3 settled. Nothing in the file currently fails if the Google mappers grow the fill-in-when-absent `conn.password` / `conn.host` fallback: the two direct mapper tests pass `None, None` for those parameters, the `google` and `google-cloud` rows of `test_embedding_credentials_are_mapped_by_model_provider` already carry `api_key` and `base_url` in `extra` so a fill-if-absent branch never fires, and this test sets the connection fields but checks only the warning.

All three rows here limit `extra` to `model`, so each resolves to `{}` and no `provider_factory` today. Adding `mock_infer_model.assert_called_once_with(model_name)` below covers bedrock, google and google-cloud in one line and turns that mutation red.

providers/common/ai/docs/connections/pydantic_ai.rst
```diff
@@ -67,14 +88,24 @@ Host (optional)
...
-    When using the UI, the "Model" field above writes to this same location
-    automatically.
+    When using the UI, the "Model" and "Embedding Model" fields above write to
+    this same location automatically.
+
+    Bedrock-specific fields include ``api_key``, ``base_url``, ``region_name``,
+    AWS credentials and profile fields, and read/connect timeouts. ``google:``
+    accepts ``api_key`` and ``base_url``; ``google-cloud:`` additionally accepts
+    ``project``, ``location``, and ``service_account_info``. See the dedicated
+    :doc:`pydantic_ai_bedrock` and :doc:`pydantic_ai_vertex` connection pages for
```
kaxil, line 111: This sends the reader to two pages that now say the opposite of what the hook does. `pydantic_ai_bedrock.rst:26-29` and `pydantic_ai_vertex.rst:25-29` both open with those credential shapes being "none of which fit the plain ``api_key`` + ``base_url`` shape that the generic :doc:`pydantic_ai` connection assumes", which held while the mapper was picked by hook class. With the prefix table, a generic `pydanticai` connection carrying `model: "bedrock:..."` plus `region_name` and IAM keys in `extra` builds `BedrockProvider` with all of them. What is left of the difference is that the dedicated types give those fields their own form inputs and hide `password`/`host`, not that the generic one cannot reach the provider. Worth a sentence on each opener, since this new cross-reference is what will send people there.

providers/common/ai/docs/hooks/pydantic_ai.rst
```diff
@@ -56,6 +57,46 @@ The model can be specified at three levels (highest priority first):
...
+arguments return the cached instance; passing different arguments creates and caches
+a new instance.
+
+.. code-block:: python
+
+    from pydantic_ai.embeddings import EmbeddingSettings
+
+    embedder = hook.get_embedder(
+        settings=EmbeddingSettings(dimensions=512),
+        defer_model_check=False,
```
kaxil: `defer_model_check=False` is a no-op through `get_embedder()`. `Embedder.__init__` does `self._model = model if defer_model_check else infer_embedding_model(model)`, and `infer_embedding_model` returns an `EmbeddingModel` instance unchanged, which is always what the hook passes. The flag only does anything when `Embedder` is given a model string. Dropping it from the example avoids teaching a setting that cannot fire here.

kaxil:
Two blocking items in the review above, both small, and everything from the earlier rounds is now closed.

The first is the `TypeError` re-raise in `create_provider`: it names the kwargs that were supplied but not the one the provider rejected, and since `test_connection` returns `str(e)` the cause never reaches the Test button. Appending `: {e}` covers it, and also stops a `TypeError` raised inside a provider's own body being relabelled as a connection-field problem.

The second is the new ignored-fields warning. It looks only at `conn.password` and `conn.host`, which `pydanticai_vertex` hides, so a Vertex connection with Project and Location set plus a `google:` model drops both fields and reaches ADC with nothing in the log. A second tuple on `_ProviderConnectionConfig` for the extra keys a mapper is known to drop fits the structure that is already there.

Fix those two and I'll approve. The rest of the comments are optional and none of them need another review round: I can confirm the two fixes from the diff.

kaxil:
Approving this. Everything from the last round landed: the provider `TypeError` now interpolates the cause, `ignored_extra_fields` covers the Google case, the Bedrock replacement list is complete, an unprefixed `embed_model` returns early, the precedence assertion is in, and both vendor page openers are rewritten.

Five follow-ups below, none of them gating.

The first one I should have caught last round, so it is not a condition on this approve. `_validate_embedding_connection_provider` runs before the provider factory is resolved (`hooks/pydantic_ai.py:316`), so it fires on connections where nothing could reach the wrong provider. A `pydanticai_bedrock` connection hides `password` and `host`, so an `openai:` embed model falls to the base mapper, which reads exactly those two, returns `{}`, and leaves the factory `None`. That embedder would resolve through `OPENAI_API_KEY`, which is resolution step 2 in `get_conn`'s own docstring, yet the call still raises `configures different LLM and embedding providers ('bedrock' and 'openai')` and points at an `embed_conn_id` connection that would carry no credentials. Moving the call below the factory resolution and gating on a non-`None` factory matches the rationale already written at `docs/hooks/pydantic_ai.rst:66`. Gating on "has a registered mapper" instead would leave a hole, since the azure mapper is the one that does read `password` and `host`.

`hooks/pydantic_ai.py:250`: the prefixless fix covers `embed_model`, but an unprefixed `model` reaches the same branch from the other side, so `model: "gpt-4o"` with a prefixed embed model reports `(None and 'openai')` and sends you to `embed_conn_id` when the mistake is in the Model field. `test` is the one unprefixed entry in `KnownModelName`, so a working `TestModel` connection with an embedding model configured hits it too.

`hooks/pydantic_ai.py:733`: `google-cloud` names four of the five keys `_get_google_cloud_provider_kwargs` reads and leaves out `service_account_info`, which is credential-resolution step 1 in `PydanticAIVertexHook`'s own docstring. Same gap you closed for Bedrock one entry up.

`docs/hooks/pydantic_ai.rst:86`: `defer_model_check` came out of the example but is still listed as an accepted kwarg. `get_embedder` always passes an already-inferred model and `infer_embedding_model(m) is m`, so either value produces the same object.

`docs/connections/pydantic_ai.rst:193`: the resolution-order block scopes the `embed_conn_id` fallback to credentials and endpoints, but `get_embedder` reads the `embed_model` string off that connection too, so setting `embed_conn_id` and leaving `embed_model` on the LLM connection gives a "no embedding model specified" error about a field you did set.

kaxil (dismissed):
Approving. Follow-ups in the comment above, none of them gating.

jeff3071:
There are a few conflicts with #72156. 

Here are the proposed changes:
- Keep `_get_conn_and_extra`, and remove `_get_cached_connection` and `_get_cached_connection_extra_dejson`.
- Update `_seed_connection` to store the connection and its extra data in `self._connections` and `self._connection_extra_dejson`, respectively.
- Update provider_fallback.rst, since we no longer retry using the environment-variable-based provider.

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
```diff
@@ -220,8 +257,92 @@ def _seed_connection(self, conn: Connection) -> None:
...
+
+    def _validate_embedding_connection_provider(self, embed_model_name: str, extra: dict[str, Any]) -> None:
+        if self.embed_conn_id != self.llm_conn_id:
+            return
+
+        llm_model_name = self.model_id or extra.get("model", "")
+        if not llm_model_name:
+            return
+
+        llm_provider, _ = parse_model_id(llm_model_name)
```
kaxil: This one came out of my own suggestion on https://github.com/apache/airflow/pull/71437#discussion_r4032377807, so the regression is mine as much as yours. The prefixless `embed_model` half is right; the `llm_provider is None` arm that came with it skips the credential guard altogether.

Whether the guard runs now depends on whether the user typed a prefix that the connection pages call optional. Measured against HEAD on pydantic-ai 2.31.1 with the real provider classes, on a generic `pydanticai` connection holding an API key, `model: "gpt-4o"`, `embed_model: "cohere:embed-v4.0"`:

```
guard   -> passes (llm_provider is None, early return)
kwargs  -> {'api_key': 'sk-OPENAI-KEY'}
embedder-> CohereEmbeddingModel via CohereProvider, base_url=https://api.cohere.com
```

The OpenAI key is sent to Cohere. Writing `openai:gpt-4o` on that same connection raises correctly, and that is the case the guard exists for.

It rejects in the other direction too. A `pydanticai_bedrock` connection with Model `us.anthropic.claude-opus-4-6-v1:0`, the bare native id `pydantic_ai_bedrock.rst:65-74` documents, colon and all, plus Embedding Model `bedrock:amazon.titan-embed-text-v2:0`, is refused with `configures different LLM and embedding providers ('us.anthropic.claude-opus-4-6-v1' and 'bedrock')`, because `parse_model_id` splits on the id's own colon. Both are Bedrock, `get_conn()` resolves that same string to `BedrockConverseModel` behind `BedrockProvider`, and the separate `embed_conn_id` the message prescribes is not the fix.

Both directions have the same root: this line derives the LLM provider with a bare `parse_model_id`, while every other path goes through `_qualify_model_name` / `_has_recognized_provider_prefix`, which ignores a `:` that is not a real provider name and falls back to `model_provider`. Reusing that here, roughly `llm_provider = parse_model_id(llm_model_name)[0] if _has_recognized_provider_prefix(llm_model_name) else self.model_provider`, and returning early only when it is still `None`, covers both. `test_prefixless_llm_model_skips_embedding_provider_validation` pins the current arm, and no test sets a bare vendor model name together with an `embed_model` on one connection.

While you are in here, the `embed_provider == "sentence-transformers"` arm on the next line is unreachable: `_get_provider_factory_for_model` already returns `None` for that prefix, and this validator only runs where the factory is not `None`.

providers/common/ai/docs/connections/pydantic_ai.rst
```diff
@@ -72,21 +97,31 @@ Host (optional)
...
+    When using the UI, the "Model" and "Embedding Model" fields above write to
+    this same location automatically.
 
 Fallback Connections
     Other connection IDs to fail over to, in order, while this provider is
     unavailable. Stored in ``extra["fallback_conn_ids"]``. Entries may name any
     ``pydanticai`` connection type, so one chain can span vendors. See
     :doc:`/provider_fallback`.
 
+    Bedrock-specific fields include ``api_key``, ``base_url``, ``region_name``,
```
kaxil, line 112: This paragraph describes `extra` fields but is indented under the Fallback Connections item that starts at :112, so it renders as part of the fallback description. The Extra (JSON, optional) item it belongs under ends at :110.

jeff3071: fix this paragraph position in  6d70c5f

providers/common/ai/src/airflow/providers/common/ai/hooks/pydantic_ai.py
jeff3071: I think reusing `_qualify_model_name()` here is better. A generic pydanticai connection still produces `llm_provider = None` for a bare model. So it still pass the guard.

`_qualify_model_name()` already defines the behavior for relevant cases:

- recognized `provider:model` names are preserved
- bare native model IDs on vendor connections are qualified with `model_provider`, including Bedrock IDs containing their own colon;
- bare model names on a generic connection raise the same actionable `ValueError` as the normal path

PR #71441, opened 2026-08-11
Mark only a run's most recent asset event as triggering it

### LINKED ISSUE #56749: Asset Event trigger attribution is misleading
### Apache Airflow version

3.1.0

_No response_

### What happened?

When triggering a DAG via Asset Events, one of three things can happen:

 1. A single Asset Event triggers a DAG Run
 2. Multiple Asset Events trigger a single DAG Run
 3. Asset Events that haven't triggered a DAG Run, but are older than the last run are silently ignored

In this example:
Image

The consumer DAG runs twice.
It consumes 1 Asset Event on the first run, two on the second, and stops running.

However if we look at the Asset Events:
Image

We see there are 5 Asset Events, and 3 of them have "triggered" DAG Runs.

This is misleading as it makes the user believe there should be 3 DAG Runs, not 2.

I think we should add the following changes in the Asset UI page (and the corresponding internal changes):

 - `Triggered DAG Runs` should only be used on the most **recent** Asset Event that a DAG Run consumes.
 - `Included in DAG Run` mention should be used for all other consumed

As for the ignored Asset Events, I see one of two possible solutions:

 1. Either this is a bug, and they should trigger a DAG Run
 2. Or this is a feature and they should have some sort of `silently consumed` state & label attached to them.

### How to reproduce

To reproduce simply upload the following DAGs in a brand new Airflow instance:

[check_dataset_sync.py](https://github.com/user-attachments/files/22961635/check_dataset_sync.py)

make sure to use a DB other than SQlite so you can compare the difference between `max_active_runs=1` and `max_active_runs=10`.

Then use the `airflow standalone` command.
Turn all the DAGs on. And manually trigger the `asset generator` DAG once.

### Operating System

Ubuntu 24

```
apache-airflow-providers-common-compat   1.7.3
apache-airflow-providers-common-io       1.6.2
apache-airflow-providers-common-sql      1.27.5
apache-airflow-providers-postgres        6.2.3
apache-airflow-providers-smtp            2.2.0
apache-airflow-providers-standard        1.6.0
```

### Deployment

Virtualenv installation

### Deployment details

_No response_

### Anything else?

Do not use the default SQLite Database as it cannot show the parallelism issues

### Code of Conduct

pierrejeambrun:
Fixes the misleading "Triggered Dag Run" attribution in the Asset Events view.

An asset event is linked to every dag run that consumed it, so when a run consumed several events the UI showed "Triggered Dag Run" on each of them — making it look as though more runs had been created than actually were. Now only a run's most recent consumed event is labelled "Triggered Dag Run"; the earlier consumed events are labelled "Included in Dag Run".

The `/assets/events` endpoint flags the triggering event per run (ranking each run's consumed events by timestamp) and exposes it as a `triggering` field on `DagRunAssetReference`; the Asset Events UI renders the "triggered" and "included" groups accordingly.

Scope: this covers the attribution/labelling half of the issue. The separate "older events silently ignored" scheduling behaviour was addressed by #62501 and is out of scope here.

closes: #56749

##### Screenshots (Asset Events view) — _to attach_
- Before: several events under one run each showing "Triggered Dag Run"
- After: newest event "Triggered Dag Run", earlier ones "Included in Dag Run"

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
@@ -29,7 +29,7 @@ from airflow._shared.timezones import timezone
 from airflow.api_fastapi.app import get_auth_manager
 from airflow.api_fastapi.auth.managers.models.resource_details import DagAccessEntity, DagDetails
 from airflow.api_fastapi.common.dagbag import DagBagDep, get_latest_version_of_dag
-from airflow.api_fastapi.common.db.assets import resolve_triggering_event_ids
+from airflow.api_fastapi.common.db.assets import resolve_triggering_event_refs
 from airflow.api_fastapi.common.db.common import SessionDep, paginated_select
 from airflow.api_fastapi.common.parameters import (
     BaseParam,
@@ -378,16 +378,12 @@ def get_asset_events(
     # dependency-applied timeout is still active.
     assets_events = session.scalars(assets_event_select).all()
 
-    # An asset event is linked to every dag run that consumed it, so a run that consumed several
-    # events would otherwise appear as "triggered" on each one. Only a run's most recent consumed
-    # event actually triggers it; the rest were merely included.
-    triggering_event_id_by_run = resolve_triggering_event_ids(assets_events, session=session)
+    triggering_refs = resolve_triggering_event_refs(assets_events, session=session)
     asset_event_responses = []
     for event in assets_events:
-        event_response = AssetEventResponse.model_validate(event)
-        for reference, dag_run in zip(event_response.created_dagruns, event.created_dagruns):
-            reference.triggering = triggering_event_id_by_run.get(dag_run.id) == event.id
-        asset_event_responses.append(event_response)
+        for run in event.created_dagruns:
+            run.triggering = (event.id, run.dag_id, run.run_id) in triggering_refs
+        asset_event_responses.append(AssetEventResponse.model_validate(event))
 
     return AssetEventCollectionResponse(
         asset_events=asset_event_responses,
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
@@ -29,6 +29,7 @@ from airflow._shared.timezones import timezone
 from airflow.api_fastapi.app import get_auth_manager
 from airflow.api_fastapi.auth.managers.models.resource_details import DagAccessEntity, DagDetails
 from airflow.api_fastapi.common.dagbag import DagBagDep, get_latest_version_of_dag
+from airflow.api_fastapi.common.db.assets import resolve_triggering_event_ids
 from airflow.api_fastapi.common.db.common import SessionDep, paginated_select
 from airflow.api_fastapi.common.parameters import (
     BaseParam,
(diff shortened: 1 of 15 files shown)


pierrejeambrun:
## Before
Screenshot 2026-08-11 at 17 28 28

## After
Screenshot 2026-08-11 at 17 51 11

Screenshot 2026-08-11 at 17 51 28

airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
```diff
@@ -377,8 +378,19 @@ def get_asset_events(
...
 
+    # An asset event is linked to every dag run that consumed it, so a run that consumed several
+    # events would otherwise appear as "triggered" on each one. Only a run's most recent consumed
+    # event actually triggers it; the rest were merely included.
+    triggering_event_id_by_run = resolve_triggering_event_ids(assets_events, session=session)
+    asset_event_responses = []
+    for event in assets_events:
+        event_response = AssetEventResponse.model_validate(event)
+        for reference, dag_run in zip(event_response.created_dagruns, event.created_dagruns):
+            reference.triggering = triggering_event_id_by_run.get(dag_run.id) == event.id
```
bbovenzi: Why zip instead of actually checking the two ids?

airflow-ctl/src/airflowctl/api/datamodels/generated.py
```diff
@@ -533,6 +533,13 @@ class DagRunAssetReference(BaseModel):
     data_interval_start: Annotated[datetime | None, Field(title="Data Interval Start")]
     data_interval_end: Annotated[datetime | None, Field(title="Data Interval End")]
     partition_key: Annotated[str | None, Field(title="Partition Key")]
+    triggering: Annotated[
+        bool | None,
+        Field(
+            description="Whether this asset event triggered the referenced dag run. Only a run's most recent consumed asset event triggers it; earlier consumed events are included in the run but did not trigger it.",
+            title="Triggering",
+        ),
+    ] = True
```
bbovenzi: Shouldn't the default be None?

airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
```diff
@@ -282,16 +281,28 @@ def get_upstream_asset_events(
             DagRun.dag_id == dag_id,
             DagRun.run_id == dag_run_id,
         )
-        .options(joinedload(DagRun.consumed_asset_events).joinedload(AssetEvent.asset))
+        .options(
+            joinedload(DagRun.consumed_asset_events).joinedload(AssetEvent.asset),
+            joinedload(DagRun.consumed_asset_events).subqueryload(AssetEvent.created_dagruns),
```
bbovenzi, line 286: Let's add an N+1 test for this

airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
pierrejeambrun: Done,  now matching exactly ids.

pierrejeambrun: Done

airflow-ctl/src/airflowctl/api/datamodels/generated.py
pierrejeambrun: It will never be 'None' it's always popupaled, either True/False. I chose 'True' default because it mimics the current behavior in case the logic is wrong or something unexpected happens. (All events are considered triggering). 

I removed the 'default' value all together which is confusing indeed, code should set this explicitly.

airflow-core/src/airflow/api_fastapi/core_api/routes/public/dag_run.py
pierrejeambrun: done

bbovenzi (approved):

bbovenzi:
Just one CI failure we should resolve

PR #71443, opened 2026-08-11
Fix Asset Event table titles showing a raw translation key

pierrejeambrun:
Asset Events tables render their titles as raw i18n keys instead of readable text:
- Asset page (a specific asset's events): **"asset_one Events"** instead of "Asset Events"
- Run page: **"Source assetEvent_other"** instead of "Source Asset Events"
- Task Instance page: **"Created assetEvent_other"** instead of "Created Asset Events"

The shared Asset Events table (`components/Assets/AssetEvents.tsx`) resolves these titles with `useTranslation(["dashboard", "common", "dag"])`, whose default namespace is **`dashboard`**. In `en/common.json` these keys reference their labels through nested `$t()` (introduced by the translation DRY refactor #68285), e.g. `"assetEvent_other": "$t(asset_one) Events"` and `"sourceAssetEvent_other": "Source $t(assetEvent_other)"`. An **unqualified** nested `$t(...)` resolves against the consuming hook's default namespace (`dashboard`), where those keys don't exist, so the title falls back to the raw key.

This keeps the DRY nesting and fixes it by **qualifying the nested references with the `common` namespace** (`$t(common:asset_one)`, `$t(common:assetEvent_other)`, …), so they resolve in `common` regardless of the consuming component's default namespace.

Verified with i18next directly: all three keys now render "Asset Event(s)" / "Source Asset Event(s)" / "Created Asset Event(s)" for every count, both in the `dashboard`-default context and with the full namespace array.

Note: only `en` uses the nested form; the other locales currently duplicate the strings inline. Migrating them to the same nested/DRY structure is separate follow-up work.

related: #68285

---

### DIFF
--- a/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
+++ b/airflow-core/src/airflow/ui/public/i18n/locales/en/common.json
@@ -17,8 +17,8 @@
   },
   "asset_one": "Asset",
   "asset_other": "Assets",
-  "assetEvent_one": "$t(asset_one) Event",
-  "assetEvent_other": "$t(asset_one) Events",
+  "assetEvent_one": "$t(common:asset_one) Event",
+  "assetEvent_other": "$t(common:asset_one) Events",
   "assetInactive": {
     "tooltip": "Upstream asset has been deactivated; the scheduler is holding partition evaluation until it is reactivated."
   },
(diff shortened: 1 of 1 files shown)


pierrejeambrun:
## Before
Screenshot 2026-08-11 at 17 51 11

Screenshot 2026-08-11 at 17 51 28

## After 
Screenshot 2026-08-11 at 18 05 35

Screenshot 2026-08-11 at 18 07 15
Screenshot 2026-08-11 at 18 07 40

bbovenzi:
Nesting i18n keys is something we do want to do to make our translation strings more reusable. I would prefer that we actually fix nesting so that it does work for english and every other language.

pierrejeambrun:
I can make nesting work for en. And leave the other owners do the work for theirs translation, I'm not confident changing blindly 20 locals I cannot double check. 

Also we probably want to adjust the i18n scripts to detect translation keys that are using nesting in `en` but aren't updated yet in the targeted locals to use a similar nesting.

bbovenzi:
Ok, let's fix english and handle the other languages in another PR

pierrejeambrun:
> Ok, let's fix english and handle the other languages in another PR

Done

bbovenzi (approved):

Issue #62903, opened 2026-03-05
apache-airflow-providers-fab 3.4.0: Intermittent OperationalError (4031) when FAB session reuses a MySQL connection dropped due to idle timeout


davidjfrickert:
### Apache Airflow Provider(s)

fab

### Versions of Apache Airflow Providers

apache-airflow-providers-fab==3.4.0

### Apache Airflow version

3.1.7

### Operating System

Docker

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

API Server config to allow okta login. Otherwise pretty standard. Snippet below:
```
extraEnvFrom: |
  - secretRef:
      name: airflow-okta-creds
// (...)
data:
  metadataSecretName: airflow-db-conn
  resultBackendSecretName: airflow-db-conn
// (...)
env:
  - name: AIRFLOW__API__EXPOSE_CONFIG
    value: "False"
  - name: AIRFLOW__API__BASE_URL
    value: "<REDACTED>"
  - name: AIRFLOW__EMAIL__EMAIL_BACKEND
    value: "airflow.utils.email.send_email_smtp"
  - name: AIRFLOW__SMTP__SMTP_HOST
    value: "<REDACTED>"
  - name: AIRFLOW__SMTP__SMTP_MAIL_FROM
    value: "<REDACTED"
  - name: AIRFLOW__SMTP__SMTP_STARTTLS
    value: "False"
  - name: AIRFLOW__WEBSERVER__SHOW_TRIGGER_FORM_IF_NO_PARAMS
    value: "True"
  - name: AIRFLOW__WEBSERVER__WARN_DEPLOYMENT_EXPOSURE
    value: "False"
  - name: AIRFLOW__CORE__PARALLELISM
    value: "256"  
## Invalidate pool connections idle for more than 300s.
## Set below MySQL's default wait_timeout (600s) to avoid using server-closed connections.
## A lower value also speeds up recovery when the max DB connection limit is exceeded.
## NOTE: Does not apply to FAB since it does not use pool, but it uses single shared session.
  - name: AIRFLOW__DATABASE__SQL_ALCHEMY_POOL_RECYCLE
    value: "300"
// (...)
apiServer:
  apiServerConfig: |
    import os
    from flask_appbuilder.security.manager import AUTH_OAUTH

    AUTH_TYPE = AUTH_OAUTH

    # registration configs
    AUTH_USER_REGISTRATION = True  # allow users who are not already in the FAB DB

    # the list of providers which the user can choose from
[...]

PR #61339, opened 2026-02-01
AIP-84  :  Add JWT token revokation for logout invalidation (#47952)

anishgirianish:
<!--
Thank you for contributing!

Please provide above a brief description of the changes made in this pull request.
Write a good git commit message following this guide: http://chris.beams.io/posts/git-commit/

Please make sure that your code changes are covered with tests.
And in case of new features or big changes remember to adjust the documentation.

Feel free to ping (in general) for the review if you do not see reaction for a few days
(72 Hours is the minimum reaction time you can expect from volunteers) - we sometimes miss notifications.

In case of an existing issue, reference it using one of the following:

* closes: #ISSUE
* related: #ISSUE
-->

## Summary                                                                                              
                                                                                                          
 - Adds a `revoked_token` table to persist revoked JWT token JTIs on logout                              
- On logout, the token's `jti` is extracted and stored with its `exp` timestamp                         
 - On every authenticated request, `get_user_from_token` checks if the `jti` has been revoked before allowing access                                                                                         
 - Expired revoked tokens are automatically cleaned up via the existing`db_cleanup` mechanism           
                                                                                                                                           
## closes: #47952

PR #71444, opened 2026-08-11
Recover the revoked-token check from a stale DB connection (fixes idle-timeout 500)

uplsh580:
Recover the JWT revoked-token check from a stale pooled DB connection so the first authenticated request after an idle period no longer returns HTTP 500.

Issue: https://github.com/apache/airflow/issues/71395

## Problem

With MySQL as the metadata DB and the FAB auth manager, the first authenticated request after the api-server sits idle longer than MySQL's `wait_timeout` fails with HTTP 500. The exception is an `OperationalError` (MySQL error 4031, *"The client was disconnected by the server because of inactivity"*) raised from `RevokedToken.is_revoked` inside `BaseAuthManager.get_user_from_token`. The immediately following request succeeds, because the failed query resets the pooled connection — so in production it looks like intermittent 500s after overnight idle periods.

### Root cause

1. The FAB auth manager shares core's scoped `settings.Session`. A prior request (e.g. `FabAuthManager.deserialize_user`, or FAB's Flask views) can leave that scoped session bound to a connection the database later drops on idle timeout. Because the connection is never returned to the pool, there is **no checkout event** — `pool_pre_ping` / `pool_recycle` cannot help.
2. This exact failure mode was reported in #62903 and fixed by #62919, which added a discard-and-retry recovery to `FabAuthManager.deserialize_user`.
3. But 3.2.0 introduced the JWT revocation check (#61339 / AIP-84): `RevokedToken.is_revoked` in `get_user_from_token` now runs **before** `deserialize_user`, so it is the first thing to touch the poisoned scoped session — and it had no recovery logic. The 500 that #62919 fixed is back, just raised one step earlier in the auth path.

## Fix

Give the revoked-token check the same recovery `deserialize_user` already has: on `SQLAlchemyError`, discard the scoped session (`settings.Session.remove()`) and retry the check once on a fresh connection. The recovery is factored into a small `BaseAuthManager._is_token_revoked` helper so `get_user_from_token` stays readable.

This is the per-call-site variant proposed in the issue (the one that has been running in production for several weeks and eliminated these 500s). It is minimal and mirrors the established #62919 pattern rather than introducing a new session-management abstraction. A persistent DB error still surfaces — only a single transient stale-connection error is absorbed.

## Tests

Added two regression tests in `test_base_auth_manager.py`. Both raise a 4031-style `OperationalError` on the first `is_revoked` call and assert the scoped session is discarded (`settings.Session.remove()`) and the check is retried exactly once:

- retry returns *not revoked* → the request is served transparently;
- retry returns *revoked* → `InvalidTokenError("Token has been revoked")` is still raised.

related: #71395

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
+++ b/airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
@@ -22,13 +22,16 @@ import logging
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
+from contextlib import suppress
 from enum import Enum
 from functools import cache, cached_property
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar
 
 from jwt import InvalidTokenError
 from sqlalchemy import select
+from sqlalchemy.exc import SQLAlchemyError
 
+from airflow import settings
 from airflow.api_fastapi.auth.managers.models.base_user import BaseUser
 from airflow.api_fastapi.auth.managers.models.resource_details import (
     ConnectionDetails,
@@ -161,7 +164,7 @@ class BaseAuthManager(Generic[T], LoggingMixin, metaclass=ABCMeta):
             log.error("JWT token is not valid: %s", e)
             raise e
 
-        if (jti := payload.get("jti")) and RevokedToken.is_revoked(jti):
+        if (jti := payload.get("jti")) and self._is_token_revoked(jti):
             raise InvalidTokenError("Token has been revoked")
 
         try:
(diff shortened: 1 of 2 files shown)


airflow-core/newsfragments/71444.bugfix.rst
```diff
@@ -0,0 +1 @@
+Recover the JWT revoked-token check from a stale pooled database connection. On MySQL the first authenticated request after an idle period could fail with HTTP 500 (``OperationalError`` 4031) because ``RevokedToken.is_revoked`` ran on a scoped session left bound to a dropped connection; it now discards that session and retries once on a fresh connection, matching the recovery ``FabAuthManager.deserialize_user`` already had.
```
vincbeck: No need to have this

uplsh580: Thanks, removed it.

airflow-core/src/airflow/api_fastapi/auth/managers/base_auth_manager.py
```diff
@@ -161,15 +164,45 @@ async def get_user_from_token(self, token: str) -> BaseUser:
             log.error("JWT token is not valid: %s", e)
             raise e
 
-        if (jti := payload.get("jti")) and RevokedToken.is_revoked(jti):
-            raise InvalidTokenError("Token has been revoked")
+        if jti := payload.get("jti"):
+            if self._is_token_revoked(jti):
```
vincbeck:
```suggestion
        if jti := payload.get("jti") and self._is_token_revoked(jti):
```

uplsh580: Thanks, applied

vincbeck (approved):

PR #71446, opened 2026-08-11
UI: Show owning team in asset dependency popovers

vincbeck:
Tracing data lineage across a multi-team deployment means jumping between assets and the Dags and tasks that produce or schedule them, without any signal of who owns each one. Surfacing the owning team directly in the "Scheduled Dags" and "Producing Tasks" popovers on the assets list makes ownership obvious while following those links, so operators know who to reach without opening each Dag.

The team is resolved in a single batched lookup and only when multi-team support is enabled, so single-team deployments issue no extra query and see no change.

### Dags

Screenshot 2026-08-11 at 12 23 05 PM

### Tasks

Screenshot 2026-08-11 at 12 23 13 PM

 

---

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
+++ b/airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
@@ -29,6 +29,7 @@ from airflow._shared.timezones import timezone
 from airflow.api_fastapi.app import get_auth_manager
 from airflow.api_fastapi.auth.managers.models.resource_details import DagAccessEntity, DagDetails
 from airflow.api_fastapi.common.dagbag import DagBagDep, get_latest_version_of_dag
+from airflow.api_fastapi.common.db.assets import eager_load_asset_reference_teams
 from airflow.api_fastapi.common.db.common import SessionDep, paginated_select
 from airflow.api_fastapi.common.parameters import (
     BaseParam,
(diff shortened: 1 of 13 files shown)


seanghaeli:
In the same way there's a `test_query_count_with_multi_team` test for the `routes/ui/test_assets.py` path, add an analogous test for the `routes/public/test_assets.py` path. 

On the public path, `test_assets_references_include_team_name` checks that the team_name resolves correct but it seems that there isn't anything that tests that the query count is pinned. So there could still be an N+1 hiding in here that no test would surface.

vincbeck:
> In the same way there's a `test_query_count_with_multi_team` test for the `routes/ui/test_assets.py` path, add an analogous test for the `routes/public/test_assets.py` path.
> 
> On the public path, `test_assets_references_include_team_name` checks that the team_name resolves correct but it seems that there isn't anything that tests that the query count is pinned. So there could still be an N+1 hiding in here that no test would surface.

Good call, fixed

bbovenzi:
I guess we should make all of these popovers more detailed. I'm sure users will want much more than just the dag name and team name

vincbeck:
> I guess we should make all of these popovers more detailed. I'm sure users will want much more than just the dag name and team name

What are you thinking in terms of information you want to display?

vincbeck:
> > I guess we should make all of these popovers more detailed. I'm sure users will want much more than just the dag name and team name
> 
> What are you thinking in terms of information you want to display?

@bbovenzi How to you want to tackle it?
- Close this PR and you/someone will include this change as part of another bigger change to surface more information in the popover. I am just afraid of losing this change if it never happens (or in a very long time)
- Get this PR merge and we can iterate on it later to surface more information

seanghaeli:
My vote is to merge this one and iterate on it in the future if more information is desired

seanghaeli (approved):

airflow-core/src/airflow/api_fastapi/core_api/routes/public/assets.py
```diff
@@ -587,8 +587,7 @@ def get_asset(
         select(AssetModel)
         .where(AssetModel.id == asset_id)
         .options(
-            joinedload(AssetModel.scheduled_dags),
-            joinedload(AssetModel.producing_tasks),
+            *eager_load_asset_reference_teams(),
```
bbovenzi, line 593: Let's just add an `assert_queries_count` test here to make sure this doesn't blow up unexpectedly

vincbeck: Good call :) Done

bbovenzi (approved):

PR #71449, opened 2026-08-11
Fix incorrect availability versions for OpenLineage config options

ramitkataria:
Each entry was recorded against a release that had already been published by the time the change merged, so the version users were pointed at never contained the feature. That left people on those releases told a config was available when it was not, with no reliable way to tell which provider version they actually needed to upgrade to.

 

---

- [X] Yes (Claude Opus 5)

---

### DIFF
--- a/providers/openlineage/docs/emission_policy.rst
+++ b/providers/openlineage/docs/emission_policy.rst
@@ -285,7 +285,7 @@ for the full list of supported flags and examples.
 Legacy configuration options (deprecated)
 -----------------------------------------
 
-.. deprecated:: 2.18.0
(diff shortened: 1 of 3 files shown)


seanghaeli (approved):

o-nikolas (commented):
Should we not update this spot as well?:
https://github.com/apache/airflow/blob/4d3a247fc4102a0957b1adc52b3d942dc081a165/providers/openlineage/docs/emission_policy.rst?plain=1#L285-L293

ramitkataria:
Thanks for the catch! I've also added a version fix for another config option

mobuchowski (approved):

PR #71459, opened 2026-08-11
Fix DagRun.dag_versions after partial run_on_latest_version clear

### LINKED ISSUE #71454: DagRun.dag_versions silently drops versions after a partial "run on latest version" clear
## Problem

`DagRun.dag_versions` (`airflow-core/src/airflow/models/dagrun.py:515-530`) has a shortcut for bundle-versioned (pinned) runs:

```python
@property
def dag_versions(self) -> list[DagVersion]:
    """Return the DAG versions associated with the TIs of this DagRun."""
    # when the dag is in a versioned bundle, we keep the dag version fixed
    if self.bundle_version:
        return [self.created_dag_version] if self.created_dag_version is not None else []
    ...
```

This assumes a pinned run's task instances can only ever be on the single version recorded in `created_dag_version_id`. That was true when `created_dag_version_id` was immutable (set once at DagRun creation, per #49097). It is no longer true: `clear_task_instances(..., run_on_latest_version=True)` (added in #54984, extended in #65835/#66901) bumps `created_dag_version_id` to the latest version, but only bumps the `dag_version_id` of the *cleared* task instances — uncleared task instances explicitly keep their old version (see the comment at `airflow-core/src/airflow/models/taskinstance.py:477`: "Only cleared TIs get latest dag_version_id above; do not rewrite others.").

## Repro

1. Create a pinned (bundle-versioned) DagRun with tasks A and B, both created against version V1.
2. Deploy a new version V2 (no task additions/removals needed).
3. Clear only task A with `run_on_latest_version=True`. Now: `dag_run.created_dag_version_id == V2`, task A's TI is on V2, task B's TI is still on V1.
4. Read `dag_run.dag_versions` (or view the Grid UI, which reads it directly at `airflow-core/src/airflow/api_fastapi/core_api/routes/ui/grid.py:353`).

## Expected

`[V1, V2]` — the run genuinely has task instances on both versions.

## Actual

`[V2]` — task B's actual version (V1) is silently dropped, because the property trusts `created_dag_version_id` alone instead of the real per-TI versions once bundle-version pinning is set.

## Impact

The Grid UI's "Dag versions" indicator for a run can show a single version chip while some of that run's tasks are still genuinely executing under an older version — misleading when debugging "what code actually ran here."

## Suggested fix

Don't special-case bundle-versioned runs in `dag_versions`; compute it the same way as unpinned runs (union of `_ti_dag_versions`/`_tih_dag_versions`), or explicitly detect and represent the mixed-version case.

## Context

This is one concrete consequence of `created_dag_version_id`'s contract having drifted from "the version recorded at DagRun creation" (its documented meaning) to "the version this run should currently run at" (its behavior since #54984). See the companion issue tracking that broader contract problem: #71453

---
Drafted-by: Claude Code (Sonnet 5) (no human review before posting)

dgvj-work:
Fix `DagRun.dag_versions` silently dropping older versions still in use after a `run_on_latest_version` clear.

For bundle-versioned runs the property previously returned only `created_dag_version`. Clearing with `run_on_latest_version=True` bumps that pointer to latest while uncleared TaskInstances or TaskInstanceHistory keep the older `dag_version_id`, so the Grid UI could show a single version chip while tasks were still executing under mixed versions.

### Changes
- Always derive `DagRun.dag_versions` from the run's TaskInstance / TaskInstanceHistory rows (same path as unpinned runs).
- Update `attach_dag_versions_to_runs` to prefetch for bundle-versioned runs as well, so Grid/UI keeps the lightweight DISTINCT query path and still sees mixed versions.

closes: #71454

## Testing

- `uv run --project airflow-core pytest airflow-core/tests/unit/models/test_dagrun.py::TestDagRun::test_dag_run_dag_versions_method airflow-core/tests/unit/models/test_dagrun.py::TestDagRun::test_dag_run_dag_versions_with_null_created_dag_version airflow-core/tests/unit/models/test_dagrun.py::TestDagRun::test_bundled_dag_run_dag_versions_after_partial_run_on_latest_version airflow-core/tests/unit/models/test_dagrun.py::TestDagRun::test_bundled_dag_run_dag_versions_after_full_run_on_latest_version -q`

---

---

### DIFF
--- a/airflow-core/src/airflow/models/dagrun.py
+++ b/airflow-core/src/airflow/models/dagrun.py
@@ -532,10 +532,9 @@ class DagRun(Base, LoggingMixin):
     @property
     def dag_versions(self) -> list[DagVersion]:
         """Return the DAG versions associated with the TIs of this DagRun."""
-        # when the dag is in a versioned bundle, we keep the dag version fixed
-        if self.bundle_version:
-            return [self.created_dag_version] if self.created_dag_version is not None else []
-
+        # Always derive from actual TI/TIH versions. Do not special-case bundle-versioned
+        # runs via created_dag_version_id: clear(..., run_on_latest_version=True) can bump
+        # that pointer while leaving uncleared TIs on older versions (apache/airflow#71454).
         if self._prefetched_dag_version_ids is not None:
             dag_version_objects = list(self._prefetched_dag_version_ids.values())
         else:
--- a/airflow-core/src/airflow/models/dagrun.py
+++ b/airflow-core/src/airflow/models/dagrun.py
@@ -532,9 +532,7 @@ class DagRun(Base, LoggingMixin):
     @property
     def dag_versions(self) -> list[DagVersion]:
         """Return the DAG versions associated with the TIs of this DagRun."""
-        # Always derive from actual TI/TIH versions. Do not special-case bundle-versioned
-        # runs via created_dag_version_id: clear(..., run_on_latest_version=True) can bump
-        # that pointer while leaving uncleared TIs on older versions (apache/airflow#71454).
+        # Always derive from actual TI/TIH versions.
         if self._prefetched_dag_version_ids is not None:
             dag_version_objects = list(self._prefetched_dag_version_ids.values())
         else:
--- a/airflow-core/tests/unit/models/test_dagrun.py
+++ b/airflow-core/tests/unit/models/test_dagrun.py
@@ -1363,7 +1363,7 @@ class TestDagRun:
         assert dag_run.version_number == dag_v.version_number
 
     def test_dag_run_dag_versions_with_null_created_dag_version(self, dag_maker, session):
-        """Test that dag_versions returns empty list when created_dag_version is None and bundle_version is populated."""
+        """bundle_version alone must not hide TI versions when created_dag_version is None."""
         with dag_maker(
             "test_dag_run_null_created_dag_version",
             schedule=datetime.timedelta(days=1),
@@ -1372,16 +1372,71 @@ class TestDagRun:
             EmptyOperator(task_id="empty")
         dag_run = dag_maker.create_dagrun()
 
+        ti_version_ids = {ti.dag_version_id for ti in dag_run.task_instances if ti.dag_version_id is not None}
+        assert ti_version_ids
+
         dag_run.bundle_version = "some_bundle_version"
         dag_run.created_dag_version_id = None
         dag_run.created_dag_version = None
         session.merge(dag_run)
         session.flush()
 
-        # This should return empty list, not [None]
-        assert dag_run.dag_versions == []
-        assert isinstance(dag_run.dag_versions, list)
-        assert len(dag_run.dag_versions) == 0
+        # Derive from TI versions; never return [None] from a null created_dag_version shortcut.
+        versions = dag_run.dag_versions
+        assert isinstance(versions, list)
+        assert {dv.id for dv in versions} == ti_version_ids
+        assert None not in versions
+
+    def test_dag_run_dag_versions_after_partial_run_on_latest_version(self, dag_maker, session):
+        """dag_versions must include uncleared TI versions after a partial run_on_latest_version clear.
+
+        Repro for apache/airflow#71454: clearing a subset with run_on_latest_version=True bumps
+        created_dag_version_id / bundle_version to latest while uncleared TIs stay on the old
+        version. The property must report both.
+        """
+        with dag_maker(
+            "test_dag_run_dag_versions_partial_latest",
+            schedule=datetime.timedelta(days=1),
+            start_date=DEFAULT_DATE,
+            bundle_version="v1",
+        ):
+            EmptyOperator(task_id="0")
+            EmptyOperator(task_id="1")
+        dag_run = dag_maker.create_dagrun(state=State.RUNNING, run_type=DagRunType.SCHEDULED)
+
+        old_dag_version = DagVersion.get_latest_version(dag_run.dag_id)
+        ti0, ti1 = sorted(dag_run.task_instances, key=lambda ti: ti.task_id)
+        for ti in (ti0, ti1):
+            ti.state = TaskInstanceState.SUCCESS
+            session.merge(ti)
+        dag_run.state = DagRunState.SUCCESS
+        session.merge(dag_run)
+        session.flush()
+
+        with dag_maker(
+            "test_dag_run_dag_versions_partial_latest",
+            schedule=datetime.timedelta(days=1),
+            start_date=DEFAULT_DATE,
+            bundle_version="v2",
+        ):
+            EmptyOperator(task_id="0")
+            EmptyOperator(task_id="1")
+        new_dag_version = DagVersion.get_latest_version(dag_run.dag_id)
+        assert old_dag_version.id != new_dag_version.id
+
+        clear_task_instances([ti0], session, run_on_latest_version=True)
+        session.commit()
+
+        dag_run = session.scalar(select(DagRun).where(DagRun.run_id == dag_run.run_id))
+        assert dag_run.created_dag_version_id == new_dag_version.id
+        assert dag_run.bundle_version == new_dag_version.bundle_version
+
+        tis = {ti.task_id: ti for ti in dag_run.task_instances}
+        assert tis["0"].dag_version_id == new_dag_version.id
+        assert tis["1"].dag_version_id == old_dag_version.id
+
+        version_ids = {dv.id for dv in dag_run.dag_versions}
+        assert version_ids == {old_dag_version.id, new_dag_version.id}
 
     @pytest.mark.parametrize(
         "interval",
--- a/airflow-core/tests/unit/models/test_dagrun.py
+++ b/airflow-core/tests/unit/models/test_dagrun.py
@@ -50,6 +50,7 @@ from airflow._shared.observability.traces import (
     OverrideableRandomIdGenerator,
 )
 from airflow._shared.timezones import timezone
+from airflow.api_fastapi.common.db.dag_runs import attach_dag_versions_to_runs
 from airflow.callbacks.callback_requests import DagCallbackRequest, DagRunContext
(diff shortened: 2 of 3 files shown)


pierrejeambrun (commented):
~~We also have https://github.com/apache/airflow/pull/71425 to address the same issue.~~
Edit: #71425 isn't solving that issue.

@ephraimbuddy @dstandish wdyt?

This doesn't solve the contract problem mentioned in the other PR but can help sove the immediate reported issues with minimal effort. While we figure out the interface we want.

airflow-core/tests/unit/models/test_dagrun.py
```diff
@@ -1372,16 +1373,78 @@ def test_dag_run_dag_versions_with_null_created_dag_version(self, dag_maker, ses
...
+        assert isinstance(versions, list)
+        assert {dv.id for dv in versions} == ti_version_ids
+        assert None not in versions
+
+    def test_dag_run_dag_versions_after_partial_run_on_latest_version(self, dag_maker, session):
+        """dag_versions must include uncleared TI versions after a partial run_on_latest_version clear.
+
+        Repro for apache/airflow#71454: clearing a subset with run_on_latest_version=True bumps
+        created_dag_version_id / bundle_version to latest while uncleared TIs stay on the old
+        version. The property must report both.
```
pierrejeambrun:
```suggestion

```

airflow-core/tests/unit/models/test_dagrun.py
```diff
@@ -1372,16 +1373,78 @@ def test_dag_run_dag_versions_with_null_created_dag_version(self, dag_maker, ses
...
-        assert dag_run.dag_versions == []
-        assert isinstance(dag_run.dag_versions, list)
-        assert len(dag_run.dag_versions) == 0
+        # Derive from TI versions; never return [None] from a null created_dag_version shortcut.
+        versions = dag_run.dag_versions
+        assert isinstance(versions, list)
+        assert {dv.id for dv in versions} == ti_version_ids
+        assert None not in versions
+
+    def test_dag_run_dag_versions_after_partial_run_on_latest_version(self, dag_maker, session):
```
pierrejeambrun: Metion bundled version in the test name and docstring I guess

airflow-core/src/airflow/models/dagrun.py
```diff
@@ -532,10 +532,9 @@ def dag_run_data(self) -> DRDataModel:
     @property
     def dag_versions(self) -> list[DagVersion]:
         """Return the DAG versions associated with the TIs of this DagRun."""
-        # when the dag is in a versioned bundle, we keep the dag version fixed
-        if self.bundle_version:
-            return [self.created_dag_version] if self.created_dag_version is not None else []
-
+        # Always derive from actual TI/TIH versions. Do not special-case bundle-versioned
+        # runs via created_dag_version_id: clear(..., run_on_latest_version=True) can bump
+        # that pointer while leaving uncleared TIs on older versions (apache/airflow#71454).
```
pierrejeambrun:
```suggestion
        # Always derive from actual TI/TIH versions.
```

airflow-core/tests/unit/models/test_dagrun.py
pierrejeambrun: We do not specify issues number in tests.

pierrejeambrun (approved):
A few nits, looking good to me.

airflow-core/src/airflow/models/dagrun.py
dgvj-work: Shortened as suggested.

airflow-core/tests/unit/models/test_dagrun.py
dgvj-work: Renamed the test and updated the docstring to call out the bundled DAG case.

airflow-core/tests/unit/models/test_dagrun.py
dgvj-work: Removed the issue number from the test.

dgvj-work: Removed that note as well.

airflow-core/tests/unit/models/test_dagrun.py
```diff
@@ -1372,16 +1373,78 @@ def test_dag_run_dag_versions_with_null_created_dag_version(self, dag_maker, ses
...
-        assert dag_run.dag_versions == []
-        assert isinstance(dag_run.dag_versions, list)
-        assert len(dag_run.dag_versions) == 0
+        # Derive from TI versions; never return [None] from a null created_dag_version shortcut.
+        versions = dag_run.dag_versions
+        assert isinstance(versions, list)
+        assert {dv.id for dv in versions} == ti_version_ids
+        assert None not in versions
+
+    def test_bundled_dag_run_dag_versions_after_partial_run_on_latest_version(self, dag_maker, session):
```
ephraimbuddy, line 1392: This pins the partial clear, but the PR changes the full clear too and nothing
covers that path.

dgvj-work: Added `test_bundled_dag_run_dag_versions_after_full_run_on_latest_version`. After a full clear every current TI is on latest, so the older version only remains on TIH; `dag_versions` and the prefetch path now assert both.

dgvj-work:
Added coverage for a full `run_on_latest_version` clear as well.

After every TI is cleared, current tasks are all on latest and the older version only remains on TIH. `dag_versions` and `attach_dag_versions_to_runs` now have a test for that path, next to the existing partial-clear case.

@ephraimbuddy this should be ready for another look.

dgvj-work:
@ephraimbuddy could you please take another look at this ?

airflow-core/src/airflow/models/dagrun.py
```diff
@@ -532,10 +532,7 @@ def dag_run_data(self) -> DRDataModel:
     @property
     def dag_versions(self) -> list[DagVersion]:
         """Return the DAG versions associated with the TIs of this DagRun."""
-        # when the dag is in a versioned bundle, we keep the dag version fixed
-        if self.bundle_version:
-            return [self.created_dag_version] if self.created_dag_version is not None else []
-
+        # Always derive from actual TI/TIH versions.
```
ephraimbuddy, line 535: Single-run requests now load every TI and history row. Removing the shortcut makes pinned runs traverse both ORM collections when prefetching hasn’t happened. `get_dag_run()` doesn’t call the helper, so opening a large mapped run now materializes thousands of rows just to return its versions. Use the DISTINCT prefetch path there too.

dgvj-work: `get_dag_run()` now uses the same path as list/grid: `eager_load_dag_run_for_list()` plus `attach_dag_versions_to_runs`. Opening a single run no longer walks the TI/TIH collections just to return versions. Added a query-count test that stays flat from 3 to 20 task instances.

dgvj-work:
`get_dag_run()` now prefetches dag versions with the DISTINCT helper, same as the list/grid path. A large mapped run no longer materializes every TI/TIH row just to serialize versions.

@ephraimbuddy this should be ready for another look.

airflow-core/tests/unit/api_fastapi/core_api/routes/public/test_dag_run.py
```diff
@@ -438,6 +438,31 @@ def test_get_dag_run_not_found(self, test_client):
...
+            session.commit()
+
+            with count_queries() as result:
+                response = test_client.get(f"/dags/{dag_id}/dagRuns/{dag_run.run_id}")
+
+            assert response.status_code == 200
+            assert response.json()["dag_versions"]
+            query_counts.append(sum(result.values()))
+
+        assert query_counts[0] == query_counts[1], (
```
ephraimbuddy, line 461: This test still passes with the entire get_dag_run() change reverted. Query count stays flat even when each query loads every TI/TIH row, so it doesn’t catch the reported regression. Please assert that response serialization leaves those collections unloaded, including for a bundle-versioned run with history.

dgvj-work: Replaced the query-count check. It now serializes a bundle-versioned run that still has TIH after a full `run_on_latest_version` clear and asserts `task_instances` and `task_instances_histories` stay unloaded. That fails if `get_dag_run()` stops calling the DISTINCT prefetch.

dgvj-work:
The GET test now asserts serialization leaves TI/TIH collections unloaded on a bundled run with history, so it fails if the prefetch is dropped.

@ephraimbuddy this should be ready for another look.

airflow-core/src/airflow/models/dagrun.py
```diff
@@ -532,10 +532,7 @@ def dag_run_data(self) -> DRDataModel:
     @property
     def dag_versions(self) -> list[DagVersion]:
         """Return the DAG versions associated with the TIs of this DagRun."""
-        # when the dag is in a versioned bundle, we keep the dag version fixed
```
ephraimbuddy, line 535: Removing this shortcut also makes patch_dag_run() load every TI/TIH when serializing its response, even for a note-only update. Please use the DISTINCT prefetch there too, after the mutations, and add a test asserting that serialization leaves both collections unloaded.

airflow-core/tests/unit/models/test_dagrun.py
```diff
@@ -1388,16 +1374,142 @@ def test_dag_run_dag_versions_with_null_created_dag_version(self, dag_maker, ses
...
+        ):
+            EmptyOperator(task_id="0")
+            EmptyOperator(task_id="1")
+        new_dag_version = DagVersion.get_latest_version(dag_run.dag_id)
+        assert old_dag_version.id != new_dag_version.id
+
+        clear_task_instances([ti0], session, run_on_latest_version=True)
+        session.commit()
+
+        dag_run = session.scalar(select(DagRun).where(DagRun.run_id == dag_run.run_id))
```
ephraimbuddy, line 1432: Please include dag_id in this query and the equivalent query at line 1489, or reload by the primary key. run_id is only unique within a Dag.

PR #71461, opened 2026-08-11
Fix misleading error when an Anthropic agent session stops on its budget

kaxil:
## Problem

Anthropic SDK 0.121.0 adds Managed Agents session budgets, and with them a new
`budget_reached` idle stop reason. The provider funnelled every non-`end_turn` idle reason
into one message, so a budget stop reported:

> Session ... is idle but did not complete (budget_reached); configure an autonomous agent
> or use an outcome run.

That advice has nothing to do with what happened. It also raised the generic
`AnthropicAgentSessionError`, leaving Dag authors no way to tell a spend decision apart
from a fault.

## Solution

`budget_reached` gets its own branch and a new `AnthropicSessionBudgetExceeded`, a subclass
of the existing session error so current handlers keep working. The message names **both**
causes, because a session also stops this way when its usage includes a model with no list
price that a budget cannot measure -- and raising the ceiling does not unblock that case.

`poll_session_completion` returns the SDK's `stop_reason` alongside the message, and the
trigger carries it in its error event, so the deferrable path raises the same class as the
synchronous one without matching on message text.

## Verified against the live API

Not only against mocks: `sessions.create` accepts the documented budget payload, the server
really emits `stop_reason = 'budget_reached'`, and the hook classifies a genuinely
budget-stopped session and raises the new exception.

## Gotchas

**A budget is a stop trigger, not a spend cap.** The ceiling is checked *between* model
requests, so a request already in flight runs to completion. Measured against the live API,
a `"1"` ($0.01) ceiling admitted between $0.32 and $0.61 of usage across four runs. The docs
now say so, because the SDK's own phrase "a hard spend ceiling" invites the wrong
expectation.

**Dependency floor.** `anthropic` moves to `>=0.121.0`, the first release shipping
`sessions.create(budget=)` and the `budget_reached` stop reason. A per-package
`exclude-newer` override follows the existing `pydantic-ai-skills` precedent, since 0.121.0
was published inside the rolling 4-day window; it also keeps later releases quarantined.
Removable once the window has clearly advanced past 2026-08-07.

**Back-compat.** `poll_session_completion` changes from a 2-tuple to a 3-tuple. The provider
is 0.x and every in-repo caller is updated. Rolling upgrades are safe both directions: a new
trigger adds a key the old handler ignores, and a trigger serialized before `stop_reason`
existed omits it, so the operator falls back to the generic error.

### DIFF
--- a/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
@@ -20,7 +20,7 @@ import logging
 import time
 from enum import Enum
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 from anthropic import (
     Anthropic,
@@ -38,6 +38,7 @@ from airflow.providers.anthropic.exceptions import (
     AnthropicBatchJobError,
     AnthropicBatchTimeout,
     AnthropicError,
+    AnthropicSessionBudgetExceeded,
     AnthropicTriggerEventError,
 )
 from airflow.providers.common.compat.sdk import AirflowSkipException, BaseHook
@@ -118,6 +119,35 @@ class SessionStatus(str, Enum):
 
 #: ``outcome_evaluations[].result`` values that mean the outcome did NOT succeed.
 OUTCOME_FAILURE_RESULTS = frozenset({"failed", "max_iterations_reached", "interrupted"})
+
+# ``session.status_idle`` stop reason emitted when a session stops against its budget.
+BUDGET_REACHED = "budget_reached"
+
+
+def _create_session_error(message: str, stop_reason: str | None) -> AnthropicAgentSessionError:
+    """
+    Return the session error class matching an idle ``stop_reason``.
+
+    Keyed on the SDK's own ``stop_reason`` value rather than on the message text, so the
+    synchronous path and the deferrable path (which only carries the reason as a string
+    through the trigger event) raise the same type for the same cause.
+    """
+    if stop_reason == BUDGET_REACHED:
+        return AnthropicSessionBudgetExceeded(message)
+    return AnthropicAgentSessionError(message)
+
+
+class SessionPollResult(NamedTuple):
+    """
+    Verdict from one poll of a session; see :meth:`AnthropicHook.poll_session_completion`.
+
+    Named rather than a bare tuple because ``error_message`` and ``stop_reason`` are both
+    ``str | None``, so transposing them at a call site would still type-check.
+    """
+
+    done: bool
+    error_message: str | None
+    stop_reason: str | None
 
 
 def evaluate_session_state(
@@ -587,13 +617,24 @@ class AnthropicHook(BaseHook):
 
     def poll_session_completion(
         self, session_id: str, *, expect_outcome: bool = False, kickoff_event_id: str | None = None
-    ) -> tuple[bool, str | None]:
+    ) -> SessionPollResult:
         """
-        Return ``(done, error_message)`` for one poll of a session.
+        Return the :class:`SessionPollResult` for one poll of a session.
 
         Combines the session object (status / outcome verdict) with the event log
         (``stop_reason`` of the latest idle) so a ``message`` run distinguishes genuine
-        ``end_turn`` completion from ``requires_action`` / ``retries_exhausted``.
+        ``end_turn`` completion from ``requires_action`` / ``retries_exhausted`` /
+        ``budget_reached``.
+
+        ``stop_reason`` is the SDK's own idle stop reason, or ``None`` when the verdict did
+        not come from an idle event (a ``terminated`` session, or an outcome verdict). It
+        exists so callers can pick an error class without matching on the message text; pass
+        it to :func:`_create_session_error`.
+
+        .. note::
+            A budget stop is classified on ``message`` runs only. An ``outcome`` run is
+            judged from ``outcome_evaluations`` before the event log is consulted, so a
+            budget stop there surfaces as whatever verdict the outcome recorded.
         """
         session = self.get_session(session_id)
         done, error_message, needs_event_check = evaluate_session_state(
@@ -607,15 +648,34 @@ class AnthropicHook(BaseHook):
             needs_event_check,
         )
         if not needs_event_check:
-            return done, error_message
+            return SessionPollResult(done=done, error_message=error_message, stop_reason=None)
         reason = self._latest_idle_reason(session_id, kickoff_event_id)
         if reason is None:
-            return False, None
+            return SessionPollResult(done=False, error_message=None, stop_reason=None)
         if reason == "end_turn":
-            return True, None
-        return True, (
-            f"Session {session_id} is idle but did not complete ({reason}); "
-            "configure an autonomous agent or use an outcome run."
+            return SessionPollResult(done=True, error_message=None, stop_reason=reason)
+        if reason == BUDGET_REACHED:
+            # Both causes are worth naming: a session also stops with ``budget_reached``
+            # when its usage includes a model with no list price, because the budget cannot
+            # measure that spend -- and then raising the ceiling does not unblock it.
+            return SessionPollResult(
+                done=True,
+                error_message=(
+                    f"Session {session_id} stopped against its budget: the tracked list cost "
+                    "reached the configured ceiling, or its usage included a model with no "
+                    "list price (which a budget cannot measure). The operator archives the "
+                    "session on this path, so it cannot be resumed -- raise the ceiling for "
+                    "the next run, or drop the budget if a model has no list price."
+                ),
+                stop_reason=reason,
+            )
+        return SessionPollResult(
+            done=True,
+            error_message=(
+                f"Session {session_id} is idle but did not complete ({reason}); "
+                "configure an autonomous agent or use an outcome run."
+            ),
+            stop_reason=reason,
         )
 
     def wait_for_session(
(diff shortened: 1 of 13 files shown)


providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -119,6 +120,22 @@ def is_terminal(cls, status: str) -> bool:
 #: ``outcome_evaluations[].result`` values that mean the outcome did NOT succeed.
 OUTCOME_FAILURE_RESULTS = frozenset({"failed", "max_iterations_reached", "interrupted"})
 
+#: ``session.status_idle`` stop reason emitted when a session stops against its budget.
+BUDGET_REACHED = "budget_reached"
+
+
+def session_error(message: str, stop_reason: str | None) -> AnthropicAgentSessionError:
```
```suggestion
def _create_session_error(message: str, stop_reason: str | None) -> AnthropicAgentSessionError:
```

kaxil: Applied, along with the five call sites and the test class (`TestSessionError` -> `TestCreateSessionError`). The verb prefix is clearer -- the old name read as though it *was* an error rather than something that builds one.

One note in case it changes your view: it is imported by `operators/agent.py`, so the leading `_` marks it private to the provider package rather than to the module. There is precedent for that in providers, so I took your version as written -- happy to drop the `_` if you would rather it sit alongside `validate_execute_complete_event` and `evaluate_session_state` in the same file, which are public and cross-imported the same way.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -119,6 +120,22 @@ def is_terminal(cls, status: str) -> bool:
 #: ``outcome_evaluations[].result`` values that mean the outcome did NOT succeed.
 OUTCOME_FAILURE_RESULTS = frozenset({"failed", "max_iterations_reached", "interrupted"})
 
+#: ``session.status_idle`` stop reason emitted when a session stops against its budget.
```
Lee-W: https://github.com/apache/airflow/pull/71462#discussion_r3765824885

We can also change it here. (not including the existing ones)

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -587,13 +604,24 @@ def _latest_idle_reason(self, session_id: str, kickoff_event_id: str | None) ->
 
     def poll_session_completion(
         self, session_id: str, *, expect_outcome: bool = False, kickoff_event_id: str | None = None
-    ) -> tuple[bool, str | None]:
+    ) -> tuple[bool, str | None, str | None]:
```
Lee-W: Could we make the return type a NamedTuple? This seems easy to get wrong, especially when the second and third values have the same type.

Lee-W (commented):
I think the tuple is my last major comment. everything else looks good

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -636,12 +696,13 @@ def wait_for_session(
...
         :param poll_interval: Seconds to sleep between polls.
         :param timeout: Maximum seconds to wait before raising :class:`AnthropicAgentSessionTimeout`.
+        :raises AnthropicSessionBudgetExceeded: If the session stopped against its budget.
         """
         start = time.monotonic()
         consecutive_failures = 0
         while True:
             try:
-                done, error_message = self.poll_session_completion(
+                poll = self.poll_session_completion(
```
```suggestion
                poll_result = self.poll_session_completion(
```

providers/anthropic/src/airflow/providers/anthropic/triggers/agent.py
```diff
@@ -94,7 +94,7 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
         while True:
             try:
                 # poll_session_completion does blocking SDK HTTP calls; run off the event loop.
-                done, error_message = await asyncio.to_thread(
+                poll = await asyncio.to_thread(
```
Lee-W:
```suggestion
                poll_result = await asyncio.to_thread(
```

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -156,39 +159,77 @@ class TestPollSessionCompletion:
     def test_terminated_is_error(self):
         hook, client = _make_hook()
         client.beta.sessions.retrieve.return_value = _session("terminated")
-        done, err = hook.poll_session_completion("s")
-        assert done is True
-        assert err is not None
+        poll = hook.poll_session_completion("s")
```
```suggestion
        poll_result = hook.poll_session_completion(session_id="s")
```

or let's try something like

```python
    session_id = "s"

   ...

        poll_result = hook.poll_session_completion(self.session_id)
```

it would be easier to understand

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -156,39 +159,77 @@ class TestPollSessionCompletion:
...
+        # is wrong, and the no-list-price cause is invisible without being named.
+        hook, client = _make_hook()
+        client.beta.sessions.retrieve.return_value = _session("idle")
+        client.beta.sessions.events.list.return_value = [_idle_event("budget_reached")]
+        poll = hook.poll_session_completion("s", kickoff_event_id="evt_kick")
+        assert poll.done is True
+        assert poll.stop_reason == "budget_reached"
+        assert "budget" in poll.error_message
+        assert "no list price" in poll.error_message
+        assert "autonomous agent" not in poll.error_message
```
```suggestion
        assert poll == SessionPollResult(
            done=True, error_message=ANY, stop_reason="budget_reached"
        )
        
        for msg_keyword in ["budget", "no list price", "autonomous agent"]:
            assert msg_keyword in poll.error_message        
```

Lee-W (approved):
some minor suggestion but nothing really blocking

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Done -- plain `#` now. This is the only `#:` the PR adds, so the pre-existing ones in the file stay untouched, as you asked.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Added `SessionPollResult(done, error_message, stop_reason)`.

One thing worth reporting, because it changed how much of the change was needed: the annotation on its own does not buy the protection. mypy accepts a transposed unpack of the old signature silently --

```python
done, stop_reason, error_message = hook.poll_session_completion(...)
```

-- since both slots are `str | None`, and it accepts the same positional unpack of a NamedTuple just as quietly.

So the guard is the NamedTuple *plus* dropping positional unpacking at the consumers. `wait_for_session` and the trigger's `run` now hold the result and read `poll.done` / `poll.error_message` / `poll.stop_reason`; that mypy does check, a wrong field name being an `attr-defined` error. The returns inside the hook use keywords for the same reason, and the tests follow, so a mocked poll result now names its fields instead of being a bare triple.

`evaluate_session_state` stays a plain tuple: its second and third elements are `str | None` and `bool`, and mypy does reject a transposed unpack there (`Argument 1 ... has incompatible type "bool"; expected "str | None"`). Happy to convert it too for uniformity, but the failure you spotted is specific to the two same-typed slots.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Renamed -- `poll_result` here, in the trigger, and at the test call sites.

providers/anthropic/src/airflow/providers/anthropic/triggers/agent.py
kaxil: Done here too.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Renamed to `poll_result`, and took your second option in spirit -- there was a real mismatch behind the confusion, not just a short name.

`_session()` builds the mocked session with `id = "sess_1"`, but this class was passing `"s"`. The two ids reach the assertions by different routes: `evaluate_session_state` interpolates `session.id`, while `poll_session_completion` interpolates the argument. So the messages under test carried `"sess_1"` on one path and `"s"` on the other, and an assertion could have matched the wrong one.

There is now a module-level `SESSION_ID = "sess_1"` used by both `_session()` and every call in this class, so they cannot drift apart.

I left the pre-existing `"sess_1"` literals elsewhere in the file as they are, to keep the diff off tests this PR does not otherwise touch -- happy to move them onto the constant too. I also skipped the explicit `session_id=` keyword, since the value now names itself; say the word if you would still rather have it spelled out.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Took the whole-object assertion with `error_message=mock.ANY`, and the loop for the causes that must be named.

One correction on the loop, though: it inverts the last assertion. The original line was

```python
assert "autonomous agent" not in poll.error_message
```

`not in`, and that negative is the point of the test -- a budget stop must *not* carry the generic "configure an autonomous agent or use an outcome run" advice, because that advice is wrong for this case and was the bug the PR fixes. Folding it into a loop of `in` checks asserts the opposite of what the test is for.

I ran your version verbatim to be sure rather than argue from reading:

```
FAILED test_budget_reached_names_both_causes - AssertionError: assert 'autonomous agent' in
'Session s stopped against its budget: the tracked list cost reached the configured ceiling,
or its usage included a model with no list price (which a budget cannot measure). ...'
```

So it is the loop over the two causes that must be present, with the negative kept as its own line and a comment saying why it is there.

I applied the same whole-object shape to `test_message_blocked_is_error` just above, since it had the same stack of single-field asserts.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
Lee-W: yep, don't need that keyword. `SESSION_ID` along is clear enough. thanks!

Lee-W (approved):

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
Lee-W: > the annotation on its own does not buy the protection.

yep, I don't think we need to go that far unless this actaully becomes a problem. I raise it because it was not easy to read back then and it's way better now.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
Lee-W: oops, didn't even notice the `in`, `not in`. now they're separate and easier to read. Thanks!

PR #71462, opened 2026-08-11
Add a budget argument to AnthropicAgentSessionOperator

kaxil:
## Problem

Session budgets were reachable only as a raw payload through `session_kwargs`:

```python
session_kwargs={
    "budget": {"type": "limit", "max_list_cost": {"amount": "2500", "currency": "USD"}}
}
```

Minor units as an integer string is the right wire format but a poor authoring surface, and
being buried in a passthrough dict meant the ceiling could not be templated.

## Solution

`budget` is now a first-class templated argument. A number or numeric string is read as
**US dollars**, so the ceiling can come from a Variable and change without editing the Dag:

```python
AnthropicAgentSessionOperator(
    task_id="research",
    agent_id="agt_...",
    environment_id="env_...",
    message="Summarise yesterday's incidents.",
    budget=25.00,
    retries=0,
)
```

A mapping is still accepted verbatim for a payload shape the provider has not caught up
with.

Conversion runs through `Decimal` built from `str()`, never binary float, so `0.07` yields
`"7"` rather than a rounding artifact. An amount finer than a cent is rejected instead of
silently rounded: quietly changing an amount of money is worse than refusing it. The
operator normalizes before creating the session, so a bad amount fails without leaving a
server-side container allocated. Setting the budget both ways at once is rejected rather
than one silently winning.

`AnthropicHook.update_session` fills a gap that predates budgets -- the hook wrapped
create, retrieve and archive but never update. It forwards only the keywords the caller
passes, because the API distinguishes *omitted* (preserve) from `None` (clear):

```python
hook.update_session(session_id, budget=50.00)  # raise the ceiling
hook.update_session(session_id, budget=None)  # remove it entirely
```

Removing it is the only escape for a session stopped by a model with no list price, which
raising the ceiling cannot unblock. The same call carries the `agent={"tools": [...]}`
replacement behind the mid-conversation-tool-changes beta.

## Gotchas

**Units differ between the two forms.** A scalar is dollars; a mapping's `amount` is minor
units, so `budget=25` and `{"amount": "25"}` are 100x apart. Both are documented at the
point of use.

**`AnthropicSessionBudgetExceeded` fires on `message` runs only.** An `outcome` run is
judged from `outcome_evaluations` before the idle event is consulted, so a budget stop there
surfaces as the outcome verdict and the generic `AnthropicAgentSessionError`.

**Retries multiply spend.** Each Airflow retry starts a new session with a fresh budget, so
`retries=2` with a $25 ceiling can spend $75. Prefer `retries=0`, or raise the budget on the
existing session rather than retrying.

### DIFF
--- a/providers/anthropic/docs/operators/anthropic.rst
+++ b/providers/anthropic/docs/operators/anthropic.rst
@@ -147,9 +147,13 @@ Parameters
 * ``poll_interval`` — seconds between session status checks.
 * ``timeout`` — seconds to wait for a terminal status; defaults to 24 hours.
 * ``vault_ids`` — vault IDs providing MCP/credential access to the session.
+* ``budget`` -- spend ceiling for the session, in US dollars (``25.00``) or as the raw API
+  payload (a mapping). Templated, so it can come from a Variable, a params entry, or
+  an upstream XCom. See `Session budgets`_ below.
 * ``session_resources`` — files, GitHub repos, or memory stores to mount (forwarded to
   ``sessions.create`` as ``resources``; renamed to avoid the reserved ``BaseOperator.resources``).
-* ``session_kwargs`` — extra keyword arguments forwarded to ``sessions.create``.
+* ``session_kwargs`` — extra keyword arguments forwarded to ``sessions.create``. Setting
+  ``budget`` here as well as via the ``budget`` argument is rejected.
 
 .. note::
 
@@ -163,10 +167,9 @@ Parameters
 Session budgets
 """""""""""""""
 
-Pass a `session budget
-<https://platform.claude.com/docs/en/managed-agents/overview>`__ through
-``session_kwargs`` to cap what a single session may spend. The session stops issuing new
-model requests once its tracked list cost reaches the ceiling:
+``budget`` bounds what a single session may spend. The session stops issuing new model
+requests once its tracked list cost reaches the ceiling. A number or numeric string is read
+as **US dollars**:
 
 .. code-block:: python
 
@@ -175,14 +178,33 @@ model requests once its tracked list cost reaches the ceiling:
         agent_id="agt_...",
         environment_id="env_...",
         message="Summarise yesterday's incidents.",
-        session_kwargs={
-            "budget": {
-                "type": "limit",
-                # Minor units as an integer string: "2500" is $25.00.
-                "max_list_cost": {"amount": "2500", "currency": "USD"},
-            }
-        },
+        budget=25.00,  # 25.00 USD
+        retries=0,
     )
+
+The field is templated, so the ceiling can come from a Variable
+(``budget="{{ var.value.max_agent_spend }}"``) and be changed without editing the Dag.
+Amounts are converted through :class:`~decimal.Decimal`, never binary float, and anything
+finer than a cent is rejected rather than silently rounded.
+
+Pass a mapping instead to send the raw API payload, for a budget shape the provider has not
+caught up with. Today the API accepts only ``type: "limit"`` and ``currency: "USD"``, so the
+mapping form is an escape hatch for future additions rather than something needed now:
+
+.. code-block:: python
+
+    budget = {"type": "limit", "max_list_cost": {"amount": "2500", "currency": "USD"}}
+
+To raise or remove a ceiling on a session that is already running, use
+:meth:`~airflow.providers.anthropic.hooks.anthropic.AnthropicHook.update_session`. Only the
+keywords you pass are sent, because the API distinguishes *omitted* (preserve) from
+``None`` (clear):
+
+.. code-block:: python
+
+    hook = AnthropicHook()
+    hook.update_session(session_id, budget=50.00)  # raise the ceiling
+    hook.update_session(session_id, budget=None)  # remove it entirely
 
 On a ``message`` run, a session that stops this way raises
 :class:`~airflow.providers.anthropic.exceptions.AnthropicSessionBudgetExceeded`, a subclass
@@ -196,6 +218,13 @@ rather than treated as a fault.
     continues until ``timeout`` (24 hours by default), and the task then fails with
     ``AnthropicAgentSessionTimeout`` -- a misleading error for a session that stopped
     deliberately. Set a shorter ``timeout`` when combining ``outcome`` with a budget.
+
+.. note::
+
+    On an ``outcome`` run, completion is judged from the session's ``outcome_evaluations``
+    before the idle event is consulted, so a budget stop is reported as whatever verdict the
+    outcome recorded and raises the generic ``AnthropicAgentSessionError``. Catch
+    ``AnthropicSessionBudgetExceeded`` only on ``message`` runs.
 
 .. warning::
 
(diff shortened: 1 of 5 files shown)


providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -147,9 +147,12 @@ Parameters
 * ``poll_interval`` — seconds between session status checks.
 * ``timeout`` — seconds to wait for a terminal status; defaults to 24 hours.
 * ``vault_ids`` — vault IDs providing MCP/credential access to the session.
+* ``budget`` -- spend ceiling for the session, in US dollars (``25.00``) or as the raw API
+  payload (a mapping). Templated. See `Session budgets`_ below.
```
Lee-W: **Templated** here is correct in Airflow context, but one might be surprised seeing only this word in the middle of other descriptions. might be better to extend the description of templated a bit

providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -175,15 +177,34 @@ model requests once its tracked list cost reaches the ceiling:
...
         environment_id="env_...",
         message="Summarise yesterday's incidents.",
-        session_kwargs={
-            "budget": {
-                "type": "limit",
-                # Minor units as an integer string: "2500" is $25.00.
-                "max_list_cost": {"amount": "2500", "currency": "USD"},
-            }
-        },
+        budget=25.00,  # $25.00
```
Lee-W: While $25,00 means USD, I feel something like `USD$ 25.00` might make this comment actually helpful.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -123,6 +126,49 @@ def is_terminal(cls, status: str) -> bool:
 #: ``session.status_idle`` stop reason emitted when a session stops against its budget.
 BUDGET_REACHED = "budget_reached"
 
+#: What a caller may pass as a session budget: an amount in USD, or the raw API payload.
```
Lee-W: Just wnat ot make sure we want to use `#:` syntax (I know there's already one above). But this is not that frequently used in airflow yet.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -123,6 +126,49 @@ def is_terminal(cls, status: str) -> bool:
...
+BudgetSpec = str | int | float | Decimal | Mapping[str, Any]
+
+
+def build_budget(budget: BudgetSpec) -> dict[str, Any]:
```
Lee-W: Should we make it return a TypedDict instead?

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -123,6 +126,49 @@ def is_terminal(cls, status: str) -> bool:
...
+        # Deep, not ``dict()``: a shallow copy leaves the nested ``max_list_cost`` aliased
+        # to the caller's object, so a later edit of the returned payload would reach back
+        # into a templated operator field.
+        return deepcopy(dict(budget))
+    if isinstance(budget, bool):
+        raise ValueError(f"Invalid budget {budget!r}: expected an amount in USD or a mapping.")
+    try:
+        dollars = Decimal(str(budget))
+    except (InvalidOperation, ValueError) as e:
+        raise ValueError(f"Invalid budget {budget!r}: not a decimal amount in USD.") from e
```
```suggestion
    try:
        dollars = Decimal(str(budget))
    except (InvalidOperation, ValueError) as e:
        raise ValueError(f"Invalid budget {budget!r}: not a decimal amount in USD.") from e
```

Should we just go with this? "True" is not a decimal amount either

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
```diff
@@ -199,6 +199,67 @@ def test_deferrable_defers_with_trigger(self, mock_hook_prop):
         hook.wait_for_session.assert_not_called()
 
 
+class TestBudgetParam:
+    @staticmethod
+    def _op(**kwargs):
```
Lee-W:
```suggestion
    def _make_op(**kwargs):
```

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Done -- and better than a local one, because the SDK already publishes this shape at the version this provider floors on.

`build_budget` now returns `BetaManagedAgentsBudgetLimitParam` (`{"type": Literal["limit"], "max_list_cost": BetaMonetaryAmountParam}`), imported under `TYPE_CHECKING`. That also gets `currency` typed as the SDK's `BetaCurrency` literal rather than a bare `str`, which a hand-rolled TypedDict here would have lost.

Worth noting for anyone checking: the type is new in `anthropic` 0.121.0, the same release that added `sessions.create(budget=...)`. On 0.120.2 neither exists, so introspecting the previous release makes it look as though the payload is untyped. 0.121.0 is this provider's declared floor, so the published type is available.

The raw-mapping branch casts rather than validates. Accepting a raw payload exists precisely so a caller can reach a field the provider does not model yet, so checking its shape there would defeat the purpose.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Fair point, and the counts back it up: `#:` appears 24 times across all of `providers/`, 0 in `airflow-core/src`, and 3 in `task-sdk/src` -- and most of the provider hits are a single Amazon hook. It is not a convention this codebase actually uses.

Switched the line this PR adds to a plain `#`. The other 13 in this file predate the PR, so I left them rather than widen the diff -- happy to sweep them in a follow-up if you would rather the file be consistent.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Applied -- you are right that it was dead code. `Decimal(str(True))` raises `InvalidOperation` (`str(True)` is `'True'`), so bools were already rejected by the parse below; the guard only changed which message they got.

One knock-on: `test_rejects_bad_amounts` had a case pinning the old `expected an amount` message for `True`, so that param now matches `not a decimal` instead. Without that the suite fails.

providers/anthropic/docs/operators/anthropic.rst
kaxil: Expanded -- it now reads "Templated, so it can come from a Variable, a params entry, or an upstream XCom." Agreed that the bare word assumes the reader already knows what it buys them.

providers/anthropic/docs/operators/anthropic.rst
kaxil: Changed to `# 25.00 USD`. Putting the currency after the number avoids the `$25,00` ambiguity you flagged without the odd `USD$` doubling.

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
kaxil: Applied, along with its five call sites in that class.

amoghrajesh:
@kaxil I wanna take a look at this one too, if you could hold it till tomorrow, that'd be great.

Lee-W (approved):

PR #71463, opened 2026-08-11
Record Anthropic agent session token usage and cost in XCom

kaxil:
## Problem

A session budget is a stop trigger, not a spend cap: the ceiling is checked between model
requests, so a request already in flight runs to completion. Measured against the live API,
a $0.01 ceiling admitted between $0.32 and $0.61 across four runs. Setting a budget
therefore tells you nothing about what a run actually cost, and the operator returned only a
session ID.

## Solution

The session's usage is pushed to XCom under `usage` on **both** success and failure, so cost
per Dag run is queryable and a budget-stopped run still records what it consumed:

```python
{
    "input_tokens": 814,
    "output_tokens": 14002,
    "cache_read_input_tokens": 0,
    "cache_creation": {"ephemeral_5m_input_tokens": 0, "ephemeral_1h_input_tokens": 0},
    "server_tool_use": {"web_search_requests": 0, "web_fetch_requests": 0},
    "active_seconds": 213.666,
    "list_cost": {"amount": "36", "currency": "USD"},
    "try_number": 1,
}
```

`amount` stays the API's minor-unit string rather than becoming a float, because a cost
figure must not pick up binary rounding. `list_cost` is `None` when usage includes a model
with no list price -- which is exactly when a caller has to price the run from usage
instead, so every billable dimension is reported: cache *writes* are billed above base
input, and server tool calls are billed per request.

Teardown runs first on the failure paths. It is the time-critical call, and
`sessions.archive` returns the session carrying its final usage, so spend is reported
without a second request against an API that may be the reason the task is failing. If
archiving fails, usage falls back to a fetch. The whole push is best effort: it runs
immediately before re-raising, so anything that throws there would otherwise replace the
exception the task should fail with.

## Verified end to end

A real deferred run, resumed by a real triggerer against postgres, failed with
`AnthropicSessionBudgetExceeded` and left the usage above in XCom, read back over the REST
API.

## Gotchas

Airflow clears a task's XCom at the start of every attempt, so `usage` holds the **final
attempt only** and `try_number` records which one that was. With retries enabled, total
spend across attempts is not recoverable from this key and must be summed from the session
records. `retries=0` keeps this and the fresh-budget-per-retry problem away together.

### DIFF
--- a/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
@@ -660,6 +660,40 @@ class AnthropicHook(BaseHook):
         self._require_first_party("Managed Agents")
         return self._first_party_conn.beta.sessions.retrieve(session_id)
 
+    def get_session_usage(self, session_id: str) -> dict[str, Any]:
+        """
+        Return a JSON-serializable token/cost summary for a session.
+
+        Plain scalars and a nested ``list_cost`` mapping rather than SDK models, so the
+        result survives XCom serialization and can be queried across runs. ``amount`` is
+        kept as the API's **minor-unit string** (``"44"`` is $0.44) rather than converted to
+        a float, so no rounding is applied to a cost figure.
+
+        Every field is optional server-side -- ``list_cost`` is absent when usage includes a
+        model with no list price -- so missing values come back as ``None``. That absence is
+        why every billable dimension is reported and not just the token totals: it is
+        exactly when a caller has to reconstruct cost from usage that the breakdown must be
+        complete. Cache *writes* (``cache_creation``) are billed above base input, and
+        server tool calls are billed per request.
+        """
+        return self.summarize_usage(self.get_session(session_id))
+
+    @staticmethod
+    def summarize_usage(session: BetaManagedAgentsSession) -> dict[str, Any]:
+        """
+        Flatten an already-retrieved session's usage; see :meth:`get_session_usage`.
+
+        Split out because ``sessions.archive`` also returns the session, so a caller that
+        is tearing a session down can report its usage without a second request.
+
+        Dumps the model rather than copying a fixed list of fields. The usage model sets
+        ``extra="allow"``, so a billable dimension added by the API is kept on the object --
+        an allowlist here would drop it silently, which is worst precisely when ``list_cost``
+        is ``None`` and a caller has to price the run from the breakdown. ``mode="json"``
+        keeps the result XCom-safe and leaves ``amount`` a minor-unit string.
+        """
+        return session.usage.model_dump(mode="json")
+
     def send_event(self, session_id: str, event: dict[str, Any]) -> Any:
         """Send a single event (e.g. a ``user.message`` or ``user.define_outcome``)."""
         self._require_first_party("Managed Agents")
@@ -668,8 +702,13 @@ class AnthropicHook(BaseHook):
             session_id, events=cast("list[BetaManagedAgentsEventParams]", [event])
         )
 
-    def archive_session(self, session_id: str) -> Any:
-        """Archive a session (frees the server-side container). Best-effort teardown."""
+    def archive_session(self, session_id: str) -> BetaManagedAgentsSession:
+        """
+        Archive a session (frees the server-side container). Best-effort teardown.
+
+        Returns the archived session, which carries its final ``usage`` -- so a caller
+        tearing a session down does not need a separate retrieve to report what it spent.
+        """
         self._require_first_party("Managed Agents")
         return self._first_party_conn.beta.sessions.archive(session_id)
 
--- a/providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
@@ -33,6 +33,8 @@ from airflow.providers.anthropic.triggers.agent import AnthropicAgentSessionTrig
 from airflow.providers.common.compat.sdk import BaseOperator, conf
 
 if TYPE_CHECKING:
+    from anthropic.types.beta import BetaManagedAgentsSession
+
     from airflow.providers.anthropic.hooks.anthropic import BudgetSpec
     from airflow.providers.common.compat.sdk import Context
 
@@ -66,6 +68,11 @@ class AnthropicAgentSessionOperator(BaseOperator):
 
     Outputs the agent writes to ``/mnt/session/outputs/`` are retrieved afterwards via the
     Files API (``scope_id=<session_id>``); the operator returns the **session ID only**.
+
+    The session's token counts and list cost are pushed to XCom under ``usage`` on both
+    success and failure, so cost per Dag run can be queried. This matters because a budget is
+    a stop trigger rather than a cap: the ceiling is checked between model requests, so it
+    does not tell you what the session actually spent.
 
     .. seealso::
         For more information, take a look at the guide:
@@ -178,7 +185,10 @@ class AnthropicAgentSessionOperator(BaseOperator):
                 )
         except Exception:
             # send_event failed after create_session allocated the container; tear it down.
-            self._archive_session(session.id)
+            # The event may still have been accepted server-side, so the agent can already be
+            # spending -- record usage here too rather than leaving this the one failure path
+            # with no cost trail.
+            self._tear_down(context, session.id)
             raise
         # Correlate completion against the kickoff event so a message run is not fooled by a
         # just-created idle session (the start race).
@@ -216,8 +226,11 @@ class AnthropicAgentSessionOperator(BaseOperator):
         except Exception:
             # Any failure after the session starts (timeout, SDK 5xx, auth expiry) leaves
             # the server-side container running; archive it best-effort before failing.
-            self._archive_session(session.id)
+            # Teardown goes first: it is the time-critical call, and its response carries
+            # the usage, so reporting spend costs no extra request.
+            self._tear_down(context, session.id)
             raise
+        self._push_usage(context, session.id)
         return session.id
 
     def execute_complete(self, context: Context, event: Any = None) -> str:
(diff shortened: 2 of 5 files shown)


providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -646,6 +646,40 @@ def get_session(self, session_id: str) -> BetaManagedAgentsSession:
         self._require_first_party("Managed Agents")
         return self._first_party_conn.beta.sessions.retrieve(session_id)
 
+    def get_session_usage(self, session_id: str) -> dict[str, Any]:
```
Lee-W, line 663: should we make it a property instead?

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
```diff
@@ -178,7 +183,11 @@ def execute(self, context: Context) -> str | None:
                 )
         except Exception:
             # send_event failed after create_session allocated the container; tear it down.
-            self._archive_session(session.id)
+            # The event may still have been accepted server-side, so the agent can already be
+            # spending -- record usage here too rather than leaving this the one failure path
+            # with no cost trail.
+            archived = self._archive_session(session.id)
+            self._push_usage(context, session.id, session=archived)
```
Lee-W: should we make this a method? it appears a few times already.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -308,6 +315,94 @@ def test_drops_a_none_budget(self):
...
+            "output_tokens",
+            "cache_read_input_tokens",
+            "cache_creation",
+            "server_tool_use",
+            "active_seconds",
+            "list_cost",
+        }
+
+    @staticmethod
+    def _session(list_cost, cache_creation=None, server_tool_use=None):
```
```suggestion
    def _create_session(list_cost, cache_creation=None, server_tool_use=None):
```

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
```diff
@@ -34,8 +34,10 @@
 pytest.importorskip("anthropic")
 
 
-def _context():
-    return {"ti": mock.MagicMock()}
+def _context(try_number=1):
```
Lee-W:
```suggestion
def _create_context(try_number=1):
```

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
```diff
@@ -260,10 +262,143 @@ def test_budget_is_templated(self):
...
+    USAGE = {
+        "input_tokens": 827,
+        "output_tokens": 17065,
+        "cache_read_input_tokens": 0,
+        "active_seconds": 91.2,
+        "list_cost": {"amount": "44", "currency": "USD"},
+    }
+
+    @staticmethod
+    def _op(**kwargs):
```
Lee-W:
```suggestion
    def _create_op(**kwargs) -> AnthropicAgentSessionOperator:
```

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -308,6 +315,94 @@ def test_drops_a_none_budget(self):
...
+        # expensive side (cache writes, server tool calls) cannot be missing.
+        hook, client = _make_hook()
+        client.beta.sessions.retrieve.return_value = self._session(
+            None,
+            cache_creation=BetaManagedAgentsCacheCreationUsage(
+                ephemeral_5m_input_tokens=120, ephemeral_1h_input_tokens=340
+            ),
+            server_tool_use=BetaManagedAgentsServerToolUsage(web_search_requests=3, web_fetch_requests=5),
+        )
+        usage = hook.get_session_usage("s")
```
Lee-W: should we just assert the whole usage dict insetad?

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
```diff
@@ -98,7 +100,7 @@ def test_message_sends_user_message_and_waits(self, mock_hook_prop):
             "sess_1", {"type": "user.message", "content": [{"type": "text", "text": "summarize"}]}
         )
         hook.wait_for_session.assert_called_once()
-        context["ti"].xcom_push.assert_called_once_with(key="session_id", value="sess_1")
+        context["ti"].xcom_push.assert_any_call(key="session_id", value="sess_1")
```
Lee-W, line 122: why do we make this change

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Not able to -- `get_session_usage(session_id)` takes the session to look up, and a property cannot take an argument. It is also a network call rather than an attribute read, so a property would hide a round trip behind attribute access.

There is a related no-argument accessor though: `summarize_usage(session)` flattens a session object you already hold, which is what the failure paths use so they do not pay a second retrieve. If the concern was the call looking heavier than it is, that one is the cheap path.

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
kaxil: Taken, but one step further, because the duplication turned out to be the real problem.

`TestBudgetParam` and `TestUsageXCom` each had a byte-identical builder -- one named `_make_op` (from your suggestion on #71462) and this one named `_op`. Renaming this one would have left two copies of the same function in the same file under two different prefixes.

So it is now a single module-level `_create_op(**kwargs) -> AnthropicAgentSessionOperator` with the signature you suggested, and both classes use it. That also settles the prefix: `_create_` matches `_create_session`, `_create_context` and `_create_session_error` from your other comments, so the file has one convention rather than two.

Flagging it because it means #71462 changed too -- the helper is defined there now.

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
kaxil: Good catch -- it was four sites, not a few. Extracted `_tear_down(context, session_id)`, which archives and then records usage; all four failure paths call it.

Keeping the order inside the helper is the point: teardown runs first because it is the time-critical call on a failing task, and `sessions.archive` returns the session carrying its final usage, so recording spend costs no extra request. Having that in one place means the ordering cannot drift between paths.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Applied. It also disambiguates it from the module-level `_session(status, ...)` helper further up the file, which is a different thing entirely.

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
kaxil: Applied, across all 30 call sites in the file.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Done -- it now asserts the whole dict. Better than my version: the per-key asserts would not have caught an unexpected extra key, and this pins the exact shape that lands in XCom.

providers/anthropic/tests/unit/anthropic/operators/test_agent.py
kaxil: Because this PR adds a second `xcom_push`. The task pushed only `session_id` before; it now pushes `usage` as well, so `assert_called_once_with` fails outright.

Swapping to `assert_any_call` was the minimum fix but it quietly weakened the test -- it would pass no matter how many other pushes happened. Replaced with an assertion on the pushed keys in order:

```python
assert [c.kwargs["key"] for c in context["ti"].xcom_push.call_args_list] == [
    "session_id",
    "usage",
]
```

That pins exactly two pushes and their order, so it is stricter than what was there before this PR rather than looser. Thanks for pushing on it -- I had accepted the weaker assertion without noticing what it stopped checking.

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
```diff
@@ -226,24 +237,83 @@ def execute_complete(self, context: Context, event: Any = None) -> str:
...
+            cost = usage.get("list_cost")
+            self.log.info(
+                "Session %s used %s input / %s output tokens; list cost %s",
+                session_id,
+                usage.get("input_tokens"),
+                usage.get("output_tokens"),
+                f"{cost['currency']} {cost['amount']} (minor units)" if cost else "unavailable",
+            )
+        except Exception as e:
+            self.log.warning("Could not record usage for session %s: %s", session_id, e)
```
Lee-W:
```suggestion
            self.log.exception("Could not record usage for session %s", session_id)
```

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
```diff
@@ -226,24 +237,83 @@ def execute_complete(self, context: Context, event: Any = None) -> str:
...
+
+    def _archive_session(self, session_id: str | None) -> Any:
```
Lee-W:
```suggestion
    def _archive_session(self, session_id: str | None) -> BetaManagedAgentsSession | None:
```

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -308,7 +315,100 @@ def test_drops_a_none_budget(self):
...
+            server_tool_use=server_tool_use,
+        )
+        return SimpleNamespace(usage=usage)
+
+    def test_flattens_to_json_safe_scalars(self):
+        hook, client = _make_hook()
+        client.beta.sessions.retrieve.return_value = self._create_session(
+            BetaMonetaryAmount(amount="44", currency="USD")
+        )
+        assert hook.get_session_usage("s") == {
```
Lee-W: should we make "s" a `session_id` variable? I was confused first seeing "s" and woundering where this "s" came from

Lee-W (approved):
LGTM. the remaining are nits.

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
kaxil: Applied -- `warning(..., e)` was throwing the traceback away, and this is the branch where you would want it.

Two things I did not change, in case you want them changed:

It also lifts the level from WARNING to ERROR. I think that is right (a failed usage push is genuinely unexpected, and the task succeeds either way), but `self.log.warning(..., exc_info=True)` keeps the traceback at the old level if you would rather it not page anyone.

The sibling line in `_archive_session` still has the same `warning("Failed to archive session %s: %s", session_id, e)` shape. I left it deliberately -- archiving fails in expected ways (a session already archived, a race with teardown), so promoting that one to ERROR felt wrong without asking. Happy to match them if you would rather they were consistent.

providers/anthropic/src/airflow/providers/anthropic/operators/agent.py
kaxil: Applied, with `BetaManagedAgentsSession` imported under `TYPE_CHECKING`. It lines up with the hook, whose `archive_session` already declares `-> BetaManagedAgentsSession`, so the `Any` was losing type information that existed one call away.

For context: the `-> Any` is not something this PR introduced -- it came in with the original provider, so this tightens existing code.

One heads-up for when you reach #71465: that PR adds `**archive_kwargs` to the same signature, so there it reads

```python
def _archive_session(
    self, session_id: str | None, **archive_kwargs: Any
) -> BetaManagedAgentsSession | None:
```

wrapped only because the one-line form runs past the line limit.

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Yes -- done, and the confusion had a real cause rather than just a terse name.

`_session()` builds the mocked session with `id = "sess_1"`, while these tests passed `"s"`. The two ids reach the assertions by different routes, so the strings under test carried whichever id the code path happened to interpolate. There is now a module-level `SESSION_ID = "sess_1"` used by both `_session()` and every call site here, so they cannot drift.

Separately, while working in this file I found this PR was renaming `TestCreateSessionError` back to `TestSessionError` -- quietly undoing the rename you asked for earlier in the stack. It was a rebase artifact: inserting `TestGetSessionUsage` directly above it made git anchor the hunk on the old class line and re-add the old name. Fixed here, and I checked the name is now consistent from `main` through to the tip of the stack rather than just in this diff.

PR #71464, opened 2026-08-11
Document Anthropic advisor rosters and pinned inference regions

kaxil:
## Summary

Anthropic SDK 0.121.0 adds two agent-level settings: an `advisor` entry for a coordinator's
multiagent roster, and `inference_geo` on the model config for pinning which region serves
inference. Both reach the API through `AnthropicHook.create_agent`'s existing keyword
passthrough, so they need no provider support -- only documentation, since nothing in the
guide says they are available.

Pinning the inference region, which is what a data-residency requirement usually comes down
to:

```python
hook.create_agent(
    name="eu-only-analyst",
    model={"id": "claude-opus-4-8", "inference_geo": "eu"},
)
```

Adding an advisor the coordinator may consult mid-turn:

```python
hook.create_agent(
    name="coordinator",
    model="claude-opus-4-8",
    multiagent={
        "type": "coordinator",
        "agents": [worker_agent_id, {"type": "advisor", "model": "claude-opus-4-8"}],
    },
)
```

## Why the shapes are verified rather than read off the types

Both payloads were exercised against the live API. That mattered: the first attempt omitted
`multiagent.type: "coordinator"` and was rejected with `multiagent.type: Field required`, so
documenting from the SDK type definitions alone would have shipped a payload that 400s.

The guide also records the constraints the API enforces (1 to 20 distinct roster entries, at
most one advisor, referenced agents may not themselves set `multiagent`) and that `model` is
whole-object replacement on update, so omitting `inference_geo` clears it rather than
preserving it.

## One code change

`create_agent`'s `model` annotation widens from `str | None` to accept a mapping. It already
worked at runtime, but a user type-checking their Dag would have got an error on the exact
snippet the guide tells them to write.

### DIFF
--- a/providers/anthropic/docs/operators/anthropic.rst
+++ b/providers/anthropic/docs/operators/anthropic.rst
@@ -283,6 +283,65 @@ task's real outcome is preserved and a warning is logged.
     session records instead. This is the same scenario as the retry warning above, so
     ``retries=0`` keeps both problems away.
 
+Configuring the agent
+"""""""""""""""""""""
+
+Agent-level settings are not operator arguments: they belong to the agent, which is created
+once and referenced by ID on every run.
+:meth:`~airflow.providers.anthropic.hooks.anthropic.AnthropicHook.create_agent` forwards
+keyword arguments to the API unchanged, so these need no provider support.
+
+**Pinning the inference region.** Pass ``model`` as a config object instead of a bare id to
+confine inference to one region:
+
+.. code-block:: python
+
+    hook.create_agent(
+        name="us-only-analyst",
+        model={"id": "claude-opus-5", "inference_geo": "us"},
+    )
+
+An unsupported value is rejected with a 400 naming the accepted set; see `Data residency
+<https://platform.claude.com/docs/en/manage-claude/data-residency>`__ for the regions
+Anthropic currently serves and the workspace-level controls. When ``inference_geo`` is unset,
+requests fall through to the workspace's
+``default_inference_geo``. On an update, ``model`` is whole-object replacement, so omitting
+``inference_geo`` clears it rather than preserving it.
+
+The pin is re-checked against the workspace allowlist when the agent is saved, when a
+session is created, and on every turn a session serves -- so narrowing the allowlist stops
+running sessions, not just new ones.
+
+In a ``multiagent`` configuration the coordinator's pin and every roster member's must all
+be set to the same value, or all be unset -- a mismatch is rejected. Following both this and
+the roster example below on one agent is the easy way to trip that.
+
+**Adding an advisor.** A coordinator agent can consult a second model mid-turn by adding an
+``advisor`` entry to its ``multiagent`` roster:
+
+.. code-block:: python
+
+    hook.create_agent(
+        name="coordinator",
+        model="claude-opus-5",
+        multiagent={
+            "type": "coordinator",
+            "agents": [
+                worker_agent_id,
+                {"type": "advisor", "model": "claude-opus-5"},
+            ],
+        },
+    )
+
+``type: "coordinator"`` on the ``multiagent`` object is required and the request is rejected
+without it. The roster takes 1 to 20 entries, each an agent ID
+string, a versioned ``{"type": "agent", "id": ..., "version": ...}`` reference,
+``{"type": "self"}`` for recursive self-invocation, or an ``advisor``. Referenced agents
+must exist, must be distinct, must not be archived, and must not themselves set
+``multiagent`` (depth limit 1); at most one ``self`` and at most one ``advisor``. The
+advisor occupies the roster name ``anthropic.advisor``, and its model must be permitted as
+an advisor for the coordinator's own model.
+
 .. exampleinclude:: /../tests/system/anthropic/example_anthropic_agent.py
     :language: python
     :dedent: 4
--- a/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
@@ -57,6 +57,7 @@ if TYPE_CHECKING:
         BetaManagedAgentsAgent,
         BetaManagedAgentsBudgetLimitParam,
         BetaManagedAgentsSession,
+        agent_create_params,
         environment_create_params,
     )
     from anthropic.types.beta.sessions import BetaManagedAgentsEventParams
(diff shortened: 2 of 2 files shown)


providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -282,6 +282,60 @@ task's real outcome is preserved and a warning is logged.
...
+
+.. code-block:: python
+
+    hook.create_agent(
+        name="us-only-analyst",
+        model={"id": "claude-opus-4-8", "inference_geo": "us"},
+    )
+
+The accepted values are ``"us"`` and ``"global"``; anything else is rejected with
+``400 inference_geo: must be one of ["global" "us"]``. When ``inference_geo`` is unset,
```
Lee-W:
```suggestion
``400 inference_geo: must be one of ["global", "us"]``. When ``inference_geo`` is unset,
```

Lee-W: if we're doing this, does that mean we'll manually update the list when the upstream/sdk add more support?

providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -282,6 +282,60 @@ task's real outcome is preserved and a warning is logged.
...
+the roster example below on one agent is the easy way to trip that.
+
+**Adding an advisor.** A coordinator agent can consult a second model mid-turn by adding an
+``advisor`` entry to its ``multiagent`` roster:
+
+.. code-block:: python
+
+    hook.create_agent(
+        name="coordinator",
+        model="claude-opus-4-8",
```
Lee-W:
```suggestion
        model="claude-opus-5",
```

providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -282,6 +282,60 @@ task's real outcome is preserved and a warning is logged.
...
+        multiagent={
+            "type": "coordinator",
+            "agents": [
+                worker_agent_id,
+                {"type": "advisor", "model": "claude-opus-4-8"},
```
Lee-W:
```suggestion
                {"type": "advisor", "model": "claude-opus-5"},
```

let's use the latest

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -575,16 +576,21 @@ def wait_for_batch(
...
+    def create_agent(
+        self, name: str, model: str | dict[str, Any] | None = None, **kwargs: Any
+    ) -> BetaManagedAgentsAgent:
         """
         Create a (reusable, versioned) Managed Agents agent. One-time setup.
 
         ``model`` defaults to :attr:`default_model` (the connection's ``extra['model']``
-        or :data:`DEFAULT_MODEL`).
+        or :data:`DEFAULT_MODEL`). Pass a mapping instead of a bare id to set the model
+        config, e.g. ``{"id": "claude-opus-4-8", "inference_geo": "us"}``.
```
```suggestion
        config, e.g. ``{"id": "claude-opus-5", "inference_geo": "us"}``.
```

providers/anthropic/docs/operators/anthropic.rst
Lee-W: I thought we're doing additional logic check here. But looks like the worst case is doc drift. or should we reference to the doc that list down what's supported?

kaxil: Not taking this one, but you were right that the line was wrong -- it just needed removing rather than fixing.

That string was quoting the API's response verbatim, and the API really does return `["global" "us"]` with no comma. Adding the comma would have made the quote inaccurate. But it reads as a typo, which is the problem: an error string copied verbatim invites exactly this correction, and it had already drawn one round of review over backslash escaping.

So the quote is gone. The text now says an unsupported value is rejected with a 400 naming the accepted set, and links to the page that owns the list -- which also answers your drift question on the same line.

kaxil: Good call, and I've gone with your second option -- the list is gone rather than maintained here.

Enumerating the regions means this guide owns a list Anthropic controls, so it drifts silently the moment a region is added. It now says an unsupported value is rejected with a 400 naming the accepted set, and links to [Data residency](https://platform.claude.com/docs/en/manage-claude/data-residency) for the current regions and the workspace-level controls.

On the "additional logic check" question: there is no validation in the provider, deliberately. `inference_geo` is `Optional[str]` in the SDK, not a Literal, so a client-side allowlist would be a second copy of the same list with the same drift problem -- and it would reject a newly-supported region before the API did. Letting the 400 through is the fail-open direction here, but it is the API's list that decides, and its error already names the accepted set.

While confirming this I also picked up something worth documenting, now added: the pin is re-validated against the workspace allowlist when the agent is saved, when a session is created, **and on every turn a session serves**. So narrowing the allowlist stops in-flight sessions, not just new ones.

kaxil: Answered on the sibling thread just above -- short version: no client-side check, because it would be a second copy of a list Anthropic owns and would reject a newly-supported region before the API did. The guide now links to the page that owns the list instead of enumerating it.

providers/anthropic/docs/operators/anthropic.rst
kaxil: Done, here and in the coordinator, the advisor entry, and the `create_agent` docstring.

Checked both roles against the API rather than assuming, since advisor eligibility is a separate policy from model availability: `claude-opus-5` is accepted as the agent model with `inference_geo: "us"`, and as an `advisor` entry in a coordinator roster.

Worth noting the provider's `DEFAULT_MODEL` is still `claude-opus-4-8`, so these examples are now ahead of the default. That seems right for docs, but it is a deliberate difference rather than an oversight.

providers/anthropic/docs/operators/anthropic.rst
kaxil: Applied. Verified against the API that `claude-opus-5` is accepted as an `advisor` roster entry, not just as an agent model -- advisor eligibility is a separate policy, so it was worth the call rather than assuming.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Applied to the docstring too, so it matches the guide.

providers/anthropic/docs/operators/anthropic.rst
```diff
@@ -282,6 +282,65 @@ task's real outcome is preserved and a warning is logged.
...
+
+    hook.create_agent(
+        name="us-only-analyst",
+        model={"id": "claude-opus-5", "inference_geo": "us"},
+    )
+
+An unsupported value is rejected with a 400 naming the accepted set; see `Data residency
+<https://platform.claude.com/docs/en/manage-claude/data-residency>`__ for the regions
+Anthropic currently serves and the workspace-level controls, rather than relying on a list
+copied here. When ``inference_geo`` is unset, requests fall through to the workspace's
```
Lee-W:
```suggestion
Anthropic currently serves and the workspace-level controls. When ``inference_geo`` is unset, requests fall through to the workspace's
```

This description feel like internal to us instead of something user should know

Lee-W (approved):
one nit

providers/anthropic/docs/operators/anthropic.rst
kaxil: Applied your suggestion. You are right that it was internal reasoning -- it explained why *we* chose not to copy the list, which tells a user nothing about the field. The sentence now just points at the page that owns the values.

PR #71465, opened 2026-08-11
Interrupt a running Anthropic session before archiving it

kaxil:
## Problem

The API refuses to archive **or** delete a session while its status is `running`, rejecting
both with a 400. `AnthropicAgentSessionOperator` treats archiving as best-effort teardown on
every failure path, so a session that will not stop on its own is left behind with no way to
release it -- and it keeps accruing `active_seconds`, which the SDK describes as the
duration the session's runtime cost is priced on.

## Evidence

Two sessions halted by a $0.01 budget sat at status `running` for over 40 minutes. Model
spend was frozen at the ceiling, but `active_seconds` kept climbing across successive
samples (777 to 838 seconds over one 60-second window). Both `archive` and `delete` returned
400. Sending `user.interrupt` moved them to `idle` immediately, after which archiving
succeeded.

## Solution

`AnthropicHook.archive_session` now interrupts and retries when the first archive attempt
fails, and `interrupt_session` exposes the event on its own. The retry is bounded, so a
session that genuinely cannot be archived still surfaces its error rather than looping.

This is reachable well beyond budgets: any session still working when a task fails, times
out, or is killed hits the same 400. A budget halt just makes it routine, because that is a
session which has stopped spending without stopping.

## Scope

The interrupt-then-archive sequence was verified by hand against the live API on two
genuinely stuck sessions. Its wiring through `archive_session` is covered by unit tests
rather than a live run.

### DIFF
--- a/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
+++ b/providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
@@ -31,6 +31,7 @@ from anthropic import (
     AnthropicBedrock,
     AnthropicFoundry,
     AnthropicVertex,
+    BadRequestError,
     IdentityTokenFile,
     WorkloadIdentityCredentials,
 )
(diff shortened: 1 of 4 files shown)


providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
```diff
@@ -266,6 +268,42 @@ def test_rejects_bad_amounts(self, amount, match):
             build_budget(amount)
 
 
+def _bad_request(message="cannot be archived while its status is running"):
```
```suggestion
def _make_bad_request(message="cannot be archived while its status is running") -> BadRequestError:
```

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
```diff
@@ -694,15 +695,58 @@ def send_event(self, session_id: str, event: dict[str, Any]) -> Any:
             session_id, events=cast("list[BetaManagedAgentsEventParams]", [event])
         )
 
-    def archive_session(self, session_id: str) -> BetaManagedAgentsSession:
+    def interrupt_session(self, session_id: str) -> Any:
```
Lee-W, line 712: do we want to make it a public method?

Lee-W (approved):

providers/anthropic/tests/unit/anthropic/hooks/test_anthropic.py
kaxil: Applied verbatim, plus both call sites. `BadRequestError` was already imported at module level, so no import change was needed.

providers/anthropic/src/airflow/providers/anthropic/hooks/anthropic.py
kaxil: Kept public, deliberately. `archive_session` is the only in-tree caller today, but the two have different time budgets: archiving retries for roughly 25s at the defaults, while a task's `on_kill` is SIGKILLed a few seconds in. A caller that just wants to stop billable runtime and let the session settle has a real use for the interrupt on its own.

It is also a thin wrapper over the already-public `send_event`, so making it private would hide the name without removing the capability. Happy to flip it if you would rather keep the surface minimal.

PR #65661, opened 2026-04-22
Pass try_number to extra links API

shashbha14:
In case a user chooses a retry from past attempts in the Details tab, the extra links were being created only for the latest attempt and not for the chosen retry attempt. This PR fixes that by passing try_number to the extra links API. This closes: #65354
Was generative AI tooling used to co-author this PR?
- [X] Yes — Claude Sonnet 4.6
Generated-by: Claude Sonnet 4.6 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)
'@

PR #71433, opened 2026-08-11
feat: add --ssl-ciphers option for API server [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

waterWang:
### Description

Adds support for Uvicorn `ssl_ciphers` and Gunicorn `ciphers` configuration options, allowing users to specify allowed SSL ciphers for the Airflow API server.

### Changes
- `cli_config.py`: Add `ARG_SSL_CIPHERS` CLI argument (`--ssl-ciphers`)
- `config.yml`: Add `ssl_ciphers` config option under `[api]` section
- `api_server_command.py`: Pass `ssl_ciphers` to Uvicorn as `ssl_ciphers` kwarg and to Gunicorn via `create_gunicorn_app`
- `gunicorn_app.py`: Accept `ssl_ciphers` parameter and set `options["ciphers"]`
- `test_api_server_command.py`: Add test coverage for `--ssl-ciphers` option

### Related Issue
Closes: #71426

PR #71472, opened 2026-08-11
fix: preserve extra link XCom keys across task retries

### LINKED ISSUE #71471: Operator extra links don't work per attempt
**TL;DR — operator extra links don't work per attempt.** After a retry, asking for an earlier
attempt's link returns the latest attempt's URL, so the button under attempt 1 opens attempt
2's logs.

It hits anyone whose link carries a job id the service mints per submission — every bundled
provider linking to per-attempt logs, so EMR, Glue, Databricks, Dataproc, Livy.

### Why

A link is one XCom row under `BaseOperatorLink.xcom_key` (`_link_<ClassName>` by default).

1. `XComOperatorLink.get_link` gets a `TaskInstanceKey` and ignores its `try_number`.
2. That row only ever holds the latest attempt: the run endpoint clears every XCom for the
   task before each attempt ([task_instances.py#L285-L293](https://github.com/apache/airflow/blob/537feafb3697338b3988d826b9c9b45da06b1f23/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py#L285-L293)),
   and `XComModel.set` deletes before it inserts.

\#65661 made the endpoint resolve the requested attempt, which is what makes the wrong answer
reachable from the UI.

No workaround exists: `xcom_key` is a plain property with no access to `ti_key`, the `xcom`
table has no `try_number` column, and the clear takes every key regardless.

### Reproduce

Run a task with an extra link, make it retry, ask the extra-links endpoint for
`try_number=1`. You get attempt 2's URL, and the metastore holds one `_link_*` row.

### Fix

#71522 — each attempt's link goes to the task state store, which a retry does not clear. That
keeps the XCom clear absolute instead of carving an exemption into it, and it is where
cross-retry state already lives (#69914, #71211).

Two alternatives were drafted and closed: a row per attempt in XCom (#71518) and the same with
the link opting in (#71521). #71518 is the one to revive if this needs to reach the 3.2 line,
since the state store landed after 3.2.2.

### Apache Airflow version

3.2.2. The code path is unchanged on `main` at 537feafb3697338b3988d826b9c9b45da06b1f23.

### Code of Conduct

waterWang:
When a task instance transitions to running, the execution API collects every XCom key for that task instance into `TIRunContext.xcom_keys_to_clear`, and the worker deletes each one. There is no filter — the only exemption is deferral.

This means a retry deletes the `_link_*` rows written by every previous attempt. An extra link can only ever resolve for the attempt that ran last. This collides with #65661 (try-aware extra-links endpoint), which made the read path attempt-aware while the write path still guarantees only the latest attempt's row exists.

**Fix**: Filter out XCom keys starting with `_link_` (the default `xcom_key` prefix for `BaseOperatorLink`) from the `xcom_keys_to_clear` list. This preserves per-attempt extra link rows across retries, allowing the try-aware endpoint to resolve links for any attempt.

Fixes #71471.

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -291,6 +291,10 @@ def ti_run(
                 xcom_query = xcom_query.where(XComModel.map_index == map_index)
 
             xcom_keys = list(session.scalars(xcom_query))
+            # Preserve extra link XCom keys across retries. Extra links are backed
+            # by XCom rows with keys starting with "_link_", and clearing them on
+            # retry would lose the per-attempt links for earlier attempts.
+            xcom_keys = [k for k in xcom_keys if not k.startswith("_link_")]
         task_reschedule_count = (
             session.scalar(
                 select(func.count(TaskReschedule.id)).where(TaskReschedule.ti_id == task_instance_id)


kaxil:
Closing it for the same reason as https://github.com/apache/airflow/pull/71433#issuecomment-5259029926

PR #71477, opened 2026-08-12
Link registry modules to the guide section that documents them

Lee-W:
A module card offers only generated API reference, which lists arguments but never says how the thing is meant to be used. The prose guides say it, and they already mark which class each section is about by titling that section with the class name -- so the pointer exists in the docs and just wasn't being read.

Resolving it from the guides rather than from a declared list is deliberate: a hand-maintained class-to-guide table goes stale silently every time a guide is split or renamed, and a link that lands on the wrong section is worse than no link at all. A class documented only in prose gets no link.

Both extraction paths resolve it, since a superseded version's page is rendered only from its own per-version metadata.

image

 

---

---

### DIFF
--- a/dev/registry/extract_parameters.py
+++ b/dev/registry/extract_parameters.py
@@ -54,6 +54,7 @@ from pathlib import Path
 import yaml
 from extract_metadata import fetch_provider_inventory, read_inventory
 from registry_contract_models import validate_modules_catalog, validate_provider_parameters
+from registry_tools.docs_guides import attach_guide_urls, collect_guide_anchors
 from registry_tools.types import (
     BASE_CLASS_IMPORTS,
     CLASS_LEVEL_CATEGORY_OVERRIDES,
@@ -95,6 +96,7 @@ class Module:
     provider_id: str
     provider_name: str
     supports_durable_execution: bool
+    guide_url: str | None = None
 
 
 def get_category(integration_name: str) -> str:
@@ -453,6 +455,16 @@ def _resolve_dotted_path(class_path: str) -> tuple[str, str, object] | None:
     return module_path, name, obj
 
 
+def read_guide_docs(docs_dir: Path) -> dict[str, str]:
+    """Read a provider's authored reST docs from the working tree, keyed by path relative to ``docs_dir``."""
+    if not docs_dir.is_dir():
+        return {}
+    return {
+        path.relative_to(docs_dir).as_posix(): path.read_text(encoding="utf-8")
+        for path in sorted(docs_dir.rglob("*.rst"))
+    }
+
+
 def discover_classes_from_provider(
     provider_yaml_path: Path,
     base_classes: dict[str, type],
@@ -690,6 +702,9 @@ def discover_classes_from_provider(
                 "provider_name": provider_name,
             }
         )
+
+    guide_docs = read_guide_docs(provider_yaml_path.parent / "docs")
+    attach_guide_urls(discovered, collect_guide_anchors(guide_docs), base_docs_url)
 
     return discovered
 
--- a/dev/registry/extract_versions.py
+++ b/dev/registry/extract_versions.py
@@ -59,6 +59,7 @@ except ImportError:
     sys.exit(1)
 
 from extract_metadata import fetch_provider_inventory, read_connection_urls, resolve_connection_docs_url
+from registry_tools.docs_guides import attach_guide_urls, collect_guide_anchors
 from registry_tools.types import (
     CLASS_LEVEL_CATEGORY_OVERRIDES,
     CLASS_LEVEL_SECTIONS,
@@ -129,6 +130,21 @@ def git_show(tag: str, path: str) -> str | None:
         return None
 
 
+def git_ls_tree(tag: str, prefix: str) -> list[str]:
+    """List the file paths under a prefix at a specific git tag."""
+    try:
+        result = subprocess.run(
+            ["git", "ls-tree", "-r", "--name-only", tag, "--", prefix],
+            capture_output=True,
+            text=True,
+            cwd=AIRFLOW_ROOT,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return []
+    return [line for line in result.stdout.splitlines() if line]
+
+
 def git_tag_exists(tag: str) -> bool:
     """Check if a git tag exists locally."""
     result = subprocess.run(
@@ -179,6 +195,27 @@ def get_source_file_path(layout: str, dir_path: str, module_path: str) -> str:
     if layout == "new":
         return f"providers/{dir_path}/src/{rel_file}"
     return f"providers/src/{rel_file}"
+
+
+def read_guide_docs(tag: str, layout: str, dir_path: str) -> dict[str, str]:
+    """Read a provider's authored reST docs at a tag, keyed by path relative to its docs dir.
+
+    Only the per-provider layout keeps docs beside the provider; under the old flat
+    layout they lived in a top-level ``docs/`` tree, so those tags get no guide
+    links rather than links guessed from a path that moved.
+    """
+    if layout != "new":
+        return {}
+
+    docs_prefix = f"providers/{dir_path}/docs/"
+    docs: dict[str, str] = {}
+    for path in git_ls_tree(tag, docs_prefix):
+        if not path.endswith(".rst"):
+            continue
+        content = git_show(tag, path)
+        if content:
+            docs[path[len(docs_prefix) :]] = content
+    return docs
 
 
 def parse_pyproject_toml_content(content: str, layout: str) -> dict[str, Any]:
@@ -381,6 +418,8 @@ def extract_modules_from_yaml(
                     "category": category,
                 }
             )
+
+    attach_guide_urls(modules, collect_guide_anchors(read_guide_docs(tag, layout, dir_path)), base_docs_url)
 
     return modules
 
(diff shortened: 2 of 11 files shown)


dev/registry/registry_tools/docs_guides.py
```diff
@@ -0,0 +1,124 @@
...
+from collections.abc import Mapping
+from typing import Any
+
+# reST underlines an (optionally overlined) section title with a run of one
+# punctuation character, at least as long as the title itself.
+_ADORNMENT_CHARS = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
+
+# Only titles opening with a class name as an inline literal are treated as
+# documenting it, so prose headings ("Bounded query results") never produce a link.
+_LEADING_CLASS_LITERAL = re.compile(r"^``([A-Za-z_][A-Za-z0-9_]*)``")
```
kaxil: The catalog carries decorators as modules named `@task.agent`, `@task.llm` and so on, and `operators/agent.rst` titles its section ``AgentOperator`` & ``@task.agent`` precisely because it documents both. Only the leading literal is captured, so the decorator half never gets a Guide link even though the title names it (`llm_file_analysis.rst` is the same shape). Could this collect every inline literal in the title, with the pattern widened enough to admit the `@task.x` form?

dev/registry/extract_parameters.py
```diff
@@ -453,6 +455,16 @@ def _resolve_dotted_path(class_path: str) -> tuple[str, str, object] | None:
     return module_path, name, obj
 
 
+def read_guide_docs(docs_dir: Path) -> dict[str, str]:
+    """Read a provider's authored reST docs from the working tree, keyed by path relative to ``docs_dir``."""
+    if not docs_dir.is_dir():
+        return {}
+    return {
+        path.relative_to(docs_dir).as_posix(): path.read_text(encoding="utf-8")
+        for path in sorted(docs_dir.rglob("*.rst"))
```
kaxil: `rglob` picks up everything under `docs/`, including the autoapi-generated `_api/` tree. common/ai has 93 `.rst` on disk but only 36 tracked, so the anchor set here depends on whether a docs build has run in the tree this executes against. Include-only files have the same problem (`providers/amazon/docs/_partials/` is never built as a page), and since `_`-prefixed paths sort first they would win the first-page tie-break in `collect_guide_anchors` and emit a link to an `.html` that does not exist. Skipping path parts starting with `_` closes both.

dev/registry/extract_versions.py
```diff
@@ -181,6 +197,26 @@ def get_source_file_path(layout: str, dir_path: str, module_path: str) -> str:
...
+    layout they lived in a top-level ``docs/`` tree, so those tags get no guide
+    links rather than links guessed from a path that moved.
+    """
+    if layout != "new":
+        return {}
+
+    docs_prefix = f"providers/{dir_path}/docs/"
+    docs: dict[str, str] = {}
+    for path in git_ls_tree(tag, docs_prefix):
+        if not path.endswith(".rst"):
```
kaxil, line 213: This spawns one `git show` per `.rst`, and most of that set can never yield an anchor. At `providers-amazon/9.17.0` it is 102 files and 1.45s, of which `changelog.rst` alone is 140KB, and `--all-versions` pays that per provider-version. Filtering to the pages that are actually built (skip `_`-prefixed dirs plus `changelog.rst` and `commits.rst`) would cut most of the cost and fixes the same dead-page tie-break as the working-tree reader.

registry/AGENTS.md
```diff
@@ -462,6 +462,24 @@ They run inside Breeze where all providers are installed. `extract_metadata.py`
...
+### How a module gets a "Guide" link
+
+A module card links to the how-to guide section that documents it, alongside the
+generated API reference. Nothing declares that link: `registry_tools/docs_guides.py`
+reads the provider's own `docs/*.rst` and matches a class to a section when the
+section's title *opens with the class name as an inline literal* — ``` ``HookToolset`` ```
+or ``` ``AgentOperator`` & ``@task.agent`` ```. The anchor is derived from the whole
+title the way docutils derives its HTML id.
+
+That convention is what the guides already do, and it is deliberately the only
```
kaxil, line 474: I checked this across the tree: common/ai is the only provider whose guides title sections with a leading inline literal, 17 of them. informatica has 3, but they name config options rather than classes, and the other 92 providers have none. Worth saying so here, since it means extending Guide links to another provider is a matter of editing that provider's headings rather than touching this extractor.

dev/registry/extract_parameters.py
```diff
@@ -95,6 +96,7 @@ class Module:
     provider_id: str
     provider_name: str
     supports_durable_execution: bool
+    guide_url: str | None = None
```
kaxil, line 99: That makes 13 fields, so `discover_classes_from_provider`'s docstring below ("all 12 Module fields") is now off by one.

dev/registry/registry_tools/docs_guides.py
Lee-W: `_LEADING_LITERAL_NAME_RUN` now takes a run of inline literals joined by `&`, `,` or `/`, and the name pattern admits a leading `@` — so ``AgentOperator`` & ``@task.agent`` gives both halves the same anchor. `common/ai` resolves 19 names now, `@task.agent` and `@task.llm_file_analysis` among them.

dev/registry/extract_parameters.py
Lee-W: drops paths with `_` prefix, `changelog.rst` and `commits.rst`

dev/registry/extract_versions.py
Lee-W: Same `is_guide_page` filter here, applied to the `git_ls_tree` listing before `git_show`, so skipped pages are never read at all.

registry/AGENTS.md
Lee-W: Added a paragraph for it — growing Guide-link coverage is a matter of editing that provider's section titles, not this extractor.

dev/registry/extract_parameters.py
Lee-W: fixed it. thanks!

kaxil:
Round-1 items all landed, thanks. Four follow-ups and two nits, none blocking.

**`extract_versions.py`: the `is_guide_page` filter runs before `git_show` now, but it removes very little at a release tag.** At `providers-amazon/9.36.0` the docs tree holds 111 `.rst`, of which 2 are `_`-prefixed and 2 are `changelog.rst`/`commits.rst`, so 107 `git show` spawns remain and the per-provider-version cost from round 1 is still paid. Feeding `<tag>:<path>` lines to a single `git cat-file --batch` on stdin builds a byte-identical dict in 0.019s against 1.254s for the loop, with `git_ls_tree` and `is_guide_page` unchanged. `--all-versions` pays the current cost once per provider-version.

**`extract_parameters.py:481`: my round-1 off-by-one prompted the wrong correction.** `attach_guide_urls` does `if not anchor: continue`, so `guide_url` is set only for classes a guide documents, which is what the new `test_module_contract_omits_guide_url_for_undocumented_classes` asserts. Entries for undocumented classes carry 12 keys, not 13. "all 12 Module fields, plus `guide_url` when a how-to guide documents the class" would be accurate.

**`test_module_contract_omits_guide_url_for_undocumented_classes` cannot fail.** `_validate` in `registry_contract_models.py` validates and then returns the input payload rather than `model_dump()`, so the assertion checks that a dict built without `guide_url` has no `guide_url`. `ModuleContract.model_validate(_module_payload()).guide_url is None` pins the field itself. `test_module_contract_round_trips_guide_url` is not vacuous, since `extra="forbid"` rejects an undeclared field, but its name promises a round-trip it does not perform.

**`AGENTS.md:468` "Nothing declares that link" is not quite right.** `provider.yaml` declares `integrations[].how-to-guide` and `transfers[].how-to-guide`, and `check_doc_files` in `run_provider_yaml_files_check.py` set-compares that declared mapping against a glob of the operator, sensor and transfer guide paths, so it fails CI rather than rotting silently. The justification that does hold is the one worth writing down: the declared field names a page and never a section, and it does not reach the toolset, hook or decorator pages this reads. Worth correcting because `AGENTS.md` is a durable instruction file.

Two nits. `.provider-detail-page .module-actions` in `main.css` is a flex row with no `flex-wrap`, and this PR takes it from two links to three, so `flex-wrap: wrap` is cheap insurance at narrow viewports or a raised root font size. And `git_ls_tree` in `extract_versions.py` runs with `core.quotePath` at its default, which C-quotes a non-ASCII path so it fails the `.endswith(".rst")` test while the working-tree reader accepts it; `-c core.quotePath=false` closes that divergence. No such path exists under `providers/` today.

Lee-W:
`read_guide_docs` now uses `git cat-file --batch`: at `providers-amazon/9.36.0` that is one subprocess and ~0.05s against 107 and ~1.9s, for a dict that compares equal to the loop's key for key. Parsing stays on bytes since the header counts them, and content decodes as explicit UTF-8 rather than following locale, with a decode failure left to propagate.

The `discover_classes_from_provider` docstring now uses "all 12 Module fields, plus `guide_url` when a how-to guide documents the class".

`test_module_contract_omits_guide_url_for_undocumented_classes` now asserts on
`ModuleContract.model_validate(_module_payload()).guide_url`, and the round-trip one is renamed `test_module_contract_preserves_guide_url_value` and reads the attribute off the model as well. `_validate` still returns the payload it was handed — switching it to `model_dump()` would inject defaults into every emitted entry — so the tests moved to the model instead.

`AGENTS.md` drops "Nothing declares that link" for the justification that holds:
`provider.yaml`'s `how-to-guide` is CI-enforced by `check_doc_files`, but it names a whole page and never a section, and it covers only operators, sensors and transfers — not the toolset, hook and decorator pages this reads.

---

`.provider-detail-page .module-actions` gets `flex-wrap: wrap`, and `git_ls_tree` runs with `core.quotePath=false`. I kept your framing on the second one: no
non-ASCII path exists under `providers/` today, so it closes the divergence between the two readers rather than fixing an observed break.

PR #71478, opened 2026-08-12
Find providers in the registry by the external services they reach

Lee-W:
## Why

Users often search the service they want to interfact in registry instead of the airflow provider name. (e.g., "openai" or "ollama" rather than "common-ai") Those names were declared in `provider.yaml` and rendered in a table on the provider page, but neither search path read them.

## What
- Both search paths now match on the external services a provider's connections reach. -> "openai" or "ollama" now finds Common AI.
- A result surfaced this way carries a badge naming the service that matched. Previously the badge took the first entry the query appeared anywhere inside, so "ai" credited OpenAI and "llm" credited vLLM; a hyphenated query like "mistral-ai" matched no service at all and got no badge.
- The Explore page's **ai-ml** category — and the matching entry in the `/providers/` category dropdown — now also matches providers through their external services, so a provider reachable only that way can be found by filtering, not just by typing.

 

---

---

### DIFF
--- a/registry/scripts/build-pagefind-index.mjs
+++ b/registry/scripts/build-pagefind-index.mjs
@@ -20,6 +20,8 @@
 import fs from 'fs';
 import * as pagefind from "pagefind";
 
+import { collectExternalServices } from '../src/_data/providerExternalServices.js';
+
 async function buildPagefindIndex() {
   console.log('Building PageFind index with custom records...');
 
@@ -44,13 +46,14 @@ async function buildPagefindIndex() {
       .map((category) => category.name)
       .filter(Boolean)
       .join(' ');
+    const externalServices = collectExternalServices(provider).join(' ');
 
     await index.addCustomRecord({
       url: `/providers/${provider.id}/${provider.version}/`,
       // The id is indexed on its own, not as part of the distribution name:
       // the `apache-airflow-providers-` prefix is shared by every provider and
       // would make 'apache' or 'airflow' match all of them.
-      content: `${provider.name} ${provider.id} ${provider.description} ${integrations}`,
+      content: `${provider.name} ${provider.id} ${provider.description} ${integrations} ${externalServices}`,
       language: 'en',
       meta: {
         type: 'provider',
--- a/registry/src/_data/providerExternalServices.js
+++ b/registry/src/_data/providerExternalServices.js
@@ -1,0 +1,43 @@
+/*!
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Shared by the pagefind index builder and the `externalServices` Eleventy
+// filter (used for the /providers/ filter box), so the two search paths read
+// the same field off the same shape and can't drift apart.
+
+// The services a provider's connection types reach -- for Common AI these are
+// the LLM providers behind each `pydanticai` / `langchain` connection, which are
+// the names people actually search for ("openai", "ollama") even though no
+// provider is called that.
+function collectExternalServices(provider) {
+  const seen = new Set();
+  const services = [];
+  for (const connection of provider.connection_types || []) {
+    for (const service of connection.external_services || []) {
+      const key = service.toLowerCase();
+      if (!seen.has(key)) {
+        seen.add(key);
+        services.push(service);
+      }
+    }
+  }
+  return services;
+}
+
+module.exports = { collectExternalServices };
--- a/registry/.eleventy.js
+++ b/registry/.eleventy.js
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+const { collectExternalServices } = require("./src/_data/providerExternalServices");
+
 module.exports = function(eleventyConfig) {
   // Copy static assets
   eleventyConfig.addPassthroughCopy("src/assets");
@@ -119,6 +121,14 @@ module.exports = function(eleventyConfig) {
     const match = types.find((t) => t.id === typeId);
     if (match && match.icon) return match.icon;
     return typeId.charAt(0).toUpperCase();
+  });
+
+  // Flattens a provider's per-connection external services into one deduped
+  // list, so the /providers/ filter box can match them the way the pagefind
+  // index already does.
+  eleventyConfig.addFilter("externalServices", (provider) => {
+    if (!provider || typeof provider !== "object") return [];
+    return collectExternalServices(provider);
   });
 
   eleventyConfig.addShortcode("year", () => `${new Date().getFullYear()}`);
--- a/registry/src/js/provider-filters.js
+++ b/registry/src/js/provider-filters.js
@@ -62,11 +62,13 @@
       const id = item.dataset.id || '';
       const categories = item.dataset.categories || '';
       const integrations = item.dataset.integrations || '';
+      const services = item.dataset.services || '';
 
       const matchesLifecycle = currentLifecycle === 'all' || lifecycle === currentLifecycle;
       const matchesSearch = normalize(name).includes(search) ||
         normalize(id).includes(search) ||
-        normalize(integrations).includes(search);
+        normalize(integrations).includes(search) ||
+        normalize(services).includes(search);
       const matchesCategory = currentCategory === 'all' ||
         categories.split(',').includes(currentCategory);
 
--- a/registry/src/providers.njk
+++ b/registry/src/providers.njk
@@ -70,7 +70,8 @@ mainClass: providers-page
           data-downloads="{{ provider.pypi_downloads.monthly }}"
           data-updated="{{ provider.last_updated or '' }}"
           data-categories="{{ cats | join(',') }}"
-          data-integrations="{{ (provider.categories or []) | selectattr('name') | join(',', 'name') | lower }}">
+          data-integrations="{{ (provider.categories or []) | selectattr('name') | join(',', 'name') | lower }}"
+          data-services="{{ provider | externalServices | join(',') | lower }}">
         {% include "provider-card.njk" %}
       </li>
       {% endfor %}


registry/src/js/search.js
```diff
@@ -110,13 +111,25 @@
...
 
+      // A provider often matches because of an external service its connections
+      // reach (e.g. "anthropic" -> Common AI), not the provider's own name --
+      // badge the specific service so that's visible instead of just the
+      // provider it lives under.
+      let matchedService = '';
+      if (type === 'provider' && currentQuery) {
+        const services = (result.meta.externalServices || '').split(',').filter(Boolean);
+        const normalizedQuery = currentQuery.toLowerCase();
+        matchedService = services.find((service) => service.toLowerCase().includes(normalizedQuery)) || '';
```
kaxil: `find` returns the first service the query is a substring of, not the one that actually made pagefind match, so with Common AI's service list `ai` badges OpenAI and `llm` badges vLLM. It also skips the `normalize()` collapse that `provider-filters.js` and `providerKeywordMatch.js` both do, so `mistral-ai` gets no badge at all. Matching on a word start after the same normalize would cover both.

registry/src/_data/providerExternalServices.js
```diff
@@ -0,0 +1,43 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Shared by the pagefind index builder and the `externalServices` Eleventy
+// filter (used for the /providers/ filter box), so the two search paths read
+// the same field off the same shape and can't drift apart.
```
kaxil: There's a third place doing this job: `collectSearchableValues()` in `providerKeywordMatch.js`, whose comment says "every string a provider is searchable by" and reads only the id plus integration names. The `ai-ml` explore category keys on `openai` and `anthropic`, so a provider that reaches those only through `external-services` would match the text box on /providers/ but not the category dropdown next to it. No behaviour change today, since common.ai already lands in `ai-ml` via "LangChain" and "Pydantic AI", but should external services feed that matcher too?

Lee-W: Yes, but opt-in per category rather than for all of them.

`collectSearchableValues()` now takes `{ includeExternalServices }` and `exploreCategories.js` sets it only on `ai-ml`; the default is off, so a new category has to opt in.

Feeding every category would misfile providers. `cloud`'s keywords (`amazon`, `google`, `microsoft-azure`, …) are provider distribution ids — an identity claim — so Common AI's bare `Google` service name would land it in `cloud`. `ai-ml` is the only category keyed on integration names (`langchain`, `llamaindex`, `pydantic ai`), which is the same axis `external-services` describes. The two consumers now share `categoryMatchesProvider()` so the options can't drift between them.

registry/src/js/search.js
Lee-W: The badge now runs the query through the same `normalize()` collapse, then matches a word start instead of a substring — `ai` no longer credits OpenAI, `llm` credits nothing, and `mistral-ai` resolves to Mistral AI.

registry/scripts/build-pagefind-index.mjs
```diff
@@ -44,20 +46,26 @@ async function buildPagefindIndex() {
...
+    const externalServicesList = collectExternalServices(provider);
+    const externalServices = externalServicesList.join(' ');
 
     await index.addCustomRecord({
       url: `/providers/${provider.id}/${provider.version}/`,
       // The id is indexed on its own, not as part of the distribution name:
       // the `apache-airflow-providers-` prefix is shared by every provider and
       // would make 'apache' or 'airflow' match all of them.
-      content: `${provider.name} ${provider.id} ${provider.description} ${integrations}`,
+      content: `${provider.name} ${provider.id} ${provider.description} ${integrations} ${externalServices}`,
```
kaxil, line 57: Putting the service names in `content` flips ranking against the providers that actually implement the service, because pagefind's BM25 normalises by document length. Common AI's record goes 24 -> 40 words while Amazon's stays 153 and Google's 219, each with a single occurrence of the term. A/B on the real pagefind runtime, Providers tab: `bedrock` goes amazon 0.604 (#1) -> common-ai 1.659 (#1, amazon 0.537); `vertex` goes google 0.433 -> common-ai 1.659; `vertex ai` goes google 1.162 -> common-ai 7.546. `google`, `aws`, `azure` and `openai` are untouched -- the incumbents keep #1 by a wide margin. The gain is real too: `ollama` and `groq` went from zero results to Common AI, and `anthropic` from noise (openlineage) to the right provider. `addCustomRecord` has no per-term weighting, so the lever is which services go into `content`: keeping a service out of `content` when it collides with another provider's own name or id (leaving it in `meta` and `data-services`) would preserve both effects. Either way it's worth a line in the description so the tradeoff is signed off rather than discovered.

registry/src/js/search.js
```diff
@@ -110,13 +120,29 @@
...
 
+      // A provider often matches because of an external service its connections
+      // reach (e.g. "anthropic" -> Common AI), not the provider's own name --
+      // badge the specific service so that's visible instead of just the
+      // provider it lives under.
+      let matchedService = '';
+      if (type === 'provider' && currentQuery) {
+        const services = (result.meta.externalServices || '').split(',').filter(Boolean);
+        const normalizedQuery = normalize(currentQuery);
+        matchedService = services.find((service) => {
```
kaxil, line 131: The word-start fix does what I asked and `mistral-ai`, `llm` and `ai`-vs-OpenAI all behave now, but the badge still derives from `currentQuery` alone and never from what pagefind matched, so it can still name a service that had nothing to do with the hit. Searching `ai` returns Common AI on its own name and badges it "Mistral AI"; `a` returns it #1 and badges "Anthropic". Skipping the lookup when the provider's own identity already explains the hit fixes it without losing any legitimate badge (openai, anthropic, bedrock, vertex, azure, ollama, vllm, mistral-ai all still badge): `const selfText = normalize(name + ' ' + (result.meta.providerId || ''))` and only run the `find` when `!selfText.includes(normalizedQuery)`.

registry/src/_data/providerKeywordMatch.js
```diff
@@ -39,20 +45,30 @@ function fuzzyIncludes(value, keyword) {
...
 
 // Every string a provider is searchable by: its id/slug and its declared
-// integration names (provider.categories[].name — e.g. "LangChain",
-// "Pydantic AI").
-function collectSearchableValues(provider) {
+// integration names (provider.categories[].name) are always included.
+// Its declared `external_services` are included only when the caller opts
+// in via `{ includeExternalServices: true }` — see exploreCategories.js
+// for which category does and why.
+function collectSearchableValues(provider, { includeExternalServices = false } = {}) {
```
kaxil, line 52: `registry/README.md` still documents this matcher as keywords "matched against `provider.id` and the provider's declared integration names", in both the `keywords` field bullet and the paragraph below it, and the field list doesn't mention `includeExternalServices` at all. #70498 widened this same matcher and updated that exact paragraph, so it seems worth the same treatment here.

registry/src/_data/exploreCategories.js
```diff
@@ -57,6 +57,13 @@ module.exports = [
     color: 'amber',
     keywords: ['openai', 'cohere', 'anthropic', 'huggingface', 'mlflow', 'pinecone', 'qdrant', 'weaviate', 'pgvector', 'langchain', 'llamaindex', 'mcp', 'pydantic ai'],
     description: 'OpenAI, vector DBs, and ML platforms',
+    // Opt-in only: this category's keywords are capability claims ("this
+    // provider integrates with X"), the same axis as `external-services`
+    // ("upstream provider/service names this connection type reaches" --
+    // provider.yaml.schema.json:475), so external services may also match
+    // here. Every other category's keywords are identity claims (this
+    // distribution's own directory id) and must not opt in.
+    includeExternalServices: true,
```
kaxil, line 66: The conclusion is right, but the rule stated for it doesn't hold in either direction, and this comment is the only written rationale for the new flag anywhere. `dingtalk` isn't a directory id (the distribution is `dingding`; it matches only the integration name "DingTalk", which is why #70498 added it), and neither are `dataflow`, `couchbase` or `twilio` -- while 7 of `ai-ml`'s own 13 keywords are directory ids (openai, cohere, anthropic, pinecone, qdrant, weaviate, pgvector). The concrete version is stronger: flipping the flag on for every category across all 107 `provider.yaml` files changes exactly one, `cloud`, which picks up Common AI through keyword `google` matching the services "Google" and "Google Vertex AI". Worth saying that instead. Related: on today's data the `ai-ml` opt-in is a 0-delta, since Common AI already matches through langchain, llamaindex, mcp and pydantic ai, so a sentence in the description saves a reviewer hunting for a changed Explore page.

kaxil:
Approving this. I rebuilt the site at `81d4d470` with real dependencies and queried pagefind's own runtime rather than reasoning about the index, and the collision filter does what it set out to do: no incumbent loses its top spot (`openai` 11.30 vs common-ai 5.07, `anthropic` 12.15 vs 3.66, `google` 1.55 vs 0.46, `azure` 5.51 vs 1.75), while `ollama`, `groq`, `deepseek`, `vllm` and `mistral` went from zero results to Common AI.

The category refactor came out clean: generating the category map at this head and on `main` over the same 107 providers gives zero differences, so routing both consumers through `categoryMatchesProvider` changed no existing assignment. I also checked the statement in the `exploreCategories.js` comment and it holds exactly, forcing the flag on for all seven categories changes only `cloud`.

Four small things I'd take or leave, none of them worth holding this up:

1. The first bullet in the description says "Both search paths now match on the external services a provider's connections reach". That holds for 9 of the 11 services, but `AWS Bedrock` and `Google Vertex AI` are deliberately kept out of the pagefind text, so `bedrock` and `vertex` find Common AI on /providers/ and return nothing in Cmd-K. Both examples you give are accurate, it's the general sentence that could use the caveat, since it becomes the commit message.

2. `providerExternalServices.js` line 74 names my handle and cites numbers that only exist in a review thread, which nobody reading the file later can check. Could you drop the attribution? The mechanism reads fine on its own: single-token vendor names stay in the indexed text because the incumbent still ranks first on those queries. The clause after it isn't right either, since `openai`, `anthropic` and `google` are separate providers in this tree, so a single-token name would collide with them rather than with Common AI itself.

3. The `gcp` alias never fires. No service name or `integration-name` across the 107 `provider.yaml` files contains that token, and removing the entry leaves the filter output byte for byte identical, whereas removing `aws` stops "AWS Bedrock" being caught. The comment above it also points at "the collision-detection PR" for design notes, and this diff is that PR.

4. `tokenizeForCollisionCheck` splits on whitespace only, while all three `normalize()` copies and the matching rule the README now states collapse `-`, `_` and whitespace together. `external-services` is free-form, so `AWS-Bedrock` or `AWS_Bedrock` would slip past the check that exists to catch `AWS Bedrock`. I ran `/[-_\s]+/` there against all 107 files and the output is identical on today's data, so it's a safe swap whenever you next touch the file.

Thanks for the README and comment updates from the last round, both went in cleanly.

kaxil (approved):

PR #71480, opened 2026-08-12
Anchor run_id to logical_date when create_cron_data_intervals is enab…

### LINKED ISSUE #71187: A way to switch create_cron_data_interval=true back to 2.x run_id logic
### Description

[apache/airflow#59115](https://github.com/apache/airflow/pull/59115) (merged in 3.2.0, as part of AIP-76
partition scheduling support) changed how `run_id` is generated for scheduled DAG runs. This is documented
plainly in [apache/airflow#65688](https://github.com/apache/airflow/pull/65688):

> Airflow 3.2.0 (related PR: #59115) changed how run_id is generated for scheduled DAG runs:
> - 3.1.x: `run_id = scheduled__<logical_date>`
> - 3.2.0: `run_id = scheduled__<run_after>`

This is confirmed intentional, not a bug, see the base `Timetable.generate_run_id()` implementation:
https://github.com/apache/airflow/blob/3.3.0/airflow-core/src/airflow/timetables/base.py#L460-L474

The issue: `AIRFLOW__SCHEDULER__CREATE_CRON_DATA_INTERVALS` exists specifically to let cron-scheduled DAGs
opt back into Airflow-2-style semantics (logical_date == interval start, run_after offset). It restores
`logical_date`/data-interval behavior, but does **not** cover `run_id` generation, `CronDataIntervalTimetable`
(the class this setting selects) doesn't override `generate_run_id()`:
https://github.com/apache/airflow/blob/3.3.0/airflow-core/src/airflow/timetables/interval.py#L133

So users who explicitly opt into `CREATE_CRON_DATA_INTERVALS=True` for full Airflow-2 semantics still end up
with `run_id` decoupled from `logical_date`,  a partial, easy-to-miss gap in what that setting promises.

Related, currently open: [apache/airflow#70167](https://github.com/apache/airflow/pull/70167) is updating
`dag-run.rst` to document the new default `run_id`/`logical_date` behavior, confirming this is the accepted,
documented state going forward, but doesn't address the `CREATE_CRON_DATA_INTERVALS` coverage gap raised here.

What should happen instead:
Two possible directions, in rough order of scope:

1. Extend `CREATE_CRON_DATA_INTERVALS` (or a new, separate flag) to also anchor `run_id` to the data
   interval start when enabled.
2. A more general mechanism e.g. a `dag_run_policy`-style cluster policy hook, that would let users
   customize `run_id` generation (and potentially other DagRun attributes) without a purpose-built config
   flag. This was suggested informally by an Airflow core maintainer as possibly the cleaner long-term
   design, though not something planned near-term.

Either would close the gap for deployments that rely on `CREATE_CRON_DATA_INTERVALS` for backward
compatibility during 2→3 migrations.

Workaround:
A custom `Timetable` subclass overriding `generate_run_id()` to key off `data_interval.start` works today
and requires no core changes:

​```python
from airflow.timetables.interval import CronDataIntervalTimetable

class LogicalDateRunIdTimetable(CronDataIntervalTimetable):
    def generate_run_id(self, *, run_type, run_after, data_interval=None, **extra) -> str:
        if data_interval is not None:
            return run_type.generate_run_id(suffix=data_interval.start.isoformat())
        return super().generate_run_id(
            run_type=run_type, run_after=run_after, data_interval=data_interval, **extra
        )
​```

This is DAG-level, not deployment-wide, so it doesn't fully substitute for a config-level fix for users
with many DAGs.

### Use case/motivation

Airflow deployments migrating from 2 to 3 that rely on `CREATE_CRON_DATA_INTERVALS` for backward
compatibility expect that setting to fully restore Airflow-2-style semantics. Currently `run_id` silently
falls outside that guarantee, which can break anything downstream that parses `run_id` and assumes it
reflects `logical_date`, without any warning that the setting's coverage is partial.

### Related issues

_No response_

### Code of Conduct

bujjibabukatta:
Closes #71187.

### Problem
`[scheduler] create_cron_data_intervals` restores Airflow 2-style `logical_date`/data-interval
semantics for `CronDataIntervalTimetable` and `DeltaDataIntervalTimetable`, but it doesn't extend
to `run_id`. `generate_run_id()` is only defined once, on the shared base `Timetable` class, and
always anchors to `run_after` — the AIP-76 default. So enabling the flag restores `logical_date`
but silently leaves `run_id` on the new scheme, which breaks any workflow that parses `run_id` to
recover the logical date (a common pattern carried over from Airflow 2).

This is a documented gap, not a regression — `_create_timetable()` in
`airflow.sdk.definitions.dag` confirms `_DataIntervalTimetable` subclasses are only instantiated
when the flag is on, so the fix is naturally scoped by construction rather than an extra
conditional.

### Fix
Adds a `generate_run_id()` override on the shared `_DataIntervalTimetable` base in
`airflow-core/src/airflow/timetables/interval.py`, covering both `CronDataIntervalTimetable` and
`DeltaDataIntervalTimetable` symmetrically. It anchors to `data_interval.start` when a data
interval is available, and falls back to the base (`run_after`-anchored) behavior otherwise, so it
never raises on an unexpected `None`. `CronTriggerTimetable`/`DeltaTriggerTimetable` (the AIP-76
default path) are untouched.

This mirrors the `LogicalDateRunIdTimetable` workaround already posted in the issue by the
reporter, who confirmed it works in their environment — promoted here into core so it applies
without a custom DAG-level subclass.

### Tests
Added three regression tests to `airflow-core/tests/unit/timetables/test_interval_timetable.py`:
- `test_generate_run_id_anchors_to_data_interval_start` — `run_id` anchors to `data_interval.start`
  for both `CronDataIntervalTimetable` and `DeltaDataIntervalTimetable`
- `test_generate_run_id_falls_back_without_data_interval` — falls back correctly when
  `data_interval=None`
- `test_generate_run_id_default_timetable_unaffected` — confirms `CronTriggerTimetable` (default)
  is unaffected, i.e. zero regression to the majority code path

### Docs
- `config.yml`: extended the `create_cron_data_intervals` description to note it now also covers
  `run_id`.
- `authoring-and-scheduling/timetable.rst`: added a scoping note to the `run_id`/`logical_date`
  section — that section previously described this behavior as already true, which it wasn't until
  this PR.
- `installation/upgrading_to_airflow3.rst`: added a note for 2→3 migrators who rely on `run_id`
  reflecting the logical date.

### Verification performed
- `python -m py_compile` on changed files
- `ruff check` / `ruff format --check` against the repo's actual `pyproject.toml` — clean
- Installed the real dependency chain and imported the actual patched classes (not stand-ins) to
  confirm `run_id` anchoring, the `None` fallback, and that `CronTriggerTimetable` is unaffected —
  all as expected
- Best-effort `mypy` (couldn't install the internal `airflow_mypy` plugin outside CI, so treat as a
  strong signal, not equivalent to the CI job)
- **Not run**: the full pytest suite via Breeze (`tests/conftest.py` shells out to `uv` for
  provider dependency metadata, which needs the Breeze/CI environment) — please run
  `breeze testing tests airflow-core/tests/unit/timetables/test_interval_timetable.py` before
  merging

### Open question for maintainers
The issue outlines two possible directions: (1) extend this flag to cover `run_id` (what this PR
does), or (2) a more general `dag_run_policy`-style cluster policy hook, noted in the issue as
possibly the preferred long-term design. This PR takes the smaller-scoped option since it's a pure
addition with low regression risk, but happy to rework toward the hook approach if that's the
preferred direction.

One incidental finding while researching this: `[scheduler] create_delta_data_intervals` is
documented in `config.yml` and listed in the CLI config commands, but I couldn't find anywhere in
`_create_timetable()` (or elsewhere) that actually reads it to gate `DeltaDataIntervalTimetable`
selection — `create_cron_data_intervals` appears to gate both cron and delta branches today.
Flagging in case that's a known issue or I'm missing a call site; didn't want to fix it as a
drive-by in this PR.

Newsfragment will be added as a follow-up commit once this PR has a number, per the contribution
guide.

---

**AI disclosure:** All code, documentation, and test changes in this PR were written by me. An AI
coding tool (Claude) was used only PR description and to run the new regression test cases locally and verify they
pass against the patched code.

### DIFF
--- a/airflow-core/docs/authoring-and-scheduling/timetable.rst
+++ b/airflow-core/docs/authoring-and-scheduling/timetable.rst
@@ -345,6 +345,12 @@ The time when a Dag run is triggered
 Both trigger and data interval timetables trigger Dag runs at the same time. However, the timestamp for the
 ``run_id`` is different for each. This is because ``run_id`` is based on ``logical_date``.
 
+.. note::
+    For **CronDataIntervalTimetable** and **DeltaDataIntervalTimetable**, ``run_id`` anchoring to
+    ``logical_date`` requires ``[scheduler] create_cron_data_intervals`` to be enabled — this single
+    flag currently gates both. Without it, Dags default to a trigger timetable, whose ``run_id`` is
+    anchored to ``run_after`` instead.
+
 For example, suppose there is a cron expression ``@daily`` or ``0 0 * * *``, which is scheduled to run at 12AM every day. If you enable Dags using the two timetables at 3PM on January
 31st,
 
--- a/airflow-core/docs/installation/upgrading_to_airflow3.rst
+++ b/airflow-core/docs/installation/upgrading_to_airflow3.rst
@@ -384,6 +384,11 @@ These include:
   ``create_cron_data_intervals=True`` explicitly to keep ``CronDataIntervalTimetable``.
   If you don't, the new ``False`` default is fine.
 
+  Setting ``create_cron_data_intervals=True`` also anchors the scheduled ``run_id`` to
+  ``logical_date``, matching Airflow 2's ``run_id`` semantics. If your workflows parse
+  ``run_id`` and assume it reflects the logical date, set this flag rather than relying on
+  the ``run_after``-anchored default.
+
   Set this **before** the upgrade. If you instead change the flag after some
   Airflow 3 dagruns already exist (going
   ``CronTriggerTimetable`` -> ``CronDataIntervalTimetable``), one scheduled run
--- a/airflow-core/src/airflow/config_templates/config.yml
+++ b/airflow-core/src/airflow/config_templates/config.yml
@@ -2851,6 +2851,11 @@ scheduler:
         try to schedule, while for **CronDataIntervalTimetable**, the logical date is the beginning of
         the data interval, but the DAG Run will try to schedule at the end of the data interval.
 
+        Enabling this setting also anchors the scheduled ``run_id`` to ``logical_date`` (the data
+        interval start), matching the Airflow 2 ``run_id`` semantics this flag is meant to restore.
+        Deployments that leave this setting at its default (``False``) are unaffected; their ``run_id``
+        continues to be anchored to ``run_after``
+
         When a DAG is switched from **CronTriggerTimetable** to **CronDataIntervalTimetable** (for example,
         by flipping this setting from ``False`` to ``True``), the next scheduled run skips one period past
         the most recent **CronTriggerTimetable** run to avoid colliding with its logical date.
--- a/airflow-core/src/airflow/timetables/interval.py
+++ b/airflow-core/src/airflow/timetables/interval.py
@@ -29,6 +29,7 @@ from airflow.timetables.base import DagRunInfo, DataInterval, Timetable
 
 if TYPE_CHECKING:
     from airflow.timetables.base import TimeRestriction
+    from airflow.utils.types import DagRunType
 
 Delta = datetime.timedelta | relativedelta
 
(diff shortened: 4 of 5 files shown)


uranusjr:
Please refer to the original issue for discussions.

bujjibabukatta:
> Please refer to the original issue for discussions.

Thank you for the pointer. I've reviewed the original issue #71187 and implemented the fix with the correct scope.
Key clarification: The issue title specifically mentions "when create_cron_data_intervals is enabled" — this flag controls which timetable class is instantiated:

When enabled: CronDataIntervalTimetable / DeltaDataIntervalTimetable
When disabled (default): CronTriggerTimetable / DeltaTriggerTimetable
The initial implementation placed generate_run_id in the shared _DataIntervalTimetable base class, which incorrectly applied the logical_date anchoring to both cron and delta timetables.
Updated implementation (latest commit):

Moved generate_run_id override only to CronDataIntervalTimetable — this correctly limits the behavior to cron expressions when the flag is enabled> - DeltaDataIntervalTimetable (used for @daily, @hourly, timedelta()) now correctly retains AIP-76 run_after anchoring> - Added test test_generate_run_id_uses_run_after_for_delta_and_default_cron to verify this> - CronTriggerTimetable (AIP-76 default) remains unaffected as verified by test_generate_run_id_default_timetable_unaffected
This aligns with the factory logic in _create_timetable (airflow/sdk/definitions/dag.py) where the config flag only controls cron vs trigger timetable selection.

PR #71487, opened 2026-08-12
Chart: Default airflow version to 3.3.1

vatsrahul1001:
Update the default Airflow image used by the Helm chart to 3.3.1 (from 3.3.0),


kaxil (approved):

amoghrajesh (approved):

Miretpl:
Hi @vatsrahul1001, @kaxil, @amoghrajesh, should I update a different file than https://github.com/apache/airflow/blob/main/dev/README_RELEASE_AIRFLOW.md#update-default-airflow-version-in-the-helm-chart to include Helm Chart 1.2x during the Airflow release process? I've updated the mentioned document after 3.3.0, when it was omitted, and #70046 was created.

potiuk:
> Hi @vatsrahul1001, @kaxil, @amoghrajesh, should I update a different file than https://github.com/apache/airflow/blob/main/dev/README_RELEASE_AIRFLOW.md#update-default-airflow-version-in-the-helm-chart to include Helm Chart 1.2x during the Airflow release process? I've updated the mentioned document after 3.3.0, when it was omitted, and #70046 was created.

I would rather say that Helm Chart release process should be updated to bump latest airflow version (it happens much less frequently than airflow release) - and cherry-picked to chart 1.2 branch.

Miretpl:
> I would rather say that Helm Chart release process should be updated to bump latest airflow version (it happens much less frequently than airflow release) - and cherry-picked to chart 1.2 branch.

It is, but the case is that we might see issues in the Helm Chart CI right before release for an unknown number of Airflow versions if we change the order, as we will not test against the current latest version. In the end, it might postpone the release of the Helm Chart. Because of that, I would prefer to stay with the current setup, just including the tag. I can take care of merging all PRs into the 1.2x line.

PR #71492, opened 2026-08-12
Make KPO durable execution inert below Airflow 3.3

amoghrajesh:
---

### Summary

`durable`, KPO's crash-recovery parameter, currently behaves identically on every supported Airflow version -- including below 3.3, where the task state store it's meant to rely on doesn't exist. This PR makes `durable` a genuine no-op below 3.3 and fixes the scope of its deprecation warning.

### Why

Raised on this PR by elad:

> Durable is functionality that works only in 3.3+ while we made the parameter working for earlier versions. I'd like to avoid a case where users of 3.1 expect durable functionality just because of the parameter name.

Before this change, `durable=True` on a pre-3.3 install fell back to the same label-search reattach mechanism `reattach_on_restart` always used -- so the parameter technically "worked," but under a name that implies a stronger, task-state-store-backed guarantee it can't actually deliver there. Below 3.3, that's a coincidence of implementation, not a documented contract.

Separately, the deprecation warning for `reattach_on_restart` was gated to fire only on Airflow 3.3+. Provider deprecations are scoped to this provider's own release schedule, not the caller's Airflow version -- the parameter will be removed in a future amazon-provider-style release regardless of which Airflow version anyone runs. Gating the warning meant a 3.0-3.2 user could stay silent right up until removal, then hit a `TypeError` with no warning.

### What changed

- `durable` is fully inert below Airflow 3.3: setting it explicitly (`True` or `False`) only emits a `UserWarning` that it has no effect there, and its value is otherwise ignored. `reattach_on_restart` (default `True`, unchanged) is the only lever on those versions, using the same label-search mechanism it always has -- no behavior change for anyone who never touches either flag.
- `reattach_on_restart`'s deprecation warning now fires on **every** supported Airflow version, not just 3.3+. The message itself is version-aware: below 3.3 it points at a future upgrade ("On Airflow 3.3+, use `durable` instead"), since telling that user to switch right now would be bad advice.
- On Airflow 3.3+, `durable` now wins outright when explicitly set -- including `durable=False` beating a conflicting `reattach_on_restart=True` -- rather than `reattach_on_restart` always winning as it did before. `reattach_on_restart` only fills in when `durable` itself was never touched.
- Both messages state the removal is tied to this provider's minimum supported Airflow version reaching 3.3, not to the caller's own version.

### Behavior

See the combination table below. Full test coverage added in `TestKubernetesPodOperatorDurableBelow3_3` (new) and updated in the existing `TestKubernetesPodOperatorDurableExecution` deprecation tests.

---

## Behavior for every combination

`D` = `durable` param as passed, `R` = `reattach_on_restart` param as passed. Both `self.durable` and `self.reattach_on_restart` always end up equal (one mirrors the other).

### Airflow 3.3+

| `D` | `R` | Result | Warnings |
|---|---|---|---|
| not set | not set | `True` (default) | none |
| not set | `True` | `True` | deprecation |
| not set | `False` | `False` | deprecation |
| `True` | not set | `True` | none |
| `True` | `True` | `True` | deprecation |
| `True` | `False` | **`True`** (`D` wins over conflicting `R`) | deprecation |
| `False` | not set | `False` | none |
| `False` | `True` | **`False`** (`D` wins over conflicting `R`) | deprecation |
| `False` | `False` | `False` | deprecation |

### Airflow below 3.3

| `D` | `R` | Result | Warnings |
|---|---|---|---|
| not set | not set | `True` (old default, unchanged) | none |
| not set | `True` | `True` | deprecation |
| not set | `False` | `False` | deprecation |
| `True` | not set | `True` (coincidentally matches) | `durable` inert |
| `True` | `True` | `True` | deprecation + inert |
| `True` | `False` | `False` (`D`'s `True` ignored) | deprecation + inert |
| `False` | not set | **`True`** (`D`'s `False` discarded) | `durable` inert |
| `False` | `True` | `True` | deprecation + inert |
| `False` | `False` | `False` | deprecation + inert |

**Rule of thumb:** on 3.3+, `durable` governs and wins on conflict, defaulting to `True`. Below 3.3, `durable`'s value never matters at all -- the result is always computed purely from `reattach_on_restart` (`True` if untouched, else its value) -- and touching `durable` there always draws a warning that it did nothing.

---

### DIFF
--- a/providers/cncf/kubernetes/docs/operators.rst
+++ b/providers/cncf/kubernetes/docs/operators.rst
@@ -216,9 +216,11 @@ To always create a fresh pod on retry rather than reattaching, set ``durable=Fal
         durable=False,
     )
 
-Durable execution requires Airflow 3.3 or newer, since it relies on the task state store. On
-earlier Airflow versions ``durable=True`` (the default) falls back to the same label-search
-reattach behavior this operator has always used.
+Durable execution requires Airflow 3.3 or newer, since it relies on the task state store. Below
+3.3, ``durable`` has no effect at all: setting it explicitly only emits a warning, and its value is
+ignored either way. The deprecated ``reattach_on_restart`` parameter (default ``True``) is the
+only lever there, and it falls back to the same label-search reattach behavior this operator has
+always used -- unchanged from before this feature existed.
 
 The pod identity persisted in task state store isn't deleted automatically, that only happens
 when someone runs ``airflow state-store clean``. If a task's ``retry_delay`` is longer than
@@ -229,9 +231,11 @@ search can often still find the same pod, but it loses the unambiguous reconnect
 exposure to ``FoundMoreThanOnePodFailure`` if a genuine duplicate pod exists by then. Avoid
 running cleanup on a schedule shorter than your longest ``retry_delay``.
 
-``durable`` supersedes the deprecated ``reattach_on_restart`` parameter -- passing
-``reattach_on_restart`` still works but emits ``AirflowProviderDeprecationWarning`` (on Airflow
-3.3+) and maps its value onto ``durable``.
+``durable`` supersedes the deprecated ``reattach_on_restart`` parameter on Airflow 3.3+, where
+passing ``reattach_on_restart`` still works and maps its value onto ``durable``. Below 3.3,
+``reattach_on_restart`` remains the only working option, since ``durable`` is a no-op there.
+Either way, passing it emits an ``AirflowProviderDeprecationWarning``, since the parameter will be
+removed once this provider's minimum supported Airflow version reaches 3.3.
 
 How does XCom work?
 ^^^^^^^^^^^^^^^^^^^
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
@@ -86,7 +86,7 @@ from airflow.providers.cncf.kubernetes.utils.pod_manager import (
     PodPhase,
     detect_pod_terminate_early_issues,
 )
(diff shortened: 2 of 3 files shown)


eladkal:
> durable is a functional drop-in for reattach_on_restart on every supported version,

Is it? If I run Airflow 3.1 I won't have the durable parameter.

amoghrajesh:
@eladkal durable is a plain, unconditional parameter on `KubernetesPodOperator.__init__`- no version check guards it, so functionally or documentation wise its not suggested but it's can potentially be provided.

So in practice, `durable=True` on Airflow 3.1 does exactly what `reattach_on_restart=True` did

amoghrajesh:
Let me move this to draft, I am clearing up some confusion in my head re this too.

eladkal:
> So in practice, durable=True on Airflow 3.1 does exactly what reattach_on_restart=True did

I am a bit histated here. Durable is functionality that works only in 3.3+ while we made the parameter working for earlier versions. I'd like to avoid case where users of 3.1 expect durable functionality just because of the parameter name.

amoghrajesh:
@eladkal thanks for the concern, its totally fair. I gave it some thought and fixed it properly now.

`durable` is now inert below Airflow 3.3, ie: setting it explicitly (True or False) only emits a warning that it has no effect there, and its value is otherwise ignored. `reattach_on_restart` (still defaulting to True, unchanged from before) is the only thing that controls the older / label search behaviour below 3.3. 

So `durable` as a name is now reserved for the real, it is task-state-store based solution that _only exists on 3.3+_, it can't accidentally imply that on an earlier version. `reattach_on_restart` itself still emits the deprecation warning on every version, since its removal is tied to this provider's release schedule, not the caller's Airflow version and the warning message is just version aware about what to do next.

eladkal (approved):

PR #71495, opened 2026-08-12
Fail Glue job tasks stopped mid-run in deferrable verbose mode

### LINKED ISSUE #71490: GlueJobCompleteTrigger's verbose branch still treats STOPPED as success, disagreeing with the waiter and the operator
`GlueJobOperator`'s handling of a Glue job run ending in `STOPPED` (e.g. cancelled manually from the AWS console, or by this operator's own `on_kill`) is inconsistent across code paths -- three different answers exist for the same state today:

| Path | Where | `STOPPED` classified as | Result |
|---|---|---|---|
| Deferrable, non-verbose | `providers/amazon/src/airflow/providers/amazon/aws/waiters/glue.json`, waiter `job_complete` | Explicit `"state": "failure"` acceptor | Task fails |
| Deferrable, verbose | `GlueJobCompleteTrigger.run()`'s hand-rolled polling loop (`providers/amazon/src/airflow/providers/amazon/aws/triggers/glue.py`) | Bundled with `SUCCEEDED`: `if job_run_state in ("SUCCEEDED", "STOPPED"): ... "status": "success"` | Task succeeds |
| Synchronous | `GlueJobOperator.poll_until_complete` (`JOB_RUN_SUCCESS_STATES = ("SUCCEEDED",)`) | Same bucket as `FAILED`/`TIMEOUT` | Task fails |

The synchronous path was made to agree with the deferrable non-verbose path's waiter as part of #71211 (closing a "self-inflicted stop silently reported as success" gap). The deferrable *verbose* path was not touched and still disagrees with both -- it's a separate, pre-existing bug in the trigger's custom polling loop, not something #71211 introduced.

Fix direction: make the verbose branch of `GlueJobCompleteTrigger.run()` treat `STOPPED` as a failure, matching the waiter used by the non-verbose branch and the operator's own synchronous handling.

Raised during review of #71211: https://github.com/apache/airflow/pull/71211#discussion_r3764425797

rjgoyln:
## Summary

A Glue job run ending in `STOPPED` — cancelled from the AWS console, or by the operator's own `on_kill` — fails the task on every path but this one, where the verbose polling loop buckets it with `SUCCEEDED` and lets downstream work run on output the run never produced. The operator docs already state the intended behaviour without qualification: a run that ends in `STOPPED` is treated as a failure, not a success.

## Scope

`GlueJobHook._handle_state` still counts `STOPPED` as finished and is left alone, since #71211 moved the operator's success decision out of the hook. `GlueJobSensor` defers to the same trigger, so its verbose deferrable mode changes too, toward its own `FAILURE_STATES`.

closes: #71490

---

### DIFF
--- a/providers/amazon/tests/unit/amazon/aws/triggers/test_glue.py
+++ b/providers/amazon/tests/unit/amazon/aws/triggers/test_glue.py
@@ -174,7 +174,7 @@ class TestGlueJobTrigger:
     @pytest.mark.parametrize("job_run_state", ["FAILED", "TIMEOUT", "STOPPED"])
     @mock.patch.object(AwsLogsHook, "get_async_conn")
     @mock.patch.object(GlueJobHook, "get_async_conn")
-    async def test_verbose_run_job_failed(self, mock_glue_conn, mock_logs_conn, job_run_state):
+    async def test_verbose_run_failure_states(self, mock_glue_conn, mock_logs_conn, job_run_state):
         """When verbose=True and the run ends in a failure state, the trigger yields an error event."""
         glue_client = AsyncMock()
         glue_client.get_job_run = AsyncMock(
--- a/providers/amazon/tests/unit/amazon/aws/triggers/test_glue.py
+++ b/providers/amazon/tests/unit/amazon/aws/triggers/test_glue.py
@@ -171,13 +171,14 @@ class TestGlueJobTrigger:
         assert logs_client.get_log_events.call_count >= 2
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize("job_run_state", ["FAILED", "TIMEOUT", "STOPPED"])
     @mock.patch.object(AwsLogsHook, "get_async_conn")
     @mock.patch.object(GlueJobHook, "get_async_conn")
-    async def test_verbose_run_job_failed(self, mock_glue_conn, mock_logs_conn):
-        """When verbose=True and the job fails, the trigger yields an error event."""
+    async def test_verbose_run_job_failed(self, mock_glue_conn, mock_logs_conn, job_run_state):
+        """When verbose=True and the run ends in a failure state, the trigger yields an error event."""
         glue_client = AsyncMock()
         glue_client.get_job_run = AsyncMock(
-            return_value={"JobRun": {"JobRunState": "FAILED", "LogGroupName": "/aws-glue/python-jobs"}}
+            return_value={"JobRun": {"JobRunState": job_run_state, "LogGroupName": "/aws-glue/python-jobs"}}
         )
         mock_glue_conn.return_value.__aenter__ = AsyncMock(return_value=glue_client)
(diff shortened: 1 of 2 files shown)


henry3260 (approved):
`ERROR` is the one state where the verbose branch still disagrees with the `job_complete`
waiter that the non-verbose branch uses. The waiter has an explicit
`{"expected": "ERROR", "state": "failure"}` acceptor and the operator's
`JOB_RUN_TERMINAL_STATES` lists it, but this loop classifies only
`FAILED`/`TIMEOUT`/`STOPPED` as failure and `SUCCEEDED` as success, so an `ERROR` run falls
through to "keep polling" and only surfaces after `waiter_max_attempts` (75 × 60s by default
for `GlueJobOperator`, 60 × 120s for `GlueJobSensor`) as `waiter exceeded max attempts` which reads as a timeout rather than as a job error.

It is worth to open a pr to a follow-up. Do you want to take it?

providers/amazon/tests/unit/amazon/aws/triggers/test_glue.py
```diff
@@ -171,13 +171,14 @@ async def test_verbose_run_success(self, mock_glue_conn, mock_logs_conn):
         assert logs_client.get_log_events.call_count >= 2
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize("job_run_state", ["FAILED", "TIMEOUT", "STOPPED"])
     @mock.patch.object(AwsLogsHook, "get_async_conn")
     @mock.patch.object(GlueJobHook, "get_async_conn")
-    async def test_verbose_run_job_failed(self, mock_glue_conn, mock_logs_conn):
-        """When verbose=True and the job fails, the trigger yields an error event."""
+    async def test_verbose_run_job_failed(self, mock_glue_conn, mock_logs_conn, job_run_state):
```
henry3260:
```suggestion
    async def test_verbose_run_failure_states(self, mock_glue_conn, mock_logs_conn, job_run_state):
```

rjgoyln:
> `ERROR` is the one state where the verbose branch still disagrees with the `job_complete` waiter that the non-verbose branch uses. The waiter has an explicit `{"expected": "ERROR", "state": "failure"}` acceptor and the operator's `JOB_RUN_TERMINAL_STATES` lists it, but this loop classifies only `FAILED`/`TIMEOUT`/`STOPPED` as failure and `SUCCEEDED` as success, so an `ERROR` run falls through to "keep polling" and only surfaces after `waiter_max_attempts` (75 × 60s by default for `GlueJobOperator`, 60 × 120s for `GlueJobSensor`) as `waiter exceeded max attempts` which reads as a timeout rather than as a job error.
> 
> It is worth to open a pr to a follow-up. Do you want to take it?

Great catch! I'd be happy to take care of it. I will open a follow-up PR to handle the ERROR state shortly.
Thanks for your review! Really appreciate it.

PR #71503, opened 2026-08-12
Show the teams of the signed-in user in multi-team mode

vincbeck:
When an Airflow environment runs in multi-team mode, users had no way to see which teams they belong to.

`GET /ui/auth/me` now returns a `teams` field, and the UI lists those teams in the user menu below the user name. The field is `null`, and the UI section hidden, when the environment is not in multi-team mode, so nothing changes for single-team deployments and auth managers that don't support teams are never queried.

## Screenshot

Screenshot 2026-08-12 at 11 01 06 AM

### Long name

Screenshot 2026-08-12 at 11 39 45 AM

### More than 5 teams

Screenshot 2026-08-12 at 11 51 46 AM

 

---

---

### DIFF
--- a/airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
+++ b/airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { Box, Icon, useDisclosure } from "@chakra-ui/react";
+import { Box, HStack, Icon, useDisclosure, VStack } from "@chakra-ui/react";
 import { useTranslation } from "react-i18next";
 import {
   FiKey,
@@ -33,7 +33,7 @@ import {
 } from "react-icons/fi";
 
 import { useAuthLinksServiceGetCurrentUserInfo } from "openapi/queries";
-import { Menu } from "src/components/ui";
+import { Menu, Tooltip } from "src/components/ui";
 import { RouterLink } from "src/components/ui/RouterLink";
 import { useColorMode } from "src/context/colorMode/useColorMode";
 import type { NavItemResponse } from "src/utils/types";
@@ -43,6 +43,9 @@ import LogoutModal from "./LogoutModal";
 import { NavButton } from "./NavButton";
 import { PluginMenuItem } from "./PluginMenuItem";
 import TokenGenerationModal from "./TokenGenerationModal";
+
+// Beyond a handful of teams the list would push the menu items below the fold.
+const MAX_VISIBLE_TEAMS = 5;
 
 const COLOR_MODES = {
   DARK: "dark",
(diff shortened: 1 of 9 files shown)


airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
```diff
@@ -97,6 +100,31 @@ export const UserSettingsButton = ({ externalViews }: { readonly externalViews:
                 <Box fontSize="md" fontWeight="semibold">
                   {`${currentUser.username} (id: ${currentUser.id})`}
                 </Box>
+                {Array.isArray(currentUser.teams) ? (
```
pierrejeambrun, line 103: Why `Array.isArray` ? Can this be something else than an array?

airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
```diff
@@ -44,6 +44,9 @@ import { NavButton } from "./NavButton";
 import { PluginMenuItem } from "./PluginMenuItem";
 import TokenGenerationModal from "./TokenGenerationModal";
 
+// Beyond a handful of teams the list would push the menu items below the fold.
+const MAX_VISIBLE_TEAMS = 5;
```
pierrejeambrun, line 48: Can you provide a screenshot with both 5 team names, as well as super long team names such as.

"theteamwithasuperlongnameeeeeeeeeeeeeeeeeeeeeeeeeee" or something like it.

pierrejeambrun: Also with team name > 5

pierrejeambrun (commented):
Nice, looking good overall. 

Just a few questions / suggestions.


Also it means that it's impossible to see all of your teams ? (If you have more than 5, maybe a clickable 'page' to get details could be a nice follow up.

airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
vincbeck: It can be null if the environment is not multi-team

airflow-core/src/airflow/ui/src/layouts/Nav/UserSettingsButton.tsx
vincbeck: Added in the description

vincbeck:
> Nice, looking good overall.
> 
> Just a few questions / suggestions.
> 
> Also it means that it's impossible to see all of your teams ? (If you have more than 5, maybe a clickable 'page' to get details could be a nice follow up.

Good point. I updated the code to display the additional teams in a tooltip. Let me know what you think

bbovenzi (approved):

PR #71518, opened 2026-08-12
Operator extra links per attempt (draft A): a row per attempt in XCom

1fanwang:
**TL;DR — operator extra links don't work per attempt** (#71471). After a retry the button
under attempt 1 opens attempt 2's logs. It hits every bundled provider linking to per-attempt
logs: EMR, Glue, Databricks, Dataproc, Livy.

Draft **A of three** for the same bug — pick a shape and I'll finish it and close the others.
**Recommended: C** (#71522). See the issue for why; A is the fallback if this needs to ship on the 3.2 line.

| | Approach | Trade |
|---|---|---|
| **A** | **this PR — a row per attempt in XCom, exempt from the clear** | **smallest; core owns a key convention for every link** |
| B | #71521 — the link opts in | no extra rows when unneeded; changes the serialized DAG format |
| C | #71522 — the task state store | nothing about the clear moves; 3.3+ only |

## How

The worker already caches each rendered link as an XCom row under the link's `xcom_key`. It
now also writes one under a key carrying the attempt, the reader prefers that row and falls
back to the bare key, and those rows survive the clear a retry issues.

The prefix is owned by Airflow rather than derived from `xcom_key`, because a link may
override that with any name — Databricks uses `databricks_job_run_link`, `BaseGoogleLink`
proxies to `self.key`.

## Testing

Red to green on a real metastore, rows present for both attempts.

Before, on 3.2.2 — attempt 1 resolves to attempt 2's URL:

```
$ AIRFLOW__DATABASE__SQL_ALCHEMY_CONN=sqlite:///af.db python read_test.py
try 1 -> https://logs/attempt-2
try 2 -> https://logs/attempt-2
```

After:

```
$ AIRFLOW__DATABASE__SQL_ALCHEMY_CONN=sqlite:///af.db python read_test.py
try 1 -> https://logs/attempt-1
try 2 -> https://logs/attempt-2
```

Purge predicate, including keys that would be spared if the `_` were left as a LIKE wildcard:

```
spared : ['_link_attempt_1__link_Foo', '_link_attempt_2_databricks_job_run_link']
cleared: ['_link_Foo', 'databricks_job_run_link', 'return_value',
          'Xlink_attemptX1_y', 'my_link_attempt_1_thing']
```

</details>

Tests cover the reader preferring the requested attempt, the fallback, and the run endpoint
keeping those rows out of `xcom_keys_to_clear`. The full suite runs in CI.

### DIFF
--- a/airflow-core/newsfragments/71471.bugfix.rst
+++ b/airflow-core/newsfragments/71471.bugfix.rst
@@ -1,0 +1,1 @@
+Return each attempt's own URL from operator extra links, instead of the latest attempt's.
--- a/airflow-core/newsfragments/71471.bugfix.rst
+++ b/airflow-core/newsfragments/71471.bugfix.rst
@@ -1,1 +1,0 @@
-Return each attempt's own URL from operator extra links, instead of the latest attempt's.
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -94,6 +94,7 @@ from airflow.models.taskinstancehistory import TaskInstanceHistory as TIH
 from airflow.models.taskreschedule import TaskReschedule
 from airflow.models.trigger import Trigger, handle_event_submit
 from airflow.models.xcom import XComModel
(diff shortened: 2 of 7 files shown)


1fanwang:
Going with #71522 instead: the task state store is where cross-retry state already lives, so the XCom clear stays absolute rather than gaining an exemption. Closing this one.

PR #71521, opened 2026-08-12
Operator extra links per attempt (draft B): the link opts in

1fanwang:
**TL;DR — operator extra links don't work per attempt** (#71471). After a retry the button
under attempt 1 opens attempt 2's logs. It hits every bundled provider linking to per-attempt
logs: EMR, Glue, Databricks, Dataproc, Livy.

Draft **B of three** for the same bug — pick a shape and I'll finish it and close the others.
**Recommended: C** (#71522). See the issue for why; A is the fallback if this needs to ship on the 3.2 line.

| | Approach | Trade |
|---|---|---|
| A | #71518 — a row per attempt in XCom | smallest; core owns a key convention for every link |
| **B** | **this PR — the link opts in** | **no extra rows when unneeded; changes the serialized DAG format** |
| C | #71522 — the task state store | nothing about the clear moves; 3.3+ only |

## How

`BaseOperatorLink` gains `keeps_a_link_per_attempt` and `xcom_key_for_try`. A link that sets
the flag gets a row per attempt, exempt from the retry clear; everything else behaves as it
does today.

The flag has to survive DAG serialization, and that is the cost. The api-server never sees a
provider's link class — it rebuilds every link as `XComOperatorLink` from a `{name: xcom_key}`
mapping carrying a string and nothing else. That mapping now carries
`{"key": ..., "per_attempt": true}` for opted-in links and stays a bare string otherwise, so
existing serialized DAGs read back unchanged.

## Testing

Read path against a real metastore on 3.2.2, rows present for both attempts, asking an
opted-in link and a plain one.

```
$ AIRFLOW__DATABASE__SQL_ALCHEMY_CONN=sqlite:///af.db python b_test.py
opted-in  try 1 -> https://logs/attempt-1
opted-in  try 2 -> https://logs/attempt-2
not opted try 1 -> https://logs/attempt-2
```

Line 3 is the before state on the same path: without the flag, attempt 1 still returns the
latest attempt's URL. Lines 1 and 2 are the same read with it set.

Key derivation:

```
$ python -c "from airflow.serialization.definitions.operatorlink import XComOperatorLink; ..."
default  try1 -> _link_X                 | try2 -> _link_X
opted-in try1 -> _link_attempt_1__link_X | try2 -> _link_attempt_2__link_X
```

Serialization round-trip, the risky part:

```
encoded: {'plain': '_link_P', 'opted': {'key': '_link_O', 'per_attempt': True}}
  plain: key=_link_P per_attempt=False  (legacy string, unchanged)
  opted: key=_link_O per_attempt=True
```

</details>

### DIFF
--- a/airflow-core/newsfragments/71471.bugfix.rst
+++ b/airflow-core/newsfragments/71471.bugfix.rst
@@ -1,0 +1,1 @@
+Return each attempt's own URL from operator extra links, for links that opt in.
--- a/airflow-core/newsfragments/71471.bugfix.rst
+++ b/airflow-core/newsfragments/71471.bugfix.rst
@@ -1,1 +1,0 @@
-Return each attempt's own URL from operator extra links, for links that opt in.
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -94,6 +94,7 @@ from airflow.models.taskinstancehistory import TaskInstanceHistory as TIH
 from airflow.models.taskreschedule import TaskReschedule
 from airflow.models.trigger import Trigger, handle_event_submit
 from airflow.models.xcom import XComModel
+from airflow.sdk.bases.operatorlink import ATTEMPT_LINK_XCOM_KEY_PREFIX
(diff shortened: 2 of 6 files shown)


1fanwang:
Going with #71522 instead: the task state store is where cross-retry state already lives, so the XCom clear stays absolute rather than gaining an exemption. Closing this one.

PR #71523, opened 2026-08-13
Stop presenting multi-team tagged metrics as an exhaustive list

ramitkataria:
The list read as an authoritative enumeration of every metric carrying a team_name tag, so it drifts out of date as metrics change and misleads anyone building dashboards or alerts from it. A stale entry naming a callback failure counter that is never emitted is the symptom that surfaced the problem.

Readers need to know which areas of Airflow carry the tag; the emission sites in the source are the right authority for individual metric names.

 

---

- [X] Yes (Claude Opus 5)

---

### DIFF
--- a/airflow-core/docs/core-concepts/multi-team.rst
+++ b/airflow-core/docs/core-concepts/multi-team.rst
@@ -1063,26 +1063,24 @@ Dags, and global components emit the same metrics without a ``team_name`` tag.
     When Multi-Team mode is disabled, metrics are emitted with no ``team_name`` tag whatsoever, exactly
     as they have always been emitted for a single-team Airflow environment.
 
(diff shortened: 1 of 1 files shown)


o-nikolas (commented):
I've never really loved this being an authoritative and exhaustive source of the metrics which are team-ified. I think the best place is to just look at the code (easy task for any modern agent). We're just going to be stuck updating these docs constantly as they inevitably become out of date as changes happen. Can we just this to just mention a handful of metrics and then a "..., etc."? Or maybe have a `Note` that says this isn't an exhaustive list?

ramitkataria:
That makes sense. I've updated the PR

vincbeck (approved):

uros-b (approved):

o-nikolas (approved):

PR #71527, opened 2026-08-13
Build lang-SDK k8s test artifacts from the branch when targeting main

jason810496:
## Why

The k8s lang-SDK test builds its Go bundle and Java jar from upstream `main`'s `go-sdk`/`java-sdk` unconditionally, no matter which branch is checked out or which branch the run targets.

Example fail case: #71057 renames `Dag` to `DagDef` updating `K8sBundleBuilder.java` to match made `compileJava` fail. Since the k8s test still pull from the `upstream/main` instead of respecting the current branch change.

## What

This PR scopes the upstream-main fallback to the case it was written for:

| Target branch | Go/Java SDK sources |
| --- | --- |
| `main` | the checked-out branch's own `go-sdk/` and `java-sdk/` |
| anything else (`v3-*-test`, …) | upstream `main`, as before |

---

- [X] Yes — Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
@@ -2515,10 +2515,14 @@ LANG_SDK_AWS_CONN_URI = (
 # sources. See kubernetes-tests/lang_sdk/README.md.
 LANG_SDK_UPSTREAM_GIT_URL = "https://github.com/apache/airflow.git"
 LANG_SDK_UPSTREAM_REF = "main"
+# The target branch for which the lang-SDK artifacts are built from this checkout's own SDK
+# sources rather than fetched from upstream. Distinct from LANG_SDK_UPSTREAM_REF (the ref fetched
+# from upstream) even though both are "main" today -- one names a git ref, the other a sentinel.
+LANG_SDK_LOCAL_SOURCE_BRANCH = "main"
 
 
 def _lang_sdk_target_branch() -> str:
-    """Branch this run targets. GITHUB_BASE_REF for PRs, DEFAULT_BRANCH in CI, else this checkout's."""
+    """Branch this run targets: GITHUB_BASE_REF or DEFAULT_BRANCH if set, else this checkout's."""
     return os.environ.get("GITHUB_BASE_REF") or os.environ.get("DEFAULT_BRANCH") or AIRFLOW_BRANCH
 
 
@@ -2531,13 +2535,13 @@ def _lang_sdk_resolve_sdk_sources(staging: Path, output: Output | None) -> tuple
     any SDK rename breaks the build. Runs targeting anything else fall back to upstream main.
     """
     target = _lang_sdk_target_branch()
-    if target == LANG_SDK_UPSTREAM_REF:
+    if target == LANG_SDK_LOCAL_SOURCE_BRANCH:
         get_console(output=output).print(
             f"[info]Run targets {target}: building the lang-SDK Go/Java artifacts from this branch"
         )
         return AIRFLOW_ROOT_PATH / "go-sdk", AIRFLOW_ROOT_PATH / "java-sdk"
     get_console(output=output).print(
-        f"[info]Run targets {target}, not {LANG_SDK_UPSTREAM_REF}: building the lang-SDK Go/Java "
+        f"[info]Run targets {target}, not {LANG_SDK_LOCAL_SOURCE_BRANCH}: building the lang-SDK Go/Java "
         f"artifacts from upstream {LANG_SDK_UPSTREAM_REF}"
     )
     return _lang_sdk_fetch_upstream_sdk_sources(staging, output)
--- a/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
@@ -32,6 +32,7 @@ from typing import Any
 import click
 import yaml
 
+from airflow_breeze.branch_defaults import AIRFLOW_BRANCH
 from airflow_breeze.commands.common_options import (
     option_answer,
     option_debug_resources,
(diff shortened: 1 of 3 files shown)


dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
```diff
@@ -2509,12 +2510,39 @@ def deploy_cluster(
...
+    are what the k8s test exercises -- without this, ``java_example``/``go_example`` (harness code
+    that tracks the branch) is compiled against a different SDK than the branch it belongs to, and
+    any SDK rename breaks the build. Runs targeting anything else fall back to upstream main.
+    """
+    target = _lang_sdk_target_branch()
+    if target == LANG_SDK_UPSTREAM_REF:
+        get_console(output=output).print(
+            f"[info]Run targets {target}: building the lang-SDK Go/Java artifacts from this branch"
+        )
+        return AIRFLOW_ROOT_PATH / "go-sdk", AIRFLOW_ROOT_PATH / "java-sdk"
```
uranusjr: `LANG_SDK_UPSTREAM_REF` is doing double duty; it's "the ref to fetch from the upstream remote" and, via this comparison, "the sentinel meaning use local sources." I would separate the two semantics; this one here can maybe be `LANG_SDK_LOCAL_SOURCE_BRANCH`.

uranusjr (approved):

dev/breeze/tests/test_kubernetes_lang_sdk_commands.py
```diff
@@ -374,3 +376,56 @@ def fake_parallel(steps, output):
...
+        [
+            pytest.param({"GITHUB_BASE_REF": "main"}, "main", id="pr-targeting-main"),
+            pytest.param({"GITHUB_BASE_REF": "v3-3-test"}, "v3-3-test", id="pr-targeting-release"),
+            pytest.param({"DEFAULT_BRANCH": "v3-3-test"}, "v3-3-test", id="ci-default-branch"),
+            pytest.param(
+                {"GITHUB_BASE_REF": "main", "DEFAULT_BRANCH": "v3-3-test"},
+                "main",
+                id="pr-target-wins-over-default-branch",
+            ),
+            pytest.param({}, "main", id="falls-back-to-this-checkouts-branch"),
```
potiuk: With both env vars unset `_lang_sdk_target_branch()` returns `AIRFLOW_BRANCH`, which is `"v3-3-test"` on the branch this PR is labelled to backport to — so this case asserts `"main"` against `"v3-3-test"` and fails there. The test file's pre-image blob on `v3-3-test` is identical to main's, so the hunk cherry-picks cleanly and lands the assertion verbatim.

`branch_defaults.py` exists for exactly this ("This file is different in every branch"), and `test_shell_params.py:25` already imports it rather than hardcoding:

```suggestion
            pytest.param({}, AIRFLOW_BRANCH, id="falls-back-to-this-checkouts-branch"),
```

(plus `from airflow_breeze.branch_defaults import AIRFLOW_BRANCH` at the top of the file)

dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
```diff
@@ -2509,12 +2510,39 @@ def deploy_cluster(
...
-# go-sdk/java-sdk copies still test current SDK sources. See kubernetes-tests/lang_sdk/README.md.
+# Runs targeting a branch other than main build the Go/Java SDKs from upstream main, so
+# release/backport branches with stale or missing go-sdk/java-sdk copies still test current SDK
+# sources. See kubernetes-tests/lang_sdk/README.md.
 LANG_SDK_UPSTREAM_GIT_URL = "https://github.com/apache/airflow.git"
 LANG_SDK_UPSTREAM_REF = "main"
 
 
+def _lang_sdk_target_branch() -> str:
+    """Branch this run targets. GITHUB_BASE_REF for PRs, DEFAULT_BRANCH in CI, else this checkout's."""
```
potiuk: This documents `DEFAULT_BRANCH` as the CI-side source, but `_setup_lang_sdk_test` runs on the host and I can't find anything that sets `DEFAULT_BRANCH` there: the `tests-kubernetes-lang-sdk` job in `k8s-tests.yml` doesn't set it, and `shell_params.py:667` injects it only into the container env dict — and from `self.airflow_branch`, so even when it is set it equals the `AIRFLOW_BRANCH` fallback anyway.

Harmless either way, but either wire it into the job env or drop the claim from the docstring and the README. Happy to be corrected if there's a path I missed.

potiuk (approved):
LGTM — scoping the upstream-main fallback to non-`main` targets is the right fix, and the `K8S Lang-SDK` job going green on this PR is good evidence the new path actually works end to end. Approving; two things below that should be fixed, plus one observation — none of them need another review round, but the first one will bite on the backport.

### Should be fixed: test hardcodes `main` and will fail on the backport branch

See the inline comment on `dev/breeze/tests/test_kubernetes_lang_sdk_commands.py`. Short version: with both env vars unset `_lang_sdk_target_branch()` returns `AIRFLOW_BRANCH`, which is `"v3-3-test"` on the branch this PR is labelled to backport to — so that parametrised case asserts `"main"` against `"v3-3-test"`. The test file's pre-image blob on `v3-3-test` is identical to main's, so the hunk cherry-picks cleanly and lands the assertion verbatim, breaking `Breeze unit tests` there.

### Should be fixed: two comments left asserting the old unconditional behaviour

Both sit on unchanged lines, so there's no inline anchor for them:

- `dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py:2645` — "go_example's go.sum is tidied against the in-repo go-sdk, but the bundle is built against the upstream-main go-sdk copied in above". That's now only true when the run targets something other than `main`. The docstring immediately above it was updated for exactly this case; this one was missed.
- `dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py:2749` — "...outside `/repo/java-sdk`, which is remounted to the (per-run) upstream copy below". Same: on a main-targeting run that remount is the in-repo `java-sdk/` itself.

> Comments should explain *why* when it is non-obvious, not narrate *what*.

— [`.github/instructions/code-review.instructions.md` § AI-Generated Code Signals](https://github.com/apache/airflow/blob/main/.github/instructions/code-review.instructions.md#ai-generated-code-signals)

A comment that asserts something no longer true costs more than one that merely narrates.

### Smaller observations

- See the inline comment on `kubernetes_commands.py:2521` about the `DEFAULT_BRANCH` link, which looks like it never fires in the process that reads it.
- The open thread about `LANG_SDK_UPSTREAM_REF` doing double duty is still unresolved — leaving that one between you and the reviewer who raised it; it doesn't change my read of the change.

---

> will follow up.*
>

---

jason810496:
Thanks all for the review, I just resolved the comments in https://github.com/apache/airflow/pull/71527/changes/463da970692eb94366953758b4ff08f0ea05bfe9.

PR #67229, opened 2026-05-20
Enforce execution_timeout in deferrable KubernetesPodOperator

paultmathew:
### Why + What

`KubernetesPodOperator(deferrable=True)` does not enforce `execution_timeout`. Once the operator defers, the synchronous `execute()` returns and the `signal.alarm`-based timeout context wrapping it exits cleanly — there is no further `execution_timeout` enforcement for the lifetime of the deferral. Pods continue running well past `execution_timeout`, bounded only by `active_deadline_seconds` (which defaults to ~1h or whatever the operator passed).

The framework gap is acknowledged by `# TODO: handle timeout in case of deferral` at [`task-sdk/.../task_runner.py:1782`](https://github.com/apache/airflow/blob/main/task-sdk/src/airflow/sdk/execution_time/task_runner.py#L1782).

This PR fixes the symptom for `KubernetesPodOperator`, mirroring the pattern already merged for `AirbyteTriggerSyncOperator` ([PR #64051](https://github.com/apache/airflow/pull/64051)) and `DbtCloudRunJobOperator` ([PR #66449](https://github.com/apache/airflow/pull/66449)).

### Approach

1. **Operator (`pod.py`)**: in `invoke_defer_method`, translate `execution_timeout` into an absolute deadline anchored on `ti.start_date`:
   - `execution_deadline = ti.start_date.timestamp() + execution_timeout.total_seconds()`
   - Pass `execution_deadline` to `KubernetesPodTrigger`.
   - Pass `timeout=remaining` (timedelta) to `self.defer()` so the framework's `trigger_timeout` also bounds the trigger lifetime as a backstop.
   - Anchoring on `ti.start_date` keeps the deadline stable across re-deferrals (e.g. `logging_interval` re-entries), since Airflow preserves the original `start_date` when a task resumes from defer.
   - Re-pass `context` from `trigger_reentry` → `invoke_defer_method` so the deadline is recomputed correctly on each re-defer.

2. **Trigger (`pod.py`)**: at the top of `_wait_for_container_completion`, check `time.time() >= execution_deadline` and emit a `status="timeout"` event when the deadline is crossed. The operator's existing `trigger_reentry` terminal-event path already handles `status in ("error", "failed", "timeout", "success")` — the operator fails the task and `_clean()` runs `on_finish_action` (default: delete pod).

### Impact

[...]

PR #71528, opened 2026-08-13
Fix GKEStartPodOperator ignoring execution_timeout when deferrable

### LINKED ISSUE #71514: GKEStartPodOperator does not respect execution_timeout when deferrable=True

Providers

### Apache Airflow version

3.3.0

```python
GKEStartPodOperator(
  image="debian:bullseye",
  cmds=["bash"],
  arguments=["-c", "sleep 480"],
  deferrable=True,
  execution_timeout=timedelta(minutes=2),
  # ...
)
```
instead of the execution timeout causing the job to fail at the 2 minute marks, it chugs along for all 4 minutes and then succeeds.

Deferred pods should respect their execution timeout, similar to what was implemented in the Kubernetes provider here:
https://github.com/apache/airflow/pull/67229

### Operating System

_No response_

### Deployment

_No response_

google

apache-airflow-providers-cncf-kubernetes==10.20.0
apache-airflow-providers-google==22.2.2

1.18.0

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

Every single deferrable pod with an execution timeout set does not respect it.

### Code of Conduct

AyushkhatiDev:
## Summary

Fixes #71514.

`GKEStartPodOperator` did not respect `execution_timeout` when
`deferrable=True` because its custom `invoke_defer_method` did not
propagate the execution deadline to `GKEStartPodTrigger`.

### What changed

- Translate `execution_timeout` into an absolute `_execution_deadline`
  anchored to the task instance `start_date`.
- Pass the deadline through `trigger_kwargs` to `GKEStartPodTrigger`.
- Add a buffered `defer.timeout` backstop so the trigger can emit its
  timeout event before the framework timeout fires.
- Preserve `trigger_kwargs` in `GKEStartPodTrigger.serialize()` so the
  deadline survives trigger serialization/restarts.
- Add regression tests covering deadline propagation, no-timeout
  behavior, existing `trigger_kwargs`, and trigger serialization.

### Testing

- `TestGKEStartPodOperator`: 16 passed
- `GKEStartPodTrigger` serialization test: passed
- Ruff format/check: passed

One broader trigger test run encountered an unrelated local
`ModuleNotFoundError: airflow_shared` in the test infrastructure.
The affected serialization test and all operator tests for this change
passed successfully.

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
@@ -801,20 +801,14 @@ class GKEStartPodOperator(GKEOperatorMixin, KubernetesPodOperator):
         if type(on_finish_action) is str and self.on_finish_action not in [i.value for i in OnFinishAction]:
             on_finish_action = self.on_finish_action.split(".")[-1].lower()  # type: ignore[assignment]
 
-        # Translate ``execution_timeout`` into an absolute deadline plumbed to
-        # the trigger via ``trigger_kwargs["_execution_deadline"]``. Anchoring
-        # on ``ti.start_date`` keeps the deadline stable across re-deferrals
-        # (``logging_interval`` re-entries), since Airflow preserves the
-        # original ``start_date`` when a task resumes from defer.
+        # Anchor on ti.start_date so the deadline survives re-deferrals.
         trigger_kwargs = dict(self.trigger_kwargs or {})
         defer_timeout: datetime.timedelta | None = None
         if self.execution_timeout is not None and context is not None:
             ti_start_date = context["ti"].start_date
             execution_deadline = int(ti_start_date.timestamp() + self.execution_timeout.total_seconds())
             trigger_kwargs["_execution_deadline"] = execution_deadline
-            # Pad ``defer.timeout`` past the deadline so the framework
-            # backstop doesn't preempt the trigger's ``status="timeout"``
-            # emission and orphan the pod via the ``__fail__`` path.
+            # Allow the trigger to emit its timeout event before the framework backstop fires.
             remaining = execution_deadline - time.time()
             poll_buffer = max(60, int(self.poll_interval * 2))
             defer_timeout = datetime.timedelta(seconds=max(remaining, 0) + poll_buffer)
--- a/providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
+++ b/providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
@@ -19,6 +19,7 @@
 
 from __future__ import annotations
 
+import datetime
 import time
 import warnings
 from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
```diff
@@ -799,6 +800,25 @@ def invoke_defer_method(
         on_finish_action = self.on_finish_action
         if type(on_finish_action) is str and self.on_finish_action not in [i.value for i in OnFinishAction]:
             on_finish_action = self.on_finish_action.split(".")[-1].lower()  # type: ignore[assignment]
+
+        # Translate ``execution_timeout`` into an absolute deadline plumbed to
```
jroachgolf84: Can you please reduce the comments, as not to just restate what the code is doing?

jroachgolf84 (changes requested):

providers/google/src/airflow/providers/google/cloud/operators/kubernetes_engine.py
AyushkhatiDev: Thanks! I've reduced the comments to focus on the reasoning rather than restating the implementation. The changes are pushed in the latest commit.

providers/google/tests/unit/google/cloud/operators/test_kubernetes_engine.py
```diff
@@ -1060,12 +1061,145 @@ def test_invoke_defer_method(
...
+    @mock.patch(GKE_OPERATORS_PATH.format("timezone.utcnow"))
+    def test_invoke_defer_method_passes_execution_deadline_when_execution_timeout_set(
+        self, mock_utcnow, mock_trigger, mock_cluster_hook, mock_fetch_cluster_info, mock_defer
+    ):
+        """
+        ``execution_timeout`` is converted into an absolute ``_execution_deadline``
+        anchored on ``ti.start_date`` and propagated to the trigger via
+        ``trigger_kwargs``.
+        """
+        import time_machine
```
shahar1: `import time_machine` is inside the test body here and again at line 1170. Please move it to the module imports. AGENTS.md says: *"Imports at top of file. Valid exceptions: circular imports, lazy loading for worker isolation, `TYPE_CHECKING` blocks."* None of those exceptions applies to a test module.

providers/google/tests/unit/google/cloud/operators/test_kubernetes_engine.py
```diff
@@ -1060,12 +1061,145 @@ def test_invoke_defer_method(
...
             method_name="trigger_reentry",
+            timeout=None,
         )
 
+    @mock.patch(GKE_OPERATORS_PATH.format("GKEStartPodOperator.defer"))
+    @mock.patch(GKE_OPERATORS_PATH.format("GKEClusterAuthDetails.fetch_cluster_info"))
+    @mock.patch(GKE_OPERATORS_PATH.format("GKEHook"))
+    @mock.patch(GKE_OPERATORS_PATH.format("GKEStartPodTrigger"))
+    @mock.patch(GKE_OPERATORS_PATH.format("timezone.utcnow"))
+    def test_invoke_defer_method_passes_execution_deadline_when_execution_timeout_set(
```
shahar1: These three tests share about 25 lines of operator setup. They differ only in `execution_timeout`, the pre-existing `trigger_kwargs`, the frozen time, and the expected values. AGENTS.md says: *"Use `@pytest.mark.parametrize` for multiple similar inputs — consolidate tests that only differ in input/expected values into a single parametrized test."*

When you consolidate them, please also add a case for the `max(remaining, 0)` clamp at `kubernetes_engine.py:814`: freeze the time after the deadline and expect `timeout == datetime.timedelta(seconds=60)`. No test covers that branch now, and AGENTS.md asks that *"Every changed or added behaviour must have a test."*

shahar1 (commented):
## AI Review

The fix looks right: it mirrors the deferrable `KubernetesPodOperator` change from #67229. `GKEStartPodTrigger` inherits the deadline handling from `KubernetesPodTrigger`, the inherited `trigger_reentry` already handles `status="timeout"`, and adding `trigger_kwargs` to `serialize()` keeps the deadline across triggerer restarts. There is also a useful side effect: user-supplied `trigger_kwargs` and the re-defer counter (`_redefer_count`) now reach the GKE trigger, which was not the case before. I have left a few test-related comments inline (on `test_kubernetes_engine.py:1077` and `:1085`). None of them is blocking. The earlier request to trim the comments has been addressed.

### Smaller observations

- **Minimum `cncf.kubernetes` version.** The trigger-side timeout only works with `apache-airflow-providers-cncf-kubernetes>=10.18.0`, the first release that contains #67229. The Google provider's `cncf.kubernetes` extra still declares `>=10.1.0`. With versions 10.6–10.17, the trigger ignores `_execution_deadline`, so only the `defer(timeout=...)` backstop fires. The task then fails through the deferral-timeout path, and the pod is not cleaned up. This is still better than running forever, and you should not change the floor in this PR (per `providers/AGENTS.md`, only release managers raise `>=` floors). I am mentioning it so the release manager can see it.
- **Unspecced mocks.** In the new tests, `ti` is a bare `mock.MagicMock()` (`:1105`, `:1150`, `:1191`). AGENTS.md says: *"Use `spec`/`autospec` when mocking."* `mock.MagicMock(spec=["start_date"])` would be enough here.
- **Gen-AI disclosure.** The PR description replaces the repository's PR template, which removes the *"Was generative AI tooling used to co-author this PR?"* section. Please restore that section and tick it if it applies. See [Gen-AI Assisted contributions](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions).

### Worth a second look from

This change mirrors the `cncf.kubernetes` deferrable-timeout logic. These people have the most context on it:

- `@jscheffl` — `CODEOWNERS` owner for `providers/cncf/kubernetes` (committer); approved #67229
- `@paultmathew` — authored #67229, the KPO change this PR ports

None of them have been notified. Asking any of them for an extra pass is the maintainer's call, and optional.

---

>

PR #68393, opened 2026-06-11
Simplify OTEL metrics config to mirror traces pattern

dsuhinin:
Mirror `airflow_shared.observability.traces.configure_otel`'s shape in the metrics module so both halves of the OTel config story have one entry point, one conf-driven setup function, and the same backcompat-bridging contract.

  ## What changes

  - **New**: `airflow_shared.observability.metrics.configure_otel(conf)` — a single conf-driven entry point that bridges deprecated Airflow configs (`metrics.otel_host` / `otel_port` / `otel_ssl_active` / `otel_service` / `otel_interval_milliseconds`) into the standard OTel env vars, loads the exporter via
   the SDK's entry-point mechanism, installs the global `MeterProvider`, and returns a `SafeOtelLogger` for the Stats factory.
  - **Removed**: the 11-parameter `get_otel_logger(host=, port=, ssl_active=, prefix=, conf_interval=, debug=, ...)` function and the helper `get_otel_data_exporter` that duplicated the OTel SDK's own exporter dispatch.
  - **Removed**: the 26-line shim file in `airflow-core` and the near-identical one in `task-sdk`. Both `stats_utils.py` files now wire directly to `configure_otel`.
  - **Removed**: `otel_env_config.py` (entirely unused after the bridging consolidated into `_get_backcompat_config`).
  
  OTel environment variables take precedence over the deprecated Airflow configs in every code path, matching the rule already used in `traces/configure_otel`. IPv6 host bracketing in the bridged URL is preserved via `_format_url_host`.
  ## Why

  Before: 11-parameter shared function + 4-parameter exporter helper + a 26-line shim duplicated across two distributions + an unused env-config dataclass module. Net ~700 lines for what is, conceptually, "read conf, bridge to env, install provider."
  
  After: one `configure_otel(conf)` function that follows the same shape as the existing traces module. ~225 fewer lines net. New developers who learn one module understand the other immediately.
  
  ## Verified

  - Unit tests: 41/41 pass (8 parametrize cases for the new bridging contract; protocol-dispatch sub-tests dropped because they exercised SDK internals, not our code).
  - End-to-end: metrics emitted via the deprecated `AIRFLOW__METRICS__OTEL_*` configs (no env vars set) arrived at a real OTel collector — proving the bridging works end-to-end through the OTel SDK pipeline.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @dsuhinin** · by `@potiuk` · 2026-07-15 16:20 UTC
>
[...]

Issue #64690, opened 2026-04-03
OTel task-level metrics (ti.finish, ti.start) lost — forked processes and KubernetesExecutor


MichaelRBlack:
### Apache Airflow version

3.1.8

### What happened

Task-level OTel metrics (e.g. `ti.finish`, `ti.start`) are silently lost and never reach the OTel collector. This affects **all executors** through two distinct failure modes:

1. **Forking executors** (LocalExecutor, CeleryExecutor): The OTel SDK's `Once()` guard prevents `set_meter_provider()` from working in forked child processes, leaving the child with a stale `MeterProvider` whose export thread is dead after fork.

2. **KubernetesExecutor**: Each task pod initializes a fresh `MeterProvider` (no fork issue), but the pod terminates before the `PeriodicExportingMetricReader` fires. `ti.finish` is recorded at the very end of task execution, and the pod exits before the export interval (default 60s, commonly configured to 30s) elapses — so the metric is buffered but never sent.

The result in both cases: `ti.finish` and `ti.start` metrics are missing from Prometheus/Grafana, while scheduler-level metrics (`scheduler_heartbeat`, `executor.running_tasks`, `pool.*`, etc.) work fine since they're emitted from long-lived processes.

### What you think should happen instead

Task-level metrics emitted from any executor should be reliably exported to the OTel collector before the task process or pod exits.

#### Forking executors (LocalExecutor, CeleryExecutor)

1. Configure Airflow 3.x with OTel metrics enabled (`otel_on = True`)
2. Run any DAG with LocalExecutor or CeleryExecutor
3. Observe task logs show:
   ```
   INFO - Stats instance was created in PID 7 but accessed in PID 19. Re-initializing.
   INFO - [Metric Exporter] Connecting to OpenTelemetry Collector at http://...
   WARNING - Overriding of current MeterProvider is not allowed
   ```
4. Check Grafana/Prometheus — `ti.finish` metrics are missing

#### KubernetesExecutor

1. Configure Airflow 3.x with OTel metrics enabled and KubernetesExecutor
2. Run any DAG — tasks execute in individual pods
3. No warning is logged (the `MeterProvider` initializes successfully in a fresh process)
4. Check Grafana/Prometheus — `ti.finish` metrics are still missing because the pod exited before the periodic export

#### Failure mode 1: Forking executors — `Once()` guard survives fork

[...]

PR #71529, opened 2026-08-13
Keep the MeterProvider configured via OTEL_CONFIG_FILE

1fanwang:
Running Airflow under `opentelemetry-instrument` with `OTEL_CONFIG_FILE` set silently discards the whole file. Every view, reader and exporter it declares is thrown away and the deployment gets Airflow's defaults instead, with nothing in the logs.

`get_otel_logger()` resets the SDK's `Once` guard and calls `set_meter_provider()` unconditionally, so whatever the agent bootstrap installed is replaced. The [declarative configuration spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk/#declarative-configuration) makes that file the sole source of SDK construction, so this is a spec violation rather than a matter of precedence.

The fix declines to replace a provider that came from `OTEL_CONFIG_FILE`. The guard reset stays on every path where Airflow owns the provider, so the post-fork case in #64690 is unaffected.

Scoped to `OTEL_CONFIG_FILE` on purpose: the env-var bootstrap path carries no equivalent "sole source" rule, and Airflow's `[metrics] otel_*` settings legitimately compete with it there.

related: #64690

Overlaps textually with https://github.com/apache/airflow/pull/68393, which restructures this module but leaves the clobber in place. Happy to rebase on it if that lands first.

### Testing Done

A script mirroring what `opentelemetry-instrument airflow scheduler` does: run `_OTelSDKConfigurator()._configure()` with `OTEL_CONFIG_FILE` pointing at a config that declares a `*_duration` view, then call `get_otel_logger()` and inspect the installed provider.

```yaml
# otel-config.yaml
file_format: "1.0-rc.1"
meter_provider:
  readers:
    - periodic:
        interval: 60000
        exporter:
          console: {}
  views:
    - selector:
        instrument_name: "*_duration"
      stream:
        aggregation:
          explicit_bucket_histogram:
            boundaries: [0.5, 1, 2, 4, 8]
```

```python
# repro.py
import os
os.environ["OTEL_CONFIG_FILE"] = "otel-config.yaml"

from opentelemetry import metrics
from opentelemetry.sdk._configuration import _OTelSDKConfigurator

def views_of(provider):
    sdk_config = getattr(provider, "_sdk_config", None)
    return [] if sdk_config is None else [str(v._instrument_name) for v in sdk_config.views]

_OTelSDKConfigurator()._configure()          # what opentelemetry-instrument runs
before = metrics.get_meter_provider()
print("bootstrap views:", views_of(before))

from airflow_shared.observability.metrics.otel_logger import get_otel_logger
get_otel_logger(host="localhost", port=4318)
after = metrics.get_meter_provider()
print("after get_otel_logger:", views_of(after))
print("same provider object?", before is after)
```

On `main`, the declarative view is gone and the provider has been swapped for Airflow's, whose only view is the instrument-type baseline:

```
$ pip install opentelemetry-configuration
$ python repro.py
after opentelemetry-instrument bootstrap: MeterProvider
  declarative views: ['*_duration']
after get_otel_logger():                   MeterProvider
  declarative views: ['None']

same provider object? False
RESULT: declarative view was DISCARDED
```

</details>

With this change, the provider and its view survive:

```
$ python repro.py
after opentelemetry-instrument bootstrap: MeterProvider
  declarative views: ['*_duration']
after get_otel_logger():                   MeterProvider
  declarative views: ['*_duration']

same provider object? True
RESULT: declarative view survived
```

</details>

`test_declaratively_configured_provider_is_not_replaced` covers this through the real `get_otel_logger()` against a real SDK `MeterProvider`, and fails on unpatched sources:

```
# revert just the fixed file to the released behaviour
$ git checkout origin/main -- shared/observability/src/airflow_shared/observability/metrics/otel_logger.py
$ uv run --project shared/observability pytest \
    shared/observability/tests/observability/metrics/test_otel_logger.py -q -k declarative
FAILED ...::test_declaratively_configured_provider_is_not_replaced
    assert logger.otel is configured_provider
E   assert <...MeterProvider object at 0x10a252ec0> is <...MeterProvider object at 0x109fb0190>
1 failed, 1 passed, 51 deselected

# restore the fix
$ git checkout HEAD -- shared/observability/src/airflow_shared/observability/metrics/otel_logger.py
$ uv run --project shared/observability pytest \
    shared/observability/tests/observability/metrics/test_otel_logger.py -q -k declarative
2 passed, 51 deselected
```

</details>

`test_provider_is_replaced_without_declarative_config` pins the other side, so the fix can't quietly widen into "never replace any provider".

The #64690 re-init path still exports after a second `get_otel_logger()` call:

```
$ python -c "
from airflow_shared.observability.metrics.otel_logger import get_otel_logger, flush_otel_metrics
get_otel_logger(debug=True)
logger = get_otel_logger(debug=True)
logger.incr('post_fork_stat')
flush_otel_metrics()" | grep -c post_fork_stat
3
```

</details>

### DIFF
--- a/shared/observability/src/airflow_shared/observability/metrics/otel_logger.py
+++ b/shared/observability/src/airflow_shared/observability/metrics/otel_logger.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import atexit
 import datetime
 import logging
+import os
 import random
 import warnings
 from collections.abc import Callable
@@ -75,6 +76,11 @@ UP_DOWN_COUNTERS = {"airflow.dag_processing.processes"}
 DEFAULT_METRIC_NAME_PREFIX = "airflow"
 # Delimiter is placed between the universal metric prefix and the unique metric name.
 DEFAULT_METRIC_NAME_DELIMITER = "."
+
+# Not imported from ``opentelemetry.sdk.environment_variables``: the constant postdates the
+# ``opentelemetry-api>=1.27.0`` floor this distribution declares, while the name is spec-fixed.
+# https://opentelemetry.io/docs/specs/otel/configuration/sdk/#declarative-configuration
+OTEL_CONFIG_FILE = "OTEL_CONFIG_FILE"
 
 
 def full_name(name: str, *, prefix: str = DEFAULT_METRIC_NAME_PREFIX) -> str:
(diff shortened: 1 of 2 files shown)


uros-b (approved):

uros-b:
Nice, thank you @1fanwang!

potiuk (approved):

PR #71534, opened 2026-08-13
Make SparkSubmitOperator durable execution inert below Airflow 3.3

amoghrajesh:
---

`durable` was already wired straight into the operator's YARN cluster-mode validation and, before this, was silently accepted below 3.3 with no signal that it does nothing there since this provider still supports Airflow 2.11 through 3.x and ResumableJobMixin (and the task state store it relies on) only exists on 3.3+.

The compat stub now warns if durable is set explicitly below 3.3 and always resolves it to False, matching the same fix already applied to https://github.com/apache/airflow/pull/71531

`reconnect_on_retry` -- the parameter's original name, introduced in the same PR as ResumableJobMixin itself and renamed to durable almost immediately -- gets a plain map-and-deprecate rather than the legacy-flag treatment Glue and KPO need. It was never a distinct pre-existing mechanism with its own below-3.3 behavior to preserve, so telling a caller on any version to "use durable instead" is not bad advice the way it would be for a flag with real legacy behavior. The
deprecation warning fires on every Airflow version regardless, since removal happens on this provider's own release schedule, not the caller's.

The YARN `yarn_track_via_rm_api` requirement needed no code change: it already reads `self.durable`, so once durable is properly inert below 3.3, the requirement naturally stops applying there too -- there's no durable-execution prerequisite to enforce when durable execution itself isn't active.

---


amoghrajesh:
cc @nailo2c would love to get a review if you can

eladkal (approved):

PR #71535, opened 2026-08-13
Fetch missing Git bundle tracking refs

### LINKED ISSUE #71388: GitDagBundle: promoting a SHA-pinned tracking_ref can fail with 'reference is not a tree' if local storage isn't cleared

CAOShurong:
Promoting a SHA-pinned Git Dag bundle can reuse a working clone whose object database predates the new commit, even though the bundle's local bare mirror has already fetched it. The subsequent checkout then fails with `reference is not a tree` unless local bundle storage is cleared.

Fetch the working clone from its local bare origin only when the configured `tracking_ref` cannot already be resolved, then continue with the existing checkout flow. This keeps the normal branch and tag paths unchanged and avoids an unconditional extra fetch.

The regression test now promotes a second commit into the same local bundle storage and verifies both the resolved version and the newly added file.

closes: #71388

Testing:

- `pytest providers/git/tests/unit/git/bundles/test_git.py::TestGitDagBundle::test_tracking_ref_commit_sha_promote_without_clearing_storage -xvs` (fails on `main`, passes with this change)
- `pytest providers/git/tests/unit/git/bundles/test_git.py -q -s --disable-warnings` (`105 passed`)
- focused test after installing the built provider wheel (`1 passed`)
- `ruff check` and `ruff format --check` for both changed files
- `mypy providers/git/src/airflow/providers/git/bundles/git.py`
- `uv build --project providers/git --wheel`
- `prek run --from-ref b2ee721a2cd30181b2759c00f359d115e2859a8c --stage pre-commit`
- `prek run mypy-providers --from-ref b2ee721a2cd30181b2759c00f359d115e2859a8c --stage manual`

---

- [X] Yes (OpenAI Codex)

---

Drafted-by: OpenAI Codex (no human review before posting)

### DIFF
--- a/providers/git/src/airflow/providers/git/bundles/git.py
+++ b/providers/git/src/airflow/providers/git/bundles/git.py
@@ -212,9 +212,8 @@ class GitDagBundle(BaseDagBundle):
                 raise RuntimeError("Error cloning repository") from e
             except InvalidGitRepositoryError as e:
                 raise RuntimeError(f"Invalid git repository at {self.repo_path}") from e
-            # If tracking_ref was just repointed to a SHA that predates this working clone's
-            # last fetch, this checkout fails until the clone is fetched or storage is cleared;
-            # tracked at https://github.com/apache/airflow/issues/71388
+            if not self._has_version(self.repo, self.tracking_ref):
+                self.repo.remotes.origin.fetch()
             self.repo.git.checkout(self.tracking_ref)
             self._log.debug("bundle initialize", version=self.version)
             if self.version:
--- a/providers/git/tests/unit/git/bundles/test_git.py
+++ b/providers/git/tests/unit/git/bundles/test_git.py
@@ -734,14 +734,7 @@ class TestGitDagBundle:
         assert {"test_dag.py"} == files_in_repo
 
     @mock.patch("airflow.providers.git.bundles.git.GitHook")
-    def test_tracking_ref_commit_sha_promote_fails_without_clearing_storage(self, mock_githook, git_repo):
-        """A SHA created after the bundle's local storage was first populated can't be
-        promoted to in-place: the working clone never fetches it before checkout.
-
-        This documents a known limitation rather than desired behavior -- it should start
-        passing once the fix tracked at https://github.com/apache/airflow/issues/71388 lands,
-        at which point this test should be updated to assert success instead.
-        """
+    def test_tracking_ref_commit_sha_promote_without_clearing_storage(self, mock_githook, git_repo):
         repo_path, repo = git_repo
         mock_githook.return_value.repo_url = repo_path
         first_commit = repo.head.commit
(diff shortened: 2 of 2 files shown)


8silvergun:
Following up from #72759, which describes a different way to reach the same failing `checkout`: if a process is killed while a bundle version is being cloned, `versions/<sha>/` is left with a `.git` that has no references and an empty working tree, and it is reused on every later initialization.

Your change looks like it recovers that state too, for a reason it was not written for:

- `_has_version(repo, tracking_ref)` cannot resolve `tracking_ref` in a repository with no refs, so it returns `False` and the fetch runs.
- `git clone` writes the remote before it transfers objects, so a leftover can still have `remote.origin.url` configured, pointing at the bundle's healthy local bare mirror. The fetch is local and repopulates `refs/remotes/origin/*`, after which the checkout resolves.

If you agree, would you like a regression test for that path added to this PR? It initializes a versioned bundle with a named branch such as `main` as `tracking_ref`, removes all loose and packed refs from `versions/<sha>/.git`, leaves HEAD pointing at a nonexistent branch, empties the working tree, verifies that `origin` remains configured, then initializes the same version again and asserts that both the expected version and the Dag files are restored. It fails on `main` and should pass with this change.

I have it written against the current `TestGitDagBundle` fixtures and can push it to your branch or post it here as a patch, whichever you prefer. Happy to leave it out if you would rather keep this PR scoped to #71388.

PR #71536, opened 2026-08-13
Pass native datetime and UUID literals to stub tasks

jason810496:
- related: #69757 shipped the `@task.stub` TaskFlow arg-binding contract; this fills in the value half of it.
- next: https://github.com/apache/airflow/pull/71809

## Why

A `@task.stub` TaskFlow call may only carry values `json.dumps` already knows, so the most natural arguments to pass a foreign-language task were rejected outright:

```python
@task.stub(queue="golang")
def schedule_window(starts_at: datetime, every: timedelta, trace_id: UUID): ...


schedule_window(datetime(2024, 1, 2, 3, 4, 5), timedelta(minutes=5), UUID("6ba7b810-..."))
```

```
ValueError: @task.stub task 'schedule_window' parameter 'starts_at' received a literal of
type datetime that is not JSON-serializable, so it cannot be passed to the foreign runtime
```

The parameter's `value_schema` already told the runtime this was `{"type": "string", "format": "date-time"}` but the *value* could not travel.

## What

Temporal and UUID literals are now rendered through the **same Pydantic adapter that produced their `value_schema`**, so the value and schema cannot disagree by construction. The conversion recurses through typed lists, sequences, tuples, mapping values, sets, and unions, so supported values use the same wire spelling at every depth. Sets containing supported leaves become deterministically sorted JSON arrays while `uniqueItems: true` tells the runtime it may reconstruct set semantics.

| Python literal | Wire value after this PR | `value_schema` | Rejected before? | Rejected after? |
| --- | --- | --- | --- | --- |
| `datetime(2024, 1, 2, 3, 4, 5)` | `"2024-01-02T03:04:05Z"` | `{"type": "string", "format": "date-time"}` | Yes | No |
| `date(2024, 1, 2)` | `"2024-01-02"` | `{"type": "string", "format": "date"}` | Yes | No |
| `time(3, 4, 5)` | `"03:04:05"` | `{"type": "string", "format": "time"}` | Yes | No |
| `timedelta(days=1, hours=2)` | `"P1DT2H"` | `{"type": "string", "format": "duration"}` | Yes | No |
| `UUID("6BA7B810-...")` | `"6ba7b810-..."` | `{"type": "string", "format": "uuid"}` | Yes | No |
| `[datetime(2024, 1, 2, 3, 4, 5)]` | `["2024-01-02T03:04:05Z"]` | `{"type": "array", "items": {"type": "string", "format": "date-time"}}` | Yes | No |
| `{datetime(2024, 1, 2), datetime(2024, 1, 3)}` | `["2024-01-02T00:00:00Z", "2024-01-03T00:00:00Z"]` | `{"type": "array", "items": {"type": "string", "format": "date-time"}, "uniqueItems": true}` | Yes | No |
| `Status.READY` for plain `Enum` | rejected, no wire value | `{"type": "string", "enum": ["ready"], "title": "Status"}` | Yes | Yes |
| `Status.READY` for `str`-backed Enum | `"ready"` | `{"type": "string", "enum": ["ready"], "title": "Status"}` | No | No |
| `Priority.HIGH` for `int`-backed Enum | `1` | `{"type": "integer", "enum": [1], "title": "Priority"}` | No | No |
| `Decimal("1.20")` | rejected, no wire value | number or numeric string | Yes | Yes |
| `Path("/tmp/example")` | rejected, no wire value | `{"type": "string", "format": "path"}` | Yes | Yes |

Only temporal and UUID leaves opt into Pydantic serialization. Plain Enum members, `Decimal`, and `Path` retain their previous rejection behavior; string- and integer-backed Enums retain their existing JSON-native behavior, and bytes remain unsupported. Schema generation and literal conversion are deliberately separate: Pydantic may describe a type even when its Python object cannot be emitted as JSON.

## Timezone-naive timestamps

A naive `datetime` is pinned to an explicit offset before serializing, via the same `coerce_datetime` the rest of Airflow uses. An offset-less timestamp is a *different instant* to each language runtime, so leaving it naive would make a task's behaviour depend on which language happens to run it.

| Wire value | Go | Java `Instant.parse` | JS `new Date` |
| --- | --- | --- | --- |
| `2024-01-02T03:04:05Z` | `03:04:05Z` | `03:04:05Z` | `03:04:05Z` |
| `2024-01-02T03:04:05` | `03:04:05Z` | throws | `08:04:05Z` (worker-local) |

---

- [x] Yes, with help of Claude Code following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
@@ -43,18 +43,28 @@ by pydantic from the stub annotation, carried verbatim, unknown keywords ignored
 lang SDK should decode the value into. Every SDK is expected to follow the same table,
 so a Dag author sees one behaviour regardless of the task's language:
 
-===================  ==========  ====================================  =========================
-Python annotation    ``type``    ``format`` / wire spelling            Native target
-===================  ==========  ====================================  =========================
-``datetime``         string      ``date-time`` ``2024-01-02T03:04:05Z``  timestamp
-``date``             string      ``date`` ``2024-01-02``                 date
-``time``             string      ``time`` ``03:04:05``                    time of day
-``timedelta``        string      ``duration`` ``P1DT2H3M4S`` ``-PT1M30S`` duration
-``UUID``             string      ``uuid`` ``6ba7b810-9dad-...-...``       UUID
-``bytes``            string      ``binary`` (raw text, **not** base64)   byte string
-``int``              integer     ``int64``                                64-bit integer
-``float``            number      ``double``                               64-bit float
-===================  ==========  ====================================  =========================
+=========================  ========================  ====================================  ==================  =================================
+Python annotation          JSON-schema signal        Wire spelling                         Native target       Inline literal handling
+=========================  ========================  ====================================  ==================  =================================
+``datetime``               string + ``date-time``    ``2024-01-02T03:04:05Z``              timestamp           converted
+``date``                   string + ``date``         ``2024-01-02``                         date                converted
+``time``                   string + ``time``         ``03:04:05``                           time of day         converted
+``timedelta``              string + ``duration``     ``P1DT2H3M4S`` or ``-PT1M30S``         duration            converted
+``UUID``                   string + ``uuid``         ``6ba7b810-9dad-...-...``              UUID                converted
+``int``                    integer + ``int64``       ``42``                                 64-bit integer      JSON-native
+``float``                  number + ``double``       ``1.5``                                64-bit float        JSON-native
+``Enum`` (string value)    string + ``enum``         ``"value"``                            enum                member rejected; pass ``.value``
+``str``-backed ``Enum``    string + ``enum``         ``"value"``                            enum                JSON-native
+``int``-backed ``Enum``    integer + ``enum``        ``1``                                  enum                JSON-native
+``Decimal``                number or string          ``1.2`` or ``"1.20"``                   decimal             rejected; pass number/string
+``Path``                   string + ``path``         ``"/tmp/example"``                     path or string      rejected; pass string
+``set[datetime]``          array + ``uniqueItems``   ``["2024-01-02T03:04:05Z"]``           set of timestamps   converted in stable order
+=========================  ========================  ====================================  ==================  =================================
+
+``Inline literal handling`` describes a value captured directly from the Python Dag. The
+schema also accompanies XCom bindings, where the value is produced later. Schema generation
+does not itself serialize a literal: unsupported Python objects are rejected before a binding
+is emitted even when pydantic can describe their eventual JSON representation.
 
 Timestamps always carry an explicit offset -- a naive ``datetime`` is pinned to Airflow's
 default timezone at serialization time -- because an offset-less timestamp means different
--- a/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
@@ -37,7 +37,33 @@ ArgValueSchema = TypeAliasType(
     "ArgValueSchema", Annotated[dict[str, JsonValue], Field(title="ArgValueSchema")]
 )
 """JSON-schema fragment constraining the value a stub-task argument binds to; generated
-by pydantic from the stub annotation, carried verbatim, unknown keywords ignored."""
+by pydantic from the stub annotation, carried verbatim, unknown keywords ignored.
+
+``format`` carries the part of the contract ``type`` alone cannot: which native type a
+lang SDK should decode the value into. Every SDK is expected to follow the same table,
+so a Dag author sees one behaviour regardless of the task's language:
+
+===================  ==========  ====================================  =========================
+Python annotation    ``type``    ``format`` / wire spelling            Native target
+===================  ==========  ====================================  =========================
+``datetime``         string      ``date-time`` ``2024-01-02T03:04:05Z``  timestamp
+``date``             string      ``date`` ``2024-01-02``                 date
+``time``             string      ``time`` ``03:04:05``                    time of day
+``timedelta``        string      ``duration`` ``P1DT2H3M4S`` ``-PT1M30S`` duration
+``UUID``             string      ``uuid`` ``6ba7b810-9dad-...-...``       UUID
+``bytes``            string      ``binary`` (raw text, **not** base64)   byte string
+``int``              integer     ``int64``                                64-bit integer
+``float``            number      ``double``                               64-bit float
+===================  ==========  ====================================  =========================
+
+Timestamps always carry an explicit offset -- a naive ``datetime`` is pinned to Airflow's
+default timezone at serialization time -- because an offset-less timestamp means different
+instants to different runtimes (UTC in Go, worker-local in JavaScript, unparsable in Java).
+
+A ``string`` target is always acceptable for any of the string formats: an SDK that does
+not model a format hands the raw text to the task and lets it parse. Unions serialize as
+``anyOf``, and a ``null`` branch means the argument may arrive absent, so the native
+parameter has to be nullable."""
 
 
 class _ArgBindingBase(BaseModel):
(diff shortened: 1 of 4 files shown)


airflow-core/src/airflow/serialization/stub_arg_bindings.py
```diff
@@ -108,36 +109,76 @@ def _infer_value_schema(annotation: Any) -> dict[str, Any] | None:
...
+    Values pydantic cannot render for this annotation pass through untouched, leaving
+    the JSON-serializability check to reject them.
+    """
+    if annotation is Parameter.empty or annotation is None or annotation is Any:
+        return value
+    if isinstance(value, datetime.datetime):
+        # A naive timestamp is ambiguous once it leaves Python: Go would read it as UTC,
+        # JavaScript as the worker's local time, and Java would refuse to parse it. Pin
+        # the offset here, using the same default timezone the rest of Airflow applies.
+        value = coerce_datetime(value)
```
uranusjr: This probably also need to go into collection types? Otherwise it would miss e.g. `list[datetime]`.

airflow-core/src/airflow/serialization/stub_arg_bindings.py
```diff
@@ -274,6 +319,8 @@ def build_arg_bindings(op: DecoratedOperator) -> list[dict[str, Any]] | None:
                 xcom_entry["value_schema"] = value_schema
             spec.append(xcom_entry)
             continue
+        _reject_nested_xcom(value, task_id, name)
+        value = _to_json_value(value, annotations[name])
```
uranusjr, line 439: This runs _every_ annotated literal through `adapter.dump_python(..., mode="json")`. An e.g. Enum or Decimal literal that previously round-tripped as-is (or was rejected) will now come out as whatever pydantic's json-mode serializer produces for it. The implied behavioural change may be greater than it seems.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,33 @@
...
+
+===================  ==========  ====================================  =========================
+Python annotation    ``type``    ``format`` / wire spelling            Native target
+===================  ==========  ====================================  =========================
+``datetime``         string      ``date-time`` ``2024-01-02T03:04:05Z``  timestamp
+``date``             string      ``date`` ``2024-01-02``                 date
+``time``             string      ``time`` ``03:04:05``                    time of day
+``timedelta``        string      ``duration`` ``P1DT2H3M4S`` ``-PT1M30S`` duration
+``UUID``             string      ``uuid`` ``6ba7b810-9dad-...-...``       UUID
+``bytes``            string      ``binary`` (raw text, **not** base64)   byte string
```
uranusjr: Are we sure this is correct? There’s no explicit bytes-handling, so this is handled directly by Pydantic? Do we want to be explicit in the implementation so we’re safe if Pydantic changes in the future?

airflow-core/src/airflow/serialization/stub_arg_bindings.py
jason810496: Good point. I had not considered collection types. Addressed in 108dbf5f8b, thanks.

airflow-core/src/airflow/serialization/stub_arg_bindings.py
jason810496: I had not considered Enum or Decimal initially. The table in the PR description now shows the behavior before and after this PR. The behavior of Enum and Decimal remains unchanged after this PR.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
jason810496: I overlooked bytes in my review. We do not need to support bytes at all. I also added the Enum, Decimal, and Path cases to the docstring table.

jason810496 (commented):
Thanks TP for the review.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,43 @@
...
+by pydantic from the stub annotation, carried verbatim, unknown keywords ignored.
+
+``format`` carries the part of the contract ``type`` alone cannot: which native type a
+lang SDK should decode the value into. Every SDK is expected to follow the same table,
+so a Dag author sees one behaviour regardless of the task's language:
+
+=========================  ========================  ====================================  ==================  =================================
+Python annotation          JSON-schema signal        Wire spelling                         Native target       Inline literal handling
+=========================  ========================  ====================================  ==================  =================================
+``datetime``               string + ``date-time``    ``2024-01-02T03:04:05Z``              timestamp           converted
```
uranusjr, line 49: Nit: this line is misaligned

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,43 @@
...
+``time``                   string + ``time``         ``03:04:05``                           time of day         converted
+``timedelta``              string + ``duration``     ``P1DT2H3M4S`` or ``-PT1M30S``         duration            converted
+``UUID``                   string + ``uuid``         ``6ba7b810-9dad-...-...``              UUID                converted
+``int``                    integer + ``int64``       ``42``                                 64-bit integer      JSON-native
+``float``                  number + ``double``       ``1.5``                                64-bit float        JSON-native
+``Enum`` (string value)    string + ``enum``         ``"value"``                            enum                member rejected; pass ``.value``
+``str``-backed ``Enum``    string + ``enum``         ``"value"``                            enum                JSON-native
+``int``-backed ``Enum``    integer + ``enum``        ``1``                                  enum                JSON-native
+``Decimal``                number or string          ``1.2`` or ``"1.20"``                   decimal             rejected; pass number/string
```
uranusjr, line 59: Same on alignment.

(Unrelatedly, I'm slightly worried about this one since decimal may be quite useful for specific data use cases. But we can always add support to this if someone raises an issue.)

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,43 @@
...
+``Path``                   string + ``path``         ``"/tmp/example"``                     path or string      rejected; pass string
+``set[datetime]``          array + ``uniqueItems``   ``["2024-01-02T03:04:05Z"]``           set of timestamps   converted in stable order
```
uranusjr, line 61: Does the ordering point apply to any kind of set? It seems weird set[datetime] is specifically mentioned.

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,43 @@
...
+=========================  ========================  ====================================  ==================  =================================
+
+``Inline literal handling`` describes a value captured directly from the Python Dag. The
```
uranusjr, line 64: This shouldn't be a code block (double backticks)

airflow-core/src/airflow/api_fastapi/execution_api/datamodels/task_arg_binding.py
```diff
@@ -37,7 +37,43 @@
...
+schema also accompanies XCom bindings, where the value is produced later. Schema generation
+does not itself serialize a literal: unsupported Python objects are rejected before a binding
+is emitted even when pydantic can describe their eventual JSON representation.
+
+Timestamps always carry an explicit offset -- a naive ``datetime`` is pinned to Airflow's
+default timezone at serialization time -- because an offset-less timestamp means different
+instants to different runtimes (UTC in Go, worker-local in JavaScript, unparsable in Java).
```
uranusjr, line 71: I would simply say they "have different offset handling semantics" instead. Saying "instants" may be confusing since an instant also mat means slightly different things in different languages.

uranusjr (commented):
Mostly nits on documentation wording

PR #71544, opened 2026-08-13
Time-box and retry pre-extras SDK downloads

potiuk:
The lowest-dependency providers job fails when IBM's download server stops answering: the `ibm.mq` pre-extras manifest fetch passed no socket timeout, so each of the five routes (DNS plus the four published anycast IPs) sat in the kernel's TCP connect timeout, and the single pass over them had no retry. One transient upstream outage took the whole job down — see the `Errno 110` failures in [this run](https://github.com/apache/airflow/actions/runs/31479252918/job/93745832553).

Every attempt now uses a 20s socket timeout, and the full set of routes is retried in 3 rounds with a short sleep between them, so a stalled transfer or a blackholed IP recovers in seconds. A checksum mismatch is still fatal on the first attempt — that means the manifest disagrees with what upstream serves, which retrying cannot fix.

---


potiuk:
Green and self-contained: the pre-extras SDK download now has a time box and retries instead of being able to hang the build.

@dabla  you have both worked on this script — could one of you take a look?

---

Issue #71489, opened 2026-08-12
Bound and harden the task-UUID scan fallback


amoghrajesh:
`GlueJobOperator._find_job_run_id_by_task_uuid` (the task-UUID scan fallback, used when neither `task_state_store` nor a cached XCom id has a prior run) has a few pre-existing quality issues, carried over verbatim from `main`, that are worth cleaning up:

1. **Unbounded pagination.** It's a `while True` over `get_job_runs(MaxResults=50)` with no page cap and no age cutoff. The no-match case — the prior attempt died before it ever called `StartJobRun`, which is the common retry shape — is exactly the one that walks the job's entire run history before giving up.
2. **Failure is silently swallowed.** The two `except Exception` blocks around this scan (and the XCom lookup before it) log at `warning` level and return `None`, after which the operator submits fresh — straight into `ConcurrentRunsExceededException` against a run that's still alive. There's no error-level signal telling anyone why.
3. **The `except Exception` is broader than it needs to be.** It should be narrowed to `except ClientError`, so a real bug in the surrounding code doesn't get silently absorbed the same way as a transient AWS-side issue.
4. **Missing IAM documentation.** Anyone relying on this fallback (via `resume_glue_job_on_retry=True` today, or explicit `durable=True` on Airflow <3.3 going forward) needs `glue:GetJobRuns` in their task policy, on top of `StartJobRun`/`GetJobRun`. The docs don't currently call that out.

None of this is new — it's present on `main` today, gated behind the opt-in `resume_glue_job_on_retry` flag. Raised during review of #71211, initially flagged as more urgent because that PR was going to make the scan reachable by default; the default-flip was scoped back to explicit opt-in for Airflow <3.3 (and to a narrow crash-recovery window on 3.3+) in the same PR, so the original urgency no longer applies — but the underlying issues are still real for anyone who does hit this path.

Fix direction: bound the walk (a page cap, or stop once a run's `StartedOn` predates the DAG run), narrow the exception handling, log the fallthrough at `error` level with the actual reason, and document the IAM requirement.

See https://github.com/apache/airflow/pull/71211#discussion_r3764425788 for the original discussion.

MannXo:
I'd like to give this a shot if it's unassigned.

atanishka308:
I would like to work on this issue.

Johnkothapalli:
[...]

PR #71546, opened 2026-08-13
Limit Glue retry task UUID scan

fat-catTW:
## Why
`GlueJobOperator` can reconnect to a previously submitted Glue job run on retry by tagging the Glue run arguments with the Airflow task UUID and later scanning `GetJobRuns` for that UUID.

That fallback is useful when the operator does not have a stored Glue run id, but the no-match case currently has no upper bound. If a Glue job has a long run history, a retry can keep paginating through the full history before deciding to submit a fresh run.

The reconnect lookup also catches broad exceptions. That can hide non-AWS bugs and make AWS lookup failures harder to diagnose, because the operator may continue by submitting a new Glue job run and surface a later error such as `ConcurrentRunsExceededException`.

relates: #71489

## Solution
Bound the task UUID fallback scan to 10 `GetJobRuns` pages. The scan keeps the existing page size of 50 runs, so a retry checks at most 500 recent Glue job runs before treating the previous run as not found.

This keeps the reconnect path useful for normal retry and crash-recovery cases while preventing an unbounded walk through the full Glue job history when there is no match.

Also narrow reconnect lookup exception handling to AWS `ClientError`. These failures are now logged with exception details, while non-AWS errors bubble up instead of being swallowed.

Finally, update the Glue operator documentation to mention that this retry fallback requires `glue:GetJobRuns`, in addition to the usual Glue job run permissions.

---


providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -180,6 +180,8 @@ class GlueJobOperator(ResumableJobMixin, AwsBaseOperator[GlueJobHook]):
 
     operator_extra_links = (GlueJobRunDetailsLink(),)
     TASK_UUID_ARG = "--airflow_task_uuid"
+    TASK_UUID_SCAN_PAGE_SIZE = 50
```
o-nikolas: Why not max this out to the 200 limit? You'll avoid more possible throttling errors if you request as many as possible each time through.  Also you'll have a higher chance of getting a hit if the job is busy and you need to go far into history to find the existing uuid match.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
```diff
@@ -447,8 +452,10 @@ def _find_previous_job_run(self, context: Context, task_uuid: str) -> str | None
                 self.log.info("Previous Glue job_run_id: %s, state: %s", previous_job_run_id, state)
                 if self.is_job_active(state):
                     return previous_job_run_id
-            except Exception:
-                self.log.warning("Failed to get previous Glue job run state", exc_info=True)
+            except ClientError:
```
o-nikolas, line 455: These now catch much less than before. So things that used to just cause a fresh Glue Job to be created now fail the entire task? Was that intentional?

fat-catTW: Yes, this was intentional. AWS lookup failures are still handled via `ClientError`, so a Glue API failure during reconnect lookup still logs the exception and falls back to submitting a fresh run.

The behavior change is for non-AWS exceptions. Those are more likely to be programming errors or unexpected local assumptions, and silently converting them into a fresh Glue job submission can hide the real bug and potentially create a duplicate run. That was one of the issues this PR is trying to address.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
fat-catTW: Thanks for review.
Good point, I updated `TASK_UUID_SCAN_PAGE_SIZE` to 200, matching the `GetJobRuns` `MaxResults` limit. The scan remains bounded by page count while checking more recent runs per request.

providers/amazon/src/airflow/providers/amazon/aws/operators/glue.py
o-nikolas: I think this is mostly fair yeah, definitely a behaviour change though. Is this a bugfix justifiable one though? Probably? What do you think @vincbeck?

vincbeck: Yeah I would qualify it a bug fix. Catching `Exception` should be prohibited unless very specific reason so at the end I prefer the version in this PR than the one in main.

jkasireddy:
@fat-catTW  I like the bounded lookup direction. One edge case I’m wondering about is whether a fixed 500-run cap can miss a valid previous attempt on high-frequency Glue jobs shared across multiple DAGs. Since GetJobRuns is returned newest-first, would stopping once StartedOn predates the relevant DAG run provide a stronger semantic boundary than the fixed page count alone? A regression test where the matching run falls beyond the first page may also help document the intended behavior.

fat-catTW:
Good point. 

I agree that a `StartedOn` cutoff could be a stronger semantic boundary, but I’m hesitant to use the Dag run `logical_date` for that because it is not the actual execution time, especially for manual runs and backfills. Using an actual Dag run or task start time may be a good follow-up, but I’d prefer to keep this PR focused on replacing the unbounded scan with a fixed worst-case bound.

BTW. The page size was bumped to 200 after an earlier review comment, so the current hard cap is 2,000 runs rather than 500.
Also, there is already a regression test for a match beyond the first page: `test_task_uuid_scan_paginates_until_match`.

PR #71548, opened 2026-08-13
Stop CI duration alerts firing on one-off image rebuilds

potiuk:
The per-job duration trend excludes one step, `Prepare breeze & CI image`, from a job's wall clock. The image-cache push jobs do not have that step — they build and push the image in a step of their own — so when a change to an early image layer invalidates everything after it, the resulting full rebuild and re-push is counted as work time and reported as a regression. The same CI-only prefix also missed every job that prepares a PROD image.

Seen on the [2026-08-13 canary](https://github.com/apache/airflow/actions/runs/31661917904), the first scheduled run after [`516e4afba1`](https://github.com/apache/airflow/commit/516e4afba1) added an apt package to `install_os_dependencies.sh`: `Push CI latest images: 3.12` went 666s -> 1778s, and all ten cache-push jobs went from 7-13m to 14-31m. Image build time is meant to be handled by the persistence check, which only reports a slowdown lasting days rather than a single night.

Known gap left alone: the dedicated `Build CI images` / `Build PROD images` jobs are still exposed to the same false positive. Their heavy step name is produced by a template with an empty interpolation (`Build  linux/amd64:3.10 image`), so matching it by prefix is fragile; covering them properly means feeding those jobs into the image-build persistence check instead.

---

### DIFF
--- a/scripts/ci/analyze_ci_job_durations.py
+++ b/scripts/ci/analyze_ci_job_durations.py
@@ -38,8 +38,8 @@ with noisy timings do not trigger spurious alerts.
 Steps that build, pull or push images (``IMAGE_WORK_STEP_PREFIXES`` — preparing the
 CI/PROD image in a test job, and the build+push of the image-cache jobs) occasionally
 balloon on a one-off cache miss, so their time is *excluded* from the run and per-job
-durations used for the trend above. The image build is instead watched on its own and
-only reported when it has stayed slow for longer than ``IMAGE_BUILD_PERSISTENCE_DAYS``
+durations used for the trend above. That same image time is instead watched on its own
+and only reported when it has stayed slow for longer than ``IMAGE_BUILD_PERSISTENCE_DAYS``
 (so a single slow night never alerts).
 
 Environment variables (required):
@@ -78,23 +78,22 @@ from pathlib import Path
 from typing import TypedDict
 
 ISO_SUFFIX_Z = "Z"
-PREPARE_BREEZE_STEP_PREFIX = "Prepare breeze & CI image"
 # Steps that build, pull or push a Docker image rather than doing test/work. A change to
 # an early image layer (an apt package, a Dockerfile line) invalidates every layer after
 # it, so on the first run after such a merge these steps take multiples of their usual
 # time - a one-off that must not read as a work-time regression. Their names come from
 # .github/workflows/push-image-cache.yml and .github/actions/prepare_breeze_and_image.
 IMAGE_WORK_STEP_PREFIXES = (
-    PREPARE_BREEZE_STEP_PREFIX,
+    "Prepare breeze & CI image",
     "Prepare breeze & PROD image",
     "Push CI ",
     "Push PROD ",
 )
+IMAGE_WORK_LABEL = "Image build, pull & push"
 
 
 class JobDuration(TypedDict):
     duration: float
-    prepare_breeze_duration: float | None
     image_work_duration: float | None
 
 
@@ -262,16 +261,6 @@ def get_recent_runs(
     return runs
 
 
-def get_prepare_breeze_step_duration(job: dict) -> float | None:
-    """Return the prepare breeze step duration for a job, when the step exists."""
-    for step in job.get("steps", []):
-        name = step.get("name", "")
-        if not name.startswith(PREPARE_BREEZE_STEP_PREFIX):
-            continue
-        return duration_seconds(step.get("startedAt"), step.get("completedAt"))
-    return None
-
-
 def get_image_work_seconds(job: dict) -> float | None:
     """Return the total time a job spent building, pulling or pushing images.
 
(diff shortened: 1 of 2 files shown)


potiuk:
Small and green, and it has been sitting a while. It stops the CI duration alert from firing when a run rebuilt the CI image — that is a one-off cost rather than a regression, so alerting on it is noise.

@gopidesupavan you were the last to work on this script; would you mind taking a look?

---

scripts/ci/analyze_ci_job_durations.py
```diff
@@ -78,11 +79,23 @@
...
+    PREPARE_BREEZE_STEP_PREFIX,
+    "Prepare breeze & PROD image",
+    "Push CI ",
+    "Push PROD ",
+)
 
 
 class JobDuration(TypedDict):
     duration: float
     prepare_breeze_duration: float | None
```
Miretpl: As breeze duration is part of image work, I think we could delete that.

scripts/ci/analyze_ci_job_durations.py
```diff
@@ -259,17 +272,33 @@ def get_prepare_breeze_step_duration(job: dict) -> float | None:
     return None
 
 
+def get_image_work_seconds(job: dict) -> float | None:
```
Miretpl, line 264: A similar function already exists for `Prepare breeze & CI image`. We should rather not hold duplicates when there is almost one difference in the function itself.

scripts/ci/analyze_ci_job_durations.py
```diff
@@ -321,6 +350,7 @@ def get_run_jobs(repo: str, run_id: int) -> dict[str, JobDuration]:
             durations[name] = {
                 "duration": seconds,
                 "prepare_breeze_duration": get_prepare_breeze_step_duration(job),
+                "image_work_duration": get_image_work_seconds(job),
```
Miretpl, line 341: Image work consists of the `prepare_breeze_duration` inside. Looking at the changes only, I don't see the change which would change the check logic, which would include time for `image_work_duration` instead of/next to `prepare_breeze_duration`.

scripts/tests/ci/test_analyze_ci_job_durations.py
```diff
@@ -266,7 +266,9 @@ def test_parses_successful_jobs(self, durations_module):
         completed = subprocess.CompletedProcess(args=[], returncode=0, stdout=payload, stderr="")
         with patch.object(subprocess, "run", return_value=completed):
             jobs = durations_module.get_run_jobs("apache/airflow", 2)
-        assert jobs == {"Tests": {"duration": 20 * 60, "prepare_breeze_duration": 5 * 60}}
+        assert jobs == {
+            "Tests": {"duration": 20 * 60, "prepare_breeze_duration": 5 * 60, "image_work_duration": 5 * 60}
```
Miretpl: Is it realistic to have always the same value for `prepare_breeze_duration` and `image_work_duration`?

Miretpl (changes requested):

scripts/ci/analyze_ci_job_durations.py
potiuk: Done — the field is gone, `image_work_duration` is the only one now.

---

scripts/ci/analyze_ci_job_durations.py
potiuk: Merged — `get_prepare_breeze_step_duration` is deleted, `get_image_work_seconds` is the only extractor.

---

scripts/ci/analyze_ci_job_durations.py
potiuk: Good catch, and it was more than duplication: the persistence check only ever saw the prepare-breeze step, so a slow build in a cache-push job — the case this PR is about — never reached it. `calculate_image_build_seconds` now medians `image_work_duration`. The Slack and step-summary lines are relabelled "Image build, pull & push", since they were still naming a single step for a figure that now covers all of them.

---

scripts/tests/ci/test_analyze_ci_job_durations.py
potiuk: No, it wasn't realistic — a cache-push job has no prepare-breeze step but plenty of image work. With one field it can't arise; the fixtures now carry `image_work_duration` alone.

---

potiuk:
@Miretpl -> issues addressed :)

scripts/tests/ci/test_analyze_ci_job_durations.py
```diff
@@ -416,6 +455,41 @@ def test_image_build_spike_alone_does_not_flag_a_job(self, durations_module):
         )
         assert regressions == []
 
+    def test_cache_push_job_is_not_flagged_when_only_the_image_rebuild_grew(self, durations_module):
+        """The image-cache push jobs have no prepare-breeze step - their build+push is the image.
+
+        Modelled on the 2026-08-13 canary: an apt package added to an early image layer
+        invalidated every layer after it, so the build+push step went 11m -> 30m while the
+        job did no more work than usual.
```
Miretpl, line 463: I remember that there was something, but without a link, it doesn't bring any value really and even link, due to retention, would not make any difference. I would remove this comment.

Miretpl (approved):
Small comment, generally LGTM

PR #71549, opened 2026-08-13
Fix task duration increasing forever after Dag run timeout skips it

### LINKED ISSUE #58536: Task duration keeps increasing for skipped tasks due to DAG timeout
### Discussed in https://github.com/apache/airflow/discussions/53868

<div type='discussions-op-text'>

<sup>Originally posted by **Szjs-z** July 29, 2025</sup>
Hi Airflow community,

I've encountered an issue where task duration keeps increasing for tasks that are skipped due to DAG timeout:

**Problem Description:**
When a DAG run times out due to `dagrun_timeout`, the unfinished tasks are marked as `SKIPPED`. However, in the Web UI, the task duration continues to increase even though the task state is `SKIPPED` :

Screenshot 2025-07-29 at 17 42 11

**Root Cause Analysis:**
From code inspection in `airflow/jobs/scheduler_job.py`, the `_schedule_dag_run` method handles DAG timeout by:

```python
        if (
            dag_run.start_date
            and dag.dagrun_timeout
            and dag_run.start_date < timezone.utcnow() - dag.dagrun_timeout
        ):
            dag_run.set_state(DagRunState.FAILED)
            unfinished_task_instances = session.scalars(
                select(TI)
                .where(TI.dag_id == dag_run.dag_id)
                .where(TI.run_id == dag_run.run_id)
                .where(TI.state.in_(State.unfinished))
            )
            for task_instance in unfinished_task_instances:
                task_instance.state = TaskInstanceState.SKIPPED
                session.merge(task_instance)
            session.flush()
            self.log.info("Run %s of %s has timed-out", dag_run.run_id, dag_run.dag_id)

```

1. Setting DAG run state to FAILED
2. Marking unfinished tasks as SKIPPED
3. But it doesn't set `end_date` for these task instances

I'm wondering if this is a known issue or if there's a recommended way to handle this.

Thanks,
Chris Zhao</div>

yuseok89:
closes: #58536

When a Dag run times out, unfinished tasks are marked SKIPPED but left with no `end_date`, causing the UI to show their duration growing indefinitely. Fixed by routing through `set_state()` which properly sets both `end_date` and `duration`.

## Screenshots
Triggered a Dag with dagrun_timeout=60s and a long-running task, so the run times out while the task is unfinished and gets skipped.
- Before: the skipped task's duration keeps growing on refresh (no end_date).
- After: duration is fixed (~1m) with end_date set.

### Before
image

### After
image

---

  - Opus 5

---

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2946,8 +2946,7 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 default=None,
             )
             for task_instance in unfinished_task_instances:
-                task_instance.state = TaskInstanceState.SKIPPED
-                session.merge(task_instance)
+                task_instance.set_state(TaskInstanceState.SKIPPED, session=session)
             session.flush()
             self.log.info("Run %s of %s has timed-out", dag_run.run_id, dag_run.dag_id)
 
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -2945,8 +2945,13 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
                 key=lambda ti: ti.start_date or timezone.make_aware(datetime.min),
                 default=None,
             )
+            # Mirror TI.set_state inline so the whole batch flushes once instead of per task.
+            now = timezone.utcnow()
             for task_instance in unfinished_task_instances:
-                task_instance.set_state(TaskInstanceState.SKIPPED, session=session)
+                task_instance.state = TaskInstanceState.SKIPPED
+                task_instance.start_date = task_instance.start_date or now
+                task_instance.end_date = task_instance.end_date or now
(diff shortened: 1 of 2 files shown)


airflow-core/src/airflow/jobs/scheduler_job_runner.py
```diff
@@ -2946,8 +2946,7 @@ def _schedule_dag_run(
                 default=None,
             )
             for task_instance in unfinished_task_instances:
-                task_instance.state = TaskInstanceState.SKIPPED
-                session.merge(task_instance)
+                task_instance.set_state(TaskInstanceState.SKIPPED, session=session)
```
steveahnahn: nit: `set_state` flushes per call, so this flushes once per skipped TI instead of the single batched flush after the loop so in a wide timeout that's N round-trips. Set `end_date`/`duration` inline and keep the one `flush()`?

steveahnahn (approved):
LGTM. Fixes the growing-duration bug, and good test coverage, just one nit

airflow-core/src/airflow/jobs/scheduler_job_runner.py
yuseok89: Good catch, thanks. Inlined it keeping set_state semantics, one flush after the loop. Also dropped the redundant session.merge and added a test for the preserve case.

Issue #69360, opened 2026-07-03
Fix execution API returning non-RFC 9457 error for invalid TI state


KushagraB424:
### Under which category would you file this issue?

Airflow Core

### Apache Airflow version

3.4.0

### What happened and how to reproduce it?

**Issue Description**
In `airflow/api_fastapi/execution_api/routes/task_instances.py`, when a client tries to start a task instance that is in an invalid state, it throws an `HTTPException` with a 409 status code. However, the `detail` object does not comply with the RFC 9457 format for problem details, as noted by an existing `TODO` in the codebase.

**Steps to reproduce**
This is a code-level inconsistency marked by a `TODO` comment at line 217 in `airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py`.

### What you think should happen instead?

The code should be updated to resolve that `TODO` by ensuring the error detail dictionary contains the standard `type`, `title`, `status`, `detail`, and `instance` fields as per RFC 9457, in addition to the existing custom keys (`reason` and `previous_state`).

### Deployment

None

### Versions of Apache Airflow Providers

Not Applicable

### Official Helm Chart version

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

Not Applicable

### Docker Image customizations

Not Applicable

### Anything else?

I already have a fix ready on my fork and will be submitting a PR for this shortly!

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

pierrejeambrun:
Looked into this and I don't think we should take a fix here. The `TODO` predates the shape that ended up becoming the execution API's error convention, and every other error response in the module today
  uses the same non-RFC-9457 shape:

  - `{"reason": "not_found" | "invalid_state" | "already_exists" | ...,  "message": "..."}` (with optional extras like `previous_state`) — used across `dag_runs.py`, `xcoms.py`, `connections.py`,
  `variables.py`, `dags.py`, `assets.py`, `asset_events.py`, `asset_state_store.py`, `hitl.py`, `connection_tests.py`, `task_state_store.py`, and `task_instances.py`.
  - Plain string `detail` in `security.py` (FastAPI default).

  Zero endpoints emit `type` / `title` / `status` / `instance`. The endpoint this issue points at is *consistent* with the rest — the `TODO` is stale.

[...]

PR #71552, opened 2026-08-13
Drop the stale RFC 9457 TODO on the task-instance-run endpoint

pierrejeambrun:
The whole execution API returns errors as `{"reason": …, "message": …, ...extras}` today — that shape is the de-facto contract, applied uniformly across every route in `airflow-core/src/airflow/api_fastapi/execution_api/routes/` (task instances, dag runs, xcoms, connections, variables, dags, assets, asset events, asset state store, hitl, connection tests, task state store), and it's parsed by the task SDK client (`task-sdk/src/airflow/sdk/api/client.py:255` raises `TaskAlreadyRunningError` on `detail["reason"] == "invalid_state"` and `detail["previous_state"] == "running"`).

The `TODO: Pass a RFC 9457 compliant error message` comment on this one 409 site predates that convention and is now misleading. Any single-endpoint migration to RFC 9457 would (a) break the SDK parser and (b) make this endpoint the only RFC-9457 endpoint in the module, which is a worse inconsistency than the current uniform shape.

Removing the comment so it stops attracting drive-by "fix RFC 9457" PRs. If we ever want the execution API on RFC 9457, that's a cross-cutting migration of every route + every SDK parser + docs — an AIP-shaped conversation, not a one-liner.

related: #69360

- [X] Yes — Claude Code (Opus 4.7)

Generated-by: Claude Code (Opus 4.7) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/task_instances.py
@@ -221,11 +221,6 @@ def ti_run(
             previous_state=previous_state,
         )
 
-        # TODO: Pass a RFC 9457 compliant error message in "detail" field
-        # https://datatracker.ietf.org/doc/html/rfc9457
-        # to provide more information about the error
-        # FastAPI will automatically convert this to a JSON response
-        # This might be added in FastAPI in https://github.com/fastapi/fastapi/issues/10370
         raise HTTPException(
             status_code=status.HTTP_409_CONFLICT,
             detail={


uros-b (approved):

uros-b:
LGTM

amoghrajesh (approved):
Yeah migrating just this endpoint would break task sdk clients of potentially all versions. Not worth the work. LGTM

Andrushika (approved):
Thanks for the PR, LGTM. That TODO comment is really misleading and led contributors to submit similar PRs that we don't need...

Andrushika (approved):
Thanks for the PR, LGTM. That TODO comment is really misleading and led contributors to submit similar PRs that we don’t need…

PR #71554, opened 2026-08-13
UI: Consolidate table controls into consistent slots and improve visual presentation

ryanahamilton:
Consolidates the shared `DataTable` chrome into a single header row above the table, and fixes the defects that surfaced along the way.

## Why

In general, the `DataTable` component is heavily used throughout the UI, so it's flaws compounded, negatively impacting the overall user experience. These changes are intended to resolve those flaws and improve the UX with a cleaner and more consistent visual presentation.

The column visibility menu was rendered inside the last `<th>`, so its position shifted with column order and it disappeared entirely whenever a table had no rows. Because hidden columns persist in `localStorage`, a user who filtered down to an empty result was left with no control that could bring them back.

Every list page also hand-rolled its own row of controls, which left the display toggle in a different row from the heading it belongs beside, and several pages duplicated the row count heading `DataTable` already renders.

## What changed

**Layout.** `DataTable` owns one header row: `[heading + headingExtra] … [actions][columns menu][display toggle]`. New `actions` and `headingExtra` slots let pages contribute their own controls (`ExpandCollapseButtons`, sort selects) rather than building a parallel row. The columns menu moved out of `TableList`, so it stays reachable on an empty table.

**Empty state** moved down into `TableList`/`CardList`. In table display the column headers stay visible and the message renders in a cell spanning every column.

**Props.** `allowFiltering` → `showColumnsMenu` (no call site passed it, and the menu toggles column *visibility*, not filtering). `showRowCountHeading` → `hideRowCountHeading`, since an optional boolean defaulting to `true` reads badly at call sites. `total` is now optional: a table whose endpoint reports no count names its model instead of claiming zero rows.

**i18n.** Six tables named their model with a page-*title* key, so counts read "1 Jobs" or "950 XCom" regardless of the count. Added proper `_one`/`_other` keys and repointed them. The count-free plural label now reads the `_other` key directly, because Russian CLDR never selects `_other` for an integer count — the form translators supplied was unreachable.

**Bug fixes.** Removed the unreachable row-expansion machinery (`getRowCanExpand`/`renderSubComponent`: `TableState` has no `expanded` field, so `getIsExpanded()` was always false and no call site passed either prop). Offset pagination no longer renders with `count={0}` on tables that report no total. `Tasks` shows table skeletons instead of 10 empty rows while loading.

### Before:

Table headers don't align with the content in their columns, `ToggleTableDisplay` is hacked into the table header row (and can be scrolled out of view), Checkboxes are visually too large.

Zight 2026-08-13 at 11 28 26 AM

### After:

file-770f6d982dc4ce6c845e60ddb452cce9

### Before - control placements:

Inconsistent placements that don't make good use of whitespace and cause jumping when toggling between card and table views.

Zight Recording 2026-08-13 at 11 34 40 AM

### After - control placements:

Utilizing consistent slots in the table component.

Zight Recording 2026-08-13 at 11 33 54 AM

### Before - empty states:

Sparse view with inconsistently placed header.

file-24b5d23044b0d05a1570b3763ffcfee8

### After - empty states:

file-63a40255f341ef02cf86c015eb0a3336

## Tests

117 files / 916 unit tests pass on this branch. Every new test was mutation-verified: each fix was reverted in turn to confirm the intended test failed and nothing else did.

`modelName.test.ts` is worth a look — it scans the source for `modelName="…"` literals and resolves each through a real i18next instance. Nothing previously validated code-side translation keys (the i18n lint rules only compare locale files against each other), which is how the six mis-keyed tables went unnoticed. It found two more on its first run.

Also fixed 11 test files whose `react-i18next` mock omitted `i18n` (latent crashes for any future test rendering a real `DataTable`), and excluded the new empty-state row from 19 e2e row locators behind a shared `DATA_ROWS` selector — otherwise several "at least one row exists" assertions would pass against an empty table.

---


airflow-core/src/airflow/ui/src/components/DataTable/DataTable.tsx
```diff
@@ -85,10 +101,10 @@ export const DataTable = <TData,>({
   onDisplayToggleChange,
   onStateChange,
   previousCursor,
+  showColumnsMenu,
   showDisplayToggle,
-  showRowCountHeading = true,
```
ryanahamilton, line 89: Replaced this with `hideRowCountHeading`. In general, optional boolean props that default to `true` are a bad practice, so I inverted it to default to `false`.

### Implementation - before:
```tsx
<DataTable showRowCountHeading={false} />
```
### Implementation - after:
```tsx
<DataTable hideRowCountHeading />
```

airflow-core/src/airflow/ui/src/components/SearchBar.tsx
```diff
@@ -112,7 +112,9 @@ export const SearchBar = ({
       <Input
         data-testid="search-dags"
         onChange={onSearchChange}
-        placeholder={`${placeholder}${hotkeyDisabled ? undefined : ` (${metaKey}${translate("search.hotkey")})`}`}
+        placeholder={
+          hotkeyDisabled ? placeholder : `${placeholder} (${metaKey}${translate("search.hotkey")})`
+        }
```
ryanahamilton, line 117: This fix is not directly related to this PR, but I found it while testing. It was introduced by me in [#71470 ](https://github.com/apache/airflow/pull/71470) yesterday.

Image

bbovenzi (approved):

bbovenzi:
Looks like our E2E tests are angry

ryanahamilton:
@bbovenzi I fixed the E2E tests, and cleaned up some orphan translation keys (hence the myriad of new reviewers requested)

bbovenzi (approved):

PR #68520, opened 2026-06-13
feat(sftp): add deferrable mode to SFTPOperator

venkata-pavani:
## Summary

- Add `deferrable` parameter to `SFTPOperator`, allowing `get` and `put` transfers of a single file (or list of single-file pairs, when `concurrency == 1`) to be deferred to a trigger instead of blocking a worker slot while the transfer is in progress.
- Add `SFTPOperatorTrigger` (in `airflow.providers.sftp.triggers.sftp`) which performs the async file transfer via `SFTPHookAsync` and yields a `TriggerEvent` on success or failure.
- `deferrable=True` raises `ValueError` at execute-time for unsupported configurations:
  - `operation == "delete"`
  - `concurrency > 1`
  - no `ssh_conn_id` available (neither explicit `ssh_conn_id` nor an `sftp_hook` with one set)

This re-implements the design from #65480 (closed unmerged due to a rebase issue), scoped down per discussion on #65475.

Closes #65475

## Test plan

- [x] New unit tests added in `providers/sftp/tests/unit/sftp/triggers/test_sftp.py` (`TestSFTPOperatorTrigger`) covering serialization, get/put success, intermediate directory creation, and failure events.
- [x] New unit tests added in `providers/sftp/tests/unit/sftp/operators/test_sftp.py` covering deferral behavior, the `ValueError` guard rails, and `execute_complete`.
- [ ] CI test suite (could not run pytest locally on this Windows checkout due to POSIX-only dependencies in airflow-core: `fcntl`, `os.register_at_fork`, and symlinked `_shared` packages).

PR #65480, opened 2026-04-19
Add deferrable mode to SFTPOperator

sunildataengineer:
What this PR does
Adds `deferrable=True` parameter to `SFTPOperator` which allows the
operator to defer execution to `SFTPOperatorTrigger`, freeing the
worker slot during file transfers instead of blocking it.

Currently, `SFTPOperator` blocks a worker slot for the entire duration
of every file transfer. For large file transfers, this wastes worker
resources unnecessarily.

Changes
- Added `SFTPOperatorTrigger` class in `triggers/sftp.py`
- Added `deferrable` parameter to `SFTPOperator.__init__()` with
  default `False` (fully backward compatible)
- Added `self.defer()` call in `execute()` when `deferrable=True`
- Added `execute_complete()` callback method to handle trigger result
- Added 3 unit tests covering deferrable mode

Tests
- `test_sftp_operator_defers_when_deferrable_true`
- `test_sftp_operator_execute_complete_success`
- `test_sftp_operator_execute_complete_raises_on_error`

Closes #65475

PR #71557, opened 2026-08-13
Add deferrable mode to SFTPOperator

### LINKED ISSUE #65475: Add deferrable mode to SFTPOperator
### Description

The SFTPOperator currently runs in synchronous mode only, blocking 
a worker slot for the entire duration of every file transfer. 
For large file transfers, this wastes worker resources unnecessarily.

Proposed Solution
Add deferrable=True parameter to SFTPOperator following the same 
pattern used in other Airflow providers. When deferrable=True, 
the operator will defer execution to a trigger, freeing the worker 
slot immediately and allowing the Triggerer to poll for completion.

 Implementation Plan
- Add SFTPOperatorTrigger class in triggers/sftp.py
- Add deferrable param to SFTPOperator.__init__()
- Add defer() call in execute() when deferrable=True  
- Add execute_complete() callback method
- Add unit tests

 Related
Similar to deferrable mode already implemented in SFTPSensor.

I am willing to submit a PR for this.

### Use case/motivation

_No response_

### Related issues

_No response_

### Code of Conduct

tannnmayy:
Add optional deferrable mode to `SFTPOperator` so a single-file (or list of file pairs) get/put transfer can run on the Triggerer instead of occupying a worker slot for the whole transfer.

closes: #65475

## Why

`SFTPOperator` is synchronous today. Large transfers hold a worker for the entire copy. `SFTPSensor` is already deferrable; this applies the same operator/trigger split to get/put.

Three earlier attempts (#65480, #68520, #68298) did not land. #68298 was closed for a polluted diff and a security regression in `hooks/sftp.py`. This branch is a fresh rebase onto current `main` and contains only the feature.

## What this PR does

- Add `deferrable` to `SFTPOperator`, defaulting to `[operators] default_deferrable` (`False` if unset)
- Add `SFTPTransferTrigger` in the existing triggers module. `provider.yaml` already lists that module, so no extra registration is needed
- `execute()` validates first, then `self.defer(...)` when `deferrable=True`. The existing synchronous transfer loop is not rewritten
- `execute_complete()` resumes from the trigger event
- Unit tests for serialize/run, deferral, guard rails, `execute_complete`, and the default sync path

Unsupported in this first cut (raise `ValueError` before deferring):

- `operation="delete"`
- `concurrency > 1`
- directory PUT (local path is a directory)
- missing `ssh_conn_id` (a hook object cannot be serialized)

## Intentional limits

- Local paths are read/written on the Triggerer host. Workers and Triggerer must share that filesystem
- Directory GET is not pre-checked (that would require a worker-side SFTP call); `retrieve_file` fails at trigger time
- `SFTPHookAsync.store_file` already creates missing remote parents, so PUT may create remote directories even when `create_intermediate_dirs=False`
- `confirm` and `prefetch` are not on the async hook methods, so they are unused in the deferred path
- `hooks/sftp.py` is untouched, including `_validate_within_directory`
- No new `AirflowException` (allowlist stays at 2). Trigger/operator errors use `ValueError`
- No `pyproject.toml`, generated docs, newsfragment, `constants.py`, or `uv.lock` changes

## Tests

- Isolated checks of trigger serialize/run and operator defer/guard/`execute_complete` (this checkout cannot import Airflow on native Windows: `os.register_at_fork`, `_shared` gitlinks)
- `ruff check` / `ruff format --check` on the four files
- `git diff --check`

CI on Linux should run:

- `pytest providers/sftp/tests/unit/sftp/triggers/test_sftp.py`
- `pytest providers/sftp/tests/unit/sftp/operators/test_sftp.py`

Draft until CI is green.

- [x] Yes (Grok 4.6 / xAI)

Generated-by: Grok 4.6 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


sunildataengineer:
"Good to see this feature moving forward in #71557. I spent months on this in my PR, hit some walls with the implementation details and review cycles, but the work was worth doing. @tannnmayy  fresh approach looks solid — clean rebase, proper scope, and learning from what didn't land before.

I'm rooting for this to get merged. Happy to help test or review if needed."

dabla (changes requested):
### 🔴 Critical

**1. `SFTPHookAsync` is untouched — deferrable mode offers no real performance improvement**
`providers/sftp/src/airflow/providers/sftp/hooks/sftp.py` (unchanged from `main`)

This PR adds a trigger and wires up `deferrable=True` on the operator, but `hooks/sftp.py` is **byte-for-byte identical to `main`**. As a result the deferrable path:

- Opens a **new SSH connection + new SFTP session on every individual file transfer**
  (`retrieve_file()` lines 878–880, `store_file()` lines 915–916 each do
  `async with await self._get_conn()` + `start_sftp_client()`).
- Transfers files **sequentially** in a plain `for` loop in the trigger — no
  `asyncio.gather`, no semaphore, no bounded concurrency.
- Has no native `async def transfer()` on the hook — transfer logic is reimplemented
  inside the trigger instead of being shared with the sync path.

The net effect is: the worker slot is freed while waiting, but the actual transfer
throughput is the same as (or worse than) the sync path because of per-file connection overhead and no parallelism.

There is also a deeper architectural problem: **the current `SFTPOperator` violates the Airflow operator design principle that operators must be pure facilitators — they must contain zero business logic and delegate everything to the hook.** Looking at the current operator, file transfer logic (path handling, remote dir creation, stat calls, looping over files) lives directly in the operator rather than in the hook. This is the root cause of the duplication: once deferrable mode is added, the trigger has to reimplement the same logic all over again because there is no central hook method to call.

**Please look at [PR](https://github.com/apache/airflow/pull/68298) #68298** by @sunildataengineer as a reference for how this should be done:
- A proper `async def transfer()` method was added to `SFTPHookAsync`, encapsulating all
  transfer logic (connection reuse, stat, path resolution, directory handling).
- `asyncio.gather` + a semaphore were used for bounded concurrent multi-file transfers.
- The operator was cleaned up to be a thin facilitator that simply calls
  `hook.transfer(...)` — both the sync and deferrable paths call the same hook method,
  which is the only correct way to keep behavior consistent and the code DRY.

Without those hook-level changes, every new execution mode (deferrable, async, future modes) will keep reimplementing the same transfer logic independently, and bugs or behavior differences will accumulate across paths. This PR should not land until the hook is properly structured first.

---

### 🔴 Major

**2. Deferrable PUT silently changes behavior: `create_intermediate_dirs` ignored**
`providers/sftp/src/airflow/providers/sftp/triggers/sftp.py:218–219`

The deferrable PUT always calls `SFTPHookAsync.store_file()`, which auto-creates remote parent directories (`hooks/sftp.py:907–920`). The synchronous path fails when parent dirs are missing unless `create_intermediate_dirs=True`. This is a silent behavioral regression — users relying on `create_intermediate_dirs=False` (the default) get different results depending on whether `deferrable=True`.

**3. Deferrable mode silently drops `confirm` and `prefetch`**
`operators/sftp.py:162–170`, `triggers/sftp.py:153–154, 217–219`

`confirm` and `prefetch` are accepted by the operator but never forwarded to the trigger or honored by the async path. `deferrable=True` therefore changes transfer semantics, not just scheduling behavior. The operator should raise `ValueError` upfront when unsupported non-default values are passed together with `deferrable=True`.

---

### 🟡 Minor

**4. Docstring claims directory PUT raises `ValueError` — directory GET does not**
`operators/sftp.py:82–86, 274–286` / `triggers/sftp.py:211–217`

The new docstring states that directory transfers in deferrable mode "raise `ValueError`", but only PUT is validated before deferral. A directory GET still defers and fails later inside the trigger, giving a confusing error.

**5. Test asserts on raw `caplog.text`**
`tests/unit/sftp/operators/test_sftp.py:857–859`

Airflow's test guidelines deprecate raw log-text matching. Prefer structured caplog
assertions: `{"event": ..., "field": ...} in caplog`.

---

### ⚠️ Unaddressed comments from previous PRs (#65480, #68298, #68520)

**A. Duplicate transfer logic in the trigger (blocker from #65480 / #68298)**
`triggers/sftp.py:200–221`

Reviewers repeatedly asked to avoid re-implementing transfer logic inside the trigger and instead delegate to hook-level transfer logic. The new PR still contains a second transfer implementation, so sync/deferrable behavior can drift over time. This was a repeated blocker that is still not resolved.

**B. `prefetch` silently dropped (blocker from #65480, @srchilukoori)**
`operators/sftp.py:162–170` / `triggers/sftp.py:153–154, 217–219`

The original blocker about dropped parameters still applies: `prefetch` is dropped in
deferrable mode, and `confirm` is now dropped too.

**C. Directory handling parity (#65480)**
`operators/sftp.py:274–286` / `triggers/sftp.py:211–219`

Directory handling remains inconsistent between sync and deferrable paths. The concern raised in #65480 is only partially addressed.

---

### ✅ What looks good
- License headers present
- No direct `raise AirflowException(...)`
- No `time.time()` used for duration logic
- Commit messages follow Airflow conventions
- Trigger and operator scaffolding is present
- `provider.yaml` updated

---

**Top priorities before merge:** Critical #1 (the hook-level async optimisations from #68298 must land first — without them this PR is just scaffolding), then #3 (semantics change, silent param drop) and unaddressed blocker **B** from #65480.

---
Drafted-by: Claude Sonnet 4.6 and reviewed by @dabla before posting

dabla:
> "Good to see this feature moving forward in #71557. I spent months on this in my PR, hit some walls with the implementation details and review cycles, but the work was worth doing. @tannnmayy fresh approach looks solid — clean rebase, proper scope, and learning from what didn't land before.
> 
> I'm rooting for this to get merged. Happy to help test or review if needed."

I've re-opened a new [PR](https://github.com/apache/airflow/pull/72336) which recovers your work.

tannnmayy:
Closing this in favour of #72336.

@dabla was right that the hook needed restructuring first. Without an `async def transfer()` on `SFTPHookAsync`, the deferrable path was always going to end up duplicating transfer logic in the trigger and freeing the worker slot without actually improving throughput. #72336 does that properly, and building on @sunildataengineer's original work is the right foundation for it.

Thanks @dabla for the unusually detailed review — the breakdown of the `create_intermediate_dirs` / `confirm` / `prefetch` divergence between the sync and deferred paths was the part I had wrong, and it was useful to have it spelled out. And thanks @sunildataengineer for the kind words earlier in this thread.

I'll follow #72336 and am happy to help test it if that's useful.

PR #68547, opened 2026-06-14
feat(ui): add aggregate Dag schedule overview view (#68532)

anmolxlight:
Fixes #68532

## What

Adds a new `/dags/schedule` page that visualizes the **typical (mean / median) start and end time of each Dag over a 24-hour timeline**, aggregated from recent successful Dag runs. This complements the existing per-Dag Gantt view, which only shows a single run at a time. The new view operates at a different level: it aggregates across all Dags to answer "across my whole deployment, when during the day does each Dag usually run?"

## Why

Operators running many Dags currently have no at-a-glance way to see how scheduled load is distributed across the day. The new view makes the whole schedule visible at once, which helps to:
- spot overlapping heavy windows where many Dags pile up at the same hour,
- find quiet windows for maintenance / deploys / resource-intensive jobs,
- notice drift when a Dag's typical run time shifts over recent runs,
- plan capacity and SLAs across the fleet.

### Backend
- New `/ui/dag_schedule_overview` endpoint (in `api_fastapi/core_api/routes/ui/dag_schedule_overview.py`) gated by `requires_access_dag(method="GET", access_entity=DagAccessEntity.RUN)`.
- New `DagScheduleOverviewService` (`api_fastapi/core_api/services/ui/dag_schedule_overview.py`) that:
  - fetches all DagModel rows (so dags with no runs still appear as zero-statistic rows),
  - pulls up to 200 most-recent successful `DagRun`s per dag, optionally scoped by a `run_after` window,
  - computes mean/median of `start_date` and `end_date` mapped to seconds-of-day in UTC, plus mean/median duration,
  - supports `dag_id_pattern` and `dag_display_name_pattern` filters.
- New datamodels `DagScheduleOverviewEntry` + `DagScheduleOverviewCollectionResponse` (`api_fastapi/core_api/datamodels/ui/dag_schedule_overview.py`).
- 11 backend tests in `tests/unit/api_fastapi/core_api/routes/ui/test_dag_schedule_overview.py`: 200/401/403, all-three-dag fixture, morning/evening/no-runs statistics, FAILED-run exclusion, pattern filters, run_after window edges.
- OpenAPI `_private_ui.yaml` updated; `pnpm codegen` regenerated.

### Frontend
- New page `src/pages/DagsList/ScheduleOverview/` mounted at `/dags/schedule` with:
  - one row per Dag (linkable to the Dag detail page),
  - a 24h tick rail at 0/3/6/9/12/15/18/21/24 hours,
[...]

Issue #68532, opened 2026-06-14
Add an aggregate Dag schedule view: typical daily run times across all Dags


choo121600:
### Description

Add a built-in UI view that visualizes the **typical (mean / median) start and end time of each Dag over a 24-hour timeline**, aggregated from recent successful Dag runs. 
This is **not** the existing Gantt view. The current Gantt is *per Dag run* — it shows when each task started/ended within a single run. 
The proposed view operates at a different level: it aggregates across **all Dags** to answer "across my whole deployment, when during the day does each Dag usually run?"

#### DEMO
<img width="2168" height="1402" alt="Image" src="https://github.com/user-attachments/assets/6db18d2f-5561-4b8b-996b-6bda55f83c7d" />

### Use case/motivation

Operators running many Dags currently have no at-a-glance way to see how scheduled load is distributed across the day. To understand it today you have to open each Dag's runs one by one. 
This view makes the whole schedule visible at once, which helps to: 
- spot **overlapping heavy windows** where many Dags pile up at the same hour, 
- find **quiet windows** for maintenance / deploys / resource-intensive jobs, 
- notice **drift** when a Dag's typical run time shifts over recent runs, 
- plan capacity and SLAs across the fleet.

### Discussion before implementation

Before implementing this feature, I'd like to gather feedback on whether this functionality would be valuable to the broader Airflow community and, if so, what the best UI/UX approach would be.

In particular, it would be helpful to discuss where this feature should live in the UI, how it should be presented to users, and whether the proposed design solves the underlying problem in the most intuitive way.

I'd prefer to reach some level of consensus on these questions before moving forward with implementation.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

venkatamandavilli-code:
Thanks for raising this enhancement.

In large enterprise environments with hundreds or even thousands of DAGs, it can be difficult for platform teams to understand workload distribution throughout the day. An aggregate schedule view would provide valuable operational visibility by helping teams identify peak execution windows, scheduling bottlenecks, and opportunities to better distribute workloads across available resources.

[...]

PR #71558, opened 2026-08-13
UI: Add cross-Dag Time Schedule view

minyeamer:
Add a **Time Schedule** tab for comparing Dag start times and durations across an Airflow environment.

Features:

- Day and Week timelines with overlap handling and fixed headers
- Mean, Max, and Min aggregation within zoom-dependent time buckets
- Pointer-anchored zoom using buttons, the mouse wheel, or arrow keys
- Existing Dag run, tag, timetable type, and optional team filters
- Configurable limits from the latest 200 runs through all matching runs
- Contiguous API pagination using the configured page size
- Gray expected-run bars for scheduled Dags without matching Dag runs
- Readable tooltips and links to the relevant Dag or Dag run
- Saved view, aggregation, run-limit, and scheduled-only preferences
- Light and dark theme support

This is a UI-only change. It reuses existing endpoints and adds no schema or backend API changes.

Earlier work in #68532 and #68547 informed the time-bucket design. The view placement follows #45497, while bounded loading follows the work in #47900, #60241, and #65388.

Related: #22001

---

Screenshots:

- Day view:

time_schedule-day-tooltip<br>

- Week view:

time_schedule-week-tooltip

---

Videos:

- Zoom (Day view):

https://github.com/user-attachments/assets/91fb712a-7492-43ed-a5b0-969f4ddbda36

- Zoom (Week view):

https://github.com/user-attachments/assets/30f2dfb1-a0d0-4250-92f0-024968f457d8

- Dag run links (Day view):

https://github.com/user-attachments/assets/a7a306b7-e816-48aa-8659-b2a59962b834

- Dag run limits (Week view):

https://github.com/user-attachments/assets/b3cf1d48-eda5-4751-b194-ab2e0b8ca16d

- Aggregation (Day view):

https://github.com/user-attachments/assets/5ce3fff4-9d5c-4ca4-bd47-8e3a943a5600

---

- [x] Yes (GPT-5.6 Sol)

Generated-by: Codex (GPT-5.6 Sol) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---


bbovenzi:
Very cool! I could see a world where we could even show arrows between runs that are connected via partition keys. Or highlighting when Max active dag runs is being hit.

Love the minute detailing. That's great for dags that run every minute! Just needs a label.

1. Let's move this as a tab inside the Dashboard home page instead of in the tabled dags/runs/tis view.
2. Let's add the state icons to the bars so that they can be distinguished by something other than color
3. We need labels and/or tooltips to help explain some of these filters:
a) Latest 200 should probably be "Limit" to be consistent with other views
b) Mean/Max/Min means nothing to me without context
4. Day/Week modes are different types of views not a simple setting. It should probably match the tabs we have to change the calendar view on the dag details page.
5) The filters are a bit inconsistent. We have the Filter Pill UX that is used elsewhere but separately we also have timetable filters, scheduled dags only filter, and tag filters. Let's consolidate them all together. I also recommend rebasing since @ryanahamilton's recent changes add some great examples.

airflow-core/src/airflow/ui/src/pages/TimeSchedule/useTimeScheduleData.ts
```diff
@@ -0,0 +1,287 @@
...
+  const additionalDagPages = useQueries({
+    queries: getAdditionalPageOffsets({
+      pageSize,
+      requestedEntryCount: dagsData?.total_entries ?? 0,
+      totalEntries: dagsData?.total_entries ?? 0,
+    }).map((offset) => ({
+      queryFn: () => DagService.getDagsUi({ ...dagQueryParams, offset }),
+      queryKey: ["time-schedule-dags", dagQueryParams, offset],
+    })),
+  });
```
bbovenzi: Oh we're just firing off tons of API requests per dag? That is not going to scale well at all. This will need a custom API endpoint to more efficiently pull this data.

airflow-core/docs/ui.rst
```diff
@@ -461,6 +461,100 @@ Displays runtime metadata about the task instance, including:
...
+
+The Time Schedule view shows typical Dag start times and durations across an environment. Dag runs are grouped by time
+of day, making schedule patterns and congested periods easier to identify.
+
+Display Options
+'''''''''''''''
+
+- **Latest 200** loads the 200 most recent Dag runs by default.
+- **Latest 600**, **Latest 1000**, **Latest 2000**, and **Latest 5000** increase the number of runs loaded.
+- **All Dag runs** loads all matching runs.
```
bbovenzi: There should be no All Dag Runs option. This table can have millions of entries

airflow-core/src/airflow/ui/src/pages/TimeSchedule/useTimeScheduleData.ts
bbovenzi: You should look at how we stream the data for the grid view. I think we will want to stream this data in batches as well so that the page can show some useful information while other parts load.

airflow-core/src/airflow/ui/src/pages/TimeSchedule/useTimeScheduleData.ts
```diff
@@ -0,0 +1,287 @@
...
+  const visibleDagIds = new Set(scheduledDags.map((dag) => dag.dag_id));
+  const metadataFilterApplied =
+    selectedTags.length > 0 || selectedTeams.length > 0 || selectedTimetableTypes.length > 0;
+  const visibleDagRuns = [
+    ...(dagRunsData?.dag_runs ?? []),
+    ...additionalDagRunPages.flatMap((page) => page.data?.dag_runs ?? []),
+  ].filter(
+    (dagRun) =>
+      (!metadataFilterApplied || visibleDagIds.has(dagRun.dag_id)) &&
+      (!showScheduledOnly || scheduledDagIds.has(dagRun.dag_id)),
```
bbovenzi: Filtering needs to be done server-side too.

airflow-core/src/airflow/ui/src/pages/TimeSchedule/useTimeScheduleData.ts
```diff
@@ -0,0 +1,287 @@
...
+    timeScale,
+    viewMode: "day",
+  });
+  const aggregatedWeekItems = aggregateTimelineItems({
+    aggregationMode,
+    items: timelineItems,
+    selectedTimezone,
+    timeScale,
+    viewMode: "week",
+  });
```
bbovenzi: When we move this to the backend. We would only want to return a single day or week mode at a time. No need to have a whole expensive aggregation built if the user doesn't even use it.

bbovenzi (changes requested):
This is a cool demo. But it definitely needs some work to "productionize" the code to be usable for deployments with any degree of scale.

minyeamer:
@bbovenzi
Thanks for the detailed review. I interpreted the four main concerns as scalability and
data-flow issues rather than only UI adjustments.

Since b13f993, excluding merge commits, I made the following changes:

- Replaced client-side pagination and request fan-out with a dedicated streaming endpoint.
- Removed the unbounded "All Dag runs" option and enforced a maximum of 5,000 runs on the backend.
- Moved Dag and Dag run filtering into the backend query construction.
- Aggregate only the selected Day or Week view on the server.
- Stream results in batches so the UI can render progressively.
- Updated the filter implementation after the shared FilterBar changes were merged.
- Kept the private OpenAPI contract and generated client aligned with the backend route.

The current flow is:

```mermaid
flowchart TD
    UI[Time Schedule UI] --> Hook[useTimeScheduleData]
    Hook -->|One GET request| API[GET /ui/time-schedule]
    API --> Filter[Apply authorization and filters in SQL]
    Filter --> Limit[Select up to 5,000 Dag runs]
    Limit --> Batch[Process 25 Dags per batch]
    Batch --> Aggregate[Aggregate selected Day or Week view]
    Aggregate --> Stream[Return NDJSON TimeScheduleBatch]
    Stream --> Hook
    Hook --> Render[Progressively render timeline bars]
```

The browser now receives only the filtered data needed for the selected view, while the backend controls the result size and aggregation cost.

I initially intended this change to remain UI-only and did not plan to add a backend API. Because the review identified the scalability limitations of the original approach, I needed to first understand Airflow's backend API structure, streaming session lifecycle, authorization filters, and OpenAPI generation flow. I also spent additional time reviewing the AI-assisted code carefully, which is why the follow-up changes took about 5 days.

The implementation is now focused on the production-scale concerns from the review while preserving the existing Airflow UI/API patterns.

minyeamer:
I missed these additional suggestions in my previous follow-up. I plan to address the following:

- Move Time Schedule into the Dashboard as a tab.
- Add state icons to timeline bars so states are not distinguished by color alone.
- Improve labels and tooltips:
  - Rename `Latest 200` to `Limit 200`.
  - Add context for the duration aggregation options.
- Replace the Day/Week selector with tabs.
- Consolidate the remaining filters through the shared Filter Pill UX.

For the partition-key arrows, I understand this may include dependencies between Dag runs, such as those created through `TriggerDagRun`. However, with the timeline sorted alphabetically by `dag_id`, the connections could easily cross and become difficult to follow.

I’m happy to implement this with more specific guidance on which relationships to show and how they should be laid out. With the current context, I’m hesitant to add a potentially confusing visualization.

---
Drafted-by: GPT-5

minyeamer:
@bbovenzi I've addressed the remaining Time Schedule suggestions and wrapped up the final refinements.

### Dashboard entry point

Time Schedule is no longer surfaced from the Dags area. It is now available from the **Stats** section on the Dashboard home page.

home_stats

### State icons on timeline bars

Timeline bars now include the shared Airflow state icons alongside their labels, so Dag run state is distinguishable without relying on color alone. The same icon treatment is also used in the bar tooltip.

Screenshot 2026-09-06 at 8 58 44 PM

### Labels and tooltips

- **Latest 200** is now **Limit 200**.
- The aggregation choices now use descriptive names: **Average duration**, **Full time range**, and **Shortest run**.
- Added contextual help for zoom controls, the Dag run limit, and duration aggregation.

Screenshot 2026-09-06 at 8 59 19 PM

Screenshot 2026-09-06 at 9 00 21 PM

Screenshot 2026-09-06 at 9 00 30 PM

### Day and Week as view tabs

Day and Week are now presented as tabs rather than as a simple setting, matching the interaction pattern used for calendar-style view changes elsewhere in the UI.

Screenshot 2026-09-06 at 9 01 31 PM

### Documentaion

I also updated the UI documentation and screenshots to reflect the Dashboard entry point and the current Time Schedule controls.

PR #66810, opened 2026-05-12
feat(metrics): emit pool slot histograms alongside existing gauges

1fanwang:
On our LinkedIn DI deployment we run schedulers in HA, and the pool slot gauges are unreliable there. Each scheduler runs `_emit_pool_metrics` independently and samples the metadata DB a moment apart, so they emit different counts for the same metric (e.g. one reports `open_slots=128`, another `126`). The pool metrics are tagged only by `pool_name` — there is no per-scheduler attribute — so on OpenTelemetry the per-scheduler series collide and the backend keeps whichever export landed last. `pool.open_slots` then flaps between schedulers' samples and is often wrong.

### Fix

Emit a `stats.timing` (histogram) alongside each existing `stats.gauge` — same value, same `pool_name` tag, name = the gauge name + a `_histogram` suffix:

- `pool.open_slots_histogram`
- `pool.queued_slots_histogram`
- `pool.running_slots_histogram`
- `pool.deferred_slots_histogram`
- `pool.scheduled_slots_histogram`

A histogram accumulates every scheduler's sample in the interval instead of overwriting, so the backend can aggregate to the correct value — `min` for `open_slots`, `max` for `queued`/`running`/`scheduled`/`deferred` — per interval. The gauges are unchanged, so existing single-scheduler scrapers keep working. (A histogram also avoids the series-cardinality blow-up of adding a per-scheduler tag to the gauge — it folds every scheduler into one series per pool.)

`stats.timing` is the histogram primitive in the existing stats API (statsd `timer`/`|ms`; OTel `Histogram`), registered as `type: timer` in `metrics_template.yaml`.

### Tests

Unit: `test_emit_pool_metrics_emits_gauge_and_histogram` asserts the gauge and histogram fire with the same value and tags for all five slot states.

End-to-end against the real OTel logger — two schedulers report divergent samples for the same series (`B=126`, `A=128`) into one in-memory reader (the backend):

```
BEFORE - gauge pool.open_slots:               backend keeps last writer = 128   (WRONG)
AFTER  - histogram pool.open_slots_histogram: count=2 sum=254 min=126 max=128
         take min(open_slots) = 126           (CORRECT - both samples preserved)
```

The gauge loses all but the last scheduler's value; the histogram preserves every sample, so the true `min`/`max` is recoverable. (StatsD: the same `_histogram` metric lands as a `|ms` timer, verified on the wire.)

Closes #66800

---

> [!IMPORTANT]
[...]

PR #71561, opened 2026-08-13
Emit pool slot histograms alongside the existing gauges

### LINKED ISSUE #66800: Emit histograms alongside pool slot gauges to capture distribution between scrapes
### What's the problem

In HA (multi-scheduler) deployments the pool slot gauges are unreliable. Each scheduler runs `_emit_pool_metrics` independently and samples the metadata DB a moment apart, so they emit different counts for the same metric (e.g. one reports `open_slots=128`, another `126`). The pool metrics carry only a `pool_name` tag — there is no per-scheduler attribute — so on OpenTelemetry the per-scheduler series collide and the backend keeps whichever export arrived last. `pool.open_slots` (and the other slot gauges) then flap between schedulers' samples and can be wrong.

### Proposal

Emit a histogram alongside each existing pool slot gauge — same value, same tag, `_histogram` suffix. A histogram accumulates every scheduler's sample in the interval rather than overwriting, so operators can aggregate to the correct value — `min(open_slots)`, `max(queued/running/...)` per interval. The gauges stay unchanged for backwards compatibility.

Verified end-to-end through the real OTel logger: with two schedulers reporting `126` and `128` for the same series, the gauge keeps only the last (`128`, wrong) while the histogram preserves both (`min=126`, correct).

Implemented in #66810.

1fanwang:
In HA deployments every scheduler runs `_emit_pool_metrics` on its own loop and samples the metadata DB a moment apart. Pool metrics are tagged only by `pool_name` (plus `team_name` under multi-team), so every scheduler publishes the *same* series and a gauge keeps whichever export landed last. `pool.open_slots` flaps between schedulers' samples and is frequently wrong.

Tagging the gauge per scheduler would fix the collision but multiplies pool series by the scheduler count. A histogram instead folds every scheduler into one series per pool and retains every sample, so the true value stays recoverable — `min` for `open_slots`, `max` for the queued/running/deferred/scheduled counts.

This emits a `stats.timing` alongside each existing gauge: same value, same tags, `_histogram` suffix. The gauges are unchanged, so existing scrapers keep working.

An unbounded pool (`slots=-1`) reports `open=inf`, which is skipped rather than recorded — OpenTelemetry drops a non-finite observation but logs a warning on every scheduler emission, and StatsD would put a literal `inf|ms` on the wire.

closes: https://github.com/apache/airflow/issues/66800

Replaces https://github.com/apache/airflow/pull/66810, which was approved by @xBis7 and @ferruzzi and then auto-closed as a stale draft. Both review points raised there are folded in here: the registry descriptions now state these are unit-less slot counts rather than durations, and the infinite-pool case is handled.

Still open for a wider discussion (not this PR): `timing` is the only histogram primitive the stats API exposes, so a slot count rides the timer type. On OpenTelemetry that is the intended `Histogram`; on StatsD it renders as `|ms`.

# Testing Done

`_emit_pool_metrics` is exercised against a real `Pool` row, and the collision itself is reproduced through the real `SafeOtelLogger` and OpenTelemetry SDK.

| # | Scenario | Result |
|---|---|---|
| 1 | Two schedulers report divergent `open_slots` for one pool | Gauge keeps last writer (wrong); histogram preserves both samples (correct) |
| 2 | Gauge/histogram pair for all five slot states | Same value and tags |
| 3 | Unbounded pool (`slots=-1`) | Gauge carries `inf`; `open_slots_histogram` not emitted |

**1. The bug, through the real OTel logger.** Two `SafeOtelLogger` instances share one `InMemoryMetricReader`, standing in for two schedulers exporting to one backend:

```python
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import InMemoryMetricReader

from airflow_shared.observability.metrics.otel_logger import SafeOtelLogger

reader = InMemoryMetricReader()
provider = MeterProvider(metric_readers=[reader])

scheduler_a = SafeOtelLogger(otel_provider=provider)
scheduler_b = SafeOtelLogger(otel_provider=provider)

tags = {"pool_name": "default_pool"}
for scheduler, open_slots in ((scheduler_b, 126), (scheduler_a, 128)):
    scheduler.gauge("pool.open_slots", open_slots, tags=tags)
    scheduler.timing("pool.open_slots_histogram", open_slots, tags=tags)

# an unbounded pool would report inf into the same histogram
scheduler_a.timing("pool.queued_slots_histogram", 4, tags=tags)
scheduler_a.timing("pool.queued_slots_histogram", float("inf"), tags=tags)

for metric in reader.get_metrics_data().resource_metrics[0].scope_metrics[0].metrics:
    point = next(iter(metric.data.data_points))
    if metric.name.endswith("_histogram"):
        print(f"{metric.name}: count={point.count} sum={point.sum} min={point.min} max={point.max}")
    else:
        print(f"{metric.name}: value={point.value}")
```

```
$ uv run --project shared/observability python dev/repro_pool_histogram.py
Record amount inf is not finite on Histogram airflow.pool.queued_slots_histogram, ignoring measurement.
airflow.pool.open_slots: value=128
airflow.pool.open_slots_histogram: count=2 sum=254.0 min=126.0 max=128.0
airflow.pool.queued_slots_histogram: count=1 sum=4.0 min=4.0 max=4.0
```

Scheduler B's `126` is gone from the gauge — the backend reports `128`. The histogram kept both, so `min` recovers the correct `126`.

The `inf` sample also shows why the guard is worth having: the SDK discards the measurement but logs that line on every scheduler emission. StatsD has no such filter — the raw client puts it on the wire:

```
$ python -c "import statsd; ...; real.timing('pool.open_slots_histogram', float('inf'))"
statsd wire payload: call(b'pool.open_slots_histogram:inf|ms', ('127.0.0.1', 8125))
```

</details>

**2. Regression tests, red then green.** Both new tests fail on unmodified `scheduler_job_runner.py`:

```
$ git checkout origin/main -- airflow-core/src/airflow/jobs/scheduler_job_runner.py
$ uv run --project airflow-core pytest airflow-core/tests/unit/jobs/test_scheduler_job.py -k histogram -q
E   AssertionError: timing('pool.queued_slots_histogram', 0, tags={'pool_name': 'unbounded_pool'}) call not found
FAILED ...::test_emit_pool_metrics_emits_histogram_alongside_gauge
FAILED ...::test_emit_pool_metrics_skips_open_slots_histogram_for_unbounded_pool
2 failed, 409 deselected, 1 warning in 5.76s

$ git checkout HEAD -- airflow-core/src/airflow/jobs/scheduler_job_runner.py
$ uv run --project airflow-core pytest airflow-core/tests/unit/jobs/test_scheduler_job.py -k emit_pool_metrics -q
airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_emit_pool_metrics_team_name[with_team] PASSED
airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_emit_pool_metrics_team_name[without_team] PASSED
airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_emit_pool_metrics_emits_histogram_alongside_gauge PASSED
airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_emit_pool_metrics_skips_open_slots_histogram_for_unbounded_pool PASSED
4 passed, 407 deselected, 1 warning in 18.60s
```

</details>

Registry sync and type checks are clean (`check-metrics-synced-with-registry`, `mypy-airflow-core`, `prek run --from-ref origin/main --stage pre-commit`).

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import itertools
 import logging
+import math
 import multiprocessing
 import operator
 import os
@@ -3356,6 +3357,15 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             stats.gauge("pool.running_slots", slot_stats["running"], tags=metric_tags)
             stats.gauge("pool.deferred_slots", slot_stats["deferred"], tags=metric_tags)
             stats.gauge("pool.scheduled_slots", slot_stats["scheduled"], tags=metric_tags)
+
+            # An unbounded pool (slots=-1) reports open=inf, which would poison the
+            # histogram's sum and max for every other sample in the bucket.
+            if math.isfinite(slot_stats["open"]):
+                stats.timing("pool.open_slots_histogram", slot_stats["open"], tags=metric_tags)
+            stats.timing("pool.queued_slots_histogram", slot_stats["queued"], tags=metric_tags)
+            stats.timing("pool.running_slots_histogram", slot_stats["running"], tags=metric_tags)
+            stats.timing("pool.deferred_slots_histogram", slot_stats["deferred"], tags=metric_tags)
+            stats.timing("pool.scheduled_slots_histogram", slot_stats["scheduled"], tags=metric_tags)
 
     @provide_session
     def adopt_or_reset_orphaned_tasks(self, *, session: Session = NEW_SESSION) -> int:
--- a/airflow-core/tests/unit/jobs/test_scheduler_job.py
+++ b/airflow-core/tests/unit/jobs/test_scheduler_job.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import contextlib
 import datetime
 import logging
+import math
 import os
 import re
 import time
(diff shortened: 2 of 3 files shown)


1fanwang:
Closing this - double checked the collision it fixes is already solvable with standard OTel configuration

Airflow builds the meter provider with `Resource.create()` ([otel_logger.py:465](https://github.com/apache/airflow/blob/main/shared/observability/src/airflow_shared/observability/metrics/otel_logger.py#L465)), which merges `OTEL_RESOURCE_ATTRIBUTES`. Give each scheduler a distinct `service.instance.id` and the per-scheduler samples become separate series, so nothing is overwritten:

```
airflow.pool.open_slots{service.instance.id=scheduler-a, pool_name=default_pool} = 128
airflow.pool.open_slots{service.instance.id=scheduler-b, pool_name=default_pool} = 126
```

`min by (pool_name)` then recovers the same 126 the histogram was there to provide. `service.instance.id` is also what the spec maps to Prometheus' `instance` label, so backends already understand it — and it covers every gauge Airflow emits, not the five pool ones.

I also had the cardinality argument backwards in the description above: histograms here use exponential bucket aggregation, so they cost considerably more stored series than N per-instance gauges, not fewer. That was the main justification for preferring a histogram over a per-scheduler dimension, and it doesn't hold.

What's actually missing is documentation. Nothing in the metrics docs tells an HA operator that every scheduler publishes the same series, or how to separate them. Sending a docs PR instead.

PR #71564, opened 2026-08-13
Document service.name and service.instance.id for OpenTelemetry metrics

1fanwang:
# Rationale for this change

OpenTelemetry identifies a metric's source by two resource attributes: `service.name` says which component reported it, `service.instance.id` says which instance of that component. Airflow honours both, and the metrics docs mention neither.

Without them every process reports as `airflow`, so a deployment running several processes of one component publishes to the same series and the backend keeps whichever export arrived last. Gauges show it most clearly: each scheduler samples the metadata database on its own loop, so `pool.open_slots` and its siblings look like they are flapping.

No code change is needed. The gap is that nothing in the docs tells an operator to set them, or which variable sets which.

# What changes are included in this PR?

A subsection of *Setup - OpenTelemetry* covering what each attribute identifies, the environment variable behind each, how the samples are combined at query time, and how the two surface as `job` and `instance` on Prometheus-compatible backends.

A note in the StatsD section that StatsD has no resource concept, so several schedulers overwrite each other's gauges, pointing at the OpenTelemetry subsection.

Docs only, one file.

# Are these changes tested?

The documented mechanism was run rather than assumed.

**Airflow honours both variables.** `OTEL_SERVICE_NAME` is read in `otel_env_config.py` and takes precedence over the `otel_service` option; every other attribute arrives through `Resource.create()`, which merges `OTEL_RESOURCE_ATTRIBUTES`:

```console
$ OTEL_SERVICE_NAME="airflow-scheduler" \
  OTEL_RESOURCE_ATTRIBUTES="service.instance.id=scheduler-a" python -c "
from opentelemetry.sdk.resources import SERVICE_NAME, Resource
r = Resource.create(attributes={SERVICE_NAME: 'airflow'})
for k, v in sorted(r.attributes.items()): print(f'{k} = {v}')"
service.instance.id = scheduler-a
service.name = airflow
telemetry.sdk.language = python
telemetry.sdk.name = opentelemetry
telemetry.sdk.version = 1.44.0
```

**It separates the processes.** Two `SafeOtelLogger` instances built the way `get_otel_logger()` builds one, differing only in `service.instance.id`, each reporting a different `open_slots` for the same pool:

<details><summary>Script and output</summary>

```python
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import InMemoryMetricReader
from opentelemetry.sdk.resources import SERVICE_NAME, Resource

from airflow_shared.observability.metrics.otel_logger import SafeOtelLogger


def build_scheduler(instance_id):
    resource = Resource.create(
        attributes={SERVICE_NAME: "airflow-scheduler", "service.instance.id": instance_id}
    )
    reader = InMemoryMetricReader()
    return SafeOtelLogger(otel_provider=MeterProvider(metric_readers=[reader], resource=resource)), reader


tags = {"pool_name": "default_pool"}
schedulers = [build_scheduler("scheduler-a"), build_scheduler("scheduler-b")]

for (scheduler, _), open_slots in zip(schedulers, (128, 126)):
    scheduler.gauge("pool.open_slots", open_slots, tags=tags)

for _, reader in schedulers:
    rm = reader.get_metrics_data().resource_metrics[0]
    instance = rm.resource.attributes["service.instance.id"]
    for metric in rm.scope_metrics[0].metrics:
        point = next(iter(metric.data.data_points))
        print(f"{metric.name}{{service.instance.id={instance}, {tags}}} = {point.value}")
```

```
airflow.pool.open_slots{service.instance.id=scheduler-a, {'pool_name': 'default_pool'}} = 128
airflow.pool.open_slots{service.instance.id=scheduler-b, {'pool_name': 'default_pool'}} = 126
```

Both samples survive as separate series, so `min by (pool_name)` returns 126 rather than an arbitrary one of the two. Without the attribute the second export overwrites the first.

</details>

**The page itself.** `prek run --files airflow-core/docs/.../metrics.rst` passes, including the RST and codespell hooks, and parsing with docutils reports no structural problems — that caught a short heading underline before this push. The linked spec returns `200`. I could not run the full `breeze build-docs` render locally, since it wanted an interactive CI image rebuild; CI covers it.

---


Miretpl (changes requested):
Could you update your PR description to follow our project guidelines - https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst?

1fanwang:
> Could you update your PR description to follow our project guidelines - https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst?

Thanks for the pointer @Miretpl , updated!

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+    min by (pool_name) (airflow_pool_open_slots)
+
+Backends implementing the OpenTelemetry `Prometheus compatibility
+<https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/>`_ spec expose
+``service.instance.id`` as the ``instance`` label.
+
+.. note::
+
+    This applies to OpenTelemetry only. The StatsD protocol has no resource concept, so metrics
+    from replicated components cannot be distinguished this way.
```
Miretpl: So we could just move this section under `OpenTelemetry` and remove that note, or add a note to `StatsD` that if there is a need for separating information regarding pods, the user has to use OpenTelemetry.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+
+Gauges are where this shows up in the data. Each scheduler samples the metadata database on its
+own loop and exports a value for the same series, so ``pool.open_slots`` and its siblings keep
+whichever export arrived last and appear to flap between replicas' samples.
+
+Give each replica a unique ``service.instance.id`` so its samples form their own series:
+
+.. code-block:: bash
+
+    # on each scheduler, triggerer or Dag processor
```
Miretpl:
```suggestion
    # On every Airflow component which has more than one instance, like scheduler, triggerer, etc.
```

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
     `SDK environment variable documentation <https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#periodic-exporting-metricreader>`_ for more information.
 
 
+Replicated components
+---------------------
+
+Every Airflow process reports the same ``service.name``, so a deployment running more than one
+replica of a component — schedulers in high availability, several triggerers or Dag processors —
```
Miretpl:
```suggestion
replica of a component — Schedulers in high availability, several Triggerers or Dag Processors —
```

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+sends telemetry the backend cannot attribute to an individual process.
+
+Gauges are where this shows up in the data. Each scheduler samples the metadata database on its
+own loop and exports a value for the same series, so ``pool.open_slots`` and its siblings keep
```
Miretpl:
```suggestion
own loop and exports a value for the same series, so e.g. ``pool.open_slots`` and its siblings keep
```

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+configuring. Replicas can then be aggregated at query time instead of overwriting one another —
+for example, the lowest number of open slots any scheduler observed:
+
+.. code-block:: text
+
+    min by (pool_name) (airflow_pool_open_slots)
+
+Backends implementing the OpenTelemetry `Prometheus compatibility
+<https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/>`_ spec expose
+``service.instance.id`` as the ``instance`` label.
```
Miretpl: Not sure if we should have it here - it is not the only compatibility backend with OpenTelemetry, and I'm not sure if we should promote one over another by mentioning it in our docs.

Miretpl (changes requested):
@xBis7 if you have a moment, I would appreciate your take on this PR too

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
     `SDK environment variable documentation <https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#periodic-exporting-metricreader>`_ for more information.
 
 
+Replicated components
+---------------------
+
+Every Airflow process reports the same ``service.name``, so a deployment running more than one
```
xBis7: `service.name` and `service.instance.id` aren't the same. So for example, if I have this in my docker-compose file

```yaml
  airflow-scheduler:
    <<: *airflow-common
    command: scheduler
    environment:
      <<: *airflow-common-env
      OTEL_SERVICE_NAME: airflow-scheduler
      OTEL_RESOURCE_ATTRIBUTES: service.instance.id=airflow-scheduler-1
    ...

  airflow-scheduler-2:
    <<: *airflow-common
    command: scheduler
    environment:
      <<: *airflow-common-env
      OTEL_SERVICE_NAME: airflow-scheduler
      OTEL_RESOURCE_ATTRIBUTES: service.instance.id=airflow-scheduler-2
    ...
```

Then in prometheus I would get

```
airflow_scheduler_critical_section_duration{
  exported_instance="airflow-scheduler-1", exported_job="airflow-scheduler",
  instance="otel-collector:8889", job="otel-collector",
  otel_scope_name="airflow.metrics.otel_logger"
}
```

and

```
airflow_scheduler_critical_section_duration{
  exported_instance="airflow-scheduler-2", exported_job="airflow-scheduler",
  instance="otel-collector:8889", job="otel-collector",
  otel_scope_name="airflow.metrics.otel_logger"
}
```

I think it's worth distinguishing and mentioning both.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
     `SDK environment variable documentation <https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#periodic-exporting-metricreader>`_ for more information.
 
 
+Replicated components
```
xBis7: This is confusing. There isn't actual replication because we don't have a leader-follower relationship between the schedulers. They are both independent.

It doesn't necessarily have to do with scheduler HA, because there is value in distinguishing between different components such as worker and scheduler as well.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+
+Every Airflow process reports the same ``service.name``, so a deployment running more than one
+replica of a component — schedulers in high availability, several triggerers or Dag processors —
+sends telemetry the backend cannot attribute to an individual process.
+
+Gauges are where this shows up in the data. Each scheduler samples the metadata database on its
+own loop and exports a value for the same series, so ``pool.open_slots`` and its siblings keep
+whichever export arrived last and appear to flap between replicas' samples.
+
+Give each replica a unique ``service.instance.id`` so its samples form their own series:
```
xBis7: You use the word `replica` and then refer to a scheduler, triggerer, dag processor, which is inaccurate.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +98,42 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
 
 
+Replicated components
+---------------------
+
+Every Airflow process reports the same ``service.name``, so a deployment running more than one
+replica of a component — schedulers in high availability, several triggerers or Dag processors —
+sends telemetry the backend cannot attribute to an individual process.
+
+Gauges are where this shows up in the data. Each scheduler samples the metadata database on its
```
xBis7: > Gauges are where this shows up in the data.

We should be more generic and describe what will happen if they don't use distinct service ids.

We should say that the OpenTelemetry service name defaults to `Airflow` if unset and explain when it's good to set it to something else. Also explain why and when to set the `service.instance.id`.

xBis7 (changes requested):
@1fanwang Thanks for the PR. This is a valid issue that could occur with multiple schedulers and this PR fills the documentation gap.

We should explain the differences between  `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES: service.instance.id=` and when to use each and why.

The text should be more generic rather than focus on the discussed issue.

When the users read the section, they should get the idea that when configuring OTel, they might also need to set a distinct service name and id but they should also understand when and why.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Done in https://github.com/apache/airflow/pull/71564/commits/f6214bbc380e01ccdb4923971090715c80215aaa

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: done in https://github.com/apache/airflow/pull/71564/commits/f6214bbc380e01ccdb4923971090715c80215aaa

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: outdated removed in a rewrite https://github.com/apache/airflow/pull/71564/commits/f6214bbc380e01ccdb4923971090715c80215aaa based on https://github.com/apache/airflow/pull/71564#discussion_r3796351150

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: done

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Reframed in https://github.com/apache/airflow/commit/f6214bbc380e01ccdb4923971090715c80215aaa. It now leads with the fact that how the attributes surface depends on the backend, and cites the OpenTelemetry Prometheus-compatibility spec rather than any specific backend

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Thanks! agreed, 

it's `service.name` from `OTEL_SERVICE_NAME` (falling back to the `otel_service` option, then `airflow`), `service.instance.id` from `OTEL_RESOURCE_ATTRIBUTES`

done in https://github.com/apache/airflow/commit/f6214bbc380e01ccdb4923971090715c80215aaa

1fanwang:
> @1fanwang Thanks for the PR. This is a valid issue that could occur with multiple schedulers and this PR fills the documentation gap.
> 
> We should explain the differences between `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES: service.instance.id=` and when to use each and why.
> 
> The text should be more generic rather than focus on the discussed issue.
> 
> When the users read the section, they should get the idea that when configuring OTel, they might also need to set a distinct service name and id but they should also understand when and why.

Good point, rewritten along those lines in https://github.com/apache/airflow/commit/f6214bbc380e01ccdb4923971090715c80215aaa

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: agreed this is confusing, fixed in https://github.com/apache/airflow/pull/71564/changes/4ef1a18d8ba9003c77954541de7f774fb89ee417

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: fixed in https://github.com/apache/airflow/pull/71564/changes/4ef1a18d8ba9003c77954541de7f774fb89ee417

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Restructured as a definition list so each attribute maps to what it identifies https://github.com/apache/airflow/pull/71564/changes/4ef1a18d8ba9003c77954541de7f774fb89ee417

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -53,6 +53,12 @@ custom StatsD client. This module must be available on your :envvar:`PYTHONPATH`
 
 See :doc:`../modules_management` for details on how Python and Airflow manage modules.
 
+.. note::
+
+    StatsD has no resource concept, so metrics cannot be attributed to the process that
+    produced them. Use OpenTelemetry if you need that — see
```
xBis7: Should we also say that in scheduler HA it's recommended to use OTel otherwise, you end up with the values from the scheduler that exports last?

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +104,51 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+decide how much of a deployment can be told apart:
+
+``service.name``
+    Which component reported the metric. It defaults to ``airflow`` for every Airflow process, so
+    a scheduler, a triggerer and a worker arrive under one name. Set it per component to attribute
+    a metric to the kind of process that produced it.
+
+``service.instance.id``
+    Which process of that component reported the metric. It is unset by default, so processes
+    running the same component are indistinguishable. Set it per process to attribute a metric to
```
xBis7:
```suggestion
    running the same component (e.g. 2+ schedulers) are indistinguishable. Set it per process to attribute a metric to
```

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
```diff
@@ -98,6 +104,51 @@ Add the Collector details to your configuration file e.g. ``airflow.cfg``
...
+``OTEL_RESOURCE_ATTRIBUTES``:
+
+.. code-block:: bash
+
+    # on one of the schedulers
+    export OTEL_SERVICE_NAME="airflow-scheduler"
+    export OTEL_RESOURCE_ATTRIBUTES="service.instance.id=$(hostname)"
+
+Processes that share a resource also share a series, and the backend keeps whichever export
+arrived last. Where several processes run the same component this loses data rather than
```
xBis7: > Where several processes run the same component this loses data rather 

This is confusing. Maybe you meant `when`.

```suggestion
arrived last. When several processes run the same component, data are lost instead of aggregated.
```

xBis7 (commented):
@1fanwang This looks much better. I added some small comments.

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Makes sense! Clarified in https://github.com/apache/airflow/pull/71564/commits/b32b9e4d55be9162c2e9445238c439116a7a7ee3

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: Done in https://github.com/apache/airflow/pull/71564/commits/b32b9e4d55be9162c2e9445238c439116a7a7ee3

airflow-core/docs/administration-and-deployment/logging-monitoring/metrics.rst
1fanwang: The suggested wording does look a LOT better then before, it was indeed confusing.
Done in https://github.com/apache/airflow/pull/71564/commits/b32b9e4d55be9162c2e9445238c439116a7a7ee3 and https://github.com/apache/airflow/pull/71564/commits/3a2e3dda8a269e3238c397c8c2cfb588099a3443

xBis7 (approved):
LGTM!

uros-b (approved):

uros-b:
+1, LGTM!

Miretpl (approved):
Looks nice! Thanks everyone for the review!

Miretpl:
@1fanwang, it is in the history of the PR description, but the latest edit to it made it not follow our guidelines. Could you fix that?

1fanwang:
> @1fanwang, it is in the history of the PR description, but the latest edit to it made it not follow our guidelines. Could you fix that?

Thanks! Fixed.

PR #71565, opened 2026-08-13
Add support for sending emails through Microsoft Graph

### LINKED ISSUE #71541: Add support for sending emails on Outlook 365 via Microsoft Graph
### Description

Airflow provides built-in functionality to send email notifications on task state changes via SMTP. This works with on-premises mail servers and SMTP relays to cloud providers, however the recommended way to send emails in Outlook 365 programmatically is now via Microsoft Graph. With the deprecation of EWS this year it is possible that Microsoft stops supporting the SMTP relay route in the future as well

### Use case/motivation

Add support for sending emails using Office365 / Outlook cloud accounts via Microsoft Graph to the Azure provider.

### Related issues

_No response_

### Code of Conduct

rjgoyln:
## Summary

Sending mail from an Office 365 or Outlook cloud mailbox is expected to go through Microsoft Graph, and with EWS gone the SMTP relay into those mailboxes is the last legacy path standing. Airflow could only reach them over SMTP, so a deployment on a Microsoft tenant had no supported way to alert on task failures from its own mailbox.

## Change

The Azure provider sends mail through the Graph `sendMail` endpoint over an existing `msgraph` connection, covering both places email is configured: `[email] email_backend` for `email_on_failure` and `email_on_retry`, and `MSGraphNotifier` for the state-change callbacks. The notifier implements `async_notify`, so it also runs in the triggerer.

`from_email` selects the mailbox to send from, and Graph addresses that mailbox in the request path — hence the `parseaddr`, since the option is conventionally configured as `Display name <user@example.com>`, the form SMTP and SES accept. Attachments ride inline in the request body rather than through an upload session, which is what bounds them at 3 MB.

The email backend contract is synchronous while the Graph hook is async-only. The sync entry point therefore runs the coroutine on its own loop and closes the request adapter before that loop goes away; the cached adapter would otherwise outlive `asyncio.run` bound to a closed loop and break the next email.

closes: #71541

---

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -531,6 +531,7 @@ Drillbit
 driverless
 dropdown
 druidHook
+dryrun
 ds
 dsl
 dsn
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -531,7 +531,6 @@ Drillbit
 driverless
 dropdown
 druidHook
-dryrun
 ds
 dsl
 dsn
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
@@ -44,7 +44,12 @@ def send_email(
     custom_headers: dict[str, Any] | None = None,
     **kwargs,
 ) -> None:
-    """Email backend for Microsoft Graph."""
+    """
+    Email backend for Microsoft Graph.
+
+    .. note::
+        For more information, see :ref:`email-configuration-msgraph`
+    """
     if not from_email:
         raise ValueError(
             "The `from_email` configuration has to be set for the Microsoft Graph emailer, as it "
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
@@ -57,7 +57,7 @@ def send_email(
         )
 
     if dryrun:
-        log.info("Dryrun, not sending email with subject %r to %s", subject, to)
+        log.info("Dry run, not sending email with subject %r to %s", subject, to)
         return
 
     # ``mime_subtype`` and ``mime_charset`` are part of the email backend contract but have no
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
@@ -14,7 +14,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-"""Airflow module for email backend using Microsoft Graph."""
+"""
+Airflow module for email backend using Microsoft Graph.
+
+.. spelling:word-list::
+
+    dryrun
+"""
 
 from __future__ import annotations
 
--- a/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
+++ b/providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
@@ -1,0 +1,70 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Airflow module for email backend using Microsoft Graph."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+from airflow.providers.microsoft.azure.hooks.msgraph import MSGraphMailHook
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+log = logging.getLogger(__name__)
+
+
+def send_email(
+    to: str | Iterable[str],
+    subject: str,
+    html_content: str,
+    files: list[str] | None = None,
+    dryrun: bool = False,
+    cc: str | Iterable[str] | None = None,
+    bcc: str | Iterable[str] | None = None,
+    mime_subtype: str = "mixed",
+    mime_charset: str = "utf-8",
+    conn_id: str | None = None,
+    from_email: str | None = None,
+    custom_headers: dict[str, Any] | None = None,
+    **kwargs,
+) -> None:
+    """Email backend for Microsoft Graph."""
+    if not from_email:
+        raise ValueError(
+            "The `from_email` configuration has to be set for the Microsoft Graph emailer, as it "
+            "determines which mailbox the message is sent from."
+        )
+
+    if dryrun:
+        log.info("Dryrun, not sending email with subject %r to %s", subject, to)
+        return
+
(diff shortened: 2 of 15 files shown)


airflow-core/docs/howto/email-config.rst
eladkal, line 1: This will need to be adjusted after we merge 
https://github.com/apache/airflow/pull/71571

rjgoyln: OK. I got it. Thank for the review.

eladkal: Pr is merged. You can now build on top of it

rjgoyln: Ok. I have updated the files. Thank you!

providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
```diff
@@ -0,0 +1,75 @@
...
+        For more information, see :ref:`email-configuration-msgraph`
+    """
+    if not from_email:
+        raise ValueError(
+            "The `from_email` configuration has to be set for the Microsoft Graph emailer, as it "
+            "determines which mailbox the message is sent from."
+        )
+
+    if dryrun:
+        log.info("Dryrun, not sending email with subject %r to %s", subject, to)
```
eladkal:
```suggestion
        log.info("Dry run, not sending email with subject %r to %s", subject, to)
```

docs/spelling_wordlist.txt
```diff
@@ -531,6 +531,7 @@ Drillbit
 driverless
 dropdown
 druidHook
+dryrun
```
eladkal: can be removed after earlier comment is addressed

rjgoyln: Removed `dryrun` from `docs/spelling_wordlist.txt`, restoring the file to match `main`.

One note: the spelling error comes from AutoAPI rendering the required `dryrun` parameter in the generated docs, not from the log message itself. Since `send_email` passes `dryrun=` as part of the backend contract, the parameter name cannot be changed.

So I scoped the spelling exception to the module instead, consistent with other modules in this provider:

```python
"""
Airflow module for email backend using Microsoft Graph.

.. spelling:word-list::

    dryrun
"""
```

providers/microsoft/azure/src/airflow/providers/microsoft/azure/emailer.py
eladkal, line 1: This file is not needed.
send_email should just be a function in the hook

rjgoyln: Done — `send_email` now lives in `hooks/msgraph.py`, and `emailer.py` has been removed.

dabla:
I will try to review this PR tomorrow as I may have some remarks, but I like this, very nice and needed addition!

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -789,3 +807,264 @@ def error_mapping() -> dict[str, type[ParsableFactory]]:
...
+        For more information, see :ref:`email-configuration-msgraph`
+    """
+    if not from_email:
+        raise ValueError(
+            "The `from_email` configuration has to be set for the Microsoft Graph emailer, as it "
+            "determines which mailbox the message is sent from."
+        )
+
+    if dryrun:
+        log.info("Dry run, not sending email with subject %r to %s", subject, to)
```
dabla: Maybe the dry run could call the `build_message` method, so at least it does something except sending the actual mail?

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
rjgoyln: Thanks for the suggestion! I initially changed the dry run to call `build_message()` before returning. While reviewing this, I noticed that `extract_sender()` was only performed on the actual send path, so the dry run could still behave differently from a real send.

I therefore moved the `dryrun` handling into the hook, so both paths perform the same preparation (`extract_sender()` and `build_message()`), and only the actual Graph request is skipped during a dry run.

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
```diff
@@ -789,3 +804,276 @@ def error_mapping() -> dict[str, type[ParsableFactory]]:
...
+
+        if not request_adapter:
+            return
+
+        adapter = cast("HttpxRequestAdapter", request_adapter)
+        await adapter._http_client.aclose()
+        provider = cast("BaseBearerTokenAuthenticationProvider", adapter._authentication_provider)
+        access_token_provider = cast("AzureIdentityAccessTokenProvider", provider.access_token_provider)
+        credential = cast("CachedAsyncTokenCredential", access_token_provider._credentials)
+        await credential._credential.close()
```
dabla: This method should be renamed as close and should move to the parent hook `KiotaRequestAdapterHook`. Also we know the `KiotaRequestAdapterHook` uses the `CachedAsyncTokenCredential` which is a decorator holding a reference to an `AsyncTokenCredential` instance (e.g. `CertificateCredential` or `ClientSecretCredential`) and call the close there like you already did.

So we could write the async `close` method like this in `KiotaRequestAdapterHook`:

```
    async def close(self) -> None:
        """Close the request adapter cached for this connection and evict it from the cache."""
        _, request_adapter = self.cached_request_adapters.pop(self.conn_id, (None, None))

        if not request_adapter:
            return

        try:
            adapter = cast("HttpxRequestAdapter", request_adapter)
            await adapter._http_client.aclose()
        finally:
            provider = cast("BaseBearerTokenAuthenticationProvider", adapter._authentication_provider)
            access_token_provider = cast("AzureIdentityAccessTokenProvider", provider.access_token_provider)
            credential = cast("AsyncTokenCredential", access_token_provider._credentials)
            await credential._credential.close()
```

Then in the `send_request` method of `KiotaRequestAdapterHook`, we could write it like this:

```
    async def send_request(self, request_info: RequestInformation, response_type: str | None = None):
        conn = await self.get_async_conn()

        try:
            self.log.info("Executing url '%s' as '%s'", request_info.url, request_info.http_method)

            if response_type:
                return await conn.send_primitive_async(
                    request_info=request_info,
                    response_type=response_type,
                    error_map=self.error_mapping(),
                )
            return await conn.send_no_response_content_async(
                request_info=request_info,
                error_map=self.error_mapping(),
            )
        except (PermissionError, RuntimeError, ValueError) as e:
            self.log.warning(
                "Request failed for conn_id '%s': %s. Invalidating cached request adapter.",
                self.conn_id,
                e,
            )
            await self.close()
            raise
```

dabla (changes requested):

providers/microsoft/azure/src/airflow/providers/microsoft/azure/hooks/msgraph.py
dabla: I've already implemented that async close method in `KiotaRequestAdapterHook` to fix an issue there in another [PR](https://github.com/apache/airflow/pull/72688), so it would only become a matter of calling that method once that [PR](https://github.com/apache/airflow/pull/72688) is merged instead of calling the `close_async_conn` on the `MSGraphMailHook` which can be removed.

rjgoyln: Thanks, I understand. I'll wait for #72688 to land and then remove `close_async_conn()` from `MSGraphMailHook` and use the inherited `close()` instead.

dabla: It's been merged.

rjgoyln: Thanks! I removed `close_async_conn()`, and updated `send_and_close()` to use the inherited `close()`.

The inherited implementation also uses `try/finally`, so the credential is still closed if closing the HTTP client fails.

dabla (approved):

dabla:
Congratulations @rjgoyln, nice work on this PR as this was a most needed feature especially with notifications.

rjgoyln:
Thanks for the review. Really appreciate it!

PR #71567, opened 2026-08-13
[KubernetesPodOperator] Add on_deferrable_finish_log_tail_lines param

raghavan-arvind:
## AI Usage
Model: Fable 5

Used AI to assist in writing code + PR summary, but hand verified the results and the PR summary is correct. Can provide any more detail as needed.

## Why

Supporting an explicit max tail lines will help improve UI performance for high-volume jobs. We do want _some logs_, on failure we print the error near the end, so that is helpful. For the rest we have a separate log-ingestion solution.

More context (below is AI-generated, but accurate to my knowledge):

In deferrable mode, when the trigger fires its terminal event, `trigger_reentry` calls `_write_logs`, which re-emits the **entire pod log** into the task log in one bulk fetch. For chatty workloads this is problematic:

- Most production setups already ship pod stdout to a log aggregator directly from the node; the bulk replay double-ingests every line (and stores a second copy in remote task logs).
- The "full log" is illusory anyway: the kubelet rotates container logs (`containerLogMaxSize`, default 10Mi), and the pod logs API only serves the current file — so long-running chatty pods already get a silent, size-based, node-config-dependent tail today.

## What

Adds `on_deferrable_finish_log_tail_lines: int | None = None` to `KubernetesPodOperator`:

- `None` (default) preserves existing behavior exactly.
- When set, `_write_logs` passes `tail_lines` through to `read_namespaced_pod_log`, so only the last N lines are written to the task log in the single bulk fetch after the pod finishes.
- Deliberately scoped to that one fetch: it does **not** affect live log streaming (non-deferrable mode) or the periodic `logging_interval` relay, where tailing each window would silently drop lines. `_write_logs` takes `tail_lines` as a plain parameter, so call sites stay explicit.
- When `logging_interval` is set, the finish-time fetch already passes `since_time`; the Kubernetes API applies both (time window first, tail within it), which keeps the semantics consistent.

Happy to adjust the parameter name if reviewers prefer something shorter.

## Testing

Verified in a production-like staging deployment (Airflow 3.3.0, provider vendored with this patch): a deferrable pod task with `on_deferrable_finish_log_tail_lines=5` wrote exactly the last 5 lines of the container log after resume, with defer/resume/XCom/cleanup behavior unchanged.

### DIFF
--- a/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
+++ b/providers/cncf/kubernetes/src/airflow/providers/cncf/kubernetes/operators/pod.py
@@ -272,6 +272,10 @@ class KubernetesPodOperator(BaseOperator):
     :param logging_interval: max time in seconds that task should be in deferred state before
         resuming to fetch the latest logs. If ``None``, then the task will remain in deferred state until pod
         is done, and no logs will be visible until that time.
+    :param on_deferrable_finish_log_tail_lines: only write the last N lines of the container log
+        to the task log in the single bulk log fetch after the pod finishes (deferrable mode).
+        Does not affect log streaming (non-deferrable mode) or the periodic log relay
+        (``logging_interval``). If ``None`` (default), the full container log is written.
     :param trigger_kwargs: additional keyword parameters passed to the trigger
     :param container_name_log_prefix_enabled: if True, will prefix container name to each log line.
         Default to True.
@@ -394,6 +398,7 @@ class KubernetesPodOperator(BaseOperator):
         ) = None,
         progress_callback: Callable[[str], None] | None = None,
         logging_interval: int | None = None,
+        on_deferrable_finish_log_tail_lines: int | None = None,
         trigger_kwargs: dict | None = None,
         container_name_log_prefix_enabled: bool = True,
         log_formatter: Callable[[str, str], str] | None = None,
@@ -503,6 +508,7 @@ class KubernetesPodOperator(BaseOperator):
         self.termination_message_policy = termination_message_policy
         self.active_deadline_seconds = active_deadline_seconds
         self.logging_interval = logging_interval
+        self.on_deferrable_finish_log_tail_lines = on_deferrable_finish_log_tail_lines
         self.trigger_kwargs = trigger_kwargs
 
         self._config_dict: dict | None = None  # TODO: remove it when removing convert_config_file_to_dict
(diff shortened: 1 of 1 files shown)


potiuk (changes requested):
The change is small, correctly scoped to the single post-completion bulk fetch in `trigger_reentry`, and `None` is a true no-op (the kubernetes client drops `None` query params) — but it adds a user-facing parameter with no test at all, so it cannot go in as it stands.

### Missing test for the new behaviour

> Flag any changed or added behaviour without a corresponding test, and flag tests that a reviewer cannot fail by reverting the PR's change. The target is exactly 100% coverage of what the PR changes — no more, no less.

```python
        logs = self.client.read_namespaced_pod_log(
            ...
            since_seconds=since_seconds,
            tail_lines=tail_lines,          # ← new kwarg, nothing asserts it
            _preload_content=False,
        )
```

The PR touches only `pod.py`. Please add, in `providers/cncf/kubernetes/tests/unit/cncf/kubernetes/operators/test_pod.py` next to `test_write_logs_with_valid_since_time` (which already inspects `read_namespaced_pod_log.call_args`):

1. a unit test that `k._write_logs(pod, tail_lines=5)` produces `call_kwargs["tail_lines"] == 5`, and that the default produces `None`;
2. a `trigger_reentry`-level test (e.g. parametrize `test_async_write_logs_should_execute_successfully` on `on_deferrable_finish_log_tail_lines`) asserting the operator forwards its own attribute to `read_namespaced_pod_log`.

Both fail with the change reverted, which is exactly the coverage we want here.

### The option is a silent no-op on `KubernetesJobOperator`

`KubernetesJobOperator` subclasses `KubernetesPodOperator` (`job.py:63`), so it inherits the new argument and its docstring — but its finish path calls `self._write_logs(pods_by_name[pod_name])` (`job.py:345`) without `tail_lines`. A user who sets the option on the Job operator gets the full log and no warning. Either forward `tail_lines=self.on_deferrable_finish_log_tail_lines` there too (with a `test_job.py` case), or state in the docstring that the option applies only to `KubernetesPodOperator.trigger_reentry`.

### Smaller observations

- PR quality rule 2 in `contributing-docs/05_pull_requests.rst`: "never use ... any prefixes in the title". Please drop the `[KubernetesPodOperator]` prefix from the PR title and the commit headline, e.g. `Add on_deferrable_finish_log_tail_lines to KubernetesPodOperator` — the provider changelog is generated from it.
- The Gen-AI disclosure is present in substance (tool and model named, hand-verified), but the PR template's `- [ ] Yes (please specify the tool below)` checkbox and `Generated-by:` line were removed. Restoring that section keeps the PR parseable by the project's tooling.
- Naming (you invited it): `deferrable_log_tail_lines` would sit more naturally next to `logging_interval` / `trigger_kwargs`; the "final bulk fetch only" nuance is already in the docstring.
- When `logging_interval` is set, the terminal fetch passes both `since_seconds` and `tail_lines`; the kubelet returns the intersection of the two windows, which matches your description. The docstring's "does not affect the periodic log relay" is true for the re-defer windows, but it may be worth stating that the final windowed fetch is still tailed.
- Airflow CI has not run on this PR at all yet (only the `Mergeable` / `WIP` apps have reported), so nothing here is verified by CI. A maintainer will approve the workflow run once the above is addressed; in the meantime `prek run --from-ref main` locally is worth a pass.

---

>

PR #71570, opened 2026-08-13
Fail Glue job tasks promptly when a verbose job run errors

rjgoyln:
## Summary

A Glue job in `ERROR` is already finished and cannot reach `SUCCEEDED`, but the verbose polling loop continues until it exhausts its attempts. With the default settings, this means waiting up to 75 minutes for `GlueJobOperator` or two hours for `GlueJobSensor` before reporting `waiter exceeded max attempts`, which looks like a timeout instead of the actual job failure.

The `job_complete` waiter used by the non-verbose path already treats `ERROR` as a failure, so this change makes the two deferrable paths consistent.

## Scope

Stacked on #71495, whose commit is included here until it merges.

`EXPIRED` is intentionally out of scope. It is terminal for `GlueJobOperator`, but the waiter does not currently handle it either. Handling it here would make the verbose path behave differently from the non-verbose path.

Raised during review of #71495.

---


henry3260 (approved):
Thanks!

PR #71571, opened 2026-08-13
Move SendGrid and SES email backend setup docs to their providers

eladkal:
Split the SendGrid and Amazon SES email backend setup instructions out
of airflow-core/docs/howto/email-config.rst into dedicated pages under
each provider's docs (providers/sendgrid/docs/email-backend.rst and
providers/amazon/docs/email-backend.rst).

The core doc now only covers the general email configuration mechanism
(SMTP defaults, templates, from_email, the email_backend/email_conn_id
pattern) and links out to the provider pages for setup. Each provider
page links back to the core doc for the shared concepts, and the SES
emailer docstring gets the same :ref: cross-link SendGrid's already had.

This avoids duplicating provider setup steps in core docs and keeps
each provider's docs as the source of truth for its own setup.

---

- [X] Yes — Claude (Sonnet 5)

Generated-by: Claude (Sonnet 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/airflow-core/docs/howto/email-config.rst
+++ b/airflow-core/docs/howto/email-config.rst
@@ -93,7 +93,5 @@ community provider to send email through a third-party service's API. To use one
 provider distribution, set ``email_backend`` to the dotted path of its ``send_email`` function, and set
 ``email_conn_id`` to a connection holding the credentials it needs.
 
-Community providers currently offering an email backend:
-
-* **SendGrid** — see :doc:`apache-airflow-providers-sendgrid:email-backend` for setup instructions.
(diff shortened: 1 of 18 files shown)


vincbeck (approved):
A follow-up PR could be creating a page that automatically discover email backends from the providers, the same way we do with auth managers https://airflow.apache.org/docs/apache-airflow-providers/core-extensions/auth-managers.html. But this is already a nice improvement so non-blocking.

eladkal:
@vincbeck done

vincbeck (approved):
Love it!

PR #71574, opened 2026-08-13
Bound and harden the Glue job-UUID scan fallback

### LINKED ISSUE #71489: Bound and harden the task-UUID scan fallback

bramhanandlingala:
Closes #71489

## Summary
Bounds and hardens the Glue task-UUID scan fallback used on retry.

## Root Cause
The scan looped over `GetJobRuns` with no page cap. A no-match run —
the common retry case — walked the job's entire run history before
giving up. Failures were also swallowed by a broad `except Exception`
and logged only at `warning`, so a fallback failure gave no clear
signal before the operator submitted a fresh run into a still-active
one.

## Fix
- Added a page cap (`TASK_UUID_SCAN_MAX_PAGES`) and an early exit once
  a run's `StartedOn` predates the DAG run (runs come back newest-first).
- Narrowed `except Exception` to `except ClientError`.
- Raised the fallthrough log level to `error` with the actual reason.
- Documented the `glue:GetJobRuns` IAM requiremen

- [X] Yes - Claude

### DIFF
--- a/providers/amazon/docs/operators/glue.rst
+++ b/providers/amazon/docs/operators/glue.rst
@@ -198,7 +198,7 @@ existed unless ``resume_glue_job_on_retry`` is set. On Airflow 3.3+, ``durable``
 .. note::
   The task-UUID scan fallback (used on retry whenever XCom has no cached run id -- always the case
   on Airflow 3.0-3.2, and on Airflow 2.x once XCom itself has expired or been cleared) calls
- ``glue:GetJobRuns`` against the job's run history. Grant this action, in addition to
+  ``glue:GetJobRuns`` against the job's run history. Grant this action, in addition to
   ``glue:StartJobRun`` and ``glue:GetJobRun``, to any IAM policy relying on ``durable=True`` (or
   ``resume_glue_job_on_retry=True`` below Airflow 3.3) for crash recovery -- without it, the scan
   fails, is logged as an error, and the operator submits a fresh run instead of reconnecting.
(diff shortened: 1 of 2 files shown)


o-nikolas:
https://github.com/apache/airflow/pull/71546 already exists

PR #71575, opened 2026-08-13
Do not replay cached AI steps that cannot be verified

YAshhh29:
When durable execution replays a cached model response or tool result, it compares a fingerprint of the current request against the fingerprint stored with the cache entry. Requests that cannot be serialized to JSON fingerprint as `None`, and `None == None` is `True` — so the guard passed for precisely the requests it could not verify, and a stale entry was replayed.

```diff
-            if cached_fingerprint == fingerprint:
+            if fingerprint is not None and cached_fingerprint == fingerprint:
```

**This is a behaviour change, not only a bug fix.** A step whose request cannot be fingerprinted now re-runs live on retry rather than replaying, which costs a model call. The existing code was ambiguous about which behaviour was intended: `fingerprint.py` logged *"cached responses for this step replay without verification"* (fail open), while the `reason=` branch immediately below the comparison explains *"entry predates fingerprinting or the request could not be fingerprinted"* — a message that could never be reached, because that case short-circuited into the replay path. This PR resolves the contradiction in favour of not replaying what cannot be verified, and documents the contract in `agent.rst`.

### Behaviour before and after

The screenshots below drive the real `CachingModel` and `CachingToolset` directly (no pytest, no mocked guard). Each run prints the guard line straight out of `caching_model.py`, so the version under test is visible in the output. The scenario is identical in both: a cached answer for Q3 revenue is stored with a `None` fingerprint, then the step is retried with a request about Q4 revenue that also cannot be fingerprinted.

**Before — guard as it is on `main`**

image

The cached Q3 answer is returned in response to the Q4 question, with `live calls: 0` and `replayed steps: 1`. The model is never consulted, and nothing verified that the two requests matched.

**After — guard with this PR**

image

The unverifiable step runs live: `live calls: 1`, `replayed steps: 0`, and the answer matches the question that was actually asked.

### Testing

`test_unverifiable_current_request_treated_as_miss` and `test_unverifiable_current_call_treated_as_miss` fail without the change and pass with it.

| `providers/common/ai/tests/unit/common/ai/durable` | result |
| --- | --- |
| guard as it is on `main` | `2 failed, 85 passed` |
| guard with this PR | `87 passed` |

---

- [X] Yes — GitHub Copilot (Claude Opus 5)

Generated-by: GitHub Copilot (Claude Opus 5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/common/ai/docs/operators/agent.rst
+++ b/providers/common/ai/docs/operators/agent.rst
@@ -300,6 +300,9 @@ cache:
    never replays responses that belong to a different conversation.
 4. After successful completion, the cached steps are deleted.
 
+If a model request or tool call cannot be fingerprinted, that step runs live on
+retry rather than replaying an unverified cache entry.
+
 Replay verification compares the **requests** sent to models and tools, not
 the code behind them. Editing a tool's implementation between attempts does
 not invalidate an already-cached result for an identical call, and pointing
--- a/providers/common/ai/docs/operators/agent.rst
+++ b/providers/common/ai/docs/operators/agent.rst
@@ -300,8 +300,16 @@ cache:
    never replays responses that belong to a different conversation.
 4. After successful completion, the cached steps are deleted.
 
-If a model request or tool call cannot be fingerprinted, that step runs live on
-retry rather than replaying an unverified cache entry.
+If a model request or tool call cannot be fingerprinted -- it carries a value
+that will not serialize to JSON -- that step is not cached, and on retry it runs
+live rather than replaying an unverified entry. This is rarely confined to a
+single step: the usual causes are a non-serializable value in ``model_settings``,
+which is attached to every request, or in the message history, which every later
+request carries forward. Either one degrades all subsequent model steps the same
+way, leaving durable execution with nothing to replay, so the retry re-runs the
+agent at full cost. The ``could not fingerprint model request`` warning in the
+task log marks where this began; making the offending value JSON-serializable
+restores replay.
 
 Replay verification compares the **requests** sent to models and tools, not
 the code behind them. Editing a tool's implementation between attempts does
--- a/providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
@@ -95,7 +95,7 @@ class CachingModel(WrapperModel):
 
         cached, cached_fingerprint = self.storage.load_model_response(key)
         if cached is not None:
-            if cached_fingerprint == fingerprint:
+            if fingerprint is not None and cached_fingerprint == fingerprint:
                 self.counter.replayed_model += 1
                 log.debug("Durable: replayed cached model response", step=step)
                 return cached
--- a/providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
@@ -49,7 +49,9 @@ class CachingModel(WrapperModel):
     returns the cached response without calling the underlying model.
     Otherwise, calls the model and caches the response. A fingerprint
     mismatch means the agent changed between attempts; the stale entry is
-    discarded and the step re-runs live.
+    discarded and the step re-runs live. A request that cannot be
+    fingerprinted is neither replayed nor cached: an entry stored without a
+    fingerprint could never be verified on a later attempt.
     """
 
     storage: DurableStorageProtocol = field(repr=False)
@@ -112,6 +114,11 @@ class CachingModel(WrapperModel):
             )
 
         response = await self.wrapped.request(messages, model_settings, model_request_parameters)
+        if fingerprint is None:
+            # Storing this would write an entry the guard above can never accept,
+            # once per step, each write rewriting the whole cache blob.
+            log.debug("Durable: not caching model response that cannot be verified on replay", step=step)
+            return response
         self.storage.save_model_response(key, response, fingerprint=fingerprint)
         self.counter.cached_model += 1
         log.debug("Durable: cached model response", step=step)
--- a/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
@@ -34,9 +34,8 @@ longer match.
 Fields that pydantic-ai regenerates on every attempt (message-level
 ``timestamp``/``run_id``/``conversation_id`` and part-level ``timestamp``)
 are excluded from the fingerprint.  Requests that cannot be serialized to
-JSON fingerprint as ``None``, which degrades that step to unverified
-positional replay (the pre-fingerprint behavior) rather than disabling
-caching.
+JSON fingerprint as ``None`` and re-run live instead of replaying without
+verification.
 """
 
 from __future__ import annotations
@@ -119,9 +118,8 @@ def fingerprint_model_request(
     output mode and schema, native tools, ...) so any change to what is sent
     to the model invalidates the cached response.
 
-    Returns ``None`` when the request cannot be serialized; ``None`` compares
-    equal to ``None``, so requests that cannot be fingerprinted degrade to
-    unverified positional replay rather than disabling caching.
+    Returns ``None`` when the request cannot be serialized, which prevents a
+    cached response from being replayed for that step.
     """
     try:
         dumped = ModelMessagesTypeAdapter.dump_python(messages, mode="json")
@@ -136,10 +134,7 @@ def fingerprint_model_request(
         )
     except (TypeError, ValueError):
         # TypeError from json.dumps, ValueError covers PydanticSerializationError
-        log.warning(
-            "Durable: could not fingerprint model request; cached responses for this "
-            "step replay without verification"
-        )
+        log.warning("Durable: could not fingerprint model request; this step will execute live on retry")
         return None
 
 
@@ -155,8 +150,7 @@ def fingerprint_tool_call(name: str, tool_args: dict[str, Any], tool_call_id: st
         return _digest({"name": name, "args": tool_args, "tool_call_id": tool_call_id})
     except (TypeError, ValueError):
         log.warning(
-            "Durable: could not fingerprint tool call; cached results for this "
-            "step replay without verification",
+            "Durable: could not fingerprint tool call; this step will execute live on retry",
             tool=name,
         )
         return None
--- a/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
@@ -34,8 +34,11 @@ longer match.
 Fields that pydantic-ai regenerates on every attempt (message-level
 ``timestamp``/``run_id``/``conversation_id`` and part-level ``timestamp``)
 are excluded from the fingerprint.  Requests that cannot be serialized to
-JSON fingerprint as ``None`` and re-run live instead of replaying without
-verification.
+JSON fingerprint as ``None``: the step is neither replayed nor cached, and
+re-runs live instead of replaying without verification.  This is seldom
+confined to one step -- the usual causes (a non-JSON value in model settings,
+or in the message history) are carried into every later request, so durable
+execution stops contributing anything for the rest of the run.
 """
 
 from __future__ import annotations
@@ -62,9 +65,13 @@ _VOLATILE_MESSAGE_KEYS = ("timestamp", "run_id", "conversation_id")
 
 # Settings that control transport, not response content. Excluded from the
 # fingerprint: changing them should not invalidate a cached response, and some
-# (``timeout`` can be an ``httpx.Timeout``) are not JSON-serializable, which
-# would otherwise force the whole fingerprint to ``None`` and silently disable
-# replay verification for every step.
+# (``timeout`` can be an ``httpx.Timeout``) are not JSON-serializable.
+#
+# This frozenset is load-bearing. Model settings ride along with every request,
+# so a single non-JSON member fingerprints every model step as ``None`` -- which
+# now costs durable execution entirely for the run (nothing is cached, nothing
+# is replayed), not merely the verification of a replay. Any non-JSON setting
+# must be listed here or normalized before it reaches the fingerprint.
 _TRANSPORT_ONLY_SETTINGS = frozenset({"timeout"})
 
 
@@ -118,8 +125,10 @@ def fingerprint_model_request(
     output mode and schema, native tools, ...) so any change to what is sent
     to the model invalidates the cached response.
 
-    Returns ``None`` when the request cannot be serialized, which prevents a
-    cached response from being replayed for that step.
+    Returns ``None`` when the request cannot be serialized, which prevents the
+    step from being replayed or cached. Because model settings and message
+    history are carried into every later request, an unserializable value in
+    either usually degrades every subsequent model step of the run the same way.
     """
     try:
         dumped = ModelMessagesTypeAdapter.dump_python(messages, mode="json")
@@ -134,7 +143,11 @@ def fingerprint_model_request(
         )
     except (TypeError, ValueError):
         # TypeError from json.dumps, ValueError covers PydanticSerializationError
-        log.warning("Durable: could not fingerprint model request; this step will execute live on retry")
+        log.warning(
+            "Durable: could not fingerprint model request; this step will not be cached and will "
+            "execute live on retry. If the cause is in model settings or message history, every "
+            "later model step of this run is affected too"
+        )
         return None
 
 
@@ -150,7 +163,8 @@ def fingerprint_tool_call(name: str, tool_args: dict[str, Any], tool_call_id: st
         return _digest({"name": name, "args": tool_args, "tool_call_id": tool_call_id})
     except (TypeError, ValueError):
         log.warning(
-            "Durable: could not fingerprint tool call; this step will execute live on retry",
+            "Durable: could not fingerprint tool call; this step will not be cached and will "
+            "execute live on retry",
             tool=name,
         )
         return None
--- a/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
@@ -127,7 +127,7 @@ def fingerprint_model_request(
 
     Returns ``None`` when the request cannot be serialized, which prevents the
     step from being replayed or cached. Because model settings and message
-    history are carried into every later request, an unserializable value in
+    history are carried into every later request, a non-serializable value in
     either usually degrades every subsequent model step of the run the same way.
     """
     try:
--- a/providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
+++ b/providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
@@ -159,6 +159,25 @@ class TestCachingModelReplayVerification:
         mock_model.request.assert_called_once()
 
     @pytest.mark.asyncio
+    async def test_unverifiable_current_request_treated_as_miss(
+        self, mock_model, mock_storage, counter, sample_response
+    ):
+        stale = ModelResponse(parts=[TextPart(content="stale")])
+        mock_storage.load_model_response.return_value = (stale, None)
+        mock_model.request = AsyncMock(return_value=sample_response)
+        mock_model.prepare_request = lambda settings, params: ({"extra_body": object()}, params)
+        caching = CachingModel(mock_model, storage=mock_storage, counter=counter)
+
+        result = await caching.request([], None, ModelRequestParameters())
+
+        assert result is sample_response
+        mock_model.request.assert_called_once()
+        assert counter.replayed_model == 0
+        mock_storage.save_model_response.assert_called_once_with(
+            f"{P}model_step_0", sample_response, fingerprint=None
+        )
+
+    @pytest.mark.asyncio
     async def test_fingerprint_uses_prepared_request_not_raw_arguments(
         self, mock_storage, counter, sample_response
     ):
--- a/providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
+++ b/providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
@@ -173,9 +173,24 @@ class TestCachingModelReplayVerification:
         assert result is sample_response
         mock_model.request.assert_called_once()
         assert counter.replayed_model == 0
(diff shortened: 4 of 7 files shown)


providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
```diff
@@ -95,7 +95,7 @@ async def request(
 
         cached, cached_fingerprint = self.storage.load_model_response(key)
         if cached is not None:
-            if cached_fingerprint == fingerprint:
+            if fingerprint is not None and cached_fingerprint == fingerprint:
```
kaxil, line 100: An entry stored with `fingerprint=None` can never be replayed now: this line needs `fingerprint is not None`, and a real digest won't equal a stored `None` either. Line 115 still saves with `fingerprint=fingerprint` (same in `caching_toolset.py:92`), so a run that can't fingerprint writes a dead entry per step and reads it back next attempt only to log the warning below, and `ObjectStorageDurableCache._save_cache` rewrites the whole blob each time. Worth skipping the save when `fingerprint is None`?

providers/common/ai/docs/operators/agent.rst
```diff
@@ -300,6 +300,9 @@ cache:
    never replays responses that belong to a different conversation.
 4. After successful completion, the cached steps are deleted.
 
+If a model request or tool call cannot be fingerprinted, that step runs live on
```
kaxil: "That step" undersells it. The realistic causes are a non-JSON value in model settings (present on every request) or in the message history (which stays there), so it's every model step from that point on, not one, and durable execution ends up doing nothing for the run. I ran both through `fingerprint_model_request` and every subsequent step came back `None`. Worth saying that here so a full-price retry isn't a surprise.

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
```diff
@@ -34,9 +34,8 @@
 Fields that pydantic-ai regenerates on every attempt (message-level
 ``timestamp``/``run_id``/``conversation_id`` and part-level ``timestamp``)
 are excluded from the fingerprint.  Requests that cannot be serialized to
-JSON fingerprint as ``None``, which degrades that step to unverified
-positional replay (the pre-fingerprint behavior) rather than disabling
-caching.
+JSON fingerprint as ``None`` and re-run live instead of replaying without
```
kaxil: The comment on `_TRANSPORT_ONLY_SETTINGS` (line 66) still says an unserializable setting would "silently disable replay verification for every step", which was the old consequence. After this change it disables replay entirely for the run, which makes that frozenset load-bearing: `ModelSettings.timeout` being `float | httpx.Timeout` was already one member that needed stripping, and any other non-JSON member now costs the whole feature rather than just verification. Worth updating that comment along with these docstrings.

YAshhh29:
Thanks @kaxil — all three addressed in `98004ea265`.

**1. Skipping the save when `fingerprint is None`** — done, in both `caching_model.py` and `caching_toolset.py`. The entry is unreachable exactly as you describe: the guard needs a non-`None` current fingerprint, and a real digest never equals a stored `None`, so the write only bought a dead entry per step plus a full rewrite of the cache blob each time. The step now returns before the save.
One consequence worth noting: the warning below the comparison is now reachable only for entries that predate fingerprinting, since a `None` fingerprint is no longer something we can write.I also stopped incrementing `cached_model` / `cached_tool` on that path — otherwise the INFO summary reports steps as cached when nothing was persisted. 
Happy to drop that part if you'd rather keep this commit strictly to the dead-write fix.

**2. "That step" undersells it** — agreed, and it matches what I see putting those cases through `fingerprint_model_request`. The docs now state that the realistic causes — a non-JSON value in `model_settings`, which accompanies every request, or in the message history, which every later request carries forward — degrade every subsequent model step, leaving durable execution with nothing to replay and a retry that re-runs the agent at full cost. It also points at the `could not fingerprint model request` warning as the marker for where it began.

**3. The `_TRANSPORT_ONLY_SETTINGS` comment** — updated to the current consequence: the frozenset is load-bearing, because one non-JSON member now costs caching and replay for the whole run rather than just the verification. I carried the same correction into the module docstring, `fingerprint_model_request`'s docstring, both `log.warning` strings (they still read "this step will execute live on retry"), and the docstring on `test_httpx_timeout_does_not_disable_fingerprint`, which had the same stale framing.

**Tests** — the two existing unverifiable-step tests now assert that nothing is written, plus one new test per wrapper for the fresh-cache case. All four fail without this change and pass with it.

One adjacent thing I noticed but deliberately left out of scope: `DurableStorage.save_tool_result` skips the write for a non-serializable result, but the caller still counts it as cached, so the same counter skew exists on that path. Want that folded in here, or kept separate?

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
```diff
@@ -63,9 +65,13 @@
...
-# (``timeout`` can be an ``httpx.Timeout``) are not JSON-serializable, which
-# would otherwise force the whole fingerprint to ``None`` and silently disable
-# replay verification for every step.
+# (``timeout`` can be an ``httpx.Timeout``) are not JSON-serializable.
+#
+# This frozenset is load-bearing. Model settings ride along with every request,
+# so a single non-JSON member fingerprints every model step as ``None`` -- which
+# now costs durable execution entirely for the run (nothing is cached, nothing
+# is replayed), not merely the verification of a replay. Any non-JSON setting
+# must be listed here or normalized before it reaches the fingerprint.
```
kaxil: Following up on the comment rewrite: two things violate this invariant today and they share one cause. `_digest` calls `json.dumps` with no `default=`, so it rejects pydantic-native values that would hash perfectly deterministically. `tool_choice` is one, since it accepts `ToolOrOutput`, a plain dataclass, and it can't go in the frozenset because it does affect the response. The other is on the tool side: `tool_args` reaches `fingerprint_tool_call` after pydantic has already coerced them, so an ordinary tool with a `datetime` or `Decimal` parameter fingerprints as `None` on every call, and after this PR that tool is never cached at all. Both reproduce on 2.23.0 (the floor) and 2.42.0. Passing the payload through `to_jsonable_python` before `json.dumps` covers both, leaves plain values byte-identical so in-flight entries don't shift, and still raises `PydanticSerializationError` on `extra_body=object()`, which the existing `except (TypeError, ValueError)` catches.

providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
```diff
@@ -112,6 +114,11 @@ async def request(
             )
 
         response = await self.wrapped.request(messages, model_settings, model_request_parameters)
+        if fingerprint is None:
+            # Storing this would write an entry the guard above can never accept,
+            # once per step, each write rewriting the whole cache blob.
+            log.debug("Durable: not caching model response that cannot be verified on replay", step=step)
+            return response
```
kaxil, line 125: On your question about whether to keep the counter change: I think this half needs to go the other way. `cached_model` and `cached_tool` are rendered in `operators/agent.py` as `executed %d new steps`, and agent.rst describes that line as reporting steps replayed vs executed fresh, so the field means executed, not persisted. Skipping the increment here makes a run that replays step 0 and runs an unverifiable step 1 live print `executed 0 new steps` after paying for a model call, which is the retry cost your new docs paragraph is trying to make visible. The persisted reading doesn't hold anyway, since `save_tool_result` returns without writing on a non-serializable result and the counter still increments there. Either keep the increment and reword the log line, or add a separate not-persisted counter; the two new tests assert `cached_model == 0` so they'd move with it.

providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
```diff
@@ -112,6 +114,11 @@ async def request(
             )
 
         response = await self.wrapped.request(messages, model_settings, model_request_parameters)
+        if fingerprint is None:
+            # Storing this would write an entry the guard above can never accept,
+            # once per step, each write rewriting the whole cache blob.
```
kaxil: The blob rewrite is specific to the `ObjectStorage` backend. `TaskStateStoreDurableStorage.save_model_response` sets a single key instead, and this class holds a `DurableStorageProtocol`, so the first half of the comment carries the point on its own. Separately, this debug line is the only place `step` gets attached: the `fingerprint.py` warning that the new docs paragraph points at as the marker has no `step`, so at production log levels there is nothing to say which step stopped fingerprinting. Threading `step` into that warning would be worth more than the debug line.

providers/common/ai/docs/operators/agent.rst
```diff
@@ -300,6 +300,17 @@ cache:
    never replays responses that belong to a different conversation.
 4. After successful completion, the cached steps are deleted.
 
+If a model request or tool call cannot be fingerprinted -- it carries a value
+that will not serialize to JSON -- that step is not cached, and on retry it runs
+live rather than replaying an unverified entry. This is rarely confined to a
+single step: the usual causes are a non-serializable value in ``model_settings``,
```
kaxil: Neither of these causes applies to a tool call, though the paragraph opens by covering both. `fingerprint_tool_call` only sees the tool name, args and call id, so `model_settings` and message history cannot reach it, and the warning it emits is `could not fingerprint tool call`, not the string quoted below. The dominant tool-call cause is an ordinary typed tool parameter, which is the `_digest` point on fingerprint.py. If that gets normalized this mostly resolves itself; otherwise the causes here need scoping to the model path.

providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
```diff
@@ -158,6 +158,40 @@ async def test_legacy_entry_without_fingerprint_treated_as_miss(
         assert result is sample_response
         mock_model.request.assert_called_once()
 
+    @pytest.mark.asyncio
+    async def test_unverifiable_current_request_treated_as_miss(
+        self, mock_model, mock_storage, counter, sample_response
+    ):
+        stale = ModelResponse(parts=[TextPart(content="stale")])
+        mock_storage.load_model_response.return_value = (stale, None)
```
kaxil, line 166: One case none of the four new tests reaches: a real stored fingerprint paired with a current request that cannot be fingerprinted. The read side was already safe there before this PR, since a real hash never equalled `None`, but the write side changed. That good entry used to get clobbered with a `fingerprint=None` one and now it survives for a later attempt to replay, which is the one new write behaviour with no coverage.

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
YAshhh29: Confirmed, and I'm glad you caught this one. It's a pre-existing bug, but my last commit turned it from "replays unverified" into "never cached at all", which is the worse of the two.
I reproduced all three on 2.29.0 before touching anything: a `datetime` argument, a `Decimal` argument, and `ToolOrOutput(function_tools=[...])` in `tool_choice` each come back `None` today. The tool one bothers me most — an ordinary typed parameter isn't an edge case, and it fails silently.

`to_jsonable_python` before `json.dumps`, as you suggested. I checked the three properties it leans on rather than taking them on trust: `PydanticSerializationError` subclasses `ValueError`, so the existing handler catches it unchanged; `extra_body=object()` still raises, so the unverifiable path is intact; and plain values round-trip byte-identically. That last one now has a test asserting `_digest` of a plain payload equals its pre-normalization sha256, so in-flight entries can't drift if pydantic changes later.

`timeout` as an `httpx.Timeout` still fails under pydantic, so I've kept `_TRANSPORT_ONLY_SETTINGS` and just reworded the comment to draw the line at what pydantic can't serialize rather than what isn't JSON.

providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
YAshhh29: You're right, and I've reverted that half. "Executed" is the correct reading — `agent.py` renders these as `executed %d new steps` and `agent.rst` describes the line as replayed vs executed fresh, so suppressing the increment would print `executed 0 new steps` right after paying for a model call. That hides exactly the retry cost the new docs paragraph is there to surface.

Both counters now increment before the early return, and the two debug lines distinguish a step that was cached from one that ran without being persisted. The two tests that asserted `cached_model == 0` now assert `== 1`.

I kept the increment rather than adding a separate not-persisted counter — the summary line reads to me as being about cost, not storage accounting. Happy to add the second counter if you'd rather see both numbers.

That also settles the `save_tool_result` question I raised: under the "executed" reading its silent skip isn't an inconsistency, so there's nothing to fold in.

providers/common/ai/src/airflow/providers/common/ai/durable/caching_model.py
YAshhh29: Both done. You're right that the blob rewrite is an `ObjectStorage` detail and this class only holds a `DurableStorageProtocol`, with `TaskStateStoreDurableStorage` setting a single key — so that half is gone. What's left ("an entry the guard can never accept") makes the point without naming a backend.

Agreed the marker was the more useful half. `fingerprint_model_request` and `fingerprint_tool_call` now both take a keyword-only `step` and attach it to their warnings, so the string the docs point at names the step at production log levels instead of only the debug line carrying it. It defaults to `None`, so existing calls still work.

providers/common/ai/docs/operators/agent.rst
YAshhh29: Fair — the `_digest` normalization takes care of the dominant tool-call cause, so that part mostly resolves itself, but the paragraph was still wrong to open by covering both paths and then only give model-path causes.
It's now three parts: 
a) normalization first, so a `datetime` or `Decimal` argument isn't presented as a hazard at all,
b) then the model path, where `model_settings` and message history make it run-wide, 
c) then tool calls on their own, noting they're fingerprinted from name, args and call id alone and report `could not fingerprint tool call`.

The quoted strings now match what each path actually emits.

providers/common/ai/tests/unit/common/ai/durable/test_caching_model.py
YAshhh29: I've Added `test_unverifiable_request_leaves_a_verifiable_entry_intact` plus the toolset equivalent:
a real stored fingerprint against a current request that can't be fingerprinted, asserting the step runs live, doesn't replay, and that `save_*` isn't called — so the usable entry survives for a later attempt instead of being clobbered with a `fingerprint=None` one.

YAshhh29:
Thanks @kaxil — all five are addressed, split one commit per point: `e1de6303` (the `_digest` normalization), `b1fd367b` (counter), `d00158f8` (step in the warnings), `1fd22218` (the missing coverage case), `f85947d4` (the docs paragraph). The `_digest` one was the real find: it's a pre-existing bug, but my previous commit made it worse rather than introducing it, so I've treated it as the root fix and let the rest follow from there.


I reproduced each case before changing anything, on pydantic-ai 2.29.0:

| case | fingerprint today |
|---|---|
| `datetime` tool argument | `None` |
| `Decimal` tool argument | `None` |
| `ToolOrOutput` in `tool_choice` | `None` |

`_digest` now runs the payload through `to_jsonable_python` before `json.dumps`. I checked the three properties that make that safe instead of assuming them: `PydanticSerializationError` subclasses `ValueError`, so the existing `except (TypeError, ValueError)` catches it unchanged; `object()` still raises, so an unverifiable request still fingerprints as `None`; and plain JSON values come back byte-identical, so entries already written keep matching. There's a test pinning that last one to the pre-normalization sha256, so a future pydantic change can't quietly shift stored fingerprints.

Worth flagging: `httpx.Timeout` still fails under pydantic too, so `_TRANSPORT_ONLY_SETTINGS` is still load-bearing. I've narrowed that comment to draw the line at what pydantic can't serialize, since merely non-JSON values are fine now.

No new dependency — this provider already imports from `pydantic_core` in `utils/tool_definition.py`.

Tests: 98 in `providers/common/ai/tests/unit/common/ai/durable`, nine of them new here. Reverting just the `to_jsonable_python` call fails exactly the five pydantic-native tests while the stability and `object()` tests still pass, which is the signature I was after — the change only moves what it's meant to.

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
```diff
@@ -99,10 +114,18 @@ def _strip_volatile(messages_dump: list[dict[str, Any]]) -> list[dict[str, Any]]
 
 
 def _digest(payload: Any) -> str:
```
kaxil: One thing the normalization changes on the tool path that the old `json.dumps` did not: pydantic validates an `Iterable[int]` (or any `Iterable[T]`) tool parameter lazily and hands the tool a `ValidatorIterator`, and `to_jsonable_python` drains it to build the list. `call_tool` fingerprints `tool_args` on line 73 and passes the same dict to the wrapped tool on line 93, so the tool sees an exhausted iterator. I ran it in breeze with the installed pydantic-ai and `def total(values: Iterable[int]) -> int: return sum(values)`: the validated arg is a `ValidatorIterator`, after `to_jsonable_python` it yields nothing and the tool returns 0, and that 0 gets cached under the fingerprint of `[1, 2, 3]`. On the merge-base `json.dumps` raised `TypeError` without advancing it, so the tool still got 6. Worth either rejecting iterator-valued args before normalizing (return `None`, which is the not-cached path), or normalizing a deep copy so the original objects reach the tool untouched, plus a regression test with a real `FunctionToolset` and an `Iterable[int]` parameter.

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
```diff
@@ -99,10 +114,18 @@ def _strip_volatile(messages_dump: list[dict[str, Any]]) -> list[dict[str, Any]]
...
+    # ``datetime`` or ``Decimal`` tool argument (pydantic has already coerced tool
+    # arguments by the time they arrive) and a dataclass in ``tool_choice`` are
+    # ordinary, and hash identically on every attempt. Plain JSON values pass
+    # through unchanged, so fingerprints written by earlier versions still match.
+    #
+    # Still no ``default=`` fallback: a value pydantic cannot serialize either
+    # raises ``PydanticSerializationError`` (a ``ValueError``), so callers degrade
+    # to an unverifiable ``None`` fingerprint rather than hashing process-local
+    # reprs like ``<object at 0x...>`` that never match on retry.
+    canonical = json.dumps(to_jsonable_python(payload), sort_keys=True)
```
kaxil: `to_jsonable_python` turns a `set` or `frozenset` into a list in iteration order, and for `str` members that order depends on the process hash seed; `sort_keys=True` only sorts dict keys. Since every task attempt is a fresh process, a tool with a `set[str]` parameter now stores a digest on attempt 1 that attempt 2 never matches. Measured in breeze: six seeds gave six different digests for one `set[str]` payload, and a real pydantic-ai run stored `ba98e1f0...` then under another seed logged `reason='the conversation diverged from the previous attempt' step=1` and re-ran the tool live, while the same-seed control replayed. Before this PR that argument fingerprinted as `None`, so this is the one input where the new code costs a live re-run every retry rather than just declining to cache. A dataclass or `BaseModel` argument with a `set[str]` field, a nested set inside a list, and a tool that *returns* a set (via `ToolReturnPart.content` into the next model fingerprint) all behave the same way; `set[int]` and single-member sets happen to be stable. A recursive canonicalizer that sorts set members by their JSON encoding before hashing (dump in `mode="python"` first so dataclasses unwrap but sets stay sets) gave one digest across all seeds and left plain payloads byte-identical when I tried it. No current test can catch this because every stability test compares two digests inside one process; a subprocess pair with different `PYTHONHASHSEED`, or asserting `_digest({"a": {"x", "y"}}) == _digest({"a": ["x", "y"]})`, would.

providers/common/ai/src/airflow/providers/common/ai/durable/fingerprint.py
```diff
@@ -137,26 +164,40 @@ def fingerprint_model_request(
     except (TypeError, ValueError):
         # TypeError from json.dumps, ValueError covers PydanticSerializationError
         log.warning(
-            "Durable: could not fingerprint model request; cached responses for this "
-            "step replay without verification"
+            "Durable: could not fingerprint model request; this step will not be cached and will "
+            "execute live on retry. If the cause is in model settings or message history, every "
+            "later model step of this run is affected too",
+            step=step,
```
kaxil: Now that `step` is here, the one thing still missing from this warning is what could not be serialized. `PydanticSerializationError` carries the offending type in its message ("Unable to serialize unknown type: <class 'httpx.Timeout'>"), so `except (TypeError, ValueError) as exc` plus `error=str(exc)` here and on the tool warning below would let someone find the setting or argument without a debugger.

providers/common/ai/docs/operators/agent.rst
```diff
@@ -300,6 +300,23 @@ cache:
    never replays responses that belong to a different conversation.
 4. After successful completion, the cached steps are deleted.
 
+Fingerprints are computed from values normalized through pydantic, so ordinary
```
kaxil: Two things on this hunk. First, it now conflicts with main: #73523 moved this whole section to `providers/common/ai/docs/durable_execution.rst` (the anchor line "After successful completion, the cached steps are deleted." is at line 103 there, and `agent.rst` now just says "Moved to :doc:`../durable_execution`"), so these three paragraphs need to land there on rebase. Second, "still fingerprint normally" is broader than what the code delivers: a `set` argument fingerprints, but not stably across attempts (see the `_digest` comment). If the canonicalization lands this sentence becomes true as written; otherwise it needs the module docstring's "serialize to the same bytes on every attempt" hedge with `set` named as the exception.

providers/common/ai/docs/operators/agent.rst
```diff
@@ -300,6 +300,23 @@ cache:
...
+On the model path this is rarely confined to a single step: the causes are such a
+value in ``model_settings``, which is attached to every request, or in the message
+history, which every later request carries forward. Either one degrades all
+subsequent model steps the same way, leaving durable execution with nothing to
+replay, so the retry re-runs the agent at full cost. The
+``could not fingerprint model request`` warning names the step where this began.
+
+A tool call is fingerprinted from its name, arguments and call id alone, so
+neither of those causes reaches it. One that cannot be fingerprinted is reported
+as ``could not fingerprint tool call`` and costs only that call.
```
kaxil: "costs only that call" holds only when the live re-run returns the same content as the original attempt. The return lands in the next request's `ToolReturnPart.content`, which is part of the message history the following model step fingerprints, so a tool that reads mutable state or returns a timestamp makes every later model step miss too. Step 3 above already describes that cascade for divergence; a clause like "and, if its result differs, the model steps after it" keeps this sentence consistent with it.

PR #68741, opened 2026-06-19
Ignore stale executor success when TI is queued after defer

goingforstudying-ctrl:
Fixes a race where the trigger resumes a deferred task to `queued` (rather than `scheduled`) before the scheduler processes the executor SUCCESS from the worker defer exit. The scheduler then treated `queued` vs executor `success` as a state mismatch and failed the TI (#67287).

The fix for #66374 (#66431) added handling for the `scheduled` variant. Under load the trigger may reschedule into either `queued` or `scheduled` depending on scheduler loop timing, so both states must be treated as stale defer-exit events when `next_method` is set.

Changes:
- Extend `ti_requeued` condition in `process_executor_events` to include `TaskInstanceState.QUEUED`
- Add `test_process_executor_events_stale_success_when_queued_after_defer` (positive + negative without `next_method`)
- Add newsfragment `67287.bugfix.rst`

Testing:
```bash
pytest airflow-core/tests/unit/jobs/test_scheduler_job.py::TestSchedulerJob::test_process_executor_events_stale_success_when_queued_after_defer -q
```

* closes: #67287

PR #71579, opened 2026-08-13
Ignore stale executor success when a deferred task has resumed to running

### LINKED ISSUE #71580: Deferred task marked as failed if triggerer moves state to running

Providers

### Apache Airflow version

3.3.0

Events from audit logs:

* 2026-08-13 18:25:07 - running
* 2026-08-13 18:25:13 - deferred
* 2026-08-13 18:25:18 - running
* 2026-08-13 18:25:20 - state mismatch - `Executor LocalExecutor(parallelism=200) reported that the task instance <TaskInstance: DAG.TASK__2026-08-13T15:00:00+00:00 [running] ti_id=019ffd71-91ba-7380-865a-671e464f41ee> finished with state success, but the task instance's state attribute is running. Learn more: https://airflow.apache.org/docs/apache-airflow/stable/troubleshooting.html#task-state-changed-externally`
* 2026-08-13 18:25:20 - failed

The task should be allowed to complete by the triggerer.

### Operating System

linux

### Deployment

None

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

raghavan-arvind:
## What
Follow-up to #66431 and #68741: extends the resume-after-defer guard in `_process_executor_events` to the running state, for triggers that fire immediately after deferral because scheduler has processed success event from executor.

## Why
Ran into the following issue today, timeline from audit log:

* 2026-08-13 18:25:07 - running
* 2026-08-13 18:25:13 - deferred
* 2026-08-13 18:25:18 - running
* 2026-08-13 18:25:20 - state mismatch - `Executor LocalExecutor(parallelism=200) reported that the task instance <TaskInstance: DAG.TASK__2026-08-13T15:00:00+00:00 [running] ti_id=019ffd71-91ba-7380-865a-671e464f41ee> finished with state success, but the task instance's state attribute is running. Learn more: https://airflow.apache.org/docs/apache-airflow/stable/troubleshooting.html#task-state-changed-externally`
* 2026-08-13 18:25:20 - failed

Inline with the previous few changes to handle the race for tasks already in the `scheduled` and `queued` states. If the task is `running`, we want to handle the same way, allow the `triggerer` to finish the task instead of killing it.

## Test
Expand the previous tests to ensure tasks in the `running` state are handled correctly. All three tests are testing the same thing -- collapse them and parameterize over the resumed state. Relying on unit tests + intuition based on the observed audit logs above.

## AI Disclosure
Used AI to help write the PR. Description written by hand.

Closes #71580.

### DIFF
--- a/airflow-core/src/airflow/jobs/scheduler_job_runner.py
+++ b/airflow-core/src/airflow/jobs/scheduler_job_runner.py
@@ -1534,6 +1534,8 @@ class SchedulerJobRunner(BaseJobRunner, LoggingMixin):
             # from the worker exit after defer() has not been processed yet - should not fail it.
             # 4) the trigger already put the TI back to queued (resume after defer) but the executor success
             # from the worker exit after defer() has not been processed yet - should not fail it.
+            # 5) the trigger fired fast enough that the resumed TI is already running again (next_method
+            # set) when the executor success from the defer exit arrives - should not fail it.
 
             # All of this could also happen if the state is "running",
             # but that is handled by the scheduler detecting task instances without heartbeats.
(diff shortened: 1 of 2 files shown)


raghavan-arvind:
cc: @ashb from #66431, another one in the chain :(. I wonder if we want to just drop the event entirely if `state == SUCCESS` and `next_method is not None` (i.e. deferred), at this point, this event has nothing to do with the status of this task. We could also keep parity with above, add "RESTARTING" to this block as well.

Pedrinhonitz:
Just for the record, I’m leaving my question here as well.

Can we guarantee that the combination of `RUNNING` and `next_method != None` never corresponds to a legitimate success result from the current attempt's executor?

If that is the case, we could simply ignore the wrong execution.

Wouldn't it be more robust to consider an alternative that uses an ID associated with the execution? I provided more details on this in #72767

Issue #71577, opened 2026-08-13
[Tech Debt] Improve Unit Test DB Isolation


ferruzzi:
### Body

This is a direct follow-up to https://github.com/apache/airflow/issues/68374 with all the same reasoning: We have a lot of unit tests which start by cleaning the DB in the setup phase.  This is a clear indication that _something_ is leaking artifacts into the database causing other tests to fail.  All unit test fixtures which create a row should own that row and clean it up.  A blanket bd cleanup in the unit test's teardown should be an exception when there is no fixture, not the norm.  No unit test should require pre-cleaning in setup in order to avoid another test's garbage.  As before, I'll get to this eventually, but I'm putting up an Issue in case someone wants to get to it before I do.

After @seanghaeli pointed out other spots we didn't clean up, I set Claude on the task of finding any remaining unit tests that match the pre-cleaning pattern and it found A LOT.  The numbers below are unverified Claude output, but should serve as a solid launching point.

---
```
  Tables cleared in setup position, by frequency:
  
   91  clear_db_runs               7  clear_db_pools
   53  clear_db_dags               7  clear_db_logs
   26  clear_db_assets             7  clear_db_import_errors
   18  clear_db_connections        6  clear_db_deadline
   17  clear_db_serialized_dags    5  clear_db_backfills
   14  clear_db_jobs               4  clear_db_apdr / pakl / callbacks / triggers
    9  clear_db_variables          3  clear_db_revoked_tokens / xcom
    9  clear_db_dag_bundles        2  clear_db_deadline_alert
    8  clear_db_connection_tests   1  clear_db_dag_warnings / dag_code / task_reschedule
    8  clear_db_teams
```
---

Notably, `dag_maker.cleanup()` has the same gap for assets that it had for bundles before #69093: it deletes `AssetEvent` but not the `AssetModel` / `AssetActive` / `AssetAliasModel` rows a Dag's schedule creates.  Fixing it the same way (track what was created, delete when unreferenced, FK-safe order) should account for the ~26 `clear_db_assets` pre-cleans on its own.

### Proposal
I'm using that term very loosely, these are just Thoughts.  Cleaning up the codebase is not very useful without enforcing the change, so some ideas in addition to the actual clean up:

1. A new enforcer:  I don't know if this is a pre-commit or a ruff check or what this actually looks like, but we should be preventing tests from calling db_clean in the setup.  If a new test needs that to pass, something is already wrong and we're just masking an existing issue.

[...]

PR #71589, opened 2026-08-14
Prevent new setup-time database pre-cleaning in unit tests

shivaam:
Unit tests sometimes call `clear_db_*()` during setup. This can hide database rows leaked by an earlier test instead of fixing the fixture that created them.

This PR adds a small prek check that:

- detects direct `clear_db_*` calls in pytest fixture setup and xUnit setup methods;
- blocks new affected files;
- temporarily allows the 79 existing affected files;
- removes a file from the allowlist once its cleanup calls are fixed;
- prevents removed files from being added back.

Existing allowlisted files are exempt as a whole. This keeps the first version simple and avoids failures caused by line movement. Future PRs can fix those files incrementally.

When the check fails, it explains that the fixture or test creating the rows must clean them during teardown and links to #71577.

This changes test tooling only. It does not change Airflow runtime behavior.

Related: #71577

Validation:

- 978 prek tests passed
- full-repository checker passed
- current-main baseline simulation passed
- formatting, mypy, hook-registration, and commit checks passed

---

- [X] Yes — OpenAI Codex

Generated-by: [OpenAI Codex](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


jroachgolf84:
@shivaam - after this PR is merged, what would be the next-steps/additional changes needed?

jroachgolf84:
I assume it would be the list here?

`generated/known_clear_db_setup.txt`

shivaam:
Yes, broadly. `generated/known_clear_db_setup.txt` is the migration list for the patterns this guard detects. It contains one entry per affected file, not per cleanup call.

Follow-up PRs can take small groups from that list, fix cleanup in the fixture or test that owns the rows, and remove the setup-time `clear_db_*` calls. Once a file has no remaining detected calls, regenerating the list removes it, and the guard prevents it from being added back.

We’re also experimenting separately with a runtime detector and the `dag_maker` asset cleanup gap. Those may help prioritize later work, but they aren’t part of this PR or committed next steps yet.

PR #71592, opened 2026-08-14
Explicitly convert map_index to str

### LINKED ISSUE #71586: `map_index_template` that produces numbers fails

Airflow Core

### Apache Airflow version

3.2.0

I have a dynamic mapped task that uses `map_index_template` to make it easier to distinguish tasks in the UI.

However, when the result of map_index_template is convertible to int, Airflow/Jinja2 does that, which then fails downstream in the code.

Example that triggers a failure: `map_index_template="123"`

Error:

```
TypeError: object of type 'int' has no len()
File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1279 in run
File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1695 in _render_map_index
```

The value should not be coerced into an int so that `map_index_template` results that look like numbers are valid.

### Operating System

Debian 12 (but presumably irrelevant)

### Deployment

_No response_

_No response_

Not Applicable

### Kubernetes Version

Not Applicable

### Helm Chart configuration

_No response_

_No response_

### Anything else?

I can confirm that the value produced by the template _are_ strings. I've also tried adding `| string` to the template to force it to a string, but Jinja still forces a coercion to int. Adding some non-integer extra text to the template (e.g. `map_index_template="id: {{ ... }}"` works, but is an annoying workaround that doesn't add any additional value.

### Code of Conduct

Prajwal7842:
Previous: Dynamic task map index was using jinja's native env, so any numeric string like "123" was being coverted to int and subsequently failing. 

Fix: Explicitly render this field as text, independent of the DAG’s native-rendering setting. A test has been added to validate the fix. 

---

closes: #71586 

---

---

### DIFF
--- a/task-sdk/src/airflow/sdk/execution_time/task_runner.py
+++ b/task-sdk/src/airflow/sdk/execution_time/task_runner.py
@@ -69,6 +69,7 @@ from airflow.sdk.definitions.asset import (
     AssetUniqueKey,
     AssetUriRef,
 )
+from airflow.sdk.definitions.context import render_template_to_string
 from airflow.sdk.definitions.mappedoperator import MappedOperator
 from airflow.sdk.definitions.param import process_params
 from airflow.sdk.exceptions import (
@@ -2259,7 +2260,7 @@ def _render_map_index(context: Context, ti: RuntimeTaskInstance, log: Logger) ->
         return None
     log.debug("Rendering map_index_template", template_length=len(template))
     jinja_env = ti.task.dag.get_template_env()
-    rendered_map_index = jinja_env.from_string(template).render(context)
+    rendered_map_index = render_template_to_string(jinja_env.from_string(template), context)
     log.debug("Map index rendered", length=len(rendered_map_index))
     return rendered_map_index
 
--- a/task-sdk/tests/task_sdk/execution_time/test_task_runner.py
+++ b/task-sdk/tests/task_sdk/execution_time/test_task_runner.py
@@ -2111,6 +2111,30 @@ def test_execute_success_task_with_rendered_map_index(create_runtime_ti, mock_su
     assert ti.rendered_map_index == "Hello! test_run"
 
 
+def test_execute_success_task_with_numeric_rendered_map_index(create_runtime_ti, mock_supervisor_comms):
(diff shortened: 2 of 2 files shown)


jroachgolf84 (approved):
LGTM, I like the fix.

Prajwal7842:
Hi @amoghrajesh Can you help with the review for this fix. Thank you.

msumit (approved):

task-sdk/src/airflow/sdk/execution_time/task_runner.py
```diff
@@ -2259,7 +2260,7 @@ def _render_map_index(context: Context, ti: RuntimeTaskInstance, log: Logger) ->
         return None
     log.debug("Rendering map_index_template", template_length=len(template))
     jinja_env = ti.task.dag.get_template_env()
-    rendered_map_index = jinja_env.from_string(template).render(context)
+    rendered_map_index = render_template_to_string(jinja_env.from_string(template), context)
```
kaxil, line 2263: This forces string-mode rendering (`native=False`, `"".join(nodes)` in `render_template`), but the template is compiled by whatever `dag.get_template_env()` returns. That's a jinja2 `NativeEnvironment` when `render_template_as_native_obj=True`. `NativeCodeGenerator` skips wrapping output nodes in `str()`, so a template like `map_index_template="index-{{ ti.try_number }}"` yields `['index-', 1]` from `root_render_func`, and `"".join(['index-', 1])` raises `TypeError: sequence item 1: expected str instance, int found`. Even a bare `"{{ ti.try_number }}"` crashes the same way, turning an otherwise-successful task into a failed one.

The new test only covers a bare literal `"123"` (no `{{ }}`). That stays a plain-text `str` node regardless of which code generator compiled it, so it doesn't catch this case: a label combined with a mapped or try index, usually an `int`.

A fix that holds under both env modes: always render through the native path (`render_template_as_native`) and `str()`-cast the result if it isn't `None`. `native_concat` already does per-node `str()` coercion when there's more than one node, and passes through an already-`str` single-node result untouched. The only missing piece is the final cast.

PR #71621, opened 2026-08-14
Cut the per-file cost of persisting Dag parse results

ephraimbuddy:
The Dag processor persists parse results one file at a time, and on small deployments that is the CPU bottleneck. Two per-file costs are removed.

Import errors were read by loading the whole `import_error` table, even for files with none. The lookup is now skipped when there are none, and scoped to the keys being written otherwise. There is no index on `(bundle_name, filename)`, so this saves materialising every row rather than the scan itself.

The serialized Dag was hashed twice per write — once for change detection, once in the model constructor — and hashing re-sorts and re-encodes the whole structure.

The budget pinned in #71720 moves with it: a clean parse now costs 9 statements per call instead of 10. The import-error path was unpriced, so it gets a case of its own.

Also makes the compressed-storage test parametrization real. It never ran, because the module binds that flag at import, and this change moves code into that branch.

From @seanmuth's investigation of the Airflow 2-to-3 dag-processor CPU regression: github.com/seanmuth/af2-af3-scheduler-cpu-repro. Two of the smaller levers; per-sweep batching is follow-up.

 

---

### DIFF
--- a/airflow-core/src/airflow/dag_processing/collection.py
+++ b/airflow-core/src/airflow/dag_processing/collection.py
@@ -382,16 +382,26 @@ def _update_dag_warnings(
 
 def _update_import_errors(
     files_parsed: set[tuple[str, str]],
-    bundle_name: str,
     import_errors: dict[tuple[str, str], str],
     session: Session,
 ):
     from airflow.listeners.listener import get_listener_manager
 
-    # Check existing import errors BEFORE deleting, so we can determine if we should update or create
-    existing_import_error_files = set(
-        session.execute(select(ParseImportError.bundle_name, ParseImportError.filename))
-    )
+    # Check existing import errors BEFORE deleting, so we can determine if we should update or create.
+    # Only the rows about to be written are needed, so scope the lookup to those keys instead of
+    # loading the whole table into memory on each parse result. The table is not indexed on
+    # (bundle_name, filename), so this still scans; what it saves is building a Row object and a
+    # set entry for every import error the deployment is carrying.
+    existing_import_error_files: set[tuple[str | None, str | None]] = set()
+    if import_errors:
+        existing_import_error_files = {
+            (bundle, filename)
+            for bundle, filename in session.execute(
+                select(ParseImportError.bundle_name, ParseImportError.filename).where(
+                    tuple_(ParseImportError.bundle_name, ParseImportError.filename).in_(list(import_errors))
+                )
+            )
+        }
 
     # Delete errors for files that were parsed but don't have errors in import_errors
     # (i.e., files that were successfully parsed without errors)
@@ -439,7 +449,7 @@ def _update_import_errors(
         else:
             import_error = ParseImportError(
                 filename=relative_fileloc,
-                bundle_name=bundle_name,
+                bundle_name=bundle_name_,
                 timestamp=utcnow(),
                 stacktrace=stacktrace,
             )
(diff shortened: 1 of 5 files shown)


airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -346,15 +346,23 @@ class SerializedDagModel(Base):
...
+            stands now — recompute it first if the data was mutated after hashing, or change
+            detection on the next parse silently misbehaves.
+        """
         self.dag_id = dag.dag_id
         dag_data = dag.data
-        self.dag_hash = SerializedDagModel.hash(dag_data)
-
-        # partially ordered json data
-        dag_data_json = json.dumps(dag_data, sort_keys=True).encode("utf-8")
+        self.dag_hash = SerializedDagModel.hash(dag_data) if _dag_hash is None else _dag_hash
```
jedcunningham:
```suggestion
        self.dag_hash = _dag_hash or SerializedDagModel.hash(dag_data)
```

Likely safe?

airflow-core/tests/unit/dag_processing/test_collection.py
```diff
@@ -1437,6 +1438,105 @@ def test_max_consecutive_failed_dag_runs_defaults_from_conf_when_none(
...
+        )
+
+        selects = self._selects(import_error_statements)
+        assert selects, "expected the existing-error lookup to run"
+        assert all("where" in stmt.lower() for stmt in selects), (
+            f"import_error must never be scanned unfiltered, got: {selects}"
+        )
+
+        # A list, not a dict: import_error has no unique constraint on (bundle_name, filename), so
+        # keying by it would silently collapse a duplicate row the lookup failed to match.
```
jedcunningham: This almost feels like a gap. Either way, probably dont need this comment.

airflow-core/src/airflow/dag_processing/collection.py
```diff
@@ -394,16 +394,26 @@ def _update_dag_warnings(
...
     import_errors: dict[tuple[str, str], str],
     session: Session,
 ):
     from airflow.listeners.listener import get_listener_manager
 
-    # Check existing import errors BEFORE deleting, so we can determine if we should update or create
-    existing_import_error_files = set(
-        session.execute(select(ParseImportError.bundle_name, ParseImportError.filename))
-    )
+    # Check existing import errors BEFORE deleting, so we can determine if we should update or create.
```
jedcunningham: We can probably simplify this comment a bit.

airflow-core/src/airflow/models/serialized_dag.py
```diff
@@ -346,15 +346,23 @@ class SerializedDagModel(Base):
     load_op_links = True
     __table_args__ = (Index("idx_serialized_dag_dag_id_created_at", dag_id, created_at),)
 
-    def __init__(self, dag: LazyDeserializedDAG) -> None:
+    def __init__(self, dag: LazyDeserializedDAG, *, _dag_hash: str | None = None) -> None:
+        """
+        Build a serialized Dag row.
+
+        :param _dag_hash: hash of ``dag.data`` when the caller has already computed it, used by
```
jedcunningham: Here too, simpler is better.

jedcunningham (approved):

airflow-core/src/airflow/models/serialized_dag.py
ephraimbuddy: Yeah, safe

airflow-core/tests/unit/dag_processing/test_collection.py
ephraimbuddy: Yeah, there's no unique constraint but duplicate won't be possible still with this change except down below where there's a listener lookup. We can tighten it later in a different PR

PR #71256, opened 2026-08-06
fix(sensors): only skip timeout-related exceptions on deferred sensor soft_fail

waterWang:
## Description

When `soft_fail=True`, deferred sensors currently skip on **all** exceptions, while non-deferred (`poke`/`reschedule` mode) sensors only skip on timeout-related exceptions (`AirflowSensorTimeout`, `AirflowTaskTimeout`, `AirflowFailException`) and let unexpected errors propagate.

This behavior gap means operators cannot distinguish between a sensor timeout (legitimate skip) and a sensor misconfiguration or unexpected error when using deferred mode with `soft_fail=True`.

## Fix

In `BaseSensorOperator.resume_execution()`, split the exception handling into two tiers matching the `execute()` method:

1. **Timeout-related exceptions** (`AirflowSensorTimeout`, `AirflowTaskTimeout`, `AirflowFailException`) → `soft_fail` and `never_fail` apply (skip)
2. **Other exceptions** (`AirflowException`, `TaskDeferralError`) → only `never_fail` applies (not `soft_fail`)

## Related issue

Closes #71255

PR #71625, opened 2026-08-14
Add DatadogMonitorSensorAsync waiting on a monitor's overall state

raghavan-arvind:
## Why

The existing `DatadogSensor` queries the event stream for transition events in a time window. That shape cannot answer "is this monitor healthy right now": a monitor that entered Alert before the window (or recovered before it) emits no event inside it, so gating a pipeline on monitor health requires reconstructing state from transitions.

This adds `DatadogMonitorSensorAsync`, which checks a monitor's current `overall_state` — an already-alerting or already-recovered monitor is observed regardless of when it last changed state.

## What

- `DatadogMonitorSensorAsync`: waits for a monitor to reach one of `target_states` (default `("OK",)`). Deferrable-only, and named `*Async` to make that explicit: it waits in the triggerer via a new `DatadogMonitorTrigger` and never occupies a worker slot while waiting; there is deliberately no poke-mode fallback.
- `DatadogMonitorTrigger`: polls the monitor state off the event loop (`asyncio.to_thread`, since the datadog client is sync) and fires when a target state is reached.
- `soft_fail` on the deferral path applies only to timeouts, matching poke/reschedule semantics elsewhere: a trigger crash (bad monitor id, auth failure) fails the task loudly instead of being converted into a skip that silently skips the downstream branch. Works on both Airflow 3.x (`TaskDeferralTimeout`) and 2.11 (timeout arrives as `TaskDeferralError("Trigger/execution timeout")`).
- The datadog library returns API errors as a payload instead of raising; they surface as a clear `AirflowException` (e.g. `Datadog monitor 999999999: ['Monitor not found']`).
- provider.yaml gains a `triggers` section; unit tests for the sensor (defer, completion, soft_fail semantics) and trigger (serialize round-trip, poll-until-target, API error).

## Testing

`uv run --project providers/datadog pytest providers/datadog/tests/unit/datadog`: 18 passed. Also validated end to end in a production-like Airflow 3.3 deployment with canary tasks covering pass / timeout / timeout-with-soft_fail / missing-monitor / missing-with-soft_fail.

### DIFF
--- a/providers/datadog/src/airflow/providers/datadog/sensors/datadog.py
+++ b/providers/datadog/src/airflow/providers/datadog/sensors/datadog.py
@@ -17,13 +17,33 @@
 # under the License.
 from __future__ import annotations
 
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
+from datetime import timedelta
 from typing import TYPE_CHECKING, Any
 
 from datadog import api
 
-from airflow.providers.common.compat.sdk import AirflowException, BaseSensorOperator
+from airflow.providers.common.compat.sdk import (
+    AirflowException,
+    AirflowSensorTimeout,
+    AirflowSkipException,
+    BaseSensorOperator,
+)
 from airflow.providers.datadog.hooks.datadog import DatadogHook
+from airflow.providers.datadog.triggers.datadog import DatadogMonitorTrigger
+
+try:
+    from airflow.sdk.exceptions import TaskDeferralError, TaskDeferralTimeout
+except ImportError:  # Airflow 2.x
+    from airflow.exceptions import TaskDeferralError
+
+    try:
+        from airflow.exceptions import TaskDeferralTimeout
+    except ImportError:
+
+        class TaskDeferralTimeout(TaskDeferralError):  # type: ignore[no-redef]
+            """Not raised before Airflow 3; timeouts arrive as TaskDeferralError."""
+
 
 if TYPE_CHECKING:
     from airflow.providers.common.compat.sdk import Context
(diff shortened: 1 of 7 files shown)


raghavan-arvind:
Verified all these sensors return the correct result:

```
# Datadog sensor canaries: every (case x mode) combination, alerting when
# the outcome differs from the expected one encoded in the task name.
DATADOG_CANARY_CASES = {
    # name -> (monitor_id, target_states, soft_fail, expected outcome)
    'ok_should_pass': (DATADOG_OK_MONITOR_ID, ('OK',), False, 'success'),
    'alert_timeout_should_fail': (DATADOG_OK_MONITOR_ID, ('Alert',), False, 'failed'),
    'alert_timeout_soft_fail_should_skip': (DATADOG_OK_MONITOR_ID, ('Alert',), True, 'skipped'),
    'missing_should_fail': (DATADOG_INVALID_MONITOR_ID, ('OK',), False, 'failed'),
    'missing_soft_fail_should_fail': (DATADOG_INVALID_MONITOR_ID, ('OK',), True, 'failed'),
}

for case_name, (monitor_id, target_states, soft_fail, expected) in DATADOG_CANARY_CASES.items():
  for sensor_mode in ('poke', 'reschedule', 'deferrable'):
    callbacks = {
        'success': {},
        'failed': {
            'on_failure_callback': None,
            'on_success_callback': alert_on_failure,
            'on_skipped_callback': alert_on_failure,
        },
        'skipped': {'on_success_callback': alert_on_failure},
    }[expected]
    DatadogMonitorSensor(
        dag=dag,
        task_id=f'datadog_monitor_{case_name}_{sensor_mode}',
        monitor_id=monitor_id,
        target_states=target_states,
        soft_fail=soft_fail,
        timeout=timedelta(seconds=90),
        poke_interval=30,
        deferrable=sensor_mode == 'deferrable',
        **({'mode': sensor_mode} if sensor_mode != 'deferrable' else {}),
        **callbacks,
    )
```

raghavan-arvind:
The soft_fail tests are reliant on some fix like:
https://github.com/apache/airflow/pull/71256

Would like to discuss that PR first before this one

providers/datadog/src/airflow/providers/datadog/sensors/datadog.py
```diff
@@ -98,3 +119,80 @@ def poke(self, context: Context) -> bool:
...
+            return super(BaseSensorOperator, self).resume_execution(next_method, next_kwargs, context)
+        except TaskDeferralError as e:
+            timed_out = isinstance(e, TaskDeferralTimeout) or str(e) == "Trigger/execution timeout"
+            if timed_out and self.soft_fail:
+                raise AirflowSkipException(str(e)) from e
+            if getattr(self, "never_fail", False):
+                raise AirflowSkipException(str(e)) from e
+            if timed_out:
+                raise AirflowSensorTimeout(*e.args) from e
+            raise
```
raghavan-arvind, line 188: Ideally we can remove this and merge some variant of #71256 first.

raghavan-arvind:
Would like to only support `deferrable` for this sensor -- is that okay? Any existing patterns for this? For retrying transient errors, `500`, `429`, the semantics of changing the poll interval temporarily into an exponential backoff are more straightforward that way

I know `TimeSensorAsync` is deprecated in favor of `TimeSensor` with `deferrable=True`, but would prefer to write a `DatadogMonitorSensorAsync` to standardize on `deferrable` going forward.

edit: updated

PR #71628, opened 2026-08-14
Fix Dags with deadlines failing to parse after a database retry

ephraimbuddy:
Writing a Dag with a deadline rewrites its serialized deadline entries from definitions into UUID references, and it did so on the object the caller passed in. When update_dag_parsing_results_in_db hit a transient database error it rolled back and retried the same objects, which by then held UUID strings where definitions were expected. The retry raised AttributeError while re-encoding them, and since that is not a database error it escaped the retry entirely, so the manager discarded the whole file's parse results rather than one Dag's.

Rewrite a copy instead. Only the "dag" sub-dict is copied, so the bulk of the serialized structure is still shared, and the Dag the caller holds stays in the shape it arrived in however many times the write is attempted.

 

---

### DIFF
--- a/airflow-core/src/airflow/models/serialized_dag.py
+++ b/airflow-core/src/airflow/models/serialized_dag.py
@@ -346,9 +346,16 @@ class SerializedDagModel(Base):
     load_op_links = True
     __table_args__ = (Index("idx_serialized_dag_dag_id_created_at", dag_id, created_at),)
 
-    def __init__(self, dag: LazyDeserializedDAG) -> None:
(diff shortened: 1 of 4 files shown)


ferruzzi (approved):
LGTM, nice catch

ephraimbuddy:
Closing this as https://github.com/apache/airflow/pull/70148 fixed same issue and is merged now

ferruzzi:
To be fair, I think your tests are more thorough and still worth saving, but the bug itself is fixed.

Issue #60770, opened 2026-01-19
Cleanup: Move remaining provider tests from airflow-core to support standalone uv sync


MadhuTiwari-345:
---

## Apache Airflow version

3.1.6

## If "Other Airflow 3 version" selected, which one?

main (development)

---

## What happened?

As part of the Airflow 3.x development effort, the repository is transitioning toward a stricter separation between `airflow-core` and provider packages.

Running:

```bash
cd airflow-core
uv sync
```

successfully creates a minimal core-only environment.

However, running the test suite in this environment reveals failures because **some tests located in `airflow-core` still import provider modules** (e.g. `airflow.providers.*`). Since providers and their dependencies are not installed in a core-only setup, those tests fail.

There is an existing TODO in `contributing-docs/07_local_virtualenv.rst` noting that standalone core development is not fully supported until remaining provider-dependent tests are moved out of `airflow-core`.

The current state prevents developers from running the core test suite independently without installing all providers.

---

## What you think should happen instead?

`airflow-core` should be fully testable in isolation.

Specifically:

* All tests under `airflow-core` should depend only on core functionality.
* Tests that require provider functionality should be moved to the appropriate `providers/*` directories.
* After cleanup, developers should be able to:

```bash
cd airflow-core
uv sync
pytest
```

and have the full core test suite pass without installing provider packages.

This aligns with the architectural goal of strict separation between core and providers and improves developer experience by reducing setup complexity and dependency surface area.

---

## How to reproduce

1. Clone the current `main` branch of the Airflow repository.
2. Navigate to the core directory:

```bash
cd airflow-core
```

3. Synchronize the environment:

```bash
uv sync
```

4. Run the test suite:

```bash
pytest
```

Observe: Some tests fail due to imports from `airflow.providers.*`, indicating that provider-dependent tests still exist in the core test suite.

---

## Operating System

Linux / macOS

---

## Versions of Apache Airflow Providers

Not applicable (core-only environment)

---

## Deployment

Other

---

## Anything else?

This task is part of the broader Airflow 3.x effort to enforce clean separation between core and providers.

This PR will:

* Identify provider-dependent tests currently under `airflow-core`
* Move them to the appropriate provider packages
* Ensure `airflow-core` tests pass in a standalone core-only environment

---

## Are you willing to submit PR?

* [x] Yes I am willing to submit a PR!

---

MadhuTiwari-345:
[...]

PR #71637, opened 2026-08-14
Fix airflow-core tests that silently required the full provider set

coleheflin:
## What

- Fixes 3 `tests_common` files (`test_utils/logs.py`, `api_fastapi.py`, `providers.py`) that imported `airflow_shared.secrets_masker`/`module_loading` — a private, never-published distribution only resolvable in a full local workspace checkout — instead of the vendored `airflow._shared.*` copy airflow-core bakes into its own installed package (same pattern already used elsewhere, e.g. this file's own `pytest_plugin.py`).
- Adds a `skip_unless_all_providers_installed` marker (plus a couple of targeted `pytest.importorskip`s) to 11 tests in `test_providers_manager.py`/`test_configuration.py` that assumed the full ~100+ provider set is always installed.
- Links the existing TODO in `airflow-core/pyproject.toml` (about eventually dropping its own hardcoded provider dev-dependencies) to a new tracking issue, #71641, since it had none.

## Why

addresses (partially): #60770

`uv sync --project airflow-core` (a scoped, core-only sync) succeeds, but the test suite didn't run clean: collection crashed with `ModuleNotFoundError: No module named 'airflow_shared'` (it only worked before as a side effect of an unrelated full workspace sync), and 11 more tests failed on full-provider-set assumptions. Fixing both gets `pytest airflow-core/tests/unit -m "not db_test"` passing cleanly in that scoped environment.

(An earlier version fixed the import crash by adding those private packages as explicit `devel-common` dependencies. That broke the `Compat 3.1.8/3.2.2/3.3.1` CI jobs instead — they deliberately strip local editable packages before installing a released `apache-airflow` wheel, so the new dependency reintroduced the same error there. Using the vendored `airflow._shared.*` path avoids the dependency entirely and survives that environment too.)

**This only partially addresses #60770.** Its title ("move remaining provider tests from airflow-core") implies a bigger migration: `airflow-core`'s own dev group still hardcodes 6 providers (`amazon`, `celery`, `cncf-kubernetes`, `fab`, `git`, `ftp`) so tests depending on them still pass — nothing was moved or removed here. Untangling which tests genuinely need which of those 6, and moving vs. guarding each, is a much larger multi-PR effort, so it's tracked separately in #71641 rather than folded into this PR. Leaving #60770 open for a maintainer to decide whether this is enough to close it.

## Testing performed

- Reproduced both failure modes via `uv sync --project airflow-core` + `pytest`; after the fix, collection is clean (0 errors) and the non-db unit suite passes fully (0 failed, was 11 failed) — with no new dependency added.
- Restored a full `uv sync` and reran the same files to confirm all tests still pass for real (marker correctly no-ops when all providers are present).
- Reproduced the CI regression locally (scoped `shared/observability` venv, and a compat-style env with airflow-core but no standalone shared packages) to verify the intermediate approach failed and the final one passes.
- Confirmed on CI: all 5 `Compat` jobs pass (including the two that broke on the intermediate approach), `Shared observability tests` passes, no failures anywhere.
- Rebased on latest `main`; `prek run --from-ref upstream/main --stage pre-commit` (ruff, mypy) passes.

---

### DIFF
--- a/airflow-core/pyproject.toml
+++ b/airflow-core/pyproject.toml
@@ -290,7 +290,7 @@ dev = [
     "apache-airflow-devel-common",
     "apache-airflow-task-sdk",
     # TODO(potiuk): eventually we do not want any providers nor apache-airflow extras to be needed for
-    # airflow-core tests
+    # airflow-core tests; tracked at https://github.com/apache/airflow/issues/71641
     "apache-airflow[pandas,polars]",
     "apache-airflow-providers-amazon",
     "apache-airflow-providers-celery",
--- a/airflow-core/tests/unit/always/test_providers_manager.py
+++ b/airflow-core/tests/unit/always/test_providers_manager.py
@@ -40,7 +40,10 @@ from airflow.providers_manager import (
     RemoteLoggingInfo,
 )
 
-from tests_common.test_utils.markers import skip_if_force_lowest_dependencies_marker
+from tests_common.test_utils.markers import (
+    skip_if_force_lowest_dependencies_marker,
+    skip_unless_all_providers_installed,
+)
 
 if TYPE_CHECKING:
     from unittest.mock import MagicMock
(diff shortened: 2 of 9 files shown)


shahar1 (commented):
Please resolve conflicts

coleheflin:
Rebased onto the latest `main` and resolved the conflicts; the PR is now mergeable.

PR #71639, opened 2026-08-14
Fix lang-SDK k8s test on release branches by using their own SDKs

potiuk:
The lang-SDK k8s test fetched `go-sdk/`/`java-sdk/` from upstream `main` on any
run not targeting `main`, on the assumption that a release branch either lacks
them or carries a stale copy.

That copy isn't stale — it's the matching one. A packed bundle declares a dated
`supervisor_schema_version` and the task-SDK supervisor refuses a bundle whose
version it doesn't know, so upstream `main`'s SDK cannot run against a release
branch's Airflow at all. `_lang_sdk_resolve_sdk_sources()` now builds from the
checkout's own `go-sdk`/`java-sdk` whenever it has them; the upstream-`main`
fetch stays only as the fallback for a branch cut before those directories
existed. No behaviour change for `main`, which always has them.

Seen on `v3-3-test`, where every run of `K8S Lang-SDK` has failed since `main`
moved to supervisor schema `2026-10-30` while the branch is on `2026-06-16`:
`go_extract` dies with `cannot find executable bundle with usable
supervisor_schema_version`.

related: #71527

---

### DIFF
--- a/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
+++ b/dev/breeze/src/airflow_breeze/commands/kubernetes_commands.py
@@ -32,7 +32,6 @@ from typing import Any
 import click
 import yaml
 
-from airflow_breeze.branch_defaults import AIRFLOW_BRANCH
 from airflow_breeze.commands.common_options import (
     option_answer,
     option_debug_resources,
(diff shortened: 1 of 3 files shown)


jason810496 (approved):
> Seen on v3-3-test, where every run of K8S Lang-SDK has failed since main
moved to supervisor schema 2026-10-30 while the branch is on 2026-06-16:
go_extract dies with cannot find executable bundle with usable supervisor_schema_version.

Yes, I just realized the same root cause yesterday but haven't got chance to fix it. Thanks.

PR #71645, opened 2026-08-15
Add SSL cipher list option to the api server

### LINKED ISSUE #71426: Missing SSL options for API Server

Airflow Core

### Apache Airflow version

3.3.0

The command used to launch the API web server only allow a set of SSL options for [Uvicorn](https://uvicorn.dev/settings/#https) & [Gunicorn](https://gunicorn.org/reference/settings/#ssl). Some are missing, including Uvicorn's `ssl-ciphers` in my case.

Uvicorn's `ssl-ciphers` and Gunicorn's `ciphers` options should be available for the API web server through the CLI args in api_server_command.py
Arguably, we should be able to overload any option on these server.

### Operating System

Red Hat 9.8

### Deployment

Virtualenv installation

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

ColtenOuO:
### Sumarry

The api server exposes the certificate, key, CA file and client-verification mode for its TLS listener, but not the cipher list. Deployments with a TLS policy that requires pinning an explicit set of cipher suites have no way to express it, and are stuck with whatever Python's `ssl` module defaults to.

This adds an `[api] ssl_ciphers` configuration option and the matching `airflow api-server --ssl-ciphers` flag, wired into both server backends: `ssl_ciphers` for uvicorn and `ciphers` for gunicorn.

Leaving the option empty keeps the current behaviour, so existing deployments are unaffected. Both uvicorn and gunicorn default their cipher setting to `None` (meaning "use the Python default"), so an unset value is passed through unchanged rather than substituting a value of our own.

A cipher list that OpenSSL cannot resolve is rejected during argument handling with an `AirflowConfigException`, instead of surfacing later as a bare `ssl.SSLError: No cipher can be selected` while the server binds its socket.

The linked issue also suggests being able to override *any* uvicorn/gunicorn option. That is a much larger design question about how much of the underlying servers' surface Airflow should expose, so this PR is deliberately scoped to the cipher list that was actually blocking the reporter.

### Changes

- **`airflow-core/src/airflow/config_templates/config.yml`**
Adds the `[api] ssl_ciphers` option (`version_added: 3.4.0`, default empty), documented as an OpenSSL cipher list with a link to the cipher-list format reference.

- **`airflow-core/src/airflow/cli/cli_config.py`**
Adds `ARG_SSL_CIPHERS` (`--ssl-ciphers`), defaulting to the config value, and registers it on the `api-server` command next to the existing `--ssl-*` flags.

- **`airflow-core/src/airflow/cli/commands/api_server_command.py`**
Adds a `_ssl_ciphers()` helper alongside the existing `_ssl_cert_reqs()` / `_get_ssl_filepaths()` helpers. It normalises an empty value to `None` and validates a non-empty list by applying it to a throwaway `SSLContext`, raising `AirflowConfigException` if OpenSSL cannot resolve it. Both the uvicorn and the gunicorn code paths call it.

- **`airflow-core/src/airflow/api_fastapi/gunicorn_app.py`**
`create_gunicorn_app()` takes a new `ssl_ciphers` argument and maps it to gunicorn's `ciphers` option. It is set inside the existing `if ssl_cert and ssl_key:` block, matching how `ca_certs` and `cert_reqs` are already handled, so ciphers never leak into a plain-HTTP configuration.

closes: #71426

---

- [X] Yes — Claude Code (Opus 5) for writting unit test.

### DIFF
--- a/airflow-core/tests/unit/cli/commands/test_api_server_command.py
+++ b/airflow-core/tests/unit/cli/commands/test_api_server_command.py
@@ -158,8 +158,28 @@ class TestCliApiServer(_CommonCLIUvicornTestClass):
                     "ssl_certfile": "ssl_cert_path_placeholder",
                     "ssl_ca_certs": "ssl_ca_file_placeholder",
                     "ssl_cert_reqs": ssl.CERT_REQUIRED,
+                    "ssl_ciphers": None,
                 },
                 id="api-server with SSL cert and key",
+            ),
+            pytest.param(
+                [
+                    "api-server",
+                    "--ssl-cert",
+                    "ssl_cert_path_placeholder",
+                    "--ssl-key",
+                    "ssl_key_path_placeholder",
+                    "--ssl-ciphers",
+                    "ECDHE-RSA-AES256-GCM-SHA384",
+                ],
+                {
+                    "ssl_keyfile": "ssl_key_path_placeholder",
+                    "ssl_certfile": "ssl_cert_path_placeholder",
+                    "ssl_ca_certs": None,
+                    "ssl_cert_reqs": ssl.CERT_NONE,
+                    "ssl_ciphers": "ECDHE-RSA-AES256-GCM-SHA384",
+                },
+                id="api-server with SSL ciphers",
             ),
             pytest.param(
                 [
@@ -172,6 +192,7 @@ class TestCliApiServer(_CommonCLIUvicornTestClass):
                     "ssl_certfile": None,
                     "ssl_ca_certs": None,
                     "ssl_cert_reqs": ssl.CERT_NONE,
+                    "ssl_ciphers": None,
                     "log_config": "my_log_config.yaml",
                 },
                 id="api-server with log config",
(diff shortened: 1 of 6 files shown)


fat-catTW (approved):
Awesome!

airflow-core/tests/unit/cli/commands/test_api_server_command.py
```diff
@@ -209,6 +228,7 @@ def test_args_to_uvicorn(self, ssl_cert_key_and_ca, cli_args, expected_additiona
                     "log_level": "info",
                     "proxy_headers": args.proxy_headers,
                     "log_config": None,
+                    "ssl_ciphers": None,
```
pierrejeambrun: This isn't correct. THis shouldn't always be None, we are parametrizing this. `ssl_ciphers` isn't forwarded properly to the command.

pierrejeambrun (approved):
Thanks LGTM.

One test need fixing though, cf comment below.

airflow-core/tests/unit/cli/commands/test_api_server_command.py
ColtenOuO: Thanks for reviewing and catch this, I've fixed this porblem!

pierrejeambrun (approved):
Nice thanks.

PR #71646, opened 2026-08-15
Use the configured region for deferred Neptune cluster tasks

SEPURI-SAI-KRISHNA:
All three Neptune triggers declare a `region_name` parameter, document it, and read
`self.region_name` when building their hook — but never forward it to
`AwsBaseWaiterTrigger.__init__`, which is what actually assigns the attribute:

```python
def __init__(self, *, db_cluster_id: str, ..., region_name: str | None = None, **kwargs):
    super().__init__(
        ...,
        aws_conn_id=aws_conn_id,
        # region_name never passed
        **kwargs,
    )

def hook(self) -> AwsGenericHook:
    return NeptuneHook(aws_conn_id=self.aws_conn_id, region_name=self.region_name, ...)
```

Because `region_name` is an explicit named parameter it is not swept up by the `**kwargs`
forwarding either — it is captured and discarded, so `self.region_name` is always `None`
and the deferred waiter polls the account's default region.

The failure is quiet and slow: the waiter simply never finds the cluster in the region it
is looking at, so the task burns through `waiter_max_attempts` and then fails with a
message that does not mention regions at all. `AwsBaseWaiterTrigger.serialize()` writes
`self.region_name`, so the wrong value also survives triggerer restarts.

There is a second, independent layer in the operators. Two of the four defer sites already
pass the hook configuration; the two in `NeptuneStartDbClusterOperator.execute` and
`NeptuneStopDbClusterOperator.execute` pass none of it, so fixing only the triggers would
still leave those paths on the default region. Both are brought in line with the existing
sites, which is why `verify` and `botocore_config` are included alongside `region_name`.

Affects `NeptuneClusterAvailableTrigger`, `NeptuneClusterStoppedTrigger` and
`NeptuneClusterInstancesAvailableTrigger`.

Tests cover both layers — that the triggers retain and serialize `region_name`, and that
the operators hand it to the trigger they defer with. All five fail without the change.

---

### DIFF
--- a/providers/amazon/tests/unit/amazon/aws/operators/test_neptune.py
+++ b/providers/amazon/tests/unit/amazon/aws/operators/test_neptune.py
@@ -39,6 +39,10 @@ from unit.amazon.aws.utils.test_template_fields import validate_template_fields
 
 CLUSTER_ID = "test_cluster"
 REGION_NAME = "eu-west-2"
+VERIFY = False
+BOTOCORE_CONFIG = {"read_timeout": 42}
+WAITER_DELAY = 12
+WAITER_MAX_ATTEMPTS = 34
 
 EXPECTED_RESPONSE = {"db_cluster_id": CLUSTER_ID}
 
@@ -67,7 +71,7 @@ def _create_cluster(hook: NeptuneHook):
     ],
 )
 @mock.patch.object(NeptuneHook, "conn")
-def test_deferred_trigger_receives_region_name(mock_conn, operator_class, trigger_class):
+def test_deferred_trigger_receives_the_operator_configuration(mock_conn, operator_class, trigger_class):
     operator = operator_class(
         task_id="task_test",
         db_cluster_id=CLUSTER_ID,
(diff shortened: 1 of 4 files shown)


vincbeck:
@seanghaeli @ramitkataria

o-nikolas (approved):
Seems reasonable to me

providers/amazon/tests/unit/amazon/aws/operators/test_neptune.py
```diff
@@ -54,6 +59,31 @@ def _create_cluster(hook: NeptuneHook):
...
 
+@pytest.mark.parametrize(
+    ("operator_class", "trigger_class"),
+    [
+        (NeptuneStartDbClusterOperator, NeptuneClusterAvailableTrigger),
+        (NeptuneStopDbClusterOperator, NeptuneClusterStoppedTrigger),
+    ],
+)
+@mock.patch.object(NeptuneHook, "conn")
+def test_deferred_trigger_receives_region_name(mock_conn, operator_class, trigger_class):
```
ramitkataria: Maybe we should check for all the params instead of just region?

providers/amazon/tests/unit/amazon/aws/triggers/test_neptune.py
```diff
@@ -30,6 +30,22 @@
...
+
+@pytest.mark.parametrize(
+    "trigger_class",
+    [
+        NeptuneClusterAvailableTrigger,
+        NeptuneClusterStoppedTrigger,
+        NeptuneClusterInstancesAvailableTrigger,
+    ],
+)
+def test_region_name_reaches_the_hook_configuration(trigger_class):
```
ramitkataria: Here as well

SEPURI-SAI-KRISHNA:
Thanks @ramitkataria, done in both places.

The operator test now asserts the whole serialized trigger payload instead of picking out one field:

```python
assert deferred.value.trigger.serialize()[1] == {
    "db_cluster_id": CLUSTER_ID,
    "aws_conn_id": "aws_default",
    "region_name": REGION_NAME,
    "verify": VERIFY,
    "botocore_config": BOTOCORE_CONFIG,
    "waiter_delay": WAITER_DELAY,
    "waiter_max_attempts": WAITER_MAX_ATTEMPTS,
}
```

Exact-dict equality rather than a set of individual asserts, so an argument going missing on the way to the triggerer fails the suite, which is exactly the failure mode this PR is about. The same assertion is now in the trigger test across all three trigger classes.

One nuance the wider assertions surfaced, worth recording: of the three arguments the operators now forward, only `region_name` was actually being lost. `verify` and `botocore_config` are not named parameters on the Neptune trigger `__init__`s, so they already reached `AwsBaseWaiterTrigger` through `**kwargs`. `region_name` *is* named, so the subclass bound it and never passed it up. The widened assertions still fail without the source change, and they now also pin down the two that were only working by accident.

Verified locally: 31 passed with the change, and exactly the 5 new parametrizations fail with it reverted (26 pre-existing tests unaffected either way).

---
Drafted-by: Claude Code (Opus 5); reviewed by @SEPURI-SAI-KRISHNA before posting

ramitkataria (approved):

ramitkataria:
Thanks for the addition!

PR #71647, opened 2026-08-15
Keep OpenAPI error responses in sync with the statuses routes raise

Pushkal-Gupta:
Follow-up to #67570, #67571, #70992 and #71011, which each found and patched instances of
the same drift by hand: a route handler raises an HTTP status that
`create_openapi_http_exception_doc(...)` never declares, so the generated spec — and every
client built from it — has no model for a response the API really returns.

The helper's own docstring names the problem:

> There is no easy way to introspect the code and automatically see what HTTPException are
> actually raised by the endpoint implementation. This piece of documentation needs to be
> kept in sync with the endpoint code manually.

This adds the check that makes it automatic, so the next divergence fails in CI instead of
shipping. Running it over `api_fastapi/**/routes/` found 15 statuses still undeclared after
the four manual passes:

| Endpoint | Undeclared | Reached when |
| --- | --- | --- |
| `POST /connections/test` | `400` | the host/port to test differs from the stored connection and no password is supplied |
| `GET /dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances/{task_id}/logs/{try_number}` | `400` | the configured task log handler does not support reading |
| `POST /variables` | `404` | the variable is not readable back after being written |
| `GET /ui/next_run_assets/{dag_id}` | `404` | the Dag does not exist |
| `GET /ui/partitioned_dag_runs` | `404` | filtered by a `dag_id` that does not exist |
| `GET /ui/pending_partitioned_dag_run/{dag_id}` | `404` | no pending partitioned run for that Dag/partition key |
| `GET /ui/teams` | `403` | multi-team mode is not configured |
| `GET /execution/asset-events/by-asset` | `400` | neither `name` nor `uri` is supplied |
| `GET /execution/store/asset/by-name/value` | `404` | the asset is unknown, or has no value for the key |
| `GET /execution/store/asset/by-uri/value` | `404` | the asset is unknown, or has no value for the key |
| `PATCH /execution/hitlDetails/{task_instance_id}` | `409` | a response has already been received |
| `PATCH /execution/task-instances/{task_instance_id}/run` | `500` | the serialized TaskFlow arg spec for a stub task fails validation |
| `GET /execution/store/ti/{task_instance_id}/{key}` | `404` | the task instance is unknown, or has no value for the key |
| `HEAD /execution/xcoms/{dag_id}/{run_id}/{task_id}/{key}` | `400` | `map_index` is passed to a HEAD request |
| `POST /execution/xcoms/{dag_id}/{run_id}/{task_id}/{key}` | `400` | the key is empty, the value is too large to map, or is unserializable |

The regenerated spec and UI client show the effect: seven public/UI endpoints gain error
models their clients previously had no type for.

### About the check

It is deliberately conservative — it is meant to be trustworthy enough to gate CI, so it
under-reports rather than over-reports, and stays silent whenever it cannot see the whole
picture:

- Only `HTTPException(...)` raised in the handler's own body counts. Statuses raised by a
  shared dependency or a service helper are not required to be declared.
- `422` is never required (FastAPI documents validation errors itself), and neither are
  `401`/`403` — routers contribute those wholesale via their auth dependencies, and the
  router that does so is often built in another module (`routes/public/__init__.py`
  declares both for every public route).
- A route inherits whatever its own router declares, so the `responses=` on
  `APIRouter(...)` (as in `execution_api/routes/xcoms.py`) counts as declared.
- Statuses that are not resolvable constants, and `responses=` blocks that are not a
  literal `create_openapi_http_exception_doc([...])` call or mapping, are skipped rather
  than guessed at.

The `GET /ui/teams` `403` above sits outside what the check enforces, for the `401`/`403`
reason just described; it is included because it is a real gap in `_private_ui.yaml` that
the same audit turned up.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
@@ -38,7 +38,6 @@ from sqlalchemy import select
 
 from airflow._shared.state import AssetScope, AssetStateStoreWriterKind
 from airflow.api_fastapi.common.db.common import SessionDep
-from airflow.api_fastapi.core_api.openapi.exceptions import create_openapi_http_exception_doc
 from airflow.api_fastapi.execution_api.datamodels.asset_state_store import (
     AssetStateStorePutBody,
     AssetStateStoreResponse,
@@ -107,12 +106,7 @@ def _resolve_asset_id_by_uri(uri: str, session: SessionDep) -> int:
     return asset_id
 
 
-@router.get(
-    "/by-name/value",
-    responses=create_openapi_http_exception_doc(
-        [(status.HTTP_404_NOT_FOUND, "Asset not found, or it has no value for the key")]
-    ),
-)
+@router.get("/by-name/value")
 def get_asset_state_store_by_name(
     name: Annotated[str, Query(min_length=1)],
     key: Annotated[str, Query(min_length=1)],
@@ -201,12 +195,7 @@ def clear_asset_state_store_by_name(
     get_state_backend().clear(AssetScope(asset_id=asset_id), session=session)
 
 
-@router.get(
-    "/by-uri/value",
-    responses=create_openapi_http_exception_doc(
-        [(status.HTTP_404_NOT_FOUND, "Asset not found, or it has no value for the key")]
-    ),
-)
+@router.get("/by-uri/value")
 def get_asset_state_store_by_uri(
     uri: Annotated[str, Query(min_length=1)],
     key: Annotated[str, Query(min_length=1)],
--- a/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
+++ b/airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
@@ -38,6 +38,7 @@ from sqlalchemy import select
 
 from airflow._shared.state import AssetScope, AssetStateStoreWriterKind
 from airflow.api_fastapi.common.db.common import SessionDep
+from airflow.api_fastapi.core_api.openapi.exceptions import create_openapi_http_exception_doc
 from airflow.api_fastapi.execution_api.datamodels.asset_state_store import (
     AssetStateStorePutBody,
     AssetStateStoreResponse,
@@ -106,7 +107,12 @@ def _resolve_asset_id_by_uri(uri: str, session: SessionDep) -> int:
     return asset_id
 
 
-@router.get("/by-name/value")
+@router.get(
+    "/by-name/value",
+    responses=create_openapi_http_exception_doc(
+        [(status.HTTP_404_NOT_FOUND, "Asset not found, or it has no value for the key")]
+    ),
+)
 def get_asset_state_store_by_name(
     name: Annotated[str, Query(min_length=1)],
     key: Annotated[str, Query(min_length=1)],
@@ -195,7 +201,12 @@ def clear_asset_state_store_by_name(
     get_state_backend().clear(AssetScope(asset_id=asset_id), session=session)
 
 
-@router.get("/by-uri/value")
+@router.get(
+    "/by-uri/value",
+    responses=create_openapi_http_exception_doc(
+        [(status.HTTP_404_NOT_FOUND, "Asset not found, or it has no value for the key")]
+    ),
+)
 def get_asset_state_store_by_uri(
     uri: Annotated[str, Query(min_length=1)],
     key: Annotated[str, Query(min_length=1)],
--- a/scripts/ci/prek/check_openapi_exception_doc_in_sync.py
+++ b/scripts/ci/prek/check_openapi_exception_doc_in_sync.py
@@ -1,0 +1,254 @@
+#!/usr/bin/env python
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Check API route handlers declare every HTTP status they raise.
+
+``create_openapi_http_exception_doc(...)`` feeds the ``responses=`` block of a
+route, which is what the generated OpenAPI spec — and every client built from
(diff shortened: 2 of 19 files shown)


scripts/ci/prek/check_openapi_exception_doc_in_sync.py
```diff
@@ -0,0 +1,254 @@
...
+dependency or a service helper are not required to be declared. The hook
+therefore under-reports rather than over-reports.
+"""
+
+# /// script
+# requires-python = ">=3.10,<3.11"
+# dependencies = [
+#   "rich>=13.6.0",
+# ]
+# ///
```
pierrejeambrun, line 40: To remove?

airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
```diff
@@ -106,7 +107,12 @@ def _resolve_asset_id_by_uri(uri: str, session: SessionDep) -> int:
     return asset_id
 
 
-@router.get("/by-name/value")
+@router.get(
+    "/by-name/value",
+    responses=create_openapi_http_exception_doc(
+        [(status.HTTP_404_NOT_FOUND, "Asset not found, or it has no value for the key")]
+    ),
+)
```
pierrejeambrun: execution_api routes are built on cadwyn's VersionedAPIRouter, which isn't in ROUTER_CLASSES, so the check can't see router-level responses= there. That's why it reported the task_state_store /  asset_state_store 404s as undeclared even though both routers already declare HTTP_404_NOT_FOUND (task_state_store.py:43, asset_state_store.py:85) — the local additions here are actually redundant with the router. Could we add it?

pierrejeambrun (commented):
LGTM, a couple of suggestions before we can merge.

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
jason810496: IMO, we need to keep PEP 723 inline dependencies, almost all the prek hook follow this convention.

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
```diff
@@ -0,0 +1,254 @@
...
+* A status that is not a resolvable constant (``status.HTTP_404_NOT_FOUND``,
+  a bare ``HTTP_404_NOT_FOUND``, or a literal ``404``) is skipped.
+* A handler whose ``responses=`` is not a ``create_openapi_http_exception_doc``
+  call over a literal list is skipped entirely, as are handlers where any
+  declared entry cannot be resolved.
+
+Because only the handler's own body is inspected, statuses raised by a shared
+dependency or a service helper are not required to be declared. The hook
+therefore under-reports rather than over-reports.
+"""
```
jason810496, line 33: Let's make the docstring more concise.

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+        if console:
+            console.print()
+            console.print(f"[red]{message}[/red]")
+        else:
+            print()
+            print(message)
+        return 1
+    return 0
+
```
jason810496, line 236: It think we should leverge LibCST to support the nested fastapi.HTTPException resolution.
The current implementation can't handle the following example: 
```
route.handler -> common.validate -> common.not_found raises HTTPException(404)
```

More context in:

LibCST — not a complete call-graph package, but probably the best building block here. Its FullyQualifiedNameProvider resolves aliases and relative imports across a repository, and Airflow already depends on LibCST for scripts. LibCST name resolution
    (https://libcst.readthedocs.io/en/latest/_modules/libcst/metadata/name_provider.html#FullyQualifiedNameProvider)

jason810496 (commented):
Thanks for the improvement.

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
pierrejeambrun: You're right, I wasn't aware of this PEP, thanks

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
Pushkal-Gupta: Agreed this is the real limitation, and the LibCST pointer is the right one —
`FullyQualifiedNameProvider` is the piece that makes cross-module resolution
tractable, and the dependency is already there.

I'd like to propose it as a follow-up PR rather than folding it in here, for two
reasons:

1. **Scope.** It changes the check from a single-file AST pass to inter-procedural
   analysis across the repo, which brings a call graph, cycle handling, a depth
   bound, and a per-file cost that has to stay inside the pre-commit budget. That
   deserves its own review rather than riding along with a set of declaration
   fixes.
2. **It interacts with the router fix above.** Now that inheritance works on
   `VersionedAPIRouter`, following calls will move the finding set in both
   directions — new true positives from helpers like `_resolve_asset_id_by_name`,
   but also statuses that turn out to be already covered by a router. Landing the
   corrected baseline first makes that delta measurable.

The current behaviour is documented as under-reporting rather than over-reporting,
so this is strictly better than the status quo today and does not block the deeper
version.

Happy to open the follow-up issue and take it, if that split works for you.

scripts/ci/prek/check_openapi_exception_doc_in_sync.py
Pushkal-Gupta: Trimmed from 30 lines to 15. Kept the what/why and a one-paragraph summary of the conservatism rules; dropped the per-rule bullet list, which is better placed in the PR description than re-read on every file open.

airflow-core/src/airflow/api_fastapi/execution_api/routes/asset_state_store.py
Pushkal-Gupta: Good catch, and you're right — that's a real defect in the check, not just in this file. `ROUTER_CLASSES` was `{"APIRouter", "AirflowRouter"}`, so router inheritance silently did not apply to any Cadwyn-based execution API route.

Added `VersionedAPIRouter` and re-audited all six execution API route modules:

| Router | Declares | Was added | Now |
|---|---|---|---|
| `asset_state_store` (Versioned) | 401, **404** | 404 ×2 | **reverted** |
| `task_state_store` (Versioned) | 401, 403, **404** | 404 | **reverted** |
| `hitl` (Versioned) | *(none)* | 409 | kept |
| `task_instances` (Versioned) | *(none)* | 500 | kept |
| `asset_events` (plain) | 401, 404 | 400 | kept |
| `xcoms` (plain) | 401, 403, 404 | 400 ×2 | kept |

So three of the fifteen were false positives. `asset_state_store.py` and `task_state_store.py` are now byte-identical to `main`. The other twelve stand, and no public-spec change is affected — all three reverts were execution API, and regenerating both specs produces no diff.

Re-running the corrected check against the pre-PR tree now reports 11 violations rather than 15; with the `/ui/teams` 403 that the audit found but the check exempts, that is the 12 declarations this PR keeps.

Added a regression test (`cadwyn-versioned-router-declares-the-status`) so this cannot silently regress.

PR #71648, opened 2026-08-15
Apply impersonation_chain to deferred BigQuery existence checks

SEPURI-SAI-KRISHNA:
`BigQueryTableExistenceSensor` and `BigQueryTablePartitionExistenceSensor` hand their
impersonation chain to the trigger inside `hook_params`:

```python
trigger=BigQueryTableExistenceTrigger(
    ...,
    hook_params={"impersonation_chain": self.impersonation_chain},
)
```

but the trigger authenticates from a different attribute entirely:

```python
def _get_async_hook(self) -> BigQueryTableAsyncHook:
    return BigQueryTableAsyncHook(
        gcp_conn_id=self.gcp_conn_id, impersonation_chain=self.impersonation_chain
    )
```

Nothing ever sets `self.impersonation_chain` from those call sites, so it stays `None` and
the deferred existence check runs as the connection's service account. `hook_params` is
stored and serialized but never read anywhere in the module — it carries the value and
then drops it.

`poke()` builds its hook correctly, so the sensor honours impersonation when
`deferrable=False` and ignores it when `deferrable=True`. That divergence is the awkward
part: turning deferrable off appears to "fix" the permissions error, which points
investigation away from the real cause.

The trigger already grew a proper `impersonation_chain` parameter in #36341; the sensors
were simply never moved onto it. This change passes it explicitly at both defer sites.

`hook_params` is left untouched here. It is a required argument on the trigger's public
`__init__`, and now that nothing reads it, removing or deprecating it is a separate
decision — happy to follow up if maintainers would like it gone.

For reviewers: `GoogleBaseHook.__init__` falls back to the connection's
`impersonation_chain` extra when the argument is absent, so deployments that configure
impersonation on the connection were never affected. This only bites when impersonation is
set on the sensor, which is the documented parameter.

---

### DIFF
--- a/providers/google/tests/unit/google/cloud/sensors/test_bigquery.py
+++ b/providers/google/tests/unit/google/cloud/sensors/test_bigquery.py
@@ -43,39 +43,6 @@ TEST_PARTITION_ID = "20200101"
 TEST_IMPERSONATION_CHAIN = ["ACCOUNT_1", "ACCOUNT_2", "ACCOUNT_3"]
 
 
-@pytest.mark.parametrize(
-    ("sensor_class", "trigger_class", "extra_kwargs"),
-    [
-        (BigQueryTableExistenceSensor, BigQueryTableExistenceTrigger, {}),
-        (
-            BigQueryTablePartitionExistenceSensor,
-            BigQueryTablePartitionExistenceTrigger,
-            {"partition_id": TEST_PARTITION_ID},
-        ),
-    ],
-)
-@mock.patch("airflow.providers.google.cloud.sensors.bigquery.BigQueryHook")
-def test_deferred_trigger_receives_impersonation_chain(mock_hook, sensor_class, trigger_class, extra_kwargs):
-    task = sensor_class(
-        task_id="check_existence",
-        project_id=TEST_PROJECT_ID,
-        dataset_id=TEST_DATASET_ID,
-        table_id=TEST_TABLE_ID,
-        gcp_conn_id=TEST_GCP_CONN_ID,
-        impersonation_chain=TEST_IMPERSONATION_CHAIN,
-        deferrable=True,
-        **extra_kwargs,
-    )
-    mock_hook.return_value.table_exists.return_value = False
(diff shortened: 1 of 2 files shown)


providers/google/tests/unit/google/cloud/sensors/test_bigquery.py
```diff
@@ -43,6 +43,39 @@
...
+        (BigQueryTableExistenceSensor, BigQueryTableExistenceTrigger, {}),
+        (
+            BigQueryTablePartitionExistenceSensor,
+            BigQueryTablePartitionExistenceTrigger,
+            {"partition_id": TEST_PARTITION_ID},
+        ),
+    ],
+)
+@mock.patch("airflow.providers.google.cloud.sensors.bigquery.BigQueryHook")
+def test_deferred_trigger_receives_impersonation_chain(mock_hook, sensor_class, trigger_class, extra_kwargs):
```
SameerMesiah97: Could we move this test under the respective sensor/trigger test classes rather than having it as a standalone parametrized test?

SEPURI-SAI-KRISHNA:
Thanks @SameerMesiah97, moved.

The standalone parametrized test is gone. Each sensor now has its own `test_deferred_trigger_receives_impersonation_chain` sitting next to that class's existing deferral test — in `TestBigqueryTableExistenceSensor` and in `TestBigqueryTablePartitionExistenceSensor`.

I dropped the `isinstance(exc.value.trigger, ...)` assertion during the move: the sibling `test_execute_deferred` / `test_execute_with_deferrable_mode` in each class already covers the trigger type, so the new tests assert only the argument that was being lost.

Verified locally: 25 passed with the change, and exactly the 2 new tests fail with it reverted (23 pre-existing tests unaffected either way).

---
Drafted-by: Claude Code (Opus 5); reviewed by @SEPURI-SAI-KRISHNA before posting

SEPURI-SAI-KRISHNA:
Gentle ping, the review feedback here was addressed on 18 August and CI has been green since. Anything else you'd like changed before this can go in?

Issue #59625, opened 2025-12-19
Switch to hardened Docker images in Dockerfiles


shahar1:
### Description

Following the fresh new [statement](https://www.docker.com/blog/docker-hardened-images-for-every-developer/) from Docker, hardened container images are now free and open source - which **might** mean that we could eventually utilize it in Airflow to enhance security.

The main challenge is that it requires authentication, which could be problematic legally & practically for local builds. If we get to overcome this one, implementation will be a great security improvement.

### Use case/motivation

Improve container's security

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

alessandro-peyrachia-rnc:
Hi,
coming in from the Slack channel.
This comes with an exceptional timing for us.
We are very new to airflow, we literally started 1 month ago, and because of the market we operate in, we need secured/hardened images.
I was starting to understand how to do it properly, then I saw your message.
We are on airflow 3 in AKS, I'm building experience on it, please count me in to help out on testing.

potiuk:
will do

suii2210:
Hi! I’m interested in working on this. For hardened images, should we aim for XYZ base image? Are there specific guidelines we must follow?

shahar1:
> Hi! I’m interested in working on this. For hardened images, should we aim for XYZ base image? Are there specific guidelines we must follow?

Thank you for your will to help in this area! I assigned you to this task.
Basically, it should be simple as switching the base image in the Docker file to the [hardened version](https://hub.docker.com/hardened-images/catalog/dhi/debian-base) - see instructions in the link.
Building this image requires logging it to Dockerhub using a Personal Access Token (PAT) before the build step. In the GitHub Actions, the step should be:

```yaml
      - name: Login to Docker Hub
        uses: docker/login-action@v3
        with:
          registry: dhi.io
          username: ${{ vars.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PAT }}
```

I tested the above in another repository and it works - but it will be problematic for now to configure it in Airflow's CI/CD, because I cannot use my personal user for the credentials (and AFAIK, we currently don't have a user dedicated for ASF organization).
Therefore, you could start by testing locally and see that nothing breaks in their image, and if it all works - create a documentation for instructing others how to make it work.
[...]

PR #71650, opened 2026-08-15
Switch official Docker images to Debian Trixie

### LINKED ISSUE #71494: Upgrading docker images debian from version 12(bookworm) to 13(trixie).
### Description

Bookworm regular support ended 11 Jul 2026; Airflow’s own policy is to switch ~6 months before that — so this is overdue by policy, even though Bookworm LTS runs to 2028.
Trixie has been stable since Aug 2025 (~1 year).

### Use case/motivation

I noticed 3 critical vulnerabilities related to the unbound package: CVE-2026-42960, CVE-2026-33278, CVE-2026-50252 that for bookworm have no upgrade. Would be much easier to jump forward with the version instead.

### Related issues

_No response_

### Code of Conduct

ytoprakc:
Switch official Airflow Docker images from Debian Bookworm to Debian Trixie.

closes: #71494

### Why

- Airflow’s Docker image policy says the Debian base should be upgraded at or before 6 months before the previous release reaches end of regular support.
- Debian Bookworm regular support ended on July 11, 2026.
- Debian Trixie has been stable since August 2025 and is now the appropriate Debian baseline for newly built Airflow images.
- Moving to Trixie keeps Airflow aligned with its documented support policy and Debian’s current package/security stream.

### What Changed

- Updated the default Docker base image:

  - `debian:bookworm-slim` -> `debian:trixie-slim`
  - applies to both production and CI image builds
  - deprecated Bookworm-based custom image builds remain available during the Airflow 3.4.x transition period

- Updated build tooling and CI defaults:

  - Breeze now defaults to Trixie
  - GitHub Actions image-build inputs now default to Trixie, while still allowing Bookworm explicitly
  - release-management image build paths now use Trixie
  - generated Breeze command-output docs were refreshed

- Updated Debian package mappings for Trixie:

  - `libldap-2.5-0` -> `libldap2`
  - `libssl3` -> `libssl3t64`
  - `lzma-dev` is kept only for Bookworm builds
  - moved `lcov` installation until after Airflow’s custom Python build

- Made system-Python detection version-independent:

  - previously the diagnostic path assumed Bookworm’s system Python package naming
  - now it detects any installed `libpython*` package
  - this keeps the guard useful across Debian releases and Python versions

- Updated MariaDB client installation:

  - Trixie builds use MariaDB `11.8`
  - Bookworm builds continue to use MariaDB `10.11`
  - added `mariadb-client-compat` where needed for MariaDB 11.x

- Updated MSSQL client installation:

  - refreshed Microsoft apt signing key
  - keeps `msodbcsql18` installation working with the Trixie base

- Fixed CI SSH startup logic:

  - previously waited for exactly 3 SSH host keys
  - now waits for any available SSH host key
  - avoids Trixie/OpenSSH-related CI timeouts where SSH was usable but exposed a different number of host key types

- Updated Java example/e2e images:

  - use Temurin 17 JRE instead of Debian `openjdk-17-jre-headless`
  - needed because Trixie no longer ships that package
  - keeps Spark examples on Java 17 rather than moving them to Java 21, which Spark 3.5.x does not support

- Updated Amazon provider optional `python3-saml` dependency floor:

  - `xmlsec>=1.3.14` -> `xmlsec>=1.3.16`
  - Airflow’s lowest-dependencies CI selected `xmlsec 1.3.14`, and that version fails to build from source with Debian Trixie’s GCC 14

- Updated IBM DB2 provider dependency floor:

  - `ibm-db>=3.2.0` -> `ibm-db>=3.2.8`
  - Airflow’s all-version lowest-dependencies CI selected `ibm-db 3.2.0`, which fails to build from source on newer Python/Trixie CI images

- Updated documentation:

  - Docker stack docs
  - installation/support docs
  - security model platform reference
  - Docker Compose docs
  - Docker customization examples
  - PyPI README
  - spelling wordlist
  - release newsfragment

---

- [x] Yes — OpenAI Codex


airflow-e2e-tests/docker/Dockerfile.java
```diff
@@ -19,12 +19,15 @@
 # can spawn JVM subprocesses for @task.stub tasks.
 #
 # Pin Java 17 (rather than default-jre-headless): the Scala Spark example runs
-# Apache Spark 3.5.x, which supports Java 8/11/17 but not Java 21.
+# Apache Spark 3.5.x, which supports Java 8/11/17 but not Java 21. Debian Trixie
```
raphaelauv, line 22: since version 6 of the spark provider ( https://airflow.apache.org/docs/apache-airflow-providers-apache-spark/stable/changelog.html#breaking-changes ) 

the Apache spark version minimum is now 4.0.0 ( jdk 21 compatible )

and apache spark 4.2.0 is jdk25 compatible -> `eclipse-temurin:25-jre-resolute`

ytoprakc: Thanks for pointing this out @raphaelauv.

I agree with the general direction for the Spark provider. Since `apache-airflow-providers-apache-spark` 6.x now uses Spark dependencies `>=4.0.0`, Java 21 is a valid baseline for Spark 4.x.

For this specific file, I think there is a small distinction worth keeping. `airflow-e2e-tests/docker/Dockerfile.java` is used by the Java SDK e2e tests, not directly by the Spark provider package. That path currently includes `java-sdk/scala_spark_example`, which still pins `org.apache.spark:spark-sql_2.13:3.5.9` directly and has its runtime/module-opening setup written around Spark 3.5.x.

So for this PR, I kept that Java SDK e2e image aligned with the existing Spark 3.5.x example and only changed how Java 17 is supplied on Trixie. Moving the Java SDK Scala Spark example to Spark 4.x / Java 21 sounds like a good follow-up, but I think it deserves its own PR because it would touch the Gradle dependency, Java toolchain, e2e build image, JVM module options, docs, and e2e validation together. Spark 4.2 adds Java 25 compatibility, but provider 6.x only gives us a Spark `>=4.0.0` baseline, so Java 25 feels ahead of the follow-up. 

This PR is already fairly large for a base image migration, so I would rather keep this part scoped to making the existing Java 17-based e2e setup work on Trixie instead of also changing the Spark version and Java SDK test baseline.

potiuk (changes requested):
Updated the labels to make sure all tests and all versions of python are run. However there is one more important change  needed - if you look at past changes to docker base image we supported both - previous and new verison of theb base debian - for until next minor version (see the changelog) - this should be done similarly

ytoprakc:
Thanks @potiuk. I updated the PR to follow the same transition pattern used in the earlier Debian base-image migrations.

Trixie is still the default for newly built CI and production images, but Bookworm remains explicitly selectable for custom image builds during the Airflow 3.4.x line. I also updated the Docker image changelog to call this out and to make clear that Bookworm-based custom image build support is expected to be removed in Airflow 3.5.0.

potiuk:
Thank you for this — and I want to be clear up front that closing it is not a judgement on the work. I went through it properly before deciding, and it does what I asked for in my last review: the dual-base transition follows the shape of the previous Debian migrations, `ALLOWED_DEBIAN_VERSIONS` keeps Bookworm selectable, the changelog documents both the transition and the 3.5.0 removal, and there are tests covering both bases. I also verified the Microsoft signing-key rotation against the vendor rather than taking it on trust — the key you added matches `https://packages.microsoft.com/keys/microsoft-2025.asc` exactly (`AA86F75E427A19DD33346403EE4D7792F748182B`), and Microsoft's own key README gives the reason, which is precisely our reason: the old `BE1229CF` key carries SHA1 signatures that newer Linux releases no longer accept. That was the highest-consequence line in the diff and it was right.

### Why I'm closing it anyway

The base-image story has moved underneath this PR. We are switching the CI and PROD images to **Docker Hardened Images** first, and only then revisiting the Debian release. That work is in https://github.com/apache/airflow/pull/73038, with the background in https://github.com/apache/airflow/issues/59625.

That changes the ground this PR is standing on. Once the images are based on a hardened Python image, we no longer compile CPython on top of `debian:bookworm-slim` at all — `install_python()`, the tarball download and the sigstore/GPG verification all go away, along with the fifteen apt packages that existed only to build CPython. Several of those are exactly the packages this PR is carefully making conditional per Debian release. So the `bookworm`/`trixie` branching added here to `Dockerfile`, `Dockerfile.ci` and `install_os_dependencies.sh` would have to be rewritten on top of a different base rather than merged and then adapted, and the two changes would conflict heavily in the meantime.

Doing hardened first and Debian second is the order that costs the least total work — but it does mean this PR would be largely redone rather than rebased, which is not a fair thing to ask of you.

### What is worth keeping, and is independent of all of the above

Several pieces here are not coupled to the base-image question at all and are useful on their own. If you would like to carry any of them forward as separate, much smaller PRs, they would be very welcome:

- **The Microsoft signing-key rotation to the 2025 key.** Needed regardless of which base we end up on, for the SHA1 reason above. One request if you do: add a comment next to the key ID in `scripts/ci/prek/update_docker_gpg_keys.py` citing `https://packages.microsoft.com/keys/microsoft-2025.asc` and the full fingerprint. That hook fetches keys **by ID from public keyservers**, which accept uploads from anyone with any UID — so the hardcoded ID is the actual trust root and cannot validate itself. Recording where it came from makes it re-verifiable in seconds.
- **The two provider dependency floors** — `xmlsec>=1.3.16` in `providers/amazon` and `ibm-db>=3.2.8` in `providers/ibm/db2`. Both are justified on their own terms and have nothing to do with the Debian base. Worth noting if you split them out: provider changelogs are regenerated from `git log` at release time rather than from newsfragments, so name the bumps in the commit message or they will be missed during release prep.
- **The MariaDB 10.11 → 11.8 selection**, which will be needed by whichever base ships a newer Debian.
- **The `openjdk-17` fix in the extending example.** One question there rather than a request: Trixie ships `openjdk-21-jre-headless`, so is `eclipse-temurin` needed because the pinned Spark provider requires 17 specifically? If 21 works, staying on the distro package keeps our canonical example Debian-only; if 17 is genuinely required, a line in the example saying so would help, since readers copy these verbatim.

Sorry to close a PR you have already revised once in response to my review. If you would like to pick the Debian work back up after the hardened images land, please do — and feel free to follow https://github.com/apache/airflow/pull/73038 to see when that is.

---

PR #71652, opened 2026-08-15
Add deferrable mode to EmrServerlessJobSensor

saadbelgi:
Add deferrable mode to `EmrServerlessJobSensor` so waiting on a long-running EMR Serverless job does not occupy a worker slot.

Follows `EmrJobFlowSensor` / `EmrStepSensor`: poke once, then `defer()` to a dedicated `EmrServerlessJobSensorTrigger`. The trigger uses the existing `serverless_job_completed` waiter with acceptors overridden to preserve the sensor's configured `target_states`. Poke/reschedule behaviour is unchanged when `deferrable=False`.

Docs: class docstring plus `providers/amazon/docs/operators/emr/emr_serverless.rst`.

### Implementation notes

- **Waiter configuration overrides:** The trigger reuses the existing `serverless_job_completed` waiter and overrides only its acceptors. A static waiter cannot represent runtime-configurable `target_states`, so this avoids adding separate waiter JSON for each possible target-state combination.
- **Failure exceptions:** `poke()` still raises `AirflowException` on `JOB_FAILURE_STATES`, as before. `execute_complete()` raises `RuntimeError` because new direct `raise AirflowException` usages are blocked by the `check-no-new-airflow-exceptions` pre-commit hook. This also follows the existing DynamoDB and SSM sensor patterns.
- **Dedicated trigger:** `EmrServerlessStartJobTrigger` remains unchanged for `EmrServerlessStartJobOperator`. The new `EmrServerlessJobSensorTrigger` is sensor-specific because it must support runtime-configurable `target_states` and sensor timeout semantics.

---

- [X] Yes — Cursor Grok 4.5

Generated-by: Cursor Grok 4.5 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### Tests run locally

- **prek (pre-commit):** passed, including ruff, ruff-format, and `check-no-new-airflow-exceptions` on the changed files.
- **mypy:** `breeze run mypy providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py providers/amazon/src/airflow/providers/amazon/aws/triggers/emr.py` — passed
- **Sensor + EMR trigger unit tests:** `breeze testing providers-tests --python 3.10 providers/amazon/tests/unit/amazon/aws/sensors/test_emr_serverless_job.py providers/amazon/tests/unit/amazon/aws/triggers/test_emr.py providers/amazon/tests/unit/amazon/aws/triggers/test_serialization.py -xvs` — passed
- **Amazon provider suite:** `breeze testing providers-tests --python 3.10 --test-type "Providers[amazon]"` — passed

---

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
@@ -26,7 +26,7 @@ from airflow.providers.amazon.aws.links.emr import EmrClusterLink, EmrLogsLink, 
 from airflow.providers.amazon.aws.sensors.base_aws import AwsBaseSensor
 from airflow.providers.amazon.aws.triggers.emr import (
     EmrContainerTrigger,
-    EmrServerlessStartJobTrigger,
+    EmrServerlessJobSensorTrigger,
     EmrStepSensorTrigger,
     EmrTerminateJobFlowTrigger,
 )
@@ -116,7 +116,7 @@ class EmrBaseSensor(AwsBaseSensor[EmrHook]):
 
 class EmrServerlessJobSensor(AwsBaseSensor[EmrServerlessHook]):
     """
-    Poll the state of the job run until it reaches a terminal state; fails if the job run fails.
+    Poll the state of the job run until it reaches one of the target states; fails if the job run fails.
 
     .. seealso::
         For more information on how to use this sensor, take a look at the guide:
@@ -124,10 +124,7 @@ class EmrServerlessJobSensor(AwsBaseSensor[EmrServerlessHook]):
 
     :param application_id: application_id to check the state of
     :param job_run_id: job_run_id to check the state of
-    :param target_states: a set of states to wait for, defaults to ``SUCCESS``. In deferrable mode
-        the waiter waits for a terminal success state regardless of ``target_states``.
-    :param max_attempts: Maximum waiter attempts in deferrable mode. Ignored when
-        ``deferrable=False``; poke mode uses ``timeout`` (default ``sensors.default_timeout``, 7 days).
+    :param target_states: a set of states to wait for, defaults to ``SUCCESS``.
     :param deferrable: Run sensor in the deferrable mode.
     :param aws_conn_id: The Airflow connection used for AWS credentials.
         If this is ``None`` or empty then the default boto3 behaviour is used. If
@@ -151,14 +148,12 @@ class EmrServerlessJobSensor(AwsBaseSensor[EmrServerlessHook]):
         application_id: str,
         job_run_id: str,
         target_states: set | frozenset = frozenset(EmrServerlessHook.JOB_SUCCESS_STATES),
-        max_attempts: int = 60,
         deferrable: bool = conf.getboolean("operators", "default_deferrable", fallback=False),
         **kwargs: Any,
     ) -> None:
         self.target_states = target_states
         self.application_id = application_id
         self.job_run_id = job_run_id
-        self.max_attempts = max_attempts
         self.deferrable = deferrable
         super().__init__(**kwargs)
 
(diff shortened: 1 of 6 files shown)


providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
```diff
@@ -123,7 +124,11 @@ class EmrServerlessJobSensor(AwsBaseSensor[EmrServerlessHook]):
 
     :param application_id: application_id to check the state of
     :param job_run_id: job_run_id to check the state of
-    :param target_states: a set of states to wait for, defaults to 'SUCCESS'
+    :param target_states: a set of states to wait for, defaults to ``SUCCESS``. In deferrable mode
+        the waiter waits for a terminal success state regardless of ``target_states``.
```
SameerMesiah97: Why are the termination semantics of deferrable mode not symmetrical to non-deferrable? As implemented, a sensor configured with e.g. `target_states={"RUNNING"}` can succeed at `RUNNING` in synchronous mode but wait until job completion in deferrable mode. Can we preserve `target_states` semantics in the trigger?

providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
```diff
@@ -123,7 +124,11 @@ class EmrServerlessJobSensor(AwsBaseSensor[EmrServerlessHook]):
...
+    :param max_attempts: Maximum waiter attempts in deferrable mode. Ignored when
+        ``deferrable=False``; poke mode uses ``timeout`` (default ``sensors.default_timeout``, 7 days).
+    :param deferrable: Run sensor in the deferrable mode.
```
SameerMesiah97, line 128: I am not sure about `max_attempts` only applying when `deferrable=True`. This gives the sensor different timeout semantics depending on execution mode: non-deferrable execution is bounded by the sensor timeout, while deferrable execution is bounded by `max_attempts` . Could we use the same timeout semantics for both modes rather than introducing a deferrable-specific limit?

SameerMesiah97 (commented):
I think there are 2 main concerns that should be addressed. I have left comments.

providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
saadbelgi: I have updated the sensor to use the same sensor timeout in deferrable and non-deferrable modes instead of having separate max_attempts.

providers/amazon/src/airflow/providers/amazon/aws/sensors/emr.py
saadbelgi: This was done because I was trying to reuse `EmrServerlessStartJobTrigger` (used in `EmrServerlessStartJobOperator` which doesn't accept `target_states` and only waits for completion in deferrable mode).    


I have created a new trigger class - `EmrServerlessJobSensorTrigger`- to replicate this sensor's behaviour in non-deferrable mode. Now, it will fail if it's a failure state (even if it's in target states), and succeed if any state in the target state is reached. This behaviour is similar to the one implemented in `poke` method of this sensor.     


One callout is that `EmrServerlessJobSensorTrigger` inherits `AwsBaseWaiterTrigger` and uses the same waiter as `EmrServerlessStartJobTrigger`, but overrides the failure and success states dynamically based on the input target states. This is also done in `airflow.providers.amazon.aws.triggers.mwaa.MwaaDagRunCompletedTrigger` and `airflow.providers.amazon.aws.triggers.mwaa.MwaaTaskCompletedTrigger`. This is done here because I could not find a way to use static waiters defined in json file because target_states input is a set. If this is to be avoided, I would have to avoid inheriting from `AwsBaseWaiterTrigger`.

SameerMesiah97 (approved):
Approved pending green CI. But this needs one of the AWS maintainers to look at it before merge.

o-nikolas (approved):

PR #69042, opened 2026-06-26
Fix SageMakerTransformOperator succeeding on a failed deferred job

steveahnahn:
## Problem

`SageMakerTransformOperator` with `deferrable=True` (which is the default whenever a deployment sets `[operators] default_deferrable=True`) can mark a task **successful even though the SageMaker batch-transform job failed**. The failed job's description is pushed as the XCom result, so downstream tasks run against missing/invalid transform output with no error surfaced.

## Root cause

The operator defers while the job is still `InProgress`, so the job can fail *during* the deferred wait. The trigger is `SageMakerTrigger`, which since #68927 inherits `AwsBaseWaiterTrigger.run()`. On waiter failure that base **yields** `TriggerEvent({"status": "error", ...})` instead of raising — so `execute_complete` is resumed with an error event. `SageMakerTransformOperator.execute_complete` never checked the status: it logged `"SageMaker job ... completed."` and returned `serialize_result(...)`, so the task succeeded.

Every other SageMaker operator's `execute_complete` already guards this (`if status != "success": raise`). Transform was the only one missing it. Before #68927 the previous trigger *raised* on failure, so the missing guard was dormant; migrating to the yielding base class turned it into a live silent-success.

## Fix

Reject a non-success status in `SageMakerTransformOperator.execute_complete`, matching the sibling operators, and add a regression test (a deferred transform that fails during the wait now fails the task instead of reporting success).

**Note on the exception type:** the sibling operators raise `AirflowException`, but the `check-no-new-airflow-exceptions` hook forbids *new* `raise AirflowException` usages, so this raises `RuntimeError` (identical task-failure/retry semantics, and already used elsewhere in this provider, e.g. `dms.py`).

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #48946, opened 2025-04-08
Handle NoCredentialsError in waiter_with_logging.py

nrobinson-intelycare:
Fixes https://github.com/apache/airflow/issues/45622

PR #71653, opened 2026-08-15
Preserve SageMaker job lifecycle states in trigger events

SameerMesiah97:
**Description**

This change updates `SageMakerTrigger` to preserve SageMaker job lifecycle outcomes as explicit trigger event statuses.

The trigger now translates structured AWS waiter outcomes into `failed`, `stopped`, and `timeout` events, while successful jobs continue to emit `success`. Polling, retries, timeout detection, and AWS error handling remain managed by the shared `AwsBaseWaiterTrigger` implementation.

This is a follow-up to PR #69042 and the shared AWS waiter changes that allow individual triggers to translate waiter outcomes without reimplementing the polling lifecycle.

**Rationale**

`SageMakerTrigger` previously represented unsuccessful waiter outcomes using the generic `error` status, preventing consumers from distinguishing between a failed SageMaker job, an explicitly stopped job, and waiter exhaustion.

Using the exception-to-event translation provided by `AwsBaseWaiterTrigger` allows `SageMakerTrigger` to preserve these lifecycle states while continuing to rely on the shared AWS waiter implementation.

**Tests**

* Failed and stopped SageMaker jobs are translated into `failed` and `stopped` events.
* Failed jobs use the SageMaker failure reason when available.
* Waiter exhaustion is translated into a `timeout` event.
* Unrecognized terminal states and generic errors fall back to the default `error` event.

The stopped-job path was also manually validated against a real SageMaker training job by stopping the underlying job while the Airflow task was deferred.

**Backwards Compatibility**

This change does not modify the public API or constructor of `SageMakerTrigger`.

Successful jobs continue to emit `success`. Non-successful outcomes use more specific statuses instead of the generic `error` status. Existing SageMaker operators already treat any status other than `success` as a task failure, so their task-level behavior remains unchanged.

Generated-by: [GPT 5.6] following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
@@ -24,7 +24,7 @@ from collections.abc import AsyncIterator
 from enum import IntEnum
 from typing import TYPE_CHECKING, Any
 
-from botocore.exceptions import WaiterError
+from botocore.exceptions import NoCredentialsError, WaiterError
 
 from airflow.exceptions import AirflowProviderDeprecationWarning
 from airflow.providers.amazon.aws.hooks.sagemaker import SageMakerHook
@@ -118,22 +118,40 @@ class SageMakerTrigger(AwsBaseWaiterTrigger):
                 client=conn,
             )
 
-            for _ in range(self.attempts):
+            for attempt in range(self.attempts):
+                if attempt:
+                    await asyncio.sleep(self.waiter_delay)
+
                 try:
                     await waiter.wait(
                         **self.waiter_args,
                         WaiterConfig={"MaxAttempts": 1},
                     )
+
+                except NoCredentialsError as error:
+                    self.log.info(str(error))
+                    continue
+
                 except WaiterError as error:
                     response = error.last_response
                     status = response.get(self._get_response_status_key(self.job_type))
+
+                    if status is None:
+                        yield TriggerEvent(
+                            {
+                                "status": "error",
+                                "message": str(error),
+                                "job_name": self.job_name,
+                            }
+                        )
+                        return
 
                     if status == "Failed":
                         yield TriggerEvent(
                             {
                                 "status": "failed",
                                 "job_name": self.job_name,
-                                "message": response.get("FailureReason"),
+                                "message": response.get("FailureReason") or str(error),
                             }
                         )
                         return
@@ -143,21 +161,22 @@ class SageMakerTrigger(AwsBaseWaiterTrigger):
                             {
                                 "status": "stopped",
                                 "job_name": self.job_name,
+                                "message": str(error),
                             }
                         )
                         return
 
                     self._log_job_status(response)
-                    await asyncio.sleep(self.waiter_delay)
                     continue
 
-                yield TriggerEvent(
-                    {
-                        "status": "success",
-                        "job_name": self.job_name,
-                    }
-                )
-                return
+                else:
+                    yield TriggerEvent(
+                        {
+                            "status": "success",
+                            "job_name": self.job_name,
+                        }
+                    )
+                    return
 
         yield TriggerEvent(
             {
--- a/providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
+++ b/providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
@@ -22,7 +22,7 @@ import warnings
 from collections import Counter
 from collections.abc import AsyncIterator
 from enum import IntEnum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from botocore.exceptions import WaiterError
 
@@ -106,6 +106,90 @@ class SageMakerTrigger(AwsBaseWaiterTrigger):
             region_name=self.region_name,
             verify=self.verify,
             config=self.botocore_config,
+        )
+
+    async def run(self) -> AsyncIterator[TriggerEvent]:
+        hook = self.hook()
+
+        async with await hook.get_async_conn() as conn:
+            waiter = hook.get_waiter(
+                self.waiter_name,
+                deferrable=True,
+                client=conn,
+            )
+
+            for _ in range(self.attempts):
+                try:
+                    await waiter.wait(
+                        **self.waiter_args,
+                        WaiterConfig={"MaxAttempts": 1},
+                    )
+                except WaiterError as error:
+                    response = error.last_response
+                    status = response.get(self._get_response_status_key(self.job_type))
+
+                    if status == "Failed":
+                        yield TriggerEvent(
+                            {
+                                "status": "failed",
+                                "job_name": self.job_name,
+                                "message": response.get("FailureReason"),
+                            }
+                        )
+                        return
+
+                    if status == "Stopped":
+                        yield TriggerEvent(
+                            {
+                                "status": "stopped",
+                                "job_name": self.job_name,
+                            }
+                        )
+                        return
+
+                    self._log_job_status(response)
+                    await asyncio.sleep(self.waiter_delay)
+                    continue
+
+                yield TriggerEvent(
+                    {
+                        "status": "success",
+                        "job_name": self.job_name,
+                    }
+                )
+                return
+
+        yield TriggerEvent(
+            {
+                "status": "timeout",
+                "job_name": self.job_name,
+                "message": (
+                    f"SageMaker {self.job_type} job {self.job_name!r} "
+                    f"did not complete after {self.attempts} attempts."
+                ),
+            }
+        )
+
+    def _log_job_status(self, response: dict[str, Any]) -> None:
+        status = response.get(self._get_response_status_key(self.job_type))
+
+        if self.job_type.lower() == "training":
+            secondary_status = response.get("SecondaryStatus")
+            if secondary_status:
+                self.log.info(
+                    "SageMaker %s job %s is in state %s (%s)",
+                    self.job_type,
+                    self.job_name,
+                    status,
+                    secondary_status,
+                )
+                return
+
+        self.log.info(
+            "SageMaker %s job %s is in state %s",
+            self.job_type,
+            self.job_name,
+            status,
         )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


vincbeck:
@seanghaeli @ramitkataria

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
```diff
@@ -108,6 +108,90 @@ def hook(self) -> AwsGenericHook:
...
+
+            for _ in range(self.attempts):
+                try:
+                    await waiter.wait(
+                        **self.waiter_args,
+                        WaiterConfig={"MaxAttempts": 1},
+                    )
+                except WaiterError as error:
+                    response = error.last_response
+                    status = response.get(self._get_response_status_key(self.job_type))
```
ramitkataria, line 137: What's in `last_response` when the `DescribeTrainingJob` call itself fails? Bad job name, or a role without `sagemaker:DescribeTrainingJob`? Does `status` get a value on that path, and where does the loop go next? Worth checking against the endpoint operator, which doesn't pass `waiter_max_attempts`.

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
```diff
@@ -108,6 +108,90 @@ def hook(self) -> AwsGenericHook:
...
+                client=conn,
+            )
+
+            for _ in range(self.attempts):
+                try:
+                    await waiter.wait(
+                        **self.waiter_args,
+                        WaiterConfig={"MaxAttempts": 1},
+                    )
+                except WaiterError as error:
```
ramitkataria, line 135: `async_wait` also catches `NoCredentialsError` (added in #48946 for credential refresh during long jobs). Where does it go now?

  Also, `error` is bound but never used, and `message` is only `FailureReason`. What does the operator's `AirflowException` show when `FailureReason` is absent?

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
```diff
@@ -108,6 +108,90 @@ def hook(self) -> AwsGenericHook:
...
+                        yield TriggerEvent(
+                            {
+                                "status": "stopped",
+                                "job_name": self.job_name,
+                            }
+                        )
+                        return
+
+                    self._log_job_status(response)
+                    await asyncio.sleep(self.waiter_delay)
```
ramitkataria: `async_wait` sleeps at the top of the loop, skipped on the first pass. This sleeps at the bottom. On the final iteration, what is the sleep waiting for?

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
SameerMesiah97: `NoCredentialsError` is now being caught and if `FailureReason` is absent, the error itself is being provided.

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
SameerMesiah97: The below has now been added:

```
if status is None:
      yield TriggerEvent(
          {
              "status": "error",
              "message": str(error),
              "job_name": self.job_name,
          }
      )
      return
```

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
SameerMesiah97: The first iteration is now skipped and the sleep has been moved to the beginning of the loop.

SameerMesiah97:
Requesting review for this.

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
o-nikolas: We're also losing handling for other common errors the api can throw by trying to hand roll a solution here instead of using the async_wait ThrottlingError for example, how does this code handle that? Will it result in the Trigger reporting the task as failing?

o-nikolas (commented):
I think we really need to be careful when moving away from very well tested and delicate code that all Triggers share (async_wait) in this case and trying to re-implement it. I know you have good intentions and are trying to provide more information after the Trigger returns, but you're opening yourself up to a lot of new code to maintain (and almost duplicate) as well as all the bugs and edge cases that come along with that.

SameerMesiah97:
> I think we really need to be careful when moving away from very well tested and delicate code that all Triggers share (async_wait) in this case and trying to re-implement it. I know you have good intentions and are trying to provide more information after the Trigger returns, but you're opening yourself up to a lot of new code to maintain (and almost duplicate) as well as all the bugs and edge cases that come along with that.

After reading your comment, I have realized the core issue is not with the SageMakerTrigger but AwsBaseWaiterTrigger flattening the waiter outcome into success / error, so `SageMakerTrigger `can't preserve states like `Failed` and `Stopped` without taking over the polling loop.

What are your thoughts on adding a overridable translation method  i.e. `_event_from_exception` to `AwsBaseWaiterTrigger`, which the `SageMakerTrigger` uses to surface more granular `TriggerEvent` payloads? I believe this should preserve centralizied handling of `ThrottlingError`, the polling loop, timeouts etc whilst keeping the state machine for this trigger (and any other AWS trigger) explicit.

o-nikolas:
> > I think we really need to be careful when moving away from very well tested and delicate code that all Triggers share (async_wait) in this case and trying to re-implement it. I know you have good intentions and are trying to provide more information after the Trigger returns, but you're opening yourself up to a lot of new code to maintain (and almost duplicate) as well as all the bugs and edge cases that come along with that.
> 
> After reading your comment, I have realized the core issue is not with the SageMakerTrigger but AwsBaseWaiterTrigger flattening the waiter outcome into success / error, so `SageMakerTrigger `can't preserve states like `Failed` and `Stopped` without taking over the polling loop.
> 
> What are your thoughts on adding a overridable translation method i.e. `_event_from_exception` to `AwsBaseWaiterTrigger`, which the `SageMakerTrigger` uses to surface more granular `TriggerEvent` payloads? I believe this should preserve centralizied handling of `ThrottlingError`, the polling loop, timeouts etc whilst keeping the state machine for this trigger (and any other AWS trigger) explicit.

Yupp, I think this is the right approach, a generalized one. CC @ramitkataria @vincbeck

SameerMesiah97:
> > > I think we really need to be careful when moving away from very well tested and delicate code that all Triggers share (async_wait) in this case and trying to re-implement it. I know you have good intentions and are trying to provide more information after the Trigger returns, but you're opening yourself up to a lot of new code to maintain (and almost duplicate) as well as all the bugs and edge cases that come along with that.
> > 
> > 
> > After reading your comment, I have realized the core issue is not with the SageMakerTrigger but AwsBaseWaiterTrigger flattening the waiter outcome into success / error, so `SageMakerTrigger` can't preserve states like `Failed` and `Stopped` without taking over the polling loop.
> > What are your thoughts on adding a overridable translation method i.e. `_event_from_exception` to `AwsBaseWaiterTrigger`, which the `SageMakerTrigger` uses to surface more granular `TriggerEvent` payloads? I believe this should preserve centralizied handling of `ThrottlingError`, the polling loop, timeouts etc whilst keeping the state machine for this trigger (and any other AWS trigger) explicit.
> 
> Yupp, I think this is the right approach, a generalized one. CC @ramitkataria @vincbeck

I opened PR #72455 to add the overridable translation method i.e. _event_from_exception to `AwsBaseWaiterTrigger` which just got merged very recently. This PR now only overrides that method via its own implementation for `SageMakerTrigger`. Everything else is being handled by the shared AWS abstraction.

o-nikolas (approved):

providers/amazon/src/airflow/providers/amazon/aws/triggers/sagemaker.py
```diff
@@ -108,6 +109,41 @@ def hook(self) -> AwsGenericHook:
             config=self.botocore_config,
         )
 
+    def _event_from_exception(self, error: AirflowException) -> TriggerEvent:
+
+        if isinstance(error, WaiterMaxAttemptsError):
```
ramitkataria, line 114: Does this branch also catch the case where every attempt failed with `NoCredentialsError`?

ramitkataria: Continued from https://github.com/apache/airflow/pull/71653#discussion_r3806008895

SameerMesiah97: Yes. `async_wait` catches `NoCredentialsError` on each attempt and increments the attempt counter, so if every attempt fails due to missing credentials it will eventually raise `WaiterMaxAttemptsError`. I understand that missing credentials will be transmuted into a generic timeout event but it might be okay here.

ramitkataria: Fair, let's keep that out of scope for this PR. Might be worth a follow-up in `async_wait` to raise a distinct error when every attempt failed on credentials, so triggers can report it as `error` instead of `timeout`.

providers/amazon/tests/unit/amazon/aws/triggers/test_sagemaker.py
```diff
@@ -121,6 +123,91 @@ async def test_sagemaker_trigger_run_all_job_types(self, mock_async_conn, mock_g
...
+        trigger = SageMakerTrigger(
+            job_name=JOB_NAME,
+            job_type=JOB_TYPE,
+            waiter_delay=WAITER_DELAY,
+            waiter_max_attempts=WAITER_MAX_ATTEMPTS,
+            aws_conn_id=AWS_CONN_ID,
+        )
+        error = WaiterTerminalFailure(
+            "SageMaker job failed",
+            last_response={"TrainingJobStatus": aws_status},
```
ramitkataria, line 156: Should we add a parametrize case for the `FailureReason` path?

ramitkataria (approved):
Added a minor nit inline but otherwise looks good

PR #71654, opened 2026-08-15
Fix Dag run filtering for null start dates

fat-catTW:
## Why
`GET /dags/~/dagRuns` can return old finished Dag runs when filtering with `start_date_gte` or `start_date_gt` if those runs have `start_date=NULL`.

The nullable datetime filter treated `NULL` start dates as matching lower-bound filters unconditionally. That is useful for unfinished runs whose start date may still be populated later, but it is incorrect for runs that have already ended.

relates: #70627

## Solution
Limit the `NULL` lower-bound match for `DagRun.start_date` to runs whose `end_date` is also `NULL`.

This keeps the existing nullable datetime behavior for unfinished Dag runs, while excluding already-ended runs from recent start-date lower-bound queries.

---

### DIFF
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -1209,7 +1209,7 @@ class NullableDatetimeRangeFilter(RangeFilter):
         super().__init__(value, attribute)
         self.null_lower_bound_clause = null_lower_bound_clause
 
-    def _get_null_lower_bound_clause(self):
+    def _get_null_lower_bound_clause(self) -> ColumnElement[bool]:
         null_clause = self.attribute.is_(None)
         if self.null_lower_bound_clause is not None:
             null_clause = and_(null_clause, self.null_lower_bound_clause)
@@ -1239,7 +1239,11 @@ class NullableDatetimeRangeFilter(RangeFilter):
 
 
 def datetime_range_filter_factory(
-    filter_name: str, model: Base, attribute_name: str | None = None
+    filter_name: str,
+    model: Base,
+    attribute_name: str | None = None,
+    *,
+    null_lower_bound_clause: ColumnElement[bool] | None = None,
 ) -> Callable[[datetime | None, datetime | None, datetime | None, datetime | None], RangeFilter]:
     def depends_datetime(
         lower_bound_gte: datetime | None = Query(alias=f"{filter_name}_gte", default=None),
@@ -1254,13 +1258,7 @@ def datetime_range_filter_factory(
             upper_bound_lte=upper_bound_lte,
             upper_bound_lt=upper_bound_lt,
         )
-        attr_name = attribute_name or filter_name
-        if filter_name in ("start_date", "end_date") or (model is DagRun and attr_name == "start_date"):
-            null_lower_bound_clause: ColumnElement[bool] | None = None
-            if model is DagRun:
-                null_lower_bound_clause = DagRun.id.is_not(None)
-                if attr_name == "start_date":
-                    null_lower_bound_clause = and_(null_lower_bound_clause, DagRun.end_date.is_(None))
+        if filter_name in ("start_date", "end_date"):
             return NullableDatetimeRangeFilter(
                 range_val, attr, null_lower_bound_clause=null_lower_bound_clause
             )
--- a/airflow-core/src/airflow/api_fastapi/common/parameters.py
+++ b/airflow-core/src/airflow/api_fastapi/common/parameters.py
@@ -1210,7 +1210,7 @@ class NullableDatetimeRangeFilter(RangeFilter):
         self.null_lower_bound_clause = null_lower_bound_clause
 
     def _get_null_lower_bound_clause(self) -> ColumnElement[bool]:
-        null_clause = self.attribute.is_(None)
+        null_clause: ColumnElement[bool] = self.attribute.is_(None)
         if self.null_lower_bound_clause is not None:
             null_clause = and_(null_clause, self.null_lower_bound_clause)
         return null_clause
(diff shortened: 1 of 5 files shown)


ColtenOuO (commented):
Nice fix! This closes the exact repro from #70627 for `/dags/~/dagRuns`.

The narrowing only fires when `filter_name == "start_date"` , keyed off the query parameter name. 

But `GET /dags` filters the same `DagRun.start_date`/`DagRun.end_date` columns via
`datetime_range_filter_factory("dag_run_start_date", DagRun, "start_date")`
(dags.py:104), where `filter_name` is `"dag_run_start_date"`, not `"start_date"`.

That call goes through the identical join-to-latest-run + `NullableDatetimeRangeFilter`
path in `generate_dag_with_latest_run_query()` (db/dags.py:86-96), so a Dag whose
latest run is terminal with `start_date=NULL` and a populated `end_date` will still
match any `dag_run_start_date_gte`/`_gt` filter unconditionally.

It is the same bug, but in different endpoint.

Would it make sense to key the guard off the actual attribute being filtered instead
of the query param name, e.g.:

```python
    if model is DagRun and (attribute_name or filter_name) == "start_date":
        null_lower_bound_clause = DagRun.end_date.is_(None)
```

That would also cover this case.

fat-catTW:
Thanks for the review.

I updated the guard to key off the actual filtered attribute instead of only the query parameter name, so `dag_run_start_date` now gets the same null-start-date handling as `start_date`.

I also kept `dag_run_end_date` on the plain range filter path to avoid changing the existing `/dags` end-date behavior, and added `DagRun.id IS NOT NULL` to the start-date null guard so the outer join does not accidentally include Dags with no latest run.

Added coverage for the `/dags` case with a latest Dag run that has `start_date=NULL` and a populated `end_date`.

Verified locally with the focused regression tests, the existing `/dags` filter matrix, and `mypy-airflow-core`.

ColtenOuO (approved):
Thanks! Great work!

airflow-core/src/airflow/api_fastapi/common/parameters.py
```diff
@@ -1239,8 +1254,16 @@ def depends_datetime(
             upper_bound_lte=upper_bound_lte,
             upper_bound_lt=upper_bound_lt,
         )
-        if filter_name in ("start_date", "end_date"):
-            return NullableDatetimeRangeFilter(range_val, attr)
+        attr_name = attribute_name or filter_name
+        if filter_name in ("start_date", "end_date") or (model is DagRun and attr_name == "start_date"):
+            null_lower_bound_clause: ColumnElement[bool] | None = None
```
pierrejeambrun: We probably shouldn't have an `model is DagRun and attr_name == "start_date")`. Other models have the same problem `TaskInstance` and `Job`. 

This will just grow the 'if' when we actually fix those as follow up.

pierrejeambrun: Maybe move that parameter to the kwarg params of the factory intead.

```python
  # routes/public/dag_run.py
  start_date_range: Annotated[RangeFilter, Depends(datetime_range_filter_factory(
      "start_date", DagRun,
      null_lower_bound_clause=DagRun.end_date.is_(None),
  ))]

  # routes/public/dags.py — where the LEFT JOIN makes the id guard load-bearing
  dag_run_start_date_range: Annotated[RangeFilter, Depends(datetime_range_filter_factory(
      "dag_run_start_date", DagRun, "start_date",
      null_lower_bound_clause=and_(DagRun.id.is_not(None), DagRun.end_date.is_(None)),
  ))]
```

airflow-core/tests/unit/api_fastapi/common/test_parameters.py
```diff
@@ -504,14 +504,11 @@ def test_end_date_returns_nullable_filter(self):
         rf = _make_datetime_filter("end_date")
         assert isinstance(rf, NullableDatetimeRangeFilter)
 
-    def test_aliased_filter_name_returns_plain_filter(self):
-        """dag_run_start_date uses attribute_name='start_date' via outer join; NULL means 'no run',
-        not 'currently running', so it must return a plain RangeFilter to avoid inflating counts."""
+    def test_aliased_start_date_returns_nullable_filter(self):
         rf = _make_datetime_filter("dag_run_start_date", model=DagRun, attribute_name="start_date")
-        assert type(rf) is RangeFilter
+        assert isinstance(rf, NullableDatetimeRangeFilter)
```
pierrejeambrun, line 512: This was an explicit design decision and this is reverting it. Is that intentional?

Basically before this, a dag with 'queued' latest run (not started) wouldn't return. Now a dag with latest dag run in queued state will return for `/dags?dag_run_start_date_gte=2026-08-21T00:00:00Z`

WDYT is that a preferable behavior?

pierrejeambrun (commented):
Thanks for the PR. 

A few comments/suggestions

airflow-core/tests/unit/api_fastapi/common/test_parameters.py
fat-catTW: Good catch. Thanks
This behavior change was not intentional. I restored the plain `RangeFilter` for `dag_run_start_date`, so a Dag whose latest run is queued and has not started remains excluded from this filter.

airflow-core/src/airflow/api_fastapi/common/parameters.py
fat-catTW: Thanks, for review!!
I moved `null_lower_bound_clause` to an explicit keyword-only factory argument and now pass `DagRun.end_date.is_(None)` from the Dag run route, removing the `DagRun`-specific condition from the shared factory.

ColtenOuO (commented):
Thanks! I think we are almost there. Just a small suggestion

Would it make sense to add a targeted unit test for `null_lower_bound_clause` here?

Currently, `null_lower_bound_clause` is only tested indirectly through the endpoint in `test_dag_run.py`. We're missing a direct unit test in `TestDatetimeRangeFilterFactory` asserting that `to_orm()` actually incorporates it into the NULL branch -- e.g. checking the compiled WHERE clause contains the extra AND condition.

Adding this would make it much easier and faster to pinpoint issues if this gets refactored in the future.

WDYT

PR #71662, opened 2026-08-15
helm - allow empty pass for metadataConnection ( token authent )

raphaelauv:
I want to be able to set an empty password for the `metadataConnection`

```yaml
data:
  metadataConnection:
    user: airflow_iam_user
    pass: ""
    host: xxx.yyyy.rds.amazonaws.com
    port: 5432
    db: airflow_db
```

to get `postgresql://airflow_iam_user@xxx.yyyy.rds.amazonaws.com:5432/airflow_db`

currently I get `postgresql://airflow_iam_user:@xxx.yyyy.rds.amazonaws.com:5432/airflow_db`

### DIFF
--- a/chart/templates/secrets/metadata-connection-secret.yaml
+++ b/chart/templates/secrets/metadata-connection-secret.yaml
@@ -53,15 +53,21 @@ metadata:
 type: Opaque
 data:
   {{- with .Values.data.metadataConnection }}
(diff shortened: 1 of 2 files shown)


raphaelauv:
airflowLocalSettings and apiServerConfig already works with the `airflow_iam_user:` , so it's not a necessary change

but if the user do not set the pass , it put the default password `postgres`

PR #71665, opened 2026-08-15
Fix HttpHook double slash when base_url and endpoint both have slashes

### LINKED ISSUE #71658: HttpHook: double slash in URL when connection host has trailing slash and endpoint has leading slash

Providers

### Apache Airflow version

main (development)

`HttpHook._url_from_endpoint()` (in `providers/http/src/airflow/providers/http/hooks/http.py`) only guards against a *missing* slash between `base_url` and `endpoint` — it doesn't guard against a *double* one. If a connection's host is configured with a trailing `/` (common) and an endpoint is passed with a leading `/` (also common REST convention), the resulting URL has `//` in it.

Repro:

```python
from airflow.providers.http.hooks.http import HttpHook

hook = HttpHook()
hook.base_url = "https://api.example.com/v1/"
hook._base_url_initialized = True
print(hook.url_from_endpoint("/users"))
# https://api.example.com/v1//users  <- double slash
```

Many API frameworks (Flask, FastAPI, Django) do not normalize `//` in a path and will 404 on it, so this produces a silent, confusing failure for anyone hitting this combination. Affects both `HttpHook` and `HttpAsyncHook`, since both call the shared `_url_from_endpoint()` helper.

Exactly one `/` should separate the base URL and endpoint regardless of which side(s) already have a slash. `_url_from_endpoint()` should strip the trailing slash from `base_url` and the leading slash from `endpoint` before joining, rather than only checking for the missing-slash case.

Happy to open a PR with the fix and a test covering this slash combination — will link it here.

### Operating System

_No response_

### Deployment

None

http

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

cliam23:
_url_from_endpoint() only guarded against a missing slash between base_url and endpoint, not a double one. When a connection's host has a trailing slash and the endpoint passed to run()/url_from_endpoint() has a leading slash, the resulting URL contained '//', which many API frameworks (Flask, FastAPI, Django) do not normalize and will 404 on.

Fix by stripping at most one trailing slash from base_url and at most one leading slash from endpoint before joining. A full rstrip('/')/ lstrip('/') was tried first but is too aggressive: base_url can legitimately end in '//' on its own (e.g. a bare 'http://' with an empty host), and stripping all trailing slashes there eats the scheme separator itself, breaking an existing test
(test_build_request_url_from_endpoint_param). Stripping at most one slash from each side fixes the double-slash bug without that regression. Affects both HttpHook and HttpAsyncHook since both call this shared helper.

Closes #71658

 

---

---

### DIFF
--- a/providers/http/src/airflow/providers/http/hooks/http.py
+++ b/providers/http/src/airflow/providers/http/hooks/http.py
@@ -47,9 +47,18 @@ if TYPE_CHECKING:
 
 def _url_from_endpoint(base_url: str | None, endpoint: str | None) -> str:
     """Combine base url with endpoint."""
-    if base_url and not base_url.endswith("/") and endpoint and not endpoint.startswith("/"):
-        return f"{base_url}/{endpoint}"
-    return (base_url or "") + (endpoint or "")
+    if not base_url:
+        return endpoint or ""
+    if not endpoint:
+        return base_url
+    # Strip at most one slash from each side, not all of them: base_url can
+    # legitimately end in "//" (e.g. a bare "http://" with an empty host),
+    # and a full rstrip("/")/lstrip("/") would eat the scheme separator too.
+    if base_url.endswith("/"):
+        base_url = base_url[:-1]
+    if endpoint.startswith("/"):
+        endpoint = endpoint[1:]
+    return f"{base_url}/{endpoint}"
 
 
 def _process_extra_options_from_connection(
--- a/providers/http/tests/unit/http/hooks/test_http.py
+++ b/providers/http/tests/unit/http/hooks/test_http.py
@@ -599,6 +599,12 @@ class TestHttpHook:
         [
             pytest.param("https://example.org", "/v1/test", "https://example.org/v1/test", id="both-set"),
             pytest.param("", "http://foo/bar/v1/test", "http://foo/bar/v1/test", id="only-endpoint"),
+            pytest.param(
+                "https://example.org/",
+                "/v1/test",
+                "https://example.org/v1/test",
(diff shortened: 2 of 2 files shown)


providers/http/src/airflow/providers/http/hooks/http.py
```diff
@@ -47,9 +47,18 @@
 
 def _url_from_endpoint(base_url: str | None, endpoint: str | None) -> str:
     """Combine base url with endpoint."""
-    if base_url and not base_url.endswith("/") and endpoint and not endpoint.startswith("/"):
-        return f"{base_url}/{endpoint}"
-    return (base_url or "") + (endpoint or "")
+    if not base_url:
+        return endpoint or ""
+    if not endpoint:
+        return base_url
```
jroachgolf84, line 53:
```suggestion
        return base_url
        
```

providers/http/tests/unit/http/hooks/test_http.py
```diff
@@ -599,6 +599,12 @@ def test_keep_alive_disabled(self):
         [
             pytest.param("https://example.org", "/v1/test", "https://example.org/v1/test", id="both-set"),
             pytest.param("", "http://foo/bar/v1/test", "http://foo/bar/v1/test", id="only-endpoint"),
+            pytest.param(
```
jroachgolf84, line 602: Can you also add an example with just `https://` as the `base_url` (and whatever else you'd like after it), since you call it out in your comment?

PR #55568, opened 2025-09-12
Implement async version of databricks_conn in BaseDatabricksHook

BasPH:
I bumped into this error when running the DatabricksSubmitRunOperator on Airflow 3.0.6 using apache-airflow-providers-databricks==7.7.1:
```
ERROR - Trigger failed:
Traceback (most recent call last):

  File "/usr/local/lib/python3.12/site-packages/airflow/jobs/triggerer_job_runner.py", line 963, in cleanup_finished_triggers
    result = details["task"].result()
             ^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/site-packages/airflow/jobs/triggerer_job_runner.py", line 1072, in run_trigger
    async for event in trigger.run():

  File "/usr/local/lib/python3.12/site-packages/airflow/providers/databricks/triggers/databricks.py", line 90, in run
    run_state = await self.hook.a_get_run_state(self.run_id)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/site-packages/airflow/providers/databricks/hooks/databricks.py", line 514, in a_get_run_state
    response = await self._a_do_api_call(GET_RUN_ENDPOINT, json)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/site-packages/airflow/providers/databricks/hooks/databricks_base.py", line 713, in _a_do_api_call
    url = self._endpoint_url(full_endpoint)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/site-packages/airflow/providers/databricks/hooks/databricks_base.py", line 623, in _endpoint_url
    port = f":{self.databricks_conn.port}" if self.databricks_conn.port else ""
                                              ^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/functools.py", line 998, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.12/site-packages/airflow/providers/databricks/hooks/databricks_base.py", line 142, in databricks_conn
    return self.get_connection(self.databricks_conn_id)  # type: ignore[return-value]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #71667, opened 2026-08-16
Document Airflow 3 requirement for Databricks deferral

### LINKED ISSUE #71525: DatabricksRunNowOperator(deferrable=True) crashes in triggerer: AsyncToSync used inside running event loop (Airflow 3.0.6)
### Apache Airflow version

3.0.6

_No response_

### What happened?

Running `DatabricksRunNowOperator(deferrable=True)` (and this likely applies to `DatabricksSubmitRunOperator` too, since both share the same trigger/hook code) submits the Databricks job successfully and the task defers correctly (worker slot released). But the triggerer then crashes while polling job status, before the run ever reaches a terminal state:

```
ERROR - Trigger failed:
  File ".../airflow/jobs/triggerer_job_runner.py", line 963, in cleanup_finished_triggers
    result = details["task"].result()
  File ".../airflow/jobs/triggerer_job_runner.py", line 1072, in run_trigger
    async for event in trigger.run():
  File ".../airflow/providers/databricks/triggers/databricks.py", line 90, in run
    run_state = await self.hook.a_get_run_state(self.run_id)
  File ".../airflow/providers/databricks/hooks/databricks.py", line 514, in a_get_run_state
    response = await self._a_do_api_call(GET_RUN_ENDPOINT, json)
  File ".../airflow/providers/databricks/hooks/databricks_base.py", line 713, in _a_do_api_call
    url = self._endpoint_url(full_endpoint)
  File ".../airflow/providers/databricks/hooks/databricks_base.py", line 623, in _endpoint_url
    port = f":{self.databricks_conn.port}" if self.databricks_conn.port else ""
  File ".../functools.py", line 998, in __get__
    val = self.func(instance)
  File ".../airflow/providers/databricks/hooks/databricks_base.py", line 142, in databricks_conn
    return self.get_connection(self.databricks_conn_id)
  File ".../airflow/hooks/base.py", line 64, in get_connection
    conn = Connection.get_connection_from_secrets(conn_id)
  File ".../airflow/models/connection.py", line 478, in get_connection_from_secrets
    conn = TaskSDKConnection.get(conn_id=conn_id)
  File ".../airflow/sdk/definitions/connection.py", line 144, in get
    return _get_connection(conn_id)
  File ".../airflow/sdk/execution_time/context.py", line 160, in _get_connection
    msg = SUPERVISOR_COMMS.send(GetConnection(conn_id=conn_id))
  File ".../airflow/jobs/triggerer_job_runner.py", line 740, in send
    return async_to_sync(self.asend)(msg)
  File ".../asgiref/sync.py", line 186, in __call__
    raise RuntimeError(
RuntimeError: You cannot use AsyncToSync in the same thread as an async event loop - just await the async function directly.
```

Root cause: `DatabricksHook.databricks_conn` is a synchronous `@cached_property` (`hooks/databricks_base.py`) that fetches the connection via the TaskSDK's `SUPERVISOR_COMMS.send()`, which internally bridges to sync via `asgiref.async_to_sync`. The trigger's async polling path (`DatabricksExecutionTrigger.run()` -> `hook.a_get_run_state()` -> `_a_do_api_call()` -> `_endpoint_url()`) calls this same synchronous property, but this time from *inside* the triggerer's already-running asyncio event loop. `async_to_sync` detects the running loop and refuses to bridge, raising `RuntimeError` instead of fetching the connection.

Since the trigger constructs a fresh `DatabricksHook` instance (`DatabricksExecutionTrigger.__init__`), `databricks_conn` has never been cached beforehand, so the first (and every) access inside `run()` hits this failure — the deferred task can never successfully poll.

I verified this is not fixed in the latest released provider: `apache-airflow-providers-databricks==7.18.1`'s `hooks/databricks_base.py` still defines `databricks_conn` as a plain synchronous `@cached_property` and `_a_do_api_call`/`_endpoint_url` are structurally unchanged from 7.7.1 — so bumping the provider version alone does not resolve this.

This looks like the same class of bug as #53447 (BigQuery hit the identical `AsyncToSync`-in-running-loop error). A Databricks-specific fix was proposed in #55568 ("Implement async version of databricks_conn in BaseDatabricksHook") but it stalled and was auto-closed as stale without merging, despite multiple people confirming the bug is still present ("this is not solved, please reopen"). A more general core-level fix (#55799, follow-up #57154) exists for the `ExecutionAPISecretsBackend` fallback path, but targets Airflow core >= 3.1.0 - we're not yet in a position to confirm whether that alone resolves the Databricks hook's specific synchronous `cached_property` pattern, since `_endpoint_url` calls `self.databricks_conn` directly rather than going through an async-safe path.

`DatabricksRunNowOperator`/`DatabricksSubmitRunOperator` with `deferrable=True` should be able to poll run status from the triggerer without crashing. Either:
- Reopen and land #55568 (or an equivalent), giving `BaseDatabricksHook` an async-safe connection-fetch path that `_a_do_api_call`/`_endpoint_url` use when running inside the triggerer, or
- Confirm the core-level fix (#57154, Airflow >= 3.1.0) actually resolves this specific code path, and document the minimum core version required for Databricks deferrable operators to work.

### How to reproduce

1. Airflow 3.0.6, `apache-airflow-providers-databricks==7.7.1` (also reproduces on 7.18.1's unchanged code path per the above).
2. Any DAG using `DatabricksRunNowOperator(deferrable=True, job_name=..., databricks_conn_id=...)`.
3. Trigger the DAG. The job submits successfully (log shows "Run submitted with run_id: ..." and "Pausing task as DEFERRED"), then the triggerer immediately crashes with the `AsyncToSync` `RuntimeError` above on its first status poll.

### Operating System

N/A (AWS MWAA managed Airflow 3.0.6 environment)

apache-airflow-providers-databricks==7.7.1 (also verified 7.18.1 has the same unchanged code path)

### Deployment

Amazon (AWS) MWAA

### Deployment details

_No response_

### Anything else?

Related: #53447, #55568, #55799, #57154

### Code of Conduct

breadtitor:
Document the Airflow 3 version requirement for deferrable Databricks run-now and submit-run operators.

Airflow 3.0.x cannot fetch the Databricks connection from the Triggerer during asynchronous polling, while Airflow 3.1.0 and later handle this path in core. Documenting the boundary gives affected users an actionable upgrade path instead of suggesting an unnecessary provider-side fix.

closes: #71525

Validation:

- RST syntax checked with `rstcheck`
- Spelling checked with `codespell`
- `git diff --check`

---

- [X] Yes — Codex (GPT-5)

---

### DIFF
--- a/providers/databricks/docs/operators/run_now.rst
+++ b/providers/databricks/docs/operators/run_now.rst
@@ -146,4 +146,8 @@ DatabricksRunNowDeferrableOperator
 
 Deferrable version of the :class:`~airflow.providers.databricks.operators.DatabricksRunNowOperator` operator.
 
+.. note::
+  When using Airflow 3, deferrable execution requires Airflow 3.1.0 or later. Airflow 3.0.x
+  cannot fetch the Databricks connection from the Triggerer during asynchronous polling.
+
 It allows to utilize Airflow workers more effectively using `new functionality introduced in Airflow 2.2.0 <https://airflow.apache.org/docs/apache-airflow/2.2.0/concepts/deferring.html#triggering-deferral>`_
--- a/providers/databricks/docs/operators/submit_run.rst
+++ b/providers/databricks/docs/operators/submit_run.rst
@@ -225,4 +225,8 @@ DatabricksSubmitRunDeferrableOperator
 
 Deferrable version of the :class:`~airflow.providers.databricks.operators.DatabricksSubmitRunOperator` operator.
 
+.. note::
+  When using Airflow 3, deferrable execution requires Airflow 3.1.0 or later. Airflow 3.0.x
+  cannot fetch the Databricks connection from the Triggerer during asynchronous polling.
+
 It allows to utilize Airflow workers more effectively using `new functionality introduced in Airflow 2.2.0 <https://airflow.apache.org/docs/apache-airflow/2.2.0/concepts/deferring.html#triggering-deferral>`_


providers/databricks/docs/operators/run_now.rst
```diff
@@ -146,4 +146,8 @@ DatabricksRunNowDeferrableOperator
 
 Deferrable version of the :class:`~airflow.providers.databricks.operators.DatabricksRunNowOperator` operator.
 
+.. note::
+  When using Airflow 3, deferrable execution requires Airflow 3.1.0 or later. Airflow 3.0.x
+  cannot fetch the Databricks connection from the Triggerer during asynchronous polling.
```
eladkal, line 151: Reading https://github.com/apache/airflow/issues/71525

I am not sure if this is a doc issue. Shouldn't we check on runtime the airflow version and if it is 3.0 then alert users / fallback to the supported functionality?

cc @moomindani

moomindani (commented):
Thanks for picking this up, and for pinning the note to a concrete core version rather than leaving it vague.

First, the boundary in the note is right. I ran the matrix on official images with the provider pinned at 7.18.1 in every run, a DB-backed Databricks connection and a real job: 2.11.0 succeeds, 3.0.6 fails with the `AsyncToSync` `RuntimeError` (reproduced twice), and 3.1.0, 3.1.1 and 3.2.2 all succeed. So the affected window is exactly the 3.0.x line. For the record on 2.x, `get_connection_from_secrets` there is a plain synchronous secrets lookup with no Execution API and no `async_to_sync`, which is why the 2.x triggerer is unaffected.

Three things about the note as placed:

1. Both sections it was added to are dead. `DatabricksRunNowDeferrableOperator` and `DatabricksSubmitRunDeferrableOperator` were removed in provider 7.0.0 (`changelog.rst:753-754`); on `main` those names survive only as the two headings this note now sits under. The live `deferrable=True` prose is at `run_now.rst:135-137` and `submit_run.rst:214-216`, which is where someone setting `deferrable=True` will actually be reading.

2. The failure is in the shared hook (`databricks_base.py:193-195`, reached from `_a_do_api_call:1387` via `_endpoint_url:1304`), so it hits every deferrable surface in the provider — including `DatabricksSQLStatementsOperator` and `DatabricksSQLStatementsSensor`, both documented in `sql_statements.rst:34` and `:84` where deferrable mode is explicitly advertised, and `DatabricksNotebookOperator` / `DatabricksTaskOperator` via `DatabricksTaskBaseOperator:1714`. Worth covering those pages too.

3. Including the literal `RuntimeError: You cannot use AsyncToSync in the same thread as an async event loop` would help, since that string is what people search for when they land here.

On the runtime-check question: I looked into it, and I think there is a better answer than either docs or a version guard.

A hard raise or a warning would need `AIRFLOW_V_3_1_PLUS` (one line in `version_compat.py`), and the warn-and-disable pattern already exists in this same file as `_warn_and_disable_durable_pre_3_3` (`operators/databricks.py:75-83`). But I would avoid silently downgrading `deferrable=True` to synchronous: there is no precedent for that in `providers/*/src`, and it converts a released worker slot into one held for the entire Databricks run, which is the opposite of what the user asked for. `deferrable` also defaults from `[operators] default_deferrable`, so a silent downgrade would be invisible at Dag level.

More usefully, the provider can just fix this. `databricks_conn` is a `cached_property`, and both triggers already enter the hook through `async with self.hook:` (`triggers/databricks.py:132` and `:298`), so priming that cache once from async context makes every later synchronous read a cache hit:

```python
async def __aenter__(self):
    self._session = aiohttp.ClientSession()
    if "databricks_conn" not in self.__dict__:
        self.__dict__["databricks_conn"] = await sync_to_async(self.get_connection)(self.databricks_conn_id)
    return self
```

`sync_to_async` runs the getter on a worker thread where no loop is running, so the SDK's internal `async_to_sync` bridge is legal again. I tested exactly this against Airflow 3.0.6 with provider 7.18.1 and a DB-backed connection: the task went deferred and then succeeded, the trigger log shows `Connection Retrieved 'databricks_default'` followed by normal `life_cycle_state` polling to `TERMINATED`, and zero `AsyncToSync` errors — where the unpatched run on the same image fails on its first poll. That is about ten lines and no signature churn, compared with #55568's approach of threading a `conn` local through roughly fourteen async methods. Adding the same guard at the top of `_a_do_api_call` would also cover any caller that skips the context manager.

One thing worth stating for anyone finding this from the issue: there is no configuration workaround on 3.0.x. I tried supplying the connection as an `AIRFLOW_CONN_DATABRICKS_DEFAULT` env var instead of a DB row, on the theory that a triggerer-local secrets backend would avoid the bridge. It does not: the worker submits the run fine, but the trigger still fails twice over — once inside the env-var backend path and again on the fall-through to `SUPERVISOR_COMMS.send` — so upgrading core, or the hook change above, are the only real options.

So my suggestion: keep this note (it is worth having either way) but move it next to the `deferrable=True` prose and extend it to the other deferrable pages. If maintainers want the hook fix, I am happy to open that PR, in which case this note becomes informational rather than the only mitigation.

---

PR #71669, opened 2026-08-16
Fix slow triggerer log draining for large records

### LINKED ISSUE #66158: Large string logs from trigger are not written to log file intermittently.

Airflow Core

### Apache Airflow version

3.1.8

We noticed reports where large string logs from trigger are not displayed in UI intermittently. The lines are also not present in the log file on the disk. We have custom Spark and Kubernetes based triggers where we download the container log and pod log to be logged using `self.log.info(f"Output : {pod_log}")` inside custom triggers. The pod_log sometimes are large as 4-5 MB as a single string passed through socket. As per our understanding from Airflow 3 the logs pass through sockets and then reach the structlog logger in process_log_messages_from_subprocess compared to Airflow 2 where it directly got written to the file from the trigger through file handler. So while processing large strings and when the triggerer is also busy the messages seem to get lost to be not logged in the triggerer. Sometimes the large string is broken and `msgspec.json.decode(line)` raises exception while decoding causing them to be skipped. This is not consistently reproducible as it depends on the triggerer load and also the large string.

In Airflow 3 is there a way to skip the socket layer and directly write to the trigger.log file for the trigger similar to Airflow 2.

https://github.com/apache/airflow/blob/44b79fe53981bee4c9ccf163acd927be34042774/task-sdk/src/airflow/sdk/execution_time/supervisor.py#L2008-L2047

_No response_

### Operating System

_No response_

### Deployment

Virtualenv installation

_No response_

_No response_

Not Applicable

### Kubernetes Version

_No response_

### Helm Chart configuration

_No response_

_No response_

### Anything else?

_No response_

### Code of Conduct

korex-f:
`make_buffered_socket_reader` re-scanned the entire accumulated buffer from offset 0 on every socket read, making newline scanning O(N²) in the size of a single record. A 5 MiB single-line trigger log record, as reported in #66158, could cost gigabytes of cumulative prefix scanning before the terminating newline was found, delaying drainage of supervised sockets under triggerer load.

This PR:

* Adds a persistent scan cursor (`search_from`) so each byte in the buffer is examined once per delivery, rather than once per chunk.
* Sends each completed record to the generator as an immutable `bytes` value.
* Drops incomplete records at EOF only for the newline-delimited JSON log channel and reports bounded metadata without exposing the fragment content.
* Preserves the existing behavior for stdout and stderr, where a final line without a newline is valid output and must still be forwarded.
* Adds deterministic socket tests covering large records split across many reads and both EOF policies.

The newline-delimited wire protocol is unchanged. The helper's existing consumers retain their previous behavior unless they explicitly opt into dropping an incomplete JSON record at EOF.

A separate GCS-handler symptom mentioned later in #66158, involving missing files for ordinary-sized messages, is outside the scope of this change.

closes: #66158

---

- [X] Yes — Claude Code and OpenAI Codex (GPT-5) assisted with planning and review; the implementation was written and reviewed by the contributor.

Generated-by: Claude Code and OpenAI Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

---

### DIFF
--- a/task-sdk/src/airflow/sdk/execution_time/supervisor.py
+++ b/task-sdk/src/airflow/sdk/execution_time/supervisor.py
@@ -2314,11 +2314,8 @@ def make_buffered_socket_reader(
         nonlocal search_from
         # We could have read multiple lines in one go, yield them all
         while (newline_pos := buffer.find(b"\n", search_from)) != -1:
-            # Copy the record out as immutable bytes before sending it: the generator's
-            # local may still reference this value at its next suspension point, so it
-            # must not alias the buffer we're about to resize below -- resizing a
-            # bytearray while a memoryview onto it is live raises BufferError.
-            line = bytes(buffer[: newline_pos + 1])
+            # Avoid an intermediate bytearray copy while producing the immutable record.
+            line = bytes(memoryview(buffer)[: newline_pos + 1])
             gen.send(line)
             buffer = buffer[newline_pos + 1 :]  # Update the buffer with remaining data
             search_from = 0
--- a/task-sdk/src/airflow/sdk/execution_time/supervisor.py
+++ b/task-sdk/src/airflow/sdk/execution_time/supervisor.py
@@ -863,7 +863,9 @@ class WatchedSubprocess:
             logs,
             selectors.EVENT_READ,
             make_buffered_socket_reader(
-                process_log_messages_from_subprocess(target_loggers), on_close=self._on_socket_closed
+                process_log_messages_from_subprocess(target_loggers),
+                on_close=self._on_socket_closed,
+                drop_incomplete_at_eof=True,
             ),
         )
         self.selector.register(
@@ -1993,7 +1995,9 @@ class ActivitySubprocess(WatchedSubprocess):
             read_logs,
             selectors.EVENT_READ,
             make_buffered_socket_reader(
-                process_log_messages_from_subprocess(target_loggers), on_close=self._on_socket_closed
+                process_log_messages_from_subprocess(target_loggers),
+                on_close=self._on_socket_closed,
+                drop_incomplete_at_eof=True,
             ),
         )
         # We don't explicitly close the old log socket, that will get handled for us if/when the other end is
@@ -2291,6 +2295,7 @@ def make_buffered_socket_reader(
     data: bytes = b"",
     on_close: Callable[[socket], None],
     buffer_size: int = 4096,
+    drop_incomplete_at_eof: bool = False,
 ):
     buffer = bytearray(data)  # This will hold our accumulated binary data
     read_buffer = bytearray(buffer_size)  # Temporary buffer for each read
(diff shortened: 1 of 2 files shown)


task-sdk/src/airflow/sdk/execution_time/supervisor.py
```diff
@@ -2291,19 +2295,34 @@ def make_buffered_socket_reader(
...
     def process(buffer: bytearray) -> bytearray:
+        nonlocal search_from
         # We could have read multiple lines in one go, yield them all
-        while (newline_pos := buffer.find(b"\n")) != -1:
-            line = buffer[: newline_pos + 1]
+        while (newline_pos := buffer.find(b"\n", search_from)) != -1:
+            # Copy the record out as immutable bytes before sending it: the generator's
+            # local may still reference this value at its next suspension point, so it
+            # must not alias the buffer we're about to resize below -- resizing a
+            # bytearray while a memoryview onto it is live raises BufferError.
```
ColtenOuO: I think there 4 lines comment feels a bit off.

```python
buffer = buffer[newline_pos + 1 :]
```
In Python, this operation actually creates a new object, so the original buffer isn't being resized.

task-sdk/src/airflow/sdk/execution_time/supervisor.py
```diff
@@ -2291,19 +2295,34 @@ def make_buffered_socket_reader(
...
+            line = bytes(buffer[: newline_pos + 1])
```
ColtenOuO: Would it be better to use `memoryview` here to save an intermediate copy?

```python
line = bytes(memoryview(buffer)[: newline_pos + 1])
```

ColtenOuO (commented):
Thanks for improvement! 

I've left some suggestions in the code for your reference. If anything seems off or incorrect, feel free to bring it up for discussion! ><

By the way:

The title reads a bit terse for what this PR actually does — it's mostly a keyword string plus a bare issue number, which doesn't tell a reader what user-visible problem got fixed. Since `closes: #66158` already links the issue, the title could focus on describing the fix itself, e.g.:

- "Fix slow triggerer log draining on large single-line records"
- "Fix quadratic log-buffer scanning in triggerer under large records"

(Just a suggestion — happy to leave it as-is if you'd rather keep it short.)

task-sdk/src/airflow/sdk/execution_time/supervisor.py
korex-f: Good catch. The slice creates a new bytearray rather than resizing the original, so I removed the inaccurate explanation and replaced it with a concise comment describing the copy avoidance. I also updated the PR title to describe the user-visible problem more clearly.

task-sdk/src/airflow/sdk/execution_time/supervisor.py
korex-f: Thanks for the suggestion. I adopted the memoryview approach so conversion to immutable bytes no longer creates an intermediate bytearray copy.

PR #71400, opened 2026-08-11
Verify Java SDK build dependencies

jason810496:
## Why

The Java SDK build trusted whatever its configured repositories returned, so a replaced or shadowed artifact could execute during a build without anything failing.

## How

- Pin SHA-256 checksums for dependencies, plugins, and metadata in `gradle/verification-metadata.xml`. Gradle enables strict verification automatically when that file is present, so no build invocation needs a new flag.
- Centralize repositories in `settings.gradle.kts` and reject project-added ones.
- Reject dynamic and changing versions for project dependency configurations.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Codex GPT-5.6 following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

PR #71670, opened 2026-08-16
Bump Java SDK dependencies with Java 11 compatibility

jason810496:
- related: https://github.com/apache/airflow/pull/71605
- related: https://github.com/apache/airflow/pull/71636

## Why

The grouped Java SDK dependency update fails CI because its verification metadata is incomplete, two requested Jackson versions are unpublished, and some upgraded Gradle plugins require Java 17 while the SDK build runs on Java 11.

## What

- Keep `org.jsonschema2pojo` at 1.2.2 and the Shadow plugin at 9.1.0, the latest compatible versions for the Java 11 build.
- Use published Jackson 2.22.1 artifacts and retain the valid dependency updates, including Ktor 3.5.2.
- Regenerate strict Gradle verification metadata from a clean file so only resolved artifacts remain trusted.
- Align the standalone dependency-submission jobs with Gradle 8.14.5.

---

- [x] Yes, with help of Codex (GPT-5) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)


java-sdk/gradle/verification-metadata.xml
uranusjr, line 1: Do we have CI checks for this file? We should if not. This can be a separate PR.

uranusjr (approved):

java-sdk/gradle/verification-metadata.xml
jason810496: The verification is introduced in https://github.com/apache/airflow/pull/71400. But there isn't a good way for dependent itself to regenerate the hash, so I still need to cherry-pick and manually generate like this PR.

Thanks!

FrankYang0529: Create a PR for automatic checking in CI.

https://github.com/apache/airflow/pull/71869

PR #68847, opened 2026-06-22
Add SandboxToolset for sandboxed agent shell and file access

zozo123:
## Summary

Adds `SandboxToolset` to the Common AI provider: an agent gets a shell and a filesystem inside a disposable sandbox that runs off the Airflow worker process.

| Tool | What it does |
|---|---|
| `run_command` | Runs a shell command. Pipes, redirection, `&&` and globs work. A non-zero exit is returned as output, not raised, so the model reads `stderr` and corrects itself. |
| `read_file` | Reads a text file head-first and reports the next `offset`, so the model can page through a long file. |
| `write_file` | Writes text, creating parent directories. |
| `list_directory` | Lists a directory, marking directories with a trailing `/`. |

These are the same four names and shapes pydantic-ai's own sandbox capabilities use, so a model that has seen one already knows this one, and a vendor with an adapter for one is close to having written this one.

One backend ships: `SbxSandboxBackend`, driving Docker Sandboxes through the `sbx` CLI. It needs no Python dependency, so this PR adds no runtime requirement and leaves `uv.lock` untouched.

## Which boundary this is

"Sandboxed" means different things at different layers, and picking the wrong layer is the common way to end up with less protection than you expected.

| Boundary | What moves inside | Protects against |
|---|---|---|
| A tool call | What these four tools do. **This PR.** | Model-written code damaging the worker host, reading its files, or reaching the network from it |
| The glue between tools | Generated orchestration code, via [code mode](https://github.com/apache/airflow/pull/68407) | Generated code touching anything but registered tools. Still runs in the worker process |
| The agent process | The agent loop, its LLM credentials, its message history | The agent's own credentials leaking. Not available today |
| The whole task | The complete Airflow task, as `KubernetesExecutor` does | Everything, including Airflow's worker context |

This is the first row, and the docs say so plainly. Airflow puts none of its context, connections, variables or worker environment into the sandbox. **It does not contain the agent**: the agent loop and every other toolset on the same agent still run in the worker with the worker's credentials, so an agent holding `SandboxToolset` alongside a toolset that can reach connections has a contained code tool sitting next to an uncontained credential path.

## Design decisions

**Four tools rather than one.** An earlier revision exposed a single `run_python_in_sandbox`, which let an agent run code but never get a result back out. Files are how work leaves a sandbox, so the file tools are not a convenience.

[...]

PR #71672, opened 2026-08-16
Add Islo backend for sandbox tools

zozo123:
Adds a hosted Islo microVM backend for `SandboxToolset`, giving containerized Airflow workers a production-capable backend without a local daemon or KVM.

### What changed

- maps `SandboxSpec` environment and deny-all/allow-all network policy at creation, rejecting unsupported domain allowlists;
- uses native streaming file transfer and bounds command output inside the microVM;
- destroys timed-out sandboxes and configures a server-side TTL as the orphan-cleanup backstop;
- resolves the optional Islo SDK and Airflow connection lazily; and
- documents the `sandbox-islo` extra and adds unit plus system coverage.

### Validation

- 45 focused unit tests pass with the current SDK and minimum `islo==0.3.9`;
- the complete system test passed against a real Islo microVM, covering all four tools, persistence, failures, and teardown;
- provider import succeeds without the optional Islo SDK; and
- Ruff, formatting, codespell, provider metadata checks, and relevant pre-commit hooks pass locally.

Related: #68847

---

- [x] Yes — Codex (GPT-5.6)

Generated-by: Codex (GPT-5.6) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/providers/common/ai/docs/toolsets.rst
+++ b/providers/common/ai/docs/toolsets.rst
@@ -799,7 +799,7 @@ Islo backend
 :class:`~airflow.providers.common.ai.sandbox.IsloSandboxBackend` runs each
 sandbox in an `islo.dev <https://islo.dev>`__ microVM. Unlike ``sbx``, the
 worker talks to a hosted API and needs neither a local daemon nor host
-virtualisation, so it can run from a containerised worker.
+virtualization, so it can run from a containerized worker.
 
 Requires the ``sandbox-islo`` extra::
 
--- a/providers/common/ai/docs/toolsets.rst
+++ b/providers/common/ai/docs/toolsets.rst
@@ -793,6 +793,60 @@ Constructor parameters:
   guarantee this backend cannot make. Set ``"deny-all"`` after running
   ``sbx policy init deny-all``, or ``"allow-all"`` to state that egress is open.
 
+Islo backend
+^^^^^^^^^^^^
+
+:class:`~airflow.providers.common.ai.sandbox.IsloSandboxBackend` runs each
+sandbox in an `islo.dev <https://islo.dev>`__ microVM. Unlike ``sbx``, the
+worker talks to a hosted API and needs neither a local daemon nor host
+virtualisation, so it can run from a containerised worker.
+
+Requires the ``sandbox-islo`` extra::
+
+    pip install "apache-airflow-providers-common-ai[sandbox-islo]"
+
+.. code-block:: python
+
+    from airflow.providers.common.ai.sandbox import IsloSandboxBackend
+
+    SandboxToolset(IsloSandboxBackend(islo_conn_id="islo_default"))
+
+Credentials come from a generic Airflow connection, resolved lazily on first
+use, so the API key lives in your configured secrets backend rather than the
+worker environment:
+
+- ``password``: the Islo API key. Required.
+- ``host``: the compute URL. Optional.
+- Extra: optional ``base_url`` and ``timeout`` (request timeout in seconds).
+
+Constructor parameters:
+
+- ``islo_conn_id``: Connection ID. Default ``"islo_default"``. ``None`` lets the
+  SDK resolve credentials from its own environment variables.
+- ``image``, ``vcpus``, ``memory_mb``: image and sizing. ``None`` (default) uses
+  the server default for each.
+- ``delete_after``: Server-side TTL in seconds, after which the sandbox is
+  deleted even if the worker never got to destroy it. Default ``3600``. This is
+  the backstop the ``sbx`` backend lacks.
+
+``SandboxSpec.env`` is passed at creation, and ``block_network`` maps to the
+API's ``internet_enabled``. A per-domain ``allow_egress_to`` is refused: the API
+can turn outbound access on or off, not scope it to named hosts.
+
+File reads and writes use Islo's native streaming APIs. Directory listings and
+command-output bounding require common Unix command-line tools in the sandbox
+image, including ``sh``, ``mkfifo``, ``tail``, ``tee`` and a ``find``
+implementation with ``-printf`` support. Command output is capped inside the
+microVM before the SDK returns it to the worker, keeping the tail of each stream
+without storing the full output on disk.
+
+The backend enforces the command deadline itself because the API's
+``timeout_secs`` is only a hint. If no terminal state arrives by the deadline,
+the backend deletes the microVM before reporting a timeout; if deletion cannot
+be confirmed, the task fails instead of claiming that the sandbox stopped. The
+server-side ``delete_after`` policy remains the backstop for a worker killed
+mid-run.
+
 Bringing your own backend
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/providers/common/ai/pyproject.toml
+++ b/providers/common/ai/pyproject.toml
@@ -96,6 +96,7 @@ dependencies = [
     "apache-airflow-providers-git>=0.4.0",
     "pydantic-ai-skills>=1.2.0",
 ]
+"sandbox-islo" = ["islo>=0.3.9"]
 "avro" = [
     'fastavro>=1.10.0; python_version < "3.14"',
     'fastavro>=1.12.1; python_version >= "3.14"',
--- a/providers/common/ai/src/airflow/providers/common/ai/sandbox/__init__.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/sandbox/__init__.py
@@ -26,9 +26,13 @@ from airflow.providers.common.ai.sandbox.base import (
     SandboxSpec,
     SandboxTerminalError,
 )
+
+# SbxSandboxBackend only shells out to the `sbx` CLI (stdlib imports), so it is
+# always importable; IsloSandboxBackend needs the optional `islo` SDK.
 from airflow.providers.common.ai.sandbox.sbx import SbxSandboxBackend
 
 __all__ = [
+    "IsloSandboxBackend",
     "SandboxBackend",
     "SandboxError",
     "SandboxExecResult",
@@ -37,3 +41,15 @@ __all__ = [
     "SandboxTerminalError",
     "SbxSandboxBackend",
 ]
+
+
+def __getattr__(name: str):
+    if name == "IsloSandboxBackend":
+        try:
+            from airflow.providers.common.ai.sandbox.islo import IsloSandboxBackend
+        except ImportError as e:
+            from airflow.providers.common.compat.sdk import AirflowOptionalProviderFeatureException
+
+            raise AirflowOptionalProviderFeatureException(e)
+        return IsloSandboxBackend
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
@@ -24,10 +24,6 @@ import time
 from contextlib import contextmanager, suppress
 from typing import TYPE_CHECKING, Any
 
-from islo import Islo
-from islo.core.api_error import ApiError
-from islo.errors import NotFoundError
-
 from airflow.providers.common.ai.sandbox.base import (
     SandboxBackend,
     SandboxError,
@@ -41,6 +37,9 @@ from airflow.providers.common.compat.sdk import BaseHook
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+
+    from islo import Islo
+    from islo.errors import NotFoundError
 
     from airflow.providers.common.ai.sandbox.base import SandboxSpec
 
@@ -81,10 +80,16 @@ def _translate_islo_errors(operation: str) -> Iterator[None]:
         yield
     except SandboxError:
         raise
-    except ApiError as e:
-        status = f" (HTTP {e.status_code})" if e.status_code is not None else ""
-        raise SandboxTerminalError(f"Islo could not {operation}{status}.") from e
     except Exception as e:
+        try:
+            from islo.core.api_error import ApiError
+        except ImportError:
+            raise SandboxTerminalError(
+                'The Islo SDK is not installed. Install "apache-airflow-providers-common-ai[sandbox-islo]".'
+            ) from e
+        if isinstance(e, ApiError):
+            status = f" (HTTP {e.status_code})" if e.status_code is not None else ""
+            raise SandboxTerminalError(f"Islo could not {operation}{status}.") from e
         raise SandboxTerminalError(f"Islo could not {operation}: {type(e).__name__}.") from e
 
 
@@ -107,7 +112,7 @@ class IsloSandboxBackend(SandboxBackend):
     """
     Sandbox backend that runs agent commands in an `islo.dev <https://islo.dev>`__ microVM.
 
-    Islo is a hosted API with no local daemon or host-virtualisation requirement,
+    Islo is a hosted API with no local daemon or host-virtualization requirement,
     so this backend works from an Airflow worker running in a container.
     Credentials resolve lazily from an Airflow connection on first use.
 
@@ -158,6 +163,8 @@ class IsloSandboxBackend(SandboxBackend):
         if self._client is not None:
             return self._client
         with _translate_islo_errors("initialize its client"):
+            from islo import Islo
+
             if self._islo_conn_id is None:
                 self._client = Islo()
                 return self._client
@@ -193,27 +200,27 @@ class IsloSandboxBackend(SandboxBackend):
         return options
 
     def create(self, *, spec: SandboxSpec | None = None) -> str:
-        from islo.types import LifecyclePolicy
-
         if spec is not None and spec.allow_egress_to:
             raise SandboxTerminalError(
                 "The Islo backend cannot apply a per-domain egress allowlist; it can only turn "
                 "outbound access on or off. Drop allow_egress_to, or use a backend with "
                 "per-domain network rules."
             )
-        kwargs: dict[str, Any] = {
-            "internet_enabled": False if spec is None else not spec.block_network,
-            "lifecycle": LifecyclePolicy(delete_after=self._delete_after),
-        }
-        if self._image is not None:
-            kwargs["image"] = self._image
-        if self._vcpus is not None:
-            kwargs["vcpus"] = self._vcpus
-        if self._memory_mb is not None:
-            kwargs["memory_mb"] = self._memory_mb
-        if spec is not None and spec.env:
-            kwargs["env"] = dict(spec.env)
         with _translate_islo_errors("create a sandbox"):
+            from islo.types import LifecyclePolicy
+
+            kwargs: dict[str, Any] = {
+                "internet_enabled": False if spec is None else not spec.block_network,
+                "lifecycle": LifecyclePolicy(delete_after=self._delete_after),
+            }
+            if self._image is not None:
+                kwargs["image"] = self._image
+            if self._vcpus is not None:
+                kwargs["vcpus"] = self._vcpus
+            if self._memory_mb is not None:
+                kwargs["memory_mb"] = self._memory_mb
+            if spec is not None and spec.env:
+                kwargs["env"] = dict(spec.env)
             sandbox = self._get_client().sandboxes.create_sandbox(
                 name=_new_sandbox_name(),
                 request_options=self._request_options(timeout=_FILE_OP_TIMEOUT),
@@ -312,10 +319,13 @@ class IsloSandboxBackend(SandboxBackend):
 
     def read_file(self, sandbox: str, path: str, *, max_bytes: int) -> bytes:
         _validate_positive_finite(max_bytes, "max_bytes")
+        client = self._get_client()
+        from islo.errors import NotFoundError
+
         chunks = None
         data = bytearray()
         try:
-            chunks = self._get_client().sandboxes.download_file(
+            chunks = client.sandboxes.download_file(
                 sandbox,
                 path=path,
                 request_options=self._request_options(
@@ -372,8 +382,11 @@ class IsloSandboxBackend(SandboxBackend):
         return entries
 
     def destroy(self, sandbox: str) -> None:
+        client = self._get_client()
+        from islo.errors import NotFoundError
+
         try:
-            self._get_client().sandboxes.delete_sandbox(
+            client.sandboxes.delete_sandbox(
                 sandbox_name=sandbox,
                 request_options=self._request_options(timeout=_FILE_OP_TIMEOUT),
             )
--- a/providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
+++ b/providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
@@ -1,0 +1,384 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""islo.dev microVM backend for :class:`~airflow.providers.common.ai.toolsets.sandbox.SandboxToolset`."""
+
+from __future__ import annotations
+
+import math
+import shlex
+import time
+from contextlib import contextmanager, suppress
+from typing import TYPE_CHECKING, Any
+
+from islo import Islo
+from islo.core.api_error import ApiError
+from islo.errors import NotFoundError
+
+from airflow.providers.common.ai.sandbox.base import (
+    SandboxBackend,
+    SandboxError,
+    SandboxExecResult,
+    SandboxFileTooLargeError,
+    SandboxTerminalError,
+    _new_sandbox_name,
+    _validate_positive_finite,
+)
+from airflow.providers.common.compat.sdk import BaseHook
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from airflow.providers.common.ai.sandbox.base import SandboxSpec
+
+_TERMINAL_EXEC_STATUSES = frozenset({"completed", "failed", "timeout"})
+_POLL_INITIAL = 0.2
+_POLL_MAX = 2.0
+_POLL_BACKOFF = 1.5
+_FILE_OP_TIMEOUT = 120.0
+_HELPER_OUTPUT_CAP = 1024 * 1024
+_COMMAND_WRAPPER = """\
+umask 077
+dir="${TMPDIR:-/tmp}/airflow-sandbox-$$"
+mkdir "$dir" || exit 70
+trap 'rm -rf "$dir"' EXIT HUP INT TERM
+mkfifo "$dir/out" "$dir/err" "$dir/out-tail" "$dir/err-tail" || exit 70
+tail -c "$2" <"$dir/out-tail" >"$dir/out-result" &
+out_tail_pid=$!
+tail -c "$2" <"$dir/err-tail" >"$dir/err-result" &
+err_tail_pid=$!
+tee "$dir/out-tail" <"$dir/out" | wc -c >"$dir/out-count" &
+out_drain_pid=$!
+tee "$dir/err-tail" <"$dir/err" | wc -c >"$dir/err-count" &
+err_drain_pid=$!
+sh -lc "$1" >"$dir/out" 2>"$dir/err"
+status=$?
+wait "$out_drain_pid" "$err_drain_pid" "$out_tail_pid" "$err_tail_pid" || exit 70
+if [ "$(cat "$dir/out-count")" -gt "$2" ]; then printf '1\\n'; else printf '0\\n'; fi
+cat "$dir/out-result"
+if [ "$(cat "$dir/err-count")" -gt "$2" ]; then printf '1\\n' >&2; else printf '0\\n' >&2; fi
+cat "$dir/err-result" >&2
+exit "$status"
+"""
+
+
+@contextmanager
+def _translate_islo_errors(operation: str) -> Iterator[None]:
+    try:
+        yield
+    except SandboxError:
+        raise
+    except ApiError as e:
+        status = f" (HTTP {e.status_code})" if e.status_code is not None else ""
+        raise SandboxTerminalError(f"Islo could not {operation}{status}.") from e
+    except Exception as e:
+        raise SandboxTerminalError(f"Islo could not {operation}: {type(e).__name__}.") from e
+
+
+def _bound_result_stream(text: str, max_bytes: int, *, server_truncated: bool) -> tuple[str, bool]:
+    flag, separator, payload = text.partition("\n")
+    if separator and flag in {"0", "1"}:
+        truncated = flag == "1" or server_truncated
+    else:
+        payload = text
+        truncated = True
+
+    encoded = payload.encode("utf-8")
+    if len(encoded) > max_bytes:
+        payload = encoded[-max_bytes:].decode("utf-8", errors="ignore")
+        truncated = True
+    return payload, truncated
+
+
+class IsloSandboxBackend(SandboxBackend):
+    """
+    Sandbox backend that runs agent commands in an `islo.dev <https://islo.dev>`__ microVM.
+
+    Islo is a hosted API with no local daemon or host-virtualisation requirement,
+    so this backend works from an Airflow worker running in a container.
+    Credentials resolve lazily from an Airflow connection on first use.
+
+    Connection fields: ``password`` is the Islo API key (required), ``host`` the
+    compute URL (optional), and the extra may set ``base_url`` and ``timeout``
+    (request timeout in seconds).
+
+    File reads and writes use Islo's native streaming APIs. Directory listings
+    and command-output bounding require common Unix command-line tools in the
+    sandbox image, including ``sh``, ``mkfifo``, ``tail``, ``tee`` and a ``find``
+    implementation with ``-printf`` support.
+
+    :param islo_conn_id: Airflow connection ID for Islo. ``None`` lets the SDK
+        resolve credentials from its own environment variables (``ISLO_API_KEY``).
+    :param image: Sandbox image. ``None`` (default) uses the server default.
+    :param vcpus: Number of virtual CPUs. ``None`` uses the server default.
+    :param memory_mb: Memory in MB. ``None`` uses the server default.
+    :param delete_after: Server-side TTL in seconds after which the sandbox is
+        deleted even if the worker never got to destroy it. Default ``3600``.
+    """
+
+    name = "islo"
+
+    def __init__(
+        self,
+        islo_conn_id: str | None = "islo_default",
+        *,
+        image: str | None = None,
+        vcpus: int | None = None,
+        memory_mb: int | None = None,
+        delete_after: int = 3600,
+    ) -> None:
+        _validate_positive_finite(delete_after, "delete_after")
+        if vcpus is not None:
+            _validate_positive_finite(vcpus, "vcpus")
+        if memory_mb is not None:
+            _validate_positive_finite(memory_mb, "memory_mb")
+        if image == "":
+            raise ValueError("image must not be empty.")
+        self._islo_conn_id = islo_conn_id
+        self._image = image
+        self._vcpus = vcpus
+        self._memory_mb = memory_mb
+        self._delete_after = delete_after
+        self._client: Islo | None = None
+
+    def _get_client(self) -> Islo:
+        if self._client is not None:
+            return self._client
+        with _translate_islo_errors("initialize its client"):
+            if self._islo_conn_id is None:
+                self._client = Islo()
+                return self._client
+            conn = BaseHook.get_connection(self._islo_conn_id)
+            api_key = (conn.password or "").strip()
+            if not api_key:
+                raise SandboxTerminalError(
+                    f"Connection {self._islo_conn_id!r} has no password; set it to the Islo API key."
+                )
+            kwargs: dict[str, Any] = {"api_key": api_key}
+            if conn.host:
+                kwargs["compute_url"] = conn.host
+            extra = conn.extra_dejson
+            if extra.get("base_url"):
+                kwargs["base_url"] = extra["base_url"]
+            if extra.get("timeout") is not None:
+                try:
+                    request_timeout = float(extra["timeout"])
+                    _validate_positive_finite(request_timeout, "connection extra timeout")
+                except (TypeError, ValueError) as e:
+                    raise SandboxTerminalError(
+                        "The Islo connection extra timeout must be a positive finite number."
+                    ) from e
+                kwargs["timeout"] = request_timeout
+            self._client = Islo(**kwargs)
+            return self._client
+
+    @staticmethod
+    def _request_options(*, timeout: float, chunk_size: int | None = None) -> dict[str, int]:
+        options = {"timeout_in_seconds": max(1, math.ceil(timeout)), "max_retries": 0}
+        if chunk_size is not None:
+            options["chunk_size"] = chunk_size
+        return options
+
+    def create(self, *, spec: SandboxSpec | None = None) -> str:
+        from islo.types import LifecyclePolicy
+
+        if spec is not None and spec.allow_egress_to:
+            raise SandboxTerminalError(
+                "The Islo backend cannot apply a per-domain egress allowlist; it can only turn "
+                "outbound access on or off. Drop allow_egress_to, or use a backend with "
+                "per-domain network rules."
+            )
+        kwargs: dict[str, Any] = {
+            "internet_enabled": False if spec is None else not spec.block_network,
+            "lifecycle": LifecyclePolicy(delete_after=self._delete_after),
+        }
+        if self._image is not None:
+            kwargs["image"] = self._image
+        if self._vcpus is not None:
+            kwargs["vcpus"] = self._vcpus
+        if self._memory_mb is not None:
+            kwargs["memory_mb"] = self._memory_mb
+        if spec is not None and spec.env:
+            kwargs["env"] = dict(spec.env)
+        with _translate_islo_errors("create a sandbox"):
+            sandbox = self._get_client().sandboxes.create_sandbox(
+                name=_new_sandbox_name(),
+                request_options=self._request_options(timeout=_FILE_OP_TIMEOUT),
+                **kwargs,
+            )
+        return sandbox.name
+
+    def _await_exec(self, sandbox: str, exec_id: str, *, deadline: float) -> Any:
+        client = self._get_client()
+        interval = _POLL_INITIAL
+        while time.monotonic() < deadline:
+            remaining = deadline - time.monotonic()
+            with _translate_islo_errors("poll a sandbox command"):
+                result = client.sandboxes.get_exec_result(
+                    sandbox,
+                    exec_id,
+                    request_options=self._request_options(timeout=remaining),
+                )
+            if result.status in _TERMINAL_EXEC_STATUSES:
+                return result
+            time.sleep(min(interval, max(0.0, deadline - time.monotonic())))
+            interval = min(interval * _POLL_BACKOFF, _POLL_MAX)
+        return None
+
+    def _destroy_after_timeout(self, sandbox: str) -> None:
+        try:
+            self.destroy(sandbox)
+        except SandboxError as e:
+            raise SandboxTerminalError(
+                "The Islo command timed out and deletion of its sandbox could not be confirmed."
+            ) from e
+
+    def run_command(
+        self, sandbox: str, command: str, *, timeout: float, max_output_bytes: int
+    ) -> SandboxExecResult:
+        _validate_positive_finite(timeout, "timeout")
+        _validate_positive_finite(max_output_bytes, "max_output_bytes")
+        client = self._get_client()
+        deadline = time.monotonic() + timeout
+        with _translate_islo_errors("start a sandbox command"):
+            response = client.sandboxes.exec_in_sandbox(
+                sandbox,
+                command=["sh", "-c", _COMMAND_WRAPPER, "airflow-sandbox", command, str(max_output_bytes)],
+                timeout_secs=max(1, math.ceil(timeout)),
+                request_options=self._request_options(timeout=timeout),
+            )
+        result = self._await_exec(sandbox, response.exec_id, deadline=deadline)
+        if result is None:
+            self._destroy_after_timeout(sandbox)
+            return SandboxExecResult(
+                exit_code=-1, stdout="", stderr="", timed_out=True, sandbox_terminated=True
+            )
+
+        server_truncated = bool(getattr(result, "truncated", False))
+        stdout, out_truncated = _bound_result_stream(
+            result.stdout or "", max_output_bytes, server_truncated=server_truncated
+        )
+        stderr, err_truncated = _bound_result_stream(
+            result.stderr or "", max_output_bytes, server_truncated=server_truncated
+        )
+        if result.status == "timeout":
+            self._destroy_after_timeout(sandbox)
+            return SandboxExecResult(
+                exit_code=-1,
+                stdout=stdout,
+                stderr=stderr,
+                timed_out=True,
+                stdout_truncated=out_truncated,
+                stderr_truncated=err_truncated,
+                sandbox_terminated=True,
+            )
+        return SandboxExecResult(
+            exit_code=result.exit_code if result.exit_code is not None else -1,
+            stdout=stdout,
+            stderr=stderr,
+            stdout_truncated=out_truncated,
+            stderr_truncated=err_truncated,
+        )
+
+    def _run_helper(self, sandbox: str, script: str, *, operation: str) -> str:
+        result = self.run_command(
+            sandbox, script, timeout=_FILE_OP_TIMEOUT, max_output_bytes=_HELPER_OUTPUT_CAP
+        )
+        if result.timed_out or result.sandbox_terminated:
+            raise SandboxTerminalError(f"The sandbox was destroyed after it timed out while {operation}.")
+        if result.exit_code:
+            raise SandboxError(result.stderr.strip() or f"Could not {operation}.")
+        return result.stdout
+
+    def _raise_file_not_found(self, sandbox: str, path: str, error: NotFoundError) -> None:
+        with _translate_islo_errors("check a sandbox after a missing file response"):
+            self._get_client().sandboxes.get_sandbox(
+                sandbox, request_options=self._request_options(timeout=_FILE_OP_TIMEOUT)
+            )
+        raise SandboxError(f"{path!r} does not exist in the sandbox, or is not readable.") from error
+
+    def read_file(self, sandbox: str, path: str, *, max_bytes: int) -> bytes:
+        _validate_positive_finite(max_bytes, "max_bytes")
+        chunks = None
+        data = bytearray()
+        try:
+            chunks = self._get_client().sandboxes.download_file(
+                sandbox,
+                path=path,
+                request_options=self._request_options(
+                    timeout=_FILE_OP_TIMEOUT, chunk_size=min(65536, max_bytes + 1)
+                ),
+            )
+            for chunk in chunks:
+                data.extend(chunk[: max_bytes + 1 - len(data)])
+                if len(data) > max_bytes:
+                    raise SandboxFileTooLargeError(path, len(data), max_bytes)
+        except SandboxFileTooLargeError:
+            raise
+        except NotFoundError as e:
+            self._raise_file_not_found(sandbox, path, e)
+        except Exception:
+            with _translate_islo_errors("download a sandbox file"):
+                raise
+        finally:
+            close = getattr(chunks, "close", None)
+            if close is not None:
+                with suppress(Exception):
+                    close()
+        return bytes(data)
+
+    def write_file(self, sandbox: str, path: str, content: bytes) -> None:
+        quoted = shlex.quote(path)
+        self._run_helper(
+            sandbox,
+            f'mkdir -p -- "$(dirname -- {quoted})"',
+            operation=f"create the parent directory for {path!r}",
+        )
+        with _translate_islo_errors("upload a sandbox file"):
+            self._get_client().sandboxes.upload_file(
+                sandbox,
+                path=path,
+                file=("upload", content, "application/octet-stream"),
+                request_options=self._request_options(timeout=_FILE_OP_TIMEOUT),
+            )
+
+    def list_directory(self, sandbox: str, path: str) -> list[tuple[str, bool]]:
+        quoted = shlex.quote(path)
+        listing = self._run_helper(
+            sandbox,
+            f"find -- {quoted} -maxdepth 1 -mindepth 1 -printf '%y %f\\0'",
+            operation=f"list {path!r}",
+        )
+        entries: list[tuple[str, bool]] = []
+        for record in listing.split("\0"):
+            if not record:
+                continue
+            kind, _, name = record.partition(" ")
+            if name:
+                entries.append((name, kind == "d"))
+        return entries
+
+    def destroy(self, sandbox: str) -> None:
+        try:
+            self._get_client().sandboxes.delete_sandbox(
+                sandbox_name=sandbox,
+                request_options=self._request_options(timeout=_FILE_OP_TIMEOUT),
+            )
+        except NotFoundError:
+            return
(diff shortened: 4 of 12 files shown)


providers/common/ai/docs/toolsets.rst
```diff
@@ -793,6 +793,60 @@ Constructor parameters:
   guarantee this backend cannot make. Set ``"deny-all"`` after running
   ``sbx policy init deny-all``, or ``"allow-all"`` to state that egress is open.
 
+Islo backend
+^^^^^^^^^^^^
+
+:class:`~airflow.providers.common.ai.sandbox.IsloSandboxBackend` runs each
```
kaxil, line 799: Same as other one, working demo would be ideal to see

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+out_tail_pid=$!
+tail -c "$2" <"$dir/err-tail" >"$dir/err-result" &
+err_tail_pid=$!
+tee "$dir/out-tail" <"$dir/out" | wc -c >"$dir/out-count" &
+out_drain_pid=$!
+tee "$dir/err-tail" <"$dir/err" | wc -c >"$dir/err-count" &
+err_drain_pid=$!
+sh -lc "$1" >"$dir/out" 2>"$dir/err"
+status=$?
+wait "$out_drain_pid" "$err_drain_pid" "$out_tail_pid" "$err_tail_pid" || exit 70
```
kaxil, line 68: This `wait` only returns once the tee/wc drains hit EOF, and any process the command backgrounds inherits the fifo write end, so EOF never comes. Running the wrapper under `sh` in a Debian slim image, `sleep 25 & echo started` took 25s to return and `nohup sleep 300 & echo server-started` never returned at all. Past the deadline `_destroy_after_timeout` deletes the microVM, so the model is told its sandbox was replaced and its files are gone, for a command whose foreground part finished instantly. `sbx.py` hits the same wall and bounds it with daemon drain threads plus a deadline-limited join; the wrapper needs a similar bound, such as a watchdog that kills the drains once `$status` is captured.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+    from airflow.providers.common.ai.sandbox.base import SandboxSpec
+
+_TERMINAL_EXEC_STATUSES = frozenset({"completed", "failed", "timeout"})
+_POLL_INITIAL = 0.2
+_POLL_MAX = 2.0
+_POLL_BACKOFF = 1.5
+_FILE_OP_TIMEOUT = 120.0
+_HELPER_OUTPUT_CAP = 1024 * 1024
+_COMMAND_WRAPPER = """\
+umask 077
```
kaxil, line 53: `umask 077` guards the scratch dir but it is still in effect for `sh -lc "$1"`, so it changes the mode of everything the agent creates. Same image and command: through the wrapper `touch f; mkdir d` gives `-rw-------` and `drwx------`; without it, `-rw-r--r--` and `drwxr-xr-x`. Resetting the umask before running the command, or dropping it and doing `chmod 700 "$dir"` after `mkdir`, keeps the scratch dir private without that side effect.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+dir="${TMPDIR:-/tmp}/airflow-sandbox-$$"
+mkdir "$dir" || exit 70
+trap 'rm -rf "$dir"' EXIT HUP INT TERM
```
kaxil, line 56: A POSIX trap handler with no `exit` hands control back to the script, so on `TERM` this removes `$dir` and then keeps running. Sending `TERM` mid-command gave me stdout of just the flag line, stderr filled with `cat: .../out-count: No such file or directory` and `[: Illegal number:`, and an exit status of 0. Whether that reaches the model depends on the status Islo reports for a `timeout_secs` kill, but `trap 'rm -rf "$dir"; exit 143' HUP INT TERM`, keeping the bare `EXIT` trap for cleanup, would take the question off the table.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+mkfifo "$dir/out" "$dir/err" "$dir/out-tail" "$dir/err-tail" || exit 70
+tail -c "$2" <"$dir/out-tail" >"$dir/out-result" &
```
kaxil, line 58: `tail -c` cuts at a byte boundary, so the first record it keeps is normally a fragment, and neither `_bound_result_stream` nor `truncate_output` drops it. Pushing 1000 numbered lines through the wrapper with a 100 byte cap, the model ends up with `88` on a line of its own, which is the tail of `line988`. `output.py`'s own header says truncation never emits a partial line, and `sbx.py`'s drain drops the leading partial line for exactly this reason. The second cut at line 106 has the same gap, and `list_directory` shares the path, where a mid-record cut can yield a mangled entry name.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+                    raise SandboxTerminalError(
+                        "The Islo connection extra timeout must be a positive finite number."
+                    ) from e
+                kwargs["timeout"] = request_timeout
+            self._client = Islo(**kwargs)
+            return self._client
+
+    @staticmethod
+    def _request_options(*, timeout: float, chunk_size: int | None = None) -> dict[str, int]:
+        options = {"timeout_in_seconds": max(1, math.ceil(timeout)), "max_retries": 0}
```
kaxil, line 197: `max_retries: 0` also lands on `delete_sandbox`, which is idempotent, and that is the one call whose failure `_destroy_after_timeout` converts into a task failure. So a single transient error during cleanup fails a run whose command merely timed out, even though `delete_after` is going to reclaim the microVM regardless. `sbx.py` only logs a warning on the same path, and it has no TTL to fall back on. Letting the delete keep the SDK's retries, at least on the cleanup path, would fit the backstop this backend already has.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+                "lifecycle": LifecyclePolicy(delete_after=self._delete_after),
+            }
+            if self._image is not None:
+                kwargs["image"] = self._image
+            if self._vcpus is not None:
+                kwargs["vcpus"] = self._vcpus
+            if self._memory_mb is not None:
+                kwargs["memory_mb"] = self._memory_mb
+            if spec is not None and spec.env:
+                kwargs["env"] = dict(spec.env)
```
kaxil, line 223: Is `env` set at creation visible to later `exec_in_sandbox` calls, or only to the sandbox's own init? `exec_in_sandbox` takes its own `env`, so carrying `spec.env` on the backend and passing it per exec would remove the doubt. Worth pinning down, because `base.py` treats silently dropping a `SandboxSpec` field as a contract violation, and neither the unit tests nor the system test ever set one.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from islo import Islo
+    from islo.errors import NotFoundError
+
+    from airflow.providers.common.ai.sandbox.base import SandboxSpec
+
+_TERMINAL_EXEC_STATUSES = frozenset({"completed", "failed", "timeout"})
```
kaxil, line 46: The SDK types the exec `status` as a plain `str` with `extra="allow"`, and the vendor's other status enums carry values like `cancelled` and `dead`, so these three may not be the full vocabulary. Anything outside the set reads as still-running here, so `_await_exec` polls to the deadline and then destroys the sandbox and reports a timeout. Treating an unrecognised status as terminal-failed instead would keep a mislabelled status from costing the agent its files.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,397 @@
...
+            if self._image is not None:
+                kwargs["image"] = self._image
+            if self._vcpus is not None:
+                kwargs["vcpus"] = self._vcpus
+            if self._memory_mb is not None:
+                kwargs["memory_mb"] = self._memory_mb
+            if spec is not None and spec.env:
+                kwargs["env"] = dict(spec.env)
+            sandbox = self._get_client().sandboxes.create_sandbox(
+                name=_new_sandbox_name(),
```
kaxil, line 225: `_new_sandbox_name()` is generated inline, so if `create_sandbox` raises after the server has already provisioned the microVM (a response timeout, a reset, a 5xx after creation), the handle is gone and nothing can delete it or even log which sandbox leaked. `sbx.py` binds the name first and best-effort removes it under `except BaseException`. `delete_after` does bound the damage, but it is documented as the backstop for a worker killed mid-run, and here the worker is alive and able to clean up immediately.

providers/common/ai/docs/toolsets.rst
```diff
@@ -793,6 +793,60 @@ Constructor parameters:
   guarantee this backend cannot make. Set ``"deny-all"`` after running
   ``sbx policy init deny-all``, or ``"allow-all"`` to state that egress is open.
 
+Islo backend
```
kaxil, line 796: The `sbx` warning box just above (line 753) still says a hosted backend plugs in through `SandboxBackend` but none ships with the provider yet, and `sbx.py`'s class docstring says the same; both are now false. The credentials paragraph below also reads as unconditional, but `islo_conn_id=None` sends the SDK to its own environment variable, which is the worker environment the sentence says the key stays out of, and it is the path this PR's own system test and README use.

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def test_api_failure_is_terminal(self):
+        backend, client = _backend_with_client()
+        client.sandboxes.create_sandbox.side_effect = ApiError(status_code=503)
+
+        with pytest.raises(SandboxTerminalError, match="HTTP 503"):
+            backend.create()
+
+
+class TestRunCommand:
```
kaxil, line 201: `_backend_with_client` mocks `exec_in_sandbox` outright, so `_COMMAND_WRAPPER` is never executed by anything except the system test, which needs a live API key and so will not run in ordinary CI. It is the riskiest code in the PR (four background jobs, four fifos, a trap, and a `wait`), and running it through a local `sh` subprocess needs no Islo access at all. A test like that would have caught the drain that blocks on a backgrounded child, the umask reaching the agent's command, and the partial leading line.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Confirmed, and it reproduced in a real microVM too: `sleep 20 & echo started` took 20.1s, because the backgrounded child inherits the fifo write end so the drains never see end-of-input.

Rather than bound the drains, the wrapper no longer has any. Each stream is redirected to a scratch file and the last `max_output_bytes` are emitted with `tail -c` afterwards, so only the foreground command is waited on — the fifos, the `tee`/`wc` drains and the `wait` are all gone. Same command is now 0.3s against a live sandbox, and `nohup sleep 300 & echo server-started` returns immediately.

The trade is that total output lands on the sandbox’s own ephemeral disk instead of streaming through a fixed-size window; that is called out in the class docstring and the docs. Covered by `test_a_backgrounded_process_does_not_hold_the_command_open` and `test_a_long_lived_daemon_does_not_hold_the_command_open`, which run the wrapper through a local `sh`.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Right, and it was visible exactly as you describe — `touch f; mkdir d` came back `-rw-------`/`drwx------` through the wrapper versus `-rw-r--r--`/`drwxr-xr-x` without it.

Replaced with `mkdir -m 700 "$dir"`, which is atomic and never puts a umask in front of the agent’s command. `test_does_not_change_the_permissions_of_what_the_agent_creates` asserts the normal modes, and `test_the_scratch_directory_is_private_while_in_use_and_gone_after` stats the scratch directory mid-run to confirm it is still `0700` and removed afterwards.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Fixed as suggested: `trap 'rm -rf "$dir"; exit 143' HUP INT TERM` with the bare `EXIT` trap kept for cleanup.

Worth noting the old behaviour was environment-dependent — signalling the process group locally gave exit 70 from the failed `wait` rather than your 0 — but either way it cleaned up and then kept running with its scratch files gone. It is now an unambiguous 143 with no output. `test_a_terminated_command_exits_nonzero_without_emitting_garbage` covers it by signalling the whole process group, which is what stopping the microVM looks like from inside.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Reproduced exactly: with a 100 byte cap the model got `88`, the tail of `line988`.

Truncation now drops the leading partial line when it cuts, the same way `sbx.py`’s drain does, so `output.py`’s promise holds here too. Live check with a 100 byte cap now starts at `line1990` and ends at `line2000`.

The `list_directory` path you flagged needed separate handling, since its records are NUL-separated and have no newline for the cut to align on: `_run_helper` now returns the `SandboxExecResult` and `list_directory` drops a leading partial record when the stream was truncated. `test_truncation_never_emits_a_partial_leading_line` and `test_keeps_the_tail_and_never_a_partial_leading_line` cover the command path.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Agreed. `_request_options` takes a `max_retries` argument now and `destroy` passes `2`, so the one call that can strand a microVM is no longer the one call that never retries.

`_destroy_after_timeout` also no longer converts a failed cleanup into a task failure — it warns and leaves `delete_after` to reclaim the sandbox, matching `sbx.py`’s handling of the same path. `test_delete_keeps_the_sdk_retries` and `test_timeout_cleanup_failure_warns_and_leaves_the_ttl_to_reclaim` cover both.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Pinned it down against a live microVM: environment set at creation **is** visible to later `exec_in_sandbox` calls, including through the wrapper’s `sh -lc`. So the spec is honored for the sandbox’s whole life and the create-time call is enough.

Left it at creation rather than moving to per-exec, since one authoritative place beats repeating it on every call, and added a comment recording that this was verified rather than assumed. `test_spec_env_is_forwarded_so_it_is_never_silently_dropped` guards the `base.py` contract, and the E2E now asserts `$SPEC_MARKER` is readable from inside the sandbox.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Good catch — the SDK types `status` as a plain `str` on a model with `extra="allow"`, so the vocabulary is genuinely open.

Inverted the check: `_RUNNING_EXEC_STATUSES` lists the statuses that mean "not finished yet" (`pending`, `queued`, `starting`, `running`), and anything else is terminal. An unknown value now surfaces a failure instead of polling to the deadline and destroying the sandbox. `test_an_unrecognised_status_is_terminal_rather_than_still_running` parametrises `cancelled`, `dead` and an unseen value, and `test_in_flight_statuses_keep_polling` pins the other side so the two cannot drift.

Related finding from the live run: the server does not appear to enforce `timeout_secs` at all — `sleep 30` with `timeout_secs=2` ran the full 30s and reported `completed` with exit 0. The client-side polling deadline is what actually bounds a command, and that path already destroys the sandbox and reports a timeout.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Fixed. The name is generated before the call and a `BaseException` handler best-effort deletes it, so a 5xx or reset after the server provisioned the microVM leaves a handle that can still reclaim it — the `sbx.py` pattern.

`test_a_failed_create_deletes_the_name_it_had_already_bound` asserts the delete targets the same name that was requested, and `test_a_cleanup_failure_does_not_mask_the_original_create_error` makes sure the original error is what the caller sees.

providers/common/ai/docs/toolsets.rst
zozo123: Both corrected.

The `sbx` warning box and the `SbxSandboxBackend` docstring now point at `IsloSandboxBackend` instead of saying no hosted backend ships.

The credentials paragraph now reads "By default, credentials come from a generic Airflow connection…" and spells out that `islo_conn_id=None` hands resolution to the SDK, which reads `ISLO_API_KEY` from the worker environment — convenient for a local trial, but outside the secrets backend, so a connection is preferred in a deployment. The system test README already used that path and now the docs match it.

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
zozo123: Agreed, and this was the gap that mattered — every one of your findings was real and none of the 45 tests could see them, because `exec_in_sandbox` was mocked.

Added `TestCommandWrapper`, which runs `_COMMAND_WRAPPER` through a local `sh` with no Islo access, covering the backgrounded child, the umask, the signal path, whole-record truncation, the scratch directory mode and its cleanup, and stream/exit-status separation. I wrote them against the old wrapper first and confirmed they fail on it.

65 tests now pass, up from 45.

providers/common/ai/docs/toolsets.rst
zozo123: Here it is, against a real Islo microVM:

SandboxToolset on an islo.dev microVM

A deterministic pydantic-ai `FunctionModel` drives it, so the run is reproducible and needs no LLM. Each of the four tools is called in turn, and the steps double as evidence for the findings in your other review: the backgrounded process returns immediately, the agent's files keep normal permissions, deny-all egress holds, and a nonzero exit comes back as a retryable error. Source and recording are on [this branch](https://github.com/zozo123/airflow/tree/demo/islo-sandbox) — deliberately not part of the PR.

The shipped system test (`example_sandbox_toolset_islo.py`) passes against the same live backend:

```console
$ ISLO_API_KEY=... pytest --system providers/common/ai/tests/system/common/ai/example_sandbox_toolset_islo.py
providers/common/ai/tests/system/common/ai/example_sandbox_toolset_islo.py::test_run PASSED [100%]
1 passed in 11.25s
```

zozo123:
Rebased onto current `main` (the `uv.lock` conflict is resolved — regenerated with the pinned `uv` 0.12.10 so the diff stays at the 19 lines the new extra needs) and pushed fixes for every thread from @kaxil's review. Replies are inline on each one.

SandboxToolset on an islo.dev microVM

### The main change

The output-bounding wrapper is much smaller. It used four background jobs, four fifos and a `wait`, and that `wait` could only finish once the capture reached end-of-input — so a command that backgrounds a process stalled until the child exited. This was not theoretical: `sleep 20 & echo started` took **20.1s** in a real microVM, and past the deadline the backend destroys the sandbox, so the agent lost its files over a command whose foreground part had already finished.

Each stream is now redirected to a scratch file and the last `max_output_bytes` emitted with `tail -c`, so only the foreground command is waited on. The same command is **0.3s** live. The fifos, drains and `wait` are gone, along with the `0`/`1` header protocol — the wrapper is asked for one byte over the budget, so an over-budget stream is itself the proof that bytes were dropped. The trade is that total output lands on the sandbox's own ephemeral disk rather than streaming through a fixed window, which is now stated in the docstring and the docs.

Also fixed: `mkdir -m 700` instead of a `umask` left in force for the agent's command; signal traps that `exit` instead of cleaning up and carrying on; truncation that drops a partial leading line (and a partial leading *record* in `list_directory`, which has no newline to align on); the sandbox name bound before creation with best-effort delete if creation fails ambiguously; retries kept on `delete_sandbox`; a failed post-timeout cleanup downgraded from task failure to a warning, since `delete_after` reclaims it; and unrecognised exec statuses treated as terminal rather than still-running.

### Two things the live runs turned up

- **`SandboxSpec.env` set at creation is visible to later execs**, including through the wrapper's `sh -lc`. That answers the open question on that thread — the create-time call is sufficient, and there is now a test so it cannot be dropped silently.
- **The server does not appear to enforce `timeout_secs`.** `sleep 30` with `timeout_secs=2` ran the full 30s and reported `completed` with exit 0. The client-side polling deadline is what actually bounds a command, and that path already destroys the sandbox and reports a timeout, so behaviour is correct — but the `"timeout"` status is not something to rely on.

### Verification

- **65 unit tests pass**, up from 45. The new `TestCommandWrapper` executes `_COMMAND_WRAPPER` through a local `sh` with no API key, so the riskiest code is covered in ordinary CI — that was the gap that let all of these through. I wrote those tests against the old wrapper first and confirmed they fail on it.
- **The shipped system test passes** against a live microVM, as does a broader end-to-end pass covering spec env, backgrounded commands, permissions, truncation, persistence, deny-all egress, the timeout path and idempotent teardown.
- Provider still imports without the optional SDK, and first use raises the actionable "install the extra" error.
- Ruff check and format clean at the repo's pinned version. No sandboxes were left behind by any of the runs.

providers/common/ai/docs/toolsets.rst
zozo123: Here is the Airflow UI view from a fresh UI-triggered run against a real Islo microVM using Islo SDK 0.3.19.

The deterministic `FunctionModel` needs no LLM or model credentials. The task log shows all five sandbox-boundary steps: `write_file`, `run_command` validating file persistence + `SandboxSpec.env` propagation + arithmetic, `read_file`, the expected exit-code-3/stderr path, `list_directory`, and teardown. The UI-triggered Dag and task both finished **Success**.

Airflow Islo sandbox E2E

The recording-only [Dag and browser script](https://github.com/zozo123/airflow/tree/demo/islo-sandbox) are kept off the PR branch.

---
Drafted-by: Codex (GPT-5.6) (no human review before posting)

zozo123:
### Working demo: Airflow UI driving an Islo microVM

Fresh UI-triggered run against a real Islo microVM, using Islo SDK 0.3.19.

Airflow Islo sandbox E2E

The GIF walks through the Dags list, the Trigger dialog, the run, and the task log. A deterministic pydantic-ai `FunctionModel` drives it, so the run is reproducible and needs no LLM or model credentials.

The task log shows every sandbox-boundary step:

1. `write_file` — native streaming upload into the microVM
2. `run_command` — file persistence, `SandboxSpec.env` propagation (`SPEC_MARKER=forwarded-at-create`), and arithmetic
3. `read_file` — state persists across tool calls
4. `run_command` — the expected exit-code-3 / stderr failure path
5. `list_directory` — the persistent workspace is visible

...then teardown. The Dag run and task both finished **Success** (11.3s).

The recording-only Dag and browser script live on [this branch](https://github.com/zozo123/airflow/tree/demo/islo-sandbox) and are deliberately not part of the PR.

---
Drafted-by: Claude (Sonnet 4.5) (no human review before posting)

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+# end-of-input would block until that process exits -- ``sleep 20 & echo
+# started`` took 20s in a real microVM before this. Redirecting to files means
+# only the foreground command is waited on. The cost is that the scratch file
+# grows with total output, on the sandbox's own ephemeral disk.
+_COMMAND_WRAPPER = """\
+dir="${TMPDIR:-/tmp}/airflow-sandbox-$$"
+mkdir -m 700 "$dir" || exit 70
+trap 'rm -rf "$dir"' EXIT
+trap 'rm -rf "$dir"; exit 143' HUP INT TERM
+sh -lc "$1" >"$dir/out" 2>"$dir/err"
```
kaxil, line 77: Following up on the `spec.env` thread above, which I think is only half closed. The live check used `SPEC_MARKER`, a name no profile touches, so it never exercised a collision. Because this is a *login* shell, `/etc/profile` runs after the create-time env is already in place and wins for anything it also exports. In `debian:bookworm-slim` with `PATH=/spec/wins` in the process env, `sh -lc` reports `PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games`; `alpine:3` behaves the same; `sh -c` keeps the spec value. So `SandboxSpec.env={"PATH": ...}` is silently dropped on the two most common image families, and `base.py` calls a silently-dropped spec field a contract violation. `sbx` escapes this by appending its exports to the end of `/etc/profile`, so they win. This backend has a native `env=` on `exec_in_sandbox` and does not need `-l` for env at all, so either drop `-l` or re-export the spec vars inside the wrapper after the profile has run. `toolsets.rst:849` states the guarantee flatly too.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+                    raise SandboxTerminalError(
+                        "The Islo connection extra timeout must be a positive finite number."
+                    ) from e
+                kwargs["timeout"] = request_timeout
+            self._client = Islo(**kwargs)
+            return self._client
+
+    @staticmethod
+    def _request_options(
+        *, timeout: float, chunk_size: int | None = None, max_retries: int = 0
```
kaxil, line 213: `max_retries` defaults to 0 and only `destroy` overrides it, so six of the seven `_request_options` call sites (253, 271, 315, 363, 378, 412) still have no retry at all. The poll loop is where that costs most: `_translate_islo_errors` sits inside the `while`, so one 502 or reset on any poll leaves the loop as a `SandboxTerminalError` and fails the task even with the whole deadline left, and with `_POLL_INITIAL`/`_POLL_MAX` a 300s command makes a few hundred of those requests. Same line has a second edge: `timeout=remaining` floors at `max(1, ceil(remaining))`, so the last poll before the deadline gets a 1 second HTTP budget, and an API slower than that turns a command timeout into a task failure rather than the timeout result the code below is trying to return. Continuing to poll on a transient error instead of translating it would fix both.

providers/common/ai/docs/toolsets.rst
```diff
@@ -807,6 +808,64 @@ Constructor parameters:
...
+image: ``sh``, ``tail`` and a ``find`` implementation with ``-printf`` support.
+Command output is capped inside the microVM before the SDK returns it to the
+worker, keeping the tail of each stream -- where a traceback and the exit status
+live -- rather than the head that the vendor's own 1 MB cap would keep. Each
+stream is captured to a scratch file in the sandbox first, so total output is
+bounded by the sandbox's own ephemeral disk rather than by worker memory.
+
+The backend enforces the command deadline itself because the API's
+``timeout_secs`` is only a hint. If no terminal state arrives by the deadline,
+the backend deletes the microVM before reporting a timeout; if deletion cannot
```
kaxil, line 864: This describes the pre-fix behaviour and is now inverted on both halves. `_destroy_after_timeout` warns and returns, and `run_command` then returns `sandbox_terminated=True`, so an unconfirmed deletion does not fail the task, and the result does claim the sandbox stopped. Looks like it was just missed when the cleanup path changed.

providers/common/ai/docs/toolsets.rst
```diff
@@ -807,6 +808,64 @@ Constructor parameters:
...
+
+    pip install "apache-airflow-providers-common-ai[sandbox-islo]"
+
+.. code-block:: python
+
+    from airflow.providers.common.ai.sandbox import IsloSandboxBackend
+
+    SandboxToolset(IsloSandboxBackend(islo_conn_id="islo_default"))
+
+By default, credentials come from a generic Airflow connection, resolved lazily
```
kaxil, line 829: "a generic Airflow connection" is all the guidance there is on what to create, and `provider.yaml`'s `connection-types:` has no Islo entry, so there is no hook, no `ui-field-behaviour`, no `connections/*.rst` page and no toctree entry. Every other connection-backed integration in this provider ships all four, which is what gives them a labelled form in Add Connection; here a reader has to land on core's built-in `generic` type and then know that `password` means the API key and `host` means the compute URL. Is the connection worth keeping in this PR at all? Ambient credentials only, which `islo_conn_id=None` already does, would be a smaller surface and leaves the conn type to whoever can own it properly; the SDK reads `ISLO_BASE_URL` and `ISLO_COMPUTE_URL` as well as `ISLO_API_KEY`, so that path also covers the endpoint config the connection carries in `host` and `extra.base_url`. If the connection stays, it wants the `connection-types` entry and a docs page here rather than later.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+                sandbox,
+                path=path,
+                request_options=self._request_options(
+                    timeout=_FILE_OP_TIMEOUT, chunk_size=min(65536, max_bytes + 1)
+                ),
+            )
+            for chunk in chunks:
+                data.extend(chunk[: max_bytes + 1 - len(data)])
+                if len(data) > max_bytes:
+                    raise SandboxFileTooLargeError(path, len(data), max_bytes)
```
kaxil, line 385: `len(data)` is always `max_bytes + 1` here, since the loop raises the moment it crosses, and the toolset renders both numbers through `format_size` into the `ModelRetry` text. At the default `max_read_bytes` of 5 MB the model is told `'x' is 5.0MB, over the 5.0MB read limit`, and the same collapse happens at 256 KB and at 50 KB. `base.py` avoids it by reporting `max(size, len(data))` from the guest's own `stat`, so this is a regression from going native. The SDK has no stat or file-info endpoint, so rather than inventing a number I would say the file is larger than the limit and drop the size. `test_oversized_read_stops_and_closes_the_stream` only checks the exception type, so nothing pins this either way.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+        )
+        if result.timed_out or result.sandbox_terminated:
+            raise SandboxTerminalError(f"The sandbox was destroyed after it timed out while {operation}.")
+        if result.exit_code:
+            raise SandboxError(result.stderr.strip() or f"Could not {operation}.")
+        return result
+
+    def _raise_file_not_found(self, sandbox: str, path: str, error: NotFoundError) -> None:
+        with _translate_islo_errors("check a sandbox after a missing file response"):
+            self._get_client().sandboxes.get_sandbox(
```
kaxil, line 362: This treats "`get_sandbox` did not raise" as "the sandbox is healthy, so the path is missing", and returns the recoverable `SandboxError` that the toolset turns into a `ModelRetry`. But `SandboxResponse` carries a required `status` plus `deleted_at` and `creation_phase`, and `LifecyclePolicy` offers `pause_after` / `pause_after_idle` / `auto_resume`, so a paused or stopped sandbox is a real state that still answers 200. The model would then spend its retries on file operations against a sandbox that cannot serve them, rather than getting the terminal error that lets Airflow retry the task. Worth reading `status` here. `create` at line 251 receives the same object and ignores it too, and those are the only two places a `SandboxResponse` reaches this module.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+                "The Islo backend cannot apply a per-domain egress allowlist; it can only turn "
+                "outbound access on or off. Drop allow_egress_to, or use a backend with "
+                "per-domain network rules."
+            )
+        with _translate_islo_errors("create a sandbox"):
+            from islo.types import LifecyclePolicy
+
+            kwargs: dict[str, Any] = {
+                "internet_enabled": False if spec is None else not spec.block_network,
+                "lifecycle": LifecyclePolicy(delete_after=self._delete_after),
```
kaxil, line 232: This is set once at create and never renewed, and the SDK names the idle variant separately (`pause_after_idle` alongside `pause_after`), which reads like `delete_after` counts from creation rather than from last use. If that is right, a run past the hour has its still-in-use microVM deleted under it and the next tool call fails, because one sandbox backs the whole agent run. The docs sell the TTL purely as the backstop `sbx` lacks. Which semantic is it, and would `pause_after_idle` plus `auto_resume` be the better default for a sandbox that is idle between tool calls anyway?

providers/common/ai/docs/toolsets.rst
```diff
@@ -807,6 +808,64 @@ Constructor parameters:
...
+  deleted even if the worker never got to destroy it. Default ``3600``. This is
+  the backstop the ``sbx`` backend lacks.
+
+``SandboxSpec.env`` is passed at creation, and ``block_network`` maps to the
+API's ``internet_enabled``. A per-domain ``allow_egress_to`` is refused: the API
+can turn outbound access on or off, not scope it to named hosts.
+
+File reads and writes use Islo's native streaming APIs. Directory listings and
+command-output bounding require common Unix command-line tools in the sandbox
+image: ``sh``, ``tail`` and a ``find`` implementation with ``-printf`` support.
```
kaxil, line 855: `-printf` is GNU-only, and this is the backend that exposes `image=`, so it is worth naming the families the way the `sbx` section does a few paragraphs up ("any Debian or Ubuntu based image, including `python:*-slim`"). Measured: `debian:bookworm-slim` works, `alpine:3` does not, with busybox `find` exiting 1 and writing 2395 bytes of usage text to stderr, which `_run_helper` hands straight to the model as the `ModelRetry` body (the helper cap is 1 MB, so nothing trims it). Saying whether the server-default image qualifies would help too.

providers/common/ai/docs/toolsets.rst
```diff
@@ -807,6 +808,64 @@ Constructor parameters:
...
+Command output is capped inside the microVM before the SDK returns it to the
+worker, keeping the tail of each stream -- where a traceback and the exit status
+live -- rather than the head that the vendor's own 1 MB cap would keep. Each
```
kaxil, line 858: Where is the vendor's 1 MB head-keeping cap documented? I could not find a cap or a truncation direction anywhere in `islo==0.3.19`, only the bare `truncated: bool` on `ExecResultResponse`. It matters past the docs: `list_directory` drops the *leading* record when `stdout_truncated` is set, but that flag is also set by the server, and if the server keeps the head then the mangled record is the trailing one, so the code would drop a good entry and report a bad one. Nothing covers that branch either, since no test sets `stdout_truncated` on the `list_directory` path.

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
```diff
@@ -0,0 +1,453 @@
...
+                request_options=self._request_options(timeout=timeout),
+            )
+        result = self._await_exec(sandbox, response.exec_id, deadline=deadline)
+        if result is None:
+            self._destroy_after_timeout(sandbox)
+            return SandboxExecResult(
+                exit_code=-1, stdout="", stderr="", timed_out=True, sandbox_terminated=True
+            )
+
+        server_truncated = bool(getattr(result, "truncated", False))
```
kaxil, line 324: `ExecResultResponse.truncated` is one bool for the whole result and it is fed to both streams, so a command whose stdout the server trimmed comes back with `stderr_truncated=True` on an empty stderr, and the toolset then appends a truncation note to a stream that lost nothing. `sbx` tracks the two separately.

providers/common/ai/docs/toolsets.rst
```diff
@@ -807,6 +808,64 @@ Constructor parameters:
   guarantee this backend cannot make. Set ``"deny-all"`` after running
   ``sbx policy init deny-all``, or ``"allow-all"`` to state that egress is open.
 
+Islo backend
+^^^^^^^^^^^^
+
+:class:`~airflow.providers.common.ai.sandbox.IsloSandboxBackend` runs each
```
kaxil, line 814: `IsloSandboxBackend` leaves the package only through `__getattr__`, and sphinx-autoapi reads the AST, so this target does not exist and the role renders as literal text (`nitpicky` is not set, so there is no warning to catch it either). The two other lazily-exported classes in this provider are referenced through their submodule for the same reason, on lines 32 and 34: `...toolsets.sql.SQLToolset` and `...toolsets.mcp.MCPToolset`. `main` has no package-path reference to a lazy name. `...sandbox.islo.IsloSandboxBackend` resolves. Same for line 768 and the new reference in `sbx.py`.

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
```diff
@@ -0,0 +1,687 @@
...
+
+        # The wrapper is asked for cap+1 bytes, so the backend can tell the
+        # stream was over budget; the first record must still be whole.
+        assert len(result.stdout.encode()) == cap + 1
+        payload, truncated = _bound_result_stream(result.stdout, cap, server_truncated=False)
+        assert truncated
+        assert payload.endswith("line1000\n")
+        assert all(line.startswith("line") for line in payload.splitlines())
+
+    def test_a_terminated_command_exits_nonzero_without_emitting_garbage(self):
```
kaxil, line 671: This one does not discriminate. I extracted `_COMMAND_WRAPPER` and ran three variants under the same `killpg(SIGTERM)`: as shipped, with the `HUP INT TERM` trap deleted outright, and with the trap present but its `exit 143` removed, which is the exact "cleans up and then falls through" bug the comment below names. All three gave a nonzero returncode and an empty stdout, on macOS and on Linux dash, so all three pass these two assertions. `status=$?` captures the foreground command's own signal death (143) before the trap runs, so falling through only adds stderr noise to a nonzero exit. The trap fix from the last round is real, but nothing in the suite would catch its removal.

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
```diff
@@ -0,0 +1,687 @@
...
+
+def _exec_result(status="completed", exit_code=0, stdout="", stderr="", truncated=False):
+    return SimpleNamespace(
+        status=status, exit_code=exit_code, stdout=stdout, stderr=stderr, truncated=truncated
+    )
+
+
+def _backend_with_client(**kwargs) -> tuple[IsloSandboxBackend, mock.MagicMock]:
+    backend = IsloSandboxBackend(**kwargs)
+    client = mock.MagicMock(spec=["sandboxes"])
```
kaxil, line 64: A string-list spec validates attribute names but not call signatures, so a renamed keyword or a new required parameter in the SDK passes silently here. This file already uses `autospec=True` for `Islo` itself, and since the module is gated by `pytest.importorskip("islo")` the real `SandboxesClient` is importable wherever these tests run, so `mock.create_autospec(SandboxesClient, instance=True)` is available. Worth using, given the backend mixes positional and keyword forms against this client (`delete_sandbox(sandbox_name=...)` versus `get_sandbox(sandbox)`).

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
```diff
@@ -0,0 +1,687 @@
...
+        # would lose its files over a command that already finished.
+        start = time.monotonic()
+        result = _run_wrapper("sleep 20 & echo started", 1024)
+        elapsed = time.monotonic() - start
+
+        assert result.stdout == "started\n"
+        assert result.returncode == 0
+        assert elapsed < 5.0
+
+    def test_a_long_lived_daemon_does_not_hold_the_command_open(self):
```
kaxil, line 617: This starts a real `sleep 300` and, by design, the wrapper no longer waits for it, so the process outlives the test with nothing to reap it. Fine on an ephemeral runner, a stray every run on a dev machine. Could the daemon be something self-limiting like `nohup sleep 300 &` replaced with a short sleep, or killed by pid in a fixture?

providers/common/ai/pyproject.toml
```diff
@@ -98,6 +98,7 @@ dependencies = [
     "apache-airflow-providers-git>=0.4.0",
     "pydantic-ai-skills>=1.2.0",
 ]
+"sandbox-islo" = ["islo>=0.3.19"]
```
kaxil, line 101: `islo` is not in `[dependency-groups].dev`, so `importorskip` skips all 65 tests in the main unit-test lane and in a plain local `breeze run pytest`; they run only in "Low dep tests: providers", which is the lane doing `uv sync --all-extras`. `langchain`, `llama-index-core`, `pydantic-ai-skills` and `sqlglot` are all in `dev` for exactly this reason. One line here makes the suite run where people look at it.

providers/common/ai/src/airflow/providers/common/ai/sandbox/__init__.py
```diff
@@ -37,3 +41,15 @@
     "SandboxTerminalError",
     "SbxSandboxBackend",
 ]
+
+
+def __getattr__(name: str):
```
kaxil, line 46: This is the path most users hit first, since the extra is opt-in, and it is untested. Six sibling test files in this provider cover their own `AirflowOptionalProviderFeatureException` branches, so there is a pattern to follow.

kaxil: @zozo123  ^^^ this is pending

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
kaxil: ^^^ @zozo123 This too

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: You were right, and the live check I had run was too weak to see it. Measured on the server default image (`ghcr.io/islo-labs/islo-runner:latest`, Debian 12): its `/etc/profile` exports only `PATH`, and `LANG` and `HOME` set at creation survive both `sh -c` and `sh -lc`. `PATH` itself is owned by the runner -- a `PATH` given at creation is gone before the process starts under either shell, and so is an exec-level `env` override of it.

So the wrapper now runs `sh -c`, and a `PATH` in `SandboxSpec.env` is refused at `create` rather than dropped, which is what `base.py` asks for (`test_refuses_a_path_the_runner_would_drop`). The comment on the create-time `env` and the docs now state what is actually guaranteed, and `TestCommandWrapper` runs the real wrapper with a `PATH` override and checks the first component survives -- that fails under `-lc` on both Debian's profile and macOS's `path_helper`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Right on both, and passing `max_retries: 0` was worse than omitting it: the SDK retries transport errors and 5xx twice by default (`islo/core/http_client.py:274`), and `0` switched that off. `_request_options` now leaves `max_retries` to the SDK unless asked for, and `destroy` still asks for 2.

In `_await_exec` a transient failure -- `ApiError` with a 5xx, 429 or no status, or an `httpx.TransportError` -- keeps polling until the deadline; a 4xx or any other exception still translates. If the deadline passes and the last poll had failed, that error is raised, since a timeout the backend never observed would be a guess. The poll's HTTP timeout has a 5s floor rather than the 1s you spotted.

This bit while I was measuring, which is the best argument for it: `get_exec_result` answered `503 RESOURCE_UNAVAILABLE` twice in one session. Tests: `test_transient_poll_errors_are_ridden_out`, `test_a_non_transient_poll_error_is_terminal`, `test_transient_errors_lasting_past_the_deadline_are_terminal_not_a_timeout`, `test_the_last_poll_keeps_a_minimum_http_budget`, `test_requests_leave_the_sdk_retries_in_place`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/docs/toolsets.rst
zozo123: Missed when the cleanup path changed -- fixed to match the code. An unconfirmed deletion is logged, the result reports `sandbox_terminated=True`, and the lifecycle policy reclaims the VM. (This paragraph now lives in `docs/sandbox/backends.rst`; `toolsets.rst` was split up on `main` in the meantime.)

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/docs/toolsets.rst
zozo123: Kept, and shipped properly rather than later: `IsloHook` in `hooks/islo.py` (conn type `islo`, default `islo_default`, `get_ui_field_behaviour`, and a `test_connection` that lists one sandbox and creates nothing), `connection-types` and `hooks` entries in `provider.yaml` with `get_provider_info.py` regenerated, `docs/connections/islo.rst` with its toctree entry, `docs/hooks/islo.rst`, and `tests/unit/common/ai/hooks/test_islo.py`.

The backend resolves its client through the hook; `islo_conn_id=None` still hands resolution to the SDK's `ISLO_API_KEY` / `ISLO_BASE_URL` / `ISLO_COMPUTE_URL`, and the docs name all three. The `extra.timeout` knob has no env-var equivalent, which was the other reason to keep the connection rather than drop it. I exercised the connection path live -- hook `test_connection`, create, command, write, read, list, destroy.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Done, and you are right that it was a regression from going native. On overflow the backend runs `stat -Lc %s` in the guest and reports `max(size, bytes read)`; if `stat` cannot answer, it raises a recoverable `SandboxError` that says only that the file is larger than the limit and how to read part of it, rather than reporting the budget back as the size. Tests: `test_oversized_read_stops_closes_the_stream_and_reports_the_real_size`, `test_oversized_read_without_a_size_says_so_rather_than_inventing_one`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Done. `_raise_file_not_found` now reads the `SandboxResponse`: `deleted_at` set, a known-unusable status, or `paused` without `auto_resume=on_activity` is a `SandboxTerminalError`; otherwise the missing-file `SandboxError` stands.

Measured: a stopped sandbox answers 200 with `status: stopped` and refuses exec with `400 ... has no running VM (status: stopped)`; a deleted one is a 404. The unusable set is `stopping`/`stopped`/`deleting`/`deleted`/`error`/`failed`; an unknown status reads as usable, because the sandbox was reachable a moment earlier -- the opposite bias from the exec-status set, and the comment says why. `create` runs the same check on the response it gets and destroys a sandbox that fails it. Tests: `test_missing_file_on_a_sandbox_that_cannot_serve_is_terminal`, `test_a_paused_sandbox_is_usable_only_when_it_resumes_on_activity`, `test_a_sandbox_that_cannot_serve_after_creation_is_destroyed_and_terminal`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Your reading is right, and it is documented (docs.islo.dev/concepts/sandbox-lifecycle): `delete_after` counts from creation, `pause_after_idle` from the last exec or file operation, and `auto_resume: on_activity` resumes a paused sandbox on the next one.

The default is now `LifecyclePolicy(pause_after_idle=600, auto_resume="on_activity", delete_after=86400)`: a run that outlives an hour keeps its microVM, a dead worker's VM stops costing compute after ten idle minutes, and storage is reclaimed a day after creation. All three are constructor parameters, `None` disables a timer, and the docs say plainly that `delete_after` counts from creation and is not renewed. The policy round-trips on a live create.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/docs/toolsets.rst
zozo123: Done, and thank you for measuring it. The docs now say the server default image is Debian based, that any Debian or Ubuntu image including `python:*-slim` has the GNU `find`, `stat` and `tail` the file operations use, and that Alpine and other busybox images do not. The server default is `ghcr.io/islo-labs/islo-runner:latest`, Debian 12 with findutils 4.9.0, so it qualifies.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/docs/toolsets.rst
zozo123: It was not documented anywhere -- you could not find it because the claim was wrong. I measured it against the compute API instead: each stream is capped at exactly 1 MiB and the **tail** is kept. A 300,000-line `seq` came back as 1,048,576 bytes ending in `299999\n300000\n`, stderr is capped independently at the same size, and a 3 MB single line behaves the same. One flag covers both streams. The "vendor keeps the head" sentence is gone from the docs and the comment.

That also settles the `list_directory` branch you were worried about. The wrapper never asks the server for more than 1 MiB per stream (a larger `max_output_bytes` is clamped, and `_HELPER_OUTPUT_CAP` sits just under), so server truncation is not reachable by wrapper output; the server's flag is only ever attributed to a stream actually at the cap, never to a short one; and because both the wrapper and the server keep the tail, a truncated listing's mangled record is always the leading one, which is the one dropped. Tests: `test_the_server_flag_marks_only_a_stream_at_the_server_cap`, `test_the_server_flag_alone_does_not_mark_streams_that_fit`, `test_budget_above_the_server_cap_is_clamped_to_it`, `test_a_listing_cut_at_the_head_drops_only_the_leading_record`, `test_the_server_flag_alone_leaves_a_short_listing_intact`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/islo.py
zozo123: Fixed -- the server's single flag is now attributed only to a stream that is actually at the server cap, so a trimmed stdout no longer marks an empty stderr as truncated. Test: `test_the_server_flag_marks_only_a_stream_at_the_server_cap`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/docs/toolsets.rst
zozo123: Taken a different way, which removes the problem rather than working around it. `islo.py` never imported the SDK at import time -- only inside methods and under `TYPE_CHECKING` -- so the lazy `__getattr__` never had a case to handle. The package now imports `IsloSandboxBackend` eagerly, exactly like `SbxSandboxBackend`, so the package path resolves for autoapi the same way theirs does, and the untested optional-feature branch is gone rather than covered. A missing SDK surfaces as an actionable `SandboxTerminalError` on first use: `test_missing_sdk_error_is_actionable`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
zozo123: You are right, and the three-variant check is convincing -- `status=$?` captures the foreground command's own signal death before the trap runs, so falling through only adds stderr noise to an already-nonzero exit.

The test now waits for the wrapper's `sleep` child to exist, so the traps are installed even on a loaded runner, before `killpg`, then asserts exit status 143, an empty stdout, and no "No such file" on stderr. Without the trap the wrapper dies of the signal itself and Python reports a negative return code; a trap that cleans up and falls through reaches `tail` with the scratch files gone and says exactly that. bash's own "Terminated: 15" notice for the child is allowed, since it is the shell's and not the wrapper's -- I checked what the shipped wrapper emits under load, and 12/12 runs gave `(143, '', 'Terminated: 15')`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
zozo123: Done: `mock.create_autospec(SandboxesClient, instance=True)`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/tests/unit/common/ai/sandbox/test_islo.py
zozo123: Fair -- a stray per run on a dev machine is not acceptable. The daemon now prints its pid and the test kills it in `finally`.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/pyproject.toml
zozo123: Done: `islo>=0.3.19` is in `[dependency-groups].dev`, so the suite runs in the main lane rather than being skipped. It is 100 items now with the hook tests.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/sandbox/__init__.py
zozo123: Gone rather than tested -- see the reply on the `:class:` target thread. With the eager export there is no optional-feature path left to cover, and the first-use error has a test of its own.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

zozo123:
Pushed the round for the 2026-09-17 review; replies are inline on each thread.

Two of the threads needed facts neither the SDK nor the API spec carries, so I measured them against the compute API rather than guess:

- Each output stream is capped at exactly 1 MiB and the **tail** is kept, not the head, and one `truncated` flag covers both streams. A 300,000-line `seq` came back as 1,048,576 bytes ending in `299999/300000`; stderr caps independently; a 3 MB single line behaves the same. The "vendor keeps the head" claim in the docs was wrong and is gone, which also settles the `list_directory` branch -- a truncated listing's mangled record is always the leading one, which is the one dropped.
- `delete_after` counts from creation, as you read it; `pause_after_idle` counts from last use and `auto_resume: on_activity` resumes on the next call. The default policy is now `pause_after_idle=600, auto_resume="on_activity", delete_after=86400`, so a run past the hour keeps its microVM while a dead worker's VM stops costing compute after ten idle minutes.

Also measured, and each drove a change: the runner owns `PATH` and drops one given at creation under both `sh -c` and `sh -lc`, so the wrapper is `sh -c` and a `PATH` in `SandboxSpec.env` is refused at `create` rather than silently dropped; and a stopped sandbox answers `get_sandbox` with 200 and `status: stopped` while exec returns 400, which is what the missing-file check now reads before deciding between recoverable and terminal.

The rest: polling rides out transient 5xx rather than failing the task on one (the API produced `503 RESOURCE_UNAVAILABLE` twice during the session, so this is not hypothetical), and `max_retries: 0` is gone since it was switching off the SDK's own retries; oversized reads report the guest's `stat` size; the connection has its own `islo` type with a hook, a docs page and `test_connection`; the lazy export is replaced by an eager one so the autoapi reference resolves and the untested optional branch disappears; `islo` is in the dev group so the suite runs in the main lane; and the trap test now discriminates between the shipped wrapper and the two ways it could regress.

The system test passed end to end against islo.dev (`1 passed in 12.08s`), as did the connection-backed path -- hook `test_connection`, create, command, write, read, list, destroy.

Rebased onto current `main`, which reorganised the provider docs twice today, so the backend is now documented in `docs/sandbox/backends.rst` beside Modal and `sbx`, and the connection and hook pages are in the `docs/index.rst` toctree rather than the removed `docs/hooks/index.rst`.

100 unit tests, ruff and the applicable prek hooks pass locally; mypy is clean on the changed files with the SDK installed.

---
Drafted-by: Claude Code (Opus 5); reviewed by @zozo123 before posting

providers/common/ai/src/airflow/providers/common/ai/hooks/islo.py
```diff
@@ -0,0 +1,105 @@
...
+import math
+from typing import TYPE_CHECKING, Any
+
+from airflow.providers.common.compat.sdk import BaseHook
+
+if TYPE_CHECKING:
+    from islo import Islo
+
+
+class IsloHook(BaseHook):
```
kaxil, line 30: Hmm unsure how I missed this. Islo hooks shouldn't be present on root of `common/ai/hooks` . Ideally soon we will have an Islo provider like the proposed Modal provider..

Until then all the Islo stuff should be restricted to sandbox dir. wdyt?

PR #71675, opened 2026-08-16
Warn when ignore_unknown_values is a no-op in GCSToBigQueryOperator

jonathangosling:
`ignore_unknown_values=True` silently does nothing under the operator's default `autodetect=True`. The load job's schema is inferred from the same source data the values are checked against, so no source field can be unknown to it. Users set the flag expecting fields absent from the destination table to be dropped, and instead get a schema mismatch with nothing in the logs indicating the flag was ignored.

This warns when the assembled load configuration has `autodetect` truthy, `ignoreUnknownValues` set, and no `schema`, pointing at `autodetect=None` - the setting that makes BigQuery use the destination table's own schema.

The check reads the assembled `configuration["load"]` rather than the operator attributes, because `src_fmt_configs` and `extra_config` can override any of those three keys after they are set (all three are in `valid_configs` for the relevant source formats). Two of the added test cases cover exactly that.

It also fixes the `autodetect` docstring, which actively pointed the wrong way: it said the parameter "must be set to True if 'schema_fields' and 'schema_object' are undefined" (`None` is also valid, and is the documented way to load into an existing table), and that it is "suggested to set to True if table are create outside of Airflow" (for a table created outside Airflow, `True` is precisely what triggers the mismatch - `None` is what works). All three states are now documented, and `autodetect` gets a `bool | None` annotation matching the existing `is False` check at the schema-object guard.

Deliberately out of scope: changing the default, making `autodetect=False` fall back to the table schema, and redesigning the flag. The three-state parameter is an artifact of `"autodetect": self.autodetect` being serialized unconditionally, so `None` is the only way to express "omit the field" - worth revisiting separately, but not in a bugfix PR.

 

---

---

### DIFF
--- a/providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
+++ b/providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
@@ -164,21 +164,10 @@ class GCSToBigQueryOperator(BaseOperator):
         by one or more columns. BigQuery supports clustering for both partitioned and
         non-partitioned tables. The order of columns given determines the sort order.
         Not applicable for external tables.
-    :param autodetect: [Optional] Indicates whether the options and schema should be inferred
-        from the source data for CSV and JSON sources. (Default: ``True``). This parameter is
-        three-state:
-
-        * ``True`` - infer the schema from the source files. The inferred schema is sent to
-          BigQuery as the load job's schema, so ``ignore_unknown_values`` has no effect: no field
-          in the source data can be unknown to a schema derived from that same data. If the
-          destination table already exists and the inferred schema differs from it, the load
-          fails with a schema mismatch unless ``schema_update_options`` permits the change.
-        * ``False`` - do not infer the schema. One of ``schema_fields`` or ``schema_object``
-          must be provided, otherwise an exception is raised.
-        * ``None`` - neither infer a schema nor supply one, so the destination table's own
-          schema is used. The table must already exist. This is the setting to use when you want
-          ``ignore_unknown_values`` to drop fields that are present in the source data but absent
-          from the destination table.
+    :param autodetect: [Optional] Whether to infer the schema from the source data for CSV and
+        JSON sources. If ``True``, the schema is inferred from the source data. If ``False``,
+        either ``schema_fields`` or ``schema_object`` must be provided. If ``None``, no schema is
+        supplied and the existing destination table's schema is used. (Default: ``True``).
     :param encryption_configuration: [Optional] Custom encryption configuration (e.g., Cloud KMS keys).
 
         .. code-block:: python
@@ -649,23 +638,6 @@ class GCSToBigQueryOperator(BaseOperator):
         self.log.info("External table created successfully: %s", self.destination_project_dataset_table)
         return table_obj_api_repr
 
-    def _warn_if_ignore_unknown_values_is_ineffective(self) -> None:
-        # Checked against the assembled load config rather than the operator attributes, because
-        # src_fmt_configs and extra_config can still override any of these three keys.
-        load_config = self.configuration["load"]
-        if (
-            load_config.get("autodetect")
-            and load_config.get("ignoreUnknownValues")
-            and "schema" not in load_config
-        ):
-            self.log.warning(
-                "`ignore_unknown_values` is set but will have no effect. With `autodetect` enabled "
-                "and no schema supplied, the load job's schema is inferred from the source data "
-                "itself, so no source field can be unknown to it. To drop source fields that are "
-                "absent from an existing destination table, set `autodetect=None` so that the "
-                "destination table's own schema is used instead."
-            )
-
     def _use_existing_table(self):
         destination_project_id, destination_dataset, destination_table = self.hook.split_tablename(
             table_input=self.destination_project_dataset_table,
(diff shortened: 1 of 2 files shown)


providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
```diff
@@ -162,12 +164,21 @@ class GCSToBigQueryOperator(BaseOperator):
...
+          BigQuery as the load job's schema, so ``ignore_unknown_values`` has no effect: no field
+          in the source data can be unknown to a schema derived from that same data. If the
+          destination table already exists and the inferred schema differs from it, the load
+          fails with a schema mismatch unless ``schema_update_options`` permits the change.
+        * ``False`` - do not infer the schema. One of ``schema_fields`` or ``schema_object``
+          must be provided, otherwise an exception is raised.
+        * ``None`` - neither infer a schema nor supply one, so the destination table's own
+          schema is used. The table must already exist. This is the setting to use when you want
+          ``ignore_unknown_values`` to drop fields that are present in the source data but absent
+          from the destination table.
```
SameerMesiah97: This is too long. There is no need to go deep into the interaction between this and the other parameters here. I would suggest the below instead:

`:param autodetect: [Optional] Whether to infer the schema from the source data for CSV and JSON sources. If ``True``, the schema is inferred from the source data. If ``False``, either ``schema_fields`` or ``schema_object`` must be provided. If ``None``, no schema is supplied and the existing destination table's schema is used. (Default: ``True``).`

providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
```diff
@@ -638,6 +649,23 @@ def _create_external_table(self):
...
+            and load_config.get("ignoreUnknownValues")
+            and "schema" not in load_config
+        ):
+            self.log.warning(
+                "`ignore_unknown_values` is set but will have no effect. With `autodetect` enabled "
+                "and no schema supplied, the load job's schema is inferred from the source data "
+                "itself, so no source field can be unknown to it. To drop source fields that are "
+                "absent from an existing destination table, set `autodetect=None` so that the "
+                "destination table's own schema is used instead."
+            )
```
SameerMesiah97: 1) This helper is not needed as it is only called once. I would inline it. 
2) The warning is too verbose. I would suggest the below:

```
self.log.warning(
    "`ignore_unknown_values` has no effect when `autodetect=True` and no schema is provided. "
    "Set `autodetect=None` to use the existing destination table's schema instead."
)
```

providers/google/tests/unit/google/cloud/transfers/test_gcs_to_bigquery.py
```diff
@@ -2124,6 +2131,110 @@ def test_schema_fields_is_templated(self):
...
+                False,
+                id="schema_supplied_by_extra_config",
+            ),
+            pytest.param(
+                {"ignore_unknown_values": False, "extra_config": {"ignoreUnknownValues": True}},
+                True,
+                id="ignore_unknown_values_set_by_extra_config",
+            ),
+        ],
+    )
```
SameerMesiah97, line 2160: Since autodetect is explicitly a three-state parameter, could we also cover `autodetect=False `here?

providers/google/tests/unit/google/cloud/transfers/test_gcs_to_bigquery.py
```diff
@@ -2124,6 +2131,110 @@ def test_schema_fields_is_templated(self):
...
+            project_id=JOB_PROJECT_ID,
+            **operator_kwargs,
+        )
+
+        with mock.patch.object(operator.log, "warning") as mock_warning:
+            operator.execute(context=MagicMock())
+
+        assert mock_warning.call_args_list.count(call(IGNORE_UNKNOWN_VALUES_WARNING)) == (
+            1 if expects_warning else 0
+        )
```
SameerMesiah97, line 2205: Do these parametrized cases need to assert the complete warning text? The behaviour under test seems to be whether the warning is emitted. Using `assert_called_once() / assert_not_called()` would make these cases less coupled to the wording, with the message itself asserted in a single test if we want to protect it.

SameerMesiah97 (commented):
I have left some comments.

providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
jonathangosling: Good point, I've updated the docstring with your suggestion.

providers/google/src/airflow/providers/google/cloud/transfers/gcs_to_bigquery.py
jonathangosling: Updated: 
- Inlined the check and warning. 
- Made the message less verbose. I've used your suggestion, which covers the problem and the suggested fix more concisely.

providers/google/tests/unit/google/cloud/transfers/test_gcs_to_bigquery.py
jonathangosling: I've added a test case for `autodetect=False`, supplying the `schema_fields` as required.

providers/google/tests/unit/google/cloud/transfers/test_gcs_to_bigquery.py
jonathangosling: Yes, good suggestion. I've updated to use `assert_called_once()` / `assert_not_called()`. On protecting the message, rather than pinning the full text, one test (`test_ignore_unknown_values_no_op_warning_names_the_fix`) now asserts only that the warning names `autodetect=None`, since that's the actionable part.
One thing worth flagging - these assertions don't discriminate which warning fired, so they'll break if an unrelated warning is added. Filtering on/asserting against a stable part of the message (i.e.  `ignore_unknown_values` or `autodetect=None`), rather than the full message, could avoid that. Let me know if you think it's worth it.

